PR #2, opened 2019-05-23
Add SCP code, D binding, update scripts and build system

Geod24:
```
This commit add the following components:
- SCP code from `stellar-core/src/scp`
- Multiple dependencies from the code (includes)
- A `update.d` script to simplify future changes (when a PR affecting the code is merged)
- A `build.d` script to build the C++ code as a bunch of object files
- D bindings for STL types and said SCP code
```

I'm not 100% sure the bindings are correct... But we'll find out soon enough.


Geod24:
Erg, draft PRs don't trigger CI...

dub.json
```diff
@@ -10,8 +10,35 @@
 
     "targetType": "executable",
     "targetPath": "build",
+    "excludedSourceFiles": [ "source/scpp/*.d" ],
+    "preBuildCommands": [ "source/scpp/build.d" ],
+    "sourceFiles": [
```
Geod24, line 15: This should be automated, and it's ridiculous that `dub` does not allow globbing here.
I used `ls -1 build/ | xargs -I '{}' echo '"source/scpp/build/{}"',` to generate this list.

source/scpd/Cpp.d
```diff
@@ -0,0 +1,186 @@
+/*******************************************************************************
+
+    Types currently missing from `core.stdcpp` and some additions
```
Geod24, line 3: Long term goal: nuke this module

source/scpd/scp/BallotProtocol.d
```diff
@@ -0,0 +1,273 @@
...
+
+    // bool federatedAccept(StatementPredicate voted, StatementPredicate accepted);
+    // bool federatedRatify(StatementPredicate voted);
+
+    void startBallotProtocolTimer();
+    void stopBallotProtocolTimer();
+    void checkHeardFromQuorum();
+}
+
+static assert(BallotProtocol.sizeof == 136);
```
Geod24, line 290: Long term goal: Make this check automatic

source/scpd/types/Stellar_types.d
```diff
@@ -0,0 +1,71 @@
...
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module scpd.types.Stellar_types;
```
Geod24, line 15: In modules like this, we should generate `static assert` which check the size of the types...

AndrejMitrovic:
> +19,615

LGTM.

Geod24:
Test Github integration

Geod24:
This includes the following changes to Stellar:
- https://github.com/stellar/stellar-core/pull/2111
- https://github.com/Geod24/stellar-core/commit/39e38abeb1737a7c913cb93dc323a1c66f4ccd48
- https://github.com/Geod24/constrictor/commit/123d8189372ef6da66f31b8bbe6e71349aeb3ad9

Will PR the two commits to stellar-core once I'm sure it doesn't break anything.

Geod24:
Submitted https://github.com/stellar/stellar-core/pull/2113 as it seems necessary

Geod24:
Also https://github.com/stellar/stellar-core/pull/2114

AndrejMitrovic:
> Submitted stellar/stellar-core#2113 as it seems necessary

How about a `status-dep-blocked` to mark PRs which are blocked by an upstream dependency? We used this label in internal projects at some previous company.

source/scpd/scp/BallotProtocol.d
```diff
@@ -0,0 +1,273 @@
...
+    int mCurrentMessageLevel; // number of messages triggered in one run
+
+    /// last envelope generated by this node
+    shared_ptr!SCPEnvelope mLastEnvelope;
+
+    /// last envelope emitted by this node
+    shared_ptr!SCPEnvelope mLastEnvelopeEmit;
+
+  public:
+    //this(Slot& slot);
```
AndrejMitrovic: Can you add a comment around all the dead code in this PR? Or just remove the code if it will never be needed (if it's needed in the future we should make an issue to revisit this later).

source/scpd/scp/BallotProtocol.d
AndrejMitrovic: Do you know about clicking the "..." on the right and clicking on "Reference in new issue"? I'm about to do that.

Geod24:
So LDC uses `gold` instead of `bfd` by default, and that's why it fails: https://github.com/ldc-developers/ldc/blob/5e5ce199f493091e6aea39f7f19f609dcf683da2/driver/linker-gcc.cpp#L558-L561

Geod24:
~This is good to go~

EDIT: No I need to address Drey's comments

Geod24:
@AndrejMitrovic : Did a clean up on commented-out code. Would like this merged sooner than later. I'm 100% sure we'll find bugs in the bindings as we start to use it, but we'll do that progressively.

AndrejMitrovic (approved):

PR #7, opened 2019-05-28
Travis: Remove duplicated source addition

Geod24:
Let's see if that works


AndrejMitrovic:
> Let's see if that works

Btw, it doesn't. :P

PR #9, opened 2019-05-30
Implement size & layout checks for the glue layer

### LINKED ISSUE #4: Add automatic size checks between C++ and D types (structs, etc)
Long term goal: Make this check automatic

_Originally posted by @Geod24 in https://github.com/bpfkorea/agora/pull/2#discussion_r286777408_

AndrejMitrovic:
As noted in the commit comment, I've tried to use friend declarations, but it's problematic due to circular #include's which end up breaking compilation in subtle ways (if only C++ had modules..).

Fixes #4

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -37,6 +37,16 @@
     ],
     "lflags": [ "-lsodium", "-lstdc++" ],
 
+    "configurations": [
+        {
+            "name": "unittest",
+            "sourceFiles": [
+                "source/scpp/build/DSizeChecks.o",
+                "source/scpp/build/DLayoutChecks.o"
+            ]
+        }
+    ],
+
     "dependencies": {
         "bitblob": "~>1.1.1",
         "dyaml": "~>0.7.1",
--- a/source/scpd/tests/GlueTypes.d
+++ b/source/scpd/tests/GlueTypes.d
@@ -1,0 +1,67 @@
+/*******************************************************************************
+
+    Contains types used for size & ABI object layout checks.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module scpd.tests.GlueTypes;
+
+import core.stdc.stdint;
+import std.meta;
+
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types;
+import scpd.types.Utils;
+import scpd.types.XDRBase;
+
+// todo: replace with __allMembers tricks in the future
+alias TypesWithLayout = AliasSeq!
+(
+    /// scpd.types.Stellar_SCP
+    SCPBallot,
+    SCPNomination,
+    SCPStatement._pledges_t._prepare_t,
+    SCPStatement._pledges_t._confirm_t,
+    SCPStatement._pledges_t._externalize_t,
+    SCPStatement._pledges_t,
+    SCPStatement,
+    SCPEnvelope,
+    SCPQuorumSet,
+
+    /// scpd.types.Stellar_types
+    PublicKey,
+    Hash,
+
+    /// scpd.types.Utils
+    ByteSlice,
+);
+
+/// For these types we cannot do layout checks as we require
+/// access to private std::array fields to get the offset.
+/// However we can still do sizeof() checks.
+alias TypesNoLayout = AliasSeq!
+(
+    /// scpd.types.Stellar_SCP
+    Value,
+    SCPStatementType,
+    CryptoKeyType,
+    PublicKeyType,
+    SignerKeyType,
+    Signature,
+    SignatureHint,
+
+    /// scpd.types.XDRBase
+    xvector!Value,
+    xvector!PublicKey,
+    xvector!SCPQuorumSet,
+);
+
+/// All the glue types have sizeof(), but some may not have accessible layout information
+alias GlueTypes = AliasSeq!(TypesWithLayout, TypesNoLayout);
--- a/source/scpd/tests/LayoutTest.d
+++ b/source/scpd/tests/LayoutTest.d
@@ -1,0 +1,70 @@
+/*******************************************************************************
+
+    Contains runtime field size & ABI object layout checks.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module scpd.tests.LayoutTest;
+
+import scpd.tests.GlueTypes;
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types;
+
+import std.digest.sha;
+import std.random;
+import std.stdio;
+import std.string;
+import std.traits;
+
+/// Contains the size and offset of a field within a struct
+extern(C++) struct FieldInfo
+{
+    long size;
+    long offset;
+    //const(char)* mangleof;  // todo: could this work?
+}
+
+static foreach (Type; GlueTypes)
+{
+    extern(C++) FieldInfo cppFieldInfo (ref Type, const(char)*);
+}
+
+/// Special-case as we implement our own Hash type that is binary-compatible with SCP's Hash
+extern(C++) bool isHashTypeCompatible (ref Hash, const(char*));
+
+/// size & layout checks for C++ structs / objects
+unittest
+{
+    foreach (Type; TypesWithLayout)
+    {
+        foreach (idx, field; Type.init.tupleof)
+        {
+            auto object = Type.init;
+            auto field_info = cppFieldInfo(object,
+                Type.tupleof[idx].stringof.toStringz);
+
+            assert(typeof(field).sizeof == field_info.size,
+                format("Field '%s' of '%s' size mismatch: %s (D) != %s (C++)",
+                    Type.tupleof[idx].stringof, Type.stringof,
+                    typeof(field).sizeof, field_info.size));
+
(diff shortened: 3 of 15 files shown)


source/scpd/tests/GlueTypes.d
```diff
@@ -0,0 +1,74 @@
...
+    SCPStatement._pledges_t._prepare_t,
+    SCPStatement._pledges_t._confirm_t,
+    SCPStatement._pledges_t._externalize_t,
+    SCPStatement._pledges_t,
+    SCPStatement,
+    SCPEnvelope,
+    SCPQuorumSet,
+
+    /// scpd.types.Stellar_types
+    //Hash,  // todo (linker error)
```
Geod24: `Hash` is aliased to `BitBlob!256`
But I would double check if they are binary compatible (esp. w.r.t. endianess)

source/scpd/tests/GlueTypes.d
```diff
@@ -0,0 +1,74 @@
...
+    //PublicKey,
+    //SignatureHint,  // todo: opaque_array!4 is not instantiated on the C++ side,
+                      // could be a type or mangling mismatch
```
Geod24: https://github.com/stellar/stellar-core/search?q=SignatureHint&unscoped_q=SignatureHint

Geod24:
Random idea: We could add `offsetoff` tests (or better... GENERATE THEM!)

source/scpp/extra/DSizeChecks.cpp
```diff
@@ -0,0 +1,60 @@
...
+#include "xdrpp/types.h"
+#include "xdr/Stellar-SCP.h"
+#include "xdr/Stellar-types.h"
+#include "scp/Slot.h"
+#include "crypto/ByteSlice.h"
+
+using namespace xdr;
+using namespace stellar;
+
+/// note: tried to use template but linking to it is a nightmare
```
Geod24: Care to elaborate ?

AndrejMitrovic:
> Random idea: We could add offsetoff tests (or better... GENERATE THEM!)

What do you mean? It already contains `.offsetof` tests.

Geod24:
Right, I was looking at the `static assert` and thinking in addition to size it could also be `offsetof`. But that actually does not test anything... Disregard :D

AndrejMitrovic:
Btw, this is missing checks for passing structs around and testing functions. I will add those tests as well. First let me fix the failure on linux.

Geod24:
Since you are testing offset and size, I'd say pass the struct by ref. Because if the size's wrong you might crash before reaching your test function.

source/scpd/tests/GlueTypes.d
AndrejMitrovic: Ah I could just pass it as a void* then. Yeah I'll add those checks as well.

Geod24: Or you could pass it as a `ref Hash` on D side and `Hash&` on the C++ side?

source/scpp/extra/DSizeChecks.cpp
AndrejMitrovic: C++ side:

```C++
template <typename T>
unsigned long long cppSizeOf ( T &value ) { return sizeof(value); }

// instantiate
template unsigned long long cppSizeOf<Value>(Value &value);
```

D side:

```D
extern(C++) ulong cppSizeOf (T)(ref T);
alias cppSizeOf!Value cppSizeOf_;
```

Error:

```
  "unsigned long cppSizeOf<xdr::xvector<unsigned char, 4294967292u> >(xdr::xvector<unsigned char, 4294967292u>&)", referenced from:
      __D4scpd5tests8SizeTest17__unittest_L31_C1FZv in agora-test-application.o
```

I'm not sure how well D links with C++ templates. Maybe I was missing something?

AndrejMitrovic:
Pushed linux fix and added size & layout checks for the `Hash` type.

Edit: also rebased on v0.x.x

dub.json
```diff
@@ -15,6 +15,8 @@
     "sourceFiles": [
         "source/scpp/build/BallotProtocol.o",
         "source/scpp/build/ByteSliceHasher.o",
+        "source/scpp/build/DSizeChecks.o",
+        "source/scpp/build/DLayoutChecks.o",
```
Geod24: Can we move this in tests only ?
See https://dub.pm/package-format-json.html#configurations

source/scpd/tests/GlueTypes.d
```diff
@@ -0,0 +1,68 @@
+
```
Geod24: Extra empty line

source/scpd/tests/GlueTypes.d
```diff
@@ -0,0 +1,68 @@
...
+/*******************************************************************************
+
+    Contains runtime size & ABI object layout checks and
+    function argument passing in the D glue layer works properly.
```
Geod24: "that the D glue layer..." ?

source/scpd/tests/LayoutTest.d
```diff
@@ -0,0 +1,80 @@
...
+import std.random;
+import std.stdio;
+import std.string;
+import std.traits;
+
+/// Contains the size and offset of a field within a struct
+extern(C++) struct FieldInfo
+{
+    long size;
+    long offset;
```
Geod24, line 30: So... Is `long` mangled correctly ? 🤔

source/scpd/tests/LayoutTest.d
```diff
@@ -0,0 +1,80 @@
...
+
+static foreach (Type; GlueTypes)
+{
+    extern(C++) FieldInfo cppFieldInfo (ref Type, const(char)*);
+}
+
+/// Special-case as we implement our own Hash type that is binary-compatible with SCP's Hash
+/// Note: cannot pass with proper type due to mangling issues:
+///  "isHashTypeCompatible(BitBlob<256ul>&)", referenced from:
+///      __D4scpd5tests10LayoutTest17__unittest_L43_C1FZv in agora-test-application.o
```
Geod24: One question I have is: "how does mangling behave w.r.t `alias`. Aliases are transparent, so we should probably use `struct Name { RealType data; alias data this; }` however it has its problems (namely there's a Name -> RealType path but not the other way around).

Geod24: Yeah I think we might need to do this instead of aliases in the binding. Care to try ?

source/scpd/tests/LayoutTest.d
AndrejMitrovic: Yeah long seems fine. But not `unsigned lang lang`.

source/scpd/tests/LayoutTest.d
AndrejMitrovic: Hmm, are there any benefits to this? Because implicit conversions will still happen. Genuinely curious as I rarely use subtyping.

Geod24: Well if you declare a `void fun(ref Hash)` function on the D side and a `void fun(Hash&)` on the C++ side and it doesn't link I'd say it's a pretty annoying bug.

AndrejMitrovic: ok so you want me to try declaring Hash as:

```D
struct Hash
{
    BitBlob!(...) _internal;
    alias _internal this;
}
```

And see how far we can take this?

dub.json
AndrejMitrovic: Linking to the dub docs is cruel.

AndrejMitrovic:
Update: added a commit changing the `Hash` definition so it's mangle-compatible with C++. This means it can now be passed around to C++ functions.

I have yet to fix https://github.com/bpfkorea/agora/pull/9/files#r289269364 because dub documentation is terrible.

AndrejMitrovic:
Moved the layout & size C++ test support file sources to the `unittest` target, so they don't get unconditionally compiled & linked to.

source/scpd/tests/GlueTypes.d
Geod24: Ping

source/scpd/tests/GlueTypes.d
Geod24: This just declares the types right ?

source/scpd/tests/SizeTest.d
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
+
+    Contains runtime size & ABI object layout checks and
```
Geod24: Just size tho

source/scpd/tests/LayoutTest.d
```diff
@@ -0,0 +1,71 @@
+/*******************************************************************************
+
+    Contains runtime size & ABI object layout checks and
+    function argument passing in the D glue layer works properly.
```
Geod24: Just layout tho

AndrejMitrovic:
Fixed docs and extra blank line (forgot to merge from another branch).

source/scpd/types/XDRBase.d
```diff
@@ -88,34 +91,14 @@ class xdr_wrong_union : logic_error
 
 extern(D) static immutable XDR_MAX_LEN = 0xffff_fffc;
 
-//struct xarray(T, uint32_t N) { public array!(T, N) data; alias data this; }
-struct xarray(T, uint32_t N)
+///
+struct opaque_array(uint32_t N = XDR_MAX_LEN)
```
Geod24, line 95: It's only used for `Hash` right ?

AndrejMitrovic: Also `uint256`, `SignatureHint` and some other types I guess. What are you thinking?

Geod24 (approved):
This is grrrrrreat

*auto-merge toggled on*

PR #11, opened 2019-06-13
Implement basic skeleton & networking

AndrejMitrovic:
I've stripped out everything from the draft project by @Geod24 and left in the networking. It's basic stuff, but it works, and we can get started implementing other things like interacting with SCP, etc.

I can demonstrate how multiple nodes connect with each other in person.

Edit: I wish to make everything unit-testable, perhaps with the framework @Geod24 previously implemented.

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -1,0 +1,94 @@
+################################################################################
+## This is a sample, documented configuration file for Agora in YAML format
+##
+## It is not intended for usage on any network
+##
+## A node has two interfaces: one for network communication,
+## and one for administration.
+## The one communicating with the network is public,
+## while the administrative is unsecured, disabled by default,
+## and should not be exposed to the outside world.
+## The administrative one is unsecured and should not be exposed to the outside
+## world. It is disabled by default for security purpose.
+################################################################################
+
+################################################################################
+##                             Network interface                              ##
+################################################################################
+node:
+  # The server can operate in two modes: full node and validator node.
+  # The full node does not participate in consensus,
+  # it only replicates the state of the blockchain.
+  # When validating, the `seed` of an eligible account is required
+  # An eligible account has at least 4 units frozen in it
+  is_validator: true
+  # Maximum number of non-validating nodes to connect to
+  max_listeners: 10
+  # Address to which we bind
+  address: 0.0.0.0 # Any node can bind - default value
+  # Port on which we bind
+  port:    2826    # 0xB0A, default value
+  # Number of seconds to wait between retrying requests (floating-point)
+  retry_delay: 1.5
+  # This is a randomly generated keypair
+  # If this node is not a validator, this will be ignored
+  #
+  # DO NOT USE THOSE VALUES ANYWHERE
+  # Private seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  # Public address:  GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
+  seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+
+################################################################################
+##                          Administrative interface                          ##
+################################################################################
+admin:
+  enabled: true      # `false` by default
+  address: 127.0.0.1 # Private
+  port:    2827      # 0xB0B
+
+################################################################################
+##                               Node discovery                               ##
+##                                                                            ##
+## When the network first starts, we need to connect to some peers to learn   ##
+## the topology and find a safe intersection to listen to, and, if we are     ##
+## a validator, to insert ourselves.                                          ##
+################################################################################
+network:
+  # Supported value: IPv4, IPv6
+  - http://192.168.1.42:2828
+  - http://192.168.0.44:2828
+
+################################################################################
+##                               Quorum slices                                ##
+################################################################################
+quorum:
+  # name of the quorum
+  - name: quorum-1
+    # threshold as a percentage
+    threshold: 66%
+    # the list of nodes and sub-quorums in this quorum
+    nodes:
+      - GBFDLGQQDDE2CAYVELVPXUXR572ZT5EOTMGJQBPTIHSLPEOEZYQQCEWN
+      - GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5
+      # a subquorum is referenced with a $ symbol
+      - $quorum-1.1
+  - name: quorum-1.1
+    threshold: 66%
+    nodes:
+      - GBFDLGQQDDE2CAYVELVPXUXR572ZT5EOTMGJQBPTIHSLPEOEZYQQCEWN
+      - GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5
+      # references another sub-quorum
+      - $quorum-1.1.1
+  - name: quorum-1.1.1
+    threshold: 66%
+    nodes:
+      - GBFDLGQQDDE2CAYVELVPXUXR572ZT5EOTMGJQBPTIHSLPEOEZYQQCEWN
+      - GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5
+      - GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5
+
+################################################################################
+##                               Logging options                              ##
+################################################################################
+logging:
+  # Values: trace, debugV, debug_, diagnostic, info, warn, error, critical, fatal, none (default)
+  level: none
--- a/source/agora/app.d
+++ b/source/agora/app.d
@@ -1,6 +1,11 @@
 /*******************************************************************************
 
-    Entry point for Agora node
+    Entry point for the Agora node
+
+    Each node is a Vibe.d server application.
+
+    On startup, it tries to connect with a range of known hosts to join the
+    quorum. It then starts to listen for requests, using a REST interface.
 
     Copyright:
         Copyright (c) 2019 BOS Platform Foundation Korea
@@ -13,11 +18,32 @@
 
 module agora.app;
 
-void main()
+import agora.common.Config;
+import agora.node.Node;
+
+import vibe.core.core;
+
+import std.getopt;
+import std.stdio;
+
+/// Application entry point
+private int main (string[] args)
 {
-	// TODO:
-    // - Config parsing
-    // - Network interface setup
-    // - Event loop
-    // - End world hunger
+    CommandLine cmdln;
+    auto help = parseCommandLine(cmdln, args);
+    if (help.helpWanted)
+    {
+        defaultGetoptPrinter("The Agora node", help.options);
+        return 0;
+    }
+
+    auto config = parseConfigFile(cmdln);
+    if (cmdln.config_check)
+    {
+        writefln("Config file '%s' succesfully parsed.", cmdln.config_path);
+        return 0;
+    }
+
+    auto node = new Node(config);
+    return runEventLoop();
 }
--- a/source/agora/app.d
+++ b/source/agora/app.d
@@ -26,6 +26,10 @@ import vibe.core.core;
 import std.getopt;
 import std.stdio;
 
+/// Workaround for issue likely related to dub #225,
+/// expects a main() function and invokes it after unittesting.
+version (unittest) void main () { } else:
+
 /// Application entry point
 private int main (string[] args)
 {
--- a/source/agora/common/API.d
+++ b/source/agora/common/API.d
@@ -1,0 +1,93 @@
+/*******************************************************************************
+
+    Contains the REST API interface.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.API;
+
+import agora.common.crypto.Key;
+import agora.common.Data;
+import agora.common.Set;
+
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types;
+
+import vibe.data.json;
+import vibe.web.rest;
+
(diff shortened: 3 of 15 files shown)


doc/config.example.yaml
```diff
@@ -0,0 +1,82 @@
+################################################################################
+## This is a sample, documented configuration file for <PROJECT-NAME>
```
Geod24: <CURRENT-YEAR>

source/agora/app.d
```diff
@@ -12,12 +20,53 @@
...
+import vibe.web.rest;
 
-void main()
+import std.digest.sha;
+import std.stdio;
+import core.stdc.time;
+import core.time;
+
+/// todo: work around this for 'dub test'
+version (unittest) void main () { } else:
```
Geod24, line 31: Shouldn't be necessary if you specify `mainSourceFile` in `dub.json`

source/agora/common/API.d
```diff
@@ -0,0 +1,127 @@
...
+    Contains the REST API interface.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+
```
Geod24: W H I T E S P A C E

source/agora/common/Hash.d
```diff
@@ -0,0 +1,50 @@
...
+    static if (is(T : E[], E))
+        const bs = ByteSlice.make(arg);
+    else static if (is(typeof(arg.serialize(sink))))
+    {
+        arg.serialize(sink);
+        const bs = ByteSlice.make(buffer);
+    }
+    else
+        const bs = ByteSlice.make(XDRToOpaque(arg));
+    return sha256(bs);
```
Geod24: Is that ?... I mean ? I'm scared.

source/agora/common/Set.d
```diff
@@ -0,0 +1,150 @@
...
+/// simplified Set() code with some convenience methods,
+/// could use a drop-in implementation later.
+struct Set (T)
+{
+    ///
+    bool[][T] _set;
+    alias _set this;
+
+    /// Get and remove one element from the set
+    public T stealOne ( ) ( )
```
Geod24: Why is this a template ?

source/agora/common/Hash.d
AndrejMitrovic: This seems to be unused leftover code from _that other repo of yours_. I used it during testing here and there, I'll remove it.

AndrejMitrovic:
Update:

- Fixed documented config to mention Agora.
- Fixed extra blank line
- Added config verification which should be called both after parsing, and after creating a manual config for unittests
- Made array fields in the config immutable (see next line)
- Added shared aliasing tests to make sure these configs can always be used with the LocalRest library
- Removed unecessary key_pair field which can just be part of the Config file instead now that it's immutable
- Changed retry-delay from integer to floating-point. This allows setting it to a few milliseconds which is useful in order to speed up unittests.
- Fixed uneccessary "() ()" template
- Removed unused Hash module
- Fixed a node from trying to connect to itself during discovery
- Fixed fiber sleep calls to consistently use the configured sleep time
- Moved creation of the initial task directly into the Node class. This allows the node to auto-connect in unittests without having to manually run its node.start() method.

And:
- Added a commit implementing a very basic network discovery test-case. Much more will be added, this was just a proof-of-concept. But the test already uncovered a few bugs that are now fixed.

AndrejMitrovic:
> @AndrejMitrovic AndrejMitrovic force-pushed the AndrejMitrovic:basic-networking  branch from ace94d4 to 04b9a1d 1 minute ago

Unfortunately this shows the wrong diff because I force-pushed the wrong changes at first.

doc/config.example.yaml
```diff
@@ -0,0 +1,81 @@
+################################################################################
+## This is a sample, documented configuration file for Agora in YAML format
+##
+## It is not intended for usage on any network
+##
+## A node has two interface: a working and an administrative one
```
```suggestion
## A node has two interface: one for network communication and one for administration
```

doc/config.example.yaml
```diff
@@ -0,0 +1,81 @@
...
+## The public one is used when communicating with the network.
```
Geod24: And then rephrase this as:
```
The one communicating with the network is public, while the administrative is unsecured, disabled by default, and should not be exposed to the outside world.
```

doc/config.example.yaml
```diff
@@ -0,0 +1,81 @@
...
+## The administrative one is unsecured and should not be exposed to the outside
+## world. It is disabled by default for security purpose.
+################################################################################
+
+################################################################################
+##                             Network interface                              ##
+################################################################################
+node:
+  # The server can operate in two modes: watcher and validator
```
Geod24: `s/watcher/full node/g`

doc/config.example.yaml
```diff
@@ -0,0 +1,81 @@
...
+################################################################################
+##                               Node discovery                               ##
+##                                                                            ##
+## When the network first starts, we need to connect to some peers to learn   ##
+## the topology and find a safe intersection to listen to, and, if we are     ##
+## a validator, to insert ourselves.                                          ##
+################################################################################
+network:
+  # Supported value: domain name, IPv4, IPv6
+  - http://0.0.0.0:2828
```
```suggestion
  - http://192.168.1.42:2828
  - http://192.168.0.44:2828
```

source/agora/app.d
```diff
@@ -12,12 +20,38 @@
...
+
+/// Application entry point
+public int main (string[] args)
```
Geod24: `private`

source/agora/app.d
```diff
@@ -12,12 +20,38 @@
...
 {
-	// TODO:
-    // - Config parsing
-    // - Network interface setup
-    // - Event loop
-    // - End world hunger
+    CommandLine cmdln;
+    if (!parseCommandLine(cmdln, args))
+        return 0;  // help shown, early exit
```
Geod24:
```suggestion
        return 1;  // help shown, early exit
```

source/agora/common/Config.d
```diff
@@ -0,0 +1,310 @@
...
+    string config_path = "config.yaml";
+
+    /// check state of config file and exit early
+    bool config_check;
+}
+
+/// Main config
+public struct Config
+{
+    static assert(!hasUnsharedAliasing!(typeof(this)));
```
Geod24: Add a problem explaining the comment why

source/agora/node/Node.d
```diff
@@ -0,0 +1,99 @@
...
+
+module agora.node.Node;
+
+import agora.common.API;
+import agora.common.Config;
+import agora.common.crypto.Key;
+import agora.common.Version;
+import agora.node.Network;
+
+import vibe.d;
```
Geod24, line 21: Nope!

source/agora/node/NodeTest.d
```diff
@@ -0,0 +1,364 @@
...
+
+    //case NetworkTopology.Balanced:
+    //    assert(nodes.length == 4, "Balanced topology requires 4 nodes");
+    //    assert(0, "Not implemented");
+
+    //case NetworkTopology.Tiered:
+    //    assert(nodes.length == 10, "Tiered topology requires 10 nodes");
+    //    assert(0, "Not implemented");
+
+    //case NetworkTopology.Cyclic:
```
Geod24: Remove dead code

source/agora/node/NodeTest.d
```diff
@@ -0,0 +1,364 @@
...
+    //    assert(nodes.length == 7, "SinglePoF topology requires 7 nodes");
+    //    assert(0, "Not implemented");
+    }
+}
+
+///
+unittest
+{
+    import core.thread;
+    import std.stdio;
```
Geod24, line 316: Why the imports ?

source/agora/node/NodeTest.d
```diff
@@ -0,0 +1,364 @@
...
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.NodeTest;
```
Geod24, line 14: I think this should live in a separate package (`agora.test` maybe ?)

source/agora/node/NodeTest.d
```diff
@@ -0,0 +1,364 @@
...
+
+        enforce(fully_discovered.length == this.registry.length,
+            format("Got %s/%s discovered nodes. Missing nodes: %s",
+                fully_discovered.length,
+                this.registry.length,
+                this.registry.byKey.filter!(a => !!(a in fully_discovered))));
+
+        // todo: trace logging during unittesting?
+        // writefln("Discovered %s/%s nodes.", fully_discovered.length,
+        //     this.registry.length);
```
Geod24: Remove dead code

source/agora/node/NodeTest.d
```diff
@@ -0,0 +1,364 @@
...
+
+    ///
+    public override API[] getAllAPIs ( )
+    {
+        return this.registry.byValue.array;
+    }
+}
+
+///
+class TestNodeRegistry : TestRegistry
```
Geod24, line 120: For the moment let's keep it as it, but the separation was done so that you could have different registry for different tests (different registry == different node behavior)

source/agora/node/Network.d
```diff
@@ -0,0 +1,240 @@
...
+import agora.common.crypto.Key;
+import agora.common.Config;
+import agora.common.Data;
+import agora.common.Set;
+import agora.node.Client;
+
+import libsodium.crypto_sign_ed25519;
+import libsodium.randombytes;
+
+import vibe.d;
```
Geod24: Nope, this is fine in test code but not here

source/agora/node/Client.d
```diff
@@ -0,0 +1,193 @@
...
+module agora.node.Client;
+
+import agora.common.API;
+import agora.common.Config;
+import agora.common.crypto.Key;
+import agora.common.Data;
+import agora.common.Set;
+import agora.common.Version;
+
+import libsodium.crypto_sign_ed25519;
```
Geod24: Leftover ?

source/agora/common/Version.d
```diff
@@ -0,0 +1,37 @@
...
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.Version;
```
Geod24: Why this module ?

source/agora/node/Network.d
AndrejMitrovic: You mean using the package module? Why is it not ok?

source/agora/common/Version.d
AndrejMitrovic: Useless actually, can remove it for now.

source/agora/node/NodeTest.d
AndrejMitrovic: Btw, how exactly do we plan on simulating a "bad" node? In theory you can only interact with it through its publicly exposed API.

AndrejMitrovic: I'm guessing through subclassing the Node class somehow?

source/agora/common/Config.d
AndrejMitrovic: _because immutable sucks_ :PP

source/agora/app.d
AndrejMitrovic: I can change it if you want but I don't understand why or what for?

doc/config.example.yaml
AndrejMitrovic: It's really cute when you review your own text. :P

source/agora/app.d
Geod24: Just a matter of style I guess. Why would you have it public ? No one should call it.

source/agora/node/NodeTest.d
Geod24: Yes

source/agora/app.d
AndrejMitrovic: Conceptually it's the entry point of the application. In practice the real main function is in DRuntime.. anyway I'll make it private.

source/agora/node/Network.d
Geod24: Too many dependencies being pulled in

source/agora/node/NodeTest.d
Geod24: You subclass the node class and implement the behavior. Then in `registry`, you give it a name, e.g. `byzantine` or `dead` and return the appropriate node.

source/agora/node/Network.d
AndrejMitrovic: Good point.

source/agora/node/NodeTest.d
AndrejMitrovic: Ah, I see now why you have that name string passed as a parameter. Got it.

source/agora/app.d
Geod24: ^ After reading the code, forget about this

AndrejMitrovic: I could make it an enum to make it clear?

Geod24: Some command line tools returns success on `--help` (e.g. `git`), some don't.
Let's follow git here

source/agora/app.d
AndrejMitrovic: I tried doing this:

```

"configurations": [
    {
        "name": "agora",
        "targetName": "agora",
        "targetType": "executable",
        "mainSourceFile": "source/agora/app.d",
    },
    {
        "name": "unittest",
        "sourceFiles": [
            "source/scpp/build/DSizeChecks.o",
            "source/scpp/build/DLayoutChecks.o"
        ],
        "excludedSourceFiles": [ "source/agora/app.d" ]
    }
],
```

But then I get a linker error with `dub test`.

AndrejMitrovic:
~I fixed most stuff except the `dub test` target which still fails. Why does dub have to be so difficult..~ Edit: fixed

AndrejMitrovic:
There is also some missing documentation. Can be added soon or later after merging if I don't make it today.

AndrejMitrovic:
Ok, managed to fix the dub test target.

AndrejMitrovic:
I also renamed `Client` to `RemoteNode` to make it a bit clearer.

AndrejMitrovic:
> Ok, managed to fix the dub test target.

~Nope..~

AndrejMitrovic:
So the reason why I was confused that my tests didn't always build/link is because in the first commit `dub test` fails, but in the second one it doesn't. It's really weird. Anyway it doesn't matter, second one works fine.

Geod24 (approved):

PR #13, opened 2019-06-18
Temporarily disable LDC testing

Geod24:
It fails on some dependencies


AndrejMitrovic (approved):

AndrejMitrovic:
I'm guessing we can't enable it for non-mac OSes?

Geod24:
We could be I'll just fix LocalRest instead

PR #14, opened 2019-06-18
Implement basic metadata caching

AndrejMitrovic:
Might require some test-case as well. I'll consult with the test master @AndrejMitrovic if he wants me to write any.

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -37,6 +37,8 @@ node:
   # Private seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
   # Public address:  GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
   seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  # Path to the data directory (if the path doesn't exist it will be created)
+  data_dir: .cache
 
 ################################################################################
 ##                          Administrative interface                          ##
--- a/source/agora/common/Metadata.d
+++ b/source/agora/common/Metadata.d
@@ -1,0 +1,137 @@
+/*******************************************************************************
+
+    Defines metadata which is stored to disk (e.g. peer info)
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.Metadata;
+
+import agora.common.Data;
+import agora.common.Set;
+
+import vibe.core.log;
+import vibe.data.json;
+
+import std.array;
+import std.file;
+import std.format;
+import std.meta;
+import std.stdio;
+import std.typecons;
+
+/*******************************************************************************
+
+    Contains the serialized set of peers a node has connected to.
+    Note: Cannot be abstract, vibe.d requirement.
+
+*******************************************************************************/
+
+public class Metadata
+{
+    /// The set of peers we previously established connection to
+    public Set!Address peers;
+
+    /// Deserialize the Metadata from a Json string
+    public static void fromJsonString (Metadata metadata, string json)
+    {
+        metadata.peers = deserializeJson!(Set!Address)(json);
+    }
+
+    /// Dump the Metadata to a Json string
+    public string toJsonString ()
+    {
+        auto app = appender!string();
+        serializeToJson(app, this.peers);
+        return app.data[];
+    }
+
+    /// Load the metadata
+    public void load() { }
+
+    /// Dump the metadata
+    public void dump() { }
+}
+
+/// Metadata stored in memory (for unittests)
+public class MemMetadata : Metadata
+{
+    ///
+    public this ( )
+    {
+    }
+
+    ///
+    public override void load ()
+    {
+    }
+
+    ///
+    public override void dump ()
+    {
+    }
+}
+
+/// Metadata stored on disk (for persistence)
+public class DiskMetadata : Metadata
+{
+    /// Path to the metadata on disk
+    private string file_path;
+
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            data_dir = path to the data directory
+
+    ***************************************************************************/
+
+    public this ( string data_dir )
+    {
+        this.file_path = format("%s/%s", data_dir, "metadata.dat");
+    }
+
+    /// Load metadata from disk
+    public override void load ()
+    {
+        try
+        {
+            string raw_data = cast(string)std.file.read(file_path);
+            Metadata.fromJsonString(this, raw_data);
+        }
+        catch (Exception ex)
+        {
+            // ignored for now, missing metadata is not critical
+            logError(ex.message);
+        }
+    }
+
+    /// Dump metadata to disk
+    public override void dump ()
+    {
+        std.file.write(this.file_path, this.toJsonString());
+    }
+}
+
+///
+unittest
+{
+    auto meta = new MemMetadata();
+    const peer = "http://127.0.0.1:4567";
+    meta.peers.put(peer);
+    assert(peer in meta.peers);
+
+    auto str = meta.toJsonString();
+    meta.peers.clear();
+
+    Metadata.fromJsonString(meta, str);
+    assert(peer in meta.peers);
+}
--- a/source/agora/node/Network.d
+++ b/source/agora/node/Network.d
@@ -26,6 +26,7 @@ import agora.common.API;
 import agora.common.crypto.Key;
 import agora.common.Config;
 import agora.common.Data;
+import agora.common.Metadata;
 import agora.common.Set;
 import agora.node.RemoteNode;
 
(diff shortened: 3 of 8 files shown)


source/agora/common/Metadata.d
```diff
@@ -0,0 +1,109 @@
...
+    public bool hasLoaded ( )
+    {
+        return this.loaded;
+    }
+}
+
+///
+unittest
+{
+    auto meta = new Metadata(tempDir());
```
AndrejMitrovic: this might fail in Travis, I should probably just use a MemoryDevice buffer like we used to use in ocean.

AndrejMitrovic:
> std.file.FileException@std/file.d(1057): Failed to remove file /tmp/metadata.dat: No such file or directory

~I'll fix it up to use a memory device tomorrow.~

Edit: I went with the Appender approach because the Phobos File is a struct so it's not as easy to provide a custom "File" that writes to a buffer.

AndrejMitrovic:
It's green now.

MichaelKim20 (approved):

doc/config.example.yaml
```diff
@@ -37,6 +37,8 @@ node:
   # Private seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
   # Public address:  GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
   seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  # Path to the data directory (if the path doesn't exist it will be created)
+  data_dir: /Users/you/agora/data
```
Geod24: I'd put something more usable, like ".cache" so there are less items to modify.

source/agora/common/Metadata.d
```diff
@@ -0,0 +1,132 @@
...
+            scope file = File(this.file_name, "w");
+            auto output = file.lockingTextWriter();
+        }
+
+        serializeToJson(output, tuple(validators, listeners));
+    }
+
+    /// Load any existing metadata from disk (if it exists),
+    /// and set 'this.loaded' to true if at least one validator was loaded
+    public void load ( )
```
Geod24: This class shouldn't do IO. Make it accept a string and take the IO out of it.

source/agora/common/Metadata.d
```diff
@@ -0,0 +1,132 @@
...
+        return this.loaded;
+    }
+}
+
+///
+unittest
+{
+    auto meta = new Metadata(tempDir());
+    const Validator = "http://127.0.0.1:1234";
+    const Listener = "http://127.0.0.1:4567";
```
Geod24: Yeah that's not scalable. Since we serialize to `json`, you should be able to do:
```d
unittest
{
    auto meta = new Metadata("{ ... }");
}
```

The API here does not matter to much though. You might want to make your ctor non-validating and provide a `static Metadata fromFile(string path)` function as well as a `static Metadata fromString(string)`.

source/agora/node/Network.d
```diff
@@ -25,6 +26,7 @@ import libsodium.randombytes;
 
 import vibe.core.core;
 import vibe.core.log;
+import vibe.data.json;
```
Geod24: why?

Geod24 (changes requested):
Need to take the IO out of the class, or at least make it more modular so that we don't have to provide a special `version` for testing purposes.

source/agora/common/Metadata.d
AndrejMitrovic: Alright. I didn't think it was that important since it just stores IPs.

source/agora/node/Network.d
AndrejMitrovic: Leftover code, will remove.

source/agora/common/Metadata.d
AndrejMitrovic: Alrighty, I'll put these as helper functions outside the Metadata class.

AndrejMitrovic:
> Need to take the IO out of the class, or at least make it more modular so that we don't have to provide a special version for testing purposes.

The thing is that just moves the IO responsibility somewhere else. So, who gets to do IO if it's not encapsulated in this class? Do I need to store a File handle in the Network class side-by-side with the Metadata class? Because that's dirty to me.

AndrejMitrovic:
How about something like this:

```D
class Metadata
{
     T ips;  // the data to dump
     Writer writer;  // in-memory or file-based writer

     this (Writer writer) { this.writer = writer; }
}
```

And then in the Network class we instantiate it either with a file-based writer or an in-memory writer based on whether we compile in -unittest mode?

Can you think of a better or cleaner way to do this? With Ocean/Tango it was a bit simpler since File is a subclass of an IODevice, and you can just drop in a MemoryDevice to write to memory rather than to disk.

AndrejMitrovic:
Ok now that you've made your PRs I know what to do here.

source/agora/node/Network.d
```diff
@@ -257,4 +281,22 @@ class Network
     {
         return new RestInterfaceClient!API(address);
     }
+
+    /***************************************************************************
+
+        Reads the metadata from the provided disk path.
+
+        Subclasses could override this method and return
```
AndrejMitrovic: So I made something similar to `getMetadata`. The unittest code will override this function and return a `new MemMetadata` rather than `new DiskMetadata`.

Now I just need your other PR to be finished so I can add some tests on top of it. @Geod24

AndrejMitrovic:
Updated PR. Feel free to destroy.

source/agora/test/Base.d
```diff
@@ -258,9 +261,7 @@ public NetworkT makeTestNetwork (NetworkT : TestNetworkManager)
             configs ~= conf;
         }
 
-
-        auto net = new NetworkT(NodeConfig.init, node_configs.map!(
-            c => c.key_pair.address.toString()).array);
```
Geod24: I think that's where our approaches differ. I think that those arguments should be used / useful, and I thought they currently were, at least for the second one ?

Geod24: Well it makes sense in the context of your previous PR which only uses `apis`. But if we go this way, we do not need, nor do we want, to have a `NetworkManager`, because we're essentially conflating 2 use cases into one.

AndrejMitrovic: Yeah that's why I suggested splitting up these into two different classes. But if keep them the same class, the thing is these arguments that we pass here are just not used at all in the test-suite. Which approach do we take?

source/agora/node/Network.d
```diff
@@ -46,7 +47,7 @@ import std.random;
 public class NetworkManager
 {
     /// Config instance
-    private const NodeConfig node_config;
+    private const NodeConfig node_config = NodeConfig.init;
```
Geod24, line 49: Wat

source/agora/common/Metadata.d
```diff
@@ -0,0 +1,137 @@
...
+
+*******************************************************************************/
+
+public class Metadata
+{
+    /// The set of peers we previously established connection to
+    public Set!Address peers;
+
+    /// Deserialize the Metadata from a Json string
+    public static void fromJsonString (Metadata metadata, string json)
```
Geod24, line 42: I'm not soooooo found of that design. Mixing `class` and JSON object usually doesn't end well.

The way I tend to look at it is that classes have an "identity" as two bitwise-equal objects are different, but value types have no identity as they are defined by the sum of their content.
For JSON object you want value type, not something with an identity.

source/agora/test/Metadata.d
```diff
@@ -0,0 +1,69 @@
...
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.Metadata;
```
Geod24, line 14: Does it need its own test ? It should be more or less ignored in unittest for the time being.

Geod24 (changes requested):
Needs a rebase and I think we can simplify it

source/agora/test/Metadata.d
AndrejMitrovic: I'm confused. Are we doing TDD or not?

source/agora/common/Metadata.d
AndrejMitrovic: Well I can't use structs if I want to be able to provide Metadata instance that doesn't use IO and is used during unittests. What do you propose instead?

AndrejMitrovic:
Rebased.

AndrejMitrovic:
@Geod24 ready?

source/agora/test/Metadata.d
Geod24: We are but the point of abstracting away this is that it does nothing in unittests. So testing that it does nothing is a bit odd. In fact this test looks very much like the `Network` test.

source/agora/common/Metadata.d
```diff
@@ -0,0 +1,137 @@
...
+
+    /// Load the metadata
+    public void load() { }
+
+    /// Dump the metadata
+    public void dump() { }
+}
+
+/// Metadata stored in memory (for unittests)
+public class MemMetadata : Metadata
```
Geod24, line 63: I think this should live in the `Base` module.
One thing I really try to avoid is mixing test code with "regular code".
There's two reason for that:
- Better separation of concern. The "regular" code should not know about the test code, but the other way around is fine.
- The regular code should not have any special "unittest" path. Any such path is a special behavior that might break one of the class' assumption.

Take the example of the `version (unittest)` constructor: having a different constructor might break some assumptions the class makes (now or later), by -e.g.- leaving some fields empty.

source/agora/node/Network.d
Geod24: I'm pretty sure this can be removed ?

source/agora/node/Network.d
```diff
@@ -64,33 +65,53 @@ public class NetworkManager
...
+    private Metadata metadata;
+
+    /// Initial seed peers
+    const(string)[] seed_peers;
+
     /// DNS seeds
     private const(string)[] dns_seeds;
 
+    /// no-op. for use with unittests only
+    version (unittest) public this () { }
```
Geod24, line 78: Please no special-path for unittests :(

Geod24 (changes requested):
Remove the `unittest` special casing and we're G2G

Geod24:
Sorry for the delay!

source/agora/test/Metadata.d
AndrejMitrovic: How do you make it do nothing in unittests then? You do one of these:

- Do a `version(unittest) { } else { this.metadata.loadFromDisk()` }`
- Subclass Metadata and then provide a File-based one by default, and an in-memory one in the unittests

Is there a third option? I really don't understand.

AndrejMitrovic:
Please just show me how you want it done?

Geod24 (approved):

PR #18, opened 2019-06-20
Fix the Set() implementation to not use arrays

AndrejMitrovic:
The value of the hashmap should just be a simple type like bool,
not bool[].

We use the "bool[Type]" hashmap as a Set,
where we ignore the value (the bool).

Thanks to @MukeunKim for finding the bug.


Geod24 (approved):
Nice

PR #21, opened 2019-06-20
Split the tests in their own files

Geod24:
Have a 'Base' file for utilities that are most likely to be used accross tests,
and per-topic modules for tests.


MichaelKim20 (approved):

AndrejMitrovic:
Alright, I started doing something similar.

PR #22, opened 2019-06-20
add to 'put /message' of API

TrustHenry:
add to 'put /message' of API

### DIFF
--- a/source/agora/common/API.d
+++ b/source/agora/common/API.d
@@ -17,9 +17,6 @@ import agora.common.crypto.Key;
 import agora.common.Data;
 import agora.common.Set;
 
-import scpd.types.Stellar_SCP;
-import scpd.types.Stellar_types;
-
 import vibe.data.json;
 import vibe.web.rest;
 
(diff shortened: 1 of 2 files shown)


MichaelKim20 (approved):

source/agora/common/API.d
```diff
@@ -90,4 +90,22 @@ public interface API
     ***************************************************************************/
 
     public PublicConfig getPublicConfig();
+
+    /***************************************************************************
+
+        Send message to this node
+
+        Parameter: 
+            string message(Hex)
```
AndrejMitrovic: If you want to pass a hash around, you could just use `Hash` from `scpd.types.Stellar_types`. For example:

```D
public void setMessage(Hash hash)
```

source/agora/common/API.d
```diff
@@ -90,4 +90,22 @@ public interface API
...
+
+        Returns:
```
AndrejMitrovic: When a function is a `void` function, you don't have to document `Returns: void`, so you can remove this part.

AndrejMitrovic:
It looks better now. Maybe @Geod24 can add his review.

PR #24, opened 2019-06-21
Update dyaml dependency: v0.7.1 -> v0.8.0

Geod24:


source/agora/common/Config.d
```diff
@@ -177,8 +177,11 @@ public Config parseConfigFile (CommandLine cmdln)
 
     Node root = Loader.fromFile(cmdln.config_path).load();
 
-    enforce("network" in root && root["network"].isSequence,
-        "The 'network' section is mandatory and must specify at least one peer");
+    if (auto node = "network" in root)
```
AndrejMitrovic, line 180: Unrelated change?

Geod24: No they removed the property, look at the `enforce`.
Since I'm changing that code I might as well remove the double lookup.

AndrejMitrovic: oh sorry I misread the dependency

Geod24: https://github.com/dlang-community/D-YAML/commit/b63ea1aaae1f366d325c58aa36b19df4ec6c1130#r34022924

AndrejMitrovic (approved):

PR #27, opened 2019-06-21
Fix error message in tests

Geod24:
The message actually only printed those in fully_discovered, not missing in it.

### DIFF
--- a/source/agora/test/Network.d
+++ b/source/agora/test/Network.d
@@ -92,7 +92,7 @@ class TestNodeRegistry : TestRegistry
             format("Got %s/%s discovered nodes. Missing nodes: %s",
                 fully_discovered.length,
                 this.registry.length,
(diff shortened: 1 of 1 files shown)


source/agora/test/Network.d
```diff
@@ -92,7 +92,7 @@ class TestNodeRegistry : TestRegistry
             format("Got %s/%s discovered nodes. Missing nodes: %s",
                 fully_discovered.length,
                 this.registry.length,
-                this.registry.byKey.filter!(a => !!(a in fully_discovered))));
+                this.registry.byKey.filter!(a => !(a in fully_discovered))));
```
AndrejMitrovic, line 95: Oh right, this is because at first I printed out found nodes and then decided to switch this around. Without fixing this. ¯\\\_(ツ)_/¯

AndrejMitrovic (approved):

PR #32, opened 2019-06-23
Update LocalRest to v0.1.0

Geod24:
The new version changes the API so that the type is less
dependent on the internals of the instantiated node.

### DIFF
--- a/source/agora/test/Network.d
+++ b/source/agora/test/Network.d
@@ -38,12 +38,12 @@ class TestNodeRegistry : TestRegistry
         string id = config.node.key_pair.address.toString();
         auto tid = std.concurrency.locate(id);
         if (tid != tid.init)
-            return new RemoteAPI!(API, Node)(tid);
+            return new RemoteAPI!(API)(tid);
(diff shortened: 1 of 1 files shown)


source/agora/test/Network.d
```diff
@@ -38,12 +38,12 @@ class TestNodeRegistry : TestRegistry
         string id = config.node.key_pair.address.toString();
         auto tid = std.concurrency.locate(id);
         if (tid != tid.init)
-            return new RemoteAPI!(API, Node)(tid);
```
AndrejMitrovic, line 41: Nice. I was wondering if we could do something like this.

AndrejMitrovic (approved):

AndrejMitrovic:
Btw, I think Travis didn't trigger a rebuild after I merged another PR first. I'm guessing you need to manually rebase ths PR before Travis retests it?

Geod24:
Yep, Travis doesn't rebuild after a merge (on purpose, it's in their FAQ)

Issue #15, opened 2019-06-19
Re-enable LDC testing


Geod24:
Make sure [this](https://github.com/bpfkorea/agora/pull/13) is really temporary.
Depends on https://github.com/Geod24/localrest/issues/1

PR #35, opened 2019-06-24
Make unittests communicate via `localrest`, not local network

Geod24:
This stream of change, better reviewed commit-by-commit, changes ~network~ communication in unittests to be thread based RPC instead of actual network call (which messes up with Vibe.d).
It also fixes the LDC failure.
I removed `Registry` and `NodeSet` as they serve the very same purpose as `Network`. We might want to re-introduce that level of abstraction in the future, but for the moment it works ™️ .

### DIFF
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -17,7 +17,9 @@ import agora.common.API;
 import agora.common.Config;
 import agora.common.crypto.Key;
 
-import vibe.d;
+import vibe.core.log;
+import vibe.data.json;
+import vibe.web.rest;
 
 
 /*******************************************************************************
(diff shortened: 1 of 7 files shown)


Geod24:
Fixes https://github.com/bpfkorea/agora/issues/15

source/agora/node/Node.d
```diff
@@ -21,15 +21,15 @@ import agora.node.Network;
 import vibe.d;
 
 /// Ditto
-public class Node : API
+public class Node (Net) : API
```
AndrejMitrovic: Docs for `Net`?

AndrejMitrovic:
Just a small note but otherwise LGTM.

AndrejMitrovic (approved):

Geod24:
Added documentation

AndrejMitrovic:
There is something missing here.

Namely, calling `node.start()`. The unittests pass because `enforce(fully_discovered.length == this.listeners.length);` succeeds, and both numbers are 0.

AndrejMitrovic:
Of course the slight issue here is that `start` is not exposed in the public API (and it shouldn't be). That's the reason I moved that call into the constructor instead of into main().

AndrejMitrovic:
There is a small issue where a node tries to connect to itself. However I can work on this.

AndrejMitrovic:
It's because of this:

```
        // add our own IP to the list of banned IPs to avoid
        // the node communicating with itself
        this.banned_addresses.put(
            format("http://%s:%s", this.node_config.address, this.node_config.port));
```

The code assumes it uses IPs.

AndrejMitrovic:
Want me to merge it and I'll fix it?

Geod24:
That would work

PR #42, opened 2019-06-24
Remove enforce on network argument from Node

Geod24:
Node should not care what is a valid argument for Network,
it should just pass it along.
In addition, since this enforcement is done at the config parsing level,
we can make it an assert in Network's constructor.

### DIFF
--- a/source/agora/node/Network.d
+++ b/source/agora/node/Network.d
@@ -70,7 +70,8 @@ class Network
 
     /// Ctor
     public this (in NodeConfig node_config, in string[] peers)
-    {
+    in { assert(peers.length > 0, "No network option found"); }
+    do {
         this.node_config = node_config;
         this.addAddresses(Set!Address.from(peers));
 
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -50,7 +50,6 @@ public class Node (Network) : API
     public this (const Config config)
     {
         this.config = config;
-        enforce(this.config.network.length > 0, "No network option found");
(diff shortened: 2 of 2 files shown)


source/agora/node/Network.d
```diff
@@ -70,7 +70,8 @@ class Network
 
     /// Ctor
     public this (in NodeConfig node_config, in string[] peers)
-    {
+    in { assert(peers.length > 0, "No network option found"); }
+    do {
```
AndrejMitrovic, line 74: `do`? I thought this was supposed to be `body`?

AndrejMitrovic: ah they changed it, didn't they..

AndrejMitrovic: There was some DIP that simplified `in` expressions, but I'm not sure if it was approved or implemented. It would avoid these ugly braces. :)

AndrejMitrovic (approved):

source/agora/node/Network.d
Geod24: It's implemented (but its ugly and not all editors support it)

AndrejMitrovic: What is the actual syntax?

Geod24: https://dlang.org/spec/contracts.html#pre_post_contracts

AndrejMitrovic: ah so it would be something like:

```D
public this ( in string[] peers) 
    in (peers.length > 0, "No peers")
{
}
```

Similar to template constraints.

PR #45, opened 2019-06-25
Don't interfere with networking during unittests

AndrejMitrovic:

PR #46, opened 2019-06-25
Fix visibility issue with calling a protected method

AndrejMitrovic:
`createNewNode` might be overriden in a subclass, but if it is we can't directly call it because it might be defined in another module.

Perhaps it should be made public instead? Or this could even be a drawback in the visibility rules in D / a bug.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -201,10 +201,14 @@ public NetworkT makeTestNetwork (NetworkT : TestNetwork)
 
         auto net = new NetworkT(NodeConfig.init, node_configs.map!(
             c => c.key_pair.address.toString()).array);
+
+        // Workaround https://issues.dlang.org/show_bug.cgi?id=20002
+        TestNetwork base_net = net;
+
         foreach (idx, ref conf; configs)
         {
             const address = node_configs[idx].key_pair.address;
-            net.apis[address] = net.createNewNode(address.toString(), conf);
+            net.apis[address] = base_net.createNewNode(address.toString(), conf);
         }
 
         net.apis.each!(a => a.start());


source/agora/test/Base.d
```diff
@@ -200,9 +200,16 @@ public NetworkT makeTestNetwork (NetworkT : TestNetwork)
...
+        // accessibility workaround: Base module has access to protected
+        // createNewNode because it's defined in TestNetwork in *this* module,
+        // however it cannot access NetworkT's override because it usually
+        // lives in another module
+        auto base_net = cast(TestNetwork)net;
+
         foreach (idx, ref conf; configs)
         {
-            net.apis ~= net.createNewNode(node_configs[idx].key_pair.address.toString(), conf);
+            net.apis[node_configs[idx].key_pair.address] = base_net.createNewNode(node_configs[idx].key_pair.address.toString(), conf);
```
Geod24: Bad rebase ?

AndrejMitrovic: right.. well I'll need to rebase it again if https://github.com/bpfkorea/agora/pull/45 is merged first.

source/agora/test/Base.d
```diff
@@ -200,9 +200,16 @@ public NetworkT makeTestNetwork (NetworkT : TestNetwork)
 
         auto net = new NetworkT(NodeConfig.init, node_configs.map!(
             c => c.key_pair.address.toString()).array);
+
+        // accessibility workaround: Base module has access to protected
+        // createNewNode because it's defined in TestNetwork in *this* module,
+        // however it cannot access NetworkT's override because it usually
+        // lives in another module
+        auto base_net = cast(TestNetwork)net;
```
Geod24: Can you report the DMD bug ?
Then just:
```d
auto ret = ....
// Workaround LINK-TO-BUGZILLA
TestNetwork net = ret;
[...]
return ret;
```

I don't like `cast` because it is not future proof: If you change the type, an implicit conversion will fail, but a `cast` will just return `null`.

AndrejMitrovic: I have a feeling that it's already reported somewhere. I'll try to find it.

AndrejMitrovic: And yeah I shouldn't cast.

AndrejMitrovic: It seems it's similar to https://issues.dlang.org/show_bug.cgi?id=2417, but there they just changed the spec as a "wontfix". (even though it says "fixed")

AndrejMitrovic: However I will file a new one.

AndrejMitrovic: Filed as https://issues.dlang.org/show_bug.cgi?id=20002 and updated the PR.

AndrejMitrovic:
Rebased.

Geod24 (approved):

PR #47, opened 2019-06-26
Fix naming of unittest classes and file names

AndrejMitrovic:
I often find myself looking at the wrong module when trying to find the Network class. I think this naming convention is easier for humans to distinguish between test modules and core modules, even if it's a little bit repetitive.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -86,6 +86,20 @@ public class TestNetwork : Network
 
     /***************************************************************************
 
+        Start each of the nodes
+
+        Params:
+            count = Expected number of nodes
+
+    ***************************************************************************/
+
+    public void start ()
+    {
+        this.apis.each!(a => a.start());
+    }
+
+    /***************************************************************************
+
         Wait until the nodes have reached discovery.
         If after 10 query attempts they haven't connected
         to each other, throw an exception.
@@ -260,7 +274,6 @@ public NetworkT makeTestNetwork (NetworkT : TestNetwork)
             net.apis[address] = base_net.createNewNode(address.toString(), conf);
         }
 
-        net.apis.each!(a => a.start());
         return net;
 
     case NetworkTopology.Balanced:
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -104,12 +104,9 @@ public class TestNetwork : Network
         If after 10 query attempts they haven't connected
         to each other, throw an exception.
 
-        Params:
-            count = Expected number of nodes
-
     ***************************************************************************/
 
-    public void waitUntilConnected (size_t count)
+    public void waitUntilConnected ()
     {
         import core.thread;
         import std.stdio;
@@ -132,19 +129,19 @@ public class TestNetwork : Network
             }
 
             // we're done
-            if (fully_discovered.length == count)
+            if (fully_discovered.length == this.apis.length)
                 return;
 
             // try again
             auto sleep_time = 1.seconds;  // should be enough time
             writefln("Sleeping for %s. Discovered %s/%s nodes", sleep_time,
-                fully_discovered.length, count);
+                fully_discovered.length, this.apis.length);
             Thread.sleep(sleep_time);
         }
 
         assert(fully_discovered.length == count,
                format("Got %s/%s discovered nodes. Missing nodes: %s",
-                   fully_discovered.length, count, this.todo_addresses));
+                   fully_discovered.length, this.apis.length, this.todo_addresses));
     }
 }
 
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -83,6 +83,55 @@ public class TestNetwork : Network
         std.concurrency.register(address, api.tid());
         return api;
     }
+
+    /***************************************************************************
+
+        Wait until the nodes have reached discovery.
+        If after 10 query attempts they haven't connected
+        to each other, throw an exception.
+
+        Params:
+            count = Expected number of nodes
+
+    ***************************************************************************/
+
+    public void waitUntilConnected (size_t count)
+    {
+        import core.thread;
+        import std.stdio;
+
+        const attempts = 10;
+
+        bool[PublicKey] fully_discovered;
+
+        foreach (_; 0 .. attempts)
+        {
+            foreach (key, api; this.apis)
+            try
+            {
+                if (api.getNetworkInfo().state == NetworkState.Complete)
+                    fully_discovered[key] = true;
+            }
+            catch (Exception ex)
+            {
+                // just continue
+            }
+
+            // we're done
+            if (fully_discovered.length == count)
+                return;
+
+            // try again
+            auto sleep_time = 1.seconds;  // should be enough time
+            writefln("Sleeping for %s. Discovered %s/%s nodes", sleep_time,
+                fully_discovered.length, count);
+            Thread.sleep(sleep_time);
+        }
+
+        assert(fully_discovered.length == count,
+               format("Got %s/%s discovered nodes. Missing nodes: %s",
+                   fully_discovered.length, count, this.todo_addresses));
+    }
 }
 
 /// Temporary hack to work around the inability to do 'start' from main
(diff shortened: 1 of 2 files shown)


Geod24:
Hum, okay, not fan of the repetition, but there seem to be a consensus among the team on that.
Note that I expect that subclassing `TestNetwork` will be a quite common operation. The reason is that `TestNetwork` is the place where the node behavior is defined (as in, do we provide a byzantine node, a regular node, etc...)

source/agora/test/Base.d
```diff
@@ -83,6 +83,48 @@ public class TestNetwork : Network
         std.concurrency.register(address, api.tid());
         return api;
     }
+
+    /// Wait a certain time until the nodes have reached discovery
+    /// If after 10 query attempts they still all haven't discovered => assert
+    /// Params:
+    ///   count = Expected number of nodes
```
Geod24: Too many lines for `///`

source/agora/test/TestNetwork.d
```diff
@@ -0,0 +1,35 @@
...
+import agora.common.Config;
+import agora.common.Data;
+import agora.common.crypto.Key;
+import agora.node.Network;
+import agora.node.Node;
+import agora.test.Base;
+
+import std.algorithm.iteration;
+import std.exception;
+import std.format;
```
Geod24: All those import don't seem necessary ?

AndrejMitrovic:
> Hum, okay, not fan of the repetition, but there seem to be a consensus among the team on that.
Note that I expect that subclassing TestNetwork will be a quite common operation.

There is a small usability issue with this. When you do inherit from `TestNetwork`, you must ensure to override `createNewNode`, otherwise the node that's instantiated will still be templated on the base class. For example in here:

```D
/// Initialize a new node
protected TestAPI createNewNode (Address address, Config conf)
{
    auto api = RemoteAPI!TestAPI.spawn!(TestNode!TestNetwork)(conf);
    std.concurrency.register(address, api.tid());
    return api;
}
```

In the derived class you would have to override createNewNode to do `spawn!(TestNode!DerivedTestNetwork)`. I ran into it a few times already.

source/agora/test/Base.d
AndrejMitrovic: I guess we are going to use `/** */` everywhere for multiline comments? I'm 50% sure that using multiline `///` comments works too. Anyway I will fix it.

Geod24:
True, that's not ideal. Ideally I'd like to get rid of the template arg and just instantiate it from outside the class though.

AndrejMitrovic:
Aha so something like `getNetwork()` inside of the Node?

Geod24:
I have to look at alternatives, but more like passing `Network` to the ctor, and having `LocalRest` support an instantiating function instead of `new Type`

AndrejMitrovic:
Hmm there is one issue, `waitUntilConnected` will assert in a place other than the function call. Let me rework this a little bit.

AndrejMitrovic:
I've added 3 more commits. Maybe you want me to split this up into two PRs, or is that overkill?

source/agora/test/Base.d
```diff
@@ -104,12 +104,9 @@ public class TestNetwork : Network
         If after 10 query attempts they haven't connected
         to each other, throw an exception.
 
-        Params:
-            count = Expected number of nodes
-
     ***************************************************************************/
 
-    public void waitUntilConnected (size_t count)
+    public void waitUntilConnected ()
```
Geod24: Well the point of doing this is that, if for some reason the `apis` array is empty, this just passes.

AndrejMitrovic: Why should it pass?

source/agora/test/Base.d
```diff
@@ -139,9 +141,7 @@ public class TestNetwork : Network
             Thread.sleep(sleep_time);
         }
 
-        assert(fully_discovered.length == count,
-               format("Got %s/%s discovered nodes. Missing nodes: %s",
-                   fully_discovered.length, this.apis.length, this.todo_addresses));
```
Geod24: Honestly I don't see much of an improvement here... We do have stack traces so the assert being here shouldn't be a problem?

AndrejMitrovic: Aren't stack traces still broken?

source/agora/test/Base.d
Geod24: You'll enter the `foreach` for `attempts`. Then the `foreach` on `this.apis` will not be executed because `this.apis.length == 0`. Then L132 will pass (0 == 0) and it returns.

source/agora/test/Base.d
AndrejMitrovic:
```
core.exception.AssertError@source/agora/test/Base.d(145): Got 4/5 discovered nodes. Missing nodes: ["GD6GUK2NDHNTAS5TAKCIE6BFSLG5WELOE2GLRMZTORCY4FPWZOY5EPAZ":true, "GCI6VCQJG53N3HWQVENP6WED56YREJTDRQOW2VC2UBEADCST42GDE3QB":true, "GBDESH4FGFCEMTDADJMCFKPCHHZSFKZ4U5BYLX6NIB6ED3RDVZ23UE2Z":true, "GBCJQKB4QRTPUX4Y2H6PAKKRKIKH7RV4K5PK5CGTUI2EDH2SD3NAQY7Z":true]
----------------
??:? _d_assert_msg [0x1026c86c6]
/usr/local/opt/dmd/include/dlang/dmd/object.d:2911 void agora.test.Base.TestNetwork.waitUntilConnected(ulong) [0x1023c8051]
/usr/local/opt/dmd/include/dlang/dmd/object.d:2911 void agora.test.TestNetwork.__unittest_L19_C1() [0x1023ca805]
source/agora/app.d:36 void agora.test.TestNetwork.__modtest() [0x10234e5e8]
??:? int core.runtime.runModuleUnitTests().__foreachbody2(object.ModuleInfo*) [0x1026c93fb]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)).__lambda2(immutable(object.ModuleInfo*)) [0x1026bd9ce]
```

Yeah they're still broken.

source/agora/test/Base.d
AndrejMitrovic: I mean the `assert(network.getDiscoveredNodes().length == NodeCount);` will fail. And then you get the assert message where it's used.

AndrejMitrovic:
@Geod24 I think Michael's only issue was with the subclass `Network => TestNetwork => Network`, so I could revert the commit which renames the files. That alright?

AndrejMitrovic:
Reverted that commit.

source/agora/test/Network.d
```diff
@@ -13,76 +13,11 @@
 
 module agora.test.Network;
 
-version (unittest):
```
Geod24: That should stay though

source/agora/test/Base.d
```diff
@@ -84,6 +84,20 @@ public class TestNetwork : Network
         return api;
     }
 
+    /***************************************************************************
+
+        Start each of the nodes
+
+        Params:
+            count = Expected number of nodes
```
Geod24, line 92: What ? :D

source/agora/test/Base.d
```diff
@@ -84,6 +84,20 @@ public class TestNetwork : Network
...
+
+    ***************************************************************************/
+
+    public void startNodes ()
```
Geod24: Maybe just `start`?

source/agora/test/Base.d
Geod24: But then you rely on the caller checking that you actually connected to all the nodes.

source/agora/test/Base.d
Geod24: They are, but moving this responsibility to the caller does not appeal to me too much... But okay, let's give it a try.

source/agora/test/Base.d
AndrejMitrovic: This was moved from one place to another. And also, I remove it in a separate commit!

AndrejMitrovic:
Applied fixes.

Geod24 (approved):
Heh let's try

PR #37, opened 2019-06-24
Create GossipProtocol

MichaelKim20:
Implement class GossipProtocol
Implement class Cache
Add unittest for GossipProtocol

PR #51, opened 2019-06-30
Expose LocalRest's event loop and make use of it for asynchronous tasks

Geod24:
It *looks like* we could also use the Vibe.d event loop but I have little trust in the whole thing.
And this works with #37

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -52,6 +52,7 @@ public class TestNetworkManager : NetworkManager
 {
     static import std.concurrency;
     import geod24.LocalRest;
+    import core.time;
 
     /// Used by the unittests in order to directly interact with the nodes,
     /// without trying to handshake or do any automatic network discovery.
@@ -145,6 +146,20 @@ public class TestNetworkManager : NetworkManager
 
         return fully_discovered.byKey.array;
     }
+
+    /// This is a hack
+    protected override void runTask ( void delegate() dg)
+    {
+        static import geod24.LocalRest;
+        geod24.LocalRest.runTask(dg);
+    }
+
+    /// Ditto
+    protected override void wait (Duration dur)
+    {
+        static import geod24.LocalRest;
+        geod24.LocalRest.sleep(dur);
+    }
 }
 
 /// Temporary hack to work around the inability to do 'start' from main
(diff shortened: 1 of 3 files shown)


source/agora/test/Base.d
```diff
@@ -58,6 +58,9 @@ public class TestNetworkManager : NetworkManager
     /// Also kept here to avoid any eager garbage collection.
     public RemoteAPI!TestAPI[PublicKey] apis;
 
+    /// Workaround compiler bug that triggers in `std.concurrency`
+    protected __gshared std.concurrency.Tid[string] tbn;
```
AndrejMitrovic, line 63: Is there an associated tracking bug for this? It would be useful to link it here.

source/agora/test/Base.d
```diff
@@ -56,7 +56,7 @@ public class TestNetworkManager : NetworkManager
     /// Used by the unittests in order to directly interact with the nodes,
     /// without trying to handshake or do any automatic network discovery.
     /// Also kept here to avoid any eager garbage collection.
-    public TestAPI[PublicKey] apis;
```
AndrejMitrovic: Where and how ill they be used directly? I mean if you could just describe a use-case?

AndrejMitrovic: This comment should have been on the right side.

source/agora/test/Base.d
Geod24: No because I don't have simple steps to reproduce.

source/agora/test/Base.d
Geod24: I don't understand the question?

AndrejMitrovic: "Expose something because it will be used". How will it be used?

AndrejMitrovic (approved):

AndrejMitrovic:
Let me know if you want to add anything, otherwise I'll merge.

source/agora/test/Base.d
Geod24: Oh right. That's an artifact.

Geod24:
Removed offending commit

Geod24:
Also rebased on `v0.x.x`

PR #52, opened 2019-07-01
Actually use gdc-9 for testing

Geod24:
```
Because `d: dmd` is applied after `DC=gdc` is exported,
it overrode DC and we ended up testing with DMD.
```

Since `gdc-9` does not have a critical bug to Phobos that I made, this is blocked on a new release of `gdc-9`. Iain, the maintainer, might include it in 9.2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,22 @@ matrix:
     - d: dmd-nightly
     - d: ldc-latest-ci
   fast_finish: true
+  include:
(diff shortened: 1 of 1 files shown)


Geod24:
That will not happen for a while, we need at least GDC10, which is not released yet, and if we want to use all the ~nice new features~ bug fixes we made I don't see us using GDC for a while.

PR #55, opened 2019-07-02
Add basic transaction data layout

Geod24:
As simple as that

### DIFF
--- a/source/agora/common/Transaction.d
+++ b/source/agora/common/Transaction.d
@@ -1,0 +1,89 @@
+/*******************************************************************************
+
+    Defines the data structure of a transaction
+
+    The current design is heavily influenced by Bitcoin: as we have a UTXO,
+    starting with a simple input/output approach seems to make the most sense.
+    Script is not implemented: instead, we currently only have a simple signature
+    verification step.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.Transaction;
+
+import agora.common.Data;
+import agora.common.crypto.Key;
+
+
+/*******************************************************************************
+
+    Currency amount type to use
+
+    This is simply an alias to `long` at the moment, and should be made smarter.
+    Currently this has two major drawbacks:
+    - It does not do any overflow / underflow checks (over the total amount)
+    - It can go negative
+    We should probably wrap this in a smarter type to do currency operations,
+    and move it to a shared location, as currency operations might be performed
+    independently of `Transaction`.
+
+*******************************************************************************/
+
+public alias Amount = long;
+
+/*******************************************************************************
+
+    Represents a transaction (shortened as 'tx')
+
+    Agora uses a UTXO model for transactions, so the transaction format is
+    originally derived from that of Bitcoin.
+
+*******************************************************************************/
+
+public struct Transaction
+{
+    /// The list of unspent `outputs` from previous transaction(s) that will be spent
+    public Input[] inputs;
+
+    /// The list of newly created outputs to put in the UTXO
+    public Output[] outputs;
+}
+
+/*******************************************************************************
+
+    Represents an entry in the UTXO
+
+    This is created by a valid `Transaction` and is added to the UTXO after
+    a transaction is confirmed.
+
+*******************************************************************************/
+
+public struct Output
+{
+    /// The monetary value of this output, in 1/10^7
+    public Amount value;
+
+    /// The public key that can redeem this output (A = pubkey)
+    /// Note that in Bitcoin, this is an address (the double hash of a pubkey)
+    public PublicKey address;
+}
+
+/// The input of the transaction, which spends a previously received `Output`
+public struct Input
+{
+    /// The hash of a previous transaction containing the `Output` to spend
+    public Hash previous;
+
+    /// Index of the `Output` in the `previous` `Transaction`
+    public uint index;
+
+    /// A signature that should be verified using the `previous[index].address` public key
+    public Signature signature;
+}
--- a/source/agora/common/Data.d
+++ b/source/agora/common/Data.d
@@ -24,3 +24,6 @@ public alias Hash = BitBlob!256;
 
 /// A network address
 public alias Address = string;
+
+/// The type of a signature
+public alias Signature = BitBlob!256;


source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
+/*******************************************************************************
+
+    Defines the data structure of a transaction
+
+    The current design is heavily influenced by Bitcoin: as we have an UTXO,
```
AndrejMitrovic:
```suggestion
    The current design is heavily influenced by Bitcoin: as we have a UTXO,
```

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+    starting with a simple input/output approach seems to make the most sense.
+    Script is not implemented: instead, we currently only do a simple signature
+    verification.
```
AndrejMitrovic:
```suggestion
    verification step.
```

Alternatively: `only do simple signature verification`

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+    Currently this has two major drawbacks:
+    - It does not do any overflow / underflow checks (over the total amount)
+    - It can go negative
+    We should probably wrap this in a smarter type to do currency operations,
+    and move it to a shared location, as currency operations might be performed
+    independently of `Transaction`.
+
+*******************************************************************************/
+
+public alias Amount = long;
```
AndrejMitrovic, line 39: Bitcoin seems to allow negative values (at least internally?), but I'm not sure why. Do you know?

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+
+/*******************************************************************************
+
+    Represents a transaction (shortened as 'tx')
+
+    Agora uses an UTXO model for transactions, so the transaction format is
```
AndrejMitrovic:
```suggestion
    Agora uses a UTXO model for transactions, so the transaction format is
```

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+    originally derived from that of Bitcoin.
+
+*******************************************************************************/
+
+public struct Transaction
+{
+    /// The list of `outputs` from previous transaction(s) being spent
+    Input[] inputs;
```
AndrejMitrovic:
```suggestion
    Input[] inputs;

```

I know we don't have a style guide, but I find it hard to look a fields when they're so tightly packed..

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+    This is created by a valid `Transaction` and is added to the UTXO after
+    a transaction is confirmed.
+
+*******************************************************************************/
+
+public struct Output
+{
+    /// The monetary value of this output, in 1/10^7
+    Amount value;
+    /// The public key that can redeem this output (A = pubkey)
```
AndrejMitrovic, line 73: `(A = pubkey)` - what? where does "A" come from?

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+    /// The list of newly created outputs to put in the UTXO
+    Output[] outputs;
+}
+
+/*******************************************************************************
+
+    Represents an entry in the UTXO
+
+    This is created by a valid `Transaction` and is added to the UTXO after
+    a transaction is confirmed.
```
AndrejMitrovic, line 64: What does it mean that a transaction is _confirmed_, in this context? It should be documented.

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+
+*******************************************************************************/
+
+public struct Output
+{
+    /// The monetary value of this output, in 1/10^7
+    Amount value;
```
AndrejMitrovic:
```suggestion
    Amount value;

```

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+    /// The public key that can redeem this output (A = pubkey)
+    /// Note that in Bitcoin, this is an address (the double hash of a pubkey)
+    PublicKey address;
+}
+
+/// The input of the transaction, which spends a previously received `Output`
+public struct Input
+{
+    /// The hash of a previous transaction containing the `Output` to spend
+    Hash previous;
```
AndrejMitrovic:
```suggestion
    Hash previous;

```

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+    /// Index of the `Output` in the `previous` `Transaction`
+    uint index;
```
AndrejMitrovic:
```suggestion
    uint index;

```

source/agora/common/Transaction.d
AndrejMitrovic: So w.r.t. style guide, do we use explicit protection attributes? Or just make it implied that things are public?

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+    Represents a transaction (shortened as 'tx')
+
+    Agora uses an UTXO model for transactions, so the transaction format is
+    originally derived from that of Bitcoin.
+
+*******************************************************************************/
+
+public struct Transaction
+{
+    /// The list of `outputs` from previous transaction(s) being spent
```
AndrejMitrovic:
```suggestion
    /// The list of unspent `outputs` from previous transaction(s) being spent
```

AndrejMitrovic: Wait actually that's double confusing.

AndrejMitrovic: How about:

```suggestion
    /// The list of unspent `outputs` from previous transaction(s) that will be spent
```

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+
+module agora.common.Transaction;
+
+import agora.common.Data;
+import agora.common.crypto.Key;
+
+
+/*******************************************************************************
+
+    Currency amount to use
```
AndrejMitrovic:
```suggestion
    Currency amount type to use
```

source/agora/common/Transaction.d
Geod24: Oups shall remove

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// The input of the transaction, which spends a previously received `Output`
+public struct Input
+{
+    /// The hash of a previous transaction containing the `Output` to spend
+    Hash previous;
+    /// Index of the `Output` in the `previous` `Transaction`
+    uint index;
+    /// A signature done using `previous[index].address`
```
AndrejMitrovic: I think this should be

```suggestion
    /// A signature that should be verified using the `previous[index].address` public key
```

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
...
+{
+    /// The monetary value of this output, in 1/10^7
+    Amount value;
+    /// The public key that can redeem this output (A = pubkey)
+    /// Note that in Bitcoin, this is an address (the double hash of a pubkey)
+    PublicKey address;
+}
+
+/// The input of the transaction, which spends a previously received `Output`
+public struct Input
```
AndrejMitrovic, line 79: Mmmhm.. I dislike that you've ordered them as:

```
public struct Output { ... }
public struct Input { ... }
```

Can we make them Input and then Output instead? I know I'm nitpicking.

source/agora/common/Transaction.d
AndrejMitrovic: `uint index`

I think in the future we should probably make this type smaller.

AndrejMitrovic:
Done with review.

source/agora/common/Transaction.d
Geod24: Agreed

source/agora/common/Transaction.d
Geod24: The reasoning is that, while your approach is the natural approach, `Input` is actually defined in terms of `Output`, not the other way around. An `Output` never references an `Input`, but an `Input` always reference an `Input`. `Output` can exists without `Input` (Coinbase) but not the other way around.
It sounds backwards but that's the way it's done in BTC and I can't think of something better.

Geod24: That being said, I don't mind changing the ordering at all, just wanted to provide a reasoning for the ordering.

AndrejMitrovic: Yeah that's fair.

source/agora/common/Transaction.d
```diff
@@ -0,0 +1,85 @@
+/*******************************************************************************
+
+    Defines the data structure of a transaction
+
+    The current design is heavily influenced by Bitcoin: as we have an UTXO,
+    starting with a simple input/output approach seems to make the most sense.
+    Script is not implemented: instead, we currently only do a simple signature
```
```suggestion
    Script is not implemented: instead, we currently only have a simple signature
```

source/agora/common/Transaction.d
Geod24: Not exactly. I imagine that might be related to dealing with `a - b` (remember Don hating on unsigned types ? ;) )

source/agora/common/Transaction.d
AndrejMitrovic: The problem isn't the `do` or `have`, it's the `a`. English is complicated, I know. And I'm evil for nitpicking here. I'll stop. :P

source/agora/common/Transaction.d
AndrejMitrovic: Ah that's a very good point. I remember the implicit `ulong` => `uint` conversion in D1 as well. That was a fun deployment.

source/agora/common/Transaction.d
Geod24: I don't mind packed when things are self explanatory, but I don't mind the other way either.

AndrejMitrovic: I'm ok with packing if there's no comments in-between. Because the comments and the fields they comment should be grouped together. And you do that with spacing.

```
/// comment1
int field1;

/// comment2
float field2;
```

But I'm totally ok with this:

```
int field1;  /// hello
float field2;  /// hello 2
```

source/agora/common/Transaction.d
Geod24: Yeah `an` / `a` is one of the most common mistake I do :'(

source/agora/common/Transaction.d
Geod24: We do. Oups.

Geod24:
Updated according to review. Also added protection attributes.

source/agora/common/Transaction.d
Geod24: I think that's pretty standard Bitcoin jargon.

AndrejMitrovic: I still don't understand though, what does it mean in this context? I know about the "6 confirmations rule", but is that related to this at all?

Geod24: Confirmed == Included in a block

AndrejMitrovic: got it!

AndrejMitrovic:
@TrustHenry / @MukeunKim, do you want to review this?

AndrejMitrovic (approved):

PR #58, opened 2019-07-03
test-case for linker bug

AndrejMitrovic:


Geod24:
https://issues.dlang.org/show_bug.cgi?id=20022

PR #60, opened 2019-07-03
Implement transaction verification

MichaelKim20:
Contains tests for transaction verification

### DIFF
--- a/source/agora/common/Transaction.d
+++ b/source/agora/common/Transaction.d
@@ -22,6 +22,8 @@ import agora.common.Data;
 import agora.common.Hash;
 import agora.common.crypto.Key;
 
+import std.algorithm;
+
 
 /*******************************************************************************
 
@@ -149,4 +151,171 @@ public struct Input
         hashPart(this.previous, dg);
         hashPart(this.index, dg);
     }
+}
+
+/*******************************************************************************
+
+    Checks whether the transaction is coinbase
+
+    Params:
+        tx = Type of `Transaction`
+
+    Return:
+        Return true if this transaction is a coinbase
+
+*******************************************************************************/
+
+public bool isCoinbaseTx (Transaction tx)
+{
+    return tx.inputs.length == 1 && tx.inputs[0] == Input.init;
+}
+
+/*******************************************************************************
+
+    Creates a new coinbase transaction
+
+    Params:
+        address = The public key that can redeem this output
+        value = The initial value
+
+    Return:
+        Return coinbase transation
+
+*******************************************************************************/
+
+public Transaction newCoinbaseTX (PublicKey address, Amount value = 0)
+{
+    return Transaction(
+        [Input(Hash.init, 0)],
+        [Output(value, address)]
+    );
+}
+
+/*******************************************************************************
+
+    Get sum of `Input`
+
+    Params:
+        tx = `Transaction`
+        findOutput = delegate for finding `Output`
+
+    Return:
+        Sum of `Input` in the `Transaction`
+
+*******************************************************************************/
+
+public Amount getSumInput (Transaction tx, Output* delegate (Hash hash, size_t index) findOutput)
+{
+    return tx.inputs.map!(a => findOutput(a.previous, a.index)).filter!(a => a !is null).map!(a => a.value).sum();
+}
+
+/*******************************************************************************
+
+    Get sum of `Output`
+
+    Params:
+        tx = `Transaction`
+
+    Return:
+        Sum of `Output` in the `Transaction`
+
+*******************************************************************************/
+
+public long getSumOutput (Transaction tx)
+{
+    return tx.outputs.map!(a => a.value).sum();
+}
+
+/*******************************************************************************
+
+    Get result of transaction verification
+
+    Params:
+        tx = `Transaction`
+        findOutput = delegate for finding `Output`
+
+    Return:
+        Return true if this transaction is verified.
+
+*******************************************************************************/
+
+public bool verifyData (Transaction tx, Output* delegate (Hash hash, size_t index) findOutput)
+{
+    if (tx.inputs.length == 0)
+        return false;
+
+    if (tx.outputs.length == 0)
+        return false;
+
+    if (tx.isCoinbaseTx())
+    {
+        return true;
+    }
+    else
+    {
+        if (tx.getSumOutput() > tx.getSumInput(findOutput))
+            return false;
+        else
+            return true;
+    }
+}
+
+/// verify transaction data
+unittest
+{
+    import std.format;
+
+    Transaction[Hash] storage;
+    KeyPair[] key_pairs = [KeyPair.random, KeyPair.random, KeyPair.random, KeyPair.random];
+
+    // Creates the first transaction.
+    Transaction previousTx = newCoinbaseTX(key_pairs[0].address, 100);
+
+    // Save
+    Hash previousHash = hashFull(previousTx);
+    storage[previousHash] = previousTx;
+
+    // Creates the second transaction.
+    Transaction secondTx = Transaction(
+        [
+            Input(previousHash, 0)
+        ],
+        [
+            Output(50, key_pairs[1].address)
+        ]
+    );
+
+    // delegate for finding `Output`
+    auto findOutput = (Hash hash, size_t index)
+        {
+            if (auto tx = hash in storage)
+            {
+                if (index < tx.outputs.length)
+                {
+                    return &tx.outputs[index];
+                }
+                else
+                {
+                    return null;
+                }
+            }
+            else
+            {
+                return null;
+            }
+        };
+
+    // It is validated. (the sum of `Output` < the sum of `Input`)
+    assert(secondTx.verifyData(findOutput), format("Transaction data is not validated %s", secondTx));
+
+    secondTx.outputs ~= Output(50, key_pairs[2].address);
+
+    // It is validated. (the sum of `Output` == the sum of `Input`)
+    assert(secondTx.verifyData(findOutput), format("Transaction data is not validated %s", secondTx));
+
+    secondTx.outputs ~= Output(50, key_pairs[3].address);
+
+    // It isn't validated. (the sum of `Output` > the sum of `Input`)
+    assert(!secondTx.verifyData(findOutput), format("Transaction data is not validated %s", secondTx));
+
 }


source/agora/test/Transaction.d
```diff
@@ -0,0 +1,264 @@
...
+            tx = `Transaction`
+
+        Return:
+            Return true if this transaction is verified.
+
+    *******************************************************************************/
+
+    public bool verification (const Transaction tx)
+    {
+        if (tx.inputs.length <= 0)
```
AndrejMitrovic: The input and outputs also need to be validated so they're not negative.

MichaelKim20: Yes, it is. I made a mistake.

AndrejMitrovic:
When @Geod24 mentioned that we should have code and tests together, what he meant was that they should be made at the same time and be submitted in the same pull request.

But the verification code itself should live in another module, not in a unittest one. Because the verification code will be used both in tests and within the node itself.

Does that make sense?

MichaelKim20:
> When @Geod24 mentioned that we should have code and tests together, what he meant was that they should be made at the same time and be submitted in the same pull request.
> 
> But the verification code itself should live in another module, not in a unittest one. Because the verification code will be used both in tests and within the node itself.
> 
> Does that make sense?

Yeah, I got it.

MichaelKim20:
Added class TransactionManager.
This class has the ability to create and store transactions.
It can also verify transactions.
When @Geod24  improves the hash function, I will use that hash function.
It will be connected to the recipient of the transaction that @AndrejMitrovic  is working on. So, I will add the transaction so that it can be verified at the receiving end of the transaction.

AndrejMitrovic:
> It will be connected to the recipient of the transaction that @AndrejMitrovic is working on. So, I will add the transaction so that it can be verified at the receiving end of the transaction.

I think this part can be added in a separate pull request later, so your work is not blocked on my work.

So the only blocker right now for both of us is hashing.

MichaelKim20:
> > It will be connected to the recipient of the transaction that @AndrejMitrovic is working on. So, I will add the transaction so that it can be verified at the receiving end of the transaction.
> 
> I think this part can be added in a separate pull request later, so your work is not blocked on my work.
> 
> So the only blocker right now for both of us is hashing.

We need to find 'Output' in 'Block' using 'Input' ('previous' and 'index').
This should also be done later after the two PR's are Merge. What do you think?

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+
+    *******************************************************************************/
+
+    public Output newTxOut (PublicKey address, Amount value = 0)
+    {
+        Output txOut;
+        txOut.address = address;
+        txOut.value = value;
+        return txOut;
+    }
```
Geod24: This does not use any of the class' variable. So it could be made `static` to allow it to be called without an instance. Or it could be made a free function.

However, I think it can be removed. Because you can use:
```d
auto txout = TxOut(address, value);
```

the same way you would use:
```d
auto txout = txmanager.newTxOut(address, value);
```

D generate a default constructor for `struct` which is equivalent to:
```d
this(PublicKey _1 = PublicKey.init, Amount _2 = Amount.init)
```

So you can call `TxOut`, `TxOut(address)` and `TxOut(address, amount)`.

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+            previous = The hash of a previous transaction containing the `Output` to spend
+            index = Index of the `Output` in the `previous` `Transaction`
+            signature = A signature that should be verified using the `previous[index].address` public key
+
+        Return:
+            The input of the transaction
+
+    *******************************************************************************/
+
+    public Input newTxIn (Hash previous, uint index, Signature signature = TransactionManager.nullSignature)
```
Geod24: Same comment as `newTxOut`, this can be removed

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+            tx = Type of `Transaction`
+
+        Return:
+            `Hash` of `Transaction`
+
+    *******************************************************************************/
+
+    public Hash getTxHash (Transaction tx)
+    {
+        Hash hash = sha256Of(serializeToJsonString(tx));
```
Geod24: We should not serialize to JSON here, the struct should be directly hashed. See the hashing primitive PR for an example on how to do this.

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.TransactionManager;
+import agora.common.Data;
+import agora.common.crypto.Key;
+import agora.common.Transaction;
+import std.digest.sha;
+import vibe.data.json : serializeToJsonString;
```
Geod24: There shouldn't be any dependency to Vibe.d here

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+
+*******************************************************************************/
+
+public class TransactionManager
+{
+    /// Store of previously processed transactions
+    private Transaction[Hash] txs;
+
+    /// null value of transaction hash
+    public static Hash nullTxHash;
```
Geod24: I see that this is never initialized. So it always have the default value, which can be represented as `Hash.init`.
It is better to use `Hash.init` instead of `nullTxHash`, as `Hash.init` can be optimized (https://en.wikipedia.org/wiki/Constant_folding) while `nullTxHash` cannot.

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+
+    /// null value of signature
+    public static Signature nullSignature;
```
Geod24: Same comment as `nullTxHash`: Use `Signature.init`

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+    *******************************************************************************/
+
+    public bool isCoinbaseTx (Transaction tx)
+    {
+        return
+            (
+                (tx.inputs.length == 1) &&
+                (tx.inputs[0].previous == TransactionManager.nullTxHash) &&
+                (tx.inputs[0].index == 0) &&
+                (tx.inputs[0].signature == TransactionManager.nullSignature)
```
Geod24: You can simplify this with:
```d
return tx.inputs.length == 1 && tx.inputs[0] == TxIn.init;
```

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+        Params:
+            hash = `Hash` of `Transaction`
+            index = Index of the `Output` in the `Transaction`
+
+        Return:
+            `Output` in the `Transaction`
+
+    *******************************************************************************/
+
+    public Output * findTxOutput (Hash hash, uint index)
```
Geod24: Style nitpick:
```suggestion
    public Output* findTxOutput (Hash hash, uint index)
```

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+    {
+        if (hash in this.txs)
```
Geod24: This is doing a double lookup. The builtin hashmap returns a pointer to the element if it exists, or `null`.
Additionally, you can define a new variable in `if` statement, so the following code:
```d
if (auto tx = hash in this.txs)
{
    // Use `tx`
}
else
{
    // Cannot use `tx` from here, but we know it is `null`
    return null;
}
```

Works just fine.

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,283 @@
...
+
+    public long getInputSum (Transaction tx)
+    {
+        if (isCoinbaseTx(tx))
+            return 0;
+
+        long sum = 0;
+        Output *txOut;
+
+        foreach (idx, ref txIn; tx.inputs)
```
Geod24: I think you should take a look at [`std.algorithm`](https://dlang.org/phobos/std_algorithm.html).
Many of those loop could be turned into calls to `each`, `map` & co.

Geod24 (commented):
First round of review done

MichaelKim20:
> First round of review done

@Geod24 All the requested modifications have been completed.

Geod24:
Great work !

One point though, if a function does not use any member of a class, and its usage is not specific to a class, it should not be part of a class. Other parts of the program might need to get the `sum` of the input. They should not have to have an instance of a `TransactionManager` to do so.

For example, imagine I want to print the sum of all transactions in a debug logging statement in the network code.

Using https://github.com/bpfkorea/agora/blob/740d54a677a209dabcb8fbd1893c8f2950089b37/source/agora/node/Network.d#L300-L306 as a base

```d
    public void sendMessage (Transaction tx) @safe
    {
        scope tm = new TransactionManager();
        foreach (ref node; this.peers)
        {
            logDebug("Sending a transaction with %d input and %d output", tm.getInputSum(tx), tm.getOutputSum(tx));
            node.sendMessage(tx);
        }
    }
```

Instead, if you make it a free function:
```d
public long getOutputSum (Transaction tx)
{
    return tx.outputs.map!(a => a.value).sum();
}
```

I can call it like this:
```d
    public void sendMessage (Transaction tx) @safe
    {
        foreach (ref node; this.peers)
        {
            logDebug("Sending a transaction with %d input and %d output", tx.getInputSum(), tx.getOutputSum());
            node.sendMessage(tx);
        }
    }
```

Which I think is much clearer. It does not have to be a member of `Transaction` because we can use UFCS, as you already did with `map` and `sum`.

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,211 @@
...
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.TransactionManager;
+import agora.common.Data;
+import agora.common.crypto.Key;
+import agora.common.Transaction;
+import agora.common.Hash;
+import  std.algorithm;
```
AndrejMitrovic:
```suggestion

import std.algorithm;
```

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,211 @@
...
+
+        Params:
+            tx = Type of `Transaction`
+
+        Return:
+            `Hash` of `Transaction`
+
+    *******************************************************************************/
+
+    public Hash getTxHash (Transaction tx)
```
AndrejMitrovic: Do you plan on ever overriding and changing what this method does? Because if not, we don't really need it. Unittest code can simply call `hashFull(tx)`.

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,211 @@
...
+        Return:
+            Sum of `Input` in the `Transaction`
+
+    *******************************************************************************/
+    public long getInputSum (Transaction tx)
+    {
+        return
+            tx.inputs
+                .map!(a => findTxOutput(a.previous, a.index))
+                .filter!(a => a != null)
```
AndrejMitrovic:
```suggestion
                .filter!(a => a !is null)
```

@Geod24: wasn't there some warning / deprecation in the past about using `!=` with null?

AndrejMitrovic:
I agree with @Geod24's comments in https://github.com/bpfkorea/agora/pull/60#issuecomment-508912767.

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,212 @@
...
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.TransactionManager;
+import agora.common.Data;
```
Geod24:
```suggestion

import agora.common.Data;
```

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,212 @@
...
+import agora.common.crypto.Key;
+import agora.common.Transaction;
+import agora.common.Hash;
+
+import std.algorithm;
+
+/*******************************************************************************
+
+Class for the transactions verification
```
Geod24:
```suggestion
    Class for the transactions verification
```

source/agora/common/TransactionManager.d
```diff
@@ -0,0 +1,212 @@
...
+
+        Get sum of `Input`
+
+        Params:
+            tx = `Transaction`
+
+        Return:
+            Sum of `Input` in the `Transaction`
+
+    *******************************************************************************/
```
Geod24:
```suggestion
    *******************************************************************************/
    
```

source/agora/common/TransactionManager.d
MukeunKim: > Great work !
> 
> One point though, if a function does not use any member of a class, and its usage is not specific to a class, it should not be part of a class. Other parts of the program might need to get the `sum` of the input. They should not have to have an instance of a `TransactionManager` to do so.
> 
> For example, imagine I want to print the sum of all transactions in a debug logging statement in the network code.
> 
> Using
> 
> https://github.com/bpfkorea/agora/blob/740d54a677a209dabcb8fbd1893c8f2950089b37/source/agora/node/Network.d#L300-L306
> 
> as a base
> ```d
>     public void sendMessage (Transaction tx) @safe
>     {
>         scope tm = new TransactionManager();
>         foreach (ref node; this.peers)
>         {
>             logDebug("Sending a transaction with %d input and %d output", tm.getInputSum(tx), tm.getOutputSum(tx));
>             node.sendMessage(tx);
>         }
>     }
> ```
> 
> Instead, if you make it a free function:
> 
> ```d
> public long getOutputSum (Transaction tx)
> {
>     return tx.outputs.map!(a => a.value).sum();
> }
> ```
> 
> I can call it like this:
> 
> ```d
>     public void sendMessage (Transaction tx) @safe
>     {
>         foreach (ref node; this.peers)
>         {
>             logDebug("Sending a transaction with %d input and %d output", tx.getInputSum(), tx.getOutputSum());
>             node.sendMessage(tx);
>         }
>     }
> ```
> 
> Which I think is much clearer. It does not have to be a member of `Transaction` because we can use UFCS, as you already did with `map` and `sum`.

The function is dependent on the internal data structure.
Please check in detail to see if you can do this.
Wouldn't it be better to make improvements after the blocks are made in the future?

MichaelKim20: Modified~

source/agora/common/Transaction.d
```diff
@@ -21,6 +21,7 @@ module agora.common.Transaction;
 import agora.common.Data;
 import agora.common.Hash;
 import agora.common.crypto.Key;
+import std.algorithm;
```
Geod24, line 25:
```suggestion

import std.algorithm;
```

source/agora/common/Transaction.d
```diff
@@ -150,3 +151,112 @@ public struct Input
...
+
+    Params:
+        tx = Type of `Transaction`
+
+    Return:
+        Return true if this transaction is a coinbase
+
+*******************************************************************************/
+
+bool isCoinbaseTx (Transaction tx)
```
```suggestion
public bool isCoinbaseTx (Transaction tx)
```

source/agora/common/Transaction.d
```diff
@@ -150,3 +151,112 @@ public struct Input
...
+    Params:
+        address = The public key that can redeem this output
+        value = The initial value
+
+    Return:
+        Return coinbase transation
+
+*******************************************************************************/
+
+Transaction newCoinbaseTX (PublicKey address, Amount value = 0)
```
```suggestion
public Transaction newCoinbaseTX (PublicKey address, Amount value = 0)
```

source/agora/common/Transaction.d
```diff
@@ -150,3 +151,112 @@ public struct Input
...
+
+    Params:
+        tx = `Transaction`
+        findOutput = delegate for finding `Output`
+
+    Return:
+        Sum of `Input` in the `Transaction`
+
+*******************************************************************************/
+long getSumInput (Transaction tx, Output* delegate (Hash hash, uint index) findOutput)
```
```suggestion

public Amount getSumInput (Transaction tx, Output* delegate(Hash hash, size_t index) findOutput)
```

Changes:
- `public`
- No space after `delegate` keyword
- Indexes are always `size_t`

source/agora/common/Transaction.d
```diff
@@ -150,3 +151,112 @@ public struct Input
...
+
+    Params:
+        tx = `Transaction`
+
+    Return:
+        Sum of `Output` in the `Transaction`
+
+*******************************************************************************/
+
+long getSumOutput (Transaction tx)
```
```suggestion
public long getSumOutput (Transaction tx)
```

source/agora/common/Transaction.d
```diff
@@ -150,3 +151,112 @@ public struct Input
...
+    Params:
+        tx = `Transaction`
+        findOutput = delegate for finding `Output`
+
+    Return:
+        Return true if this transaction is verified.
+
+*******************************************************************************/
+
+public bool verifyData (Transaction tx, Output* delegate (Hash hash, uint index) findOutput)
```
```suggestion
public bool verifyData (Transaction tx, Output* delegate (Hash hash, size_t index) findOutput)
```

source/agora/common/Transaction.d
```diff
@@ -150,3 +151,112 @@ public struct Input
...
+        else
+            return true;
+    }
+    else
+    {
+        if (tx.inputs.length == 0)
+            return false;
+
+        if (tx.outputs.length == 0)
+            return false;
```
Geod24: Move this `if` before `if (tx.isCoinbaseTx())` ?

source/agora/test/Transaction.d
```diff
@@ -0,0 +1,167 @@
...
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.TransactionManager;
```
Geod24: Those unittests don't use the network. They are good though, and should be moved to the `TransactionManager` code. They are "simple" `unittest` as opposed to ["Network unittests"](https://github.com/bpfkorea/agora/pull/74/files#diff-f69cd4b5588d09bb541f85596fdacad3R193).

Geod24 (commented):
Pretty good, a few comments, mostly style

source/agora/common/TransactionManager.d
MichaelKim20: This file is deleted, moved to source/agora/common/Transaction.d

source/agora/common/TransactionManager.d
MichaelKim20: This file is deleted, moved to source/agora/common/Transaction.d

source/agora/common/TransactionManager.d
MichaelKim20: This file is deleted, moved to source/agora/common/Transaction.d

source/agora/common/Transaction.d
MichaelKim20: Modified

source/agora/test/Transaction.d
MichaelKim20: Modified

source/agora/common/Transaction.d
MichaelKim20: Modified

Geod24 (approved):

MichaelKim20:
@Geod24 @AndrejMitrovic
I have organized it into a single file(source/agora/common/Transaction.d).

PR #63, opened 2019-07-03
Add Homebrew repository to the cache

Geod24:
It means our cache gets invalidated much more frequently,
but fetching those commits is actually what takes time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,6 +17,7 @@ os:
 cache:
   directories:
     - $HOME/Library/Caches/Homebrew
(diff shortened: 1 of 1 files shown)


AndrejMitrovic (approved):

Geod24:
Self merging since it has been approved and it doesn't touch actual code (and I need to test stuff)

Geod24:
Given what I'm seeing on #64 , this behaves as expected!
It does remove a few minutes from our build time. I think I can improve it further, will look into it tomorrow.

PR #65, opened 2019-07-03
Add some pure/nothrow/`@safe`/`@nogc` and remove class-level @trusted

Geod24:
Because ain't nobody going to trust your code if you use `@trusted`!

### DIFF
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -40,8 +40,6 @@ import vibe.web.rest;
 *******************************************************************************/
 public class Node (Network) : API
 {
-@trusted:
-
     /// Config instance
     private const Config config;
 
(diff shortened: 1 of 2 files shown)


source/agora/node/Node.d
```diff
@@ -40,8 +40,6 @@ import vibe.web.rest;
 *******************************************************************************/
 public class Node (Network) : API
 {
-@trusted:
```
AndrejMitrovic, line 43: Thank you. I hated this.

AndrejMitrovic (approved):

AndrejMitrovic:
Is there a trusted-equivalent for `pure`? Because these changes severely limit which code I can call.

For example, in testing code I added throwaway hashing code (using libsodium), changed the Set implementation to use `hash()` under-the-hood. But now I can't use it here because everything is pure.

I just feel like adding these attributes everywhere is going to slow down development time.. just my 2 cents.

Geod24:
That is true. I am not fan of those attributes TBH, I think they compose badly. But `@safe` is a requirement of the REST generator, which I think is very limiting. I think `libsodium` is not "attributed" enough and I'll work on that soon.

PR #67, opened 2019-07-04
Implement basic Ledger class which creates Blocks

AndrejMitrovic:


source/agora/node/GossipProtocol.d
```diff
@@ -61,6 +66,7 @@ public class GossipProtocol
 
         this.tx_cache[tx_hash] = tx;
         this.network.sendTransaction(tx);
+        this.ledger.receiveTransaction(tx);
```
Geod24, line 69: Local storage before network communication

source/agora/common/Block.d
```diff
@@ -55,3 +56,51 @@ public struct Block
...
+    Returns:
+        the genesis block
+
+*******************************************************************************/
+
+public Block getGenesisBlock ( )
+{
+    Block block;
+
+    block.header.height = 0;  // genesis block is 0
```
Geod24: That's the default?

Geod24: However, there should be a transaction giving 40k to a `random()` keypair.

Geod24:
Looks pretty neat

MichaelKim20 (approved):

source/agora/node/GossipProtocol.d
AndrejMitrovic: Yeah I was wondering about that actually. Could we add it to the style guide?

source/agora/common/Block.d
AndrejMitrovic: Makes sense.

AndrejMitrovic: Speaking of which, we need to define the unitary model or just a helper function (so 1 BOA = 1,000,000 units or whatever we defined it as). So I could do something like `output.amount = 1.boa` and `boa` is defined as `long boa (long amount) { return amount * 1_000_000; }`.

That could be a separate task in the backlog.

Geod24: Correct

AndrejMitrovic:
Rebased and reworked it a little bit.

The unittests could perhaps use some refinement.

source/agora/common/Block.d
```diff
@@ -95,3 +95,78 @@ public struct Block
     ///
     public Transaction tx;
 }
+
+/*******************************************************************************
+
+    Creates the genesis block.
+    Note: The output specifies a randomly-generated address,
```
Geod24:
```suggestion

    The output specifies a randomly-generated address,
```

Otherwise you create a DDOC section

source/agora/common/Block.d
```diff
@@ -95,3 +95,78 @@ public struct Block
...
+    later to be replaced with proper outputs.
+
+    Returns:
+        the genesis block
+
+*******************************************************************************/
+
+public Block getGenesisBlock ( )
```
Geod24:
```suggestion
public Block getGenesisBlock ()
```

source/agora/common/Block.d
```diff
@@ -95,3 +95,78 @@ public struct Block
...
+{
+    import agora.common.crypto.Key;
+
+    // note: not using random() as the genesis block should always be the same
+    auto address = KeyPair.fromSeed(
+        Seed.fromString("SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4")).address;
```
Geod24, line 116: I should make this CTFE-able so we can just use `static immutable`

source/agora/common/Block.d
```diff
@@ -95,3 +95,78 @@ public struct Block
...
+
+/*******************************************************************************
+
+    Create a new block, referencing the provided previous block.
+
+    Params:
+        prev_block = the previous block
+        tx = the transaction that will be contained in the new block
+
+    In the current preliminary design a block contains a single transaction.
```
Geod24: Move this before `Params`

source/agora/common/Block.d
```diff
@@ -95,3 +95,78 @@ public struct Block
...
+
+*******************************************************************************/
+
+public Block makeNewBlock ( Block prev_block, Transaction tx ) @nogc nothrow @safe
```
Geod24:
```suggestion
public Block makeNewBlock (Block prev_block, Transaction tx) @nogc nothrow @safe
```

source/agora/node/Ledger.d
```diff
@@ -0,0 +1,99 @@
...
+
+module agora.node.Ledger;
+
+import agora.common.Block;
+import agora.common.Data;
+import agora.common.Hash;
+import agora.common.Transaction;
+
+/// Ditto
+class Ledger
```
```suggestion
public class Ledger
```

source/agora/node/Ledger.d
```diff
@@ -0,0 +1,99 @@
...
+{
+    /// data storage for all the blocks
+    private Block[Hash] ledger;
+
+    /// latest block height
+    private Block last_block;
```
Geod24: Pointer ?

source/agora/node/Ledger.d
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the highest block
+
+    ***************************************************************************/
+
+    public Block getLastBlock ( ) @safe nothrow @nogc
```
```suggestion
    public Block getLastBlock () @safe nothrow @nogc
```

source/agora/node/Ledger.d
```diff
@@ -0,0 +1,99 @@
...
+    /***************************************************************************
+
+        Add a block to the ledger
+
+        Params:
+            block = the block to add
+
+    ***************************************************************************/
+
+    private void addNewBlock ( Block block ) @safe nothrow
```
```suggestion
    private void addNewBlock (Block block) @safe nothrow
```

source/agora/node/Ledger.d
```diff
@@ -0,0 +1,99 @@
...
+{
+    import agora.common.crypto.Key;
+
+    scope ledger = new Ledger;
+    assert(ledger.getLastBlock() == getGenesisBlock());
+
+    auto rand_pair = KeyPair.random();
+    auto input = Input(hashFull(42));
+    input.signature = rand_pair.secret.sign(input.hashFull[]);
+    Transaction tx = { inputs : [input], outputs: [Output(100, rand_pair.address)]};
```
```suggestion
    Transaction tx = { inputs : [input], outputs: [ Output(100, rand_pair.address) ] };
```

source/agora/node/Ledger.d
```diff
@@ -0,0 +1,99 @@
...
+
+        Creates a new block containing the transaction,
+        and adds the block to the ledger.
+
+        Params:
+            tx = the received transaction
+
+    ***************************************************************************/
+
+    public void receiveTransaction (Transaction tx) @trusted
```
Geod24, line 50: That's very network-oriented. What about loading from disk ? Will the name stick ?

AndrejMitrovic: Hmm I mean I could name it `addTransaction`, but I was worried about this name because the transaction might be invalid, therefore it will **not** be added, despite the name.

So what do we call it?

source/agora/common/Block.d
AndrejMitrovic: Ah, makes it easier to grep? Add it to the style guide!

source/agora/common/Block.d
AndrejMitrovic: That node is outdated anyway. :D

source/agora/common/Block.d
AndrejMitrovic: yessss please.

source/agora/node/Ledger.d
AndrejMitrovic: That's an extra dereference I guess, but we're fine with it?

source/agora/common/Block.d
Geod24: It already is there: https://github.com/bpfkorea/agora/pull/74/files#diff-f69cd4b5588d09bb541f85596fdacad3R57

source/agora/node/Ledger.d
Geod24: Damn hard questions. We can go with this now and change later.

source/agora/node/Ledger.d
Geod24: I take a dereference over a copy any time of the day.

AndrejMitrovic:
Updated PR with the suggested changes.

source/agora/node/Ledger.d
Geod24: Pin g

AndrejMitrovic:
Missed out the public thing. Done.

source/agora/node/Ledger.d
```diff
@@ -0,0 +1,101 @@
...
+        Params:
+            block = the block to add
+
+    ***************************************************************************/
+
+    private void addNewBlock (Block block) @safe nothrow
+    {
+        auto block_hash = block.header.hashFull();
+        this.ledger[block_hash] = block;
+        this.last_block = block_hash in this.ledger;
```
Geod24, line 81: This is a very crude approach but will work for now :)

Geod24 (approved):

PR #68, opened 2019-07-05
Add hashing primitives

Geod24:
```
Those hashing primitives use `BLAKE2b` for hashing.
They allow struct to have control over the way hashing is done.
The results are currently not tested against other implementation,
but they are stable and will unblock the team.
```

We need a few fixes in `libsodiumd` for attributes to fully work, but that should unblock everyone.

### DIFF
--- a/source/agora/common/Hash.d
+++ b/source/agora/common/Hash.d
@@ -1,0 +1,200 @@
+/*******************************************************************************
+
+    Function definition and helper related to hashing
+
+    The actual definition of the `Hash` type is in `agora.common.Data`,
+    as parts of the system might pass along `Hash` without having to know
+    what / how they were produced.
+
+    However, this module expose functionalities for modules that want to do
+    hashing. The interface is designed so that this module knows about the
+    hash types, and aggregates implementing the interface only deal with their
+    members.
+
+    Due to a language limitation (one can't overload based on return value),
+    this module expose two main functions:
+    - `hashFull(T)`, which returns a `Hash`
+    - `hashPart(T, HasherDg)` which does not return anything and should be used
+      to accumulate data to hash.
+
+    For safety reason, a data structure currently need to explicitly define
+    a `hash` function to support hashing.
+    This limitation might be lifted in the future, but a few things
+    need to be taken into account:
+    - Alignment / packing: We can't have unaligned / unpacked structures
+      as it would create malleability issues
+    - Non-public members: We can't deal with anything non-public (we can use
+      `.tupleof` but it's impossible to tell the usage of the member,
+      e.g. it could be a cache of the hash)
+    - Indirections / references types
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.Hash;
+
+import agora.common.Data;
+
+import libsodium;
+
+
+///
+nothrow @nogc @safe unittest
+{
+    static struct Struct
+    {
+        ulong foo;
+        string bar;
+
+        void computeHash (scope HashDg dg) const nothrow @safe @nogc
+        {
+            // We can hash in any order we want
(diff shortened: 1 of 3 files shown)


source/agora/common/Hash.d
```diff
@@ -0,0 +1,159 @@
...
+
+module agora.common.Hash;
+
+import agora.common.Data;
+
+import libsodium;
+
+
+///
+nothrow @nogc @safe unittest
```
AndrejMitrovic, line 48: It's a little weird to put the unittest first and then the implementation?

Geod24: Module documentation

source/agora/common/Hash.d
```diff
@@ -0,0 +1,159 @@
...
+    crypto_generichash_init(&state, null, 0, Hash.sizeof);
+    scope HashDg dg = (scope const(ubyte)[] data) @trusted
+                     => cast(void)crypto_generichash_update(&state, data.ptr, data.length);
+    hashPart(record, dg);
+    crypto_generichash_final(&state, hash[].ptr, Hash.sizeof);
+    return hash;
+}
+
+/// Ditto
+public void hashPart (T) (scope const auto ref T record, scope HashDg state)
```
AndrejMitrovic, line 113: Shouldn't these be private?

source/agora/common/Hash.d
```diff
@@ -0,0 +1,159 @@
...
+*******************************************************************************/
+
+public Hash hashFull (T) (scope const auto ref T record)
+    nothrow @nogc @trusted
+{
+    Hash hash = void;
+    crypto_generichash_state state;
+    crypto_generichash_init(&state, null, 0, Hash.sizeof);
+    scope HashDg dg = (scope const(ubyte)[] data) @trusted
+                     => cast(void)crypto_generichash_update(&state, data.ptr, data.length);
```
AndrejMitrovic, line 106: What does `cast(void)` do?

source/agora/common/Hash.d
AndrejMitrovic: Ah no, I get it now. It's called by the struct's implementation of `computeHash`.

source/agora/common/Hash.d
Geod24: Avoid DMD erroring on "Cannot return non-`void` from `void` function"...

AndrejMitrovic: I'm guessing it discards the return value, to help in type inference for the declaration of `dg`.

Geod24: That's a more accurate way to put it, yes :)

AndrejMitrovic:
Can you just move the unittests after the functions? The first thing I want to see is the code when I open the module, not the tests.

Other than that it LGTM and I think it's ready to go.

Geod24:
The very first `unittest` should be documentation. I agree it's not good doc. Will rework.

AndrejMitrovic:
> The very first unittest should be documentation. I agree it's not good doc. Will rework.

But that's the point of `///` unittests, they document the symbol that's defined before the unittest.

AndrejMitrovic:
So the documentation looks good with the unittest in place.

AndrejMitrovic (approved):

AndrejMitrovic:
@MukeunKim @TrustHenry want to review this?

source/agora/common/Hash.d
```diff
@@ -0,0 +1,200 @@
...
+    if (is(T == struct))
+{
+    static assert(is(typeof(T.init.computeHash(HashDg.init))),
+                  "Struct `" ~ T.stringof ~
+                  "` does not implement `computeHash(scope HasherDg)` function");
+    record.computeHash(state);
+}
+
+/// Ditto
+public void hashPart () (scope const auto ref Hash record, scope HashDg state)
```
AndrejMitrovic, line 124: What's the extra `()` for?

Geod24: DMD complains otherwise (`auto ref`)

PR #69, opened 2019-07-05
Tx signature verification unittest

TrustHenry:
This creates a new transaction and signs it as a publickey
of the previous transaction to create and validate the input.
Required PR : Implement transaction verification #60

### DIFF
--- a/source/agora/test/GossipProtocol.d
+++ b/source/agora/test/GossipProtocol.d
@@ -42,3 +42,36 @@ unittest
         assert(node.hasMessage(h1));
     }
 }
+
+/// This creates a new transaction and signs it as a publickey
+/// of the previous transaction to create and validate the input.
+unittest
+{
+    TransactionManager manager = new TransactionManager();
+
+    immutable(KeyPair)[] key_pairs;
+    key_pairs ~= KeyPair.random;
+    key_pairs ~= KeyPair.random;
+
+    // Creates the first transaction.
+    Transaction tx0 = manager.newCoinbaseTX(key_pairs[0].address, 100);
+    Hash h0 = manager.getTxHash(tx0);
+    manager.saveTransaction(h0, tx0);
+
+    // Creates the second transaction.
+    Transaction tx1;
+    Input input = manager.newTxIn(h0, 0 );
+    Hash inputHash = sha256Of(serializeToJsonString(input));
+
+    // Signs at the previous hash value.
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
There's a few issues with this PR:

- The test code should not live here. It should live in `agora.common.Transaction` because it is self-contained and doesn't require networking code.
- It does not need the "TransactionManager"
- It doesn't seem to use the public key in the `outputs` array for testing the signature, instead it directly uses the key in the key_pairs array (`PublicKey pubkey = key_pairs[0].address; assert(pubkey.verify(input.signature, inputHash[]));`).

So I suggest moving unittest code to `agora.common.Transaction`, and writing something similar to:

```D
/// This creates a new transaction and signs it as a publickey
/// of the previous transaction to create and validate the input.
unittest
{
    immutable(KeyPair)[] key_pairs;
    key_pairs ~= KeyPair.random();
    key_pairs ~= KeyPair.random();

    // first transaction (coinbase to first public key)
    Transaction tx0 =
    {
        inputs  : [Input(Hash.init, 0)],
        outputs : [Output(40_000, key_pairs[0].address)]
    };

    Input input =
    {
        previous : hashFull(tx0),  // hash of previous utxo
        index : 0  // first item in the outputs array
    };

    // Signs the previous hash value.
    Signature signature = key_pairs[0].secret.sign(hashFull(input)[]);
    input.signature = signature;

    Output output =
    {
        value : 20_000,
        address : key_pairs[1].address,
    };

    // second transaction (spends the utxo from the first transaction)
    Transaction tx1 =
    {
        inputs : [input],
        outputs : [output]
    };

    // Verify the transaction's only output
    assert(tx0.outputs[0].address.verify(
        tx1.inputs[0].signature,
        hashFull(tx1.inputs[0])[]));
}
```

@Geod24 please verify my logic in this unittest. But I think it should be reasonable..

HenryXtremer:
@AndrejMitrovic `agora.common.Transaction` You're talking about me adding it to this file?

TrustHenry:
@AndrejMitrovic `agora.common.Transaction` You're talking about me adding it to this file?

AndrejMitrovic:
Yes to move the unittest there. The `source/agora/test` folder is only for code that needs networking support.

TrustHenry:
I uploaded it again. I'll close this PR.

PR #72, opened 2019-07-05
Emit only error logging during unittests

AndrejMitrovic:


source/agora/node/Node.d
```diff
@@ -63,6 +63,7 @@ public class Node (Network) : API
         this.gossip = new GossipProtocol(this.network);
         this.exception = new RestException(
             400, Json("The query was incorrect"), string.init, int.init);
+        setLogLevel(config.logging.log_level);
```
Geod24: No!

Geod24 (changes requested):
The `Node` constructor is not the place to do this. `setLogLevel` affects global data and there might be multiple instances of `Node`, two things which are essentially incompatible.

AndrejMitrovic:
hmm.. damn global data. I forgot about that.

AndrejMitrovic:
Updated. I've changed it to set the error level in `makeTestNetwork`, since it appears to be the entry-point to all networking-related unittests.

The calling code can then easily set it to something else via another `setLogLevel` call.

Geod24 (approved):

PR #74, opened 2019-07-06
Add a style guide, with formatting and design approach

Geod24:
I'm sure there is still a lot to spell out, so please send suggestions!

### DIFF
--- a/doc/Style.md
+++ b/doc/Style.md
@@ -1,0 +1,266 @@
+# Agora coding style
+
+This document describes the style used within Agora.
+It covers the style itself, which should be clear and unambiguous,
+as well as more design aspects on how we approach / solve problems.
+
+## Formatting
+
+We mostly follow [the D style](https://dlang.org/dstyle.html).
+
+The following is a list of exceptions:
+
+### Modules
+
+All modules should have a module documentation. If you create a new module,
+just look at how another module is documented.
+
+Module names are CamelCase. While the D style recommend `std.exception`,
+we use `std.Exception`.
+This makes the module more prominent and disambiguate any symbol when
+reading fully qualified names, as the first CamelCase name is the module.
+
+Imports are organized by packages: `agora` first, dependencies second, `std` third,
+`core` last. Within packages, alphabetical order is prefered, with modules
+taking preference over packages, e.g.:
+```
+import agora.foo.Bar;
+import agora.foo.Zyx; // In alphabetical order, this would be last
+import agora.foo.pkg.Bar;
+```
+
+### Explicit, readable code
+
+Visibility attribute (`public`, `protected`, `package`, `private`) is always specified.
+Statement covering a scope without indentation (`private:`) should be avoided.
+This made code much easier to read, and code is read much more than it is written.
+
+When implementing an interface, while `override` is not enforced by the language,
+it is mandatory in our style, as changing the interface might lead to silent
+breakage of implementers without `override` being used.
+It also informs the reader that there is an additional context needed.
+
+Prefer longer, more descriptive names over short names in API (e.g. parameters),
+as it makes the documentation more readable.
+Likewise, when writing unittests for documentation purpose, avoid `auto` and
+`const` without a type spelled out.
+
+When importing outside modules, selective import (e.g. `import std.exception : enforce`)
+are prefered. This does not matter as much for `agora`, as modules tend to
+be smaller, with less dependencies.
+
+Put public symbols first in a module / aggregate whenever possible.
+
+### Functions
+
+Functions have a space between their name and their opening parenthesis
+E.g. a function is: `void foobar ()`, not `void foobar()`.
+With parameters: `void foobar (int p1, void* p2)`.
+Templated functions are written as `void foobar (T) (T arg)`.
+This makes functions easy to `grep` for.
+
+### Aggregates (`class` & `struct`)
+
+The order in which members of an aggregate are ordered is as follow:
+- `alias this`
+- Fields
+- Constructors
+- Destructors
+- Operator overloads
+- `public` functions
+- `protected` functions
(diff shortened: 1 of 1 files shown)


doc/Style.md
```diff
@@ -0,0 +1,248 @@
...
+
+## Formatting
+
+We mostly follow [the D style](https://dlang.org/dstyle.html).
+
+The following is a list of exceptions:
+
+### Modules
+
+All module should have a module documentation. If you create a new module,
```
AndrejMitrovic:
```suggestion
All modules should have a module documentation. If you create a new module,
```

doc/Style.md
```diff
@@ -0,0 +1,248 @@
+# Agora coding style
+
+This document describes the style used in within Agora.
```
AndrejMitrovic:
```suggestion
This document describes the style used within Agora.
```

doc/Style.md
```diff
@@ -0,0 +1,248 @@
...
+Likewise, when writing unittests for documentation purpose, avoid `auto` and
+`const` without a type spelled out.
+
+When importing outside modules, selective import (e.g. `import std.exception : enforce`)
+are prefered. This does not matter as much for `agora`, as modules tend to
+be smaller, with less dependencies.
+
+Put public symbols first in a module / aggregate whenever possible.
+
+### Functions
```
AndrejMitrovic, line 54: There should also be a `Classes / Structs` section mentioning that fields should be placed before any methods.

doc/Style.md
```diff
@@ -0,0 +1,248 @@
...
+
+Functions have a space between their name and their opening parenthesis
+E.g. a function is: `void foobar ()`, not `void foobar()`.
+With parameters: `void foobar (int p1, void* p2)`.
+This makes functions easy to `grep` for.
```
AndrejMitrovic, line 60: Maybe also make a note about templates. Do we prefer `void template (T)(T arg)` or `void template (T) (T arg)`?

doc/Style.md
```diff
@@ -0,0 +1,248 @@
...
+
+### Documentation
+
+Documentation matters. Ideally, all non-private symbols should be documented,
+and private symbols which are fairly complicated should be as well.
+
+`/// Ditto` can be used to document two symbols share the same purpose or are
```
AndrejMitrovic:
```suggestion
`/// Ditto` can be used to document that two symbols share the same purpose or are
```

AndrejMitrovic: Or:

```suggestion
`/// Ditto` can be used to document two symbols which share the same purpose or are
```

doc/Style.md
```diff
@@ -0,0 +1,248 @@
...
+********************************************************************************/
+
+bool foobar (int param);
+
+/// Ditto
+bool foobar (int param, string message);
+```
+It might look heavy on the eyes, but it actually separate the code well enough
+that it is readable.
+Documentation, unlike code, has a strict 80 columns rule
```
AndrejMitrovic, line 117: I thought code has that limit as well?

doc/Style.md
```diff
@@ -0,0 +1,248 @@
...
+///
+unittest
+{
+    const NodeCount = 4;
+    auto network = makeTestNetwork!TestNetworkManager(NetworkTopology.Simple, NodeCount);
+    network.start();
+    assert(network.getDiscoveredNodes().length == NodeCount);
+
+   /// Use API as needed
+   foreach (/*PublicKey*/ key, ref /*API*/ node; network.apis)
```
AndrejMitrovic, line 248: Btw, having these `/* */` in the style example is probably a bad idea. It's better to either explicitly use types, or to not use them at all. Because documentation is rarely updated so this might get outdated.

AndrejMitrovic:
Looks fairly good for starters.

TrustHenry (commented):
I think it's good. Let's do this.
I think we can define what we need later.

doc/Style.md
Geod24: The reason it's in `/* */` is because I'm wary of DMD `foreach` type inference errors...

AndrejMitrovic:
Squash? Or merge as-is?

doc/Style.md
Geod24: It's a soft limit for code

Geod24:
Applied the code suggestions plus the following:
```diff
 diff --git a/doc/Style.md b/doc/Style.md
index 4d3f4e4..55521e6 100644
--- a/doc/Style.md
+++ b/doc/Style.md
@@ -56,8 +56,26 @@ Put public symbols first in a module / aggregate whenever possible.
 Functions have a space between their name and their opening parenthesis
 E.g. a function is: `void foobar ()`, not `void foobar()`.
 With parameters: `void foobar (int p1, void* p2)`.
+Templated functions are written as `void foobar (T) (T arg)`.
 This makes functions easy to `grep` for.

+### Aggregates (`class` & `struct`)
+
+The order in which members of an aggregate are ordered is as follow:
+- Fields
+- Constructors
+- Destructors
+- `alias this`
+- Operator overloads
+- `public` functions
+- `protected` functions
+- `package` functions
+- `private` function
+The intent is to have the interface presented to the world grouped together
+and visible with minimal effort.
+An additional suggestion is to put `override`ing functions first.
+
+
 ### Documentation

 Documentation matters. Ideally, all non-private symbols should be documented,
```

AndrejMitrovic:
I'm not sure about `alias this`. I very often completely miss that it's there when it's sandwiched between methods. And it's an important feature.

I would prefer it either at the very end, or at the beginning..

Geod24:
Made `alias this` first in da list

AndrejMitrovic:
Thanks, LGTM!

AndrejMitrovic (approved):

PR #76, opened 2019-07-08
Implement the getBlocksFrom() method

AndrejMitrovic:


AndrejMitrovic:
The Mac build timed out for some reason. Restarting..

AndrejMitrovic:
Mac + ldc 1.16.0

AndrejMitrovic:
The mac build is green yet github status still shows it as yellow..

source/agora/node/Ledger.d
```diff
@@ -13,11 +13,46 @@
...
 import agora.common.Transaction;
 
+import std.algorithm;
+import std.range;
+
+/// Input range of Blocks
+public struct BlockRange
+{
+    ///
+    private Block[Hash] ledger;
```
Geod24: I think we should move storage to be `Block*[Hash]` and backed by a contiguous list of items, possibly a memory mapped file. Because accessing by height seems to be a very, very common use case, while hash is less common.

AndrejMitrovic: Did you mean `Block*[ulong]` then if you meant accessing by height?

AndrejMitrovic: Well I suppose we should have both Hash and height lookups..

Geod24: Yeah, accessing by height with an AA lookup could be useful as well. The main point is that it needs to be backed by a contiguous storage, otherwise that's never going to scale. Unless you have other suggestions ?

AndrejMitrovic: I'll think up of something.

AndrejMitrovic:
Ok for starters I will split up the getBlockHeight() into another PR so it can be unblocked while the rest of the work is being done on the backing store of the ledger.

Geod24:
So unblocked?

AndrejMitrovic:
What I meant was, to unblock getBlockHeight() from having to wait for this PR. Because this PR needs more work.

AndrejMitrovic:
Fixed the PR. Now it's very simple, and we don't need the hash => block lookup yet, that can be part of a PR that needs it.

source/agora/node/Ledger.d
```diff
@@ -31,6 +33,7 @@ public class Ledger
     /// Ctor
     public this ()
     {
+        this.ledger.reserve(2 ^^ 20);  // reserve ~1 million blocks
```
Geod24: Yeeeeeaaaaaaaaaahhhhhhhhhhhh... About that... I have mixed feelings about allocating 151Mbs like this with the GC, even if it's just a reserve.

Geod24:
> Fixed the PR. Now it's very simple, and we don't need the hash => block lookup yet, that can be part of a PR that needs it.

It would be less than 10 LoCs to do it in this PR, right ?

source/agora/node/Ledger.d
```diff
@@ -65,6 +71,30 @@ public class Ledger
...
+            block_height = the starting block height to begin retrieval from
+            max_blocks   = the maximum blocks to return at once
+
+        Returns:
+            the array of blocks starting from block_height,
+            up to `max_blocks`
+
+    ***************************************************************************/
+
+    public Block[] getBlocksFrom ( ulong block_height, size_t max_blocks )
```
```suggestion
    public Block[] getBlocksFrom (ulong block_height, size_t max_blocks)
```

AndrejMitrovic:
I've removed the reserve call.

source/agora/node/Ledger.d
Geod24: Actually:

```suggestion
    public Block[] getBlocks ( ulong from, size_t max_blocks )
```

Wouldn't this be better ?

source/agora/node/Ledger.d
```diff
@@ -65,6 +70,30 @@ public class Ledger
...
+    {
+        assert(max_blocks > 0);
+
+        if (block_height > this.ledger.length)
```
Geod24, line 92: It should be specified in the doc that it's an inclusive range.

source/agora/node/Ledger.d
```diff
@@ -99,3 +129,71 @@ unittest
...
+    scope ledger = new Ledger;
+    assert(ledger.getLastBlock() == getGenesisBlock());
+    assert(ledger.ledger.length == 1);
+
+    void generateBlocks (size_t count)
+    {
+        foreach (_; 0 .. count)
+        {
+            auto rand_pair = KeyPair.random();
+            auto input = Input(hashFull(42));
```
Geod24, line 148: Use the predicted height here ?

Geod24 (changes requested):
Can you add a Network Test ?

AndrejMitrovic:
> Can you add a Network Test ?

Oh good point!

source/agora/node/Ledger.d
AndrejMitrovic: Inclusive in what regards? It starts at the block height and goes from there. And the ledger always has at least one block (genesis).

source/agora/node/Ledger.d
AndrejMitrovic: > Use the predicted height

Sorry what?

source/agora/node/Ledger.d
Geod24: That the provided `block_height` is included.
So if I want to catch-up, I should send `my_block_height + 1`

AndrejMitrovic:
> Can you add a Network Test ?

Btw should I add a `Ledger` module in /test? I'd hate to jam even more test code into GossipProtocol.d or elsewhere.

source/agora/node/Ledger.d
Geod24: Well it's obv... Nvm
I originally wrote "use the index" to have a bit more randomness.
Then realized that if you call this function twice, you'll have twice the same set of txs.
Hence, use the height at which this tx is supposed to be included.

source/agora/node/Ledger.d
AndrejMitrovic: I mean it literally says that in the docs. `Get the array of blocks starting from the provided block height.` I guess "starting from" is not descriptive enough. I'll add another note.

Geod24:
Go with the way you feel most comfortable with. I expect us to refactor / consolidate tests at some point.

AndrejMitrovic:
Ah wait.. I didn't add the API part, it got lost in another branch lol.

AndrejMitrovic:
Updated. @Geod24 I added a limit to the number of blocks that can be sent at once. I think it's useful to have, and also it should probably be configurable later.

source/agora/common/API.d
```diff
@@ -136,4 +140,26 @@ public interface API
...
+            block_height = the starting block height to begin retrieval from
+            max_blocks   = the maximum blocks to return at once
+
+        Returns:
+            the array of blocks starting from block_height,
+            up to `max_blocks`
+
+    ***************************************************************************/
+
+    public Block[] getBlocksFrom ( ulong block_height, size_t max_blocks );
```
Geod24: Are you for real?

source/agora/common/API.d
```diff
@@ -22,6 +23,9 @@ import vibe.data.json;
 import vibe.web.rest;
 import vibe.http.common;
 
+/// Maximum number of blocks that can be sent in a call to getBlocksFrom()
+enum MaxBatchBlocksSent = 1000;
```
Geod24:
```suggestion
private enum MaxBatchBlocksSent = 1000;
```

But it's an implementation details so it should go in `Node` IMO

source/agora/common/API.d
```diff
@@ -136,4 +140,26 @@ public interface API
     ***************************************************************************/
 
     public ulong getBlockHeight ();
+
+
+
+    /***************************************************************************
+
+        Get the array of blocks starting from the provided block height.
+        The block at `block_height` is included in the array.
```
Geod24, line 146:
```suggestion

        The block at `block_height` is included in the array.
        Note that an implementation is free to return less blocks than asked for.
        It is however required to never return more blocks than asked for.        
```

AndrejMitrovic:
Ok fixed the styling.

source/agora/node/Ledger.d
AndrejMitrovic: > I originally wrote "use the index" to have a bit more randomness.

"Every time I hear about randomness in blockchain.."

source/agora/test/Ledger.d
```diff
@@ -0,0 +1,72 @@
...
+version (unittest):
+
+import agora.common.Block;
+import agora.common.Data;
+import agora.common.Hash;
+import agora.common.Transaction;
+import agora.test.Base;
+
+///
+unittest
```
Geod24, line 25: This is beautiful

source/agora/common/API.d
AndrejMitrovic: Well not really, because clients have to know what the limit is when they request blocks. Otherwise they'll get errors.

source/agora/test/Ledger.d
AndrejMitrovic: _suspicious_

source/agora/common/API.d
Geod24: Just make it return less instead of erroring.

source/agora/test/Ledger.d
```diff
@@ -0,0 +1,72 @@
...
+    assert(blocks[0] == getGenesisBlock());
+
+    // exclude genesis block
+    assert(blocks[1 .. $].map!(block => block.tx).equal(txes[]));
+
+    blocks = node_1.getBlocksFrom(0, 1);
+    assert(blocks.length == 1 && blocks[0] == getGenesisBlock());
+
+    blocks = node_1.getBlocksFrom(100, 1);
+    assert(blocks.length == 1 && blocks[0].tx == txes[$ - 1]);
```
Geod24, line 71: Try to get more blocks than there is
Try to start > the max height

AndrejMitrovic:
Thanks, updated.

source/agora/test/Ledger.d
Geod24: That comment was in the network part :'(

AndrejMitrovic: Huh?

Geod24: You added the tests to the `unittests`, not the network UT.

AndrejMitrovic: See this is why I dislike having modules with the same name..

AndrejMitrovic:
Added the tests to the network unittest.

MichaelKim20 (approved):

AndrejMitrovic:
@Geod24 anything else?

Geod24 (approved):

PR #77, opened 2019-07-08
Ledger: Get rid of double-lookup in addNewBlock

AndrejMitrovic:
The require() method returns a reference to the value.

However using it means we lose safety and the method
is no longer nothrow.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -74,11 +74,11 @@ public class Ledger
 
     ***************************************************************************/
 
-    private void addNewBlock (Block block) @safe nothrow
+    private void addNewBlock (Block block) @trusted nothrow
     {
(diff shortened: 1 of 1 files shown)


Geod24 (approved):

Geod24:
By the way if a method never throw, you can do:
```d
void meth () nothrow
{
    scope (exit) assert(0);
    // Whatever you want
}

AndrejMitrovic:
Wait was the previous comment a joke, or you meant scope(failure)?

Geod24:
I got confused with `@attribute(noreturn)` :D

AndrejMitrovic:
Ok not a bad tip, I ammended the PR.

Geod24:
I never give bad tips.

PR #80, opened 2019-07-09
Implement catch-up phase

AndrejMitrovic:
It's not ready for merging yet, but it can be looked at.

This still needs a bit of refactoring, separating atomic changes into their own commits.

### DIFF
--- a/source/agora/node/Network.d
+++ b/source/agora/node/Network.d
@@ -24,11 +24,13 @@ module agora.node.Network;
 
 import agora.common.API;
 import agora.common.crypto.Key;
+import agora.common.Block;
 import agora.common.Config;
 import agora.common.Data;
 import agora.common.Metadata;
 import agora.common.Set;
 import agora.common.Transaction;
+import agora.node.Ledger;
 import agora.node.RemoteNode;
 
 import vibe.core.log;
@@ -41,6 +43,7 @@ import std.array;
 import std.exception;
 import std.format;
 import std.random;
+import std.typecons;
 
 
 /// Ditto
@@ -135,6 +138,72 @@ public class NetworkManager
                 this.wait(this.node_config.retry_delay.msecs);
             }
         });
+    }
+
+    /***************************************************************************
+
+        Periodically retrieve the latest blocks and apply them to the
+        provided ledger.
+
+        Params:
+            ledger = the Ledger to apply received blocks to
+
+    ***************************************************************************/
+
+    public void retrieveLatestBlocks (Ledger ledger)
+    {
+        this.runTask(
+        ()
+        {
+            // periodic task
+            while (1)
+            {
+                this.getBlocksFrom(
+                    ledger.getLastBlock().header.height + 1,
+                    (blocks) { blocks.each!(block => ledger.addNewBlock(block)); });
+
+                this.wait(2.seconds);
+            }
+        });
+    }
+
+    /***************************************************************************
+
+        Retrieve blocks starting from block_height up to the highest block
+        that's available from the connected nodes.
+
+        Params:
+            block_height = the starting block height to begin retrieval from
+
+    ***************************************************************************/
+
+    private void getBlocksFrom (ulong block_height,
+        scope void delegate(Block[]) @safe onReceivedBlocks)
+    {
+        auto node_pair = this.peers.byValue
+            .map!(node => tuple(node.getBlockHeight(), node))
+            .filter!(pair => pair[0] != ulong.max)  // request failed
+            .reduce!((a, b) => a[0] > b[0] ? a : b);
+
+        const highest_block = node_pair[0];
+        if (highest_block < block_height)
+            return;  // we're up to date
+
+        logInfo("Retrieving latest blocks..");
+        auto node = node_pair[1];
+        const MaxBlocks = 1024;
+
+        do
+        {
+            // todo: if any block fails verification, we have to try another node
+            auto blocks = node.getBlocksFrom(block_height, MaxBlocks);
+            logInfo("Received blocks [%s..%s] out of %s..", block_height,
+                block_height + blocks.length, highest_block + 1);  // genesis block
+
+            onReceivedBlocks(blocks);
+            block_height += blocks.length;
+        }
+        while (block_height < highest_block);
     }
 
     /// Dump the metadata
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -89,6 +89,8 @@ public class Node (Network) : API
         logInfo("Doing network discovery..");
         this.network.discover();
 
+        this.network.retrieveLatestBlocks(this.ledger);
+
         logInfo("Dumping metadata..");
         this.network.dumpMetadata();
     }
--- a/source/agora/node/RemoteNode.d
+++ b/source/agora/node/RemoteNode.d
@@ -170,4 +170,25 @@ class RemoteNode
             logError(ex.message);
         }
     }
+
+    /***************************************************************************
+
+        Returns:
+            the height of the node's ledger,
+            or ulong.max if the request failed
+
+    ***************************************************************************/
+
+    public ulong getBlockHeight ()
+    {
+        try
+        {
+            return this.api.getBlockHeight();
+        }
+        catch (Exception ex)
+        {
+            logError(ex.message);
+            return ulong.max;
+        }
+    }
 }
--- a/source/agora/node/RemoteNode.d
+++ b/source/agora/node/RemoteNode.d
@@ -14,6 +14,7 @@
 module agora.node.RemoteNode;
 
 import agora.common.API;
+import agora.common.Block;
 import agora.common.Config;
 import agora.common.crypto.Key;
 import agora.common.Data;
@@ -189,6 +190,36 @@ class RemoteNode
         {
             logError(ex.message);
             return ulong.max;
+        }
+    }
+
+    /***************************************************************************
+
+        Get the array of blocks starting from the provided block height.
+        The block at block_height is included in the array.
+
+        Params:
+            block_height = the starting block height to begin retrieval from
+            max_blocks   = the maximum blocks to return at once
+
+        Returns:
+            the array of blocks starting from block_height,
+            up to `max_blocks`.
+
+            If the request failed, returns an empty array
+
+    ***************************************************************************/
+
+    public Block[] getBlocksFrom (ulong block_height, size_t max_blocks)
+    {
+        try
+        {
+            return this.api.getBlocksFrom(block_height, max_blocks);
+        }
+        catch (Exception ex)
+        {
+            logError(ex.message);
+            return null;
         }
     }
 }
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -31,6 +31,7 @@ import agora.common.Metadata;
 import agora.common.Set;
 import agora.common.Transaction;
 import agora.common.crypto.Key;
+import agora.node.Ledger;
 import agora.node.Network;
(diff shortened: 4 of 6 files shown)


AndrejMitrovic:
Also this only implements catch-up during booting up. The other definitions of done in the Catchup Phase Issue need to be solved.

source/agora/test/Ledger.d
```diff
@@ -76,3 +77,47 @@ unittest
...
     assert(node_1.getBlocksFrom(1000, 10).length == 0);
 }
+
+/// test catch-up phase
+unittest
+{
+    import std.algorithm;
+    import std.conv;
+    import std.format;
+    import std.digest.sha;
```
Geod24: Should be removed

AndrejMitrovic: damn where did that sneak in from lol

AndrejMitrovic:
I've atomic'ed (atomized?) the commits, it should look much nicer now.

I probably need to think of edge cases to add to the unittests.

source/agora/node/RemoteNode.d
```diff
@@ -170,4 +170,25 @@ class RemoteNode
...
+    public ulong getBlockHeight ( )
+    {
+        try
+        {
+            return this.api.getBlockHeight();
+        }
+        catch (Exception ex)
+        {
+            logError(ex.message);
+            return 0;
```
Geod24: `ulong.max`

source/agora/test/Base.d
```diff
@@ -183,10 +188,14 @@ public final class TestNode (Net) : Node!Net, TestAPI
...
     MemMetadata mem_metadata;
 
+    /// The ledger used for the tests
+    Ledger ledger;
+
     ///
     public this (Config config)
     {
         this.mem_metadata = new MemMetadata;
+        this.ledger = new Ledger;
```
Geod24: That code doesn't make sense to me. You instantiate a new `Ledger` just to override it with itself in the parent constructor. You also shadow the parent `ledger` variable ?

AndrejMitrovic: The shadowing was an oversight. I'll simplify this a bit tomorrow.

source/agora/node/RemoteNode.d
AndrejMitrovic: Good point, block height is zero for genesis blocks. I'll add a test-case. Also like I said before I dislike this class, but I plan on making it better at a later point. There's no retry mechanism for example.

AndrejMitrovic:
Update: Added periodic catch-up phase, and added a network test for it. This completes all of the tasks and should cover everything in the definition of done.

source/agora/test/Base.d
```diff
@@ -175,6 +177,9 @@ public interface TestAPI : API
 
     ///
     public abstract void metaAddPeer(string peer);
+
+    ///
+    public abstract void addBlockToLedger (Block block);
```
Geod24: Why do you need this ? I don't see an override anywhere ?

source/agora/test/Ledger.d
```diff
@@ -118,3 +118,48 @@ unittest
         assert(node.getBlockHeight() == 100);
     }
 }
+
+/// test catch-up phase after initial booting (periodic catch-up)
+unittest
+{
+    import core.thread;
+    import std.conv;
+    import vibe.core.log;
```
Geod24: Two leftover imports

source/agora/node/RemoteNode.d
Geod24: Ping

Geod24 (changes requested):
I would merge the last two commits. Also, do you want the ability to arbitrarily delay a node ? Because I can give you that power easily in `LocalRest`.

source/agora/test/Base.d
AndrejMitrovic: Well, the question is how do you get access to the Ledger from the unittest. As it's required right now. edit: reworded

Geod24: You never access the internals of a node at the moment.

AndrejMitrovic: Well the first test is:

Instantiate the nodes. Don't start networking yet. One node (node A) needs to have 100 blocks, the 3 are empty.

Then, make them connect to each other. The 3 other nodes will retrieve blocks from the 1st node on boot.

But to make this work, I need to add those blocks to node A before it connects to the 3 other nodes.

AndrejMitrovic:
I've squashed the two commits.

I'm thinking about how to get rid of those methods in the TestAPI now.

AndrejMitrovic:
> Also, do you want the ability to arbitrarily delay a node ? Because I can give you that power easily in LocalRest.

Hmm.. that might be very useful.

AndrejMitrovic:
All the Mac builds fail. Hmm..

AndrejMitrovic:
It's a timing issue.

source/agora/test/Base.d
AndrejMitrovic: To be more specific here. I can't just use the `sendTransaction` API here because this will trigger the gossip protocol to send the transaction over the whole network, and then each node will create their own block. So by the time the node tries to retrieve the blocks from the other nodes, it will already have all the blocks.

Delaying the node might not help. I would need some way to temporarily disable the gossip protocol, but then this becomes very messy to have all these special cases for unittesting.

The more features the node has, the more complicated it is to test a specific feature in the node without triggering an avalanche of other features.

AndrejMitrovic:
Ok I made a more stable version of the tests that shouldn't suffer too much from timing issues. I added polling and a timeout of 4 seconds.

AndrejMitrovic:
Very peculiar that macOS dmd 2.085.1 seems to consistently be failing now..

source/agora/test/Base.d
AndrejMitrovic: I will be able to replace this code once LocalRest has the ability to filter requests. Then I can do something like:

- Use the sendTransaction API to one node
- Make the other nodes ignore transaction gossiping
- The other nodes will then catch-up using the periodic `getBlocksFrom` call.

This is way nicer than messing with Node's internals.

AndrejMitrovic:
~@Geod24 I'm having consistent time-outs in the MacOS build: https://travis-ci.com/bpfkorea/agora/jobs/214998228

It seems it gets stuck trying to install DMD & Dub.~

And it's green.

Geod24:
I guess you want to rebase ?

AndrejMitrovic:
> I guess you want to rebase ?

Ah I thought this one was rebased.

AndrejMitrovic:
I mean merged.

AndrejMitrovic:
Rebased.

source/agora/node/Network.d
```diff
@@ -137,6 +140,72 @@ public class NetworkManager
...
+
+    public void retrieveLatestBlocks (Ledger ledger)
+    {
+        this.runTask(
+        ()
+        {
+            // periodic task
+            while (1)
+            {
+                logInfo("Retrieving latest blocks..");
```
Geod24: I think this logging should only happen if we are actually catching up

source/agora/node/RemoteNode.d
```diff
@@ -170,4 +171,55 @@ class RemoteNode
...
+
+    /***************************************************************************
+
+        Returns:
+            the height of the node's ledger,
+            or ulong.max if the request failed
+
+    ***************************************************************************/
+
+    public ulong getBlockHeight ( )
```
Geod24:
```suggestion
    public ulong getBlockHeight ()
```

source/agora/test/Ledger.d
```diff
@@ -22,13 +23,24 @@ import agora.common.Hash;
 import agora.common.Transaction;
 import agora.test.Base;
 
+/// Returns: true if all node's block heights are equal to the provided height
+private bool allHeightsEqual ( TestAPI[] nodes, size_t height )
```
Geod24:
```suggestion
private bool allHeightsEqual (TestAPI[] nodes, size_t height)
```

Geod24 (changes requested):
As discussed IRL: Bad rebase, some code is missing

AndrejMitrovic:
Fixed the suggested changes.

AndrejMitrovic:
Rebased to use new LocalRest sleep() feature.

However it's a bit ugly because now I have to store `RemoteAPI!TestAPI` in `TestNetworkManager`, not just `TestAPI`. And then there's the issue of `RemoteAPI!TestAPI[]` not implicitly converting to `TestAPI[]`.

AndrejMitrovic:
Edit: Pushed a simplification https://github.com/bpfkorea/agora/compare/3c38af6c675d18cab84aa8c89c702d6b3cf8bb7c..bf03c16e9bc1267934450648f55a9f583ffbc9e1

source/agora/node/Network.d
```diff
@@ -137,6 +140,72 @@ public class NetworkManager
...
+        Retrieve blocks starting from block_height up to the highest block
+        that's available from the connected nodes.
+
+        Params:
+            block_height = the starting block height to begin retrieval from
+
+    ***************************************************************************/
+
+    private void getBlocksFrom (ulong block_height,
+        void delegate(Block[]) @safe onReceivedBlocks)
```
Geod24:
```suggestion
        scope void delegate(Block[]) @safe onReceivedBlocks)
```

source/agora/node/Node.d
```diff
@@ -174,4 +176,18 @@ public class Node (Network) : API
...
+        Returns a new Ledger instance. Can be overriden in unittests.
+
+        Returns:
+            a Ledger instance
+
+    ***************************************************************************/
+
+    protected Ledger getLedger ()
+    {
+        return new Ledger();
```
Geod24: I don't think you need this anymore ?

source/agora/test/Ledger.d
```diff
@@ -71,3 +83,79 @@ unittest
...
+        Thread.sleep(100.msecs);
+    }
+}
+
+/// test catch-up phase after initial booting (periodic catch-up)
+unittest
+{
+    import core.thread;
+    import std.algorithm;
+    import std.format;
```
Geod24: Those 2 imports are not necessary anymore

source/agora/test/Ledger.d
```diff
@@ -22,13 +23,24 @@ import agora.common.Hash;
 import agora.common.Transaction;
 import agora.test.Base;
 
+/// Returns: true if all node's block heights are equal to the provided height
+private bool allHeightsEqual (API)(API[] nodes, size_t height)
```
Geod24: Suggestion: `isConsistent` and check the block headers for equality ?

source/agora/test/Ledger.d
AndrejMitrovic: I need `each` still.

AndrejMitrovic:
Updated with suggested changes.

source/agora/test/Ledger.d
AndrejMitrovic: Ok.

source/agora/test/Ledger.d
```diff
@@ -71,3 +83,78 @@ unittest
...
     assert(node_1.getBlocksFrom(1000, 10).length == 0);
 }
+
+/// test catch-up phase during booting
+unittest
+{
+    import core.thread;
+    import std.algorithm;
+    import std.conv;
+    import std.format;
```
Geod24: conv and format superfluous

Geod24 (approved):
LGTM

AndrejMitrovic:
Pushed a small update.

Geod24 (approved):

PR #85, opened 2019-07-10
Disallow negative amounts in the outputs of a transaction

AndrejMitrovic:


source/agora/common/Transaction.d
```diff
@@ -317,5 +327,44 @@ unittest
...
+            }
+            else
+            {
+                return null;
+            }
+        }
+        else
+        {
+            return null;
+        }
```
Geod24: Can we simplify this ?
```d
 if (auto tx = hash in storage)
        if (index < tx.outputs.length)
            return &tx.outputs[index];
return null;
```

AndrejMitrovic:
Fixed styling of both unittests.

Geod24 (approved):

PR #87, opened 2019-07-11
Move transaction verification into the Ledger

AndrejMitrovic:
I've named it `acceptTransaction` rather than `acceptsTransaction` because it's a verb (it has side-effects), and not just a question (do you accept this transaction). 

I wanted to avoid having two separate methods `isTransactionAccepted` and `putTransaction`, because then you could circumvent the check outside the `Ledger` class.

PR #88, opened 2019-07-11
Add some helper functions to simplify writing test code

AndrejMitrovic:
It will also simplify the tests in https://github.com/bpfkorea/agora/pull/87 (but it doesn't have to block it).

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -25,8 +25,10 @@ version (unittest):
 import agora.common.API;
 import agora.common.Config;
 import agora.common.Data;
+import agora.common.Hash;
 import agora.common.Metadata;
 import agora.common.Set;
+import agora.common.Transaction;
 import agora.common.crypto.Key;
 import agora.node.Network;
 import agora.node.Node;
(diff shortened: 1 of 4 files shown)


source/agora/test/Base.d
```diff
@@ -36,6 +39,70 @@ import std.algorithm.iteration;
...
+*******************************************************************************/
+
+public Block[] getChainedBlocks (Block root, Transaction[] txes)
+{
+    Block[] blocks;
+
+    foreach (tx; txes)
+    {
+        root = root.makeNewBlock(tx);
+        blocks ~= root;
```
Geod24: I have a feeling you deleted some code here, because the `root` assignment is useless

AndrejMitrovic: Perhaps it's UFCS that's confusing you.

```D
root = makeNewBlock(root, tx);
```

AndrejMitrovic: I will make it clearer maybe by adding a `temp` block up there?

AndrejMitrovic:
Pushed a little change. Does it make more sense now?

source/agora/test/Base.d
Geod24: `blocks ~= blocks[$-1].makeNewBlock(tx)` ? Ofc you need to handle the first loop, but...

Geod24: `temp` is also good tho

AndrejMitrovic:
Rebased and fixed build failure.

Geod24:
https://dlang.org/dstyle.html#naming_general

Geod24 (approved):
Usually I'd like to see a rule of three, but this is test code, so let's give it a shot :)

PR #89, opened 2019-07-11
Use the new unittest helper routine

AndrejMitrovic:
So this is the missing part, because one PR got rebased before the other.

### DIFF
--- a/source/agora/common/Transaction.d
+++ b/source/agora/common/Transaction.d
@@ -193,6 +193,46 @@ public Transaction newCoinbaseTX (PublicKey address, Amount value = 0)
 
 /*******************************************************************************
 
+    Create a set of chained transactions, where each transaction spends the
+    entire sum of the previous transaction.
+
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
However, I am very unsure about importing into the `test` package in here.. so maybe the routines should be moved somewhere else.

AndrejMitrovic:
Update: Moved helper functionality into their proper location.

Geod24:
Arg... This is an interesting conundrum. Those functions should IMO not be used in production. Let's discuss on Monday.

AndrejMitrovic:
So `getChainedBlocks` was removed. `getChainedTransactions` was moved, but I now versioned it into `version(unittest)`, this should ensure it's never used in production. Is this still ok?

AndrejMitrovic:
Rebased. W.r.t. "rule of three", this function is already used in the networking tests, now it's also being used here.

Geod24 (approved):

PR #90, opened 2019-07-11
Simplify and speed up transaction verification

AndrejMitrovic:
I am not 100% sure about the design, but it gets rid of the double-lookup, and more importantly it makes sure code doesn't accidentally forget to call one of two functions (so it only has to call one function instead).

### DIFF
--- a/source/agora/common/Transaction.d
+++ b/source/agora/common/Transaction.d
@@ -232,7 +232,7 @@ public Amount getSumOutput (Transaction tx) nothrow pure @safe @nogc
 
 /*******************************************************************************
 
-    Get result of transaction data verification
+    Get result of transaction data and signature verification
 
     Params:
         tx = `Transaction`
@@ -243,7 +243,7 @@ public Amount getSumOutput (Transaction tx) nothrow pure @safe @nogc
 
 *******************************************************************************/
 
-public bool verifyData (Transaction tx,
+public bool verify (Transaction tx,
     Output* delegate (Hash hash, size_t index) @safe findOutput) @safe
 {
     if (tx.inputs.length == 0)
@@ -257,37 +257,23 @@ public bool verifyData (Transaction tx,
         if (output.value < 0)
             return false;
 
-    return tx.getSumOutput() <= tx.getSumInput(findOutput);
-}
+    long sum_unspent;
 
-/*******************************************************************************
-
-    Get result of signature verification
-
-    Params:
-        tx = `Transaction` to verify the signature of
-        findOutput = delegate for finding `Output`
-
-    Return:
-        Return true if this signature is verified.
-
-*******************************************************************************/
-
-public bool verifySignature (Transaction tx,
-    Output* delegate (Hash hash, size_t index) @safe findOutput) @safe
-{
-    const Hash txHash = hashFull(tx);
+    const tx_hash = hashFull(tx);
     foreach (input; tx.inputs)
     {
-        if (auto output = findOutput(input.previous, input.index))
-        {
-            if (!output.address.verify(input.signature, txHash[]))
-                return false;
-        }
-        else
+        // all referenced outputs must be present
+        auto output = findOutput(input.previous, input.index);
+        if (output is null)
             return false;
+
+        if (!output.address.verify(input.signature, tx_hash[]))
+            return false;
+
+        sum_unspent += output.value;
     }
-    return true;
+
+    return tx.getSumOutput() <= sum_unspent;
 }
 
 /// verify transaction data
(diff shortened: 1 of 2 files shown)


TrustHenry (approved):

source/agora/node/Ledger.d
```diff
@@ -144,13 +134,14 @@ public class Ledger
...
-    private inout(Transaction)* findTransaction (Hash tx_hash) inout @safe
+    private Output* findOutput (Hash tx_hash, size_t index) @safe
     {
         foreach (ref block; this.ledger)
         {
             if (block.header.tx_hash == tx_hash)
             {
-                return &block.tx;
+                if (index < block.tx.outputs.length)
+                    return &block.tx.outputs[index];
```
Geod24, line 144: What happen if I send a transaction which reference a previous tx but with `index` == `size_t.max` ?

AndrejMitrovic: Watch the next episode of Pokemon to find out!

source/agora/common/Transaction.d
```diff
@@ -243,7 +243,7 @@ public Amount getSumOutput (Transaction tx) nothrow pure @safe @nogc
 
 *******************************************************************************/
 
-public bool verifyData (Transaction tx,
+public bool verify (Transaction tx,
```
Geod24, line 246: I wish `in` meant `scope const ref`... Because `scope const ref`

source/agora/common/Transaction.d
```diff
@@ -257,37 +257,25 @@ public bool verifyData (Transaction tx,
...
 
-    return tx.getSumOutput() <= tx.getSumInput(findOutput);
-}
-
-/*******************************************************************************
+    static struct Pair { Input input; Output* output; }
 
-    Get result of signature verification
+    auto prev_txes = tx.inputs
+        .map!(input => Pair(input, findOutput(input.previous, input.index)));
```
Geod24: Ranges are `lazy` so this actually doesn't speed things up ?

source/agora/common/Transaction.d
```diff
@@ -365,68 +357,7 @@ unittest
...
-
-    Transaction tx2 = Transaction(
-        [
-            Input(tx1Hash, 0)
-        ],
-        [
-            Output(1_000, key_pairs[1].address)
-        ]
-    );
+    tx_2.inputs[0].signature = key_pairs[0].secret.sign(hashFull(tx_2)[]);
```
Geod24, line 358: It's weird to merge the signature verification check into a PR which documentation is:
> /// negative output amounts disallowed

Geod24 (changes requested):
As a bigly man would say: Looks good, doesn't work.

source/agora/common/Transaction.d
```diff
@@ -257,37 +257,25 @@ public bool verifyData (Transaction tx,
...
+        return false;
 
-public bool verifySignature (Transaction tx,
-    Output* delegate (Hash hash, size_t index) @safe findOutput) @safe
-{
-    const Hash txHash = hashFull(tx);
-    foreach (input; tx.inputs)
+    // all signatures must be correct
+    if (!prev_txes.all!(pair =>
+        pair.output.address.verify(pair.input.signature, hashFull(tx)[])))
```
TrustHenry:
```suggestion
    const Hash txHash = hashFull(tx);
    if (!prev_txes.all!(pair =>
        pair.output.address.verify(pair.input.signature, txHash[]))
```

TrustHenry (commented):
having is expensive.

source/agora/common/Transaction.d
AndrejMitrovic: into a unittest you mean. Yeah I'll fix this up.

AndrejMitrovic:
> having is expensive

having what?

AndrejMitrovic:
I know how to fix this, will do monday.

source/agora/node/Ledger.d
AndrejMitrovic: Hmm what did you mean by this? That we should limit the index? the `if` check here should do the job for now, no?

source/agora/common/Transaction.d
AndrejMitrovic: Do you need me to const-ify the parameters?

AndrejMitrovic:
I need to re-add the old unittest and just modify it a bit.

I've pushed a change (hopefully) fixing the code, but I can't properly test it yet because I'm using Windows (well, I'm using WSL1.0 and the test-suite fails there regardless of my changes). Stay tuned..

source/agora/node/Ledger.d
Geod24: I mean that you'll keep iterating the ledger, even though there's no way you get another match.

source/agora/common/Transaction.d
Geod24: It would be great

AndrejMitrovic:
The bug was at my end. Fixed.

source/agora/common/Transaction.d
```diff
@@ -257,37 +257,23 @@ public bool verifyData (Transaction tx,
...
-        else
+        // all referenced outputs must be present
+        auto output = findOutput(input.previous, input.index);
+        if (output is null)
+            return false;
+
+        if (!output.address.verify(input.signature, tx_hash[]))
             return false;
+
+        sum_unspent += output.value;
```
Geod24, line 273: Hello overflow. Let's go with this right now because we don't have a type for monetary value, but it's `#2` on my priority list.

Geod24 (approved):

PR #94, opened 2019-07-17
Fix some networking & parsing bugs

AndrejMitrovic:


AndrejMitrovic:
`data_dir` used to be parsed, but the code to do it got lost in a rebase of an older PR.

We need integration tests soon.

AndrejMitrovic:
Btw, the networking part will be improved once we start working on the settlement epic. The rules for min/max listeners will probably be changed, and it all depends on our consensus protocol & the tech paper.

source/agora/node/Network.d
```diff
@@ -289,7 +289,7 @@ public class NetworkManager
     ///
     private bool allPeersConnected ( )  pure nothrow @safe @nogc
     {
-        return this.todo_addresses.length == 0;
+        return this.peers.length >= this.node_config.min_listeners;
```
Geod24, line 320: Well this function is badly named now

Geod24 (commented):
Isn't the second commit supposed to go before the first ? Otherwise the CI will fail on this specific commit ?

AndrejMitrovic:
> Isn't the second commit supposed to go before the first ? Otherwise the CI will fail on this specific commit ?

I was under the impression it only tests the whole PR. Maybe that was Jenkins though.

source/agora/node/Network.d
AndrejMitrovic: Yes I'll rename it.

AndrejMitrovic:
I've renamed the function. And while doing that I've realized what you meant, I also reintroduced some code I initially removed (so it keeps connecting with new nodes until max is reached).

Also reordered those first two commits.

Geod24:
> I was under the impression it only tests the whole PR. Maybe that was Jenkins though.

It is not enforced by Travis, but each commit should compile & pass test. That's the only way to make `git bisect` work with rebase & merge.

AndrejMitrovic:
> It is not enforced by Travis, but each commit should compile & pass test. That's the only way to make git bisect work with rebase & merge.

Great. I'll add it to the style guide.

AndrejMitrovic:
Rebased.

source/agora/node/RemoteNode.d
```diff
@@ -85,15 +85,21 @@ class RemoteNode
...
+            allPeersConnected = will be periodically called by this method to
+                                determine if we should keep polling the node
+                                for more network info (to connect to more nodes).
 
     ***************************************************************************/
 
     public void getReady ( void delegate(NetworkInfo) receiveNetInfo,
-        void delegate(RemoteNode) onNodeConnected )
+        void delegate(RemoteNode) onNodeConnected,
+		bool delegate() allPeersConnected )
```
Geod24: Broken indent

Geod24 (commented):
I still see the first 2 commits out of order

AndrejMitrovic:
> I still see the first 2 commits out of order

I think this is Github ordering commits by creation date, but I did move them IIRC. I'll check.

Geod24:
> I think this is Github ordering commits by creation date, but I did move them IIRC. I'll check.

Last time I checked, Github does order them by creation date, but if you go on a commit-by-commit view, they are properly ordered when it comes to `Previous` and `Next` commit (as in, click on a commit and the button at the top right allows you to navigate).

But maybe that changed?

AndrejMitrovic:
```
~/dev/agora connection-bugs $ git log origin/v0.x.x..HEAD

* bc02675 Parse the mandatory data_dir config option
* 2b71d6c Ensure Metadata data dir is never empty
* b4dbd59 Remove connected address from "connecting" set
* 1b6f735 Don't keep connecting if peer info is complete
* 45c79de Finish discovery when minimum listeners reached
         5 commits between origin/v0.x.x and HEAD
```

However, maybe I should just squash the first two commits?

Geod24 (approved):

PR #96, opened 2019-07-17
Add basic block height verification

AndrejMitrovic:
This must be done before asynchronous putTransaction requests are implemented, otherwise a node can end up building blocks indefinitely as it accepts any blocks it receives from a call to `getBlocksFrom`.

### DIFF
--- a/source/agora/test/Ledger.d
+++ b/source/agora/test/Ledger.d
@@ -26,6 +26,7 @@ import agora.test.Base;
 /// Returns: the entire ledger from the provided node
 private Block[] getAllBlocks (TestAPI node)
 {
+    import std.range;
     Block[] blocks;
 
     // note: may return less than asked for, hence the loop
@@ -36,8 +37,11 @@ private Block[] getAllBlocks (TestAPI node)
         if (new_blocks.length == 0)  // no blocks left
             break;
 
+        // ensure sequential consistency
+        foreach (block; new_blocks)
+            assert(block.header.height == starting_block++);
+
         blocks ~= new_blocks;
-        starting_block = blocks[$ - 1].header.height + 1;
     }
 
     return blocks;
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -19,6 +19,8 @@ import agora.common.Data;
 import agora.common.Hash;
 import agora.common.Transaction;
 
+import vibe.core.log;
+
 import std.algorithm;
 
 /// Ditto
@@ -106,7 +108,9 @@ public class Ledger
 
     /***************************************************************************
 
-        Add a block to the ledger
+        Add a block to the ledger.
+
+        If the block fails verification, it is not added to the ledger.
 
         Params:
             block = the block to add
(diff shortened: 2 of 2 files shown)


source/agora/test/Ledger.d
```diff
@@ -36,6 +37,21 @@ private Block[] getAllBlocks (TestAPI node)
...
             break;
 
+        auto prev_height = new_blocks[0].header.height;
+
+        // consistency check: new blocks must continue from where we last called
+        // first loop: start with 0, subsequent: start from saved blocks position
+        if (blocks.length)
+            assert(prev_height == blocks.back.header.height + 1);
+        else
+            assert(prev_height == 0);  // genesis block
```
Geod24: Why not just `new_blocks[0].header.height == starting_block` ?

source/agora/test/Ledger.d
```diff
@@ -36,6 +37,21 @@ private Block[] getAllBlocks (TestAPI node)
...
+
+        foreach (block; new_blocks[1 .. $])
+        {
+            // sequential consistency check
+            assert(block.header.height == ++prev_height);
```
Geod24: Yeah you could just reuse `starting_height` (maybe also rename it?) because those two have roughly the same function

AndrejMitrovic: The second function verifies that the chain of blocks has an increasing height.

AndrejMitrovic: So it's not like `[height0, height2, height1, height3]`.

AndrejMitrovic:
Rebased with the first proposed change. But I left the sequential consistency check in.

source/agora/test/Ledger.d
Geod24:
```d
size_t starting_block = 0;
while (1)
{
    auto new_blocks = node.getBlocksFrom(starting_block, uint.max);
    if (new_blocks.length == 0)  // no blocks left
        break;
    assert(new_blocks[0].header.height == starting_block)
    foreach (ref block; new_blocks[1 .. $])
        assert(block.header.height == ++starting_block);
    blocks ~= new_blocks;
}
```

What's wrong with this ?

AndrejMitrovic: Nothing wrong, that's what reviews are for. Thanks!

AndrejMitrovic: > What's wrong with this ?

I've solved the riddle. A missing `++starting_block` at the very end before the while loop ends, otherwise it's an infinite loop.

AndrejMitrovic:
Pushed the suggested changes, + a bugfix for the suggested change.

source/agora/test/Ledger.d
```diff
@@ -36,8 +37,13 @@ private Block[] getAllBlocks (TestAPI node)
         if (new_blocks.length == 0)  // no blocks left
             break;
 
+        // ensure sequential consistency
+        assert(new_blocks[0].header.height == starting_block);
+        foreach (ref block; new_blocks[1 .. $])
+            assert(block.header.height == ++starting_block);
+
+        ++starting_block;
```
Geod24: Would:
```D
        foreach (ref block; new_blocks)
            assert(block.header.height == starting_block++);
```

Work ?

AndrejMitrovic:
That works.

It made me think. A cool project would be to come up with a libdparse app which tries to refactor code while retaining the same semantics.

Geod24 (approved):

PR #97, opened 2019-07-17
Make sendTransaction() asynchronous

AndrejMitrovic:
Requires:
- https://github.com/bpfkorea/agora/pull/96

Note that this also contains a TaskManager abstraction which is related to the backlog issue "Build a replaceable task abstraction". That issue is not in the current sprint. But I don't consider this solving the aforementioned issue entirely yet.

The reason I've included this is because I do not want to have to pass the `NetworkManager` class around to the `RemoteNode` class or anywhere else, when the only thing the code wants to do is start tasks.

### DIFF
--- a/source/agora/common/Task.d
+++ b/source/agora/common/Task.d
@@ -1,0 +1,54 @@
+/*******************************************************************************
+
+    Contains a task manager backed by vibe.d's event loop.
+
+    Overriding classes can implement task routines to run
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
Unblocked.

Geod24 (approved):

AndrejMitrovic:
Merge?

PR #100, opened 2019-07-17
Implement MerkleTree ( Merkle-Path, Merkle-Proof )

MichaelKim20:
Multiple transactions are structured in binary tree, and a node consists of a hash of the transaction or nodes.

MerkleTree_BuildMerkleTree

MerkleTree_MakePath

MerkleTree_Proof


source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,215 @@
...
+
+    // Create the second Merkle Node.
+    node = MerkleNode(null, null, &tree.txs[1]);
+    node.hash = node.getHash();
+    assert(node.hash == hashOfTxs[1]);
+    tree.nodes ~= node;
+
+    Hash mergeHash(Hash h1, Hash h2)
+    {
+        struct merge_hash_struct
```
Geod24:
```suggestion
        static struct MergeHash
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,215 @@
...
+    assert(node.hash == hashOfTxs[0]);
+    tree.nodes ~= node;
+
+    // Create the second Merkle Node.
+    node = MerkleNode(null, null, &tree.txs[1]);
+    node.hash = node.getHash();
+    assert(node.hash == hashOfTxs[1]);
+    tree.nodes ~= node;
+
+    Hash mergeHash(Hash h1, Hash h2)
```
Geod24:
```suggestion
    static Hash mergeHash(Hash h1, Hash h2)
```

Great usage of nested functions BTW!

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,215 @@
...
+    {
+        struct merge_hash_struct
+        {
+            Hash left;
+            Hash right;
+
+            public void computeHash (scope HashDg dg) const nothrow @safe @nogc
+            {
+                hashPart(left, dg);
```
Geod24:
```suggestion
                hashPart(this.left, dg);
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,215 @@
...
+                hashPart(right, dg);
```
Geod24:
```suggestion
                hashPart(this.right, dg);
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,215 @@
...
+
+        return hashFull(merge_hash_struct(h1, h2));
+    }
+
+    // Create the third Merkle Node. This is root
+    node = MerkleNode(&tree.nodes[0], &tree.nodes[1]);
+    node.hash = node.getHash();
+    tree.nodes ~= node;
+
+    assert(node.hash == mergeHash( hashOfTxs[0], hashOfTxs[1] ));
```
Geod24:
```suggestion
    assert(node.hash == mergeHash(hashOfTxs[0], hashOfTxs[1]));
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,215 @@
...
+
+    /// Array of `Transaction`
+    /// All transactions in the `MerkleTree` are stored here,
+    /// and refer to its memory everywhere.
+    public Transaction[] txs;
+
+    /// Array of MerkleNode,
+    /// All `MerkleNode` in the `MerkleTree` are stored here,
+    /// and refer to its memory everywhere.
+    public MerkleNode[] nodes;
```
Geod24: I don't see why you need a binary tree if you store everything in an array ? In fact I don't see why you would need the `MerkleNode` structure at all if you have an array of `Transaction` ?
Because as long as your tree is balanced (and it should be), you know how many hashes you are going to need.

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,215 @@
...
+    public MerkleNode* left;
+
+    /// The right branch's ptr of the tree.
+    public MerkleNode* right;
+
+    /// The transaction's ptr
+    public Transaction* tx;
+
+    /// The hash of this `MerkleNode`
+    public Hash hash;
```
Geod24: This is never written to from within this struct. So it is assumed that the caller will set it ?

Geod24 (commented):
Pretty good start! I think there's a bit of redundancy that could be reduced. Also I'd like to see tree-handling functions in addition to the data structure, so they can be called from the unittests.

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,208 @@
...
+
+    static MerkleNode opCall(MerkleNode* left, MerkleNode* right, Transaction* tx = null)
```
Geod24: Missing protection attribute and documentation.
Also why a `static opCall` over a constructor ?

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,212 @@
...
+            left = The left branch's ptr of the tree.
+            right = The right branch's ptr of the tree.
+            tx = The transaction's ptr
+
+    ***************************************************************************/
+
+    /// Ctor
+    this(MerkleNode* left, MerkleNode* right, Transaction* tx = null)
+    {
+        if ((left !is null) && (right !is null))
```
AndrejMitrovic: This can be simplified.

```D
import std.algorithm;

if (left !is null && 
    right !is null &&
    right.hash < left.hash)
{
    swap(left, right);
}

this.left = left;
this.right = right;
```

AndrejMitrovic: But I don't know if @Geod24 likes it.

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,233 @@
...
+
+import agora.common.Data;
+import agora.common.Hash;
+import agora.common.Transaction;
+
+import std.algorithm;
+
+
+/// Ditto
+public struct MerkleTree
```
Geod24: This seems to be redundant with `MerkleNode` now, isn't it ?

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,233 @@
...
+
+    /// The transaction's ptr
+    public Transaction* tx;
+
+    /// The hash of this `MerkleNode`
+    public Hash hash;
+
+    /***************************************************************************
+
+        Create
```
Geod24: Incomplete doc ?

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,233 @@
...
+
+        Params:
+            left = The left branch's ptr of the tree.
+            right = The right branch's ptr of the tree.
+            tx = The transaction's ptr
+
+    ***************************************************************************/
+
+    /// Ctor
```
Geod24: You can remove this line

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,233 @@
...
+    this (MerkleNode* left, MerkleNode* right, Transaction* tx = null)
```
Geod24:
```suggestion
    public this (MerkleNode* left, MerkleNode* right, Transaction* tx = null)
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,233 @@
...
+    {
+        this.left = left;
+        this.right = right;
+        this.tx = tx;
+        this.hash = this.getHash();
```
Geod24: `getHash` is essentially doing what `computeHash` is doing.
You might want to do:
```suggestion
        this.hash = hashFull(this);
```

Instead, remove `getHash`, and define `computeHash` as:
```D
    {
        // Our node contains either a transaction, in which case it is a leaf node,
        // or 2 childs (which might be the same if the tree is unbalanced)
        if (this.transaction !is null)
            hashPart(*this.transaction, dg);
        else
        {
            hashPart(this.left.hash, dg);
            hashPart(this.right.hash, dg);
        }
}
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,233 @@
...
+        Params:
+            left = The left branch's ptr of the tree.
+            right = The right branch's ptr of the tree.
+            tx = The transaction's ptr
+
+    ***************************************************************************/
+
+    /// Ctor
+    this (MerkleNode* left, MerkleNode* right, Transaction* tx = null)
+    {
```
Geod24: You might want to add a contract that `left` and `right` are both non `null` OR that `tx` is null.

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,233 @@
...
+        nothrow @nogc @trusted
+{
+    static struct MergeHash
+    {
+        Hash left;
+        Hash right;
+
+        public void computeHash (scope HashDg dg) const nothrow @safe @nogc
+        {
+            if (this.left < this.right)
```
Geod24: Why ?

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,233 @@
...
+                hashPart(this.right, dg);
+                hashPart(this.left, dg);
+            }
+        }
+    }
+
+    return hashFull(MergeHash(h1, h2));
+}
+
+/// If there is no transaction, the hash value will be the initial value.
```
Geod24: I don't think we should be able to create a tree without transaction. There is no use case for it.

source/agora/common/MerkleTree.d
MichaelKim20: It is to change the order of the two so that the same value can be obtained.
We compared the values first, and we got the hash value ahead of the small one.

source/agora/common/MerkleTree.d
MichaelKim20: Ok

source/agora/common/MerkleTree.d
Geod24: It however means you need to change the order of the transaction at the lower layer.

source/agora/common/MerkleTree.d
MichaelKim20: @Geod24 I added several method

source/agora/common/MerkleTree.d
MichaelKim20: @Geod24  Is this a `document` ?

Geod24: I'm sorry I don't understand the question.

I mean to add:
```d
assert((left is null) == (right is null), "`right` and `left` should be both `null` or both non-`null`");
assert((left is null) != (tx is null), "Either `tx` should be `null`, or there should be no child nodes");
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+            txs = Array of `Transaction`
+
+    ***************************************************************************/
+
+    public this (ref Transaction[] txs)
+    {
+        MerkleNode[] nodes;
+
+        foreach (ref tx; txs)
+            nodes ~= MerkleNode(null, null, &tx);
```
Geod24: This is quite inefficient.
When you append (`~`) to an array, you ask the compiler to add more contiguous memory.
As a result, the compiler might re-allocate the array somewhere else, move the data, and add the new structure.
You can imagine how adding a lot of nodes would be inefficient, because you will reallocate the memory and move the data many times.

Instead, if you already know the size of the array, you can do the following:
```D
MerkleNode[] nodes = new MerkleNode[](txs.length);
foreach (ref node; nodes)
    node = MerkleNode(null, null, &tx);
```

This is already a much more efficient approach that the current one.
In addition, you can add [uninitializedArray](https://dlang.org/library/std/array/uninitialized_array.html) to further reduce the overhead (it avoids overriding `node`).

In addition, if you wanted to, you could pre-emptively allocate all nodes, as you know in advance how many there are.

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+
+        const max_level = to!long(ceil(log2(nodes.length)));
+        for (int i = 0; i < max_level; i++)
```
Geod24: In D you can write:
```suggestion
        foreach (i; 0 .. max_level)
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+            MerkleNode[] nodesInLevel;
+            for (int j = 0; j < nodes.length; j += 2)
+                nodesInLevel ~= MerkleNode(&nodes[j], &nodes[j + 1]);
+            nodes = nodesInLevel[];
+        }
+
+        if (nodes.length > 0)
+            this.root = &nodes[0];
+        else
+            this.root = null;
```
Geod24: That `else` branch is unnecessary

source/agora/common/MerkleTree.d
Geod24: Those methods would be better on `MerkleNode` I think.
It makes the code more composable and reusable, as any `MerkleNode` is also a `MerkleTree`.

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+
+        Params:
+            hash = looking hash.
+
+        Returns:
+            Return node if found the hash. Return null otherwise.
+
+    *******************************************************************************/
+
+    public MerkleNode* searchNode (Hash hash)
```
Geod24: When a method returns a value internal to it, it is good to use `inout`.
In this case:
```suggestion
    public inout(MerkleNode)* searchNode (Hash hash) inout
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+        if (this.hash == hash) return &this;
+
+        MerkleNode* res;
+        res = this.right.searchNode(hash);
+        if (res != null) return res;
+
+        res = this.left.searchNode(hash);
+        if (res != null) return res;
+
+        return null;
```
Geod24: This could be greatly simplified:
```d
if (this.hash == hash) return &this;
if (auto res = this.left.search(hash)
    return res;
return this.right.search(hash)
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+    KeyPair[] key_pairs = [
+        KeyPair.fromSeed(Seed.fromString("SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4")),
+        KeyPair.fromSeed(Seed.fromString("SBYEHWKMZCXY2JIWVNFRLJQ6KTTEZSNOYMA7AV3L4RKIARX2U7D5IRIA")),
+        KeyPair.fromSeed(Seed.fromString("SC337M34RPZ5US6FMHDPPOSQJYURJU2LSELNFUF2ANZH3IF4MSFEOQ6Q")),
+        KeyPair.fromSeed(Seed.fromString("SDEJK352SOMTSJEMT7KNS4NBC3SWNDHU7VQMFCGGABMUEI4242XQK4VI")),
+        KeyPair.fromSeed(Seed.fromString("SC3X4L4GWP5BSPKAUU7ZBWWWQXLHTZRKY2YJ6HA5B7SB6YLYV5OCLWJJ")),
+        KeyPair.fromSeed(Seed.fromString("SBDJ3WL3BFNJ4GE4MCQSEJC6YMCGV636QOACHFW43UGVAFQIY3IAQOB4")),
+        KeyPair.fromSeed(Seed.fromString("SBFQVTPNSLB3PAGJPGPCMUYYBTXF3FOAOBCEOSYMOSAFBHJUNHCNZUO4")),
+        KeyPair.fromSeed(Seed.fromString("SAQK7BMZG4VKJZ2J46U44M4VQFAC2LPCJPUGUWJL3O3R4BGFJNQHOOD5")),
+        KeyPair.fromSeed(Seed.fromString("SBKDH3GHUECKVA3ZHVLANW2B27RBASQHCBK5TCKBG72CVVVSGWTWGXCD"))
```
Geod24: Why not a random keypair ?

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+
+    *******************************************************************************/
+
+    static public bool proveMerklePath (Hash hash, Hash[] path, Hash root, int index)
+    {
+        if (path.length == 0) return false;
+        if (index < 0) return false;
+
+        Hash h = hash;
+        int idx = index;
```
Geod24: Since `Hash` and `int` are value types, those two variable are superfluous

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    /*******************************************************************************
+
+        Prove merkle path
+
+        Params:
+            hash = `Hash` of `Transaction`
+            path = `Hash` of Merkle path
+            root = `Hash` of Merkle root
```
Geod24: `index` not documented

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+
+        Returns:
+            Return true if the Merkle proof is completed. Return false otherwise.
+
+    *******************************************************************************/
+
+    static public bool proveMerklePath (Hash hash, Hash[] path, Hash root, int index)
+    {
+        if (path.length == 0) return false;
+        if (index < 0) return false;
```
Geod24: If it cannot be `< 0`, make it unsigned

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+            hash = `Hash` of `Transaction`
+            path = `Hash` of Merkle path
+            root = `Hash` of Merkle root
+
+        Returns:
+            Return true if the Merkle proof is completed. Return false otherwise.
+
+    *******************************************************************************/
+
+    static public bool proveMerklePath (Hash hash, Hash[] path, Hash root, int index)
```
Geod24:
```suggestion
   public static bool proveMerklePath (Hash hash, const scope Hash[] path, Hash root, int index)
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,384 @@
...
+    {
+        if (path.length == 0) return false;
+        if (index < 0) return false;
+
+        Hash h = hash;
+        int idx = index;
+        foreach (ref p; path)
+        {
+            if (idx % 2 == 0)
```
Geod24: I'm not sure I understand the need for `index` here ?

source/agora/common/MerkleTree.d
MichaelKim20: I used a fixed key to validate this test, but I don't need it anymore. So let's do it randomly.

source/agora/common/MerkleTree.d
MichaelKim20: That's what I've thought. I did this to make it clear.^^

source/agora/common/MerkleTree.d
MichaelKim20: It's my mistake. I'll modify it.

source/agora/common/MerkleTree.d
MichaelKim20: It's my mistake.

source/agora/common/MerkleTree.d
MichaelKim20: To distinguish between parameters and internal variables.

source/agora/common/MerkleTree.d
MichaelKim20: Thank you~

source/agora/common/MerkleTree.d
MichaelKim20: OK

source/agora/common/MerkleTree.d
MichaelKim20: OK

source/agora/common/MerkleTree.d
MichaelKim20: OK

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,345 @@
...
+    Create MerkleTree
+
+    Params:
+        txs = Array of `Transaction`
+
+    Returns:
+        Return root node.
+
+***************************************************************************/
+public MerkleNode* MakeMerkleTree (ref Transaction[] txs)
```
Geod24: Why is it passed by `ref` ? An array is already a reference type.

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,345 @@
...
+
+    Create MerkleTree
+
+    Params:
+        txs = Array of `Transaction`
+
+    Returns:
+        Return root node.
+
+***************************************************************************/
```
Geod24:
```suggestion
***************************************************************************/

```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,345 @@
...
+
+        MerkleNode[] next_nodes = new MerkleNode[](nodes.length / 2);
+
+        for (int j = 0; j < nodes.length; j += 2)
+            next_nodes[j / 2] = MerkleNode(&nodes[j], &nodes[j + 1]);
+
+        nodes = next_nodes[];
+    }
+
+    if (nodes.length > 0)
```
Geod24: `nodes.length` can only be `0` if `txs.length` is `0` right ? In which case we might want to handle it explicitly at the beginning of the function

Geod24 (commented):
@AndrejMitrovic : Could you give this PR a review ?

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,345 @@
...
+        txs = Array of `Transaction`
+
+    Returns:
+        Return root node.
+
+***************************************************************************/
+public MerkleNode* MakeMerkleTree (ref Transaction[] txs)
+{
+    MerkleNode[] nodes = new MerkleNode[](txs.length);
+    foreach (size_t idx, ref node; nodes)
```
AndrejMitrovic: I think you can write this as:

```D
MarkleNode[] nodes;
nodes.reserve(txs.length);
foreach (ref tx; txs)
    nodes ~= MerkleNode(null, null, &tx);
```

Geod24: The current version is more efficient though as it does not go into the GC

AndrejMitrovic: ok @Geod24 is right

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,345 @@
...
+            if (getPath(node.left, hash, path))
+            {
+                path ~= node.right.hash;
+                return true;
+            }
+
+            return false;
+        }
+
+        Hash [] path;
```
AndrejMitrovic:
```suggestion
        Hash[] path;
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,345 @@
...
+            hash = `Hash` of `Transaction`
+
+        Returns:
+            Return array of `Hash` of Merkle path
+
+    *******************************************************************************/
+
+    public Hash[] getMerklePath (Hash hash)
+    {
+        bool getPath (MerkleNode* node, Hash hash, ref Hash [] path)
```
AndrejMitrovic:
```suggestion
        static bool getPath (MerkleNode* node, Hash hash, ref Hash [] path)
```

source/agora/common/MerkleTree.d
```diff
@@ -0,0 +1,345 @@
...
+        else
+        {
+            hashPart(this.left.hash, dg);
+            hashPart(this.right.hash, dg);
+        }
+    }
+}
+
+/// Merge two hash
+public Hash mergeHash (Hash h1, Hash h2)
```
AndrejMitrovic: Will this function be used outside this module? If not, it should be `private`. Also it's missing documentation.

source/agora/common/MerkleTree.d
AndrejMitrovic:
```suggestion
public MerkleNode* makeMerkleTree (ref Transaction[] txs)
```

By the styling convention, we start methods with lowercase I think.

AndrejMitrovic:
I've completed my review.

source/agora/common/MerkleTree.d
MichaelKim20: OK

AndrejMitrovic (approved):

PR #102, opened 2019-07-19
Added ledger serialization function

TrustHenry:
Function definition and helper related to serialization

### DIFF
--- a/source/agora/common/Block.d
+++ b/source/agora/common/Block.d
@@ -19,6 +19,7 @@ import agora.common.crypto.Key;
 import agora.common.Data;
 import agora.common.Hash;
 import agora.common.Transaction;
+import agora.node.BlockSerialize;
 
 /*******************************************************************************
 
@@ -53,6 +54,22 @@ public struct BlockHeader
     {
         dg(this.prev_block[]);
         hashPart(this.height, dg);
+        dg(this.tx_hash[]);
+    }
+
+    /***************************************************************************
+
+        Block Header Serialization
+
+        Params:
+             dg = serialize function accumulator
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) pure const nothrow @safe
+    {
+        dg(this.prev_block[]);
+        serializePart(this.height, dg);
         dg(this.tx_hash[]);
     }
 }
@@ -94,6 +111,21 @@ public struct Block
 
     ///
     public Transaction tx;
+
+    /***************************************************************************
+
+        Block serialization
+
+        Params:
+            dg = serialize function accumulator
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) pure const nothrow @safe
+    {
+        serializePart(header, dg);
+        serializePart(tx, dg);
+    }
 }
 
 /*******************************************************************************
--- a/source/agora/common/Transaction.d
+++ b/source/agora/common/Transaction.d
@@ -21,6 +21,7 @@ module agora.common.Transaction;
 import agora.common.Data;
 import agora.common.Hash;
 import agora.common.crypto.Key;
+import agora.node.BlockSerialize;
 
 import std.algorithm;
 
@@ -76,6 +77,25 @@ public struct Transaction
         foreach (output; this.outputs)
             hashPart(output, dg);
     }
+
+
+    /***************************************************************************
+
+        Transactions Serialization
+
+        Params:
+            dg = Serialize function
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) pure const nothrow @safe
+    {
+        foreach (const ref input; this.inputs)
+            serializePart(input, dg);
+
+        foreach (const ref output; this.outputs)
+            serializePart(output, dg);
+    }
 }
 
 ///
@@ -122,6 +142,21 @@ public struct Output
         hashPart(this.value, dg);
         hashPart(this.address, dg);
     }
+
+    /***************************************************************************
+
+        Output Serialization
+
+        Params:
+            dg = Serialize function
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) pure const nothrow @safe
+    {
+        serializePart(this.value, dg);
+        serializePart(this.address, dg);
+    }
 }
 
 /// The input of the transaction, which spends a previously received `Output`
@@ -150,6 +185,22 @@ public struct Input
     {
         hashPart(this.previous, dg);
         hashPart(this.index, dg);
+    }
+
+    /***************************************************************************
+
+        Input Serialization
+
+        Params:
+             dg = serialize function
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) pure const nothrow @safe
+    {
+        serializePart(this.previous, dg);
+        serializePart(this.index, dg);
+        serializePart(this.signature, dg);
     }
 }
 
--- a/source/agora/common/crypto/Key.d
+++ b/source/agora/common/crypto/Key.d
@@ -16,6 +16,7 @@ module agora.common.crypto.Key;
 import agora.common.Data;
 import agora.common.crypto.Crc16;
 import agora.common.Hash;
+import agora.node.BlockSerialize;
 
 import geod24.bitblob;
 import base32;
@@ -171,6 +172,20 @@ public struct PublicKey
     ***************************************************************************/
 
     public void computeHash (scope HashDg dg) const nothrow @safe @nogc
+    {
+        dg(this.data[]);
+    }
+
+    /***************************************************************************
+
+        Key Serialization
+
+        Params:
+            dg = Serialize function
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) pure const nothrow @safe
     {
         dg(this.data[]);
     }
--- a/source/agora/node/BlockSerialize.d
+++ b/source/agora/node/BlockSerialize.d
@@ -1,0 +1,92 @@
+/*******************************************************************************
+
+    Function definition and helper related to serialization
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.BlockSerialize;
+
+import agora.common.Data;
+
+/// Type of delegate SerializeDg
+public alias SerializeDg = void delegate(scope const(ubyte)[]) pure nothrow @safe;
+
+public ubyte[] serializeFull (T) (scope const auto ref T record)
+    nothrow @safe
+{
+    ubyte[] res;
+    scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
+    {
+        res ~= data;
+    };
+    serializePart(record, dg);
+
+    return res;
+}
+
+/// Ditto
+public void serializePart (T) (scope const auto ref T record, scope SerializeDg dg)
+    pure nothrow @safe
+    if (is(T == struct))
+{
+    static assert(is(typeof(T.init.serialize(SerializeDg.init))),
+                "Struct `" ~ T.stringof ~
+                "` does not implement `serialize(scope SerializeDg) const nothrow` function");
+    record.serialize(dg);
+}
+
+/// Ditto
+public void serializePart () (scope const auto ref Hash record, scope SerializeDg dg)
+    pure nothrow @safe
+{
+    dg(record[]);
+}
+
+/// Ditto
+public void serializePart (scope const Signature record, scope SerializeDg dg)
+    pure nothrow @trusted
+{
+   dg(record[]);
+}
+
+/// Ditto
+public void serializePart (ubyte record, scope SerializeDg dg)
+    pure nothrow @trusted
+{
+    dg((cast(ubyte*)&record)[0 .. ubyte.sizeof]);
+}
+
+/// Ditto
+public void serializePart (ushort record, scope SerializeDg dg)
+    pure nothrow @trusted
+{
+    dg((cast(ubyte*)&record)[0 .. ushort.sizeof]);
+}
+
+/// Ditto
+public void serializePart (uint record, scope SerializeDg dg)
+    pure nothrow @trusted
+{
+    dg((cast(ubyte*)&record)[0 .. uint.sizeof]);
+}
+
+/// Ditto
+public void serializePart (ulong record, scope SerializeDg dg)
+    pure nothrow @trusted
+{
+    dg((cast(ubyte*)&record)[0 .. ulong.sizeof]);
+}
+
+/// Ditto
+public void serializePart (scope cstring record, scope SerializeDg dg)
+    pure nothrow @trusted
+{
+    dg(cast(const ubyte[])record);
+}
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -18,6 +18,7 @@ import agora.common.Block;
 import agora.common.Data;
 import agora.common.Hash;
 import agora.common.Transaction;
+import agora.node.BlockSerialize;
 
 import std.algorithm;
 
@@ -153,6 +154,7 @@ public class Ledger
 unittest
 {
     import agora.common.crypto.Key;
+    import std.digest : toHexString;
 
     scope ledger = new Ledger;
     assert(ledger.getLastBlock() == getGenesisBlock());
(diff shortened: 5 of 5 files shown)


source/agora/node/LederOutputStream.d
```diff
@@ -0,0 +1,130 @@
...
+        offset += ulong.sizeof;
+    }
+
+    // get Data bytes
+    public ubyte[] getData() @trusted
+    {
+        return this.data;
+    }
+
+    // TOBE
```
Geod24: I think you mean `TODO` ?

source/agora/node/LederOutputStream.d
```diff
@@ -0,0 +1,130 @@
...
+
+/// Ditto
+public class LedgerOutputStream
+{
+    private ubyte[] data;
+
+    private size_t offset;
+
+    /// Ctor
+    public this () @trusted
```
Geod24: Why does this need to be `@trusted` ? In general, we should use `@trusted` as little as possible, and every usage should be carefully considered.

source/agora/node/LederOutputStream.d
```diff
@@ -0,0 +1,130 @@
...
+
+    // TOBE
+    public void clear() @trusted
+    {
+        data.length = 0;
+        offset = 0;
+    }
+
+    // TOBE
+    public void destroy() @trusted
```
Geod24: Having 3 functions with different name that do the same thing is a red flag ;)

source/agora/node/LederOutputStream.d
```diff
@@ -0,0 +1,130 @@
...
+    // Assign additional data size bytes length and store bytes in data
+    // and add offset location bytes size.
+    void write(in ubyte[] bytes) @trusted
+    {
+        reserve(bytes.length);
+        data[offset .. offset + bytes.length] = bytes[];
+        offset += bytes.length;
+    }
+
+    void write(in byte[] bytes) @trusted
```
Geod24: I don't think we should ever need this.

source/agora/node/LederOutputStream.d
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    void writeLong(const long value) @trusted
+    {
+        reserve(long.sizeof);
+        *cast(long*)&data[offset] = value;
+        offset += long.sizeof;
+    }
+
+    void writeAmount(const Amount value) @trusted
```
Geod24: `Amount` is currently an `alias` to `long`, so one could just call `writeLong`

source/agora/node/LederOutputStream.d
```diff
@@ -0,0 +1,130 @@
...
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.LedgerOutputStream;
+
+import agora.common.Transaction;
+
+/// Ditto
+public class LedgerOutputStream
```
Geod24: It seems that this whole class is re-inventing a GC-backed dynamic array.
The `write` methods use a pattern of `reserve` (which re-allocates the data), then write to a slice.
Instead, one can just use the appending operator (`~=`) to obtain the same effects. I commented on one of the methods with an example.

Additionally, we should try to use `struct` over `class` when there is no need for OOP.

source/agora/node/LederOutputStream.d
```diff
@@ -0,0 +1,130 @@
...
+            data = cast(ubyte[]) vdata;
+        }
+    }
+
+    // Assign additional data size bytes length and store bytes in data
+    // and add offset location bytes size.
+    void write(in ubyte[] bytes) @trusted
+    {
+        reserve(bytes.length);
+        data[offset .. offset + bytes.length] = bytes[];
```
Geod24:
```suggestion
        this.data ~= bytes;
```
You can use this instead, and remove the `reserve` call, to achieve the same results.

Geod24 (commented):
I noted a few things. In addition, I'll add that we should use verbs, not nouns, for method name. So `serialization` would be better named `serialize` (because it takes the action).
I think we have way too much `@trusted` in our code. `@trusted` is telling the compiler "Don't check that code". It is dangerous and should be used when there is no other way. The same is true for `cast`.

In addition to the length test, it would be good to have a test of the serialized values.

source/agora/node/LederOutputStream.d
TrustHenry: That's right. TODO, I deleted it in this PR.

TrustHenry: The function is in development, so this PR has been deleted.

source/agora/node/LederOutputStream.d
TrustHenry: The function is in development, so this PR has been deleted.

source/agora/node/LederOutputStream.d
TrustHenry: I don't think we need byte[] here.

source/agora/node/LederOutputStream.d
TrustHenry: Yes, deleted

source/agora/node/LederOutputStream.d
TrustHenry: I use this it.

source/agora/node/LederOutputStream.d
TrustHenry: For some reason, I used `@trusted`
It cannot be used in the event of a pointer change.

source/agora/node/LederOutputStream.d
AndrejMitrovic: For example:

```D
auto bytes = cast(ubyte[long.sizeof]*)&value;
this.data ~= (*bytes)[];
```

Geod24: `s/auto/const/`

TrustHenry: What do you mean?

AndrejMitrovic: `s/word1/word` is a common way of expressing substitution. So, just replace `auto bytes` with `const bytes` here.

Personally I don't think it matters when there's only two lines of code.

AndrejMitrovic: Well actually it could matter if the backing store's element type is const I suppose. Anyway yeah, just make it const.

source/agora/node/Ledger.d
```diff
@@ -147,12 +149,28 @@ public class Ledger
...
+    /***************************************************************************
+
+        Ledger serialization
+
+        Params:
+            block = LedgerOutputStream
+
+    ***************************************************************************/
+
+    public void serialize (LedgerOutputStream stream) @safe
```
Geod24: I would not add a `serialize` method to `Ledger`. Because it is not a data structure that needs serialization, it is a owner of the data (which is likely to know when serialization is needed), thus it does not fall under the same abstraction.

In other word: How often do we want to serialize the full blockchain ?

source/agora/node/Ledger.d
```diff
@@ -18,8 +18,10 @@ import agora.common.Block;
 import agora.common.Data;
 import agora.common.Hash;
 import agora.common.Transaction;
+import agora.node.LedgerOutputStream;
 
 import std.algorithm;
+import std.mmfile;
```
Geod24: Superfluous?

source/agora/common/Transaction.d
```diff
@@ -76,6 +77,25 @@ public struct Transaction
...
+        Transactions Serialization
+
+        Params:
+            stream = LedgerOutputStream
+
+    ***************************************************************************/
+
+    public void serialize(LedgerOutputStream stream) @safe
+    {
+        foreach (Input input; this.inputs)
```
```suggestion
        foreach (const ref input; this.inputs)
```

Geod24: Actually I am not sure it works. If not, just `ref`

Geod24 (commented):
I think the "Stream" class could be better replaced with a delegate.
Note that a "stream" is a C++ concept, while D uses ranges. A delegate is an output range, and pretty similar to a stream all things considered.

source/agora/node/Ledger.d
TrustHenry: It's a necessary part of the next version, but I'll delete it now.

source/agora/node/Ledger.d
TrustHenry: So I deleted it :)

source/agora/common/Block.d
```diff
@@ -55,6 +56,22 @@ public struct BlockHeader
...
+    /***************************************************************************
+
+        Block Header Serialization
+
+        Params:
+             dg = serialize function accumulator
+
+    ***************************************************************************/
+
+    public void serialize(scope SerializeDg dg) const nothrow @safe
```
Geod24:
```suggestion
    public void serialize (scope SerializeDg dg) const nothrow @safe
```

source/agora/common/Block.d
```diff
@@ -94,6 +111,21 @@ public struct Block
...
+    /***************************************************************************
+
+        Block serialization
+
+        Params:
+            dg = serialize function accumulator
+
+    ***************************************************************************/
+
+    public void serialize(scope SerializeDg dg) const nothrow @safe
```
Geod24:
```suggestion
    public void serialize (scope SerializeDg dg) const nothrow @safe
```

source/agora/common/Transaction.d
```diff
@@ -76,6 +77,25 @@ public struct Transaction
...
+    /***************************************************************************
+
+        Transactions Serialization
+
+        Params:
+            dg = Serialize function
+
+    ***************************************************************************/
+
+    public void serialize(scope SerializeDg dg) const nothrow @safe
```
```suggestion
    public void serialize (scope SerializeDg dg) const nothrow @safe
```

source/agora/common/Transaction.d
```diff
@@ -122,6 +142,21 @@ public struct Output
...
+    /***************************************************************************
+
+        Output Serialization
+
+        Params:
+            dg = Serialize function
+
+    ***************************************************************************/
+
+    public void serialize(scope SerializeDg dg) const nothrow @safe
```
```suggestion
    public void serialize (scope SerializeDg dg) const nothrow @safe
```

source/agora/common/Transaction.d
```diff
@@ -151,6 +186,22 @@ public struct Input
...
+    /***************************************************************************
+
+        Input Serialization
+
+        Params:
+             dg = serialize function
+
+    ***************************************************************************/
+
+    public void serialize(scope SerializeDg dg) const nothrow @safe
```
```suggestion
    public void serialize (scope SerializeDg dg) const nothrow @safe
```

source/agora/common/crypto/Key.d
```diff
@@ -175,6 +176,20 @@ public struct PublicKey
...
+    /***************************************************************************
+
+        Key Serialization
+
+        Params:
+            dg = Serialize function
+
+    ***************************************************************************/
+
+    public void serialize(scope SerializeDg dg) const nothrow @safe
```
Geod24:
```suggestion
    public void serialize (scope SerializeDg dg) const nothrow @safe
```

source/agora/node/BlockSerialize.d
```diff
@@ -0,0 +1,81 @@
...
+import geod24.bitblob;
+
+/// Type of delegate SerializeDg
+public alias SerializeDg = void delegate(scope const(ubyte)[]) /*pure*/ nothrow @safe;
+
+public ubyte[] serializeFull (T) (scope const auto ref T record)
+    nothrow @trusted
+{
+    ubyte[] res;
+    scope SerializeDg dg = (scope const(ubyte []) data) nothrow @safe
```
Geod24: I think you can just do:
```suggestion
    scope SerializeDg dg = (v)
```

If not, just a slight change:
```suggestion
    scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
```

source/agora/node/BlockSerialize.d
```diff
@@ -0,0 +1,81 @@
...
+        res ~= data;
+    };
+
+    serializePart(record, dg);
+    return res;
+}
+
+/// Ditto
+public void serializePart (T) (scope const auto ref T record, scope SerializeDg dg)
+    /*pure*/ nothrow
```
Geod24:
```suggestion
    pure nothrow @safe
```

source/agora/node/BlockSerialize.d
```diff
@@ -0,0 +1,81 @@
...
+module agora.node.BlockSerialize;
+
+import agora.common.Data;
+import geod24.bitblob;
+
+/// Type of delegate SerializeDg
+public alias SerializeDg = void delegate(scope const(ubyte)[]) /*pure*/ nothrow @safe;
+
+public ubyte[] serializeFull (T) (scope const auto ref T record)
+    nothrow @trusted
```
Geod24: Why is it `@trusted` ? Also this can be made `pure` (unlike the hashing part)

source/agora/node/BlockSerialize.d
```diff
@@ -0,0 +1,81 @@
...
+    if (is(T == struct))
+{
+    static assert(is(typeof(T.init.serialize(SerializeDg.init))),
+                "Struct `" ~ T.stringof ~
+                "` does not implement `serialize(scope SerializeDg) const nothrow` function");
+    record.serialize(dg);
+}
+
+/// Ditto
+public void serializePart () (scope const auto ref BitBlob!512LU record, scope SerializeDg dg)
```
Geod24: Why `BitBlob!512U` ?

source/agora/node/BlockSerialize.d
```diff
@@ -0,0 +1,81 @@
...
+    /*pure*/ nothrow @safe
+{
+    dg(record[]);
+}
+
+/// Ditto
+public void serializePart (ubyte record, scope SerializeDg dg) /*pure*/ nothrow @trusted
```
Geod24: You might want to add overload for each basic types, because there are implicit conversions happening.

source/agora/node/Ledger.d
```diff
@@ -175,6 +177,21 @@ unittest
...
+     "DDDC3C5D5CB1018DDAD191A1AE3D80DC1AF350B535CFE9F1732EF321AE0F15709F08168648E88686"
+    ~"22DE1DE55525650E9A9BB02605BBA9DA11173A6D491853F50100000000000000DD9AAF1064DA6746"
+    ~"2DC1CC496E0374084058403429A1F74B72663F6CCA35B80AECD85063383D372F9DBCD770EF654B96"
+    ~"99F8C8801E6FD8041E9F517DF0B7CD2DE8D87AE11BDB8A861F7B0BD6EE6D7D37657FDD47CF8ED340"
+    ~"CDE27387366F9A64ABC91788162EFE5B3C9DAA4104113445994C7F0497199FF442E88F78C582A2CF"
+    ~"0000000066D002DC3D4F02352BF1E9478501CCF78FEA23AAFBAD4652630297C030A2F53E028BE2CD"
+    ~"CEEBADCCAD09908A375495919B49F263127B9BCC06DC8519A51EF80D005A6202000000009D0238E0"
+    ~"A171400BC6D68A9D9B316ACD5109649113A05C284F4296D2B30122F5";
+
+    assert(cast(string)data.toHexString() == serializeData);
```
Geod24: Why the `cast` ?

Geod24 (commented):
Pretty good, just some nits

source/agora/common/Block.d
TrustHenry: added the space.

source/agora/common/Block.d
TrustHenry: I modified it according to the style guide.

source/agora/common/Transaction.d
TrustHenry: I modified it according to the style guide.

source/agora/common/Transaction.d
TrustHenry: I modified it according to the style guide.

source/agora/common/Transaction.d
TrustHenry: I modified it according to the style guide.

source/agora/common/crypto/Key.d
TrustHenry: I modified it according to the style guide.

source/agora/node/BlockSerialize.d
TrustHenry: Apply as Suggested

source/agora/node/BlockSerialize.d
TrustHenry: use @safe & pure

source/agora/node/BlockSerialize.d
TrustHenry: use all pure

source/agora/node/BlockSerialize.d
TrustHenry: add the overload Type Hash,  Signature

source/agora/node/BlockSerialize.d
TrustHenry: ^

source/agora/node/Ledger.d
TrustHenry: It was removed.

Geod24 (approved):
LGTM

PR #106, opened 2019-07-23
Implement simple request retrying

AndrejMitrovic:
Blocked by https://github.com/bpfkorea/agora/pull/117

### DIFF
--- a/source/agora/network/NetworkClient.d
+++ b/source/agora/network/NetworkClient.d
@@ -14,6 +14,7 @@
 module agora.network.NetworkClient;
 
 import agora.common.API;
+import agora.common.BanManager;
 import agora.common.Block;
 import agora.common.Config;
 import agora.common.crypto.Key;
@@ -42,6 +43,9 @@ class NetworkClient
     /// so whatever implements `API` should be handling this
     private const Duration retry_delay;
 
+    /// Max request retries before a request is considered failed
+    private const size_t max_retries;
+
     /// Task manager
     private TaskManager taskman;
 
@@ -50,6 +54,9 @@ class NetworkClient
 
     /// The key of the node as retrieved by getPublicKey()
     public PublicKey key;
+
+    /// Reusable exception
+    private Exception exception;
 
 
     /***************************************************************************
@@ -61,15 +68,20 @@ class NetworkClient
             address = used for logging and querying by external code
             api = the API to issue the requests with
             retry = the amout to wait between retrying failed requests
+            max_retries = max number of times a failed request should be retried
 
     ***************************************************************************/
 
-    public this (TaskManager taskman, Address address, API api, Duration retry)
+    public this (TaskManager taskman, Address address, API api, Duration retry,
+        size_t max_retries)
     {
         this.taskman = taskman;
         this.address = address;
         this.api = api;
         this.retry_delay = retry;
+        this.max_retries = max_retries;
+        this.exception = new Exception(
+            format("Request failure after %s attempts", max_retries));
     }
 
     /***************************************************************************
@@ -81,40 +93,14 @@ class NetworkClient
 
         Note that it is currently blocking until handshake is considered complete.
 
+        Throws:
+            Exception if the handshake did not complete.
+
     ***************************************************************************/
 
     public void handshake ()
     {
-        while (!this.getPublicKey())
-        {
-            logInfo("[%s] Couldn't retrieve public key. Will retry in %s..",
-                this.address, this.retry_delay);
-            this.taskman.wait(this.retry_delay);
-        }
-    }
-
-    /***************************************************************************
-
-        Get the public key of this node, stored in the
-        `key` field if the request succeeded.
-
-        Returns:
-            true if the request succeeded
-
-    ***************************************************************************/
-
-    private bool getPublicKey ()
-    {
-        try
-        {
-            this.key = this.api.getPublicKey();
-            return true;
-        }
-        catch (Exception ex)
-        {
-            logError(ex.message);
-            return false;
-        }
+        this.key = this.attemptRequest(this.api.getPublicKey(), this.exception);
     }
 
     /***************************************************************************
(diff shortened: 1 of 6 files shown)


source/agora/network/NetworkClient.d
```diff
@@ -208,18 +179,73 @@ class NetworkClient
...
-            logError(ex.message);
-            return null;
+            try
+            {
+                return dg();
+            }
+            catch (Exception ex)
+            {
+                this.banman.onFailedRequest(this.address);
+                logError(ex.message);
```
AndrejMitrovic: This should probably be configurable, maybe through a parameter.

source/agora/network/NetworkClient.d
```diff
@@ -208,18 +179,73 @@ class NetworkClient
...
+            throw ex;
+
+        // ex is null, failure is ignored
+        static if (!is(T == void))
+            return T.init;
+    }
+}
+
+/// Thrown when a request fails after a number of retries
+public class RequestFailure : Exception
```
AndrejMitrovic: Question: continue using this or use an Exception?

I would like to avoid programming errors to throw exceptions and be caught as if they were request exceptions. 

For example, if I have a call like `string stuff = format("%s %s", 1);`, this will throw an Exception. But it's really a programmer bug, not a "failed request". See what I mean?

AndrejMitrovic:
I think this could use some networking tests.

source/agora/network/NetworkClient.d
```diff
@@ -61,15 +68,20 @@ class NetworkClient
             address = used for logging and querying by external code
             api = the API to issue the requests with
             retry = the amout to wait between retrying failed requests
+            max_retries = max number of times a failed request should be retried
 
     ***************************************************************************/
 
-    public this (TaskManager taskman, Address address, API api, Duration retry)
+    public this (TaskManager taskman, Address address, API api, Duration retry,
+        size_t max_retries )
```
Geod24:
```suggestion
        size_t max_retries)
```

source/agora/network/NetworkClient.d
```diff
@@ -208,18 +176,58 @@ class NetworkClient
...
+            T = the type of delegate
+            dg = the delegate to call
+            ex = the exception to throw if all attempts fail (if not null)
+
+        Returns:
+            the return value of dg(), which may be void
+
+    ***************************************************************************/
+
+    private T attemptRequest (LogLevel log_level = LogLevel.error, T)(
```
Geod24, line 209: Does log level has to be a template parameter ?

AndrejMitrovic:
Fixed styling. The log level is a CT parameter in vibe.d. Well I could do a switch/case here I guess but I don't know if that improves anything.

AndrejMitrovic:
Added network unittests.

AndrejMitrovic:
There are some weird timing issues with this. I'm investigating.

AndrejMitrovic:
> There are some weird timing issues with this. I'm investigating.

This PR will go green after its dependencies are resolved. I'll rebase it when those are ready.

AndrejMitrovic:
Rebased.

source/agora/network/NetworkClient.d
AndrejMitrovic: Answered IRL before, but yes because `log` takes it as a CT param, while `doDebug` which takes it as a runtime parameter is private.

source/agora/network/NetworkManager.d
```diff
@@ -241,28 +241,46 @@ public class NetworkManager
...
+        {
+            try
+            {
+                node.handshake();
+                this.onNodeConnected(node);
+                break;
+            }
+            catch (Exception ex)
+            {
+               // keep trying until node is connected
```
AndrejMitrovic, line 257: Btw, this part is changed in the node banning support code. For now it has the same semantics as before, keep trying to handshake.

AndrejMitrovic: As in, it's changed in one of the other PRs.

source/agora/network/NetworkClient.d
```diff
@@ -208,18 +176,58 @@ class NetworkClient
...
-            return null;
+            try
+            {
+                return dg();
+            }
+            catch (Exception ex)
+            {
+                log!log_level(ex.message);
+
+                if (idx + 1 != this.max_retries)  // don't wait after final attempt
```
AndrejMitrovic: I'll change this to `idx + 1 < this.max_retries` to make it more obvious.

AndrejMitrovic: Fixed.

source/agora/network/NetworkClient.d
```diff
@@ -208,18 +176,57 @@ class NetworkClient
...
+            dg = the delegate to call
+            ex = the exception to throw if all attempts fail (if not null)
+
+        Returns:
+            the return value of dg(), which may be void
+
+    ***************************************************************************/
+
+    private T attemptRequest (LogLevel log_level = LogLevel.error, T)(
+        lazy T dg, Exception ex = null)
```
Geod24: If `ex is null` means "ignore the failure", then the default parameter should not be there, because we don't want the default to be "ignore failure".

source/agora/test/NetworkClient.d
```diff
@@ -0,0 +1,68 @@
...
+import agora.common.Hash;
+import agora.common.Transaction;
+import agora.consensus.Genesis;
+import agora.test.Base;
+
+/// test retrying requests after failure
+unittest
+{
+    import core.thread;
+    import vibe.core.log;
```
AndrejMitrovic, line 29: poor import ordering

source/agora/network/NetworkManager.d
```diff
@@ -241,28 +241,46 @@ public class NetworkManager
...
+                // attempt later
             }
 
+            // if it's incomplete give the client some time to connect
+            // with other peers and try again later
             logInfo("[%s] (%s): Peer info is incomplete. Retrying in %s..",
-                node.address, node.key, this.node_config.retry_delay);
-            this.taskman.wait(this.node_config.retry_delay.msecs);
+                node.address, node.key, 30.seconds);
+            this.taskman.wait(30.seconds);
```
Geod24: Is that on purpose ??

AndrejMitrovic: Well yes, there is a comment right above it.

AndrejMitrovic: The idea is, let the node have some time too boot. I don't know if `retry_delay` is enough, because that value should be kept fairly low (a few seconds).

Geod24: Some time /= 30 seconds. But IIRC it already retry internally and that's why the delay is so high ?

AndrejMitrovic: For things like `putTransaction` I've set it to follow the configured `retry_delay`. While for the periodic `getNetworkInfo` I feel like we should wait longer for the target node to establish a connection with other nodes, before asking it for its peers again.

I'm not sure, maybe I was overthinking it and we should just use `retry_delay`? `getNetworkInfo` could be pretty expensive to call, it returns an array of all known peers.

AndrejMitrovic:
Fixed exception and import ordering.

source/agora/network/NetworkManager.d
Geod24: Yeah I'd stay let's stick to retry_delay for now, there's a few optimization we'll need to do in the future anyways.

AndrejMitrovic:
Fixed delay.

Geod24 (approved):

PR #108, opened 2019-07-23
Add some BanManager improvements

AndrejMitrovic:
This will be required for my next PR which actually implements banning. I wanted to split it up into two to make it more digestible.

### DIFF
--- a/source/agora/common/BanManager.d
+++ b/source/agora/common/BanManager.d
@@ -86,8 +86,22 @@ public class BanManager
         if (status.fail_count >= this.config.max_failed_requests)
         {
             status.fail_count = 0;
-            status.banned_until = this.getCurTime() + this.config.ban_duration;
+            this.ban(address);
         }
+    }
+
+    /***************************************************************************
+
+        Manually ban an address using the configured ban time in the ban config.
+
+        Params:
+            address = the address to ban
+
+    ***************************************************************************/
+
+    public void ban (Address address)
+    {
+        this.banFor(address, this.config.ban_duration);
     }
 
     /***************************************************************************
@@ -243,4 +257,8 @@ unittest
     assert(banman.isBanned("node-1"));
     banman.time++;
     assert(!banman.isBanned("node-1"));
+
+    banman.time = 0;
+    banman.ban("node-2");  // use default ban time
+    assert(banman.getUnbanTime("node-2") == 86400);
 }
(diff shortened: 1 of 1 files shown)


source/agora/common/BanManager.d
```diff
@@ -86,7 +86,7 @@ public class BanManager
         if (status.fail_count >= this.config.max_failed_requests)
         {
             status.fail_count = 0;
-            status.banned_until = this.getCurTime() + this.config.ban_duration;
+            this.banFor(address, this.config.ban_duration);
```
Geod24: Why not `ban` ?

source/agora/common/BanManager.d
```diff
@@ -210,8 +210,9 @@ public class BanManager
 
     ***************************************************************************/
 
-    private Status* get (Address address)
+    private Status* get (Address address) @trusted nothrow pure
     {
+        scope(failure) assert(0);  // it will never throw
```
Geod24, line 215: `require` is not `nothrow` ??

AndrejMitrovic: It creates an item if it doesn't exist. The naming of `require` is very counter-intuitive.

source/agora/common/BanManager.d
AndrejMitrovic: Good eye.

Geod24 (approved):
Side note: What happens if `ban` or `banFor` is called on an existing banned node ? Does `banFor` takes the max of both bans ?

source/agora/common/BanManager.d
Geod24: Ping

AndrejMitrovic:
Updated.

> Side note: What happens if ban or banFor is called on an existing banned node ? Does banFor takes the max of both bans ?

Ah very good question.. Should it add to the existing ban time, or use the max? Currently it just overrides whatever was set.

AndrejMitrovic:
Maybe I should simplify this:

```
public void banForever() /// e.g. known bad nodes
private void ban()  // called internally in onFailedRequest
```

Geod24:
No I don't think we should ever ban forever. IPs are dynamic.
As for the ban time, a conservative approach would be to implicitly extend and explicitly reduce.
In other word, if you ban an already ban client, unless you do it via a specially-designed functions, you only ever extend.
And at the moment I don't see a reason to extend / reduce bans ?

Geod24 (approved):

PR #111, opened 2019-07-24
Fix check in the Ledger unittest

AndrejMitrovic:
I've just noticed this part missing, as we periodically poll for the node height until the desired height is reached and then exit early. But if the same height isn't reached, we should assert.

### DIFF
--- a/source/agora/test/Ledger.d
+++ b/source/agora/test/Ledger.d
@@ -187,4 +187,6 @@ unittest
         // let them do catch-up after boot
         Thread.sleep(100.msecs);
     }
+
+    assert(0, "Nodes do not contain the same blocks");
 }


source/agora/test/Ledger.d
```diff
@@ -178,13 +178,17 @@ unittest
...
-            return;
+        blocks_equal = containSameBlocks(nodes, 10);
+        if (blocks_equal)
+            break;
 
         // let them do catch-up after boot
         Thread.sleep(100.msecs);
     }
+
+    assert(blocks_equal, "Nodes do not contain the same blocks");
```
Geod24: This could just be `assert(0)` if you left the `return` instead of `break`

Geod24: Ping

AndrejMitrovic: Alright.

AndrejMitrovic:
Fixed.

Geod24 (approved):

PR #113, opened 2019-07-25
Refactor test-code for multi-transaction blocks

AndrejMitrovic:
This PR is big but it is atomic.

There were several issues with https://github.com/bpfkorea/agora/commit/3226a71f2bccbc695135b418b70780fe67925d1b:

- The unittests were fixed by changing how getChainedTransactions() works, but what it ended up doing was creating a bunch of transaction which spent the same `Output`. Note: we still don't have checks against spending the same output multiple times.
- The test code implied it knows how many transactions there are in a block. So there are various magic numbers being used everywhere like `period = 8`.

Upon closer look I've realized that the existing tests are a bit messy. For example, the network test for the gossip protocol was testing transaction propagation **and** block creation. The block creation tests belong in the Ledger module, not in the GossipProtocol.

I've made all of the unittest code support any number of transactions in the block (although there is now a defined constant in the form of `Block.TxsInBlock`).

Note that for example, to create 10 transactions, those 10 transactions must refer to a UTXO in the blockchain. Therefore there must exist 10 Outputs in the Blockchain already. Currently transactions which try to spend UTXOs which are not in the Blockchain will be rejected. (I believe these are called self-referential transactions, or ones which refer to other ones in-flight and not yet in the blockchain).


AndrejMitrovic:
This also seems to fix my request retrying PR (at least on my machine), so that's very good news.

AndrejMitrovic:
Squashed the first two commits, and created a bunch of more atomic ones as well.

Geod24:
It's red :'(

source/agora/common/Transaction.d
```diff
@@ -259,48 +259,26 @@ public Transaction newCoinbaseTX (PublicKey address, Amount value = 0)
 *******************************************************************************/
 
 version (unittest)
-public Transaction[] getChainedTransactions (Transaction root, size_t count, KeyPair key_pair, size_t period = 1)
+public Transaction[] getChainedTransactions (Transaction root, size_t count,
```
Geod24: Commit fed184f52e4d4b1e19811c9cc23d6f78e588ae08 "Revert changes to getChainedTransactions()".
You delete a bunch of code but don't change any test. Does that mean that code is unused, or do the tests fail ?

source/agora/node/Ledger.d
```diff
@@ -66,26 +66,28 @@ public class Ledger
     {
         if (!tx.verify(&this.findOutput))
             return false;
+
         this.storage ~= tx;
-        if (this.storage.length >= 8)
+        if (this.storage.length == Block.TxsInBlock)
```
Geod24, line 71: Looks like the commit is doing much more than "Encode the block size in the Block and use it"

source/agora/node/Ledger.d
```diff
@@ -264,13 +267,12 @@ unittest
     // each tx currently creates one block
     void genTransactions (size_t count)
     {
-        auto txes = getChainedTransactions(last_tx, count, gen_key_pair, 1);
```
Geod24, line 267: This commit is clearly not atomic then

source/agora/common/Transaction.d
```diff
@@ -242,6 +242,28 @@ public Transaction newCoinbaseTX (PublicKey address, Amount value = 0)
...
 }
 
+/*******************************************************************************
+
+    Returns:
+        the genesis transaction
+
+*******************************************************************************/
+
+public Transaction getGenesisTx ()
```
Geod24, line 252: `version (unittest)`

Geod24 (changes requested):
Okay that's a lot of changes. You might want to break it into a few PRs so we can move fast on this. Some commit are clearly not atomic though.

source/agora/common/Transaction.d
AndrejMitrovic: Yes this is the non-atomic part. But if I want to make it atomic then I have to squash 3 commits together, which kind of sucks.

I wish there was a way to make "sub-atomic" commits.

AndrejMitrovic:
I have a suspicion the failure reason is because the transactions show up out-of-order, so they're not placed in the block in the same order in all of the nodes. The code was actually testing whether the blocks are the same, not just that the block height is the same. I'm looking into it.

And yeah I'll split this up then.

AndrejMitrovic:
The bug was in my code. Fixing it now.

Geod24:
Plz rebase

AndrejMitrovic:
Replaced by simpler PRs.

PR #112, opened 2019-07-24
Implemented API for Merkle proof

MichaelKim20:
Implemented to enable the Merkle proof in the network.
Register API Interface
Use Merkle Proof

MerkleTree_Proof

PR #117, opened 2019-07-26
Fix tests so they're reliable against block size changes

AndrejMitrovic:
This should be merged after ~https://github.com/bpfkorea/agora/pull/113~ https://github.com/bpfkorea/agora/pull/112, so I can fix up the unittests in that code as well.

It's hard to split up the second commit into multiple commits, because the change in `getChainedTransactions` affects most unittests (and it was broken before, referencing the same Input multiple times, and hardcoded to 8 transactions per block).

### DIFF
--- a/source/agora/common/Transaction.d
+++ b/source/agora/common/Transaction.d
@@ -23,6 +23,7 @@ import agora.common.Data;
 import agora.common.Deserializer;
 import agora.common.Hash;
 import agora.common.Serializer;
+import agora.consensus.Genesis;
 
 import std.algorithm;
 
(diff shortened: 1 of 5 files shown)


Geod24:
> This should be merged after #113, so I can fix up the unittests in that code as well.

You closed #113 ? 😕

AndrejMitrovic:
I linked to the wrong PR.

AndrejMitrovic:
Rebased. The diff looks involved, but it's hard to split this up while still making it atomic.

AndrejMitrovic:
Hmm I should add a unittest to the ~`getBlocksFrom`~ `getChainedTransactions` function.

AndrejMitrovic:
Err I meant `getChainedTransactions`

AndrejMitrovic:
Updated with a unittest.

source/agora/common/Transaction.d
```diff
@@ -306,66 +307,113 @@ public Transaction newCoinbaseTX (PublicKey address, Amount value = 0)
...
+        prev_txs = the previous transactions to refer to
         count = the number of transactions to create
         key_pair = the key pair used to sign transactions and to send
                    the output to
 
 *******************************************************************************/
 
 version (unittest)
-public Transaction[] getChainedTransactions (Transaction root, size_t count, KeyPair key_pair, size_t period = 1)
+public Transaction[] getChainedTransactions (KeyPair key_pair, size_t count,
```
Geod24: Since `count` *always* needs to be a multiple of `Block.TxsInBlock`, why not simplify the code and multiply `code` by `Block.TxsInBlock` ?
Also I don't like the `get` verb here. It's not "getting" anything, it is "make
ing, "build"ing, "create"ing, etc...

AndrejMitrovic: Yes, so maybe `makeBlockTransactions`, and the `count` would specify the number of blocks?

AndrejMitrovic: Ah actually it still returns transactions, so I'll just name it `makeTransactions` and document it better.

Geod24: I had something like `makeTransactionsToFillBlock` in mind, only it's too long.

AndrejMitrovic:
Rebased.

Geod24 (approved):

PR #120, opened 2019-07-28
Function to save ledger to file

TrustHenry:
Add the ability to save and load the ledger into a file using a serializer delegate function and a deserializer delegate function.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -17,12 +17,17 @@ import agora.common.API;
 import agora.common.Block;
 import agora.common.Data;
 import agora.common.Hash;
+import agora.common.Serializer;
 import agora.common.Transaction;
 import agora.consensus.Genesis;
 
 import vibe.core.log;
 
 import std.algorithm;
+import std.exception;
+import std.file;
+import std.format;
+import std.stdio;
 
 /// Ditto
 public class Ledger
@@ -228,6 +233,49 @@ public class Ledger
             return block.getMerklePath(index);
         else
             return null;
+    }
+
+    /***************************************************************************
+
+        Write a block to the ledger file on disk.
+
+        Params:
+            ledger_file = Instance of file for ledgerFile
+            index_file = Instance of file for indexFile
+            block = Type of struct to writeFile
+
+    ***************************************************************************/
+
+    public void writeBlockFile (File ledger_file, File index_file, ref const Block block) @safe
+    {
+        size_t block_size = 0;
+
+        scope SerializeDg dg = (scope const (ubyte[]) data) nothrow @safe
+        {
+            try
+            {
+                ledger_file.rawWrite(data);
+                block_size += data.length;
+            }
+            catch (Exception ex)
+            {
+                logError("writeBlockFile: %s", () @trusted { return ex.message;}());
+            }
+        };
+        serializePart(block, dg);
+
+        scope SerializeDg index_dg = (scope const (ubyte[]) data) nothrow @safe
+        {
+            try
+            {
+                index_file.rawWrite(data);
+            }
+            catch (Exception ex)
+            {
+                logError("writeBlockFile: %s", () @trusted { return ex.message;}());
+            }
+        };
+        serializePart(block_size, index_dg);
     }
 }
 


Geod24:
Looks like you need to rebase on `v0.x.x`

TrustHenry:
Rebase done :)

Geod24:
Screen Shot 2019-07-28 at 23 56 19

TrustHenry:
Have dependencies of #121.
remove serializer @pure

source/agora/node/Ledger.d
```diff
@@ -16,12 +16,16 @@ module agora.node.Ledger;
 import agora.common.API;
 import agora.common.Block;
 import agora.common.Data;
+import agora.common.Deserializer;
+import agora.common.Serializer;
```
AndrejMitrovic: This import goes after `Hash`, alphabetic ordering.

source/agora/node/Ledger.d
```diff
@@ -16,12 +16,16 @@ module agora.node.Ledger;
...
 import agora.common.Hash;
 import agora.common.Transaction;
 
 import vibe.core.log;
 
 import std.algorithm;
+import std.stdio;
```
AndrejMitrovic: exception before stdio

AndrejMitrovic:
Is this PR ready? Because it looks incomplete. Exceptions are caught and not even logged.

source/agora/node/Ledger.d
```diff
@@ -229,6 +233,84 @@ public class Ledger
...
+        size_t offset = 0;
+
+        DeserializeDg dg = (size) nothrow @safe
+        {
+            ubyte [] res = data[offset .. offset + size];
+            offset += size;
+            return res;
+        };
+
+        bool eof = true;
```
AndrejMitrovic: Shouldn't this be false by default? `eof` stands for `end of file`. You only reach the end of file at the very end.

source/agora/node/Ledger.d
```diff
@@ -229,6 +233,84 @@ public class Ledger
...
+            dg  = Deserialization delegate, when this struct is a nested struct
+
+        Return:
+            Return Block[]
+
+    ***************************************************************************/
+
+    public Block[] readLedgerFile () @safe
+    {
+        scope ubyte[] data;
```
AndrejMitrovic: `scope` has no meaning here.

source/agora/node/Ledger.d
```diff
@@ -229,6 +233,84 @@ public class Ledger
...
+        try
+        {
+            ubyte[] buffer;
+            buffer.length = 1024;
+            auto file = File("ledger.dat", "r"); // TODO move to config.yaml
+            data = file.rawRead(buffer);
+        }
+        catch (ErrnoException ex)
+        {
+            //logError("readLedgerFile: %s", ex.message);
```
AndrejMitrovic: You are ignoring the exception here.

Then in the delegate `dg` below, you will try to slice `data[offset .. offset + size];`, which will throw another exception because `data` is empty and you try to index more data than exists.

This will make the node shut down on startup when it tries to read the blocks.

source/agora/node/Ledger.d
TrustHenry: You'r Right :)

source/agora/node/Ledger.d
TrustHenry: You'r Right :)

source/agora/node/Ledger.d
TrustHenry: remove it `scop`

source/agora/node/Ledger.d
TrustHenry: Change the meaning

source/agora/node/Ledger.d
```diff
@@ -230,6 +234,84 @@ public class Ledger
...
+            ubyte [] res = data[offset .. offset + size];
+            offset += size;
+            return res;
+        };
+
+        bool eof = false;
+        int cnt = 0;
+        while(!eof){
+            block[cnt].deserialize(dg);
+            if(offset >= data.sizeof){
```
Geod24: Did you test this ? `data.sizeof` is `ubyte[].sizeof` which is 16 on 64 bits machines. I doubt this works.

source/agora/node/Ledger.d
```diff
@@ -230,6 +234,84 @@ public class Ledger
...
     }
+
+    /***************************************************************************
+
+        Write the block on the ledgerFile.
+
+        Params:
+            T = Type of struct to writeFile
+            record = Instance of `T` to serialize
+            dg  = Serialization delegate, when this struct is a nested struct
```
AndrejMitrovic: There is no `dg` here.

source/agora/node/Ledger.d
```diff
@@ -230,6 +234,84 @@ public class Ledger
         else
             return null;
     }
+
+    /***************************************************************************
+
+        Write the block on the ledgerFile.
```
```suggestion
        Write a block to the ledger file on disk.
```

source/agora/node/Ledger.d
```diff
@@ -230,6 +234,84 @@ public class Ledger
...
+
+        Params:
+            T = Type of struct to writeFile
+            record = Instance of `T` to serialize
+            dg  = Serialization delegate, when this struct is a nested struct
+
+    ***************************************************************************/
+
+    public void writeBlockFile (T) (auto ref T record) @safe
```
AndrejMitrovic: If this only writes blocks, why is it templated? You could just write `( Block record)`

source/agora/node/Ledger.d
```diff
@@ -230,6 +234,84 @@ public class Ledger
...
+    if (is(T == struct))
+    {
+        auto fd = File("ledger.dat", "w"); // TODO move to config.yaml
```
AndrejMitrovic: Shouldn't this open the file for appending, not writing? Because otherwise you will keep writing one block and overwriting it in this function each time it's called with a block.

source/agora/node/Ledger.d
TrustHenry: length -> sizeof -> length,

source/agora/node/Ledger.d
TrustHenry: apply. :)

source/agora/node/Ledger.d
TrustHenry: apply :)

TrustHenry: public void writeBlockFile (ref Block block) @safe

source/agora/node/Ledger.d
TrustHenry: changed "a" / append access
if the file does not exist, it is created as empty
if the file already exists, its contents are preserved and it is opened to be written at the end

TrustHenry:
The file read section is being modified.

TrustHenry:
[without]delete Function readLedgerFile.  
I'll add it in the next PR.

AndrejMitrovic:
You committed `ledger.dat` in your PR by mistake. :)

TrustHenry:
add param config data_dir in Ledger Ctor

source/agora/node/Ledger.d
```diff
@@ -42,11 +44,16 @@ public class Ledger
...
+    /// ledger data dir
+    private string data_dir;
+
     /// Ctor
-    public this ()
+    public this (const string data_dir = ".cache")
     {
         auto block = getGenesisBlock();
         this.addNewBlock(block);
+        if(this.data_dir != data_dir)
```
AndrejMitrovic: You don't need this check, you are in the constructor, therefore you know for sure `this.data_dir` is not initialized.

source/agora/node/Ledger.d
```diff
@@ -42,11 +44,16 @@ public class Ledger
     /// Temporary storage where transactions are stored until blocks are created.
     private Transaction[] storage;
 
+    /// ledger data dir
+    private string data_dir;
```
AndrejMitrovic: `const`

source/agora/node/Ledger.d
```diff
@@ -42,11 +44,16 @@ public class Ledger
...
+
     /// Ctor
-    public this ()
+    public this (const string data_dir = ".cache")
```
AndrejMitrovic: There shouldn't be a default value here.

source/agora/node/Ledger.d
```diff
@@ -247,7 +254,13 @@ public class Ledger
 
     public void writeBlockFile (ref Block block) @safe
     {
-        auto fd = File("ledger.dat", "a"); // TODO move to config.yaml
+        auto dir = this.data_dir;
+        // Make sure we're in the right directory
+        if (!std.file.exists(dir) || !std.file.isDir(dir))
+        std.file.mkdir(dir);
```
AndrejMitrovic: This should probably be moved somewhere in `main()` or in the `Node` constructor. The data directory should be created once, some time after parsing the `data_dir` config option.

source/agora/node/Ledger.d
```diff
@@ -247,7 +254,13 @@ public class Ledger
 
     public void writeBlockFile (ref Block block) @safe
     {
-        auto fd = File("ledger.dat", "a"); // TODO move to config.yaml
+        auto dir = this.data_dir;
+        // Make sure we're in the right directory
+        if (!std.file.exists(dir) || !std.file.isDir(dir))
```
AndrejMitrovic: And this should assert that it's a directory.

source/agora/node/Ledger.d
TrustHenry: delete it

source/agora/node/Ledger.d
```diff
@@ -247,7 +254,13 @@ public class Ledger
 
     public void writeBlockFile (ref Block block) @safe
```
AndrejMitrovic: This function should probably take the File as a parameter. Opening a file costs performance, it should only be opened once, and this method should write to the file using the existing File instance.

source/agora/node/Ledger.d
TrustHenry: delete it

source/agora/node/Ledger.d
TrustHenry: add it

source/agora/node/Ledger.d
TrustHenry: In many unittest, I will put the default value in the part where I create the bedget.

source/agora/node/Ledger.d
TrustHenry: add param (ref File fd, ref Block block)

source/agora/node/Ledger.d
TrustHenry: add assert directory

TrustHenry:
I created an index file that stores the size of the block.

AndrejMitrovic:
> I created an index file that stores the size of the block.

How will that help you determine at which byte offset is the block?

Let's say I want to read the block at block height 50. So I go to the index file, and it says:

| Block Height   |      Block Length      |  
|----------------|:----------------------:|
| 1             |    100                 |
| 2             |    120                 |
| 3             |    110                 |
| ...            |                        |
| 50            |    130                 |

So I have to read 130 bytes for block height 50. But at which byte offset in the blockchain file?

TrustHenry:
File division is likely in the next sprint.
Maybe we need fileIndex.

TrustHenry:
atomic commit rebase Done.

AndrejMitrovic:
You still have `ledger.dat` in the PR.

source/agora/node/Ledger.d
```diff
@@ -229,6 +235,57 @@ public class Ledger
...
+            }
+            catch (ErrnoException ex)
+            {
+                //logError("writeBlockFile: %s", ex.message);
+            }
+        };
+        serializePart(block, dg);
+
+         string file_path = format("%s/%s", dir, "index.dat");
+        scope File index_fd = File(file_path, "a");
```
AndrejMitrovic: `scope` has no meaning here.

It is only useful when you do `scope object = new Class()`, which will allocate the class no the stack.

source/agora/node/Ledger.d
```diff
@@ -229,6 +235,57 @@ public class Ledger
...
+
+        Write a block to the ledger file on disk.
+
+        Params:
+            fd = Instance of LedgerFile
+            block = Type of struct to writeFile
+
+    ***************************************************************************/
+
+    public void writeBlockFile (ref File fd, ref Block block) @trusted
```
AndrejMitrovic: You do not have to pass the file by reference. The `File` struct has an internal reference counter, which means it acts as a reference. So you can pass it by value.

Also could you rename it to `File file;`? It makes it a bit clearer, it's not just a file descriptor but a wrapper around it.

source/agora/node/Ledger.d
AndrejMitrovic: Now that you've added an assertion that it exists, this whole part should be changed into:

```
assert(dir.exists && std.file.isDir(dir));
```

There is no need to do `mkdir` if it already exists.

source/agora/node/Ledger.d
TrustHenry: Fixed

source/agora/node/Ledger.d
TrustHenry: Fixed

TrustHenry:
> You still have `ledger.dat` in the PR.

remove it

source/agora/node/Ledger.d
```diff
@@ -229,14 +239,65 @@ public class Ledger
...
+        std.file.mkdir(dir);
+
+         scope SerializeDg dg = (scope const (ubyte[]) data) @safe
+        {
+            try
+            {
+                file.rawWrite(data);
+                block_size += data.length;
+            }
+            catch (ErrnoException ex)
```
AndrejMitrovic: If you don't want this function to throw, then you have to catch `Exception`, not just `ErrnoException`. Then, you can mark it as `nothrow`. If you do that, you don't have to remove `nothrow` from the serialization functions. So something like:

```diff
+++ b/source/agora/node/Ledger.d
@@ -260,14 +260,14 @@ public class Ledger
         if (!std.file.exists(dir) || !std.file.isDir(dir))
         std.file.mkdir(dir);

-         scope SerializeDg dg = (scope const (ubyte[]) data) @safe
+         scope SerializeDg dg = (scope const (ubyte[]) data) @safe nothrow
         {
             try
             {
                 file.rawWrite(data);
                 block_size += data.length;
             }
-            catch (ErrnoException ex)
+            catch (Exception ex)
             {
                 //logError("writeBlockFile: %s", ex.message);
             }
@@ -277,13 +277,13 @@ public class Ledger
         string file_path = format("%s/%s", dir, "index.dat");
         File index_file = File(file_path, "a");

-         scope SerializeDg index_dg = (scope const (ubyte[]) data) @safe
+         scope SerializeDg index_dg = (scope const (ubyte[]) data) @safe nothrow
         {
             try
             {
                 index_file.rawWrite(data);
             }
-            catch (ErrnoException ex)
+            catch (Exception ex)
             {
                 //logError("writeIndexFile: %s", ex.message);
             }
```

source/agora/node/Ledger.d
TrustHenry: Fixed.

source/agora/node/Ledger.d
TrustHenry: only use Exception.

TrustHenry:
@Geod24 @AndrejMitrovic ping

source/agora/node/Ledger.d
```diff
@@ -229,14 +239,63 @@ public class Ledger
...
+
+        Write a block to the ledger file on disk.
+
+        Params:
+            fd = Instance of LedgerFile
+            block = Type of struct to writeFile
+
+    ***************************************************************************/
+
+    public void writeBlockFile (File file, ref Block block) @trusted
```
Geod24: Please don't use `@trusted` on a function, especially this big.
Mark it `@safe` and use `@trusted` delegates (E.g. `int x = () @trusted { return mySystemFunc(); }();`) when necessary.

source/agora/node/Ledger.d
```diff
@@ -229,14 +239,63 @@ public class Ledger
         else
             return null;
     }
+
+    /***************************************************************************
+
+        Write a block to the ledger file on disk.
+
+        Params:
+            fd = Instance of LedgerFile
```
Geod24: What is `LedgerFile` ? It sounds like a type, but I can't find it anywhere.

source/agora/node/Ledger.d
```diff
@@ -229,14 +239,63 @@ public class Ledger
...
+            {
+                //logError("writeBlockFile: %s", ex.message);
+            }
+        };
+        serializePart(block, dg);
+
+        string file_path = format("%s/%s", dir, "index.dat");
+        File index_file = File(file_path, "a");
+
+         scope SerializeDg index_dg = (scope const (ubyte[]) data) nothrow @safe
```
Geod24: Indentation is off. Also no space after `const`

source/agora/node/Ledger.d
```diff
@@ -229,14 +239,63 @@ public class Ledger
...
+        scope SerializeDg dg = (scope const (ubyte[]) data) nothrow @safe
+        {
+            try
+            {
+                file.rawWrite(data);
+                block_size += data.length;
+            }
+            catch (Exception ex)
+            {
+                //logError("writeBlockFile: %s", ex.message);
```
Geod24: Errors should never be silently ignored

source/agora/node/Ledger.d
```diff
@@ -229,6 +235,55 @@ public class Ledger
...
+
+    ***************************************************************************/
+
+    public void writeBlockFile (ref File fd, ref Block block) @trusted
+    {
+        const string dir = this.data_dir;
+        size_t block_size = 0;
+
+         // sure we're in the right directory
+        assert(dir.exists && std.file.isDir(dir));
```
Geod24: Why do you do IO here if you have a `File` instance ?

source/agora/node/Ledger.d
```diff
@@ -229,6 +235,55 @@ public class Ledger
...
+                block_size += data.length;
+            }
+            catch (Exception ex)
+            {
+                //logError("writeBlockFile: %s", ex.message);
+            }
+        };
+        serializePart(block, dg);
+
+         string file_path = format("%s/%s", dir, "index.dat");
```
Geod24: Wrong indentation

source/agora/node/Ledger.d
Geod24: Your fix is not in the right commit.

source/agora/node/Ledger.d
```diff
@@ -229,6 +235,55 @@ public class Ledger
...
+
+        Params:
+            fd = Instance of LedgerFile
+            block = Type of struct to writeFile
+
+    ***************************************************************************/
+
+    public void writeBlockFile (ref File fd, ref Block block) @trusted
+    {
+        const string dir = this.data_dir;
```
Geod24: This commit is not atomic: `data_dir` does not exist until the next commit.

source/agora/node/Ledger.d
```diff
@@ -258,7 +262,7 @@ public class Ledger
         {
             try
             {
-                fd.rawWrite(data);
+                file.rawWrite(data);
```
Geod24: This commit is not atomic as it changes previously introduced code.

source/agora/node/Ledger.d
```diff
@@ -268,14 +272,14 @@ public class Ledger
         };
         serializePart(block, dg);
 
-         string file_path = format("%s/%s", dir, "index.dat");
-        scope File index_fd = File(file_path, "a");
+        string file_path = format("%s/%s", dir, "index.dat");
+        File index_file = File(file_path, "a");
```
Geod24: It does not make sense to me to get a `File` as an argument and then build the index file in this function. Instead both `File` should be passed, otherwise you have mixed responsibilities.

source/agora/test/Base.d
```diff
@@ -324,6 +324,7 @@ public NetworkT makeTestNetwork (NetworkT : TestNetworkManager)
                 retry_delay : 100, // msecs
                 max_retries : 20,
                 min_listeners : nodes - 1,
+                data_dir : format("%s/%s%d", ".cache", "ledger.dat", idx)
```
Geod24: Unittests should never do IO, which this does.

Geod24 (changes requested):
Don't do IO from unittests. Also the commits are not atomic.

source/agora/test/Base.d
TrustHenry: just config not IO But deleted

source/agora/node/Ledger.d
TrustHenry: Fixed; use @safe

source/agora/node/Ledger.d
TrustHenry: added index_File param

source/agora/node/Ledger.d
TrustHenry: delete it

source/agora/node/Ledger.d
Geod24: Also the `Block` parameter should be made `const`

source/agora/node/Ledger.d
```diff
@@ -38,11 +44,15 @@ public class Ledger
     /// Temporary storage where transactions are stored until blocks are created.
     private Transaction[] storage;
 
+    /// ledger data dir
+    private const string data_dir;
+
     /// Ctor
-    public this ()
+    public this (const string data_dir)
```
Geod24: I am not sure I understand why `data_dir` is a parameter of the constructor. It does not need to be needed anywhere, as `writeBlockFile` takes `File` arguments, so it does not need to open them.

source/agora/node/Ledger.d
TrustHenry: ledger_file = Instance of file for ledgerFile

source/agora/node/Ledger.d
TrustHenry: But now we don't need File place because you get it as a parameter.

TrustHenry:
This PR has a comment of 75.

source/agora/node/Ledger.d
TrustHenry: added

TrustHenry:
Rebase done  - made very simple code :)

TrustHenry:
Better mmfile will be used.

PR #126, opened 2019-07-30
Add agora.common.Amount to handle monetary operations

Geod24:
This type should be used for monetary operations in the future, to prevent any chance of overflow or underflow.
Currently the total supply is 500M, but in the future it should be extended to the real value.
Since this value is likely to be dependent on the chain state, we might need to move some verifications outside of this type. But for now it'll do.
Also the amount of boilerplate we need to add to support our [de]serialization, hashing and Vibe.d is already annoying. Will check how I can fix that soon.

### DIFF
--- a/source/agora/common/Amount.d
+++ b/source/agora/common/Amount.d
@@ -1,0 +1,243 @@
+/*******************************************************************************
+
+   Defines a monetary type used in the blockchain
+
+   This struct can hold an amount between 0 and 500_000_000 * 10^7,
+   as there is an absolute maximum supply of 500M and each coin can be divided
+   in up to 10^7 units.
+
+   This struct does not expose operation overloading on purpose.
+   Having operator overloading would mean that any error should be reported
+   as either an `assert` being triggered or an `Exception` being thrown.
+   The convenience of using `a + b` would likely mean some places wouldn't
+   be properly bound checked, which would in turn open a DoS.
+   Instead, we provide two kind of functions for operations:
+   `bool OPNAME(ref Amount res, Type)` and `Amount mustOPNAME(Type)`
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.Amount;
+
+/// Ditto
+public struct Amount
+{
+    import agora.common.Deserializer;
+    import agora.common.Hash;
+    import agora.common.Serializer;
+
+    import core.checkedint;
+
+    /// Number of units ('cents') per coins
+    public static immutable Amount UnitPerCoin   = Amount(10_000_000, true);
+    /// Maximum number of BOA coins that can ever be in circulation
+    public static immutable ulong  MaxCoinSupply = 500_000_000;
+    /// Maximum amount of money that can ever be in circulation
+    public static immutable Amount MaxUnitSupply =
+        Amount(UnitPerCoin.value * MaxCoinSupply, true);
+
+    /// Helper type for `toString`
+    private alias SinkT = void delegate(scope const(char)[] v) @safe;
+
+    /// Internal data storage
+    private ulong value;
+
+    /***************************************************************************
+
+        Construct an instance of an `Amount`
+
+        Params:
+            units = The monetary amount this new instance represent.
+                    The unit is the smaller unit of currency (see `UnitPerCoin`)
+
+    ***************************************************************************/
+
+    public this (ulong units) nothrow pure @nogc @safe
+    {
+        assert(Amount.isInRange(units));
+        this.value = units;
+    }
+
+    /// Support for hashing
+    public void computeHash (scope HashDg dg) const nothrow @nogc @safe
+    {
+        hashPart(this.value, dg);
+    }
+
+    /// Support for serialization
+    public void serialize (scope SerializeDg dg) const nothrow @safe
+    {
+        serializePart(this.value, dg);
+    }
+
+    /// Support for deserialization
+    public void deserialize (scope DeserializeDg dg) nothrow @safe
+    {
+        deserializePart(this.value, dg);
+    }
+
+    /// Pretty-print this value
+    public void toString (SinkT dg) const @safe
+    {
+        import std.format;
+        formattedWrite(dg, "%d", this.value);
+    }
+
+    /// Also support for Vibe.d serialization to JSON
+    public string toString () const @safe
+    {
+        string ret;
+        scope SinkT dg = (scope v) { ret ~= v; };
+        this.toString(dg);
+        return ret;
+    }
+
+    /// Support for Vibe.d deserialization
+    public static Amount fromString (scope const(char)[] str) pure @safe
+    {
+        import std.conv : to;
+        immutable ul = str.to!ulong;
+        if (!Amount.isInRange(ul))
+            throw new Exception("Invalid input value to Amount");
+        return Amount(ul, true);
+    }
+
+    nothrow pure @nogc @safe:
+
+    /***************************************************************************
+
+        Returns:
+            `true` if `value` is in the range `[0; Amount.MaxUnitSupply]`
+            and can be safely used to make an `Amount`
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public static bool isInRange (ulong value)
+    {
+        return value <= MaxUnitSupply.value;
+    }
+
+    /// `isInRange` but as member function
+    pragma(inline, true)
+    public bool isValid () const
+    {
+        return isInRange(this.value);
+    }
+
+    /***************************************************************************
+
+        Add another value to `this` `Amount`
+
+        Params:
+            other = Another `Amount` value. If the value is invalid,
+                    `false` will be returned.
+        Returns:
+            `true` if the addition returned a number within bounds.
+            `false` if the number is out of the [0; MaxUnitSupply] bound.
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public bool add (Amount other)
+    {
+        bool overflow;
+        this.value = addu(this.value, other.value, overflow);
+        if (overflow || !this.isValid())
+        {
+            // If we overflow, make sure to poison the return value
+            this.value = ulong.max;
+            return false;
+        }
+        return true;
+    }
+
+    /***************************************************************************
+
+        Substract another value from `this` `Amount`
+
+        Params:
+            other = Another `Amount` value. If the value is invalid,
+                    `false` will be returned.
+        Returns:
+            `true` if the addition returned a number within bounds.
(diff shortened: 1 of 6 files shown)


source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
+/*******************************************************************************
+
+    Defines a monetary type used in the blockchain
```
AndrejMitrovic:
```suggestion
   Defines a monetary type used in the blockchain
```

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+    public static immutable ulong  MaxCoinSupply = 500_000_000;
+    /// Maximum amount of money that can ever be in circulation
+    public static immutable Amount MaxUnitSupply =
+        Amount(UnitPerCoin.value * MaxCoinSupply, true);
+
+    /// Internal data storage
+    private ulong value;
+
+    /// Support for hashing
+    public void computeHash(scope HashDg dg) const nothrow @nogc @safe
```
AndrejMitrovic: https://github.com/bpfkorea/agora/blob/v0.x.x/doc/Style.md#functions

:P

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+    /// Also support for Vibe.d serialization to JSON
+    public string toString () const @safe
+    {
+        string ret;
+        scope SinkT dg = (scope v) { ret ~= v; };
+        this.toString(dg);
+        return ret;
+    }
+
+    private alias SinkT = void delegate(scope const(char)[] v) @safe;
```
AndrejMitrovic: Could you move this above all the field declarations? It's hard to find here.

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+
+        Construct an instance of an `Amount`
+
+        Params:
+            units = The monetary amount this new instance represent.
+                    The unit is the smaller unit of currency (see `UnitPerCoin`)
+
+    ***************************************************************************/
+
+    public this (ulong units)
```
AndrejMitrovic: Could we make a rule to move the constructor before any other methods?

Geod24: We could, but the reason it's here is because of the attributes declaration on top. Bitblob works the same.

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+                    `false` will be returned.
+        Returns:
+            `true` if the addition returned a number within bounds.
+            `false` if the number is out of the [0; MaxUnitSupply] bound.
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public bool sub (Amount other)
+    {
```
AndrejMitrovic, line 176: Hmm.. What about checking if `this.isValid` is false at the start of this function? Because the value may already be poisoned.

AndrejMitrovic: And the subtracting it "unpoisons" it by mistake.

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+        }
+        return true;
+    }
+
+    /// Convenience version of `add` which asserts in case of overflow
+    /// Prefer using this only in `unittest`s
+    public ref Amount mustAdd (Amount other)
+    {
+        if (!this.add(other))
+            assert(0);
```
AndrejMitrovic, line 196: Seems like it could be a DoS vulnerability to keep this public, if you're never supposed to use it. It crashes the app right..

Geod24: Yep it does. It is "practical" but should be flagged in any code review. I could also make it `version (unittest)` but I wanted to see what the diff would look like

source/agora/common/Amount.d
Geod24: Make the suggestion instead you dingus

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+    private ulong value;
+
+    /// Support for hashing
+    public void computeHash(scope HashDg dg) const nothrow @nogc @safe
+    {
+        hashPart(this.value, dg);
+    }
+
+    /// Support for serialization
+    public void serialize(scope SerializeDg dg) const nothrow @safe
```
Geod24:
```suggestion
    public void serialize (scope SerializeDg dg) const nothrow @safe
```

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+    {
+        serializePart(this.value, dg);
+    }
+
+    /// Support for deserialization
+    public void deserialize(scope DeserializeDg dg) nothrow @safe
```
Geod24:
```suggestion
    public void deserialize (scope DeserializeDg dg) nothrow @safe
```

Geod24: Like THIS

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+    public this (ulong units)
+    {
+        assert(Amount.isInRange(units));
+        this.value = units;
+    }
+
+    /// Make a value without checking bounds
+    /// Used to initialize the bounds themselves, and to make invalid values
+    /// in unittests
+    private this (ulong units, bool dummy) { this.value = units; }
```
AndrejMitrovic:
```suggestion
        private this (ulong units, Flag!"YouDingus") { this.value = units; }
```

And then use it like:

```D
auto amount = Amount(100, Yes.YouDingus);
```

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+    /***************************************************************************
+
+        Returns:
+            `true` if `value` is in the range `[0; Amount.MaxUnitSupply]`
+            and can be safely used to make an `Amount`
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public static bool isInRange (ulong value)
```
AndrejMitrovic, line 122: Technically.. this would never be allowed. An amount should never spend the entire money supply. Except maybe the genesis transaction?

source/agora/common/Amount.d
Geod24: You have terrible indentation tastes

AndrejMitrovic: Whoa what happened there. Ignore that please.

Geod24: Too late. Can't unsee.

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+        return  value <= MaxUnitSupply.value;
+    }
+
+    /// `isInRange` but as member function
+    pragma(inline, true)
+    public bool isValid () const
+    {
+        return isInRange(this.value);
+    }
+
```
AndrejMitrovic, line 133: Two blank lines? What are you compensating for?

Geod24: https://www.youtube.com/watch?v=Dkk9gvTmCXY

source/agora/common/Amount.d
AndrejMitrovic:
```suggestion
    public void computeHash (scope HashDg dg) const nothrow @nogc @safe
```

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+    private alias SinkT = void delegate(scope const(char)[] v) @safe;
+
+    /// Support for Vibe.d deserialization
+    public static Amount fromString (scope const(char)[] str) pure @safe
+    {
+        import std.conv : to;
+        immutable ul = str.to!ulong;
+        if (!Amount.isInRange(ul))
```
AndrejMitrovic, line 106: So in some cases we poison the value, in other cases we throw an Exception. Is there a general rule where one or the other should be preferred?

AndrejMitrovic: Also I feel this check should be in the constructor?

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+        Returns:
+            `true` if `value` is in the range `[0; Amount.MaxUnitSupply]`
+            and can be safely used to make an `Amount`
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public static bool isInRange (ulong value)
+    {
+        return  value <= MaxUnitSupply.value;
```
AndrejMitrovic:
```suggestion
        return value <= MaxUnitSupply.value;
```

source/agora/common/Transaction.d
```diff
@@ -441,15 +427,20 @@ public Amount getSumInput (Transaction tx,
 
     Params:
         tx = `Transaction`
+        acc = Accumulator value. Pass a default-initialized value to get this
+              transaction's sum of output.
 
     Return:
         Sum of `Output` in the `Transaction`
```
AndrejMitrovic: Wrong docs.

source/agora/common/Amount.d
Geod24: This is a bit of an odd one: `fromString` is required for Vibe.d. But actually, Vibe.d has a serialization framework which offers multiple alternatives. I'd like to unify everything under this architecture ASAP, because every new type we introduce will need to define 5 methods...

Geod24: And the check is not in the constructor because the constructor being called with wrong parameter is a programming bug. While `fromString` is called for deserialization of parameters sent to the `API`.

source/agora/common/Amount.d
```diff
@@ -0,0 +1,239 @@
...
+   as there is an absolute maximum supply of 500M and each coin can be divided
+   in up to 10^7 units.
+
+   This struct does not expose operation overloading on purpose.
+   Having operator overloading would mean that any error should be reported
+   as either an `assert` being triggered or an `Exception` being thrown.
+   The convenience of using `a + b` would likely mean some places wouldn't
+   be properly bound checked, which would in turn open a DoS.
+   Instead, we provide two kind of functions for operations:
+   `bool OPNAME(ref Amount res, const Type)` and `Amount mustOPNAME(const Type)`
```
AndrejMitrovic: They're not `const` though. :)

AndrejMitrovic:
It LGTM for the most part.

source/agora/common/Amount.d
Geod24: Very good point. Fixed and added a test that substracts `invalid` to `invalid`.

source/agora/common/Amount.d
Geod24: But what be a good value then ?...

Geod24:
Updated, and removed the `getSumInput` function since it's actually unused and untested.

AndrejMitrovic:
The build failure was a random segfault, I've restarted that build.

AndrejMitrovic:
Is this ready?

AndrejMitrovic (approved):

PR #128, opened 2019-08-01
Memory mapped file of a block is Implemented using a single file.

MichaelKim20:
Use single file.
We will expand to use multiple files in the next step.

### DIFF
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -1,0 +1,268 @@
+/*******************************************************************************
+
+    Defines the memory mapped file of a block.
+
+    The number of files is one.
+    We will expand to use multiple files in the next step.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.BlockStorage;
+
+import agora.common.Amount;
+import agora.common.Block;
+import agora.common.Deserializer;
+import agora.common.Serializer;
+
+import std.algorithm.comparison;
+import std.file;
+import std.format;
+import std.mmfile;
+import std.path;
+
+
+/// Ditto
+public class BlockStorage
+{
+    /// Instance of memory mapped file
+    private MmFile file;
+
+    /// Path of block data
+    private string path;
+
+    /// Index is block height, Value is position of file
+    /// We will change to index file in the next step.
+    private size_t[] block_positions;
+
+    /// Ctor
+    public this (string path)
+    {
+        this.path = path;
+    }
+
+    /***************************************************************************
+
+        Make a file name using the index.
+
+        Returns:
+            Returns the file name.
+
+    ***************************************************************************/
+
+    private string getFileName ()
+    {
+        return buildPath(this.path, "block.dat");
+    }
+
+    /***************************************************************************
+
+        Open memory mapped file.
+
+        If a size change occurs, open it again.
+
+        Params:
+            resize = The size of the file to change. If 0 does not change it.
+
+    ***************************************************************************/
+
+    private void map (size_t resize = 0)
+    {
+        if ((this.file !is null) && (resize == 0))
+            return;
+
+        if (this.file !is null)
+            this.unMap();
+
+        this.file =
+            new MmFile(
+                this.getFileName(),
+                MmFile.Mode.readWrite,
+                resize,
+                null
+            );
+    }
+
+    /***************************************************************************
+
+        Close memory mapped file.
+
+    ***************************************************************************/
+
+    private void unMap ()
+    {
+        if (this.file is null)
+            return;
+        import core.memory : GC;
+
+        destroy(this.file);
+        GC.free(&this.file);
+        this.file = null;
+    }
+
+    /***************************************************************************
+
+        Save block to the file.
+
+        Params:
+            block = `Block` to save
+
+        Returns:
+            Returns true if success, otherwise returns false.
+
+    ***************************************************************************/
+
+    public bool saveBlock (const ref Block block)
+    {
+        if (block.header.height != this.block_positions.length)
+            return false;
+
+        const ubyte[] serialized_block = serializeFull(block);
+
+        const size_t bidx = block.header.height;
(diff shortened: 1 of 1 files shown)


source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,292 @@
...
+                    name,
+                    MmFile.Mode.readWrite,
+                    resize,
+                    null
+                );
+        else
+            this.file =
+                new MmFile(
+                    name,
+                    MmFile.Mode.readWriteNew,
```
AndrejMitrovic: It seems that the `readWrite` mode already creates the file if it doesn't exist? From the documentation:

```
/**
 * The mode the memory mapped file is opened with.
 */
enum Mode
{
    read,            /// Read existing file
    readWriteNew,    /// Delete existing file, write new file
    readWrite,       /// Read/Write existing file, create if not existing
    readCopyOnWrite, /// Read/Write existing file, copy on write
}
```

Unless I'm wrong, I think this part can just be:

```diff
diff --git a/source/agora/common/BlockStorage.d b/source/agora/common/BlockStorage.d
index f8ad8a0..a0a1456 100644
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -90,22 +90,13 @@ public class BlockStorage

         const string name = this.getFileName();

-        if (name.exists())
-            this.file =
-                new MmFile(
-                    name,
-                    MmFile.Mode.readWrite,
-                    resize,
-                    null
-                );
-        else
-            this.file =
-                new MmFile(
-                    name,
-                    MmFile.Mode.readWriteNew,
-                    max(1, resize),
-                    null
-                );
+        this.file =
+            new MmFile(
+                name,
+                MmFile.Mode.readWrite,
+                resize,
+                null
+            );

         this.is_mapped = true;
     }
```

MichaelKim20: I just checked. I'll fix it. Thank you.

source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,277 @@
...
+
+    // test of random access
+    import std.random;
+
+    auto rnd = rndGen;
+
+    Block random_block;
+    foreach (idx; 0 .. 100)
+    {
+        auto height = uniform(0, 999, rnd);
```
AndrejMitrovic: If you want to test with all the heights, but in random order, then you can do this:

```
foreach (height; iota(999).randomCover(rnd))
{
}
```

source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,277 @@
...
+        else
+        {
+            this.map();
+            pos = file.length;
+            // increase size by serialized_block.length
+            this.map(file.length + serialized_block.length);
+        }
+
+        // add information to the look up table.
+        if (this.block_positions.length < block.header.height + 1)
```
AndrejMitrovic: Shouldn't this always be true? Because you have the check above `if (block.header.height != this.block_positions.length) return false`. So you can do:

```D
this.block_positions ~= pos;
```

source/agora/common/BlockStorage.d
MichaelKim20: It's a very good function that I haven't seen anywhere else, and it's amazing. Thank you.

source/agora/common/BlockStorage.d
MichaelKim20: That's a good point. I'll apply it.

source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,274 @@
...
+
+    private void unMap ()
+    {
+        if (!this.is_mapped)
+            return;
+        import core.memory : GC;
+
+        destroy(this.file);
+        GC.free(&this.file);
+        this.is_mapped = false;
```
Geod24: Instead of having `is_mapped` boolean, you can just set `this.file = null` and use that as a test for being mapped.

source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,274 @@
...
+    import agora.common.crypto.Key;
+    import agora.common.Data;
+    import agora.common.Hash;
+    import agora.common.Transaction;
+    import agora.consensus.Genesis;
+
+    import std.algorithm.comparison;
+
+
+    string path = buildPath(getcwd, ".cache");
```
Geod24, line 205: Please don't do IO in unittests

Geod24 (changes requested):
Pretty good! I'll suggest something for the unittests soon.

source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,274 @@
...
+        assert(from < to);
+
+        block = deserializeBlock(cast(ubyte[])this.file[from .. to]);
+
+        return true;
+    }
+}
+
+/// Create and store 1000 blocks and read again. And read 100 random.
+unittest
```
Geod24:
```suggestion
version(none) unittest
```

Let's start with that so we don't hold you back.

source/agora/common/BlockStorage.d
Geod24: Ping

source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,274 @@
...
+        If a size change occurs, open it again.
+
+        Params:
+            resize = The size of the file to change. If 0 does not change it.
+
+    ***************************************************************************/
+
+    private void map (size_t resize = 0)
+    {
+        if (this.is_mapped && (resize == 0))
```
Geod24:
```suggestion
        if (this.file !is null && (resize == 0))
```

Geod24 (approved):

source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,268 @@
...
+
+        if (next_height >= this.block_positions.length)
+            to = file.length;
+        else
+            to = this.block_positions[next_height];
+
+        assert(from < file.length);
+        assert(from < to);
+
+        block = deserializeBlock(cast(ubyte[])this.file[from .. to]);
```
Geod24:
```suggestion
        block = deserialize!Block(cast(ubyte[])this.file[from .. to]);
```

PR #129, opened 2019-08-01
Apply `binary search algorithm` to Block.findHashIndex

MichaelKim20:
The hash of the transaction is aligned within the block.
Use the `binary search algorithm` to find out how many times the hash is contained.

### DIFF
--- a/source/agora/common/Block.d
+++ b/source/agora/common/Block.d
@@ -307,9 +307,26 @@ public struct Block
     {
         assert(this.merkle_tree.length != 0, "Block hasn't been fully initialized");
 
-        foreach (idx; 0 .. min(this.txs.length, this.merkle_tree.length))
-            if (hash == this.merkle_tree[idx])
-                return idx;
+        const size_t size = min(this.txs.length, this.merkle_tree.length);
+        if (size == 0) return this.txs.length;
+        long left = 0;
+        long right = size - 1;
+        long mid;
+        while (left <= right)
+        {
+            mid = (left + right) / 2;
+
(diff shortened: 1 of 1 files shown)


Geod24:
Good catch !
In dlang, there is a simpler way to do this. Look for `Sortedness` [here](https://dlang.org/phobos/std_range.html).
You can just call `assumeSorted` and `find` to have the binary search done by the library.

Geod24:
Hum, it seems that it is only implemented for sort. So forget about `std.algorithm` and let's just implement it here.

source/agora/common/Block.d
```diff
@@ -307,9 +307,25 @@ public struct Block
...
+        const size_t size = min(this.txs.length, this.merkle_tree.length);
+        if (size == 0) return this.txs.length;
+        long left = 0;
+        long right = size - 1;
+        long mid;
+        while (left <= right)
+        {
+            mid = (left + right) / 2;
+
+            if (this.merkle_tree[mid] == hash)
```
AndrejMitrovic: you have a double lookup here. How about storing it in a temporary:

```D
auto tgt_hash = this.merkle_tree[mid];
if (tgt_hash == hash)
   ...
else
    if (hash < tgt_hash)
...
```

AndrejMitrovic:
Just one comment, otherwise it looks good to me.

source/agora/common/Block.d
MichaelKim20: Modified~ Thanks~

AndrejMitrovic (approved):

PR #131, opened 2019-08-01
Upgrade to use multiple memory map files

MichaelKim20:
The file is divided into multi parts.
The number of blocks in a file is fixed.
Store the position of each block using an array.
Replace with an index file later.

MMFile-BlockStorage-Index

### DIFF
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -23,12 +23,16 @@ import agora.common.Deserializer;
 import agora.common.Hash;
 import agora.common.Serializer;
 
+import vibe.core.log;
+
 import std.algorithm;
 import std.container.rbtree;
+import std.exception;
 import std.file;
 import std.format;
 import std.mmfile;
 import std.path;
+import std.stdio;
 
 
 /// The size of header
@@ -90,6 +94,8 @@ public class BlockStorage
 
         this.height_idx = new IndexHeight();
         this.hash_idx = new IndexHash();
+
+        this.loadAllIndexes();
     }
 
     /***************************************************************************
@@ -280,6 +286,8 @@ public class BlockStorage
 
         // write to memory
         this.write(MFILE_HEAD_SIZE + pos + size_t.sizeof, serialized_block);
+
+        this.saveIndex(block.header.height, hash_bytes, block_position);
 
         return true;
     }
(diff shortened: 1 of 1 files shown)


source/agora/common/BlockStorage.d
```diff
@@ -0,0 +1,292 @@
...
+    /// Ctor
+    public this (string path)
+    {
+        this.is_mapped = false;
+        this.path = path;
+    }
+
+    public ~this ()
+    {
+        this.unMap();
```
Geod24: You cannot use the GC from a destructor.
Since D has a garbage collector, when a reference is collected by the GC, you have no guarantee that the GC-allocated references it holds are still alive.
If you had this guarantee, it would be impossible to finalize objects with cyclic references.

MichaelKim20: Okay. I'll remove this part.

source/agora/common/BlockStorage.d
```diff
@@ -62,7 +71,40 @@ public class BlockStorage
...
+            length = The size of data
+
+    ***************************************************************************/
+
+    private void writeDataLength (size_t length)
+    {
+        ubyte[] values;
+        values.length = size_t.sizeof;
+        *cast(size_t*)&(values[0]) = length;
+        foreach (idx, ref e; values)
```
Geod24:
```suggestion
        foreach (idx, e; (cast(const ubyte*)&length)[0 .. size_t.sizeof])
```

Geod24:
I think it would be great to break this PR done into smaller commits.

Geod24:
It's pretty hard to follow, as the commits are not atomic. For example the first 2 commits override each other quite a lot.
Style wise, use `private immutable ConstantName = 42;` instead of `const CONSTANT_NAME = 42;`.

source/agora/common/BlockStorage.d
```diff
@@ -159,35 +298,346 @@ public class BlockStorage
...
+
+        Read type of `size_t` data
+
+        Params:
+            pos = position of memory mapped file
+            value = type of `size_t`
+
+    ***************************************************************************/
+
+    private void writeSizeT (size_t pos, size_t value) @trusted
```
Geod24, line 423: W.r.t. `readDataLength`, `writeDataLength`, `writeSizeT`, `readSizeT`: This is what the serialization / deserialization module is for.

source/agora/common/BlockStorage.d
```diff
@@ -37,56 +73,125 @@ public class BlockStorage
...
+                    MmFile.Mode.readWrite,
+                    resize + MFILE_HEAD_SIZE + MFILE_RESERVE_SIZE,
+                    null
+                );
+            this.writeDataLength(resize);
+            this.data_size = resize;
+        }
+        else
+        {
+            if (name.exists())
```
Geod24, line 171: Can be turned into an `else if`

source/agora/common/BlockStorage.d
```diff
@@ -159,35 +298,346 @@ public class BlockStorage
...
+        size_t height;
+        ubyte[HASH_SIZE] hash;
+        size_t pos;
+        string file_name = buildPath(this.path, "index.dat");
+
+        this.height_idx.clear();
+        this.hash_idx.clear();
+
+        if (file_name.exists)
+        {
```
Geod24, line 563: Use early returns to reduce indentation.
In this case, instead of wrapping the rest of the function in a `if`, use:
```d
if (!file_name.exists())
    return;
```

This way the code is simpler to read.

source/agora/common/BlockStorage.d
```diff
@@ -159,35 +298,346 @@ public class BlockStorage
...
+        Params:
+            block = `Block` to read
+            hash = `Hash` of `Block`
+
+        Returns:
+            Returns true if success, otherwise returns false.
+
+    ***************************************************************************/
+
+    public bool readBlock (ref Block block, Hash hash) @safe
```
Geod24, line 354: The whole point of memory mapping is to return a `const` pointer, so that there is only ever one copy of the block. This copy the block.

source/agora/common/BlockStorage.d
```diff
@@ -118,37 +234,60 @@ public class BlockStorage
...
 
-    public bool saveBlock (const ref Block block)
+    public bool saveBlock (const ref Block block) @safe
     {
-        if (block.header.height != this.block_positions.length)
+        if ((this.height_idx.length > 0) &&
+            (this.height_idx.back.height >= block.header.height))
             return false;
 
         const ubyte[] serialized_block = serializeFull(block);
```
Geod24, line 243: Instead of allocating memory to serialize, it should be done straight to the file.

Geod24:
Also can you rebase against `v0.x.x` so we get coverage report ?

Geod24 (approved):
Actually on a second take, this does not break anything, so let's move forward and fix as needed.

PR #132, opened 2019-08-01
Test with dmd-beta and dmd-2.087.1

Geod24:
DMD 2.087.1 should solve all the C++ regressions that we had.
    In addition, testing with dmd-beta should allow us to catch regressions early.
    It will slow down our CI, however if dmd-beta introduce a regression it won't fail the build.
    We do look at the logs often enough that we'd catch any issue.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,20 @@
 language: d
 
 d:
-  - dmd-2.086.1 # Pinned due to https://issues.dlang.org/show_bug.cgi?id=20022
+  - dmd-2.087.1 # 2.087.0 has regressions
+  - dmd-2.086.1
   - ldc-1.16.0  # Older version don't work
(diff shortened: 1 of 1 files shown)


Geod24:
This uncover a DMD regression. Fix on its way at: https://github.com/dlang/dmd/pull/10256
Let's wait until the beta actually passes to merge this.

Geod24:
And also https://github.com/dlang/dmd/pull/10257 because I'm a dingus.

Geod24:
DMD v2.087.1 has been released, so I added it to this patch.

Geod24:
💚

AndrejMitrovic (approved):

Geod24:
`dmd-beta` is red because it's the old beta (without my fixes).
But `2.087.1` is newer

PR #135, opened 2019-08-01
Fix broken test in Ledger.d

AndrejMitrovic:
This problem showed up once I started using a tx pool in the Ledger. It started erroring about duplicate transactions. Turns out the test was wrong here.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -252,7 +252,8 @@ unittest
                 assert(ledger.acceptTransaction(tx));
             });
 
-        last_txs = txes;
+        // keep track of last tx's to chain them to
+        last_txs = txes[$ - Block.TxsInBlock .. $];
     }
 
     genBlockTransactions(2);
--- a/source/agora/common/Transaction.d
+++ b/source/agora/common/Transaction.d
@@ -330,6 +330,7 @@ public Transaction[] makeChainedTransactions (KeyPair key_pair,
     import agora.common.Block;
     import std.conv;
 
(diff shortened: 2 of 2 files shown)


source/agora/node/Ledger.d
```diff
@@ -252,7 +252,8 @@ unittest
                 assert(ledger.acceptTransaction(tx));
             });
 
-        last_txs = txes;
+        // keep track of last tx's to chain them to
+        last_txs = txes[$ - Block.TxsInBlock .. $];;
```
Geod24: Shouldn't it be:
```suggestion
        last_txs = txes[$ - (Block.TxsInBlock * count) .. $];
```

AndrejMitrovic: I just want the last 8 transactions, regardless of how many are there before that.

Geod24: Ah right: https://github.com/bpfkorea/agora/blob/f5bed737df2d19a1e83b6bf2ea4b5121f27d7903/source/agora/common/Transaction.d#L348

Maybe the function should reflect that

Geod24:
```suggestion
        last_txs = txes[$ - Block.TxsInBlock .. $];
```

AndrejMitrovic: > Maybe the function should reflect that

You mean to add docs? Sure.

AndrejMitrovic:
Added assertion in `getChainedTransactions` and removed double semicolon.

AndrejMitrovic:
Green.

Geod24 (approved):

PR #136, opened 2019-08-01
Make node shutdown explicit

AndrejMitrovic:
I need this for https://github.com/biozic/d2sqlite3/blob/fea5016f1b08f4e3135a0f4ebcb0ef5f522be4cd/source/d2sqlite3/database.d#L89

If you don't explicitly call `database.close()` **before** the destructors kick-in, it will throw an exception thanks to that check above which calls this: https://github.com/biozic/d2sqlite3/blob/08e77f5f2da3ead57d62c80226bd85e41c27fbd7/source/d2sqlite3/internal/memory.d#L70

I think we will probably need shutdown functionality at some point, so in my mind it makes sense to add it.

### DIFF
--- a/source/agora/app.d
+++ b/source/agora/app.d
@@ -72,6 +72,8 @@ private int main (string[] args)
         auto router = new URLRouter();
 
         auto node = new Node!NetworkManager(config);
+        scope(exit) node.shutdown();
+
         router.registerRestInterface(node);
         runTask( { node.start(); });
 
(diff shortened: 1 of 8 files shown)


source/agora/app.d
```diff
@@ -72,6 +72,8 @@ private int main (string[] args)
         auto router = new URLRouter();
 
         auto node = new Node!NetworkManager(config);
+        scope (exit) node.shutdown();
```
Geod24:
```suggestion
        scope(exit) node.shutdown();
```

AndrejMitrovic: Ah, really? Ok, I used it like this before because it's not technically a function call. :)

AndrejMitrovic:
Fixed scopes.

Geod24 (approved):
Please rebase

PR #137, opened 2019-08-01
Implement and use a Transaction Pool

AndrejMitrovic:
This is the most basic implementation of a transaction pool.

There's a bunch of things that will probably need to be improved:
- caching (e.g. caching .length() calls)
- avoiding excessive memory allocations

The first three commits are part of other PRs. Needed here to make this go green, I'll rebase it after the other PRs are merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,13 +38,15 @@ addons:
     packages:
       - libsodium-dev
       - g++-9
+      - libsqlite3-dev
 
 env:
-  - PATH=$HOME/bin/:$PATH
+  - PATH=$HOME/bin/:$PATH; export PKG_CONFIG_PATH="/usr/local/opt/sqlite/lib/pkgconfig"
 
 # No way to install the apt package, so...
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "osx" ];   then brew install libsodium; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ];   then brew install sqlite3; fi
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then mkdir -p $HOME/bin/; fi
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then ln -s `which gcc-9` $HOME/bin/gcc; fi # /usr/bin/gcc-9
   - if [ "$TRAVIS_OS_NAME" = "linux" ]; then ln -s `which g++-9` $HOME/bin/g++; fi # /usr/bin/g++-9
--- a/dub.json
+++ b/dub.json
@@ -35,7 +35,7 @@
         "source/scpp/build/numeric.o",
         "source/scpp/build/uint128_t.o"
     ],
-    "lflags": [ "-lsodium", "-lstdc++" ],
+    "lflags": [ "-lsodium", "-lstdc++", "-lsqlite3" ],
 
     "mainSourceFile": "source/agora/node/main.d",
 
@@ -52,6 +52,7 @@
     "dependencies": {
         "base32":           { "path": "submodules/base32/", "version": "*" },
         "bitblob":          { "path": "submodules/bitblob/", "version": "*" },
+        "d2sqlite3":        { "path": "submodules/d2sqlite3/", "version": "*" },
         "dyaml":            { "path": "submodules/dyaml/", "version": "*" },
         "libsodiumd":       { "path": "submodules/libsodiumd/", "version": "*" },
         "localrest":        { "path": "submodules/localrest/", "version": "*" },
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -1,0 +1,277 @@
+/*******************************************************************************
+
+    Contains a transaction pool that is serializable to disk,
+    using SQLite as a backing store.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.TransactionPool;
+
+import agora.common.Data;
+import agora.common.Deserializer;
+import agora.common.Hash;
+import agora.common.Serializer;
+import agora.common.Transaction;
+
+import d2sqlite3.database;
+import d2sqlite3.library;
+import d2sqlite3.results;
+import d2sqlite3.sqlite3;
+
+import vibe.core.log;
+
+import std.algorithm;
+import std.conv : to;
+import std.exception : collectException, enforce;
+import std.file : exists;
+import std.range;
+
+/// Ditto
+public class TransactionPool
+{
+    /***************************************************************************
+
+        Initialization function. Must be called once per-process
+        (not per-thread!) before the class can be used.
+
+    ***************************************************************************/
+
+    public static void initialize ()
+    {
+        .shutdown();
+        .config(SQLITE_CONFIG_MULTITHREAD);
+        version (unittest) {} else
+            .config(SQLITE_CONFIG_LOG, &loggerCallback, null);
+        .initialize();
+    }
+
+    /// SQLite db instance
+    private Database db;
+
+
+    /***************************************************************************
+
+        Params:
+            db_path = path to the database file, or in-memory storage if
+                      :memory: was passed
+
+    ***************************************************************************/
+
+    public this (in string db_path)
+    {
+        const db_exists = db_path.exists;
+        if (db_exists)
+            logInfo("Loading database from: %s", db_path);
+
+        // note: can fail. we may want to just recover txes from the network instead.
+        this.db = Database(db_path);
+
+        if (db_exists)
+            logInfo("Loaded database from: %s", db_path);
+
+        // create the table if it doesn't exist yet
+        this.db.execute("CREATE TABLE IF NOT EXISTS tx_pool " ~
+            "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
+    }
+
+    /***************************************************************************
+
+        Callback used
+
+        Params:
+            arg = unused (null)
+            code = the error code
+            msg = the error message
+
+    ***************************************************************************/
+
+    private static extern(C) void loggerCallback (void *arg, int code,
+        const(char)* msg) nothrow
+    {
+        try
+        {
+            logError("SQLite error: (%s) %s", code, msg.to!string);
+        }
+        catch (Exception ex)
+        {
+            // should not propagate exceptions to C code
+        }
+    }
+
+    /***************************************************************************
+
+        Shut down the database
+
+        Note: this method must be called explicitly, and not inside of
+        a destructor.
+
+    ***************************************************************************/
+
+    public void shutdown ()
+    {
+        this.db.close();
+    }
+
+    /***************************************************************************
+
+        Add a transaction to the pool
+
+        Params:
+            tx = the transaction to add
+
+        Throws:
+            SqliteException if the transaction failed to be added
+
+    ***************************************************************************/
+
+    public void add (Transaction tx) @safe
+    {
+        static ubyte[] buffer;
+        buffer.length = 0;
+        () @trusted { assumeSafeAppend(buffer); }();
+
+        scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
+        {
+            buffer ~= data;
+        };
+
+        serializePart(tx, dg);
+
+        () @trusted {
+            db.execute("INSERT INTO tx_pool (key, val) VALUES (?, ?)",
+                hashFull(tx)[], buffer);
+        }();
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the number of transactions in the pool
+
+    ***************************************************************************/
+
+    public size_t length () @safe
+    {
+        // todo: optimize / cache this
+        return () @trusted {
+            return db.execute("SELECT count(*) FROM tx_pool").oneValue!size_t;
+        }();
+    }
+
+    /***************************************************************************
+
+        Params:
+            amount = how many transactions to return
+
+        Returns:
+            an array of 'amount' transactions
+
+        Note:
+            @trusted because most of the body executes non-@safe code,
+            which is assumed to be safe
+
+    ***************************************************************************/
+
+    public Transaction[] take (size_t amount) @trusted
+    {
+        assert(amount <= this.length());
+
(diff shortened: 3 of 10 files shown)


AndrejMitrovic:
I tested GC collection cycles, and it looks like the GC can't reach the pointers to the transaction arrays when they're stored in the database. I've confirmed it with a test.

So we need to take one of several approaches:
- Use `GC.addRoot` for each pointer inside of a Transaction when it's added to the database.
- Store transactions in serialized form. However this has performance implications.
- Use a fast and compact form of serialization for the transactions. For example, using the `Contiguous` feature from the ocean library: https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/src/ocean/util/serialize/contiguous/package_.d

Note that in the last case we might have to extract the relevant code out of the Ocean library and use it directly. This is because Ocean is a large library, and it does not support platforms other than Linux.

AndrejMitrovic:
Added tests & improved the code a bit. Currently I'm using `GC.addRoot`, but this will probably not scale well.

source/agora/common/TransactionPool.d
```diff
@@ -0,0 +1,226 @@
...
+
+    ***************************************************************************/
+
+    public void add (Transaction tx) @trusted
+    {
+        import core.memory;
+        GC.addRoot(tx.inputs.ptr);
+        GC.addRoot(tx.outputs.ptr);
+        db.execute("INSERT INTO tx_pool (key, val) VALUES (:k, :v)",
+            hashFull(tx)[], (cast(ubyte*)&tx)[0 .. Transaction.sizeof]);
```
AndrejMitrovic: This was a start, but it can't work this way. We can't store pointers to disk.

We should probably incorporate (some form) of Contiguous support, and store that serialization here.

Also, SQLite supports memory-mapping through a pragma. I'm looking into that.

AndrejMitrovic:
I need to figure out a way to properly link with sqlite so it works on Travis. Hate this part.

Geod24:
Need a hand ?

.travis.yml
```diff
@@ -42,6 +42,8 @@ env:
 # No way to install the apt package, so...
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "osx" ];   then brew install libsodium; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ];   then brew install sqlite3; fi
+  - if [ "$TRAVIS_OS_NAME" = "linux" ]; then sudo apt-get install sqlite3; fi
```
Geod24: Put it under `packages` (L35)

Geod24:
Export the following variable on OSX:
`export PKG_CONFIG_PATH="/usr/local/opt/sqlite/lib/pkgconfig"`

dub.json
```diff
@@ -35,7 +35,7 @@
         "source/scpp/build/numeric.o",
         "source/scpp/build/uint128_t.o"
     ],
-    "lflags": [ "-lsodium", "-lstdc++" ],
+    "lflags": [ "-lsodium", "-lstdc++", "-L/usr/local/opt/sqlite/lib", "-lsqlite3" ],
```
```suggestion
    "lflags": [ "-lsodium", "-lstdc++", "-lsqlite3" ],
```

AndrejMitrovic:
Still fails on linux. I wish I could run a VM on this machine..

.travis.yml
```diff
@@ -35,13 +35,16 @@ addons:
...
+      - sqlite3
 
 env:
   - PATH=$HOME/bin/:$PATH
 
 # No way to install the apt package, so...
 before_install:
   - if [ "$TRAVIS_OS_NAME" = "osx" ];   then brew install libsodium; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ];   then brew install sqlite3; fi
+  - if [ "$TRAVIS_OS_NAME" = "osx" ];   then export PKG_CONFIG_PATH="/usr/local/opt/sqlite/lib/pkgconfig"; fi
```
Geod24: Doesn't work.
Use:
```yaml
env:
  - PKG_CONFIG_PATH=/usr/local/opt/sqlite/lib/pkgconfig
```

At the top level

AndrejMitrovic: Ah I missed the env part.

AndrejMitrovic:
> Export the following variable on OSX:
> export PKG_CONFIG_PATH="/usr/local/opt/sqlite/lib/pkgconfig"

So anyone who wants to build agora locally will have to add this to their .bashrc?

AndrejMitrovic:
Fixed segfault due to https://issues.dlang.org/show_bug.cgi?id=20110

Now there are a bunch of SQL errors during uninitialization due to https://github.com/biozic/d2sqlite3/issues/54

Geod24:
> So anyone who wants to build agora locally will have to add this to their .bashrc?

Yep. We could do something about it in the future, to detect and warn.

AndrejMitrovic:
Ok I think I fixed the linux build.

I can reproduce the InvalidMemoryOperation (I was missing a rebase against v0.x.x).

AndrejMitrovic:
Fixed segfault, it was a missing `shutdown` call.

AndrejMitrovic:
Fixed issue with shutdown, https://github.com/biozic/d2sqlite3/issues/54 fixed it. I've updated the submodule HEAD in the first commit with the fix.

AndrejMitrovic:
Not sure why it doesn't link on Linux, it links ok locally in virtualbox (beaver).

AndrejMitrovic:
Eh I think I know why:

```
/* Version 3.8.7 and later */
#define sqlite3_auto_extension         sqlite3_api->auto_extension
#define sqlite3_bind_blob64            sqlite3_api->bind_blob64
```

The trusty version of sqlite3 is probably old.

AndrejMitrovic:
Cool, osx passes. I just need to figure out how to make this work on trusty..

AndrejMitrovic:
I found a backported version of sqlite3 which seems to work on Trusty: https://linuxhint.com/install-sqlite-ubuntu-linux-mint/

However, I'm unsure about just using random people's ppa's.

AndrejMitrovic:
Looks like Travis doesn't install the one from the ppa. grr..

Geod24:
You can't `sudo` in Travis.

AndrejMitrovic:
> You can't sudo in Travis.

Does it "just work" without sudo?

.travis.yml
```diff
@@ -25,6 +25,12 @@ cache:
     - $HOME/Library/Caches/Homebrew
     - /usr/local/Homebrew
 
+# workaround: force installing the ppa version of libsqlite3 (for trusty)
+before_install:
+  - sudo add-apt-repository ppa:jonathonf/backports -y
+  - sudo apt-get update -q
+  - sudo apt-get install libsqlite3-dev
```
Geod24: Change the addons part to be like this:

```yaml
addons:
  apt:
    sources:
      - ubuntu-toolchain-r-test
      - travis-ci/sqlite3
    packages:
      - libsodium-dev
      - g++-9
      - sqlite3
```

And remove this.

Geod24: This should give you 3.7

AndrejMitrovic:
Weird, now I'm getting hangs on my local machine too.

```
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x00007fff610c986a libsystem_kernel.dylib`__psynch_cvwait + 10
    frame #1: 0x00007fff6118856e libsystem_pthread.dylib`_pthread_cond_wait + 722
    frame #2: 0x000000010d61b7d1 agora`_D4core4sync9condition9Condition4waitMFZv + 25
    frame #3: 0x000000010d2a1495 agora`_D3std11concurrency10MessageBox__T3getTDFNaNbNiNfS6geod249LocalRest8ResponseZvTPFNaNiNfCQDiQDh14LinkTerminatedZvTPFNaNiNfCQEqQEp15OwnerTerminatedZvTPFNfSQFv7variant__T8VariantNVmi32ZQpZvZQFxMFMQFyMQEoMQDkMQCfZb at concurrency.d:2148
    frame #4: 0x000000010d2a121b agora`_D3std11concurrency__T11receiveOnlyTS6geod249LocalRest8ResponseZQBqFZQBh at concurrency.d:775
    frame #5: 0x000000010d31c7c5 agora`_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPIZQBl5startMFNfZ9__lambda2MFNeZSQDjQDf8Response at LocalRest.d-mixin-698:715
    frame #6: 0x000000010d31c646 agora`_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPIZQBl5startMFNfZv at LocalRest.d-mixin-698:706
    frame #7: 0x000000010d322481 agora`_D5agora4test4Base18TestNetworkManager5startMFZ__T9__lambda1TC6geod249LocalRest__T9RemoteAPITCQDoQDlQDj7TestAPIZQBeZQCoFNfQCjZv at Base.d:155
```

I see an `OnwerTerminated` in there.

AndrejMitrovic:
The issue was with initialization. `shutdown/initialize` should only be called once per process, not once per thread.

AndrejMitrovic:
Most are green except the random segfault, restarting that machine.

.travis.yml
```diff
@@ -36,13 +36,15 @@ addons:
     packages:
       - libsodium-dev
       - g++-9
+      - libsqlite3-dev
 
 env:
-  - PATH=$HOME/bin/:$PATH
+  - PATH=$HOME/bin/:$PATH; export PKG_CONFIG_PATH="/usr/local/opt/sqlite/lib/pkgconfig"
```
Geod24, line 44: `export` is not necessary, nor is `;`. Just a space.

source/agora/common/TransactionPool.d
```diff
@@ -0,0 +1,248 @@
...
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.TransactionPool;
```
Geod24, line 15: Shouldn't be in `common` if it has a dependency on `Transaction`. Let's talk about this IRL.

source/agora/common/TransactionPool.d
```diff
@@ -0,0 +1,248 @@
...
+import std.conv : to;
+import std.exception : collectException, enforce;
+import std.file : exists;
+import std.range;
+
+/// Ditto
+public class TransactionPool
+{
+    // should only be initialized once per process
+    shared static this()
```
Geod24: Is a `shared static this` the only way to do this ? (shared) module ctor have quite a lot of drawback / friction with other features.

source/agora/common/TransactionPool.d
```diff
@@ -0,0 +1,248 @@
...
+
+        Params:
+            tx = the transaction to add
+
+        Throws:
+            SqliteException if the transaction failed to be added
+
+    ***************************************************************************/
+
+    public void add (Transaction tx) @trusted
```
Geod24: No `@trusted` at top level please

source/agora/test/Base.d
```diff
@@ -264,6 +265,12 @@ public final class TestNode (Net) : Node!Net, TestAPI
         return new MemMetadata();
     }
 
+    /// Return a transaction pool backed by an in-memory SQLite db
+    public override TransactionPool getPool (string _unused) @system
```
Geod24: There's a simple way to tell an argument is unused: don't give it a name ;)

Geod24 (commented):
Wow that was easy

source/agora/common/TransactionPool.d
AndrejMitrovic: Hmm not necessarily. We could do it in the main module and also the main test runner module.

source/agora/common/TransactionPool.d
AndrejMitrovic: What do you mean "top level" though?

source/agora/test/Base.d
AndrejMitrovic: Ah I seem to remember ddoc complaining about it.. but maybe that was years ago. I'll fix.

source/agora/common/TransactionPool.d
Geod24: I mean, at the function level (unless the function is trivial).
Here you have no idea if `serializeFull` is safe for example.

AndrejMitrovic: Right, so trusted delegates it is then. :)

AndrejMitrovic:
I've moved the shared module constructor in the main.d/Base.d modules.

I could get rid of the one in main.d and just initialize everything inside the `main()` function, but in unittests I don't have a main function that's called before the tests, so I still think we need to use a module ctor there.

Do you have better suggestions?

source/agora/common/TransactionPool.d
```diff
@@ -0,0 +1,267 @@
...
+            tx = the transaction to add
+
+        Throws:
+            SqliteException if the transaction failed to be added
+
+    ***************************************************************************/
+
+    public void add (Transaction tx) @safe
+    {
+        auto tx_bytes = serializeFull(tx);
```
Geod24: For the future: This should really use a buffer

source/agora/common/TransactionPool.d
```diff
@@ -0,0 +1,267 @@
...
+        Returns:
+            an array of 'amount' transactions
+
+        Note:
+            @trusted because most of the body executes non-@safe code,
+            which is assumed to be safe
+
+    ***************************************************************************/
+
+    public Transaction[] take (size_t amount) @trusted
```
Geod24, line 182: Urg

Geod24 (approved):

Geod24:
LGTM besides the `Urg`

source/agora/common/TransactionPool.d
AndrejMitrovic: Ah yeah, forgot about the API there.

source/agora/common/TransactionPool.d
AndrejMitrovic: I'm not sure what to do about this. Almost everything inside calls non-safe code..

AndrejMitrovic: There is a comment right above it.

Geod24: Right. Okay.

AndrejMitrovic:
Updated to use a buffer for the transaction serialization.

Geod24 (approved):

PR #138, opened 2019-08-02
[temporary] Add printouts to test where the segfault originates

AndrejMitrovic:
We want to figure out which of the unittests fail in v0.x.x. Hopefully this uncovers it.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -235,6 +235,8 @@ public class Ledger
 ///
 unittest
 {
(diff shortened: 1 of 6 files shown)


AndrejMitrovic:
In the near future we should provide our own test-runner function which does these printouts and allows things like filtering.

AndrejMitrovic:
Found the offending unittest: Running unittest source/agora/test/Ledger.d:193

The question is if it's always the same unittest. I'll do a few more runs.

PR #141, opened 2019-08-05
Disable OSX testing

Geod24:
OSX tests have been failing on slow machines lately.
However they do not fail for the developers.
Disable them to not block progress while we work on a fix.


AndrejMitrovic (approved):

Geod24:
💚

PR #144, opened 2019-08-06
Add tryUntil() function to simplify writing tests

AndrejMitrovic:
I've also noticed one of the tests just exits after the timeout, it failed to actually assert that the condition was not met.


AndrejMitrovic:
Yeah the random failure is definitely back.

Geod24 (approved):

Geod24:
The name is kinda weird tho. `cond().retryFor(1.seconds)` sounds better to me. But potato, potahto

PR #147, opened 2019-08-06
Move agora.common.API to agora.node.API

Geod24:
API was out of place in common, as it depends on a lot of types.
Since it basically defines the communication interface for a node,
it is better located in agora.node itself.


AndrejMitrovic (approved):

Geod24:
💚

PR #148, opened 2019-08-06
newCoinBaseTX: Pass an `Amount` as default parameter, not `int`

Geod24:
It was only working because no one was calling it with the default value.


AndrejMitrovic:
> It was only working because no one was calling it with the default value.

So the compiler doesn't even check if it makes sense? How odd.

AndrejMitrovic (approved):

Geod24:
> So the compiler doesn't even check if it makes sense? How odd.

I'm as surprised as you. Well it started to fail after some random code reshuffle.

PR #155, opened 2019-08-08
Test with dmd-nightly instead of beta

Geod24:
Because that's really what I wanted to do originally.
Beta are not frequent enough to be in the CI.
Also do the same thing for LDC.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,9 @@ dist: xenial
 d:
   - dmd-2.087.1 # 2.087.0 has regressions
   - dmd-2.086.1
-  - ldc-1.16.0  # Older version don't work
-  - dmd-beta    # Last so that `fast_finish` is useful
+  - ldc-1.16.0     # Older version don't work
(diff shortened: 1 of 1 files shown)


AndrejMitrovic (approved):

.travis.yml
```diff
@@ -6,7 +6,8 @@ d:
   - dmd-2.087.1 # 2.087.0 has regressions
   - dmd-2.086.1
   - ldc-1.16.0  # Older version don't work
-  - dmd-beta    # Last so that `fast_finish` is useful
+  - dmd-nightly   # Last so that `fast_finish` is useful
```
AndrejMitrovic: no longer last

Geod24:
Green

AndrejMitrovic:
Why is codecov "failing". Is there a rule set that it should never go negative?

AndrejMitrovic:
The problem is then on the PR page the PR will look red when it's actually green..

Geod24:
> The problem is then on the PR page the PR will look red when it's actually green..

Yep. I'll see what I can do.

PR #156, opened 2019-08-08
Implement a basic integration test

Geod24:


Geod24:
I am truly amazed how this just works.

AndrejMitrovic:
> I am truly amazed how this just works.

Put that on the Bosagora T-shirt.

AndrejMitrovic:
> Do not use Vibe.d logger from BlockStorage
> Otherwise we have to link in Vibe.d in the integration test

I don't understand this comment. Why can't we use vibe.d logging in integration tests?

Geod24:
Now that #160 is merged this need a rebase

Geod24:
> I don't understand this comment. Why can't we use vibe.d logging in integration tests?

Otherwise it attempts to link in Vibe.d's static ctors...

AndrejMitrovic:
Ah. And there's the reason why we don't like them. :)

Geod24:
Updated, should be ready to go

source/agora/common/BlockStorage.d
```diff
@@ -504,8 +502,7 @@ public class BlockStorage
         }
         catch (Exception ex)
         {
-            logError("saveIndex: %s",
-                () @trusted { return ex.message; }());
+            () @trusted { stderr.writeln("saveIndex: ", ex.message); }();
```
AndrejMitrovic, line 505: I assume we are going to replace this eventually with non-vibe.d logging?

Geod24: Yes

tests/unit/BlockStorage.d
```diff
@@ -0,0 +1,134 @@
...
+
+/// The maximum number of block in one file
+private immutable ulong MFILE_MAX_BLOCK = 100;
+
+///
+private void main ()
+{
+    const size_t count = 300;
+
+    string path = buildPath(getcwd, ".cache");
```
AndrejMitrovic, line 39: Hmm.. this doesn't make the test parallelizable yet though, since multiple tests will end up overwriting each others data.

But we can fix this later.

AndrejMitrovic:
Good start. We could add filtering and parallel test runs later.

AndrejMitrovic (approved):

tests/unit/BlockStorage.d
Geod24: Yep, at this point I wanted to just move (and execute) the test.
We should go with something similar to Ocean (sandbox), but let's add a few tests first.

PR #158, opened 2019-08-08
Modify to a contiguous memory map file using multiple files

MichaelKim20:
A memory map file consists of several files.
Previously, there was a gap between the files. We got rid of it and connected it as one.
And when we save a block or read a block, we can serialize it from the inside.
We can expect better performance because we can reduce memory replication.
MMFile

### DIFF
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -502,7 +502,7 @@ public class BlockStorage
         }
         catch (Exception ex)
         {
-            () @trusted { stderr.writeln("saveIndex: ", ex.message); }();
+            this.writeLog("BlockStorage.saveIndex: ", ex);
         }
     }
 
@@ -552,7 +552,7 @@ public class BlockStorage
         }
         catch (Exception ex)
         {
-            () @trusted { stderr.writeln("loadAllIndexes: ", ex.message); }();
+            this.writeLog("BlockStorage.loadAllIndexes: ", ex);
         }
     }
 
@@ -567,6 +567,27 @@ public class BlockStorage
         string name = buildPath(path, "index.dat");
         if (name.exists)
             name.remove();
+    }
+
+    /***************************************************************************
+
+        Write error message
+
+        Params:
+            func = function name
+            ex = Instance of `Exception`
+
+    ***************************************************************************/
+
+    private void writeLog (string func, Exception ex) @trusted nothrow
+    {
+        try
+        {
+            stderr.writeln(func, ex.message);
+        }
+        catch (Exception ex)
+        {
+        }
     }
 }
 
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -87,6 +87,9 @@ public class BlockStorage
         this.height_idx = new IndexHeight();
         this.hash_idx = new IndexHash();
 
+        if (!this.path.exists)
+            mkdirRecurse(this.path);
+
         this.loadAllIndexes();
     }
 
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -33,15 +33,6 @@ import std.path;
 import std.stdio;
 
 
-/// The size of header
-private immutable size_t HeadSize = size_t.sizeof;
-
-/// Amount of reserved size
-private immutable size_t ReserveSize = 64 * 1024;
-
-/// The maximum number of block in one file
-private immutable ulong MaxBlock = 100;
-
 /// The map file size
 private immutable size_t MapSize = 640 * 1024;
 
@@ -74,21 +65,24 @@ public class BlockStorage
     /// Index of current file
     private size_t file_index;
 
-    /// The size of data. Exclude the size of the header and the reserved size
-    private size_t data_size;
-
     /// Index is block height
     private IndexHeight height_idx;
 
     /// Index is block hash
     private IndexHash hash_idx;
 
+    /// Size of Block Data
+    private size_t length;
+
+    /// Base Position of current file
+    private size_t file_base;
+
     /// Ctor
     public this (string path)
     {
         this.path = path;
-        this.file_index = 0;
-        this.data_size = 0;
+        this.file_index = ulong.max;
+        this.length = ulong.max;
 
         this.height_idx = new IndexHeight();
         this.hash_idx = new IndexHash();
@@ -119,79 +113,44 @@ public class BlockStorage
 
         If it was mapping the same file, just return.
         If it was mapping the other file, close previously mapped file.
-        If a size change occurs, open it again.
 
         Params:
             findex = the index of the file.
-            resize = The size of the file to change. If 0 does not change it.
 
     ***************************************************************************/
 
-    private void map (size_t findex, size_t resize = 0) @trusted
+    private bool map (size_t findex) @trusted nothrow
     {
-        // It's already mapped,
-        if ((this.file !is null) && (findex == this.file_index))
-        {
-            // if it doesn't change its size
-            if (resize == 0)
-                return;
+        try {
+            if ((this.file !is null) && (findex == this.file_index))
+                return true;
 
-            // if can use reserved memory
-            if ((this.data_size < resize) &&
-                (HeadSize + resize < this.getFileLength()))
-            {
-                this.writeDataLength(resize);
-                this.data_size = resize;
-                return;
-            }
-        }
+            if (this.file !is null)
+                this.release();
 
-        if (this.file !is null)
-            this.release();
+            this.file_index = findex;
 
-        this.file_index = findex;
-        const string name = this.getFileName(this.file_index);
-
-        if (resize != 0)
-        {
             this.file =
                 new MmFile(
-                    name,
+                    this.getFileName(this.file_index),
                     MmFile.Mode.readWrite,
-                    resize + HeadSize + ReserveSize,
+                    MapSize,
                     null
                 );
-            this.writeDataLength(resize);
-            this.data_size = resize;
+
+            this.file_base = MapSize * this.file_index;
+            return true;
         }
-        else if (name.exists())
+        catch (Exception ex)
         {
-            this.file =
-                new MmFile(
-                    name,
-                    MmFile.Mode.readWrite,
-                    0,
-                    null
-                );
-            this.data_size = this.readDataLength();
-        }
-        else
-        {
-            this.file =
-                new MmFile(
-                    name,
-                    MmFile.Mode.readWrite,
-                    max(HeadSize+ReserveSize, resize),
-                    null
-                );
-            this.writeDataLength(0);
-            this.data_size = 0;
+            this.writeLog("BlockStorage.map: ", ex);
+            return false;
         }
     }
 
     /***************************************************************************
 
-        Close memory mapped file.
+        Release memory mapped file.
 
     ***************************************************************************/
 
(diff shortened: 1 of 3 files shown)


source/agora/common/BlockStorage.d
```diff
@@ -382,6 +397,50 @@ public class BlockStorage
...
+    ***************************************************************************/
+
+    public bool readBlockMerkleRoot (ref Block block, Hash merkle_root) @safe
+    {
+        ubyte[HASH_SIZE] hash_bytes = merkle_root[0..HASH_SIZE];
+
+        auto finds
+            = this.merkle_root_idx[].find!((a, b) => a.hash == b)(hash_bytes);
+
+        if (!finds.empty)
```
AndrejMitrovic: It's better to return early. So:

```D
if (finds.empty)
    return false;
```

And then the rest of the code can be unindented by one level. It makes it easier to reason about code this way.

MichaelKim20: Yes, I'll modify it.

source/agora/common/BlockStorage.d
```diff
@@ -781,5 +781,5 @@ version(none) unittest
         assert(random_block.header.merkle_root == merkle_roots[idx]);
     }
 
-    other.release();
+    other.unMap();
```
Geod24: Sorry, I noticed that after. I thought `release` was unused and so removed it. But it is a better name, all things considered, so the integration test PR fixes this.

Geod24:
You should rebase on v0.x.x

MichaelKim20:
A memory map file consists of several files.
Previously, there was a gap between the files. We got rid of it and connected it as one.
And when we save a block or read a block, we can serialize it from the inside.
We can expect better performance because we can reduce memory replication.

MichaelKim20:
Std.mmfile was modified to obtain the address of the memory map file.
However, it was difficult to secure adjacent memory continuously.

So whenever a new memory map is created. 
It was possible to change the memory address of the base point.

However, I thought it was not appropriate to have to modify std.mmfile.
This was resolved by the concept of position of the file.

Then, when accessing the portion at the edge of the file in the memory map, it was done one by one.

I think I've met all the requirements now.

source/agora/common/BlockStorage.d
```diff
@@ -33,17 +33,8 @@ import std.path;
...
 
 
-/// The size of header
-private immutable size_t MFILE_HEAD_SIZE = size_t.sizeof;
-
-/// Amount of reserved size
-private immutable size_t MFILE_RESERVE_SIZE = 64 * 1024;
-
-/// The maximum number of block in one file
-private immutable ulong MFILE_MAX_BLOCK = 100;
```
Geod24: In this commit (the second) those constants are removed, not renamed, so this commit does not compile.

source/agora/common/BlockStorage.d
```diff
@@ -113,73 +113,37 @@ public class BlockStorage
...
-                    0,
+                    MapSize,
                     null
                 );
-            this.data_size = this.readDataLength();
+
+            this.file_base = MapSize * this.file_index;
         }
-        else
+        catch (Exception ex)
```
Geod24, line 147: So the failure does not get reported to the caller. Does that mean we are okay with this component failing ?

source/agora/common/BlockStorage.d
```diff
@@ -149,19 +149,24 @@ public class BlockStorage
...
     public void release () @trusted
     {
-        if (this.file is null)
-            return;
         import core.memory : GC;
 
-        destroy(this.file);
-        GC.free(&this.file);
-        this.file = null;
+        try
```
Geod24: Did you see `release` throw ? Because it should never happen.

source/agora/common/BlockStorage.d
```diff
@@ -244,6 +244,7 @@ public class BlockStorage
 
         return true;
     }
+
```
Geod24, line 257: Unrelated change (commit 926779701294e16eba7c744344ad30f976d02033)

source/agora/common/BlockStorage.d
```diff
@@ -359,12 +360,33 @@ public class BlockStorage
...
+            (to / MapSize == this.file_index))
+        {
+            ubyte[] data;
+            try
+            {
+                const size_t x0 = from - this.file_base;
+                const size_t x1 = to - this.file_base;
+                data = cast(ubyte[])this.file[x0 .. x1];
+            }
+            catch (Exception ex)
```
Geod24, line 417: Flat-out ignoring an exception does not seem like the best way to go

source/agora/common/BlockStorage.d
```diff
@@ -377,13 +399,76 @@ public class BlockStorage
...
+                foreach (idx, e; data)
+                    this.file[x0+idx] = e;
+            }
+            else
+            {
+                foreach (idx, e; data)
+                    this.writeByte(pos + idx, e);
+            }
+        }
+        catch (Exception ex)
```
Geod24, line 467: Same here, why are we ignoring Exception ?

Geod24 (changes requested):
Good commit separation. The review was much simpler. Thanks for that 👍 
My recommendation: The commit title should indicate what & why something is being done. A commit message like "Modify function read, write" is not informative. "Remove ability to read unmapped file" gives much more info, for example (that's just an example). However, things like "Remove unused variable xxx" or "Fix typo in documentation", for example, are fine because it is self evident from the diff.
Last thing, there have been a lot of empty `try` / `catch` added. They seem to have been added to make the function `nothrow`, however that means that errors are now ignored, and we cannot guarantee the integrity of the data on disk, which is pretty bad.

Geod24:
Can you submit the first 4 commits in an independent PR ? They are good to go.

source/agora/common/BlockStorage.d
MichaelKim20: fixed

source/agora/common/BlockStorage.d
MichaelKim20: fixed

source/agora/common/BlockStorage.d
MichaelKim20: fixed

source/agora/common/BlockStorage.d
MichaelKim20: fixed

source/agora/common/BlockStorage.d
MichaelKim20: fixed

source/agora/common/BlockStorage.d
MichaelKim20: fixed

source/agora/common/BlockStorage.d
MichaelKim20: fixed

source/agora/common/BlockStorage.d
MichaelKim20: fixed

Geod24:
@MukeunKim :
Please do not ignore `Exception`.
The pattern:
```d
try { /* ... */ }
catch (Exception e) {}
```
Is very bad because if an exception is triggered, we won't know.
I understand that is done because there are some `nothrow` in place.
In this case, for the time being, I recommend to use:
```d
void func () nothrow
{
    scope (failure) assert(0);
    // ...
}
```

This will make the function `nothrow`

Geod24:
Ah, it looks like the last commit should be merged with the first one. But I recommend to just `assert` for the time being.

Geod24:
Ready for a new review

AndrejMitrovic:
What about the test-case including multiple transactions in the block?

MukeunKim:
@AndrejMitrovic

MichaelKim20:
Added test casings for multi-transaction to block.
I have also verified normal operation for large blocks.
I'm trying to add @AndrejMitrovic  to coworker for the test code, but it doesn't work.

AndrejMitrovic:
> I'm trying to add @AndrejMitrovic to coworker for the test code, but it doesn't work.

It doesn't matter, I just want us to have a test.

source/agora/common/BlockStorage.d
```diff
@@ -301,8 +280,17 @@ public class BlockStorage
         if (finds.empty)
             return false;
 
-        this.readBlockAtPosition(block, finds.front.position);
-        return true;
+        bool success = true;
```
AndrejMitrovic: In general, it is much better to assign failure by default, not success.

However there is no need for this boolean. This can be rewritten to:

```D
try
{
    this.readBlockAtPosition(block, finds.front.position);
    return true;
}
catch (Exception ex)
{
    this.writeLog("BlockStorage.readBlock: ", ex);
    return false;
}
```

source/agora/common/BlockStorage.d
```diff
@@ -328,26 +316,33 @@ public class BlockStorage
...
+        try
+        {
+            this.readBlockAtPosition(block, finds.front.position);
```
AndrejMitrovic, line 320: Same comment here as above.

source/agora/common/BlockStorage.d
```diff
@@ -441,30 +457,79 @@ public class BlockStorage
...
-            return 0;
+        ubyte data;
+        try
+        {
+            if (this.map(pos / MapSize))
+                data = this.file[pos - this.file_base];
+        }
+        catch (Exception ex)
+        {
+            writeLog("BlockStorage.readBytes: ", ex);
```
AndrejMitrovic, line 497: So in case of a failure, it will return a ubyte with the value 0 and the error will be ignored?

source/agora/common/BlockStorage.d
```diff
@@ -441,30 +457,79 @@ public class BlockStorage
...
+    private void writeByte (size_t pos, ubyte data) nothrow @trusted
+    {
+        try
+        {
+            if (this.map(pos / MapSize))
+                this.file[pos - this.file_base] = data;
+        }
+        catch (Exception ex)
+        {
+            this.writeLog("BlockStorage.writeByte: ", ex);
```
AndrejMitrovic, line 525: Why is it ok to fail to write and then ignore the error? If the node can't write to disk, wouldn't it be better to abort the whole process and shut down the node? The data that was written will be invalid when its being read later, because some of the data is missing (the checksum will fail, once it's implemented).

source/agora/common/BlockStorage.d
```diff
@@ -441,30 +457,79 @@ public class BlockStorage
...
+            }
+            else
+            {
+                foreach (idx, e; data)
+                    this.writeByte(pos + idx, e);
+            }
+        }
+        catch (Exception ex)
+        {
+            this.writeLog("BlockStorage.write: ", ex);
```
AndrejMitrovic, line 469: Also same comment about ignoring write errors.

MichaelKim20:
Review Please!

AndrejMitrovic:
It looks great now. How about adding `nothrow` everywhere where there a function returns a status code?

AndrejMitrovic:
@MukeunKim if you get too many errors trying to make them `nothrow`, or if it's too much work, then leave it the way it is. We can add `nothrow` later.

AndrejMitrovic (approved):

MukeunKim:
Ready for a new review

MichaelKim20:
@Geod24 Ready for a new review ^^

source/agora/common/BlockStorage.d
```diff
@@ -119,79 +116,44 @@ public class BlockStorage
...
-        If a size change occurs, open it again.
 
         Params:
             findex = the index of the file.
-            resize = The size of the file to change. If 0 does not change it.
 
     ***************************************************************************/
 
-    private void map (size_t findex, size_t resize = 0) @trusted
+    private bool map (size_t findex) @trusted nothrow
```
Geod24, line 125: Missing `Returns:`

Geod24 (approved):
LGTM

PR #162, opened 2019-08-09
Integrate BlockStorage with Ledger

TrustHenry:
Integrate BlockStorage with Ledger
Using BlockStorage  Dependency PR #158

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -14,6 +14,7 @@
 module agora.node.Ledger;
 
 import agora.common.Amount;
+import agora.common.BlockStorage;
 import agora.consensus.data.Block;
 import agora.common.Data;
 import agora.common.Hash;
@@ -30,22 +31,20 @@ import std.algorithm;
 /// Ditto
 public class Ledger
 {
-    /// data storage for all the blocks,
-    /// currently a single contiguous region to
-    /// improve locality of reference
-    private const(Block)[] ledger;
-
-    /// pointer to the latest block
-    private const(Block)* last_block;
+    /// data storage for all the blocks
+    private IBlockStorage storage;
 
     /// Pool of transactions to pick from when generating blocks
     private TransactionPool pool;
 
     /// Ctor
-    public this (TransactionPool pool)
+    public this (TransactionPool pool, IBlockStorage storage)
     {
         this.pool = pool;
-        this.addGenesisBlock();
+        this.storage = storage;
+        auto block = getGenesisBlock();
+        if (!this.storage.saveBlock(block))
+            assert(0);
     }
 
     /***************************************************************************
@@ -64,15 +63,17 @@ public class Ledger
 
     public bool acceptBlock (const ref Block block) nothrow @safe
     {
-        if (!block.isValid(this.last_block.header.height,
-            this.last_block.header.hashFull, &this.findUTXO))
+        Block last_block;
+        if (!this.storage.readLastBlock(last_block))
+            assert(0);
+        if (!block.isValid(last_block.header.height,
+            last_block.header.hashFull, &this.findUTXO))
         {
             logDebug("Rejected block. %s", block);
             return false;
         }
-
-        this.ledger ~= block;
-        this.last_block = &this.ledger[$ - 1];
+        if (!this.storage.saveBlock(block))
+            return false;
         return true;
     }
 
@@ -115,20 +116,27 @@ public class Ledger
     {
         auto txs = this.pool.take(Block.TxsInBlock);
         assert(txs.length == Block.TxsInBlock);
-        auto block = makeNewBlock(*this.last_block, txs);
-        assert(this.acceptBlock(block));
+        Block last_block;
+        if (!this.storage.readLastBlock(last_block))
+            assert(0);
+        auto block = makeNewBlock(last_block, txs);
+        if (!this.acceptBlock(block))
+            assert(0);
     }
 
     /***************************************************************************
 
         Returns:
-            the highest block
+            latest block height
 
     ***************************************************************************/
 
-    public const(Block)* getLastBlock () @safe nothrow @nogc
+    public ulong getBlockHeight () @safe nothrow
     {
-        return this.last_block;
+        Block last_block;
+        if (!this.storage.readLastBlock(last_block))
+            assert(0);
+        return last_block.header.height;
     }
 
     /***************************************************************************
@@ -147,14 +155,23 @@ public class Ledger
     ***************************************************************************/
 
     public const(Block)[] getBlocksFrom (ulong block_height, size_t max_blocks)
-        @safe nothrow @nogc
+        @safe nothrow
     {
         assert(max_blocks > 0);
 
-        if (block_height > this.ledger.length)
-            return null;
+        const MaxHeight
+            = min(block_height + max_blocks, this.getBlockHeight() + 1);
+        const(Block)[] res;
+        foreach (height; block_height .. MaxHeight)
+        {
+            Block block;
+            if (!this.storage.readBlock(block, height))
+                assert(0);
 
-        return this.ledger[block_height .. min(block_height + max_blocks, $)];
+            res ~= block;
+        }
+
+        return res;
     }
 
     /***************************************************************************
@@ -170,10 +187,14 @@ public class Ledger
     ***************************************************************************/
 
     private const(Output)* findUTXO (Hash tx_hash, size_t index)
-        @safe nothrow @nogc
+        @safe nothrow
     {
-        foreach (ref block; this.ledger)
+        foreach (height; 0 .. this.getBlockHeight() + 1)
         {
+            Block block;
+            if (!this.storage.readBlock(block, height))
+                assert(0);
+
             foreach (ref tx; block.txs)
             {
                 if (hashFull(tx) == tx_hash)
@@ -204,45 +225,35 @@ public class Ledger
 
     public Hash[] getMerklePath (ulong block_height, Hash hash) @safe
     {
-        if (block_height >= this.ledger.length)
+        if (this.getBlockHeight() < block_height)
             return null;
 
-        const(Block)* block = &this.ledger[block_height];
-
+        Block block;
+        if (!this.storage.readBlock(block, block_height))
+            assert(0);
         size_t index = block.findHashIndex(hash);
-        if (index < block.txs.length)
-            return block.getMerklePath(index);
-        else
+        if (index >= block.txs.length)
             return null;
-    }
-
-    /***************************************************************************
-
-        Add the genesis block to the ledger.
-
-    ***************************************************************************/
-
-    private void addGenesisBlock ()
-    {
-        assert(this.ledger.length == 0);
-        auto block = getGenesisBlock();
-        this.ledger ~= block;
-        this.last_block = &this.ledger[$ - 1];
+        return block.getMerklePath(index);
     }
 }
 
 ///
 unittest
 {
+    import agora.common.BlockStorage;
     import agora.common.crypto.Key;
     import agora.common.Data;
     import agora.common.Hash;
 
+    auto storage = new MemBlockStorage();
     auto pool = new TransactionPool(":memory:");
     scope(exit) pool.shutdown();
-    scope ledger = new Ledger(pool);
-    assert(*ledger.getLastBlock() == getGenesisBlock());
-    assert(ledger.ledger.length == 1);
+    scope ledger = new Ledger(pool, storage);
+    assert(ledger.getBlockHeight() == 0);
+
+    const(Block)[] blocks = ledger.getBlocksFrom(0, 10);
+    assert(blocks[$ - 1] == getGenesisBlock());
 
     auto gen_key_pair = getGenesisKeyPair();
     Transaction[] last_txs;
@@ -261,15 +272,14 @@ unittest
     }
 
     genBlockTransactions(2);
-    const(Block)[] blocks = ledger.getBlocksFrom(0, 10);
+    blocks = ledger.getBlocksFrom(0, 10);
     assert(blocks[0] == getGenesisBlock());
     assert(blocks[0].header.height == 0);
     assert(blocks.length == 3);  // two blocks + genesis block
 
     /// now generate 98 more blocks to make it 100 + genesis block (101 total)
     genBlockTransactions(98);
-
-    assert(ledger.getLastBlock().header.height == 100);
+    assert(ledger.getBlockHeight() == 100);
 
     blocks = ledger.getBlocksFrom(0, 10);
     assert(blocks[0] == getGenesisBlock());
@@ -317,9 +327,10 @@ unittest
     import agora.common.Data;
     import agora.common.Hash;
 
+    auto storage = new MemBlockStorage();
     auto pool = new TransactionPool(":memory:");
     scope(exit) pool.shutdown();
-    scope ledger = new Ledger(pool);
+    scope ledger = new Ledger(pool, storage);
 
     Block invalid_block;  // default-initialized should be invalid
     assert(!ledger.acceptBlock(invalid_block));
@@ -335,11 +346,13 @@ unittest
 /// Merkle Proof
 unittest
 {
+    import agora.common.BlockStorage;
     import agora.common.crypto.Key;
 
+    auto storage = new MemBlockStorage();
     auto pool = new TransactionPool(":memory:");
     scope(exit) pool.shutdown();
-    scope ledger = new Ledger(pool);
+    scope ledger = new Ledger(pool, storage);
 
     auto gen_key_pair = getGenesisKeyPair();
     auto txs = makeChainedTransactions(gen_key_pair, null, 1);
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -15,6 +15,7 @@ module agora.node.Node;
 
 import agora.consensus.data.Block;
 import agora.common.BanManager;
+import agora.common.BlockStorage;
 import agora.common.Config;
 import agora.common.Metadata;
 import agora.common.crypto.Key;
@@ -69,6 +70,8 @@ public class Node : API
     ///
     private Ledger ledger;
 
+    /// Blockstorage
+    private IBlockStorage storage;
 
     /// Ctor
     public this (const Config config)
@@ -78,8 +81,9 @@ public class Node : API
         this.config = config;
         this.network = this.getNetworkManager(config.node, config.banman,
             config.network, config.dns_seeds, this.metadata);
+        this.storage = this.getBlockStorage(config.node.data_dir);
         this.pool = this.getPool(config.node.data_dir);
-        this.ledger = new Ledger(this.pool);
+        this.ledger = new Ledger(this.pool, this.storage);
         this.gossip = new GossipProtocol(this.network, this.ledger);
         this.exception = new RestException(
             400, Json("The query was incorrect"), string.init, int.init);
@@ -149,7 +153,7 @@ public class Node : API
     /// GET: /block_height
     public ulong getBlockHeight ()
     {
-        return this.ledger.getLastBlock().header.height;
+        return this.ledger.getBlockHeight();
     }
 
     /// GET: /blocks_from
@@ -230,6 +234,32 @@ public class Node : API
         return new DiskMetadata(data_dir);
     }
 
+    /***************************************************************************
+
+        Returns an instance of a BlockStorage or MemoryStorage
+
+        Note: not exposed in the API.
+
+        Params:
+            data_dir = path to the blockdata directory
+
+        Returns:
+            Returns instance of `MemoryStorage` if data_dir is empty,
+            otherwise returns instance of `BlockStorage`
+
+    ***************************************************************************/
+
+    protected IBlockStorage getBlockStorage (string data_dir) @system
+    {
+        version (unittest)
+        {
+            return new MemBlockStorage();
+        }
+        else
+        {
+            return new BlockStorage(data_dir);
+        }
+    }
 
     /// GET: /merkle_path
     public Hash[] getMerklePath (ulong block_height, Hash hash) @safe
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -165,7 +165,7 @@ public class NetworkManager
             while (1)
             {
                 this.getBlocksFrom(
-                    ledger.getLastBlock().header.height + 1,
+                    ledger.getBlockHeight() + 1,
                     blocks => blocks.all!(block => ledger.acceptBlock(block)));
 
                 this.taskman.wait(2.seconds);


source/agora/node/Ledger.d
```diff
@@ -25,17 +26,21 @@ import agora.node.API;
...
-    private const(Block)[] ledger;
+    private BlockStorage ledger;
 
-    /// pointer to the latest block
-    private const(Block)* last_block;
+    /// latest block of `BlockStorage`
+    private Block last_block;
+
+    /// is vaild `last_block`
+    private bool is_valid_last_block;
```
AndrejMitrovic: This is unnecessary. There will always be at least one valid block in the ledger, the genesis block.

source/agora/node/Ledger.d
```diff
@@ -142,14 +155,21 @@ public class Ledger
...
+        foreach (height;  block_height .. block_height + max_blocks)
+        {
+            if (this.ledger.readBlock(block, height))
+                res ~= block;
+            else
+                break;
+        }
 
-        return this.ledger[block_height .. min(block_height + max_blocks, $)];
+        return cast(const(Block)[])res;
```
AndrejMitrovic: You could just change `Block[] res;` into `const(Block)[]` res, and return it here without casting.

source/agora/node/Ledger.d
```diff
@@ -15,6 +15,7 @@ module agora.node.Ledger;
```
AndrejMitrovic, line 15: typo in commit message: `Leedger`

source/agora/node/Ledger.d
```diff
@@ -25,17 +26,21 @@ import agora.node.API;
...
 
 /// Ditto
 public class Ledger
 {
-    /// data storage for all the blocks,
+    /// Blockstorage for all the blocks,
     /// currently a single contiguous region to
-    /// improve locality of reference
-    private const(Block)[] ledger;
+    private BlockStorage ledger;
```
AndrejMitrovic: This should ideally be an interface, something like `IBlockStorage`, where in unittest code it uses in-memory storage for blocks, while in production it uses `BlockStorage` that is disk-based.

cc @Geod24 what do you think?

source/agora/node/Ledger.d
```diff
@@ -59,16 +71,17 @@ public class Ledger
...
         if (!this.isValidBlock(block))
         {
             logDebug("Rejected block. %s", block);
             return;
         }
 
-        this.ledger ~= block;
-        this.last_block = &this.ledger[$ - 1];
+        this.ledger.saveBlock(block);
+        () @trusted { this.last_block = cast(Block) block; }();
```
AndrejMitrovic: Why is this cast necessary?

source/agora/node/Ledger.d
TrustHenry: modified

source/agora/node/Ledger.d
TrustHenry: remove is_valid_last_block

source/agora/node/Ledger.d
TrustHenry: @AndrejMitrovic Can you suggest a good way?

AndrejMitrovic: yes, don't cast

AndrejMitrovic: @Geod24 The issue here was that `this.last_block` was changed from:

```D
const(Block)* last_block;
```

to:

```D
Block last_block;
```

So you can't assign const(Block) to Block.

Geod24: Why is it not a pointer anymore ?

AndrejMitrovic:
The reason why we had `last_block` in the `Ledger` is because:

1. When creating a new block, it has to refer to the hash of the previous block.
2. To know the ledger's current block height (used by other functions).

The question is if you want to keep the last block as a field, or maybe just the block height? You could do something like this:

```diff
diff --git a/source/agora/common/BlockStorage.d b/source/agora/common/BlockStorage.d
index c1b82e7..b581b15 100644
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -304,7 +304,7 @@ public class BlockStorage
 
     ***************************************************************************/
 
-    public bool readBlock (ref Block block, size_t height) @safe
+    public bool readBlock (ref Block block, size_t height) @safe nothrow
     {
         if ((this.height_idx.length == 0) ||
             (this.height_idx.back.height < height))
@@ -316,8 +316,16 @@ public class BlockStorage
         if (finds.empty)
             return false;
 
-        this.readBlockAtPosition(block, finds.front.position);
-        return true;
+        try
+        {
+            this.readBlockAtPosition(block, finds.front.position);
+            return true;
+        }
+        catch (Exception ex)
+        {
+            // todo: log error
+            return false;
+        }
     }
 
     /***************************************************************************
diff --git a/source/agora/node/Ledger.d b/source/agora/node/Ledger.d
index 588339f..64d8922 100644
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -37,10 +37,7 @@ public class Ledger
     private BlockStorage ledger;
 
     /// latest block of `BlockStorage`
-    private Block last_block;
-
-    /// is vaild `last_block`
-    private bool is_valid_last_block;
+    private size_t block_height;
 
     /// Pool of transactions to pick from when generating blocks
     private TransactionPool pool;
@@ -79,9 +76,8 @@ public class Ledger
             return;
         }
 
+        this.block_height = block.header.height;
         this.ledger.saveBlock(block);
-        () @trusted { this.last_block = cast(Block) block; }();
-        this.is_valid_last_block = true;
     }
 
     /***************************************************************************
@@ -123,7 +119,7 @@ public class Ledger
     {
         auto txs = this.pool.take(Block.TxsInBlock);
         assert(txs.length == Block.TxsInBlock);
-        auto block = makeNewBlock(this.last_block, txs);
+        auto block = makeNewBlock(this.getLastBlock(), txs);
         this.acceptBlock(block);
     }
 
@@ -134,9 +130,12 @@ public class Ledger
 
     ***************************************************************************/
 
-    public const(Block)* getLastBlock () @safe nothrow @nogc
+    public ref const(Block) getLastBlock () @safe nothrow
     {
-        return &this.last_block;
+        static Block block;
+        if (!this.ledger.readBlock(block, this.block_height))
+            assert(0);  // should never happen
+        return block;
     }
 
     /***************************************************************************
@@ -186,12 +185,9 @@ public class Ledger
 
     ***************************************************************************/
 
-    private bool isValidBlock (const ref Block block) pure nothrow @safe @nogc
+    private bool isValidBlock (const ref Block block) nothrow @safe
     {
-        const expected_height = this.is_valid_last_block
-            ? (this.last_block.header.height + 1)
-            : 0;
-
+        const expected_height = this.block_height + 1;
         return block.header.height == expected_height;
     }
 
@@ -210,10 +206,8 @@ public class Ledger
     private const(Output)* findOutput (Hash tx_hash, size_t index) @safe
     {
         Block block;
-        if (!this.is_valid_last_block)
-            return null;
 
-        foreach (height; 0 .. this.last_block.header.height + 1)
+        foreach (height; 0 .. this.block_height + 1)
         {
             if (this.ledger.readBlock(block, height))
                 break;
```

It depends what you think is the best and most optimal solution.

AndrejMitrovic:
Also, please be aware that this:

```D
const ref Block func { }
``` 

does **not** make it return a constant reference. It marks the function itself as const.

This returns a constant reference:

```D
ref const(Block) func { }
```

It's one of the edge-cases in D's design that you have to do it this way.

source/agora/node/Ledger.d
TrustHenry: modified

TrustHenry:
Apply As @AndrejMitrovic  Suggested

AndrejMitrovic:
Why does this PR now have 14 commits?

Geod24:
@TrustHenry : https://travis-ci.com/bpfkorea/agora/jobs/224247564

source/agora/common/BlockStorage.d
```diff
@@ -301,8 +301,16 @@ public class BlockStorage
         if (finds.empty)
             return false;
 
-        this.readBlockAtPosition(block, finds.front.position);
-        return true;
+        bool success = true;
```
AndrejMitrovic:
```D
try
{
    this.readBlockAtPosition(block, finds.front.position);
    return true;
}
catch (Exception ex)
{
    return false;
}
```

source/agora/node/Ledger.d
TrustHenry: delete last_block then Point to Last Block at block height.

source/agora/node/Ledger.d
```diff
@@ -142,14 +153,21 @@ public class Ledger
...
 
     public const(Block)[] getBlocksFrom (ulong block_height, size_t max_blocks)
-        @safe nothrow @nogc
+        @safe
     {
         assert(max_blocks > 0);
 
-        if (block_height > this.ledger.length)
-            return null;
+        const(Block)[] res;
```
AndrejMitrovic, line 164:
```D
const MaxHeight = min(this.block_height, start_height + max_blocks);

const(Block)[] res;
foreach (height; start_height .. MaxHeight + 1)
{
    Block block;
    if (!this.ledger.readBlock(block, height))
        assert(0);  // should not happen

    res ~= block;
}
```

source/agora/node/Ledger.d
```diff
@@ -166,12 +184,9 @@ public class Ledger
...
     ***************************************************************************/
 
-    private bool isValidBlock (const ref Block block) pure nothrow @safe @nogc
+    private bool isValidBlock (const ref Block block) nothrow @safe
     {
-        const expected_height = this.last_block !is null
-            ? (this.last_block.header.height + 1)
-            : 0;
-
+        const expected_height = this.block_height + 1;
```
AndrejMitrovic: You can remove the constant if it's only used in one place. Just write:

```D
return block.header.height == this.block_height + 1;
```

source/agora/node/Ledger.d
```diff
@@ -189,8 +204,12 @@ public class Ledger
 
     private const(Output)* findOutput (Hash tx_hash, size_t index) @safe
     {
-        foreach (ref block; this.ledger)
+        Block block;
```
AndrejMitrovic: Move the block declaration inside of the `foreach` loop.

source/agora/node/Ledger.d
```diff
@@ -221,16 +240,19 @@ public class Ledger
...
     {
-        if (block_height >= this.ledger.length)
-            return null;
-
-        const(Block)* block = &this.ledger[block_height];
-
-        size_t index = block.findHashIndex(hash);
-        if (index < block.txs.length)
-            return block.getMerklePath(index);
+        Block block;
```
AndrejMitrovic, line 231: If the wanted block height is greater than our `this.block_height`, then this function should return null.

Also, add an assertion like before so `block.getMerklePath()` must return true.

source/agora/node/Ledger.d
```diff
@@ -244,8 +266,8 @@ unittest
     auto pool = new TransactionPool(":memory:");
     scope(exit) pool.shutdown();
     scope ledger = new Ledger(pool);
-    assert(*ledger.getLastBlock() == getGenesisBlock());
-    assert(ledger.ledger.length == 1);
+    assert(ledger.getLastBlock() == getGenesisBlock());
+    assert(ledger.block_height + 1 == 1);
```
```suggestion
    assert(ledger.block_height == 0);
```

source/agora/node/Ledger.d
```diff
@@ -326,8 +348,8 @@ unittest
...
```
```suggestion
    assert(ledger.block_height == 0);
```

AndrejMitrovic (changes requested):

AndrejMitrovic:
@team: There is something we need to keep in mind:

If a node consistently returns "no data" because the Ledger has issues reading from the disk, then that node may end up getting banned by other nodes. We don't want that to happen. If there's an issue with the hard drive then the node should should ideally just shut down and the node operator can fix the hard drive and restart the node.

Also when receiving requests, the arguments to our API calls need to be validated. So if some "bad node" requests block #999999999 and it doesn't exist, we don't want to just iterate over each block height and request the ledger to retrieve the block. It's a form of denial-of-service attack.

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

source/agora/common/BlockStorage.d
TrustHenry: Applied the suggestion

TrustHenry:
@AndrejMitrovic ping

source/agora/node/Ledger.d
```diff
@@ -25,17 +26,18 @@ import agora.node.API;
...
 
 import std.algorithm;
+import std.file;
+import std.path;
 
 /// Ditto
 public class Ledger
 {
-    /// data storage for all the blocks,
+    /// Blockstorage for all the blocks,
```
```suggestion
    /// Blockstorage for all the blocks
```

And remove the next line which says `/// currently a single contiguous region to`, because the comment is outdated (and it's cut off, because the `/// improve locality of reference` part was removed).

source/agora/node/Ledger.d
```diff
@@ -25,17 +26,18 @@ import agora.node.API;
...
     /// currently a single contiguous region to
-    /// improve locality of reference
-    private const(Block)[] ledger;
+    private BlockStorage ledger;
```
AndrejMitrovic: I don't like calling this object instance `ledger`. The `Ledger` class is already the ledger, this should be named `storage` or `block_storage` or something else.

source/agora/node/Ledger.d
```diff
@@ -44,7 +46,15 @@ public class Ledger
     public this (TransactionPool pool)
     {
         this.pool = pool;
+
+        string path = buildPath(getcwd, ".cache");
+        if (!path.exists)
+            mkdir(path);
+        this.ledger = new BlockStorage(path);
+        BlockStorage.removeIndexFile(path);
```
AndrejMitrovic: I don't understand why the index file is deleted after loading? Also `removeIndexFile ` should be private and called from within the `BlockStorage` constructor.

source/agora/node/Ledger.d
```diff
@@ -44,7 +46,15 @@ public class Ledger
...
+
         auto block = getGenesisBlock();
+        this.block_height = size_t.max;
```
AndrejMitrovic: Why?

AndrejMitrovic: This initializer will be replaced as soon as the genesis block is created. So it's unnecessary.

AndrejMitrovic: I can see why this was added. It's because of the code in `isValidBlock` which checks that the header's height is `this.block_height + 1`. However what you are depending on here is integer overflow to wrap around to zero.

It would be much better to just replace the call `this.acceptBlock` with `this.ledger.saveBlock(block);`, and leave `this.block_height` to be initialized with zero (the default).

source/agora/node/Ledger.d
```diff
@@ -221,16 +242,22 @@ public class Ledger
 
     public Hash[] getMerklePath (ulong block_height, Hash hash) @safe
     {
-        if (block_height >= this.ledger.length)
+        if(this.block_height < block_height)
```
AndrejMitrovic: Styling:

```suggestion
        if (this.block_height < block_height)
```

source/agora/node/Ledger.d
```diff
@@ -189,8 +206,12 @@ public class Ledger
 
     private const(Output)* findOutput (Hash tx_hash, size_t index) @safe
     {
-        foreach (ref block; this.ledger)
+        foreach (height; 0 .. this.block_height + 1)
         {
+            Block block;
+            if (this.ledger.readBlock(block, height))
```
AndrejMitrovic: If `readBlock` succeeds, you break out of the foreach loop and you return null. So the output will never be found. I think you meant to do `if (!this.ledger.readBlock(block, height)) return null`.

source/agora/node/Ledger.d
TrustHenry: change name `storage`

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

source/agora/node/Ledger.d
TrustHenry: Fixed

source/agora/node/Ledger.d
TrustHenry: deleted `removeIndexFile`

TrustHenry: However, we have to declare it separately during the unittest.

source/agora/node/Ledger.d
TrustHenry: `this.block_height = block.header.height;
        this.storage.saveBlock(block);` 
Applied the suggestion

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

source/agora/node/Ledger.d
TrustHenry: Why it without? Thank you for finding me.

source/agora/node/Ledger.d
```diff
@@ -25,27 +26,31 @@ import agora.node.API;
...
+    public this (TransactionPool pool, BlockStorage storage)
     {
         this.pool = pool;
+        this.storage = storage;
         auto block = getGenesisBlock();
-        assert(this.acceptBlock(block));
+        if (this.storage.saveBlock(block))
+            this.block_height = block.header.height;
+        else
+            assert(0);
```
Geod24, line 47: Suggestion:
```
        if (!this.storage.saveBlock(block))
            assert(0);
        this.block_height = block.header.height;
```

source/agora/node/Ledger.d
```diff
@@ -62,16 +67,15 @@ public class Ledger
...
         if (!this.isValidBlock(block))
         {
             logDebug("Rejected block. %s", block);
             return false;
         }
-
-        this.ledger ~= block;
-        this.last_block = &this.ledger[$ - 1];
+        if (this.storage.saveBlock(block))
+            this.block_height = block.header.height;
```
Geod24: This should handle error

source/agora/node/Ledger.d
```diff
@@ -125,9 +129,24 @@ public class Ledger
 
     ***************************************************************************/
 
-    public const(Block)* getLastBlock () @safe nothrow @nogc
+    public ref const(Block) getLastBlock () @safe nothrow
+    {
+        static Block block;
```
Geod24: This is a dangerous pattern: It means that when calling this function again, the previous reference will be invalidated.

source/agora/node/Ledger.d
```diff
@@ -227,31 +254,43 @@ public class Ledger
...
+        if (this.block_height < block_height)
             return null;
 
-        const(Block)* block = &this.ledger[block_height];
-
-        size_t index = block.findHashIndex(hash);
-        if (index < block.txs.length)
-            return block.getMerklePath(index);
+        Block block;
+        if (this.storage.readBlock(block, block_height))
```
Geod24: @TrustHenry : From experience, returning early leads to more readable code.
For example:
```d
if (!this.storage.readBlock(block, block_height))
    assert(0);
size_t index = block.findHashIndex(hash);
if (index >= block.txs.length)
    return null;
return block.getMerklePath(index);
```

In this example, we return early when there's an error, so it is easier for the reader to follow the flow.

source/agora/node/Ledger.d
```diff
@@ -227,31 +254,43 @@ public class Ledger
...
 unittest
 {
+    import agora.common.BlockStorage;
     import agora.common.crypto.Key;
     import agora.common.Data;
     import agora.common.Hash;
 
+    string path = buildPath(getcwd, ".cache");
+    if (path.exists)
+        rmdirRecurse(path);
```
Geod24: We should not do IO in unittests. Do we have a way to have in memory BlockStorage @AndrejMitrovic ?

source/agora/test/Base.d
```diff
@@ -467,6 +477,10 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         auto net = new APIManager();
         foreach (idx, ref conf; configs)
         {
+            string path = conf.node.data_dir;
+            if (std.file.exists(path))
+                rmdirRecurse(path);
```
Geod24: That looks wrong, we should not do IO in unittests

Geod24 (commented):
Looks pretty good overall, but there's still some things to address.
We might want to move forward with this to get the ball rolling tho.

source/agora/node/Ledger.d
AndrejMitrovic: I already made a comment about this. https://github.com/bpfkorea/agora/pull/162#discussion_r312346572

source/agora/node/Ledger.d
Geod24: Yes

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

source/agora/node/Ledger.d
TrustHenry: added handle error

source/agora/node/Ledger.d
TrustHenry: A block ref parameter was added.

source/agora/node/Ledger.d
TrustHenry: Applied the suggestion

Geod24:
I guess this needs a rebase ?

TrustHenry:
Ready for a new review

source/agora/common/BlockStorage.d
```diff
@@ -758,3 +758,253 @@ public size_t readSizeType (File file) @trusted
...
+        {
+            stderr.writeln(func, ex.message);
+        }
+        catch (Exception ex)
+        {
+        }
+    }
+}
+
+
```
AndrejMitrovic: Could you please replace this unittest with the following one? It creates valid blocks with no double-spend transactions:

```D
///
unittest
{
    import std.algorithm.comparison;

    const size_t BlockCount = 50;
    MemoryStorage storage = new MemoryStorage();

    const(Block)[] blocks;
    Hash[] block_hashes;

    auto gen_block = getGenesisBlock();
    blocks ~= gen_block;
    storage.saveBlock(gen_block);
    block_hashes ~= hashFull(gen_block.header);
    Transaction[] last_txs;

    void genBlocks (size_t count)
    {
        while (--count)
        {
            auto txs = makeChainedTransactions(gen_key_pair, last_txs, 1);
            auto block = makeNewBlock(blocks[$ - 1], txs);
            last_txs = txs;

            blocks ~= block;
            block_hashes ~= hashFull(block.header);
            storage.saveBlock(block);
        }
    }

    genBlocks(BlockCount);

    // load
    Block[] loaded_blocks;
    loaded_blocks.length = BlockCount;
    foreach (idx; 0 .. BlockCount)
        storage.readBlock(loaded_blocks[idx], idx);

    // compare
    assert(equal(blocks, loaded_blocks));

    // test of random access
    import std.random;
    import std.range;

    auto rnd = rndGen;

    Block random_block;
    foreach (height; iota(BlockCount).randomCover(rnd))
    {
        storage.readBlock(random_block, height);
        assert(random_block.header.height == height);
    }

    foreach (idx; iota(BlockCount).randomCover(rnd))
    {
        storage.readBlock(random_block, block_hashes[idx]);
        assert(hashFull(random_block.header) == block_hashes[idx]);
    }
}
```

source/agora/common/BlockStorage.d
```diff
@@ -758,3 +758,253 @@ public size_t readSizeType (File file) @trusted
...
+/*******************************************************************************
+
+    Define the Memory Storage for Blocks
+
+    Implemented using only memory without file IO.
+
+*******************************************************************************/
+
+public class MemoryStorage : IBlockStorage
```
AndrejMitrovic: `MemoryStorage` is too generic of a name. I would rename it to something like `MemBlockStorage` instead.

source/agora/common/BlockStorage.d
```diff
@@ -758,3 +758,253 @@ public size_t readSizeType (File file) @trusted
...
+
+    public this ()
+    {
+        this.height_idx = new IndexHeight();
+        this.hash_idx = new IndexHash();
+    }
+
+    /***************************************************************************
+
+        Save block to the file.
```
AndrejMitrovic: It doesn't store to the file.

source/agora/common/BlockStorage.d
```diff
@@ -758,3 +758,253 @@ public size_t readSizeType (File file) @trusted
...
+        catch (Exception ex)
+        {
+            this.writeLog("MemoryStorage.saveBlock: ", ex);
+            return false;
+        }
+    }
+
+    /***************************************************************************
+
+        Read block from the file.
```
AndrejMitrovic: It doesn't load from the file.

source/agora/common/BlockStorage.d
```diff
@@ -758,3 +758,253 @@ public size_t readSizeType (File file) @trusted
...
+        catch (Exception ex)
+        {
+            this.writeLog("MemoryStorage.readBlock: ", ex);
+            return false;
+        }
+    }
+
+    /***************************************************************************
+
+        Read block from the file.
```
AndrejMitrovic: Doesn't read from the file.

source/agora/common/BlockStorage.d
```diff
@@ -221,6 +221,9 @@ public class BlockStorage : IBlockStorage
         Params:
             findex = the index of the file.
 
+        Returns:
+            Returns true if success, otherwise returns false.
+
```
AndrejMitrovic: This looks good, but could you also rename the commit comment `"Fix document"` to `Add missing Return docs to BlockStorage`?

AndrejMitrovic:
These two commits should be merged into one:

```
- Implement an interface to create different types of block storage
- Add function 'getLastBlock()' to 'IBlockStorage'
```

Because this pull request introduces the interface, so it can already introduce `getLastBlock` in the commit that defines the interface.

So there should only be `- Implement an interface to create different types of block storage`.

source/agora/common/BlockStorage.d
```diff
@@ -32,6 +32,73 @@ import std.mmfile;
...
+            the highest block
+
+    ***************************************************************************/
+
+    public const(Block)* getLastBlock ();
+
+
+    /***************************************************************************
+
+        Save block to the file.
```
AndrejMitrovic: This could just say `Save block to the storage`, since it doesn't require it to be a file.

source/agora/common/BlockStorage.d
```diff
@@ -32,6 +32,73 @@ import std.mmfile;
...
+            Returns true if success, otherwise returns false.
+
+    ***************************************************************************/
+
+    public bool saveBlock (const ref Block block);
+
+
+    /***************************************************************************
+
+        Read block from the file.
```
AndrejMitrovic: Also `Read block from the storage.`

source/agora/common/BlockStorage.d
```diff
@@ -32,6 +32,73 @@ import std.mmfile;
...
+            Returns true if success, otherwise returns false.
+
+    ***************************************************************************/
+
+    public bool readBlock (ref Block block, size_t height);
+
+
+    /***************************************************************************
+
+        Read block from the file.
```
AndrejMitrovic: `Read block from the storage.`

source/agora/common/BlockStorage.d
```diff
@@ -110,6 +186,31 @@ public class BlockStorage
...
     }
 
+    /***************************************************************************
+
+        Returns:
+            the highest block
+
+    ***************************************************************************/
+
+    public const(Block)* getLastBlock () @safe nothrow
```
AndrejMitrovic: Ok, I just realized that my comment about merging the commits may be wrong. However, I am not sure this function is actually safe.

This function was implemented as a workaround for the `Ledger` using a `static Block` as a temporary, right? But here we are again returning a pointer to a temporary which will be overwritten as soon as `getLastBlock` is called again.

I'll have to think about how we can make this better.

MichaelKim20: Do you want me to do something?

AndrejMitrovic: I am writing a proposal, I will let you know soon!

AndrejMitrovic:
I would suggest having the same interface as `readBlock`. So something like `readLastBlock`.

Here's what `BlockStorage.d` might look like: [BlockStorage.d](https://gist.github.com/AndrejMitrovic/abf191ea3984f1cc55e2495f9e4aa68b)

And `Ledger.d`: [Ledger.d](https://gist.github.com/AndrejMitrovic/a51ad1f9a9fb2ae52c02910c034e7d25)

The diff:

```diff
diff --git a/source/agora/common/BlockStorage.d b/source/agora/common/BlockStorage.d
index 6c5aa78..857473e 100644
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -42,16 +42,6 @@ public interface IBlockStorage
 {
     @safe nothrow :
 
-    /***************************************************************************
-
-        Returns:
-            the highest block
-
-    ***************************************************************************/
-
-    public const(Block)* getLastBlock ();
-
-
     /***************************************************************************
 
         Save block to the storage.
@@ -82,7 +72,6 @@ public interface IBlockStorage
 
     public bool readBlock (ref Block block, size_t height);
 
-
     /***************************************************************************
 
         Read block from the storage.
@@ -97,6 +86,20 @@ public interface IBlockStorage
     ***************************************************************************/
 
     public bool readBlock (ref Block block, Hash hash);
+
+    /***************************************************************************
+
+        Read the last block from the storage.
+
+        Params:
+            block = will contain the block if the read was successful
+
+        Returns:
+            Returns true if success, otherwise returns false.
+
+    ***************************************************************************/
+
+    public bool readLastBlock (ref Block block);
 }
 
 
@@ -186,31 +189,6 @@ public class BlockStorage : IBlockStorage
         return buildPath(this.path, format("B%012d.dat", index));
     }
 
-    /***************************************************************************
-
-        Returns:
-            the highest block
-
-    ***************************************************************************/
-
-    public const(Block)* getLastBlock () @safe nothrow
-    {
-        if (this.height_idx.length == 0)
-            return null;
-
-        try {
-            if (this.length == ulong.max)
-                this.readBlockAtPosition(
-                    this.last_block, this.height_idx.back.position);
-            return &this.last_block;
-        }
-        catch (Exception ex)
-        {
-            this.writeLog("BlockStorage.getLastBlock: ", ex);
-            return null;
-        }
-    }
-
     /***************************************************************************
 
         Open memory mapped file.
@@ -366,6 +344,26 @@ public class BlockStorage : IBlockStorage
         }
     }
 
+    /***************************************************************************
+
+        Read the last block from the storage.
+
+        Params:
+            block = will contain the block if the read was successful
+
+        Returns:
+            Returns true if success, otherwise returns false.
+
+    ***************************************************************************/
+
+    public bool readLastBlock (ref Block block) @safe nothrow
+    {
+        if (this.height_idx.length == 0)
+            return false;
+
+        return this.readBlock(block, this.height_idx.back.position);
+    }
+
     /***************************************************************************
 
         Read block from the file.
@@ -821,9 +819,6 @@ public class MemBlockStorage : IBlockStorage
     // storage for all the blocks,
     private const(Block)[] blocks;
 
-    /// pointer to the latest block
-    private const(Block)* last_block;
-
     /// Index is block height
     private IndexHeight height_idx;
 
@@ -837,18 +832,6 @@ public class MemBlockStorage : IBlockStorage
         this.hash_idx = new IndexHash();
     }
 
-    /***************************************************************************
-
-        Returns:
-            the highest block
-
-    ***************************************************************************/
-
-    public const(Block)* getLastBlock () @safe nothrow
-    {
-        return this.last_block;
-    }
-
     /***************************************************************************
 
         Save block to array.
@@ -871,7 +854,6 @@ public class MemBlockStorage : IBlockStorage
             size_t block_position = this.blocks.length;
 
             this.blocks ~= block;
-            this.last_block = &this.blocks[$ - 1];
 
             // add to index of heigth
             this.height_idx.insert(
@@ -937,6 +919,26 @@ public class MemBlockStorage : IBlockStorage
         }
     }
 
+    /***************************************************************************
+
+        Read the last block from the storage.
+
+        Params:
+            block = will contain the block if the read was successful
+
+        Returns:
+            Returns true if success, otherwise returns false.
+
+    ***************************************************************************/
+
+    public bool readLastBlock (ref Block block) @safe nothrow
+    {
+        if (this.height_idx.length == 0)
+            return false;
+
+        return this.readBlock(block, this.height_idx.back.position);
+    }
+
     /***************************************************************************
 
         Read block from array.
diff --git a/source/agora/network/NetworkManager.d b/source/agora/network/NetworkManager.d
index b021565..be94183 100644
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -165,7 +165,7 @@ public class NetworkManager
             while (1)
             {
                 this.getBlocksFrom(
-                    ledger.getLastBlock().header.height + 1,
+                    ledger.getBlockHeight() + 1,
                     blocks => blocks.all!(block => ledger.acceptBlock(block)));
 
                 this.taskman.wait(2.seconds);
diff --git a/source/agora/node/Ledger.d b/source/agora/node/Ledger.d
index c930edb..f10c7a7 100644
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -113,20 +113,12 @@ public class Ledger
     {
         auto txs = this.pool.take(Block.TxsInBlock);
         assert(txs.length == Block.TxsInBlock);
-        auto block = makeNewBlock(*this.getLastBlock(), txs);
-        assert(this.acceptBlock(block));
-    }
-
-    /***************************************************************************
 
-        Returns:
-            the highest block
-
-    ***************************************************************************/
+        Block last_block;
+        assert(this.storage.readLastBlock(last_block));
+        auto block = makeNewBlock(last_block, txs);
 
-    public const(Block)* getLastBlock () @safe nothrow
-    {
-        return this.storage.getLastBlock();
+        assert(this.acceptBlock(block));
     }
 
     /***************************************************************************
@@ -138,11 +130,9 @@ public class Ledger
 
     public size_t getBlockHeight () @safe nothrow
     {
-        auto last_block = this.getLastBlock();
-        if (last_block !is null)
-            return last_block.header.height;
-        else
-            return 0;
+        Block last_block;
+        assert(this.storage.readLastBlock(last_block));
+        return last_block.header.height;
     }
 
     /***************************************************************************
@@ -196,11 +186,9 @@ public class Ledger
 
     private bool isValidBlock (const ref Block block) nothrow @safe
     {
-        auto last_block = this.getLastBlock();
-        const expected_height = last_block !is null
-            ? (last_block.header.height + 1)
-            : 0;
-        return block.header.height == expected_height;
+        Block last_block;
+        assert(this.storage.readLastBlock(last_block));
+        return block.header.height == last_block.header.height + 1;
     }
 
     /***************************************************************************
@@ -279,8 +267,10 @@ unittest
     auto pool = new TransactionPool(":memory:");
     scope(exit) pool.shutdown();
     scope ledger = new Ledger(pool, storage);
-    assert(*ledger.getLastBlock() == getGenesisBlock());
+
     assert(ledger.getBlockHeight() == 0);
+    const(Block)[] blocks = ledger.getBlocksFrom(0, 10);
+    assert(blocks[$ - 1] == getGenesisBlock());
 
     auto gen_key_pair = getGenesisKeyPair();
     Transaction[] last_txs;
@@ -299,14 +289,14 @@ unittest
     }
 
     genBlockTransactions(2);
-    const(Block)[] blocks = ledger.getBlocksFrom(0, 10);
+    blocks = ledger.getBlocksFrom(0, 10);
     assert(blocks[0] == getGenesisBlock());
     assert(blocks[0].header.height == 0);
     assert(blocks.length == 3);  // two blocks + genesis block
 
     /// now generate 98 more blocks to make it 100 + genesis block (101 total)
     genBlockTransactions(98);
-    assert(ledger.getLastBlock().header.height == 100);
+    assert(ledger.getBlockHeight() == 100);
 
     blocks = ledger.getBlocksFrom(0, 10);
     assert(blocks[0] == getGenesisBlock());
```

AndrejMitrovic:
The important thing here is that we don't return temporaries to values on the stack.

AndrejMitrovic:
Rephrased: I mean it's important that the pointer to the value that we return is not changed between function calls. Otherwise we might overwrite one value which might still be read somewhere else.

source/agora/common/BlockStorage.d
```diff
@@ -793,3 +793,263 @@ public size_t readSizeType (File file) @trusted
...
+        auto finds
+            = this.height_idx[].find!( (a, b) => a.height == b)(height);
+
+        if (finds.empty)
+            return false;
+
+        try
+        {
+            () @trusted {
+                block = cast(Block)this.blocks[finds.front.position];
```
AndrejMitrovic: This is the problematic part, because it's casting away const.

However I have an idea how to solve it. Using serialization.

AndrejMitrovic: @Geod24 this code is in `MemBlockStorage`, which is used in unittests. 

I only see two options here:

- Instead of storing `const(Block)[]`, store a serialized `ubyte[]` and deserialize the block here.
- Do a deep copy of the struct to get a mutable Block

Can you think of another solution?

MichaelKim20: It's a good solution. I checked the review and have already fixed it.

AndrejMitrovic (dismissed):

AndrejMitrovic:
Great work!

MichaelKim20:
@AndrejMitrovic It's thanks to you.

TrustHenry:
It's green :)

AndrejMitrovic:
cc @Geod24

TrustHenry:
It's dockercloud green.

AndrejMitrovic:
There is a small conflict. But all you have to do is rename `findOutput` to `findUTXO` when rebasing.

source/agora/common/BlockStorage.d
AndrejMitrovic: What happened to the code which fixed the cast? It seems to have disappeared.

TrustHenry:
rebase done :)

AndrejMitrovic:
```D
                block = cast(Block)this.blocks[finds.front.position];
```

This part needs to be fixed. There was already code which fixed this, but it got (accidentally?) reverted. Could you please restore the code which used serialization to avoid casting?

AndrejMitrovic:
Take a look here: https://github.com/bpfkorea/agora/compare/8ef93bec68650f1b26b986a90cfdb8ee6302eda5..c1d467552fa13b880f2be7250e9c27e9c6776369

At the very end of the diff in `source/agora/common/BlockStorage.d`. Those changes need to be reverted.

AndrejMitrovic (approved):

TrustHenry:
It's all green.........

AndrejMitrovic:
> It’s all green……

I know, but I want @Geod24 to look at your PR before merging it.

source/agora/common/BlockStorage.d
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
 
-    Defines the memory mapped file of a block.
+    Define the Storage for Blocks
```
Geod24: English nitpick: Why are `Storage` and `Blocks` capitalized ?

source/agora/common/BlockStorage.d
```diff
@@ -32,6 +32,78 @@ import std.mmfile;
 import std.path;
 import std.stdio;
 
+/*******************************************************************************
+
+    Define the Storage for Blocks
```
Geod24: Same thing w.r.t. capitalization

source/agora/common/BlockStorage.d
```diff
@@ -793,3 +793,263 @@ public size_t readSizeType (File file) @trusted
     file.rawRead(buffer);
     return *cast(size_t*)&(buffer[0]);
 }
+
+
+/*******************************************************************************
+
+    Define the Memory Storage for Blocks
```
Geod24: Capitalization

source/agora/common/BlockStorage.d
Geod24: Indeed this cast is problematic.

source/agora/common/BlockStorage.d
```diff
@@ -793,3 +793,263 @@ public size_t readSizeType (File file) @trusted
...
+            block = `Block` to save
+
+        Returns:
+            Returns true if success, otherwise returns false.
+
+    ***************************************************************************/
+
+    public bool saveBlock (const ref Block block) @safe nothrow
+    {
+        try
```
Geod24: Where do you expect this to `throw` ?

source/agora/node/Ledger.d
```diff
@@ -63,16 +62,15 @@ public class Ledger
 
     ***************************************************************************/
 
-    public bool acceptBlock (const ref Block block) nothrow @safe
+    public bool acceptBlock (const ref Block block) @safe
```
Geod24: Why was `nothrow` removed ?

source/agora/node/Ledger.d
```diff
@@ -115,20 +113,24 @@ public class Ledger
     {
         auto txs = this.pool.take(Block.TxsInBlock);
         assert(txs.length == Block.TxsInBlock);
-        auto block = makeNewBlock(*this.last_block, txs);
+        Block last_block;
+        assert(this.storage.readLastBlock(last_block));
```
```d
if (!this.storage.readLastBlock(last_block))
    assert(0);
```

The reason it is written this way is that `assert` are removed in `release` mode.
Which means `readLastBlock` would disappear (!) and the program would not perform correctly.
With `if` + `assert(0)` this does not happen, as `assert(0)` are never removed.

source/agora/node/Ledger.d
```diff
@@ -147,14 +149,23 @@ public class Ledger
     ***************************************************************************/
 
     public const(Block)[] getBlocksFrom (ulong block_height, size_t max_blocks)
-        @safe nothrow @nogc
+        @safe
```
Geod24: Why not `nothrow` ?

source/agora/node/Ledger.d
```diff
@@ -171,13 +182,11 @@ public class Ledger
...
+    private bool isValidBlock (const ref Block block) nothrow @safe
     {
-        const expected_height = this.last_block !is null
-            ? (this.last_block.header.height + 1)
-            : 0;
-
-        return block.header.height == expected_height;
+        Block last_block;
+        assert(this.storage.readLastBlock(last_block));
+        return block.header.height == last_block.header.height + 1;
```
Geod24: Can't we just use `getBlockHeight` here ?

source/agora/node/Ledger.d
```diff
@@ -193,10 +202,14 @@ public class Ledger
...
     private const(Output)* findUTXO (Hash tx_hash, size_t index)
-        @safe nothrow @nogc
+        @safe nothrow
     {
-        foreach (ref block; this.ledger)
+        foreach (height; 0 .. this.getBlockHeight() + 1)
         {
+            Block block;
+            if (!this.storage.readBlock(block, height))
+                break;
```
Geod24: Make it an `assert(0)` for now

source/agora/node/Node.d
```diff
@@ -230,6 +234,28 @@ public class Node : API
...
+        Returns:
+            Returns instance of `MemoryStorage` if data_dir is empty,
+            otherwise returns instance of `BlockStorage`
+
+    ***************************************************************************/
+
+    protected IBlockStorage getBlockStorage (string data_dir) @system
+    {
+        if (data_dir == "")
+            return new MemBlockStorage();
```
Geod24, line 256: It shouldn't be based on the runtime data, because someone could make a mistake and pass a string which is `null` by accident. We usually use `version (unittest)` or override.

Geod24 (commented):
A few nitpicks

Geod24:
It seems we need a fix to add `genesis` into a `BlockStorage`.

source/agora/common/BlockStorage.d
Geod24: Hum, Phobos' `RedBlackTree` is not `nothrow`.

AndrejMitrovic:
There are some styling issues, but we can fix this in another PR.

AndrejMitrovic (approved):

Geod24 (approved):

MichaelKim20 (approved):

PR #166, opened 2019-08-12
Improve retryFor() and use it everywhere

AndrejMitrovic:


AndrejMitrovic:
Hold on, I didn't rebase against v0.x.x, there's more tests to fix.

AndrejMitrovic:
Good, using this function actually uncovered there is something wrong with the BanManager tests / implementation. I'm looking into it.

AndrejMitrovic:
Ok, hopefully this doesn't fail, unless there's more stuff to fix here.

AndrejMitrovic:
hmm failure. ok.

AndrejMitrovic:
Ah I see another test-case we have here that spawns a lot of transactions. It creates 100 blocks, so that's 800 transactions. It just seems too much to send them all at once for that Travis machine. I'll reduce this.

AndrejMitrovic:
🥗

source/agora/test/Base.d
```diff
@@ -489,8 +492,13 @@ public void retryFor (lazy bool check, Duration timeout,
         Thread.sleep(SleepTime.msecs);
     }
 
-    throw new AssertError(format("Check condition failed after timeout of %s",
-        timeout), file, line);
+    auto assert_msg = format("Check condition failed after timeout of %s " ~
+        "and %s attempts", timeout, TotalAttempts);
+
+    if (msg.length)
+        assert_msg ~= ": " ~ msg;
```
Geod24, line 499: This actually evaluate `msg` twice

source/agora/test/GossipProtocol.d
```diff
@@ -52,12 +52,10 @@ unittest
         // send it to one node
         txs.each!(tx => node_1.putTransaction(tx));
 
-        Thread.sleep(50.msecs);  // await gossip
```
Geod24, line 55: Oh yeah I can see the problem here :D

source/agora/test/Ledger.d
```diff
@@ -53,10 +53,9 @@ unittest
         // send it to one node
         txs.each!(tx => node_1.putTransaction(tx));
 
-        Thread.sleep(50.msecs);  // await gossip and block creation
-
         nodes.enumerate.each!((idx, node) =>
-            assert(node.getBlockHeight() == block_idx + 1,
+            retryFor(node.getBlockHeight() == block_idx + 1,
```
Geod24, line 57: Note: If you still want to use UFCS, you can just remove the `assert` and leave the parenthesis.
But I'm not sure we should go down this road 🤔

source/agora/test/NetworkManager.d
```diff
@@ -43,19 +43,9 @@ unittest
     auto txes = makeChainedTransactions(getGenesisKeyPair(), null, 1);
     txes.each!(tx => node_1.putTransaction(tx));
 
-    Thread.sleep(2000.msecs);
```
Geod24, line 46: Ouch

source/agora/test/BanManager.d
```diff
@@ -69,71 +70,41 @@ unittest
...
-    assert(node_2.getBlockHeight() == 5);
-
-    new_tx = genBlockTransactions(1);
-    left_txs ~= new_tx;
-    new_tx.each!(tx => node_1.putTransaction(tx));
-    Thread.sleep(500.msecs);
-    assert(node_1.getBlockHeight() == 6);
-    assert(node_2.getBlockHeight() == 5);  // filter is active
+    // wait for node 2 to be banned and all putTransaction requests to time-out
+    Thread.sleep(2.seconds);
```
Geod24, line 81: The timeout is `10` (msecs?) so why wait 2 seconds ?

source/agora/test/Base.d
AndrejMitrovic: Interesting to think about. Will fix.

source/agora/test/Ledger.d
AndrejMitrovic: Yeah I know, I just found it a bit weird that way.

source/agora/test/BanManager.d
AndrejMitrovic: It keeps polling with a 10 msec timeout until a total of 2 seconds are reached. It will either finish early or throw.

Geod24: I might have forgotten the details here, so my bad, but what will lead to node 2 being banned ?

AndrejMitrovic: 32 failed putTransaction's.

Geod24 (approved):

PR #170, opened 2019-08-12
Move Block and Transaction to agora.consensus.data

Geod24:
Not perfect yet but better than in `common`.
What do you think @AndrejMitrovic ?


AndrejMitrovic:
I think this looks like a good start.

AndrejMitrovic (approved):

PR #168, opened 2019-08-12
Change memory map size & fix up styling

Geod24:
Just a few refactoring we can already integrate.

PR #172, opened 2019-08-12
Increase the size of the memory map to 640KByte

MichaelKim20:
Increase the size of the memory map to 640.
To solve problems that arise from large multi-transaction blocks. Unable to put 100 transactions in 64KByte's memory.
This problem will soon be fixed.

### DIFF
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -43,7 +43,7 @@ private immutable size_t ReserveSize = 64 * 1024;
 private immutable ulong MaxBlock = 100;
 
 /// The map file size
-private immutable size_t MapSize = 64 * 1024;
+private immutable size_t MapSize = 640 * 1024;
 
 private struct HeightPosition
 {


AndrejMitrovic:
Where is the test-case?

AndrejMitrovic:
And, I don't think this will solve the issue. A transaction can contain a large number of inputs and outputs. I don't see how this will allow supporting arbitrary "100" transactions in the block.

AndrejMitrovic:
So from my understanding, [this commit here](https://github.com/bpfkorea/agora/pull/168/commits/dc2d56e32b7f27bb52a815dbed6062647e22a0c7) should not have been included when being merged?

Then I suggest doing a revert commit instead of a regular commit:

$ git checkout mfile-size
$ git reset --hard HEAD~1
$ git revert bb28d627b01f621fd61e89ed137a3a12db779273

And then add a message after the "Reverted commit" part.

AndrejMitrovic:
Thanks, looks good now!

AndrejMitrovic (approved):

PR #175, opened 2019-08-13
Get rid of templated Network parameter

AndrejMitrovic:
I will need this for some dependency injection in a future unittest. While refactoring, I've realized we don't actually need this templated parameter anymore.


Geod24 (approved):
Love it

PR #177, opened 2019-08-13
Retrieve blocks from multiple nodes and add a test-case when bad blocks are retrieved

AndrejMitrovic:


source/agora/test/Base.d
```diff
@@ -351,6 +354,17 @@ public final class TestNode : Node, TestAPI
         this.ledger = new TestLedger(pool);
         return this.ledger;
     }
+
+    /// Add a set of invalid blocks to the Ledger, skipping validation
+    public override void makeBadBlocks (size_t height, size_t count)
```
Geod24: This just makes empty blocks right ? Not "meaningfully bad" blocks.

Geod24:
I'm not sure about the dependency injection on the `Ledger` level. It does not sound like something that should be replaced.

AndrejMitrovic:
Actually come to think of it, I could make a fake node that returns invalid data directly in the `getBlocksFrom` API, without touching the Ledger at all.

Geod24:
Commit 1 LVGTM.
Commit 2 I'm skeptical because just generating empty blocks does not sound generic enough to be the default implementation. I'd much rather have a `Block` being sent from the client. I *think* we could make `makeBlock` be `pure` (so that we get `immutable` blocks).
Commit 3 is not used.
Commit 4 LGTM

AndrejMitrovic:
3 is technically used by 4, but I will rewrite this to use a customly inherited Node. It's better than dependency injection.

AndrejMitrovic:
Reworked the PR.

Instead of injecting dependencies, I made a custom `TestNode` subclass which returns invalid blocks.

Note that currently there is limited block validation (only the block height is checked). This will be improved in future PRs.

AndrejMitrovic:
Rebased.

source/agora/network/NetworkManager.d
```diff
@@ -235,35 +241,34 @@ public class NetworkManager
...
+            .each!(pair => node_pairs ~= pair);
 
-            const highest_block = node_pair[0];
-            if (highest_block < block_height)
-                return;  // we're up to date
+        node_pairs.sort!((a, b) => a.height < b.height);
 
-            logInfo("Retrieving latest blocks..");
-            auto node = node_pair[1];
+        LNextNode: foreach (pair; node_pairs.retro) try
```
Geod24: You could just change the ordering and wouldn't need `retro`, no ?

source/agora/test/NetworkManager.d
```diff
@@ -49,3 +49,114 @@ unittest
...
+
+    static class BadAPIManager : TestAPIManager
+    {
+        // base class uses a hashmap, can't depend on the order of nodes
+        public RemoteAPI!TestAPI[] nodes;
+
+        /// Initialize a new node
+        public override void createNewNode (PublicKey address, Config conf)
+        {
+            static bool first_node = true;
```
Geod24: `first_node == (this.nodes.length == 0)`

source/agora/test/NetworkManager.d
```diff
@@ -49,3 +49,114 @@ unittest
...
+    // enable filtering first
+    node_good.filter!(API.getBlocksFrom);
+    node_bad.filter!(API.getBlocksFrom);
+    node_test.filter!(API.putTransaction);
+
+    // make 10 good blocks
+    auto txes = makeChainedTransactions(getGenesisKeyPair(), null, 10);
+    txes.each!(tx => node_good.putTransaction(tx));
+
+    // at this point both the good node and bad node have same amount of blocks
```
Geod24:
```suggestion
    // at this point both the good node and bad node have same amount of blocks, but bad node pretends to have + 10
```

source/agora/test/Base.d
```diff
@@ -131,7 +131,7 @@ public class TestAPIManager
     public RemoteAPI!TestAPI[PublicKey] apis;
 
     /// Initialize a new node
-    protected void createNewNode (PublicKey address, Config conf)
+    public void createNewNode (PublicKey address, Config conf)
```
Geod24, line 134: It seems `protected` is enough, according to the next commit

source/agora/test/Base.d
```diff
@@ -245,7 +245,7 @@ public class TestNetworkManager : NetworkManager
     import geod24.LocalRest;
 
     /// Workaround compiler bug that triggers in `std.concurrency`
-    protected __gshared std.concurrency.Tid[string] tbn;
+    public __gshared std.concurrency.Tid[string] tbn;
```
Geod24, line 248: Same thing about `protected`

source/agora/network/NetworkManager.d
AndrejMitrovic: LOL

source/agora/test/NetworkManager.d
AndrejMitrovic: It's a static initializer, that can't work.

AndrejMitrovic: Ah nevermind.

AndrejMitrovic: See what I mean? My brain..

source/agora/test/Base.d
AndrejMitrovic: `createNewNode` needs to be public as it's accessed from the `Base` module, and DMD has a bug: https://issues.dlang.org/show_bug.cgi?id=20002

source/agora/test/Base.d
AndrejMitrovic: Well `BadAPIManager` is not a subclass of `TestNetworkManager`, but of `TestAPIManager`, so it needs to be public. 

This is after I split the two classes into `TetAPIManager` and `TestNetworkManager`. I didn't feel it made sense to have it one class just in order to keep one `__gshared` field `protected.`

Geod24: Fair enough

AndrejMitrovic:
Fixed the rest of the bugs.

Geod24 (approved):

PR #179, opened 2019-08-13
Short-circuit when blocks fail validation.

AndrejMitrovic:
If one in a series of blocks which another node sent to a node fail validation, then there is no reason to continue reading the rest of the batch of blocks as they will most likely fail validation too.

I don't have a test-case for this yet until https://github.com/bpfkorea/agora/pull/177 is merged.


AndrejMitrovic:
This is now part of https://github.com/bpfkorea/agora/pull/177

PR #180, opened 2019-08-13
Going away from Object Oriented Interface imposed by Vibe.d

denizzzka:
(Do not merge! Tests are broken by now!)

### DIFF
--- a/source/agora/network/NetworkClient.d
+++ b/source/agora/network/NetworkClient.d
@@ -21,7 +21,7 @@ import agora.common.Data;
 import agora.common.Set;
 import agora.common.Task;
 import agora.consensus.data.Transaction;
-import agora.node.API;
+import agora.node.Node: Node;
 
 import vibe.core.log;
 
@@ -31,6 +31,23 @@ import std.format;
 import std.random;
 
 import core.time;
+
+/// The network state (completed when sufficient validators are connected to)
+public enum NetworkState
+{
+    Incomplete,
+    Complete
+}
+
+/// Contains the network info (state & addresses)
+public struct NetworkInfo
+{
+    /// Whether the node knows about the IPs of all its quorum set nodes
+    public NetworkState state;
+
+    /// Partial or full view of the addresses of the node's quorum (based on is_complete)
+    public Set!string addresses;
+}
 
 /// Used for communicating with a remote node
 class NetworkClient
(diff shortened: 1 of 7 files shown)


source/agora/network/NetworkManager.d
```diff
@@ -287,7 +288,8 @@ public class NetworkManager
 
         logInfo("Establishing connection with %s...", address);
         auto node = new NetworkClient(this.taskman, this.banman, address,
-            this.getClient(address, this.node_config.timeout.msecs),
+            null, //FIXME: temporary
+            //~ this.getClient(address, this.node_config.timeout.msecs), //FIXME: what's going on here?!
```
Geod24, line 292: This line is one of the most important of the module.
It instantiate a client to send requests to remote nodes.

denizzzka: yes, that's why the tests are broken now.

source/agora/network/NetworkClient.d
```diff
@@ -52,8 +69,7 @@ class NetworkClient
     /// Ban manager
     private BanManager banman;
 
-    /// API client to the node
-    private API api;
```
Geod24, line 56: I don't see how this could work. Node is the business code of the server. `API` is the public interface, which can be used to instantiate a client. A `NetworkClient` is just an instance owned by `NetworkManager` that uses an `API` in order to perform queries.

If you store a `Node` here, you basically store another node in a node, instead of storing a client.

denizzzka: You're right, it's still not finished - api substituted by node just for compilation check

source/agora/network/rest_api.d
```diff
@@ -80,8 +118,12 @@ public struct NetworkInfo
 
 *******************************************************************************/
 @path("/")
-public interface API
+interface VibeRestAPI
 {
+    void metaAddPeer (string peer);
+    public void start ();
+    public void shutdown ();
```
Geod24, line 125: Those don't belong to the public interface. Since this is the `VibeRestAPI`, you can't just expose `start` and `shutdown`...

denizzzka: it is temporary so that the compilation does not break

denizzzka:
decided as wrong idea

PR #150, opened 2019-08-06
Adds support of HTTP clients timeouts

denizzzka:
Do not merge before https://github.com/vibe-d/vibe.d/pull/2344 will be merged

PR #181, opened 2019-08-13
Splits API interface to API and VibeRestAPI

denizzzka:
It is only cosmetic change what isn't touches anything but makes things easier to understand and maybe more flexible for further wire protocols implementation.

### DIFF
--- a/source/agora/network/RestApi.d
+++ b/source/agora/network/RestApi.d
@@ -1,28 +1,6 @@
 /*******************************************************************************
 
-    Definitions of the nodes (full node & validator) APIs
-
-    Two kinds of nodes exist: full nodes, and validators.
-    A full node follows the network as a passive actor, but does validation
-    on the data it receives, and can forward that data to other nodes.
-    A validator is a full node which participates in consensus.
-
-    An `API` is used as an interface to communicate with a node.
-    As such, a class that implements `API` exists (in `agora.node.Node`),
-    and in order to communicate with other nodes, it holds an `API` for
-    each of those nodes.
-
-    `API` are defined as D interfaces, following what is done in Vibe.d.
-    Those interfaces can be read by a generator to build a client or a server.
-    One such generator is Vibe.d's `vibe.web.rest`. `RestInterfaceClient`
-    allows to query a REST API, while `registerRestInterface` will route queries
-    and deserialize parameters according to the interface's definition.
-
-    Another generator which we use for unittests is "LocalRest".
-    It allows to start a node per thread, and uses `std.concurrency`
-    to do message passing between nodes.
-
-    Lastly, we plan to implement a generator which works directly on TCP/IP.
+    Definitions of the nodes (full node & validator) REST APIs
 
     Copyright:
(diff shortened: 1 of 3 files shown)


source/agora/network/RestApi.d
```diff
@@ -0,0 +1,147 @@
...
+        Params:
+            block_height = Height of the block that contains the transaction hash
+            hash         = transaction hash
+
+        Returns:
+            the array of hashes which form the merkle path
+
+    ***************************************************************************/
+
+    public Hash[] getMerklePath (ulong block_height, Hash hash);
```
Geod24: Those functions don't need to be redefined since they are in the base API.

Geod24:
This makes much more sense, yes. However, I don't see it bringing any value at the moment. The REST attributes being on the `interface` can be seen as inelegant, but they are not so annoying that they require so much duplication.

Geod24:
To clarify: If it becomes a necessity in the future, we can do this. But there's no clear use case for it at the moment.

source/agora/network/RestApi.d
denizzzka: Fixed

source/agora/network/RestApi.d
```diff
@@ -0,0 +1,61 @@
...
+        Returns:
+            Return true if the node has this transaction hash.
+
+        API:
+            GET /hasTransactionHash
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.GET)
+    public bool hasTransactionHash (Hash tx);
```
Geod24: Does `override` works here ?

denizzzka: Yes, works with and without it

denizzzka:
Weird PR, yes. But let it be - for some elusive reason I’m so calmer with it. :-)

Geod24:
> To clarify: If it becomes a necessity in the future, we can do this. But there's no clear use case for it at the moment.

denizzzka:
Value of this PR is in adding clarity.

AndrejMitrovic:
It's not adding any clarity whatsoever. There should be a single interface defining all the public-facing API entry-points for the node.

denizzzka:
> It's not adding any clarity whatsoever. There should be a single interface defining all the public-facing API entry-points for the node.

...and another one for Vibe rest generator to distinguish between API and Vibe.d REST API implementation (what accidentally looks now like D's interface description block equal to API interface block).

This PR leads to avoid exposing symbols that are unnecessary in a particular interface, add necessary ones, change their ​​if necessary (https://github.com/bpfkorea/agora/pull/150#discussion_r311352084). Say, different wiring schemes can use different addresses or may have additional calls.

Related: https://github.com/Geod24/localrest/pull/20 - in my opinion, including therefore no one knew that LocalRest actually is not implements REST

Geod24:
As there is not any consensus in favor of this, I am going to close this.
As mentioned in our discussions, we can re-discuss the issue once the TCP communication protocol is ready.

PR #182, opened 2019-08-13
TODO about address added

denizzzka:


AndrejMitrovic:
Could you just make this a github issue instead? That way we can track it.

PR #184, opened 2019-08-14
Refactor Transaction verification & improve attributes

AndrejMitrovic:
The UTXO finder becoming public will be used in a separate PR which will add the Block verification function which also uses the UTXO finder.


Geod24 (approved):
Neat!

Geod24:
By the way, the `Transaction` param could also be made `ref` to avoid unnecessary copies

AndrejMitrovic:
> By the way, the Transaction param could also be made ref to avoid unnecessary copies

Yes I thought about it, however it only contains two dynamic arrays. I'm not sure whether `ref` would buy us anything here (although we could go with auto ref I suppose).

AndrejMitrovic:
The `@nogc` was over-eager. I will end up removing it (or Henry might do it too in his PR).

We cannot really avoid allocations at this point.

PR #185, opened 2019-08-14
Rename mergeHash() to hashMulti() and make it generic

AndrejMitrovic:
If you have an idea for a better name, please let me know.

### DIFF
--- a/source/agora/common/Hash.d
+++ b/source/agora/common/Hash.d
@@ -233,3 +233,15 @@ public Hash mergeHash (Hash h1, Hash h2) nothrow @nogc @safe
 
     return hashFull(MergeHash(h1, h2));
 }
+
+///
+unittest
+{
+    Hash foo = hashFull("foo");
+    Hash bar = hashFull("bar");
+    const merged = Hash(
+        "0xe0343d063b14c52630563ec81b0f91a84ddb05f2cf05a2e4330ddc79bd3a06e57" ~
+        "c2e756f276c112342ff1d6f1e74d05bdb9bf880abd74a2e512654e12d171a74");
+
+    assert(mergeHash(foo, bar) == merged);
+}
--- a/source/agora/common/Hash.d
+++ b/source/agora/common/Hash.d
@@ -162,36 +162,6 @@ public void hashPart (scope const(char)[] record, scope HashDg state) /*pure*/ n
     state(cast(const ubyte[])record);
 }
 
-/*******************************************************************************
-
-    Merges two hashes together
-
-    Params:
-        h1 = the first hash
-        h2 = the second hash
-
-    Returns:
-        the merged hash
-
-*******************************************************************************/
-
-public Hash mergeHash (Hash h1, Hash h2) nothrow @nogc @safe
-{
-    static struct MergeHash
-    {
-        Hash left;
-        Hash right;
-
-        public void computeHash (scope HashDg dg) const nothrow @safe @nogc
-        {
-            hashPart(this.left, dg);
-            hashPart(this.right, dg);
-        }
-    }
-
-    return hashFull(MergeHash(h1, h2));
-}
-
 // Test that the implementation actually matches what the RFC gives
 nothrow @nogc @safe unittest
 {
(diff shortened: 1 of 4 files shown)


source/agora/common/Hash.d
```diff
@@ -233,3 +203,65 @@ nothrow @nogc @safe unittest
...
+    Params:
+        T = variadic argument types
+        args = the arguments
+
+    Returns:
+        the hash of all the arguments
+
+*******************************************************************************/
+
+public Hash hashMulti (T...)(T args) nothrow @nogc @safe
```
Geod24:
```suggestion
public Hash hashMulti (T...)(auto ref T args) nothrow @nogc @safe
```

source/agora/common/Hash.d
```diff
@@ -233,3 +203,65 @@ nothrow @nogc @safe unittest
...
+{
+    static struct HashMulti
+    {
+        T fields;
+
+        public void computeHash (scope HashDg dg) const nothrow @safe @nogc
+        {
+            foreach (const ref field; this.fields)
```
Geod24: Whouldn't we be able to do:
```d
static foreach (idx, ref _; args)
    hashPart(args[idx], dg);
```

Or something similar ?

AndrejMitrovic: Yep.

AndrejMitrovic: Ok so I can do it, but only if I make the struct non-static. And that means:

```
source/agora/common/Hash.d(220,13): Error: function `agora.common.Hash.hashMulti!(BitBlob!512LU, BitBlob!512LU).hashMulti` is @nogc yet allocates closures with the GC
```

So I think s/Yep/Nope.

Geod24: I mean you can foreach on the tuple, so you don't need the struct to begin with.

AndrejMitrovic: I could duplicate some of the stuff done in hashFull like:

```D
    Hash hash = void;
    crypto_generichash_state state;
    crypto_generichash_init(&state, null, 0, Hash.sizeof);
    scope HashDg dg = (scope const(ubyte)[] data) @trusted
                     => cast(void)crypto_generichash_update(&state, data.ptr, data.length);
    hashPart(record, dg);
    crypto_generichash_final(&state, hash[].ptr, Hash.sizeof);
    return hash;
```

That could work.

AndrejMitrovic:
Updated. I had to make the function `@trusted` because it's dealing with pointers on almost every statement level.

source/agora/common/Hash.d
```diff
@@ -233,3 +203,62 @@ nothrow @nogc @safe unittest
...
+
+public Hash hashMulti (T...)(T args) nothrow @nogc @trusted
+{
+    Hash hash = void;
+    crypto_generichash_state state;
+    crypto_generichash_init(&state, null, 0, Hash.sizeof);
+    scope HashDg dg = (scope const(ubyte)[] data) @trusted
+        => cast(void)crypto_generichash_update(&state, data.ptr, data.length);
+    static foreach (arg; args)
+        hashPart(arg, dg);
```
Geod24: That static foreach can't be `@trusted`

AndrejMitrovic: Pourquoi?

AndrejMitrovic: ah

AndrejMitrovic: I see what you mean.

source/agora/common/Hash.d
```diff
@@ -233,3 +203,62 @@ nothrow @nogc @safe unittest
...
+*******************************************************************************/
+
+public Hash hashMulti (T...)(T args) nothrow @nogc @trusted
+{
+    Hash hash = void;
+    crypto_generichash_state state;
+    crypto_generichash_init(&state, null, 0, Hash.sizeof);
+    scope HashDg dg = (scope const(ubyte)[] data) @trusted
+        => cast(void)crypto_generichash_update(&state, data.ptr, data.length);
+    static foreach (arg; args)
```
Geod24:
```suggestion
    static foreach (ref arg; args)
```

~But as I mentioned, `static foreach (idx, ref _; args)` because https://issues.dlang.org/show_bug.cgi?id=16521~

source/agora/common/Hash.d
Geod24: It still is.
You need the rest of the function to be `@trusted`, but `arg` could be anything, including a `system` `opCall`.

AndrejMitrovic: Indeed. And also hashPart too. I'll fix this in a bit.

source/agora/common/Hash.d
AndrejMitrovic: If I make it `auto ref` here, then I can't seem to use `ref` in the foreach loop. I'm guessing I'm hitting a language limitation here?

AndrejMitrovic: Ah, I think this is a limitation with `static foreach`.

AndrejMitrovic:
Ok done pushing, this one should work.

Geod24 (approved):

PR #190, opened 2019-08-14
What's wrong with this code?

denizzzka:


AndrejMitrovic (commented):
You need to call shutdown() to release the SQLite resources. It cannot be done from a destructor due to the d2sqlite wrapper's design. But I hope this is fixed one day.

denizzzka:
Thanks!

```D
unittest
{
    import d2sqlite3: shutdown;

    auto tn = makeTestNetwork(NetworkTopology.Simple, 2);
    scope(exit) tn.shutdown();
}
```

PR #192, opened 2019-08-16
test-runner: Add -vcolumns to RDMD

AndrejMitrovic:


tests/runner.d
```diff
@@ -35,16 +35,17 @@ private int main (string[] args)
...
     foreach (test; dirEntries(UnitPath, SpanMode.shallow))
     {
         writeln("Running test on ", test);
         auto pp = pipeProcess(Args ~ test);
         if (pp.pid.wait() != 0)
         {
-            pp.stdout.byLine.each!(a => writeln("[stdout]\t", a));
-            pp.stderr.byLine.each!(a => writeln("[stderr]\t", a));
+            pp.stdout.byLine.each!(writeln);
+            pp.stderr.byLine.each!(line => stderr.writeln(line));
```
AndrejMitrovic: I can't seem to do `pp.stderr.byLine.each!(stderr.writeln)` here. I get an unusual error:

```D
/Library/D/dmd/src/phobos/std/algorithm/iteration.d(936,48): Error: need `this` for `writeln` of type `@safe void(char[] _param_0)`
```

AndrejMitrovic:
I mainly did this so I can press F2 and go to the offending line when there's a compilation error.

tests/runner.d
```diff
@@ -35,16 +35,17 @@ private int main (string[] args)
         return 1;
     const Args = ["rdmd", ] ~ args[1 .. $] ~
         importPaths.map!(v => "-I" ~ v).array ~
-        lflags.map!(v => "-L" ~ v).array;
+        lflags.map!(v => "-L" ~ v).array ~
+        "-vcolumns";
```
Geod24: Move it after `"rdmd"`

tests/runner.d
Geod24: It's not an error message coming from the runner hence it shouldn't go to the runner's `stderr` output.

Geod24:
> I mainly did this so I can press F2 and go to the offending line when there's a compilation error.

Can't you Command+click ?

tests/runner.d
AndrejMitrovic: I mean.. the runner itself should just not error at all.

Geod24: I guess we could remove it but having the distinction seemed useful originally.

tests/runner.d
```diff
@@ -33,7 +33,7 @@ private int main (string[] args)
     const lflags = getLflags();
     if (!importPaths.length || !lflags.length)
         return 1;
-    const Args = ["rdmd", ] ~ args[1 .. $] ~
+    const Args = ["rdmd", ] ~ "-vcolumns" ~ args[1 .. $] ~
```
Geod24: I meant:
```suggestion
    const Args = ["rdmd", "-vcolumns", ] ~ args[1 .. $] ~
```

tests/runner.d
Geod24: Okay let's just go with this.

AndrejMitrovic:
Fixed, and removed second commit since I can just `grep` in my build script to strip out the `[stderr]` stuff.

Geod24 (approved):

PR #193, opened 2019-08-16
Implement full block validation

AndrejMitrovic:
Note: Checking for double-spend requires a UTXO set, and this is part of my upcoming PR.

However, the unittests do include tests for double-spend.

### DIFF
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -16,6 +16,7 @@ module agora.consensus.Validation;
 import agora.common.Amount;
 import agora.common.crypto.Key;
 import agora.common.Hash;
+import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
 
 /// Delegate to find an unspent UTXO
@@ -216,4 +217,44 @@ unittest
     storage[tx2Hash] = tx2;
     // Signature verification must be error
     assert(!tx2.isValid(findUTXO), format("Transaction signature is not validated %s", tx2));
+}
+
+/*******************************************************************************
+
+    Check the validity of a block.
+
+    Currently only the height of the block is
+    checked against the previous height.
+
+    Params:
+        block = the block to check
+        prev_height = the height of the previous block which this
+                      block should point to
+
+    Returns:
+        true if the block is considered valid
+
+*******************************************************************************/
+
+public bool isValid (const ref Block block, in ulong prev_height)
+    pure nothrow @safe @nogc
+{
+    if (block.header.height != prev_height + 1)
+        return false;
+
+    return true;
+}
+
+///
+unittest
+{
+    import agora.consensus.Genesis;
+    auto gen_block = getGenesisBlock();
+    Block block;
+    block.header.prev_block = gen_block.header.hashFull();
+    block.header.height = gen_block.header.height + 1;
+    assert(block.isValid(gen_block.header.height));
+
+    block.header.height = 100;
+    assert(!block.isValid(gen_block.header.height));
 }
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -18,6 +18,7 @@ import agora.common.crypto.Key;
 import agora.common.Hash;
 import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
+import agora.consensus.Genesis;
 
 /// Delegate to find an unspent UTXO
 public alias UTXOFinder = scope const(Output)* delegate (Hash hash, size_t index)
@@ -223,23 +224,57 @@ unittest
 
     Check the validity of a block.
 
-    Currently only the height of the block is
-    checked against the previous height.
+    A block is considered valid if:
+        - its height is the previous block height + 1
+        - its prev_hash is the previous block header's hash
+        - the number of transactions in the block are equal to Block.TxsInBlock
+        - the merkle root in the header matches the re-built merkle tree root
+          based on the included transactions in the block
+        - all the the transactions pass validation, which implies:
+            - signatures are authentic
+            - the inputs spend an output which must be found with the
+              findUTXO() delegate
+
+    Note that checking for transactions which double-spend is the responsibility
+    of the findUTXO() delegate. During validation, whenever this delegate is
+    called it should also keep track of the used UTXOs, thereby marking
+    it as a spent output. See the `findNonSpent` function in the
+    unittest for an example.
 
     Params:
         block = the block to check
         prev_height = the height of the previous block which this
                       block should point to
+        findUTXO = delegate to find the referenced unspent UTXOs with
 
     Returns:
         true if the block is considered valid
 
 *******************************************************************************/
 
-public bool isValid (const ref Block block, in ulong prev_height)
-    pure nothrow @safe @nogc
+public bool isValid (const ref Block block, in ulong prev_height,
+    in Hash prev_hash, UTXOFinder findUTXO) nothrow @safe
 {
+    import std.algorithm;
+
+    // special case for the genesis block
+    if (block.header.height == 0)
+        return block == getGenesisBlock();
+
     if (block.header.height != prev_height + 1)
+        return false;
+
+    if (block.header.prev_block != prev_hash)
+        return false;
+
+    if (block.txs.length != Block.TxsInBlock)
+        return false;
+
+    if (block.txs.any!(tx => !tx.isValid(findUTXO)))
+        return false;
+
+    Hash[] merkle_tree;
+    if (block.header.merkle_root != Block.buildMerkleTree(block.txs, merkle_tree))
         return false;
 
     return true;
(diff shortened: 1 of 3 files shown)


source/agora/consensus/Validation.d
```diff
@@ -217,3 +219,236 @@ unittest
...
+    if (block.header.prev_block != prev_hash)
+        return false;
+
+    if (block.txs.length != Block.TxsInBlock)
+        return false;
+
+    if (block.txs.any!(tx => !tx.isValid(findOutput)))
+        return false;
+
+    static Hash[] merkle_tree;
```
AndrejMitrovic: I think I need to make this a parameter, to make it at least somewhat re-entrant. What do you think @Geod24?

source/agora/consensus/Validation.d
```diff
@@ -217,3 +219,236 @@ unittest
...
+
+    foreach (idx, output; gen_tx.outputs)
+        utxo_set[hashMulti(gen_tx.hashFull, idx)] = output;
+
+    assert(utxo_set.length != 0);
+    const utxo_set_len = utxo_set.length;
+
+    // contains the used set of UTXOs during validation (to prevent double-spend)
+    Output[Hash] used_set;
+    UtxoFinder findNotDoubleSpent = (Hash hash, size_t index)
```
AndrejMitrovic: I'll rename this function.

source/agora/node/Ledger.d
```diff
@@ -238,6 +233,19 @@ public class Ledger
...
+
+    /***************************************************************************
+
+        Add the genesis block to the ledger.
+
+    ***************************************************************************/
+
+    private void addGenesisBlock ()
+    {
+        auto block = getGenesisBlock();
```
Geod24, line 228: `assert(this.ledger.length == 0);`?

AndrejMitrovic:
Unblocked.

source/agora/consensus/Validation.d
```diff
@@ -217,3 +218,43 @@ unittest
...
+        block = the block to check
+        prev_height = the height of the previous block which this
+                      block should point to
+
+    Returns:
+        true if the block is considered valid
+
+*******************************************************************************/
+
+public bool isValid (const ref Block block, in ulong prev_height)
```
Geod24: More a remark than something that needs addressing, but this will soon need an instance of `Ledger`, right ?

Geod24:
Looks like the commits show in the wrong order in Github ?

AndrejMitrovic:
```
* 6a1df5f Add the genesis block directly to the ledger
* 4f011b7 Move block verification to the Validation module
* 8ce4d72 Make the spend amount configurable for the tests
* 9a18dc1 Implement full block validation
```

Actual order.

AndrejMitrovic:
Hmm.. the "random" segfault seems to happen consistently with this PR.

AndrejMitrovic:
@Geod24 is it possible to extract the binary built by Travis? I want to compare the binaries of a successful and failed run.

Because this morning I could easily reproduce the issue locally, yet when I rebuilt the binary again the problem just went away and I can no longer reproduce it..

AndrejMitrovic:
It's green finally.

AndrejMitrovic:
There was a typo in the commit message so I started a rebuild.

AndrejMitrovic:
~Still fails. Damn what is this bug..~

The bug was in my code.

AndrejMitrovic:
It's green, but I will restart the build a few more times on ldc + osx to verify.

AndrejMitrovic:
Travis is mostly ok. So I'd say this is ready (for review at least).

source/agora/consensus/Validation.d
```diff
@@ -217,3 +219,230 @@ unittest
...
+        return false;
+
+    if (block.txs.length != Block.TxsInBlock)
+        return false;
+
+    if (block.txs.any!(tx => !tx.isValid(findOutput)))
+        return false;
+
+    Hash[] merkle_tree;
+    if (block.header.merkle_root != Block.buildMerkleTree(block.txs, merkle_tree))
```
AndrejMitrovic, line 277: Btw, I was thinking about this.

If the transactions in the block were already sorted lexicographically by their hash, then I could simply merge the hashes of the transactions together to create the merkle root here and compare that to the merkle root in the header. 

Making them sorted would break a bunch of test code though.

Currently the transactions are added to the block in arbitrary order. Is this desired?

AndrejMitrovic:
Why is dockercloud still failing?

Geod24:
> Currently the transactions are added to the block in arbitrary order. Is this desired?

No. But the order might end up being a bit more complicated than lexicographical.

source/agora/consensus/data/Block.d
```diff
@@ -220,7 +220,11 @@ public struct Block
         ref Hash[] merkle_tree) nothrow @safe
     {
         () @trusted { merkle_tree.assumeSafeAppend(); }();
-        merkle_tree.length = (txs.length * 2) - 1;
+
+        // workaround for segfault issue with ldc2 on osx
```
Geod24: Issue plz

source/agora/consensus/Validation.d
```diff
@@ -223,38 +224,225 @@ unittest
...
+    A block is considered valid if:
+        - its height is the previous block height + 1
+        - its prev_hash is the previous block header's hash
+        - the number of transactions in the block are equal to Block.TxsInBlock
+        - the merkle root in the header matches the re-built merkle tree root
+          based on the included transactions in the block
+        - all the the transactions pass validation, which implies:
+            - signatures are authentic
+            - the inputs spend an output which must be found with the
+              findOutput() delegate
```
Geod24: Just say `UTXO` ?

source/agora/consensus/Validation.d
```diff
@@ -223,38 +224,225 @@ unittest
...
 
     Returns:
         true if the block is considered valid
 
 *******************************************************************************/
 
-public bool isValid (const ref Block block, in ulong prev_height)
-    pure nothrow @safe @nogc
+public bool isValid (const ref Block block, in ulong prev_height,
+    in Hash prev_hash, UtxoFinder findOutput) nothrow @safe
```
Geod24: Style note for the future: acronyms like UTXO should be all caps (`HTTP`, `UTXO`, ...)

source/agora/consensus/Validation.d
Geod24: Another thing: We should make this an `InputRange`.

source/agora/consensus/Validation.d
```diff
@@ -223,38 +224,225 @@ unittest
...
+    assert(block.isValid(gen_block.header.height, gen_hash, findOutput));
 
     block.header.height = 100;
-    assert(!block.isValid(gen_block.header.height));
+    assert(!block.isValid(gen_block.header.height, gen_hash, findOutput));
+
+    block.header.height = gen_block.header.height + 1;
+    assert(block.isValid(gen_block.header.height, gen_hash, findOutput));
+
+    /// .pref_block check
```
Geod24: Typo

source/agora/consensus/Validation.d
```diff
@@ -223,38 +224,225 @@ unittest
...
-    assert(block.isValid(gen_block.header.height));
+    assert(gen_block.isValid(gen_block.header.height, Hash.init, null));
+    auto gen_hash = gen_block.header.hashFull();
+
+    auto gen_tx = gen_block.txs[0];
+    tx_map[gen_tx.hashFull()] = [gen_tx];
+    auto txs = makeChainedTransactions(gen_key, null, 1);
+    auto block = makeNewBlock(gen_block, txs);
+
+    // height checks
```
Geod24: `checks` then `check` ? Me confused

source/agora/consensus/Validation.d
```diff
@@ -223,38 +224,225 @@ unittest
...
+    assert(block.isValid(gen_block.header.height, gen_hash, findOutput));
+
+    /// .pref_block check
+    block.header.prev_block = block.header.hashFull();
+    assert(!block.isValid(gen_block.header.height, gen_hash, findOutput));
+
+    block.header.prev_block = gen_hash;
+    assert(block.isValid(gen_block.header.height, gen_hash, findOutput));
+
+    /// .txs length check
```
Geod24, line 329: Since you are not rebuilding the merkle root it would not succeed anyway, would it ?

source/agora/consensus/Validation.d
```diff
@@ -223,38 +224,225 @@ unittest
...
+
+    txs = makeChainedTransactions(gen_key, prev_txs, 1);
+    block = makeNewBlock(gen_block, txs);
+    assert(block.isValid(gen_block.header.height, gen_block.header.hashFull(),
+        findOutput));
+
+    // modify the last hex byte of the merkle root
+    block.header.merkle_root = Hash("0x18cb29d05b784548fe385401df4e0a920d254" ~
+        "dd3c59ab8942888e40211acf05581ad8547a2c7d37d532a1465cdba0acaa2492f49" ~
+        "f30331cabebe2464c89e1554");
```
Geod24: You can store `Hash` and change it (it has `opSlice`) to make the diff simpler to spot

source/agora/consensus/Validation.d
AndrejMitrovic: Yeah I just tried to visually separate it from the rest. Makes sense to capitalize though.

source/agora/consensus/Validation.d
AndrejMitrovic: Ah, I thought it was immutable.. cool.

source/agora/consensus/Validation.d
AndrejMitrovic: No it won't.

AndrejMitrovic: https://github.com/bpfkorea/agora/blob/a2962faa71b9d9f6f8af9bb45342ad4da6076f97/source/agora/consensus/data/UtxoSet.d#L111

And then:

https://github.com/bpfkorea/agora/blob/a2962faa71b9d9f6f8af9bb45342ad4da6076f97/source/agora/consensus/data/UtxoSet.d#L135

A simple approach for making sure there are no double-spends.

source/agora/consensus/Validation.d
AndrejMitrovic: Well, the length check comes before the expensive merkle root check, so I've kept it in. And I can't rebuild the merkle root because it also contains the .length check. Once we remove the transaction length limit, we can remove this check too.

AndrejMitrovic:
> Another thing: We should make this an InputRange.

Hmm, not sure how this can be done.

AndrejMitrovic:
Fixed most issues. I still don't like having to rebuild the merkle tree though. I mean just the array allocation. If the txs were sorted when inserted in the block, then a merkle root could easily be re-created during validation without having to build the array.

Geod24 (approved):

AndrejMitrovic:
> Geod24 approved these changes 3 hours ago, 5:49 AM

Oh my.

PR #195, opened 2019-08-18
Just DB related code cleanup

denizzzka:
Not changes behaviour of Agora

### DIFF
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -43,9 +43,8 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public static void initialize ()
+    shared static this()
     {
-        .shutdown();
         .config(SQLITE_CONFIG_MULTITHREAD);
         version (unittest) {} else
             .config(SQLITE_CONFIG_LOG, &loggerCallback, null);
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -70,7 +70,6 @@ public class TransactionPool
         if (db_exists)
             logInfo("Loading database from: %s", db_path);
 
-        // note: can fail. we may want to just recover txes from the network instead.
         this.db = Database(db_path);
 
         if (db_exists)
@@ -105,16 +104,7 @@ public class TransactionPool
         }
     }
 
-    /***************************************************************************
-
-        Shut down the database
-
-        Note: this method must be called explicitly, and not inside of
-        a destructor.
-
-    ***************************************************************************/
-
-    public void shutdown ()
+    ~this ()
     {
         this.db.close();
     }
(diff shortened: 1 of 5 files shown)


Geod24:
Regarding the second commit: `shared static this` are nasty, and from my experience lead to issues in libraries (e.g. we have initialization code just to deal with Vibe.d's `shared static this` messing with global variables).

denizzzka:
> Regarding the second commit: shared static this are nasty,

This is beautiful part of a great modules subsystem of D language

And at other hand `.shutdown()` which used end-to-end inside of all codebase is an implementation of hidden state of any class where it is used.

source/agora/common/TransactionPool.d
```diff
@@ -70,7 +69,6 @@ public class TransactionPool
         if (db_exists)
             logInfo("Loading database from: %s", db_path);
 
-        // note: can fail. we may want to just recover txes from the network instead.
         this.db = Database(db_path);
```
AndrejMitrovic, line 72: Please replace this note with a github issue, since the error isn't handled yet.

source/agora/common/TransactionPool.d
```diff
@@ -43,9 +43,8 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public static void initialize ()
+    shared static this()
     {
-        .shutdown();
```
AndrejMitrovic, line 48: There is something which causes logging configuration to fail if shutdown is not called first. I'm not sure what it is yet. I'll add an integration test later to confirm it.

source/agora/common/TransactionPool.d
```diff
@@ -105,16 +103,7 @@ public class TransactionPool
...
-
-        Shut down the database
-
-        Note: this method must be called explicitly, and not inside of
-        a destructor.
-
-    ***************************************************************************/
-
-    public void shutdown ()
+    ~this ()
```
AndrejMitrovic, line 106: Missing documentation.

denizzzka: How do you imagine a good documentation for this destructor?

source/agora/common/TransactionPool.d
denizzzka: Done: https://github.com/bpfkorea/agora/issues/196

source/agora/common/TransactionPool.d
denizzzka: <s>First by compared to what? Nothing in calling order was changed.</s>

Shutdown currently contains only `this.db.close();`

source/agora/common/TransactionPool.d
AndrejMitrovic: Maybe just put it in the class header. "Must call GC.destroy on an instance of this class manually"

denizzzka: Not agreed. Then someone will forget to remove it after this behaviour will be fixed.

AndrejMitrovic: Then file an issue on github so we can track this and make a comment there that the comment which you will add should be removed.

AndrejMitrovic:
Shutting down, dumping data to disk, etc, should all be done before any destructors are called. So I disagree that calling `destroy(object)` is better than `object.shutdown`. After all, the destructor itself ends up calling `db.shutdown`.

As for the initialization part, we'll improve on this by moving it to another module, since the UTXO will use it as well.

PR #216, opened 2019-08-21
Simplify the findHashIndex() function

AndrejMitrovic:


source/agora/consensus/data/Block.d
```diff
@@ -334,25 +335,11 @@ public struct Block
 
         const size_t size = min(this.txs.length, this.merkle_tree.length);
```
AndrejMitrovic: I am not sure why this constant is here though. The merkle tree's length should be `(tx.length * 2) - 1`, so I think this should be moved to the assert above as a check and this `min` should be removed.

AndrejMitrovic:
Added a commit removing the `size` constant.

Geod24 (dismissed):
The Merkle Tree is not sorted. The transactions are.

AndrejMitrovic:
Well in the current implementation the merkle tree array is sorted from [0 .. txs.length], whereas the actual transaction array is not sorted at all.

AndrejMitrovic:
Yeah, I've fixed the commit message now.

Geod24 (approved):

PR #217, opened 2019-08-21
Don't send a transaction if a node already has it

### LINKED ISSUE #210: Network manager sends a transaction even if a node already has the transaction
The node implements a `hasTransactionHash` API, but the network manager gossips the transaction to a node even if that node already contains the transaction. 

The node should instead query the target node's `hasTransactionHash()`, and only send the transaction if `hasTransactionHash()` returns false.

AndrejMitrovic:
Logging confirms it works. Devising a specialized test-case just for this functionality is likely not possible.

Fixes #210

### DIFF
--- a/source/agora/network/NetworkClient.d
+++ b/source/agora/network/NetworkClient.d
@@ -143,8 +143,16 @@ class NetworkClient
 
     public void sendTransaction (Transaction tx) @trusted
     {
+        import agora.common.Hash;
+        const tx_hash = tx.hashFull();
+
         this.taskman.runTask(
         {
+            // if the node already has this tx, don't send it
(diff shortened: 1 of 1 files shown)


Geod24:
Interestingly, all Mac builds failed

AndrejMitrovic:
Rebased.

Geod24 (approved):
Random idea: We cache this using bloom filters in the future.

AndrejMitrovic:
> Random idea: We cache this using bloom filters in the future.

Can you elaborate?

Geod24:
Originally I was thinking we would end up calling a node more than once to know if they have a transactions, but realized we would just reject a tx we already have, so it's not a problem.

PR #218, opened 2019-08-21
Use a checksum in BlockStorage for data integrity validation

TrustHenry:
If there is damage to the block file, calculate the checksum in a way that can verify it and save it to the beginning of the file.
We can check the damage by reading the checksum and comparing it.

Requires #162

### DIFF
--- a/source/agora/node/BlockStorage.d
+++ b/source/agora/node/BlockStorage.d
@@ -26,6 +26,7 @@ import agora.consensus.Genesis;
 
 import std.algorithm;
 import std.container.rbtree;
+import std.digest.crc;
 import std.exception;
 import std.file;
 import std.format;
@@ -109,6 +110,12 @@ public interface IBlockStorage
 /// The map file size
 private immutable size_t MapSize = 640 * 1024;
 
+/// The block data size
+private immutable size_t DataSize = MapSize - ChecksumSize;
+
+/// The CRC32 checksum size
+private immutable size_t ChecksumSize = 4;
+
 private struct HeightPosition
 {
     size_t              height;
@@ -156,12 +163,16 @@ public class BlockStorage : IBlockStorage
     /// Base Position of current file
     private size_t file_base;
 
+    /// Saving current block
+    private bool is_saving;
+
     /// Ctor
     public this (string path)
     {
         this.path = path;
         this.file_index = ulong.max;
         this.length = ulong.max;
+        this.is_saving = false;
 
         this.height_idx = new IndexHeight();
         this.hash_idx = new IndexHash();
@@ -235,13 +246,19 @@ public class BlockStorage : IBlockStorage
     private bool map (size_t findex) @trusted nothrow
     {
         try {
-            if ((this.file !is null) && (findex == this.file_index))
-                return true;
-
             if (this.file !is null)
+            {
+                if (findex == this.file_index)
+                    return true;
+
+                if (this.is_saving && !this.writeChecksum())
+                    assert(0);
+
                 this.release();
+            }
 
             this.file_index = findex;
+            bool file_exist = std.file.exists(this.getFileName(this.file_index));
 
             this.file =
                 new MmFile(
@@ -251,7 +268,10 @@ public class BlockStorage : IBlockStorage
                     null
                 );
 
-            this.file_base = MapSize * this.file_index;
+            this.file_base = DataSize * this.file_index;
+            if (file_exist)
+                this.validateChecksum();
+
             return true;
         }
         catch (Exception ex)
@@ -321,6 +341,8 @@ public class BlockStorage : IBlockStorage
             block_position = this.length;
             data_position = block_position + size_t.sizeof;
 
+            this.is_saving = true;
+            scope(exit) this.is_saving = false;
             size_t block_size = 0;
             scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
             {
@@ -337,6 +359,9 @@ public class BlockStorage : IBlockStorage
                 return false;
 
             this.length += size_t.sizeof + block_size;
+
+            if (!this.writeChecksum())
+                assert(0);
 
             // add to index of heigth
             this.height_idx.insert(
@@ -517,13 +542,13 @@ public class BlockStorage : IBlockStorage
     {
         if (
             (this.file !is null) &&
-            (from / MapSize == this.file_index) &&
-            (to / MapSize == this.file_index))
+            (from / DataSize == this.file_index) &&
+            (to / DataSize == this.file_index))
         {
             try
             {
-                const size_t x0 = from - this.file_base;
-                const size_t x1 = to - this.file_base;
+                const size_t x0 = from - this.file_base + ChecksumSize;
+                const size_t x1 = to - this.file_base + ChecksumSize;
                 data = cast(ubyte[])this.file[x0 .. x1];
                 return true;
             }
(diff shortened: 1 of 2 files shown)


source/agora/common/BlockStorage.d
```diff
@@ -155,12 +163,16 @@ public class BlockStorage : IBlockStorage
     /// Base Position of current file
     private size_t file_base;
 
+    /// Saving current block
+    private bool is_savaing;
```
AndrejMitrovic: `is_saving`

source/agora/common/BlockStorage.d
```diff
@@ -761,6 +788,80 @@ public class BlockStorage : IBlockStorage
...
+            data = file data bytes
+
+        Returns:
+            calculate the checksum
+
+    *******************************************************************************/
+
+    private ubyte[] checksum (const ubyte[] data) @safe nothrow
+    {
+        auto crc32 = new CRC32Digest();
```
AndrejMitrovic: `auto` => `scope` to allocate on the stack instead of the heap.

source/agora/common/BlockStorage.d
```diff
@@ -761,6 +788,80 @@ public class BlockStorage : IBlockStorage
...
+            const actual = this.checksum(fileData);
+            if (expected != actual)
+            {
+                stderr.writefln("[ERROR] %s Block file is corrupted and will be terminated." , this.getFileName(this.file_index));
+                abort();
+            }
+        }
+        catch (Exception ex)
+        {
+            this.writeLog("BlockStorage.compareChecksum: ", ex);
```
AndrejMitrovic: If the checksum is read, but it doesn't match, we abort. But if the checksum fails to be read, we just ignore the error. We should be consistent and abort in all cases.

source/agora/common/BlockStorage.d
```diff
@@ -761,6 +788,80 @@ public class BlockStorage : IBlockStorage
...
+
+        Read the file data, calculate checksum,
+        and write checksum to file at start point
+
+        Returns:
+            Returns true if success, otherwise returns false.
+
+    *******************************************************************************/
+
+    private bool writeChecksum () @trusted nothrow
```
AndrejMitrovic: Why is the return value of this function ignored?

source/agora/common/BlockStorage.d
TrustHenry: Fixed rename

source/agora/common/BlockStorage.d
TrustHenry: add assert(writeChecksum());

source/agora/common/BlockStorage.d
TrustHenry: add abort();

source/agora/common/BlockStorage.d
TrustHenry: Fixed

source/agora/common/BlockStorage.d
```diff
@@ -120,18 +230,27 @@ public class BlockStorage
...
-            if ((this.file !is null) && (findex == this.file_index))
-                return true;
-
             if (this.file !is null)
+            {
+                if (findex == this.file_index)
+                    return true;
+
+                if (this.is_saving)
+                    assert(this.writeChecksum());
```
AndrejMitrovic: Like Mathias mentioned, this assert will disappear in `-release` mode. It needs to be replaced with:

```
if (this.is_saving && !this.writeChecksum())
    assert(0);
```

`assert(0);` is special. It is similar to `abort();`.

source/agora/common/BlockStorage.d
```diff
@@ -226,6 +352,8 @@ public class BlockStorage
 
             this.length += size_t.sizeof + block_size;
 
+            assert(this.writeChecksum());
```
AndrejMitrovic: Same comment about assert.

AndrejMitrovic:
Why is the checksum written twice? I see two calls to `writeChecksum`.

TrustHenry:
If it save block data to a file during the block storage function and then move on to the next file, record the checksum for the next file.

source/agora/common/BlockStorage.d
TrustHenry: Applied the suggestion

source/agora/common/BlockStorage.d
TrustHenry: Applied the suggestion

Geod24:
This needs a rebase

TrustHenry:
rebase done

tests/unit/BlockStorageMultiTx.d
```diff
@@ -72,6 +74,42 @@ private void main ()
...
+    assert(result == expected);
+
+    // Noise Test
+    File block_file;
+    string file_name = buildPath(path, storage.getFileName(0));
+    block_file = File(file_name, "r+b");
+    block_file.seek(5, SEEK_SET);
+    ubyte[1] noise;
+    foreach (i;0 .. noise.length)
+        noise[i] = cast(ubyte)(0);
```
AndrejMitrovic: What was the intention here? Because this doesn't do anything. `ubyte[1]` is already a fixed-length array with all values set to zero.

source/agora/common/BlockStorage.d
Geod24: Use `assert(0)` instead of `abort()`

TrustHenry: Applied the suggestion

tests/unit/BlockStorageMultiTx.d
TrustHenry: Yes, It's 0 So, I will changed the value to 9

tests/unit/BlockStorageMultiTx.d
```diff
@@ -72,6 +74,42 @@ private void main ()
...
+    immutable ubyte[4] expected = cast(ubyte[4]) [0xfd, 0xa0, 0xb6, 0xa0];
+    const ubyte[] result = storage.checksum(data);
+    assert(result == expected);
+
+    // Noise Test
+    File block_file;
+    string file_name = buildPath(path, storage.getFileName(0));
+    block_file = File(file_name, "r+b");
+    block_file.seek(5, SEEK_SET);
+    ubyte[1] noise;
```
AndrejMitrovic:
```suggestion
    ubyte[1] noise = [9];
```

Simpler! :)

TrustHenry: Yes, simple

AndrejMitrovic:
Here are the changes I think would be good to make here:

- Remove the unittest commit, and replace it with a new integration test file: https://gist.github.com/AndrejMitrovic/4b02269d6b545bc564d1586b3f35ec61
- Make `checksum` private. It should not be used by any other module except `BlockStorage.d`, the tests should not call this function.
- Change the `checksum` code a bit so it doesn't allocate:

```diff
diff --git a/source/agora/node/BlockStorage.d b/source/agora/node/BlockStorage.d
index a5fd2a61..f7db747d 100644
--- a/source/agora/node/BlockStorage.d
+++ b/source/agora/node/BlockStorage.d
@@ -814,7 +814,8 @@ public class BlockStorage : IBlockStorage
     {
         scope crc32 = new CRC32Digest();
         crc32.put(data);
-        return crc32.finish();
+        static ubyte[4] buffer;
+        return () @trusted { return crc32.finish(buffer); }();
     }
```

Later we will remove the `nothrow` requirement from our serialization routines, which means we will be able to add a proper integration test (notice that this part of the code is still commented out: `// storage.readBlock(block, 0);  // will halt due to checksum failure`).

We also need to merge https://github.com/bpfkorea/agora/pull/254.

TrustHenry:
Applied the suggestion

MichaelKim20 (approved):

AndrejMitrovic:
I did some minor changes and squashed the two commits: https://github.com/bpfkorea/agora/compare/f2461a43e1f30f43b01d8f86024cc438f0dae7a9..79e8bb9c85c1c6e36b6e1bcf8bc032d4a241c241

I will merge this once it's green.

AndrejMitrovic (approved):

PR #223, opened 2019-08-21
Implement IBlockStorage interface (Split from #162)

Geod24:
Since we are close to the end of the sprint, and the code is good to go, I decided to extract it in another PR and fix the details that were caught during review.


AndrejMitrovic (approved):

AndrejMitrovic:
Looks good, thanks Michael for the work.

PR #224, opened 2019-08-22
Make Genesis block & transaction manifest constants

Geod24:
So there is only one single instance in the program,
and we refer to it.

### DIFF
--- a/source/agora/consensus/Genesis.d
+++ b/source/agora/consensus/Genesis.d
@@ -20,56 +20,50 @@ import agora.common.Hash;
 import agora.common.crypto.Key;
 import agora.consensus.data.Transaction;
 
-/*******************************************************************************
 
-    Creates the genesis block.
-    The output address is currently hardcoded to a randomly generated value,
-    it will be replaced later with the proper address.
-
-    Returns:
-        the genesis block
-
-*******************************************************************************/
-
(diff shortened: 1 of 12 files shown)


source/agora/consensus/Genesis.d
```diff
@@ -20,56 +20,50 @@ import agora.common.Hash;
...
-        ).array;
-
-    return Transaction(
-        [Input(Hash.init, 0)],
-        outputs
-    );
-}
+    inputs: [ Input(Hash.init, 0) ],
+    outputs: [
+        Output(Amount(40_000_000 / Block.TxsInBlock), GenesisOutputAddress),
```
AndrejMitrovic, line 57: Hmm any way to generalize this so it still works if the `TxsInBlock` constant is changed? Because this manually makes 8 transactions here.

AndrejMitrovic (approved):

source/agora/consensus/Genesis.d
Geod24: I was thinking about that, and originally was going for `std.algorithm`, but we'll remove the hard limit on txs / blocks in the future, so would it be a generalization ?

AndrejMitrovic: No it's not a big issue.

PR #226, opened 2019-08-22
Implement additional serialization routines

AndrejMitrovic:
Split off from https://github.com/bpfkorea/agora/pull/219


Geod24 (approved):
LGTM
The general idea was to have 1 method per basic-type size, so signed and unsigned would be deserialized by the same routine, but I see how that's a problem now.

Geod24:
Please rebase

PR #230, opened 2019-08-22
Move Travis setup / execution to a shell script

Geod24:
One issue with Travis is that if a test fails early in the pipeline,
it does not immediately exit, instead continuing the other tests.
By setting '-e' we enable this behavior.


AndrejMitrovic (approved):
Somehow all these declarative configs & languages end up being replaced with ol' reliable shell scripts.

PR #235, opened 2019-08-22
Fix BlockStorage to properly work from an empty disk

Geod24:
This is the issue that triggered during the sprint demo.

### DIFF
--- a/source/agora/common/BlockStorage.d
+++ b/source/agora/common/BlockStorage.d
@@ -22,6 +22,7 @@ import agora.common.Deserializer;
 import agora.common.Hash;
 import agora.common.Serializer;
 import agora.consensus.data.Block;
+import agora.consensus.Genesis;
 
 import std.algorithm;
 import std.container.rbtree;
@@ -169,6 +170,10 @@ public class BlockStorage : IBlockStorage
             mkdirRecurse(this.path);
 
         this.loadAllIndexes();
+
+        // Add Genesis if the storage is empty
+        if (this.height_idx.length == 0)
+            this.saveBlock(GenesisBlock);
     }
 
     /***************************************************************************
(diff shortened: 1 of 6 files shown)


source/agora/common/BlockStorage.d
```diff
@@ -817,6 +822,7 @@ public class MemBlockStorage : IBlockStorage
     {
         this.height_idx = new IndexHeight();
         this.hash_idx = new IndexHash();
+        this.saveBlock(GenesisBlock);
```
AndrejMitrovic, line 825: Hmm.. I need to add the genesis block's UTXOs to the UTXO set. But I'll work on this in my PR after we rebase yours.

AndrejMitrovic: What I mean is, the Ledger class will manager the UTXO set, not the BlockStorage.

source/agora/node/BlockStorage.d
```diff
@@ -14,7 +14,7 @@
 
 *******************************************************************************/
 
-module agora.common.BlockStorage;
+module agora.node.BlockStorage;
```
AndrejMitrovic, line 17: Could also move it to a `storage` package in the future.

AndrejMitrovic (approved):

source/agora/common/BlockStorage.d
Geod24: Well, since you can assume the BlockStorage always have at least one block, can't you just load it from there ?

source/agora/node/BlockStorage.d
Geod24: Yep, I was considering `agora.node.network` and `agora.node.storage` as well. But later.

PR #273, opened 2019-08-26
Limit CRC32 checksum data length to 1MB

AndrejMitrovic:
The CRC algorithm's effeciveness is based on the length of its data input, as well as the size of
the CRC (e.g. 8-bit, 16-bit, etc).

With CRC32, a 1MB limit is reasonable.

### DIFF
--- a/source/agora/node/BlockStorage.d
+++ b/source/agora/node/BlockStorage.d
@@ -811,7 +811,16 @@ public class BlockStorage : IBlockStorage
     *******************************************************************************/
 
     private static ubyte[] makeChecksum (const ubyte[] data) @safe nothrow
+    out(result)
     {
+        assert(result.length + DataSize <= MapSize,
+            "Checksum size is too large to fit in the map");
+    }
+    body
+    {
+        assert(data.length < 1 << 20,
+            "Data length for checksum should not exceed 1MB");
+
         scope crc32 = new CRC32Digest();
         crc32.put(data);
         static ubyte[4] buffer;


AndrejMitrovic:
@TrustHenry can you Review & Approve this PR?

AndrejMitrovic:
Ping @TrustHenry

source/agora/node/BlockStorage.d
```diff
@@ -812,6 +812,9 @@ public class BlockStorage : IBlockStorage
 
     private static ubyte[] makeChecksum (const ubyte[] data) @safe nothrow
     {
+        assert(data.length < 1 << 20,
```
TrustHenry, line 821: ` assert((MapSize <= 1 << 20) && ((data.length + ChecksumSize) <= 1 << 20),`
We can make file up to 1MB.
If the map size is changed, I think we can catch the error.

AndrejMitrovic: This isn't really testing for the size of the map, it's testing the size of the data that is being checksummed. That's why this function is `static`, it doesn't care about what the "map" size is. The reason for this limit is because CRC32 has a limited efficiency based on the size of its input data.

There is more information here: https://stackoverflow.com/questions/2321676/data-length-vs-crc-length

TrustHenry: As you say, mapsize doesn't seem right to do check here.
However, I wish the data size + checksum size would be 1MB.
I think it will be good to expand to 1MB when changing the mapsize in the future. 
It is convenient to change the mapsize to `size_t MapSize = 1024 * 1024;`

AndrejMitrovic: I see, in that case I'll add a check for that.

AndrejMitrovic:
Updated with an assert check for the checksum size.

TrustHenry (approved):

AndrejMitrovic:
고맙습니다 for the good review!

PR #194, opened 2019-08-16
Implement UTXO set data structure

AndrejMitrovic:
Things to consider:

~- The storage is split in two, the hot cache (hashmap) and the cold storage (SQLite). Moving from the hot cache to the cold storage is done by maintaining an array acting as a queue. But this is very basic, and we could go with a linked-list instead.~
~- On shutdown, I just save everything into the cold store (the SQLite database). But ideally we should be able to restore the hot items (might not be super-important on node startup, we'll see..).~

Edit: I've removed the "hot cache", it was premature optimization. And it actually hid a bug in SQLite indexing.

- If the UTXO database couldn't be loaded from disk, we should attempt to either retrieve it from other nodes (introduce getUtxoSet() entry-point), or rebuild it from the blockchain data we have (much more costly though).

PR #274, opened 2019-08-27
Introduce opApply() and remove() for the TransactionPool

AndrejMitrovic:
I considered making it an input range, but then there are issues with `@system` destructors being called from within wherever the input range is being used at the call site. Maybe I can make this work better in the future, but for no `opApply` works well.

Required for https://github.com/bpfkorea/agora/pull/194

### DIFF
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -167,46 +167,79 @@ public class TransactionPool
 
     /***************************************************************************
 
+        Walk over the transactions in the pool and call the provided delegate
+        with each hash and transaction
+
         Params:
-            amount = how many transactions to return
+            dg = the delegate to call
 
         Returns:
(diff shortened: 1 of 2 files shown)


source/agora/common/TransactionPool.d
```diff
@@ -221,7 +225,15 @@ unittest
     txs.each!(tx => pool.add(tx));
     assert(pool.length == txs.length);
 
-    auto pool_txs = pool.take(txs.length);
```
Geod24: Rule of three: provide a `version(unittest)` wrapper ?

AndrejMitrovic: Well now that we have code coverage, I'm on board with that idea.

AndrejMitrovic:
Reintroduced `take` as a unittest function.

Geod24 (approved):

PR #275, opened 2019-08-27
Add the genesis block from within the Ledger

AndrejMitrovic:
It's not correct to add it within the BlockStorage, because then we lose track whether the block was just added, or if it was loaded from disk. The difference is important, because the Ledger keeps track of UTXOs based on newly-added blocks to the ledger (and the serialized UTXO on disk).

Required for https://github.com/bpfkorea/agora/pull/194

### DIFF
--- a/source/agora/node/BlockStorage.d
+++ b/source/agora/node/BlockStorage.d
@@ -46,6 +46,15 @@ public interface IBlockStorage
 
     /***************************************************************************
 
+        Returns:
+            true if the storage is empty (height index was not loaded)
+
+    ***************************************************************************/
+
+    public bool isEmpty ();
+
+    /***************************************************************************
+
         Read the last block from the storage.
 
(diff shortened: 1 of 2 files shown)


TrustHenry (approved):
It is right to make Genesis blocks from the ledger.

AndrejMitrovic:
Thank you!

Geod24:
Not a fan of this to be honest. The `isEmpty` method seems to be single-purpose. After initialization, it will always return `true`, which is a red flag.
The two options I can think of is:
- Hard-error if the storage was not previously initialized and force the user to call an init method. This is less user friendly, interacts badly with tests, but is more configurable.
- Provide a delegate that is called on first initialization.

AndrejMitrovic:
Yes I agree. Let's go with the second option. Maybe we can come up with something better later.

PR #279, opened 2019-08-28
[POC] Testing IP banning support

AndrejMitrovic:
@Geod24 I had this in my branch I was experimenting with. I wonder if this is the right approach, or if we would have to implement this in a deeper lower level in vibe.d as this only filters HTTP requests but not necessarily connections.

In any case, I don't know why vibe.d's URLRouter is marked `final`.

The only interesting code is in `module agora.network.FilteredURLRouter;`

### DIFF
--- a/source/agora/network/FilteredURLRouter.d
+++ b/source/agora/network/FilteredURLRouter.d
@@ -1,0 +1,82 @@
+/*******************************************************************************
+
+    Contains a URL router which can filter incoming requests.
+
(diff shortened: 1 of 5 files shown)


AndrejMitrovic:
Ah yeah, actually this is too high-level. We need to ban on the connection level.

AndrejMitrovic:
Closing this one down as it's very stale and was just a test.

PR #285, opened 2019-08-30
Fix some 'Params' documentation In API (hasTransactionHash)

TrustHenry:
Corrected incorrect comments.

### DIFF
--- a/source/agora/node/API.d
+++ b/source/agora/node/API.d
@@ -115,7 +115,7 @@ public interface API
             Return true if the node has this transaction hash.
 
         API:
-            GET /hasTransactionHash
+            GET /has_transaction_hash
 
     ***************************************************************************/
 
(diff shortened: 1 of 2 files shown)


source/agora/node/API.d
```diff
@@ -115,7 +115,7 @@ public interface API
             Return true if the node has this transaction hash.
 
         API:
-            GET /hasTransactionHash
+            GET /has_transaction_hash?tx=
```
Geod24: Other places don't include the query parameters. If we want to add them, it should be done for all of them (in a separate commit).

TrustHenry: I Fixed /has_transaction_hash

MichaelKim20 (approved):

Geod24 (approved):

PR #289, opened 2019-09-02
Improve debugging experience by allowing "human readable" output

Geod24:
As I am working on re-adding the integration test once and for all (and making sure it tells us why it fails next time), I found it very hard to debug Agora. This adds a few QoL improvements which should make debugging much easier in the future. Some of them might prove controversial, so feel free to go ahead and shoot.

This also includes a few improvements for the CI. If this PR lingers for too long, or if it's too much to review, I can split them out.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -23,6 +23,7 @@ import agora.consensus.Genesis;
 import agora.consensus.Validation;
 import agora.node.API;
 import agora.node.BlockStorage;
+import agora.utils.PrettyPrinter;
 
 import vibe.core.log;
 
@@ -75,10 +76,10 @@ public class Ledger
 
     public bool acceptBlock (const ref Block block) nothrow @safe
     {
-        if (!block.isValid(this.last_block.header.height,
+        if (auto reason = block.isInvalidReason(this.last_block.header.height,
             this.last_block.header.hashFull, &this.findUTXO))
         {
-            logDebug("Rejected block. %s", block);
+            logDebug("Rejected block: %s: %s", reason, prettify(block));
             return false;
         }
 
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -25,6 +25,7 @@ import agora.network.NetworkManager;
 import agora.node.API;
 import agora.node.BlockStorage;
 import agora.node.Ledger;
+import agora.utils.PrettyPrinter;
 
 import agora.node.GossipProtocol;
 
@@ -141,6 +142,7 @@ public class Node : API
 
     public override void putTransaction (Transaction tx) @safe
     {
+        logDebug("Receveid Transaction: %s", prettify(tx));
         this.gossip.receiveTransaction(tx);
     }
 
--- a/source/agora/utils/PrettyPrinter.d
+++ b/source/agora/utils/PrettyPrinter.d
@@ -1,0 +1,317 @@
+/*******************************************************************************
+
+    Defines method to format types to their human readable representation
+
+    This approach allows to take the string formatting out of the type,
+    simplifying the implementation and reducing dependencies.
+    It is assumed that a human-readable representation will not rely on
+    non-visible (`private`, `package`, `protected`) data,
+    or data that cannot be accessed without mutation.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.utils.PrettyPrinter;
+
+import agora.common.Amount;
+import agora.common.Data;
+import agora.common.crypto.Key;
+import agora.consensus.data.Block;
+import agora.consensus.data.Transaction;
+
+import std.algorithm;
+import std.format;
+import std.range;
+
+/// Ditto
+public auto prettify (T) (const ref T input)
+{
+    static if (is(T : const Amount))
+        return AmountFmt(input);
+    else static if (is(T : const Amount))
+        return HashFmt(input);
+    else static if (is(T : const Input))
+        return InputFmt(input);
+    else static if (is(T : const Output))
+        return OutputFmt(input);
+    else static if (is(T : const Transaction))
+        return TransactionFmt(input);
+    else static if (is(T : const Block))
+        return BlockFmt(input);
+    else
+        return input;
+}
+
+/// Formatting struct for `Amount`
+private struct AmountFmt
+{
+    private Amount value;
+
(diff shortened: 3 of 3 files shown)


Geod24:
Side note: Currently the logging is very inefficient. It should take a pointer, but it looks like there's some memory corruption happening with `std.algorithm : map`. I choose to go the easy way and copy everything for the time being.

source/agora/utils/HumanReadable.d
```diff
@@ -0,0 +1,284 @@
+/*******************************************************************************
```
AndrejMitrovic: `reduce` => `improves` in the commit message imo. Unless you also means it makes it more compact.

source/agora/utils/HumanReadable.d
```diff
@@ -0,0 +1,284 @@
...
+
+    Defines method to format types to their human readable representation
+
+    This approach allows to take the string formatting out of the type,
+    simplifying the implementation and reducing dependencies.
+    It is assumed that a human-readable representation will not rely on
+    non-visible data, or data that cannot be accessed without mutation.
```
AndrejMitrovic: Could you elaborate what "non-visible data" means?

source/agora/utils/HumanReadable.d
```diff
@@ -0,0 +1,284 @@
...
+import agora.common.crypto.Key;
+import agora.consensus.data.Block;
+import agora.consensus.data.Transaction;
+
+import std.algorithm;
+import std.format;
+import std.range;
+
+/// Ditto
+public auto humanReadable (T) (const ref T input)
```
AndrejMitrovic: In many libraries this is called "pretty" printing. So something like `toPretty` is short and concise. But you can keep `humanReadable` too if you want. I'm just thinking of making this function shorter to write.

source/agora/utils/HumanReadable.d
```diff
@@ -0,0 +1,284 @@
...
+        this.value = r;
+    }
+
+    public void toString (scope void delegate(scope const(char)[]) @safe sink) @safe
+    {
+        // Only format `0xABCD..EFGH`
+        enum StartUntil = 6;
+        enum EndFrom    = Hash.StringBufferSize - 4;
+        size_t count;
+        scope void delegate(scope const(char)[]) @safe wrapper = (scope data) @safe {
```
AndrejMitrovic: ~Egyptian style braces?~

AndrejMitrovic: Actually I see it in the style guide. Very well, carry on.

source/agora/utils/HumanReadable.d
```diff
@@ -0,0 +1,284 @@
...
+        this.value.toString(wrapper);
+    }
+}
+
+@safe unittest
+{
+    static immutable Hash SomeHash =
+        "0x000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f"
+        ~ "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f";
+    assert(format("%s", HashFmt(SomeHash)) == "0x0000...e26f");
```
AndrejMitrovic: Do you think we could make it all-caps? I'm thinking it might improve the readability.

AndrejMitrovic: Ah, actually you are using all-caps for public keys. It makes sense to keep this lowercase.

source/agora/utils/HumanReadable.d
```diff
@@ -0,0 +1,284 @@
...
+    {
+        formattedWrite(sink, "%s(%s)",
+            AmountFmt(this.value.value), PubKeyFmt(this.value.address));
+    }
+}
+
+@safe unittest
+{
+    Output output;
+    assert(format("%s", OutputFmt(output)) == "0.0000,000(GAAA...AWHF)");
```
AndrejMitrovic: Hmm.. I would prefer Address(Value), ordering by importance from left to right. What do you think?

source/agora/utils/HumanReadable.d
```diff
@@ -0,0 +1,284 @@
...
+        formattedWrite(sink, "%s, Transactions: %d\n",
+            BlockHeaderFmt(this.value.header), this.value.txs.length);
+        formattedWrite(sink, "%(%s\n%)", this.value.txs.map!(v => TransactionFmt(v)));
+    }
+}
+
+@safe unittest
+{
+    import agora.consensus.Genesis;
+    static immutable GenesisStr = `Height: 0, Prev: 0x0000...0000, Root: 0x893a...d115, Transactions: 1
```
AndrejMitrovic: Could you add a test with multiple transactions? I want to see how it formats.

source/agora/utils/HumanReadable.d
Geod24: SGTM

source/agora/consensus/Validation.d
```diff
@@ -33,23 +33,24 @@ public alias UTXOFinder = scope const(Output)* delegate (Hash hash, size_t index
...
 
     Return:
-        Return true if this transaction is verified.
+        `null` if the transaction is valid, a string explaining the reason it
+        is invalid otherwise.
 
 *******************************************************************************/
 
-public bool isValid (const Transaction tx, UTXOFinder findUTXO)
+public string invalidReason (const Transaction tx, UTXOFinder findUTXO)
```
AndrejMitrovic: I dislike the name of this function. It seems to imply (on first glance) that the transaction/block is already invalid. And it's also not a verb, and hard to memorize. `validate` or `checkValid` or something like that might be better, but I can't suggest a really good name for this..

AndrejMitrovic:
The rest LGTM.

source/agora/utils/HumanReadable.d
Geod24: `private` data. Because nowadays D is using "visibility" rather than "protection".

Geod24: Well I'll add it to the docs

source/agora/utils/HumanReadable.d
Geod24: Yes, the point is not so much to have something "sensible", but something that doesn't take a whole screen.

Geod24:
Fixed all comments

AndrejMitrovic:
Looks good. I wonder if it would be useful to also output a Transaction's Hash in the output for debugging purposes?

Geod24:
> Looks good. I wonder if it would be useful to also output a Transaction's Hash in the output for debugging purposes?

Probably very, same for blocks.

Geod24:
Green

AndrejMitrovic (approved):

Issue #209, opened 2019-08-19
Integrate scpd code into Agora


bpalaggi:
As a developer, I need the SCP code to be integrated into Agora in order for the basic block of consensus to work.

**Definition of Done:**
- [x] There is a subclass of SCPDriver
- [x] Agreement is made with the quorum slice using SCP

PR #299, opened 2019-09-04
Build SCPP with debugging symbols

AndrejMitrovic:
It makes debugging easier.


Geod24 (approved):
We should probably make it conditional on a parameter, and enable `-O2` and friends in release mode, but we're still far from that.

AndrejMitrovic:
#209

PR #305, opened 2019-09-05
Debug regression with nightlies

Geod24:


source/agora/utils/PrettyPrinter.d
```diff
@@ -62,7 +62,8 @@ private struct AmountFmt
 @safe unittest
 {
     immutable one = Amount(1);
-    assert(format("%s", AmountFmt(Amount.MaxUnitSupply)) == "500,000,000.0000,000");
```
AndrejMitrovic, line 65: So can you help me understand the existing formatting? Because I didn't notice this before and it's very confusing to me.

Why are there 4 zeros before the comma? Shouldn't it be something like:

```
500,000,000.0,000,000
```

Help me understand the logic.

Geod24: To be fair, I expected exactly what you describe when I did the PR, but somehow the results were wrong so I just went with it. Looks like someone fixed it.

Geod24:
Probably due to https://github.com/dlang/phobos/pull/7112

PR #306, opened 2019-09-06
Write the Genesis block in the BlockStorage

### LINKED ISSUE #286: Ledger: Replace isEmpty() with a callback to prepare UTXOs
https://github.com/bpfkorea/agora/pull/275#issuecomment-526000739

AndrejMitrovic:
Fixes #286

This or https://github.com/bpfkorea/agora/pull/194 will need a rebase depending on which gets pulled first. So I'll mark this a blocked for now.

### DIFF
--- a/source/agora/node/BlockStorage.d
+++ b/source/agora/node/BlockStorage.d
@@ -46,12 +46,15 @@ public interface IBlockStorage
 
     /***************************************************************************
 
-        Returns:
-            true if the storage is empty (height index was not loaded)
+        Load the block storage. If there was nothing to load,
+        a Genesis block will be added to the ledger. In this case the calling
+        code should treat the block as new and update the set of UTXOs, etc.
+
+        If the blockchain data is corrupt the application will be halted.
 
     ***************************************************************************/
 
-    public bool isEmpty ();
+    public void load ();
 
     /***************************************************************************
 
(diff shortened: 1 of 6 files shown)


AndrejMitrovic:
Fixed the integration tests.

source/agora/node/BlockStorage.d
```diff
@@ -188,21 +194,32 @@ public class BlockStorage : IBlockStorage
 
         if (!this.path.exists)
             mkdirRecurse(this.path);
```
Geod24: I'd say this should be moved to `load` as well

source/agora/node/BlockStorage.d
```diff
@@ -962,14 +979,17 @@ public class MemBlockStorage : IBlockStorage
...
+        Adds the genesis block to the in-memory blockchain storage.
+
         Returns:
-            true if the storage is empty (height index was not loaded)
+            false as the storage was empty before the genesis block was added
 
     ***************************************************************************/
 
-    public bool isEmpty ()
+    public bool load ()
```
Geod24: It looks like the return value is only used in tests ? I don't think we need this information to be honest, it looks like a leaky abstraction...

AndrejMitrovic: Like I mentioned before, I need to add the UTXO set from the genesis block in the UTXO PR.

I should only add the UTXOs if the genesis block was added for the very first time.

AndrejMitrovic:
Ok put this on hold, I've got a simpler idea I'll work on later.

AndrejMitrovic:
So, there is a better way to do what I wanted.

The ledger should check if the UTXO database is empty. If it's empty, it can walk through the ledger storage and add each block's UTXOs. Simple.

Unblocked.

AndrejMitrovic:
Of course, it may be more efficient to get the mempool from the network in some cases, but for starters we can do it this way.

AndrejMitrovic:
Updated. There are no boolean return values anymore. It should be looked at commit by commit.

I still need to add one more test: if the UTXOSet was empty but the blockchain in the storage had multiple blocks, the UTXOSet should be properly re-built.

source/agora/node/Ledger.d
```diff
@@ -162,6 +164,26 @@ public class Ledger
...
+            block = the block to update the UTXO set with
+
+    ***************************************************************************/
+
+    private void updateUTXOSet (const ref Block block) nothrow @safe
+    {
+        scope (failure) assert(0);
+
+        // add the new UTXOs
+        block.txs.each!(tx => this.utxo_set.updateUtxoCache(tx));
```
Geod24: `UTXO`, will need a rename

Geod24 (approved):

Geod24 (approved):
LGTM, just waiting for the last test you mentioned.

AndrejMitrovic:
Updated with a test-case.

PR #310, opened 2019-09-06
Tests: Make starting the nodes asynchronous

### LINKED ISSUE #276: TestAPIManager.start() is not asynchronous
This is an issue, because it's actually calling `start()` for each created node, but it waits for a response back from the node before it starts another node. 

What happens is, this method is called in the child thread (and the main unittest thread *waits* for a response, even if it's a void function!):

```
/// The first task method, loading from disk, node discovery, etc
public void start ()
{
    logInfo("Doing network discovery..");
    this.network.discover();

}
```

The `discover()` method will not return until it is complete. What we have to do, is make sure `start` spawns a task. This will make the discovery phase asynchronous.

AndrejMitrovic:
Previously what happened during boot in the tests was:

- for each node N in M:
=> node.start();
  => calls discover();
    => tries to reach other nodes
      => other nodes haven't had their start() method called yet, therefore they are not ready to communicate (blocks are not loaded from disk yet, etc)

This is all because in LocalRest after we send a message to a node's thread we also wait for a response back.

With this PR the asynchronous starting of nodes is more thoroughly tested, and this also speeds up the tests a bit.

P.S.: Is there a better way to do this without using a `__gshared`?

Fixes #276

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -141,7 +141,7 @@ public class TestAPIManager
 
     /***************************************************************************
 
-        Start each of the nodes
+        Asynchronously start each of the nodes
 
         Params:
             count = Expected number of nodes
@@ -150,7 +150,17 @@ public class TestAPIManager
 
     public void start ()
     {
-        this.apis.each!(a => a.start());
+        import std.concurrency;
+
+        __gshared RemoteAPI!TestAPI[] nodes;
+        static void runNode (size_t index)
+        {
+            nodes[index].start();
+        }
+
+        nodes = this.apis.values;
+        foreach (idx; 0 .. nodes.length)
+            spawn(&runNode, idx);
     }
 
     /***************************************************************************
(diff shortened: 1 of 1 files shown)


Geod24 (changes requested):
I'd much rather have this done in `localrest`

AndrejMitrovic:
Yes me too, but it's a bit more involved. There is also a choice:

A) Should the calling thread just ignore the return value? If so, the called thread needs to know not to send the value back.
B) Should the calling thread eventually receive the value (via a callback), while not blocking the call?

AndrejMitrovic:
This PR doesn't have to be merged right now, it isn't necessarily required right now. So it can wait.

AndrejMitrovic:
> => other nodes haven't had their start() method called yet, therefore they are not ready to communicate (blocks are not loaded from disk yet, etc)

This part is no longer true. The block loading is done in the constructor.

Issue #204, opened 2019-08-19
Add freezing capability to UTXO


bpalaggi:
As a node operator, I can create a transaction that marks my UTXO as “frozen” so that I can use it as collateral for running a validator node.

**Definition of Done:**
- [x] An UTXO can be marked as "frozen"
- [x] When a transaction spending a frozen UTXO comes in, it marks the UTXO as "melting" for 2016 blocks
- [x] Transactions attempting to spend melting UTXO are invalid
- [x] After the UTXO has melt, the new UTXO can be spent without delay
- [x] The amount must be exactly 40,000 BOA (for the moment)

MichaelKim20:
### Transaction & UTXO in Freeze
The design was based on **Definition of Done**.
Illustrated the changing situation over time for **Block / Transaction / UTXOSet.**

<img width="920" alt="Transaction & UTXO in Freeze" src="https://user-images.githubusercontent.com/11639939/65112733-9c22a880-da1b-11e9-9f1b-55f4b00098c5.png">
<img width="896" alt="" src="https://user-images.githubusercontent.com/11639939/65110791-82319780-da14-11e9-9367-2f8eac1c87f0.png">



![Block_Tx_UTXO 001](https://user-images.githubusercontent.com/11639939/65367182-a7b6df00-dc68-11e9-9e56-224fe5c6b5ff.jpeg)

![Block_Tx_UTXO](https://user-images.githubusercontent.com/11639939/65367153-219a9880-dc68-11e9-89ae-39ac288ef2ee.gif)

![FreezingAndUTXO](https://user-images.githubusercontent.com/11639939/65403828-b213dd00-de10-11e9-9ac9-1e0bc328e7b5.gif)

MichaelKim20:
### Define of Subtask
1. Add type to struct `Transaction`
2. Implement UTXOSetValue to be used by UTXOSet
3. Implement  validation when receive transaction 'freeze'  in melted and , receive transaction 'payment'  in frozen.(TX2, TX3 in upper diagram)
4. Implement  validation when receive transaction 'payment' in various status (melting, melted)(TX4, TX5 in upper diagram)

Geod24:
Well done!

PR #311, opened 2019-09-06
Add type parameters to the transaction output struct.

TrustHenry:
The function of transaction output is divided into types.
The type currently defined is the same as 

/// General payment type
public immutable int TXO_PAYMENT = 0;

/// Freezing capability to UTXO type
public immutable int TXO_FREEZE = 1;

https://github.com/bpfkorea/agora/issues/204

### DIFF
--- a/source/agora/cli/SendTxProcess.d
+++ b/source/agora/cli/SendTxProcess.d
@@ -192,7 +192,7 @@ public int sendTxProcess (string[] args, ref string[] outputs,
     Transaction tx =
     {
         [Input(Hash.fromString(op.txhash), op.index)],
-        [Output(Amount(op.amount), PublicKey.fromString(op.address))]
+        [Output(Amount(op.amount), PublicKey.fromString(op.address), TXO_PAYMENT)]
     };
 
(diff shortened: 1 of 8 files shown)


AndrejMitrovic:
When using a field that specifies a type of something where it can have several pre-defined values, we should use an enum. For example:

```D
enum TxType : ubyte
{
    Payment,
    Freeze
}
```

The `: ubyte` means it will only consume a single byte, so you can have up to 255 values, which is more than enough (there are only two types here for now, Payment & Freeze).

AndrejMitrovic:
Then in the `Transaction` you might have a field like `TxType type;`

AndrejMitrovic:
I have updated the description to include a link to https://github.com/bpfkorea/agora/issues/204, and added `type-feature`, set the Milestone, and assigned this to the Kanban project board.

Don't worry about doing this on your own yet, we'll have some documentation written how to use Issues / Assignments on Github.

AndrejMitrovic:
Also good use of the `Draft PR` feature! 👍

TrustHenry:
#316 A new PR was registered to reflect the suggest.

Issue #312, opened 2019-09-06
Client connection to node blocks node internal communication


Geod24:
A bug that I discovered while working on the integration test: the transactions / block will only be gossiped *after* the connection to nodes are closed. This is probably because the re-entrancy of `vibe.web.rest` is not defined, so any network communication will just keep the object busy, which is terrible.
I'll need to come up with a simple example and report this upstream. Fortunately for us, the new RCP generator will fix this.

Geod24:
I just witnessed this again: A node was retrieving a large amount of blocks (720) and the metric endpoint was unresponsive for few seconds (around 1m15s according to the logs).

ferencdg:
I am wondering wether this unresponsiveness is caused by us instead of vibe.d. We are retrieving a maximum of 1024 blocks(720 in this case) in a sync call(as getBlocksFrom is not wrapped inside a task)

ferencdg:
never mind, it is wrapped inside a task at a higher level

Geod24:
It has recently come to my attention that `requestHTTP` [*is documented as synchronous*](https://vibed.org/api/vibe.http.client/requestHTTP).

ferencdg:
do we have the logs somewhere? I also want to verify how long it takes to process 720 blocks...

Geod24:
```
Jan 08 06:37:12 na-002 docker[22304]: [main(Ky6v) INF] Listening for requests on http://0.0.0.0:7826/
Jan 08 06:37:12 na-002 docker[22304]: 2021-01-08 06:37:12,175 Info [agora.network.NetworkManager] - Doing network discovery..
Jan 08 06:37:12 na-002 docker[22304]: [main(+O1h) INF] Listening for requests on http://[::]:9111/
Jan 08 06:37:12 na-002 docker[22304]: Failed to listen on 0.0.0.0:9111
Jan 08 06:37:12 na-002 docker[22304]: 2021-01-08 06:37:12,182 Info [agora.consensus.protocol.Nominator] - Starting nominating timer..
Jan 08 06:37:12 na-002 docker[22304]: 2021-01-08 06:37:12,183 Info [agora.network.NetworkManager] - Found new Validator: http://na-002.bosagora.io:6826 (key: GDNODE6ZXW2NNOOQIGN24MBEZRO5226LSMHGQA3MUAMYQSTJVR7XT6GH)
[...]

Issue #270, opened 2019-08-23
Add the ability to override any config file field from the command line


bpalaggi:
As a developer, I need the ability to override a config file field in order to provide me with flexibility while doing system integration test.

PR #313, opened 2019-09-06
Re-add system integration tests

Geod24:
```
They were disabled in a previous PR because they were inadvertently broken.
This commit also allows them to be run locally multiple times without manual intervention.
Additionally, they can be re-run without rebuilding the Docker image.
Finally, the output is much more readable now.
```

Lost a lot of time before I could find https://github.com/bpfkorea/agora/issues/312 , which is going to be a big blocker for everyone.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,16 +1,11 @@
-version: 2
+version: 2.1
 
 jobs:
   # Build the Docker image
   build:
-    docker:
-      - image: bpfk/circleci:latest
+    machine:
+      image: ubuntu-1604:201903-01
     steps:
-      - setup_remote_docker:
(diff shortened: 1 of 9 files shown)


Geod24:
Goddam it, we're hitting a limitation of CircleCI, where we can't mount volumes.
So yeah, we need #270

Geod24:
Okay I managed to work around the limitation. Let's see if it consistently passes the test.
Now I just need to add coverage export.

Geod24:
Okay, I think we can go ahead with this for the time being, there's much debugging to do on Agora side before I can add code coverage.

Geod24:
Rebased on v0.x.x

AndrejMitrovic (approved):

PR #316, opened 2019-09-09
Add type parameters to the transaction struct

TrustHenry:
Add freezing capability to UTXO #204 of part
The function of transaction is divided into types.
The type currently defined is the same as
(Payment, Freeze)
Unittest code change and hash changed due to structure change of transaction.

### DIFF
--- a/source/agora/common/Deserializer.d
+++ b/source/agora/common/Deserializer.d
@@ -59,6 +59,17 @@ public void deserializePart (T) (ref T record, scope DeserializeDg dg)
     record.deserialize(dg);
 }
 
+/// Enum support
+public void deserializePart (T)(ref T record, scope DeserializeDg dg)
+    nothrow @trusted
+    if (is(T == enum))
+{
+    import std.traits;
+    OriginalType!T orig_val;
+    deserializePart(orig_val, dg);
+    record = cast(T)(orig_val);
+}
+
 /// Ditto
 public void deserializePart (ref Hash record, scope DeserializeDg dg)
     nothrow @safe
--- a/source/agora/consensus/data/Transaction.d
+++ b/source/agora/consensus/data/Transaction.d
@@ -27,7 +27,14 @@ import agora.common.Serializer;
 import agora.consensus.Genesis;
 
 import std.algorithm;
+import std.traits;
 
+/// The Agora Transaction Type constans
+public enum TxType : ubyte
+{
+    Payment,
+    Freeze
+}
 
 /*******************************************************************************
 
@@ -40,6 +47,9 @@ import std.algorithm;
 
 public struct Transaction
 {
+    /// Transation type
+    public TxType type;
+
     /// The list of unspent `outputs` from previous transaction(s) that will be spent
     public Input[] inputs;
 
@@ -58,6 +68,9 @@ public struct Transaction
 
     public void computeHash (scope HashDg dg) const nothrow @safe @nogc
     {
+        OriginalType!TxType tx_type;
+        hashPart(tx_type, dg);
+
         foreach (input; this.inputs)
             hashPart(input, dg);
 
@@ -77,6 +90,9 @@ public struct Transaction
 
     public void serialize (scope SerializeDg dg) const @safe
     {
+        OriginalType!TxType tx_type;
+        serializePart(tx_type, dg);
+
         serializePart(this.inputs.length, dg);
         foreach (const ref input; this.inputs)
             serializePart(input, dg);
@@ -97,6 +113,8 @@ public struct Transaction
 
     public void deserialize (scope DeserializeDg dg) nothrow @safe
     {
+        deserializePart(this.type, dg);
+
         size_t input_size;
         deserializePart(input_size, dg);
         this.inputs.length = input_size;
@@ -132,9 +150,9 @@ nothrow @safe @nogc unittest
 {
     Transaction tx;
     auto hash = hashFull(tx);
-    auto exp_hash = Hash("0xcee29bfe1a706fd555b748145b683a904bb04e9344648913" ~
-        "5358eeaf31105ed219541ff717e2868a614758e140472f9172d2522585fdc6c6035" ~
-        "90142f7026a78");
+    auto exp_hash = Hash("0x4b6e507a0519827d48792e6f27f3a0f5b4bc284c69c83a2a" ~
+        "ebe35f11443e96364c90f780d4448320440efe68808fe4b6c7c4745d2e7c7e16956" ~
+        "987df86f6a32f");
     assert(hash == exp_hash);
 }
 
@@ -305,6 +323,7 @@ public bool isCoinbaseTx (Transaction tx) nothrow pure @safe @nogc
 public Transaction newCoinbaseTX (PublicKey address, Amount value = Amount(0))
 {
     return Transaction(
+        TxType.Payment,
         [Input(Hash.init, 0)],
         [Output(value, address)]
     );
@@ -371,6 +390,7 @@ public Transaction[] makeChainedTransactions (KeyPair key_pair,
 
         Transaction tx =
         {
+            TxType.Payment,
             [input],
             [Output(AmountPerTx, key_pair.address)]  // send to the same address
         };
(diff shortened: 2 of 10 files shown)


source/agora/consensus/data/Transaction.d
```diff
@@ -97,6 +110,9 @@ public struct Transaction
 
     public void deserialize (scope DeserializeDg dg) nothrow @safe
     {
+        ubyte type_size;
+        deserializePart(type_size, dg);
```
AndrejMitrovic: ?

you're not storing it to this.type.

TrustHenry: aH. Your right,

AndrejMitrovic:
> add Transaction Type constans …

I suggest renaming this to: `Introduce freezing transactions and TxType`

Then in the commit message, you can document what the different types of transactions exist.

Also, I suggest either renaming `Added type to print from PrettyPrinter` to `Add pretty-printing support to TxType` or even squashing the two commits together since it makes sense as an atomic change.

source/agora/consensus/data/Transaction.d
```diff
@@ -40,6 +46,9 @@ import std.algorithm;
 
 public struct Transaction
 {
+    /// Transation Type(Payment, Freeze)
```
AndrejMitrovic:
```suggestion
    /// Transation type
```

I suggest removing the stuff in the parentheses because they will go out of date quickly when we add new types to `TxType`. This comment will be forgotten about.

TrustHenry: Suggested change

source/agora/consensus/data/Transaction.d
```diff
@@ -97,6 +110,10 @@ public struct Transaction
 
     public void deserialize (scope DeserializeDg dg) nothrow @safe
     {
+        ubyte type;
+        deserializePart(type, dg);
+        this.type = cast(TxType)type;
```
AndrejMitrovic: This cast should not be here. Instead, I suggest writing an explicit deserialization function in `Deserialize.d`:

```D
/// Ditto
public void deserializePart (ref TxType record, scope DeserializeDg dg)
    nothrow @trusted
{
    import std.traits;
    OriginalType!TxType orig_val;
    deserializePart(orig_val, dg);
    record = *cast(TxType*)(orig_val);
}
```

TrustHenry: Thanks for suggest.

source/agora/consensus/data/Transaction.d
```diff
@@ -97,6 +110,10 @@ public struct Transaction
 
     public void deserialize (scope DeserializeDg dg) nothrow @safe
     {
+        TxType type_val;
+        deserializePart(type_val, dg);
```
AndrejMitrovic: You don't need to declare a temporary anymore. You can write `deserializePart(this.type, dg);`

TrustHenry: Suggested change

AndrejMitrovic (approved):

TrustHenry:
@Geod24 ping.

source/agora/common/Deserializer.d
```diff
@@ -66,6 +67,16 @@ public void deserializePart (ref Hash record, scope DeserializeDg dg)
     record = Hash(dg(Hash.sizeof));
 }
 
+/// Ditto
+public void deserializePart (ref TxType record, scope DeserializeDg dg)
```
Geod24: I would say we handle this business within `Transaction`. This module should not depend on `consensus` module

source/agora/consensus/data/Transaction.d
```diff
@@ -77,6 +88,8 @@ public struct Transaction
 
     public void serialize (scope SerializeDg dg) const @safe
     {
+        serializePart(this.type, dg);
```
Geod24: Deserialize it here directly using `OriginalType`

source/agora/utils/PrettyPrinter.d
```diff
@@ -222,6 +222,9 @@ private struct TransactionFmt
         enum InputPerLine = 3;
         enum OutputPerLine = 3;
 
+        formattedWrite(sink, "Type : %s\n",
+            this.value.type);
+
         formattedWrite(sink, "Inputs (%d): %(%(%s, %),\n%)\n",
```
Geod24:
```suggestion
        formattedWrite(sink, "Type: %s, Inputs (%d): %(%(%s, %),\n%)\n",
```

The goal is to have it be compact, as we might print a lot of data otherwise.

Geod24 (commented):
I think it might be worth it to introduce a constructor for transaction which takes the `TxType` as last, default argument. Making a transaction is a very common action in code (e.g. tests), but the overwhelming majority will be payment. That would make this diff trivial also.

source/agora/consensus/data/Transaction.d
Geod24: Ah sorry, I see that's what you did initially. @AndrejMitrovic : The problem with the explicit deserialization overload is that it brings in dependency. With this approach, the `Serialization` module will depend on every type in the program.

AndrejMitrovic: Well I just worry about this code segment being duplicated anywhere TxType is used. For example, in the UTXO set Michael is working on.

Geod24: Random idea: `public void deserializePart (T) (ref T TxType ...) if (is(T == enum)` ?

AndrejMitrovic:
```D
/// Enum support
public void deserializePart (T)(ref T record, scope DeserializeDg dg) nothrow @trusted
    if (is(T == enum)) 
{
    import std.traits;
    OriginalType!T orig_val;
    deserializePart(orig_val, dg);
    record = *cast(T*)(orig_val);
}
```

source/agora/common/Deserializer.d
TrustHenry: I've remove the `Transaction`  It's make not depend on consensus module.

source/agora/consensus/data/Transaction.d
TrustHenry: Using OriginalType

source/agora/utils/PrettyPrinter.d
TrustHenry: Removed line feed.

source/agora/consensus/data/Transaction.d
TrustHenry: > ```d
> /// Enum support
> public void deserializePart (T)(ref T record, scope DeserializeDg dg) nothrow @trusted
>     if (is(T == enum)) 
> {
>     import std.traits;
>     OriginalType!T orig_val;
>     deserializePart(orig_val, dg);
>     record = *cast(T*)(orig_val);
> }
> ```

Is this the best?

Geod24: That addresses both @AndrejMitrovic and my concerns, so I think so. Sorry for the back and forth.

TrustHenry: not depend on consensus module  :)

TrustHenry:
It's addresses both @AndrejMitrovic and @Geod24 concerns. :)
@Geod24 @AndrejMitrovic ready for review.

Geod24:
@TrustHenry : What about https://github.com/bpfkorea/agora/pull/316#pullrequestreview-285907712 ?

TrustHenry:
> @TrustHenry : What about [#316 (review)](https://github.com/bpfkorea/agora/pull/316#pullrequestreview-285907712) ?

For example
`public Transaction DefalutTransaction =
{
    TxType.Payment,
    inputs: [ Input.init ],
    outputs: [ Output.init ]
};`

TrustHenry:
Or

TrustHenry:
Or
public struct Transaction
{
    /// Transation type
    public TxType type = TxType.Payment;

Geod24 (approved):

source/agora/common/Deserializer.d
```diff
@@ -59,6 +59,17 @@ public void deserializePart (T) (ref T record, scope DeserializeDg dg)
     record.deserialize(dg);
 }
 
+/// Enum support
+public void deserializePart (T)(ref T record, scope DeserializeDg dg)
+    nothrow @trusted
+    if (is(T == enum))
```
MichaelKim20, line 65: What a great way!

MichaelKim20 (approved):

TrustHenry:
It's Green

Geod24:
According to our vision.

PR #319, opened 2019-09-09
Collection of minor fixes to BlockStorage

Geod24:


source/agora/node/BlockStorage.d
```diff
@@ -50,11 +50,12 @@ public interface IBlockStorage
         a Genesis block will be added to the ledger. In this case the calling
         code should treat the block as new and update the set of UTXOs, etc.
 
-        If the blockchain data is corrupt the application will be halted.
+        Returns:
+            `false if the data couldn't be loaded, `true` otherwise.
```
AndrejMitrovic:
```
`false
```

 => 

```
`false`
```

source/agora/node/BlockStorage.d
```diff
@@ -194,24 +208,30 @@ public class BlockStorage : IBlockStorage
...
         if (!this.loadAllIndexes())
-            assert(0);
+            return false;
 
         // Add Genesis if the storage is empty
         if (this.height_idx.length == 0)
         {
             if (!this.saveBlock(GenesisBlock))
-                assert(0);
+                return false;
```
AndrejMitrovic, line 237: This isn't about loading, but saving. So I think it should stay as `assert(0);`

AndrejMitrovic: Although I see you do the same in the calling code.. It's a bit of an interesting case..

AndrejMitrovic: Ok based on the commit message it makes sense, let it be like this.

source/agora/node/BlockStorage.d
```diff
@@ -809,22 +808,6 @@ public class BlockStorage : IBlockStorage
...
-    /***************************************************************************
-
-        Remove the index file.
-
-        Params:
-            path = path to the data directory
-
-    ***************************************************************************/
-
-    public static void removeIndexFile (string path)
```
AndrejMitrovic: Was this function unused? It seems out of place in this commit.

AndrejMitrovic (approved):
LG except the minor comments.

source/agora/node/BlockStorage.d
Geod24: Yep it was. And as I was grepping for `buildPath` I came accross this. Want it in another commit ?

AndrejMitrovic: I guess it doesn't matter much if it was no-op.

Geod24: It was actually used in the integration tests. So I reverted the `rm`.

Geod24:
There was a stupid mistake that the integration test caught, fixed now

PR #320, opened 2019-09-09
Fix #270: Add ability to override any config field from the command line

### LINKED ISSUE #270: Add the ability to override any config file field from the command line

Geod24:
```
This is a crude approach, and it does not actually cover quorum config,
but it lays the foundation to do it and unblocks our integration tests.
```

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -228,16 +228,13 @@ private Config parseConfigFileImpl (CommandLine cmdln)
         return result;
     }
 
-    enforce("node" in root, "The 'node' section is required");
-    enforce("quorum" in root, "The 'quorum' section is required");
-
     Config conf =
     {
         banman : parseBanManagerConfig(root["banman"]),
-        node : parseNodeConfig(root["node"]),
+        node : parseNodeConfig("node" in root),
         network : assumeUnique(parseSequence("network")),
         dns_seeds : assumeUnique(parseSequence("dns", true)),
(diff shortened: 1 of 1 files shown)


source/agora/common/Config.d
```diff
@@ -164,7 +180,12 @@ public GetoptResult parseCommandLine (ref CommandLine cmdline, string[] args)
 
         "config-check",
             "Check the state of the config and exit",
-            &cmdline.config_check);
+            &cmdline.config_check,
+
+        "override|O",
+            "Override a config file value",
```
AndrejMitrovic: It would be nice to show an example how it's actually done, in this help message. is it just `-O validator=false` or `-O node:validator=false`, etc?

AndrejMitrovic:
```suggestion
            "Override a config file value.\n" ~
            "Example: ./agora -O node.validator=true -o dns=1.1.1.1 -o dns=2.2.2.2\n" ~
            "Array values are additive, other items are set to the last override",
```

AndrejMitrovic:
Will merge after conflict is resolved.

Geod24:
Fixed / rebased, but my use case is gone now. This could definitely use some tests.

Geod24:
Ping ?

AndrejMitrovic (approved):

PR #325, opened 2019-09-11
Fix some of the SCP binding code

AndrejMitrovic:
I'm not sure why only some methods were commented with `// Slot in the vtable`, on the C++ side all the methods are marked as `virtual`

### DIFF
--- a/source/scpd/scp/SCPDriver.d
+++ b/source/scpd/scp/SCPDriver.d
@@ -24,13 +24,10 @@ extern(C++, `stellar`):
 
 public abstract class SCPDriver
 {
-    void D0 () {} // First dtor
-    void D1 () {} // Second dtor
-    //~this();
+    ~this() {}
 
     // Envelope signature/verification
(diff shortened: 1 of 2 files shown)


source/scpd/scp/SCPDriver.d
```diff
@@ -63,29 +60,27 @@ public abstract class SCPDriver
         kFullyValidatedValue, // value is valid for sure
         kMaybeValidValue      // value may be valid
     }
-    ValidationLevel validateValue(uint64_t slotIndex, ref const(Value) value, bool nomination)
-    {
-        return ValidationLevel.kMaybeValidValue;
```
Geod24, line 68: Why did you remove the default impl ?

Geod24 (commented):
I think I upgrade the SCP code to HEAD just before committing the initial version.
Looks like I missed some binding updates 🤷‍♂ 
Why the default implementation removal ?

Geod24 (approved):

PR #327, opened 2019-09-16
Automatically deploy documentation through Github Actions

Geod24:
There's a 50% chance this doesn't work on the first try, as I need to check if `GITHUB_TOKEN` has proper access, and can't test this in my fork. But we'll only see after merging 🤷‍♂

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -36,6 +36,7 @@
         "source/scpp/build/uint128_t.o"
     ],
     "lflags": [ "-lsodium", "-lstdc++", "-lsqlite3" ],
+    "buildRequirements": [ "allowWarnings" ],
 
     "configurations": [
         {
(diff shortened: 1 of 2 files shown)


dub.json
```diff
@@ -36,6 +36,7 @@
         "source/scpp/build/uint128_t.o"
     ],
     "lflags": [ "-lsodium", "-lstdc++", "-lsqlite3" ],
+    "buildRequirements": [ "allowWarnings" ],
```
AndrejMitrovic, line 39: Would it be possible to make this versioned, e.g. only use this flag for the `ddox` target?

Geod24: Not that I know of. But while they're allowed, we'll still see them on the console.

AndrejMitrovic (approved):

PR #329, opened 2019-09-17
Rename 'deserialize' to 'deserializeFull'

Geod24:
There are two reason for this:
- We do not want to encourage the use of 'deserialize', as it allocates memory.
  However having such a prominent name makes it the first tool a developer will reach for.
- The previous behaviour wasn't friendly to generic code, and only supported aggregates,
  so one had to 'switch' over the types in the calling code.
  This now mirror the approach of 'serializeFull' and is usable in generic code.


Geod24:
Rebased on v0.x.x

AndrejMitrovic (approved):

PR #331, opened 2019-09-17
Allow DeserializeDg to throw

Geod24:
Unfortunately, `DeserializeDg` could potentially `throw`:
The (badly named) delegate provides data to the deserialization function,
according to the requested size.
This data could potentially be read from a network socket.
In the event said socket terminate, an Exception could be thrown.

### DIFF
--- a/source/agora/common/Amount.d
+++ b/source/agora/common/Amount.d
@@ -77,7 +77,7 @@ public struct Amount
     }
 
     /// Support for deserialization
-    public void deserialize (scope DeserializeDg dg) nothrow @safe
+    public void deserialize (scope DeserializeDg dg) @safe
     {
         deserializePart(this.value, dg);
     }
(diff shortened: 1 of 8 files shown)


AndrejMitrovic:
What if we made the serialization / deserialization routines templated on the delegate type? We don't (currently) use overridable serialization methods, but maybe we should keep it that way.

Geod24:
Yeah it's one of the possibilities I considered. I think it would work in the present case, but I am not done with the RPC generator yet so I wanted to go with the least invasive approach, which also mirrors what we done with the Serializer.

To be fair, having the `DeserializeDg` returns a `ubyte[]` of the provided size is not a nice interface to work with, and I think it needs a deeper refactoring at some point.

AndrejMitrovic (approved):

PR #333, opened 2019-09-18
Implement support for walking through an std::set

AndrejMitrovic:
This is mainly needed for `SCPDriver.combineCandidates()` which selects one value out of a given set of candidate values.

C++ can't (easily) call D delegates, so I worked around this with a wrapper function.

### DIFF
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -113,13 +113,13 @@ extern(C++, `std`) {
     }
 }
 
-private extern(C++) set!uint makeTestSet();
+private extern(C++) set!uint* makeTestSet();
 
 unittest
 {
     auto set = makeTestSet;
     uint[] values;
-    foreach (val; set)
+    foreach (val; *set)
         values ~= val;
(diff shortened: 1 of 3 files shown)


Geod24 (approved):
Sweet

AndrejMitrovic:
Damn, it segfaults on ldc.

AndrejMitrovic:
~Ah it's unrelated to this PR, it's "that" segfault bug..~ Investigating..

AndrejMitrovic:
I can't reproduce this failure in Docker.

AndrejMitrovic:
The issue is something to do with returning a `std::set` by value from a function (which I only used in a unittest). I'll work around it, and reduce a test-case later.

AndrejMitrovic:
Ah right, it probably uses RAII on the C++ side and destroys the object.

AndrejMitrovic:
Great, now DMD & osx fails.

Geod24 (approved):

PR #336, opened 2019-09-18
Add ability to convert between QuorumConfig <-> SCPQuorumSet

AndrejMitrovic:
Alternatively, we could replace QuorumConfig and use SCPQuorum directly? Let me know what you think.

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -20,6 +20,9 @@ import agora.common.crypto.Key;
 import agora.common.Set;
 import agora.utils.Log;
 
+import scpd.types.Stellar_SCP;
+import scpd.types.Utils;
+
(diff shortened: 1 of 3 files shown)


Geod24:
> Alternatively, we could replace QuorumConfig and use SCPQuorum directly?

Is there any downside to this ?

AndrejMitrovic:
I can't think of any, other than depending more on SCP types.

AndrejMitrovic:
Hmm well there is one problem, trying to make SCPQuorumSet immutable. Because everything in the `Config` has to be shared (to be used with LocalRest..).

Geod24:
Or we could use the same trick we use with function arguments: JSON serialization.

Geod24 (approved):

PR #326, opened 2019-09-11
Add freezing capability to UTXO

MichaelKim20:
Add freezing capability to UTXO #204 of part

### Transaction & UTXO in Freeze
The design was based on **Definition of Done**.
Illustrated the changing situation over time for **Block / Transaction / UTXOSet.**
Transaction & UTXO in Freeze

Implement validation when receive transaction 'freeze' in melted and , receive transaction 'payment' in frozen.(TX2, TX3 in upper diagram)
Implement validation when receive transaction 'payment' in various status (melting, melted)(TX4, TX5 in upper diagram)

PR #337, opened 2019-09-18
Implement UTXOSetValue for freezing

MichaelKim20:
Add freezing capability to UTXO #204 of part

UTXOSetValue is used by UTXOSet
To freeze UTXO, add the following information to UTXOSet:
The height of block.
The type of transaction.

UTXO 008

Block_Tx_UTXO

Transaction & UTXO in Freeze

### DIFF
--- a/source/agora/consensus/data/UTXOSet.d
+++ b/source/agora/consensus/data/UTXOSet.d
@@ -191,38 +191,6 @@ public class UTXOSet
 
     /***************************************************************************
 
-        Find an UTXOSetValue in the UTXO set.
-
-        Params:
-            hash = the hash of transation
-            index = the index of the output
-            value = will contain the UTXOSetValue if found
-
-        Return:
-            Return true if the UTXO was found
-
-    ***************************************************************************/
-
-    private bool findUTXOSetValue (Hash hash, size_t index,
-                                        out UTXOSetValue value)
-        nothrow @safe
-    {
-        auto utxo_hash = getHash(hash, index);
-
-        if (utxo_hash in this.used_utxos)
-            return false;  // double-spend
-
-        if (this.utxo_db.find(utxo_hash, value))
-        {
-            this.used_utxos.put(utxo_hash);
-            return true;
-        }
-
-        return false;
-    }
-
-    /***************************************************************************
-
         Prepare tracking double-spent transactions and
         return the UTXOFinder delegate
 
@@ -239,19 +207,19 @@ public class UTXOSet
 
     /***************************************************************************
 
-        Find an unspent Output in the UTXO set.
+        Find an UTXOSetValue in the UTXO set.
 
         Params:
             hash = the hash of transation
             index = the index of the output
-            output = will contain the UTXO if found
+            output = will contain the UTXOSetValue if found
 
         Return:
             Return true if the UTXO was found
 
     ***************************************************************************/
 
-    private bool findUTXO (Hash hash, size_t index, out Output output)
+    private bool findUTXO (Hash hash, size_t index, out UTXOSetValue value)
         nothrow @safe
     {
         auto utxo_hash = getHash(hash, index);
@@ -259,10 +227,8 @@ public class UTXOSet
         if (utxo_hash in this.used_utxos)
             return false;  // double-spend
 
-        UTXOSetValue value;
         if (this.utxo_db.find(utxo_hash, value))
         {
-            output = value.output;
             this.used_utxos.put(utxo_hash);
             return true;
         }
--- a/source/agora/consensus/data/UTXOSet.d
+++ b/source/agora/consensus/data/UTXOSet.d
@@ -27,6 +27,51 @@ import std.file;
 
 mixin AddLogger!();
 
+/// The structure of spendable transaction output
+public struct UTXOSetValue
+{
+    /// Height of the block to be unlock
+    ulong unlock_height;
+
+    /// Transaction type
+    TxType type;
+
+    /// Unspend transaction output
+    Output output;
+
+    /***************************************************************************
+
+        Input Serialization
+
+        Params:
+             dg = serialize function
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) const @safe
+    {
+        serializePart(this.unlock_height, dg);
+        serializePart(this.type, dg);
+        serializePart(this.output, dg);
+    }
+
+    /***************************************************************************
+
+        Input deserialization
+
+        Params:
+             dg = deserialize function
+
+    ***************************************************************************/
+
+    public void deserialize (scope DeserializeDg dg) @safe
+    {
+        deserializePart(this.unlock_height, dg);
+        deserializePart(this.type, dg);
+        deserializePart(this.output, dg);
+    }
+}
+
 /// ditto
 public class UTXOSet
 {
--- a/source/agora/consensus/data/UTXOSet.d
+++ b/source/agora/consensus/data/UTXOSet.d
@@ -126,11 +126,30 @@ public class UTXOSet
 
         Params:
             tx = the transaction
+            height = Height of the block where UTXO will be unlocked
 
     ***************************************************************************/
 
-    public void updateUTXOCache (const ref Transaction tx) @safe
+    public void updateUTXOCache (const ref Transaction tx, ulong height) @safe
     {
+        import std.algorithm : any;
+
+        // defaults to next block
+        ulong unlock_height = height + 1;
+
+        // for payments of frozen transactions, it will melt after 2016 blocks
+        if ((tx.type == TxType.Payment)
+            && tx.inputs.any!(input =>
+                (
+                    (input.previous != Hash.init) &&
+                    (this.getUTXOSetValue(input.previous, input.index).type == TxType.Freeze)
+                )
+            )
+        )
+        {
+            unlock_height = height + 2016;
+        }
+
         foreach (const ref input; tx.inputs)
         {
             auto utxo_hash = getHash(input.previous, input.index);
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
This pull request introduces `findUTXOSetValue`, and then removes it in another commit. It should just be removed from the commit it was originally added in, and then the commit `Remove unused function UTXOSetValue.findUTXOSetValue` can be removed.

source/agora/consensus/data/UTXOSet.d
```diff
@@ -27,6 +27,51 @@ import std.file;
 
 mixin AddLogger!();
 
+/// The structure of spendable transaction output
+public struct UTXOSetValue
+{
+    /// Height of the block to be activated
+    ulong height;
```
Geod24: This could use a more descriptive name.
I can suggest "height_lock", "frozen_until", "locked_until", "min_height"... But I'm open to suggestion @TrustHenry @AndrejMitrovic @MukeunKim

AndrejMitrovic: I suggest snake_case.

Geod24:
> This pull request introduces findUTXOSetValue, and then removes it in another commit.b

Damn Github fix your commit ordering.

source/agora/consensus/data/UTXOSet.d
TrustHenry: I suggest "height_lock".

source/agora/consensus/data/UTXOSet.d
```diff
@@ -81,25 +126,116 @@ public class UTXOSet
...
+                    return;
+            }
+
+            foreach (const ref input; tx.inputs)
+            {
+                auto utxo_hash = getHash(input.previous, input.index);
+                this.utxo_db.remove(utxo_hash);
+            }
+
+            ulong actived_height = height + 1;
```
AndrejMitrovic: Why +1? The UTXO will be in the block at the provided height that's in the parameter, not height + 1.

source/agora/consensus/data/UTXOSet.d
```diff
@@ -81,25 +126,116 @@ public class UTXOSet
...
-        foreach (const ref input; tx.inputs)
+        if (tx.type == TxType.Freeze)
         {
-            auto utxo_hash = getHash(input.previous, input.index);
-            this.utxo_db.remove(utxo_hash);
+            UTXOSetValue utxo_value;
+            foreach (const ref input; tx.inputs)
+            {
+                if (!this.getUTXOSetValue(input.previous, input.index, utxo_value))
+                    return;
```
AndrejMitrovic: These should `assert(0)`, not fail silently.

source/agora/consensus/data/UTXOSet.d
```diff
@@ -81,25 +126,116 @@ public class UTXOSet
...
+                else if (utxo_value.type != TxType.Payment)
+                    return;
```
AndrejMitrovic: Same comment about assert.

source/agora/consensus/data/UTXOSet.d
```diff
@@ -81,25 +126,116 @@ public class UTXOSet
...
+                    else if (input.previous == Hash.init)
+                        n_payment++;
+                }
+
+                if (n_payment == tx.inputs.length)
+                    tx_type = TxType.Payment;
+                else if (n_freeze == tx.inputs.length)
+                    tx_type = TxType.Freeze;
+                else
+                    return;
```
AndrejMitrovic: `assert(0)`

source/agora/consensus/data/UTXOSet.d
MichaelKim20: I suggest "locked_until"~

source/agora/consensus/data/UTXOSet.d
MichaelKim20: Next Block height

source/agora/consensus/data/UTXOSet.d
AndrejMitrovic: I would go with @MukeunKim's suggestion, or maybe `unlock_height`

source/agora/consensus/data/UTXOSet.d
MichaelKim20: I entered the available height, i.e. unlock height.
I wanted to specify what could be used in the next block, not in the current block.

source/agora/consensus/data/UTXOSet.d
MichaelKim20: I like `unlock_height`

source/agora/consensus/data/UTXOSet.d
MukeunKim: I use log.error()

source/agora/consensus/data/UTXOSet.d
MukeunKim: I use log.error()

MichaelKim20:
Ready fo Review

source/agora/consensus/data/UTXOSet.d
TrustHenry: Me too `unlonk_height`  because  `height + 1;`

source/agora/consensus/data/UTXOSet.d
MichaelKim20: Remove this part. I have confirmed that this is not necessary.

source/agora/node/Ledger.d
```diff
@@ -542,7 +543,138 @@ unittest
...
+                tx_type,
+                last_txs,
+                GenesisTransaction);
+
+            txes.each!((tx)
+                {
+                    assert(ledger.acceptTransaction(tx) == is_valid);
+                });
+
+            if (is_valid)
```
AndrejMitrovic: `is_valid` is always true, because it is never set to false.

MichaelKim20: It's not used here, so we'll have to remove it

MichaelKim20: However, if an error occurs and the block is not created, then the values of these variables must not be changed.

MichaelKim20: Ready for review

AndrejMitrovic:
Unless I'm missing something, this is what `updateUTXOCache` should look like:

```D
public void updateUTXOCache (const ref Transaction tx, ulong height) @safe
{
    ulong unlock_height = (tx.type == TxType.Freeze)
        ? height + 2016 : height + 1;

    foreach (const ref input; tx.inputs)
    {
        auto utxo_hash = getHash(input.previous, input.index);
        this.utxo_db.remove(utxo_hash);
    }

    Hash tx_hash = tx.hashFull();
    foreach (idx, output; tx.outputs)
    {
        auto utxo_hash = getHash(tx_hash, idx);
        const UTXOSetValue value = UTXOSetValue(unlock_height, tx.type, output);
        this.utxo_db[utxo_hash] = value;
    }
}
```

And also, this code passes the test-suite you wrote.

The validation rules for payment => freeze => melting should live in `isInvalidReason`, and not be scattered across multiple functions. `updateUTXOCache()`'s only purpose is to Remove spent UTXOs, and add new ones.

AndrejMitrovic:
My previous comment was wrong with regards to unfreezing. Let me see if it can still be simplified..

source/agora/consensus/data/UTXOSet.d
```diff
@@ -81,25 +126,105 @@ public class UTXOSet
...
+                        if (utxo_value.type == TxType.Payment)
+                            n_payment++;
+                        else
+                            n_freeze++;
+                    }
+                    else if (input.previous == Hash.init)
+                        n_payment++;
+                }
+
+                if (n_freeze == tx.inputs.length)
```
AndrejMitrovic: This is the part I find a bit confusing. The new outputs (UTXOs) will be be melting only if **all** of the inputs are frozen UTXOs. But if one of the was a Payment, then all of the outputs will not be melting, they will be instantly melted?

MichaelKim20:
### There are three more commit after that.
It is PR #326.

**Includes:**
Implement validation of transactions associated with freezing
Add test for freezing and melting while creating blocks

MichaelKim20:
Ready for review

source/agora/consensus/data/UTXOSet.d
```diff
@@ -81,25 +126,89 @@ public class UTXOSet
...
+                    {
+                        tx_type = TxType.Freeze;
+                        break;
+                    }
+                }
+            }
+
+            foreach (const ref input; tx.inputs)
+            {
+                auto utxo_hash = getHash(input.previous, input.index);
```
AndrejMitrovic: This part seems to be the same for the freeze transactions too. Maybe it could be moved before the `if (tx.type == TxType.Freeze)` check so the code doesn't have to be duplicated.

AndrejMitrovic:
I suggest rewriting these two functions as follows:

```D

    Add all of a transaction's outputs to the Utxo set,
    and remove the spent outputs in the transaction from the set.

    Params:
        tx = the transaction
        height = Height of the block where UTXO will be unlocked

public void updateUTXOCache (const ref Transaction tx, ulong height) @safe
{
    import std.algorithm : any;

    foreach (const ref input; tx.inputs)
    {
        auto utxo_hash = getHash(input.previous, input.index);
        this.utxo_db.remove(utxo_hash);
    }

    // defaults to next block
    ulong unlock_height = height + 1;

    // for payments of frozen transactions, it will melt after 2016 blocks
    if (tx.type == TxType.Payment
        && tx.inputs.any!(input => this.getUTXOSetValue(
            input.previous, input.index).type == TxType.Freeze))
    {
        unlock_height = height + 2016;
    }

    Hash tx_hash = tx.hashFull();
    foreach (idx, output; tx.outputs)
    {
        auto utxo_hash = getHash(tx_hash, idx);
        auto utxo_value = UTXOSetValue(unlock_height, tx.type, output);
        this.utxo_db[utxo_hash] = utxo_value;
    }
}

    get an UTXOSetValue in the UTXO set.

    Params:
        hash = the hash of transation
        index = the index of the output
        value = will contain the UTXOSetValue if found

    Return:
        Return true if the UTXO was found

private UTXOSetValue getUTXOSetValue (Hash hash, size_t index)
    nothrow @safe
{
    auto utxo_hash = getHash(hash, index);

    UTXOSetValue value;
    if (this.utxo_db.find(utxo_hash, value))
        assert(0);  // the UTXO must exist
    return value;
}
```

It gets rid of code duplication. And also, `getUTXOSetValue` must return a valid value because `updateUTXOCache` is called after validation. We know at this point a previous UTXO must exist.

MichaelKim20:
To use 'any!' I think it's better to change it.

MichaelKim20:
Ready for review

Geod24 (approved):

AndrejMitrovic:
The bug in my suggestion was the over-eager removal of inputs. So this can be fixed to this:

```D
public void updateUTXOCache (const ref Transaction tx, ulong height) @safe
{
    import std.algorithm : any;

    // defaults to next block
    ulong unlock_height = height + 1;

    // for payments of frozen transactions, it will melt after 2016 blocks
    if (tx.type == TxType.Payment
        && tx.inputs.any!(input => this.getUTXOSetValue(
            input.previous, input.index).type == TxType.Freeze))
    {
        unlock_height = height + 2016;
    }

    Hash tx_hash = tx.hashFull();
    foreach (idx, output; tx.outputs)
    {
        auto utxo_hash = getHash(tx_hash, idx);
        auto utxo_value = UTXOSetValue(unlock_height, tx.type, output);
        this.utxo_db[utxo_hash] = utxo_value;
    }

    foreach (const ref input; tx.inputs)
    {
        auto utxo_hash = getHash(input.previous, input.index);
        this.utxo_db.remove(utxo_hash);
    }
}
```

AndrejMitrovic:
However, this assertion will fail:

```D
private UTXOSetValue getUTXOSetValue (Hash hash, size_t index)
    nothrow @safe
{
    auto utxo_hash = getHash(hash, index);

    UTXOSetValue value;
    if (this.utxo_db.find(utxo_hash, value))
        assert(0);  // assertion triggers
    return value;
}
```

That means either something with the tests is wrong, or the validation functions aren't catching an edge-case.

MichaelKim20:
@AndrejMitrovic I fixed it. Thank you.

MichaelKim20:
The full test code using `UTXSet.updateUTXOCache` is in PR #326.
So we had a hard time proceeding with the review here.
Anyway, thank you all.

PR #338, opened 2019-09-18
Change genesis amount to 500,000,000 BOA

MichaelKim20:
The amount was increased to test the frozen transaction.

### DIFF
--- a/source/agora/consensus/Genesis.d
+++ b/source/agora/consensus/Genesis.d
@@ -20,7 +20,6 @@ import agora.common.Hash;
 import agora.common.crypto.Key;
 import agora.consensus.data.Transaction;
 
-
 /// The genesis block
 public immutable Block GenesisBlock =
 {
@@ -46,8 +45,8 @@ unittest
 
 ///
 private immutable Hash GenesisMerkleRoot =
-    Hash(`0xd04f8dd627ec5245129527b2327ad73d1157d483441a249d7eecd086ee1c4775`
-         ~ `3ad14d410f2bf128d8737366038c85d223e5701b73c7238089f52898ecd0e1b2`);
+    Hash(`0xdb6e67f59fe0b30676037e4970705df8287f0de38298dcc09e50a8e85413959ca`
+         ~ `4c52a9fa1edbe6a47cbb6b5e9b2a19b4d0877cc1f5955a7166fe6884eecd2c3`);
 
 /// The single transaction that are part of the genesis block
 public immutable Transaction GenesisTransaction =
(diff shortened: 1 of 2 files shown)


source/agora/consensus/Genesis.d
```diff
@@ -46,23 +46,55 @@ unittest
...
-        Output(Amount(40_000_000 / Block.TxsInBlock), GenesisOutputAddress),
-        Output(Amount(40_000_000 / Block.TxsInBlock), GenesisOutputAddress),
-        Output(Amount(40_000_000 / Block.TxsInBlock), GenesisOutputAddress),
-        Output(Amount(40_000_000 / Block.TxsInBlock), GenesisOutputAddress),
-        Output(Amount(40_000_000 / Block.TxsInBlock), GenesisOutputAddress),
-        Output(Amount(40_000_000 / Block.TxsInBlock), GenesisOutputAddress),
-        Output(Amount(40_000_000 / Block.TxsInBlock), GenesisOutputAddress),
+        Output(
+            Amount(
+                Amount.MaxCoinSupply * Amount.Unit / Block.TxsInBlock
```
AndrejMitrovic: @Geod24 there is already a `MaxUnitSupply` that we could use here, but we can't divide an `Amount`. Because I wanted to originally suggest here:

```D
Amount.MaxUnitSupply / Block.TxsInBlock
```

So what should we do here? Because we can't access `.value` to divide the amount (it's private).

MichaelKim20: Amount.value is private

MichaelKim20:
@Geod24 
Ready for review

source/agora/consensus/Genesis.d
AndrejMitrovic: @Geod24 how about my question here? `Amount.MaxUnitSupply` already exists, but it's an `Amount`, which is not divisible.

Geod24:
Hum I see two ways:
- Making Genesis output's value a constant in Amount. It seems cleaner than duplicating the 10M constant, but will only be used in one place, but will be in another module;
- Supporting division in `Amount`;

I lean heavily for the second. However, as we are building the validation rules, we'll run into a problem: Genesis will need to create a frozen transaction, otherwise we'll end up with a blockchain without validator! So I think we should put it off for a bit (mark this as blocked) and revisit soon, unless it is blocking something, @MukeunKim ?

MichaelKim20:
@Geod24 Ready for review :)

Geod24 (approved):

AndrejMitrovic (approved):

PR #339, opened 2019-09-18
Emit reason for failed block transaction validation

AndrejMitrovic:
If one of the transactions in a block failed validation,
it's useful to know which validation rule failed.

The parent logger already emits that the Block itself failed,
and the block is dumped to the logger.

### DIFF
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -310,8 +310,11 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
     if (!block.txs.isSorted())
         return "Block: Transactions are not sorted";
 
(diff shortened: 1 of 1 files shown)


source/agora/consensus/Validation.d
```diff
@@ -310,8 +310,11 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
     if (!block.txs.isSorted())
         return "Block: Transactions are not sorted";
 
-    if (block.txs.any!(tx => !tx.isValid(findUTXO)))
-        return "Block: Some transactions are invalid";
+    foreach (tx; block.txs)
```
```suggestion
    foreach (const ref tx; block.txs) 8326 
```

AndrejMitrovic:
Fixed.

Geod24 (approved):

Issue #203, opened 2019-08-19
Implement a TCP, binary-based interconnection protocol


bpalaggi:
As a developer, I need to be able to use a protobuffer generator that uses simple TCP (UDP?) connections and protobuff for arguments in order to have efficient communication between nodes.

**Definition of Done:**
- [ ] The generator has a way to listen to a port and do "routing"
- [ ] "Routing" is the act of directing a message to one of the interface's method according to the first byte (or some other data, open to discussion)
- [ ] Once the route has been chosen, the arguments are deserialized according to the method chosen
- [ ] The act of routing does not allocate
- [ ] Routing is nothrow
- [ ] There is a client class that can be instantiated with an `API` as template parameter and an address / port as parameter
- [ ] The client class has support for timeouts
- [ ] The client class can perform asynchronously (either by itself or through `runTask`)
- [ ] Both client and server code are critical for performance, hence the code is performant.
- [ ] There is support to reject inbound peers (agora #202 )

Geod24:
Since I wrote LocalRest, which is very similar, I'll assign this one to myself.

Geod24:
After #2520 is there something that is left in this issue ? #2526 is an optimization (which we want), but I think all the features are here @omerfirmak ?

omerfirmak:
Yeah, all I can think of are optimizations. Closing this.

PR #340, opened 2019-09-19
Fix parsing non-native / non string type in Config

Geod24:
```
The code wasn't doing a conversion, and would throw for e.g. LogLevel.
Unfortunately, the use of 'opt' meant such an Exception was swallowed,
and the default value was always used instead.
```

Found while working on #203


Geod24:
Also added a fix to make CircleCI more reliable, thanks @AndrejMitrovic for pointing it out.

AndrejMitrovic (approved):

PR #342, opened 2019-09-19
Provide default implementation for [de]serialization

Geod24:
This should simplify a bunch of code, as many of the structures
we are dealing with should be PoD.

### DIFF
--- a/source/agora/common/Serializer.d
+++ b/source/agora/common/Serializer.d
@@ -18,6 +18,43 @@ import agora.common.Data;
 /// Type of delegate SerializeDg
 public alias SerializeDg = void delegate(scope const(ubyte)[]) @safe;
 
+/// Default serializer implementation
+public mixin template DefaultSerializer ()
+{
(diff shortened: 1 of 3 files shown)


source/agora/common/Serializer.d
```diff
@@ -18,6 +18,43 @@ import agora.common.Data;
...
+        Serialize all members of this `struct`, in the order they appear.
+
+        Params:
+            dg = serialize function accumulator
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) const @safe
+    {
+        foreach (ref entry; this.tupleof)
```
AndrejMitrovic: `const ref`?

AndrejMitrovic:
Yep this is nice.

source/agora/common/Serializer.d
Geod24: It's `const` by default because it's part of `this`, but yeah

AndrejMitrovic (approved):

PR #344, opened 2019-09-20
Auto-implement interface methods for unittested class

AndrejMitrovic:
This ensures that if there are any new APIs added, the unittest class won't break.

### DIFF
--- a/source/agora/cli/SendTxProcess.d
+++ b/source/agora/cli/SendTxProcess.d
@@ -217,64 +217,24 @@ public int sendTxProcess (string[] args, ref string[] outputs,
 /// Test of send transaction
 unittest
 {
-    class TestCLINode : API
+    // BlackHole auto-implements interface methods and returns typeof(return).init
+    import std.typecons;
+    class TestCLINode : BlackHole!API
     {
+        @safe:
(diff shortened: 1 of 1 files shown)


source/agora/cli/SendTxProcess.d
```diff
@@ -217,65 +217,24 @@ public int sendTxProcess (string[] args, ref string[] outputs,
 /// Test of send transaction
 unittest
 {
-    class TestCLINode : API
+    import std.typecons;
+    class TestCLINode : BlackHole!API
```
Geod24, line 222:
```suggestion
    // Any non-explicitly implement API function will do nothing and return `typeof(return).init`
    class TestCLINode : BlackHole!API
```

Just to make it clear what happens, `BlackHole` is not the most popular function around here.

AndrejMitrovic: Unfortunately `WhiteHole` is broken: https://issues.dlang.org/show_bug.cgi?id=20232 ;)

AndrejMitrovic:
Added comment.

Geod24 (approved):

PR #348, opened 2019-09-20
Remove support for serializing signed integers [for VarInt support]

TrustHenry:
Our Serializer and Deserializer are used only by unsigned Integers. 
So delete unused int,long and Modifying the int used in unittest to uint

### DIFF
--- a/source/agora/common/Deserializer.d
+++ b/source/agora/common/Deserializer.d
@@ -56,7 +56,7 @@ unittest
     // test of various field types
     static struct S
     {
-        int i;
+        uint i;
         string s;
 
         void serialize (scope SerializeDg dg) const @safe
@@ -192,13 +192,6 @@ public void deserializePart (ref ushort record, scope DeserializeDg dg)
 }
 
 /// Ditto
-public void deserializePart (ref int record, scope DeserializeDg dg)
-    @trusted
-{
-    record = *cast(int*)(dg(record.sizeof).ptr);
-}
-
-/// Ditto
 public void deserializePart (ref uint record, scope DeserializeDg dg)
     @trusted
 {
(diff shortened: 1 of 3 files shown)


source/agora/common/Deserializer.d
```diff
@@ -103,7 +103,7 @@ public void deserializePart (T)(ref T record, scope DeserializeDg dg)
 {
     import std.traits;
     OriginalType!T orig_val;
-    deserializePart(orig_val, dg);
+    deserializePart(cast(OriginalType!T) orig_val, dg);
```
AndrejMitrovic: `orig_val` is already typed as `OriginalType!T`.

TrustHenry: Error: none of the overloads of deserializePart are callable using argument types (int, ubyte[] delegate(ulong size) @safe), candidates are:
Is there a way to solve this error?

Geod24: Don't use `OriginalType`. Pass the member directly.

AndrejMitrovic: Explained IRL.

So the actual issue was that `NetworkState` was another enum but it defaulted to `int`. So I suggested to make it `NetworkState : ubyte` in a separate commit, and getting rid of this cast here.

TrustHenry:
ready for review.

AndrejMitrovic:
I've renamed the PR.

AndrejMitrovic:
> ready for review.

I don't know about that. All the test systems failed.

MichaelKim20:
Signed element does not need to be treated with VarInt. Why do you have to delete it?

TrustHenry:
> > ready for review.
> 
> I don't know about that. All the test systems failed.

Fixed it

Geod24 (approved):

PR #351, opened 2019-09-23
Integrate the SCP nomination protocol

AndrejMitrovic:
This requires ~https://github.com/bpfkorea/agora/pull/364~.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -1,0 +1,430 @@
+/*******************************************************************************
+
+    Contains the SCP consensus driver implementation.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.protocol.Nominator;
+
+import agora.common.crypto.Key;
+import agora.common.Deserializer;
+import agora.common.Hash : hashFull;
+import agora.common.Serializer;
+import agora.common.Set;
+import agora.common.Task;
+import agora.consensus.data.Block;
+import agora.consensus.data.Transaction;
+import agora.network.NetworkClient;
+import agora.node.Ledger;
+import agora.utils.Log;
+import agora.utils.PrettyPrinter;
+
+import scpd.Cpp;
+import scpd.scp.SCP;
+import scpd.scp.SCPDriver;
+import scpd.scp.Utils;
+import scpd.types.Stellar_types;
+import scpd.types.Stellar_types : StellarHash = Hash;
+import scpd.types.Stellar_SCP;
+import scpd.types.Utils;
+import scpd.Util;
+
+import core.stdc.stdint;
+import core.time;
+
+mixin AddLogger!();
+
+/// Ditto
+public extern (C++) class Nominator : SCPDriver
+{
+    /// SCP instance
+    private SCP* scp;
+
+    /// Key pair of this node
+    private KeyPair key_pair;
+
+    /// Task manager
+    private TaskManager taskman;
+
+    /// Ledger instance
+    private Ledger ledger;
+
+    /// This node's quorum node clients
+    private NetworkClient[PublicKey] peers;
+
+    /// The set of active timers
+    /// Todo: SCPTests.cpp uses fake timers,
+    /// Similar to how we use FakeClockBanManager!
+    private Set!ulong timers;
+
+    /// The set of externalized slot indices
+    private Set!uint64_t externalized_slots;
+
+    /// The quorum set
+    private SCPQuorumSetPtr[Hash] quorum_set;
+
+extern(D):
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            key_pair = the key pair of this node
+            ledger = needed for SCP state restoration & block validation
+            taskman = used to run timers
+            peers = the set of clients to the peers in the quorum
+            quorum_set = the quorum set of this node
+
+    ***************************************************************************/
+
+    public this (KeyPair key_pair, Ledger ledger,
+        TaskManager taskman, NetworkClient[PublicKey] peers,
+        SCPQuorumSet quorum_set)
+    {
+        this.key_pair = key_pair;
+        auto node_id = NodeID(StellarHash(key_pair.address[]));
+        const IsValidator = true;
+        this.scp = createSCP(this, node_id, IsValidator, quorum_set);
+        this.taskman = taskman;
+        this.ledger = ledger;
+        this.peers = peers;
+
+        // cast: see makeSharedSCPQuorumSet() in Cpp.d
+        auto local_quorum_set = this.scp.getLocalQuorumSet();
+        auto localQSet = makeSharedSCPQuorumSet(local_quorum_set);
+
+        const bytes = ByteSlice.make(XDRToOpaque(*localQSet));
+        auto quorum_hash = sha256(bytes);
+        this.quorum_set[quorum_hash] = localQSet;
+
+        this.restoreSCPState(ledger);
+        this.ledger.setNominator(&this.nominateTransactionSet);
+    }
+
+    /***************************************************************************
+
+        Nominate a new transaction set to the quorum.
+        Failure to nominate is only logged.
+
+        Params:
+            slot_idx = the index of the slot to nominate for
+            prev = the transaction set of the previous slot
+            next = the proposed transaction set for the provided slot index
+
+    ***************************************************************************/
+
+    private void nominateTransactionSet (ulong slot_idx, Set!Transaction prev,
+        Set!Transaction next) @trusted
+    {
+        log.info("{}(): Proposing tx set for slot {}", __FUNCTION__, slot_idx);
+
+        auto prev_value = prev.serializeFull().toVec();
+        auto next_value = next.serializeFull().toVec();
+        if (this.scp.nominate(slot_idx, next_value, prev_value))
+        {
+            log.info("{}(): Tx set nominated", __FUNCTION__);
+        }
+        else
+        {
+            log.info("{}(): Tx set rejected nomination", __FUNCTION__);
+        }
+    }
+
+    /***************************************************************************
+
+        Restore SCP's internal state based on the provided ledger state
+
+        Params:
+            ledger = the ledger instance
+
+    ***************************************************************************/
+
+    private void restoreSCPState (Ledger ledger)
+    {
+        import agora.common.Serializer;
+        import scpd.types.Stellar_SCP;
+        import scpd.types.Utils;
+        import scpd.types.Stellar_types : StellarHash = Hash, NodeID;
+        import std.range;
+
+        auto key = StellarHash(this.key_pair.address[]);
+        auto pub_key = NodeID(key);
+
+        foreach (block_idx, block; ledger.getBlocksFrom(0).enumerate)
+        {
+            Value block_value = block.serializeFull().toVec();
+
+            SCPStatement statement =
+            {
+                nodeID: pub_key,
+                slotIndex: block_idx,
+                pledges: {
+                    type_: SCPStatementType.SCP_ST_EXTERNALIZE,
+                    externalize_: {
+                        commit: {
+                            counter: 0,
+                            value: block_value,
+                        },
+                        nH: 0,
+                    },
+                },
+            };
+
+            SCPEnvelope env = SCPEnvelope(statement);
+            this.scp.setStateFromEnvelope(block_idx, env);
+            if (!this.scp.isSlotFullyValidated(block_idx))
+                assert(0);
+        }
+
+        // there should at least be a genesis block
+        if (this.scp.empty())
+            assert(0);
+    }
+
+    /***************************************************************************
+
+        Called when a new SCP Envelope is received from the network.
+
+        Params:
+            envelope = the SCP envelope
+
+        Returns:
+            true if the SCP protocol accepted this envelope
+
+    ***************************************************************************/
+
+    public bool receiveEnvelope (SCPEnvelope envelope) @trusted
+    {
+        return this.scp.receiveEnvelope(envelope) == SCP.EnvelopeState.VALID;
+    }
+
+    extern (C++):
+
+
+    /***************************************************************************
+
+        Signs the SCPEnvelope with the node's private key.
+
+        todo: Currently not signing yet. To be done.
+
+        Params:
+            envelope = the SCPEnvelope to sign
+
+    ***************************************************************************/
+
+    public override void signEnvelope (ref SCPEnvelope envelope)
+    {
+    }
+
+    /***************************************************************************
+
+        Validates the provided transaction set for the provided slot index,
+        and returns a status code of the validation.
+
+        Params:
+            slot_idx = the slot index we're currently reaching consensus for
+            value = the transaction set to validate
+            nomination = unused, seems to be stellar-specific
+
+    ***************************************************************************/
+
+    public override ValidationLevel validateValue (uint64_t slot_idx,
+        ref const(Value) value, bool nomination) nothrow
+    {
+        scope (failure) assert(0);
+
+        try
+        {
+            auto tx_set = deserializeFull!(Set!Transaction)(
+                cast(ubyte[])value[]);
+
+            if (auto fail_reason = this.ledger.validateTxSet(tx_set))
+            {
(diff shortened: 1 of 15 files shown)


AndrejMitrovic:
I'll split up the 3 non-controversial commits into another PR as they're ready to go.

AndrejMitrovic:
The `Nominator` class doesn't have to exist, it can be rolled into `Driver`. Initially I split them up thinking that the nominating part and the driver part (which actually sends messages) would benefit from being separate, but there's not much additional functionality the `Nominator` has.

AndrejMitrovic:
Updated with latest fixes. The test-suite passes now (if the LocalRest change is included).

The big SCP commit needs to be further broken down. And I'll focus on writing additional tests, probably porting some test-cases from stellar-core.

AndrejMitrovic:
I've got a race-condition in my code. The following triggers a segfault, as it tries to query the database after it's been shut down:

```diff
diff --git a/source/agora/consensus/data/UTXOSet.d b/source/agora/consensus/data/UTXOSet.d
index 8c8b06d9..de5aeb94 100644
--- a/source/agora/consensus/data/UTXOSet.d
+++ b/source/agora/consensus/data/UTXOSet.d
@@ -105,6 +105,7 @@ public class UTXOSet
     public void shutdown ()
     {
         this.utxo_db.shutdown();
+        this.utxo_db = null;
     }

     /***************************************************************************
diff --git a/source/agora/utils/Log.d b/source/agora/utils/Log.d
index 5b043acc..5c6e2281 100644
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -42,12 +42,8 @@ public template AddLogger (string moduleName = __MODULE__)
 /// Convenience alias
 public alias LogLevel = Level;

-version (unittest) {}
-else
+/// Initialize the logger
+static this ()
 {
-    /// Initialize the logger
-    static this ()
-    {
-        Log.defaultConfig();
-    }
+    Log.defaultConfig();
 }
```

AndrejMitrovic:
Updated: Now we are nominating transaction sets, rather than whole blocks. This will allow implementing tx set merging rules in the future, for example: take first 8 sorted txs out of the union of all the proposed tx sets.

AndrejMitrovic:
This code includes https://github.com/bpfkorea/agora/pull/364, but I'm thinking of a way to avoid this. Maybe by implementing `setupTimer()` on the C++ side and avoiding the need to change the function signatures in https://github.com/bpfkorea/agora/pull/364/files.

Geod24:
```
Linking...
.dub/build/unittest-unittest-cov-linux.posix-x86_64-dmd_2088-60157B2BF871D74430F234139CD56BE4/agora-unittests.o: In function `_D5agora9consensus8protocol9NominatorQk22nominateTransactionSetMFNemSQCp6common3Set__TQhTSQDkQDh4data11TransactionQnZQBmQCaZv':
/home/travis/build/bpfkorea/agora/source/agora/consensus/protocol/Nominator.d:130: undefined reference to `stellar::SCP::nominate(unsigned long, xdr::xvector<unsigned char, 4294967292u> const&, xdr::xvector<unsigned char, 4294967292u> const)'
.dub/build/unittest-unittest-cov-linux.posix-x86_64-dmd_2088-60157B2BF871D74430F234139CD56BE4/agora-unittests.o:(.data._D5agora9consensus8protocol9NominatorQk6__vtblZ+0x58): undefined reference to `stellar::SCPDriver::computeValueHash(unsigned long, xdr::xvector<unsigned char, 4294967292u> const&, int, xdr::xvector<unsigned char, 4294967292u> const)'
```

AndrejMitrovic:
2.088.1 linker failure due to https://issues.dlang.org/show_bug.cgi?id=20336

We can't revert to 2.088.0 because it also has a mangling problem which 2.088.1 solved.

AndrejMitrovic:
I just hard a random segfault (DMD) and test failure (LDC) locally, and it doesn't always reproduce. I have to debug this.

AndrejMitrovic:
@Geod24 https://github.com/dlang/dmd/pull/10477 fixed the linking issue, but when are they going to release 2.088.2 with the fix?

Geod24:
It's going to be part of 2.089.1, as it missed the deadline for 2.089.0 ... https://github.com/dlang/dmd/pull/10527#issuecomment-549096588

AndrejMitrovic:
Resolved conflicts. It will still fail due to linking issues that will hopefully be resolved soon once the new DMD comes out.

Geod24:
Rebase ?

AndrejMitrovic:
> Rebase ?

Done. But it will still fail to link on ldc until they make a new release based on the bugfixes in 2.089.1.

Geod24:
So we need to drop LDC 1.18.0 for this to move forward.
In order to do that, we need to update the LDC package on Alpine linux.
On it: https://gitlab.alpinelinux.org/alpine/aports/merge_requests/2703

Geod24:
The fix to the package is done, we can drop LDC 1.18.0 from the CI

Geod24:
https://github.com/bpfkorea/agora/pull/463

AndrejMitrovic:
Rebased.

AndrejMitrovic:
There is a random stack overflow on LDC, restarting the build. SCP does use more stack space, so it might be that it's caused by the same thing as https://github.com/dlang/dub/issues/1812#issuecomment-568981831

source/agora/consensus/protocol/Nominator.d
```diff
@@ -0,0 +1,373 @@
...
+            if (timerID !in this.timers)
+                return;
+
+            callCPPDelegate(cb);
+        });
+    }
+}
+
+/// workaround for logger not being nothrow
+private void logCall (T...)(lazy T args, string func = __FUNCTION__) nothrow
```
Geod24: Can we make a `nothrow` wrapper in Agora an call it a day ?

AndrejMitrovic: Like a wrapper around `Logger` with an `alias logger this` inside?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -0,0 +1,373 @@
...
+
+        this.timers.put(timerID);
+
+        this.taskman.runTask(
+        {
+            this.taskman.wait(timeout.msecs);
+            if (timerID !in this.timers)
+                return;
+
+            callCPPDelegate(cb);
```
Geod24: ~You never remove the delegate from `this.timers` though, or did I miss something ?~

So this function is called both to setup and remove timers, it seems. Could you document it a bit ?

source/scpd/Cpp.d
```diff
@@ -73,6 +74,10 @@ extern(C++, (StdNS!())) {
 /// C++ support for foreach
 extern(C++) private int cpp_set_foreach(T)(void* set, void* ctx, void* cb);
 
+// todo fix: SCPQuorumSet should be const, but triggers a wild vibe.d error
```
Geod24: Can you add more details about the error ?

source/scpd/Cpp.d
```diff
@@ -27,6 +27,7 @@
 
 module scpd.Cpp;
 
+import scpd.types.Stellar_SCP;
```
Geod24: This should actually be in `scpd.Utils`.
The point of this module is that it provides "C++" wrapper, as in, not specific to Stellar, and that ultimately we'll be able to remove it.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -0,0 +1,373 @@
...
+            quorum_set = the quorum set of this node
+
+    ***************************************************************************/
+
+    public this (KeyPair key_pair, Ledger ledger,
+        TaskManager taskman, NetworkClient[PublicKey] peers,
+        SCPQuorumSet quorum_set)
+    {
+        this.key_pair = key_pair;
+        import scpd.types.Stellar_types : StellarHash = Hash;
```
Geod24: This also exists at module scope

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: https://github.com/bpfkorea/agora/pull/464

AndrejMitrovic:
> Can we make a nothrow wrapper in Agora an call it a day ?

Are you still reviewing? =)

It's also green, for the first time ever. Yay.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -0,0 +1,373 @@
...
+
+        assert(0);  // should not reach here
+    }
+
+    ///
+    public override void setupTimer (ulong slotIndex, int timerID,
+        milliseconds timeout, CPPDelegate!SCPCallback* cb)
+    {
+        scope (failure) assert(0);
+        logCall("{} {}", this.key_pair.address.prettify);
```
Geod24: That's just disgusting

source/agora/node/Ledger.d
```diff
@@ -91,6 +99,43 @@ public class Ledger
...
+    /***************************************************************************
+
+        Set the transaction set nominator callback.
+
+        Params:
+            nominateTxSet = the nominator callback
+
+    ***************************************************************************/
+
+    public void setNominator ( NominatorDg nominateTxSet ) @safe
```
Geod24: Extra whitespace after parenthesis

source/agora/node/Ledger.d
```diff
@@ -49,6 +54,9 @@ public class Ledger
     /// UTXO set
     private UTXOSet utxo_set;
 
+    /// Callback when a transaction set is ready to be nominated
+    public NominatorDg nominateTxSet;
```
Geod24: Why not just make it a method and pass its address when needing a delegate ?

source/agora/node/Node.d
```diff
@@ -107,10 +114,47 @@ public class Node : API
...
     public void start ()
     {
+        import scpd.scp.QuorumSetUtils;
         log.info("Doing network discovery..");
-        this.network.discover();
-
+        auto peers = this.network.discover();
         this.network.retrieveLatestBlocks(this.ledger);
+
+        if (this.config.node.is_validator)
```
Geod24:
```D
if (!this.config.node.is_validator)
    return;
```

source/agora/test/Base.d
```diff
@@ -485,9 +486,12 @@ public class TestNode : Node, TestAPI
 /// Describes a network topology for testing purpose
 public enum NetworkTopology
 {
-    /// 4 nodes which all know about each other. Figure 9 in the SCP paper.
+    /// A number nodes which all know about each other. Figure 9 in the SCP paper.
```
Geod24: A number of... ?

source/agora/test/Base.d
```diff
@@ -559,55 +564,98 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
...
+    case NetworkTopology.Cyclic:
+        node_configs = iota(nodes).map!(_ => makeNodeConfig()).array;
+        node_configs.each!((ref conf) => conf.min_listeners = 1);
+        configs = iota(nodes)
+            .map!(idx => makeCyclicConfig(idx, node_configs[idx], node_configs))
+                .array;
 
+        break;
+
+    case NetworkTopology.SinglePoF:
```
Geod24: Why the move ?

Geod24 (commented):
First round of review complete. I need a second round for sure though. In the morning. After a coffee. ~And some speed~

source/agora/test/Base.d
AndrejMitrovic: The diff seems to make it harder to see, but previously we only created nodes for the `NetworkTopology.Simple` config, now we create configs for `NetworkTopology.Simple`, `NetworkTopology.OneNonValidator`, and `NetworkTopology.Cyclic`, and we move the instantiation of the nodes after the configs were built (so we don't duplicate node creation code).

source/scpd/Cpp.d
AndrejMitrovic: It seems to be gone, and now const works. I don't remember what the error was. :/

source/agora/node/Ledger.d
AndrejMitrovic: This was accidentally public, it should be private. I'll fix it.

AndrejMitrovic:
Rebased, fixed comments, documented, and removed the `logCall` stuff too.

source/scpd/types/Utils.d
```diff
@@ -18,6 +18,10 @@ import scpd.types.Stellar_SCP;
 import scpd.types.Stellar_types;
 import scpd.types.XDRBase;
 
+// todo fix: SCPQuorumSet should be const, but triggers a wild vibe.d error
```
Geod24: What's the error again ? Is it RT or CT ?

source/agora/node/Ledger.d
Geod24: I mean it could still be a method instead of a delegate no?

source/agora/test/Base.d
Geod24: I meant that L608 (`case NetworkTopology.SinglePoF:`) has just been reordered and is still doing the same (assert(0)), but not a big deal.

source/scpd/types/Utils.d
```diff
@@ -18,9 +18,8 @@ import scpd.types.Stellar_SCP;
 import scpd.types.Stellar_types;
 import scpd.types.XDRBase;
 
-// todo fix: SCPQuorumSet should be const, but triggers a wild vibe.d error
 extern(C++) public shared_ptr!SCPQuorumSet makeSharedSCPQuorumSet (
-    ref SCPQuorumSet);
+    ref const(SCPQuorumSet));
```
Geod24, line 22: Seems to belong to a previous commit

source/agora/test/GossipProtocol.d
```diff
@@ -51,22 +51,29 @@ unittest
...
-        2.seconds,
-        format("Node %s has block height %s. Expected: %s",
-        idx,
-        node.getBlockHeight().to!string,
-        1)));
+    version (none)
+    {
+        // todo: race condition: when the 8th transaction is added to one node,
+        // it begins the nomination protocol before the transaction had the
+        // chance to be propagated to other nodes
```
Geod24: That's a pretty big bug heh

Geod24:
Looks good except for the `version (none)`. What's the plan for it ?

source/agora/test/GossipProtocol.d
AndrejMitrovic: Yeah so, Stellar nomination is based on a timer, but we don't want to physically wait 10 minutes in the **test-suite** before each nomination round starts. So I've temporarily implemented it in a way that the nomination starts as soon as there's enough transactions in the pool. 

This is not ideal. We probably want to use, ehh, a virtual timer of sorts instead? :>

source/scpd/types/Utils.d
AndrejMitrovic: The issue is I can't reproduce it anymore. /shrug

source/agora/test/Base.d
AndrejMitrovic: Oh good eye! I'll fix that up.

AndrejMitrovic:
> Looks good except for the version (none). What's the plan for it ?

I just fixed the "timing issue" bug now. But the BanManager test is still versioned out. I think I can fix it too, today.

AndrejMitrovic:
Fixed the BanManager test. I just added a non-validator node and tested banning behavior with that one.

AndrejMitrovic:
Fixed a bug just now. I used `%s` by mistake and it threw an exception.

Geod24:
Needs a rebase now

AndrejMitrovic:
@Geod24 ready to go.

source/agora/test/BanManager.d
```diff
@@ -25,30 +25,34 @@ import agora.consensus.Genesis;
 import agora.test.Base;
 
 /// test node banning after putTransaction fails a number of times
+/// note: temporarily disabled as node banning needs to be rethinked with regards
+/// to the node being in another node's quorum set
```
Geod24: Not valid anymore

source/agora/test/GossipProtocol.d
```diff
@@ -51,6 +51,7 @@ unittest
     auto send_txs = txs[0..$-1];
     // send it to tx to node
     send_txs.each!(tx => node_1.putTransaction(tx));
+
```
Geod24: ?

Geod24 (approved):
LGTM just 2 nits

AndrejMitrovic:
Yep, sorry for missing these. Fixed now.

Geod24:
💣💥

PR #352, opened 2019-09-23
Make getBlocksFrom() in Ledger return a range

AndrejMitrovic:
This makes it easier to optimize & cache the result at the call site (which is not done yet, but it enables it).

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -28,6 +28,7 @@ import agora.utils.Log;
 import agora.utils.PrettyPrinter;
 
 import std.algorithm;
+import std.range;
 
 mixin AddLogger!();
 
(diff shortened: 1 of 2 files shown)


source/agora/node/Ledger.d
```diff
@@ -283,37 +284,30 @@ public class Ledger
...
         Returns:
-            the array of blocks starting from block_height,
-            up to `max_blocks`
+            the range of blocks starting from start_height
 
     ***************************************************************************/
 
-    public const(Block)[] getBlocksFrom (ulong block_height, size_t max_blocks)
-        @safe
+    public auto getBlocksFrom ( size_t start_height ) @safe nothrow
```
```suggestion
    public auto getBlocksFrom (size_t start_height) @safe nothrow
```

MichaelKim20 (commented):
This PR is expected to have a lot of effect on performance.

AndrejMitrovic:
Fixed styling issue.

> This PR is expected to have a lot of effect on performance.

Not yet, but it will enable caching in the future. I will need this code for SCP though, to restore the SCP state when a validator node boots up.

MichaelKim20 (approved):

Geod24 (approved):
Neat

Issue #301, opened 2019-09-05
Make sure double-spend transactions are not added to the Pool


AndrejMitrovic:
## Current behavior

A double-spend transaction is a transaction which tries to spend the same unspent `Output` which another transaction also spends.

In the implementation of full block verification [in this PR](https://github.com/bpfkorea/agora/pull/193), the filtering of double-spend transactions was done in the Ledger class itself, [as seen here](https://github.com/bpfkorea/agora/blob/095cde407384d2afa9482162c240bb989af2dfe0/source/agora/node/Ledger.d#L217).

## Problem

Because the filtering of double-spends is done at the Ledger level **after** the transactions are added to the TransactionPool, this means the TransactionPool will always keep the double-spend transactions in its storage. This enables a form of DDoS, as the pool may end up growing indefinitely with double-spend transactions.

## Wanted behavior

Filter out double-spend transactions before they are added to the TransactionPool.

linked0:
My implementations will be as follows
1. The checking code will be in the acceptTransaction function of Ledger class. The transaction parameter of the function will be compared against each Transaction object of the Transaction pool, by checking values of both hashes and those of both indices of the ’inputs’ member variables of the both Transaction objects. If The transaction parameter has same hash and index equal to those values of any transaction in the pool, It will be discarded and the rest of the function will not be executed.
2. I will use the unittest code having the comment “test behavior of receiving double-spend transaction” already written in agora/test/Ledger.d

Any comment would be helpful.

Geod24:
Comparing the transaction hash won't work, because there can be an infinity of transactions that spends the same input.

AndrejMitrovic:
There is way to do this in a more optimal way:

The UTXO set can in addition to tracking UTXOs also track referenced UTXOs by the transactions in the pool. `findUTXO` would have to be updated, to return false if a transaction tries to spend a UTXO which will already be spent by another transaction in the pool.

AndrejMitrovic:
@Geod24 suggested to update the Pool instead of the UTXO set. That makes more sense, so I would suggest taking that approach if possible.

linked0:
[...]

PR #356, opened 2019-09-25
Implement double-spending checking and write unit test code

linked0:
- Add a member variable map to store Input objects into TransactionPool class.
- Add the isValidTransaction function checking double-spending into TransactionPool class.
- Add the code calling isValidTransaction of the TransactionPool into the acceptTransaction of the Ledger.
- Add unit test code into agora.node.Ledger module.

### DIFF
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -18,6 +18,7 @@ import agora.common.Types;
 import agora.common.Deserializer;
 import agora.common.Hash;
 import agora.common.Serializer;
+import agora.common.Set;
 import agora.consensus.data.Transaction;
 import agora.utils.Log;
 
@@ -56,6 +57,8 @@ public class TransactionPool
     /// SQLite db instance
     private Database db;
 
+    /// Map for hashes for Input objects of each transaction in TX pool
+    private Set!Hash input_set;
 
     /***************************************************************************
 
@@ -80,6 +83,17 @@ public class TransactionPool
         // create the table if it doesn't exist yet
         this.db.execute("CREATE TABLE IF NOT EXISTS tx_pool " ~
             "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
+
+        // insert Input objects to the set of Input hashes from the transactions stored in DB.
+        auto results = this.db.execute("SELECT key, val FROM tx_pool");
+
+        foreach (row; results)
+        {
+            auto tx = deserializeFull!Transaction(row.peek!(ubyte[])(1));
+
+            foreach (input; tx.inputs)
+                input_set.put(input.hashFull());
+        }
     }
 
     /***************************************************************************
@@ -127,15 +141,24 @@ public class TransactionPool
         Params:
             tx = the transaction to add
 
-        Throws:
-            SqliteException if the transaction failed to be added
+        Returns:
+            true if the transaction has been added to the pool
 
     ***************************************************************************/
 
-    public void add (Transaction tx) @safe
+    public bool add (Transaction tx) @safe
     {
         static ubyte[] buffer;
         buffer.length = 0;
+
+        // check double-spend
+        if (!isValidTransaction(tx))
+            return false;
+
+        // insert each input information of the transaction
+        foreach (input; tx.inputs)
+            this.input_set.put(input.hashFull());
+
         () @trusted { assumeSafeAppend(buffer); }();
 
         scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
@@ -149,6 +172,8 @@ public class TransactionPool
             db.execute("INSERT INTO tx_pool (key, val) VALUES (?, ?)",
                 hashFull(tx)[], buffer);
         }();
+
+        return true;
     }
 
     /***************************************************************************
@@ -201,12 +226,17 @@ public class TransactionPool
         Remove the transaction with the given key from the pool
 
         Params:
-            hash = the hash of the transaction
+            tx = the transaction to remove
 
     ***************************************************************************/
 
-    public void remove (Hash hash) @trusted
+    public void remove (const ref Transaction tx) @trusted
     {
+        // delete inputs of transaction from the set of Input hashes
+        foreach (ref input; tx.inputs)
+            this.input_set.remove(input.hashFull());
+
+        auto hash = tx.hashFull();
         this.db.execute("DELETE FROM tx_pool WHERE key = ?", hash[]);
     }
 
@@ -226,20 +256,50 @@ public class TransactionPool
     {
         const len_prev = this.length();
         assert(len_prev >= count);
-        Hash[] hashes;
         Transaction[] txs;
 
         foreach (hash, tx; this)
         {
-            hashes ~= hash;
             txs ~= tx;
             if (txs.length == count)
                 break;
         }
 
-        hashes.each!(hash => this.remove(hash));
+        txs.each!(tx => this.remove(tx));
         assert(this.length() == len_prev - count);
         return txs;
+    }
+
+    /***************************************************************************
+
+        Check if the input transaction has any double spending input.
+
+        Params:
+            tx = the transaction to validate
+
+        Returns:
+            true if the transaction has no double spending input.
+
+    ***************************************************************************/
+
+    public bool isValidTransaction (const ref Transaction tx) @trusted
+    {
+        auto txHash = tx.hashFull();
+
+        auto results = this.db.execute("SELECT key, val FROM tx_pool " ~
+            "WHERE key = ?", txHash[]);
+
+        if (!results.empty)
+            return false;
+
+        foreach (input; tx.inputs)
+        {
+            auto hash = input.hashFull();
+            if (hash in this.input_set)
+                return false;
+        }
+
+        return true;
     }
 }
 
@@ -269,9 +329,9 @@ unittest
     assert(half_txs.length == txs.length / 2);
     assert(pool.length == txs.length / 2);
 
-    // adding duplicate tx hash => exception throw
+    // adding duplicate tx hash => return false
     pool.add(txs[0]);
-    assertThrown!SqliteException(pool.add(txs[0]));
+    assert(!pool.add(txs[0]));
 }
 
 /// memory reclamation tests
@@ -317,4 +377,59 @@ unittest
     auto pool_txs = pool.take(txs.length);
     assert(pool.length == 0);
     assert(txs == pool_txs);
+}
+
+/// test double-spending on the Transaction pool
+unittest
+{
+    import agora.common.Amount;
+    import agora.common.crypto.Key;
+    import agora.common.Hash;
+
+    auto seed1 = "SCFPAX2KQEMBHCG6SJ77YTHVOYKUVHEFDROVFCKTZUG7Z6Q5IKSNG6NQ";
+    auto seed2 = "SCTTRCMT7DVZHQS375GWIKYQYHKA3X4IC4EOBNPRGV7DFR3X6OM5VIWL";
+    auto seed3 = "SAI4SRN2U6UQ32FXNYZSXA5OIO6BYTJMBFHJKX774IGS2RHQ7DOEW5SJ";
+    auto key_pair1 = KeyPair.fromSeed(Seed.fromString(seed1));
+    auto key_pair2 = KeyPair.fromSeed(Seed.fromString(seed2));
+    auto key_pair3 = KeyPair.fromSeed(Seed.fromString(seed3));
+
+    // create first transaction pool
+    auto pool = new TransactionPool(":memory:");
+    scope(exit) pool.shutdown();
+
+    // create first transaction
+    Transaction tx1 =
+    {
+        TxType.Payment,
+        [Input(Hash.init, 0)],
+        [Output(Amount(0), key_pair2.address)]
+    };
+    auto signature = key_pair1.secret.sign(hashFull(tx1)[]);
+    tx1.inputs[0].signature = signature;
+
+    // create second transaction
+    Transaction tx2 =
+    {
+        TxType.Payment,
+        [Input(Hash.init, 0)],
+        [Output(Amount(0), key_pair3.address)]
+    };
+    signature = key_pair1.secret.sign(hashFull(tx2)[]);
+    tx2.inputs[0].signature = signature;
+
+    // add first tx to the pool
+    assert(pool.add(tx1));
+
+    // add duplicate tx => return false
+    assert(!pool.add(tx2));
+
+    // create second transaction pool
+    auto pool2 = new TransactionPool(":memory:");
+    scope(exit) pool2.shutdown();
+
+    // add duplicate tx into second pool => return true
+    assert(pool2.add(tx2));
+
+    // add first tx => return false
+    assert(!pool2.add(tx1));
 }
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -134,13 +134,12 @@ public class Ledger
 
     public bool acceptTransaction (Transaction tx) @safe
     {
-        if (!this.isValidTransaction(tx))
+        if (!this.isValidTransaction(tx) || !this.pool.add(tx))
         {
             log.info("Rejected tx: {}", tx);
             return false;
         }
 
-        this.pool.add(tx);
         if (this.pool.length >= Block.TxsInBlock)
             this.tryCreateBlock();
 
@@ -188,7 +187,7 @@ public class Ledger
         block.txs.each!(tx => this.utxo_set.updateUTXOCache(tx, height));
 
         // remove the TXs from the Pool
-        block.txs.each!(tx => this.pool.remove(tx.hashFull()));
+        block.txs.each!(tx => this.pool.remove(tx));
     }
 
     /***************************************************************************


source/agora/common/TransactionPool.d
```diff
@@ -241,6 +276,30 @@ public class TransactionPool
...
+            true if the transaction has no double spending input.
+
+    ***************************************************************************/
+
+    public bool isValidTransaction (Transaction tx) @trusted
+    {
+        foreach (input; tx.inputs)
+        {
+            auto hash = hashMulti(input.previous, input.index);
+            if(hash in input_map)
```
TrustHenry: `if (hash in input_map)`

source/agora/common/TransactionPool.d
```diff
@@ -149,6 +165,12 @@ public class TransactionPool
             db.execute("INSERT INTO tx_pool (key, val) VALUES (?, ?)",
                 hashFull(tx)[], buffer);
         }();
+
+        // insert each input information of the transaction
+        foreach(input; tx.inputs) {
```
TrustHenry:
```suggestion
        foreach (input; tx.inputs) {
```

source/agora/common/TransactionPool.d
TrustHenry:
```suggestion
            if (hash in input_map)
```

source/agora/common/TransactionPool.d
```diff
@@ -207,6 +229,19 @@ public class TransactionPool
 
     public void remove (Hash hash) @trusted
     {
+        // delete inputs of transaction from input_map
+        auto results = this.db.execute("SELECT key, val FROM tx_pool " ~
+            "WHERE key = ?", hash[]);
+        foreach(row; results)
+        {
+            auto tx = deserializeFull!Transaction(row.peek!(ubyte[])(1));
+            foreach(input; tx.inputs)
```
TrustHenry:
```suggestion
            foreach (input; tx.inputs)
```

source/agora/node/Ledger.d
```diff
@@ -132,6 +132,12 @@ public class Ledger
 
     public bool acceptTransaction (Transaction tx) @safe
     {
+        if (!this.pool.isValidTransaction(tx))
```
AndrejMitrovic: Could you move this check into the `isValidTransaction` method in the `Ledger`? That way we have a single method where we check transaction validation.

source/agora/common/TransactionPool.d
```diff
@@ -56,6 +56,8 @@ public class TransactionPool
     /// SQLite db instance
     private Database db;
 
+    // map for Input objects of each transaction.
+    private Input[Hash] input_map;
```
AndrejMitrovic: The `Input` itself doesn't seem to be used. So this could be a `Set!Hash` instead. There is a Set implementation in `agora.common.Set`

source/agora/common/TransactionPool.d
```diff
@@ -80,6 +82,20 @@ public class TransactionPool
         // create the table if it doesn't exist yet
         this.db.execute("CREATE TABLE IF NOT EXISTS tx_pool " ~
             "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
+
+        // insert Input objects to input_map from the transactions stored in DB.
```
AndrejMitrovic: We could serialize the map (or Set) to disk and load it on start-up instead of recreating it each time the transaction pool is loaded. Iterating over the entire transaction pool on load can be costly.

source/agora/common/TransactionPool.d
```diff
@@ -149,6 +165,12 @@ public class TransactionPool
             db.execute("INSERT INTO tx_pool (key, val) VALUES (?, ?)",
                 hashFull(tx)[], buffer);
         }();
+
+        // insert each input information of the transaction
+        foreach (input; tx.inputs) {
```
AndrejMitrovic: `{` on its own line. We follow this style guide: https://github.com/bpfkorea/agora/blob/v0.x.x/doc/Style.md

source/agora/common/TransactionPool.d
```diff
@@ -207,6 +229,19 @@ public class TransactionPool
 
     public void remove (Hash hash) @trusted
     {
+        // delete inputs of transaction from input_map
```
AndrejMitrovic: You could change the `remove` method to take the `Transaction` as a parameter, and then pass it as an argument from within `updateUTXOSet` in the Ledger. It avoids a lookup in the pool.

source/agora/common/TransactionPool.d
```diff
@@ -241,6 +276,30 @@ public class TransactionPool
...
+
+        Params:
+            transaction = the transaction to validate
+
+        Returns:
+            true if the transaction has no double spending input.
+
+    ***************************************************************************/
+
+    public bool isValidTransaction (Transaction tx) @trusted
```
AndrejMitrovic: `@safe` instead of `@trusted`. `@safe` means the compiler will verify all the code inside the function only calls other `@safe` functions.

`@trusted` tells the compiler to ignore the function body, and to make `isValidTransaction` appear as if it's `@safe`.

source/agora/node/Ledger.d
```diff
@@ -675,3 +681,57 @@ unittest
...
+
+/// test double-spending on the Transaction pool
+unittest
+{
+    import agora.common.crypto.Key;
+    import agora.common.Deserializer;
+    import agora.common.Serializer;
+    import agora.common.Types;
+    import agora.common.Hash;
+    import std.stdio;
```
AndrejMitrovic: Unused import.

AndrejMitrovic:
I have a suggestion:

Replace the `isValidTransaction` in the `TransactionPool`. Move that check inside of the `add` method, and make the `add` method return a status code (true if the transaction was added).

Currently the abstraction is leaky. If we forget to call `pool.isValidTransaction` before calling `pool.add(tx)`, then we will have added a double-spend transaction by mistake.

AndrejMitrovic:
Also, the unittests should test the TransactionPool, not the Ledger. The Ledger already has tests for double-spend transactions. So I suggest moving the unittest into the `TransactionPool` module, where the only thing being tested is instantiating the Pool and trying to add double-spend transactions, and checking the contents of the transaction pool verifying no double-spend transactions were added.

AndrejMitrovic (changes requested):
Made a few comments.

AndrejMitrovic:
Good work so far!

linked0:
I fully understand all your comments, I'll try.

source/agora/node/Ledger.d
```diff
@@ -675,3 +681,57 @@ unittest
...
+    scope(exit) pool.shutdown();
+    auto utxo_set = new UTXOSet(":memory:");
+    scope (exit) utxo_set.shutdown();
+    scope ledger = new Ledger(pool, utxo_set, storage);
+
+    // third set of transactions
+    auto new_txs = makeChainedTransactions(gen_key, txs, 1);
+
+    // create a deep-copy of the first tx
+    auto backup_tx = deserializeFull!Transaction(serializeFull(new_txs[0]));
```
MichaelKim20: This variable(`backup_tx`) is not used.

source/agora/node/Ledger.d
```diff
@@ -675,3 +681,57 @@ unittest
...
+    new_txs[0].inputs[0].signature = signature;
+
+    // check block height, it must be 2
+    assert(ledger.getBlockHeight() == 2);
+
+    // add 8 transactions containing two same transactions to leger
+    foreach(tx; new_txs)
+    {
+        ledger.acceptTransaction(tx);
+    }
```
```suggestion
    foreach(tx; new_txs)
        ledger.acceptTransaction(tx);
```

AndrejMitrovic:
I already suggested replacing the Ledger unittest with a TransactionPool unittest in https://github.com/bpfkorea/agora/pull/356#issuecomment-534853592

MichaelKim20:
> I already suggested replacing the Ledger unittest with a TransactionPool unittest in [#356 (comment)](https://github.com/bpfkorea/agora/pull/356#issuecomment-534853592)
Yes, I see. If I made a mistake, please understand.

AndrejMitrovic:
Sure, I just didn't want you to do extra work on the review. :)

source/agora/common/TransactionPool.d
linked0: I think there are some issues about serializing the set to disk (I'm now changing the data structure using Set!Hash as you suggested). First, There is a possibility that the set of Input hashes and the database do not have same information in case a node crashes before serializing the hash set. Second, The tx pool database already have the data for hash set for Input objects, so it's kind of redundant. I know your concerns about loading the Input hashes by iterating over the all transactions, but I don't think it's so big. 
Any opinion would be helpful.

AndrejMitrovic: Ok, then for starters keep it the way it is. We can optimize it later, and ensure we add checksums and things like that later on.

linked0:
Ready for review.

source/agora/common/TransactionPool.d
```diff
@@ -205,8 +232,13 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public void remove (Hash hash) @trusted
+    public void remove (const Transaction tx) @trusted
```
Geod24:
```suggestion
    public void remove (const ref Transaction tx) @trusted
```

source/agora/common/TransactionPool.d
```diff
@@ -205,8 +232,13 @@ public class TransactionPool
...
     {
+        // delete inputs of transaction from the set of Input hashes
+        foreach (input; tx.inputs)
```
Geod24:
```suggestion
        foreach (ref input; tx.inputs)
```

source/agora/common/TransactionPool.d
```diff
@@ -226,21 +258,52 @@ public class TransactionPool
...
+
+    public bool isValidTransaction (const ref Transaction tx) @trusted
+    {
+        auto txHash = tx.hashFull();
+
+        auto results = this.db.execute("SELECT key, val FROM tx_pool " ~
+            "WHERE key = ?", txHash[]);
+
+        if (!results.empty) {
+            return false;
```
Geod24, line 293:
```D
if (!results.empty)
    return false;
```

source/agora/common/TransactionPool.d
```diff
@@ -318,3 +381,41 @@ unittest
...
+    txs = makeChainedTransactions(gen_key, txs, 1);
+
+    // create a double-spend tx
+    txs[0].inputs[0] = txs[1].inputs[0];
+    txs[0].outputs[0].value = Amount(100);
+    auto signature = getGenesisKeyPair().secret.sign(hashFull(txs[0])[]);
+    txs[0].inputs[0].signature = signature;
+
+    // add first tx to the pool
+    pool.add(txs[0]);
```
Geod24: `assert` that they are added

source/agora/common/TransactionPool.d
```diff
@@ -318,3 +381,41 @@ unittest
...
+    scope(exit) pool.shutdown();
+
+    // first set of transactions
+    auto txs = makeChainedTransactions(gen_key, null, 1);
+
+    // second set of transactions
+    txs = makeChainedTransactions(gen_key, txs, 1);
+
+    // create a double-spend tx
+    txs[0].inputs[0] = txs[1].inputs[0];
```
Geod24: Does this works ? It looks like `txs[1]` would be a descendant of `txs[0]`, not the other way around ?

Geod24 (commented):
The unittest looks wrong. Did you check that it passes without the new code ?

source/agora/common/TransactionPool.d
Geod24: After IRL conversation:
- Get rid of `makeChainedTransactions` and just make 2 transactions
- The first transaction must be accepted
- The second transaction has the same inputs as the first but a different output (and is signed)
- The second should be rejected
- Make a second pool, and check that the second transaction is accepted (but the first one isn't)

linked0:
Ready for review.

linked0:
Ready for review.

AndrejMitrovic:
The second commit should be squashed with the first. In general, when you want to merge changes in from the `v0.x.x`  branch, you have to `rebase`, not `merge`. For example, `git pull --rebase origin HEAD`

source/agora/common/TransactionPool.d
```diff
@@ -127,15 +141,26 @@ public class TransactionPool
         Params:
             tx = the transaction to add
 
-        Throws:
-            SqliteException if the transaction failed to be added
+        Returns:
+            true if the transaction may be added to the pool
```
AndrejMitrovic: `may be added` => `has been added`

source/agora/common/TransactionPool.d
```diff
@@ -127,15 +141,26 @@ public class TransactionPool
...
     ***************************************************************************/
 
-    public void add (Transaction tx) @safe
+    public bool add (Transaction tx) @safe
     {
         static ubyte[] buffer;
         buffer.length = 0;
+
+        // check double-spend
+        if (!isValidTransaction(tx))
```
AndrejMitrovic, line 155: `this.isValidTransaction`. And you could do away with braces:

```D
if (!this.isValidTransaction(tx))
    return false;
```

source/agora/common/TransactionPool.d
```diff
@@ -127,15 +141,26 @@ public class TransactionPool
...
+        {
+            return false;
+        }
+
+        // insert each input information of the transaction
+        foreach (input; tx.inputs)
+            input_set.put(input.hashFull());
```
AndrejMitrovic: `this.input_set`

source/agora/common/TransactionPool.d
```diff
@@ -205,8 +232,13 @@ public class TransactionPool
```
AndrejMitrovic, line 230: The parameter documentation needs to be fixed.

source/agora/common/TransactionPool.d
```diff
@@ -205,8 +232,13 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public void remove (Hash hash) @trusted
+    public void remove (const ref Transaction tx) @trusted
     {
+        // delete inputs of transaction from the set of Input hashes
+        foreach (ref input; tx.inputs)
+            input_set.remove(input.hashFull());
```
AndrejMitrovic: `this.input_set`

source/agora/common/TransactionPool.d
```diff
@@ -226,21 +258,51 @@ public class TransactionPool
...
         assert(this.length() == len_prev - count);
         return txs;
     }
+
+    /***************************************************************************
+
+        Check if the input transaction has any double spending input.
+
+        Params:
+            transaction = the transaction to validate
```
AndrejMitrovic: `transaction` => `tx`

source/agora/common/TransactionPool.d
```diff
@@ -226,21 +258,51 @@ public class TransactionPool
...
+        auto results = this.db.execute("SELECT key, val FROM tx_pool " ~
+            "WHERE key = ?", txHash[]);
+
+        if (!results.empty)
+            return false;
+
+        foreach (input; tx.inputs)
+        {
+            auto hash = input.hashFull();
+            if (hash in input_set)
```
AndrejMitrovic: `this.input_set`

AndrejMitrovic:
I believe there is one more part that needs to be updated, in Ledger.d in `acceptTransaction`:

```D
public bool acceptTransaction (Transaction tx) @safe
{
    if (!this.pool.add(tx) || !this.isValidTransaction(tx))
    {
        log.info("Rejected tx: {}", tx);
        return false;
    }
}
```

AndrejMitrovic (changes requested):
Looks good overall, just a few issues left addressing.

linked0:
After pulling new merged sources from Github and doing unit tests, I found out that there is some more work to do. It's related to the freezing capability. There is a case that same input can be included into a payment transaction and a freezing transaction. So the transanction pool have to use the tx type value to make the hash for inputs. I think It might take more time than expected.

AndrejMitrovic:
> There is a case that same input can be included into a payment transaction and a freezing transaction.

I don't think that should be possible? Once a UTXO is referenced by a transaction's input in the blockchain, it's no longer possible to reference it again. Can you show the code example which fails?

Geod24:
> There is a case that same input can be included into a payment transaction and a freezing transaction.

I'm not sure I quite follow, too. The txtype shouldn't matter in that case, only if an output is referenced multiple times.

linked0:
Sorry for my misunderstanding. Anyway, After appying Andrej's suggestion( https://github.com/bpfkorea/agora/pull/356#issuecomment-535324931), I found that the unittest relating to the freezing ability have failed. So, I'm now debugging it. The code failed at the line 806 of agora/node/Ledger.d. 
`
txes.each!((tx)
    {
        assert(ledger.acceptTransaction(tx) == is_valid);
    });
`

AndrejMitrovic:
@linked0 I think my suggestion was slightly wrong:

```
if (!this.pool.add(tx) || !this.isValidTransaction(tx))
```

This should be re-ordered, so `this.isValidTransaction` is checked first, and then `this.pool.add`.

linked0:
It works. Thanks.

AndrejMitrovic:
Excellent. Sorry for causing confusion!

Is the PR ready for another review now?

linked0:
Yes. Just 20 minutes to be ready for the PR. Would you please the reason why it works just by reordering the calls later?

AndrejMitrovic:
Because in this case:

```D
if (!this.pool.add(tx) || !this.isValidTransaction(tx))
```

A transaction will be added to the pool even if it's invalid. The pool itself doesn't do any transaction validation. So if you reorder them:

```D
if (!this.isValidTransaction(tx) || !this.pool.add(tx))
```

Edit: Fixed typo && to ||.

Validation is done first, followed by transaction pool's duplicate transaction filtering.

linked0:
Ready for review.

AndrejMitrovic (approved):

linked0:
> Because in this case:
> 
> ```d
> if (!this.pool.add(tx) || !this.isValidTransaction(tx))
> ```
> 
> A transaction will be added to the pool even if it's invalid. The pool itself doesn't do any transaction validation. So if you reorder them:
> 
> ```d
> if (!this.isValidTransaction(tx) && !this.pool.add(tx))
> ```
> 
> Validation is done first, followed by transaction pool's duplicate transaction filtering.

You mean "if (!this.isValidTransaction(tx) || !this.pool.add(tx))".
Just in order to avoid confusion.

AndrejMitrovic:
Yes, that's right.

source/agora/common/TransactionPool.d
```diff
@@ -127,15 +141,24 @@ public class TransactionPool
...
+
+        // check double-spend
+        if (!isValidTransaction(tx))
+            return false;
+
+        // insert each input information of the transaction
+        foreach (input; tx.inputs)
+            this.input_set.put(input.hashFull());
+
         () @trusted { assumeSafeAppend(buffer); }();
```
Geod24, line 162: Please move this line directly after L152. It needs to follow the length reset.

Geod24 (approved):

source/agora/common/TransactionPool.d
Geod24: Actually we can take care of that later.

Geod24:
Fixes #301

PR #364, opened 2019-09-27
Add bindings to std::function & a workaround for calling it in a timer

AndrejMitrovic:
The API change might make it more difficult to apply new SCP change-sets from upstream.

I've tried to find a way to increment the refcount for `std::function` with a bit of googling, but with no luck. Maybe there is a better way to do this though, I'm open to suggestions.

### DIFF
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -111,6 +111,24 @@ extern(C++, `std`) {
     {
         void*[3] ptr;
     }
+
+    // only used at compile-time on the C++ side, here for mangling
+    struct ratio (int _Num, int _Den = 1)
+    {
+    }
+
+    /// Simple bindings to std::chrono
+    extern(C++, `chrono`)
+    {
+        public struct duration (_Rep, _Period = ratio!1)
+        {
+            _Rep __r;
+            alias __r this;
+        }
+
+        alias milli = ratio!(1, 1000);
+        alias milliseconds = duration!(long, milli);
+    }
 }
 
 private extern(C++) set!uint* makeTestSet();
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -129,7 +129,22 @@ extern(C++, `std`) {
         alias milli = ratio!(1, 1000);
         alias milliseconds = duration!(long, milli);
     }
+
+    /// Simple wrapper around std::function
+    /// note: pragma(mangle) doesn't currently work on types
+    align(1) public struct CPPDelegate (Callback)
+    {
+    align(1):
+        shared_ptr!int __ptr_;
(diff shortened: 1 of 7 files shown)


source/scpd/Cpp.d
```diff
@@ -111,6 +111,16 @@ extern(C++, `std`) {
     {
         void*[3] ptr;
     }
+
+    /// Simple bindings to std::chrono
+    public struct chrono
```
Geod24: Isn't `chrono` a namespace ??

source/scpd/Cpp.d
```diff
@@ -111,6 +111,16 @@ extern(C++, `std`) {
...
+    {
+        public static struct duration
```
Geod24: Isn't it a template class ?

Geod24 (commented):
Hum, I'll need a bit more time to review this. So the timers are not recurring, they are one-shot timers, right ?

AndrejMitrovic:
> Hum, I'll need a bit more time to review this. 

No rush. This will have to carry-over to the next sprint.

> So the timers are not recurring, they are one-shot timers, right ?

They seem to be from what I can tell.

source/scpd/Cpp.d
AndrejMitrovic: Right, I just wanted the mangling to match.

Geod24 (changes requested):
- "Use workaround ..." commit LGTM
- Pending review on `std::chrono` binding
- Can you rebase with `--reset-author` ? Github f'ed the order, as usual

AndrejMitrovic:
Aha, you want me to make std.chrono a namespace?

source/scpd/Cpp.d
AndrejMitrovic: Hmm.. So I looked into the sources. To make `duration` templated I need to turn the struct into:

```D
public static struct duration (_Rep, _Period = ratio!1)
```

And then also bind to `ratio`, which is another templated struct:

```C++
namespace std _GLIBCXX_VISIBILITY(default)
{
    template<intmax_t _Num, intmax_t _Den = 1>
        struct ratio
        {
        }
}
```

Is it ok to bind to the `_GLIBCXX_VISIBILITY` namespace? Because it's specific to glibc and our D binding code might break on other platforms..

Geod24: The namespace is `std`, `_GLIBCXX_VISIBILITY` is just a macro (which does nothing in this case).

Geod24:
> Aha, you want me to make std.chrono a namespace?

Yep, or is there an issue I'm not seeing ?

AndrejMitrovic:
Fixed the bindings.

Geod24 (approved):

PR #366, opened 2019-09-30
Update ocean to HEAD, reduce diff

Geod24:
```
This massively reduce the diff with upstream.
Our new approach use a (long) list of excluded files,
but makes the actual diff much easier to reason with.
This will also make future upgrade much simpler,
as rebases will only conflict with code we actually modified.
```

The diff between v5.x.x and our fork [is visible here](https://github.com/sociomantic-tsunami/ocean/compare/v5.x.x...bpfkorea:dev-2019-10-01?expand=1)


Geod24:
This triggers an LDC bug. Great..

Geod24:
This is green at last

AndrejMitrovic (approved):

PR #367, opened 2019-10-01
Add pretty-print formatting of PublicKey

AndrejMitrovic:
Similar to formatting hashes.


AndrejMitrovic:
And in fact, in unittests it would be cool to give them assigned names. For example: https://github.com/sociomantic-tsunami/ocean/blob/v4.x.x/src/ocean/io/digest/FirstName.d#L58

Geod24 (approved):

Issue #362, opened 2019-09-26
Output of `Amount(0)` should be rejected


Geod24:
Because it can lead to spam.

AndrejMitrovic:
Looks like this is resolved. @MukeunKim can you close it?

PR #369, opened 2019-10-01
Fix to remove transactions with output value of 0

MichaelKim20:
Transactions with output of `Amount(0)` should be rejected.
Because it can lead to spam.

#362

### DIFF
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -52,10 +52,16 @@ public string isInvalidReason (const Transaction tx, UTXOFinder findUTXO,
     if (tx.outputs.length == 0)
         return "Transaction: No output";
 
-    // disallow negative amounts
     foreach (output; tx.outputs)
+    {
+        // disallow negative amounts
         if (!output.value.isValid())
             return "Transaction: Output(s) overflow or underflow";
+
+        // disallow 0 amount
+        if (output.value == Amount(0))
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
@MukeunKim that was fast!

AndrejMitrovic:
In general, the commit which introduces new verification should also include the test, so I would squash the first two commits together. @Geod24 do you agree?

MichaelKim20:
> @MukeunKim that was fast!

There's a overlap with the last PR.

MichaelKim20:
> In general, the commit which introduces new verification should also include the test, so I would squash the first two commits together. @Geod24 do you agree?
I just squashed the first two commit.

AndrejMitrovic (approved):

Geod24 (approved):
Nice! This makes me think though, we need to add a way to burn, so we might accept txs without output in the future.... @AndrejMitrovic ?

Geod24:
Ah I didn't see it was too commit. Right. Next time please put the test in the same commit @MukeunKim . The idea is that we never want to revert one of those commits individually. If we revert the test, the code would be not covered, which is bad. And if we remove the implementation, the test would fail.

AndrejMitrovic:
> Nice! This makes me think though, we need to add a way to burn, so we might accept txs without output in the future.

It could just be an output to the address 0? In other words, it should already work?

Issue #360, opened 2019-09-26
Add test cases for transaction validation that aren't covered


Geod24:
According to the coverage report, those cases are not covered:
- Transactions without input
- Input overflow (duh)
- Combined input
- Unknown transaction type
- Mixed inputs

Needs to cover both payments and freeze Txs.

Geod24:
Coverage report: https://codecov.io/gh/bpfkorea/agora/src/v0.x.x/source/agora/consensus/Validation.d

![image](https://user-images.githubusercontent.com/2180215/65927568-219f5300-e435-11e9-85ac-ad7240e61259.png)

linked0:
I can't tell between combined input and mixed input. Actually, I found out that combined input means that the freezing count and the input length of tx in the code are not equal. Would you please give more information about mixed input?

linked0:
After IRL(in real life) conversation with Mathias, It is confirmed that combined inputs and mixed inputs are same things.

PR #374, opened 2019-10-07
Add validation test cases for no input, invalid type, combined input and etc

linked0:
Add unittest codes checking validation relating to:
- no input/output
- invalid transaction type
- combined inputs
- input overflow for Payment and Freeze transaction
- output overflow for Payment transaction

### DIFF
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -651,6 +651,274 @@ unittest
     }
 }
 
+/// test for transactions having no input or no output
+unittest
+{
+    import std.format;
+    import std.string;
+    import std.algorithm.searching;
+
+    Transaction[Hash] storage;
+    KeyPair key_pair = KeyPair.random;
+
+    // delegate for finding `Output`
+    scope findUTXO = (Hash hash, size_t index, out UTXOSetValue value)
+    {
+        if (auto tx = hash in storage)
+        {
+            if (index < tx.outputs.length)
+            {
+                value.unlock_height = 0;
+                value.type = TxType.Payment;
+                value.output = tx.outputs[index];
+                return true;
+            }
+        }
+        return false;
+    };
+
+    // create a transaction having no input
+    Transaction oneTx = Transaction(
+        TxType.Payment,
+        [],
+        [Output(Amount(50), key_pair.address)]
+    );
+    Hash oneHash = hashFull(oneTx);
+    storage[oneHash] = oneTx;
+
+    // test for Payment transaction having no input
+    assert(canFind(toLower(oneTx.isInvalidReason(findUTXO, 0)), "no input"), 
+        format("Tx having no input should not pass validation. tx: %s", oneTx));
+
+    // create a transaction
+    Transaction firstTx = newCoinbaseTX(key_pair.address, Amount(100_1000));
+    Hash firstHash = hashFull(firstTx);
+    storage[firstHash] = firstTx;
+    firstTx.inputs[0].signature = key_pair.secret.sign(firstHash[]);
+
+    // create a transaction having no output
+    Transaction secondTx = Transaction(
+        TxType.Payment,
+        [Input(firstHash, 0)],
+        []
+    );
+    Hash secondHash = hashFull(secondTx);
+    storage[secondHash] = secondTx;
+
+    // test for Freeze transaction having no output
+    assert(canFind(toLower(secondTx.isInvalidReason(findUTXO, 0)), "no output"), 
+        format("Tx having no output should not pass validation. tx: %s", secondTx));
+}
+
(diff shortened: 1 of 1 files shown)


linked0:
This PR is related to #360 (for later reference)

source/agora/consensus/Validation.d
```diff
@@ -631,6 +631,268 @@ unittest
...
+        TxType.Payment,
+        [],
+        [Output(Amount(50), key_pair.address)]
+    );
+    Hash oneHash = hashFull(oneTx);
+    storage[oneHash] = oneTx;
+
+    // Test for Payment transaction having no input
+    assert(!oneTx.isValid(findUTXO, 0), 
+        format("Tx having no input is not validated %s", oneTx));
```
Geod24: A general rule on assert message: Consider how it's going to show up in the console.
In this case, another developer would see: "Assertion failure: Tx having no input is not validated".
This is quite ambiguous, as if the message describes what should happen, not what is happening.
Consider the following wording:
- "Tx having no input should not pass validation"
- "Tx having no input are considered valid"

Those two wordings might not be perfect, but are much less ambiguous: the first one define what should happen, and a reader will understand that it is *not* happening (because the assertion failed), or what is happening, and hopefully that will be clear enough that the reader can infer what is wrong.

source/agora/consensus/Validation.d
```diff
@@ -631,6 +631,268 @@ unittest
     }
 }
 
+/// Test for transactions having no input or no output
```
Geod24: Looks like the coverage report is really screwed up tho.
It might come from DMD, and I don't think there's much we can do about it.

source/agora/consensus/Validation.d
```diff
@@ -631,6 +631,268 @@ unittest
...
+        [Output(Amount(50), key_pair.address)]
+    );
+    Hash oneHash = hashFull(oneTx);
+    storage[oneHash] = oneTx;
+
+    // Test for Payment transaction having no input
+    assert(!oneTx.isValid(findUTXO, 0), 
+        format("Tx having no input is not validated %s", oneTx));
+
+    // Creates a transaction having no output
```
Geod24: I think this comment would be clearer before `secondTx`.

source/agora/consensus/Validation.d
```diff
@@ -631,6 +631,268 @@ unittest
...
+    Transaction secondTx = Transaction(
+        TxType.Payment,
+        [Input(firstHash, 0)],
+        []
+    );
+    Hash secondHash = hashFull(secondTx);
+    storage[secondHash] = secondTx;
+
+    // Test for Freeze transaction having no output
+    assert(!secondTx.isValid(findUTXO, 0),
```
Geod24: I would use `invalidReason().cointains("No output")`.
You don't want to match the full message (it creates too much coupling which can be annoying), but it's very easy to have a test pass for the wrong reason.

source/agora/consensus/Validation.d
```diff
@@ -631,6 +631,268 @@ unittest
...
+    // Create the second transaction
+    auto secondTx = Transaction(
+        TxType.Payment,
+        [Input(Hash.init, 0)],
+        [Output(Amount(100), key_pairs[0].address)]
+    );
+    auto secondHash = hashFull(secondTx);
+    storage[secondHash] = secondTx;
+
+    // Cretea the third transaction
```
Geod24: Typo

Geod24 (commented):
A few comments, good work!

AndrejMitrovic:
The commit message is cut off because it's longer than the 70 or 80 character limit. You could shorten the title and write a longer message in the description. For example:

```
Add more transaction validation tests

Now includes tests for: no input, no output, invalid type, combined input, and overflow.
```

linked0:
Ready for review.
- applied Mathias and Andrej's comments.

source/agora/consensus/Validation.d
```diff
@@ -651,6 +651,273 @@ unittest
...
+    Hash secondHash = hashFull(secondTx);
+    storage[secondHash] = secondTx;
+
+    // test for Freeze transaction having no output
+    assert(canFind(toLower(secondTx.isInvalidReason(findUTXO, 0)), "no output"), 
+        format("Tx having no output should not pass validation. tx: %s", secondTx));
+}
+
+/// test for transaction having combined inputs
+unittest {
```
AndrejMitrovic: Ah, I missed this one.

```suggestion
unittest 
{
```

AndrejMitrovic (commented):
Just one styling change, otherwise it looks good to me.

linked0:
Ready for review.

AndrejMitrovic (approved):

Issue #345, opened 2019-09-20
Change GossipProtocol.hasTransactionHash() to check tx hash in the pool


AndrejMitrovic:
The periodic getBlocksFrom() call does not add transactions from the blocks to the `GossipProtocol`, so `hasTransactionHash()` will fail even if a node contains a transaction in its ledger.

Geod24:
But [`hasTransactionHash`](https://github.com/bpfkorea/agora/blob/e21d0573c6feaf8fb2d181370788dcec77c35f05/source/agora/node/GossipProtocol.d#L89) only checks the TxPool, not the ledger.

AndrejMitrovic:
Ah yeah. Well there is another issue, we never remove items from the Gossip cache..

AndrejMitrovic:
> But hasTransactionHash only checks the TxPool, not the ledger.

From what I understand, the reason `hasTransactionHash` was introduced to a avoid costly network operation when sending a tx that a node already contains.

If a tx is in a node's pool there's no need for another node to send it the tx again. But if the tx is in the blockchain there is also no need for another node to send it because that transaction won't be accepted (it double-spends UTXO).

Currently `hasTransactionHash` acts as a permanent storage of all seen transaction hashes. Well, except that it doesn't serialize its data to disk..

Question then:
- Should it only return true if a tx is in the node's tx pool?
- Or should it still keep the set of all tx hashes it's ever seen, but we should persist that data across restarts?

Geod24:
> Should it only return true if a tx is in the node's tx pool?

Yes

> Or should it still keep the set of all tx hashes it's ever seen, but we should persist that data across restarts?

No

MichaelKim20:
We should persist that data across restarts?.
When we start a node, if we don't have to maintain historical data,
We'd better use the cache.

MichaelKim20:
Of course I'll handle it in the Transaction Pool!

AndrejMitrovic:
> We should persist that data across restarts?.

As @Geod24 said in https://github.com/bpfkorea/agora/issues/345#issuecomment-536819178, we don't have to persist data. We just have to check if the transaction hash is in the TransactionPool.

TrustHenry:
@Geod24 @AndrejMitrovic 
How about a new function in TransactionPool?
The new function is `SELECT key, val FROM tx_pool WHERE key = ? ` in the TransactionPool DB, and Tx hash result have value.

[...]

PR #375, opened 2019-10-07
Change GossipProtocol.hasTransactionHash() to check tx hash in the pool

TrustHenry:
We never remove items from the Gossip cache.
So, This deletes the transaction cache of the Gossipprotocol and queries it directly from the transaction pool.
For #345

### DIFF
--- a/source/agora/test/GossipProtocol.d
+++ b/source/agora/test/GossipProtocol.d
@@ -44,20 +44,28 @@ unittest
     nodes.each!(node => assert(node.getBlockHeight() == 0));
 
     Transaction[] last_txs;
-    foreach (block_idx; 0 .. 10)  // create 10 blocks
-    {
-        // create enough tx's for a single block
-        auto txs = makeChainedTransactions(getGenesisKeyPair(), last_txs, 1);
+    // create enough tx's for a single block
+    auto txs = makeChainedTransactions(getGenesisKeyPair(), last_txs, 1);
 
-        // send it to one node
-        txs.each!(tx => node_1.putTransaction(tx));
+    auto send_txs = txs[0..$-1];
+    // send it to tx to node
+    send_txs.each!(tx => node_1.putTransaction(tx));
+    // gossip was complete
+    nodes.each!(node =>
+       send_txs.each!(tx =>
+           node.hasTransactionHash(hashFull(tx)).retryFor(2.seconds)
+    ));
+    // When a block is created, the transaction is deleted from the transaction pool.
+    node_1.putTransaction(txs[$-1]);
+    nodes.enumerate.each!((idx, node) =>
+        retryFor(node.getBlockHeight() == 1,
+        2.seconds,
+        format("Node %s has block height %s. Expected: %s",
+        idx,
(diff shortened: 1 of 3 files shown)


Geod24 (approved):

TrustHenry:
An error persists in Travis CI mac os. Does anyone know what's causing this?
https://travis-ci.com/bpfkorea/agora/jobs/242615835
`ld: warning: directory not found for option '-L@@HOMEBREW_CELLAR@@/libsodium/1.0.18/lib'`

Geod24:
That warning is not the problem. The problem happens when the unittest are run. See :

> Program exited with code -4

source/agora/test/GossipProtocol.d
```diff
@@ -44,20 +44,19 @@ unittest
...
-
-        // gossip was complete
-        nodes.each!(node =>
-            txs.each!(tx =>
-                node.hasTransactionHash(hashFull(tx)).retryFor(1.seconds)
-        ));
-
-        last_txs = txs;
-    }
+    // create enough tx's for a single block
```
AndrejMitrovic, line 47: ~Could you change the unittest documentation comment and write `/// test hasTransactionHash()`? Currently it's empty.~

AndrejMitrovic: Ah actually ignore my comment! I didn't realize this was a networking test.

source/agora/test/GossipProtocol.d
```diff
@@ -44,20 +44,19 @@ unittest
...
+    send_txs.each!(tx => node_1.putTransaction(tx));
+    // gossip was complete
+    nodes.each!(node =>
+       send_txs.each!(tx =>
+           node.hasTransactionHash(hashFull(tx)).retryFor(1.seconds)
+    ));
+    // When a block is created, the transaction is deleted from the transaction pool.
+    node_1.putTransaction(txs[$-1]);
+    assert(node_1.getBlockHeight() == 1);
+    txs.each!(tx => assert(!node_1.hasTransactionHash(hashFull(tx))));
```
AndrejMitrovic: Whenever we do networking tests we must use the `retryFor` method. This is because requests are asynchronous, and it's likely the reason why you have test-case failures. Try writing it like this:

```D
    retryFor(node.getBlockHeight() == 1,
        2.seconds,
            idx, node.getBlockHeight().to!string, 1)));
    }

nodes.each!(node =>
    txs.each!(tx =>
        !node.hasTransactionHash(hashFull(tx)).retryFor(1.seconds)
));
```

It's also important to test all nodes, and not just `node_1`, as demonstrated.

source/agora/test/GossipProtocol.d
TrustHenry: Okay :)

source/agora/test/GossipProtocol.d
TrustHenry: After small code modification
Suggested changed.

AndrejMitrovic (approved):

AndrejMitrovic:
The test failure is strange. It's not caused by the new test.

PR #377, opened 2019-10-09
Color log level in output

Geod24:
This makes large output easier to deal with, as warn / error will be more obvious.

### DIFF
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -26,6 +26,10 @@
 
 module agora.utils.Log;
 
+import ocean.text.convert.Formatter;
+import ocean.util.log.AppendConsole;
+import ocean.util.log.Appender;
+import ocean.util.log.Event;
 import ocean.util.log.Logger;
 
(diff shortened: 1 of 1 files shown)


source/agora/utils/Log.d
```diff
@@ -48,6 +52,58 @@ else
     /// Initialize the logger
     static this ()
     {
-        Log.defaultConfig();
+        auto appender = new AppendConsole();
```
AndrejMitrovic, line 55: Might want to update [this one too](https://github.com/bpfkorea/agora/blob/3df33d16faf60d682e8312c0f5b3702307e2a4b8/source/agora/common/TransactionPool.d#L49)?

Geod24: Why is that ?

AndrejMitrovic: Ah sorry I misread the diff, I thought this was inside the mixin template.

AndrejMitrovic (approved):

PR #378, opened 2019-10-09
Use CAS to make initialization of TransactionPool atomic

Geod24:
This makes initialization more predictableand easier to verify.
Additionally, the logger callback provided to SQLite could
have created problems, as it is referencing a `null` value
(at the moment the callback is supplied).


AndrejMitrovic:
That's nice.

AndrejMitrovic (approved):

PR #381, opened 2019-10-10
Implement a custom unittest runner + manual thread shutdown

AndrejMitrovic:
This should fix most issues with the random segfaults in ocean and when using SCP.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -154,7 +154,14 @@ public class TestAPIManager
 
     public void shutdown ()
     {
-        this.apis.each!(a => a.shutdown());
+        foreach (key, ref api; this.apis)
+        {
+            api.shutdown();
+            api.ctrl.shutdown();
+            api = null;
+        }
+
+        this.apis = null;
     }
 
     /// fill in the in-memory metadata with the peers before nodes are started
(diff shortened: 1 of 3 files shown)


source/agora/test/Base.d
```diff
@@ -45,8 +45,59 @@ import std.algorithm.iteration;
...
+            }
+            catch (Throwable ex)
+            {
+                writeln(ex);
+            }
+        }
+    }
+
+    //result.summarize = true;
+    result.runMain = false;
```
Geod24, line 97: <3

Geod24 (approved):
LVGTM

AndrejMitrovic:
It seems one of the builds always stalls for a long while, and it's because of frequent failures to upload the coverage report to ~circlecli~ codecov:

Screen Shot 2019-10-14 at 10 23 58

AndrejMitrovic:
Actually, how does coverage reporting work? We have multiple builds, so which coverage report is used for ~CircleCI~ codecov? Just whichever was the last build that finished?

Edit: I keep using the wrong word, sorry.

Geod24:
> It seems one of the builds always stalls for a long while, and it's because of frequent failures to upload the coverage report to circlecli codecov:

Yeah I think this machine's network is bad

> Actually, how does coverage reporting work? We have multiple builds, so which coverage report is used for CircleCI codecov? Just whichever was the last build that finished?

It merges them

AndrejMitrovic:
Great. And it's green.

PR #382, opened 2019-10-10
[do not merge] Custom unittest runner + explicit thread shutdown

AndrejMitrovic:
I want to see if the combo of not running ocean tests + having explicit shutdown will get rid of these random failures we have in travis on a mac.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -18,7 +18,7 @@
 	url = https://github.com/Geod24/libsodiumd.git
 [submodule "localrest"]
 	path = submodules/localrest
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
Travis seems to pass on this one (docker fails because I pointed the localrest submodule to my fork).

So I think the combo will work well. I'll retrigger the build a few more times.

AndrejMitrovic:
The build passed 3 times in a row, so I think this confirms it, the shutdown functionality is required.

PR #383, opened 2019-10-11
Change makeNewBlock() to accept ranges

AndrejMitrovic:
This will allow using it with transaction sets without having to allocate arrays.

Related to #209

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -352,13 +352,15 @@ public struct Block
 
 *******************************************************************************/
 
-public Block makeNewBlock (const ref Block prev_block, Transaction[] txs) @safe
+public Block makeNewBlock (Transactions)(const ref Block prev_block,
+    Transactions txs) @safe
+    if (isInputRange!Transactions)
 {
     Block block;
 
     block.header.prev_block = prev_block.header.hashFull();
     block.header.height = prev_block.header.height + 1;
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
~Fixed segfault. Apparently `foo([1].take(1))` is not safe..~

Edit: Hmm I don't think that was it. The segfault is random actually.

Geod24 (approved):

source/agora/consensus/data/Block.d
```diff
@@ -352,13 +352,15 @@ public struct Block
...
+public Block makeNewBlock (Transactions)(const ref Block prev_block,
+    Transactions txs) @safe
+    if (isInputRange!Transactions)
 {
     Block block;
 
     block.header.prev_block = prev_block.header.hashFull();
     block.header.height = prev_block.header.height + 1;
-    block.txs ~= txs;
+    txs.each!(tx => block.txs ~= tx);
```
Geod24, line 363: That's way less efficient though

AndrejMitrovic: Hmm, I didn't think of that. I could add a static if case though.

Geod24: I usually check for the presence of `length`, then do a `reserve`. But I think we can take care of this later, there are many places which are not very efficient ATM.

PR #384, opened 2019-10-15
Make logging easily toggle-able

AndrejMitrovic:
This has been an annoyance to me, because each time I wanted to log something in a test I would have to touch `Log.d` (and make sure I don't accidentally stage it).

### DIFF
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -54,11 +54,16 @@ public template AddLogger (string moduleName = __MODULE__)
 /// Convenience alias
 public alias LogLevel = Level;
 
-version (unittest) {}
+/// Whether logging is enabled (by default disabled in unittests)
+version (unittest)
+    __gshared bool EnableLogging = false;
 else
+    __gshared bool EnableLogging = true;
+
+/// Initialize the logger
+static this ()
 {
-    /// Initialize the logger
-    static this ()
+    if (EnableLogging)
     {
         auto appender = new AppendConsole();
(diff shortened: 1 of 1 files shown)


source/agora/utils/Log.d
```diff
@@ -54,11 +54,16 @@ public template AddLogger (string moduleName = __MODULE__)
 /// Convenience alias
 public alias LogLevel = Level;
 
-version (unittest) {}
+/// Whether logging is enabled (by default disabled in unittests)
+version (unittest)
+    __gshared bool EnableLogging = false;
 else
+    __gshared bool EnableLogging = true;
```
Geod24, line 61: So what exactly is the plan here ? Enabling logging before starting new threads ?
It also looks like you can never change it after you enabled / disabled it.

If you want to selectively enable/disable, and don't need the ability to affect other threads, you could go with something like:

```D
version(unittest)
{
    void disableLogging () { Log.root.level(Level.None, true); }
    void enableLogging ()  { Log.root.level(Level.Trace, true); }
}
```

AndrejMitrovic: > If you want to selectively enable/disable, and don't need the ability to affect other threads,

I do need to affect other threads. It's the child threads that run the nodes.

AndrejMitrovic:
Labeling as blocked until more work is done.

AndrejMitrovic:
Replaced by https://github.com/bpfkorea/agora/pull/391

Issue #346, opened 2019-09-20
Change the scpp logging system to use our logger instead of stdout


AndrejMitrovic:
### Description

Currently SCP code logs directly to stdout, see here: https://github.com/bpfkorea/agora/blob/e1351051e09ac71d2144edb03fdcd4dfad5d4b6a/source/scpp/src/util/Logging.h#L16

This code originally used the SCP logger:
https://github.com/stellar/stellar-core/blob/578b951aa027b850807fc15841a6eb2fffda611e/src/util/Logging.h

https://github.com/stellar/stellar-core/blob/578b951aa027b850807fc15841a6eb2fffda611e/src/util/Logging.cpp

However, we should change it to use our D ocean logger system instead. One approach to doing this is to provide D wrapper functions which forward logging to the ocean logger. We primarily use the ocean logger with this helper module: https://github.com/bpfkorea/agora/blob/v0.x.x/source/agora/utils/Log.d

The SCP code typically uses logging like this:

```
CLOG(ERROR, "SCP") << "not sane statement from self, skipping " 
    << "  e: " << mSlot.getSCP().envToStr(envelope);
```

This could be forwarded to a D function with an extern(C++) linkage, to a logger instantiated with `Log.lookup("SCP")` and called with `log.error(...)` on the D side.

The part to figure out is how to pass the arguments to the D side. Would it be a single string constructed on the C++ side, or a number of arguments (perhaps array of strings, variable-length arguments, etc).

Also, care needs to be taken to ensure there are no memory leaks.

### Definition of done:

- [ ] Make SCP use Ocean's logger instantiated on the D side. SCP only logs to a `"SCP"` logger, so you only need to instantiate this one logger instance for it to work.
- [ ] Different logging levels should be supported, so for example `CLOG(TRACE, ...)`  and `CLOG(ERROR, ...)` are using log.trace / log.error, etc, on the D side.
- [ ] Ensure there are no memory leaks in the SCP logging code. You might have to manually delete the C++ strings on the D side after logging.

As a bonus, it would be nice to not have any excessive memory allocations. However, we can improve on this later if it's unavoidable.

Geod24:
This is a good issue if anyone wants to get familiar with the C++ interop side of things

bpalaggi:
Relates to #209

linked0:
[...]

PR #388, opened 2019-10-17
Make a logger for SCP and logging functions

linked0:
The logger for SCP module is created in agora.utils.Log module.
The logging function in SCP module makes a formatted message and call a logging function in D.
Unittest code is added.

Related to #346

### DIFF
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -115,3 +115,28 @@ public class AgoraLayout : Appender.Layout
         }
     }
 }
+
+/***************************************************************************
+
+    Logging function which is only called from C++ code
+
+    It's for C++ code to use agora's logger instead of C++ stdout 
+
+    Params:
+        logger = the logger name
+        level = the logging level
+        msg = the log message
+
+***************************************************************************/
+
+private extern (C++) void writeDLog (const char* logger, int level, const char* msg)
+{
+    if (level >= Level.min && level <= Level.max)
+    {
+        import std.string;
+
+        auto log = Log.lookup(fromStringz(logger));
+        assert(log !is null);
+        log.format(cast(Level) level, fromStringz(msg));
+    }
+}
--- a/source/scpp/src/util/Logging.h
+++ b/source/scpp/src/util/Logging.h
@@ -8,12 +8,16 @@
 #include <sstream>
 #include <iostream>
 
-#define TRACE "trace"
-#define DEBUG "debug"
-#define INFO "info"
-#define ERROR "error"
-#define FATAL "fatal"
-#define CLOG(LEVEL, MOD) std::cout << "[" << LEVEL << ", " << MOD << "] "
+#define TRACE 0
+#define DEBUG 0
+#define INFO  1
+#define WARN  2
+#define ERROR 3
+#define FATAL 4
+#define CLOG(LEVEL, MOD) stellar::DLogger(LEVEL, MOD)
+
+// Logging function to D code
+void writeDLog(const char* logger, int level, const char* msg);
 
 namespace stellar
 {
@@ -26,5 +30,24 @@ class Logging
     static bool logDebug(std::string const& partition) { return true; }
     static bool logTrace(std::string const& partition) { return true; }
     static void rotate();
+};
+
+struct DLogger
+{
+  private:
+    std::string mLoggerName;
+    int mLevel;
+    std::ostringstream mOutStream;
+
+  public:
+    DLogger(int level, std::string const& loggerName);
+    ~DLogger();
+
+    template <class T>
+    DLogger &operator<<(const T &value)
+    {
+        mOutStream << value;
+        return *this;
+    }
 };
 }
--- a/dub.json
+++ b/dub.json
@@ -21,6 +21,7 @@
         "source/scpp/build/Hex.o",
         "source/scpp/build/KeyUtils.o",
         "source/scpp/build/LocalNode.o",
+        "source/scpp/build/Logging.o",
         "source/scpp/build/Math.o",
         "source/scpp/build/NominationProtocol.o",
         "source/scpp/build/QuorumSetUtils.o",
--- a/source/scpp/src/util/Logging.cpp
+++ b/source/scpp/src/util/Logging.cpp
@@ -1,0 +1,24 @@
+// Copyright 2014 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
+#include <cstdarg>
+#include <iostream>
+#include <vector>
+#include "Logging.h"
+
+using namespace std;
+
+namespace stellar
+{
+DLogger::DLogger(int level, std::string const& loggerName)
+{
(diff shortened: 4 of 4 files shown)


source/agora/utils/Log.d
```diff
@@ -115,3 +115,57 @@ public class AgoraLayout : Appender.Layout
         }
     }
 }
+
+// Log Level Definition for C++ code
+enum LogTrace = Level.Trace;
+enum LogInfo = Level.Info;
+enum LogWarn = Level.Warn;
+enum LogError = Level.Error;
+enum LogFatal = Level.Fatal;
```
Geod24: That doesn't seem to be used / useful ? Those definition are only visible from D.

source/agora/utils/Log.d
```diff
@@ -115,3 +115,57 @@ public class AgoraLayout : Appender.Layout
...
+
+// Add "SCP" logger
+public template AddSCPLogger ()
```
Geod24: There's no need to make a template unless you plan to mix it in at multiple location, which doesn't seem to be the case ?

source/agora/utils/Log.d
```diff
@@ -115,3 +115,57 @@ public class AgoraLayout : Appender.Layout
...
+    private Logger scplog;
+    static this ()
+    {
+        scplog = Log.lookup("SCP");
+    }
+}
+mixin AddSCPLogger!();
+
+// Logging function called from C++
+extern (C++) void DLogging(const char *logger, int level, const char *msg)
```
Geod24: Note that in D, you'd use `const char* ident`, not `const char *ident`.

source/agora/utils/Log.d
```diff
@@ -115,3 +115,57 @@ public class AgoraLayout : Appender.Layout
...
+{
+    import std.stdio;
+    import std.string;
+    
+    switch(level) {
```
Geod24: Style nitpicks:
- Braces on their own line
- space between `switch` and parenthesis

But that `switch` is actually not necessary. Look it [Logger.info](https://github.com/bpfkorea/ocean/blob/c053e665a127ec9b31cab00ff8dfc3354a1558e0/src/ocean/util/log/Logger.d#L504-L507) definition.

source/agora/utils/Log.d
```diff
@@ -115,3 +115,57 @@ public class AgoraLayout : Appender.Layout
...
+        case LogFatal:
+            scplog.error(fromStringz(msg));
+            break;
+        default:
+            scplog.trace(fromStringz(msg));
+    }
+}
+
+// Logging function test to C++ code 
+extern (C++) void logTestFromD();
```
Geod24: Style: This should be `private` and `version(unittest)`

Geod24 (changes requested):
I left a couple of style changes / question. But there seem to have been a misunderstanding: the core of the task was not to make an utility to call from C++, it is to replace the `#define CLOG(LEVEL, MOD)` in Logging.h at line 16 with something that uses the D logger. Since this line is untouched, the added code is essentially dead code.

linked0:
Ready for review.
Any suggestions would be good, especially about the name of the class or the function.

source/scpp/src/util/Logging.h
```diff
@@ -13,7 +13,17 @@
 #define INFO "info"
```
AndrejMitrovic: Since we are already modifying this module, it would make sense to also replace these defines with:

```
# define TRACE 0
# define DEBUG 0
# define INFO 1
```

And then you don't have to do expensive string comparisons inside of `DLogger::DLogger`, and can get rid of the `LogTrace / LogInfo / etc` defines.

source/agora/utils/Log.d
```diff
@@ -115,3 +115,43 @@ public class AgoraLayout : Appender.Layout
...
+    import std.string;
+
+    auto log = Log.lookup(fromStringz(logger));
+    if (log !is null)
+    {
+        log.format(cast(Level) level, fromStringz(msg));
+    }
+}
+
+version(unittest)
```
AndrejMitrovic: I think the `logTestFromD` and the associated unittest are unnecessary here. They're not actually testing anything other than doing a function call, which the SCP code already does. For example the `[SCP] "Hello, world 1 99.9"` string is not actually tested anywhere in this unittest.

Since the logging code is fairly trivial, I think it's ok not to add tests for it. We will notice if logging fails.

CC @Geod24

AndrejMitrovic:
Overall looks good, just a few tiny issues I've mentioned.

source/agora/utils/Log.d
```diff
@@ -115,3 +115,43 @@ public class AgoraLayout : Appender.Layout
...
+
+    It's for C++ code to use agora's logger instead of C++ stdout 
+
+    Params:
+        logger = the logger name
+        level = the logging level
+        msg = the log message
+
+***************************************************************************/
+private extern (C++) void DLogging (const char* logger, int level, const char* msg)
```
Geod24: This needs to respect the D style though

source/agora/utils/Log.d
```diff
@@ -115,3 +115,43 @@ public class AgoraLayout : Appender.Layout
...
+{
+    import std.string;
+
+    auto log = Log.lookup(fromStringz(logger));
+    if (log !is null)
```
Geod24: In D you can do the following:
```suggestion
if (auto log = Log.lookup(fromStringz(logger))
```

However, `Log.lookup` should never return `null`, so I would say don't do the check, or make it an `assert`.
Additionally, I would check the value of `level` before casting.

source/scpp/src/util/Logging.h
```diff
@@ -27,4 +37,23 @@ class Logging
...
+
+class DLogger
+{
+  private:
+    std::string mLoggerName;
+    int mLevel;
+    std::ostringstream mOutStream;
+
+  public:
+    DLogger(std::string level, std::string loggerName);
```
Geod24: const ref

source/scpp/src/util/Logging.h
```diff
@@ -27,4 +37,23 @@ class Logging
     static bool logTrace(std::string const& partition) { return true; }
     static void rotate();
 };
+
+class DLogger
```
Geod24:
```suggestion
struct DLogger
```

Geod24 (approved):
Overall LGTM. I do slightly worry about allocations, but I think it's fine for now.

source/agora/utils/Log.d
```diff
@@ -115,3 +115,28 @@ public class AgoraLayout : Appender.Layout
...
+***************************************************************************/
+
+private extern (C++) void writeDLog (const char* logger, int level, const char* msg)
+{
+    if (level >= Level.min && level <= Level.max)
+    {
+        import std.string;
+
+        auto log = Log.lookup(fromStringz(logger));
+        assert (log !is null);
```
Geod24:
```suggestion
        assert(log !is null);
```

Geod24 (approved):
LGTM. Feel free to merge once green.

AndrejMitrovic:
I restarted the build on Travis.

AndrejMitrovic (approved):

AndrejMitrovic:
Nice work.

PR #399, opened 2019-10-24
Add ability to test specific modules

AndrejMitrovic:
By simply using:

```
$ test=Gossip dub test
```

Which will only run tests within modules whose name matches the "Gossip" string.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -65,8 +65,14 @@ shared static this()
 private UnitTestResult customModuleUnitTester ()
 {
     import std.algorithm;
+    import std.process;
     import std.stdio;
     import std.string;
+    import std.uni;
+
+    //
+    auto filter = environment.get("dtest").toLower();
+    size_t filtered;
 
     UnitTestResult result;
     foreach (ModuleInfo* mod; ModuleInfo)
@@ -81,6 +87,13 @@ private UnitTestResult customModuleUnitTester ()
         if (mod.name.startsWith("agora") ||
             mod.name.startsWith("scpd"))
         {
+            if (filter.length > 0 &&
+                !canFind(mod.name.toLower(), filter.save))
+            {
+                filtered++;
+                continue;
+            }
+
             ++result.executed;
 
             try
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
It's not super advanced and not using regex, but I just needed something really simple for now so I can just test one module.

Geod24:
Few things:
- should we use case insensitive checks?
- `test` is too common, maybe `dtest`?
- Did you look at unit-threaded?

AndrejMitrovic:
> should we use case insensitive checks?

Sure.

> test is too common, maybe dtest?

Okay.

> Did you look at unit-threaded?

Yes it's a framework and seemed a bit overkill right now. I'm thinking of bringing it in later maybe. Have you used it? It seems a bit over-designed.

Geod24:
> Yes it's a framework and seemed a bit overkill right now. I'm thinking of bringing it in later maybe. Have you used it? It seems a bit over-designed.

Not directly, I just know it's an alternative, but I wish everything was built in...

AndrejMitrovic:
> Not directly, I just know it's an alternative, but I wish everything was built in...

It really should be. Actually I'm not even sure how `unit-threaded` works, because the `ModuleInfo` doesn't give you an array of unittests, just one giant function that internally calls all the unittests in a module.

AndrejMitrovic:
Updated to use case-insensitive matching and `dtest=`.

source/agora/test/Base.d
```diff
@@ -81,6 +87,13 @@ private UnitTestResult customModuleUnitTester ()
         if (mod.name.startsWith("agora") ||
             mod.name.startsWith("scpd"))
         {
+            if (filter.length > 0 &&
+                !canFind(mod.name.toLower(), filter))
```
Geod24: You see that's one thing I hate about ranges: the "copy is an implicit save" thing is nice, but it breaks ref / non copyable range.

In this case, I believe, this would be better / more correct:
```suggestion
                !canFind(mod.name.toLower(), filter.save))
```

AndrejMitrovic: I thought it returned an array. Oh well.

Geod24 (approved):

PR #400, opened 2019-10-24
Parallelize the test-suite

AndrejMitrovic:
On my machine this brings the test-suite total run time down from 23 seconds to just 11 seconds.

Requires https://github.com/bpfkorea/agora/pull/399, then I'll rebase.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -65,6 +65,7 @@ shared static this()
 private UnitTestResult customModuleUnitTester ()
 {
     import std.algorithm;
+    import std.parallelism;
     import std.process;
     import std.stdio;
     import std.string;
@@ -74,7 +75,15 @@ private UnitTestResult customModuleUnitTester ()
     auto filter = environment.get("dtest").toLower();
     size_t filtered;
 
-    UnitTestResult result;
+    // can't use ModuleInfo[], opApply returns temporaries..
+    struct ModTest
+    {
+        string name;
+        void function() test;
+    }
+
+    ModTest[] mod_tests;
+
     foreach (ModuleInfo* mod; ModuleInfo)
     {
         if (mod is null)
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
> On my machine this brings the test-suite total run time down from 23 seconds to just 11 seconds.

~Interestingly, the Github CI run time remained the same. I wonder if they're limiting tests to only one core..~

Edit: Ah I did forget that most of the time is spent fetching dependencies. So this PR will have a better impact on developers but not much impact on the CI.

source/agora/test/Base.d
```diff
@@ -74,7 +75,15 @@ private UnitTestResult customModuleUnitTester ()
...
+    // can't use ModuleInfo[], opApply returns temporaries..
+    struct ModTest
+    {
+        string name;
+        void function() test;
+    }
+
+    ModTest[] mod_tests;
+
     foreach (ModuleInfo* mod; ModuleInfo)
```
AndrejMitrovic, line 87: So, interesting thing about this. You can't store this pointer (or its value), if you try to store it and use it outside the loop, it crashes. Strange implementation of this feature..

AndrejMitrovic:
Unblocked.

Geod24:
Won't it be a mess if multiple unittests fail and have outputs tho ?

AndrejMitrovic:
Good point! As a simple solution I could use a global lock in the printLogs function.

source/agora/test/Base.d
```diff
@@ -243,13 +258,16 @@ public class TestAPIManager
 
     public void printLogs ()
     {
-        import std.stdio;
-        foreach (key, api; this.apis)
+        synchronized  // make sure logging output is not interleaved
```
Geod24, line 261: Shouldn't the function be marked `synchronized` directly ?

AndrejMitrovic: Hmm.. I didn't know you can use it like that. I'll read up on it.

AndrejMitrovic: Actually from what I can tell `synchronized` on a method means it will synchronize on the object itself. But there are multiple `TestAPIManager` instances, one per unittest. I want to synchronize globally.

Geod24: What does it synchronize on without an object ?

AndrejMitrovic: As far as I remember from TDPL, a global mutex, and I think each synchronized{} block has its own mutex.

AndrejMitrovic: https://dlang.org/spec/statement.html#SynchronizedStatement

> What mutex is used is determined by the Expression. If there is no Expression, then a global mutex is created, one per such synchronized statement. Different synchronized statements will have different global mutexes.

Geod24 (approved):

PR #403, opened 2019-10-24
[do not merge] Try to trigger & log ldc failure

AndrejMitrovic:
I'm gonna try to re-trigger this until Travis fails.

The log Mathias showed me tells me that the nodes had the same block height but actually had different blockchains (yay hard fork).


AndrejMitrovic:
This will be hard to trigger due to timing issues, but I'm pretty sure https://github.com/bpfkorea/agora/pull/404 fixes it.

PR #404, opened 2019-10-24
Fix the NetworkManager BadNode test

AndrejMitrovic:
This test was originally written before we had any block validation, so back then only the `height` was checked. But it still had issues with timing.

I've simplified the test, because all it's supposed to test is sending bad blocks in `getBlocksFrom`.

### DIFF
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -269,8 +269,12 @@ public class NetworkManager
             do
             {
                 auto blocks = pair.client.getBlocksFrom(block_height, MaxBlocks);
-                log.info("Received blocks [{}..{}] out of {}..", block_height,
-                    block_height + blocks.length, pair.height + 1);  // genesis block
+                if (blocks.length == 0)
+                    continue LNextNode;
+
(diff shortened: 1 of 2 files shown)


source/agora/network/NetworkManager.d
```diff
@@ -269,8 +269,9 @@ public class NetworkManager
             do
             {
                 auto blocks = pair.client.getBlocksFrom(block_height, MaxBlocks);
-                log.info("Received blocks [{}..{}] out of {}..", block_height,
-                    block_height + blocks.length, pair.height + 1);  // genesis block
+                log.info("Received blocks [{}..{}] out of {}..",
+                    blocks[0].header.height, blocks[$ - 1].header.height,
+                    pair.height + 1);  // +1 for genesis block
```
Geod24, line 277: Potential DoS: Returning an empty range

Geod24 (changes requested):

source/agora/network/NetworkManager.d
AndrejMitrovic: Good catch, fixed.

Geod24 (approved):

Issue #308, opened 2019-09-06
Write an alternative to `std.concurrency`


Geod24:
As a developer, I need to have a reliable way to do message passing between threads and spawn tasks in order for our tests and our nodes to work correctly.

Currently, `std.concurrency` is a major source of problem. The API and design is appealing, but it has little support for local event loop (even though it exposes types that should allow it).
We need an alternative way to do message passing, possibly something similar to Go channel.

See https://docs.google.com/document/d/1yIAYmbvL3JxOKOjuCyon7JhW4cSv1wy5hC0ApeGMV9s/pub

Also: David Nadlinger talked about Fibers at DConf 2015: https://www.youtube.com/watch?v=WzXe2kT9sEo
One of the item discussed was how we can't move Fibers accross Threads. We should look at what Vibe.d does in that regard.

Geod24:
Moved to LocalRest.

PR #405, opened 2019-10-28
Implement an alternative to `std.concurrency`

MichaelKim20:
I implemented a `channel` to resolve the issue.
A channel is a communication class using which fiber can communicate with each other.
Technically, a channel is a data transfer pipe where data can be passed into or read from.
Hence one fiber can send data into a channel, while other fiber can read that data from the same channel.
And, this channel allows communication between thread and fiber.

Issues is #308 Write an alternative to `std.concurrency`

### DIFF
--- a/source/agora/common/Channel.d
+++ b/source/agora/common/Channel.d
@@ -1,0 +1,732 @@
+/*******************************************************************************
+
+    This file contains the implementation of channels.
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.Channel;
+
+import agora.common.Queue;
+
+import std.container;
+import std.range;
+
+import core.sync.mutex;
+import core.thread;
+
+/*******************************************************************************
+
+    A interface of channel
+
+*******************************************************************************/
+
+public interface Channel (T)
+{
+    /// Send data `elem`.
+    public bool send (T elem);
+
+    /// Write the data received in `elem`
+    public bool receive (T* elem);
+}
+
+/*******************************************************************************
+
+    This channel use `NonBlockingQueue`. This is not uses `Lock`
+
+*******************************************************************************/
+
+public class NonBlockingChannel (T) : Channel!T
+{
+    /// Non blocking queue
+    private NonBlockingQueue!T queue;
+
+    /// Ctor
+    public this ()
+    {
+        this.queue = new NonBlockingQueue!T;
+    }
+
+    /***************************************************************************
+
+        Send data `elem`.
+
+        Params:
+            elem = value to send
+
+        Return:
+            true if the sending is successful, otherwise false
+
+    ***************************************************************************/
+
+    public bool send (T elem)
+    {
+        this.queue.enqueue(elem);
+        return true;
+    }
+
+    /***************************************************************************
+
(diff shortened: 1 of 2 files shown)


Geod24:
I think it would be good to implement some benchmarking for this

Geod24:
Looks like this implement the Go semantics ? Which is not bad in itself, but I think the different modes, especially the differences between buffered and non-buffered channels, should be described.

source/agora/common/Channel.d
```diff
@@ -0,0 +1,535 @@
...
+            SudoFiber!T new_sf;
+            new_sf.fiber = f;
+            new_sf.elem_ptr = null;
+            new_sf.elem = elem;
+            this.sendq.insertBack(new_sf);
+
+            this.mutex.unlock();
+
+            while (is_waiting)
+                Thread.sleep(dur!("msecs")(1));
```
Geod24: Did you test this code path ? I don't know how it could even work, the control is never passed to `f`.

source/agora/common/Channel.d
```diff
@@ -0,0 +1,535 @@
...
+
+        if (this.queue[].walkLength < this.qsize)
+        {
+            this.queue.insertBack(elem);
+            this.mutex.unlock();
+            return true;
+        }
+
+        Fiber fiber = Fiber.getThis();
+        if (fiber is null)
```
Geod24: Why is this creating a `Fiber` ? It should be possible to send a message between 2 threads without relying on Fibers.

source/agora/common/Channel.d
```diff
@@ -0,0 +1,535 @@
...
+
+        Params:
+            elem = value to send
+
+        Return:
+            true if the sending is successful, otherwise false
+
+    ***************************************************************************/
+
+    public bool send (T elem)
```
Geod24, line 189:
```suggestion
    public bool send (auto ref T elem)
```

source/agora/common/Channel.d
```diff
@@ -0,0 +1,535 @@
...
+    {
+        SudoFiber!T sf;
+        bool res;
+
+        synchronized (this.mutex)
+        {
+            this.closed = true;
+        }
+
+        while (true)
```
Geod24, line 378: This code duplication is a prime candidate for local functions

source/agora/common/Channel.d
MichaelKim20: When 'receive' is called elsewhere
It is executed when 'fiber' saved in `sendq` is 'call'.

MichaelKim20:
Two kinds of channels have been implemented.
NonBlockingChannel with Lock-free queue and WaitableChannel with queue.

**NonBlockingChannel**
This channel use `NonBlockingQueue`. This is not uses `Lock`

**WaitableChannel**
This channel has queues that senders and receivers can wait for.
With these queues, a single thread alone can exchange data with each other.

In single thread
Call 'receive' first. Then the `Fiber` will register in the queue.
Next, when the send is called, the `Fiber` in the queue receives the data.
Without a queue, it's impossible on a single thread.  (It can be confirmed at `unittest`.)

If the size of the data buffer is zero,
It blocks after calling the send. (It can be confirmed at `unittest`.)
To solve the blocking, there is a way to generate
and register Fiber from the Thread.

If the size of the data buffer is zero, If you call it using a fiber, it won't block. (It can be confirmed at `unittest`.)

Geod24:
Moved to https://github.com/Geod24/localrest/pull/25

PR #415, opened 2019-11-07
Add serialization support for various Crypto types

AndrejMitrovic:
Also adds a serialization test for the Block.

### DIFF
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -17,7 +17,10 @@
 
 module agora.common.crypto.ECC;
 
+import agora.common.Deserializer;
 import agora.common.Hash;
+import agora.common.Serializer;
+
 import geod24.bitblob;
 import libsodium;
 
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
There seem to be weird MacOS failures on GithubCI, something to do with missing symbols. And I just got a related e-mail from Github saying they are upgrading to MacOS Catalina, but we can still use Mojave. I wonder if they screwed something up. :D

Geod24:
> And I just got a related e-mail from Github saying they are upgrading to MacOS Catalina, but we can still use Mojave. I wonder if they screwed something up. :D

No they didn't, but it means we can't use < 2.087.1 because of [this](https://dlang.org/changelog/2.087.1.html#darwin-loader).

Geod24:
> No they didn't

Actually scratch that. It shouldn't fail on 10.14.

Geod24:
At this point... Should we just remove the restriction ?

AndrejMitrovic:
Restriction on what?

Geod24:
The serialization requiring to implement `serialize` instead of just `tupleof`ing stuff.

source/agora/common/crypto/ECC.d
```diff
@@ -84,6 +92,18 @@ public struct Scalar
...
     }
 
+    /// Ditto
+    public string toString () const @safe
+    {
+        return this.data.toString();
+    }
+
+    /// Vibe.d deserialization
+    static Scalar fromString (scope const(char)[] str) @safe
```
```suggestion
    public static Scalar fromString (scope const(char)[] str) @safe
```

source/agora/common/crypto/ECC.d
```diff
@@ -162,6 +223,18 @@ public struct Point
...
     }
 
+    /// Ditto
+    public string toString () const @safe
+    {
+        return this.data.toString();
+    }
+
+    /// Vibe.d deserialization
+    static Point fromString (scope const(char)[] str) @safe
```
```suggestion
    public static Point fromString (scope const(char)[] str) @safe
```

Geod24 (approved):
LGTM minus nits

AndrejMitrovic:
Fixed missing `public` attributes. Once it's green, you can merge it @TrustHenry.

PR #416, opened 2019-11-07
Implement a BitField structure

AndrejMitrovic:
There is one problem: We have to rename the `init` method in ocean's `BitArray` into something else like `initialize`, because having `.init` with parameters causes compiler errors in generic code (specifically vibe.d serialization).

So we need to either fix this upstream, or in our fork. I am not sure if `BitArray` was ever used at all in any Sociomantic code (because it lacks all types of attributes like nothrow / @safe / etc), so it might be safe to fix this upstream.

~Requires https://github.com/bpfkorea/agora/pull/415~
Requires https://github.com/sociomantic-tsunami/ocean/issues/781 to be resolved.

### DIFF
--- a/source/agora/common/BitField.d
+++ b/source/agora/common/BitField.d
@@ -1,0 +1,190 @@
+/*******************************************************************************
+
+    Contains a BitField implementation based on Ocean's BitArray struct.
+
+    The choice of Ocean's over Phobos' implementation of BitArray is
+    because the Phobos implementation requires the backing store length to be
+    a multiple of size_t, which makes it platform-specific and hard to code for.
+    This structure cannot be platform-specific as it's stored in the blockchain.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.BitField;
+
+import agora.common.Deserializer;
+import agora.common.Hash;
+import agora.common.Serializer;
+
+import ocean.core.BitArray;
+
+import std.algorithm;
+import std.math;
+
+/// Ditto
+public struct BitField
+{
+    /// Bitfield implementation
+    public BitArray bit_array;
+
+    /// Expose the bitfield API
+    public alias bit_array this;
+
+    /// Backing store (BitArray requires integer arrays)
+    private uint[] storage;
+
+    /// The number of bits used (workaround for: BitArray.length() is not @nogc)
+    private size_t num_bits;
+
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            num_bits = the number of bits to be used.
+
+    ***************************************************************************/
+
+    public this ( size_t num_bits ) @trusted
+    {
+        // BitArray uses ints as storage, we have to use multiples of 32
+        this.storage.length = (num_bits + 31) / 32;
+        this.num_bits = num_bits;
+        this.bit_array.initialize(this.storage, num_bits);
+    }
+
+    /***************************************************************************
+
+        Serialization. The data is packed tightly by
+        removing unused bytes from the uint[] storage,
+        and the length of bits is encoded instead of
+        the length of bytes.
+
+        Params:
+            dg = serialize function accumulator
+
+    ***************************************************************************/
+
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
The Phobos implementation uses `size_t[]` as its backing store, making it more complex to code for (making it platform-specific I mean). So I opted to use the Ocean implementation instead. The Phobos version might be more battle-tested though, so I'm reconsidering my choice..

AndrejMitrovic:
Unblocked and ready for review.

AndrejMitrovic:
Update: I've simplified the PR. Previously it was doing byte-level optimizations, but on second thought these are overkill. We can optimize later and just keep it simple and correct now.

source/agora/common/BitField.d
```diff
@@ -0,0 +1,191 @@
+/*******************************************************************************
+
+    Contains a BitField implementation based on Ocean's BitArray struct.
```
Geod24, line 3:
```suggestion
    Contains a BitField implementation based on Ocean's BitArray struct.
    
```

source/agora/common/BitField.d
```diff
@@ -0,0 +1,191 @@
...
+    The choice of Ocean's over Phobos' implementation of BitArray is
+    because the Phobos implementation requires the backing store length to be
+    a multiple of size_t, which makes it platform-specific and hard to code for.
```
Geod24, line 7: Maybe mention that this structure cannot be platform specific because it is stored in the block. Platform specific behavior is not bad if it's not used in data that is exchanged. I'd guess the platform-specifism of Phobos is to make it more performant.

source/agora/common/BitField.d
```diff
@@ -0,0 +1,191 @@
...
+import agora.common.Hash;
+import agora.common.Serializer;
+
+import ocean.core.BitArray;
+
+import std.algorithm;
+import std.math;
+
+/// Ditto
+struct BitField
```
Geod24:
```suggestion
public struct BitField
```

source/agora/common/BitField.d
```diff
@@ -0,0 +1,191 @@
...
+
+        Params:
+            dg = serialize function accumulator
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) const @trusted
+    {
+        serializePart(this.num_bits, dg);
+        serializePart(cast(ubyte[])this.storage, dg);
```
Geod24: This breaks on different endianness

Geod24 (commented):
A few comments. Also, since the size is dependent on the number of validators, we probably could optimize serialization at the block level, since we will repeat this information.

source/agora/common/BitField.d
AndrejMitrovic: > I'd guess the platform-specifism of Phobos is to make it more performant.

They probably used size_t without giving it a second thought.

source/agora/common/BitField.d
AndrejMitrovic: It's a shame we don't have a big endian system in the CI then.

AndrejMitrovic: Are the existing serialization routines safe w.r.t. endianness?

AndrejMitrovic: For example:

```D
dg((cast(ubyte*)&record)[0 .. long.sizeof]);
```

I think this isn't safe?

Geod24: > Are the existing serialization routines safe w.r.t. endianness?

Probably not.

> It's a shame we don't have a big endian system in the CI then.

I *think* we could set a MIPS one but it'll take and it has lots of dependency, so let's leave that fun for later.

AndrejMitrovic:
Some previous attempt at byte-level optimization: https://github.com/AndrejMitrovic/agora/blob/old-bitfield-struct/source/agora/common/BitField.d

Could roll it in later, but I wanted this PR to be simple.

Updated the PR and included `uint[]` serialization as a separate commit.

AndrejMitrovic:
~Hold on, I'm doing swapping on a const array. Of course that doesn't work. Not tested because of the `BigEndian` version block. Heheh.~

Fixed.

AndrejMitrovic:
Fixed.

Geod24 (approved):

PR #417, opened 2019-11-07
Add the validators and signature to the BlockHeader

AndrejMitrovic:
Requires:

- ~https://github.com/bpfkorea/agora/pull/415~
- ~https://github.com/bpfkorea/agora/pull/416~

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -16,7 +16,9 @@
 module agora.consensus.data.Block;
 
 import agora.common.Amount;
+import agora.common.BitField;
 import agora.common.crypto.Key;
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
@Geod24 the validators sign off on the hash of the block header, right? But the block header hash does not include the validator signature (it can't, it's self-referential). 

So I'm thinking, the validators should sign not just on the block header alone, but instead sign on `H(block_header | prev_block.signatures)`.

Does that make sense?

AndrejMitrovic:
Unblocked & ready for review.

TrustHenry (approved):

AndrejMitrovic:
Fixing integration test now.

Geod24 (approved):

Issue #406, opened 2019-10-28
Design and implement enrollment process


Geod24:
As a developer, I need the enrollment process described in the yellow paper to be implemented in order to build other issues (transaction filter, quorum balancing) on top of the random function.

Definition of done:
- [x] Nodes enroll by committing <K, X, N, R, S>
- [x] There is a function that derives randomness from `X`;
- [x] There is a function that derives randomness from the sum of all `X`;
- [ ] ~On each round, `X` is propagated and checked for validity (is it registered on the chain?);~
- [x] On each round, `R` is used for signature (see #365)

Description:
```
K: A hash of a frozen UTXO;
X: The nth image of their source of randomness;
N: A number within bounds [0; (Balance(K) % 40,000) * 2016 * 2 (tentative value)];
R: The initial R used for signing;
S: A schnorr signature for the message H(K, X, N, R) and the key K, using R.
```

Three approaches of using frozen UTXO in enrollment process
```
1) Disallow freezing if Public key ∈ Frozen set
  - Useful
2) Use everything
  - Unfreeze one(Flash!)
  - Complicates unfreezing
  - Harmless
3) Provide UTXO hash + idx on a set
  - Slight complication
  - Wanted
```

AndrejMitrovic:
Quick question: So when a node generates a random key, what it actually generates is a scalar n, and then it commits to the Nth hash of n, and also the public key generated from n. So basically our random value is a source for a point on the curve. Is this correct?

AndrejMitrovic:
> A signature for the message H(K, X, N, R) and the key K, using R.

Should this be using `r` (the private random key), rather than `R` (the public key)?

AndrejMitrovic:
I modified the issue to make it explicit: S is the schnorr signature of `H(K, X, N, R)`, just like in https://github.com/bpfkorea/agora/blob/a0e12c8abff3269293ab61e652c5ec68627ee02e/source/agora/common/crypto/Schnorr.d#L154-L158

AndrejMitrovic:
> Nodes enroll by committing <K, X, N, R, S>

The `R` is already available in the signature S, no? In [here](https://github.com/bpfkorea/agora/blob/a0e12c8abff3269293ab61e652c5ec68627ee02e/source/agora/common/crypto/Schnorr.d#L96). It seems like it doesn't have to be included separately in the enrollment transaction, unless I'm missing something?

Geod24:
That's correct.

linked0:
[...]

Issue #240, opened 2019-08-23
Design and implement quorum balancing


bpalaggi:
As a developer, I need to know what algorithm to follow for quorum assignment in order for the protocol to prevent censorship while being open.

Description

We can consider two degenerate cases for quorum balancing:
- When there is a single quorum with all nodes in it, safety is maximized, but so is communication overhead;
- When each node live in their own quorum, safety does not exists but there is no communication: a less degenerate version would be having a ring where every node belongs to two quorums, each with another node (or: two two nodes quorums).

The goal of quorum balancing is to produce a network graph with optimal safety & overhead. Note that the nodes that are quorum intersections are likely to be higher-staking nodes, as they have more to lose.

*Example* of rules:
- There is no quorum with < 3 nodes;
- There is no quorum with > 7 nodes;
- A quorum economic value must be >= 350,000 BOA;
- The top X (X=5?) stakers are placed in different quorums and used as 'seed';

Definition of done:
- [x] A new validator is assigned a quorum immediately;
- [x] Quorums get regularly shuffled (how often?);
- [x] There is a sufficient (how much?) quorum intersection;
- [x] Validators are distributed among quorums in an unpredictable manner;
- [x] Validators are distributed among quorums in a fairly balanced way;
- [x] Quorum distribution only depends on globally available state;

Tentative approach:
- Come up with a simple algorithm based on the example rules posted above;
- Build test case for working and non-working (byzantine) validators;
- Compute the economic penalty (low & high bounds) => If node A cheats, home much does it cost ? Is it aligned with the power this node has in the system ?;
- Optimize
- Repeat

Geod24:
Depends on #209

AndrejMitrovic:
`- [ ] Validators are distributed among quorums in an unpredictable manner;`

I think this sentence is not entirely correct. It should be predictable given another number X, which itself is unpredictable (and based on the preimage). So the algorithm to select the quorum is predictable, but its inputs are not predictable.

AndrejMitrovic:
> Quorums get regularly shuffled (how often?);

According to the yellow paper: `Quorum balancing events happens once every 6 rounds.`. So, every hour.

AndrejMitrovic:
I would actually add that past performance of a validator should be taken into account when assigning quorums. For example:

```
The top X (X=5?) stakers are placed in different quorums and used as 'seed';
```

[...]

PR #419, opened 2019-11-13
Add structures for the enrollment process

linked0:
- Enrollment: Defines data for enrollment process
- EnrollmentManager: Handles data and process for enrollment
- unittest code

#406

### DIFF
--- a/source/agora/common/EnrollmentManager.d
+++ b/source/agora/common/EnrollmentManager.d
@@ -1,0 +1,353 @@
+/*******************************************************************************
+
+    Contains supporting code for enrollment process.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.EnrollmentManager;
+
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
+import agora.common.crypto.Schnorr;
+import agora.common.Deserializer;
+import agora.common.Hash;
+import agora.common.Serializer;
+import agora.consensus.data.Enrollment;
+import agora.utils.Log;
+
+import d2sqlite3.database;
+import d2sqlite3.library;
+import d2sqlite3.results;
+import d2sqlite3.sqlite3;
+
+import std.algorithm;
+import std.file;
+import std.path;
+
+mixin AddLogger!();
+
+/*******************************************************************************
+
+    Handle enrollment data and manage the validators set
+
+*******************************************************************************/
+
+public class EnrollmentManager
+{
+    /// SQLite db instance
+    private Database db;
+
+    /// Node's key pair
+    private Pair key_pair;
+
+    /// Random seed
+    private Scalar random_seed_src;
+
+    /// Random key for enrollment
+    private Pair signature_noise;
+
+    /// Enrollment data object
+    private Enrollment data;
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            db_path = path to the database file, or in-memory storage if
+                        :memory: was passed
+            key_pair = the keypair of the owner node
+
+    ***************************************************************************/
+
+    public this (string db_path, KeyPair key_pair)
+    {
+        this.db = Database(db_path);
+
+        // create the table for validator set if it doesn't exist yet
+        this.db.execute("CREATE TABLE IF NOT EXISTS validator_set " ~
+            "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
+
+        // create the table for enrollment data for a node itself
+        this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
+            "(key CHAR(128) PRIMARY KEY, val BLOB NOT NULL)");
+
+        // set node's KeyPair object
+        this.key_pair.v = Scalar(key_pair.secret);
+        this.key_pair.V = this.key_pair.v.toPoint();
+
+        // load signature noise
+        auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
+            "WHERE key = ?", "signature_noise");
+
+        foreach (row; results)
+        {
+            signature_noise = deserializeFull!Pair(row.peek!(ubyte[])(0));
+            break;
+        }
+    }
+
+    /***************************************************************************
+
+        Shut down the database
+
+        Note: this method must be called explicitly, and not inside of
+        a destructor.
+
+    ***************************************************************************/
+
+    public void shutdown ()
+    {
+        this.db.close();
+    }
+
+    /***************************************************************************
+
+        Add a enrollment data to the validators set
+
+        Params:
+            enroll = the enrollment data to add
+
+        Returns:
+            true if the enrollment data has been added to the validator set
+
+    ***************************************************************************/
+
+    public bool addEnrollment (Enrollment enroll) @safe
+    {
+        static ubyte[] buffer;
+        buffer.length = 0;
+
+        // check if already exists
+        if (this.hasEnrollment(enroll.utxo_key))
+        {
+            log.info("Rejected already existing enrollment: {}", enroll);
+            return false;
+        }
+
+        () @trusted { assumeSafeAppend(buffer); } ();
+
+        scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
+        {
+            buffer ~= data;
+        };
+
+        serializePart(enroll, dg);
+
+        () @trusted {
+            this.db.execute("INSERT INTO validator_set (key, val) VALUES (?, ?)",
+                enroll.utxo_key[], buffer);
+        }();
+
+        return true;
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the number of validators in the validator set
+
+    ***************************************************************************/
+
+    public size_t getEnrollmentLength () @safe
+    {
+        return () @trusted {
+            return this.db.execute("SELECT count(*) FROM validator_set").oneValue!size_t;
+        }();
+    }
+
+    /***************************************************************************
+
+        Remove the enrollment data with the given key from the validator set
+
+        Params:
+            enroll_hash = key for an enrollment data to remove
+
+    ***************************************************************************/
+
+    public void removeEnrollment (const ref Hash enroll_hash) @trusted
+    {
+        this.db.execute("DELETE FROM validator_set WHERE key = ?", enroll_hash[]);
+    }
+
+    /***************************************************************************
+
+        Make an enrollment data for enrollment process
+
+        Params:
+            frozen_utxo_hash = the hash of a frozen UTXO used to identify 
+                                a validator and to generate a siging key
+
+        Returns:
+            an enrollment data
+
+    ***************************************************************************/
+
+    public ref const(Enrollment) createEnrollment (const ref Hash frozen_utxo_hash) @trusted
+    {
+        static ubyte[] buffer;
+        buffer.length = 0;
+
+        // K, frozen UTXO hash
+        this.data.utxo_key = frozen_utxo_hash;
+
+        // N, cycle length
+        this.data.cycle_length = 1008; // freezing period / 2
+
+        // generate random seed value
+        this.random_seed_src = Scalar.random();
+
+        // X, nth image of random seed
+        this.data.random_seed = hashFull(this.random_seed_src);
+        foreach (i; 0 .. this.data.cycle_length-1)
+            this.data.random_seed = hashFull(this.data.random_seed);
+
+        // R, signature noise
+        this.signature_noise = Pair.random();
+        () @trusted { assumeSafeAppend(buffer); }();
+
+        scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
+        {
+            buffer ~= data;
+        };
+
+        serializePart(this.signature_noise, dg);
+        
+        () @trusted {
+            auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM node_enroll_data " ~
+                "WHERE key = ?)", "signature_noise");
+            if (results.oneValue!(bool))
+            {
+                this.db.execute("UPDATE node_enroll_data SET val = ? WHERE key = ?",
+                    buffer, "signature_noise");
+            }
+            else
+            {
+                this.db.execute("INSERT INTO node_enroll_data (key, val) VALUES (?, ?)",
+                    "signature_noise", buffer);
+            }
+        }();
+
+        // signature
+        data.enroll_sig = sign(this.key_pair.v, this.key_pair.V, this.signature_noise.V,
+            this.signature_noise.v, this.data);
+
+        return this.data;
+    }
+
+    /***************************************************************************
+
+        Check if the enrollment data exists in the validator set.
+
+        Params:
+            enroll_hash = key for an enrollment data which is hash of frozen UTXO 
+
+        Returns:
+            true if the enrollment manager has the enrollment data
+
+    ***************************************************************************/
+
+    public bool hasEnrollment (const ref Hash enroll_hash) @trusted
+    {
+        auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM validator_set " ~
+            "WHERE key = ?)", enroll_hash[]);
+
+        return results.front().peek!bool(0);
+    }
+    
+    /***************************************************************************
+
+        Check if the enrollment data exists in the validator set.
+
+        Params:
+            enroll_hash = key for an enrollment data which is a hash of a frozen 
+                            UTXO
+            enroll = will contain the enrollment data if found
+
+        Returns:
+            Return true if the enrollment data was found
+
+    ***************************************************************************/
+
+    public bool getEnrollment (const ref Hash enroll_hash,
+        out Enrollment enroll) @trusted
+    {
+        auto results = this.db.execute("SELECT key, val FROM validator_set " ~
+            "WHERE key = ?", enroll_hash[]);
+
+        foreach (row; results)
+        {
+            enroll = deserializeFull!Enrollment(row.peek!(ubyte[])(1));
+            return true;
+        }
+
+        return false;
+    }
+}
+
+/// tests for member functions of EnrollmentManager
+unittest
+{
+    import agora.common.Amount;
+    import agora.consensus.data.Transaction;
+    import agora.consensus.data.UTXOSet;
+    import std.format;
+
+    KeyPair[] key_pairs = [KeyPair.random];
+
+    // create the first transaction
+    Transaction firstTx = Transaction(
+        TxType.Freeze,
+        [Input(Hash.init, 0)],
+        [Output(Amount(100_000), key_pairs[0].address)]
+    );
+    Hash firstHash = hashFull(firstTx);
+
+    // create the second transaction
+    Transaction secondTx = Transaction(
+        TxType.Freeze,
+        [Input(Hash.init, 0)],
+        [Output(Amount(50_000), key_pairs[0].address)]
+    );
+    Hash secondHash = hashFull(secondTx);
+
+    // create an EnrollmentManager object
+    auto utxo_set = new UTXOSet(":memory:");
+    scope (exit) utxo_set.shutdown();
+    auto man = new EnrollmentManager(":memory:", key_pairs[0]);
+    scope (exit) man.shutdown();
+
+    // create and add the first Enrollment object
+    auto utxo_hash = utxo_set.getHash(firstHash, 0);
+    auto enroll = man.createEnrollment(utxo_hash);
+    assert(man.hasEnrollment(utxo_hash) == false);
+    man.addEnrollment(enroll);
+    assert(man.getEnrollmentLength() == 1);
+    assert(man.hasEnrollment(utxo_hash) == true);
+    assert(man.addEnrollment(enroll) == false);
+
+    // create and add the second Enrollment object
+    auto utxo_hash2 = utxo_set.getHash(secondHash, 0);
+    auto enroll2 = man.createEnrollment(utxo_hash2);
+    man.addEnrollment(enroll2);
+    assert(man.getEnrollmentLength() == 2);
+
+    // get a stored Enrollment object
+    Enrollment stored_enroll;
+    assert(man.getEnrollment(utxo_hash2, stored_enroll));
+    assert(stored_enroll == enroll2);
+
+    // remove an Enrollment object
+    man.removeEnrollment(utxo_hash2);
+    assert(man.getEnrollmentLength() == 1);
+
+    // test for getEnrollment with removed enrollment
+    assert(!man.getEnrollment(utxo_hash2, stored_enroll));
+}
--- a/source/agora/consensus/data/Enrollment.d
+++ b/source/agora/consensus/data/Enrollment.d
@@ -1,0 +1,39 @@
+/*******************************************************************************
+
+    Contains supporting code for a enrollment data used in enrollment process.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.Enrollment;
+
+import agora.common.Types;
+import agora.common.crypto.Schnorr : Signature;
+
+/*******************************************************************************
+
+    Define enrollment data (enrollment data = validator)
+
+*******************************************************************************/
+
+public struct Enrollment
+{
+    /// K: UTXO hash, A hash of a frozen UTXO
+    public Hash utxo_key;
+
+    /// X: random seed, The nth image of random value
+    public Hash random_seed;
+
+    /// n: cycle length, the number of rounds a validator will participate in 
+    /// (currently fixed to (freezing period / 2)
+    public uint cycle_length;
+
+    /// S: A signature for the message H(K, X, n, R) and the key K, using R
+    public Signature enroll_sig;
+}


source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,138 @@
...
+
+import agora.common.Amount;
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
+import agora.common.Hash;
+import agora.common.Types;
+import agora.common.crypto.Schnorr : Pair;
+
+import std.random;
+import std.stdio;
```
TrustHenry: `import std.stdio;`
Not used here.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,138 @@
...
+
+/// Make EnrollmentData
+unittest
+{
+    KeyPair kp = KeyPair.random();
+    auto man = new EnrollmentManager(kp, Amount(10000));
+    auto enroll_data = man.getEnrollmentData();
+    auto enroll_data2 = man.getEnrollmentData();
+    assert(enroll_data.S != enroll_data2.S);
+}
```
TrustHenry: An Enter required in the last line.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,138 @@
...
+    ***************************************************************************/
+    public EnrollmentData getEnrollmentData()
+    {
+        this.data = new EnrollmentData();
+        
+        // public key
+        this.data.K = this.node_keypair.address;
+
+        // cycle length
+        this.data.n = uniform(0, 1008, this.genint);
```
AndrejMitrovic: The number of voting rounds the node participates in should not be random. It should be set by the node operator.

Currently the Yellow paper says:

> 3. Enrollment process
> n (cycle length): the number of rounds a validator will participate in (currently fixed to (freezing period / 2)];

For now, I suggest making N a constructor parameter for `EnrollmentManager`.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,138 @@
...
+        // signature
+        Hash hash = hashFull(*this.data);
+        this.data.S = this.node_keypair.secret.sign(hash[]);
+
+        return *this.data;
+    }
+
+    private static void round (ref Hash image, size_t n)
+    {
+        image = hashFull(image);
```
AndrejMitrovic: This can be replaced with a foreach loop.

```D
foreach (i; 0 .. n)
    image = hashFull(image);
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,138 @@
...
+public struct EnrollmentData{
+    /// K: UTXO key, A public key matching a frozen UTXO
+    public PublicKey K;
+
+    /// X: random seed, The nth image of random value
+    public Hash X;
+
+    /// n: cycle length, A number within bounds [0; 1008]
+    /// original spec: [0; (Balance(K) % 40,000) * freezing period * 2]
+    public ulong n;
```
AndrejMitrovic: This can be `uint`

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,138 @@
...
+        this.data = new EnrollmentData();
+        
+        // public key
+        this.data.K = this.node_keypair.address;
+
+        // cycle length
+        this.data.n = uniform(0, 1008, this.genint);
+
+        // random seed
+        this.data.X = hashFull(this.node_keypair.secret.data);
```
AndrejMitrovic: This is the part that should use the random number generator.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,138 @@
...
+        Constructor
+        Params:
+            keypair = the keypair of the owner node
+            balance = account balance of the owner node
+    ***************************************************************************/
+    public this(KeyPair keypair, Amount balance)
+    {
+        this.node_keypair = keypair;
+        this.balance = balance;
+        this.genint = Random(123);
```
AndrejMitrovic: This is not a safe way to use a random-number generator. The `123` part is the seed. The sequence of numbers you get out of a random-number generator is always the same for the given seed.

Instead, I suggest using `randombytes_uniform()` from libsodium to generate a number within a given range.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,138 @@
...
+        this.data.X = hashFull(this.node_keypair.secret.data);
+        round(this.data.X, this.data.n - 1);
+
+        // signature noise
+        this.random_key = Pair.random();
+        this.data.R = this.random_key.V;
+
+        // signature
+        Hash hash = hashFull(*this.data);
+        this.data.S = this.node_keypair.secret.sign(hash[]);
```
AndrejMitrovic: The signature part actually has to be a schnorr signature using `r` (the `random_key.v` scalar) above. If the enrollment is not signed with `r`, then you can't prove that you actually know `R`. :)

https://github.com/bpfkorea/agora/blob/5ef14fe0ce879da52cc47716d7688bf5ab0d943f/source/agora/common/crypto/Schnorr.d#L196

^ This function can be used. It's currently private, but I'll make it public in a PR in a moment.

source/agora/node/EnrollmentManager.d
linked0: Thank you for your comment.
I'll change the number as described in the yellow paper, but I'd like to set the number inside the class not by the parameter of the constrcutor because the setting the number can be a burden to other classes such as the Node class. By a burden, I mean determining which number should be set. 
Any opinion about this would be helpful.

AndrejMitrovic: Ok, you can keep this internal for now.

linked0:
Ready for review.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,130 @@
...
+import agora.common.crypto.Schnorr : Signature;
+import agora.common.crypto.Schnorr : sign;
+
+import std.random;
+
+/*******************************************************************************
+    Defines data for enrollment process
+*******************************************************************************/
+
+public struct EnrollmentData{
```
MichaelKim20:
```suggestion
public struct EnrollmentData
{
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,130 @@
...
+    /// EnrollmentData
+    private EnrollmentData* data;
+
+    /***************************************************************************
+        Constructor
+        Params:
+            keypair = the keypair of the owner node
+            balance = account balance of the owner node
+    ***************************************************************************/
+    public this(KeyPair keypair, Amount balance)
```
MichaelKim20:
```suggestion
    public this (KeyPair keypair, Amount balance)
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,130 @@
...
+        this.balance = balance;
+        this.data = null;
+    }
+
+    /***************************************************************************
+        Makes a data object for enrollment process
+        Returns:
+            a EnrollmentData object
+    ***************************************************************************/
+    public EnrollmentData createEnrollmentData()
```
MichaelKim20:
```suggestion
    public EnrollmentData createEnrollmentData ()
```

source/agora/common/crypto/ECC.d
```diff
@@ -308,6 +308,19 @@ public struct Point
...
+
+    /***************************************************************************
+
+        Implements hashing support
+
+        Params:
+            dg = Hashing function accumulator
+
+    ***************************************************************************/
+    public void computeHash(scope HashDg dg) const nothrow @safe @nogc
```
MichaelKim20:
```suggestion
    public void computeHash (scope HashDg dg) const nothrow @safe @nogc
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,130 @@
...
+module agora.node.EnrollmentManager;
+
+import agora.common.Amount;
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
+import agora.common.Hash;
+import agora.common.Types;
+import agora.common.crypto.Schnorr : Pair;
+import agora.common.crypto.Schnorr : Signature;
+import agora.common.crypto.Schnorr : sign;
```
MichaelKim20: **According to `Agora coding style`**
imports are organized by packages: agora first, dependencies second, std third, core last. Within packages, alphabetical order is prefered, with modules taking preference over packages, e.g.:

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,130 @@
...
+        this.node_keypair = keypair;
+        this.balance = balance;
+        this.data = null;
+    }
+
+    /***************************************************************************
+        Makes a data object for enrollment process
+        Returns:
+            a EnrollmentData object
+    ***************************************************************************/
```
MichaelKim20:
```
    
        Makes a data object for enrollment process
        
        Returns:
            a EnrollmentData object
            
    public EnrollmentData createEnrollmentData ()
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,130 @@
...
+
+    /// EnrollmentData
+    private EnrollmentData* data;
+
+    /***************************************************************************
+        Constructor
+        Params:
+            keypair = the keypair of the owner node
+            balance = account balance of the owner node
+    ***************************************************************************/
```
MichaelKim20:
```
    
        Constructor
        
        Params:
            keypair = the keypair of the owner node
            balance = account balance of the owner node
            
    public this (KeyPair keypair, Amount balance)
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,130 @@
+/*******************************************************************************
+    Contains supporting code for tracking the current ledger.
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+    License:
+        MIT License. See LICENSE for details.
+*******************************************************************************/
```
MichaelKim20:
```

    Contains supporting code for tracking the current ledger.

    Copyright:
        Copyright (c) 2019 BOS Platform Foundation Korea
        All rights reserved.

    License:

```

MichaelKim20 (changes requested):
Well done. I'd like to make some suggestions.
I reviewed it according to `Agora coding style`

source/agora/node/EnrollmentManager.d
AndrejMitrovic: The random seed should not be based on any existing data, it should ideally just be a random value. For example:

```D
Hash X = Scalar.random().hashFull();
```

The initial X should be stored somewhere locally so the Nth preimage can be generated from it on demand when a node needs to reveal that Nth preimage later. Either that, or you can just store an array of preimages for quick lookup later.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,426 @@
...
+        foreach (i; 0 .. this.data.n-1)
+            this.data.X = hashFull(this.data.X);
+
+        // signature noise
+        this.schnorr_pair = Pair.random();
+        this.data.R = this.schnorr_pair.V;
+
+        // signature
+        Hash hash = hashFull(*this.data);
+        this.data.S = sign(this.schnorr_pair, hash[]);
```
AndrejMitrovic: I don't think this is right. Here you are proving ownership of `R` by using `r` as a signing key, and randomly generating the actual R in the signature (see the `sign` method overload used here, it randomly generates R). But, this doesn't prove that you know `k` of `K`.

Instead I suggest signing it like so:

```D
auto pk = Scalar(this.node_keypair.seed[]);
this.data.S = sign(kp, kp.toPoint(), this.schnorr_pair.V, this.schnorr_pair.v, hash[]);
```

linked0:
Ready for review.

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,111 @@
...
+
+    Contains supporting code for a enrollment data used in enrollment process.
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
```
MichaelKim20:
```

    Contains supporting code for a enrollment data used in enrollment process.

    Copyright:
        Copyright (c) 2019 BOS Platform Foundation Korea
        All rights reserved.

    License:

```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,426 @@
+/*******************************************************************************
+
+    Contains supporting code for enrollment process.
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
```
MichaelKim20:
```

    Contains supporting code for enrollment process.

    Copyright:
        Copyright (c) 2019 BOS Platform Foundation Korea
        All rights reserved.

    License:

```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,426 @@
...
+
+    /***************************************************************************
+
+        Constructor
+        Params:
+            db_path = path to the database file, or in-memory storage if
+                        :memory: was passed
+            keypair = the keypair of the owner node
+
+    ***************************************************************************/
```
MichaelKim20:
```

        Constructor

        Params:
            db_path = path to the database file, or in-memory storage if
                        :memory: was passed
            keypair = the keypair of the owner node

```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,426 @@
...
+
+    /***************************************************************************
+
+        Check if the EnrollmentData is right
+
+    ***************************************************************************/
+
+    public bool acceptEnrollment(EnrollmentData enroll) @safe
+    {
+	    if (!this.add(enroll))
```
MichaelKim20:
```suggestion
        if (!this.add(enroll))
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,426 @@
...
+
+    }
+
+    /***************************************************************************
+
+        Check if the EnrollmentData is right
+
+    ***************************************************************************/
+
+    public bool acceptEnrollment(EnrollmentData enroll) @safe
```
MichaelKim20:
```suggestion
    public bool acceptEnrollment (EnrollmentData enroll) @safe
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,426 @@
...
+            enroll = the enrollment data hash
+
+        Returns:
+            true if the enrollment manager has the enrollment data hash
+
+    ***************************************************************************/
+
+    public bool hasEnrollmentDataHash (EnrollmentData enroll) @trusted
+    {
+	    auto enroll_hash = hashFull(enroll.K);
```
MichaelKim20:
```suggestion
        auto enroll_hash = hashFull(enroll.K); 
```

MichaelKim20 (changes requested):
I'd like to make some suggestions.

linked0:
Ready for review. Any suggestions are welcome.

source/agora/node/EnrollmentManager.d
AndrejMitrovic: any update on this?

linked0: Ah, Mathias commented that I could use the constant value, or 1008 at the moment before he left.

MichaelKim20 (approved):

source/agora/node/EnrollmentManager.d
AndrejMitrovic: I'm talking about the random seed value X, not N. We should use a random number generator, and not a hash of the private data.

AndrejMitrovic: Just to update everyone on this: Initially we planned on using the private key as the seed, which would have to be revealed at the end of the validation cycle. However, after @Geod24 had reservations (even though this part is still in the yellowpaper), so for now we can just make this a random number.

The reason why we need this X, is for the verifiable random number generation in the blockchain. This is important for many things: quorum balancing (https://github.com/bpfkorea/agora/issues/240), contracts (think games, lottery apps), etc.

linked0:
Ready for review.
I have already implemented the API endpoint on the Node class in preparation for next PR. The next PR will be opened in next Monday. I can apply other suggestions to those classes in the next PR.

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,113 @@
...
+import agora.common.Types;
+import agora.common.crypto.Schnorr : Pair;
+import agora.common.crypto.Schnorr : Signature;
+import agora.common.crypto.Schnorr : sign;
+
+/*******************************************************************************
+    Defines data for enrollment process
+*******************************************************************************/
+
+public struct EnrollmentData{
```
AndrejMitrovic:
```suggestion
public struct EnrollmentData
{
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,469 @@
...
+        Must be called once per process (not per thread!),
+        before the class can be used.
+
+    ***************************************************************************/
+
+    public static void initialize ()
+    {
+        static import core.atomic;
+        static shared bool initialized;
+        if (core.atomic.cas(&initialized, false, true))
```
AndrejMitrovic: The SQLite library is already initialized in another place. It shouldn't be initialized twice. I suggest removing this code.

source/agora/common/crypto/ECC.d
```diff
@@ -178,6 +178,20 @@ public struct Scalar
...
+    /***************************************************************************
+
+        Implements hashing support
+
+        Params:
+            dg = Hashing function accumulator
+
+    ***************************************************************************/
+
+    public void computeHash (scope HashDg dg) const nothrow @safe @nogc
```
AndrejMitrovic: This should be in a separate commit.

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,113 @@
...
+    public Hash X;
+
+    /// n: cycle length, the number of rounds a validator will participate in 
+    /// (currently fixed to (freezing period / 2)
+    public uint n;
+
+    /// R: signature noise, The initial nonce used for signing
+    public Point R;
+
+    /// S: A signature for the messge H(K, X, n, R) and the key K, using R
```
AndrejMitrovic:
```suggestion
    /// S: A signature for the message H(K, X, n, R) and the key K, using R
```

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,113 @@
...
+    public void deserialize (scope DeserializeDg dg) @safe
+    {
+        deserializePart(this.K, dg);
+        deserializePart(this.X, dg);
+        deserializePart(this.n, dg);
+        deserializePart(this.R, dg);
+        deserializePart(this.S, dg);
+    }
+
+    /// Support for sorting transactions
```
AndrejMitrovic: Why do you need to support sorting?

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,469 @@
...
+static this ()
+{
+    log = Log.lookup(__MODULE__);
+    EnrollmentManager.initialize();
+}
+
+/// Logger instance
+private Logger log;
+
+/*******************************************************************************
```
AndrejMitrovic:
```suggestion
/*******************************************************************************

```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,469 @@
...
+    Handles data and process for enrollment
```
AndrejMitrovic:
```suggestion
    Handles data and process for enrollment

```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,469 @@
...
+                return res;
+            };
+
+            this.random_seed.deserialize(dg);
+        }
+
+        // set information to member variables
+        this.db_path = db_path;
+        this.node_keypair = keypair;
+        this.data = null;
```
AndrejMitrovic:
```suggestion
```

Unnecessary, pointers are null-initialized.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,469 @@
...
+
+    ***************************************************************************/
+
+    public this (in string db_path, in string random_seed_path, KeyPair keypair)
+    {
+        const db_exists = db_path.exists;
+        if (db_exists)
+            log.info("Loading database from: {}", db_path);
+
+        // note: can fail. we may want to just recover txes from the network instead.
```
AndrejMitrovic: This was copy-pasted from existing code and does not apply here.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,469 @@
...
+        Callback used
+
+        Params:
+            arg = unused (null)
+            code = the error code
+            msg = the error message
+
+    ***************************************************************************/
+
+    private static extern(C) void loggerCallback (void *arg, int code,
```
AndrejMitrovic: This should be removed, the logger for SQLite is set once in another module.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,469 @@
...
+    {
+        // todo: optimize / cache this
+        return () @trusted {
+            return db.execute("SELECT count(*) FROM validator_set").oneValue!size_t;
+        }();
+    }
+
+    /***************************************************************************
+
+        Walk over the enrollmane data in the set and call the provided delegate
```
AndrejMitrovic:
```suggestion
        Walk over the enrollment data in the set and call the provided delegate
```

linked0:
Ready for review.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,416 @@
...
+
+        Params:
+            enroll_data = the enrollment data to add
+
+        Returns:
+            true if the enrollment data has been added to the set
+
+    ***************************************************************************/
+
+    public bool add (EnrollmentData enroll_data) @safe
```
AndrejMitrovic: This function unconditionally returns `true`. Maybe you could check `hasEnrollmentDataHash` and return `false` early if `hasEnrollmentDataHash` return true?

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,416 @@
...
+
+        Params:
+            enroll = the enrollment data hash
+
+        Returns:
+            true if the enrollment manager has the enrollment data hash
+
+    ***************************************************************************/
+
+    public bool hasEnrollmentDataHash (const ref EnrollmentData enroll) @trusted
```
AndrejMitrovic: This should be renamed to `hasEnrollmentData`. From the name it seemed like the user would provide the hash to the function, not the data.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,416 @@
...
+        return signature;
+    }
+
+    /***************************************************************************
+
+        Check if the EnrollmentData is right
+
+    ***************************************************************************/
+
+    public bool acceptEnrollment (const ref EnrollmentData enroll) @safe
```
AndrejMitrovic: Now there are two APIs: `add`, and `acceptEnrollment`. Which one should the user use?

These two functions should be merged together.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,416 @@
...
+
+        Params:
+            count = how many enrollment data to take from the set
+
+        Returns:
+            an array of 'count' enrollment data
+
+    ***************************************************************************/
+
+    version (unittest) public EnrollmentData[] take (size_t count) @safe
```
AndrejMitrovic: This can be `private`, as `private` is still usable from the same module.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,416 @@
...
+        Constructor
+
+        Params:
+            db_path = path to the database file, or in-memory storage if
+                        :memory: was passed
+            keypair = the keypair of the owner node
+
+    ***************************************************************************/
+
+    public this (in string db_path, in string seed_file_path, KeyPair keypair)
```
AndrejMitrovic: Missing docs for `seed_file_path`

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,102 @@
...
+import agora.common.crypto.Schnorr : Signature;
+import agora.common.crypto.Schnorr : sign;
+
+/*******************************************************************************
+
+    Defines data for enrollment process
+
+*******************************************************************************/
+
+public struct EnrollmentData {
```
AndrejMitrovic:
```suggestion
public struct EnrollmentData
{
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,416 @@
...
+            log.info("Rejected enroll: {}", enroll);
+            return false;
+        }
+
+        return true;
+    }
+
+    /***************************************************************************
+
+        Check if a enrollment data hash exists in the transaction pool.
```
AndrejMitrovic:
```suggestion
        Check if the enrollment data exists in the transaction pool.
```

linked0:
Ready for review.

AndrejMitrovic:
Looks good. I'll let @Geod24 review it when he's back.

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,99 @@
...
+/*******************************************************************************
+
+    Define enrollment data (enrollment data = validator)
+
+*******************************************************************************/
+
+public struct EnrollmentData
+{
+    /// K: UTXO key, A public key matching a frozen UTXO
+    public PublicKey K;
```
TrustHenry: It is correct to set the frozen UTXO hash here.
It is the only way to tell if you have frozen UTXO.

AndrejMitrovic: Are you asking to change something here?

linked0: @TrustHenry You mean that a enrollment data must have a information like the UTXO hash by which the frozen UTXO validator(you are now implementing) could find a matching UTXO?

TrustHenry: @linked0 @AndrejMitrovic  
Yes, that’s right.
In the Enrollment process, Publickey cannot be a key.
A Publickey can contain many UTXOs.
We have to choose one frozen UTXO.
This can be a key in Enrollment.

AndrejMitrovic: > We have to choose one frozen UTXO.

True. We could also sum the frozen UTXOs for the given key, but this might complicate the `N` value which is currently hardcoded by definition.

TrustHenry:
If UTXO hash becomes PREMARY KEY,
We need to implement a way to find the UTXO hash.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,394 @@
...
+
+        return !results.empty;
+    }
+}
+
+/// tests for createEnrollmentData, hasEnrollmentDataHash, add, length functions
+unittest
+{
+    KeyPair kp = KeyPair.random();
+    auto random_seed_path = buildPath(getcwd, "random_set.dat");
```
TrustHenry: FileIO is not recommended for unittest.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,394 @@
...
+    assert(man.hasEnrollmentData(enroll_data) == true);
+    assert(man.length() == 1);
+    man.shutdown();
+}
+
+/// tests for onApply, remove, take functions
+unittest
+{
+    KeyPair kp = KeyPair.random();
+    auto random_seed_path = buildPath(getcwd, "random_set.dat");
```
TrustHenry: FileIO is not recommended for unittest.

source/agora/consensus/data/EnrollmentData.d
linked0: Can I apply this request to my next PR? I have already implemented a API endpoint of Node about the enrollment process and I think we need more time to conclude how it can be implemented. So it would be better to merge the code of this PR to the Upstream and to apply Henry's request to next PR.

AndrejMitrovic: Where is the rush? :) If we know the code will be changed after merging we might as well change it before merging. It's very easy to go from the utxo hash to the PublicKey. It's just `utxo_set[utxo_hash].address`.

linked0: @Geod24 Which do you think is best way to apply @TrustHenry's request between replacing a public key with a UTXO hash or just adding a UTXO hash member in EnrollmentData structure?

linked0:
We determined that the enrollment process will use a frozeon UTXO hash instead of the public key of a node as means to identify a validator, which was output at the whiteboarding session.

linked0:
Ready for review.

Geod24:
Can you rebase on v0.x.x ?

linked0:
I have rebased on v0.x.x and I'm ready for review.

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,99 @@
...
+/*******************************************************************************
+
+    Define enrollment data (enrollment data = validator)
+
+*******************************************************************************/
+
+public struct EnrollmentData
+{
+    /// K: UTXO hash, A hash of a frozen UTXO
+    public Hash K;
```
Geod24: I used `K` in the paper for conciseness, but I think we should have more explicit names. This applies to everything in this struct, except `R` maybe, as it's the name used in the Schnorr formula.

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,99 @@
...
+    /***************************************************************************
+
+        Implements hashing support
+
+        Params:
+            dg = hashing function
+
+    ***************************************************************************/
+
+    public void computeHash (scope HashDg dg) const nothrow @safe @nogc
```
Geod24: You can remove this

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,99 @@
...
+    /***************************************************************************
+        
+        EnrollmentData serialization
+
+        Params:
+            dg = Serialize function
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) const @safe
```
Geod24: Remove

source/agora/consensus/data/EnrollmentData.d
```diff
@@ -0,0 +1,99 @@
...
+    /***************************************************************************
+        
+        EnrollmentData deserialization
+
+        Params:
+            dg = Serialize function
+
+    ***************************************************************************/
+
+    public void deserialize (scope DeserializeDg dg) @safe
```
Geod24: Remove

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+import std.path;
+import std.stdio;
+
+/*******************************************************************************
+
+    Initialize the logger 
+
+*******************************************************************************/
+
+static this ()
```
Geod24: Why not use the `mixin` ?

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+
+*******************************************************************************/
+
+module agora.common.EnrollmentManager;
+
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
+import agora.common.crypto.Schnorr : Pair;
+import agora.common.crypto.Schnorr : Signature;
+import agora.common.crypto.Schnorr : sign;
```
Geod24: In D you can do the following:
```D
import agora.common.crypto.Schnorr : Pair, Signature, sign;
```

If you want to keep the selective import.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+                        :memory: was passed
+            seed_file_path = path to the data file storing randon seed
+
+    ***************************************************************************/
+
+    public this (string db_path, string data_dir)
+    {
+        const db_exists = db_path.exists;
+        if (db_exists)
+            log.info("Loading database from: {}", db_path);
```
Geod24: I would make this a `trace` call, and I think it'll be better to check for a valid file (e.g. one with data in it) rather than the file existing. It also works better with unittests.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+    /// Database path
+    private string db_path;
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            db_path = path to the database file, or in-memory storage if
+                        :memory: was passed
```
Geod24: This is a detail of the database. Maybe just say "path to the database file (sqlite3 format)"

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+            seed_file_path = path to the data file storing randon seed
```
Geod24: Wrong name. Also, rando*m*

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+            log.info("Loading database from: {}", db_path);
+
+        this.db = Database(db_path);
+
+        // create the table for validator set if it doesn't exist yet
+        this.db.execute("CREATE TABLE IF NOT EXISTS validator_set " ~
+            "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
+
+        // load random seed
+        this.random_seed_path = buildPath(data_dir, "random_set.dat");
```
Geod24: It does IO (disk) access, so it's not suitable for an unittest.
The other part correctly abstract things using `db_path`, but this does not.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+
+        Params:
+            enroll_data = the enrollment data to add
+
+        Returns:
+            true if the enrollment data has been added to the validator set
+
+    ***************************************************************************/
+
+    public bool add (EnrollmentData enroll_data) @safe
```
Geod24:
```suggestion
    public bool add (const ref EnrollmentData enroll_data) @safe
```

Also, what @AndrejMitrovic said

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+        // check if already exists
+        if (hasEnrollmentData(enroll_data.K))
+        {
+            log.info("Rejected enroll: {}", enroll_data);
+            return false;
+        }
+
+        () @trusted { assumeSafeAppend(buffer); } ();
+
+        scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
```
Geod24: Note to myself: We should make this more idiomatic

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the number of validators in the validator set
+
+    ***************************************************************************/
+
+    public size_t length () @safe
```
Geod24: I'd say `computeLength` to make it obvious that a call isn't cheap.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+
+        Params:
+            dg = the delegate to call
+
+        Returns:
+            the status code of the delegate, or zero
+
+    ***************************************************************************/
+
+    public int opApply (scope int delegate(Hash hash, EnrollmentData enroll) dg)
```
Geod24:
```suggestion
    public int opApply (scope int delegate(const ref Hash hash, const ref EnrollmentData enroll) dg)
```

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+        Remove the enrollment data with the given key from the validator set
+
+        Params:
+            enroll = the enrollment data to remove
+
+    ***************************************************************************/
+
+    public void remove (const ref EnrollmentData enroll) @trusted
+    {
+        auto hash = hashFull(enroll.K);
```
Geod24: During the `INSERT`, `K` is inserted directly, but here it is hashed. So I don't think it works, and I think you need an extra unittest :)
Also, the API should probably be based on `K`, not the full enrollment data.

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+        saveRandomSeed();
+
+        // X, nth image of random seed
+        this.data.X = hashFull(this.random_seed);
+        foreach (i; 0 .. this.data.n-1)
+            this.data.X = hashFull(this.data.X);
+
+        // R, signature noise
+        this.schnorr_pair = Pair.random();
+        this.data.R = this.schnorr_pair.V;
```
Geod24: This also needs to be saved

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+        }
+
+        return false;
+    }
+
+    // sign enrollment data with schnorr key pairs
+    private static Signature signHash (const ref Hash frozen_utxo_hash, 
+        const ref Pair schnorr_pair, const ref Hash hash)
+    {
+        auto pk = Scalar(frozen_utxo_hash);
```
Geod24: Using the `frozen_utxo_hash` should strike anyone as a pretty bad idea

source/agora/node/EnrollmentManager.d
```diff
@@ -0,0 +1,427 @@
...
+        out EnrollmentData enroll_data) @trusted
+    {
+        auto results = this.db.execute("SELECT key, val FROM validator_set " ~
+            "WHERE key = ?", enroll_hash[]);
+
+        foreach (row; results)
+        {
+            enroll_data = deserializeFull!EnrollmentData(row.peek!(ubyte[])(1));
+
+            // break
```
Geod24:
```suggestion
```

Geod24 (changes requested):
Added a few comments:
- The IO needs to be abstracted away
- The UTXO hash cannot be used as private key
- `r` needs to be saved

linked0:
Ready for review

TrustHenry (approved):
LGTM

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,409 @@
...
+    utxo_hash = utxo_set.getHash(secondHash, 0);
+    enroll = man.createEnrollment(utxo_hash);
+    man.addEnrollment(enroll);
+    assert(man.getEnrollmentLength() == 2);
+
+    // remove an Enrollment object
+    man.removeEnrollment(utxo_hash);
+    assert(man.getEnrollmentLength() == 1);
+
+    man.shutdown();
```
AndrejMitrovic: This calls the shutdown a second time after `scope (exit) utxo_set.shutdown();` already did it. So you can just remove `man.shutdown()` here.

linked0: The `man.shutdown()` call must be called. So I will add the `scope(exit) man.shutdown() ` below the `auto man = new UnitEnrollMan()` code

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,409 @@
...
+        this.random_seed_src = Scalar.random();
+
+        // X, nth image of random seed
+        this.data.random_seed = hashFull(this.random_seed_src);
+        foreach (i; 0 .. this.data.cycle_length-1)
+            this.data.random_seed = hashFull(this.data.random_seed);
+
+        // R, signature noise
+        this.signature_noise = Pair.random();
+        this.data.R = this.signature_noise.V;
```
AndrejMitrovic: @Geod24 I think I see an optimization opportunity here. We are storing `R` in both the `EnrollmentData` and the `(R, s)` signature pair. Could we omit R from the `EnrollmentData`? After all, `R` is already hashed in the challenge.

linked0:
Ready for review.
I've removed the signature noise variable from the Enrollment structure because it is also stored in a Signature object.

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+
+*******************************************************************************/
+
+module agora.common.EnrollmentManager;
+
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
+import agora.common.crypto.Schnorr : Pair;
+import agora.common.crypto.Schnorr : Signature;
+import agora.common.crypto.Schnorr : sign;
```
Geod24: Merge those together (actually I'd say, just loose the selective import and make it a regular import)

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+import agora.common.Serializer;
+import agora.consensus.data.Enrollment;
+import agora.utils.Log;
+
+import d2sqlite3.database;
+import d2sqlite3.library;
+import d2sqlite3.results;
+import d2sqlite3.sqlite3;
+
+import ocean.util.log.Logger;
```
Geod24: Remove (it is not used AFAICS)

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+import std.algorithm;
+import std.file;
+import std.path;
+import std.stdio;
+
+mixin AddLogger!();
+
+/*******************************************************************************
+
+    Handle enrollment data and manage the validator set(db)
```
```suggestion
    Handle enrollment data and manage the validators set
```

I'm not sure what the `db` is about. Looks like an implementation detail.

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+        // create the table for validator set if it doesn't exist yet
+        this.db.execute("CREATE TABLE IF NOT EXISTS validator_set " ~
+            "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
+
+        // set node's KeyPair object
+        this.node_key_pair = key_pair;
+
+        // load random seed
+        this.node_enroll_data_path = buildPath(data_dir, "node_enroll_data.dat");
+        loadNodeEnrollmentData();
```
Geod24: My comment about mixing abstraction still stands. This is not unittestable. Why not make it part of the database ?

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+            seed_file_path = path to the data file storing randon seed
+            key_pair = the keypair of the owner node
+
+    ***************************************************************************/
+
+    public this (string db_path, string data_dir, KeyPair key_pair)
+    {
+        const db_exists = db_path.exists;
+        if (db_exists)
+            log.info("Loading database from: {}", db_path);
```
```suggestion
            log.trace("Loading enrollment database from: {}", db_path); 0 
```

Geod24: Actually I'd say just remove those 3 lines ?

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+
+        // set node's KeyPair object
+        this.node_key_pair = key_pair;
+
+        // load random seed
+        this.node_enroll_data_path = buildPath(data_dir, "node_enroll_data.dat");
+        loadNodeEnrollmentData();
+
+        // set information to member variables
+        this.db_path = db_path;
```
Geod24: This member variable is never used

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+
+    public bool addEnrollment (Enrollment enroll) @safe
+    {
+        static ubyte[] buffer;
+        buffer.length = 0;
+
+        // check if already exists
+        if (hasEnrollment(enroll.utxo_key))
+        {
+            log.info("Rejected enroll: {}", enroll);
```
Geod24: The message is too broad. Imagine trying to track down a bug in 3 months and seeing this. You at least want to know *why* the enrollment was rejected. Something like "Rejected already existing enrollment: {}" gives much more information at a glance.

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+                                a validator and to generate a siging key
+
+        Returns:
+            an enrollment data
+
+    ***************************************************************************/
+
+    public Enrollment createEnrollment (const ref Hash frozen_utxo_hash) @trusted
+    {
+        this.data = new Enrollment();
```
Geod24: Why is it a pointer ?

Geod24: Also is this function supposed to be called multiple times ? In what order ? For what purpose ?
The documentation is very generic. The fact that it is stored as a member of this class tells me that either:
- There can only be one
OR
- We cache the last one

If there can only be one, it feels like it shouldn't be called multiple times. If we cache the last created Enrollment data, why do we do that ? Neither the documentation of this function nor the variable says anything.

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+        {
+            enroll = deserializeFull!Enrollment(row.peek!(ubyte[])(1));
+            return true;
+        }
+
+        return false;
+    }
+
+    // sign enrollment data with schnorr key pairs
+    private static Signature signHash (const ref SecretKey secret_key,
```
Geod24: This function is clearly not carrying its weight and should be removed, unless I am missing something ?

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+        // X, nth image of random seed
+        this.data.random_seed = hashFull(this.random_seed_src);
+        foreach (i; 0 .. this.data.cycle_length-1)
+            this.data.random_seed = hashFull(this.data.random_seed);
+
+        // R, signature noise
+        this.signature_noise = Pair.random();
+
+        // signature
+        Hash hash = hashFull(*this.data);
```
Geod24: Why the double hashing ?

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+            enroll_hash = key for an enrollment data which is hash of frozen UTXO 
+
+        Returns:
+            true if the enrollment manager has the enrollment data
+
+    ***************************************************************************/
+
+    public bool hasEnrollment (const ref Hash enroll_hash) @trusted
+    {
+        auto results = this.db.execute("SELECT key, val FROM validator_set " ~
```
Geod24: There is an `EXISTS` statement in SQL to allow for short-circuiting, otherwise the query planner execute over the whole table even if it can find a record:
```sql
SELECT EXISTS(SELECT 1 FROM validator_set WHERE key = ?)
```

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+
+        Params:
+            dg = the delegate to call
+
+        Returns:
+            the status code of the delegate, or zero
+
+    ***************************************************************************/
+
+    public int opApply (scope int delegate(const ref Hash hash,
```
Geod24: This function is not tested

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+            enroll_hash = key for an enrollment data which is a hash of a frozen 
+                            UTXO
+            enroll = will contain the enrollment data if found
+
+        Returns:
+            Return true if the enrollment data was found
+
+    ***************************************************************************/
+
+    public bool getEnrollment (const ref Hash enroll_hash,
```
Geod24, line 279: This function is not tested

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+
+    ***************************************************************************/
+
+    public bool addEnrollment (Enrollment enroll) @safe
+    {
+        static ubyte[] buffer;
+        buffer.length = 0;
+
+        // check if already exists
+        if (hasEnrollment(enroll.utxo_key))
```
Geod24: This branch is not tested

source/agora/common/crypto/Schnorr.d
```diff
@@ -133,6 +135,36 @@ public struct Pair
...
+    /***************************************************************************
+
+        Serialization
+
+        Params:
+            dg = serialize function accumulator
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) const @safe
```
Geod24: Superfluous, no ?

Geod24 (changes requested):

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,407 @@
...
+
+        Params:
+            enroll = the enrollment data to add
+
+        Returns:
+            true if the enrollment data has been added to the validator set
+
+    ***************************************************************************/
+
+    public bool addEnrollment (Enrollment enroll) @safe
```
AndrejMitrovic, line 123: I think this method needs to be `nothrow`. Otherwise, you have to care about two cases: if the function returns false, or if it throws an exception. If it returns a status code, we need to make sure it never throws.

source/agora/common/EnrollmentManager.d
AndrejMitrovic: What Mathias is saying, is that the `sign` function already hashes the data structure. So it's not necessary to pre-hash it here, and you could even get rid of `signHash` since it's only used in one place.

source/agora/common/EnrollmentManager.d
linked0: The table, or the validator_set is for storing other validators' information, so it does not fits to the information which is just the signature noise of the enrollment data for a node itself. Do you mean another table for storing just one row for the signature noise? I'd like to make a memory storage in the EnrollmentManager in case of an unittest. What do you think about this?

Geod24: I meant another table. There are other options, but this one seems the most straightforward.

linked0: I think another table is simple and straightforward also. I will make another table named "node_enroll_data".

source/agora/common/EnrollmentManager.d
linked0: @Geod24 I'd like to discuss with you about this. I wanna split the EnrollmentManager into the ValidatorSet which manage other validators' information and the EnrollmentManager which manage the enrollment data for a node itself.

linked0:
Ready for review

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.EnrollmentManager;
+
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
+import agora.common.crypto.Schnorr;
+import agora.common.crypto.Schnorr : Signature;
```
Geod24: Redundant

source/agora/common/EnrollmentManager.d
Geod24: Ping

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+        // set node's KeyPair object
+        this.node_key_pair = key_pair;
+
+        // load signature noise
+        auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
+            "WHERE key = ?", "signature_noise");
+
+        foreach (row; results)
+        {
+            signature_noise = deserializeFull!Pair(row.peek!(ubyte[])(1));
```
Geod24: This branch is not tested

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+
+    ***************************************************************************/
+
+    public bool addEnrollment (Enrollment enroll) @safe
+    {
+        static ubyte[] buffer;
+        buffer.length = 0;
+
+        // check if already exists
+        if (hasEnrollment(enroll.utxo_key))
```
```suggestion
        if (this.hasEnrollment(enroll.utxo_key))
```

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+
+        scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
+        {
+            buffer ~= data;
+        };
+
+        serializePart(enroll, dg);
+
+        () @trusted {
+            db.execute("INSERT INTO validator_set (key, val) VALUES (?, ?)",
```
```suggestion
            this.db.execute("INSERT INTO validator_set (key, val) VALUES (?, ?)", 2 
```

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+
+        Returns:
+            the number of validators in the validator set
+
+    ***************************************************************************/
+
+    public size_t getEnrollmentLength () @safe
+    {
+        return () @trusted {
+            return db.execute("SELECT count(*) FROM validator_set").oneValue!size_t;
```
```suggestion
            return this.db.execute("SELECT count(*) FROM validator_set").oneValue!size_t; 3 
```

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+        Params:
+            frozen_utxo_hash = the hash of a frozen UTXO used to identify 
+                                a validator and to generate a siging key
+
+        Returns:
+            an enrollment data
+
+    ***************************************************************************/
+
+    public Enrollment createEnrollment (const ref Hash frozen_utxo_hash) @trusted
```
Geod24: `const ref` didn't work ?

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+            true if the enrollment manager has the enrollment data
+
+    ***************************************************************************/
+
+    public bool hasEnrollment (const ref Hash enroll_hash) @trusted
+    {
+        auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM validator_set " ~
+            "WHERE key = ?)", enroll_hash[]);
+
+        return results.oneValue!(bool);
```
Geod24: No need for the intermediate variable, return it directly (it might also be more optimal thanks to NRVO)

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+
+*******************************************************************************/
+
+public class EnrollmentManager
+{
+    /// SQLite db instance
+    private Database db;
+
+    /// Node's key pair
+    private KeyPair node_key_pair;
```
```suggestion
    private Pair node_key_pair;
```

You could also loose the `node_` part if you want, as I don't think there's any other kind of key_pair the node will know of :)

source/agora/consensus/data/Enrollment.d
```diff
@@ -0,0 +1,43 @@
...
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.Enrollment;
+
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
+import agora.common.Deserializer;
```
Geod24:
```suggestion
```

source/agora/consensus/data/Enrollment.d
```diff
@@ -0,0 +1,43 @@
...
+import agora.common.Hash;
+import agora.common.Serializer;
```
Geod24:
```suggestion
```

source/agora/consensus/data/Enrollment.d
```diff
@@ -0,0 +1,43 @@
...
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.Enrollment;
+
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
+import agora.common.Deserializer;
+import agora.common.Hash;
```
Geod24:
```suggestion
```

source/agora/consensus/data/Enrollment.d
```diff
@@ -0,0 +1,43 @@
...
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.Enrollment;
+
+import agora.common.crypto.ECC;
+import agora.common.crypto.Key;
```
Geod24: You can also remove those two and import `common.Types` instead

source/agora/common/EnrollmentManager.d
```diff
@@ -0,0 +1,353 @@
...
+        auto results = this.db.execute("SELECT key, val FROM validator_set " ~
+            "WHERE key = ?", enroll_hash[]);
+
+        foreach (row; results)
+        {
+            enroll = deserializeFull!Enrollment(row.peek!(ubyte[])(1));
+            return true;
+        }
+
+        return false;
```
Geod24, line 291: This branch is not tested

Geod24 (commented):
Looks nice, just a few comments

source/agora/common/EnrollmentManager.d
linked0: I wrote the code with reference to the add function of the TransactionPool. I have tried to add "nothrow" to the function but it output too many errors which might take me too much time. So, I'd like to modify that as Drey suggested after fully understanding the concept in other PR if possible.

source/agora/common/EnrollmentManager.d
linked0: In the unittest code, all the data saved in the database cannot be shared because the database is in memory as far as I know. So this branch cannot be tested in the unittest code. Would you please show me a clue for that?

source/agora/common/EnrollmentManager.d
linked0: The node_key_pair must be KeyPair type because it is set from a node's keypair. This variable is set in the constructor, which will be called by a node later.

AndrejMitrovic: He means to rename `node_key_pair` to `key_pair`

linked0: His suggested change is about the type of the node_key_pair variable. I have already renamed the name of the vairable to `key_pair` in my current code.

AndrejMitrovic: You're right, I read it wrong.

Geod24: > The node_key_pair must be KeyPair type because it is set from a node's keypair.

But at the signature steps you turn it into a `Pair`.

Geod24 (approved):

linked0:
@AndrejMitrovic Ping

PR #420, opened 2019-11-13
Make the Schnorr signing routines public

AndrejMitrovic:
The multi-sig version should be safe to use, as long as R is never re-used for different messages and is generated by a cryptographically-secure random number generator (e.g. libsodium's `randombytes_uniform()`, or `Pair.random()`)


linked0:
This PR has simple change and is needed for PR #419 
LGTM.

linked0 (approved):
This PR has a simple change and is needed for PR #419 
LGTM.

Issue #365, opened 2019-09-30
Implement Schnorr-based signature of the block


Geod24:
As a node operator, I need blocks that has been accepted by the consensus protocol to be signed in a compact way. In the best case scenario, the signature should use `O(1)` space & time.

The first step is to come up with a definition of the data structure used, as well as the signature collection window. For example, can a signature be added to the block after another block has been appended to the chain ?

Also, the Schnorr signature algorithm currently depends on:
- `R` (random number)
- `S` (public key)
- Hash(R || S || message)

In this case, `message` could simply be the hash of the block. There is currently no evidence that `S` being part of the `Hash` is more secure, so it could be removed. However
 this scheme, having `R` as part of the Hash, does not allow arbitrarily adding signatures.

### Schnorr multi-signature for the block signature. 
More than half the validators must sign a block which has reached consensus before it will be externalized.
Validators "gossip" their block signatures after the block is externalized, which means that even after consensus is reached, and the block is externalised, more validators can add their signature to the block.

Each validator signs the block with (R, s) where
R = (r + c).G
where
c = Hash(Block)
r = r + c where r is the private (Scalar) key of R from signature of enrollment

### s = r + (c * v)

where
c = Hash(Block)
r = r + c where r is the private (Scalar) key of R
v = private (Scalar) of validator Schnorr keyPair

To validate an individual signature:-

### S == R + (c * V)

where
S = s.G (Scalar -> point)
V = public key of validator as Point

To validate combined block signature:-

### S = R1 + R2 ... + c * (V1 + V2 ...)

where S is s.G from block signature (R, s)

<img width="1233" alt="BlockMultiSig" src="https://user-images.githubusercontent.com/11815575/103040887-201cf600-45b8-11eb-9225-dc28debe6824.png">

AndrejMitrovic:
> For example, can a signature be added to the block after another block has been appended to the chain ?

I don't think so, otherwise the chain is not immutable?

AndrejMitrovic:
> However this scheme, having R as part of the Hash, does not allow arbitrarily adding signatures.

What did you mean by this, "arbitrarily"?

AndrejMitrovic:
Ok, so I'm thinking we could have something like this in the block:

```D
Bitfield bitfield;
Signature sig;
```

[...]

Issue #205, opened 2019-08-19
Ensure that validators sign with a frozen UTXO


bpalaggi:
As a developer, I need to make sure that signatures by validators are done using a frozen UTXO in order to make sure the validator has stake in the system.

**Definition of Done:**
- [x] There is a way to link a signature and a frozen UTXO

Geod24:
Depends on #204

bpalaggi:
Depends on #209

Geod24:
Unblocked

TrustHenry:
> I need to make sure that signatures by validators are done using a frozen UTXO in order to make sure the validator has stake in the system.

Where signature by Validators means signature of BlockHeader or `enrollment.S` ?
Does frozen UTXO check the UTXO of the enrollment K (UTXO key) ?

Geod24:
> Where signature by Validators means signature of BlockHeader 

Yes

> Does frozen UTXO check the UTXO of the enrollment K (UTXO key) ?

That's a different issue

TrustHenry:
When a block is received or signed
Validators signed BitField...
The frozen UTXO shall be found and verified in utxo_db?

AndrejMitrovic:
This issue #205 only deals with the validation part of a block.

So if we have a block header which is signed, in order to validate the block then the signature must be checked. We need to make sure:
- The signature is actually valid.
- The key it was signed with belongs to a node that has frozen 40,000 Boa. If the node hasn't frozen 40,000 Boa then it has no right to sign the block, therefore the block is invalid.
- The height of the block that contains the signature of a node, it must be at least N blocks after the block height where the node staked 40,000 Boa. The "N" is defined in the Yellow paper. So, a node first freezes UTXO, then after N blocks it can start signing blocks.
- If the UTXOs for that node are actually melting, the signature is considered invalid.

I suggest implementing this issue #205 in several **steps**, depending on what kind of approach you want to take:
- At first, assume the block header can only contain one validator signature. This makes it easy to implement the checks for the frozen UTXO. You could just store a `PublicKey validator; Signature signature;` in the block header, and change `isInvalidReason` for blocks so that it tests the conditions that I listed in the previous paragraph (see also the Definition of Done).
[...]

PR #422, opened 2019-11-15
Ensure that validators sign with a frozen UTXO

TrustHenry:
#205 issues.
#365 and #406 are implemented, we can adapt my solution and change the block header so it works with the Schnorr signature scheme and with the Enrollment process.

This adds ValidatorSig to the blockhead.
Hash and sign the BlockHead of Validators to Blockheader.ValidatorSigs.
The blockhead verifies the signature of Validators of the block in Validation.isInvalidReason.
Check the Unspent Freeze UTXO associated with the Publickey used to sign Validators.

This can identify the Freeze UTXO proof of Validator, which can generate blocks.

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -48,6 +48,8 @@ public struct BlockHeader
     /// The hash of the merkle root of the transactions
     public Hash merkle_root;
 
+    /// Signing the blockheader of Validators
+    public ValidatorSig[] validator_sigs;
 
     /***************************************************************************
 
@@ -63,6 +65,9 @@ public struct BlockHeader
         dg(this.prev_block[]);
         hashPart(this.height, dg);
         dg(this.merkle_root[]);
+
+        foreach (validator_sig; this.validator_sigs)
+            hashPart(validator_sig, dg);
     }
 
     /***************************************************************************
@@ -79,6 +84,10 @@ public struct BlockHeader
         dg(this.prev_block[]);
         serializePart(this.height, dg);
         dg(this.merkle_root[]);
+
+        serializePart(this.validator_sigs.length, dg);
+        foreach (const ref validator_sig; this.validator_sigs)
+            serializePart(validator_sig, dg);
     }
 
     /***************************************************************************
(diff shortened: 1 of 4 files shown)


TrustHenry:
The unit test will be added after the dependency is resolved.

source/agora/consensus/data/Block.d
```diff
@@ -120,6 +136,72 @@ unittest
...
 }
 
+/*******************************************************************************
+
+    ValidatorSigs must be signed with a valid freeze utxo of the Validator
+    in the blockheader.
+
+*******************************************************************************/
+
+public struct ValidatorSigs
```
linked0: The structure name, ValidatorSigs is confusing because it make developers to think that the structure has multiple signatures. You should use the name like ValidatorSig.

linked0 (changes requested):

source/agora/consensus/data/Block.d
```diff
@@ -63,6 +65,9 @@ public struct BlockHeader
         dg(this.prev_block[]);
         hashPart(this.height, dg);
         dg(this.merkle_root[]);
+
+        foreach (validator_sig; this.validator_sigs)
```
AndrejMitrovic, line 69: The validators sign the block header hash, correct? But the block header hash includes the signatures. So how do you sign a block header hash when that hash includes the signature? You can't, because as soon as you sign a block header and then set the signature, the block header hash changes.

You cannot include the validator signatures as part of the block header hash.

source/agora/consensus/data/Block.d
TrustHenry: applied the suggested.

source/agora/consensus/data/Block.d
TrustHenry: ValidatorSig -
`    public void computeHash (scope HashDg dg) const nothrow @safe @nogc
    {
        hashPart(this.key, dg);
        dg(this.utxo_hash[]);
    }`
Signature information is excluded from computeHash.

linked0 (approved):

TrustHenry:
By the concept here #429.
If BitField Validators and Signature Signature; are applied,
Validate using Schnorr in the ValidateBlock and check UTXO.

TrustHenry:
this is changing the structure.

AndrejMitrovic:
> this is changing the structure.

Are you implying I should close https://github.com/bpfkorea/agora/pull/417?

TrustHenry:
Once #417 is merged I will modify the code to base it.

Geod24:
Can you rebase ?

AndrejMitrovic:
This can be unblocked now that #417 has been merged?

Geod24:
Can you rebase ?

TrustHenry:
This has become obsolete because the newly code has been merged. #467 
Therefore, I close this PR.

PR #424, opened 2019-11-15
Replace fromBinary() with deserialize() to make it consistent

TrustHenry:
We need serialize and deserialize to add a publickey in the block structure to be added.
issue #205

### DIFF
--- a/source/agora/common/crypto/Key.d
+++ b/source/agora/common/crypto/Key.d
@@ -198,20 +198,17 @@ public struct PublicKey
 
     /***************************************************************************
 
-        PublicKey fromBinary
+        Key Deserialization
 
         Params:
             dg = deserialize function
 
-        Returns:
-            `PublicKey` Public key address
-
     ***************************************************************************/
 
-    public static PublicKey fromBinary (scope DeserializeDg dg) @safe
+    public void deserialize (scope DeserializeDg dg) @safe
     {
         alias DType = typeof(this.data);
-        return PublicKey(DType(dg(DType.sizeof)));
+        this.data = DType(dg(DType.sizeof));
     }
 
     ///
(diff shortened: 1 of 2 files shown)


linked0 (approved):
LGTM!
It's needed for the issue #406 .

AndrejMitrovic (approved):

AndrejMitrovic:
Feel free to merge when it passes the test-suite.

AndrejMitrovic:
I think the `fromBinary` method should probably be removed? I don't see it used anywhere, and it looks like it should have been called `deserialize`.

TrustHenry:
I agree with you
We can replace the `fromBinary` with `deserialize`.
I'll change it.

source/agora/common/crypto/Key.d
```diff
@@ -198,20 +198,17 @@ public struct PublicKey
...
-            `PublicKey` Public key address
-
     ***************************************************************************/
 
-    public static PublicKey fromBinary (scope DeserializeDg dg) @safe
+    public void deserialize (scope DeserializeDg dg) @safe
     {
         alias DType = typeof(this.data);
-        return PublicKey(DType(dg(DType.sizeof)));
+        this.data = PublicKey(DType(dg(DType.sizeof)));
```
AndrejMitrovic:
```suggestion
        this.data = DType(dg(DType.sizeof));
```

TrustHenry: applied the suggested.

Geod24:
> I think the fromBinary method should probably be removed?

Actually, it's something I wanted to experiment more with. Because `deserialize` is really problematic: some types make no sense default-initialized (`Point`), some types could be in an invalid state (`invariant` failing), etc... So initializing a type straight from binary, be it via a ctor or a specially-named method, is IMO a better road.
I went with a specially-named method because that's where you put the validation (as opposed to constructors).

PR #425, opened 2019-11-18
Add the findUTXO feature of the UTXOSet.

TrustHenry:
Currently, the required parameters hash, index are required in the findUTXO in UTXOSet.
In addition, it is necessary to find UTXO setHash directly without index in UTXO pool.

Index is If size_t.max, find the hash parameter by UTXO hash

#205

### DIFF
--- a/source/agora/consensus/data/UTXOSet.d
+++ b/source/agora/consensus/data/UTXOSet.d
@@ -254,7 +254,7 @@ public class UTXOSet
 
     ***************************************************************************/
 
-    private static Hash getHash (Hash hash, ulong index) @safe nothrow
+    public static Hash getHash (Hash hash, ulong index) @safe nothrow
     {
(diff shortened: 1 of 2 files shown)


MichaelKim20 (commented):
All right, I'd like to have the part that uses this.

MichaelKim20 (approved):

AndrejMitrovic:
Could you add a simple unittest? You could modify one of the existing tests. This is to ensure we don't accidentally break it in the future.

TrustHenry:
added unit test code

AndrejMitrovic (approved):

AndrejMitrovic:
In general we don not want to make methods public just because they're used in a far away unittest. However I'll merge it for now, because `UtxoSet.d` does not have any unittests of its own. I'll file an issue for this.

PR #429, opened 2019-11-21
Implement proof-of-concept enrollment-based Schnorr signature scheme

AndrejMitrovic:
Part of https://github.com/bpfkorea/agora/issues/365

### DIFF
--- a/source/agora/common/BitField.d
+++ b/source/agora/common/BitField.d
@@ -119,6 +119,20 @@ public struct BitField
     {
         return this.num_bits;
     }
+
+    /// nothrow wrapper around BitArray.opApply (todo: fix BitArray)
+    public int opApply( scope int delegate(ref bool) nothrow dg ) @trusted nothrow
+    {
+        scope (failure) assert(0);
+        return this.bit_array.opApply(dg);
+    }
+
+    /// ditto
+    public int opApply (scope int delegate(ref size_t, ref bool) nothrow dg) @trusted nothrow
+    {
+        scope (failure) assert(0);
+        return this.bit_array.opApply(dg);
+    }
 }
 
 /// opEquals tests
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -14,6 +14,7 @@
 module agora.consensus.Validation;
 
 import agora.common.Amount;
+import agora.common.Config;
 import agora.common.crypto.ECC;
 import agora.common.crypto.Key;
 import agora.common.crypto.Schnorr;
@@ -23,8 +24,109 @@ import agora.consensus.data.Enrollment;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSet;
 import agora.consensus.Genesis;
+import agora.node.Ledger;
 
 import std.conv;
+
+/// Node ID => Block #index => Preimage
+public alias PreimagesMap = Hash[ulong][ushort];
+
+/// Node ID => Block #Index => Expected R value
+/// (based on R2 = R1 + X1, where X1 is the preimage for block #1)
+public alias PubRMap = Point[ulong][ushort];
+
+/*******************************************************************************
+
+    Check the validation of the block header's signature
+
+    Params:
+        header = the block header to validate
+        preimage_map = the currently revealed preimage_map of all the nodes
+        pub_R_map = the map of the expected R's based on the revealed preimage_map
+                      (e.g. R2 = R1 + X1, a map of R2 for each validator node)
+        pub_keys = public keys of all validators, sorted alphabetically
+                   to enable bitmask index lookup
+
+    Return:
+        `null` if the block is valid, otherwise a string explaining the
+        reason it is invalid.
+
+*******************************************************************************/
+
+public string isInvalidSignatureReason (BlockHeader header,
+    PreimagesMap preimages_map, PubRMap pub_R_map, Point[] pub_keys)
+    nothrow @safe
+{
+    import agora.common.crypto.Schnorr : verify;
+
+    Point P;  // sum of P for the validators which signed
+    Point R;  // ditto for R
+    size_t num_signers;
+
+    foreach (idx, has_signed; header.validators)
+    {
+        if (!has_signed)
+            continue;
+
+        /// node indices are ushort
+        assert(idx < ushort.max);
+        const node_idx = cast(ushort)idx;
+
+        /// verification: the pubkeys & preimages are known and the R's were calculated
+        if (node_idx >= pub_keys.length)
+            return "Validator is not enrolled";
+
+        auto preimages = node_idx in preimages_map;
+        if (preimages is null)
+            return "Validator has not revealed any preimages";
+
+        if (header.height !in *preimages)
+            return "Validator has not revealed the preimage for this block height";
+
+        // sanity checks: if preimages exist then the R values should have been calculated
+        auto rand_map = node_idx in pub_R_map;
+        assert(rand_map !is null);
+        auto rand_val = header.height in *rand_map;
+        assert(rand_val !is null);
+
+        /// Add the validator's P
+        if (P == Point.init)  // note: Point.init + Point != Point, must use assignment here
+            P = pub_keys[idx];
+        else
+            P = P + pub_keys[idx];
+
+        /// Add the validator's R
+        const Point node_R = *rand_val;
+        if (R == Point.init)  // note: Point.init + Point != Point, must use assignment here
+            R = node_R;
+        else
+            R = R + node_R;
+
+        num_signers++;
+    }
+
+    // todo: could have a rule: at least 50% + 1 must have signed the block
+    // in order for the signature to be considered valid
+    if (num_signers == 0)
+        return "Not enough validators signed this block";
+
+    // this avoids a signature check if R has been manipulated
+    if (header.signature.R != R)
+        return "Signature.R does not match expected R";
+
+    if (!verify(P, header.signature, header))
+        return "Signature is invalid";
+
+    return null;
+}
+
+/// Ditto but returns `bool`, only usable in unittests
+version (unittest)
+public bool isValidSignature (BlockHeader header, PreimagesMap preimage_map,
+    PubRMap pub_R_map, Point[] pub_keys) nothrow @safe
+{
+    return isInvalidSignatureReason(header, preimage_map, pub_R_map, pub_keys) is null;
+}
 
 /*******************************************************************************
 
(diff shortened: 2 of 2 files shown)


source/agora/common/crypto/Schnorr.d
```diff
@@ -54,7 +54,7 @@ import agora.common.Serializer;
 
 import std.algorithm;
 import std.range;
-
+import std.stdio;
```
TrustHenry: Not used here.
It would be better to use it in unittest code.

AndrejMitrovic:
This is a proof of concept, just a prototype. There's no need to review the styling / import issues, because this will be reworked. :)

I need a review from @Geod24 on the safety of the signature scheme itself.

AndrejMitrovic:
Update: Replaced use of an array of preimages with a hashmap of preimages. It remains a question of which data type to use to make it the most space-efficient..

TrustHenry:
When I saw the PR that's been uploaded so far,
Validators are stored in the operating manager by the operating process.
When sorting the PublicKey in the order of alignment from proven enrolment data
Is this correct to apply sequentially to bitFiled?

Here we think that when nodes with the same Publickey are enrolling multiple UTXOs, we have to sort out the second order.

AndrejMitrovic:
> Here we think that when nodes with the same Publickey are enrolling multiple UTXOs, we have to sort out the second order.

I'm not sure what you mean by enrolling multiple UTXOs. The node signs with a single key.

source/agora/common/crypto/Schnorr.d
```diff
@@ -64,6 +64,481 @@ nothrow @nogc unittest
...
+                dg = serialize function accumulator
+
+        ***************************************************************************/
+
+        public void serialize (scope SerializeDg dg) const @trusted
+        {
+            serializePart(this.validators, dg);
+            serializePart(this.signature, dg);
+
+            serializePart(this.preimages.length, dg);
```
Geod24: Here we essentially serialize the number of validators twice: first as the bitfield length (which derives exclusively from the number of validators), then as the length of `preimages`.

source/agora/common/crypto/Schnorr.d
```diff
@@ -64,6 +64,481 @@ nothrow @nogc unittest
...
+    ///
+    void signBlock (ref Block block, Pair key_pair, Pair rand_pair)
+    {
+        // the challenge (the Block hash, which includes the preimages)
+        auto msg = block.hashFull();
+
+        auto sig = signChallenge(key_pair.v, rand_pair.V, rand_pair.v, msg);
+
+        block.signature.R = combine(block.signature.R, sig.R);
+        block.signature.s = combine(block.signature.s, sig.s);
```
Geod24: Why not use `block.signature.R + sig.R` ?

source/agora/common/crypto/Schnorr.d
```diff
@@ -64,6 +64,481 @@ nothrow @nogc unittest
...
+    if (elems.length == 1)
+        return elems[0];
+    else
+        return sum(elems[1 .. $], elems[0]);  // proper accumulator, not T.init!
+}
+
+/// Returns: The accumulated values
+/// Note: be aware of accumulators, when we add to a value we must
+/// ensure the value was initialized with an accumulator (not T.init!)
+T combine (T)(T lhs, T rhs)
```
Geod24: I don't quite understand the need for `combine` ? We have operator overload and the array version doesn't seem to be used.

source/agora/common/crypto/Schnorr.d
```diff
@@ -64,6 +64,481 @@ nothrow @nogc unittest
...
+
+/// validator signature scheme
+unittest
+{
+    import agora.common.BitField;
+    import std.array;
+    import std.algorithm;
+
+    /// example block structure
+    static struct Block
```
Geod24: That looks like a lot of code that shouldn't be in `common/crypto/Schnorr.d`, no?

source/agora/common/crypto/Schnorr.d
```diff
@@ -64,6 +64,481 @@ nothrow @nogc unittest
...
+        foreach (_; 0 .. 10)
+            result ~= result[$ - 1].hashFull();
+
+        return result;
+    }
+
+    ushort getIndex (Point key)
+    {
+        auto res = pub_keys.countUntil(key);
+        assert(res != -1);
```
Geod24: Instead of checking for `-1`, check for negativity

source/agora/common/crypto/Schnorr.d
```diff
@@ -223,6 +698,29 @@ public Signature sign (T) (
     return Signature(R, s);
 }
 
+/// Overload of 'sign' which takes a specific challenge
+private Signature signChallenge (const ref Scalar k, const ref Point R,
```
Geod24: Why is this function necessary ? `sign` should do the job. Also the extra hashing will commit to `R`, while this does not.

source/agora/common/crypto/Schnorr.d
```diff
@@ -64,6 +64,481 @@ nothrow @nogc unittest
...
+
+        Pair n1_rand_pair1 = Pair(n1_rand1, n1_rand1.toPoint());
+        Pair n2_rand_pair1 = Pair(n2_rand1, n2_rand1.toPoint());
+        Pair n3_rand_pair1 = Pair(n3_rand1, n3_rand1.toPoint());
+
+        // add signatures (todo: not validating yet, need Enrollment data struct)
+        signBlock(block_1, pair_keys[0], n1_rand_pair1);
+        signBlock(block_1, pair_keys[1], n2_rand_pair1);
+        signBlock(block_1, pair_keys[2], n3_rand_pair1);
+    ///
```
Geod24: ?

source/agora/common/crypto/Schnorr.d
```diff
@@ -64,6 +64,481 @@ nothrow @nogc unittest
...
+        return verifyChallenge(Ps, block.signature, msg);
+    }
+
+    /** enrollment data for each validator (preimages) */
+    Hash[] n1_preims = genPreimages(Scalar.random());
+    Hash[] n2_preims = genPreimages(Scalar.random());
+    Hash[] n3_preims = genPreimages(Scalar.random());
+
+    /// Block #1
+        Block block_1;
```
Geod24: Alignment off ?

Geod24 (changes requested):
Not completely done with review, but my first comment is to keep everything that is blockchain-centric outside of `common`. If we can avoid the notion of blocks & the likes spilling in this package it'd be great.
Second, I'm not sure what `combine` and the `*Challenge` routines are for. Commented on them directly.
I'll go back to the test later today / tmr to make sure I understood them correctly.

source/agora/common/crypto/Schnorr.d
AndrejMitrovic: It was just to make it visually easier to distinguish, sorry.

source/agora/common/crypto/Schnorr.d
AndrejMitrovic: The problem I had was with the default initializer of Point:

```D
Point p1 = Scalar.random().toPoint;
Point p2 = Scalar.random().toPoint;

struct Block
{
    Point p;
}

Block block;
block.p = block.p + p1;
block.p = block.p + p2;

writefln("p1: %s", p1);
writefln("p2: %s", p2);
writefln("p1 + p2: %s", p1 + p2);
writefln("block.p: %s", block.p);
assert(block.p == p1 + p2);  // fails
```

```
p1: 0x9a59d8f2d8123d144bf646363f441e341352cb34210d09a90c0cab46e0bab355
p2: 0x4a65a474377deb3f200b30a69c5f9789a70f1bef1b7d693bd88f42d645baee2d
p1 + p2: 0xc764ceb8e99503bda74a56fc7806d2154f30bb89034f7e3ea6245f078d5bf668
block.p: 0xdc1a8018205117574df5011b3cc0755313122621c7b1a6d16fa00f2a0267e69a
```

source/agora/common/crypto/Schnorr.d
AndrejMitrovic: > the array version doesn't seem to be used.

Indeed, it's dead code now. I'll remove it.

source/agora/common/crypto/Schnorr.d
AndrejMitrovic: Yes. It's inside a unittest. This PR was meant to be just a PoC to see if the scheme makes sense. 👼

AndrejMitrovic:
So your idea was just to use the existing `sign` functionality already in `Schnorr.d`? But the problem with that scheme is that it only works with N of N, not X of N, where X <= N. If you agree on a common `R`, then everyone must sign with their `r` to make the signature valid. It doesn't allow a node to reveal a preimage and sign later after the block was externalized.

For example:

```D
unittest
{
    static immutable string secret = "string";

    Pair kp1 = Pair.random();
    Pair kp2 = Pair.random();
    Pair kp3 = Pair.random();
    Pair R1 = Pair.random();
    Pair R2 = Pair.random();
    Pair R3 = Pair.random();
    Point R = R1.V + R2.V + R3.V;
    Point X = kp1.V + kp2.V + kp3.V;

    auto sig1 = sign(kp1.v, X, R, R1.v, secret);
    auto sig2 = sign(kp2.v, X, R, R2.v, secret);
    auto sig3 = sign(kp3.v, X, R, R3.v, secret);

    auto sig_multi = Signature(R, sig1.s + sig2.s + sig3.s);
    assert(verify(X, sig_multi, secret));

    auto sig_partial = Signature(R, sig1.s + sig2.s);
    assert(verify(X, sig_partial, secret));  // fails
}
```

AndrejMitrovic:
Hmm maybe it would work. Give me an hour to test this.

source/agora/common/crypto/Schnorr.d
Geod24: Looks like we should disable default construction for `Point`

AndrejMitrovic: I guess so. The starting point is G, so it's not zero, haha.

AndrejMitrovic: Actually I don't think it's G, maybe it's just whatever the binary representation makes it to be..

Geod24:
> So your idea was just to use the existing sign functionality already in Schnorr.d? 

Yep

> But the problem with that scheme is that it only works with N of N, not X of N, where X <= N.

Acktually... The "Complex API" has no connection whatsoever between `x` and `X` so you could just commit to a `X` which is not `x.G`

> It doesn't allow a node to reveal a preimage and sign later after the block was externalized.

Do we want to allow this though ?

AndrejMitrovic:
> The "Complex API" has no connection whatsoever between x and X so you could just commit to a X which is not x.G

I don't think I can, if you look a the way verify is currently implemented: https://github.com/bpfkorea/agora/blob/e359de96d5e7e9c4b14c501fa6534ae50eaee92d/source/agora/common/crypto/Schnorr.d#L245 and https://github.com/bpfkorea/agora/blob/e359de96d5e7e9c4b14c501fa6534ae50eaee92d/source/agora/common/crypto/Schnorr.d#L247

We need X in `hashFull(Message!T(X, s.R, data));` to be separate from X in `Point RcX = s.R + (c * X)`, is this what you mean? That's basically what I did in my own overloads of sign/verify, although maybe I can make it nicer.

Geod24:
Oh right. Let's just take `X` out of the `Hash` then.

AndrejMitrovic:
> Oh right. Let's just take X out of the Hash then.

Great. Then I think this could work. _puts on crypto hat_

AndrejMitrovic:
Ok, so simplifying the scheme and re-using existing code in `Schnorr.d` I came up with this scheme (PR updated):

- Nodes can sign blocks only if they reveal their #Nth preimage.
- Nodes agree on the set of preimages they will put their signatures on because R is part of the hash.
- If a node failed to reveal its preimage in block 1, it cannot sign the next block 2.
- If a node reveals its preimage, it **must** also sign the block. Otherwise signature validation will fail.

Does that sound reasonable?

-----

You said it should be possible for nodes to reveal several preimages in advance. If the last point can't be worked around, if a node must sign a block it has revealed a preimage for, then is there any benefit of revealing that preimage?

source/agora/common/crypto/ECC.d
```diff
@@ -178,6 +178,20 @@ public struct Scalar
...
+    /***************************************************************************
+
+        Implements hashing support
+
+        Params:
+            dg = hashing function
+
+    ***************************************************************************/
+
+    public void computeHash (scope HashDg dg) const nothrow @safe @nogc
```
Geod24: This can go

source/agora/common/crypto/Schnorr.d
Geod24: Ping

source/agora/common/crypto/Schnorr.d
```diff
@@ -64,6 +64,447 @@ nothrow @nogc unittest
...
+        Signature signature;
+
+        /// The revealed preimages for this block
+        Hash[ushort] preimages;
+
+        /// Block data
+        string data;
+
+        /// Block hash includes the preimages
+        void computeHash (scope HashDg dg) const nothrow @nogc
```
Geod24: This can go

source/agora/common/crypto/Schnorr.d
AndrejMitrovic: The issue is specifically for `Point`, I'll remove the calls to `combine` for `Scalar` as they work ok. The issue is:

```D
unittest
{
    Point p1 = Scalar.random().toPoint();
    Point p2 = Scalar.random().toPoint();

    Point sum1 = p1 + p2;

    // the correct way is to initialize it to p1, and then add the other points
    Point sum2;
    foreach (point; [p1, p2])
        sum2 = sum2 + point;

    assert(sum2 == sum1);  // fails
}
```

It doesn't make sense to add to a point that has been default-initialized. So I wrote a helper function. I could just write this code inline without the function I guess, but I wanted to make this edge-case visible, because it seems to me like we could introduce bugs if we're not careful with using `Point.init`.

AndrejMitrovic:
Reworking this to move validation into `Validation.d` and make it more easy to digest.

AndrejMitrovic:
It's a good thing I started refactoring this. I realized my code was calculating R with the formula `R2 = R1 + X2` instead of `R2 = R1 + X1` (so it used the current preimage instead of the previous). This isn't necessarily bad, but I'll have this fixed soon anyway.

I haven't pushed any changes yet, still WIP.

AndrejMitrovic:
I will reopen this as another PR because there are many comments on outdated code.

Issue #432, opened 2019-11-27
Lift compile-time restriction on hashing / serialization / deserialization


Geod24:
Currently we have restrictions on hashing, serialization, and deserialization, like this:
https://github.com/bpfkorea/agora/blob/c38249d20122430b68a7d7d58bc8d12f1bf46868/source/agora/common/Hash.d#L122-L125

The "common" approach would have been to use `tupleof` to just iterate over all struct members. But this approach could easily be misused, by hashing/serializing some private elements of the record, such as a cached value.

However this has led to a lot of repetitive code, and even helpers to work around it: https://github.com/bpfkorea/agora/blob/c38249d20122430b68a7d7d58bc8d12f1bf46868/source/agora/common/Serializer.d#L21-L41

So should we just do away with the limitation ? Is there another approach that would work better ?

AndrejMitrovic:
how about something like `serializeEasy(data)` which will "just work", and `serializeFull` which requires hand-written support?

Geod24:
When would you use `serializeFull` then ?

AndrejMitrovic:
Well maybe it would be called `serializeSafe` or something. Basically what I mean is, you use `serializeEasy` when you just want to quickly serialize something (really useful during testing), and use `serializeFull` when you need to send data over the network to another node.

Is that distinction useful?

Other solutions might be to use UDAs. Something like `@DontSerialize` and having a single serialization function that works with .tupleof on all the fields.

Geod24:
> when you just want to quickly serialize something

Do we have case where we want to serialize something that is not serialized outside of `unittest` ?
In my experience, it only happens during development / while working on the type itself. YMMV.
If it's for testing, we could just `version (unittest)` it.

> Other solutions might be to use UDAs.

For clarify, what I suggest is to call `serialize` if it's implemented, otherwise default to `tupleof`.

AndrejMitrovic:
> In my experience, it only happens during development 

Yes, my thoughts are the same.

> For clarify, what I suggest is to call serialize if it's implemented, otherwise default to tupleof.

Aha. Ok with me.

PR #434, opened 2019-11-28
Remove hashing requirement to have `computeHash` implemented

Geod24:
Part of #432 

```
`hashPart` used to require that `computeHash` be implemented by aggregates.
However, this proved very inefficient and led to both abstraction leaks,
and repetitive code.

This commit lifts this limitation, using 'tupleof' as the default hashing mechanism.
Aggregates still have the ability to override the way hashing is done
by explicitly implementing `computeHash`.

Note that not having to implement `computeHash` makes aggregates with special requirements
(namely `Block` and `Input`) more obvious to the reader,
as otherwise the override gets lost in the noise.
```

Dev team, LGTY ?

### DIFF
--- a/source/agora/common/Hash.d
+++ b/source/agora/common/Hash.d
@@ -160,6 +160,13 @@ public void hashPart (scope const(char)[] record, scope HashDg state) /*pure*/ n
     state(cast(const ubyte[])record);
 }
 
+/// Ditto
+public void hashPart (T) (scope const auto ref T[] records, scope HashDg state)
+{
+    foreach (ref record; records)
+        hashPart(record, state);
+}
+
 // Test that the implementation actually matches what the RFC gives
 nothrow @nogc @safe unittest
 {
@@ -243,6 +250,9 @@ nothrow @nogc @safe unittest
         "c2e756f276c112342ff1d6f1e74d05bdb9bf880abd74a2e512654e12d171a74");
 
     assert(hashMulti(foo, bar) == merged);
+
+    const Hash[2] array = [foo, bar];
+    assert(hashFull(array[]) == merged);
 
     static struct S
     {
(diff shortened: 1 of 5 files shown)


source/agora/common/Hash.d
```diff
@@ -119,15 +127,13 @@ public void hashPart (T) (scope const auto ref T record, scope HashDg state)
...
+    else static if (__traits(compiles, () { const ubyte[] r = T.init[]; }))
         state(record[]);
     else
-    {
-        static assert(is(typeof(T.init.computeHash(HashDg.init))),
-                      "Struct `" ~ T.stringof ~
-                      "` does not implement `computeHash(scope HashDg) const nothrow @nogc` function");
-        record.computeHash(state);
-    }
+        foreach (ref field; record.tupleof)
```
AndrejMitrovic: `const ref` or?

AndrejMitrovic:
Lots of red, I like it.

Geod24:
Me too comrade.

source/agora/common/Hash.d
TrustHenry: Are all "tuple of type" except in case of Includes computeHash or "ubyte[]"?

Geod24: I'm not sure what you mean by "tuple of type".
Unless the type defines a compatible `computeHash` function OR can be sliced into a `const(ubyte)[]`, we call `tupleof`, which iterates over every field in sequence.
Note that this is calling `tupleof` on the instance, so it iterates on the field. If we call `tupleof` on the type, we iterate on the types of fields.

Geod24: @AndrejMitrovic : Fixed

AndrejMitrovic (approved):

MichaelKim20 (commented):
Good job.

MichaelKim20 (approved):

PR #435, opened 2019-11-28
Add opEquals to vector bindings

AndrejMitrovic:
This also makes sure any POD types using
vector internally will have the correct
behavior when doing equality checks.

### DIFF
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -235,6 +235,24 @@ extern(C++, (StdNS!())) struct vector (T, Alloc = allocator!T)
             return this._start[start .. end];
         }
 
+        public bool opEquals (const ref vector rhs) const pure nothrow @nogc @safe
+        {
+            import std.range : zip;
+            if (this.length != rhs.length)
+                return false;
+
+            // note: cannot do 'return this.innerSets[] == rhs.innerSets[];'
+            // object.d(358,64): Error: `cast(const(vector))(cast(const(vector)*)r)[i]`
+            // is not an lvalue and cannot be modified
+            foreach (const ref left, const ref right; zip(this[], rhs[]))
+            {
+                if (left != right)
+                    return false;
+            }
+
+            return true;
+        }
+
         alias opDollar = length;
 
         string toString() const @trusted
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
I also wanted to add support for `vector!ubyte vec2 = vec1`, but this doesn't call `opAssign` or the ctor, it just calls the postblit. It's a bit more complicated to implement, sigh.

Geod24:
> I also wanted to add support for 

Well copy constructors are in beta IIRC (very buggy tho)

source/scpd/Cpp.d
```diff
@@ -235,6 +235,23 @@ extern(C++, (StdNS!())) struct vector (T, Alloc = allocator!T)
             return this._start[start .. end];
         }
 
+        public bool opEquals (in vector rhs) const pure nothrow @nogc @trusted
```
Geod24: BY value ? 👎

source/scpd/Cpp.d
```diff
@@ -235,6 +235,23 @@ extern(C++, (StdNS!())) struct vector (T, Alloc = allocator!T)
...
+            if (this.length != rhs.length)
+                return false;
+
+            // note: cannot do 'return this.innerSets[] == rhs.innerSets[];'
+            // object.d(358,64): Error: `cast(const(vector))(cast(const(vector)*)r)[i]`
+            // is not an lvalue and cannot be modified
+            foreach (idx, const ref val; this[])
+            {
+                if (val != rhs[idx])
+                    return false;
```
Geod24, line 250: This line is not covered

source/scpd/Cpp.d
Geod24: Why the `@trusted` ?

Geod24 (changes requested):

source/scpd/Cpp.d
AndrejMitrovic: It's three pointers though. Isn't it more efficient to pass three pointers instead of a pointer to three pointers and do double indirection that way? I'd have to look up an assembly book to confirm. :P

AndrejMitrovic: Well, `opIndex` is not `@trusted`. The question is, should it be? I'm not sure.

Geod24: Maybe (probably?) but I'm more concerned about passing C++ structs by value. Even though we don't currently implement the copy constructor that does the RC, we might in the future.

AndrejMitrovic: Aha. That's a good point actually.

AndrejMitrovic: I always thought you were using `ref` due to performance, I forgot about the mess D has with copy ctors/postblits/etc.

Geod24: > The question is, should it be? I'm not sure

It seems to offer a `@safe` interface when assertions are on.

AndrejMitrovic:
Fixed. Using `zip` now to walk through both vectors, rather than using the unsafe `opIndex`.

Geod24 (approved):

PR #436, opened 2019-11-28
Make coverage work with integration tests

Geod24:
```
The only thing we need to do is to mount `source` and `submodules`
in the working directory so that the coverage code can access it
(otherwise it's unable to generate reports).
Since coverage won't be enabled very often, or by end user,
the option to store it in the data directory is removed
(it also became moot since the reshuffling of the docker image).
```

The Dockerfile has also been slightly improved.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -46,4 +46,11 @@ jobs:
           command: |
             source $HOME/dlang/dmd-2.088.0/activate
             ci/system_integration_test.d
+            # Work around druntime setting the permissions to 600..
+            sudo chmod 644 tests/system/node/*/*.lst
           no_output_timeout: 15m
+      - run:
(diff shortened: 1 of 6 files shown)


Geod24:
> bash: line 1334: ./tests/system/node/0/submodules-vibe-core-source-vibe-core-path.lst: Permission denied

All files are permission 600, of course...
https://github.com/dlang/druntime/blob/873fac33014c5af680c4bed69bb74cb0f192198a/src/rt/cover.d#L419

AndrejMitrovic:
Restarted the only failure on travis. Seems like a timing issue.

AndrejMitrovic (approved):

Geod24:
Merging since it was approved and is green (and yes it was a timing issue)

PR #437, opened 2019-11-28
 Remove requirement for aggregates to implement [de]serialize

### LINKED ISSUE #432: Lift compile-time restriction on hashing / serialization / deserialization

Geod24:
```
Aggregates are now default-[de]serialized to the tupleof of their members by default.
Any aggregate wishing a special handling can do so by defining '[de]serialize' as before.
Additionally, `DefaultSerializer` and `DefaultDeserializer` have been removed,
as they are now redundant with the default.
```

Fixes #432 

The `serialize` / `deserialize` methods haven't been removed yet, as there are some more room for improvements (in particular for deserialization) that require some more research.

### DIFF
--- a/source/agora/common/BanManager.d
+++ b/source/agora/common/BanManager.d
@@ -52,9 +52,6 @@ public class BanManager
         /// To set an IP as banned, we simply set its un-ban time in the future.
         /// By default it's set to the past (therefore un-banned)
         private time_t banned_until = 0;
-
-        mixin DefaultSerializer!();
-        mixin DefaultDeserializer!();
     }
 
     /// configuration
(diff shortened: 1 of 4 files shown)


AndrejMitrovic (approved):

AndrejMitrovic:
Hope it doesn't break anyone's PR. 🤞

PR #443, opened 2019-12-06
Added unit test for FreezeAmount on Amount

TrustHenry:
Make sure there are no decimal places in the FreezeAmount.
Compare the same as 40,000BOA.
It tests the function of Amount.integral

### DIFF
--- a/source/agora/common/Amount.d
+++ b/source/agora/common/Amount.d
@@ -259,6 +259,9 @@ pure @safe nothrow @nogc unittest
 
     assert(Amount(100_500_000).integral() == 10);
     assert(Amount(100_500_000).decimal() == 500_000);
+
+    assert(Amount.FreezeAmount.decimal() == 0);
+    assert(Amount.FreezeAmount.integral() == 40_000);
 }
 
 unittest


source/agora/common/Amount.d
```diff
@@ -259,6 +259,11 @@ pure @safe nothrow @nogc unittest
 
     assert(Amount(100_500_000).integral() == 10);
     assert(Amount(100_500_000).decimal() == 500_000);
+
+    assert(Amount.FreezeAmount.decimal() == 0);
+    assert(Amount(400_000_000_000, true).integral() == Amount.FreezeAmount.integral());
```
Geod24: The decimal test is good, but testing a constant is rarely adding value, as those rarely change.
I would just replace this and following lines by `assert(Amount.FreezeAmount.integral() == 40_000);`, which is testing an expectation with the integral function.

TrustHenry: applied the suggest.

Geod24 (approved):

MichaelKim20 (approved):

linked0 (approved):

PR #444, opened 2019-12-09
Changed the frozen amount to be at least 40,000 BOA

TrustHenry:
according to the Yellow White Paper
> It has at least 40,000 coins

### DIFF
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -94,8 +94,8 @@ public string isInvalidReason (const Transaction tx, UTXOFinder findUTXO,
                 return "Transaction: Can only freeze a Payment transaction";
         }
 
-        if (sum_unspent != Amount.FreezeAmount)
-            return "Transaction: Only available when the amount is 40,000";
+        if (sum_unspent.integral() < Amount.MinFreezeAmount.integral())
+            return "Transaction: available when the amount is at least 40,000 BOA";
     }
     else if (tx.type == TxType.Payment)
     {
@@ -358,7 +358,7 @@ unittest
     {
         storage.clear;
         // Create the previous transaction with type `TxType.Payment`
-        previousTx = newCoinbaseTX(key_pairs[0].address, Amount.FreezeAmount);
+        previousTx = newCoinbaseTX(key_pairs[0].address, Amount.MinFreezeAmount);
         previousHash = hashFull(previousTx);
         foreach (idx, output; previousTx.outputs)
         {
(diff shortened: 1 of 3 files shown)


source/agora/consensus/Validation.d
```diff
@@ -468,8 +468,8 @@ unittest
         );
         secondTx.inputs[0].signature = key_pairs[0].secret.sign(hashFull(secondTx)[]);
 
-        // Second Transaction is invalid.
-        assert(!secondTx.isValid(findUTXO, 0));
+        // Second Transaction is valid.
+        assert(secondTx.isValid(findUTXO, 0));
```
AndrejMitrovic, line 472: Look at the comment in line 446. That comment needs to be changed too.

source/agora/consensus/Validation.d
```diff
@@ -94,8 +94,8 @@ public string isInvalidReason (const Transaction tx, UTXOFinder findUTXO,
                 return "Transaction: Can only freeze a Payment transaction";
         }
 
-        if (sum_unspent != Amount.FreezeAmount)
-            return "Transaction: Only available when the amount is 40,000";
+        if (sum_unspent.integral() < Amount.FreezeAmount.integral())
+            return "Transaction: available when the amount is least 40,000 BOA";
```
AndrejMitrovic:
```suggestion
            return "Transaction: available when the amount is at least 40,000 BOA";
```

source/agora/consensus/Validation.d
```diff
@@ -94,8 +94,8 @@ public string isInvalidReason (const Transaction tx, UTXOFinder findUTXO,
                 return "Transaction: Can only freeze a Payment transaction";
         }
 
-        if (sum_unspent != Amount.FreezeAmount)
```
AndrejMitrovic, line 97: `Amount.FreezeAmount` should then be renamed to `Amount.MinFreezeAmount`, to make it obvious that it's the minimum freeze amount, not exact freeze amount.

TrustHenry: applied the suggest

source/agora/consensus/Validation.d
TrustHenry: applied the suggest

source/agora/consensus/Validation.d
TrustHenry: Good eye :) applied the suggest

linked0 (approved):

AndrejMitrovic (approved):

Geod24:
@TrustHenry : Please leave an empty line between the title and the commit message.
Currently your title is over 100 character, which makes it very hard to read.
This is what I see in the console:
Screen Shot 2019-12-16 at 18 26 02

PR #41, opened 2019-06-24
Add documentation for API interface

Geod24:
@team : Feel free to make any suggestion / ask any question you see fit.

PR #445, opened 2019-12-10
Add enrollments struct to the Blockheader

TrustHenry:
This is to store the enrollment data in the block when creating the block.
Enrollment data is stored only in registered blocks.

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -23,6 +23,7 @@ import agora.common.Types;
 import agora.common.Deserializer;
 import agora.common.Hash;
 import agora.common.Serializer;
+import agora.consensus.data.Enrollment;
 import agora.consensus.data.Transaction;
 import agora.consensus.Genesis;
 
@@ -56,6 +57,8 @@ public struct BlockHeader
     /// Schnorr multisig of all validators which signed this block
     public Signature signature;
 
+    /// Enrolled validators
+    public Enrollment[] enrollments;
 
     /***************************************************************************
 
@@ -74,6 +77,8 @@ public struct BlockHeader
         dg(this.prev_block[]);
         hashPart(this.height, dg);
         dg(this.merkle_root[]);
+        foreach (enrollment; this.enrollments)
+            hashPart(enrollment, dg);
     }
 
     /***************************************************************************
@@ -92,6 +97,9 @@ public struct BlockHeader
         dg(this.merkle_root[]);
         serializePart(this.validators, dg);
         serializePart(this.signature, dg);
+        serializePart(this.enrollments.length, dg);
+        foreach (enrollment; this.enrollments)
+            serializePart(enrollment, dg);
     }
 
     /***************************************************************************
@@ -110,6 +118,12 @@ public struct BlockHeader
         this.merkle_root = Hash(dg(Hash.sizeof));
         deserializePart(this.validators, dg);
         deserializePart(this.signature, dg);
+
+        size_t enrollments_size;
+        deserializePart(enrollments_size, dg);
+        this.enrollments = uninitializedArray!(Enrollment[])(enrollments_size);
+        foreach (ref val; this.enrollments)
+            deserializePart(val, dg);
     }
 }
 
@@ -388,6 +402,10 @@ unittest
     validators[2] = true;
     validators[4] = true;
 
+    Enrollment[] enrollments;
+    enrollments ~= Enrollment.init;
+    enrollments ~= Enrollment.init;
+
     const KP = Pair.random();
     Block block =
     {
@@ -397,7 +415,8 @@ unittest
             height:      0,
             merkle_root: merkle,
             validators:  validators,
-            signature:   Signature(KP.V, KP.v)
+            signature:   Signature(KP.V, KP.v),
+            enrollments: enrollments,
         },
         txs: [ tx ],
         merkle_tree: [ merkle ],
--- a/tests/system/dub.json
+++ b/tests/system/dub.json
@@ -18,6 +18,7 @@
         "../../source/agora/common/Set.d",
         "../../source/agora/common/Types.d",
         "../../source/agora/consensus/data/Block.d",
+        "../../source/agora/consensus/data/Enrollment.d",
         "../../source/agora/consensus/data/Transaction.d",
         "../../source/agora/consensus/Genesis.d",
         "../../source/agora/node/API.d",


source/agora/consensus/data/Block.d
```diff
@@ -56,6 +57,8 @@ public struct BlockHeader
     /// Schnorr multisig of all validators which signed this block
     public Signature signature;
 
+    /// Enrollmented validators
```
linked0: /// Enrolled validators

source/agora/consensus/data/Enrollment.d
```diff
@@ -16,6 +16,22 @@ module agora.consensus.data.Enrollment;
...
+
+    Define enrollment Set
+    This is an enrolment set with sort order values.
+
+*******************************************************************************/
+
+public struct EnrollmentSet
+{
+    /// the validators' key indices which enrollment validators
+    public ushort index;
```
linked0: You could transfer this variable into the Enrollment structure and remove the EnrollmentSet structure. The name, EnrollmentSet is kind of confusing also which is not a matter if you remove it.

linked0 (changes requested):

source/agora/consensus/data/Block.d
TrustHenry: applied the suggest

source/agora/consensus/data/Enrollment.d
TrustHenry: applied the suggest

Geod24:
> source/agora/consensus/data/Block.d(26,8): Error: module `Enrollment` is in file 'agora/consensus/data/Enrollment.d' which cannot be read

You need to update the `dub.json` for the integration test suite

source/agora/consensus/data/Block.d
```diff
@@ -110,6 +118,15 @@ public struct BlockHeader
...
         deserializePart(this.validators, dg);
         deserializePart(this.signature, dg);
+
+        size_t enrollments_size;
+        deserializePart(enrollments_size, dg);
+        foreach (_; 0 .. enrollments_size)
+        {
+            Enrollment enrollment;
+            deserializePart(enrollment , dg);
+            this.enrollments ~= enrollment;
```
Geod24: This will perform one allocation per one allocation. It's *very* inefficient.
Instead, since you already know the size of the whole array, you can do the following:
```D
 size_t enrollments_size;
deserializePart(enrollments_size, dg); 
this.enrollments.length = enrollments_size; // Only do one allocation
foreach (ref val; this.enrollments)
  deserializePart(val, dg);
```

Note that if `this.enrollments` is `null`, all those statements are equals:
```D
size_t len = 42; // Not this, of course :)
this.enrollments.length = len;
this.enrollments = new typeof(this.enrollments)(len);
this.enrollments = new Enrollment[](len);
```

While this is more efficient, an array allocation leads to a default-initialization of every elements, and for this reason instead of setting the length / allocation with the GC, a common optimization is to use https://dlang.org/phobos/std_array.html#uninitializedArray

source/agora/consensus/data/Enrollment.d
Geod24: Why do you need the index here ?

Geod24 (changes requested):
- Serialization / deserialization is untested
- What's the end game here ? Does `enrollments` contain *all* the currently enrolled validators, or just the ones *in this block* ?

TrustHenry:
> Why do you need the index here ?

Do you think it would be better to use an associative array?
`    /// Enrolled validators
    public Enrollment[ushort] enrollments;`

TrustHenry:
> What's the end game here ? Does enrollments contain all the currently enrolled validators, or just the ones in this block ?

The purpose of this is to store the validator enrollment in the block.
The entire valid validator can be imported at any time from the enrolementManager.
It is stored on first enrollment and reenroll.

Geod24:
> Do you think it would be better to use an associative array?

This does not answer the question. What do you use the index for / How do you use it ?

> The purpose of this is to store the validator enrollment in the block.
> The entire valid validator can be imported at any time from the enrolementManager.
> It is stored on first enrollment and reenroll.

I'm sorry that is not quite clear. From this, I understand that you only store here *new* enrollment.
So let's say `block #41` had an empty array (`[]`), then an enrollment is received, `block #42` will have an array of length 1, and then `block #43` will have an empty array again ?

TrustHenry:
> This does not answer the question. What do you use the index for / How do you use it ?

I tried to save the sort order of bitfield.
But I should think again.
I have to decide which Enrollment to store in Enrollments.

The most simple thing is to store recommended to be 1:1 matching and bitfield.

At present, my thoughts only include validators who signed.

AndrejMitrovic:
Storing pre-existing enrollments in all subsequent blocks wouldn't make sense. There is an `N` in the enrollment structure, and it needs to begin from somewhere. It should begin from the block it was initially included in, and it should not be added to subsequent blocks.

And you don't need the index in the enrollment. If `bitfield[5]` is `true`, it means get the EnrollmentData from the `validator_set` at index 5 if all the items in the `validator_set` are sorted based on the utxo key.

I think the only thing that needs to be changed to support this is to add a `SORTED BY` clause in the `CREATE TABLE` statement?

AndrejMitrovic:
> I think the only thing that needs to be changed to support this is to add a SORTED BY clause in the CREATE TABLE statement?

Ah well, you would probably need to add a secondary key? Well it's technical details at this point, it doesn't need to be part of this PR.

AndrejMitrovic:
So IMO: Just make it an array, and get rid of the index.

TrustHenry:
I deleted index at the suggestion of everyone. :)

tests/system/dub.json
```diff
@@ -18,6 +18,7 @@
         "../../source/agora/common/Set.d",
         "../../source/agora/common/Types.d",
         "../../source/agora/consensus/data/Block.d",
+        "../../source/agora/consensus/data.Enrollment",
```
AndrejMitrovic: Shouldn't this be `data/Enrollment.d`?

source/agora/consensus/data/Block.d
```diff
@@ -56,6 +57,8 @@ public struct BlockHeader
     /// Schnorr multisig of all validators which signed this block
     public Signature signature;
 
+    /// Enrolled validators
+    public Enrollment[] enrollments;
```
AndrejMitrovic, line 61: Shouldn't this be part of the hash?

AndrejMitrovic:
> This is to store the enrollment data in the block when creating the block. It is matched in the same array as the sort order of the bitfield.

So you went with the implementation where the enrollment data is copied for every block? Why?

tests/system/dub.json
TrustHenry: applied the suggest

source/agora/consensus/data/Block.d
TrustHenry: I thought of a better way, but now I'd better put it in and have a hashpart.

TrustHenry:
I'm changing to uninitializedArray.
Recognizing that many of the blocks do not have Enrollment Data, I will focus on minimizing capacity.

TrustHenry:
> So you went with the implementation where the enrollment data is copied for every block? Why?

Enrollment data is stored only in registered blocks.

source/agora/consensus/data/Enrollment.d
TrustHenry: I deleted index.

TrustHenry:
Ready for review.

source/agora/consensus/data/Block.d
```diff
@@ -74,6 +77,9 @@ public struct BlockHeader
         dg(this.prev_block[]);
         hashPart(this.height, dg);
         dg(this.merkle_root[]);
+        hashPart(this.enrollments.length, dg);
```
AndrejMitrovic: You shouldn't need to hash the length.

TrustHenry: I was thinking about whether to put it in or take it out. :)

AndrejMitrovic:
I think the two commits can be merged together? The testing commit tests something which was introduced in the previous commit, it should be part of the same commit then.

AndrejMitrovic (approved):

TrustHenry:
I merged commits.

AndrejMitrovic:
Your hardcoded hashes will probably fail, as they have changed after you removed `.length` from the hashing.

TrustHenry:
Thank you. I've changed the hash.

linked0 (approved):

TrustHenry:
Ready for review.

source/agora/consensus/data/Block.d
```diff
@@ -397,7 +415,8 @@ unittest
             height:      0,
             merkle_root: merkle,
             validators:  validators,
-            signature:   Signature(KP.V, KP.v)
+            signature:   Signature(KP.V, KP.v),
+            enrollments: enrollments
```
Geod24:
```suggestion
            enrollments: enrollments,
```

Trailing commas are legal in D

TrustHenry: applied the suggest

Geod24 (approved):

PR #447, opened 2019-12-11
Update list of supported compilers

Geod24:
Only use DMD v2.089.1 due to the swath of regressions with C++ support.
Also drop any LDC version that is not the most recent.
Once LDC 1.19.0 gets released, we can replace LDC 1.18.0 with it
(however the Docker-dependent tests will still need an upstream update).


AndrejMitrovic (approved):

Geod24:
Ah 2.089.1 is still in beta... Let's wait a few days.

Geod24:
DMD 2.089.1 has been released

PR #213, opened 2019-08-21
Small refactor & segfault workaround

AndrejMitrovic:
Split off from https://github.com/bpfkorea/agora/pull/193

PR #448, opened 2019-12-12
Add a function of getting UTXOs with a public key

linked0:
It is used to get a node's UTXOs during making a enrollment data as follows.
The EnrollmentManager get UTXOs from the UTXO set and select a frozen UTXO having 40000 BOA or above. In order to search UTXOs with public key, I change the scheme of the utxo_map table to having a column which contains the hash of a public key.

#406

### DIFF
--- a/source/agora/common/EnrollmentManager.d
+++ b/source/agora/common/EnrollmentManager.d
@@ -20,6 +20,7 @@ import agora.common.Deserializer;
 import agora.common.Hash;
 import agora.common.Serializer;
 import agora.consensus.data.Enrollment;
+import agora.consensus.data.UTXOSet;
 import agora.utils.Log;
 
 import d2sqlite3.database;
@@ -120,18 +121,28 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public bool addEnrollment (Enrollment enroll) @safe
+    public bool addEnrollment (const ref Enrollment enroll) @safe nothrow
     {
         static ubyte[] buffer;
-        buffer.length = 0;
 
         // check if already exists
-        if (this.hasEnrollment(enroll.utxo_key))
+        try
         {
-            log.info("Rejected already existing enrollment: {}", enroll);
+            if (this.hasEnrollment(enroll.utxo_key))
+            {
+                this.logMessage("Rejected already existing enrollment",
+                    enroll);
+                return false;
+            }
+        }
+        catch (Exception ex)
+        {
+            this.logMessage("Exception occured in checking if " ~
+                "the enrollment data exists", enroll, ex);
             return false;
         }
 
+        buffer.length = 0;
         () @trusted { assumeSafeAppend(buffer); } ();
 
         scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
@@ -139,12 +150,29 @@ public class EnrollmentManager
             buffer ~= data;
         };
 
-        serializePart(enroll, dg);
+        try
+        {
+            serializePart(enroll, dg);
+        }
+        catch (Exception ex)
+        {
+            this.logMessage("Serialization error", enroll, ex);
+            return false;
+        }
 
-        () @trusted {
-            this.db.execute("INSERT INTO validator_set (key, val) VALUES (?, ?)",
-                enroll.utxo_key[], buffer);
-        }();
+        try
+        {
+            () @trusted {
+                this.db.execute("INSERT INTO validator_set (key, val) VALUES (?, ?)",
+                    enroll.utxo_key[], buffer);
+            }();
+
+        }
+        catch (Exception ex)
+        {
+            this.logMessage("Database operation error", enroll, ex);
+            return false;
+        }
 
         return true;
     }
@@ -182,15 +210,16 @@ public class EnrollmentManager
         Make an enrollment data for enrollment process
 
         Params:
-            frozen_utxo_hash = the hash of a frozen UTXO used to identify
-                                a validator and to generate a siging key
+            frozen_utxo_hash = the hash of a frozen UTXO used to identify a validator
+                        and to generate a siging key
+            enroll = will contain the Enrollment if created
 
         Returns:
-            an enrollment data
+            true if the enrollment manager succeeded in creating the Enrollment
 
     ***************************************************************************/
 
-    public ref const(Enrollment) createEnrollment (const ref Hash frozen_utxo_hash) @trusted
+    public bool createEnrollment (Hash frozen_utxo_hash, out Enrollment enroll) @trusted nothrow
     {
         static ubyte[] buffer;
         buffer.length = 0;
@@ -218,39 +247,56 @@ public class EnrollmentManager
             buffer ~= data;
         };
 
-        serializePart(this.signature_noise, dg);
+        try
+        {
+            serializePart(this.signature_noise, dg);
+        }
+        catch (Exception ex)
+        {
+            this.logMessage("Serialization error", enroll, ex);
+            return false;
+        }
 
-        () @trusted {
-            auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM node_enroll_data " ~
-                "WHERE key = ?)", "signature_noise");
-            if (results.oneValue!(bool))
-            {
-                this.db.execute("UPDATE node_enroll_data SET val = ? WHERE key = ?",
-                    buffer, "signature_noise");
-            }
-            else
-            {
-                this.db.execute("INSERT INTO node_enroll_data (key, val) VALUES (?, ?)",
-                    "signature_noise", buffer);
-            }
-        }();
+        try
+        {
+            () @trusted {
+                auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM node_enroll_data " ~
+                    "WHERE key = ?)", "signature_noise");
+                if (results.oneValue!(bool))
+                {
+                    this.db.execute("UPDATE node_enroll_data SET val = ? WHERE key = ?",
+                        buffer, "signature_noise");
+                }
+                else
+                {
+                    this.db.execute("INSERT INTO node_enroll_data (key, val) VALUES (?, ?)",
+                        "signature_noise", buffer);
+                }
+            }();
+        }
+        catch (Exception ex)
+        {
+            this.logMessage("Database operation error", enroll, ex);
+            return false;
+        }
 
         // signature
         data.enroll_sig = sign(this.key_pair.v, this.key_pair.V, this.signature_noise.V,
             this.signature_noise.v, this.data);
 
-        return this.data;
+        enroll = this.data;
+        return true;
     }
 
     /***************************************************************************
 
-        Check if the enrollment data exists in the validator set.
+        Check if a enrollment data exists in the validator set.
 
         Params:
             enroll_hash = key for an enrollment data which is hash of frozen UTXO
 
         Returns:
-            true if the enrollment manager has the enrollment data
+            true if the validator set has the enrollment data
 
     ***************************************************************************/
 
@@ -264,7 +310,7 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Check if the enrollment data exists in the validator set.
+        Get the enrollment data with the key, and store it to 'enroll' if found
 
         Params:
             enroll_hash = key for an enrollment data which is a hash of a frozen
@@ -290,6 +336,35 @@ public class EnrollmentManager
 
         return false;
     }
+
+    /***************************************************************************
+
+        Logs message
+
+        Params:
+            msg = the log message to be logged
+            enroll = the Enrollment object, the information of which will be logged
+            ex = the Exception object, the message of which will be logged
+
+    ***************************************************************************/
+
+    private static void logMessage (string msg, const ref Enrollment enroll,
+        const Exception ex = null) @safe nothrow
+    {
+        try
+        {
+            if (ex !is null)
+            {
+                log.error("{}, enrollment:{}, exception:{}", msg, enroll, ex);
+            }
+            else
+            {
+                log.info("{}, enrollment:{}", msg, enroll);
+            }
+        }
+        catch (Exception ex)
+        {}
+    }
 }
 
 /// tests for member functions of EnrollmentManager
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
> The EnrollmentManager get UTXOs from the UTXO set and select a frozen UTXO having 40000 BOA or above.

@Geod24 so how does this work exactly, from a higher level? The node operator opens up the web UI, creates a freeze transaction, and then clicks a button "Enroll" and this EnrollmentManager picks whichever freeze UTXO is over 40K? I'm thinking about the user-facing side of things.

linked0:
> @Geod24 so how does this work exactly, from a higher level? The node operator opens up the web UI, creates a freeze transaction, and then clicks a button "Enroll" and this EnrollmentManager picks whichever freeze UTXO is over 40K? I'm thinking about the user-facing side of things.

We might need a web UI for this as Drey commented. At this moment, I'm writing code selecting a first matched frozon UTXO. If there will be a Web UI selecting an UTXO, the EnrollmentManager must have an API receiving the selected frozon UTXO on the Web UI.

linked0:
I also modified the EnrollmentManger's unittest to make it use the newly added function of the UTXO set.

linked0:
Ready for review.
This modification adds a unittest code getting UTXOs from the UTXO set and selecting a UTXO in creating a enrollment data.

source/agora/common/EnrollmentManager.d
```diff
@@ -290,6 +316,18 @@ public class EnrollmentManager
 
         return false;
     }
+
+    /***************************************************************************
+
+        Check if the EnrollmentData is right
+
+    ***************************************************************************/
+    public bool acceptEnrollment(Enrollment data) @safe
```
AndrejMitrovic: Is this leftover code? It doesn't seem to do anything except call `writeln`.

source/agora/common/EnrollmentManager.d
```diff
@@ -64,11 +71,14 @@ public class EnrollmentManager
...
                         :memory: was passed
             key_pair = the keypair of the owner node
+            utxo_set = the set of unspent outputs
 
     ***************************************************************************/
 
-    public this (string db_path, KeyPair key_pair)
+    public this (string db_path, KeyPair key_pair, UTXOSet utxo_set)
     {
+        this.utxo_set = utxo_set;
```
AndrejMitrovic: It's unused? Why was it added?

source/agora/common/EnrollmentManager.d
```diff
@@ -318,24 +356,45 @@ unittest
     );
     Hash secondHash = hashFull(secondTx);
 
+    // create the second transaction
+    Transaction thirdTx = Transaction(
```
AndrejMitrovic: the comment says second transaction, but the variable name is thirdTx.

source/agora/consensus/data/UTXOSet.d
```diff
@@ -381,9 +430,10 @@ private class UTXODB
         serializePart(value, dg);
 
         scope (failure) assert(0);
+        auto pubkey_hash = hashFull(value.output.address);
```
AndrejMitrovic: If you just store the public key instead of the hash of the public key, then you wouldn't need to do a hash operation on a public key every time a lookup is made. Also, a public key is 32 bytes, the hash is 64 bytes. So we would save on memory use as well.

source/agora/common/EnrollmentManager.d
```diff
@@ -190,13 +201,13 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public ref const(Enrollment) createEnrollment (const ref Hash frozen_utxo_hash) @trusted
+    public bool createEnrollment (Hash utxo_hash, out Enrollment enroll) @trusted nothrow
```
AndrejMitrovic: The documentation will need to be updated.

source/agora/common/EnrollmentManager.d
```diff
@@ -79,7 +89,8 @@ public class EnrollmentManager
         this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
             "(key CHAR(128) PRIMARY KEY, val BLOB NOT NULL)");
 
-        // set node's KeyPair object
+        // create Pair object from KeyPair object
+        this.pub_key = key_pair.address;
```
AndrejMitrovic: This one is not used anywhere.

source/agora/consensus/data/UTXOSet.d
linked0: > If you just store the public key instead of the hash of the public key, then you wouldn't need to do a hash operation on a public key every time a lookup is made. 

Do you think it is a right approach to serialize and deserialize a public key?

source/agora/common/EnrollmentManager.d
```diff
@@ -318,24 +335,45 @@ unittest
     );
     Hash secondHash = hashFull(secondTx);
 
+    // create the third transaction
+    Transaction thirdTx = Transaction(
+        TxType.Freeze,
+        [Input(Hash.init, 0)],
+        [Output(Amount(90_000), key_pairs[1].address)]
```
TrustHenry: `[Output(Amount.FreezeAmount, key_pairs[1].address)]`
The freeze amount must be a minimum of 40,000 BOA.

linked0:
Ready for review

AndrejMitrovic:
It looks good, although I think if serializing fails or adding an item to a database fails it probably means we have a bigger problem? In the current code the failures are not logged, and the function returns false.

linked0:
> In the current code the failures are not logged, and the function returns false.

After adding the `nothrow` attribute to the createEnrollment function, I must add a try/catch statement to the code in order to log failures. I mean the functions of the Logger is not `nothrow`. So It's not efficient to write the try/catch statements in every `nothrow` functions where logging is needed. So I will create another issue for making a common function for logging and make the addEnrollment function use it. If you have other opinion, please let me know.

linked0:
> So I will create another issue for making a common function for logging and make the addEnrollment function use it.

I will just add a function for this class which logs failures and request review again.

linked0:
Ready for review, though build errors occured on the Mac environments of Travis CI , which is related to the vibe.d.

Geod24:
Can you rebase ? Vibe.d errors should be fixed.

source/agora/common/EnrollmentManager.d
```diff
@@ -120,15 +121,25 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public bool addEnrollment (Enrollment enroll) @safe
+    public bool addEnrollment (const ref Enrollment enroll) @safe nothrow
     {
         static ubyte[] buffer;
         buffer.length = 0;
```
Geod24: On a side note, could you move  this just before `assumeSafeAppend` ? It's better for readability to have both of them together (however, keep it outside the `@trusted` block)

source/agora/common/EnrollmentManager.d
```diff
@@ -218,69 +247,85 @@ public class EnrollmentManager
...
+            hash = key for an enrollment data which is hash of frozen UTXO
 
         Returns:
-            true if the enrollment manager has the enrollment data
+            true if the validator set has the enrollment data
 
     ***************************************************************************/
 
-    public bool hasEnrollment (const ref Hash enroll_hash) @trusted
+    public bool hasEnrollment (const ref Hash hash) @trusted
```
Geod24: Can you split those cosmetic changes (renaming `enroll_hash` to `hash`, and the doc fixes) to another commit, possibly another PR ?

source/agora/common/EnrollmentManager.d
```diff
@@ -290,6 +335,48 @@ public class EnrollmentManager
...
+
+        Params:
+            msg = the log message to be logged
+            enroll = the Enrollment object, the information of which will be logged
+            ex = the Exception object, the message of which will be logged
+
+    ***************************************************************************/
+
+    private void logException(string msg, const ref Enrollment enroll,
+        const ref Exception ex) @safe nothrow
```
Geod24: `Exception` is a `class`, so no need for `ref`

source/agora/common/EnrollmentManager.d
```diff
@@ -290,6 +335,48 @@ public class EnrollmentManager
...
+
+        Logs message
+
+        Params:
+            msg = the log message to be logged
+            enroll = the Enrollment object, the information of which will be logged
+
+    ***************************************************************************/
+
+    private void logMessage(string msg, const ref Enrollment enroll) @safe nothrow
```
Geod24: Your goal is to have a `nothrow` log function, right ? From just looking at the function, it is not obvious.

Something like:
```
This function is a helper to deal with `log.info` not being `nothrow`, which forces to add a `try` / `catch` block when logging an exception from a `catch` block.
```

Also, you could merge those two by simply making `Exception` optional and default to `null` (which is possible if you remove `ref`).

source/agora/common/EnrollmentManager.d
```diff
@@ -290,6 +335,48 @@ public class EnrollmentManager
...
+        Logs exception message
+
+        Params:
+            msg = the log message to be logged
+            enroll = the Enrollment object, the information of which will be logged
+            ex = the Exception object, the message of which will be logged
+
+    ***************************************************************************/
+
+    private void logException(string msg, const ref Enrollment enroll,
```
Geod24: This can be `static` right ?

source/agora/common/EnrollmentManager.d
```diff
@@ -290,6 +335,48 @@ public class EnrollmentManager
...
+        const ref Exception ex) @safe nothrow
+    {
+        try
+        {
+            log.info("{}, enrollment:{}, exception:{}", msg, enroll, ex);
```
Geod24: I have a feeling an exception might warrant more than `info` ?
It depends if the exception can happen under normal operating circumstances or not.
For example, we should never log at a relevant log level when reading a stream of data from an *unauthenticated* source because it's a DoS vector.
However, when reading from an authenticated source, we can afford an `info` level.
But if it's an internal error that should not happen, `warn/error/fatal` might be warranted depending on the cases of failure. `info` / `trace` is still possible, if e.g. it's a timeout or something of the like.

Geod24 (commented):
Nice! Only a few nitpicks

source/agora/common/EnrollmentManager.d
linked0: It's right. The function is created to log messages without a `try` / `catch` block. And I will merge the two functions I created.

source/agora/common/EnrollmentManager.d
linked0: Our of curiosity,  can you explain the reason splitting those changes to another commit or PR?

Geod24: To keep aesthetic (non-functional) changes separated helps to increase the [signal-to-noise ratio](https://en.wikipedia.org/wiki/Signal-to-noise_ratio).

More precisely, if you make a commit / PR with *only* aesthetical changes, it is trivial to review: One just have to see if the name makes sense, and if there's no grammar errors or typos.
However, when such change is put along with functional changes, it creates a larger diff that draws attention away from the functional change.

Take https://github.com/bpfkorea/agora/pull/213 as an example: It contains some renames as well as functional changes. They are related to a much larger PR, which they were split from.

Now, that was the advantage for the reviewers, but there are other: since the commit is `git`'s atomic unit, when doing either a revert or a bisection, having smaller, more self contained commit will make those operation easier.

And finally, breaking down those changes helps the person making the PR get the bigger picture. People usually work incrementally on a feature: they try an approach, it doesn't work, or it's missing something, so they fix that. They will usually go through several iterations before finding a satisfying solution. When doing those iterations, one usually doesn't look at the full picture, but draws from previous experience (e.g.: "I know that this function works, so I am going to forget about its internals and just call it"). But when the iterations are over, splitting down commit make one consider the whole picture, and can lead to simplifications in already-working code (e.g. "I always call this function with a non-null parameter, so why not make it `ref` instead of a pointer ?").

That last paragraph is much more relevant for functional changes, of course. But most of the time, aesthetic changes come from refining the meaning / role of a piece of code, or getting a better understanding of it, which usually happens while doing functional changes :)

linked0: Thank you for the detailed explanation. I fully undersatand the purpose to make review process more efficient and  got to know the splitting of changes make the PR get big picture as you said.

linked0:
Ready for review.

source/agora/common/EnrollmentManager.d
```diff
@@ -79,7 +80,7 @@ public class EnrollmentManager
         this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
             "(key CHAR(128) PRIMARY KEY, val BLOB NOT NULL)");
 
-        // set node's KeyPair object
+        // create Pair object from KeyPair object
         this.key_pair.v = Scalar(key_pair.secret);
```
Geod24: I think we need to replace this with the newly merged function, don't we ?

linked0: I will do that on another PR. Can you agree with that?

source/agora/common/EnrollmentManager.d
```diff
@@ -120,31 +121,58 @@ public class EnrollmentManager
...
+            {
+                this.logMessage("Rejected already existing enrollment",
+                    enroll);
+                return false;
+            }
+        }
+        catch (Exception ex)
         {
-            log.info("Rejected already existing enrollment: {}", enroll);
+            this.logMessage("Exception occured in checking" ~
```
```suggestion
            this.logMessage("Exception occured in checking if "
```

source/agora/common/EnrollmentManager.d
```diff
@@ -218,39 +247,56 @@ public class EnrollmentManager
...
-            {
-                this.db.execute("INSERT INTO node_enroll_data (key, val) VALUES (?, ?)",
-                    "signature_noise", buffer);
-            }
-        }();
+        try
+        {
+            () @trusted {
+                auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM node_enroll_data " ~
+                    "WHERE key = ?)", "signature_noise");
```
Geod24, line 264: If the key is always the same, do we really need a key to begin with ? This is a single-entry table AFAICS.

source/agora/common/EnrollmentManager.d
```diff
@@ -290,6 +336,35 @@ public class EnrollmentManager
...
+        Logs message
+
+        Params:
+            msg = the log message to be logged
+            enroll = the Enrollment object, the information of which will be logged
+            ex = the Exception object, the message of which will be logged
+
+    ***************************************************************************/
+
+    private static void logMessage(string msg, const ref Enrollment enroll,
```
```suggestion
    private static void logMessage (string msg, const ref Enrollment enroll,
```

source/agora/common/EnrollmentManager.d
linked0: The table can have another data such as Pre-Images. So using key makes sense.

Geod24: Right, thanks for the clarification!

linked0:
Ready for review. (Build failed on linux though)

source/agora/common/EnrollmentManager.d
```diff
@@ -79,8 +80,12 @@ public class EnrollmentManager
         this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
             "(key CHAR(128) PRIMARY KEY, val BLOB NOT NULL)");
 
-        // set node's KeyPair object
+        // create Pair object from KeyPair object
+<<<<<<< HEAD
```
Geod24: Rebase error

linked0: I'm trying to resolve the error. Thank you.

linked0:
Ready for review.
Builds on Linux output errors. So the version of the setup-lang should be increased.

Geod24 (approved):
LGTM

PR #454, opened 2019-12-23
Add Key utils 'secretKeyToCurveScalar' SecretKey(Ed25519) to Scalar(X25519)

TrustHenry:
This converts from SecretKey(Ed25519) to Scalar(X25519)
Later this will be used in the Signature of enrollment verification.
#205

### DIFF
--- a/source/agora/common/crypto/Key.d
+++ b/source/agora/common/crypto/Key.d
@@ -14,6 +14,7 @@
 module agora.common.crypto.Key;
 
 import agora.common.crypto.Crc16;
+import agora.common.crypto.ECC;
 import agora.common.Types;
 import agora.common.Deserializer;
 import agora.common.Serializer;
@@ -365,4 +366,50 @@ unittest
     import std.string : representation;
     Signature sig = kp.secret.sign("Hello World".representation);
     assert(kp.address.verify(sig, "Hello World".representation));
+}
+
+/*******************************************************************************
+
+    Util to convert SecretKey(Ed25519) to Scalar(X25519)
+    The secretKeyToCurveScalar() function converts an SecretKey(Ed25519)
+    to a Scalar(X25519) secret key and stores it into x25519_sk.
+
+    Params:
+      secret = Secretkey in Ed25519 format
+
+    Returns:
+      Converted X25519 secret key
+
+*******************************************************************************/
+
+public static Scalar secretKeyToCurveScalar (SecretKey secret) nothrow @nogc
+{
+    Scalar x25519_sk;
+    if (crypto_sign_ed25519_sk_to_curve25519(x25519_sk.data[].ptr, secret[].ptr) != 0)
+        assert(0);
+    return x25519_sk;
+}
+
+///
+unittest
+{
+    import agora.common.crypto.Schnorr;
+
+    KeyPair kp = KeyPair.fromSeed(
+        Seed.fromString(
+            "SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4"));
+
+    Scalar scalar = secretKeyToCurveScalar(kp.secret);
+    assert(scalar ==
+        Scalar(`0x44245dd23bd7453bf5fe07ec27a29be3dfe8e18d35bba28c7b222b71a4802db8`));
+
+    Pair pair;
+    pair.v = scalar;
+    pair.V = pair.v.toPoint();
+
+    assert(pair.V.data == kp.address.data);
+    Signature enroll_sig = sign(pair, "BOSAGORA");
+
+    Point point_Address = Point(kp.address);
+    assert(verify(point_Address, enroll_sig, "BOSAGORA"));
 }


source/agora/common/crypto/Schnorr.d
```diff
@@ -209,6 +212,25 @@ public bool verify (T) (const ref Point X, const ref Signature s, auto ref T dat
...
+    return Point(ed25519_pk);
+}
+
+///
+unittest
+{
+    KeyPair kp = KeyPair.fromSeed(
+        Seed.fromString(
+            "SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4"));
+    Point point = pkToCurvePoint(kp.address);
```
Geod24: Can you add the `Scalar` corresponding to this `Point` ?

source/agora/common/crypto/Schnorr.d
```diff
@@ -209,6 +212,25 @@ public bool verify (T) (const ref Point X, const ref Signature s, auto ref T dat
     return S == RcX;
 }
 
+/// Publickey(Ed25519) to Point(Curve25519)
+public static Point pkToCurvePoint (PublicKey pk) nothrow @nogc
```
Geod24:
```suggestion
public Point pkToCurvePoint (PublicKey pk) nothrow @nogc
```

`static` is unnecessary as this function is at module level (unlike in C)

source/agora/common/crypto/Schnorr.d
TrustHenry: It's been solved with the help of Jay.

source/agora/common/crypto/Schnorr.d
TrustHenry: Applied the suggest.

TrustHenry:
Changed from existing code.
SKToCurveScalar method has been added.

linked0 (approved):
Nice!

source/agora/common/crypto/Schnorr.d
```diff
@@ -209,6 +212,38 @@ public bool verify (T) (const ref Point X, const ref Signature s, auto ref T dat
     return S == RcX;
 }
 
+/// SecretKey(Ed25519) to Scalar(X25519)
+public Scalar skToCurveScalar (SecretKey secret) nothrow @nogc
+{
+    ubyte[crypto_scalarmult_curve25519_BYTES] x25519_sk;
```
Geod24:
```suggestion
    Scalar x25519_sk;
```

source/agora/common/crypto/Schnorr.d
```diff
@@ -209,6 +212,38 @@ public bool verify (T) (const ref Point X, const ref Signature s, auto ref T dat
...
+
+    Scalar scalar = skToCurveScalar(kp.secret);
+    assert(scalar ==
+        Scalar(`0x44245dd23bd7453bf5fe07ec27a29be3dfe8e18d35bba28c7b222b71a4802db8`));
+
+    Pair pair;
+    pair.v = scalar;
+    pair.V = pair.v.toPoint();
+
+    Pair signature_noise = Pair.random();
```
Geod24: It was quite hard to recognize what this is actually doing. You don't need to generate a signature noise yourself, there's one overload which does it for you.
Also, it would be much simpler to just do:
```D
assert(pair.V.data == kp.address.data);
```

Geod24 (changes requested):
Amazing, it works ! After some research, turns out the curves are birationally equivalent. The author of libsodium recommends not using it, but for us, it's just a transitional measure.

But a few things need to be changed:
1) `agora.common.crypto.Schnorr` is related to *signature operation*, while the data structure live in `agora.common.crypto.Key` (Stellar-style `KeyPair`) and `agora.common.crypto.ECC` (`Pair`).
Since this function is only for conversion between data structures, it needs to go to either the `ECC` or `Key` module. Since we want to transition from `Key` to `ECC`, it is better to add the conversion function (and hence, the dependency) to the module we want to get rid of, that is, `Key`.

2) It needs better documentation ;)
3) It needs a better name... "sk" is too abbreviated.

source/agora/common/crypto/ECC.d
```diff
@@ -17,6 +17,8 @@
 
 module agora.common.crypto.ECC;
 
+import agora.common.crypto.Key;
+import agora.common.crypto.Schnorr;
```
Geod24: The import is because of `Pair` right ? If so I think it'd be better to move `Pair` here.

TrustHenry: Pair is only used in unittest.
I removed agora.common.crypto.Schnorr;

source/agora/common/crypto/Schnorr.d
TrustHenry: Applied the suggest.

source/agora/common/crypto/Schnorr.d
TrustHenry: Applied the suggest.

source/agora/common/crypto/ECC.d
TrustHenry: I moved to Key.D

TrustHenry:
Ready for review.

Geod24 (approved):
LGTM

PR #19, opened 2019-06-20
Add ability to provide DNS seeds

AndrejMitrovic:
I'm not sure we can test this in the repo until we implement integration testing. However a few manual test-runs did work with a DNS seed node.

PR #457, opened 2019-12-31
Update Github actions to v0.4.2

TrustHenry:
Cannot access 'latest' before initialization #19
The solution version is likely to be released.


TrustHenry:
It is being tested at different PRs at 0.50 release.
Re-enable LDC-master testing on Linux #462
So this PR is closed.

PR #460, opened 2020-01-02
Add the enrolled block_height column in the validatorSet DB.

TrustHenry:
Add a column to record the block height of validatorSet DB stored in the block.
This will be used as a reference for quick validation of the enrollmentData.
#205

### DIFF
--- a/source/agora/common/EnrollmentManager.d
+++ b/source/agora/common/EnrollmentManager.d
@@ -74,7 +74,7 @@ public class EnrollmentManager
 
         // create the table for validator set if it doesn't exist yet
         this.db.execute("CREATE TABLE IF NOT EXISTS validator_set " ~
-            "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
+            "(key BLOB PRIMARY KEY, val BLOB NOT NULL, enrolled_height INTEGER)");
 
         // create the table for enrollment data for a node itself
         this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
@@ -203,6 +203,72 @@ public class EnrollmentManager
     public void removeEnrollment (const ref Hash enroll_hash) @trusted
     {
         this.db.execute("DELETE FROM validator_set WHERE key = ?", enroll_hash[]);
+    }
+
+    /***************************************************************************
+
+        In validatorSet DB, return the enrolled block height.
+
+        Params:
+            enroll_hash = key for an enrollment block height
+
+        Returns:
+            the enrolled block height, or 0 if no matching key exists
+
+    ***************************************************************************/
+
+    public size_t getEnrolledHeight (const ref Hash enroll_hash) @trusted
+    {
+        try
+        {
+            auto results = this.db.execute("SELECT enrolled_height FROM validator_set" ~
+                " WHERE key = ?", enroll_hash[]);
+            if (results.empty)
+                return size_t.init;
+
+            return results.oneValue!(size_t);
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error {}", ex);
+            return size_t.init;
+        }
+    }
+
+    /***************************************************************************
+
+        Update the enrolled height of the validatorSet DB.
+
+        Params:
+            enroll_hash = enrollment blockheight to update enroll hash
+            block_height = enrolled blockheight
+
+        Returns:
+            true if the update operation was successful, false otherwise
+
+    ***************************************************************************/
+
+    public bool updateEnrolledHeight (const ref Hash enroll_hash,
+        const size_t block_height) @safe
+    {
+        try
+        {
+            if (this.getEnrolledHeight(enroll_hash) > 0)
+                return false;
+
+            () @trusted {
+                this.db.execute(
+                    "UPDATE validator_set SET enrolled_height = ? WHERE key = ?",
+                    block_height, enroll_hash[]);
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error, updateEnrolledHeight:{}, exception:{}",
+                enroll_hash, ex);
+            return false;
+        }
+        return true;
     }
 
     /***************************************************************************
@@ -443,4 +509,11 @@ unittest
 
     // test for getEnrollment with removed enrollment
     assert(!man.getEnrollment(utxo_hash2, stored_enroll));
+
+    // test for enrollment block height update
+    assert(!man.getEnrolledHeight(utxo_hash));
+    assert(man.updateEnrolledHeight(utxo_hash, 9));
+    assert(man.getEnrolledHeight(utxo_hash) == 9);
+    assert(!man.updateEnrolledHeight(utxo_hash, 9));
+    assert(man.getEnrolledHeight(utxo_hash2) == 0);
 }


source/agora/common/EnrollmentManager.d
```diff
@@ -74,7 +74,7 @@ public class EnrollmentManager
 
         // create the table for validator set if it doesn't exist yet
         this.db.execute("CREATE TABLE IF NOT EXISTS validator_set " ~
-            "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
+            "(key BLOB PRIMARY KEY, val BLOB NOT NULL, block_height INTEGER)");
```
linked0: 1. The name called `block_height` is kind of confusing. You could give clearer name to the column.

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,66 @@ public class EnrollmentManager
...
+
+    public bool updateEnrollmentBlockHeight (const ref Hash enroll_hash,
+        const size_t block_Height) @safe
+    {
+        size_t enrolled_hight = getEnrollmentBlockHeight(enroll_hash);
+        try
+        {
+            if (enrolled_hight > 0)
+            {
+                return false;
```
linked0, line 257: A logging is needed. You could use the function called `logMessage` of the EnrollmentManager class.

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,66 @@ public class EnrollmentManager
...
+                () @trusted {
+                    this.db.execute(
+                        "UPDATE validator_set SET block_height = ? WHERE key = ?",
+                        block_Height, enroll_hash[]);
+                }();
+            }
+        }
+        catch (Exception ex)
+        {
+            return false;
```
linked0, line 269: A logging is needed. You could the function called `logMessage` of the EnrollmentManager class.

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,66 @@ public class EnrollmentManager
...
+            enroll_hash = enrollment blockHeight to update enroll hash
+            block_Height = enrolled blockheight
+
+        Returns:
+            Update execution status
+
+    ***************************************************************************/
+
+    public bool updateEnrollmentBlockHeight (const ref Hash enroll_hash,
+        const size_t block_Height) @safe
```
linked0: You could use `block_height' instead of 'block_Height'. The doc should be changed also.

linked0 (changes requested):
Good idea and nice work!
My suggestions are as follows.

source/agora/common/EnrollmentManager.d
TrustHenry: The logMessage that was previously developed has many limitations. We only accept Enrollment only.

source/agora/common/EnrollmentManager.d
TrustHenry: Then what about the `enrolled_height`?

linked0: > Then what about the enrolled_height?

It's good. 'enrolled_block_height' is another option. :)

source/agora/common/EnrollmentManager.d
linked0: > The logMessage that was previously developed has many limitations. We only accept Enrollment only.

Right. The `logMessage` function should be made more general. After making the function more general I will take care of it.

source/agora/common/EnrollmentManager.d
TrustHenry: Applied the suggest. I changed 'enrolled_height'

linked0 (approved):

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,68 @@ public class EnrollmentManager
...
+            enroll_hash = key for an enrollment block height
+
+        Returns:
+            the enrolled block height.
+
+    ***************************************************************************/
+
+    public size_t getEnrolledHeight (const ref Hash enroll_hash) @safe
+    {
+        return () @trusted {
```
AndrejMitrovic: And if the enrollment doesn't exist, what is the return value?

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,68 @@ public class EnrollmentManager
...
+        Params:
+            enroll_hash = enrollment blockheight to update enroll hash
+            block_height = enrolled blockheight
+
+        Returns:
+            Update execution status
+
+    ***************************************************************************/
+
+    public bool updateEnrolledHeight (const ref Hash enroll_hash,
```
AndrejMitrovic, line 251: But why is this a public method? Shouldn't the block height be provided in the `addEnrollment` method, and then it could call this `updateEnrolledHeight` method internally?

TrustHenry: When `updateEnrolledHeight` method is executed, it is allowed to confirm the enrolled BlockHeight at the point stored in the block after `addEnrollment` method.

source/agora/common/EnrollmentManager.d
TrustHenry: Surely this creates 'no rows available' msg when there are keys don't match.
This is not an error, so we will correct it to return zero.

AndrejMitrovic: Could you add a test-case?

TrustHenry: Sure, added test-case code.

TrustHenry:
Ready for review.

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,80 @@ public class EnrollmentManager
...
+
+        Returns:
+            the enrolled block height.
+            Return size_t if no matching key exists.
+
+    ***************************************************************************/
+
+    public size_t getEnrolledHeight (const ref Hash enroll_hash) @safe
+    {
+        return () @trusted {
```
AndrejMitrovic: You can just mark the entire function as `@trusted`, it is the very same effect as making it `@safe` and having all of the code inside wrapped in a `@trusted` delegate.

TrustHenry: applied the suggestion.

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,78 @@ public class EnrollmentManager
...
+
+        Returns:
+            Update execution status
+
+    ***************************************************************************/
+
+    public bool updateEnrolledHeight (const ref Hash enroll_hash,
+        const size_t block_height) @safe
+    {
+        size_t enrolled_hight = getEnrolledHeight(enroll_hash);
```
```suggestion
        size_t enrolled_height = this.getEnrolledHeight(enroll_hash);
```

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,78 @@ public class EnrollmentManager
...
+        try
+        {
+            if (enrolled_hight > 0)
+            {
+                return false;
+            }
+            else
```
Geod24: That else is not needed since the `if` returns

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,78 @@ public class EnrollmentManager
...
+    /***************************************************************************
+
+        Update the enrolled height of the validatorSet DB.
+
+        Params:
+            enroll_hash = enrollment blockheight to update enroll hash
+            block_height = enrolled blockheight
+
+        Returns:
+            Update execution status
```
Geod24: This doesn't say anything

source/agora/common/EnrollmentManager.d
```diff
@@ -205,6 +205,78 @@ public class EnrollmentManager
...
+    /***************************************************************************
+
+        In validatorSet DB, return the enrolled block height.
+
+        Params:
+            enroll_hash = key for an enrollment block height
+
+        Returns:
+            the enrolled block height.
+            Return size_t if no matching key exists.
```
Geod24: "the enrolled block height, or 0 if no matching key exists"

TrustHenry:
Ready for review

source/agora/common/EnrollmentManager.d
```diff
@@ -443,4 +515,12 @@ unittest
 
     // test for getEnrollment with removed enrollment
     assert(!man.getEnrollment(utxo_hash2, stored_enroll));
+
+    // test for enrollment block height update
+    assert(!man.getEnrolledHeight(utxo_hash));
+    assert(man.updateEnrolledHeight(utxo_hash, 1));
```
Geod24: Use a value different from `1` (it's too common)

source/agora/common/EnrollmentManager.d
```diff
@@ -443,4 +515,12 @@ unittest
...
+    assert(man.getEnrolledHeight(utxo_hash));
```
Geod24: Check the actual value

source/agora/common/EnrollmentManager.d
TrustHenry: I don't know what your mean.
Do you mean not to return in `if`?

source/agora/common/EnrollmentManager.d
TrustHenry: Use void?

source/agora/common/EnrollmentManager.d
Geod24:
```D
if (foo())
{
    return false;
}
else
{
   some_code();
}
```

Is equivalent to:
```D
if (foo())
   return false;
some_code();
```

Except the later is usually more readable because the function flows more naturally and has less indentation.

source/agora/common/EnrollmentManager.d
TrustHenry: applied the suggestion.

source/agora/common/EnrollmentManager.d
TrustHenry: applied the suggestion.

source/agora/common/EnrollmentManager.d
TrustHenry: applied the suggestion. `true if the updated blockheight`

source/agora/common/EnrollmentManager.d
TrustHenry: applied the suggestion.

source/agora/common/EnrollmentManager.d
TrustHenry: I changed it. I like number `9` :)

source/agora/common/EnrollmentManager.d
TrustHenry: applied the suggestion.

source/agora/common/EnrollmentManager.d
Geod24: If you use `void` you'll have to `throw`.
I mean that "Update execution status" is not descriptive. If someone read it in 1 year, they won't have a clue what it is.

Something like "`true` if the update operation was successful, `false` otherwise" is a bit better.

TrustHenry: applied the suggestion.

source/agora/common/EnrollmentManager.d
Geod24: Did you push ?

TrustHenry: Yes, Why not display `Outdated`

TrustHenry: I did it.

TrustHenry:
Reday for review

source/agora/common/EnrollmentManager.d
TrustHenry: applied the suggestion.

TrustHenry:
Ready for review

source/agora/common/EnrollmentManager.d
AndrejMitrovic: What Mathias meant is to check the return value. so `assert(man.getEnrolledHeight(utxo_hash) == 8);`, or whichever the right height is supposed to be.

TrustHenry: I overlooked the obvious. 
I Modified it.

Geod24 (approved):

AndrejMitrovic (approved):

PR #461, opened 2020-01-03
Implement the feature of enrolling as a validator

linked0:
Implmented basic methods for a Node to enroll as a validator.

- Add getUTXOs API endpoint
- Add createEnrollment API endpoint
- Add enrollValidator API endpoint
- Add hasEnrollment API endpoint

#406

### DIFF
--- a/source/agora/common/EnrollmentManager.d
+++ b/source/agora/common/EnrollmentManager.d
@@ -21,6 +21,8 @@ import agora.common.Hash;
 import agora.common.Serializer;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.UTXOSet;
+import agora.consensus.Validation;
+import agora.node.Ledger;
 import agora.utils.Log;
 
 import d2sqlite3.database;
@@ -57,6 +59,12 @@ public class EnrollmentManager
     /// Enrollment data object
     private Enrollment data;
 
+    /// The ledger
+    private Ledger ledger;
+
+    /// Set of unspent transaction outputs
+    private UTXOSet utxo_set;
+
     /***************************************************************************
 
         Constructor
@@ -68,8 +76,11 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public this (string db_path, KeyPair key_pair)
+    public this (string db_path, KeyPair key_pair, Ledger ledger,
+        UTXOSet utxo_set)
     {
+        this.ledger = ledger;
+        this.utxo_set = utxo_set;
         this.db = Database(db_path);
 
         // create the table for validator set if it doesn't exist yet
@@ -124,6 +135,15 @@ public class EnrollmentManager
     public bool addEnrollment (const ref Enrollment enroll) @safe nothrow
     {
         static ubyte[] buffer;
+
+        // check validity of the enrollment data
+        if (auto reason = isInvalidEnrollmentReason(this.ledger.getBlockHeight() + 1,
+            enroll, this.utxo_set.getUTXOFinder()))
+        {
+            this.logMessage("Invalid enrollment data, Reason: " ~ reason,
+                enroll);
+            return false;
+        }
 
         // check if already exists
         try
@@ -437,42 +457,49 @@ public class EnrollmentManager
 unittest
 {
     import agora.common.Amount;
+    import agora.common.Config;
+    import agora.common.TransactionPool;
+    import agora.consensus.data.Block;
     import agora.consensus.data.Transaction;
     import agora.consensus.data.UTXOSet;
+    import agora.consensus.Genesis;
+    import agora.node.BlockStorage;
     import std.format;
+    import std.conv;
 
+    auto gen_key_pair = getGenesisKeyPair();
     KeyPair[] key_pairs = [KeyPair.random, KeyPair.random];
 
-    // create the first transaction
-    Transaction tx1 = Transaction(
-        TxType.Freeze,
-        [Input(Hash.init, 0)],
-        [Output(Amount.MinFreezeAmount, key_pairs[0].address)]
-    );
+    Transaction[] txs;
+    foreach (idx; 0 .. Block.TxsInBlock)
+    {
+        auto input = Input(hashFull(GenesisTransaction), idx.to!uint);
 
-    // create the second transaction
-    Transaction tx2 = Transaction(
-        TxType.Freeze,
-        [Input(Hash.init, 0)],
-        [Output(Amount(100_000 * 10_000_000L), key_pairs[0].address)]
-    );
+        Transaction tx =
+        {
+            TxType.Freeze,
+            [input],
+            [Output(Amount.MinFreezeAmount, key_pairs[0].address)]
+        };
 
-    // create the third transaction
-    Transaction tx3 = Transaction(
-        TxType.Freeze,
-        [Input(Hash.init, 0)],
-        [Output(Amount.MinFreezeAmount, key_pairs[1].address)]
-    );
+        auto signature = gen_key_pair.secret.sign(hashFull(tx)[]);
+        tx.inputs[0].signature = signature;
+        txs ~= tx;
+    }
 
-    // create and UTXO set and an EnrollmentManager object
+    auto storage = new MemBlockStorage();
+    auto block = makeNewBlock(GenesisBlock, txs);
+    assert(storage.saveBlock(block));
+    auto pool = new TransactionPool(":memory:");
+    scope(exit) pool.shutdown();
     auto utxo_set = new UTXOSet(":memory:");
     scope (exit) utxo_set.shutdown();
-    auto man = new EnrollmentManager(":memory:", key_pairs[0]);
-    scope (exit) man.shutdown();
+    auto config = new Config();
+    scope ledger = new Ledger(pool, utxo_set, storage, config.node);
 
-    utxo_set.updateUTXOCache(tx1, 1);
-    utxo_set.updateUTXOCache(tx2, 1);
-    utxo_set.updateUTXOCache(tx3, 1);
+    // create an EnrollmentManager object
+    auto man = new EnrollmentManager(":memory:", key_pairs[0], ledger, utxo_set);
+    scope (exit) man.shutdown();
 
     // find UTXOs to use in making enrollment data
     Hash[] utxo_hashes;
@@ -484,18 +511,18 @@ unittest
     // create and add the first Enrollment object
     auto utxo_hash = utxo_hashes[0];
     Enrollment enroll;
-    man.createEnrollment(utxo_hash, enroll);
-    assert(man.hasEnrollment(utxo_hash) == false);
-    man.addEnrollment(enroll);
+    assert(man.createEnrollment(utxo_hash, enroll));
+    assert(!man.hasEnrollment(utxo_hash));
+    assert(man.addEnrollment(enroll));
     assert(man.getEnrollmentLength() == 1);
-    assert(man.hasEnrollment(utxo_hash) == true);
-    assert(man.addEnrollment(enroll) == false);
+    assert(man.hasEnrollment(utxo_hash));
+    assert(!man.addEnrollment(enroll));
 
     // create and add the second Enrollment object
     auto utxo_hash2 = utxo_hashes[1];
     Enrollment enroll2;
-    man.createEnrollment(utxo_hash2, enroll2);
-    man.addEnrollment(enroll2);
+    assert(man.createEnrollment(utxo_hash2, enroll2));
+    assert(man.addEnrollment(enroll2));
     assert(man.getEnrollmentLength() == 2);
 
     // get a stored Enrollment object
@@ -516,73 +543,4 @@ unittest
     assert(man.getEnrolledHeight(utxo_hash) == 9);
     assert(!man.updateEnrolledHeight(utxo_hash, 9));
     assert(man.getEnrolledHeight(utxo_hash2) == 0);
-}
-
-///
-unittest
-{
-    import agora.common.Amount;
-    import agora.common.EnrollmentManager;
-    import agora.consensus.data.Enrollment;
-    import agora.consensus.data.Transaction;
-    import agora.consensus.data.UTXOSet;
-    import agora.consensus.Validation;
-
-    KeyPair key_pair = KeyPair.random;
-
-    auto utxo_set = new UTXOSet(":memory:");
-    scope (exit) utxo_set.shutdown();
-    UTXOFinder utxoFinder = utxo_set.getUTXOFinder();
-    auto man = new EnrollmentManager(":memory:", key_pair);
-    scope (exit) man.shutdown();
-
-    // normal frozen transaction
-    Transaction tx1 = Transaction(
-        TxType.Freeze,
-        [Input(Hash.init, 0)],
-        [Output(Amount.MinFreezeAmount, key_pair.address)]
-    );
-
-    // payment transaction
-    Transaction tx2 = Transaction(
-        TxType.Payment,
-        [Input(Hash.init, 0)],
-        [Output(Amount.MinFreezeAmount, key_pair.address)]
-    );
-
-    // Insufficient freeze amount transaction
-    Transaction tx3 = Transaction(
-        TxType.Freeze,
-        [Input(Hash.init, 0)],
-        [Output(Amount(1), key_pair.address)]
-    );
-
-    auto utxo_hash1 = utxo_set.getHash(hashFull(tx1), 0);
-    auto utxo_hash2 = utxo_set.getHash(hashFull(tx2), 0);
-    auto utxo_hash3 = utxo_set.getHash(hashFull(tx3), 0);
-
-    Enrollment enroll1;
-    Enrollment enroll2;
-    Enrollment enroll3;
-    assert(man.createEnrollment(utxo_hash1, enroll1));
-    assert(man.createEnrollment(utxo_hash2, enroll2));
-    assert(man.createEnrollment(utxo_hash3, enroll3));
-
-    assert(!isValidEnrollment(1, enroll1, utxoFinder));
-    assert(!isValidEnrollment(1, enroll2, utxoFinder));
-    assert(!isValidEnrollment(1, enroll3, utxoFinder));
-
-    utxo_set.updateUTXOCache(tx1, 0);
-    utxo_set.updateUTXOCache(tx2, 0);
-    utxo_set.updateUTXOCache(tx3, 0);
-
-    auto utxos = utxo_set.getUTXOs(key_pair.address);
-    assert(utxos[utxo_hash1].output.address == key_pair.address);
-    assert(utxos[utxo_hash2].output.address == key_pair.address);
-    assert(utxos[utxo_hash3].output.address == key_pair.address);
-
-    assert(isValidEnrollment(1, enroll1, utxoFinder));
-    assert(!isValidEnrollment(0, enroll1, utxoFinder));
-    assert(!isValidEnrollment(1, enroll2, utxoFinder));
-    assert(!isValidEnrollment(1, enroll3, utxoFinder));
 }
--- a/source/agora/node/API.d
+++ b/source/agora/node/API.d
@@ -37,6 +37,7 @@ module agora.node.API;
 
 import agora.common.crypto.Key;
 import agora.consensus.data.Block;
+import agora.consensus.data.Enrollment;
 import agora.common.Types;
 import agora.common.Set;
 import agora.common.Serializer;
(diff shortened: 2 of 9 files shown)


source/agora/node/Node.d
```diff
@@ -338,9 +347,67 @@ public class Node : API
...
     public Hash[] getMerklePath (ulong block_height, Hash hash) @safe
     {
         return this.ledger.getMerklePath(block_height, hash);
     }
+
+    /// GET: /get_utxos
+    public UTXOSetValue[Hash] getUTXOs(TxType type, ulong min_amount_value) @safe
+    {
+        UTXOSetValue[Hash] result_utxos;
+        auto utxos = this.utxo_set.getUTXOs(config.node.key_pair.address);
```
TrustHenry: This is to get an available UTXO list?, but not just the address of the node.
address must also be added by a parameter.

If it's for this node only, need to change the public.

TrustHenry (changes requested):

source/agora/node/API.d
```diff
@@ -198,4 +200,73 @@ public interface API
...
+        Params:
+            type = transaction type of UTXOs to find
+            min_amount_value = mininum amount of UTXOs to find
+
+        Returns:
+            the associative array for UTXOs
+
+    ***************************************************************************/
+
+    public UTXOSetValue[Hash] getUTXOs(TxType type, ulong min_amount_value);
```
TrustHenry: `public UTXOSetValue[Hash] getUTXOs (TxType type, ulong min_amount_value);`

TrustHenry:
```suggestion
    public UTXOSetValue[Hash] getUTXOs (TxType type, ulong min_amount_value);
```

source/agora/node/Node.d
```diff
@@ -338,9 +347,67 @@ public class Node : API
...
+    }
+
     /// GET: /merkle_path
     public Hash[] getMerklePath (ulong block_height, Hash hash) @safe
     {
         return this.ledger.getMerklePath(block_height, hash);
     }
+
+    /// GET: /get_utxos
+    public UTXOSetValue[Hash] getUTXOs(TxType type, ulong min_amount_value) @safe
```
TrustHenry:
```suggestion
    public UTXOSetValue[Hash] getUTXOs (TxType type, ulong min_amount_value) @safe
```

linked0:
@Geod24 I have a CI build error on Linux as follows
```
Error: clean-up of Database incorrectly depends on destructors called by the GC.
Info: validator_set.dat��
./build/agora-unittests(_D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPSQCk3sys5posix6signal9siginfo_tPvZv+0x38)[0x5650f1ab7d0c]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12890)[0x7fe50ec6e890]
```
As you see the agora/tests/EnrollmentManager.d, All the code is commented out temporarily, but it outputs incomprehensible errors.
Does this remind you of anything?

As another option, I can close this PR and make multiple PRs.

Geod24:
I think that, since you now instantiate a new EnrollmentManager for every node, some of them are not properly finalized (e.g. shutdown isn't being called).
You can probably test this by spinning up a VM and trying it.

linked0:
Ready for review although the ci/circleci tests failed.
It's not obvious to me what failed [circle/ci result page](https://app.circleci.com/jobs/github/bpfkorea/agora/396)

Geod24:
> source/agora/node/API.d(46,8): Error: module `Stellar_SCP` is in file 'scpd/types/Stellar_SCP.d' which cannot be read

You need to add it to [this array](https://github.com/bpfkorea/agora/blob/86ca56f29bcd4d2d580e2037b2e26ae622d42df6/tests/system/dub.json#L8). We should really fix this.

linked0:
The collect2 outputs error on the circle/ci test as follows after import paths being added.

> collect2: error: ld returned 1 exit status

Geod24:
Looking at the link error, you need to add the sqlite dependency in [here](https://github.com/bpfkorea/agora/blob/86ca56f29bcd4d2d580e2037b2e26ae622d42df6/tests/system/dub.json#L38-L44)

linked0:
In order to fix the ci/circleci test errors, I have modified `tests/system/dub.json` file as follows.
1. add import paths to the "importPaths" section.
2. add references to the "sourceFiles" section.
3. add references for submodules to the "dependencies" section.

linked0:
Ready for review.

source/agora/node/API.d
```diff
@@ -198,4 +200,75 @@ public interface API
...
+        Params:
+            frozen_utxo_hash = the hash of a frozen UTXO used to identify
+                        a validator or an enrollment data
+
+        Returns:
+            the Enrollment object created
+
+    ***************************************************************************/
+
+    public Enrollment createEnrollment (Hash frozen_utxo_hash);
```
AndrejMitrovic: And if the utxo doesn't exist, what happens then?

AndrejMitrovic:
Looks good, but I'm not sure about introducing several different APIs in the same commit. Maybe it makes sense if they cannot exist without being grouped together like that. I'll let others share their opinions.

source/agora/node/API.d
linked0: Your question reminds me that I should check the validity of the enrollment data in the receiveEnrollment function of the GossipProtocol class and the validity of the UTXO in the createEnrollment function that you mentioned above.

linked0:
> Looks good, but I'm not sure about introducing several different APIs in the same commit. Maybe it makes sense if they cannot exist without being grouped together like that. I'll let others share their opinions.

The APIs must have been grouped together because I had to test the entire process from start to end of the `enrollValidator` API. There are some difficulties in unit testing and fixing bugs because if there are some errors at the commit few steps before, a developer must rebase all the commits. It must be unnatural and inconvenient for developing.

Anyway, I'll make sure to pay more careful attention to what you said while working

source/agora/node/API.d
linked0: Herny is making the function for checking validity of UTXO called isInvalidValidatorUTXO. So I will use it after his work is done.

linked0:
Ready for review.
For Drey's comment, I will make another issue for validation check for enrollment data.

TrustHenry (approved):
LGTM

source/agora/node/API.d
```diff
@@ -198,4 +200,75 @@ public interface API
...
+        API:
+            GET /has_enrollment
+
+        Returns:
+            true if the validator set has the enrollment data
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.GET)
+    public bool hasEnrollment (Hash enroll_hash);
```
AndrejMitrovic, line 234: What is the use-case for this API? I see that it's used in the unittest, but other than the tests why do we need to query another node if it contains an enrollment hash?

linked0: I follow the way the code about transaction has implemented. It has the hasTransactionHash function so I add the API similar to that. Its usage is just for the unit test now.

AndrejMitrovic: But in that case, `NetworkClient.sendTransaction` calls `hasTransactionHash` before calling `putTransaction`, to avoid sending the same transaction to a node which already contains it. That was its use-case.

Then the same should be done for the enrollment, right? `NetworkManager.sendEnrollment` should first check `hasEnrollment` before calling `enrollValidator` to avoid spamming the network.

linked0: That's right. I have missed it. Thanks.

linked0: > Then the same should be done for the enrollment, right? 
> NetworkManager.sendEnrollment should first check hasEnrollment before calling enrollValidator to avoid spamming the network.

Please ignore what I commented before(=> "That's right. I have missed it.")
It is already checked in receiveEnrollment function of the GossipProtocol class, actually in the addEnrollment function in the receiveEnrollment function. The sendTransaction you mentioned is not checking other node's state also.

AndrejMitrovic: But if it's checked in `receiveEnrollment` it means the structure was already sent over the wire. The point is to not send data that doesn't have to be sent.

linked0: I got it. I will be checking the code more.

linked0: I will add the code calling the hasEnrollment function into the sendEnrollement function of the **NetworkClient** class like Transaction.

source/agora/node/Node.d
```diff
@@ -338,9 +347,70 @@ public class Node : API
...
+    }
+
     /// GET: /merkle_path
     public Hash[] getMerklePath (ulong block_height, Hash hash) @safe
     {
         return this.ledger.getMerklePath(block_height, hash);
     }
+
+    /// GET: /get_utxos
+    public UTXOSetValue[Hash] getUTXOs (PublicKey address, TxType type,
```
AndrejMitrovic: This method is allocating two associative arrays. We could add an optional `min_amount` parameter to  `UTXOSet.getUTXOs` and implement the amount filtering there?

linked0: Sure, we could. The function will be enhanced in another PR.

linked0:
Ready for review
1) Make use the hasEnrollment function in the gossip protocol.
2) Resolve an synchronization issue between nodes.(See agora/tests/EnrollmentManager.d)

linked0:
Rebased.

source/agora/node/API.d
```diff
@@ -198,4 +200,75 @@ public interface API
...
+            address = used to identify the owner of UTXOs
+            type = transaction type of UTXOs to find
+            min_amount_value = mininum amount of UTXOs to find
+
+        Returns:
+            the associative array for UTXOs
+
+    ***************************************************************************/
+
+    public UTXOSetValue[Hash] getUTXOs (PublicKey address, TxType type,
```
Geod24: Why is this in the API ?? Same question for `createEnrollment` ?

source/agora/node/API.d
```diff
@@ -198,4 +200,75 @@ public interface API
...
+        ulong min_amount_value);
```
Geod24: Any amount should be expressed using `Amount`

source/agora/test/EnrollmentManager.d
```diff
@@ -0,0 +1,97 @@
...
+/// test enrollment process
+unittest
+{
+    import std.algorithm;
+    import std.conv;
+    import core.time;
+    import core.thread;
+
+    const NodeCount = 4;
+    const SpendAmount = 40_000 * 10_000_000;
```
Geod24: Use `Amount` in order to avoid to have to do this

source/agora/test/EnrollmentManager.d
```diff
@@ -0,0 +1,97 @@
...
+    node_1.enrollValidator(enroll);
+
+    // check if a node contains enrollment data previously sent.
+    size_t times = 0; // Number of times we slept for 50 msecs
+    Start:
+        Thread.sleep(10.msecs);
+        auto has_enroll = node_2.hasEnrollment(utxo_hashes[0]);
+        // Retry if we're too early
+        if (has_enroll == false && times++ < 50)
+            goto Start;
```
Geod24: You don't need a goto here. The code I showed you as an example needed it because there was something after in the loop, but it's not the case here.
E.g.:
```D
while (!node_2.hasEnrollment(...) && times++ < 50)
    Thread.sleep(10.msecs);
assert(times < 50);
```

This will assert if it reaches the 500 msecs timeout

AndrejMitrovic: Why not just use the `retryFor` function like the rest of the test-suite?

source/agora/node/API.d
linked0: The getUTXOs can be used in the situation when a user(or Node client app) select an UTXO as a deposit for enrollment. So it should be open as API.
But, the createEnrollment could be inserted in the enrollValidator which must take an UTXO as parameter.

source/agora/test/EnrollmentManager.d
Geod24: Reasons

source/agora/node/API.d
linked0: `enrollVaidator` is an API endpoint receiving the enrollment data and cannot make enrollment data, and an user client should select an UTXO in the current scenario. So createEnrollment is needed now as a bridge between the two APIs. After the operations of an user Node client is determined, the createEnrollment API could be deleted.

Geod24: > The getUTXOs can be used in the situation when a user(or Node client app) select an UTXO as a deposit for enrollment. So it shoud be open as API.

Everything in `API` should be used *only* for interfacing with other nodes. There should not be *any* method here for convenience (by convenience, I mean used by non-node, e.g. APIs for exchanges).
Because such methods are easy DoS endpoint. Imagine a few clients calling this method if you have even [1% of the amount of UTXOs Bitcoin has today](https://www.blockchain.com/en/charts/utxo-count).

For convenience methods, we would expose *another* API, which is only enabled on demand. An API expected to be used only locally does not need to worry about this.

Geod24: > `enrollVaidator` is an API endpoint receiving the enrollment data and cannot make enrollment data, and an user client should select an UTXO in the current scenario. So `createEnrollment` is needed now as a bridge between the two APIs.

If you want to force a validator to enroll, just make an `enroll` method in [here](https://github.com/bpfkorea/agora/blob/1010eb23fdc9d702f5eea3e6dc4f30027da3dd73/source/agora/test/Base.d#L401-L415). This will not expose it to the world, but allow you to trigger specific actions on a node from an unittest.

source/agora/test/EnrollmentManager.d
```diff
@@ -0,0 +1,78 @@
...
+import agora.consensus.Genesis;
+import agora.test.Base;
+
+/// test enrollment process
+unittest
+{
+    import std.algorithm;
+    import std.conv;
+    import core.time;
+    import core.thread;
```
Geod24: The `core.thread` import is not necessary. The other ones should be moved to global scope.

source/agora/node/Node.d
```diff
@@ -387,9 +397,81 @@ public class Node : API
...
+
+        Params:
+            address = used to identify the owner of UTXOs
+
+        Returns:
+            the associative array for UTXOs
+
+    ***************************************************************************/
+
+    protected UTXOSetValue[Hash] getUTXOs (PublicKey address) @safe
```
Geod24: Since this function is only forwarding, and it's protected, you can remove it.
Instead call `this.utxo_set.getUTXOs` directly.

But I realize that most members of `Node` are `private`. Perhaps they should be made `protected`. Could you do this in another commit / PR (since it's a no brainer, another PR would get merged very quickly).

source/agora/common/EnrollmentManager.d
```diff
@@ -125,6 +136,16 @@ public class EnrollmentManager
     {
         static ubyte[] buffer;
 
+        // check validity of the enrollment data
+        auto reason = isInvalidEnrollmentReason(this.ledger.getBlockHeight() + 1,
+            enroll, this.utxo_set.getUTXOFinder());
+        if (reason !is null)
```
Geod24: One nice idiom in D is the ability to do:
```D
if (auto reason = func())
```

Which looks like a perfect fit for this.

Geod24 (approved):
Looks very good, just a couple suggestions and we're good to go

source/agora/node/Node.d
linked0: > Could you do this in another commit / PR (since it's a no brainer, another PR would get merged very quickly).

Sure, I will do that in another commit of this PR.

linked0:
Ready for review. 
Is it OK to reqeust a review in spite of failing in all tests on Github/Linux? If any problem, please tell me.

.github/workflows/github.yml
```diff
@@ -25,6 +25,8 @@ jobs:
         brew install sqlite3
     - name: Build & test agora
       run: |
+        ulimit -a
+        ulimit -Ss 16384
```
AndrejMitrovic: These two should be swapped. So we can verify the limits were increased.

source/agora/common/EnrollmentManager.d
```diff
@@ -125,6 +136,15 @@ public class EnrollmentManager
     {
         static ubyte[] buffer;
 
+        // check validity of the enrollment data
+        if (auto reason = isInvalidEnrollmentReason(this.ledger.getBlockHeight() + 1,
+            enroll, this.utxo_set.getUTXOFinder()))
+        {
+            this.logMessage("Invalid enrollment data, Reason: " ~ reason,
```
AndrejMitrovic, line 143: This will allocate a new string every time enrollment fails validation. See this comment here: https://github.com/bpfkorea/agora/pull/465#discussion_r364074502

linked0:
The log of the Github/Linux test error is like as follows.
```
./build/agora-unittests(_D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPSQCk3sys5posix6signal9siginfo_tPvZv+0x38)[0x5586650364e8]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12890)[0x7f30f1789890]
./build/agora-unittests(_D5agora9consensus4data7UTXOSetQi8findUTXOMFNbNfS6geod247bitblob__T7BitBlobVmi512ZQpmJSQDhQDeQCxQCv12UTXOSetValueZb+0xf9)[0x558664aba529]
./build/agora-unittests(_D5agora9consensus10Validation25isInvalidEnrollmentReasonFNbNfxmKxSQCnQCk4data10EnrollmentQmDFNbNfS6geod247bitblob__T7BitBlobVmi512ZQpmJSQFfQFcQCs7UTXOSet12UTXOSetValueZbZAya+0x6f)[0x558664ab076b]
./build/agora-unittests(_D5agora6common17EnrollmentManagerQt13addEnrollmentMFNbNfKxSQCg9consensus4data10EnrollmentQmZb+0x89)[0x558664a19aa1]
./build/agora-unittests(_D5agora4node14GossipProtocolQq17receiveEnrollmentMFNfSQCb9consensus4data10EnrollmentQmZv+0x29)[0x558664afa1b5]
./build/agora-unittests(_D5agora4node4NodeQf15enrollValidatorMFNfSQBo9consensus4data10EnrollmentQmZv+0xd5)[0x558664b02849]
./build/agora-unittests(_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPIZQBl13handleCommandFSQCzQCv7CommandQCiSQDrQDn9FilterAPIZv+0x1b31)[0x558664b079a9]
```

linked0:
Another error log on the Github/Linux tests
```
./build/agora-unittests(_D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPSQCk3sys5posix6signal9siginfo_tPvZv+0x38)[0x56003afa64e8]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x12890)[0x7f266b68a890]
./build/agora-unittests(_D5agora4node4NodeQf13hasEnrollmentMFNfS6geod247bitblob__T7BitBlobVmi512ZQpZb+0x21)[0x56003aa7287d]
./build/agora-unittests(_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPIZQBl13handleCommandFSQCzQCv7CommandQCiSQDrQDn9FilterAPIZv+0x1d4f)[0x56003aa77bc7]
./build/agora-unittests(_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPIZQBl__T7spawnedTCQBpQBmQBk8TestNodeZQBdFSQCn6common6ConfigQhPOSQEp8RegistryQjZ__T6handleTSQFqQFm7CommandZQyMFQuZ9__lambda2MFZv+0x52)[0x56003aa80bc2]
./build/agora-unittests(_D6geod2411concurrency14FiberScheduler6createMFNbDFZvZ4wrapMFZv+0x2b)[0x56003ac85297]
./build/agora-unittests(_D4core6thread5fiber5Fiber3runMFZv+0x2a)[0x56003afa72ae]
./build/agora-unittests(fiber_entryPoint+0x63)[0x56003afa7193]
Program exited with code -11
```

source/agora/common/EnrollmentManager.d
linked0: > This will allocate a new string every time enrollment fails validation. See this comment here: #465 (comment)

The logMessage function should be enhanced for general usage also. In the PR for the enhancement, I will change the code with your comment.

linked0:
Ready for review. 
- I added the code checking validity of an enrollment data
- I resoved the unittest error.

**Unittest Error Description**
After the flow of the unittest code in the agora.test.EnrollmentManager module ended, the following code in the sendEnrollment function of NetworkClient was running in some thread and crashed with the Segfault because the network including all nodes was shutdowned.
```
this.taskman.runTask(
{
    // if the node already has this enrollment, don't send it
    if (this.attemptRequest!(LogLevel.Trace)(
        this.api.hasEnrollment(enroll.utxo_key), null))
            return;
    this.attemptRequest(this.api.enrollValidator(enroll), null);
});
```

**Techniques used to analyze this issue**
- dub test --compiler=ldc2
- lldb ./build/agora-unittests
```
(lldb) run
(lldb) p this
(lldb) p this.enroll_man
```
- `shared` variable for selective log print
- checking the value of pointer with the following snippet
```
void** ffs = cast(void**)&this.enroll_man;
*ffs = cast(void*)0xDEAD_BEEF;
```
- and Henry's PR #375

PR #464, opened 2020-01-07
Add a nothrow wrapper around Ocean's Logger & make PrettyPrinter nothrow

AndrejMitrovic:
Reference: https://github.com/bpfkorea/agora/pull/351#discussion_r363579360

There's a few issues:
- There is a weird compiler bug with instantiating `logger.info` in one specific case where `hashmap.length` is called. I had to add an explicit `info` overload instead of using `opDispatch`.
- To fully resolve https://github.com/bpfkorea/agora/pull/351#discussion_r363579360 I had to make `PrettyPrinter` nothrow.

W.r.t. the second issue, I'm honestly not sure if we want to do this. Do we assume that all the string formatting should /always/ work, and if not then it's an Error?

### DIFF
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -279,9 +279,8 @@ public class NetworkManager
         }
         catch (Exception ex)
         {
-            scope (failure) assert(0);
             log.error("Couldn't retrieve blocks: {}. Will try again later..",
-                ex.message);
+                ex.msg);
         }
     }
 
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -201,7 +201,7 @@ public class AgoraLayout : Appender.Layout
 
     Logging function which is only called from C++ code
 
-    It's for C++ code to use agora's logger instead of C++ stdout 
+    It's for C++ code to use agora's logger instead of C++ stdout
 
     Params:
         logger = the logger name
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -41,19 +41,54 @@ import std.range : Cycle, cycle, take;
 /// Insert a logger in the current scope, named log
 public template AddLogger (string moduleName = __MODULE__)
 {
-    import ocean.util.log.Logger;
+    import Ocean = ocean.util.log.Logger;
+    import agora.common.Types;
+
+    /// nothrow wrapper around Ocean's Logger
+    public static struct Logger
+    {
+        private Ocean.Logger logger;
+
+        // workaround: weird CT bug in NetworkManager.d call:
+        // log.info("Discovery reached. {} peers connected.", this.peers.length);
+        // source/agora/network/NetworkManager.d(149,12): Error: no property info for type Logger
+        public void info (Args...) (Args args) @safe nothrow
+        {
+            try
+            {
+                this.logger.info(args);
+            }
+            catch (Exception ex)
+            {
+                assert(0, ex.msg);
+            }
+        }
+
+        public void opDispatch (string call, Args...) (Args args) @safe nothrow
+        {
+            try
+            {
+                mixin("this.logger." ~ call ~ "(args);");
+            }
+            catch (Exception ex)
+            {
+                assert(0, ex.msg);
+            }
+        }
+    }
+
     private Logger log;
     static this ()
     {
         import core.memory;
-        log = Log.lookup(moduleName);
-        GC.addRoot(cast(void*)log);
+        log = Logger(Ocean.Log.lookup(moduleName));
+        GC.addRoot(cast(void*)log.logger);
     }
 
     static ~this()
     {
         import core.memory;
-        GC.removeRoot(cast(void*)log);
+        GC.removeRoot(cast(void*)log.logger);
     }
 }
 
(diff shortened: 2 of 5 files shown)


source/agora/network/NetworkManager.d
```diff
@@ -279,9 +279,8 @@ public class NetworkManager
         }
         catch (Exception ex)
         {
-            scope (failure) assert(0);
             log.error("Couldn't retrieve blocks: {}. Will try again later..",
-                ex.message);
```
Geod24, line 284: Whoever didn't make it `nothrow` deserves a slap. I think that means either Mihails or Luca.

source/agora/utils/Log.d
```diff
@@ -41,19 +41,54 @@ import std.range : Cycle, cycle, take;
...
+    import agora.common.Types;
+
+    /// nothrow wrapper around Ocean's Logger
+    public static struct Logger
+    {
+        private Ocean.Logger logger;
+
+        // workaround: weird CT bug in NetworkManager.d call:
+        // log.info("Discovery reached. {} peers connected.", this.peers.length);
+        // source/agora/network/NetworkManager.d(149,12): Error: no property info for type Logger
```
Geod24, line 54: Would be great if you could dustmite it 👼

source/agora/utils/PrettyPrinter.d
```diff
@@ -70,22 +70,29 @@ private struct AmountFmt
 {
     private Amount value;
 
-    public void toString (scope void delegate(scope const(char)[]) @safe sink) @safe
+    public void toString (scope void delegate(scope const(char)[]) @safe sink) @safe nothrow
```
Geod24, line 73: Should the `sink` be `nothrow` ?

source/agora/utils/PrettyPrinter.d
```diff
@@ -70,22 +70,29 @@ private struct AmountFmt
...
     {
-        formattedWrite(sink, "%,d", this.value.integral());
-        if (auto dec = this.value.decimal())
+        try
         {
-            sink(".");
-            size_t mask = 1_000_000;
-            while (dec)
+            formattedWrite(sink, "%,d", this.value.integral());
```
Geod24, line 77: This should be `nothrow` if the `sink` is `nothrow`, no ?

Geod24:
@AndrejMitrovic : Line 6 of `.travis.yml`: `  - ldc-1.19.0,dub` should solve your problem.

source/agora/utils/PrettyPrinter.d
AndrejMitrovic: Discussed IRL. `formattedWrite` is not nothrow because it uses Exceptions instead of Errors.

AndrejMitrovic:
Now there are issues with github-ci.

> E: Failed to fetch https://packages.microsoft.com/repos/azure-cli/dists/bionic/main/binary-amd64/Packages.bz2  File has unexpected size (7745 != 7578). Mirror sync in progress? [IP: 40.76.35.62 443]

Someone at needs to add some CI for the CI over there at Microsoft..

source/agora/utils/Log.d
```diff
@@ -41,19 +41,54 @@ import std.range : Cycle, cycle, take;
...
+        public void info (Args...) (Args args) @safe nothrow
+        {
+            try
+            {
+                this.logger.info(args);
+            }
+            catch (Exception ex)
+            {
+                throw new Error(ex.msg);
```
Geod24: Why the `throw` instead of `assert` ?

AndrejMitrovic: Ah yeah, good point.

Geod24 (approved):

Geod24:
BTW I'm fine with either the large diff or `scope (failure) assert(0);` but I suppose you went with the former for better UX ?

AndrejMitrovic:
Yeah there is one thing that I'm missing in D, that maybe we could use a DIP for. And that is:

```D
scope (failure, exception) assert(0, exception.msg)
```

Geod24:
```
/home/circleci/dlang/dmd-2.089.1/linux/bin64/../../src/phobos/std/format.d(3962,39): Error: function `agora.utils.PrettyPrinter.BlockFmt.toString(scope void delegate(scope const(char)[]) nothrow @safe sink)` is not callable using argument types `(void delegate(scope const(char)[] s) @safe)`
/home/circleci/dlang/dmd-2.089.1/linux/bin64/../../src/phobos/std/format.d(3962,39):        cannot pass argument `__lambda4` of type `void delegate(scope const(char)[] s) @safe` to parameter `scope void delegate(scope const(char)[]) nothrow @safe sink`
/home/circleci/dlang/dmd-2.089.1/linux/bin64/../../src/phobos/std/format.d(4342,21): Error: template instance `std.format.formatObject!(LockingTextWriter, BlockFmt, char)` error instantiating
```

Is that a legit error ?

AndrejMitrovic:
Damn this is such a mess.

Screen Shot 2020-01-08 at 14 31 02

AndrejMitrovic:
I need to remove `nothrow` from the sink due to https://issues.dlang.org/show_bug.cgi?id=20487

AndrejMitrovic:
Removed `nothrow` from the sink. It's green now.

AndrejMitrovic:
I have reduced the bug I found manually (dustmite took too long): https://issues.dlang.org/show_bug.cgi?id=20488

Geod24:
> To fully resolve #351 I had to make `PrettyPrinter` `nothrow`.

played

PR #465, opened 2020-01-07
add isInvalidEnrollmentReason Function in Validation

TrustHenry:
This checks the validation of UTXO of the enrollment.

Currently, it is used to generate or add an enrollment, but later it will also be used to verify the enrollments of the blockHeader.

Has dependence on PR #460
Once you get the block height of the enrollmentData, we can validate the cycle_length of the enrollment.

Required #467 
#205

### DIFF
--- a/source/agora/common/EnrollmentManager.d
+++ b/source/agora/common/EnrollmentManager.d
@@ -517,3 +517,72 @@ unittest
     assert(!man.updateEnrolledHeight(utxo_hash, 9));
     assert(man.getEnrolledHeight(utxo_hash2) == 0);
 }
+
+///
+unittest
+{
+    import agora.common.Amount;
+    import agora.common.EnrollmentManager;
+    import agora.consensus.data.Enrollment;
+    import agora.consensus.data.Transaction;
+    import agora.consensus.data.UTXOSet;
+    import agora.consensus.Validation;
+
+    KeyPair key_pair = KeyPair.random;
+
+    auto utxo_set = new UTXOSet(":memory:");
+    scope (exit) utxo_set.shutdown();
+    UTXOFinder utxoFinder = utxo_set.getUTXOFinder();
+    auto man = new EnrollmentManager(":memory:", key_pair);
+    scope (exit) man.shutdown();
+
+    // normal frozen transaction
+    Transaction tx1 = Transaction(
+        TxType.Freeze,
+        [Input(Hash.init, 0)],
+        [Output(Amount.MinFreezeAmount, key_pair.address)]
+    );
+
+    // payment transaction
+    Transaction tx2 = Transaction(
+        TxType.Payment,
+        [Input(Hash.init, 0)],
+        [Output(Amount.MinFreezeAmount, key_pair.address)]
+    );
+
+    // Insufficient freeze amount transaction
+    Transaction tx3 = Transaction(
+        TxType.Freeze,
+        [Input(Hash.init, 0)],
+        [Output(Amount(1), key_pair.address)]
+    );
+
+    auto utxo_hash1 = utxo_set.getHash(hashFull(tx1), 0);
+    auto utxo_hash2 = utxo_set.getHash(hashFull(tx2), 0);
+    auto utxo_hash3 = utxo_set.getHash(hashFull(tx3), 0);
+
+    Enrollment enroll1;
+    Enrollment enroll2;
+    Enrollment enroll3;
+    man.createEnrollment(utxo_hash1, enroll1);
+    man.createEnrollment(utxo_hash2, enroll2);
+    man.createEnrollment(utxo_hash3, enroll3);
+
+    assert(!isValidEnrollment(1, enroll1, utxoFinder));
+    assert(!isValidEnrollment(1, enroll2, utxoFinder));
+    assert(!isValidEnrollment(1, enroll3, utxoFinder));
+
+    utxo_set.updateUTXOCache(tx1, 0);
+    utxo_set.updateUTXOCache(tx2, 0);
+    utxo_set.updateUTXOCache(tx3, 0);
+
+    auto utxos = utxo_set.getUTXOs(key_pair.address);
+    assert(utxos[utxo_hash1].output.address == key_pair.address);
+    assert(utxos[utxo_hash2].output.address == key_pair.address);
+    assert(utxos[utxo_hash3].output.address == key_pair.address);
+
+    assert(isValidEnrollment(1, enroll1, utxoFinder));
+    assert(!isValidEnrollment(0, enroll1, utxoFinder));
+    assert(!isValidEnrollment(1, enroll2, utxoFinder));
+    assert(!isValidEnrollment(1, enroll3, utxoFinder));
+}
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -14,12 +14,17 @@
 module agora.consensus.Validation;
 
 import agora.common.Amount;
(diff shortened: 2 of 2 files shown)


source/agora/consensus/Validation.d
```diff
@@ -990,6 +993,62 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
 
+/*******************************************************************************
+
+    Check the validator of an enrollment.
+
+    A Validator's enrollment is considered valid if:
+        - UTXO is unspent frozen utxo
+        - Signatures are authentic
+        - The frozen amount must be equal to or greater than 40,000 BOA
+        - The block height shall be greater than the unlock_height of the UTXO.
```
AndrejMitrovic: `shall be greater than the unlock_height` => `must be at least unlock_height or greater`

source/agora/consensus/Validation.d
```diff
@@ -990,6 +993,62 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    Params:
+        block_height = the height of the direct ancestor of this block
+        enrollment = The enrollment of the target to be verified
+        findUTXO = delegate to find the referenced unspent UTXOs with
+
+    Returns:
+        `null` if the validator's UTXO is valid, a string explaining the reason it
```
AndrejMitrovic: `a string explaining the reason it is invalid otherwise.` => `otherwise a string explaining the reason it is invalid`

source/agora/consensus/Validation.d
```diff
@@ -990,6 +993,70 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    if (!findUTXO(enrollment.utxo_key, size_t.max, utxo_set_value))
+        return "Unspent frozen UTXO not found for the validator.";
+
+    if (utxo_set_value.type != TxType.Freeze)
+        return "UTXO is not frozen.";
+
+    bool isVerify = false;
+    try
+    {
+        const Point address = Point(utxo_set_value.output.address);
```
AndrejMitrovic: We should make the crypto routines @safe or @trusted. This can be done in another PR. 

Also this can be simplified to:

```D
 if (!() @trusted { return verify(address, enrollment.enroll_sig, enrollment); }())
    return "Enrollment signature verification has an error.";
```

This gets rid of a state variable. And also `isVerify` is not according to our style guide, it would have to be `is_verify` or better yet `is_valid`.

Also, if possible this call should be moved outside the `try/catch` because `verify` is `nothrow`.

AndrejMitrovic (commented):
Few comments.

source/agora/consensus/Validation.d
TrustHenry: applied the suggest.

source/agora/consensus/Validation.d
TrustHenry: applied the suggest.

source/agora/consensus/Validation.d
TrustHenry: applied the suggest.

source/agora/consensus/Validation.d
```diff
@@ -990,6 +993,68 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    {
+        return "Error converting address to point";
+    }
+
+    if (!() @trusted { return verify(address, enrollment.enroll_sig,
+        enrollment); }())
+        return "Enrollment signature verification has an error.";
+
+    if (utxo_set_value.output.value.integral() < Amount.MinFreezeAmount.integral())
+        return "The frozen amount must be equal to or greater than 40,000 BOA.";
```
linked0: How about using the code `Amount.MinFreezeAmount.integral().to!string` instead of `40,000` as follows because the MinFreezeAmount might be changed.
```
return "The frozen amount must be equal to or greater than " ~
        Amount.MinFreezeAmount.integral().to!string ~ " BOA.";
```

AndrejMitrovic: @Geod24 do you know if the compiler will try to CTFE the above? Or will this allocate a new string every time? If it allocates, we could work around it by assigning the string to a static variable though.

Geod24: We *could* do that, but not exactly this way. 
As @AndrejMitrovic hinted, this will most likely allocate.
To force CTFE, do:
```D
if (....)
{
    static immutable Message = "The frozen amount" ...;
    return Message;
}
```

It is paramount that error path do not allocate memory (or that this memory be freed). The reason is simple: If sending an invalid transaction to a node leads to memory allocation, one could easily sends many of those to increase the workload on the validator (as it would stress the garbage collector), which is a form of denial of service attack.

AndrejMitrovic: CTFE: https://tour.dlang.org/tour/kr/gems/compile-time-function-evaluation-ctfe

linked0 (approved):
LGTM.

source/agora/consensus/Validation.d
AndrejMitrovic: If you rebase your PR, you can remove the `@trusted` workaround, because https://github.com/bpfkorea/agora/pull/466 was merged.

AndrejMitrovic:
It looks good, except the lack of tests.

source/agora/consensus/Validation.d
TrustHenry: Applied PR #466.

TrustHenry:
Added unittest code.
Ready for review.

source/agora/common/EnrollmentManager.d
```diff
@@ -444,3 +444,72 @@ unittest
...
+    utxo_set.updateUTXOCache(tx1, 0);
+    utxo_set.updateUTXOCache(tx2, 0);
+    utxo_set.updateUTXOCache(tx3, 0);
+
+    auto utxos = utxo_set.getUTXOs(key_pair.address);
+    assert(utxos[utxo_hash1].output.address == key_pair.address);
+    assert(utxos[utxo_hash2].output.address == key_pair.address);
+    assert(utxos[utxo_hash3].output.address == key_pair.address);
+
+    assert(isValidEnrollment(1, enroll1, utxoFinder));
```
Geod24, line 584: Looks like this is failing ?

TrustHenry: Required #467

TrustHenry: I rebased it.

TrustHenry:
Ready for review

source/agora/consensus/Validation.d
```diff
@@ -990,6 +995,71 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
     return null;
 }
 
+/*******************************************************************************
+
+    Check the validator of an enrollment.
```
AndrejMitrovic:
```suggestion
    Check the validity of an enrollment.
```

AndrejMitrovic:
Just found one small issue, otherwise looks good.

source/agora/consensus/Validation.d
TrustHenry: applied the suggestion.

Geod24:
Looks good here too

AndrejMitrovic:
So it looks like the unittests were moved into the `EnrollmentManager.d` module because it contains a helper function for creating enrollments. But what it is actually testing is the `isInvalidEnrollmentReason` function, which is in another module.

I don't want to block this, so I'll approve it. But we'll have to move the tests into the right location later.

AndrejMitrovic (approved):

PR #475, opened 2020-01-14
Revert "Move isValidEnrollment unittest into validation"

TrustHenry:
Reverts bpfkorea/agora#472


AndrejMitrovic:
We're reverting the change because of cyclic imports.

It looks like this code tests both the `EnrollmentManager.createEnrollment` **and** the validation with `isValidEnrollment`. 

However, Ideally we would have two unittests. One in `EnrollmentManager.d` that just tests that `createEnrollment` always creates a valid enrollment. And then **another** unittest inside of `Validation.d` which tests various valid **and** invalid Enrollment structures.

Geod24:
Can we just split it then ?

AndrejMitrovic:
Hmm actually the test is missing checks. `createEnrollment` returns a boolean but it was not tested in these unittests..

TrustHenry:
The return bool test of createEnrollment can be added in other PRs.

TrustHenry:
`CreationEnrollment` and `isValidEnrollment` tests cannot be divided. CreateEnrollment is required for testing `isValidEnrollment`.

AndrejMitrovic:
> The return bool test of createEnrollment can be added in other PRs.

Yes. As this is blocking other people's work.

> CreateEnrollment is required for testing isValidEnrollment.

`isValidEnrollment` should be tested with both valid and invalid Enrollment structures. Those can be created manually.

Issue #468, opened 2020-01-10
Move validation test to its proper location


AndrejMitrovic:
See https://github.com/bpfkorea/agora/pull/465#issuecomment-572917336

TrustHenry:
#476 improvement work.

TrustHenry:
Add isValidEnrollment unittest into Validation #476

PR #476, opened 2020-01-14
Add isValidEnrollment unittest into Validation

TrustHenry:
#475 improvement work.
#468 Move validation test to its proper location 

This removes the existing cyclic imports. 
manually switch the enrollment unittest

### DIFF
--- a/source/agora/common/EnrollmentManager.d
+++ b/source/agora/common/EnrollmentManager.d
@@ -564,9 +564,9 @@ unittest
     Enrollment enroll1;
     Enrollment enroll2;
     Enrollment enroll3;
-    man.createEnrollment(utxo_hash1, enroll1);
-    man.createEnrollment(utxo_hash2, enroll2);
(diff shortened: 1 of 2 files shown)


source/agora/common/EnrollmentManager.d
```diff
@@ -517,72 +517,3 @@ unittest
     assert(!man.updateEnrolledHeight(utxo_hash, 9));
     assert(man.getEnrolledHeight(utxo_hash2) == 0);
 }
-
```
AndrejMitrovic: I think we should keep this unittest too.

This one tests that `createEnrollment` makes a valid enrollment. And the unittest you added tests the validation function itself. Does that sound good to you?

TrustHenry: This retained the existing test and added a createEnrollment unittest.

AndrejMitrovic (approved):

linked0 (approved):
LGTM

PR #483, opened 2020-01-20
[WIP] Add a web interface for easy setup

Geod24:
```
Before we open TestNet, we need an easy to use, mobile-friendly web interface
so that first-timer can easily configure their node.
Additionally, we also want to provide a web interface for monitoring and administration.
```

Note: This uncovered a few issues with the configuration and the way we do things.
So I'm going to do a few refactor in the near future.
Additionally, there are still some issues with the interface itself:
- It is not complete (the almost full design can be seen [here](https://projects.invisionapp.com/share/PVV2WXK4TX5#/screens/395978022).
- The full design was missing a couple things we'll need to add (e.g. initial sync information);
- We are missing assets (@bpalaggi  will contact the agency about this);
- There is no feedback on user error: if the configuration is incorrect, the `alert` is quite useless;
- There's a lot of hacky code to work around small differences;

Things we should do:
- Clear up the configuration parsing so that optional things are really optional;
- Directly write YAML on the JS side: this will make things so much simpler: we can just reuse our YAML parser and provide proper error handling, and we don't even have to load a file from disk;
- Clean up the diet templates / CSS: there have been many iterations, and the code improved with my understanding of HTML/CSS/JS;
- Add bootstrap here so the installation is self-sufficient (?);


Geod24:
One of the issue that arise from this PR is how to store assets.
We want the admin interface to not have external dependency, so you can run it on a network which is not connected to the internet. However, importing `bootstrap` in this repository, or the images that make up the assets is a no-go.

I propose to build a script (preBuildCommands?) that downloads the assets if required, and is run on release.

Geod24:
Superseded by #1984

PR #488, opened 2020-01-20
CI test 2

AndrejMitrovic:


AndrejMitrovic:
Yeah this doesn't work at all. I'll just let it be copy-pasted, it's no fun trying to debug declarative languages which you can't even debug locally.

Issue #493, opened 2020-01-21
Enrollment data should be written to the block.


TrustHenry:
Writing Enrollment data to a block determines the height of the block that can participate in the validation as a validator.
Validators needed for the consensus process can be determined and the number of validators between nodes can be set.

Definition of Done:

- [x] The Enrollment data should be recorded only in the enroll start block.

- [x] When writing to a block, it must be sorted in ascending order by UTXOkey of the enrollment data.

- [x] The node store the enrolled block height.

How to implement

- create a Query to import the elements from the Enrollment manager.

- The conditions for this must be taken in order of the UTXOkey's ascending order with the null height of the block.

- Record the target enrollments at block generation.

- When the enrollment data is stored in a block, the height of that block must be renewed in the DB of the node.

Geod24:
> ascending order by UTXOkey

Do you mean ascending order of the hash, or something else (e.g. value) ?

TrustHenry:
> > ascending order by UTXOkey
> 
> Do you mean ascending order of the hash, or something else (e.g. value) ?

That's right.
It is  Enrollment.utxo_key or It's a utxo hash and a hash of index.

Geod24:
Done in https://github.com/bpfkorea/agora/pull/504 and some related PRs.

PR #494, opened 2020-01-22
Add getUnregisteredEnrollments function in EnrollmentManager

TrustHenry:
Get the unregistered enrollments in the block
And this is arranged in ascending order with the utxo_key.

For #493

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -75,7 +75,7 @@ public class EnrollmentManager
 
         // create the table for validator set if it doesn't exist yet
         this.db.execute("CREATE TABLE IF NOT EXISTS validator_set " ~
-            "(key BLOB PRIMARY KEY, val BLOB NOT NULL, enrolled_height INTEGER)");
+            "(key TEXT PRIMARY KEY, val BLOB NOT NULL, enrolled_height INTEGER)");
 
         // create the table for enrollment data for a node itself
         this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
@@ -177,7 +177,7 @@ public class EnrollmentManager
         {
             () @trusted {
                 this.db.execute("INSERT INTO validator_set (key, val) VALUES (?, ?)",
-                    enroll.utxo_key[], buffer);
+                    enroll.utxo_key.toString(), buffer);
             }();
 
         }
@@ -215,7 +215,8 @@ public class EnrollmentManager
 
     public void removeEnrollment (const ref Hash enroll_hash) @trusted
     {
-        this.db.execute("DELETE FROM validator_set WHERE key = ?", enroll_hash[]);
+        this.db.execute("DELETE FROM validator_set WHERE key = ?",
+            enroll_hash.toString());
     }
 
     /***************************************************************************
@@ -235,7 +236,7 @@ public class EnrollmentManager
         try
         {
             auto results = this.db.execute("SELECT enrolled_height FROM validator_set" ~
-                " WHERE key = ?", enroll_hash[]);
+                " WHERE key = ?", enroll_hash.toString());
             if (results.empty)
                 return size_t.init;
 
@@ -272,7 +273,7 @@ public class EnrollmentManager
             () @trusted {
                 this.db.execute(
                     "UPDATE validator_set SET enrolled_height = ? WHERE key = ?",
-                    block_height, enroll_hash[]);
+                    block_height, enroll_hash.toString());
             }();
         }
         catch (Exception ex)
@@ -382,7 +383,7 @@ public class EnrollmentManager
     public bool hasEnrollment (const ref Hash enroll_hash) @trusted
     {
         auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM validator_set " ~
-            "WHERE key = ?)", enroll_hash[]);
+            "WHERE key = ?)", enroll_hash.toString());
 
         return results.front().peek!bool(0);
     }
@@ -405,7 +406,7 @@ public class EnrollmentManager
         out Enrollment enroll) @trusted
     {
         auto results = this.db.execute("SELECT key, val FROM validator_set " ~
-            "WHERE key = ?", enroll_hash[]);
+            "WHERE key = ?", enroll_hash.toString());
 
         foreach (row; results)
         {
@@ -414,6 +415,33 @@ public class EnrollmentManager
         }
 
         return false;
+    }
+
+    /***************************************************************************
+
+        Get the unregistered enrollments in the block
+        And this is arranged in ascending order with the utxo_key
+
+        Params:
+            enrolls = will contain the unregistered enrollments data if found
+
+        Returns:
+            The unregistered enrollments data
+
+    ***************************************************************************/
+
+    public Enrollment[] getUnregisteredEnrollments (ref Enrollment[] enrolls)
+        @trusted
+    {
+        enrolls.length = 0;
+        assumeSafeAppend(enrolls);
+        auto results = this.db.execute("SELECT val FROM validator_set" ~
+            " WHERE enrolled_height is null ORDER BY key ASC");
+
+        foreach (row; results)
+            enrolls ~= deserializeFull!Enrollment(row.peek!(ubyte[])(0));
+
+        return enrolls;
     }
 
     /***************************************************************************
@@ -511,6 +539,17 @@ unittest
     assert(man.addEnrollment(0, findUTXO, enroll2));
     assert(man.getEnrollmentLength() == 2);
 
+    auto utxo_hash3 = utxo_hashes[2];
+    Enrollment enroll3;
+    assert(man.createEnrollment(utxo_hash3, enroll3));
+    assert(man.addEnrollment(0, findUTXO, enroll3));
+    assert(man.getEnrollmentLength() == 3);
+
+    Enrollment[] enrolls;
+    man.getUnregisteredEnrollments(enrolls);
+    assert(enrolls.length == 3);
+    assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
+
     // get a stored Enrollment object
     Enrollment stored_enroll;
     assert(man.getEnrollment(utxo_hash2, stored_enroll));
@@ -518,7 +557,7 @@ unittest
 
     // remove an Enrollment object
     man.removeEnrollment(utxo_hash2);
-    assert(man.getEnrollmentLength() == 1);
+    assert(man.getEnrollmentLength() == 2);
 
     // test for getEnrollment with removed enrollment
     assert(!man.getEnrollment(utxo_hash2, stored_enroll));
@@ -529,4 +568,23 @@ unittest
     assert(man.getEnrolledHeight(utxo_hash) == 9);
     assert(!man.updateEnrolledHeight(utxo_hash, 9));
     assert(man.getEnrolledHeight(utxo_hash2) == 0);
+    man.getUnregisteredEnrollments(enrolls);
+    assert(enrolls.length == 1);
+
+    man.removeEnrollment(utxo_hash);
+    man.removeEnrollment(utxo_hash2);
+    man.removeEnrollment(utxo_hash3);
+    assert(man.getUnregisteredEnrollments(enrolls).length == 0);
+
+    Enrollment[] ordered_enrollments;
+    ordered_enrollments ~= enroll;
+    ordered_enrollments ~= enroll2;
+    ordered_enrollments ~= enroll3;
+    // Reverse ordering
+    ordered_enrollments.sort!("a.utxo_key > b.utxo_key");
+    foreach (ordered_enroll; ordered_enrollments)
+        assert(man.addEnrollment(0, findUTXO, ordered_enroll));
+    man.getUnregisteredEnrollments(enrolls);
+    assert(enrolls.length == 3);
+    assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
 }


source/agora/consensus/EnrollmentManager.d
```diff
@@ -423,6 +423,35 @@ public class EnrollmentManager
...
+
+        Params:
+            enrolls = will contain the unregistered enrollments data if found
+
+        Returns:
+            Return true if the unregistered enrollment data was found
+
+    ***************************************************************************/
+
+    public bool getEnrollments (out Enrollment[] enrolls) @trusted
```
Geod24: When you use `out`, memory is not reused. Because `out` will initialize the array to its default value, which is `null`.
In order to reuse memory, you should use `ref`.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -423,6 +423,35 @@ public class EnrollmentManager
...
+    {
+        enrolls.length = 0;
```
Geod24, line 436: If your goal here is to reuse memory, you should use `assumeSafeAppend`

source/agora/consensus/EnrollmentManager.d
```diff
@@ -423,6 +423,35 @@ public class EnrollmentManager
         return false;
     }
 
+    /***************************************************************************
+
+        Get the unregistered enrollments in the block
```
Geod24, line 422: The name should reflect that it's the unregistered enrollments, not just enrollments.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -423,6 +423,35 @@ public class EnrollmentManager
...
+        enrolls.length = 0;
+        auto results = this.db.execute("SELECT val FROM validator_set" ~
+            " WHERE enrolled_height is null");
+
+        foreach (row; results)
+            enrolls ~= deserializeFull!Enrollment(row.peek!(ubyte[])(0));
+
+        if (enrolls.length > 0)
+        {
+            enrolls.sort!("a.utxo_key < b.utxo_key");
```
Geod24: I think we could move this to the query, couldn't we ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -525,6 +554,11 @@ unittest
     assert(man.addEnrollment(enroll2));
     assert(man.getEnrollmentLength() == 2);
 
+    Enrollment[] enrolls;
+    man.getEnrollments(enrolls);
+    assert(enrolls.length == 2);
+    assert(enrolls[0].utxo_key < enrolls[1].utxo_key);
```
Geod24: I think a full test would include:
- 3 enrollments, added in (utxo key) order;
- 3 enrollments, added in reverse order;

At the moment the sortedness is not really tested. It should also be tested on the block validation level, but maybe you want to keep this for another PR ?

source/agora/consensus/EnrollmentManager.d
TrustHenry: I thought I could solve it with Query.
In DB, the key is saved as Byte[]
Then it should encode again and order it into Query.
Would it be better to do this?

Geod24: So the issue is that SQLite is not able to ORDER BY on binary data ?

TrustHenry: If it's the same sort order of binary data, it's easy to fix with order by.

source/agora/consensus/EnrollmentManager.d
TrustHenry: Yes, I'll add a test at the block level in another PR to be added.

source/agora/consensus/EnrollmentManager.d
TrustHenry: I changed the function name `getUnregisteredEnrollments`

source/agora/consensus/EnrollmentManager.d
TrustHenry: I add `assumeSafeAppend(enrolls)`

source/agora/consensus/EnrollmentManager.d
TrustHenry: applied the suggest

source/agora/consensus/EnrollmentManager.d
TrustHenry: Binary data sorted by ORDER BY key ASC by SQLite

```
[14, 164, 193, 71, 177, 203, 141, 122, 96, 154, 232, 88, 253, 37, 3, 70, 159, 11, 43, 84, 191, 254, 138, 190, 72, 254, 70, 251, 59, 120, 47, 51, 45, 21, 81, 18, 208, 26, 28, 151, 112, 225, 205, 2, 10, 168, 89, 99, 223, 16, 73, 44, 166, 34, 182, 144, 194, 52, 250, 97, 16, 95, 104, 163]
[123, 195, 91, 196, 99, 236, 96, 3, 9, 136, 102, 72, 138, 254, 100, 252, 104, 23, 220, 171, 226, 215, 168, 54, 50, 240, 204, 203, 161, 107, 63, 198, 195, 254, 195, 185, 226, 63, 38, 197, 150, 40, 255, 117, 25, 242, 86, 84, 62, 31, 35, 78, 3, 232, 114, 79, 30, 182, 134, 111, 125, 171, 231, 81]
[232, 104, 205, 80, 68, 5, 170, 214, 76, 66, 173, 149, 48, 25, 96, 205, 59, 51, 251, 106, 84, 10, 53, 69, 128, 54, 153, 28, 78, 58, 218, 53, 19, 202, 39, 184, 66, 75, 77, 115, 153, 133, 128, 174, 41, 199, 175, 192, 161, 104, 94, 240, 13, 66, 5, 37, 21, 131, 133, 56, 137, 193, 218, 237]
```
deserialize from Binary data 

```
0xa3685f1061fa34c290b622a62c4910df6359a80a02cde170971c1ad01251152d332f783bfb46fe48be8afebf542b0b9f460325fd58e89a607a8dcbb147c1a40e
0x51e7ab7d6f86b61e4f72e8034e231f3e5456f21975ff2896c5263fe2b9c3fec3c63f6ba1cbccf03236a8d7e2abdc1768fc64fe8a486688090360ec63c45bc37b
0xeddac189388583152505420df05e68a1c0afc729ae808599734d4b42b827ca1335da3a4e1c99368045350a546afb333bcd60193095ad424cd6aa054450cd68e8
```

Not sorted in order of desired UTXOkey hash.

TrustHenry: Oh, I see. The binary data for SQLlite was little endian.

TrustHenry: I made a new function to reverse bytes before putting it in to DB.
It is also used in SELECT.
@Geod24 Please check to see if it's okay.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -451,6 +483,26 @@ public class EnrollmentManager
...
+
+        Return:
+            Reverse bytes and return to the new ubyte[]
+
+    ***************************************************************************/
+
+    private static ubyte[] reverseUbytes (ref const Hash data) @trusted nothrow
+    {
+        ubyte[] bytes;
+        foreach (val; data[].dup.reverse)
```
Geod24: Why the `dup` here ? It allocates data for no reason. If `reverse` is in place, then you can just `return data[].dup.reverse`, no need for the loop. If it does not allocate, then why the `dup` ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -389,7 +390,7 @@ public class EnrollmentManager
     public bool hasEnrollment (const ref Hash enroll_hash) @trusted
     {
         auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM validator_set " ~
-            "WHERE key = ?)", enroll_hash[]);
+            "WHERE key = ?)", reverseUbytes(enroll_hash));
```
Geod24: Ultimately, the library should accept either ranges, or recognize `toString`. Could you check what code is taken here ?

Geod24 (changes requested):

source/agora/consensus/EnrollmentManager.d
TrustHenry: Oh, that's right. applied the suggestion.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -451,6 +483,23 @@ public class EnrollmentManager
...
+            data = hash of target to reverse
+
+        Return:
+            Reverse bytes and return to the new ubyte[]
+
+    ***************************************************************************/
+
+    private static ubyte[] reverseUbytes (ref const Hash data) @trusted nothrow
+    {
+        return data[].dup.reverse;
```
AndrejMitrovic: What is this function for? The commit message doesn't explain it. Why is the byte array reversed?

Also, this should use a buffer instead. Like a `static ubyte[]`, set its length to `data[]` and then copy the data over with `buffer[] = data[]`.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -423,6 +424,37 @@ public class EnrollmentManager
...
+
+        enrolls.length = 0;
+        assumeSafeAppend(enrolls);
+        auto results = this.db.execute("SELECT val FROM validator_set" ~
+            " WHERE enrolled_height is null ORDER BY hex(key) ASC");
+
+        foreach (row; results)
+            enrolls ~= deserializeFull!Enrollment(row.peek!(ubyte[])(0));
+
+        if (enrolls.length > 0)
```
AndrejMitrovic: Just `return enrolls.length > 0` will do. :)

AndrejMitrovic: Actually..

@Geod24 I feel like a function should not have a success/failure return code if you can already determine at the call site if the `enrolls` array is empty or not. What do you think?

Geod24: Yeah it should probably just `return ref`.

source/agora/consensus/EnrollmentManager.d
Geod24: Ping @TrustHenry

TrustHenry: This is a library for use only here.
If we publicize this, better add a function in Hash.
Am I right?

source/agora/consensus/EnrollmentManager.d
TrustHenry: > What is this function for? The commit message doesn't explain it. Why is the byte array reversed?

Added commit message description.

> Also, this should use a buffer instead. Like a static ubyte[], set its length to data[] and then copy the data over with buffer[] = data[].

I added a buffer.

source/agora/consensus/EnrollmentManager.d
TrustHenry: I changed the return type.

source/agora/consensus/EnrollmentManager.d
Geod24: By "the library" I mean d2sqlite.

TrustHenry: Oh, I will check it.

TrustHenry:
I changed the key column to TEXT
I changed it completely by using enroll.utxo_key.toString()
Ready for review.

AndrejMitrovic (approved):

TrustHenry:
@Geod24  ping.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -451,6 +479,7 @@ public class EnrollmentManager
         catch (Exception ex)
         {}
     }
+
```
Geod24: Unrelated

source/agora/consensus/EnrollmentManager.d
```diff
@@ -525,14 +554,26 @@ unittest
...
+    auto utxo_hash3 = utxo_hashes[2];
+    Enrollment enroll3;
+    assert(man.createEnrollment(utxo_hash3, enroll3));
+    assert(man.addEnrollment(enroll3));
+    assert(man.getEnrollmentLength() == 3);
+
+    Enrollment[] enrolls;
+    man.getUnregisteredEnrollments(enrolls);
+    assert(enrolls.length == 3);
+    assert(enrolls[0].utxo_key < enrolls[1].utxo_key);
```
Geod24: You can just use https://dlang.org/phobos/std_algorithm_sorting.html#isStrictlyMonotonic

source/agora/consensus/EnrollmentManager.d
```diff
@@ -543,4 +584,22 @@ unittest
...
+    Enrollment[] ordered_enrollments;
+    ordered_enrollments ~= enroll;
+    ordered_enrollments ~= enroll2;
+    ordered_enrollments ~= enroll3;
+    ordered_enrollments.sort!("a.utxo_key < b.utxo_key");
+    foreach (ordered_enroll; ordered_enrollments)
+        assert(man.addEnrollment(ordered_enroll));
+    man.getUnregisteredEnrollments(enrolls);
+    foreach (idx, get_enroll; enrolls)
+        assert(ordered_enrollments[idx] == get_enroll);
```
Geod24: Likewise, just call https://dlang.org/phobos/std_algorithm_sorting.html#isStrictlyMonotonic

Geod24 (commented):
Made a couple comments, otherwise LGTM

Geod24 (approved):

Geod24:
Addressed the review directly

Geod24:
And rebased on master since Drey's recent PR conflicted with this.

source/agora/consensus/EnrollmentManager.d
TrustHenry: I knew this.
It was used in #480PR.
I will improve and write a new PR.

source/agora/consensus/EnrollmentManager.d
TrustHenry: ^

source/agora/consensus/EnrollmentManager.d
TrustHenry: I will improve and write a new PR.

source/agora/consensus/EnrollmentManager.d
TrustHenry: Oh, Thanks.

PR #496, opened 2020-01-28
Remove extra dependencies from EnrollmentManager

AndrejMitrovic:
This simplifies the class a bit, since it only uses one method in each class, so this has now become just a parameter in the `addEnrollment` method.

Further work could be done on other classes, such as GossipProtocol. We should take a look at it soon.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -22,7 +22,6 @@ import agora.common.Serializer;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.UTXOSet;
 import agora.consensus.Validation;
-import agora.node.Ledger;
 import agora.utils.Log;
 
 import d2sqlite3.database;
@@ -59,12 +58,6 @@ public class EnrollmentManager
     /// Enrollment data object
     private Enrollment data;
 
-    /// The ledger
-    private Ledger ledger;
-
-    /// Set of unspent transaction outputs
-    private UTXOSet utxo_set;
-
     /***************************************************************************
 
         Constructor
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
The disabling of parallel GC seems to fix the segfault, so I'll make a separate PR for this to get it merged.

Edit: done.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -457,72 +450,65 @@ public class EnrollmentManager
...
-    import agora.consensus.data.Block;
     import agora.consensus.data.Transaction;
-    import agora.consensus.data.UTXOSet;
     import agora.consensus.Genesis;
-    import agora.node.BlockStorage;
     import std.format;
     import std.conv;
 
+    Transaction[Hash] storage;
+    scope findUTXO = (Hash hash, size_t index, out UTXOSetValue value) @trusted
```
Geod24, line 459: Maybe we could factor this out into a small utility, it seems to be a common need ?

source/agora/node/GossipProtocol.d
```diff
@@ -104,12 +105,14 @@ public class GossipProtocol
 
         Params:
             enroll = the received data for the enrollment
+            finder = the UTXO finder delegate
 
     ***************************************************************************/
 
-    public void receiveEnrollment (Enrollment enroll) @safe
+    public void receiveEnrollment (Enrollment enroll, UTXOFinder finder) @safe
```
Geod24: `scope`

source/agora/consensus/EnrollmentManager.d
```diff
@@ -125,20 +115,23 @@ public class EnrollmentManager
...
+            finder = the delegate to find UTXOs with
             enroll = the enrollment data to add
 
         Returns:
             true if the enrollment data has been added to the validator set
 
     ***************************************************************************/
 
-    public bool addEnrollment (const ref Enrollment enroll) @safe nothrow
+    public bool addEnrollment (ulong block_height, UTXOFinder finder,
```
Geod24: `scope`

Geod24 (approved):
Few comments. Looks nice.

source/agora/consensus/EnrollmentManager.d
AndrejMitrovic: True. I'll take a look soon if we can combine all of them.

AndrejMitrovic:
Added `scope`. And filed #501

PR #498, opened 2020-01-29
Temporarily disable parallel GC

AndrejMitrovic:
There is a semi-frequent segfault bug in Agora that seems
to point to a bug in Druntime. A similar / identical issue
happened in Dub in:

https://github.com/dlang/dub/issues/1812

### DIFF
--- a/source/agora/node/main.d
+++ b/source/agora/node/main.d
@@ -34,6 +34,13 @@ import std.file;
 import std.getopt;
 import std.stdio;
 
+/**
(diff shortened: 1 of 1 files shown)


Geod24 (approved):

AndrejMitrovic:
Docker cloud is failing for me again. I have tried to re-link my account but it didn't fix it. It seems these failures are random.

And I still haven't heard back from the ticket I filed with them..

PR #502, opened 2020-01-29
Config: Implement better defaults

Geod24:
~Better reviewed commit-by-commit.
The large refactoring in main will be quite useful for the admin interface, where it's just a matter of adding an `if` at the top level.~

Stripped down to 1 commit...

### DIFF
--- a/source/agora/common/BanManager.d
+++ b/source/agora/common/BanManager.d
@@ -35,10 +35,10 @@ public class BanManager
     public struct Config
     {
         /// max failed requests until an address is banned
-        public size_t max_failed_requests;
+        public size_t max_failed_requests = 10;
 
-        /// the default duration of a ban
-        public time_t ban_duration;
+        /// How long does a ban lasts, in seconds (default: 1 day)
+        public time_t ban_duration = 60 * 60 * 24;
     }
 
     ///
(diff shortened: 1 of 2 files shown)


source/agora/node/main.d
```diff
@@ -75,31 +75,52 @@ private int main (string[] args)
             writefln("Config file '%s' succesfully parsed.", cmdln.config_path);
             return 0;
         }
+        runTask(() => runNode(config));
+    }
+    catch (ConfigException ex)
```
AndrejMitrovic:
```suggestion
    catch (Exception ex)
```

source/agora/node/main.d
```diff
@@ -75,31 +75,52 @@ private int main (string[] args)
             writefln("Config file '%s' succesfully parsed.", cmdln.config_path);
             return 0;
         }
+        runTask(() => runNode(config));
```
AndrejMitrovic: Clever girl.

AndrejMitrovic (approved):

Geod24:
Fixed

AndrejMitrovic:
Is that a legit error in CircleCI?

AndrejMitrovic:
> node-2_1_98233d7e6b9d | 2020-01-29T14:58:04.121257601Z Task terminated with uncaught exception: Configured quorum set is not considered valid by SCP

Hmm.. sounds like maybe the config wasn't parsed properly.

Geod24:
Stripped down to 1 single commit, since this refactor discovered a few bugs.
Going to address everything independently and gradually submit the other commits.
Since the commit has already been approved, will self-merge once green.

Issue #380, opened 2019-10-10
Make Set use an RBTree, or a sorted data structure


AndrejMitrovic:
In the current implementation the Set structure is a source of randomness. This is because internally we use a hashmap, and walking over this hashmap is semi-random. IMO, we need to have reproducible code-flow, and randomness should be limited to where it's actually desired.

Edit: might be blocked by: https://issues.dlang.org/show_bug.cgi?id=21270

AndrejMitrovic:
Hmm it's a bit complicated to use the phobos RBTree, because it's a class. I really hate the lack of default ctors for structs, you hear me Walter Bright?

AndrejMitrovic:
Update: Walking over the map is not random. However inserting into the map changes the order of things. For example:

```
import std.algorithm;
import std.range;
import std.stdio;

void main ()
{
    int[int] map_1;

    foreach (item; iota(0, 20))
    {
        map_1[item] = item;
        writefln("Map: %s", map_1.values);
    }
}
```

Produces:

```
Map: [0]
Map: [0, 1]
Map: [0, 2, 1]
Map: [0, 3, 2, 1]
Map: [0, 4, 3, 2, 1]
Map: [0, 5, 4, 3, 2, 1]
Map: [0, 6, 2, 3, 1, 5, 4]
Map: [0, 6, 7, 2, 3, 1, 5, 4]
Map: [0, 6, 7, 2, 3, 1, 8, 5, 4]
Map: [0, 6, 7, 2, 3, 1, 8, 5, 4, 9]
Map: [0, 6, 7, 2, 3, 10, 1, 8, 5, 4, 9]
Map: [0, 6, 7, 2, 11, 3, 10, 1, 8, 5, 4, 9]
Map: [0, 6, 7, 2, 11, 12, 3, 10, 1, 8, 5, 4, 9]
Map: [0, 13, 6, 7, 2, 11, 12, 3, 10, 1, 8, 5, 4, 9]
Map: [0, 13, 6, 7, 2, 11, 12, 14, 3, 10, 1, 8, 5, 4, 9]
Map: [0, 13, 15, 6, 7, 2, 11, 12, 14, 3, 10, 1, 8, 5, 4, 9]
Map: [0, 13, 15, 6, 7, 2, 11, 16, 12, 14, 3, 10, 1, 8, 5, 4, 9]
[...]

PR #504, opened 2020-01-30
Write Enrollments data to the block.

TrustHenry:
Writing Enrollment data to a block determines the height of the block that can participate in the validation as a validator.
Validators needed for the consensus process can be determined and the number of validators between nodes can be set.
- The Enrollment data should be recorded only in the enroll start block.
- The node store the enrolled block height.

~~Reach consensus on transaction and enrollment set
The enrollments to be included in a block must first
be voted on by the consensus protocol.~~

Relates to #493 `Enrollment data should be written to the block.`
~~Relates to  #565 `Make Enrollment Part of Consensus`~~

### DIFF
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -1362,7 +1362,6 @@ unittest
         tx.inputs[0].signature = keypair.secret.sign(hashFull(tx)[]);
         txs_3 ~= tx;
     }
-    auto block3 = makeNewBlock(block2, txs_3);
 
     Pair signature_noise = Pair.random;
     Pair node_key_pair;
@@ -1376,7 +1375,7 @@ unittest
     enroll1.cycle_length = 1008;
     enroll1.enroll_sig = sign(node_key_pair.v, node_key_pair.V, signature_noise.V,
         signature_noise.v, enroll1);
- 
+
     auto utxo_hash2 = utxo_set.getHash(hashFull(txs_2[1]), 0);
     Enrollment enroll2;
     enroll2.utxo_key = utxo_hash2;
@@ -1389,13 +1388,9 @@ unittest
     enrollments ~= enroll1;
     enrollments ~= enroll2;
     enrollments.sort!("a.utxo_key < b.utxo_key");
-    block3.header.enrollments.length = 0;
-    block3.header.enrollments = enrollments;
+    auto block3 = makeNewBlock(block2, txs_3, enrollments);
     assert(block3.isValid(block2.header.height, hashFull(block2.header), findUTXO));
-
     enrollments.sort!("a.utxo_key > b.utxo_key");
-    block3.header.enrollments.length = 0;
-    block3.header.enrollments = enrollments;
     findUTXO = utxo_set.getUTXOFinder();
     // Block: The enrollments are not sorted in ascending order
     assert(!block3.isValid(block2.header.height, hashFull(block2.header), findUTXO));
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -383,11 +383,12 @@ unittest
     Params:
         prev_block = the previous block
         txs = the transactions that will be contained in the new block
+        enrollments = the enrollments that will be contained in the new block
 
 *******************************************************************************/
 
 public Block makeNewBlock (Transactions)(const ref Block prev_block,
-    Transactions txs) @safe nothrow
+    Transactions txs, Enrollment[] enrollments) @safe nothrow
     if (isInputRange!Transactions)
 {
     Block block;
@@ -398,8 +399,21 @@ public Block makeNewBlock (Transactions)(const ref Block prev_block,
     block.txs.sort;
 
     block.header.merkle_root = block.buildMerkleTree();
-
+    block.header.enrollments = enrollments;
+    assert(block.header.enrollments.isStrictlyMonotonic!
+        ("a.utxo_key < b.utxo_key"));
     return block;
+}
+
+/// only used in unittests
+version (unittest)
+{
+    public Block makeNewBlock (Transactions)(const ref Block prev_block,
+        Transactions txs) @safe nothrow
+        if (isInputRange!Transactions)
+    {
+        return makeNewBlock(prev_block, txs, null);
+    }
 }
 
 ///
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -22,6 +22,7 @@ import agora.common.TransactionPool;
 import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.ConsensusData;
+import agora.consensus.data.Enrollment;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSet;
 import agora.consensus.EnrollmentManager;
@@ -118,7 +119,8 @@ public class Ledger
         nothrow @trusted
     {
         scope (failure) assert(0);
-        auto block = makeNewBlock(this.last_block, data.tx_set.byKey());
+        auto block = makeNewBlock(this.last_block, data.tx_set.byKey(),
+            data.enrolls);
         return this.acceptBlock(block);
     }
 
@@ -188,6 +190,9 @@ public class Ledger
 
         Add a validated block to the Ledger,
         and add all of its outputs to the UTXO set.
+        If there are any enrollments in the block,
+        update the enrollments' starting block heights
+        in the enrollment manager.
 
         Params:
             block = the block to add
@@ -201,6 +206,11 @@ public class Ledger
         this.updateUTXOSet(block);
         if (!this.storage.saveBlock(block))
             assert(0);
+
+        foreach (enrollment; block.header.enrollments)
+            if (!this.enroll_man.updateEnrolledHeight(enrollment.utxo_key,
+                block.header.height))
+                assert(0);
 
         // read back and cache the last block
         if (!this.storage.readLastBlock(this.last_block))
@@ -833,6 +843,202 @@ unittest
     blocks = ledger.getBlocksFrom(0).take(10).array;
     assert(blocks.length == 3);
     assert(blocks[2].header.height == 2);
+}
+
+// Create freeze transactions and create enrollments to
+// test if it is stored in a block.
+unittest
+{
+    import agora.common.crypto.ECC;
+    import agora.common.crypto.Key;
+    import agora.common.crypto.Schnorr;
+    import agora.common.Hash;
+    import agora.common.Types;
+
+    auto gen_key = getGenesisKeyPair();
+
+    KeyPair[] splited_keys = getRandomKeyPairs();
+    KeyPair[] in_key_pairs_normal;
+    KeyPair[] out_key_pairs_normal;
+    Transaction[] last_txs_normal;
+    KeyPair[] in_key_pairs_freeze;
+    KeyPair[] out_key_pairs_freeze;
+    Transaction[] last_txs_freeze;
+
+    auto storage = new MemBlockStorage();
+    auto pool = new TransactionPool(":memory:");
+    scope(exit) pool.shutdown();
+    auto utxo_set = new UTXOSet(":memory:");
+    scope (exit) utxo_set.shutdown();
+    auto config = new Config();
+    auto enroll_man = new EnrollmentManager(":memory:", gen_key);
+    scope (exit) enroll_man.shutdown();
+    config.node.is_validator = true;
+    scope ledger = new Ledger(pool, utxo_set, storage, enroll_man, config.node);
+
+    Transaction[] splited_txex;
+    // Divide 8 'Outputs' that are included in Genesis Block by 40,000
+    // It generates eight addresses and eight transactions,
+    // and one transaction has eight Outputs with a value of 40,000 values.
+    void splitGenesis ()
+    {
+        splited_txex = splitGenesisTransaction(getGenKeyPairs(), splited_keys);
+        splited_txex.each!((tx)
+        {
+            assert(ledger.acceptTransaction(tx));
+        });
+        ledger.forceCreateBlock();
+    }
+
+    in_key_pairs_normal.length = 0;
+    foreach (idx; 0 .. Block.TxsInBlock)
+        in_key_pairs_normal ~= splited_keys[0];
+
+    out_key_pairs_normal = getRandomKeyPairs();
+
+    // generate nomal transactions to form a block
+    void genNormalBlockTransactions (size_t count, bool is_valid = true)
+    {
+        foreach (idx; 0 .. count)
+        {
+            auto txes = makeTransactionForFreezing (
+                in_key_pairs_normal,
+                out_key_pairs_normal,
+                TxType.Payment,
+                last_txs_normal,
+                splited_txex[0]);
+
+            txes.each!((tx)
+                {
+                    assert(ledger.acceptTransaction(tx) == is_valid);
+                });
+            ledger.forceCreateBlock();
+
+            if (is_valid)
+            {
+                // keep track of last tx's to chain them to
+                last_txs_normal = txes[$ - Block.TxsInBlock .. $];
+
+                in_key_pairs_normal = out_key_pairs_normal;
+                out_key_pairs_normal = getRandomKeyPairs();
+            }
+        }
+    }
+
+    in_key_pairs_freeze.length = 0;
+    foreach (idx; 0 .. Block.TxsInBlock)
+        in_key_pairs_freeze ~= splited_keys[1];
+
+    out_key_pairs_freeze = getRandomKeyPairs();
+
+    // generate freezing transactions to form a block
+    void genBlockTransactionsFreeze (size_t count, TxType tx_type, bool is_valid = true)
+    {
+        foreach (idx; 0 .. count)
+        {
+            auto txes = makeTransactionForFreezing (
+                in_key_pairs_freeze,
+                out_key_pairs_freeze,
+                tx_type,
+                last_txs_freeze,
+                splited_txex[1]);
+
+            txes.each!((tx)
+                {
+                    assert(ledger.acceptTransaction(tx) == is_valid);
+                });
+            ledger.forceCreateBlock();
+
+            if (is_valid)
+            {
+                // keep track of last tx's to chain them to
+                last_txs_freeze = txes[$ - Block.TxsInBlock .. $];
+
+                in_key_pairs_freeze = out_key_pairs_freeze;
+                out_key_pairs_freeze = getRandomKeyPairs();
+            }
+        }
+    }
+
+    splitGenesis();
+    assert(ledger.getBlockHeight() == 1);
+
+    genNormalBlockTransactions(1);
+    assert(ledger.getBlockHeight() == 2);
+
+    genBlockTransactionsFreeze(1, TxType.Freeze);
+    assert(ledger.getBlockHeight() == 3);
+
+    auto blocks = ledger.getBlocksFrom(0).take(10);
+
+    // make enrollments
+    KeyPair[] enroll_key_pair;
+    foreach (txid, tx; blocks[3].txs)
+        foreach (key_pair; in_key_pairs_freeze)
+            if (tx.outputs[0].address == key_pair.address)
+                enroll_key_pair ~= key_pair;
+
+    auto utxo_hash_1 = utxo_set.getHash(hashFull(blocks[3].txs[0]),0);
+    auto utxo_hash_2 = utxo_set.getHash(hashFull(blocks[3].txs[1]),0);
+    auto utxo_hash_3 = utxo_set.getHash(hashFull(blocks[3].txs[2]),0);
+
+    Pair signature_noise = Pair.random;
+    Pair node_key_pair_1;
+    node_key_pair_1.v = secretKeyToCurveScalar(enroll_key_pair[0].secret);
+    node_key_pair_1.V = node_key_pair_1.v.toPoint();
+
+    Pair node_key_pair_2;
+    node_key_pair_2.v = secretKeyToCurveScalar(enroll_key_pair[1].secret);
+    node_key_pair_2.V = node_key_pair_2.v.toPoint();
+
+    Pair node_key_pair_3;
+    node_key_pair_3.v = secretKeyToCurveScalar(enroll_key_pair[2].secret);
+    node_key_pair_3.V = node_key_pair_3.v.toPoint();
+
+    Enrollment enroll_1;
+    enroll_1.utxo_key = utxo_hash_1;
+    enroll_1.random_seed = hashFull(Scalar.random());
+    enroll_1.cycle_length = 1008;
+    enroll_1.enroll_sig = sign(node_key_pair_1.v, node_key_pair_1.V, signature_noise.V,
+        signature_noise.v, enroll_1);
+
+    Enrollment enroll_2;
+    enroll_2.utxo_key = utxo_hash_2;
+    enroll_2.random_seed = hashFull(Scalar.random());
+    enroll_2.cycle_length = 1008;
+    enroll_2.enroll_sig = sign(node_key_pair_2.v, node_key_pair_2.V, signature_noise.V,
+        signature_noise.v, enroll_2);
+
+    Enrollment enroll_3;
+    enroll_3.utxo_key = utxo_hash_3;
+    enroll_3.random_seed = hashFull(Scalar.random());
+    enroll_3.cycle_length = 1008;
+    enroll_3.enroll_sig = sign(node_key_pair_3.v, node_key_pair_3.V, signature_noise.V,
+        signature_noise.v, enroll_3);
+
+    Enrollment[] enrollments ;
+    enrollments ~= enroll_1;
+    enrollments ~= enroll_2;
+    enrollments ~= enroll_3;
+
+    auto findUTXO = utxo_set.getUTXOFinder();
+    assert(enroll_man.add(3, findUTXO, enroll_1));
+    assert(enroll_man.add(3, findUTXO, enroll_2));
+    assert(enroll_man.add(3, findUTXO, enroll_3));
+    assert(enroll_man.hasEnrollment(utxo_hash_1));
+    assert(enroll_man.hasEnrollment(utxo_hash_2));
+    assert(enroll_man.hasEnrollment(utxo_hash_3));
+    assert(enroll_man.count() == 3);
+
+    genNormalBlockTransactions(1);
+    assert(ledger.getBlockHeight() == 4);
+
+    // Check if there are any unregistered enrollments
+    Enrollment[] unreg_enrollments;
+    assert(enroll_man.getUnregistered(unreg_enrollments) is null);
+    auto block_4 = ledger.getBlocksFrom(4);
+    enrollments.sort!("a.utxo_key < b.utxo_key");
+    assert(block_4[0].header.enrollments == enrollments);
 }
 
 version (unittest)
--- a/tests/unit/BlockStorage.d
+++ b/tests/unit/BlockStorage.d
@@ -66,7 +66,7 @@ private void main ()
             ]
         );
         tx.inputs[0].signature = gen_key_pair.secret.sign(hashFull(tx)[]);
-        blocks ~= makeNewBlock(blocks[$ - 1], [tx]);
+        blocks ~= makeNewBlock(blocks[$ - 1], [tx], null);
         block_hashes ~= hashFull(blocks[$ - 1].header);
         storage.saveBlock(blocks[$ - 1]);
     }
--- a/tests/unit/BlockStorageChecksum.d
+++ b/tests/unit/BlockStorageChecksum.d
@@ -53,7 +53,7 @@ private void writeBlocks (string path)
     foreach (block_idx; 0 .. BlockCount)
     {
         Transaction[8] txs;
-        auto block = makeNewBlock(blocks[$ - 1], txs[]);
+        auto block = makeNewBlock(blocks[$ - 1], txs[], null);
         storage.saveBlock(block);
         blocks ~= block;
     }
--- a/tests/unit/BlockStorageMultiTx.d
+++ b/tests/unit/BlockStorageMultiTx.d
@@ -53,7 +53,7 @@ private void main ()
         // create enough tx's for a single block
         auto txs = makeChainedTransactions(gen_key_pair, last_txs, 1);
 
-        auto block = makeNewBlock(blocks[$ - 1], txs);
+        auto block = makeNewBlock(blocks[$ - 1], txs, null);
         storage.saveBlock(block);
         blocks ~= block;
         last_txs = txs;


TrustHenry:
~~Writing unit test code on ledger.~~
Ready

source/agora/consensus/data/Block.d
```diff
@@ -449,6 +450,7 @@ public Block makeNewBlock (Transactions)(const ref Block prev_block,
     block.txs.sort;
 
     block.header.merkle_root = block.buildMerkleTree();
+    enrollments.each!(enrollment => block.header.enrollments ~= enrollment);
```
AndrejMitrovic: `block.header.enrollments = enrollments` should be good enough.

AndrejMitrovic:
I'm not sure if the way the commits were split makes sense. All of these seem like they atomically belong together as a single commit.

@Geod24 do you agree?

TrustHenry:
> I'm not sure if the way the commits were split makes sense. All of these seem like they atomically belong together as a single commit.

If this looks like it, I'll make these committs single-commit

source/agora/consensus/data/Block.d
TrustHenry: applied the suggestion.

source/agora/consensus/data/Block.d
```diff
@@ -434,11 +434,12 @@ unittest
     Params:
         prev_block = the previous block
         txs = the transactions that will be contained in the new block
+        enrollments = the enrollments that will be contained in the new block
 
 *******************************************************************************/
 
 public Block makeNewBlock (Transactions)(const ref Block prev_block,
-    Transactions txs) @safe nothrow
+    Transactions txs, Enrollment[] enrollments = null) @safe nothrow
```
AndrejMitrovic: Did you made it a default parameter so you don't have to change a lot of unittest code?

In general, we should not design our functions so they are specialized / convenient for existing unittest code. So I would not make this a default parameter. The only exception to that is for `version (unittest)` functions, which are only used as helper functions for unittest code.

source/agora/node/Ledger.d
```diff
@@ -68,17 +73,21 @@ public class Ledger
             pool = the transaction pool
             utxo_set = the set of unspent outputs
             storage = the block storage
+            enroll_man = the enrollmentManager
+            node_config = the node config
```
AndrejMitrovic: Nice catch. 👍

source/agora/node/Ledger.d
```diff
@@ -136,7 +145,9 @@ public class Ledger
     public bool onTXSetExternalized (Set!Transaction txs) nothrow @trusted
     {
         scope (failure) assert(0);
-        auto block = makeNewBlock(this.last_block, txs.byKey());
+        Enrollment[] enrollments;
+        auto block = makeNewBlock(this.last_block, txs.byKey(),
+            this.enroll_man.getUnregisteredEnrollments(enrollments));
```
AndrejMitrovic: `getUnregisteredEnrollments` takes a buffer, but you are passing it an empty array. If you want to reuse memory, you should mark `Enrollment[] enrollments;` as `static`.

source/agora/node/Ledger.d
```diff
@@ -222,6 +233,8 @@ public class Ledger
 
         Add a validated block to the Ledger,
         and add all of its outputs to the UTXO set.
+        and if there are additional enrollments,
```
```suggestion
        If there are any enrollments in the block,
        update the enrollments' starting block heights
        in the enrollment manager.
```

source/agora/node/Ledger.d
```diff
@@ -236,6 +249,11 @@ public class Ledger
         if (!this.storage.saveBlock(block))
             assert(0);
 
+        foreach (enrollment; block.header.enrollments)
+            if (!this.enroll_man.updateEnrolledHeight(enrollment.utxo_key,
```
AndrejMitrovic, line 211: This is a great example of what @Geod24 talked about. If `updateEnrolledHeight` fails, we shut down the node immediately with `assert(0);`.

But then the question is, why is `updateEnrolledHeight` returning a boolean status code? It should either throw an Error or just assert(0) itself, and the function should be made `nothrow` and not return any status code.

I will file a separate issue for this, so you don't have to change anything here @TrustHenry.

source/agora/node/Ledger.d
```diff
@@ -830,6 +864,7 @@ unittest
     blocks = ledger.getBlocksFrom(0).take(10).array;
     assert(blocks.length == 3);
     assert(blocks[2].header.height == 2);
+
```
AndrejMitrovic, line 973: Extra blank line.

AndrejMitrovic (changes requested):

source/agora/consensus/data/Block.d
TrustHenry: The intention of this is to assume that there may be no enrollments.
This is because most blocks don't have many cases.

AndrejMitrovic: This function is only called from a single place in Agora (I'm not talking about unittest code), and that's in the `Ledger`.  And you directly pass `his.enroll_man.getUnregisteredEnrollments(enrollments)`. So the default parameter is not necessary.

source/agora/node/Ledger.d
TrustHenry: applied the suggestion.

source/agora/node/Ledger.d
TrustHenry: applied the suggestion.

source/agora/node/Ledger.d
TrustHenry: Ah, it is unittest code lines. I'm doing now.

source/agora/node/Ledger.d
TrustHenry: I was thinking the same thing.

source/agora/consensus/data/Block.d
TrustHenry: Yes, the default parameter is not necessary.
I'll have to go back to my first commit.
I had modified many unit test codes. :)

source/agora/node/Ledger.d
TrustHenry: i'm add static.

source/agora/node/Ledger.d
TrustHenry: applied the suggestion.

TrustHenry:
Ready for review.

TrustHenry:
@Geod24 ping

TrustHenry:
@Geod24 @AndrejMitrovic ping.

TrustHenry:
I rebased it.

AndrejMitrovic:
Could you please rename the commit message?

`Write Enrollment data to the block` is the correct English. 👼

TrustHenry:
> Could you please rename the commit message?
> 
> `Write Enrollment data to the block` is the correct English. 👼

Applied the suggestion.

TrustHenry:
@Geod24 @AndrejMitrovic ping.

AndrejMitrovic (approved):

tests/unit/BlockStorageMultiTx.d
```diff
@@ -58,7 +59,7 @@ private void main ()
         // create enough tx's for a single block
         auto txs = makeChainedTransactions(gen_key_pair, last_txs, 1);
 
-        auto block = makeNewBlock(blocks[$ - 1], txs);
+        auto block = makeNewBlock(blocks[$ - 1], txs, [Enrollment.init]);
```
Geod24: Is that correct ? Why passing `Enrollment.init` (which is likely to not pass validation)

source/agora/test/NetworkManager.d
```diff
@@ -92,7 +93,7 @@ unittest
             {
                 auto txs = () @trusted { return makeChainedTransactions(prev_key, last_tx, 1); }();
                 last_tx = txs;
-                auto block = makeNewBlock(last_block, txs);
+                auto block = makeNewBlock(last_block, txs, [Enrollment.init]);
```
Geod24: Likewise, why `Enrollment.init` ?

source/agora/consensus/Validation.d
```diff
@@ -22,6 +22,7 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSet;
+import agora.consensus.EnrollmentManager;
```
Geod24: Looks like this is only used in `unittest`, hence it should be versioned accordingly

source/agora/consensus/Validation.d
```diff
@@ -993,10 +994,14 @@ unittest
...
     auto gen_hash = GenesisBlock.header.hashFull();
+    auto enroll_man = new EnrollmentManager(":memory:", gen_key);
+    scope (exit) enroll_man.shutdown();
+    Enrollment[] enrollments;
 
     tx_map[GenesisTransaction.hashFull()] = [GenesisTransaction];
     auto txs = makeChainedTransactions(gen_key, null, 1).sort.array;
-    auto block = makeNewBlock(GenesisBlock, txs);
+    auto block = makeNewBlock(GenesisBlock, txs,
+        enroll_man.getUnregisteredEnrollments(enrollments));
```
Geod24: Side note, for a future PR: Since `getUnregisteredEnrollments` is a method of `EnrollmentManager`, we should probably just remove the `Enrollments` and name it `getUnregistered`.

source/agora/consensus/data/Block.d
```diff
@@ -449,15 +450,16 @@ public Block makeNewBlock (Transactions)(const ref Block prev_block,
     block.txs.sort;
 
     block.header.merkle_root = block.buildMerkleTree();
+    block.header.enrollments = enrollments;
```
Geod24, line 402: Hum they aren't sorted ??

source/agora/consensus/data/Block.d
```diff
@@ -449,15 +450,16 @@ public Block makeNewBlock (Transactions)(const ref Block prev_block,
...
     return block;
 }
 
 ///
 @safe nothrow unittest
 {
-    auto new_block = makeNewBlock(GenesisBlock, [Transaction.init]);
-    auto rng_block = makeNewBlock(GenesisBlock, [Transaction.init].take(1));
+    auto new_block = makeNewBlock(GenesisBlock, [Transaction.init], [Enrollment.init]);
+    auto rng_block = makeNewBlock(GenesisBlock, [Transaction.init].take(1), [Enrollment.init]);
```
Geod24: Adding an invalid Enrollment doesn't seem like the best test

Geod24 (changes requested):
- It's red in CircleCI, probably because you need to make the import in `Validation`  `version(unittest)`
- Some tests seems fishy, why `Enrollment.init` is sometimes used instead of `null` ?
- Shouldn't the enrollments be sorted when added to the block ? There should be a test they are.

source/agora/consensus/Validation.d
TrustHenry: moved to unittest.

source/agora/consensus/Validation.d
TrustHenry: I will write a PR for that part.

source/agora/consensus/data/Block.d
TrustHenry: Modify to correct value.

source/agora/consensus/data/Block.d
TrustHenry: Added sort logic.

source/agora/test/NetworkManager.d
TrustHenry: Modify to correct value.

tests/unit/BlockStorageMultiTx.d
TrustHenry: Modify to correct value.

source/agora/consensus/data/Block.d
TrustHenry: The part I was worried about is whether I should align here once more.

source/agora/consensus/data/Block.d
TrustHenry: The part I was worried about is whether I should align here once more.

TrustHenry: I thought it'd be better to check if it's aligned here.

TrustHenry:
@Geod24 ping.

Geod24:
A note on design: In your case, you see that you had to add a bunch of `null` after `makeNewBlock`. Why is that ? Because the last parameter to `makeNewBlock` is optional. It's completely valid not to have any enrollment in a block. In fact, it's more likely that blocks will *not* have enrollment because it should be rare enough.

So what do we do about it ? Do we add `null` everywhere ? One option would be to add a default parameter to `makeNewBlock`, so that you can still call it with two parameters.
However, that runs the risk that we forget to pass enrollment data in production code. While it's not great, it is a kind of bug that is easily spotted, but could lead to consensus failure.

The other option is to provide a `unittest`-only function that only takes 2 parameters and forwards to `makeNewBlock`. It's usually a good first step (almost nothing can go wrong with that approach).

AndrejMitrovic:
It was originally a default parameter, I requested it be removed because of those reasons; it's easy to forget to pass the argument. It might be useful to have it as a default for unittest code.

Geod24:
Could you add a test to the system tests as well ?

source/agora/consensus/Validation.d
```diff
@@ -993,10 +994,14 @@ unittest
     auto gen_key = getGenesisKeyPair();
     assert(GenesisBlock.isValid(GenesisBlock.header.height, Hash.init, null));
     auto gen_hash = GenesisBlock.header.hashFull();
+    auto enroll_man = new EnrollmentManager(":memory:", gen_key);
+    scope (exit) enroll_man.shutdown();
+    Enrollment[] enrollments;
```
Geod24: Most of the changes in this file seems to have no effect. No `Enrollment` is every registered, thus all calls to `getUnregisteredEnrollments` will return `null`.

source/agora/consensus/Validation.d
```diff
@@ -1389,13 +1398,9 @@ unittest
     enrollments ~= enroll1;
     enrollments ~= enroll2;
     enrollments.sort!("a.utxo_key < b.utxo_key");
-    block3.header.enrollments.length = 0;
-    block3.header.enrollments = enrollments;
+    auto block3 = makeNewBlock(block2, txs_3, enrollments);
```
Geod24, line 1391: This being obviously the only exception

source/agora/node/BlockStorage.d
```diff
@@ -1107,6 +1107,7 @@ unittest
 {
     import agora.common.crypto.Key;
     import agora.consensus.data.Transaction;
+    import agora.consensus.data.Enrollment;
```
Geod24: Why the import ?

tests/unit/BlockStorageMultiTx.d
```diff
@@ -19,6 +19,7 @@ import agora.common.crypto.Key;
 import agora.common.Hash;
 import agora.common.Types;
 import agora.consensus.data.Block;
+import agora.consensus.data.Enrollment;
```
Geod24: Why the import ?

source/agora/test/NetworkManager.d
```diff
@@ -56,6 +56,7 @@ unittest
 {
     import agora.common.BanManager;
     import agora.consensus.data.Block;
+    import agora.consensus.data.Enrollment;
```
Geod24: Why the import ?

source/agora/node/Ledger.d
```diff
@@ -542,17 +564,20 @@ unittest
 {
     import agora.common.crypto.Key;
 
+    auto gen_key_pair = getGenesisKeyPair();
```
Geod24: This kind of refactoring is good, but better done in its own commit

source/agora/node/Ledger.d
```diff
@@ -589,22 +614,26 @@ unittest
...
     Block invalid_block;  // default-initialized should be invalid
     assert(!ledger.acceptBlock(invalid_block));
 
-    auto gen_key_pair = getGenesisKeyPair();
     auto txs = makeChainedTransactions(gen_key_pair, null, 1);
-
-    auto valid_block = makeNewBlock(GenesisBlock, txs);
+    Enrollment[] enrollments;
+    auto valid_block = makeNewBlock(GenesisBlock, txs,
+        enroll_man.getUnregisteredEnrollments(enrollments));
```
Geod24: Likewise, you'll alway get `null` here

Geod24:
What happens if we receive a block that has passed consensus but for which we don't have the enrollment data locally (for example, we didn't receive it when it was broadcasted) ?

AndrejMitrovic:
> What happens if we receive a block that has passed consensus but for which we don't have the enrollment data locally (for example, we didn't receive it when it was broadcasted) ?

That means the node also doesn't have the previous blocks, therefore it should do catch-up first. If it has the previous blocks, it means it has the previous enrollments.

Geod24:
> That means the node also doesn't have the previous blocks, therefore it should do catch-up first. If it has the previous blocks, it means it has the previous enrollments.

Discussed IRL, that's not actually the case if it's an *unregistered* enrollment (should have made this more precise in the previous comment).

source/agora/consensus/Validation.d
TrustHenry: I used `testMakeNewBlock`

source/agora/consensus/Validation.d
TrustHenry: Yes,

source/agora/node/BlockStorage.d
TrustHenry: I deleted it.

tests/unit/BlockStorageMultiTx.d
TrustHenry: I deleted it.

source/agora/test/NetworkManager.d
TrustHenry: I deleted it.

source/agora/node/Ledger.d
TrustHenry: Yes,

source/agora/node/Ledger.d
TrustHenry: I used `testMakeNewBlock`

TrustHenry:
> A note on design: In your case, you see that you had to add a bunch of `null` after `makeNewBlock`. Why is that ? Because the last parameter to `makeNewBlock` is optional. It's completely valid not to have any enrollment in a block. In fact, it's more likely that blocks will _not_ have enrollment because it should be rare enough.
> 
> So what do we do about it ? Do we add `null` everywhere ? One option would be to add a default parameter to `makeNewBlock`, so that you can still call it with two parameters.
> However, that runs the risk that we forget to pass enrollment data in production code. While it's not great, it is a kind of bug that is easily spotted, but could lead to consensus failure.
> 
> The other option is to provide a `unittest`-only function that only takes 2 parameters and forwards to `makeNewBlock`. It's usually a good first step (almost nothing can go wrong with that approach).

I make new function `testMakeNewBlock` for unnittest

TrustHenry:
> Could you add a test to the system tests as well ?

Do you mean to test file? `agora.test.EnrollmentManager`
Can we make it a different PR from this one?

TrustHenry:
@Geod24 ping

Geod24:
> Do you mean to test file?

No, the system tests are [here](https://github.com/bpfkorea/agora/tree/v0.x.x/tests/system).

> Can we make it a different PR from this one?

We should never add code before tests. We *can* add tests before code (e.g. "this is currently how things work" then add code and alter the tests to show the difference in behavior) but we should *never* add code without the appropriate tests.

Geod24:
W.r.t `testMakeNewBlock`. I don't know how other people feel about it, I was thinking a `version(unittest)` overload to`makeNewBlock` was enough. Why the different name ?

TrustHenry:
> W.r.t `testMakeNewBlock`. I don't know how other people feel about it, I was thinking a `version(unittest)` overload to`makeNewBlock` was enough. Why the different name ?

```
version (unittest)
{
    public Block makeNewBlock (Transactions)(const ref Block prev_block,
        Transactions txs) @safe nothrow
        if (isInputRange!Transactions)
    {
        return makeNewBlock(prev_block, txs, null);
    }
}
```

overload to `makeNewBlock`
I applied the suggestion.

source/agora/consensus/data/Block.d
```diff
@@ -449,13 +450,27 @@ public Block makeNewBlock (Transactions)(const ref Block prev_block,
...
+        if (isInputRange!Transactions)
+    {
+        return makeNewBlock(prev_block, txs, null);
+    }
+}
+
 ///
 @safe nothrow unittest
 {
+
```
Geod24:
```suggestion
```

tests/unit/BlockStorageMultiTx.d
```diff
@@ -58,7 +58,7 @@ private void main ()
         // create enough tx's for a single block
         auto txs = makeChainedTransactions(gen_key_pair, last_txs, 1);
 
-        auto block = makeNewBlock(blocks[$ - 1], txs);
+        auto block = makeNewBlock(blocks[$ - 1], txs, null);
```
Geod24, line 56:
```suggestion
        auto block = makeNewBlock(blocks[$ - 1], txs);
```

tests/unit/BlockStorageChecksum.d
```diff
@@ -55,7 +55,7 @@ private void writeBlocks (string path)
     foreach (block_idx; 0 .. BlockCount)
     {
         Transaction[8] txs;
-        auto block = makeNewBlock(blocks[$ - 1], txs[]);
+        auto block = makeNewBlock(blocks[$ - 1], txs[], null);
```
Geod24, line 56:
```suggestion
        auto block = makeNewBlock(blocks[$ - 1], txs[]);
```

tests/unit/BlockStorage.d
```diff
@@ -72,7 +72,7 @@ private void main ()
             ]
         );
         tx.inputs[0].signature = gen_key_pair.secret.sign(hashFull(tx)[]);
-        blocks ~= makeNewBlock(blocks[$ - 1], [tx]);
+        blocks ~= makeNewBlock(blocks[$ - 1], [tx], null);
```
Geod24, line 69:
```suggestion
        blocks ~= makeNewBlock(blocks[$ - 1], [tx]);
```

source/agora/consensus/data/Block.d
TrustHenry: applied the suggestion.

tests/unit/BlockStorageMultiTx.d
TrustHenry: applied the suggestion.

tests/unit/BlockStorageChecksum.d
TrustHenry: applied the suggestion.

tests/unit/BlockStorage.d
TrustHenry: applied the suggestion.

TrustHenry: Where do `makeNewBlock (blocks [$ - 1], [tx])` look?

Geod24: Oh sorry I didn't realize it was in `tests/unit`. So yeah, revert to `null`.

TrustHenry: Oh okay. I was confused for a while.

TrustHenry:
[WIP] system integration test of the enrollment.
The problem here is that the block height does not increase. :(

TrustHenry:
I made system integration test of the enrollment.
There are a lot of overlapping codes, so I'm going to try to reduce them.
But it still works.

TrustHenry:
@Geod24  Ready for review

source/agora/node/Ledger.d
```diff
@@ -790,10 +822,20 @@ private KeyPair[] getRandomKeyPairs ()
 // Use a transaction with the type 'TxType.Freeze' to create a block and test UTXOSet.
```
AndrejMitrovic: @bpfkorea/core unittests should be self-contained. We shouldn't pack every single feature test into a single unittest block. In fact, this block is commented as testing only freeze transactions and the UTXO set, but now it also tests Enrollments too.

In the future if any of these tests end up having to be changed because of some feature change, then it will become very complicated to edit and fix the tests.

source/agora/node/Ledger.d
```diff
@@ -1033,6 +1116,65 @@ unittest
     genBlockTransactionsFreeze(1, TxType.Freeze);
     assert(ledger.getBlockHeight() == 3);
 
+    // Create enrollments and add it in The block.
```
AndrejMitrovic: Same comment here about testing multitude of features in the same unittest.

AndrejMitrovic:
There is an (accidental?) update of the ocean submodule in the second commit.

tests/system/source/main.d
```diff
@@ -81,37 +111,179 @@ void main ()
...
+            tx.inputs[0].signature = signature;
+            clients[2].putTransaction(tx);
+        }
+
+        checkBlockHeight(4);
+    }
+}
+
+/// Check block generation
+private void checkBlockHeight (ulong height)
```
AndrejMitrovic: It seems like this commit does two things: refactors existing code and adds things like `checkBlockHeight`, and then also adds a test for the Enrollment.

Ideally they would be in separate commits.

source/agora/node/Ledger.d
TrustHenry: I made self-contained unitest.

source/agora/node/Ledger.d
TrustHenry: It has been processed as above.

TrustHenry:
> There is an (accidental?) update of the ocean submodule in the second commit.

Oh, accidental

tests/system/source/main.d
TrustHenry: It was separated by separate commits.

TrustHenry:
Ready for review

tests/system/source/main.d
```diff
@@ -81,36 +84,41 @@ void main ()
...
+    API[NodeCnt] clients;
+    foreach (idx, const ref addr; Addrs)
+        clients[idx] = new RestInterfaceClient!API(addr.withSchema());
+
+    Hash blockHash;
+    size_t times; // Number of times we slept for 50 msecs
+    foreach (idx, ref client; clients)
+    {
+    Start:
+        Thread.sleep(50.msecs);
```
AndrejMitrovic: Since you're refactoring, we could get rid of this use of `goto`:

```D
do
{
    Thread.sleep(50.msecs);
    const getHeight = client.getBlockHeight();
}
while (getHeight < height && times++ < 100);  // Retry if we're too early
```

I guess someone was too lazy to write a `do/while` and went with `goto` instead.

tests/system/source/main.d
```diff
@@ -88,6 +114,143 @@ void main ()
...
+
+        // clients[0] Create frozen transactions and enrolment
+        const Block[] blocks_1 = clients[0].getBlocksFrom(1,1);
+        assert(blocks_1[0].header.height == 1);
+
+        foreach (idx; 0 .. Block.TxsInBlock)
+        {
+            Transaction tx;
+
+            if (keypairs.length > idx)
```
AndrejMitrovic: It's confusing to have a value that doesn't change on the left side. It should be on the right.

```suggestion
            if (idx < keypairs.length)
```

tests/system/source/main.d
```diff
@@ -88,6 +114,143 @@ void main ()
...
+            }
+            else
+            {
+                Transaction payment_tx = {TxType.Payment, [], []};
+                payment_tx.inputs ~= Input(hashFull(blocks_1[0].txs[idx]), 0);
+                foreach (idx2; 0 .. Block.TxsInBlock)
+                    payment_tx.outputs ~= Output(Amount.MinFreezeAmount, kp.address);
+
+                payment_tx.inputs[0].signature = kp.secret.sign(hashFull(payment_tx)[]);
+                clients[1].putTransaction(payment_tx);
```
AndrejMitrovic: Why are freeze transactions sent to one node and normal payment transactions sent to a different node? They will both become part of Block with height 2, right? Is there an issue when you just try to use `client[0]`? It should just gossip everything to other nodes automatically.

tests/system/source/main.d
```diff
@@ -88,6 +114,143 @@ void main ()
...
+                };
+                freeze_tx.inputs[0].signature = kp.secret.sign(hashFull(freeze_tx)[]);
+                clients[0].putTransaction(freeze_tx);
+                freeze_txs ~= freeze_tx;
+            }
+            else
+            {
+                Transaction payment_tx = {TxType.Payment, [], []};
+                payment_tx.inputs ~= Input(hashFull(blocks_1[0].txs[idx]), 0);
+                foreach (idx2; 0 .. Block.TxsInBlock)
```
AndrejMitrovic: It took me a long time to understand what was happening here. I suggest adding a comment:

/// 
```suggestion
                // create tx's to be spent by all tx's in block 3
                foreach (idx2; 0 .. Block.TxsInBlock)
```

Also, the tests only use the transaction in the first `payment_txs` array. I understand why it was written this way, it is just more complicated to reason about. It would make more sense to have a single `payment_tx` in place of `payment_txs`, and just do `if (payment_tx == Transaction.init) { /// initialize it }`

AndrejMitrovic: It would have been simpler if there was only one Output that had the sum of the freeze amounts. So `MinFreezeAmount * Block.TxsInBlock`.

However, `Amount` does not have `operator *` or `multiply()` implemented, and using `Amount.add()` is more work. So you can keep it this way for now.

tests/system/source/main.d
```diff
@@ -88,6 +114,143 @@ void main ()
         checkBlockHeight(1);
     }
 
+    /// Make Transactions for block height 2
```
AndrejMitrovic:
```suggestion
    /// Make Transactions for block height 2,
    /// which will contain N freeze tx's, and
    /// the rest of txs will have 8 outputs to the 
    /// same address
```

tests/system/source/main.d
```diff
@@ -88,6 +114,143 @@ void main ()
...
+                signature_noises[idx].V, signature_noises[idx].v, enroll);
+            enrolls ~= enroll;
+
+            clients[0].enrollValidator(enrolls[idx]);
+        }
+
+        Thread.sleep(500.msecs);
+
+        // Checking Enrolled
+        size_t times; // Number of times we slept for 50 msecs
```
AndrejMitrovic: Unused variable.

tests/system/source/main.d
```diff
@@ -88,6 +114,143 @@ void main ()
...
+    /// Create enrollments
+    {
+        API[NodeCnt] clients;
+        foreach (idx, const ref addr; Addrs)
+            clients[idx] = new RestInterfaceClient!API(addr.withSchema());
+
+        Hash[] utxo_hashs;
+        Pair[] signature_noises;
+        Pair[] node_key_pairs;
+        Enrollment[] enrolls;
```
AndrejMitrovic: This array is not used after the `foreach` loop, so it's not needed. Just create the enrollment inside of the foreach loop and send it to the client.

tests/system/source/main.d
```diff
@@ -88,6 +114,143 @@ void main ()
...
+                inputs: [Input(hashFull(blocks_3[3].txs[idx]), 0)],
+                outputs: [Output(Amount.MinFreezeAmount, kp.address)]
+            };
+
+            auto signature = kp.secret.sign(hashFull(tx)[]);
+            tx.inputs[0].signature = signature;
+            clients[2].putTransaction(tx);
+        }
+
+        checkBlockHeight(4);
```
AndrejMitrovic: You should check that the Enrollments are in the block? Otherwise this entire test doesn't serve any real purpose.

AndrejMitrovic (changes requested):

tests/system/source/main.d
TrustHenry: applied the suggestion.

tests/system/source/main.d
TrustHenry: It is a test. If it seems strange, the test was changed to send only to the `client[0]`

tests/system/source/main.d
TrustHenry: applied the suggestion.

TrustHenry: That's right. I've looked up the `operator *`, `multiply()` in Amount, but it hasn't been implemented yet.
I think it is because there is still a risk of implementation.

TrustHenry: we can treat it as an integer and a decimal point.
But So I can keep it this way for now.

tests/system/source/main.d
TrustHenry: move to `foreach`

tests/system/source/main.d
TrustHenry: > Also, the tests only use the transaction in the first payment_txs array. I understand why it was written this way, it is just more complicated to reason about. It would make more sense to have a single payment_tx in place of payment_txs, and just do if (payment_tx == Transaction.init) { /// initialize it }

I didn't understand the meaning of using single `payment_tx`. 
Could you tell me more?

AndrejMitrovic: You can keep it the way you wrote it. 

But what I originally meant was:

```diff
- Transaction[] payment_txs;
+ Transaction payment;
```

And:

```diff
- payment_txs ~= payment_tx;
+ if (payment == Transaction.init)
    payment = payment_tx;
```

Because only one payment UTXOs are later used. It's alright to keep it your way if you want to.

TrustHenry: Thank you. applied the suggestion.

AndrejMitrovic:
Thanks. This resolves my comments.

tests/system/source/main.d
TrustHenry: I added that check the enrollments are stored in the block

TrustHenry:
@Geod24 @AndrejMitrovic  Ready for review

source/agora/node/Ledger.d
```diff
@@ -222,6 +236,11 @@ public class Ledger
         if (!this.storage.saveBlock(block))
             assert(0);
 
+        foreach (enrollment; block.header.enrollments)
```
AndrejMitrovic, line 210: There is another condition here which @Geod24 mentioned (he mentioned it in real life).

If we have received a Block from the network we try to update the heights of the enrollments that are in the block.

But what if the enrollments are not in the EnrollmentManager? For example, if we received a Block before we received Enrollments via `registerEnrollment. Then `updateEnrolledHeight` will fail, and nodes will have an inconsistent state.

I'm not sure what the best approach is here. Add the enrollment from the block to the enrollment manager via `addEnrollment`? @Geod24

TrustHenry: >If we received a Block before we received our received Enrollments via `registerEnrollment` 

Can you tell me which part of the time you're talking about?

It creates a block until it receives enrollments at the time it is created.
`makeNewBlock` in `onTXSetExternalized`

However, if there isn't received `enrolment data` until the block is created, the node will create another block.

TrustHenry:
I will rebase it.

TrustHenry:
I rebase done.

AndrejMitrovic (dismissed):

AndrejMitrovic:
@TrustHenry this is how #565 can be implemented:

https://github.com/AndrejMitrovic/agora/commit/999e03196020621476f71c025e8f18e2f0bfbf0c

It's based on your pull request #504.

The missing part is integration tests for https://github.com/AndrejMitrovic/agora/blob/999e03196020621476f71c025e8f18e2f0bfbf0c/source/agora/node/Ledger.d#L373

Feel free to take the code and make it part of your PR.

TrustHenry:
Added @AndrejMitrovic  'Reach Consensus on Transaction and Enrollment Set'.

TrustHenry:
> The missing part is integration tests for https://github.com/AndrejMitrovic/agora/blob/999e03196020621476f71c025e8f18e2f0bfbf0c/source/agora/node/Ledger.d#L373

There is no way to test 'isInvalidEnrollmentReason' in ValidateConcentusData.

It's because the `enrolmentManager.add` validates it
There is likely to be a way for DB to force in order to insert invalid enrolments.
Is this the right way to integration tests?
Ah, this is also impossible because the node is opening the DB file.
`unable to open database file`

In the current version, normal enrolment is verified by `isInvalidEnrollmentReason`

AndrejMitrovic:
> Is this the right way to test?

The proper way to do this testing is to have byzantine nodes nominate bad enrollments, similar to how it's done in https://github.com/bpfkorea/agora/pull/548/commits.

It would be a little complicated to implement. But basically:

- Add ability to provide a custom Nominator (https://github.com/bpfkorea/agora/pull/548/commits/0ac3ee37411518e5118953ad82b0f536dc482850)
- Change `Nominator.nominate` into a `protected` method, so it can be overriden in unittests.
- Write a ByzantineNominator similar to the one in https://github.com/bpfkorea/agora/pull/548/commits/065245b1050c824f0e98e391dc8cb1fa575bb800, but where the `nominate` method is overriden and where the `Enrollment` in the `ConsensusData` is replaced with a fake one.
- One node should have the `ByzantineNominator ` instance
- The other nodes should have a `filter` on the `putTransaction` (to stop them from nominating)
- Then we would feed transactions to this byzantine node. The node would start nominating, and the test would check that other nodes have rejected the ConsensusData proposal (because the Enrollments are invalid).

If this sounds very complicated, I could try to get it done. Let me know what you think.

TrustHenry:
> > Is this the right way to test?
> 
> The proper way to do this testing is to have byzantine nodes nominate bad enrollments, similar to how it's done in https://github.com/bpfkorea/agora/pull/548/commits.
> 
> It would be a little complicated to implement. But basically:
> 
> * Add ability to provide a custom Nominator ([0ac3ee3](https://github.com/bpfkorea/agora/commit/0ac3ee37411518e5118953ad82b0f536dc482850))
> * Change `Nominator.nominate` into a `protected` method, so it can be overriden in unittests.
> * Write a ByzantineNominator similar to the one in [065245b](https://github.com/bpfkorea/agora/commit/065245b1050c824f0e98e391dc8cb1fa575bb800), but where the `nominate` method is overriden and where the `Enrollment` in the `ConsensusData` is replaced with a fake one.
> * One node should have the `ByzantineNominator ` instance
> * The other nodes should have a `filter` on the `putTransaction` (to stop them from nominating)
> * Then we would feed transactions to this byzantine node. The node would start nominating, and the test would check that other nodes have rejected the ConsensusData proposal (because the Enrollments are invalid).
> 
> If this sounds very complicated, I could try to get it done. Let me know what you think.

It's a complex test method, but it's great.
I'll try to make a fake node like the above.

TrustHenry:
When adding `enrolments` on a normal simple node test network, 
assert error at the `validateValue` The cause of the error is unknown.
Surely the issue that @AndrejMitrovic was talking about before seems to continue to arise.
Consensus data does not match.

https://github.com/bpfkorea/agora/pull/504/checks?check_run_id=464275141#step:5:125

source/agora/test/EnrollmentManager.d
```diff
@@ -167,3 +188,133 @@ unittest
...
+
+/// Enrollment part of consensus
+unittest
+{
+    import std.algorithm;
+    import std.conv;
+    import core.time;
+
+    TestConf conf = { nodes : 4, threshold : 4 };
+    auto network = makeTestNetwork(TestConf.init);
```
AndrejMitrovic: You are not passing the `conf` into `makeTestNetwork`, you are passing `TestConf.init`.

AndrejMitrovic: Is this on purpose?

AndrejMitrovic: Ah the test still fails, even with that change..

AndrejMitrovic:
Yes debugging this further it seems there's an error during deserialization:

> core.exception.RangeError@source/agora/common/Deserializer.d(141): Range violation

So it might be related to https://github.com/bpfkorea/agora/issues/559 / https://github.com/bpfkorea/agora/issues/561

source/agora/test/EnrollmentManager.d
TrustHenry: Yes, I did a simple test on purpose.

TrustHenry:
> Yes debugging this further it seems there's an error during deserialization:
> 
> > core.exception.RangeError@source/agora/common/Deserializer.d(141): Range violation
> 
> So it might be related to #559 / #561

That's what I'm guessing.

TrustHenry:
```
auto next_value = next.serializeFull().toVec();
auto data = deserializeFull!ConsensusData(cast(ubyte[])next_value[]);
assert(next == data);
```

These test codes have errors.
`ToVec` erases the enrolments in the array.

AndrejMitrovic:
It's not `toVec` that is the problem.

There seems to be an issue with serialization code:

```D

    Defines the data used when reaching consensus.

    Copyright:
        All rights reserved.

    License:

module agora.consensus.data.ConsensusData;

import agora.common.Set;
import scpd.types.Utils;

import agora.common.Deserializer;

/// Consensus data which is nominated & voted on
public struct ConsensusData
{
    /// The transaction set that is being nominated / voted on
    public Set!Transaction tx_set;

    /// The enrollments that are being nominated / voted on
    public Enrollment[] enrolls;
}

///
unittest
{
    ConsensusData next;
    next.tx_set.put(Transaction.init);
    next.enrolls ~= Enrollment.init;

    auto next_value = next.serializeFull();
    auto data = deserializeFull!ConsensusData(next_value);
    assert(next == data);
}
```

This produces a Range violation.

Geod24:
Blocked by https://github.com/bpfkorea/agora/pull/504

source/agora/test/EnrollmentManager.d
AndrejMitrovic: Then please remove the `TestConf` instance and also remove the `ByzantineManager` code, unless you intend to use it here. :)

Geod24:
```
Linking...
Program exited with code -10
```

Still failing, but will give a round of review.

source/agora/consensus/data/ConsensusData.d
```diff
@@ -0,0 +1,28 @@
...
+*******************************************************************************/
+
+module agora.consensus.data.ConsensusData;
+
+import agora.common.Set;
+import agora.consensus.data.Enrollment;
+import agora.consensus.data.Transaction;
+
+/// Consensus data which is nominated & voted on
+public struct ConsensusData
```
Geod24: Please add a `testSymmetry` check

source/agora/test/EnrollmentManager.d
```diff
@@ -167,3 +196,235 @@ unittest
...
+            log.info("{}(): Tx set nominated", __FUNCTION__);
+        }
+        else
+        {
+            log.info("{}(): Tx set rejected nomination", __FUNCTION__);
+        }
+    }
+}
+
+class BynzantineNode : TestNode
```
Geod24:
```suggestion
class ByzantineNode : TestNode
```

source/agora/test/EnrollmentManager.d
```diff
@@ -167,3 +196,235 @@ unittest
...
+    override void nominate (ulong slot_idx, ConsensusData prev,
+        ConsensusData next) @trusted
+    {
+        log.info("{}(): Proposing tx set for slot {}", __FUNCTION__, slot_idx);
+        if (next.enrolls.length > 0)
+            next.enrolls.length = 0;
+
+        auto prev_value = prev.serializeFull().toVec();
+        auto next_value = next.serializeFull().toVec();
+        if (this.scp.nominate(slot_idx, next_value, prev_value))
```
Geod24: Here is your segv. `this.scp` is never initialized.

source/agora/node/Node.d
```diff
@@ -462,6 +464,32 @@ public class Node : API
...
+            taskman = the task manager
+            quorum_peers = the network clients of the validator peers
+            quorum_set = the SCP quorum set
+
+        Returns:
+            the enrollment manager
+
+    ***************************************************************************/
+
+    protected Nominator getNominator (KeyPair key_pair, Ledger ledger,
```
Geod24: I'm not sure how I feel about this. We are adding more and more of those functions. Dependency injection is good, but it looks like we're injecting quite a lot.

source/agora/test/EnrollmentManager.d
```diff
@@ -167,3 +196,235 @@ unittest
...
+        txs ~= tx;
+    }
+
+    txs.each!(tx => node_1.putTransaction(tx));
+    containSameBlocks(nodes, 1).retryFor(8.seconds);
+
+    Enrollment[] enrolls;
+    foreach (node; nodes)
+    {
+        Enrollment enroll = node.createEnrollmentData();
```
Geod24: This API needs to be removed. We can't have an API asking node to enroll.

source/agora/node/Node.d
AndrejMitrovic: I agree (I originally wrote this) and we need a better way of simulating byzantine nodes. Probably the best way is to not instantiate Node, but our own separate Node implementing the API. I'll sketch something out in the coming days.

source/agora/test/EnrollmentManager.d
TrustHenry: It's already been made by @linked0  for the test.
https://github.com/bpfkorea/agora/blob/6f9f5cf3fd8c8bac109ae836051c36088fd6477e/source/agora/test/Base.d#L446

linked0: > This API needs to be removed. We can't have an API asking node to enroll.

We need a way to broadcast enrollment information which is implemented as `enrollValidator`. And, the `createEnrollmentData` was implemented only for integration test because there is no way to make EnrollmentManager create an enrollment information.

source/agora/test/EnrollmentManager.d
TrustHenry: I changed it to `Protected SCP* scp;` on the Nominator.

source/agora/consensus/data/ConsensusData.d
TrustHenry:
```
/// ConsensusData type testSymmetry check
unittest
{
    import agora.common.Set;
    import agora.common.Types;

    struct Data
    {
        public Set!Transaction tx_set;
    }

    testSymmetry!Data();
}
```

Ah, I think I found a bug.

Geod24: What's the output ?

AndrejMitrovic: If this is too much to workaround, it can be changed to be an array of transactions.

Also in the future we will replace all of this with hashes.

AndrejMitrovic: https://github.com/bpfkorea/agora/pull/603 has been merged, so this should be possible to do now.

Geod24:
> node-0_1_c21ecaf22ec4 | 2020-03-02T05:07:30.121109835Z core.exception.AssertError@/usr/include/d/std/range/primitives.d(2428): Attempting to fetch the front of an empty array of Transaction

Geod24:
Stack trace: https://travis-ci.com/bpfkorea/agora/jobs/292840108#L1781

source/agora/consensus/data/ConsensusData.d
TrustHenry: Do you think we should use `Set!Transaction` as a consensus data type?
Or would you rather write `Transacntion[]`?

Of course, we'll go to the hash later, but now.

TrustHenry: This issue is resolved by PR #609.
And I added a unit test case.

TrustHenry:
An agreement process was required to put the enrolment data into the block, and there were some issues in the creation of ConsensusDATA and in the process serialization and deserialization were resolved.
However, further testing of the Byzantine Consensus requires arbitrary failure, but it is not working here.
Can you give @AndrejMitrovic  your opinion or review it?

AndrejMitrovic:
If you can't get a test-case to work, then please file an issue for it so we can add it later.

AndrejMitrovic:
@Geod24 have you reviewed / approved this PR? If https://github.com/bpfkorea/agora/pull/577 is merged first, it's going to cause rebase conflicts here.

Geod24:
> @Geod24 have you reviewed / approved this PR?

Reviewed yes, but some comments haven't been addressed from what I can see.
I merged your PR and will try to extract the commits piecemeal (there are some of them which I'm not really sure about and just need to test).

Geod24:
Another concern is that many of the commits are not actually atomic from what I can see.

Geod24:
@TrustHenry : Can you extract https://github.com/bpfkorea/agora/pull/504/commits/bde77405ed2ef6ed005401cfc3b3ce3ce67212e2 into its own PR ? It also needs to be called (from L192 in the commit linked).

Geod24:
You can also probably isolate `Defines the data used when reaching consensus` in its own commit.
The tests are not great though: the `init` values are all 0, so they don't test much.
For the transaction set, try to use the genesis txs (see `Set.from`) and for the Enrollment, just add twice [this entry](https://github.com/bpfkorea/agora/blob/d22f394eb21d69f012db1a0c452140b45bd0dbb1/source/agora/consensus/data/Enrollment.d#L90-L104)

Geod24:
Needs a rebase too ;)

Geod24:
This needs a rebase. And perhaps `Reach consensus on transaction and enrollment set` can be split into its own PR.

Geod24:
Looks like you removed the consensus commit ?

TrustHenry:
Except for the part that was an issue here, only the part that registered data was registered on the block was left.
Related issues are #380 , #565 
I will make another PR for the Enrollment Consensus part.

TrustHenry:
Ready to review.

Geod24:
> node-2_1_6744a39c3d6b | 2020-03-04T03:15:17.120102785Z 2020-03-04 03:15:16,842 Error [agora.consensus.protocol.Nominator] - validateValue(): Invalid tx set: Transaction: Input ref not in UTXO

Geod24:
> I will make another PR for the Enrollment Consensus part.

Could you do it now ? That commit was ready, and it's much better to make consensus before writing to the block :)

TrustHenry:
> > I will make another PR for the Enrollment Consensus part.
> 
> Could you do it now ? That commit was ready, and it's much better to make consensus before writing to the block :)

I made PR #620

TrustHenry:
I am doing solving this conflict.

Geod24:
Ping me when you have rebased, let's get this merged!

TrustHenry:
@Geod24 ping

Geod24 (approved):

PR #506, opened 2020-01-30
[maybe] Add nothrow opApply to BitField

AndrejMitrovic:
@Geod24 actually can we just start using our own fork of Ocean? Since.. well.. everyone at Sociomantic is gone soon?

### DIFF
--- a/source/agora/common/BitField.d
+++ b/source/agora/common/BitField.d
@@ -119,6 +119,20 @@ public struct BitField
     {
         return this.num_bits;
     }
+
+    /// nothrow wrapper around BitArray.opApply (todo: fix BitArray)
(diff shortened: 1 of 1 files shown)


Geod24:
> @Geod24 actually can we just start using our own fork of Ocean? Since.. well.. everyone at Sociomantic is gone soon?

I mean, we are already.

Geod24:
@AndrejMitrovic : Why did you need `opApply` ? We have a clean-room implementation now so we can probably close this...

AndrejMitrovic:
Indeed, thanks for killing it.

AndrejMitrovic:
And, it was needed because of a lack of `@safe` and other things in BitArray.

PR #507, opened 2020-01-30
Implement a Schnorr-based signature validation routine 

AndrejMitrovic:
First part of #365

This includes example code on how signing works, and the signature validation routine.


AndrejMitrovic:
This code predates the preimage support code, and it unnecessarily hashed R as a challenge in my opinion. The code can still be looked at, but I'll close the PR as we won't merge this code as-is.

PR #511, opened 2020-01-31
Remove parsing the quorum configuration from the config file

Geod24:
```
Quorum configurations were never intended to be public.
Instead, they should be dynamically adjusted by the quorum balancing algorithm.
For the moment we still need to provide them explicitly to SCP,
as the algorithm is being developped.

As a concession to simplicity we devised 2 use cases:
- Providing the quorum sets to nodes in unittests, in order to test SCP;
- Using the set of peers as validators with a fixed threshold of 67%;
```

Required for #483

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -80,29 +80,6 @@ dns:
   - seed.bosagora.io
 
 ################################################################################
-##                               Quorum slices                                ##
-################################################################################
-quorum:
-  # threshold as a percentage
-  threshold: 66%
-  # the list of nodes in this quorum
-  nodes:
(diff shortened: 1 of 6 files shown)


AndrejMitrovic:
What was blocking this? Just some test-failure? It will need a rebase.

Geod24:
Oh I rebased it locally. It needs [this to be addressed](https://github.com/bpfkorea/agora/issues/587#issuecomment-609534422). Do you have a WIP for this ? Otherwise I might tackle it.

AndrejMitrovic:
I don't have any WIP yet.

AndrejMitrovic:
Will close as it's part of https://github.com/bpfkorea/agora/pull/845 now. Thanks for your work though.

PR #512, opened 2020-01-31
Split validator and full node APIs

### LINKED ISSUE #509: Remove integration test dependency to SCP
The system integration test is supposed to be a client test of a network.
Meaning it sends a request and checks the response.

However, currently, it has way too many dependencies. It needs to build SCP (the C++ side), and link in the C++ standard library: https://github.com/bpfkorea/agora/blob/dfdcab7e5999541ca5d62d262f13f9b0a386e2df/tests/system/dub.json#L5-L43

Things like `libsodium` are required as dependency (for signing), but we should not need to build scpp unless we build something that test the validator API (but really, we don't want to do that).

Geod24:
```
This allow us to reduce dependency and not import the SCP code in clients,
as can be seen from the change of the CLI app and the system test.
```

Note that this triggers https://github.com/vibe-d/vibe.d/issues/2414 but the testsuite still pass, so... Yay ?
Fixes #509

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -51,7 +51,15 @@
             "name": "cli",
             "targetName": "agora-cli",
             "mainSourceFile": "source/agora/cli/main.d",
-            "excludedSourceFiles": [ "source/agora/node/main.d" ]
+            "excludedSourceFiles": [
+                "source/agora/api/Validator.d",
+                "source/agora/common/Config.d",
+                "source/agora/network/*",
+                "source/agora/node/*",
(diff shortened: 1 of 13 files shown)


Geod24:
Okay this works now

AndrejMitrovic:
> Okay this works now

Great, I'll merge once Travis is done.

Geod24:
Or just leave a review and I'll merge :P

AndrejMitrovic (approved):
<3

PR #516, opened 2020-01-31
System test: Remove duplicated quorum configuration

Geod24:
Since all subquorums are equal to the top quorums,
they are actually useless because if they were to
turn 'true', the top quorum would as well.


Geod24:
This passes CircleCI

AndrejMitrovic (approved):

Issue #492, opened 2020-01-21
Implement broadcasting preimage


lunalee1012:
After registering as a validator in round R, a node must broadcast the preimage every round from round R + 1 based on the white paper. It ensures true randomness of the network without ability to manipulate data.

But, as described in the white paper, a node could regularly broadcasts multiple preimages in order to survive network outages. It might be efficient to broadcast multiple preimages periodically 
 at certain intervals instead of in every round. 

Definition of Done:
- [x] The enrollment manager saves all preimages while hashing random number n times before enrollment.

- [x] A node broadcasts preimages and the round #s of the preimages through the gossip protocol.

- [x] The enrollment manager saves the received preimages and the round numbers into the DB.

How to implement:
- The enrollment manager saves preimages into the DB in creating enrollment data and make registering(=enrolling) preceed with Q(=12) preimages.(The proper value of Q should be searched)

- An API point called `revealPreimages` in every P(=1) hours with Q preimages.(The proper value of P should be searched)

- The enrollment manager saves the preimages and round numbers of other validators into the DB.

- The enrollment manager checks the validity of preimages received through the API and set an flag for an invalid validator into the DB or propagate the information of the invalid validator (it could be  another new issue)

AndrejMitrovic:
Also, according to the Yellow paper, a node can reveal multiple preimages in advance. This is in order to survive network outages.

linked0:
Based on @AndrejMitrovic comment, I changed the API name to `revealPreimages`.

AndrejMitrovic:
> Also, according to the Yellow paper, a node can reveal multiple preimages in advance. This is in order to survive network outages.

Actually it doesn't have to reveal "multiple" ones. It just has to reveal the preimage for a specific block height.

For example, if a Node wishes to reveal preimages for the range of blocks #100 to #200, It could reveal the preimage for block height #200. Because it is a preimage, other nodes can figure out the preimage at block height #100 by hashing the preimage of block height #200 a 100 times.

Does it make sense to you?

linked0:
It really makes sense. Thank you.

PR #519, opened 2020-02-03
Add an pre-image information into the Enrollment structure

linked0:
A node should reveal an pre-image in advance for effienciency during enrolling as a validator. The pre-image information contains the pre-image as a certain round and the round number, such as: (P, Nx), where P is the pre-image and Nx is: n - (rounds since enrollment), which is described in [the white paper](https://bit.ly/2U6t5O2).

#492

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -35,6 +35,19 @@ import std.path;
 
 mixin AddLogger!();
 
+
+/*******************************************************************************
+
+    Define preimage data for a preimage at a certain round and the round number
+
+*******************************************************************************/
+
+public struct PreimageInfo
+{
+    public Hash hash;
+    public uint round;
+}
+
 /*******************************************************************************
 
     Handle enrollment data and manage the validators set
@@ -60,6 +73,13 @@ public class EnrollmentManager
 
     /// Enrollment data object
     private Enrollment data;
+
+    /// The round period for revealing a preimage
+    /// It is an hour interval if a block is made in every 10 minutes
+    public static immutable uint PreimagePeriod = 6;
+
+    /// The cycle length for a valdator
+    public static immutable uint ValidatorCycleLength = 1008; // freezing period / 2
 
     /***************************************************************************
 
@@ -318,7 +338,7 @@ public class EnrollmentManager
         this.data.utxo_key = frozen_utxo_hash;
 
         // N, cycle length
-        this.data.cycle_length = 1008; // freezing period / 2
+        this.data.cycle_length = ValidatorCycleLength;
 
         // generate random seed value
         this.random_seed_src = Scalar.random();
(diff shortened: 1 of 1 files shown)


linked0:
Ready for review

source/agora/consensus/EnrollmentManager.d
```diff
@@ -331,6 +335,10 @@ public class EnrollmentManager
             this.preimages ~= hashFull(this.preimages[i]);
         this.data.random_seed = this.preimages[$-1];
 
+        // Reveal preimage in advance
+        this.data.preimage.round = this.data.cycle_length - PreimagePeriod;
```
AndrejMitrovic: So the idea is, every 6 rounds the node will release a preimage for the next 6 rounds, right?

Therefore, this data needs to be updated. So shouldn't these two statements be inside of a function that can be called later to update and reveal the preimage?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -61,6 +61,10 @@ public class EnrollmentManager
     /// Enrollment data object
     private Enrollment data;
 
+    /// The round period for revealing a preimage
+    /// It is an hour interval if a block is made in every 10 minutes
+    public static immutable uint PreimagePeriod = 6;
```
Geod24, line 79: We should make this configurable.

source/agora/consensus/data/Enrollment.d
```diff
@@ -31,6 +31,9 @@ public struct Enrollment
     /// X: random seed, The nth image of random value
     public Hash random_seed;
 
+    /// Preimage information being revealed in advance
+    public PreimageInfo preimage;
```
Geod24: The `PreImageInfo` shouldn't be part of the enrollment, as the enrollment is written on the chain. It should be completely independent. A node can of course send a `PreImageInfo` along its enrollment (we can adapt the API to make this parameter optional), but it should also be able to receive, cache, and propagate PreImageInfo independently.

Geod24 (changes requested):

source/agora/consensus/EnrollmentManager.d
linked0: The data will be of course updated during revealing the pre-image after enrollment and the code will be added in another PR. Anyway, I try to apply what you said to the code by making another function for getting an pre-image of a certain round.

source/agora/consensus/EnrollmentManager.d
linked0: I will make another issue now for configuring the period of revealing pre-images. Before the new issue gets completed, we should use that constant value.

source/agora/consensus/data/Enrollment.d
linked0: The API endpoint named sendPreimage is planned to be made in another PR. And as you suggested, I will make the `enrollValidator` API function have the second parameter for a pre-image.

linked0:
I have decided to close this PR because it makes unnecessary confusion and the code is not so helpful and it might make the work time unnecessarily longer for the issue. My original intent was to reduce network transmittion by revealing a pre-image in advance during registering as a validator. If such an concept happen to be needed some time, I will apply it to the code. From now I will start implementing the API point called sendPreimage.

Issue #501, opened 2020-01-29
See if we can avoid code duplication for the findUTXO routine


AndrejMitrovic:
It's used frequently in many places:

https://github.com/bpfkorea/agora/blob/856ca4aba69effa09e7eb5121cb8ac338ccd3b61/source/agora/consensus/Validation.d#L174
https://github.com/bpfkorea/agora/blob/856ca4aba69effa09e7eb5121cb8ac338ccd3b61/source/agora/consensus/Validation.d#L219
https://github.com/bpfkorea/agora/blob/856ca4aba69effa09e7eb5121cb8ac338ccd3b61/source/agora/consensus/Validation.d#L276
https://github.com/bpfkorea/agora/blob/856ca4aba69effa09e7eb5121cb8ac338ccd3b61/source/agora/consensus/Validation.d#L345
https://github.com/bpfkorea/agora/blob/856ca4aba69effa09e7eb5121cb8ac338ccd3b61/source/agora/consensus/Validation.d#L510
https://github.com/bpfkorea/agora/blob/856ca4aba69effa09e7eb5121cb8ac338ccd3b61/source/agora/consensus/Validation.d#L666
https://github.com/bpfkorea/agora/blob/856ca4aba69effa09e7eb5121cb8ac338ccd3b61/source/agora/consensus/Validation.d#L722
https://github.com/bpfkorea/agora/blob/856ca4aba69effa09e7eb5121cb8ac338ccd3b61/source/agora/consensus/Validation.d#L873

And so on. There are some slight variations here and there.

PR #521, opened 2020-02-03
Fix a unit test code with an invalid utxoKey

TrustHenry:


source/agora/consensus/EnrollmentManager.d
```diff
@@ -586,7 +586,7 @@ unittest
 
         auto signature = gen_key_pair.secret.sign(hashFull(tx)[]);
         tx.inputs[0].signature = signature;
-        storage[hashFull(tx)] = tx;
+        storage[hashMulti(hashFull(tx), idx)] = tx;
```
Geod24, line 589: Another example why https://github.com/bpfkorea/agora/issues/501 would be great :)

Geod24: https://github.com/bpfkorea/agora/pull/523

AndrejMitrovic:
@TrustHenry #523 was merged, so I think we won't need this fix here anymore. Can we close it?

Thanks for your work though!

Issue #518, opened 2020-02-03
Quorum consensus isn't being properly reached


AndrejMitrovic:
https://github.com/bpfkorea/agora/blob/a99d96a5dff959fd4c32b9dc24a837d47bd76713/source/agora/test/Base.d#L642

This was setting the `threshold` to `1` because that's its default value, which means that a node only cared about its own opinion while reaching consensus. Adding additional logging helped confirm this.

Setting it to a proper value however causes consensus failure. It seems that when node A tries to communicate with the other nodes through SCP, the other nodes B, C never receive any messages at all. So it might be an issue with the networking.

I'm investigating.

AndrejMitrovic:
There were a multitude of issues, I'm resolving them all now.

PR #522, opened 2020-02-03
Fix SCP quorum reaching consensus

AndrejMitrovic:
There might still be some slight timing issues. If Travis/Github CI fails, I'll see about looking into this more.

One of the test-cases would fail only when the full test-suite ran, so it seems there have to be some delays between nomination rounds (after externalization). This is something I will look into more soon.

Partial fixes for https://github.com/bpfkorea/agora/issues/518

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -250,6 +250,20 @@ public class Ledger
 
     /***************************************************************************
 
+        Returns:
+            true if there is currently a nomination round in progress.
+            Blocks should not be added to the Ledger via getBlocksFrom()
+            when this node is currently in the process of nominating a block.
+
+    ***************************************************************************/
+
+    public bool isNominating () nothrow @safe pure @nogc
+    {
+        return this.is_nominating;
+    }
+
+    /***************************************************************************
+
         Try making a new block if there are enough valid and non double-spending
         transactions in the pool
 
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -261,6 +261,10 @@ public class Ledger
 
     public void tryNominateTXSet () @safe
     {
+        // not a validator, the node should not create blocks
+        if (!this.node_config.is_validator)
+            return;
+
         // if we received another transaction while we're nominating, don't nominate again.
         // todo: when we change nomination to be time-based (rather than input-based),
         // then remove this part as it will be handled by a timer
@@ -444,6 +448,7 @@ unittest
     auto utxo_set = new UTXOSet(":memory:");
     scope (exit) utxo_set.shutdown();
     auto config = new Config();
+    config.node.is_validator = true;
     scope ledger = new Ledger(pool, utxo_set, storage, config.node);
     assert(ledger.getBlockHeight() == 0);
 
@@ -529,6 +534,7 @@ unittest
     auto utxo_set = new UTXOSet(":memory:");
     scope (exit) utxo_set.shutdown();
     auto config = new Config();
+    config.node.is_validator = true;
     scope ledger = new Ledger(pool, utxo_set, storage, config.node);
 
     // Valid case
@@ -575,6 +581,7 @@ unittest
     auto utxo_set = new UTXOSet(":memory:");
     scope (exit) utxo_set.shutdown();
     auto config = new Config();
+    config.node.is_validator = true;
     scope ledger = new Ledger(pool, utxo_set, storage, config.node);
 
     Block invalid_block;  // default-initialized should be invalid
@@ -598,6 +605,7 @@ unittest
     auto utxo_set = new UTXOSet(":memory:");
     scope (exit) utxo_set.shutdown();
     auto config = new Config();
+    config.node.is_validator = true;
     scope ledger = new Ledger(pool, utxo_set, storage, config.node);
 
     auto gen_key_pair = getGenesisKeyPair();
@@ -675,6 +683,7 @@ unittest
     auto utxo_set = new UTXOSet(":memory:");
     scope (exit) utxo_set.shutdown();
     auto config = new Config();
+    config.node.is_validator = true;
     scope ledger = new Ledger(pool, utxo_set, storage, config.node);
 
     assert(utxo_set.length == 8);
@@ -779,6 +788,7 @@ unittest
     scope (exit) utxo_set.shutdown();
 
     auto config = new Config();
+    config.node.is_validator = true;
     scope ledger = new Ledger(pool, utxo_set, storage, config.node);
 
     KeyPair[] in_key_pairs;
@@ -876,6 +886,7 @@ unittest
     scope (exit) utxo_set.shutdown();
 
     auto config = new Config();
+    config.node.is_validator = true;
     scope ledger = new Ledger(pool, utxo_set, storage, config.node);
 
     KeyPair[] splited_keys = getRandomKeyPairs();
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -211,22 +211,6 @@ public class Ledger
 
     ***************************************************************************/
 
-    public void tryNominate ()
-    {
-        if (this.pool.length >= Block.TxsInBlock)
-            this.tryNominateTXSet();
-    }
-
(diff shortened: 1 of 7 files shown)


AndrejMitrovic:
Yeah it's already failing. But that's ok, it's likely a timing thing. I'll look into it.

source/agora/node/Ledger.d
```diff
@@ -299,10 +310,12 @@ public class Ledger
         if (txs.length != Block.TxsInBlock)
             return;  // not enough valid txs
 
+        this.is_nominating = true;
```
TrustHenry, line 313: `scope(exit) this.is_nominating = false;`

Maybe should use it.

AndrejMitrovic:
Ok let me try to split this up, maybe one of the commits triggers an invalid state. The errors seem like segfaults / assertion failures in unexpected places.

```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  invalid public key type
```

```
libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: moved to a bad state (ballot protocol)
Program exited with code -6
```

AndrejMitrovic:
@Geod24 I've had to temporarily remove the commit which sets the proper threshold to > 1 in the config, there is a failure on Linux platforms (but on the Mac it works fine).

With this, the PR is green and can be merged. In the meantime I'll continue debugging and get the Linux failure fixed.

source/agora/test/Ledger.d
```diff
@@ -96,14 +96,15 @@ unittest
...
+    auto network = makeTestNetwork(NetworkTopology.OneValidator, NodeCount, true,
         100, 20, 100);  // reduce timeout to 100 msecs
     network.start();
     scope(exit) network.shutdown();
     scope(failure) network.printLogs();
     assert(network.getDiscoveredNodes().length == NodeCount);
 
-    auto nodes = network.apis.values;
+    // node_1 is the validator
+    auto nodes = network.keys.map!(key => network.apis[key]).array;
```
Geod24, line 107: Slight optimization: Since you call `array` at the end, `byKeys` would be better to avoid double allocation (but it's unittest code so heh)

source/agora/node/Ledger.d
```diff
@@ -261,6 +261,10 @@ public class Ledger
 
     public void tryNominateTXSet () @safe
```
Geod24, line 276: Third commit ("Do not nominate blocks if we're a full node"): typo "agor atest-suite"

source/agora/test/Ledger.d
```diff
@@ -63,6 +64,7 @@ unittest
 
         block_txes ~= txs.sort.array;
         last_txs = txs;
+        Thread.sleep(500.msecs);
```
Geod24: Please add a comment as to why it's required

source/agora/test/BanManager.d
```diff
@@ -79,42 +80,42 @@ unittest
...
+    {
+        auto new_tx = genBlockTransactions(1);
+        left_txs ~= new_tx;
+        new_tx.each!(tx => node_1.putTransaction(tx));
+
+        [node_1, node_2].enumerate.each!((idx, node) =>
+            retryFor(node.getBlockHeight() == 1 + (block_idx + 1),
+                4.seconds,
+                format("Block %s Node %s has block height %s. Expected: %s",
+                    block_idx + 1, idx, node.getBlockHeight().to!string, 1 + (block_idx + 1))));
```
Geod24: Why the `to!string` here ?

source/agora/test/BanManager.d
```diff
@@ -79,42 +80,42 @@ unittest
...
+
+        retryFor(node_3.getBlockHeight() == 1, 1.seconds, node_3.getBlockHeight().to!string);
+    }
 
     // wait for node 3 to be banned and all putTransaction requests to time-out
     Thread.sleep(2.seconds);
```
Geod24, line 99: Is this still relevant ?

source/agora/test/BanManager.d
```diff
@@ -79,42 +80,42 @@ unittest
...
     new_tx.each!(tx => node_1.putTransaction(tx));
-    retryFor(node_1.getBlockHeight() == 7, 1.seconds);
-    retryFor(node_2.getBlockHeight() == 7, 1.seconds);
-    retryFor(node_3.getBlockHeight() == 7, 1.seconds);  // node was un-banned
+
+    nodes.enumerate.each!((idx, node) =>
+        retryFor(node.getBlockHeight() == 6,
+            4.seconds,
+            format("Node %s has block height %s. Expected: %s",
+                idx, node.getBlockHeight().to!string, 6)));
```
Geod24: Ditto `to!striing`

source/agora/test/BanManager.d
```diff
@@ -79,42 +80,42 @@ unittest
...
     // wait for node 3 to be banned and all putTransaction requests to time-out
     Thread.sleep(2.seconds);
 
-    import std.conv;
-    retryFor(node_1.getBlockHeight() == 5, 1.seconds, node_1.getBlockHeight().to!string);
-    retryFor(node_2.getBlockHeight() == 5, 1.seconds, node_2.getBlockHeight().to!string);
-    retryFor(node_3.getBlockHeight() == 1, 1.seconds, node_3.getBlockHeight().to!string);
-
-    // clear putTransaction filter
+    // sanity check: block height should not updated, node 3 is not a validator and cannot make blocks
```
Geod24: Maybe add your explanation that the Ledger can only be manipulated through `getBlockFrom` ?

source/agora/test/NetworkManager.d
```diff
@@ -122,37 +126,53 @@ unittest
...
+        auto txs = makeChainedTransactions(getGenesisKeyPair(), last_txs, 1);
+
+        // send it to one node
+        txs.each!(tx => node_validator.putTransaction(tx));
+
+        [node_validator].enumerate.each!((idx, node) =>
+            retryFor(node.getBlockHeight() == block_idx + 1,
+                4.seconds,
+                format("Node %s has block height %s. Expected: %s",
+                    idx, node.getBlockHeight().to!string, block_idx + 1)));
```
Geod24: `to!string`

source/agora/node/Ledger.d
```diff
@@ -297,11 +311,12 @@ public class Ledger
             return;  // not enough valid txs
 
         this.is_nominating = true;
+        scope (exit) is_nominating = false;
```
Geod24: This should be in an earlier commit

source/agora/test/Ledger.d
AndrejMitrovic: Might not be necessary anymore, I'll try to remove it.

source/agora/test/BanManager.d
AndrejMitrovic: It used to be in place of the `format` call, I'll fix it.

source/agora/test/Ledger.d
AndrejMitrovic: `.keys` here refers to the `keys` field inside of `TestAPIManager`, not the `.keys()` of a hashmap.

source/agora/test/BanManager.d
AndrejMitrovic: Yeah because it still takes actual real world time for a node to be banned.

I can see two ways of making this better:
- Use a virtual clock so we can "fast-forward" events. It's still tricky to write code for this. I've attempted this once.
- Add an event handling mechanism, so we can wait until some event fires. Something like this:

```D
class Events
{
    void onBannedNode (Node node) { ... }
    void onWhatever (...) { ... }
}
```

Then in the test-suite we could provide a subclass of `Events`, give that to the `Node`, and we could listen and wait for events. Instead of this `Thread.sleep(2.seconds);` call, we could just wait for the `onBannedNode` call.

AndrejMitrovic:
Fixed up.

source/agora/test/BanManager.d
Geod24: I wonder if anyone has implemented https://en.wikipedia.org/wiki/Server-sent_events in D

Geod24 (approved):
Very nice!

PR #524, opened 2020-02-05
SCP: Replace a global buffer with a thread-local

AndrejMitrovic:


source/scpp/src/crypto/ByteSliceHasher.cpp
```diff
@@ -9,7 +9,7 @@ namespace stellar
 {
 namespace shortHash
 {
-static unsigned char sKey[crypto_shorthash_KEYBYTES];
+thread_local unsigned char sKey[crypto_shorthash_KEYBYTES];
```
Geod24, line 12: Is the initialize function ever called ?

AndrejMitrovic: It looks like it's not. It seems odd it was initialized to a random value anyway..

Geod24:
Superseeded by #526 , thanks for finding this!

PR #525, opened 2020-02-05
Fix a potential timing issue with timers in setupTimer()

AndrejMitrovic:
I noticed this piece of code while debugging. 

We don't have a proper Timer implementation yet, so we usually use:

```
runTask(() { sleep(timeout); ... }
```

But in the current code there was an issue with how older timers got deactivated. It has not triggered so far, but I'm pretty sure it could.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -399,29 +399,29 @@ extern(D):
 
         Params:
             slot_idx = the slot index we're currently reaching consensus for.
-            timer_ID = the timer ID. required in case the timer gets cancelled.
+            timer_type = the timer type (see Slot.timerIDs).
             timeout = the timeout of the timer, in milliseconds.
             callback = the C++ callback to call.
 
     ***************************************************************************/
 
-    public override void setupTimer (ulong slot_idx, int timer_ID,
+    public override void setupTimer (ulong slot_idx, int timer_type,
         milliseconds timeout, CPPDelegate!SCPCallback* callback)
     {
         scope (failure) assert(0);
 
         if (callback is null || timeout == 0)
         {
-            this.timers.remove(timer_ID);
+            this.timers.remove(timer_type);
             return;
         }
 
-        this.timers.put(timer_ID);
+        this.timers.put(timer_type);
 
         this.taskman.runTask(
         {
             this.taskman.wait(timeout.msecs);
-            if (timer_ID !in this.timers)  // timer was cancelled
+            if (timer_type !in this.timers)  // timer was cancelled
                 return;
 
             callCPPDelegate(callback);
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -29,6 +29,7 @@ import agora.utils.PrettyPrinter;
 import scpd.Cpp;
 import scpd.scp.SCP;
 import scpd.scp.SCPDriver;
+import scpd.scp.Slot;
 import scpd.scp.Utils;
 import scpd.types.Stellar_types;
 import scpd.types.Stellar_types : StellarHash = Hash;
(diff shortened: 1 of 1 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -399,29 +399,29 @@ extern(D):
 
         Params:
             slot_idx = the slot index we're currently reaching consensus for.
-            timer_ID = the timer ID. required in case the timer gets cancelled.
+            timer_type = the timer type (see Slot.timerIDs).
```
Geod24, line 412: Uuuurgh, so there's only 2 timers ? Why didn't just make 2 functions ...

AndrejMitrovic: I guess maybe they plan on making more?

Can we just delete SCP?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -70,6 +74,12 @@ public extern (C++) class Nominator : SCPDriver
     /// The quorum set
     private SCPQuorumSetPtr[Hash] quorum_set;
 
+    /// Tracks unique incremental timer IDs
+    private ulong[TimerType] last_timer_id;
```
Geod24: That's an AA, no ?

```suggestion
    private ulong[TimerType.max + 1] last_timer_id;
```

To make it a static array ? ~I really wish DMD recognized this pattern.~

source/agora/consensus/protocol/Nominator.d
Geod24: > Can we just delete SCP?

I mean... Now that we got most of the D compiler bugs fixed, it would be a shame.

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: Good idea.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -29,15 +29,19 @@ import agora.utils.PrettyPrinter;
...
 import scpd.scp.SCPDriver;
+import scpd.scp.Slot;
 import scpd.scp.Utils;
 import scpd.types.Stellar_types;
 import scpd.types.Stellar_types : StellarHash = Hash;
 import scpd.types.Stellar_SCP;
 import scpd.types.Utils;
 import scpd.Util;
 
+alias TimerType = Slot.timerIDs;
```
Geod24: I  would move it down with other declarations (after the `mixin`) and make it `private`.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -89,6 +99,12 @@ extern(D):
         TaskManager taskman, NetworkClient[PublicKey] peers,
         SCPQuorumSet quorum_set)
     {
+        foreach (type; EnumMembers!TimerType)
```
Geod24: With the static array this can go

AndrejMitrovic:
Pushed.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -399,7 +411,7 @@ extern(D):
 
         Params:
             slot_idx = the slot index we're currently reaching consensus for.
-            timer_type = the timer type (see Slot.timerIDs).
+            timer_type = the timer ID. required in case the timer gets cancelled.
```
Geod24: Looks like you disagreed with yourself

source/agora/consensus/protocol/Nominator.d
```diff
@@ -410,18 +422,21 @@ extern(D):
     {
         scope (failure) assert(0);
 
+        const type = timer_type.to!TimerType;
```
```D
const type = cast(TimerType) timer_type;
assert(type >= TimerType.min && type <= TimerType.max);
```

Allows to remove the two imports (traits, conv) also.

AndrejMitrovic:
Thanks.

Geod24 (approved):

AndrejMitrovic:
It was approved, merged it.

PR #526, opened 2020-02-05
SCP: Remove ByteSliceHasher

Geod24:
It was introduced upstream in order to provide functionalities equivalent to D's toHash,
that is, allow ByteSlice to be used in an std::unordered_map.
This usage is in Stellar's Ledger, which we do not use.


Geod24:
Superseeds #524

AndrejMitrovic (approved):

PR #530, opened 2020-02-07
Implement broadcasting a pre-image

linked0:
This PR is about broadcasting a pre-image and saving it into the validator set table. 

- sendPreimage endpoint in Validator API

#492

### DIFF
--- a/source/agora/api/Validator.d
+++ b/source/agora/api/Validator.d
@@ -43,6 +43,8 @@
 module agora.api.Validator;
 
 import agora.common.crypto.Key;
+import agora.common.Hash;
+import agora.consensus.data.PreimageInfo;
 static import agora.api.FullNode;
 
 import scpd.types.Stellar_SCP;
@@ -93,4 +95,31 @@ public interface API : agora.api.FullNode.API
     ***************************************************************************/
 
     public bool receiveEnvelope (SCPEnvelope envelope);
+
+    /***************************************************************************
+
+        Reveals a pre-image
+
+        Params:
+            preimage = a PreimageInfo object which contains a hash and a height
+
+    ***************************************************************************/
+
+    public void receivePreimage (PreimageInfo preimage);
+
+    /***************************************************************************
+
+        Checks if a pre-image exists
+
+        Params:
+            enroll_key = The key for the enrollment in which the pre-image is
+                contained.
+            height = The block height of the preimage to check existence
+
+        Returns:
+            true if the pre-image exists
+
+    ***************************************************************************/
+
+    public bool hasPreimage (Hash enroll_key, ulong height);
 }
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -33,6 +33,7 @@ import d2sqlite3.sqlite3;
 import std.algorithm;
 import std.file;
 import std.path;
+import std.string;
 
 mixin AddLogger!();
 
@@ -85,7 +86,8 @@ public class EnrollmentManager
 
         // create the table for validator set if it doesn't exist yet
         this.db.execute("CREATE TABLE IF NOT EXISTS validator_set " ~
-            "(key TEXT PRIMARY KEY, val BLOB NOT NULL, enrolled_height INTEGER)");
+            "(key TEXT PRIMARY KEY, val BLOB NOT NULL, " ~
+            "enrolled_height INTEGER, preimage BLOB)");
 
         // create the table for enrollment data for a node itself
         this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
@@ -570,6 +572,135 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Check if a pre-image exists
+
+        Params:
+            enroll_key = The key for the enrollment in which the pre-image is
+                contained.
+            height = The block height of the preimage to check existence
+
+        Returns:
+            true if the pre-image exists
+
+    ***************************************************************************/
+
+    public bool hasPreimage (const ref Hash enroll_key, ulong height) @safe
+        nothrow
+    {
+        bool result = false;
+        try
+        {
+            () @trusted {
+                auto results = this.db.execute("SELECT preimage from validator_set" ~
+                            " WHERE key = ?", enroll_key.toString());
+                if (!results.empty && results.oneValue!(byte[]).length != 0)
+                {
+                    PreimageInfo loaded_image =
+                        results.oneValue!(ubyte[]).deserializeFull!(PreimageInfo);
+                    if (height <= loaded_image.height)
+                        result = true;
+                }
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Exception occured, hasPreimage:{}, exception:{}",
+                enroll_key, ex);
+            return false;
+        }
+        return result;
+    }
+
+    /***************************************************************************
+
+        Add a pre-image information to a validator data
+
+        Params:
+            preimage = the pre-image information to add
+
+        Returns:
+            true if the pre-image information has been added to the validator
+
+    ***************************************************************************/
+
+    public bool addPreimage (const ref PreimageInfo preimage) @safe nothrow
+    {
+        static ubyte[] buffer;
+        buffer.length = 0;
+
+        // check if the enrollment data exists
+        Enrollment stored_enroll;
+        try
+        {
+            if (!this.getEnrollment(preimage.enroll_key, stored_enroll))
+            {
+                log.info("Rejected adding a pre-image for non-existing" ~
+                    "enrollment, preimage:{}", preimage);
+                return false;
+            }
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error, addPreimage:{}, exception:{}",
+                preimage, ex);
+            return false;
+        }
+
+        // check if already exists
+        try
+        {
+            if (this.hasPreimage(preimage.enroll_key, preimage.height))
+            {
+                log.info("Rejected already existing pre-image, preimage:{}",
+                    preimage);
+                return false;
+            }
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation errort, addPreimage:{}, exception:{}",
+                preimage, ex);
+            return false;
+        }
+
+        // insert the pre-image into the table
+        () @trusted { assumeSafeAppend(buffer); } ();
+
+        scope SerializeDg dg = (scope const(ubyte[]) data) nothrow @safe
+        {
+            buffer ~= data;
+        };
+
+        try
+        {
+            serializePart(preimage, dg);
+        }
+        catch (Exception ex)
+        {
+            log.error("Serialization error, addPreimage:{}, exception:{}",
+                preimage, ex);
+            return false;
+        }
+
+        try
+        {
+            () @trusted {
+                this.db.execute("UPDATE validator_set SET preimage = ? " ~
+                    "WHERE key = ?", buffer, preimage.enroll_key.toString());
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error, addPreimage:{}, exception:{}",
+                preimage, ex);
+            return false;
+        }
+
+        return true;
+    }
+
+    /***************************************************************************
+
         Logs message
 
         Params:
@@ -754,4 +885,48 @@ unittest
     assert(man.getPreimage(11, preimage));
     assert(man.getPreimage(10 + EnrollmentManager.ValidatorCycle, preimage));
     assert(!man.getPreimage(11 + EnrollmentManager.ValidatorCycle, preimage));
+}
+
+/// tests for addPreimage and hasPreimage
+unittest
+{
+    import agora.common.Amount;
+    import agora.consensus.data.Transaction;
+    import agora.consensus.Genesis;
+    import std.conv;
+
+    scope storage = new TestUTXOSet;
+    auto gen_key_pair = getGenesisKeyPair();
+    KeyPair key_pair = KeyPair.random();
+
+    foreach (idx; 0 .. 8)
+    {
+        auto input = Input(hashFull(GenesisTransaction), idx.to!uint);
+
+        Transaction tx =
+        {
+            TxType.Freeze,
+            [input],
+            [Output(Amount.MinFreezeAmount, key_pair.address)]
+        };
+
+        auto signature = gen_key_pair.secret.sign(hashFull(tx)[]);
+        tx.inputs[0].signature = signature;
+        storage.put(tx);
+    }
+
+    auto man = new EnrollmentManager(":memory:", key_pair);
+    scope (exit) man.shutdown();
+    Hash[] utxo_hashes = storage.keys;
+
+    auto utxo_hash = utxo_hashes[0];
+    Enrollment enroll;
+    assert(man.createEnrollment(utxo_hash, enroll));
+    assert(man.addEnrollment(0, &storage.findUTXO, enroll));
+    assert(man.hasEnrollment(utxo_hash));
+
+    auto preimage = PreimageInfo(utxo_hash, enroll.random_seed, 1);
+    assert(man.addPreimage(preimage));
+    assert(man.hasPreimage(utxo_hash, 1));
+    assert(!man.hasPreimage(utxo_hash, 10));
 }
--- a/source/agora/network/NetworkClient.d
+++ b/source/agora/network/NetworkClient.d
@@ -17,6 +17,7 @@ import agora.api.Validator;
 import agora.common.BanManager;
 import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
+import agora.consensus.data.PreimageInfo;
 import agora.common.crypto.Key;
 import agora.common.Types;
 import agora.common.Set;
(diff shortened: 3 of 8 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -35,6 +35,19 @@ import std.path;
...
+
+    Define preimage data for a preimage at a certain round and the round number
+
+*******************************************************************************/
+
+public struct PreimageInfo
+{
+    public Hash utxo_key;
+    public Hash hash;
+    public uint round;
```
Geod24: That's an interesting approach. When we were originally talking about it, I had in mind that this would by *block height*, not round. Because the issue is that if you re-use the same `utxo_key` (as you would), then you can do replay attacks.
Of course validation would fail, but you could potentially pretend a node sent you the wrong preimage data. We didn't factor any slashing rules that could be triggered by this tho.

A downside of using the `block_height` is that it's harder to pre-emptively broadcast preimage as you don't know whe height at which it'll be included. But that can be dealt with when the enrollment is received.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -670,4 +708,10 @@ unittest
 
     // test serialization/deserializetion for pre-images
     assert(man.preimages[] == man.loadPreimages());
+
+    // get preimage at a certain round
+    PreimageInfo preimage;
+    auto round = 1000;
+    assert(man.getPreimage(1000, preimage));
+    assert(preimage.hash == man.preimages[round-1]);
```
Geod24: Should also test with `0` and > `ValidatorCycleLength`

source/agora/consensus/EnrollmentManager.d
Geod24: In any case, this could use some documentation

source/agora/consensus/Validation.d
```diff
@@ -1400,3 +1446,26 @@ unittest
     // Block: The enrollments are not sorted in ascending order
     assert(!block3.isValid(block2.header.height, hashFull(block2.header), findUTXO));
 }
+
+///
+unittest
+{
+    uint CycleLength = 1008;
```
Geod24: Use `const` if it isn't going to be modified

source/agora/consensus/Validation.d
```diff
@@ -948,6 +948,44 @@ public string isInvalidEnrollmentReason (const ref Enrollment enrollment,
...
+        cycle_length = The cycle length of a validation
+
+    Returns:
+        `null` if the pre-image is valid, otherwise a string
+        explaining the reason it is invalid.
+
+*******************************************************************************/
+
+public string isInvalidPreimageReason (const ref Hash preimage, const uint round,
+    const ref Hash init_seed, immutable uint cycle_length) nothrow @safe
```
Geod24: For value type, `const` / `immutable` does not make sense. So just use `uint`.

source/agora/consensus/Validation.d
```diff
@@ -948,6 +948,44 @@ public string isInvalidEnrollmentReason (const ref Enrollment enrollment,
...
+{
+    // check if the round number is equal to or less than the cycle length
+    if (round > cycle_length)
+        return "The round number exceeds over the cycle length of validation";
+
+    // check the pre-image has a right hash value
+    Hash temp_hash = preimage;
+    for (int i = round; i < cycle_length; i++)
```
```suggestion
    for (uint i = round; i < cycle_length; i++)
```

source/agora/consensus/Validation.d
Geod24: Also, since you made it a constant in an earlier commit, can we just use it ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -522,18 +523,164 @@ public class EnrollmentManager
...
-    public bool getPreimage (uint round, out PreimageInfo preimage)
+    public bool getPreimage (uint round, ref PreimageInfo preimage)
         @trusted nothrow
     {
-        if (round < 1 && round > ValidatorCycleLength)
+        if (round < 1 || round > ValidatorCycleLength)
             return false;
 
-        preimage.utxo_key = this.data.utxo_key;
+        preimage.enroll_key = this.data.utxo_key;
```
Geod24: Those fixes should be in a previous commit

source/agora/consensus/EnrollmentManager.d
```diff
@@ -522,18 +523,164 @@ public class EnrollmentManager
...
+            PreimageInfo loaded_image =
+                results.oneValue!(ubyte[]).deserializeFull!(PreimageInfo);
+            if (loaded_image.round == preimage.round)
+                return true;
+        }
+        return false;
+    }
+
+    /***************************************************************************
+        Add a pre-image information to a validator data
```
Geod24, line 616:
```suggestion

        Add a pre-image information to a validator data
```

source/agora/network/NetworkClient.d
```diff
@@ -297,4 +298,24 @@ class NetworkClient
...
+
+        The request is retried up to 'this.max_retries',
+        any failures are logged and ignored.
+
+        Params:
+            preimage = the pre-image information to send
+
+    ***************************************************************************/
+
+    public void sendPreimage (PreimageInfo preimage) @trusted
```
Geod24:
```suggestion
    public void sendPreimage (ref PreimageInfo preimage) @trusted
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -35,6 +35,19 @@ import std.path;
 
 mixin AddLogger!();
 
+/*******************************************************************************
+
+    Define preimage data for a preimage at a certain round and the round number
+
+*******************************************************************************/
+
+public struct PreimageInfo
```
Geod24: This should probably be in `consensus.data`

source/agora/test/Base.d
```diff
@@ -433,6 +433,12 @@ public interface TestAPI : API
 
     ///
     public abstract Enrollment createEnrollmentData();
+
+    ///
+    public abstract PreimageInfo getNodePreimage(uint round);
```
```suggestion
    public abstract PreimageInfo getNodePreimage (uint round);
```

source/agora/api/Validator.d
```diff
@@ -93,4 +94,16 @@ public interface API : agora.api.FullNode.API
...
+
+        Reveal a pre-image
+
+        Params:
+            preimage = a PreimageInfo object which contains a hash and round
+                number
+
+    ***************************************************************************/
+
+    public void sendPreimage (PreimageInfo preimage);
```
Geod24: Naming wise, we have `receiveEnvelope` but `sendPreimage`. I don't mind either way but we should be consistent.

source/agora/test/EnrollmentManager.d
```diff
@@ -75,4 +75,12 @@ unittest
...
     nodes.each!(node =>
         retryFor(node.hasEnrollment(enroll.utxo_key) == true, 5.seconds));
+
+    // tests for revealing a preimage
+    auto preimage_1 = node_1.getNodePreimage(10000);
+    assert(preimage_1 == PreimageInfo.init);
+    preimage_1 = node_1.getNodePreimage(1000);
+    node_1.sendPreimage(preimage_1);
+    auto preimage_2 = node_2.getValidatorPreimage(preimage_1.enroll_key);
+    assert(preimage_1 == preimage_2);
```
Geod24: This is not black box testing. Adding 2 special methods means you aren't really testing the API.
A better test would be to send a pre-image and check if the (X) next block contains it.
Additionally you could filter the `sendPreimage` function of a block to see if it propagates correctly even in the absence of the preimage in one block (but that depends on Henry's PR)/

Geod24 (changes requested):

source/agora/consensus/EnrollmentManager.d
linked0: I'd like accept your thought because it's more convenient to recognize the block to which the preimage is related. So, I could add the variable named `start_block_height` into the Enrollment structure which is the block height where a validator is supposed to participate in validating. Then, The variable, `round` of the PreimageInfo structure could be replaced with a variable like `round_block_height`. How about this?

source/agora/api/Validator.d
linked0: Do you mean changing the function name into a name like receivePreimage? If you do, changing the name is not a problem for me. I mean I can change the neme.

source/agora/consensus/EnrollmentManager.d
Geod24: > start_block_height

Minor stylistic note: "height" is pretty unambiguous in blockchain code, so you can just use `start_height` (because there's no other height that block height).

> which is the block height where a validator is supposed to participate in validating

I'll have to double check, the exact flow, but bear in mind that if we go this way, it is an *implicit* value. What I mean is that, when you create an `Enrollment` structure and sign it, you cannot know which `start_height` you will be assigned, because it has to be accepted by the validators. Same goes for transactions, when you create a transaction, you cannot know at which height it will be.

Does that make sense ? Also, I think both approaches have their merits, so maybe let's have a whiteboarding session about it ?

linked0: I understand what you're considering. After a bit of thought, I can use the `enrolled_height` variable which is implemented by Henry. The value of `enrolled_height + 1` will have the same value as what I want to have with the `start_block_height` variable. It's good to have a whiteboarding session.

source/agora/test/EnrollmentManager.d
linked0: My original approach is not to save pre-images into a block. It's stored only in the validator set table. A pre-image can be restored from another nodes if the database has been crashed. In that case, another API endpoint like getPreimage is needed with the getNodePreimage and getValidatorPreimage removed. Anyway the two functions could be confusing.

linked0: There is another approach about saving pre-images. It's to store the source value of the random seed encrypted with a node's private key into the Enrollment structure. With that, if the database crashed, the source value of the random seed used for enrollment can be restored from the chain.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -496,6 +497,31 @@ public class EnrollmentManager
...
+            round = the number of a round at which the pre-image is in
+            preimage = will contain the PreimageInfo if exists
+
+        Returns:
+            true if the pre-image exists
+
+    ***************************************************************************/
+
+    public bool getPreimage (uint round, out PreimageInfo preimage)
+        @trusted nothrow
```
Geod24: Why `@trusted` ?

linked0:
I will make a smaller PR including two commits of this PR and apply reviewer's suggestion in the new PR. And then, I will handle with the remaining commits of this PR which is not inluded in the new PR.

source/agora/api/Validator.d
```diff
@@ -43,6 +43,8 @@
 module agora.api.Validator;
 
 import agora.common.crypto.Key;
+import agora.common.Hash;
```
Geod24, line 46:
```suggestion
import agora.common.Types;
```

linked0:
Ready for review.
In order to make smaller PRs, I am splitting this PR to three or four PRs inclusive of this PR. So In this PR, I only put the code about the sendPreimage endpoint in Validator API.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -755,3 +870,47 @@ unittest
...
+    Enrollment enroll;
+    assert(man.createEnrollment(utxo_hash, enroll));
+    assert(man.addEnrollment(0, &storage.findUTXO, enroll));
+    assert(man.hasEnrollment(utxo_hash));
+
+    auto preimage = PreimageInfo(utxo_hash, enroll.random_seed, 1);
+    assert(man.addPreimage(preimage));
+    assert(man.hasPreimage(utxo_hash, 1));
+    assert(!man.hasPreimage(utxo_hash, 10));
+}
```
TrustHenry: `}
`
Style guide last line is required.

TrustHenry (dismissed):

source/agora/consensus/EnrollmentManager.d
```diff
@@ -568,6 +569,120 @@ public class EnrollmentManager
...
+
+        Returns:
+            true if the pre-image exists
+
+    ***************************************************************************/
+
+    public bool hasPreimage (const ref Hash enroll_key, ulong height) @trusted
+    {
+        auto results = this.db.execute("SELECT preimage from validator_set" ~
+                    " WHERE key = ?", enroll_key.toString());
```
AndrejMitrovic: The `toString()` call is allocating. I think something like this could work:

```D
static char[] cmd;
cmd.length = 0;
assumeSafeAppend(cmd);
sformat(cmd, "SELECT preimage from validator_set WHERE key = %s", enroll_key);
auto results = this.db.execute(cmd);
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -568,6 +569,120 @@ public class EnrollmentManager
...
+        if (!results.empty && results.oneValue!(byte[]).length != 0)
```
AndrejMitrovic: If the result is not empty but the array length is zero, doesn't that mean there is a bug?

I think this should be turned into an assert.

```D
auto value = results.oneValue!(byte[]);
assert(value.length != 0);
auto loaded_image = value.deseralizeFull!PreimageInfo;
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -568,6 +569,120 @@ public class EnrollmentManager
...
+            preimage = the pre-image information to add
+
+        Returns:
+            true if the pre-image information has been added to the validator
+
+    ***************************************************************************/
+
+    public bool addPreimage (const ref PreimageInfo preimage) @safe nothrow
+    {
+        static ubyte[] buffer;
```
AndrejMitrovic, line 628: This needs to be moved right before the `buffer.length = 0;` statement. Declarations should be placed close to where they're used the first time, which is not necessarily at the start of the function.

AndrejMitrovic (changes requested):
Not done reviewing, github won't let me continue because of a push.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -568,6 +569,120 @@ public class EnrollmentManager
...
+            buffer ~= data;
+        };
+
+        try
+        {
+            serializePart(preimage, dg);
+        }
+        catch (Exception ex)
+        {
+            log.error("Serializtion error, addPreimage:{}, exception:{}",
```
AndrejMitrovic: typo: serializtion

source/agora/consensus/EnrollmentManager.d
```diff
@@ -568,6 +569,120 @@ public class EnrollmentManager
...
+                preimage, ex);
+            return false;
+        }
+
+        try
+        {
+            () @trusted {
+                this.db.execute("UPDATE validator_set SET preimage = ? " ~
+                    "WHERE key = ?", buffer, preimage.enroll_key.toString());
```
AndrejMitrovic, line 689: Same comment about memory allocation, `toString()` allocates. Can use `sformat` to write the entire string to a buffer before calling `execute`.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -568,6 +569,120 @@ public class EnrollmentManager
...
+            enroll_key = The key for the enrollment in which the pre-image is
+                contained.
+            height = The block height of the preimage to check existence
+
+        Returns:
+            true if the pre-image exists
+
+    ***************************************************************************/
+
+    public bool hasPreimage (const ref Hash enroll_key, ulong height) @trusted
```
AndrejMitrovic: Ok, so when calling `addPreimage` the function is `nothrow` and can fail so it returns a status code. But if `hasPreimage` fails it will throw an exception. This is an inconsistent API. `hasPreimage` should be `nothrow` too.

AndrejMitrovic (changes requested):
Reviewed first commit.

source/agora/network/NetworkClient.d
```diff
@@ -297,4 +298,24 @@ class NetworkClient
         });
     }
 
+    /***************************************************************************
+
+        Send a preimage asynchronously to the node.
+        Any errors are reporteed to the debugging log.
```
AndrejMitrovic:
```suggestion
        Any errors are reported to the debugging log.
```

source/agora/node/GossipProtocol.d
```diff
@@ -117,4 +118,22 @@ public class GossipProtocol
             this.network.sendEnrollment(enroll);
         }
     }
+
+    /***************************************************************************
+
+        If there is not the same pre-image on the enrollment manager,
```
AndrejMitrovic:
```suggestion
        If this is a new enrollment, add it to the enrollment manager and propagate it to the network.
```

source/agora/node/GossipProtocol.d
```diff
@@ -117,4 +118,22 @@ public class GossipProtocol
...
+        propagate it to the network.
+
+        Params:
+            enroll = the pre-image information to propagate
```
AndrejMitrovic:
```suggestion
            enroll = the pre-image information to add & propagate
```

AndrejMitrovic (changes requested):

source/agora/test/Base.d
```diff
@@ -515,6 +519,14 @@ public class TestNode : Node, TestAPI
 
         return enroll;
     }
+
+    /// Get a node's own pre-image information used when revealing it
+    public override PreimageInfo getPreimage(uint height)
```
```suggestion
    public override PreimageInfo getPreimage (uint height)
```

source/agora/test/EnrollmentManager.d
```diff
@@ -74,4 +75,12 @@ unittest
     // check if nodes contains enrollment data previously sent.
     nodes.each!(node =>
         retryFor(node.hasEnrollment(enroll.utxo_key) == true, 5.seconds));
+
+    // tests for revealing a pre-image
+    auto preimage = node_1.getPreimage(1000);
+    node_1.receivePreimage(preimage);
```
AndrejMitrovic, line 81: This node should already know about its own preimages. It is odd that it would need to send the preimage to itself in order to know about it, no?

AndrejMitrovic (dismissed):

AndrejMitrovic:
Done with review.

Geod24 (approved):
LGTM, let's move forward with this.

PR #531, opened 2020-02-11
502 with attempted fix

AndrejMitrovic:
Opening this for the CI only.


AndrejMitrovic:
Ok that went nowhere. Back to the drawing board.

PR #533, opened 2020-02-11
Update CI DMD to v2.090.1

Geod24:


AndrejMitrovic:
What about `.circleci/config.yml`? It's still using the old compiler.

Geod24:
Right, fixed

Geod24:
Updated to also include an LDC upgrade

AndrejMitrovic (approved):

PR #535, opened 2020-02-12
Update documentation for getEnrollmentLength()

AndrejMitrovic:
It was slightly misleading. Also added a test-case.


AndrejMitrovic:
@Geod24 I wonder what documentation falls under.. `type-refactoring`? We could use a `type-docs` in fact..

Geod24 (approved):

Geod24:
I'd just go for "enhancement" as the generic kind. We could make another label, but ATM we don't use them so much so I didn't bother.

PR #539, opened 2020-02-13
Refactor deserialization to be more compact and a single function

Geod24:
Also add a few features, such as full support for range, support for bool, validation for strings... See each commits.

### DIFF
--- a/source/agora/common/Deserializer.d
+++ b/source/agora/common/Deserializer.d
@@ -15,6 +15,7 @@ module agora.common.Deserializer;
 
 import agora.common.Types;
 import agora.common.crypto.Key;
+import std.range.primitives;
 
 import std.range;
 import std.traits;
@@ -152,6 +153,16 @@ public T deserializeFull (T) (scope DeserializeDg dg) @safe
 
 /// Ditto
 public void deserializePart (T) (ref T record, scope DeserializeDg dg) @safe
+    if (isInputRange!T && hasLength!T)
+{
+    size_t length = deserializeVarInt!size_t(dg);
+    record.length = length;
+    foreach (ref v; record)
+        deserializePart(v, dg);
+}
+
+/// Ditto
+public void deserializePart (T) (ref T record, scope DeserializeDg dg) @safe
     if (is(T == struct))
 {
     import geod24.bitblob;
@@ -235,17 +246,6 @@ public void deserializePart (ref uint[] record, scope DeserializeDg dg)
         foreach (ref uint elem; record)
             elem = swapEndian(elem);
     }
-}
-
-/// Ditto
-public void deserializePart (ref Hash[] record, scope DeserializeDg dg)
-    @trusted
-{
-    size_t length;
-    deserializePart(length, dg);
-    record = uninitializedArray!(Hash[])(length);
-    foreach (ref val; record)
-        deserializePart(val, dg);
 }
 
 /*******************************************************************************
--- a/source/agora/common/Deserializer.d
+++ b/source/agora/common/Deserializer.d
@@ -168,6 +168,28 @@ public void deserializePart (T) (
     else static if (is(T : BitBlob!N, size_t N))
         record = T(dg(T.Width));
 
+    // Array deserialization can be optimized in many occasions
+    else static if (is(T : E[], E) && !hasIndirections!E)
+    {
+        size_t length = deserializeVarInt!size_t(dg);
+        record = () @trusted { return cast(T) (dg(E.sizeof * length).dup); }();
+
+        // Special case for narrow string
+        debug static if (isNarrowString!T)
+        {
+            import std.utf;
+            record.validate();
+        }
+
+        // serialized in little-endian format
+        version (BigEndian)
+        {
+            static if (E.sizeof > 1)
+                foreach (ref elem; record)
+                    elem = swapEndian(elem);
+        }
+    }
+
     // Range deserialization
     else static if (isInputRange!T && hasLength!T)
     {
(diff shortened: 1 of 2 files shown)


source/agora/common/Deserializer.d
```diff
@@ -84,6 +84,10 @@ unittest
 /// Type of delegate deserializeDg
 public alias DeserializeDg = ubyte[] delegate(size_t size) @safe;
 
+/// Traits to check if a given type has an in place deserialization routine
+private enum hasDeserializeMethod (T) = is(T == struct)
```
AndrejMitrovic: @bpfkorea/core in case you are confused about this weird looking declaration, this is just a shorter way of writing a full template declaration like this:

```D
template hasDeserializeMethod(T)
{
    enum hasDeserializeMethod = is(T == struct) &&
        is(typeof(T.init.deserialize(DeserializeDg.init)));
}
```

See https://dlang.org/spec/template.html#implicit_template_properties for more info.

source/agora/common/Deserializer.d
```diff
@@ -280,18 +275,24 @@ public void deserializePart (ref Hash[] record, scope DeserializeDg dg)
...
         T = Type of unsigned integer to deserialize
-        num = Value to store the result of deserialization into
         dg = source of binary data
 
+    Returns:
+      The deserialized value, typed as `T`
+
+    Throws:
+      If the deserialized value does not fit into a `T`.
+      Not that for `ulong`, this function is `nothrow`.
```
AndrejMitrovic:
```suggestion
      Note that for `ulong`, this function is `nothrow`.
```

source/agora/common/Deserializer.d
```diff
@@ -168,48 +169,30 @@ public void deserializePart (T)(ref T record, scope DeserializeDg dg)
...
-    record = *cast(long*)(dg(record.sizeof).ptr);
-}
-
-/// Ditto
-public void deserializePart (ref ulong record, scope DeserializeDg dg, CompactMode compact = CompactMode.Yes)
-    @trusted
-{
-    if (compact == CompactMode.Yes)
-        record = deserializeVarInt!ulong(dg);
+    static if (is(Unqual!T == bool))
```
AndrejMitrovic: Was `bool` not supported before? Might be good to add a test.

source/agora/common/Serializer.d
```diff
@@ -67,18 +68,57 @@ private enum hasSerializeMethod (T) = is(T == struct)
 
 public ubyte[] serializeFull (T) (scope const auto ref T record)
     @safe
-    if (is(T == struct))
```
AndrejMitrovic, line 70: Will this end up enabling serialization for classes too?

source/agora/common/Deserializer.d
```diff
@@ -149,61 +149,63 @@ public T deserializeFull (T) (scope DeserializeDg dg) @safe
...
-    @trusted
-    if (isScalarType!T)
-{
-    static if (is(Unqual!T == bool))
+    // Enum deserialize as their base type
+    else static if (is(T == enum))
+    {
+        OriginalType!T orig_val;
+        deserializePart(orig_val, dg);
+        record = cast(T)(orig_val);
```
AndrejMitrovic, line 207: What about `assert(orig_val >= T.min && orig_val <= T.max);`? Although I'm not sure if this works well with things like `enum E { f = "foo", b = "bar" }`..

source/agora/common/Deserializer.d
```diff
@@ -286,7 +288,9 @@ private T deserializeVarInt (T) (scope DeserializeDg dg)
         return cast(T)value;
     }
 
-    if (int_size <= 0xFC)
+    static if (is(Unqual!T == ubyte))
```
AndrejMitrovic: Did you add this to evade a cast? But wouldn't the compiler already know that a cast from ubyte to ubyte is just a noop? Curious.

Geod24: No, i actually wrote it because `ubyte` are not size-encoded (it doesn't make sense). But to keep the code generic, this function is always called. Come to think of it, modifying the call site might be better...

source/agora/common/Deserializer.d
Geod24: I didn't wanna make any assumption about the type of `enum`. I don't think we support enum strings though, only enum based on integer. There are quite a few places where we need to sanitize the input: most notably, length of array (someone could DoS us with just a bit of binary data...). Maybe we should open an issue for this ?

source/agora/common/Serializer.d
Geod24: Hum yes although that wasn't the point.

source/agora/common/Deserializer.d
Geod24: Not it wasn't because it needs `!!` (`ubyte` does not convert to `bool` implicitly). Will add test.

Geod24:
Updated:
- Less commits
- Added a test for `bool`
- Reverted the `ubyte` change to `deserializeVarInt`
- Fixed `Note` typo;

Comments left "unaddressed":
- https://github.com/bpfkorea/agora/pull/539#discussion_r379246181 (was a question)
- https://github.com/bpfkorea/agora/pull/539#discussion_r379246996 (enums are weird and not well supported so far)

AndrejMitrovic:
@bpfkorea/core have you looked at this, does it make sense to you?

AndrejMitrovic:
I guess nobody disagrees with the changes. Merging..

AndrejMitrovic (approved):

source/agora/common/Serializer.d
Geod24: [Aaaaaaaaaacccckkkkktually...](https://github.com/bpfkorea/agora/blob/c0b4e7ba5e72dafab168238439e09df3b5b3e37d/source/agora/common/BanManager.d#L266-L311)

PR #540, opened 2020-02-14
Crypto: Make sure Schnorr's signature type is never used

Geod24:
The Signature type crippled in a few places already.
It should not be used directly, as a signature is, for all intents and purpose,
a binary blob to non-crypto code.

### DIFF
--- a/source/agora/common/crypto/Schnorr.d
+++ b/source/agora/common/crypto/Schnorr.d
@@ -73,9 +73,9 @@ nothrow @nogc @safe unittest
     Point R = R1.V + R2.V;
     Point X = kp1.V + kp2.V;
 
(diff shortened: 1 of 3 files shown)


TrustHenry (approved):
`const ubyte[Signature.sizeof] sig_data = 42;`

`42` It looks like meaningless data.
Does that mean anything to you?

Geod24:
Yes it is meaningless, it could be anything.
The signature is not valid (It was not valid before either), and does not affect serialization (it didn't before either).

Geod24:
Anyone else want to review this ?

AndrejMitrovic (approved):

AndrejMitrovic:
Alright.

PR #543, opened 2020-02-16
Update all compilers to v2.090.1

Geod24:
Half of the fixes in 2.090.1 were made by/for us, so might as well set it as the required version :)


Geod24:
Status are a bit broken (?)

Geod24:
I'll just add to #533, I thought it had already been merged

PR #545, opened 2020-02-17
Fix SCPEnvelope handling and re-enable quorums

### LINKED ISSUE #518: Quorum consensus isn't being properly reached

AndrejMitrovic:
This seems to work for me locally.

### DIFF
--- a/source/agora/network/NetworkClient.d
+++ b/source/agora/network/NetworkClient.d
@@ -161,25 +161,26 @@ class NetworkClient
 
     /***************************************************************************
 
-        Sends an envelope, and returns a status code if the target node
-        has processed the envelope with no errors (todo: this is blocking..)
+        Sends an SCP envelope to another node.
 
-        Returns:
-            true if the client successfully processed the envelope
+        Params:
+            envelope = the envelope to send
 
     ***************************************************************************/
 
-    public bool sendEnvelope (SCPEnvelope envelope) nothrow
+    public void sendEnvelope (SCPEnvelope envelope) nothrow
     {
         try
         {
-            return this.attemptRequest(this.api.receiveEnvelope(envelope),
-                this.exception);
+            this.taskman.runTask(
+            {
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
Github CI is already green. That's a good sign.

AndrejMitrovic:
@Geod24 it's green. Can you test this out with your PR that was failing?

source/agora/network/NetworkClient.d
```diff
@@ -161,25 +161,26 @@ class NetworkClient
...
+            {
+                this.attemptRequest(this.api.receiveEnvelope(envelope),
+                    null);
+            });
         }
         catch (Exception ex)
         {
-            log.trace("Failed to send envelope: {}", ex.msg);
-            return false;
+            assert(0, "attemptRequest should have cought it");
```
AndrejMitrovic:
```suggestion
            assert(0, "attemptRequest should have caught it");
```

source/agora/network/NetworkClient.d
```diff
@@ -161,25 +161,26 @@ class NetworkClient
 
     /***************************************************************************
 
-        Sends an envelope, and returns a status code if the target node
-        has processed the envelope with no errors (todo: this is blocking..)
```
AndrejMitrovic, line 165: I like how there was a `todo` here. 🤦‍♂

Geod24 (approved):
It can't hurt, can you just fix the typo ?

AndrejMitrovic:
Rebased, waiting for green.

Issue #547, opened 2020-02-17
SCP uses an std::vector for the signature in the SCPEnvelope (requiring malloc)


AndrejMitrovic:
The `signature` field in an `SCPEnvelope` is an `std::vector`. This is problematic because it's a dynamic array, and SCP likes to `free` it when it's done using it.

This means we can only sign the `signature` field with the use of `malloc`. It's a needless allocation, and it's not even using the GC.

There are several options going forward:
- Just use `malloc` and let SCP free it. I will have an initial PR ready with this, but it's only a temporary solution.
- Change `Signature` in `SCP` to be an `std::array<64>`. This turns out to be a non-trivial change because it breaks all the crypto routines in `scpp/src/crypto/*`

AndrejMitrovic:
Dupe of #744

PR #546, opened 2020-02-17
Add support for hashing SCPStatement & SCPEnvelope

AndrejMitrovic:
This will be needed for signing SCP statements.

Edit: Now it's also necessary to fix the issue in https://github.com/bpfkorea/agora/pull/716

I'm not sure where to place this hashing support, other than into the Nominator module because this is the only place it will be used.

I cannot move it to `scpd.types.Stellar_SCP` because this module should not import hashing routines from `agora.*`, and I'm not sure `agora.common.Hash` is a good place for it either.

Also side-note: maybe there should be a generic way of testing hashing behavior for any struct types. Something like a "fuzzer" type helper function that manipulates each field and verifies that the hash changed.

PR #548, opened 2020-02-17
Sign SCP envelopes and add a Byzantine node test

AndrejMitrovic:
Requires:
- ~https://github.com/bpfkorea/agora/pull/545~
- ~https://github.com/bpfkorea/agora/pull/546~

See also malloc `signature` issue ~#547~

Blocked by: ~https://github.com/bpfkorea/agora/issues/744~

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -13,6 +13,7 @@
 
 module agora.consensus.protocol.Nominator;
 
+import agora.common.BitField;
 import agora.common.crypto.ECC;
 import agora.common.crypto.Key;
 import agora.common.crypto.Schnorr;
@@ -25,6 +26,7 @@ import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.ConsensusData;
 import agora.consensus.data.Enrollment;
+import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.Transaction;
 import agora.consensus.EnrollmentManager;
 import agora.network.NetworkManager;
@@ -41,6 +43,9 @@ import scpd.types.Stellar_types;
 import scpd.types.Stellar_types : StellarHash = Hash;
 import scpd.types.Stellar_SCP;
 import scpd.types.Utils;
+import scpd.types.XDRBase;
+
+import geod24.bitblob;
 
 import core.stdc.stdint;
 import core.stdc.stdlib : abort;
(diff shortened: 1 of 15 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -211,9 +212,31 @@ extern(D):
 
     ***************************************************************************/
 
-    public bool receiveEnvelope (SCPEnvelope envelope) @trusted
+    public bool receiveEnvelope (SCPEnvelope envelope) @safe nothrow
```
AndrejMitrovic: This attribute change and the trusted lambda call to `receiveEnvelope` should be part of its own commit.

AndrejMitrovic: I think this PR needs a rebase. `receiveEnvelope` should no longer return a status code.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -223,15 +246,24 @@ extern(D):
...
     {
+        scope (failure) assert(0);
+        import core.stdc.stdlib;
+
+        auto msg = toHash(envelope.statement)[];
+        auto sig = this.key_pair.secret.sign(msg)[];
+
+        // note: SCP free's the memory, it must be malloc-allocated.
+        // see also: #547
+        auto mem = cast(ubyte*)malloc(sig.length);
```
AndrejMitrovic: We need to avoid malloc calls. A signature (as the way we define it) is fixed-size, so there shouldn't be a need to allocate memory.

We should change `SCPEnvelope` to embed a fixed-length signature in place of the vector which it currently uses.

File an issue for it and fix it before moving on with this PR.

source/agora/node/Node.d
```diff
@@ -160,7 +161,7 @@ public class Node : API
             .map!(item => tuple(item.key, item.value))
             .assocArray();
 
-        this.nominator = new Nominator(this.config.node.key_pair,
+        this.nominator = this.getNominator(this.config.node.key_pair,
```
AndrejMitrovic: It seems weird to add yet more dependency injection when all we really want to do is to stop signing envelopes for tests involving byzantine nodes.

It would be better to actually just subclass `Node` and use its own "fake" nominator class.

I'll write up a comment in one of the existing issues about dependency injections in the test-suite.

source/agora/test/Byzantine.d
```diff
@@ -0,0 +1,138 @@
...
+{
+    extern(D) this (KeyPair key_pair, Ledger ledger,
+        TaskManager taskman, NetworkClient[PublicKey] peers,
+        SCPQuorumSet quorum_set)
+    {
+        super(key_pair, ledger, taskman, peers, quorum_set);
+    }
+
+    // refuse to sign
+    extern(C++) override void signEnvelope (ref SCPEnvelope envelope)
```
AndrejMitrovic: There should also be a test where the signature exists but is invalid.

source/agora/test/Base.d
```diff
@@ -577,6 +578,9 @@ public struct TestConf
 
     /// max failed requests before a node is banned
     size_t max_failed_requests = 100;
+
+    /// The threshold. If not set, it will default to the number of nodes
+    size_t threshold;
```
AndrejMitrovic: I think we already added something like this in the meantime. A rebase should resolve this.

AndrejMitrovic:
This is no longer blocked. Comments should be addressed and changes rebased.

AndrejMitrovic:
Rebased. But I think it will need to be reworked again if #785 is merged first.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -306,15 +314,17 @@ extern(D):
...
-        todo: Currently not signing yet. To be done.
-
         Params:
             envelope = the SCPEnvelope to sign
 
     ***************************************************************************/
 
     public override void signEnvelope (ref SCPEnvelope envelope)
     {
+        scope (failure) assert(0);
```
AndrejMitrovic: Remove please.

AndrejMitrovic:
Rebased. But I will add another test to it.

AndrejMitrovic:
This is outdated and will be superseded by schnorr-based signatures.

Edit: Also I accidentally wrote to this branch by mistake. So the diff view is completely different to what it was before.

PR #549, opened 2020-02-18
Make isQuorumSetSane() more informative

### LINKED ISSUE #515: `isQuorumSetSane` is not informative
Currently, when this check fails, there is no mention on *why* (which check) it fails, nor what data made it fail.

AndrejMitrovic:
If this looks good to you, I will also submit a PR upstream.

However in upstream they already use a different set of rules for `isQuorumSetSane`, so we would have to update our bindings to the latest version first. And that's not trivial as of now.

Part of #515

### DIFF
--- a/source/scpp/src/scp/BallotProtocol.cpp
+++ b/source/scpp/src/scp/BallotProtocol.cpp
@@ -247,10 +247,17 @@ bool
 BallotProtocol::isStatementSane(SCPStatement const& st, bool self)
 {
     auto qSet = mSlot.getQuorumSetFromStatement(st);
-    bool res = qSet != nullptr && isQuorumSetSane(*qSet, false);
+    const char* reason = nullptr;
+    bool res = qSet != nullptr && isQuorumSetSane(*qSet, false, &reason);
     if (!res)
     {
         CLOG(DEBUG, "SCP") << "Invalid quorum set received";
+        if (reason != nullptr)
+        {
+            std::string msg(reason);
+            CLOG(DEBUG, "SCP") << msg;
+        }
+
         return false;
     }
 
--- a/source/scpp/src/scp/QuorumSetUtils.cpp
+++ b/source/scpp/src/scp/QuorumSetUtils.cpp
@@ -20,7 +20,8 @@ namespace
 class QuorumSetSanityChecker
 {
   public:
-    explicit QuorumSetSanityChecker(SCPQuorumSet const& qSet, bool extraChecks);
+    explicit QuorumSetSanityChecker(SCPQuorumSet const& qSet, bool extraChecks,
+                                    const char** reason);
     bool
     isSane() const
     {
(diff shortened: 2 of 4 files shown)


Geod24:
Probably need to use `std::string` to satisfy upstream.

AndrejMitrovic:
> Probably need to use std::string to satisfy upstream.

That would upset downstream. :>

source/scpp/src/scp/QuorumSetUtils.cpp
```diff
@@ -33,24 +34,50 @@ class QuorumSetSanityChecker
...
     : mExtraChecks{extraChecks}
 {
-    mIsSane = checkSanity(qSet, 0) && mCount >= 1 && mCount <= 1000;
+    mIsSane = checkSanity(qSet, 0, reason);
+
+    if (mCount < 1 || mCount > 1000)
+    {
+        if (reason != nullptr)
+        {
+            const char* msg = "Number of validator nodes exceeds the limit [1..1000]";
```
Geod24: Technically it's out of bounds, as it can be `0`.
You could also split it up into "Quorum set has no validator" and "Quorum set has over 1000 validators".

source/scpp/src/scp/QuorumSetUtils.cpp
```diff
@@ -33,24 +34,50 @@ class QuorumSetSanityChecker
...
 };
 
 QuorumSetSanityChecker::QuorumSetSanityChecker(SCPQuorumSet const& qSet,
-                                               bool extraChecks)
+                                               bool extraChecks,
+                                               const char** reason)
     : mExtraChecks{extraChecks}
 {
-    mIsSane = checkSanity(qSet, 0) && mCount >= 1 && mCount <= 1000;
+    mIsSane = checkSanity(qSet, 0, reason);
```
Geod24: You can simplify the code by making the `reason` optional on the API, but required in the functions (IOW provides a dummy variable if `reason` is null so you avoid bounds check).

AndrejMitrovic:
Can we already use `std::string` in D code? I see there's some support code in https://github.com/dlang/druntime/blob/master/src/core/stdcpp/string.d

Geod24:
I think we can, as long as we don't copy it and just read it.

AndrejMitrovic:
Fixed to avoid bounds / null checking. I'm using `std::string` in another branch (https://github.com/AndrejMitrovic/agora/tree/quorum-is-sane%2Bstd-string), but it fails with:

```
libstdc++ std::__cxx11::basic_string is not yet supported; the struct contains an interior pointer which breaks D move semantics!
```

source/scpp/src/scp/QuorumSetUtils.cpp
```diff
@@ -33,24 +34,50 @@ class QuorumSetSanityChecker
...
 {
-    mIsSane = checkSanity(qSet, 0) && mCount >= 1 && mCount <= 1000;
+    const char* msg;
+    if (reason == nullptr)
+        reason = &msg;  // avoid null checks in checkSanity()
+
+    mIsSane = checkSanity(qSet, 0, reason);
+    if (mCount < 1)
+    {
+        const char* msg = "Number of validator nodes is zero";
```
Geod24: Am I missing something w.r.t. the `const char* msg = ...; *reason = msg` pattern ? Wouldn't `*reason = ...` be more straightforward ?
Also assert that `reason` and `mIsSane` are consistent at the end of this constructor.

Geod24:
> but it fails with:

Sad!

source/scpp/src/scp/BallotProtocol.cpp
```diff
@@ -247,10 +247,17 @@ bool
 BallotProtocol::isStatementSane(SCPStatement const& st, bool self)
 {
     auto qSet = mSlot.getQuorumSetFromStatement(st);
-    bool res = qSet != nullptr && isQuorumSetSane(*qSet, false);
+    const char* reason;
```
AndrejMitrovic: bad bad bad! no `T.init` in C++.

AndrejMitrovic:
Updated.

source/scpp/src/scp/QuorumSetUtils.cpp
```diff
@@ -33,24 +34,49 @@ class QuorumSetSanityChecker
...
+        mIsSane = false;
+    }
+    else if (mCount > 1000)
+    {
+        *reason = "Number of validator nodes exceeds the limit of 1000";
+        mIsSane = false;
+    }
+
+    // only one of the two may be true
+    assert(mIsSane ^ (*reason != nullptr));
```
Geod24, line 62: Attaboy

Geod24 (approved):

PR #551, opened 2020-02-18
Some minor deserialization fixes, and add length check

Geod24:
Once again, review commit-by-commit.

### DIFF
--- a/source/agora/common/Deserializer.d
+++ b/source/agora/common/Deserializer.d
@@ -117,6 +117,42 @@ unittest
 
 /*******************************************************************************
 
+    Deserialize a length, checks if it's within reasonable bound, and return it
+
+    This is a convenience function that calls `deserializeFull`, and checks
+    the result matches a certain bound.
+
+    It is used everywhere an array is deserialized, for example to prevent
+    a DoS if we received crafted binary data that would lead us to allocates
+    large amounts of memory, either exceeding our available memory,
+    or overloading the GC.
+
+    Params:
+        dg         = Delegate to read binary data for deserialization
+        upperBound = The value of the upper bound (inclusive).
+                     Default to `DefaultMaxLength`.
+
+
+    Throws:
+        if the length is > to `upperBound`
+
+    Returns:
+        The deserialized length
+
+*******************************************************************************/
+
+public size_t deserializeLength (
+    scope DeserializeDg dg, size_t upperBound = DefaultMaxLength)
+    @safe
+{
+    size_t len = deserializeFull!size_t(dg);
+    if (len > upperBound)
+        throw new Exception(format("Value of 'length' exceeds upper bound (%d > %d)", len, upperBound));
+    return len;
+}
+
+/*******************************************************************************
+
     Deserialize a data type and return it
 
     Params:
@@ -146,9 +182,8 @@ public T deserializeFull (T) (scope ubyte[] data) @safe
 }
 
 /// Ditto
-public T deserializeFull (T) (
-    scope DeserializeDg dg, CompactMode compact = CompactMode.Yes)
-    @safe
+public T deserializeFull (T) (scope DeserializeDg dg,
+    DeserializerOptions opts = DeserializerOptions.init) @safe
 {
     import geod24.bitblob;
 
@@ -169,8 +204,7 @@ public T deserializeFull (T) (
     else static if (isNarrowString!T)
     {
         alias E = ElementEncodingType!T;
-        size_t length = deserializeVarInt!size_t(dg);
-
+        size_t length = deserializeLength(dg, opts.maxLength);
         T process () @trusted
         out (record)
         {
@@ -187,21 +221,21 @@ public T deserializeFull (T) (
     // If it's binary data, just copy it
     else static if (is(immutable(T) == immutable(ubyte[])))
     {
-        size_t length = deserializeVarInt!size_t(dg);
+        size_t length = deserializeLength(dg, opts.maxLength);
         return dg(ubyte.sizeof * length).dup;
     }
 
     // Array deserialization
     else static if (is(T : E[], E))
     {
-        size_t length = deserializeVarInt!size_t(dg);
-        return iota(length).map!(_ => dg.deserializeFull!(ElementType!T)).array();
+        size_t length = deserializeLength(dg, opts.maxLength);
+        return iota(length).map!(_ => dg.deserializeFull!(ElementType!T)(opts)).array();
     }
 
     // Enum deserialize as their base type
     else static if (is(T == enum))
     {
-        return cast(T) deserializeFull!(OriginalType!T)(dg);
+        return cast(T) deserializeFull!(OriginalType!T)(dg, opts);
     }
 
     // 'bool' need to be converted explicitly
@@ -217,7 +251,7 @@ public T deserializeFull (T) (
             return dg(ubyte.sizeof)[0];
         else
         {
-            if (compact == CompactMode.Yes)
+            if (opts.compact == CompactMode.Yes)
                 return deserializeVarInt!T(dg);
             else
                 return () @trusted { return *cast(T*)(dg(T.sizeof).ptr); }();
@@ -233,7 +267,7 @@ public T deserializeFull (T) (
     {
         Target convert (Target) ()
         {
-            return deserializeFull!Target(dg, compact);
+            return deserializeFull!Target(dg, opts);
         }
         return T(staticMap!(convert, Fields!T));
     }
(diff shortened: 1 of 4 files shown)


source/agora/common/Deserializer.d
```diff
@@ -136,8 +136,13 @@ unittest
 
 public T deserializeFull (T) (scope ubyte[] data) @safe
 {
+    import std.format;
     scope DeserializeDg dg = (size) @safe
     {
+        if (size > data.length)
+            throw new Exception(
```
AndrejMitrovic, line 174: So the idea is that deserialization errors are recoverable. I guess it depends entirely on the calling code, so that would make sense..

source/agora/common/Serializer.d
```diff
@@ -87,6 +87,20 @@ public void serializePart (T) (scope const auto ref T record, scope SerializeDg
         serializePart(v, dg);
 }
 
+/// Write the serialization data to a pre-allocated buffer (useful for tests)
+package(agora) ubyte[] serializeToBuffer (T) (scope const auto ref T record, scope ubyte[] buffer)
```
AndrejMitrovic: If it's for tests only, shouldn't it be `version (unittest)`?

source/agora/common/Deserializer.d
```diff
@@ -119,6 +119,46 @@ unittest
...
+
+    It is used everywhere an array is deserialized, for example to prevent
+    a DoS if we received crafted binary data that would lead us to allocates
+    large amounts of memory, either exceeding our available memory,
+    or overloading the GC.
+
+    Params:
+        dg         = Delegate to read binary data for deserialization
+        upperBound = The value of the upper bound (inclusive).
+                     Default to 0x09D0, which is a bit less than a page on most
```
AndrejMitrovic: I would move this comment to the `DefaultLengthUpperBound` declaration, because that's where the magic number appears.

source/agora/common/Deserializer.d
```diff
@@ -119,6 +119,46 @@ unittest
...
+
+    Throws:
+        if the length is > to `upperBound`
+
+    Returns:
+        The deserialized length
+
+*******************************************************************************/
+
+public size_t deserializeLength (
```
AndrejMitrovic, line 144: This still requires the programmer to use this routine when deserializing a length, which they may forget to use. I guess that's why the commit says `It's not a perfect fix..`

What about in the future replacing this with a `deserializeArray` which does all of this internally, and only allowing arrays to be deserialized with such a function?

AndrejMitrovic (commented):
Small comments.

source/agora/common/Deserializer.d
Geod24: They should be recoverable (because the only to proof that a given `ubyte[]` is valid is to feed it to the deserializer), but in this case it was to get a better error message.

source/agora/common/Serializer.d
Geod24: Good point

source/agora/common/Deserializer.d
```diff
@@ -284,6 +284,14 @@ public void deserializePart (T) (
         foreach (const ref field; record.tupleof)
             deserializePart(field, dg);
 
+    // Support for associative arrays / hashmap
```
AndrejMitrovic: Ah, the serialization support was added in another PR, without the deserialization part?

source/agora/common/Deserializer.d
```diff
@@ -284,6 +284,14 @@ public void deserializePart (T) (
...
+    else static if (is(T : V[K], K, V))
+    {
+        size_t length = deserializeLength(dg, upperBound);
```
AndrejMitrovic: Technically someone could create binary data with length = 1000, but where all the keys are the same. In the end we would have a hashmap with only one entry.

I can't really think of how this could be an attack vector. It's probably nothing to worry about..

source/agora/common/Deserializer.d
Geod24: Yes, that's exactly what I'm moving forward. One of the goal is to turn `T.init.deserialize` into `T.fromBinary` (from member function to static factory).

source/agora/common/Deserializer.d
Geod24: Ah, serialization, I knew I forgot something

source/agora/common/Deserializer.d
Geod24: Good point, I didn't think of this. Yes it opens a few possibilities. Should we just disallow it ?

AndrejMitrovic: something like `if (key in aa) throw`?

Geod24: It's a double lookup, so I was using of using `update` https://github.com/dlang/druntime/blob/9354a9207f4417d4b97c42f4a19dd93c2bfe3108/src/object.d#L2650

Geod24:
Addressed the review on the comment and removed the contentious commit (AA deserialization). Ready for another round of review.

source/agora/common/Deserializer.d
```diff
@@ -119,6 +119,54 @@ unittest
         Amount(100), [1, 2, 3]));
 }
 
+/*******************************************************************************
+
+    Default upper bound for `deserializeLength` and friends
+
+    The value is 0x09D0, which is a bit less than a page on most systems,
+    and allows to have allocations taking a full page, as the GC puts some
+    metadata on each page.
```
AndrejMitrovic: Do we know how big the metadata is? Is it defined somewhere?

IOW how much lower is this upper bound, is it just a guess?

source/agora/common/Deserializer.d
AndrejMitrovic: Should we use a reusable exception? Because `new`-ing here is also a forum of DDoS I would think.. although the GC should just collect the throwaway exceptions I suppose..

AndrejMitrovic: Anyway we can move that into a separate PR.

AndrejMitrovic:
> [stderr] core.exception.AssertError@/Users/runner/runners/2.165.2/work/agora/agora/./tests/unit/BlockStorage.d(117): Assertion failure

Is BlockStorage.d doing something nasty?

source/agora/common/Deserializer.d
Geod24: Yep it's just a guess, I'd have to clarify this. I'm banking on 256 bytes on x86_64 (2 pointers, 32 bytes of bitmap, and 16 for alignment / canary).

AndrejMitrovic: It's alright. We can fine-tune it later.

Geod24:
#594 helped:
```
[stderr]	object.Exception@/Users/geod24/projects/bpfk/agora/source/agora/node/BlockStorage.d(136): Reading block at height 54 failed
[stderr]	----------------
[stderr]	??:? @safe void agora.node.BlockStorage.IBlockStorage.readBlock(ref agora.consensus.data.Block.Block, ulong) [0x10720f06f]
[stderr]	??:? _Dmain [0x1071e686f]
Test /Users/geod24/projects/bpfk/agora/./tests/unit/BlockStorage.d failed!
```

Geod24:
Finally found the issue, thanks to a lot of other improvements, it became very obvious.
There was actually a problem, in that `CompactMode` is an `enum` with `bool` as base type.
So it implicitly converts to `size_t` and that means some places which used `CompactMode` directly set a length limit of either 0 or 1, because the signature changed from:
```D
T deserializeFull (T) (scope DeserializeDg, CompactMode = CompactMode.Yes);
```
to
```D
T deserializeFull (T) (scope DeserializeDg, size_t maxSize = DefaultValue, CompactMode = CompactMode.Yes);
```

But the worst part is that `BlockStorage.load` was returning a `bool`, that was [not checked in the tests](https://github.com/bpfkorea/agora/blob/acbd5464cd735d6c62910d49e2bac7346926ceb0/tests/unit/BlockStorage.d#L106). It's ironic because I caught [one occurence](https://github.com/bpfkorea/agora/blob/acbd5464cd735d6c62910d49e2bac7346926ceb0/tests/unit/BlockStorage.d#L47) but missed the second one.

Geod24:
Fixed

source/agora/common/Deserializer.d
```diff
@@ -235,6 +276,28 @@ public T deserializeFull (T) (
...
 }
 
+/// Options that configure the behavior of the deserializer
+public struct DeserializerOptions
+{
+    /// The bound to apply to a length deserialization (e.g. for arrays)
+    public size_t maxLength = DefaultMaxLength;
+
+    /// Whether or not to use compact notation for unsigned integer
+    public CompactMode compact = CompactMode.Yes;
```
TrustHenry, line 286: How about  `CompactMode.No` as the default? 
`CompactMode.Yes` is only used in special cases.

Geod24: `CompactMode.Yes` is currently the default, hence why I made it the default here.
I am not sure whether it would be good to make it `No` by default. Bear in mind that it only applies to unsigned integers: `ushort`, `uint`, `ulong`. We have a lot of the later (lengths), I doubt we have much of the rest.

In any case, I think changing the default should be in another PR.

TrustHenry: Yes, if we have to change it, it would be better to do it on another PR.

TrustHenry (approved):
LGTM

linked0 (approved):
Looks good to me.

PR #555, opened 2020-02-19
[trivial] Fix counting tests in parallel loop

AndrejMitrovic:
It was using non-atomic increment which can suffer from data races.


Geod24 (approved):

AndrejMitrovic:
Why does docker cloud not even show up? Hmm.

Geod24:
> Why does docker cloud not even show up? Hmm.

Maybe they banned you

Geod24:
For life

PR #558, opened 2020-02-19
Rename 'getUnregisteredEnrollments' to 'getUnregistered'

TrustHenry:
There is a reason for this:

  - Since getUnregisteredEnrollments is a method of EnrollmentManager 

https://github.com/bpfkorea/agora/pull/504#discussion_r376204501

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -530,7 +530,7 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public Enrollment[] getUnregisteredEnrollments (ref Enrollment[] enrolls)
+    public Enrollment[] getUnregistered (ref Enrollment[] enrolls)
         @trusted
     {
         enrolls.length = 0;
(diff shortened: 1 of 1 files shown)


Geod24 (approved):

Geod24:
There's a few other methods that could use the same treatment:
- `getEnrollmentLength` => Either `getLength` or `count` (I prefer count because the operation is expensive, and `count` reflect this better IMO).
- `addEnrollment`, `removeEnrollment`

The other I'm on the fence (not sure), but those seems rather straightforward, don't you think ?

TrustHenry:
> There's a few other methods that could use the same treatment:
> 
> * `getEnrollmentLength` => Either `getLength` or `count` (I prefer count because the operation is expensive, and `count` reflect this better IMO).
> * `addEnrollment`, `removeEnrollment`
> 
> The other I'm on the fence (not sure), but those seems rather straightforward, don't you think ?

I think the target will be subject to refactoring.
I was not conscious because I got used to it :)

PR #563, opened 2020-02-19
Reveal a pre-image periodically

linked0:
A validator reveals a pre-image periodically.
- Every 1 hour, or 6 blocks. 
- The revelation period could become configurable in another issue.
- The pre-image has the data for double the period because it make the validator to be able to skip to next revelation timing In case of a failure of the revelation.
- For unittest, the updateEnrolledHeight function is called manually because the PR #504 is being reviewed now. The PR has the code setting an enrolled height for an enrollment.

#492

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -66,8 +66,15 @@ public class EnrollmentManager
     /// Enrollment data object
     private Enrollment data;
 
-    /// The cycle length for a valdator
+    /// Next height for pre-image revelation
+    private ulong next_reveal_height;
+
+    /// The cycle length for a validator
     public static immutable uint ValidatorCycle = 1008; // freezing period / 2
+
+    /// The period for revealing a preimage
+    /// It is an hour interval if a block is made in every 10 minutes
+    public static immutable uint PreimageRevealPeriod = 6;
 
     /***************************************************************************
 
@@ -120,6 +127,9 @@ public class EnrollmentManager
 
         if (!results.empty)
             this.enroll_key = results.oneValue!(ubyte[]).deserializeFull!(Hash);
+
+        // load next height for preimage revelation
+        this.next_reveal_height = this.getNextRevealHeight();
     }
 
     /***************************************************************************
@@ -309,6 +319,11 @@ public class EnrollmentManager
                 enroll_hash, ex);
             return false;
         }
+
+        // set next height for revealing a pre-image
+        if (enroll_hash == this.enroll_key)
+            this.setNextRevealHeight(block_height);
+
         return true;
     }
 
@@ -546,6 +561,26 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Get a pre-image for revelation
+
+        Params:
+            preimage = will contain the PreimageInfo if exists
+
+        Returns:
+            true if the pre-image exists
+
+    ***************************************************************************/
+
+    public bool getNextPreimage (out PreimageInfo preimage) @safe
+    {
+        auto height = this.next_reveal_height + PreimageRevealPeriod * 2;
+        if (height > ValidatorCycle - 1)
+            height = ValidatorCycle - 1;
+        return getPreimage(height, preimage);
+    }
+
+    /***************************************************************************
+
         Get a pre-image at a certain height
 
         Params:
@@ -697,6 +732,100 @@ public class EnrollmentManager
         }
 
         return true;
+    }
+
+    /***************************************************************************
+
+        Check if revealing a pre-image is needed at a certain height
+
+        Params:
+            height = block height to check
+
+        Returns:
+            true if revealing a pre-image is needed
+
+    ***************************************************************************/
+
+    public bool needRevealPreimage (ulong height) @safe nothrow
+    {
+        if (height >= this.next_reveal_height)
+            return true;
+        else
+            return false;
+    }
+
+    /***************************************************************************
+
+        Increase the next reveal height by the revelation period
+
+    ***************************************************************************/
+
+    public void increaseNextRevealHeight () @safe nothrow
+    {
+        ulong next_height = this.getNextRevealHeight();
+        if (this.next_reveal_height < ulong.max)
+            this.setNextRevealHeight(next_height + PreimageRevealPeriod);
+    }
+
+    /***************************************************************************
+
+        Get the next block height to reveal a pre-image
+
+        Returns:
+            the next block height to reveal a pre-image. if any problem in
+            getting the value, it returns the MAX ulong value.
+
+    ***************************************************************************/
+
+    private ulong getNextRevealHeight () @safe nothrow
+    {
+        ulong next_height = ulong.max;
+        try
+        {
+            () @trusted {
+                auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
+                    "WHERE key = ?", "next_reveal_height");
+                if (!results.empty)
+                    next_height = results.oneValue!(size_t);
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error {}", ex);
+        }
+        return next_height;
+    }
+
+    /***************************************************************************
+
+        Set the next block height to reveal a pre-image
+
+        Params:
+            height = the next block height to reveal a pre-image
+
+    ***************************************************************************/
+
+    private void setNextRevealHeight (ulong height) @safe nothrow
+    {
+        try
+        {
+            () @trusted {
+                auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
+                    "node_enroll_data WHERE key = ?)", "next_reveal_height");
+                if (results.oneValue!(bool))
+                    this.db.execute("UPDATE node_enroll_data SET val = ? " ~
+                        "WHERE key = ?", height, "next_reveal_height");
+                else
+                    this.db.execute("INSERT INTO node_enroll_data (key, val)" ~
+                        " VALUES (?, ?)", "next_reveal_height", height);
+
+                this.next_reveal_height = height;
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error {}", ex);
+        }
     }
 
     /***************************************************************************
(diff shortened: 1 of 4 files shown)


source/agora/node/GossipProtocol.d
```diff
@@ -80,6 +80,16 @@ public class GossipProtocol
             this.network.sendTransaction(tx);
             this.ledger.tryNominateTXSet();
         }
+
+        if (this.enroll_man.needRevealPreimage(this.ledger.getBlockHeight()))
```
TrustHenry: The `needRevealPreimage` check at that location causes unnecessary calls.
I think it would be better to check when the height of the block changes as a timing for checking `next_reveal_height`

linked0:
Ready for review.
All builds and tests have passed.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -885,6 +1014,11 @@ unittest
     assert(man.getPreimage(11, preimage));
     assert(man.getPreimage(10 + EnrollmentManager.ValidatorCycle, preimage));
     assert(!man.getPreimage(11 + EnrollmentManager.ValidatorCycle, preimage));
+
+    /// test for the functions about priodic revelation of a pre-image
```
```suggestion
    /// test for the functions about periodic revelation of a pre-image
```

Geod24 (approved):

Geod24:
Rebased on v0.x.x and fixed

source/agora/consensus/EnrollmentManager.d
```diff
@@ -120,6 +127,9 @@ public class EnrollmentManager
 
         if (!results.empty)
             this.enroll_key = results.oneValue!(ubyte[]).deserializeFull!(Hash);
+
+        // load next height for preimage revelation
+        this.next_reveal_height = this.getNextRevealHeight();
```
AndrejMitrovic, line 132: How come `setNextRevealHeight` is not called here?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -309,6 +319,11 @@ public class EnrollmentManager
                 enroll_hash, ex);
             return false;
         }
+
+        // set next height for revealing a pre-image
+        if (enroll_hash == this.enroll_key)
```
AndrejMitrovic, line 324: In the future we should separate these two functionalities of the `EnrollmentManager`:

- Managing enrollments for other nodes
- Managing enrollments for *this* node

Otherwise the implementation will become complex and full of state, like it has started to show.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -544,6 +559,26 @@ public class EnrollmentManager
...
+            preimage = will contain the PreimageInfo if exists
+
+        Returns:
+            true if the pre-image exists
+
+    ***************************************************************************/
+
+    public bool getNextPreimage (out PreimageInfo preimage) @safe
+    {
+        auto height = this.next_reveal_height + PreimageRevealPeriod * 2;
```
AndrejMitrovic, line 576: Why the `* 2` part? So will this actually reveal preimages for the next 12 blocks (2 hours), rather than 6 blocks (1 hour)?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -544,6 +559,26 @@ public class EnrollmentManager
...
+        if (height > ValidatorCycle - 1)
+            height = ValidatorCycle - 1;
```
AndrejMitrovic, line 578: This looks completely wrong? If the block height is 1009 then we will simply not reveal the preimage?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -544,6 +559,26 @@ public class EnrollmentManager
...
+
+        Params:
+            preimage = will contain the PreimageInfo if exists
+
+        Returns:
+            true if the pre-image exists
+
+    ***************************************************************************/
+
+    public bool getNextPreimage (out PreimageInfo preimage) @safe
```
AndrejMitrovic, line 574: Where are the unittests for this function?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -699,6 +734,100 @@ public class EnrollmentManager
...
+    /***************************************************************************
+
+        Increase the next reveal height by the revelation period
+
+    ***************************************************************************/
+
+    public void increaseNextRevealHeight () @safe nothrow
+    {
+        ulong next_height = this.getNextRevealHeight();
+        if (this.next_reveal_height < ulong.max)
```
AndrejMitrovic, line 766: So if there is a database error then the node will silently stop emitting preimages? The only way the user will know this is if they're actively monitoring the logs. 

In the meantime the node stops emitting preimages, it gets slashed for not following the protocol, and the node owner loses funds. This is not good.

If there is an exceptional error that makes the node misbehave then we have to treat it as an error, not as a diagnostical issue. If the node can't reveal its preimages because of a database error then the node should shutdown with an error. The node operator can get notified when its node shuts down. That way the node operator gets the chance to fix the error (faulty disk or whatever else is causing the issue).

source/agora/consensus/EnrollmentManager.d
```diff
@@ -699,6 +734,100 @@ public class EnrollmentManager
...
+            () @trusted {
+                auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
+                    "WHERE key = ?", "next_reveal_height");
+                if (!results.empty)
+                    next_height = results.oneValue!(size_t);
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error {}", ex);
```
AndrejMitrovic, line 794: As mentioned before, this is a showstopper error. The node cannot reveal preimages, it should be shut down.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -699,6 +734,100 @@ public class EnrollmentManager
...
+                else
+                    this.db.execute("INSERT INTO node_enroll_data (key, val)" ~
+                        " VALUES (?, ?)", "next_reveal_height", height);
+
+                this.next_reveal_height = height;
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error {}", ex);
```
AndrejMitrovic, line 827: Same problem here. We are logging errors and then proceeding as if there was no error.

source/agora/node/Node.d
```diff
@@ -248,6 +248,16 @@ public class Node : API
             this.network.sendTransaction(tx);
             this.ledger.tryNominateTXSet();
         }
+
+        if (this.enroll_man.needRevealPreimage(this.ledger.getBlockHeight()))
```
AndrejMitrovic, line 252: This is not the right place to put this. Why is it being triggered when we receive a Transaction? There might not be enough transactions in the pool to start a nomination. It should be triggered when the ledger's state changes (new block added / externalize was called), and only if we're not catching up on outdated blocks.

AndrejMitrovic (commented):
Post-merge review because I didn't have the chance to review this before.

source/agora/node/Node.d
Geod24: https://github.com/bpfkorea/agora/issues/582

source/agora/consensus/EnrollmentManager.d
Geod24: That was discussed during the whiteboarding session. Currently it's a bit convoluted, as you mentioned, we should use a sliding windows instead.

AndrejMitrovic: Thanks, I missed the first part of the whiteboarding session when this happened.

source/agora/consensus/EnrollmentManager.d
linked0: At first, the `next_reveal_height` is set by calling `setNextRevealHeight` in `updateEnrolledHeight` which is called in `addValidatedBlock` of the ledger, which is being reviewed in the PR #504

source/agora/consensus/EnrollmentManager.d
linked0: > In the future we should separate these two functionalities of the EnrollmentManager

I totally agree with you. I have a plan to split the current `EnrollmentManager` into `EnrollmentManager` and `ValidatorSet`

source/agora/consensus/EnrollmentManager.d
linked0: > If the block height is 1009 then we will simply not reveal the preimage?

As previously mentioned, the `next_reveal_height` is set based on a current block height by calling setNextRevealHeight in updateEnrolledHeight which is called in addValidatedBlock of the ledger which is being reviewed in the PR #504

source/agora/consensus/EnrollmentManager.d
linked0: > Where are the unittests for this function?

I missed it. I will add it in a related PR.

source/agora/consensus/EnrollmentManager.d
linked0: > As mentioned before, this is a showstopper error. The node cannot reveal preimages, it should be shut down

There needs to add the code of `assert(0)` in a related PR.

source/agora/consensus/EnrollmentManager.d
linked0: > Same problem here. We are logging errors and then proceeding as if there was no error.

There needs to add the code of `assert(0)` in a related PR.

source/agora/consensus/EnrollmentManager.d
linked0: > The node operator can get notified when its node shuts down.

Right. There needs a way to notify an abnormal status of a node to the operator through client UI.

AndrejMitrovic: The UI is a separate concern. Agora doesn't have to run with any UI at all. It just has to exit with a process code `!= 0`, and then whichever initialization system is in use (systemd or otherwise) can notify the operator - whether it's through e-mail alerts, push notifications, etc. All of that stuff is external to the Agora node. But we must notify the operating system that we failed, and that means shutting down with an exit code.

linked0: I don't think the operators of nodes need to be familiar with command line environment. That's why I said about UI. Adding the code of assert or something in will be worked on one related PR or one issue.

PR #571, opened 2020-02-21
Fix verification of the quorum section

AndrejMitrovic:
The verification should be moved before `start()` is called.

While fixing this I've noticed a bunch of other issues with the tests, and included these fixes here. I also planned to add a check in the logging output of the node, but this isn't possible to do if the node throws in its constructors - because we can only reach those logs through the API, and by that point the node is dead.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -52,7 +52,7 @@ import geod24.Registry;
 
 import core.stdc.time;
 import std.array;
-import std.algorithm.iteration;
+import std.algorithm;
 import std.exception;
 import std.format;
 
@@ -209,6 +209,18 @@ public class FakeClockBanManager : BanManager
     public override void dump () { }
 }
 
+/// We use a pair of (key, client) rather than a hashmap client[key],
+/// since we want to know the order of the nodes which were configured
+/// in the makeTestNetwork() call.
+public struct NodePair
+{
+    ///
+    public PublicKey key;
+
+    ///
+    public RemoteAPI!TestAPI client;
+}
+
 /*******************************************************************************
 
     Used by unittests to send messages to individual nodes.
@@ -224,14 +236,16 @@ public class TestAPIManager
     /// Used by the unittests in order to directly interact with the nodes,
     /// without trying to handshake or do any automatic network discovery.
     /// Also kept here to avoid any eager garbage collection.
-    public RemoteAPI!TestAPI[PublicKey] apis;
+    public NodePair[] nodes;
+
+    /// convenience: returns a random-access range which lets us access clients
+    auto clients ()
+    {
+        return nodes.map!(np => np.client);
+    }
 
     /// Registry holding the nodes
     protected Registry reg;
-
-    /// Keeps the order of creation as createNewNode() was called.
-    /// Needed to know which node was the first / last one in some tests.
-    public PublicKey[] keys;
 
     ///
     public this ()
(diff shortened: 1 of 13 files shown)


Geod24:
I think you'll need to include https://github.com/Geod24/agora/commit/9c63f686db3f69d00cbe2a4b6d61c164d86f4368 before the final commit to make CircleCI pass.

Second commit (tests: Add timeout setting for the unittest thread): As discussed IRL, let's just roll with 1 single timeout.
Third commit (tests: Replace random node ordering): Maybe (in another PR), providing `opIndex` / `opSlice` could simplify the code. Perhaps `alias this` on the pair struct.

Overall LVGTM. Did similar changes to the last commit locally when trying to remove the quorum configuration...

Geod24:
> I think you'll need to include Geod24@9c63f68 before the final commit to make CircleCI pass.

Tested locally, it works 🎉

AndrejMitrovic:
> Third commit (tests: Replace random node ordering): Maybe (in another PR), providing opIndex / opSlice could simplify the code. Perhaps alias this on the pair struct.

I had the same thought about alias this before I made the PR. It would make the code a little shorter, but it might be harder to understand. In general I think tests should be as simple as possible, otherwise we run the risk of not understanding what the tests are doing somewhere down the line. _cough_ hydrosphere _cough_

Can we leave it as a future enhancement?

Geod24:
> Can we leave it as a future enhancement?

Sure

source/agora/test/Base.d
```diff
@@ -575,8 +585,11 @@ public struct TestConf
     /// max retries before a request is considered failed
     size_t max_retries = 20;
 
-    /// request timeout (in msecs)
-    long timeout = 500;
+    /// request timeout for each node (in msecs)
+    long timeout = 2000;
+
+    /// the timeout for the unittest thread issuing requests to child threads
+    long main_timeout = 10_000;
```
Geod24: Leftover?

AndrejMitrovic: Late night coding leftover. :)

Fixed now.

source/agora/test/Failures.d
```diff
@@ -0,0 +1,48 @@
...
+import agora.utils.Log;
+
+///
+unittest
+{
+    import std.algorithm;
+    import std.range;
+    import core.thread;
+
+    TestConf conf = { threshold : 100, main_timeout : 1_000 };
```
Geod24: Oops

AndrejMitrovic: Just pushed a fix.

AndrejMitrovic:
Although now circleci fails, huh.

AndrejMitrovic:
It seems that maybe CircleCI isn't picking up that I've rebased this PR against master?

Geod24:
No it's one of the other bugs.... I've seen this before. Retriggered it.

AndrejMitrovic:
No the issue is that my `v0.x.x` branch is stale. I don't quite understand why CircleCI is applying my PR on top of my own v0.x.x branch instead of the remote.

AndrejMitrovic:
It's green. :>

Geod24:
> No the issue is that my `v0.x.x` branch is stale. I don't quite understand why CircleCI is applying my PR on top of my own v0.x.x branch instead of the remote.

Ooohhhh that explains it! CircleCI somehow things that people want to test the HEAD of the PR, not the resulting merge commit (which Travis does correctly).
So we have [a special step](https://github.com/bpfkorea/agora/blob/3f8cccd76616e3a17132827c71e73e65c8020605/.circleci/config.yml#L11-L20) to do it. But it looks like it's using the wrong remote. Will look into it.

Geod24:
> It's green. :>

Not anymore :(

Geod24:
Rebased on v0.x.x.
Still fails CircleCI, but it does not locally :(

Geod24 (approved):

PR #576, opened 2020-02-22
Various test-suite fixes & proper cleanup in ctor

AndrejMitrovic:
The last commit in particular seems to get rid of bizarre "unrelated" errors whenever there is a failing test-case. It usually displays SQL cleanup errors.

### DIFF
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -111,9 +111,12 @@ public class Node : API
             config.network, config.dns_seeds, this.metadata, this.taskman);
         this.storage = this.getBlockStorage(config.node.data_dir);
         this.pool = this.getPool(config.node.data_dir);
+        scope (failure) this.pool.shutdown();
         this.utxo_set = this.getUtxoSet(config.node.data_dir);
+        scope (failure) this.utxo_set.shutdown();
         this.ledger = new Ledger(this.pool, this.utxo_set, this.storage, config.node);
         this.enroll_man = this.getEnrollmentManager(config.node.data_dir, config.node);
+        scope (failure) this.enroll_man.shutdown();
         this.exception = new RestException(
             400, Json("The query was incorrect"), string.init, int.init);
     }
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -293,6 +293,9 @@ public class TestAPIManager
 
     public void shutdown ()
     {
+        foreach (node; this.nodes)
+            enforce(this.reg.unregister(node.key.toString()));
+
         foreach (ref node; this.nodes)
         {
             node.client.shutdown();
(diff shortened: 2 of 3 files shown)


AndrejMitrovic:
> It usually displays SQL cleanup errors.

More specifically, there is a destructor call and the SQL bindings error with "shutdown not called". This causes a cascade of errors everywhere, so whenever there is 1 failing test-case in the test-suite sometimes you get a mountain of unrelated errors.

source/agora/test/Base.d
```diff
@@ -242,7 +242,7 @@ public class TestAPIManager
     public void createNewNode (PublicKey address, Config conf)
     {
         auto api = RemoteAPI!TestAPI.spawn!(TestNode)(conf, &this.reg);
-        this.reg.register(address.toString(), api.tid());
+        enforce(this.reg.register(address.toString(), api.tid()));
```
AndrejMitrovic: Actually this commit is a bit misguided. The `spawn` call there is asynchronous, so we cannot guarantee the child thread has finished doing its work by the time this enforce is called. Classic data race.

It's still good to keep the check here, but I will change it to an assert and also change the commit description.

AndrejMitrovic: Resolved.

AndrejMitrovic: Not resolved, this commit needs to be removed. As mentioned, the mbox can become closed if  the thread shuts down due to a faulty constructor. So this assert/enforcement will naturally fail.

AndrejMitrovic: Now it's removed.

AndrejMitrovic:
It's green!

AndrejMitrovic:
@Geod24 ping

source/agora/node/Node.d
```diff
@@ -111,9 +111,12 @@ public class Node : API
             config.network, config.dns_seeds, this.metadata, this.taskman);
         this.storage = this.getBlockStorage(config.node.data_dir);
         this.pool = this.getPool(config.node.data_dir);
+        scope (failure) this.pool.shutdown();
         this.utxo_set = this.getUtxoSet(config.node.data_dir);
+        scope (failure) this.utxo_set.shutdown();
         this.ledger = new Ledger(this.pool, this.utxo_set, this.storage, config.node);
         this.enroll_man = this.getEnrollmentManager(config.node.data_dir, config.node);
+	    scope (failure) this.enroll_man.shutdown();
```
Geod24: WHAT THE HECK

Geod24: I TRUSTED YOU

AndrejMitrovic:
Fixed spacing.

Geod24 (approved):

PR #577, opened 2020-02-22
Decouple Nomination and Ledger classes

### LINKED ISSUE #572: Decouple the Nominator and Ledger classes
Currently it's a little messy. The Nominator can only be instantiated with the known set of peer `API` instances which are available only **after** the handshake stage. So the `Ledger` is actually in an incomplete state and should not be used yet. And yet in the edge case that it doesn't have the nominator set yet, and a transaction is received, it will by default create a new block.

This should be changed so all classes are instantiated first, then perhaps we could have a `startNominating()` call somewhere.

AndrejMitrovic:
Includes some commits from PRs currently in review. I'll remove them as soon as those PRs get merged.

The commits which are not marked `[removed]` may be reviewed.

Requires: ~https://github.com/bpfkorea/agora/pull/576~

Fixes #572

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -168,22 +168,20 @@ extern(D):
 
     public void setupNetwork (NetworkClient[PublicKey] clients)
     {
-        import agora.common.Set;
         import std.algorithm;
         import std.array;
         import std.typecons;
 
-        void getNodes (in QuorumConfig conf, ref bool[PublicKey] nodes)
+        void getNodes (in QuorumConfig conf, ref Set!PublicKey nodes)
         {
             foreach (node; conf.nodes)
-                nodes[node] = true;
+                nodes.put(node);
 
             foreach (sub_conf; conf.quorums)
                 getNodes(sub_conf, nodes);
         }
 
-        // can't use Set(), requires serialization support
-        bool[PublicKey] quorum_keys;
+        Set!PublicKey quorum_keys;
         getNodes(this.quorum_conf, quorum_keys);
 
         this.peers = clients.byKeyValue
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -93,13 +93,12 @@ extern(D):
             key_pair = the key pair of this node
             ledger = needed for SCP state restoration & block validation
             taskman = used to run timers
-            peers = the set of clients to the peers in the quorum
             config = the quorum configuration
 
     ***************************************************************************/
 
     public this (KeyPair key_pair, Ledger ledger, TaskManager taskman,
-        NetworkClient[PublicKey] peers, ref const QuorumConfig config)
+        ref const QuorumConfig config)
     {
         this.quorum_conf = config;
         this.key_pair = key_pair;
@@ -109,7 +108,6 @@ extern(D):
         this.scp = createSCP(this, node_id, IsValidator, quorum_set);
         this.taskman = taskman;
         this.ledger = ledger;
-        this.peers = peers;
 
         auto localQSet = makeSharedSCPQuorumSet(this.scp.getLocalQuorumSet());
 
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
Unblocked!

source/agora/consensus/protocol/Nominator.d
```diff
@@ -80,6 +81,9 @@ public extern (C++) class Nominator : SCPDriver
     /// Timer IDs with >= of the active timer will be allowed to run
     private ulong[TimerType.max + 1] active_timer_ids;
 
+    /// Quorum config
+    private QuorumConfig quorum_conf;
```
```suggestion
    private const QuorumConfig quorum_conf;
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -91,17 +95,18 @@ extern(D):
...
-            quorum_set = the quorum set of this node
+            config = the quorum configuration
 
     ***************************************************************************/
 
-    public this (KeyPair key_pair, Ledger ledger,
-        TaskManager taskman, NetworkClient[PublicKey] peers,
-        SCPQuorumSet quorum_set)
+    public this (KeyPair key_pair, Ledger ledger, TaskManager taskman,
+        NetworkClient[PublicKey] peers, QuorumConfig config)
```
```suggestion
        NetworkClient[PublicKey] peers, ref const QuorumConfig config)
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -159,6 +157,42 @@ extern(D):
...
+        void getNodes (QuorumConfig conf, ref bool[PublicKey] nodes)
+        {
+            foreach (node; conf.nodes)
+                nodes[node] = true;
+
+            foreach (sub_conf; conf.quorums)
+                getNodes(sub_conf, nodes);
+        }
+
+        // can't use Set(), requires serialization support
```
Geod24: That not true anymore is it ?

Geod24: Especially since `Set` is imported

AndrejMitrovic: Aha, I've just noticed this is no longer true! Yay refactorings

source/agora/node/Ledger.d
```diff
@@ -277,44 +277,54 @@ public class Ledger
...
+        Params:
+            txs = will contain the transaction set, if the return value is true
+
+        Returns:
+            true if there were enough valid transactions found
+            to constitute a block
 
+    ***************************************************************************/
+
+    private bool prepareNominatingSet (ref Set!Transaction txs) @safe
```
Geod24: Is there ever a chance we nominate an empty set ? If not we can just simplify this signature...

source/agora/node/Ledger.d
```diff
@@ -277,44 +277,54 @@ public class Ledger
...
+        if (!this.prepareNominatingSet(txs))
             return;
 
-        Hash[] hashes;
-        Set!Transaction txs;
-        ulong expect_height = this.getBlockHeight() + 1;
+        // note: we are not passing the previous tx set as we don't really
+        // need it at this point (might later be necessary for chain upgrades)
+        auto slot_idx = this.last_block.header.height + 1;
+        this.nominate(slot_idx, Set!Transaction.init, txs);
```
Geod24: Looks like we might want to change the signature of `NominateDg` at some point if it's always called with `.init` as second arg ?

Geod24 (commented):
First round of review done, will need another one for the last two commits

source/agora/consensus/protocol/Nominator.d
Geod24: And yay self reviews!

source/agora/node/Ledger.d
AndrejMitrovic: This delegate is removed in a further commit. I originally had it part of one commit, but the changes were very hard to review, so I split it up.

AndrejMitrovic:
- Rebased against master
- Fixed definition of `prepareNominatingSet` so it doesn't needlessly return a boolean status flag.
- Replaced hashes with sets.
- Added const.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -91,17 +95,18 @@ extern(D):
...
-        TaskManager taskman, NetworkClient[PublicKey] peers,
-        SCPQuorumSet quorum_set)
+    public this (KeyPair key_pair, Ledger ledger, TaskManager taskman,
+        NetworkClient[PublicKey] peers, ref const QuorumConfig config)
     {
+        this.quorum_conf = config;
         this.key_pair = key_pair;
         auto node_id = NodeID(StellarHash(key_pair.address[]));
         const IsValidator = true;
+        auto quorum_set = verifyBuildSCPConfig(config);
```
Geod24, line 110: Pass it directly on the next line to avoid a copy.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -158,6 +156,42 @@ extern(D):
...
+        Params:
+            clients = the set of all clients we're networked with. A subset
+                      of the clients are in our quorum set, these will be
+                      the clients we exchange SCP messages with.
+
+    ***************************************************************************/
+
+    public void setupNetwork (NetworkClient[PublicKey] clients)
+    {
+        import agora.common.Set;
```
Geod24: This import is redundant

Geod24 (approved):
LGTM, only 2 nits. Feel free to self-merge after addressing them (and rebasing, sorry :D )

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: I can't, it takes a reference.

PR #579, opened 2020-02-22
integration-test: wait for node discovery first

AndrejMitrovic:
I've noticed this issue in the last week. I can't guarantee it fixes anything, but it follows the way we test the nodes in the test-suite.

We should also rework how the nodes work to ensure they reject any API requests until they are ready to handle such events. For example, a node should not attempt to gossip requests if its discovery stage is incomplete.

I've tested compiling this locally, but I can't verify it works yet. Having issues with Docker ports on WSL2. :/

### DIFF
--- a/tests/system/source/main.d
+++ b/tests/system/source/main.d
@@ -21,15 +21,28 @@ import agora.common.crypto.Key;
 import agora.common.Hash;
 import agora.common.Set;
 import agora.utils.PrettyPrinter;
+import agora.utils.Test;
 
 import vibe.web.rest;
 
 import std.algorithm;
 import std.exception;
+import std.format;
+import std.range;
 import std.stdio;
 import core.thread;
 import core.time;
 
+// keep polling the nodes for a complete network discovery, until a timeout
+private void waitForDiscovery (API[] clients, Duration timeout)
+{
+    clients.enumerate.each!((idx, api) =>
+        retryFor(api.getNetworkInfo().ifThrown(NetworkInfo.init)
+            .state == NetworkState.Complete,
+            timeout,
+            format("Node %s has not completed discovery after %s.",
+                idx, timeout)));
+}
 
 int main (string[] args)
 {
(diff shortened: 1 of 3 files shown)


tests/system/source/main.d
```diff
@@ -51,13 +54,55 @@ private immutable Address[] Addrs = [
     { host: "127.0.0.1", port: 4002, },
 ];
 
+/// keep retrying 'check' condition until it's troe or the timer expires.
+/// If the timer expires, throw an exception. See also agora.test.Base.retryFor
+private void retryFor (lazy bool check, Duration timeout,
```
AndrejMitrovic: Also this is mostly copy-pasted code. We should ideally just move it to a common module which the integration test can import, as well as the unittest modules. Would that be ok? Or do we just go with copy-pasting for now? (I prefer the former).

AndrejMitrovic:
Ok, CircleCI still fails. It seems it's consistent too.

Geod24:
But Dockercloud is back! And green...

AndrejMitrovic:
This passed CircleCI twice now.

tests/system/source/main.d
Geod24: https://github.com/bpfkorea/agora/blob/0120d6832a6112338d8588480400beb240ea60be/source/agora/utils/Test.d#L3-L19

AndrejMitrovic:
Moved `retryFor` into `agora.utils.Test`. Waiting for CircleCI..

Geod24 (changes requested):
Why do we need an empty import in all tests ? Looks like a leftover. Otherwise looks good but needs a rebase.

AndrejMitrovic:
What do you mean by "empty" import? `retryFor` is used in those tests. However I could add a public import of `agora.utils.Test` into `agora.test.Base`..

Geod24:
Ah right you moved it 🤦‍♂ 
Either works for me, what do you prefer ?

AndrejMitrovic:
People will frequently forget to import it, so I'd say let's go with public import.

AndrejMitrovic:
Updated, waiting for circle & friends.

Geod24 (approved):
Merge at will

AndrejMitrovic:
It seems the checkout command fails frequently. I'll restart the run..

AndrejMitrovic:
> ##[error]fatal: reference is not a tree: 7e6d87b35522b6b1ffbe06e39ea4f71a14995e68

Ah I bet it's my fault.

PR #580, opened 2020-02-23
Remove ConfigException, use Exception directly

Geod24:
```
ConfigException was an attempt at simplifying the code that didn't turn quite right.
The whole point of having it seemed to avoid declaring the 'Config' variable outside
of the 'try' block, but it resulted in an extra level of indentation for everything,
and a try-catch-throw which worked around the fact that only the YAML parser throws.
```

### DIFF
--- a/source/agora/node/main.d
+++ b/source/agora/node/main.d
@@ -67,6 +67,7 @@ private int main (string[] args)
             ex.message);
     }
 
+    Node node;
     try
     {
         auto config = parseConfigFile(cmdln);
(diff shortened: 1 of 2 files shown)


source/agora/node/main.d
```diff
@@ -75,31 +76,47 @@ private int main (string[] args)
             writefln("Config file '%s' succesfully parsed.", cmdln.config_path);
             return 0;
         }
+        runTask(() => node = runNode(config));
```
Geod24, line 79: Note: This is the trick I was missing earlier. Because  `shutdown` needs to be called after the event loop returns (`listenHTTP` is not a blocking call, it merely registers with the event loop).
But it was originally buried under the other errors.

AndrejMitrovic: More specifically for future reference: If shutdown is called before the event loop returns, there is a possibility the node's API gets called again **after** shutdown was called. The node should not be used after `shutdown()` is claled.

AndrejMitrovic (approved):

PR #583, opened 2020-02-24
Show compiler and dub version in Github / Travis

Geod24:
```
This is only useful when testing with dmd-nightly / ldc-latest-ci.
This will show the exact commit that is being used for testing,
helping to debug any future issue.
```

I am seeing a weird issue, where a deprecation on `body` is triggered, but it is disabled in master...

### DIFF
--- a/.github/workflows/github_push.yaml
+++ b/.github/workflows/github_push.yaml
@@ -18,6 +18,10 @@ jobs:
       with:
           compiler: ${{ matrix.dc }}
           gh_token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Print compiler and dub's version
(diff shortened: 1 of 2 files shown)


Geod24:
Note: CircleCI uses a fixed version so it is not affected, only Github and Travis use nightlies.

Geod24:
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  invalid public key type
Program exited with code -6
The command "./ci/run.sh" exited with 2.
```

With ldc2 1.20.0 on Linux... Nothing to see here.

AndrejMitrovic (approved):

Issue #510, opened 2020-01-31
Build single-binary networks


Geod24:
This is an idea I want to leave here for discussion.

Currently, we have 2 kinds of tests:
- Unittests, which use `localrest` to simulate a network.
- Full fledged network tests, using `docker-compose` (see tests/system).

However, full-fledged integration tests are quite expensive (require to spawn a container, no sharing of data, etc...). One possibility to test network would be to be able to generate a single binary which  exposes a network via the usual Node API, but internally uses LocalRest for networking.

With a sensible CLI, it could also possibly allow to test multiple network configurations (e.g. introducing bad nodes) without recompilation.

Geod24:
So currently we have system integration tests (https://github.com/bpfkorea/agora/tree/v0.x.x/tests/system) which spawns 3 docker images of the node, with 3 different configuration files (https://github.com/bpfkorea/agora/tree/v0.x.x/tests/system/node under {0,1,2}/config.yaml), and a runs a test defined [here](https://github.com/bpfkorea/agora/blob/v0.x.x/tests/system/source/main.d). The full pipeline is run via [`ci/system_integration_test.d`](https://github.com/bpfkorea/agora/blob/v0.x.x/ci/system_integration_test.d)

However, spawning 3 docker images is expensive and won't scale very well. One thing we'd like to do is to have an intermediate between the LocalRest tests we have in `source/agora/test` and the full-fledge (complete) system integration tests described here.

This can be done by having a special main that creates nodes based on its arguments. All it would need is the ability to read multiple config files, and instantiate multiple `Node` as a result. A good starting point is to look at [this code](https://github.com/bpfkorea/agora/blob/c7314c21700c526d6ee496090694bef46dc9da54/source/agora/node/main.d#L70-L89).

Geod24:
Also it is probably best to make it another "target", or configuration. This is what we are currently doing with `agora-cli`: https://github.com/bpfkorea/agora/blob/c7314c21700c526d6ee496090694bef46dc9da54/dub.json#L41-L71

MichaelKim20:
@Geod24 I am going to use LocalRest to create nodes according to the config file.
Is it correct to use LocalRest?

Geod24:
When we discussed the topic in our sprint planning, I thought so.
However, using LocalRest would prevent the client talking to any node, so all things considered I don't think it should use it.

Geod24:
Done in #592

PR #590, opened 2020-02-25
Change `agora-cli` to `agora-client`

MichaelKim20:
I would like to include various CLI programs in `source/agora/cli`.
I want to change the previously developed `agora-cli` to `agora-client`.
The detailed processing order is as follows.

1. Change a source directory of `agora-cli`; "source/cli" => "source/cli/client".
2. Change binary name `agora-cli` to `agora-client`.
3. Update `cli` to `client` in source code.
4. Update CI (run.sh).
5. Update a CI of the 'agora-client'

After this, I will add a test CLI called `agora-nettest` under the "source/cli".
The new folder is "source/cli/nettest".

### DIFF
--- a/source/agora/cli/client/DefaultProcess.d
+++ b/source/agora/cli/client/DefaultProcess.d
@@ -11,9 +11,9 @@
 
 *******************************************************************************/
 
-module agora.cli.DefaultProcess;
+module agora.client.DefaultProcess;
 
-import agora.cli.CLIResult;
+import agora.client.CLIResult;
 
 import std.getopt;
 
--- a/source/agora/cli/client/DefaultProcess.d
+++ b/source/agora/cli/client/DefaultProcess.d
@@ -1,6 +1,6 @@
 /*******************************************************************************
 
-    The Agora CLI sub-function for default command
+    The Agora client sub-function for default command
 
     Copyright:
         Copyright (c) 2019 BOS Platform Foundation Korea
@@ -13,7 +13,7 @@
 
 module agora.client.DefaultProcess;
 
-import agora.client.CLIResult;
+import agora.client.Result;
 
 import std.getopt;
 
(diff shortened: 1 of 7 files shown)


Geod24:
As mentioned, commits should be atomic, so "Update a CI of the `agora-client`" needs to be merged into "Change `agora-cli` to `agora-client`" otherwise the CI won't pass for the later.

Also I don't see how this would fix issue #510, it's merely the ground work.

source/agora/cli/client/DefaultProcess.d
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
 
-    The Agora CLI sub-function for default command
+    The Agora Client sub-function for default command
```
Geod24:
```suggestion
    The Agora client sub-function for default command
```

source/agora/cli/client/DefaultProcess.d
```diff
@@ -11,9 +11,9 @@
 
 *******************************************************************************/
 
-module agora.cli.DefaultProcess;
+module agora.client.DefaultProcess;
 
-import agora.cli.CLIResult;
+import agora.client.Result;
```
Geod24, line 16: Right, actually, this won't compile either, the changes to the module name need to be at the same time as the file path change.

It might be simpler to just merge all commits.

MichaelKim20: @Geod24 I fixed it

Geod24 (approved):
Nice, LGTM

TrustHenry (approved):
LGTM

AndrejMitrovic:
@Geod24 I was under the impression that our module names should follow path names.

Now we have `source/agora/cli/client/main.d` and the module name `agora.client.main`, which doesn't follow that rule.

I would argue that the `cli` stuff should be moved into `src`, and not within `agora`. Because tools do not form part of the Agora node.

So how about having:

```
source/cli/client/main.d
source/agora/... <node stuff>
```

or something similar to that?

AndrejMitrovic:
The point I'm trying to make is, the agora node code should always be able to import into `agora.*` modules. But now we have tools that start with the `agora` namespace, and agora code should never import into tooling software code, on purpose or on mistake.

Geod24:
Oops, I somehow missed that. Yeah that is not something desireable.

Your approach makes sense. The only downside I can see is that the client cannot access `package(agora)` symbols but that might not be an issue.

AndrejMitrovic:
Ah right, in case it needs to generate Blocks, for example. But we should probably then move the data definitions elsewhere at some point.

PR #592, opened 2020-02-27
Build single-binary networks

### LINKED ISSUE #510: Build single-binary networks

MichaelKim20:
1. Add CLI named agora-launcher for a test. This runs the node by the configuration file.
2. Add the system integration test module with a single binary.

**for test**
rdmd ci/single_binary_test.d

Relates to #510

### DIFF
--- a/ci/single_binary_test.d
+++ b/ci/single_binary_test.d
@@ -1,0 +1,75 @@
+#!/usr/bin/env rdmd
+module system_binary_test;
+
+private:
+
+import std.algorithm;
+import std.file;
+import std.format;
+import std.path;
+import std.process;
+import std.stdio;
+import core.sys.posix.signal;
+
+/// Root of the repository
+immutable RootPath = __FILE_FULL_PATH__.dirName.dirName;
+immutable IntegrationPath = RootPath.buildPath("tests").buildPath("system");
+
+/+ ***************************** Commands to run **************************** +/
+/// A simple test to ensure that the container works correctly,
+/// e.g. all dependencies are installed and the binary isn't corrupt.
+immutable BuildBinary = [ "dub", "build", "-c", "multi", ];
+immutable StartNodes = [ "build/agora-multi", IntegrationPath, "node/0/config.yaml", "node/1/config.yaml", "node/2/config.yaml"];
+immutable RunIntegrationTests = [ "dub", "--root", IntegrationPath, "--",
+                                    "http://127.0.0.1:2826",
+                                    "http://127.0.0.1:3826",
+                                    "http://127.0.0.1:4826"];
+immutable Cleanup = [ "rm", "-rf", IntegrationPath.buildPath("node/0/.cache/"),
+                      IntegrationPath.buildPath("node/1/.cache/"),
+                      IntegrationPath.buildPath("node/2/.cache/")];
+
+private int main (string[] args)
+{
+    // If the user pass `nobuild` as first argument, skip build,
+    // Build `agora-multi`
+    if (args.length < 2 || args[1] != "nobuild")
+        runCmd(BuildBinary);
+
+    // First make sure that there we start from a clean slate,
+    runCmd(Cleanup);
+
+    // Start Nodes
+    auto binary_pid = runCmdNoBlock(StartNodes);
+
+    // Run tests
+    runCmd(RunIntegrationTests);
+
+    // Shut down network cleanly
+    kill(binary_pid, SIGTERM);
+    if (auto retval = binary_pid.wait())
+        assert(retval == -SIGTERM, format("Shutting down nodes failed  with code %d!", retval));
+    return 0;
+}
+
+/// Utility function to run a command and throw on error
+private void runCmd (const string[] cmd)
+{
+    writeln(cmd);
+    auto pid = spawnProcess(cmd);
+    if (pid.wait() != 0)
+        throw new Exception(format("Command failed: %s", cmd));
+}
+
+/// Utility function to run a command
+private Pid runCmdNoBlock (const string[] cmd)
+{
+    writeln(cmd);
+    return spawnProcess(cmd);
+}
+
+/// Utility function
+private int errorOut (ProcessPipes pp)
+{
+    pp.stderr.byLine.each!(a => writeln("[fatal]\t", a));
+    return 1;
+}
--- a/dub.json
+++ b/dub.json
@@ -43,7 +43,7 @@
             "name": "server",
             "targetName": "agora",
             "mainSourceFile": "source/agora/node/main.d",
-            "excludedSourceFiles": [ "source/agora/cli/client/*" ]
+            "excludedSourceFiles": ["source/agora/cli/*/*"]
         },
         {
             "name": "client",
@@ -56,16 +56,26 @@
                 "source/agora/node/*",
                 "source/scpp/*",
                 "source/scpd/*",
-                "source/agora/consensus/protocol/*"
+                "source/agora/consensus/protocol/*",
+                "source/agora/cli/multi/*"
             ]
         },
         {
             "name": "unittest",
             "targetName": "agora-unittests",
-            "excludedSourceFiles": [ "source/agora/cli/client/main.d" ],
+            "excludedSourceFiles": ["source/agora/cli/*/main.d"],
             "sourceFiles": [
                 "source/scpp/build/DSizeChecks.o",
                 "source/scpp/build/DLayoutChecks.o"
+            ]
+        },
+        {
+            "name": "multi",
+            "targetName": "agora-multi",
+            "mainSourceFile": "source/agora/cli/multi/main.d",
+            "excludedSourceFiles": [
+                "source/agora/cli/client/*",
+                "source/agora/node/main.d"
             ]
         }
     ],
--- a/source/agora/cli/multi/main.d
+++ b/source/agora/cli/multi/main.d
@@ -1,0 +1,150 @@
+/*******************************************************************************
+
+    Entry point for running multiple agora nodes within the same binary
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.cli.multi.main;
+
+import agora.api.FullNode;
+import agora.common.Config;
+import agora.node.Node;
+import agora.utils.Log;
+
+import vibe.core.core;
+import vibe.web.rest;
+import vibe.inet.url;
+
+import std.exception;
+import std.file;
+import std.format;
+import std.path;
+import std.range;
+import std.stdio;
+import std.string;
+
+
+/*******************************************************************************
+
+    Application entry point
+
+*******************************************************************************/
+
+private int main (string[] args)
+{
+    // Validate parameters
+    if (args.length < 2)
+    {
+        printDefaultHelp();
+        return 1;
+    }
+
+    string[] files;
+    // If the root path is not set,
+    if (extension(args[1]).toLower == ".yaml")
+        files = args[1..$];
+    else
+    // If the root path is set,
+        foreach (arg; args[2 .. $])
+            files ~= buildNormalizedPath(args[1], arg);
+
+    // Reads all configurations from the file.
+    Config[] configs;
+    foreach (idx, file; files)
+    {
+        try
+        {
+            // Read
+            auto config = readFromFile(file);
+
+            // Make sure that where the configuration file is located is the root directory of the node.
+            // The data directory on the node has been changed to an absolute path.
+            config.node.data_dir = buildNormalizedPath(dirName(file), config.node.data_dir);
+
+            // Multiple nodes simultaneously output logs to the console.
+            // To reduce complexity, only errors are printed.
+            // Other improvements are needed in the future.
+            config.logging.log_level = LogLevel.Error;
+
+            // Change the network address of the nodes.
+            string[] converted_network;
+            foreach (address; config.network)
+            {
+                auto node_url = URL(address);
+                node_url.host = "127.0.0.1";
+                converted_network ~= node_url.toString;
+            }
+
+            Config converted_config =
+            {
+                banman : config.banman,
+                node : config.node,
+                network : assumeUnique(converted_network),
+                dns_seeds : config.dns_seeds,
+                quorum : config.quorum,
+                logging: config.logging,
+                admin: config.admin,
+            };
+
+            // Add
+            configs ~= converted_config;
+        }
+        catch (Exception ex)
+        {
+            writefln("Failed to parse config file '%s'. Error: %s",
+                file, ex.message);
+            return 1;
+        }
+    }
+
+    Node[] nodes;
+    foreach (const ref config; configs)
+        nodes ~= runNode(config);
+
+    scope (exit)
+    {
+        foreach (node; nodes)
+            node.shutdown();
+    }
+
+    return runEventLoop();
+}
+
+
+/*******************************************************************************
+
+    Print out help for this process.
+
+*******************************************************************************/
+
+private void printDefaultHelp ()
+{
+    writeln("usage: agora-multi [path] node0.yaml node1.yaml node2.yaml ...");
+}
+
+
+/*******************************************************************************
+
+    Read config from file
+
+    Params:
+      filename = A file name of configurationn of the node
+
+    Returns:
+      The configuration objects that are used through the node
+
+*******************************************************************************/
+
+private Config readFromFile (string filename)
+{
+    CommandLine cmdln;
+    cmdln.config_path = filename;
+    return parseConfigFile(cmdln);
+}


MichaelKim20:
@Geod24 This is sometimes an error. I'm looking for the cause. Perhaps, there is a communication problem between nodes. Maybe it is because the server socket used before has not been completely removed.

ci/single_binary_test.d
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env rdmd
+module system_integration_test;
```
Geod24: Wrong name

ci/single_binary_test.d
```diff
@@ -0,0 +1,67 @@
...
+
+    return 0;
+}
+
+/// Utility function to run a command and throw on error
+private void runCmd (const string[] cmd, bool isWait = true)
+{
+    writeln(cmd);
+    auto pid = spawnProcess(cmd);
+    if (isWait)
```
Geod24: You should always wait for a command to finish. If not here, at least return the Pid and do it in the main

ci/single_binary_test.d
```diff
@@ -0,0 +1,67 @@
...
+private int main (string[] args)
+{
+    runCmd(BinaryDown, false);
+
+    runCmd(BuildBinary);
+
+    // First make sure that there we start from a clean slate,
+    // as the docker-compose bind volumes
+    runCmd(Cleanup);
+    Thread.sleep(1.seconds);
```
Geod24: You shouldn't need to `sleep`, if wait is called correctly.

ci/single_binary_test.d
```diff
@@ -0,0 +1,67 @@
...
+immutable RootPath = __FILE_FULL_PATH__.dirName.dirName;
+immutable IntegrationPath = RootPath.buildPath("tests").buildPath("system");
+immutable LauncherPath = RootPath.buildPath("tests").buildPath("launcher");
+
+/+ ***************************** Commands to run **************************** +/
+/// A simple test to ensure that the container works correctly,
+/// e.g. all dependencies are installed and the binary isn't corrupt.
+immutable BuildBinary = [ "dub", "build", "-c", "launcher", ];
+immutable BinaryUp = [ "build/agora-launcher", "-c", LauncherPath.buildPath("configs.yaml"), ];
+immutable BinaryDown = [ "killall", "-9", "agora-launcher", ];
```
Geod24: That is not an option. We should make sure we got a clean finish

MichaelKim20: I will add a command API to agora-launcher.

tests/launcher/configs.yaml
```diff
@@ -0,0 +1,9 @@
+################################################################################
+##                             Config                                         ##
+################################################################################
+config:
+    root_path: "tests/launcher/node/"
+    files:
+        - "0/config.yaml"
+        - "1/config.yaml"
+        - "2/config.yaml"
```
Geod24: I'm a bit on the fence about this. I imagined it would just be a matter of doing `./agora-multi file1.yaml file2.yaml file3.yaml` and voila. Do you think there's a need for a specific, extra configuration ?

Geod24:
Do we use `cwd` anywhere ? You might want to abstract it away and just give them all a different `cwd`.

tests/launcher/configs.yaml
MichaelKim20: Yes, I'll think about this more.

ci/single_binary_test.d
MichaelKim20: I solved it with ctrl+c.

ci/single_binary_test.d
MichaelKim20: I removed `sleep`.

tests/launcher/configs.yaml
MichaelKim20: Now I know what you mean. That's a good idea. It will be much easier to use. I'll fix it.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,251 @@
...
+
+    Params:
+      address = An address of the node
+
+*******************************************************************************/
+
+private void testConection (string address)
+{
+    import core.time;
+    runTask({
```
Geod24: If you wrap the whole function in `runTask`, just take it out of `testConnection` and put it in the caller:
```D
runTask(&testConnection, address);
```

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,251 @@
...
+        }
+    }
+
+    scope (exit)
+    {
+        foreach (ref elem; launchers)
+            if (elem !is null)
+                elem.stop();
+
+        exitEventLoop();
```
Geod24: If you reach this, in `scope (exit)`, it means you have already exited the event loop.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,251 @@
...
+            auto config = parseConfigFile(cmdln);
+
+            auto launcher = new NodeLauncher();
+
+            //launcher.startThread(config);
+            launcher.startTask(config);
+
+            launchers ~= launcher;
+
+            testConection(format("http://%s:%d", config.node.address, config.node.port));
```
Geod24: Why are we spawning the nodes and the client in the same process ? The issue is only about spawning clients, not nodes.

MichaelKim20: I used it for testing purposes.
However, if the server port does not open, it outputs an error.
So can we get rid of `testConnection`?

Geod24: > So can we get rid of `testConnection`?

You should be able to use the client part of `test/system/` with a single-binary network.

MichaelKim20: **In the case of starting the node**
`agora-multi start node00.yaml node01.yaml node02.yaml`

**In the case of connecting to the node**
`agora-multi test node00.yaml node01.yaml node02.yaml`

MichaelKim20: Yes, I just added it

Geod24: What's the purpose of the `test` command ? That is not the issue.
The idea of `agora-multi` is to spawn a full network in one binary. The testing should still be separately done, by a client that may request one or more node, but it's orthogonal.

MichaelKim20: Test commands allow the server to wait until it runs normally.

Geod24: To shut down a node, we just send it a `SIGTERM`. That's the standard way to do it on POSIX.

MichaelKim20: I used two SIGTERM and SIGKILL, but I deleted SIGKILL now.

Geod24:
@Michael: I was thinking that you could reuse https://github.com/bpfkorea/agora/blob/500916ba63e51b5f13e7d3a8de44c0c61ddedd63/source/agora/node/main.d#L92-L122

It seems that you copied it more or less in two places:
- https://github.com/bpfkorea/agora/pull/592/files#diff-cac8a0a88f3be1f3afd86eb338f36e9cR252-R282
- https://github.com/bpfkorea/agora/pull/592/files#diff-cac8a0a88f3be1f3afd86eb338f36e9cR300-R328

Also I'm not sure why we would have `UseTask` and `UseThread` ?

Geod24:
Regarding the config files, we should be able to avoid duplication by reusing the `system` one but changing the ports.

MichaelKim20:
Ready for review

Geod24:
@MukeunKim : What about https://github.com/bpfkorea/agora/pull/592#issuecomment-593242795 ?

MichaelKim20:
> lso I'm not sure why we would have `UseTask` and `UseThread` ?

The real node operating environment is more like Thread. So I put it in there. I don't need it for the test, so I'll erase it.

MichaelKim20:
@Geod24 Hosts and ports are recorded in 'tests/systems/source/main.d'
I had no choice but to forward the port to use the above file in both docker and agora-multi.
Is there any other way?

MichaelKim20:
@Geod24 Ready for review

source/agora/cli/multi/main.d
```diff
@@ -138,7 +148,10 @@ private int main (string[] args)
 
 private void printDefaultHelp ()
 {
-    writeln("usage: agora-multi [path] node0.yaml node1.yaml node2.yaml ...");
+    writeln("usage: agora-multi command [path] node0.yaml node1.yaml node2.yaml ...");
+    writeln("These are commands:");
+    writeln("   start : Starts the nodes");
+    writeln("   test : Test connections to the nodes");
```
TrustHenry: I think you can do it generally with `Log.info`

MichaelKim20:
@AndrejMitrovic Ready for review~

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,280 @@
...
+    {
+        auto launcher = new NodeLauncher();
+        launcher.start(config);
+        launchers ~= launcher;
+    }
+
+    scope (exit)
+    {
+        foreach (ref elem; launchers)
+            if (elem !is null)
```
AndrejMitrovic: This doesn't seem necessary, `elem` can't be null here.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,280 @@
...
+    foreach (ref config; configs)
+    {
+        auto launcher = new NodeLauncher();
+        launcher.start(config);
+        launchers ~= launcher;
+    }
+
+    scope (exit)
+    {
+        foreach (ref elem; launchers)
```
AndrejMitrovic: `ref` is not necessary here because NodeLauncher is a class. Everywhere there is a `NodeLauncher object;`, `object` would be a pointer. It can be passed by value without cost.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,280 @@
...
+    Create and run a node. And handle HTTP requests.
+
+    This class is required for terminating a node, a HTTP server, and a thread
+    at the end of the main thread.
+
+*******************************************************************************/
+
+public class NodeLauncher
+{
+
```
AndrejMitrovic:
```suggestion
```

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,280 @@
...
+    if ((x2 == -1) || (x2 == x1))
+        x2 = source.length;
+
+    if (x1 == -1)
+        x1 = 0;
+    else
+        x1 += 3;
+
+    auto nodeno = (cast(string)source[x1..x2]).split("-");
```
AndrejMitrovic: I don't see why this cast is needed. `source` is already a string.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,280 @@
...
+    if (extension(args[1]).toLower == ".yaml")
+        files = args[1..$];
+    else
+    // If the root path is not set,
+        foreach (idx; 2..args.length)
+            files ~= buildNormalizedPath(args[1], args[idx]);
+
+    // Reads all configurations from the file.
+    Config[] configs;
+    foreach (idx, ref file; files)
```
AndrejMitrovic: Why the `ref`? It's not necessary here.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,280 @@
...
+
+            string[] coverted_network;
+            foreach (address; config.network)
+                coverted_network ~= replaceAddress(address, "127.0.0.1");
+
+            Config coverted_config =
+            {
+                banman : config.banman,
+                node : config.node,
+                network : cast(immutable string[])coverted_network,
```
AndrejMitrovic:
```suggestion
                network : assumeUnique(coverted_network),
```

Also:

`import std.exception`

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,280 @@
...
+        catch (Exception ex)
+        {
+            writefln("Failed to parse config file '%s'. Error: %s",
+                file, ex.message);
+            return 1;
+        }
+    }
+
+    NodeLauncher[] launchers;
+    foreach (ref config; configs)
```
AndrejMitrovic:
```suggestion
    foreach (const ref config; configs)
```

If it's not being modified, we should use `const`. And `ref` is there to avoid copying by value, which can be expensive.

source/agora/cli/multi/main.d
```diff
@@ -111,6 +111,16 @@ private int main (string[] args)
         }
     }
 
+    // In the case of to connect to the node
+    if (args[1] == "test")
+    {
+        foreach (ref config; configs)
```
AndrejMitrovic:
```suggestion
        foreach (const ref config; configs)
```

AndrejMitrovic (changes requested):
Found a few issues.

source/agora/cli/multi/main.d
MichaelKim20: Thank you very much for your opinion.
There is no variety of outputs.
The output is all about help for the input and errors during execution.
When more different types of outputs occur, it would be better to use `Log`.

dub.json
```diff
@@ -43,7 +43,10 @@
             "name": "server",
             "targetName": "agora",
             "mainSourceFile": "source/agora/node/main.d",
-            "excludedSourceFiles": [ "source/agora/cli/client/*" ]
+            "excludedSourceFiles": [
+                "source/agora/cli/client/*",
+                "source/agora/cli/multi/*"
+            ]
```
```suggestion
            "excludedSourceFiles": [ "source/agora/cli/*" ],
```

If it doesn't work, the following:
```suggestion
            "excludedSourceFiles": [ "source/agora/cli/*/*" ],
```

dub.json
```diff
@@ -56,17 +59,29 @@
...
+                "source/agora/cli/multi/*"
             ]
         },
         {
             "name": "unittest",
             "targetName": "agora-unittests",
-            "excludedSourceFiles": [ "source/agora/cli/client/main.d" ],
+            "excludedSourceFiles": [
+                "source/agora/cli/client/main.d"
+            ],
```
```suggestion
            "excludedSourceFiles": [ "source/agora/cli/*/main.d" ],
```

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,279 @@
...
+private immutable ushort ADMIN_BASE_PORT = 5000;
+
+/*******************************************************************************
+
+    Application entry point
+
+*******************************************************************************/
+
+version (unittest) {}
+else
```
Geod24: With the change mentioned to the `dub.json` file, this becomes unnecessary

dub.json
Geod24: Ah I see you have `unittest` in main... Then wait before applying this one.
But in any case, the current diff is superfluous (nothing change but indentation).

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,279 @@
+/*******************************************************************************
+
+    Entry point for the Agora multi
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
```
Geod24:
```suggestion
        Copyright (c) 2020 BOS Platform Foundation Korea
```

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,279 @@
...
+        Create and start a node
+
+        Params:
+            config = A configuration objects of the node
+
+    ***************************************************************************/
+
+    public void start (Config config)
+    {
+        void runNode ()
```
Geod24: As discussed, we should use the `runNode` we already have.

Geod24 (commented):
Nice! Just a few comments.

source/agora/node/RunNode.d
```diff
@@ -0,0 +1,65 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.RunNode;
```
Geod24: Is there a reason it's in own module, as opposed to e.g. `Node` ?

MichaelKim20: To resolve @Geod24's request to reuse `runNode()`, the main.d of agora has been modified.
Wasn't this what you wanted?

MichaelKim20: How about this? `agora.utils.RunNode`

Geod24: Why can't it just be moved to the existing `agora.node.Node` ?

MichaelKim20:
@Geod24 @AndrejMitrovic
To resolve @Geod24's request to reuse `runNode()`, the main.d of agora has been modified.
I ask you to review it.

source/agora/node/RunNode.d
MichaelKim20: Yes, that sounds good, then I'll add it at the end.

Geod24:
You can submit the first commit (about `runNode`) in a separate PR.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,312 @@
...
+        {
+            // Read
+            auto config = readFromFile(file);
+
+            // Convert data to suit the environment.
+            config.node.address = "127.0.0.1";
+            config.node.port = cast(ushort)(NODE_BASE_PORT + idx);
+            config.node.data_dir = buildNormalizedPath(dirName(file), config.node.data_dir);
+            config.admin.port = cast(ushort)(ADMIN_BASE_PORT + idx);
+            config.logging.log_level = LogLevel.Error;
```
Geod24: That might be the problem. Here the port is blindly rewritten without taking consideration what is set in the config file. Why is that ?

MichaelKim20: There's a log of three nodes. It's not easy to recognize, I think we need something new.

Geod24: The comment is more about `port` and `address` than `log_level`.
But in general, we should aim to override as little as possible, because it will surprise the user.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,329 @@
...
+    if (args[1] == "test")
+    {
+        foreach (const ref config; configs)
+            testConection(config.getNodeAddress());
+
+        return 0;
+    }
+
+    // In the case of to start the node
+    NodeLauncher[] launchers;
```
Geod24: This became a bit overkill now :)

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,329 @@
...
+            // Convert data to suit the environment.
+
+            // Change the IP of the server to remove the pop-up window when using the firewall.
+            config.node.address = "127.0.0.1";
+
+            // The configuration file is supposed to run inside the dock.
+            // Ports on nodes should not overlap each other.
+            // Ports 4000, 4001,4002 are used by `tests/system/source/main.d.`
+            // Change the port to match these rules.
+            config.node.port = cast(ushort)(NODE_BASE_PORT + idx);
```
Geod24: Oh I see why you're doing this now!
Let's just change the test to take the addresses as arguments.

MichaelKim20: Then I will modify the `tests/system/source/main.d` to get the addresses as an argument.

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,256 @@
+/*******************************************************************************
+
+    Entry point for the Agora multi
```
AndrejMitrovic:
```suggestion
    Entry point for running multiple agora nodes within the same binary
```

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,256 @@
...
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.cli.multi.main;
+
+import agora.api.FullNode;
+
```
AndrejMitrovic:
```suggestion
```

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,256 @@
...
+        printDefaultHelp();
+        return 1;
+    }
+
+    string[] files;
+    // If the root path is not set,
+    if (extension(args[1]).toLower == ".yaml")
+        files = args[1..$];
+    else
+    // If the root path is not set,
```
AndrejMitrovic: How can both the `if` and the `else` have the same comment?

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,256 @@
...
+        foreach (idx; 2..args.length)
+            files ~= buildNormalizedPath(args[1], args[idx]);
```
AndrejMitrovic:
```suggestion
        foreach (arg; args[2 .. $])
            files ~= buildNormalizedPath(args[1], arg);
```

AndrejMitrovic (changes requested):
Small changes requested.

source/agora/cli/multi/main.d
MichaelKim20: Okay, I'll just shut down the node.

MichaelKim20:
@AndrejMitrovic Thanks for the review.

MichaelKim20:
@Geod24 @AndrejMitrovic Ready for review

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,305 @@
...
+    foreach (idx, file; files)
+    {
+        try
+        {
+            // Read
+            auto config = readFromFile(file);
+
+            // Convert data to suit the environment.
+
+            // Change the IP of the server to remove the pop-up window when using the firewall.
```
Geod24: What is that ??

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,305 @@
...
+            // The data directory on the node has been changed to an absolute path.
+            config.node.data_dir = buildNormalizedPath(dirName(file), config.node.data_dir);
+
+            // Multiple nodes simultaneously output logs to the console.
+            // To reduce complexity, only errors are printed.
+            // Other improvements are needed in the future.
+            config.logging.log_level = LogLevel.Error;
+
+            // Change the network address of the nodes.
+            string[] coverted_network;
```
Geod24:
```suggestion
            string[] converted_network;
```

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,305 @@
...
+            foreach (address; config.network)
+                coverted_network ~= replaceAddress(address, "127.0.0.1");
+
+            Config coverted_config =
```
Geod24:
```suggestion
            Config converted_config =
```

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,305 @@
...
+
+/*******************************************************************************
+
+    Create and run a node. And handle HTTP requests.
+
+    This class is required for terminating a node at the end of the main thread.
+
+*******************************************************************************/
+
+public class NodeLauncher
```
Geod24: We can remove this

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,305 @@
...
+    Params:
+      source = A network address
+      host = A new hostname
+
+    Returns:
+      A network address with changed hostname
+
+*******************************************************************************/
+
+private string replaceAddress(string source, string host)
```
Geod24: Wouldn't it be better to just parse the URL with the Vibe.d class and change the host ?

Geod24: https://vibed.org/api/vibe.inet.url/URL

source/agora/cli/multi/main.d
```diff
@@ -0,0 +1,305 @@
...
+        catch (Exception ex)
+        {
+            writefln("Failed to parse config file '%s'. Error: %s",
+                file, ex.message);
+            return 1;
+        }
+    }
+
+    // In the case of to connect to the node
+    if (args[1] == "test")
```
Geod24: The test option should be removed, it is not part of the issue.

source/agora/cli/multi/main.d
MichaelKim20: @Geod24 I removed NodeLauncher, 
but only one node is running and the rest is not working.

MichaelKim20: @Geod24 I found a way. I'll organize it and then commit it again.

MichaelKim20:
@Geod24 Ready for review

ci/single_binary_test.d
```diff
@@ -0,0 +1,72 @@
...
+
+    // First make sure that there we start from a clean slate,
+    runCmd(Cleanup);
+
+    // Start Nodes
+    auto binary_pid = runCmdNoBlock(StartNodes);
+    scope (exit) kill(binary_pid, SIGTERM);
+
+    // Run tests
+    runCmd(RunIntegrationTests);
```
Geod24: Here `kill` is not enough, because `SIGTERM` can be ignored. It would be better to use `SIGTERM`, then `wait`.

```suggestion
    // Start Nodes
    auto binary_pid = runCmdNoBlock(StartNodes);
    // Run tests
    runCmd(RunIntegrationTests);
    // Shut down network cleanly
    kill(binary_pid, SIGTERM);
    if (auto retval = binary_pid.wait())
        assert(retval == -SIGTERM, format("Shutting down nodes failed  with code %d!", retval));
     return 0;
```

Note that according to the documentation, this will not be a good thing to do on Windows, but we'll deal with this when we get there.

Geod24 (approved):
Just one suggestion, otherwise LGTM

AndrejMitrovic (approved):

Issue #485, opened 2020-01-20
Move compiler HEAD tests into merge-on-master


AndrejMitrovic:

AndrejMitrovic:
See also https://github.community/t5/GitHub-Actions/continue-on-error-allow-failure-UI-indication/td-p/37033

AndrejMitrovic:
Done.

PR #593, opened 2020-02-27
Correctly print compiler's version on push

Geod24:
```
The step was missing on Linux, and the definition on Mac OSX was wrong,
because matrix.dc is 'dmd-nightly' or 'dmd-latest-ci'.
```

See https://github.com/bpfkorea/agora/runs/469345840


TrustHenry:
LGTM
Windows build github actions `github_push.yaml` , `pr.yaml` was testing yamls, and I was thinking, that's weird.

Geod24:
We did it because of https://github.com/bpfkorea/agora/issues/485

PR #594, opened 2020-02-27
BlockStorage: Provide both throwing and non-throwing alternatives of readBlock

Geod24:
```
This should make the code more robust,
as it is easy to oversee checking the return value.
```

It seems that the interface was not consistently `nothrow` (deserialization could throw and propagate from `MemBlockStorage` but not from the "real" one).

### DIFF
--- a/source/agora/node/BlockStorage.d
+++ b/source/agora/node/BlockStorage.d
@@ -89,7 +89,7 @@ public interface IBlockStorage
 
     /***************************************************************************
 
-        Read block from the storage.
+        Attempt to read a block at a specified height from the storage.
 
         Params:
             block = `Block` to read
@@ -100,12 +100,11 @@ public interface IBlockStorage
 
     ***************************************************************************/
 
-    public bool readBlock (ref Block block, size_t height);
-
+    public bool tryReadBlock (ref Block block, size_t height) nothrow;
 
     /***************************************************************************
 
-        Read block from the storage.
+        Attempt to read a block with a specified hash from the storage.
 
         Params:
             block = `Block` to read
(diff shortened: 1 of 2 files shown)


source/agora/node/BlockStorage.d
```diff
@@ -1053,24 +1054,13 @@ public class MemBlockStorage : IBlockStorage
         if (finds.empty)
             return false;
 
-        block = deserializeFull!Block(this.blocks[finds.front.position]);
+        try block = deserializeFull!Block(this.blocks[finds.front.position]);
+        catch (Exception e) return false;
```
AndrejMitrovic, line 1058: Could also do:

```suggestion
        return collectException(deserializeFull!Block(this.blocks[finds.front.position]) is null);
```

Or keep it your way if you think it makes the intention more clear.

AndrejMitrovic:
Looks good to me.

source/agora/node/BlockStorage.d
Geod24: The thing I don't like about those utilities (`enforce`, `collectException`) is that they generally don't play well with the compiler.
E.g. the following won't compile:
```D
string foo (int v)
{
    if (v == 42)
        return "42";
    enforce(false, "Expected the answer to life, universe, and the rest");
}
```

However this will:
```D
string foo (int v)
{
    if (v == 42)
        return "42";
    throw new Exception("Expected the answer to life, universe, and the rest");
}
```

Likewise, `collectException` escaping the reference to the `Exception` means it won't play well with DIP1008.

AndrejMitrovic (approved):

TrustHenry (approved):
LGTM

PR #595, opened 2020-02-27
BlockStorage: Use agora's logger instead of stdio

Geod24:
```
Additionally, an import was moved because `dmd` didn't
pick it up when emitting dependencies for `rdmd`
and thus the tests in `test/unit/` had linker failures.
```

### DIFF
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -201,11 +201,11 @@ public class CircularAppender : Appender
 /// A layout with colored LogLevel
 public class AgoraLayout : Appender.Layout
 {
+    import ocean.time.WallClock;
+
     /// Format the message
     public override void format (LogEvent event, scope FormatterSink dg)
     {
-        import ocean.time.WallClock;
-
         // convert time to field values
         const tm = event.time;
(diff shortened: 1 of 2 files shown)


AndrejMitrovic (approved):

source/agora/utils/Log.d
```diff
@@ -201,11 +201,11 @@ public class CircularAppender : Appender
 /// A layout with colored LogLevel
 public class AgoraLayout : Appender.Layout
 {
+    import ocean.time.WallClock;
```
TrustHenry, line 204: `Use agora's logger instead of stdio` Is it related to the correction?

Geod24: Yes, see the commit message:
```
Additionally, an import was moved because `dmd` didn't
pick it up when emitting dependencies for `rdmd`
and thus the tests in `test/unit/` had linker failures.
```

TrustHenry: Oh, thank you.  I understood.

MichaelKim20 (commented):
LGTM, I like the color, too.

Issue #585, opened 2020-02-24
Expire the role of a validator after the validator cycle ends


linked0:
After a cycle for the role of a validator ends, the data for that validator must be cleared. 

Definition of Done:

- [x] The data for validatator should be deleted from the validator set table

- [x] The variables in EnrollmentManager class should be cleared.

PR #597, opened 2020-02-27
Add functions for removing expired validators and getting validators

linked0:
In order to remove expired validators, the ledger must call the function of setCurrentBlockHeight in the addValidatedBlock function of the ledger like as follows.
`this.enroll_man.setCurrentBlockHeight(this.this.getBlockHeight());`

The getValidators function is also provided in this PR for getting valid validators for the ledger.

Relates #585

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -534,6 +534,78 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Get all the current validators
+
+        Params:
+            validators = will be filled with all the validators during
+                their validation cycles
+
+        Returns:
+            Return true if there is no error in getting validators
+
+    ***************************************************************************/
+
+    public bool getValidators (out Enrollment[] validators) @safe nothrow
+    {
+        try
+        {
+            () @trusted {
+                auto results = this.db.execute("SELECT val FROM validator_set" ~
+                    " WHERE enrolled_height is not null");
+                foreach (row; results)
+                {
+                    validators ~=
+                        deserializeFull!Enrollment(row.peek!(ubyte[])(0));
+                }
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error, exception:{}", ex);
+            return false;
+        }
+
+        return true;
+    }
+
+    /***************************************************************************
+
+        Set current block height
+
+        The enrollment manager can weed out expired validators from the set
+        based on the block height. Validators are deleted if their enrolled
+        height is less than or equal to the value of the passed block height
+        minus the validator cycle.
+
+        Params:
+            block_height = current block height
+
+    ***************************************************************************/
+
+    public void setCurrentBlockHeight (ulong block_height) @safe nothrow
+    {
+        // the smallest enrolled height would be 1, so the passed block height
+        // should be greater than the validator cycle for deleting validators.
+        if (block_height > ValidatorCycle)
+        {
+            try
+            {
+                () @trusted {
+                    this.db.execute("DELETE FROM validator_set " ~
+                        "WHERE enrolled_height <= ?",
+                        block_height - ValidatorCycle);
+
+                }();
+            }
+            catch (Exception ex)
+            {
+                log.error("Database operation error, exception:{}", ex);
+            }
+        }
+    }
+
+    /***************************************************************************
+
         Get the unregistered enrollments in the block
         And this is arranged in ascending order with the utxo_key
 
@@ -1019,6 +1091,33 @@ unittest
     assert(man.needRevealPreimage(10));
     man.increaseNextRevealHeight();
     assert(man.needRevealPreimage(16));
+
+    // test for getting validators
+    Enrollment[] validators;
+
+    // validator A with the `utxo_hash` and the enrolled height of 10.
+    // validator B with the 'utxo_hash2' and the enrolled height of 11.
+    // validator C with the 'utxo_hash3' and no enrolled height.
+    assert(man.updateEnrolledHeight(utxo_hash2, 11));
+
+    // current block height is 1017
+    // So, valid validator are validator A and validator B
+    man.setCurrentBlockHeight(1017);
+    assert(man.getValidators(validators));
+    assert(validators.length == 2);
+
+    // change the block height to 1018, which means validator A is expired.
+    man.setCurrentBlockHeight(1018);
+    assert(man.getValidators(validators));
+    assert(validators[0].utxo_key == utxo_hash2);
+
+    // set an enrolled height for validator C
+    // set the block height to 1019, which means validator B is expired.
+    // there is only one validator in the middle of 1020th block being made.
+    assert(man.updateEnrolledHeight(utxo_hash3, 12));
+    man.setCurrentBlockHeight(1019);
+    assert(man.getValidators(validators));
+    assert(validators[0].utxo_key == utxo_hash3);
 }
 
 /// tests for addPreimage and hasPreimage


source/agora/consensus/EnrollmentManager.d
```diff
@@ -532,6 +532,92 @@ public class EnrollmentManager
         return false;
     }
 
+    /***************************************************************************
+
+        Get all the valid validators
```
Geod24: What's an invalid validator ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -532,6 +532,92 @@ public class EnrollmentManager
...
+            return false;
+        }
+
+        return true;
+    }
+
+    /***************************************************************************
+
+        Set current block height. The enrollment manager can weed out expired
+            validators from the set based on the block height.
```
Geod24: You should break this sentence, this title is too long.
```suggestion
        Set current block height
        
        The enrollment manager can weed out expired validators from the set based on the block height.
```

^ Might need to add a line break at 80 chars, but I wanted to show the spacing required to separate the title from the description (https://dlang.org/spec/ddoc.html#summary)

source/agora/consensus/EnrollmentManager.d
```diff
@@ -532,6 +532,92 @@ public class EnrollmentManager
...
+        {
+            log.error("Database operation error, exception:{}", ex);
+        }
+
+        try
+        {
+            // weed out expired validator from the set
+            foreach (ref key, height; enrolled_heights)
+            {
+                if (block_height >= height + ValidatorCycle)
```
Geod24: I don't quite understand why you would do this over `("DELETE FROM validator_set WHERE enrolled_height <= ?", block_height - ValidatorCycle)` ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1019,6 +1105,35 @@ unittest
...
+    // So, valid validator are validator A and validator B
+    assert(man.updateEnrolledHeight(utxo_hash2, 11));
+    man.setCurrentBlockHeight(1017);
+    assert(man.getValidators(validators));
+    assert(validators.length == 2);
+
+    // Change the block height to 1018, which means validator A is expired.
+    man.setCurrentBlockHeight(1018);
+    assumeSafeAppend(validators);
+    validators.length = 0;
```
Geod24: `out` doesn't do what you think it does. We should have a whiteboarding session about it.

Geod24 (changes requested):

source/agora/consensus/EnrollmentManager.d
linked0: A valid validator means the validator that has valid `enrolled_height` value not null, and has its validator cycle not yet expired. I could find more obvious sentence.

source/agora/consensus/EnrollmentManager.d
linked0: At first, I will find and read a document for `out` and `ref`.

AndrejMitrovic: `out` just means the parameter is initialized to `Type.init` when the function runs. So if you have something like this:

```D
int value = 42;
void foo (out int var)
{
   // same as var = int.init; here
    ...
}

foo(value);
```

It will initialize `value` to `int.init`, which is `0` in this case.

source/agora/consensus/EnrollmentManager.d
linked0: Your suggestion is quite simple. I was not able to realize that the SQL statement can be made to the extent.

source/agora/consensus/EnrollmentManager.d
Geod24: >  and has its validator cycle not yet expired

The code and the documentation disagree then. Because the code only checks that `enrolled_height` is not null.

linked0: > The code and the documentation disagree then. Because the code only checks that enrolled_height is not null.

You're right but it is already checked whether the validator cycle has been expired in the `setCurrentBlockHeight` function. I have just changed the sentence to 'Get all the current validators'.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -532,6 +532,71 @@ public class EnrollmentManager
...
+
+    ***************************************************************************/
+
+    public void setCurrentBlockHeight (ulong block_height) @safe nothrow
+    {
+        try
+        {
+            () @trusted {
+                this.db.execute("DELETE FROM validator_set " ~
+                    "WHERE enrolled_height <= ?", block_height - ValidatorCycle);
```
Geod24: What happens when we call this with `block_height == 1` ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -532,6 +532,71 @@ public class EnrollmentManager
...
+
+        return true;
+    }
+
+    /***************************************************************************
+
+        Set current block height
+
+        The enrollment manager can weed out expired validators from the set
+        based on the block height.
```
Geod24: It would be good to mention that the delete is inclusive (`<=`)

source/agora/consensus/EnrollmentManager.d
```diff
@@ -532,6 +532,71 @@ public class EnrollmentManager
...
+        try
+        {
+            () @trusted {
+                auto results = this.db.execute("SELECT val FROM validator_set" ~
+                    " WHERE enrolled_height is not null");
+                foreach (row; results)
+                {
+                    auto enroll =
+                        deserializeFull!Enrollment(row.peek!(ubyte[])(0));
+                    validators ~= enroll;
```
```suggestion
                    validators ~= deserializeFull!Enrollment(row.peek!(ubyte[])(0));
```

source/agora/consensus/EnrollmentManager.d
linked0: > What happens when we call this with block_height == 1 ?

There seems no problem but I will add unit test code for that and check it.

Geod24: Really ? It substract a large unsigned value to a smaller unsigned value.
So either we get negative number, or we wrap-around and get a very high number.
In any case, not something that should happen.

linked0: > So either we get negative number, or we wrap-around and get a very high number.

It passed new unit test code as follows.
```
man.setCurrentBlockHeight(1);
assert(man.getValidators(validators));
assert(validators.length == 2);
```
I guess the sqlite get the very big number as small singed integer by the mechanism of the two's complement. But for the code to be safer, I will add the code checking the block_height.

source/agora/consensus/EnrollmentManager.d
MichaelKim20: When can the `enroled_height` have a null value?

MichaelKim20: I see. It's set when `updateEnrolledHeight()` is running.

linked0:
Ready for review

source/agora/consensus/EnrollmentManager.d
```diff
@@ -532,6 +532,78 @@ public class EnrollmentManager
...
+        Params:
+            block_height = current block height
+
+    ***************************************************************************/
+
+    public void setCurrentBlockHeight (ulong block_height) @safe nothrow
+    {
+        // the smallest enrolled height would be 1, so the passed block height
+        // should be greater than the validator cycle for deleting validators.
+        if (block_height > ValidatorCycle)
```
Geod24, line 589: A style preference: Try to always return early. If you have an `if` that covers most of the function, it's a red flag.
Instead one can write:
```D
if (block_heigth <= ValidatorCycle)
    return;
```

This way you get the error and odd cases out of the way, and it makes it easier for the reader to focus on your code.
But since it's just a style preference, let's not block the merge on this.

Geod24 (approved):

PR #598, opened 2020-02-27
BlockStorage test: Do not create the test folder in cwd

Geod24:
```
Avoids clashing with whatever the user is doing (e.g. a node could be running).
Instead use the temporary directory, but make sure we only ever write at
one unique location.
```

Yeah, no more corrupting `.cache` when doing `dmd -run ./tests/runner.d`.

Regarding the module, I have in mind to add something for the `unittest` which provides a testing env (instantiate a ledger, perhaps?). But in the meantime I didn't have a good place to put this function.

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -1,0 +1,64 @@
+/*******************************************************************************
+
+    Various utilities for testing purpose
+
+    Utilities in this module can be used in test code.
+    There are currently multiple testing approaches:
+    - Unittests in the various `agora` module, the most common, cheapest,
+      and a way to do white box testing;
+    - Unittests under `agora.test`: Those unittests rely on the LocalRest
+      library to simulate a network where nodes are thread who communicate
+      via message passing.
+    - Unit integration tests in `${ROOT}/tests/unit/` which are similar to
+      unittests but provide a way to test IO-using code.
+    - System integration tests: those are fully fledged tests that spawns
+      unmodified, real nodes within Docker containers and act as a client.
+
+    Any symbol in this module can be used by any of those method,
+    which is why this module is neither restricted by `package(agora):`
+    nor `version(unittest):`.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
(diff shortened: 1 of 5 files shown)


source/agora/utils/Test.d
```diff
@@ -0,0 +1,64 @@
...
+
+    Params:
+      postfix = A unique postfix for the calling test
+
+    Returns:
+      The path of a clean, empty directory
+
+*******************************************************************************/
+
+public string makeCleanTempDir (string postfix = __MODULE__)
```
Geod24, line 55: Probably the most documented function in agora.

@bpfkorea/core : There is a neat trick here. Default parameters are evaluated at the call site. So when this function is called without argument, the module will be `"unit.BlockStorage"`, `"unit.BanManager"`, etc...

If the default parameter was evaluated in this module, the result would always be `"agora.utils.Test"`...
This trick is used in many places, most commonly with `__LINE__`, `__FILE__`, and of course `__MODULE__`.

source/agora/utils/Test.d
```diff
@@ -0,0 +1,64 @@
...
+      unittests but provide a way to test IO-using code.
+    - System integration tests: those are fully fledged tests that spawns
+      unmodified, real nodes within Docker containers and act as a client.
+
+    Any symbol in this module can be used by any of those method,
+    which is why this module is neither restricted by `package(agora):`
+    nor `version(unittest):`.
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
```
AndrejMitrovic: Not to be Joe, but I'm not sure you can place a copyright year for a time the file didn't exist yet. I think it's based on the file creation date, and not the company creation date?

AndrejMitrovic (approved):

source/agora/utils/Test.d
Geod24: I was thinking that code that is moved should probably keep the right copyright year.
But this isn't moved :D

Geod24:
Fixed copyright year

AndrejMitrovic (approved):

PR #601, opened 2020-02-28
Fix #600: Remove dependency on ocean.core.BitArray

### LINKED ISSUE #600: Get rid of ocean's BitArray in BitField
Ocean's `BitArray` is just a pain to work with, and a thorn in our side.
It lacks attribute, duplicates most of what `BitField` is doing, and doesn't expose proper `const`-correct constructors.

Geod24:
This is a simpler implementation of BitArray, which works better with the (de)serializer,
has less overhead, is easier to work with, has a configurable granularity (type),
and proper attributes.

### DIFF
--- a/source/agora/common/BitField.d
+++ b/source/agora/common/BitField.d
@@ -1,14 +1,25 @@
 /*******************************************************************************
 
-    Contains a BitField implementation based on Ocean's BitArray struct.
+    Contains a templated BitField implementation
 
-    The choice of Ocean's over Phobos' implementation of BitArray is
-    because the Phobos implementation requires the backing store length to be
-    a multiple of size_t, which makes it platform-specific and hard to code for.
-    This structure cannot be platform-specific as it's stored in the blockchain.
+    This module implement a clean room bit field type, with a configurable
+    backing store type.
+
+    The reason to provide a configurable type is to allow users to control
+    exactly the size (and alignment) of their bit fields,
+    which is needed when the data type is serialized, either for network
+    communication or blockchain storage.
+
+    This also explains why we cannot rely on Phobos' own bitfield type,
+    as it is backed by `size_t`.
+
+    Note:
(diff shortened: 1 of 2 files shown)


source/agora/consensus/data/Block.d
```diff
@@ -51,7 +51,7 @@ public struct BlockHeader
     public Hash merkle_root;
 
     /// Bitfield containing the validators' key indices which signed the block
-    public BitField validators;
+    public BitField!(uint) validators;
```
AndrejMitrovic:
```suggestion
    public BitField!uint validators;
```

Prettier. 👀

source/agora/common/BitField.d
```diff
@@ -23,117 +34,173 @@ import agora.common.Hash;
...
+    {
+        if (index >= this.length())
+            assert(0);
+        return !!(this._storage[index / BitsPerT] & mask(index));
     }
 
-    /// Hashing support
-    public void computeHash (scope HashDg dg) const nothrow @nogc
+    /// Compare BitField of same length
+    public bool opEquals (OtherT) (const auto ref BitField!OtherT other) const
```
AndrejMitrovic, line 160: I think there is a unittest missing for comparing two different backing-store Bitfield types?

AndrejMitrovic (changes requested):
Some notes.

source/agora/common/BitField.d
Geod24: Line 57 (and after) has it. `bf3` uses `ubyte` as backing store, `bf2` uses `uint`.

source/agora/consensus/data/Block.d
Geod24: Done

Geod24:
💚

AndrejMitrovic (approved):

PR #602, opened 2020-02-28
Make QuorumConfig fields non-immutable

AndrejMitrovic:
It is only the Config class that needs immutable fields,
due to being shared accross threads in our test-suite.

An immutable QuorumConfig can easily be constructed,
but forcing the fields to always be immutable makes
it impossible to use iterative building of quorums.

Required for #240

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -82,7 +82,7 @@ public struct Config
     public immutable string[] dns_seeds;
 
     /// The quorum config
-    public QuorumConfig quorum;
+    public immutable QuorumConfig quorum;
 
     /// Logging config
     public LoggingConfig logging;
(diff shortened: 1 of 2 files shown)


source/agora/common/Config.d
```diff
@@ -488,7 +485,7 @@ unittest
                  PublicKey.fromString("GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5"),
                  PublicKey.fromString("GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5")])])]);
 
-    assert(quorum == expected);
+    assert(cast()quorum == expected);
```
Geod24: I guess you can't make `expected` immutable ?

Geod24 (approved):
Looks neat, just a nit

source/agora/common/Config.d
AndrejMitrovic: Doh, of course.

AndrejMitrovic:
Fixed.

Geod24 (approved):
Merge at will

TrustHenry (approved):
Then we can make the iterative building of quorum.
looks Good to me

PR #603, opened 2020-02-28
Support deserialization of `const` and `immutable` types

Geod24:
Here we are.
There are two things left to do after this:
- Replace `deserialize` with `T.fromBinary!T(dg)` and return new instances;
- Unify `deserializePart` / `deserializeFull` into two overloads that take either a delegate or a `ubyte[]` as first argument, and `CompactMode` as second argument

But since we can now deserialize `immutable(Block)` it is pretty neat.

### DIFF
--- a/source/agora/common/Deserializer.d
+++ b/source/agora/common/Deserializer.d
@@ -16,6 +16,7 @@ module agora.common.Deserializer;
 import agora.common.Types;
 import agora.common.crypto.Key;
 
+import std.algorithm;
 import std.range;
 import std.traits;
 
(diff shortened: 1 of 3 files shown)


source/agora/common/Deserializer.d
```diff
@@ -143,95 +144,103 @@ public T deserializeFull (T) (scope ubyte[] data) @safe
 }
 
 /// Ditto
-public T deserializeFull (T) (scope DeserializeDg dg) @safe
+public void deserializePart (T) (ref T record, scope DeserializeDg dg) @safe
```
AndrejMitrovic, line 147: I was slightly confused at the diff at first, but then I realized that you've changed the order of the functions too.

Geod24: I kept `deserializePart` as a shell for `deserializeFull`, since most `deserialize` functions defined in `struct` use it and I didn't want to diff to be too big.

source/agora/common/Deserializer.d
```diff
@@ -144,95 +144,103 @@ public T deserializeFull (T) (scope ubyte[] data) @safe
...
     @safe
 {
     import geod24.bitblob;
 
     // Custom deserialization trumps everything
     static if (hasDeserializeMethod!T)
-        record.deserialize(dg);
+    {
+        T retval = T.init;
+        retval.deserialize(dg);
```
AndrejMitrovic, line 163: Aha, so this part will also need fixing later, like mentioned in the PR comment. Making it a static method so it can return immutable.

AndrejMitrovic (approved):
LGTM.

source/agora/common/Deserializer.d
Geod24: The signature will be:
```D
struct Foo
{
    static T fromBinary(T) (scope DeserializationDg dg) @safe
     {
         // Magic
     }
}
```

Unfortunately the template is necessary because there's no way to call different function based on whether `T` is `const`, `immutable`, or not.

Geod24:
By the way, using `return` means that NRVO (should) get triggered, and it should also be as efficient.

AndrejMitrovic:
@bpfkorea/core this is what NRVO refers to: https://stackoverflow.com/questions/12953127/what-are-copy-elision-and-return-value-optimization#:~:text=(Named)%20Return%20value%20optimization%20is,since%20the%20object%20is%20named.

It's a form of copy elision, or avoiding making unnecessary copying of data.

PR #607, opened 2020-02-29
Remove redundant deserialize methods

Geod24:
Now that the deserializer is a lot smarter, we don't need all that code.
Also added tests for things which lacked them.

Note that currently, `SecretKey`, `Seed`, and `KeyPair` are not serialized, only `PublicKey`.
Nonetheless, added a test for it just in case. In the future we can probably make it smarter (e.g. for `KeyPair` one just have to serialize the seed...)

### DIFF
--- a/source/agora/common/Deserializer.d
+++ b/source/agora/common/Deserializer.d
@@ -30,8 +30,7 @@ unittest
     import agora.consensus.Genesis;
 
     ubyte[] block_bytes = serializeFull(GenesisBlock);
-    // TODO: This trigger a DMD bug about array comparison
-    assert(cast(const)deserializeFull!Block(block_bytes) == GenesisBlock);
+    assert(deserializeFull!(const(Block))(block_bytes) == GenesisBlock);
 
     // Check that there is no trailing data
     ubyte[] blocks_data = serializeFull(GenesisBlock) ~ serializeFull(GenesisBlock);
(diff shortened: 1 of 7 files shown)


TrustHenry:
https://github.com/bpfkorea/agora/blob/acbd5464cd735d6c62910d49e2bac7346926ceb0/source/agora/common/BanManager.d#L103

https://github.com/bpfkorea/agora/blob/acbd5464cd735d6c62910d49e2bac7346926ceb0/source/agora/common/BanManager.d#L295

Is this part 'BanManager' adaptable?

Geod24:
Good catch! For the moment it is not, because it serializes an  associative array: https://github.com/bpfkorea/agora/blob/acbd5464cd735d6c62910d49e2bac7346926ceb0/source/agora/common/BanManager.d#L61

And we don't support those at the moment in the (de)serializer (but will soon, I have a commit for it).

TrustHenry (approved):
LGTM

MichaelKim20:
I think you should combine Commit `2acca16` and `a976afa`.

Geod24:
> I think you should combine Commit `2acca16` and `a976afa`.

Good catch! Just did.

MichaelKim20 (approved):
LGTM

TrustHenry:
Travis CI error occurred.

Geod24:
I re-triggered it since it is spurious.

PR #608, opened 2020-02-29
Fold all serialization routines into one

Geod24:
Originally, both deserialization and serializer had multiple overloads
that each handled a specific type.
However, as the types being covered expanded, and usage of serialization grew
(it was originally intended to serialize types on which we had full control,
e.g. blocks, but will now be used for inter-node communication),
this approache did not scale: to provide any kind of generic overload,
e.g. for arrays, it is necessary to exclude the special case handled by
other overloads, for example the `string` or `ubyte[]` specialization.
The approach of merging overloads was already taken in the Deserializer.

### DIFF
--- a/source/agora/common/Serializer.d
+++ b/source/agora/common/Serializer.d
@@ -14,7 +14,9 @@
 module agora.common.Serializer;
 
 import agora.common.Types;
+
 import std.range.primitives;
+import std.traits;
 
 ///
 unittest
@@ -78,15 +80,6 @@ public ubyte[] serializeFull (T) (scope const auto ref T record)
     return res;
 }
 
-/// Ditto
-public void serializePart (T) (scope const auto ref T record, scope SerializeDg dg)
-    if (!hasSerializeMethod!T && isInputRange!T && hasLength!T)
-{
-    serializePart(record.length, dg);
-    foreach (ref v; record)
-        serializePart(v, dg);
-}
-
 ///
 unittest
 {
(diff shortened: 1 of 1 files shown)


source/agora/common/Serializer.d
```diff
@@ -117,87 +110,72 @@ unittest
...
 
-/// Ditto
-public void serializePart (scope const(Hash)[] record, scope SerializeDg dg)
-    @trusted
-{
-    serializePart(record.length, dg);
-    foreach (hash; record)
-        serializePart(hash, dg);
+    else
+        static assert(0, "Unhandled type: " ~ T.stringof);
```
TrustHenry, line 178: I'm wondering if it's right to cause an assert() error here.

Is this the point where you can attack an unacceptable type?

Geod24: The assert is static, which means it will trigger at compile time and you will know you need to fix it.

For example at the moment if you do:
```
    int[string] aa;
    serializeFull(aa);
```

You will see:
```console
agora ~serializer-improvements: building configuration "unittest"...
Running pre-build commands...
Target /Users/geod24/projects/bpfk/agora/source/scpp/build/libscp.o is up to date, nothing to do
source/agora/common/Serializer.d(181,9): Error: static assert:  "Unhandled type: const(int)[string]"
source/agora/common/Serializer.d(82,18):        instantiated from here: serializePart!(const(int)[string])
source/agora/common/Serializer.d(49,18):        instantiated from here: serializeFull!(int[string])
```

So you know the type can't be serialized and you need to add support to the serializer.

TrustHenry: Thank you for your detailed explanation.

TrustHenry (approved):
LGTM

MichaelKim20:
@Geod24 @TrustHenry  When the data type is 'ucent', there will be a problem.

PR #610, opened 2020-03-02
Ellide some copy of C++ structs

Geod24:
Pretty trivial, avoid postblitting the struct when we don't have to.

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -623,7 +623,7 @@ public SCPQuorumSet toSCPQuorumSet ( in QuorumConfig quorum_conf )
 
 *******************************************************************************/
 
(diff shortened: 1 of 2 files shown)


linked0 (approved):
Trivial change so LGTM.
But I'm curious about the difference `in` and `const ref` keywords in parameter list of the `toQuorumConfig` function  because thier meaning looks same.

TrustHenry (approved):
LGTM

Geod24:
`in` is just `const`.  Supposed to be `scope` as well but it isn't at the moment.
`ref` obviously means no copy happening. So `in ref` would also be a good substitute.
Personally I would love for `in` to mean `const ref scope` but unfortunately it does not :(

PR #617, opened 2020-03-04
Change API for 'hasPreimage' to 'getPreimage'

### LINKED ISSUE #566: Move `hasPreimage` / `getPreimage` from Validator API to Full Node API
Related: https://github.com/bpfkorea/agora/issues/567

Currently `hasPreimage` (to be changed to `getPreimage`) is part of Validator's API.
It should be part of the full node API, as preimage need to be known to full node in order to verify signature (and it also increases data availability for them).

### LINKED ISSUE #567: Change API for 'hasPreimage' to 'getPreimage'
Currently, `hasPreimage` (defined in Validator's API) has a rather inconvenient interface: It ask if a node knows about a `preImage` for a given `utxo_key` at a given `height`.

Instead, a much more useful API would be to ask for the `PreImageInfo` for a given `utxo_key` and just validate the data.

linked0:
The getPreimage API is created into the FullNode API and the hasPreimage API is removed. 

In EnrollmentManager, there are two functions getting a pre-image information for a validator or a node, which is confusing now but will be handled in bpfkorea#604 issue.

For unit test, the broadcastPreimage API is added in TestAPI.

Fixes #567
Fixes #566

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -17,6 +17,7 @@ module agora.api.FullNode;
 
 import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
+import agora.consensus.data.PreimageInfo;
 import agora.common.Types;
 import agora.common.Set;
 import agora.common.Serializer;
@@ -184,4 +185,23 @@ public interface API
 
     @method(HTTPMethod.GET)
     public bool hasEnrollment (Hash enroll_hash);
+
+    /***************************************************************************
+
+        Get validator's pre-image inforamtion
+
+        API:
+            GET /get_preimage
+
+        Params:
+            enroll_key = The key for the enrollment in which the pre-image is
+                contained.
+
+        Returns:
+            preimage information of the validator if exists, otherwise
+                PreimageInfo.init
+
+    ***************************************************************************/
+
+    public PreimageInfo getPreimage (Hash enroll_key);
 }
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -719,6 +719,43 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Get validator's pre-image from the validator set
+
+        Params:
+            enroll_key = The key for the enrollment in which the pre-image is
+                contained.
+            result_image = will contain the PreimageInfo if exists
+
+        Returns:
+            true if getting pre-image is successfully processed
+
+    ***************************************************************************/
+
+    public bool getValidatorPreimage (const ref Hash enroll_key,
+        out PreimageInfo result_image) @trusted nothrow
+    {
+        try
+        {
+            auto results = this.db.execute("SELECT preimage from validator_set" ~
+                        " WHERE key = ?", enroll_key.toString());
+            if (!results.empty && results.oneValue!(byte[]).length != 0)
+            {
+                result_image =
+                    results.oneValue!(ubyte[]).deserializeFull!(PreimageInfo);
+            }
+        }
+        catch (Exception ex)
+        {
+            log.error("Exception occured, getValidatorPreimage:{}, exception:{}",
+                enroll_key, ex);
+            return false;
+        }
+
+        return true;
+    }
+
+    /***************************************************************************
+
         Add a pre-image information to a validator data
 
         Params:
@@ -1119,7 +1156,7 @@ unittest
     assert(validators[0].utxo_key == utxo_hash3);
 }
 
-/// tests for addPreimage and hasPreimage
+/// tests for addPreimage and getValidatorPreimage
 unittest
 {
     import agora.common.Amount;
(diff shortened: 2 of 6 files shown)


linked0:
After modifying one thing, I will request review.
Please don't review now.

linked0:
Ready for review.
The ci/circleci build outputs error continuously as follows. If it should be handled, please let me know. Anyway, I will try it again.
```
object.Exception@ci/system_integration_test.d(67): Command failed: ["dub", "--root", "/home/circleci/project/tests/system"]
----------------
??:? void system_integration_test.runCmd(const(immutable(char)[][])) [0x48eb28]
??:? _Dmain [0x48e7e5]
```

Geod24:
This is not your error. You have to look at the full log, not just the last few lines ;)

source/agora/consensus/EnrollmentManager.d
```diff
@@ -717,6 +717,58 @@ public class EnrollmentManager
...
+            true if getting pre-image is successfully processed
+
+    ***************************************************************************/
+
+    public bool getValidatorPreimage (const ref Hash enroll_key, ulong height,
+        out PreimageInfo result_image) @safe nothrow
+    {
+        try
+        {
+            () @trusted {
```
AndrejMitrovic: If nearly the entire body of the function has to be trusted, then you can mark `getValidatorPreimage` as @trusted, and therefore remove this inline delegate.

AndrejMitrovic (dismissed):

linked0:
Addressed review

Geod24:
@linked0 : Still need to fix the CircleCI error though.

source/agora/api/FullNode.d
```diff
@@ -184,4 +185,24 @@ public interface API
...
+                contained.
+            height = The block height of the preimage to get
+
+        Returns:
+            preimage information of the validator if exists, otherwise
+                PreimageInfo.init
+
+    ***************************************************************************/
+
+    public PreimageInfo getPreimage (Hash enroll_key, ulong height);
```
Geod24: Why do we pass `height` here ? The following would provide a much simpler API, wouldn't it ?

```suggestion
    public PreimageInfo getPreimage (Hash enroll_key);
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -717,6 +717,56 @@ public class EnrollmentManager
...
+                {
+                    result_image.enroll_key = loaded_image.enroll_key;
+                    result_image.hash = loaded_image.hash;
+                    result_image.height = height;
+                    foreach (i; 0 .. loaded_image.height - height)
+                    {
+                        result_image.hash = hashFull(result_image.hash);
+                        result_image.height++;
+                    }
+                }
```
Geod24: Yeah, if you change the API to just return the latest preimage, instead of a user-requested height, you can remove this whole block of code and remove an attack vector.

A very real attack vector in fact, because  what happens if we're at block 100,000 and I request the preimage at height 1 for a currently enrolled validator ? There is no bound on how much we hash here, so you would end up performing 99,998 hash operations.

source/agora/test/EnrollmentManager.d
```diff
@@ -81,7 +81,11 @@ unittest
 
     // check if nodes contains the pre-image previously sent.
     nodes.each!(node =>
-        retryFor(node.hasPreimage(enroll.utxo_key, 999) == true, 5.seconds));
+        retryFor(node.getPreimage(enroll.utxo_key, 999) != PreimageInfo.init,
```
Geod24: Can we compare to the real value here ? If the node returns garbage this test will still pass when it shouldn't.

source/agora/test/EnrollmentManager.d
```diff
@@ -163,6 +167,12 @@ unittest
     // check if nodes have a pre-image newly sent
     // during creating transactions for the new block
     nodes.each!(node =>
-        retryFor(node.hasPreimage(enroll.utxo_key,
-            EnrollmentManager.PreimageRevealPeriod + 2), 5.seconds));
+        retryFor(node.getPreimage(enroll.utxo_key,
+            EnrollmentManager.PreimageRevealPeriod * 2 + 1) != PreimageInfo.init,
```
Geod24: Same comment about testing real value.

Geod24 (changes requested):

source/agora/api/FullNode.d
linked0: I agree with that. It would be simple.

source/agora/consensus/EnrollmentManager.d
linked0: The code can be removed if the `height` parameter is removed.

source/agora/test/EnrollmentManager.d
linked0: I've already thought about it. But in order to compare to the real value, it must know about the random seed or another preimage. So it needs another unecessary API like getNodePreimage for test.

linked0: But, I can add checking code for the preimage having garbage.

Geod24: That makes sense. Let's leave it as is for the time being then.

linked0:
The circleci build outputs error continuously. There seems some problem in calling `spawnProcess` with the parameter of "["dub", "--root", "/home/circleci/project/tests/system", "--", "http://127.0.0.1:4000", "http://127.0.0.1:4001", "http://127.0 .0.1:4002"]".

What should I try to solve it?

Geod24:
The error is:
```
Linking...
tests/system/.dub/build/application-debug-linux.posix-x86_64-dmd_2090-DB8FB99E8E2FF2C1325A116C295C574C/systemtest-simple.o:(.data.rel.ro+0x20): undefined reference to `_D5agora9consensus4data12PreimageInfo12__ModuleInfoZ'
tests/system/.dub/build/application-debug-linux.posix-x86_64-dmd_2090-DB8FB99E8E2FF2C1325A116C295C574C/systemtest-simple.o:(.data.rel.ro+0xa80): undefined reference to `_D5agora9consensus4data12PreimageInfo12__ModuleInfoZ'
collect2: error: ld returned 1 exit status
Error: linker exited with status 1
/home/circleci/dlang/dmd-2.090.1/linux/bin64/dmd failed with exit code 1.
```

It means one of your module is missing in `dub.json` (`agora.consensus.data.PreimageInfo`).

linked0:
> It means one of your module is missing in `dub.json` (`agora.consensus.data.PreimageInfo`).
Thank you. I'll check it out.

Geod24:
Rebasing on `v0.x.x` should fix it.

linked0:
Addressed review and thank you for supporting for the build.

Geod24 (approved):
LGTM

linked0:
Ping @AndrejMitrovic

Issue #606, opened 2020-02-29
SCP messages need to be broadcasted to all a Node's connected Validator peers


AndrejMitrovic:
In the current design a Validator is broadcasting messages to its connected peers through this function:

https://github.com/bpfkorea/agora/blob/acbd5464cd735d6c62910d49e2bac7346926ceb0/source/agora/consensus/protocol/Nominator.d#L334-L338

However these list of peers are a subset of Validators. They are the quorum slice of the node itself.

Essentially, given node A which has nodes B and C in its quorum slice:

```
A -> {B, C}
```

Node `A` will broadcast SCP messages (Nominate, etc..) to nodes `B` and `C`. In fact in this case, nodes `B` and `C` don't care about messages from node `A`, because `A` is not in either of their quorum slices.

Additionally, if there is any node `X` which has node `A` in its quorum set but node `A` does not have node `X` in its quorum set, then node `X` will not receive nomination messages from node `A`.

Example:

```
X -> A -> {B, C}
```

Node `X` must receive SCP messages from node `A`. In the current design, it does not.

In the test-suite all the nodes have each other in their quorum slice so this problem was not immediately apparent.

Reading https://www.stellar.org/developers-blog/intuitive-stellar-consensus-protocol is what made me realize there was a problem.

-----

What needs to be changed:

- [ ] The list of peers in the `Nominator` should be the list of Validator nodes which connected to that node, expecting SCP nomination / balloting statements.
- [ ] The connection may be established at any arbitrary time. Despite being possible to determine the exact quorum configuration based on https://github.com/bpfkorea/agora/issues/240, a Validator node A can only know the public key of such node, it cannot know for sure how node X with some public key _XP_ will connect to it (which IP, etc).

AndrejMitrovic:
Note that I'm still researching this, so I might be missing something else here. After I'm done researching, I'll start work next week.

Geod24:
Right. I had the same problem a few years ago when I tried implementing SCP from the paper.
A good way to test this is to implement a circular network configuration.

AndrejMitrovic:
In fact, the circular configuration is the one that fails in the test-suite when the threshold is set to 2 (self node + other node). It was not properly reaching consensus: https://github.com/bpfkorea/agora/blob/acbd5464cd735d6c62910d49e2bac7346926ceb0/source/agora/test/Base.d#L723

Doh!

It took me a while, but I'm beginning to see a much clearer picture of how SCP actually works under the hood.

[...]

PR #618, opened 2020-03-04
Refactor some Nominator & NetworkManager code

AndrejMitrovic:
This is the first part of fixing https://github.com/bpfkorea/agora/issues/606

While implementing the feature, I've realized there are leaky abstractions here. In fact SCP only requires two networking functionalities:

- receive an SCP message
- gossip it

The second part can be left to the NetworkManager, as it also deals with features such as node banning, receiving new network connections (to be done), etc.

### DIFF
--- a/source/agora/api/Validator.d
+++ b/source/agora/api/Validator.d
@@ -84,17 +84,16 @@ public interface API : agora.api.FullNode.API
 
     /***************************************************************************
 
-        Receives an SCP envelope and processes it
+        Receives an SCP envelope and processes it.
+        The node does not respond with any status code,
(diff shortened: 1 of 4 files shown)


TrustHenry (approved):
This PR is Greate.
PR #504 will be resolved.

But I see another problem #504

AndrejMitrovic:
Really? That was not my intention, but I'm glad if it's true. 👀

TrustHenry:
> Really? That was not my intention, but I'm glad if it's true. 👀

maybe Intermittent http error began to be invisible.

Geod24 (approved):
Beautiful. Just rebase and feel free to self merge.

PR #621, opened 2020-03-04
Send & receive SCP quorum sets to fix SCP message propagation

### LINKED ISSUE #606: SCP messages need to be broadcasted to all a Node's connected Validator peers

AndrejMitrovic:
Fixes https://github.com/bpfkorea/agora/issues/606

Requires: 
- ~https://github.com/bpfkorea/agora/pull/643~
- ~https://github.com/bpfkorea/agora/pull/648~

### DIFF
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -65,6 +65,9 @@ public class NetworkManager
     /// The connected nodes
     protected NetworkClient[PublicKey] peers;
 
+    /// The quorum set
+    protected SCPQuorumSet[Hash] quorum_sets;
+
     /// The addresses currently establishing connections to.
     /// Used to prevent connecting to the same address twice.
     protected Set!Address connecting_addresses;
@@ -128,6 +131,18 @@ public class NetworkManager
         // not connecting? connect later
         if (address !in this.connecting_addresses)
             this.todo_addresses.put(address);
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the quorum set associated with the given hash, or null if not found
+
+    ***************************************************************************/
+
+    public SCPQuorumSet* getQuorumSet (Hash hash) nothrow
+    {
+        return hash in this.quorum_sets;
     }
 
     /***************************************************************************
@@ -376,6 +391,9 @@ public class NetworkManager
             try
             {
                 node.getPublicKey();
+                node.getQuorumHashSet();
+                if (node.quorum_hash != Hash.init)
+                    this.quorum_sets[node.quorum_hash] = node.quorum_set;
                 this.connecting_addresses.remove(node.address);
                 this.required_peer_keys.remove(node.key);
 
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -272,6 +272,55 @@ public class Node : API
 
     /***************************************************************************
 
+        Returns:
+            The quorum hash of this node.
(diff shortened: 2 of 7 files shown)


AndrejMitrovic:
Some of these commits should be split off into another PR.

source/agora/network/NetworkManager.d
```diff
@@ -365,6 +381,8 @@ public class NetworkManager
             try
             {
                 node.handshake(this.getAddress());
+                if (node.quorum_hash != StellarHash.init)
+                    this.quorum_sets[node.quorum_hash] = node.quorum_set;
```
AndrejMitrovic, line 396: Actually this should be lazy. If the quorum hash is already known, no need to retrieve the quorum set.

AndrejMitrovic:
I will split this into two, and remove the necessity to exchange SCP structs in the API.

AndrejMitrovic:
This needs to be reworked with regards to https://github.com/bpfkorea/agora/issues/702.

source/agora/node/Node.d
```diff
@@ -270,6 +270,55 @@ public class Node : API
...
+    ***************************************************************************/
+
+    public override SCPQuorumSet getQuorumSet (Hash hash) @trusted
+    {
+        import scpd.types.Stellar_types;
+        if (this.config.node.is_validator)
+        {
+            auto stellar_hash = uint512(hash);
+            auto set = this.nominator.getQSet(stellar_hash);
+            assert(set.ptr !is null);
```
TrustHenry, line 313: If StellarHash is not new, an error occurs.
Can it be used for malicious attacks?

AndrejMitrovic:
As mentioned in https://github.com/bpfkorea/agora/pull/621#issuecomment-609649242, this PR has become unnecessary. We don't need to look up quorum sets explicitly because they will be known in advance once #240 is implemented.

PR #623, opened 2020-03-04
Sets the address of the node to the argument in system integration test

MichaelKim20:
I have improved the system integration test so that I can set the address of the node as an argument.

Fixed #592

### DIFF
--- a/tests/system/source/main.d
+++ b/tests/system/source/main.d
@@ -25,42 +25,28 @@ import agora.utils.PrettyPrinter;
 import vibe.web.rest;
 
 import std.algorithm;
+import std.exception;
 import std.stdio;
 import core.thread;
 import core.time;
 
-/// Helper struct
-private struct Address
-{
-    ///
-    string host;
-    ///
-    ushort port;
 
-    /// Helper function to call make the address vibe.inet.URL friendly
-    public string withSchema () const @safe
+int main (string[] args)
+{
+    if (args.length < 2)
     {
-        import std.format;
-        return format("http://%s:%d", this.host, this.port);
+        writeln("You must enter addresses of the nodes to connect to.");
+        writeln("   ex) http://127.0.0.1:4000 http://127.0.0.1:4001 http://127.0.0.1:4002 ...");
+        return 1;
     }
-}
 
-/// Node addresses
-private immutable Address[] Addrs = [
-    { host: "127.0.0.1", port: 4000, },
-    { host: "127.0.0.1", port: 4001, },
-    { host: "127.0.0.1", port: 4002, },
-];
+    /// Address array of nodes
+    const addresses = args[1..$];
 
-/// Test node count
-private immutable uint NodeCnt = 3;
-
-void main ()
-{
     {
-        API[NodeCnt] clients;
-        foreach (idx, const ref addr; Addrs)
-            clients[idx] = new RestInterfaceClient!API(addr.withSchema());
+        API[] clients;
+        foreach (const ref addr; addresses)
+            clients ~= new RestInterfaceClient!API(addr);
 
         foreach (idx, ref client; clients)
         {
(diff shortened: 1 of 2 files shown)


tests/system/source/main.d
```diff
@@ -25,42 +25,32 @@ import agora.utils.PrettyPrinter;
...
+private immutable uint NodeCnt = 3;
 
-    /// Helper function to call make the address vibe.inet.URL friendly
-    public string withSchema () const @safe
+/// Address array of nodes
+private string[] addresses;
+
+void main (string[] args)
+{
+    if (args.length-1 != NodeCnt)
```
Geod24: Why do we force the count to be three ? All this test do is check that a network is consistent.
So we should check that the number of args is >= 2 but I don't think we should limit it to three.

tests/system/source/main.d
```diff
@@ -97,8 +87,8 @@ private void checkBlockHeight (ulong height)
     // TODO: This is a hack because of issue #312
     // https://github.com/bpfkorea/agora/issues/312
     API[NodeCnt] clients;
-    foreach (idx, const ref addr; Addrs)
-        clients[idx] = new RestInterfaceClient!API(addr.withSchema());
+    foreach (idx, const ref addr; addresses)
```
Geod24: You have to pass it as argument to this function

tests/system/source/main.d
MichaelKim20: That's a very good idea.

tests/system/source/main.d
MichaelKim20: I thought so, too. That'd be fine.

tests/system/source/main.d
```diff
@@ -25,42 +25,32 @@ import agora.utils.PrettyPrinter;
...
-    string host;
-    ///
-    ushort port;
+/// Test node count
+private immutable uint NodeCnt = 3;
 
-    /// Helper function to call make the address vibe.inet.URL friendly
-    public string withSchema () const @safe
+/// Address array of nodes
+private string[] addresses;
```
Geod24: No global please

MichaelKim20:
@Geod24 Thanks for the review.
I have solved all the amendments.

tests/system/source/main.d
```diff
@@ -25,42 +25,28 @@ import agora.utils.PrettyPrinter;
...
 
-    /// Helper function to call make the address vibe.inet.URL friendly
-    public string withSchema () const @safe
+void main (string[] args)
+{
+    if (args.length < 2)
     {
-        import std.format;
-        return format("http://%s:%d", this.host, this.port);
+        writeln("You must enter addresses of the nodes to connect to.");
```
Geod24, line 38:
```suggestion
        writeln("You must enter at least two addresses of nodes to connect to.");
```

tests/system/source/main.d
```diff
@@ -25,42 +25,28 @@ import agora.utils.PrettyPrinter;
...
-}
 
-/// Node addresses
-private immutable Address[] Addrs = [
-    { host: "127.0.0.1", port: 4000, },
-    { host: "127.0.0.1", port: 4001, },
-    { host: "127.0.0.1", port: 4002, },
-];
+    /// Address array of nodes
+    const auto addresses = args[1..$];
```
Geod24:
```suggestion
    const addresses = args[1..$];
```

tests/system/source/main.d
```diff
@@ -25,42 +25,28 @@ import agora.utils.PrettyPrinter;
...
-    public string withSchema () const @safe
+void main (string[] args)
+{
+    if (args.length < 2)
     {
-        import std.format;
-        return format("http://%s:%d", this.host, this.port);
+        writeln("You must enter addresses of the nodes to connect to.");
+        writeln("   ex) http://127.0.0.1:4000 http://127.0.0.1:4001 http://127.0.0.1:4002 ...");
+        return;
```
Geod24: This return with a success code. That's not good, because then any script, and the CI, will see it as successful. Return 1.

Geod24:
Added a few more items, but overall LGTM

tests/system/source/main.d
```diff
@@ -25,42 +25,28 @@ import agora.utils.PrettyPrinter;
...
-private immutable uint NodeCnt = 3;
-
-void main ()
-{
     {
-        API[NodeCnt] clients;
-        foreach (idx, const ref addr; Addrs)
-            clients[idx] = new RestInterfaceClient!API(addr.withSchema());
+        API[] clients;
+        foreach (idx, const ref addr; addresses)
```
AndrejMitrovic:
```suggestion
        foreach (const ref addr; addresses)
```

tests/system/source/main.d
```diff
@@ -86,19 +72,19 @@ void main ()
...
-private void checkBlockHeight (ulong height)
+private void checkBlockHeight (const string[] addresses, ulong height)
 {
     // TODO: This is a hack because of issue #312
     // https://github.com/bpfkorea/agora/issues/312
-    API[NodeCnt] clients;
-    foreach (idx, const ref addr; Addrs)
-        clients[idx] = new RestInterfaceClient!API(addr.withSchema());
+    API[] clients;
+    foreach (idx, const ref addr; addresses)
```
AndrejMitrovic:
```suggestion
    foreach (const ref addr; addresses)
```

AndrejMitrovic (dismissed):

tests/system/source/main.d
MichaelKim20: That's a good catch. Thanks

Geod24 (approved):

Issue #264, opened 2019-08-23
Setup Windows build


bpalaggi:
Setup a build process to have Agora compile and run tests on Windows (probably through Github actions). We should test at least DMD, and hopefully LDC as well.

DoD:
- [x] Agora is built and tested on Windows on each pull requests and push to `v0.x.x`

Geod24:
So, more details about this:
1) We should definitely use Github actions
2) All the libraries we use are compatible with Windows, but most of the work will be on setting them up
3) We can add it on PR only as a first step (https://github.com/bpfkorea/agora/blob/v0.x.x/.github/workflows/github_pr.yml)

Geod24:
Github actions documentation: https://help.github.com/en/actions
One of the issue you will face is that we use G++ to build our C++ code: https://github.com/bpfkorea/agora/blob/c7314c21700c526d6ee496090694bef46dc9da54/source/scpp/build.d#L51

AndrejMitrovic:
Github might have a gcc compiler on Windows too (probably via MinGW). For example: https://github.com/onqtam/doctest/blob/875052e18cf5f5f0b9176e59f847b6205f3efb78/.github/workflows/main.yml#L46

Geod24:
Done in #624

Geod24:
Note: It is not tested on `push` currently, but that's a minor detail, and I will refactor the Github workflow so it is in the near future.

PR #624, opened 2020-03-04
Github actions windows build for PR

TrustHenry:
Setup a build process to have Agora compile and run tests on Windows

This is created to test the Windows build.
The operating environment is tested in an ldc-1.20.1 and ldc-1.19.0 
environment in Windows-2019 x64.
This PR does not support x86 environments.

Relates to  Setup Windows build #264

### DIFF
--- a/.github/workflows/github_pr.yml
+++ b/.github/workflows/github_pr.yml
@@ -22,8 +22,7 @@ jobs:
         os: [ ubuntu-18.04, macOS-10.15, windows-2019 ]
         dc: [ dmd-2.091.0, ldc-1.20.1, ldc-1.19.0 ]
         exclude:
-          # Not supported yet
-          - os: windows-2019
+          - { os: windows-2019, dc: dmd-2.091.0 }
 
     runs-on: ${{ matrix.os }}
     timeout-minutes: 30
@@ -61,12 +60,40 @@ jobs:
         sudo apt-get install libsodium-dev
         sudo apt-get install g++-9
 
+    - name: '[Windows] Install dependencies & setup environment'
+      if: runner.os == 'Windows'
+      shell: powershell
+      run: |
+        $url = "https://github.com/jedisct1/libsodium/releases/download/1.0.18-RELEASE/libsodium-1.0.18-msvc.zip"
+        $sha256hash = "C1D48D85C9361E350931FFE5067559CD7405A697C655D26955FB568D1084A5F4"
+        Write-Host ('Downloading {0} ...' -f $url)
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+        # See https://github.com/PowerShell/PowerShell/issues/2138
+        $ProgressPreference = 'SilentlyContinue'
+        New-Item -ItemType directory -Path ${{ github.workspace }}\lib\
+        Invoke-WebRequest -Uri $url -OutFile '${{ github.workspace }}\lib\libsodium.zip'
+        if ((Get-FileHash '${{ github.workspace }}\lib\libsodium.zip' -Algorithm "SHA256").Hash -ne $sha256hash) {
+          exit 1
+        }
+        Expand-Archive '${{ github.workspace }}\lib\libsodium.zip' -DestinationPath ${{ github.workspace }}\lib\
+        echo "::set-env name=LIB::${{ github.workspace }}\lib\libsodium\x64\Release\v142\static\;${{ github.workspace }}\submodules\d2sqlite3\lib\win64\;$LIB"
+        echo "::set-env name=INCLUDE::${{ github.workspace }}\lib\libsodium\include\;$INCLUDE"
+
     # Add whatever debugging information can be useful in the long run here
     - name: Print system informations
+      shell: bash
       run: |
         ${DC} --version
         dub --version
 
     # Build and run the tests
-    - name: Build & test Agora
+    - name: '[Posix] Build & test Agora'
+      if: runner.os != 'Windows'
       run: ./ci/run.sh
+
+    - name: '[Windows] Build & test Agora'
+      if: runner.os == 'Windows'
+      shell: cmd
+      run: |
+        call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        call ci\run.bat
--- a/ci/run.bat
+++ b/ci/run.bat
@@ -1,0 +1,4 @@
+call dub test --skip-registry=all --compiler=%DC%
+call rdmd --compiler=%DC% ./tests/runner.d --compiler=%DC% -cov
+call dub build --skip-registry=all --compiler=%DC%
+call dub build -c client --skip-registry=all --compiler=%DC%
--- a/dub.json
+++ b/dub.json
@@ -12,7 +12,7 @@
     "targetPath": "build",
     "excludedSourceFiles": [ "source/scpp/*.d" ],
     "preGenerateCommands": [ "$DUB source/scpp/build.d" ],
-    "sourceFiles": [
+    "sourceFiles-posix": [
         "source/scpp/build/BallotProtocol.o",
         "source/scpp/build/ByteSliceHasher.o",
         "source/scpp/build/DSCPUtils.o",
@@ -37,9 +37,35 @@
         "source/scpp/build/numeric.o",
         "source/scpp/build/uint128_t.o"
     ],
+    "sourceFiles-windows": [
+        "source/scpp/build/BallotProtocol.obj",
+        "source/scpp/build/ByteSliceHasher.obj",
+        "source/scpp/build/DSCPUtils.obj",
+        "source/scpp/build/DUtils.obj",
+        "source/scpp/build/HashOfHash.obj",
+        "source/scpp/build/Hex.obj",
+        "source/scpp/build/KeyUtils.obj",
+        "source/scpp/build/LocalNode.obj",
+        "source/scpp/build/Logging.obj",
+        "source/scpp/build/Math.obj",
+        "source/scpp/build/NominationProtocol.obj",
+        "source/scpp/build/QuorumSetUtils.obj",
+        "source/scpp/build/SCP.obj",
+        "source/scpp/build/SCPDriver.obj",
+        "source/scpp/build/SHA.obj",
+        "source/scpp/build/SecretKey.obj",
+        "source/scpp/build/Slot.obj",
+        "source/scpp/build/StrKey.obj",
+        "source/scpp/build/crc16.obj",
+        "source/scpp/build/jsoncpp.obj",
+        "source/scpp/build/marshal.obj",
+        "source/scpp/build/numeric.obj",
+        "source/scpp/build/uint128_t.obj"
+    ],
 
     "dflags": [ "-extern-std=c++14" ],
-    "lflags": [ "-lstdc++", "-lsqlite3" ],
+    "lflags-posix": [ "-lstdc++", "-lsqlite3" ],
+    "lflags-windows": [  "sqlite3.lib", "/nodefaultlib:msvcetd.lib" ],
     "buildRequirements": [ "allowWarnings" ],
 
     "configurations": [
@@ -68,9 +94,13 @@
             "name": "unittest",
             "targetName": "agora-unittests",
             "excludedSourceFiles": ["source/agora/cli/*/main.d"],
-            "sourceFiles": [
+            "sourceFiles-posix": [
                 "source/scpp/build/DSizeChecks.o",
                 "source/scpp/build/DLayoutChecks.o"
+            ],
+            "sourceFiles-windows": [
+                "source/scpp/build/DSizeChecks.obj",
+                "source/scpp/build/DLayoutChecks.obj"
             ]
         },
         {
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -192,16 +192,20 @@ extern(C++, `std`) {
 /// Type of SCP function callback called by a timer
 public alias SCPCallback = extern(C++) void function();
 
-private extern(C++) set!uint* makeTestSet();
-
-unittest
+// TODO : MSVC mangling issue w.r.t. the return type
+version (Posix)
 {
-    auto set = makeTestSet;
-    assert(!set.empty);
-    uint[] values;
-    foreach (val; *set)
-        values ~= val;
-    assert(values == [1, 2, 3, 4, 5]);
+    private extern(C++) set!uint* makeTestSet();
+
+    unittest
+    {
+        auto set = makeTestSet;
+        assert(!set.empty);
+        uint[] values;
+        foreach (val; *set)
+            values ~= val;
+        assert(values == [1, 2, 3, 4, 5]);
+    }
 }
 
 /// Can't import `core.stdcpp.allocator` because it transitively imports
--- a/source/scpd/scp/SCPDriver.d
+++ b/source/scpd/scp/SCPDriver.d
@@ -71,18 +71,42 @@ nothrow:
     // returning Value() means no valid value could be extracted
     Value extractValidValue(uint64_t slotIndex, ref const(Value) value);
 
-    // `getValueString` is used for debugging
-    // default implementation is the hash of the value
-    //std::string getValueString(ref const(Value) v) const;
-    void* getValueString(ref const(Value) v) const;
+    version (Windows)
+    {
+        // TODO: Take temporary action to support mangling in Windows MSVC
+        // MSVC mangles the return value, while POSIX does not,
+        // and `getValueString`, `toStrKey`, and `toShortString` return `std::string` for which we need bindings
+        // `getValueString` is used for debugging
+        // default implementation is the hash of the value
+        //std::string getValueString(ref const(Value) v) const;
+        pragma(mangle, `?getValueString@SCPDriver@stellar@@UEBA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@AEBU?$xvector@E$0PPPPPPPM@@xdr@@@Z`)
+        void* getValueString(ref const(Value) v) const;
 
-    // `toStrKey` returns StrKey encoded string representation
-    //std::string toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
-    void* toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
+        // `toStrKey` returns StrKey encoded string representation
+        //std::string toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
+        pragma(mangle, `?toStrKey@SCPDriver@stellar@@UEBA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@AEBUPublicKey@2@_N@Z`)
+        void* toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
 
-    // `toShortString` converts to the common name of a key if found
-    //std::string toShortString(ref const(PublicKey) pk) const;
-    void* toShortString(ref const(PublicKey) pk) const;
+        // `toShortString` converts to the common name of a key if found
+        //std::string toShortString(ref const(PublicKey) pk) const;
+        pragma(mangle, `?toShortString@SCPDriver@stellar@@UEBA?AV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@AEBUPublicKey@2@@Z`)
+        void* toShortString(ref const(PublicKey) pk) const;
+    }
+    else
+    {
+        // `getValueString` is used for debugging
+        // default implementation is the hash of the value
+        //std::string getValueString(ref const(Value) v) const;
+        void* getValueString(ref const(Value) v) const;
+
+        // `toStrKey` returns StrKey encoded string representation
+        //std::string toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
+        void* toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
+
+        // `toShortString` converts to the common name of a key if found
+        //std::string toShortString(ref const(PublicKey) pk) const;
+        void* toShortString(ref const(PublicKey) pk) const;
+    }
 
     // `computeHashNode` is used by the nomination protocol to
     // randomize the order of messages between nodes.
@@ -109,8 +133,17 @@ nothrow:
     // `computeTimeout` computes a timeout given a round number
     // it should be sufficiently large such that nodes in a
     // quorum can exchange 4 messages
-    milliseconds computeTimeout(uint32_t roundNumber);  // Slot in the vtable
-
+    version (Windows)
+    {
+        // TODO Temporary action due to MSVC mangling problem
+        // https://issues.dlang.org/show_bug.cgi?id=20700
+        pragma(mangle, `?computeTimeout@SCPDriver@stellar@@UEAA?AV?$duration@_JU?$ratio@$00$0DOI@@std@@@chrono@std@@I@Z`)
+        milliseconds computeTimeout(uint32_t roundNumber);  // Slot in the vtable
+    }
+    else
+    {
+        milliseconds computeTimeout(uint32_t roundNumber);  // Slot in the vtable
+    }
     // Inform about events happening within the consensus algorithm.
 
     // `valueExternalized` is called at most once per slot when the slot
--- a/source/scpd/scp/Slot.d
+++ b/source/scpd/scp/Slot.d
@@ -45,7 +45,8 @@ private:
 
     // keeps track of all statements seen so far for this slot.
     // it is used for debugging purpose
-    struct HistoricalStatement
+    // https://issues.dlang.org/show_bug.cgi?id=20701
+    extern(C++, struct) struct HistoricalStatement
     {
         time_t mWhen;
         SCPStatement mStatement;
--- a/source/scpd/types/Utils.d
+++ b/source/scpd/types/Utils.d
@@ -68,7 +68,6 @@ public opaque_vec!() XDRToOpaque (const ref SCPStatement arg);
 
 extern(C++, `stellar`)
 {
-    public uint256 sha256(const ref ByteSlice bin);
 
     public extern(C++, class) struct ByteSlice
     {
@@ -91,4 +90,10 @@ extern(C++, `stellar`)
             return (cast(T*)this.mData)[0 .. this.mSize / T.sizeof];
         }
     }
+
+    // Note: This needs to be after `ByteSlice` because the frontend doesn't take
+    // `extern(C++, class)` into account otherwise
+    // https://issues.dlang.org/show_bug.cgi?id=20700
+    public uint256 sha256(const ref ByteSlice bin);
+
 }
--- a/source/scpp/build.d
+++ b/source/scpp/build.d
@@ -35,7 +35,8 @@ immutable SourcePath  = RootPath;
 immutable BuildPath   = RootPath.buildPath("build");
 immutable BuildTarget = BuildPath.buildPath("libscp.o");
 
-/// Path passed to `-I`
+/// Include path for C++ dependency - libsodium must be in the include path
+/// (INCLUDE on Windows, and /usr/include/ or similar on POSIX)
 immutable Includes = [
     SourcePath,
     SourcePath.buildPath("src"),
@@ -43,19 +44,81 @@ immutable Includes = [
     SourcePath.buildPath("lib", "xdrpp"),
 ];
 
-immutable CppFlags = [
-    "-c",
-    "-g",
-    "-W",
-    "-Wall",
-    "-Wno-unused-parameter",
-    "-Wno-comment",
-    "-fPIC",
-    "-D_GLIBCXX_USE_CXX11_ABI=0",
-    "-std=c++14",
-];
-
-immutable CppCmd = ["gcc" ] ~ CppFlags;
+version (Posix)
+{
+    immutable ObjPattern = "*.o";
+    immutable CompilerIncludeFlag = "-I ";
+    immutable CppFlags = [
+        "-c",
+        "-g",
+        "-W",
+        "-Wall",
+        "-Wno-unused-parameter",
+        "-fPIC",
+        "-D_GLIBCXX_USE_CXX11_ABI=0",
+        "-std=c++14",
+    ];
+    immutable CppCmd = [ "gcc" ] ~ CppFlags;
+}
+else version (Windows)
+{
+    immutable ObjPattern = "*.obj";
+    immutable CompilerIncludeFlag = "/I ";
+    immutable CppFlags = [
+        "/JMC",
+        "/MP",
+        "/GS",
+        "/W4",
+        // https://docs.microsoft.com/en/cpp/error-messages/tool-errors/linker-tools-warning-lnk4099?view=vs-2019
+        // Remove warning : PDB file not found
+        "/wd\"4099\"",
+        // https://docs.microsoft.com/en/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4100?view=vs-2019
+        // Remove warning 'identifier' : unreferenced formal parameter
+        "/wd\"4100\"",
+        // https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4127?view=vs-2019
+        // Remove warning : conditional expression is constant
+        "/wd\"4127\"",
+        // https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4324?view=vs-2019
+        // Remove warning 'struct_name' : structure was padded due to __declspec(align())
+        "/wd\"4324\"",
+        // https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4408?view=vs-2019
+        // Remove warning : anonymous struct or union did not declare any data members
+        "/wd\"4408\"",
+        // https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4510?view=vs-2019
+        // Remove warning 'class' : default constructor could not be generated
+        "/wd\"4510\"",
+        // https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4512?view=vs-2019
+        // Remove warning 'class' : assignment operator could not be generated
+        "/Zc:wchar_t",
+        "/Zi",
+        "/Gm-",
+        "/Od",
+        "/sdl",
+        "/fp:precise",
+        "/D \"BUILD_TESTS\"",
+        "/D \"WIN32_LEAN_AND_MEAN\"",
+        "/D \"NOMINMAX\"",
+        "/D \"_WINSOCK_DEPRECATED_NO_WARNINGS\"",
+        "/D \"SODIUM_STATIC\"",
+        "/D \"_CRT_SECURE_NO_WARNINGS\"",
+        "/D \"_WIN32_WINNT=0x0601\"",
+        "/D \"WIN32\"",
+        "/D \"_MBCS\"",
+        "/D \"_CRT_NONSTDC_NO_DEPRECATE\"",
+        "/errorReport:prompt",
+        "/WX-",
+        "/Zc:forScope",
+        "/RTC1",
+        "/Gd",
+        "/std:c++14",
+        "/FC",
+        "/EHsc",
+        "/c"
+    ];
+    immutable CppCmd = [ "cl" ] ~ CppFlags;
+}
+else
+    static assert(0, "Unsupported platform");
 
 struct FileInfo
 {
@@ -78,7 +141,7 @@ int main(string[] args)
     auto sources = std.file.dirEntries(
         SourcePath, "*.c*", std.file.SpanMode.depth).array;
     auto objs = std.file.dirEntries(
-        BuildPath, "*.o", std.file.SpanMode.depth).array;
+        BuildPath, ObjPattern, std.file.SpanMode.depth).array;
 
     // If one of the obj file is older than one of the source file, rebuild
     // That's a lesser approach than the dependency tracking Makefile do,
@@ -116,7 +179,7 @@ int main(string[] args)
     else
         writeln("First build / new source files added: Doing a full build...");
 
-    auto cmd = chain(CppCmd, Includes.map!((v) => "-I " ~ v), sources);
+    auto cmd = chain(CppCmd, Includes.map!((v) => CompilerIncludeFlag ~ v), sources);
     auto strCmd = cmd.join(" ");
     // writeln(strCmd);
     auto pid = executeShell(strCmd);
--- a/source/scpp/src/scp/SCPDriver.cpp
+++ b/source/scpp/src/scp/SCPDriver.cpp
@@ -98,4 +98,37 @@ SCPDriver::computeTimeout(uint32 roundNumber)
     }
     return std::chrono::seconds(timeoutInSeconds);
 }
+
+Value SCPDriver::extractValidValue(uint64 slotIndex, Value const& value)
+{
+    return Value();
+}
+
+void SCPDriver::nominatingValue(uint64 slotIndex, Value const& value)
+{
+}
+
+void SCPDriver::updatedCandidateValue(uint64 slotIndex, Value const& value)
+{
+}
+
+void SCPDriver::startedBallotProtocol(uint64 slotIndex, SCPBallot const& ballot)
+{
+}
+
+void SCPDriver::acceptedBallotPrepared(uint64 slotIndex, SCPBallot const& ballot)
+{
+}
+
+void SCPDriver::confirmedBallotPrepared(uint64 slotIndex, SCPBallot const& ballot)
+{
+}
+
+void SCPDriver::acceptedCommit(uint64 slotIndex, SCPBallot const& ballot)
+{
+}
+
+void SCPDriver::ballotDidHearFromQuorum(uint64 slotIndex, SCPBallot const& ballot)
+{
+}
 }
--- a/source/scpp/src/scp/Slot.h
+++ b/source/scpp/src/scp/Slot.h
@@ -25,6 +25,9 @@ class Node;
  */
 class Slot : public std::enable_shared_from_this<Slot>
 {
+    // BPFK note: cannot be private as we require runtime layout checks
+  public:
+
     const uint64 mSlotIndex; // the index this slot is tracking
     SCP& mSCP;
 
--- a/source/scpp/src/scp/SCPDriver.h
+++ b/source/scpp/src/scp/SCPDriver.h
@@ -77,10 +77,7 @@ class SCPDriver
     // validateValue did not return `kFullyValidatedValue` for this value).
     // returning Value() means no valid value could be extracted
     virtual Value
-    extractValidValue(uint64 slotIndex, Value const& value)
-    {
-        return Value();
-    }
+    extractValidValue(uint64 slotIndex, Value const& value);
 
     // `getValueString` is used for debugging
     // default implementation is the hash of the value
@@ -132,9 +129,7 @@ class SCPDriver
     // ``nominatingValue`` is called every time the local instance nominates
     // a new value.
     virtual void
-    nominatingValue(uint64 slotIndex, Value const& value)
-    {
-    }
+    nominatingValue(uint64 slotIndex, Value const& value);
 
     // the following methods are used for monitoring of the SCP subsystem
     // most implementation don't really need to do anything with these
@@ -143,41 +138,29 @@ class SCPDriver
     // is included in the candidate set, the value passed in is
     // a composite value
     virtual void
-    updatedCandidateValue(uint64 slotIndex, Value const& value)
-    {
-    }
+    updatedCandidateValue(uint64 slotIndex, Value const& value);
 
     // `startedBallotProtocol` is called when the ballot protocol is started
     // (ie attempts to prepare a new ballot)
     virtual void
-    startedBallotProtocol(uint64 slotIndex, SCPBallot const& ballot)
-    {
-    }
+    startedBallotProtocol(uint64 slotIndex, SCPBallot const& ballot);
 
     // `acceptedBallotPrepared` every time a ballot is accepted as prepared
     virtual void
-    acceptedBallotPrepared(uint64 slotIndex, SCPBallot const& ballot)
-    {
-    }
+    acceptedBallotPrepared(uint64 slotIndex, SCPBallot const& ballot);
 
     // `confirmedBallotPrepared` every time a ballot is confirmed prepared
     virtual void
-    confirmedBallotPrepared(uint64 slotIndex, SCPBallot const& ballot)
-    {
-    }
+    confirmedBallotPrepared(uint64 slotIndex, SCPBallot const& ballot);
 
     // `acceptedCommit` every time a ballot is accepted commit
     virtual void
-    acceptedCommit(uint64 slotIndex, SCPBallot const& ballot)
-    {
-    }
+    acceptedCommit(uint64 slotIndex, SCPBallot const& ballot);
 
     // `ballotDidHearFromQuorum` is called when we received messages related to
     // the current `mBallot` from a set of node that is a transitive quorum for
     // the local node.
     virtual void
-    ballotDidHearFromQuorum(uint64 slotIndex, SCPBallot const& ballot)
-    {
-    }
+    ballotDidHearFromQuorum(uint64 slotIndex, SCPBallot const& ballot);
 };
 }
--- a/submodules/README.md
+++ b/submodules/README.md
@@ -13,7 +13,7 @@ This data can also be gathered by running `git submodules`.
 | diet-ng          | v1.6.1       | 2020-01-14  |     No       |
 | dyaml            | v0.8.0       | 2019-05-27  |     No       |
 | eventcore        | v0.8.50      | 2020-03-18  |     Yes      |
-| libsodiumd       | 0.1.0+1.0.18 | 2019-08-12  | Yes (Owned)  |
+| libsodiumd       | 0.1.1+1.0.18 | 2020-04-04  | Yes (Owned)  |
 | localrest        | 6da1455      | 2020-02-14  | Yes (Owned)  |
 | mir-linux-kernel | 1.0.1        | 2018-09-17  |     No       |
 | openssl          | 1.1.6+1.0.1g | 2017-11-06  |     No       |
--- a/tests/runner.d
+++ b/tests/runner.d
@@ -19,6 +19,7 @@ import std.file;
 import std.path;
 import std.process;
 import std.stdio;
+import std.string : strip;
 
 /// Directory where simple tests live
 private immutable UnitPath = __FILE_FULL_PATH__.dirName.buildPath("unit");
@@ -64,7 +65,7 @@ private string[] getImportPaths ()
         pp.stderr.byLine.each!(a => writeln("[fatal]\t", a));
         return null;
     }
-    return pp.stdout.byLineCopy.array;
+    return pp.stdout.byLineCopy.map!strip.array;
 }
 
 /// Get the linker flags from dub
@@ -80,10 +81,17 @@ private string[] getLflags ()
             pp.stderr.byLine.each!(a => writeln("[fatal]\t", a));
             return null;
         }
-        return pp.stdout.byLineCopy.array;
+        return pp.stdout.byLineCopy.map!strip.array;
     }
-    else
+    else version (Posix)
     {
         return [ "-lsodium" ];
     }
+    else version (Windows)
+    {
+        // Note: Only works with LDC, see https://github.com/dlang/dub/issues/1914
+        return [ "-llibsodium" ];
+    }
+    else
+        static assert(0, "Unhandled platform");
 }


Geod24:
```
Running pre-build commands...
'source' is not recognized as an internal or external command,
operable program or batch file.
Command failed with exit code 1: source/scpp/build.d
```

This is because our `preBuildCommand` for building SCP is: https://github.com/bpfkorea/agora/blob/f41737f3b4d51b856eacfc5c5df9fbd22f1c158a/dub.json#L14

And it uses the shebang syntax (https://github.com/bpfkorea/agora/blob/f41737f3b4d51b856eacfc5c5df9fbd22f1c158a/source/scpp/build.d#L1) which is not recognized on Windows. So you should change the dub line to `"${DC} -run source/scpp/build.d"`

Geod24:
```
Running pre-build commands...
Build failed: In file included from d:\a\agora\agora\source\scpp\lib\xdrpp/xdrpp/message.h:12,
                 from d:\a\agora\agora\source\scpp\lib\xdrpp/xdrpp/marshal.h:12,
                 from d:\a\agora\agora\source\scpp\extra\DLayoutChecks.cpp:17:
d:\a\agora\agora\source\scpp\lib\xdrpp/xdrpp/socket.h:21:10: fatal error: netdb.h: No such file or directory
 #include <netdb.h>
          ^~~~~~~~~
compilation terminated.
```

https://github.com/xdrpp/xdrpp/blob/ae501492b7eb5a829d0d98f2994ae7c98f629757/xdrpp/socket.h#L10-L27

Urgh, upstream only supports MSVC on Windows 🤦‍♂

Geod24:
Try:
```diff
diff --git a/source/scpp/lib/xdrpp/xdrpp/socket.h b/source/scpp/lib/xdrpp/xdrpp/socket.h
index 24ca400d..96b8e21a 100644
--- a/source/scpp/lib/xdrpp/xdrpp/socket.h
+++ b/source/scpp/lib/xdrpp/xdrpp/socket.h
@@ -7,7 +7,7 @@

 #include <memory>
 #include <system_error>
-#if MSVC
+#if _WIN32
 #include <WinSock2.h>
 #include <Ws2tcpip.h>
 struct iovec {
@@ -17,14 +17,14 @@ struct iovec {
 typedef int socklen_t;
 typedef unsigned __int64 ssize_t;

-#else // !MSVC
+#else // !_WIN32
 #include <netdb.h>
 #include <unistd.h>
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <sys/uio.h>
 #include <netinet/in.h>
-#endif // !MSVC
+#endif // !_WIN32

 namespace xdr {
```

AndrejMitrovic:
> Urgh, upstream only supports MSVC on Windows 🤦‍♂

Damn. Well in the worst case scenario we can make the VS Community Edition as a requirement for building on Windows, since it's free.

Geod24:
I don't see any special handling for `cl` in https://github.com/onqtam/doctest/blob/875052e18cf5f5f0b9176e59f847b6205f3efb78/.github/workflows/main.yml so it might even be simpler to setup (although our script will need some modifications).

AndrejMitrovic:
@TrustHenry have you seen https://libsodium.gitbook.io/doc/installation#compilation-on-windows?

`MingW` is the name of the GCC compiler on Windows (it is a custom distribution of GCC built specifically for Windows).

.github/workflows/github_pr.yml
```diff
@@ -56,3 +56,45 @@ jobs:
...
+      with:
+        submodules: true
+    - name: Prepare compiler
+      uses: mihails-strasuns/setup-dlang@v0.5.0
+      with:
+        compiler: ${{ matrix.dc }}
+        gh_token: ${{ secrets.GITHUB_TOKEN }}
+    - name: Print compiler and dub's version
+      run: |
+        dmd --version
```
Geod24:
```suggestion
        ${DC} --version
```

Won't make a difference right now, but for the future.

Geod24:
I'm afraid we will have to use MSVC, as DMD only mangles for MSVC on Windows: https://github.com/dlang/dmd/blob/0eb76da608e0e5c10a2db1cef3d5e40662ce6585/src/dmd/target.d#L791-L806

.github/workflows/github_pr.yml
TrustHenry:
```
Run ${DC} --version
At D:\a\_temp\cec021c4-9345-4f5c-bede-522e776aff25.ps1:2 char:9
+ ${DC} --version
+         ~~~~~~~
Unexpected token 'version' in expression or statement.
+ CategoryInfo          : ParserError: (:) [], ParseException
+ FullyQualifiedErrorId : UnexpectedToken
```

TrustHenry:
> I'm afraid we will have to use MSVC, as DMD only mangles for MSVC on Windows: https://github.com/dlang/dmd/blob/0eb76da608e0e5c10a2db1cef3d5e40662ce6585/src/dmd/target.d#L791-L806

xdrpp in socket.h 
`D:\BOA\Henry\test-action\source\scpp\lib\xdrpp/xdrpp/socket.h:18:24: error: expected initializer before 'ssize_t'
 typedef unsigned __int64 ssize_t;`

The mingw gcc compiler does not support __int64 .
Microsoft visualc++ is the only one.

AndrejMitrovic:
mingw-w64 supports 64-bit targets, and `__int64` should be available. It's possible the Github Actions is using the wrong compiler.

In any case, what I would personally suggest is: use MSVC for now, and we can add support for MinGW later.

Most Windows libraries support MSVC very well, while MinGW might be a bit harder to set up (I know this from experience building projects on Windows).

Geod24:
Also: https://github.com/bpfkorea/agora/pull/624#issuecomment-595645512

AndrejMitrovic:
@TrustHenry can you add the build command we tried yesterday to a commit in this PR? That way we can see the error in English when it runs in the CI, and we can figure out what's wrong.

TrustHenry:
Yesterday's compilation environment was local. So we're building the same environment in the github actions.

AndrejMitrovic:
@TrustHenry if you look here you will see that we modified some SCPP header files to make some fields public: https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/src/crypto/ByteSlice.h#L21

This allows our size & layout checks in https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/extra/DSizeChecks.cpp and https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/extra/DLayoutChecks.cpp to work.

So, it's possible we will have to add more `public:` declarations in other C++ structs / classes to fix those compilation errors.

TrustHenry:
For a moment, we test for the Windows version only.

Geod24:
Looks like you're making good progress!

TrustHenry:
````
cl /JMC /MP /GS /W4 /wd"4060" /wd"4100" /wd"4127" /wd"4324" /wd"4408" /wd"4510" /wd"4512" /wd"4582" /wd"4583" /wd"4592" /Zc:wchar_t /Zi /Gm- /Od /sdl /Zc:inline /fp:precise /D "BUILD_TESTS" /D "USE_POSTGRES" /D "WIN32_LEAN_AND_MEAN" /D "NOMINMAX" /D "ASIO_STANDALONE" /D "_WINSOCK_DEPRECATED_NO_WARNINGS" /D "SODIUM_STATIC" /D "ASIO_SEPARATE_COMPILATION" /D "ASIO_ERROR_CATEGORY_NOEXCEPT=noexcept" /D "_CRT_SECURE_NO_WARNINGS" /D "_WIN32_WINNT=0x0601" /D "WIN32" /D "_MBCS" /D "_CRT_NONSTDC_NO_DEPRECATE" /errorReport:prompt /WX- /Zc:forScope /RTC1 /Gd /MDd /std:c++14 /FC /EHsc /c /I d:\a\agora\agora\source\scpp /I d:\a\agora\agora\source\scpp\src /I d:\a\agora\agora\source\scpp\lib /I d:\a\agora\agora\source\scpp\lib\xdrpp /I d:\a\agora\agora\source\scpp\lib\libsodium\include d:\a\agora\agora\source\scpp\extra\DLayoutChecks.cpp d:\a\agora\agora\source\scpp\extra\DSCPUtils.cpp d:\a\agora\agora\source\scpp\extra\DSizeChecks.cpp d:\a\agora\agora\source\scpp\extra\DUtils.cpp d:\a\agora\agora\source\scpp\lib\json\jsoncpp.cpp d:\a\agora\agora\source\scpp\lib\util\crc16.cpp d:\a\agora\agora\source\scpp\lib\util\uint128_t.cpp d:\a\agora\agora\source\scpp\lib\xdrpp\xdrpp\marshal.cc d:\a\agora\agora\source\scpp\src\crypto\ByteSliceHasher.cpp d:\a\agora\agora\source\scpp\src\crypto\Hex.cpp d:\a\agora\agora\source\scpp\src\crypto\KeyUtils.cpp d:\a\agora\agora\source\scpp\src\crypto\SecretKey.cpp d:\a\agora\agora\source\scpp\src\crypto\SHA.cpp d:\a\agora\agora\source\scpp\src\crypto\StrKey.cpp d:\a\agora\agora\source\scpp\src\scp\BallotProtocol.cpp d:\a\agora\agora\source\scpp\src\scp\LocalNode.cpp d:\a\agora\agora\source\scpp\src\scp\NominationProtocol.cpp d:\a\agora\agora\source\scpp\src\scp\QuorumSetUtils.cpp d:\a\agora\agora\source\scpp\src\scp\SCP.cpp d:\a\agora\agora\source\scpp\src\scp\SCPDriver.cpp d:\a\agora\agora\source\scpp\src\scp\Slot.cpp d:\a\agora\agora\source\scpp\src\util\HashOfHash.cpp d:\a\agora\agora\source\scpp\src\util\Logging.cpp d:\a\agora\agora\source\scpp\src\util\Math.cpp d:\a\agora\agora\source\scpp\src\util\numeric.cpp
Error: unrecognized file extension o
dmd failed with exit code 1.
````

`Error: unrecognized file extension o`

The `cl` compile cpp was almost successful. (I made up to obj)
But does anyone know about that error message?
This does not seem to be a `cl` pid message. Is this a `dmd` error message?

File: source/scpp/build.d [hunk 2]
```diff
@@ -29,26 +29,64 @@ immutable SourcePath  = RootPath;
...
+        "/wd\"4592\"",
+        "/Zc:wchar_t",
+        "/Zi",
+        "/Gm-",
+        "/Od",
+        "/sdl",
+        "/Zc:inline",
+        "/fp:precise",
+        "/D \"BUILD_TESTS\"",
+        "/D \"USE_POSTGRES\"",
```
Geod24: We should be able to remove this.

source/scpp/build.d
```diff
@@ -29,26 +29,64 @@ immutable SourcePath  = RootPath;
...
+        "/D \"WIN32_LEAN_AND_MEAN\"",
+        "/D \"NOMINMAX\"",
+        "/D \"ASIO_STANDALONE\"",
+        "/D \"_WINSOCK_DEPRECATED_NO_WARNINGS\"",
+        "/D \"SODIUM_STATIC\"",
+        "/D \"ASIO_SEPARATE_COMPILATION\"",
+        "/D \"ASIO_ERROR_CATEGORY_NOEXCEPT=noexcept\"",
```
Geod24: Same for this

AndrejMitrovic:
On Windows the linker expects object files in the `.obj` format, not `.o`.

source/scpp/build.d
```diff
@@ -29,26 +29,64 @@ immutable SourcePath  = RootPath;
...
+        "/Gm-",
+        "/Od",
+        "/sdl",
+        "/Zc:inline",
+        "/fp:precise",
+        "/D \"BUILD_TESTS\"",
+        "/D \"USE_POSTGRES\"",
+        "/D \"WIN32_LEAN_AND_MEAN\"",
+        "/D \"NOMINMAX\"",
+        "/D \"ASIO_STANDALONE\"",
```
Geod24: We shouldn't have any use of ASIO so this will probably go

source/scpp/build.d
```diff
@@ -29,26 +29,64 @@ immutable SourcePath  = RootPath;
...
+        "/Gd",
+        "/MDd",
+        "/std:c++14",
+        "/FC",
+        "/EHsc",
+        "/c",
 ];
 
-immutable CppCmd = ["gcc" ] ~ CppFlags;
+immutable CppCmd = ["cl" ] ~ CppFlags;
```
Geod24:
```suggestion
version (Posix)
{
    immutable ObjPattern = "*.o";
    immutable CompilerIncludeFlag = "-I ";
    immutable CppFlags = [ /* ... */ ];
    immutable CppCmd = ["gcc" ] ~ CppFlags;
}
else version (Windows)
{
    immutable ObjPattern = "*.obj";
    immutable CompilerIncludeFlag = "/I ";
    immutable CppFlags = [ /* ... */ ];
    immutable CppCmd = ["cl" ] ~ CppFlags;
}
else
    static assert(0, "Unsupported platform");
```

Geod24:
That's going to be a problem, since object files are also hardcoded in `dub.json`.

source/scpp/build.d
```diff
@@ -29,26 +29,64 @@ immutable SourcePath  = RootPath;
...
 immutable BuildTarget = BuildPath.buildPath("libscp.o");
 
-/// Path passed to `-I`
+/// Path passed to `/I`
 immutable Includes = [
     SourcePath,
     SourcePath.buildPath("src"),
     SourcePath.buildPath("lib"),
     SourcePath.buildPath("lib", "xdrpp"),
+    SourcePath.buildPath("lib", "libsodium", "include"),
```
Geod24: Note that should shouldn't be necessary if `libsodium` is in the compiler search path

TrustHenry:
```
agora ~master: building configuration "unittest"...
submodules\ocean\src\ocean\core\array\Search.d(1251,13): Error: undefined identifier `ssize_t`, did you mean alias `size_t`?
submodules\ocean\src\ocean\time\Clock.d(167,25): Error: undefined identifier `timeval`
submodules\ocean\src\ocean\time\Clock.d(180,28): Error: undefined identifier `timeval`
submodules\ocean\src\ocean\stdc\posix\netdb.d(21,16): Error: undefined identifier `AI_PASSIVE`
submodules\ocean\src\ocean\stdc\posix\netdb.d(21,29): Error: undefined identifier `AI_CANONNAME`
submodules\ocean\src\ocean\stdc\posix\netdb.d(21,44): Error: undefined identifier `AI_NUMERICHOST`
submodules\ocean\src\ocean\stdc\posix\netdb.d(21,61): Error: undefined identifier `AI_NUMERICSERV`
submodules\ocean\src\ocean\stdc\posix\netdb.d(21,78): Error: undefined identifier `AI_ADDRCONFIG`
submodules\ocean\src\ocean\stdc\posix\netdb.d(22,19): Error: undefined identifier `AI_V4MAPPED`
submodules\ocean\src\ocean\stdc\posix\netdb.d(22,33): Error: undefined identifier `AI_ADDRCONFIG`
```

SCP compile succeeded Perhaps.
The next gate is waiting.

ocean...

TrustHenry:
```
source\agora\common\BanManager.d(115,23): Error: function agora.common.BanManager.BanManager.serialize(scope void delegate(scope const(ubyte)[]) @safe dg) const is not callable using argument types (void)
source\agora\common\BanManager.d(115,23):        cannot pass argument __lambda1 of type void to parameter scope void delegate(scope const(ubyte)[]) @safe dg
source\agora\common\BanManager.d(170,22): Error: function agora.common.BanManager.BanManager.banUntil(string address, int banned_until) is not callable using argument types (string, const(long))
source\agora\common\BanManager.d(170,22):        cannot pass argument ban_until of type const(long) to parameter int banned_until
source\agora\common\Config.d(341,63): Error: cannot implicitly convert expression get(cmdln, node) of type ulong to int
```

The `time_t` type is used as an `long` type in the `x86` window environment.

https://github.com/dlang/druntime/blob/44bcc2cf6a2a45b3c420837beb99b46ede66006d/src/core/stdc/time.d#L76

Geod24:
I will make the necessary changes to Ocean.

TrustHenry:
SourceFiles by OS of `dub.json` is divided into `sourceFiles-posix` and `sourceFiles-windows`

Geod24:
Try to change the ocean dependency to https://github.com/bpfkorea/ocean/tree/dev-2020-03-17

TrustHenry:
Updated Ocean to the dev-2020-03-17

Geod24:
As mentioned by @AndrejMitrovic , `rawRead` seems to have different attributes on Windows...
You can change `@safe` to `@trusted` (and add a comment, e.g. "// `rawRead` is not `@safe` on Windows because of `setmode`") here:
https://github.com/bpfkorea/agora/blob/dbd8a5cff0758cf13e59790dcbcc5100b63ba065/source/agora/common/BanManager.d#L94

And add `@trusted` after the delegate here: https://github.com/bpfkorea/agora/blob/dbd8a5cff0758cf13e59790dcbcc5100b63ba065/source/agora/common/BanManager.d#L114

source/agora/common/BanManager.d
```diff
@@ -108,7 +108,7 @@ public class BanManager
 
     ***************************************************************************/
 
-    public void dump ()
+    public void dump () @trusted
     {
         auto ban_file = File(this.banfile_path, "w");
         this.serialize((scope bytes) => ban_file.rawWrite(bytes));
```
Geod24: This is the line I pointed to, not L111.

```suggestion
        this.serialize((scope bytes) @trusted => ban_file.rawWrite(bytes));
```

source/agora/node/Node.d
```diff
@@ -61,14 +61,14 @@ private enum uint MaxBatchBlocksSent = 1000;
 
 /*******************************************************************************
 
-    Implementation of the Node API
+    Implementation of the Full Node API
```
Geod24: This change seems unrelated ?

source/agora/common/BanManager.d
Geod24: But the change line 94 was good...

source/agora/node/Node.d
TrustHenry: Oh, I accidentally merged.

TrustHenry:
```
agora ~master: building configuration "unittest"...
source\agora\common\BanManager.d(169,22): Error: function `agora.common.BanManager.BanManager.banUntil(string address, int banned_until)` is not callable using argument types `(string, const(long))`
source\agora\common\BanManager.d(169,22):        cannot pass argument `ban_until` of type `const(long)` to parameter `int banned_until`
source\agora\common\BitField.d(146,48): Deprecation: integral promotion not done for `~mask(index)`, use '-preview=intpromote' switch or `~cast(int)(mask(index))`
source\agora\common\BitField.d(146,48): Deprecation: integral promotion not done for `~mask(index)`, use '-preview=intpromote' switch or `~cast(int)(mask(index))`
source\agora\common\Config.d(338,63): Error: cannot implicitly convert expression `get(cmdln, node)` of type `ulong` to `int`
C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\phobos\std\range\primitives.d(174,38): Deprecation: `alias byKeyValue this` is deprecated - Iterate over .byKeyValue instead.
C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\phobos\std\range\primitives.d(176,27): Deprecation: `alias byKeyValue this` is deprecated - Iterate over .byKeyValue instead.
C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\phobos\std\range\primitives.d(174,38): Deprecation: `alias byKeyValue this` is deprecated - Iterate over .byKeyValue instead.
C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\phobos\std\range\primitives.d(176,27): Deprecation: `alias byKeyValue this` is deprecated - Iterate over .byKeyValue instead.
source\agora\node\BlockStorage.d(730,26): Error: template `agora.common.Serializer.serializePart` cannot deduce function from argument types `!()(ulong, void, CompactMode)`, candidates are:
source\agora\common\Serializer.d(346,13):        `serializePart(T)(auto ref scope const T record, scope SerializeDg dg, CompactMode compact = CompactMode.Yes)`
source\agora\node\BlockStorage.d(731,30): Error: `@safe` function `agora.node.BlockStorage.BlockStorage.saveIndex` cannot call `@system` function `std.stdio.File.rawWrite!ubyte.rawWrite`
C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\phobos\std\stdio.d(1035,10):        `std.stdio.File.rawWrite!ubyte.rawWrite` is declared here
source\agora\node\BlockStorage.d(732,26): Error: template `agora.common.Serializer.serializePart` cannot deduce function from argument types `!()(ulong, void, CompactMode)`, candidates are:
source\agora\common\Serializer.d(346,13):        `serializePart(T)(auto ref scope const T record, scope SerializeDg dg, CompactMode compact = CompactMode.Yes)`
source\agora\node\BlockStorage.d(776,29): Error: `@safe` delegate `agora.node.BlockStorage.BlockStorage.loadAllIndexes.__lambda1` cannot call `@system` function `std.stdio.File.rawRead!ubyte.rawRead`
C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\phobos\std\stdio.d(979,9):        `std.stdio.File.rawRead!ubyte.rawRead` is declared here
source\agora\node\BlockStorage.d(787,29): Error: `@safe` function `agora.node.BlockStorage.BlockStorage.loadAllIndexes` cannot call `@system` function `std.stdio.File.rawRead!ubyte.rawRead`
C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\phobos\std\stdio.d(979,9):        `std.stdio.File.rawRead!ubyte.rawRead` is declared here
Type 'Node' implements more than one interface: make sure the one describing the REST server is the first one
```

I see the end now.

Geod24:
For the call to `banUntil`, you can `cast` to `time_t` for the time being. I opened #665 about it.
And the other parts, you know how to deal with them.

TrustHenry:
```

Please report it via https://issues.dlang.org/enter_bug.cgi
with, preferably, a reduced, reproducible example and the information below.
DustMite (https://github.com/CyberShadow/DustMite/wiki) can help with the reduction.
---
DMD v2.090.1-dirty
predefs   Have_agora Have_ocean Have_base32 Have_bitblob Have_d2sqlite3 Have_dyaml Have_libsodiumd Have_localrest Have_protobuf Have_vibe_d Have_tinyendian Have_vibe_d_data Have_vibe_d_utils Have_stdx_allocator Have_vibe_d_core Have_vibe_core Have_vibe_d_crypto Have_vibe_d_http Have_vibe_d_inet Have_vibe_d_mail Have_vibe_d_mongodb Have_vibe_d_redis Have_vibe_d_stream Have_vibe_d_textfilter Have_vibe_d_tls Have_vibe_d_web Have_eventcore EventcoreWinAPIDriver Have_taggedalgebraic Have_mir_linux_kernel Have_diet_ng Have_openssl DigitalMars Windows CRuntime_Microsoft CppRuntime_Microsoft LittleEndian D_Version2 all D_SIMD D_InlineAsm_X86_64 X86_64 Win64 D_LP64 D_Coverage unittest assert D_ModuleInfo D_Exceptions D_TypeInfo D_HardFloat
binary    C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\dmd.exe
version   v2.090.1-dirty
config    C:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\sc.ini
DFLAGS    -IC:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\phobos -IC:\hostedtoolcache\windows\dc\dmd-2.090.1\x64\dmd2\windows\bin\..\..\src\druntime\import -L/OPT:NOICF
---

object.Error@(0): Access Violation
----------------
0x00710D48
0x00673439
dmd failed with exit code 1.
```

.github/workflows/github_pr.yml
```diff
@@ -56,3 +56,47 @@ jobs:
         sudo apt-get install libsodium-dev
         sudo apt-get install g++-9
         ./ci/run.sh
+
+  windows_build:
+    name: Windows
+    strategy:
+      matrix:
+        os: [windows-2019]
+        dc: [dmd-2.090.1]
```
Geod24:
```suggestion
        dc: [dmd-2.091.0]
```

TrustHenry: for God's sake ..

TrustHenry:
```
Linking...
LINK : fatal error LNK1104: cannot open file `'sodium.lib'`
Error: linker exited with status 1104
dmd failed with exit code 1.
```
Is `sodium.lib` the library used by `libsodiumd` here?
https://github.com/Geod24/libsodiumd/blob/fbe7b2d888441517b0de8461f11cc05e5a00580f/dub.json#L5

Separately, scpp is referring to the `libsodium` library.

I cannot find `sodium` library.
Where can I find this?

Geod24:
You download it and install it here:
https://github.com/bpfkorea/agora/pull/624/files#diff-0fe23b2f32fe180ba30831c917c6e237R95
This path needs to be passed to the compiler as a LIBPATH.

Geod24:
> LINK : warning LNK4098: defaultlib 'MSVCRTD' conflicts with use of other libs; use /NODEFAULTLIB:library

This seems relevant

TrustHenry:
```
Searching libraries
    Searching C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.24.28314\lib\x64\legacy_stdio_definitions.lib:
    Searching submodules\vibe.d\lib\win-amd64\libssl.lib:
    Searching submodules\vibe.d\lib\win-amd64\libcrypto.lib:
    Searching submodules\base32\.dub\build\library-unittest-windows-x86_64-dmd_2091-F96632DD24E4BB7559754C4B22AD6B51\base32.lib:
    Searching submodules\bitblob\.dub\build\library-unittest-windows-x86_64-dmd_2091-1E38AC8FF36E52FB60BB39042CDB4FD1\bitblob.lib:
    Searching submodules\d2sqlite3\.dub\build\with-lib-unittest-windows-x86_64-dmd_2091-E02DEE9CB220D073263251D62B32AB90\d2sqlite3.lib:
    Searching submodules\dyaml\.dub\build\library-unittest-windows-x86_64-dmd_2091-057AF4704C6EBC19F389F64CAFDBEF8B\dyaml.lib:
    Searching submodules\tinyendian\.dub\build\library-unittest-windows-x86_64-dmd_2091-A7E46FA0316983DAC9892BC44F8C0727\tinyendian.lib:
    Searching submodules\libsodiumd\.dub\build\library-unittest-windows-x86_64-dmd_2091-10DA5E1828D3CA10B006845BB748142C\libsodiumd.lib:
    Searching submodules\localrest\.dub\build\library-unittest-windows-x86_64-dmd_2091-8D1B163906934EC7C418967CFE852A87\localrest.lib:
    Searching submodules\protobuf-d\.dub\build\protobuf-unittest-windows-x86_64-dmd_2091-69CA6801CF9865A2E8B90960460A2A57\protobuf.lib:
    Searching submodules\vibe.d\.dub\build\vibe-core-unittest-windows-x86_64-dmd_2091-797DBCD16B2846B1F6CA88E47CCF4613\vibed.lib:
    Searching submodules\vibe.d\mail\.dub\build\library-unittest-windows-x86_64-dmd_2091-646AE7069CECC29137F275A017597111\vibe-d_mail.lib:
    Searching submodules\vibe.d\mongodb\.dub\build\library-unittest-windows-x86_64-dmd_2091-4664D1D0367DAC60DC445F7A3E69F956\vibe-d_mongodb.lib:
    Searching submodules\vibe.d\redis\.dub\build\library-unittest-windows-x86_64-dmd_2091-3EB52B6F4C586F51E3836AEDA93ADF2E\vibe-d_redis.lib:
    Searching submodules\vibe.d\web\.dub\build\library-unittest-windows-x86_64-dmd_2091-99F81FB6E794821CC98C8CFBA9515E1A\vibe-d_web.lib:
    Searching submodules\vibe.d\http\.dub\build\library-unittest-windows-x86_64-dmd_2091-9EAA8C9B32E1E3F6DC0DA5A5391E021C\vibe-d_http.lib:
    Searching submodules\diet-ng\.dub\build\library-unittest-windows-x86_64-dmd_2091-E87C6ED221625E750EAA06726D2867CD\diet-ng.lib:
    Searching submodules\vibe.d\crypto\.dub\build\library-unittest-windows-x86_64-dmd_2091-D35514AA73A79F75BF04D2E45B05260D\vibe-d_crypto.lib:
    Searching submodules\mir-linux-kernel\.dub\build\library-unittest-windows-x86_64-dmd_2091-1A7AC4D4DC6880060B40785EBFF6C675\mir-linux-kernel.lib:
    Searching submodules\vibe.d\inet\.dub\build\library-unittest-windows-x86_64-dmd_2091-F12EA0AC10AECF4B2EEFBAE0202DFD22\vibe-d_inet.lib:
    Searching submodules\vibe.d\textfilter\.dub\build\library-unittest-windows-x86_64-dmd_2091-FA3E2DA0A465C0C1B32A942AA06F3371\vibe-d_textfilter.lib:
    Searching submodules\vibe.d\tls\.dub\build\openssl-mscoff-unittest-windows-x86_64-dmd_2091-4352A905B8994A50498B44D159CD172C\vibe-d_tls.lib:
    Searching submodules\vibe.d\stream\.dub\build\library-unittest-windows-x86_64-dmd_2091-15304CD4D5F1FC63D75779C1275B51BF\vibe-d_stream.lib:
    Searching submodules\vibe-core\.dub\build\winapi-unittest-windows-x86_64-dmd_2091-0EBACCF4301781FB94E7534E9B1A8F11\vibe_core.lib:
    Searching submodules\eventcore\.dub\build\winapi-unittest-windows-x86_64-dmd_2091-6857AD3390F5F26E0300A0F74A129D3C\eventcore.lib:
    Searching submodules\taggedalgebraic\.dub\build\library-unittest-windows-x86_64-dmd_2091-81A30A978BA8067B55699884A763501B\taggedalgebraic.lib:
    Searching submodules\vibe.d\data\.dub\build\library-unittest-windows-x86_64-dmd_2091-394432F1AABCC4F34B2C9E1E30B9D9F4\vibe-d_data.lib:
    Searching submodules\vibe.d\utils\.dub\build\library-unittest-windows-x86_64-dmd_2091-F3765649E5BE94568637DB7BA1DA2F8E\vibe-d_utils.lib:
    Searching submodules\stdx-allocator\.dub\build\library-unittest-windows-x86_64-dmd_2091-D293CE79B26032EF56DDC967107D49B4\stdx-allocator.lib:
LINK : fatal error LNK1104: cannot open file 'sqlite3.lib'
Error: linker exited with status 1104
```

I set `lflags` for windowing as above.

```
"lflags-windows": [
        "/verbose:lib",
        "/nodefaultlib:msvcrt",
        "/nodefaultlib:vcruntime"
    ],

`LINK : fatal error LNK1104: cannot open file 'sqlite3.lib'`
```

    Searching submodules\d2sqlite3\.dub\build\with-lib-unittest-windows-x86_64-dmd_2091-E02DEE9CB220D073263251D62B32AB90\d2sqlite3.lib:

I wonder if d2sqlite3.lib and sqlite3.lib are not matched.

Geod24:
`d2sqlite3.lib` is the static library built from d2sqlite3 sources, not `sqlite3` itself.

TrustHenry:
```
agora-unittests.obj : error LNK2019: unresolved external symbol "void __cdecl stellar::normalizeQSet(struct stellar::SCPQuorumSet &,struct stellar::PublicKey const * const)" (?normalizeQSet@stellar@@YAXAEAUSCPQuorumSet@1@QEBUPublicKey@1@@Z) referenced in function _D5agora9consensus8protocol9NominatorQk20verifyBuildSCPConfigFKxSQCl6common6Config12QuorumConfigZS4scpd5types11Stellar_SCP12SCPQuorumSet
agora-unittests.obj : error LNK2001: unresolved external symbol "public: virtual struct xdr::xvector<unsigned char,4294967292> __cdecl stellar::SCPDriver::extractValidValue(unsigned __int64,struct xdr::xvector<unsigned char,4294967292> const &)" (?extractValidValue@SCPDriver@stellar@@UEAA?AU?$xvector@E$0PPPPPPPM@@xdr@@_KAEBU34@@Z)
agora-unittests.obj : error LNK2001: unresolved external symbol "public: virtual void * __cdecl stellar::SCPDriver::getValueString(struct xdr::xvector<unsigned char,4294967292> const &)const " (?getValueString@SCPDriver@stellar@@UEBAPEAXAEBU?$xvector@E$0PPPPPPPM@@xdr@@@Z)
agora-unittests.obj : error LNK2001: unresolved external symbol "public: virtual void * __cdecl stellar::SCPDriver::toStrKey(struct stellar::PublicKey const &,bool)const " (?toStrKey@SCPDriver@stellar@@UEBAPEAXAEBUPublicKey@2@_N@Z)
agora-unittests.obj : error LNK2001: unresolved external symbol "public: virtual void * __cdecl stellar::SCPDriver::toShortString(struct stellar::PublicKey const &)const " (?toShortString@SCPDriver@stellar@@UEBAPEAXAEBUPublicKey@2@@Z)
```

TrustHenry:
```
agora-unittests.obj : error LNK2019: unresolved external symbol "unsigned __int64 __cdecl cppSizeOf(class stellar::SCPQuorumSet &)" (unsigned __int64 __cdecl cppSizeOf(class stellar::SCPQuorumSet & __ptr64) referenced in function _D4scpd5tests8SizeTest17__unittest_L26_C1FZv
  Hint on symbols that are defined and could potentially match:
    "unsigned __int64 __cdecl cppSizeOf(struct xdr::opaque_array<4> &)" (unsigned __int64 __cdecl cppSizeOf(struct xdr::opaque_array<4> & __ptr64)
    "unsigned __int64 __cdecl cppSizeOf(struct xdr::opaque_array<32> &)" (unsigned __int64 __cdecl cppSizeOf(struct xdr::opaque_array<32> & __ptr64)
```
I don't know how to solve this mangling here. I need help.

https://github.com/bpfkorea/agora/blob/0be6c03ee54485b6adbd4116b23a1d13b926e6a0/source/scpd/tests/SizeTest.d#L22

TrustHenry:
```
agora-unittests.obj : error LNK2019: unresolved external symbol "struct FieldInfo __cdecl cppFieldInfo(class stellar::SCPQuorumSet &,char const *)" (struct FieldInfo __cdecl cppFieldInfo(class stellar::SCPQuorumSet & __ptr64,char const * __ptr64) referenced in function _D4scpd5tests10LayoutTest17__unittest_L43_C1FZv
  Hint on symbols that are defined and could potentially match:
    "struct FieldInfo __cdecl cppFieldInfo(struct xdr::opaque_array<32> &,char const *)" (struct FieldInfo __cdecl cppFieldInfo(struct xdr::opaque_array<32> & __ptr64,char const * __ptr64)
    "struct FieldInfo __cdecl cppFieldInfo(struct stellar::PublicKey &,char const *)" (struct FieldInfo __cdecl cppFieldInfo(struct stellar::PublicKey & __ptr64,char const * __ptr64)
    "struct FieldInfo __cdecl cppFieldInfo(struct stellar::SCPBallot &,char const *)" (struct FieldInfo __cdecl cppFieldInfo(struct stellar::SCPBallot & __ptr64,char const * __ptr64)
```
^ mongling problem.

https://github.com/bpfkorea/agora/blob/0be6c03ee54485b6adbd4116b23a1d13b926e6a0/source/scpd/tests/LayoutTest.d#L36

source/scpd/types/Stellar_SCP.d
```diff
@@ -161,7 +161,7 @@ struct SCPEnvelope {
 
 static assert(SCPEnvelope.sizeof == 168);
 
-struct SCPQuorumSet {
+extern(C++, class) struct SCPQuorumSet {
```
Geod24: Why?

source/scpd/types/Stellar_types.d
```diff
@@ -51,7 +51,7 @@ enum SignerKeyType : int32_t {
 /// Modified to suits scpd's needs
 /// Note: should only be used within this package
 package(scpd)
-struct PublicKey {
+extern(C++, class) struct PublicKey {
```
Geod24: Why?

source/scpd/types/XDRBase.d
```diff
@@ -102,7 +102,7 @@ struct opaque_array(uint32_t N = XDR_MAX_LEN)
 
 ///alias xvector(T, uint32_t N = XDR_MAX_LEN) = vector!(T);
 alias opaque_vec(uint32_t N = XDR_MAX_LEN) = xvector!(uint8_t, N);
-struct xvector(T, uint32_t N = XDR_MAX_LEN)
+extern(C++, class) struct xvector(T, uint32_t N = XDR_MAX_LEN)
```
Geod24: Why?

source/scpd/types/Stellar_SCP.d
TrustHenry: https://github.com/stellar/stellar-core/blob/dd7f38dba232e20ae75a95641857c0b7ed0252e1/src/xdr/Stellar-SCP.x#L80

This is not necessary because it has the same struct c++ side and d.
I'll delete it.

source/scpd/types/Stellar_types.d
TrustHenry: Incorrectly applied because it should only be used within spcd.

source/scpd/types/XDRBase.d
TrustHenry: I'll delete it.

TrustHenry:
```
The Current unresolved external symbol list

class xdr::opaque_array<32> __cdecl stellar::sha256(struct stellar::ByteSlice const &)
void __cdecl stellar::normalizeQSet(struct stellar::SCPQuorumSet &,struct stellar::PublicKey const * const)
public: virtual struct xdr::xvector<unsigned char,4294967292> __cdecl stellar::SCPDriver::extractValidValue(unsigned __int64,struct xdr::xvector<unsigned char,4294967292> const &)
public: virtual void * __cdecl stellar::SCPDriver::getValueString(struct xdr::xvector<unsigned char,4294967292> const &)const 
public: virtual void * __cdecl stellar::SCPDriver::toStrKey(struct stellar::PublicKey const &,bool)const 
public: virtual void * __cdecl stellar::SCPDriver::toShortString(struct stellar::PublicKey const &)const 
public: virtual class std::chrono::duration<__int64,class std::ratio<1,1000> > __cdecl stellar::SCPDriver::computeTimeout(unsigned int)
public: virtual void __cdecl stellar::SCPDriver::nominatingValue(unsigned __int64,struct xdr::xvector<unsigned char,4294967292> const &)
public: virtual void __cdecl stellar::SCPDriver::nominatingValue(unsigned __int64,struct xdr::xvector<unsigned char,4294967292> const &)
public: virtual void __cdecl stellar::SCPDriver::updatedCandidateValue(unsigned __int64,struct xdr::xvector<unsigned char,4294967292> const &)
public: virtual void __cdecl stellar::SCPDriver::startedBallotProtocol(unsigned __int64,struct stellar::SCPBallot const &)
public: virtual void __cdecl stellar::SCPDriver::acceptedBallotPrepared(unsigned __int64,struct stellar::SCPBallot const &)
public: virtual void __cdecl stellar::SCPDriver::confirmedBallotPrepared(unsigned __int64,struct stellar::SCPBallot const &)
public: virtual void __cdecl stellar::SCPDriver::acceptedCommit(unsigned __int64,struct stellar::SCPBallot const &)
public: virtual void __cdecl stellar::SCPDriver::ballotDidHearFromQuorum(unsigned __int64,struct stellar::SCPBallot const &)
class std::set<unsigned int> * __cdecl makeTestSet(void)
class FieldInfo __cdecl cppFieldInfo(struct stellar::SCPBallot &,char const *)
class FieldInfo __cdecl cppFieldInfo(struct stellar::SCPStatement::_pledges_t::_prepare_t &,char const *)
class FieldInfo __cdecl cppFieldInfo(struct stellar::SCPStatement::_pledges_t::_confirm_t &,char const *)
class FieldInfo __cdecl cppFieldInfo(struct stellar::SCPStatement::_pledges_t &,char const *)
class FieldInfo __cdecl cppFieldInfo(struct stellar::SCPStatement &,char const *)
class FieldInfo __cdecl cppFieldInfo(struct stellar::SCPEnvelope &,char const *)
class FieldInfo __cdecl cppFieldInfo(struct stellar::SCPQuorumSet &,char const *)
class FieldInfo __cdecl cppFieldInfo(struct stellar::PublicKey &,char const *)
class FieldInfo __cdecl cppFieldInfo(class xdr::opaque_array<32> &,char const *)
class FieldInfo __cdecl cppFieldInfo(class stellar::ByteSlice &,char const *)
bool __cdecl isHashTypeCompatible(class xdr::opaque_array<32> &,char const * const)
unsigned __int64 __cdecl cppSizeOf(class xdr::opaque_array<32> &)
unsigned __int64 __cdecl cppSizeOf(class xdr::opaque_array<4> &)
void __cdecl push_back<class stellar::Slot::HistoricalStatement,class std::vector<class stellar::Slot::HistoricalStatement,class std::allocator<class stellar::Slot::HistoricalStatement> > >(class std::vector<class stellar::Slot::HistoricalStatement,class std::allocator<class stellar::Slot::HistoricalStatement> > &,class stellar::Slot::HistoricalStatement &)
````

File: source/scpp/build.d [hunk 13]
```diff
@@ -35,27 +35,83 @@ immutable SourcePath  = RootPath;
...
+        "/wd\"4060\"",
+        "/wd\"4100\"",
+        "/wd\"4127\"",
+        "/wd\"4324\"",
+        "/wd\"4408\"",
+        "/wd\"4510\"",
+        "/wd\"4512\"",
+        "/wd\"4582\"",
+        "/wd\"4583\"",
+        "/wd\"4592\"",
```
Geod24:
```suggestion
        "/wd\"4592\"",
        "/wd\"4099\"",
```

source/scpp/build.d
```diff
@@ -35,27 +35,83 @@ immutable SourcePath  = RootPath;
...
+else version (Windows)
+{
+    immutable ObjPattern = "*.obj";
+    immutable CompilerIncludeFlag = "/I ";
+    immutable CppFlags = [
+        "/JMC",
+        "/MP",
+        "/GS",
+        "/W4",
+        "/wd\"4060\"",
```
Geod24:
```suggestion
        "/wd\"4060\"",
        "/wd\"4099\"",
```

TrustHenry:
I'm looking for this error.

```
agora-unittests.obj : error LNK2019: unresolved external symbol strncasecmp referenced in function _D5ocean4text5Ascii8icompareFAxaQdZi715
agora-unittests.obj : error LNK2019: unresolved external symbol log10l referenced in function _D5ocean4text7convert5Float8convertlFPaeiPiQciZPxa716
agora-unittests.obj : error LNK2019: unresolved external symbol ceill referenced in function _D5ocean4text7convert5Float8convertlFPaeiPiQciZPxa717
agora-unittests.obj : error LNK2019: unresolved external symbol modfl referenced in function _D5ocean4text7convert5Float8convertlFPaeiPiQciZPxa718
````

https://github.com/bpfkorea/ocean/blob/b7baf13679b39a9a2dc38801a01c9f3ddd078ab6/src/ocean/text/convert/Float.d#L50-L56
```
private extern (C)
{
    real log10l (real x);
    real ceill (real num);
    real modfl (real num, real *i);
    real powl  (real base, real exp);
}
```
https://github.com/bpfkorea/ocean/blob/b7baf13679b39a9a2dc38801a01c9f3ddd078ab6/src/ocean/stdc/string.d#L26

```
extern (C):
char *strsignal(int sig);
int strcasecmp(in char *s1, in char *s2);
int strncasecmp(in char *s1, in char *s2, size_t n);
````

**Comment by TrustHenry** [comment]
ldc2 the LLVM D compiler (1.20.1):  running X86 AMD
Found during compilation test with ldc2.
`submodules\ocean\src\ocean\math\Math.d(1718,5): Error: static assert:  `infL <= 1.79769e+308L` is false`
https://github.com/bpfkorea/ocean/blob/b7baf13679b39a9a2dc38801a01c9f3ddd078ab6/src/ocean/math/Math.d#L1718

real type ( Implementation Note: 80 bits for x86 CPUs or double size, whichever is larger )

File: .github/workflows/github_pr.yml [hunk 15]
```diff
@@ -22,8 +22,7 @@ jobs:
         os: [ ubuntu-18.04, macOS-10.15, windows-2019 ]
         dc: [ dmd-2.091.0, ldc-1.20.1, ldc-1.19.0 ]
         exclude:
-          # Not supported yet
-          - os: windows-2019
+          - { os: windows-2019, dc: dmd-2.091.0 }
```
Geod24, line 25:
```suggestion
          - ubuntu-18.04
          - macOS-10.15
          - { os: windows-2019, dc: dmd-2.091.0 }
```

For the time being, to allow you to get Windows working

source/scpd/Cpp.d
```diff
@@ -194,15 +194,16 @@ public alias SCPCallback = extern(C++) void function();
...
-unittest
-{
-    auto set = makeTestSet;
-    assert(!set.empty);
-    uint[] values;
-    foreach (val; *set)
-        values ~= val;
-    assert(values == [1, 2, 3, 4, 5]);
-}
+// TODO : For MIPS64 test
```
Geod24: Where did you read about MIPS64 ?
This is MIPS64: https://en.wikipedia.org/wiki/MIPS_architecture
I suppose it was somewhere in Ocean ? In any case, MIPS64 has *absolutely nothing to do* with your issues, so I have absolutely no idea why it is being mentioned here.

source/scpd/scp/BallotProtocol.d
```diff
@@ -55,7 +55,7 @@ extern(C++, class) public struct BallotProtocol
         SCP_PHASE_NUM
     }
     // human readable names matching SCPPhase
-    static extern __gshared const(char*)[] phaseNames;
+    extern __gshared const(char*)[SCPPhase.max] phaseNames;
```
Geod24: This can be its own PR since the binding is wrong

source/scpd/types/Utils.d
```diff
@@ -91,4 +90,11 @@ extern(C++, `stellar`)
             return (cast(T*)this.mData)[0 .. this.mSize / T.sizeof];
         }
     }
+
+    // Note: This needs to be after `ByteSlice` because the frontend doesn't take
+    // `extern(C++, class)` into account otherwise
+    // https://issues.dlang.org/show_bug.cgi?id=20700
+    // https://issues.dlang.org/show_bug.cgi?id=20701
```
Geod24: 20701 does not apply here, only 20700

source/scpd/scp/Slot.d
```diff
@@ -45,7 +45,7 @@ private:
 
     // keeps track of all statements seen so far for this slot.
     // it is used for debugging purpose
-    struct HistoricalStatement
+    extern(C++, struct) struct HistoricalStatement
```
Geod24, line 49: 20701 applies here

source/scpd/scp/SCPDriver.d
```diff
@@ -109,8 +128,17 @@ nothrow:
     // `computeTimeout` computes a timeout given a round number
     // it should be sufficiently large such that nodes in a
     // quorum can exchange 4 messages
-    milliseconds computeTimeout(uint32_t roundNumber);  // Slot in the vtable
-
+    version (Windows)
+    {
+        // TODO Temporary action due to MSVC muggling problem
```
Geod24: mangling, not muggling.
Also are you sure it's 20700 ? It *might be related* to it.

source/scpd/scp/SCPDriver.d
```diff
@@ -71,18 +71,37 @@ nothrow:
...
-    // `toStrKey` returns StrKey encoded string representation
-    //std::string toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
-    void* toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
-
-    // `toShortString` converts to the common name of a key if found
-    //std::string toShortString(ref const(PublicKey) pk) const;
-    void* toShortString(ref const(PublicKey) pk) const;
+    version(Windows)
+    {
+        // TODO: Take temporary action to support mangling in Windows MSVC
```
Geod24, line 76: Can you add a comment explaining *why* we need to take those actions on Windows and not POSIX ?

dub.json
```diff
@@ -83,7 +57,71 @@
...
+        "source/scpp/build/crc16.obj",
+        "source/scpp/build/jsoncpp.obj",
+        "source/scpp/build/marshal.obj",
+        "source/scpp/build/numeric.obj",
+        "source/scpp/build/uint128_t.obj",
+        "lib/win64/sodium.lib"
+    ],
+    "lflags-posix": [ "-lsodium", "-lstdc++", "-lsqlite3" ],
+    "dflags-windows": [
+        "-mtriple=x86_64-windows-msvc"
```
Geod24: Is this necessary now ? Also you removed the `-extern-std` flag

ci/run.bat
```diff
@@ -0,0 +1,4 @@
+dub test --skip-registry=all --compiler=%DC%
+::rdmd --compiler=%DC% ./tests/runner.d --compiler=%DC% -cov
+dub build --skip-registry=all --compiler=%DC%
+dub build -c client --skip-registry=all --compiler=%DC%
```
Geod24: Why does this need to be a `bat` script ?

source/scpp/build.d
```diff
@@ -35,27 +35,102 @@ immutable SourcePath  = RootPath;
...
+        // Remove warning 'identifier' : unreferenced formal parameter
+        "/wd\"4100\"",
+        // https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4127?view=vs-2019
+        // Remove warning : conditional expression is constant
+        "/wd\"4127\"",
+        // https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4324?view=vs-2019
+        // Remove warning 'struct_name' : structure was padded due to __declspec(align())
+        "/wd\"4324\"",
+        // https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-4-c4408?view=vs-2019
+        // Remove warning : anonymousstruct or union did not declare any data members
```
Geod24:
```suggestion
        // Remove warning : anonymous struct or union did not declare any data members
```

source/scpd/Cpp.d
TrustHenry: The problem with this is the mangling issue.
Annotation has been modified 'MSVC mangling issue'

Geod24: Mangling issue does not tell me much... What fails to mangle properly ? Is it `empty` ? `makeTestSet` ? And why ?

dub.json
```diff
@@ -83,7 +57,70 @@
             ]
         }
     ],
-
+    "sourceFiles-posix": [
```
Geod24, line 15: Why is this moved after the configurations array ? Any reason ?

source/scpd/scp/SCPDriver.d
```diff
@@ -71,18 +71,38 @@ nothrow:
...
+        // getValueString, toStrKey, toShortString: Return value missing
```
Geod24: Good, that's what I wanted to see, thanks. Here is how I would phrase it:

```suggestion
        // MSVC mangles the return value, while POSIX does not,
        // and `getValueString`, `toStrKey`, and `toShortString` return `std::string` for which we need bindings 
```

dub.json
```diff
@@ -83,7 +57,70 @@
...
+        "source/scpp/build/StrKey.obj",
+        "source/scpp/build/crc16.obj",
+        "source/scpp/build/jsoncpp.obj",
+        "source/scpp/build/marshal.obj",
+        "source/scpp/build/numeric.obj",
+        "source/scpp/build/uint128_t.obj",
+        "lib/win64/sodium.lib"
+    ],
+    "lflags-posix": [ "-lsodium", "-lstdc++", "-lsqlite3" ],
+    "dflags-windows": [
```
Geod24: This needs to be on both platforms, so just leave it as it was originally

source/scpp/src/scp/Slot.h
```diff
@@ -181,5 +180,9 @@ class Slot : public std::enable_shared_from_this<Slot>
   protected:
     std::vector<SCPEnvelope> getEntireCurrentState();
     friend class TestSCP;
+
+  private:
+    // BPFK note: cannot be private as we require runtime layout checks
+    std::vector<HistoricalStatement> mStatementsHistory;
```
Geod24: Wait, isn't this changing the layout ?

ci/run.bat
TrustHenry: vcvars64 environment variables are readily available.
run without writing `bat` script Should it be implemented in `run.sh`?

source/scpp/src/scp/Slot.h
TrustHenry: I'm not sure yet.

source/scpp/src/scp/SCPDriver.cpp
```diff
@@ -98,4 +98,37 @@ SCPDriver::computeTimeout(uint32 roundNumber)
     }
     return std::chrono::seconds(timeoutInSeconds);
 }
+
+Value SCPDriver::extractValidValue(uint64 slotIndex, Value const& value)
```
Geod24, line 102: I'm pretty sure you could revert this change if you remove `/Zc:inline`.
See the doc https://docs.microsoft.com/en-us/cpp/build/reference/zc-inline-remove-unreferenced-comdat?view=vs-2019

ci/run.bat
TrustHenry: run in bash while maintaining msbuild environment variable.Could not find a way to run sh.
Currently, this method is maintained. Let me know if there's a better way later.

.github/workflows/github_pr.yml
```diff
@@ -61,12 +60,40 @@ jobs:
...
+        $sha256hash = "C1D48D85C9361E350931FFE5067559CD7405A697C655D26955FB568D1084A5F4"
+        Write-Host ('Downloading {0} ...' -f $url)
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+        # See https://github.com/PowerShell/PowerShell/issues/2138
+        $ProgressPreference = 'SilentlyContinue'
+        Invoke-WebRequest -Uri $url -OutFile 'libsodium.zip'
+        if ((Get-FileHash libsodium.zip -Algorithm "SHA256").Hash -ne $sha256hash) {
+          exit 1
+        }
+        Expand-Archive libsodium.zip -DestinationPath d:\a\agora\agora\source\scpp\lib\
```
Geod24:
```suggestion
        Expand-Archive libsodium.zip -DestinationPath ${{ github.workspace }}\agora\source\scpp\lib\
```

SSame for the other instances of `d:\a\agora\`

TrustHenry: Applied suggest.

Geod24: Hum it looks like the replacement is for `d:\a\agora\agora\`, not `d:\a\agora\`

dub.json
```diff
@@ -37,9 +37,37 @@
...
+        "source/scpp/build/marshal.obj",
+        "source/scpp/build/numeric.obj",
+        "source/scpp/build/uint128_t.obj",
+        "lib/win64/sodium.lib"
+    ],
 
-    "dflags": [ "-extern-std=c++14" ],
-    "lflags": [ "-lsodium", "-lstdc++", "-lsqlite3" ],
+    "dflags-posix":   [ "-extern-std=c++14"],
+    "dflags-windows": [ "-extern-std=c++14", "-mtriple=x86_64-windows-msvc" ],
```
Geod24: Can you check if the triple is necessary ? That should be the default (e.g. on your machine, do `ldc2 --version` and you'll see a "default target")

TrustHenry:
```
Run ${DC} --version
LDC - the LLVM D compiler (1.20.0):
  based on DMD v2.090.1 and LLVM 9.0.1
  built with LDC - the LLVM D compiler (1.20.1)
  Default target: x86_64-pc-windows-msvc
  Host CPU: skylake
```
I don't think I need it after testing it.
Default values are set according to the platform.

Geod24 (commented):
Sweet! There is one last thing that needs cleanup, and it is the library handling.

On Linux, you usually pass two arguments to the compiler. One is the library name, which is mandatory, for example `-lsodium`. The other one is library search path, for example `-L/usr/lib/`.

Now library search path has evolved a lot over the years, and there's various things to take into account. First, some paths are included by default (e.g. `/usr/lib/`). Some of those includes are standard accross distributions, some are compiler-specific. Additionally the compiler usually read an environment variable (`LIBRARY_PATH` with GCC). And lastly, you can use the`-L/path/to/directory/` argument to provide your own, as mentioned above.

Now, since there are many distributions, filesystem hierarchy, configurations, etc... it could be a nightmare for people to write build scripts which are portable across all Linux distributions. Luckily, in order to abstract those differences (among others, like flags to pass when compiling), the open-source community created [`pkg-config`](https://en.wikipedia.org/wiki/Pkg-config).

This is what `dub` uses when you provide it with a name in `libs`. However, `pkg-config` relies on definitions file, and themselves, they have to be found... So the same problem can arise, but less likely, and it's easier to solve. If you look in the scripts, and in your `~/.bashrc`, we set `PKG_CONFIG` so that the definition file for sqlite3 can be found on Mac OSX. And on Linux, since the libraries are installed at a standard location, we don't have that problem.

Now that was for Linux/Mac, but how does it work on Windows ? Short answer: I'm not sure of the details. What your `dub.json` is doing is:
1) Directly provide `libsodium` as an argument to the file being built on line 64 (in `sourceFiles-windows`);
2) Use `-Llib\\win64` in `lflags-windows`
3) Copy `sqlite3.lib` to the binary directly, which is only required for DLL AFAIK (L118 - 120)
4) Use the `all-included` subConfiguration`

Now, from a quick look, it *seems like* the `-Llib\\win64` is not necessary (because `-L` is only passing a path, you'd need an equivalent to `-l`) and that passing the object file is the best option.
It also seems that copying `sqlite3` is not used anymore, since we use `all-included`.
However, the problem with `all-included` is that it has an adverse effect on POSIX. We would like to only use it on Windows. *However* subConfigurations  don't accept platform suffixes.

So I would say, step 1, remove the `copyFiles` and the `-L` flags and see if it's actually useful.
Step 2, see if we can move those binary we download *out* of the source tree.
And step 3, I'll handle it, rework the usage of `subConfigurations`.

dub.json
```diff
@@ -37,9 +37,37 @@
...
+        "source/scpp/build/marshal.obj",
+        "source/scpp/build/numeric.obj",
+        "source/scpp/build/uint128_t.obj",
+        "lib/win64/sodium.lib"
+    ],
 
-    "dflags": [ "-extern-std=c++14" ],
-    "lflags": [ "-lsodium", "-lstdc++", "-lsqlite3" ],
+    "dflags-posix":   [ "-extern-std=c++14"],
+    "dflags-windows": [ "-extern-std=c++14"],
```
Geod24: You can merge them then :)

TrustHenry: Yes, it's applied

TrustHenry: Oh..."lflags": [ "-lsodium", "-lstdc++", "-lsqlite3" ],   delete this?

TrustHenry: I misread the changed line.

TrustHenry:
> So I would say, step 1, remove the copyFiles and the -L flags and see if it's actually useful. Step 2, see if we can move those binary we download out of the source tree. And step 3, I'll handle it, rework the usage of subConfigurations.

Thank you for your detailed explanation.
Removed the `copyFiles` in Step1
But I couldn't find a way to replace `-Lib\\win64`
Step2 I don't understand what it means. Can you explain more?

.github/workflows/github_pr.yml
```diff
@@ -61,12 +60,40 @@ jobs:
...
+        [Net.ServicePointManager]::SecurityProtocol = [Net.SecurityProtocolType]::Tls12
+        # See https://github.com/PowerShell/PowerShell/issues/2138
+        $ProgressPreference = 'SilentlyContinue'
+        New-Item -ItemType directory -Path ${{ github.workspace }}\lib\win64
+        Invoke-WebRequest -Uri $url -OutFile '${{ github.workspace }}\lib\libsodium.zip'
+        if ((Get-FileHash '${{ github.workspace }}\lib\libsodium.zip' -Algorithm "SHA256").Hash -ne $sha256hash) {
+          exit 1
+        }
+        Expand-Archive '${{ github.workspace }}\lib\libsodium.zip' -DestinationPath ${{ github.workspace }}\lib\
+        echo "::set-env name=LIB::${{ github.workspace }}\lib\libsodium\x64\Release\v140\static\;${{ github.workspace }}\submodules\d2sqlite3\lib\win64\;$LIB"
```
Geod24: Why were you using v140 here ? It should be v142: https://docs.microsoft.com/en-us/cpp/porting/binary-compat-2015-2017?view=vs-2019

Geod24 (approved):
Green at last

ci/run.bat
```diff
@@ -0,0 +1,4 @@
+call dub test --skip-registry=all --compiler=%DC%
```
AndrejMitrovic, line 1: Why are these `call` commands instead of regular commands?

TrustHenry: The Microsoft help for the CALL command rather misleadingly states "Calls one batch program from another without stopping the parent batch program" it is true that the parent does not STOP, but it does PAUSE while the second script runs.

TrustHenry: But there's a little bit of improvement here.
Whether the ERROR LEVEL is output correctly.

AndrejMitrovic: I think these calls should not be asynchronous though. So it is right that each command should run one at a time. So the `call` is unnecessary I think..

AndrejMitrovic:
There seems to have been some test-suite failures:

```
Running .\build\agora-unittests.exe 
Program exited with code -1073741515

Test d:\a\agora\agora\.\tests\unit\BlockStorageChecksum.d failed!
```

But it was still green in the CI.

PR #627, opened 2020-03-05
Use SCPStatementType type directly in _pledges_t

AndrejMitrovic:
@Geod24 I've noticed that `log.info` doesn't format enums nicely. For example, it will not print out the name of the enum, but instead just its value. So stringification will only be better if `format` is used from Phobos.

E.g.:

```D
enum E { foo, bar }
E e = E.bar;
log.info("{}", e);  // will just print out '1' instead of 'bar'
```


Geod24:
Yeah the formatter does not support enum because D1 didn't make it possible. It should be a simple fix now that upstream dropped D1.

Geod24 (approved):

PR #629, opened 2020-03-05
Don't accept outdated SCP messages

AndrejMitrovic:
I can't really think of a way to test this, because the only thing that changes here is the logging. The state machine (Ledger) wouldn't change if for example an outdated SCP Externalize message was accepted (it would be a double-spend).

This eliminates the various "tx not found in tx set" error messages we frequently saw in the CI logs.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -67,9 +67,6 @@ public extern (C++) class Nominator : SCPDriver
     /// Similar to how we use FakeClockBanManager!
     private Set!ulong timers;
 
-    /// The set of externalized slot indices
-    private Set!uint64_t externalized_slots;
-
     /// The quorum set
     private SCPQuorumSetPtr[StellarHash] quorum_set;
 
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
This also might be a symptom of something wrong with SCP. I don't quite understand why the protocol is calling the validation routine validateValue on a slot which was already externalized. It would seem to me that SCP should log an error very early on and skip validation.

AndrejMitrovic:
Rebased.

I've added another commit removing `externalized_slots`, because we can just use the Ledger directly.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -371,9 +376,8 @@ extern(D):
     {
         scope (failure) assert(0);
 
-        if (slot_idx in this.externalized_slots)
+        if (slot_idx <= this.ledger.getBlockHeight())
```
MichaelKim20, line 379: `slot_idx` is the same as `block_height`. Doesn't slot_idx start with 0?

MichaelKim20 (commented):
I have a question.

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: `slot_idx` is set here: https://github.com/bpfkorea/agora/blob/7f52a574e07366b4c468f96f7deff207249ac01f/source/agora/consensus/protocol/Nominator.d#L167

MichaelKim20: Yeah, I got it, thank you.

MichaelKim20 (approved):
LGTM

linked0 (approved):
LGTM

PR #632, opened 2020-03-05
Deserializer: Always validate strings, introduce fromBinary, handle pointers

Geod24:
Yet another deserialization PR. Better reviewed commit-by-commit.

### DIFF
--- a/source/agora/common/Serializer.d
+++ b/source/agora/common/Serializer.d
@@ -428,7 +428,7 @@ public size_t deserializeLength (
         T = Type of data to deserialize
         data = Binary serialized representation of `T` to be deserialized
         dg   = Delegate to read binary data for deserialization
-        compact = Whether integers are serialized in variable-length form
+        opts = Deserialization options (see the type's documentation for a list)
 
     Returns:
         The deserialized data type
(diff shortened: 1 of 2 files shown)


source/agora/common/Serializer.d
```diff
@@ -380,6 +380,18 @@ public void serializePart (T) (scope const auto ref T record, scope SerializeDg
             entry.serializePart(dg);
     }
 
+    // Pointers are handled as arrays, only their size must be 0 or 1
```
AndrejMitrovic, line 383: Recursive / Cyclic pointers are not supported, right? Hopefully we'll never use them though.

Geod24: Good point, I guess that's why they aren't usually supported... And I suppose inner pointers will mess things too :/

MichaelKim20: I understand what situations are used through unittest.

source/agora/common/Serializer.d
```diff
@@ -152,11 +152,17 @@ unittest
             serializePart(this.s, dg);
         }
 
-        void deserialize (scope DeserializeDg dg) @safe
+        static QT fromBinary (QT) (
```
MichaelKim20, line 155: We must implement this as static.
And we need to use 'deserializeFull' inside.

Geod24: https://bpfkorea.github.io/agora/agora/common/Serializer.html

MichaelKim20 (approved):
LGTM

AndrejMitrovic (approved):

PR #634, opened 2020-03-06
System test: Resolve agora module dependencies automatically

Geod24:
This is a frequent source of pain when extending the system integration test.

Since we depend on the latest DMD / LDC, this will just work.


TrustHenry (approved):
Oh this will make us comfortable.

MichaelKim20 (approved):

PR #637, opened 2020-03-06
Add pretty-printing of SCPEnvelope

### LINKED ISSUE #635: Add prettify stringification for SCPEnvelope
This is really needed right now to be able to properly debug / diagnose anything that's still wrong with SCP reaching consensus.

Requirements:
- Legibility
- Hex key types into proper types, e.g. 0xAA... => GDQ6...ZY7L
- Provide a quorum getter so we can actually have the quorums printed out and not just their hashes (if they are found)

I've already started working on this.

AndrejMitrovic:
I'll wait to see the report from codecov to see if every line has been covered.

Still some things left to do:
- In some places the hash is still displayed as hexadecimal in full, we could use the `HashFmt` for this.
- Deserialization of `ConsensusData` in this case is handled gracefully. But in the future when we start using hashes, we can just provide another lookup function and print out `<tx not found>` when we're missing a tx that's being nominated.

Fixes #635

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -84,6 +84,7 @@
                 "source/agora/common/Config.d",
                 "source/agora/network/*",
                 "source/agora/node/*",
+                "source/agora/utils/PrettyPrinter.d",
                 "source/scpp/*",
                 "source/scpd/*",
                 "source/agora/consensus/protocol/*",
--- a/dub.json
+++ b/dub.json
@@ -85,6 +85,7 @@
                 "source/agora/network/*",
                 "source/agora/node/*",
                 "source/agora/utils/PrettyPrinter.d",
+                "source/agora/utils/SCPPrettyPrinter.d",
                 "source/scpp/*",
                 "source/scpd/*",
                 "source/agora/consensus/protocol/*",
--- a/source/agora/utils/SCPPrettyPrinter.d
+++ b/source/agora/utils/SCPPrettyPrinter.d
@@ -1,0 +1,685 @@
+/*******************************************************************************
+
+    Contains pretty-printing routines specialized for SCP types.
+
+    As the integration test-suite imports into PrettyPrinter,
+    we want to avoid adding a dependency to SCP.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.utils.SCPPrettyPrinter;
+
+import agora.common.Types;
+import agora.common.Amount;
+import agora.common.Config;
+import agora.common.Serializer;
+import agora.common.Types;
+import agora.common.crypto.Key;
+import agora.consensus.data.Block;
+import agora.consensus.data.ConsensusData;
+import agora.consensus.data.Enrollment;
+import agora.consensus.data.Transaction;
+import agora.utils.PrettyPrinter;
+
+import scpd.Cpp;
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types : StellarHash = Hash;
+
+import std.algorithm;
+import std.format;
+import std.range;
+
+/*******************************************************************************
+
+    Returns a formatting prettifier for SCPEnvelope.
+
+    Params:
+        env = a pointer to SCPEnvelope
+        get_qset = getter for quorum sets. If null it won't be used.
+
+*******************************************************************************/
+
+public SCPEnvelopeFmt scpPrettify (in SCPEnvelope* env,
+    in GetQSetDg get_qset = null) nothrow @trusted @nogc
+{
+    return SCPEnvelopeFmt(env, get_qset);
+}
+
+/// Formatting struct for `SCPBallot`, deserializes Value types as ConsensusData
+public struct SCPBallotFmt
+{
+    private const(SCPBallot) ballot;
+
+    public void toString (scope void delegate(scope const(char)[]) @safe sink) @trusted nothrow
+    {
+        try
+        {
+            formattedWrite(sink,
+                "{ counter: %s, ",
+                this.ballot.counter);
+
+            try
+            {
+                formattedWrite(sink,
+                    "value: %s }",
+                    prettify(  // cast: deserializer should take const(ubyte)[]
+                    (cast(ubyte[])this.ballot.value[]).deserializeFull!ConsensusData));
+            }
+            catch (Exception ex)
+            {
+                formattedWrite(sink, "value: <un-deserializable> }");
+            }
+        }
+        catch (Exception ex)
+        {
+            assert(0, ex.msg);
+        }
+    }
+}
+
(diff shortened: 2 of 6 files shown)


AndrejMitrovic:
Oh great, I love linker errors.

Geod24:
> Oh great, I love linker errors.

Move all your function-scope imports to global scope.

AndrejMitrovic:
> Move all your function-scope imports to global scope.

Thanks that works. But I thought this was just an issue with RDMD, not dub..

AndrejMitrovic:
Ah no it did not work. :>

Ok I will look into this..

Geod24:
> Thanks that works. But I thought this was just an issue with RDMD, not dub..

It's actually an issue with DMD, and we use some of the dependency tracking magic in some places (also we use rdmd in `tests/runner.d`)

AndrejMitrovic:
The linking issue is for the `client` configuration. But the client shouldn't be importing into PrettyPrinter, so I don't quite get why this is causing linker issues.

AndrejMitrovic:
Ok I figured it out. Had to add `"source/agora/utils/PrettyPrinter.d",` to the excluded source files.

Geod24:
Looks like something in the integration test suite depends from the pretty printer tho

AndrejMitrovic:
I've split the SCP pretty-printing into another module, and added several commits introducing `ConsensusData` and `Enrollment` pretty-printing support.

dub.json
```diff
@@ -54,6 +54,8 @@
                 "source/agora/common/Config.d",
                 "source/agora/network/*",
                 "source/agora/node/*",
+                "source/agora/utils/PrettyPrinter.d",
+                "source/agora/utils/SCPPrettyPrinter.d",
```
Geod24, line 88: I don't think you need to exclude `PrettyPrinter` anymore >?

source/agora/utils/SCPPrettyPrinter.d
```diff
@@ -0,0 +1,606 @@
...
+import agora.consensus.data.Block;
+import agora.consensus.data.ConsensusData;
+import agora.consensus.data.Enrollment;
+import agora.consensus.data.Transaction;
+import agora.utils.PrettyPrinter;
+
+import scpd.Cpp;
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types : StellarHash = Hash;
+alias StellarValue = Value;
```
Geod24: Why does one use renamed import and the other uses (public) alias ? Personally I prefer the former.

source/agora/utils/SCPPrettyPrinter.d
```diff
@@ -0,0 +1,606 @@
...
+    {
+        try
+        {
+            SCPQuorumSetPtr qset;
+            if (this.getQSet !is null)
+                qset = this.getQSet(this.hash);
+
+            if (qset !is null)
+            {
+                auto qconf = toQuorumConfig(*qset.ptr);
```
Geod24, line 109: Isn't that a `null` dereference ? `toQuorumConfig` takes a `ref`

source/agora/utils/SCPPrettyPrinter.d
```diff
@@ -0,0 +1,606 @@
...
+Outputs (8): GCOQ...LRIJ(62,500,000), GCOQ...LRIJ(62,500,000), GCOQ...LRIJ(62,500,000),
+GCOQ...LRIJ(62,500,000), GCOQ...LRIJ(62,500,000), GCOQ...LRIJ(62,500,000),
+GCOQ...LRIJ(62,500,000), GCOQ...LRIJ(62,500,000)], enrolls: [{ utxo: 0x0000...e26f, seed: 0x4a5e...a33b, cycles: 1008, sig: 0x0000...be78 }, { utxo: 0x0000...e26f, seed: 0x4a5e...a33b, cycles: 1008, sig: 0x0000...be78 }] } }, prep: <null>, prepPrime: <null>, nc: 100, nH: 200 } }, sig: 0x0af7...b5ab }`;
+
+    // with quorum mapping
+    static immutable PrepareRes2 = `{ statement: { node: GBUV...KOEK, slotIndex: 0, pledge: SCP_PREPARE { qset: { hash: 0xfee13ea2fd956028d31b93cbf0a47ea97c6c9868b7e911b797667113ed12bf0d, quorum: QuorumConfig(2, [GBFDLGQQDDE2CAYVELVPXUXR572ZT5EOTMGJQBPTIHSLPEOEZYQQCEWN, GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5], []) }, ballot: { counter: 42, value: { tx_set: [Type : Payment, Inputs (1): 0x0000...0000[0]:0x0000...0000
```
Geod24: Should we abbreviate `SCP_PREPARE.qset.hash` ? Also `QuorumConfig` entries ?

Geod24 (commented):
Very nice, a few comments

source/agora/utils/SCPPrettyPrinter.d
AndrejMitrovic: One was using `alias this` (`shared_ptr`) and the other explicit dereference. I'll simplify it.

source/agora/utils/SCPPrettyPrinter.d
AndrejMitrovic: Good ideas.

AndrejMitrovic:
Update:

- Renamed `SCP_EXTERNALIZE` to `Externalize`, etc.
- Added formatting for: StellarHash (32 byte hash)
- Added pretty-printing for QuorumConfig

AndrejMitrovic:
I'll force push again to reorder the commits.

AndrejMitrovic:
Fixed linking issues. I had to re-add PrettyPrinter to exclude files again.

AndrejMitrovic:
Hmm. I think I should consider changing all the `*Fmt` structs in `SCPPrettyPrinter` to store a pointer rather than a value as a field, because we want to avoid passing C++ types by value.

@Geod24 what do you think?

Geod24:
That was my original approach, however it didn't work. The details of why escape me ATM. Might be something to do with `@safe`. But yeah, we definitely want this.

AndrejMitrovic:
> Might be something to do with @safe

Have you tried @escapehatch? :P

I'll give it a go.

AndrejMitrovic:
Oops. I pinged someone. :x

AndrejMitrovic:
> source/agora/common/Config.d(26,8): Error: module `node` is in file 'dyaml/node.d' which cannot be read

hmm..

AndrejMitrovic:
Update: removed copying.

There's still the weird compilation failure on circle-ci.

AndrejMitrovic:
I know how to fix it, I'll move `QuorumConfig` out of the `Config` module.

AndrejMitrovic:
Updated.

Geod24 (approved):
Amazing, thanks!

PR #639, opened 2020-03-09
dub: try to fix build.d not being run sometimes

AndrejMitrovic:
build.d contains all the logic for tracking file timestamps, but dub sometimes just doesn't run it.

I found this forum thread about it: https://forum.dlang.org/post/mzipqtnimvexeddjtcju@forum.dlang.org

This should hopefully fix that issue.


linked0 (approved):
LGTM

Geod24 (approved):
Right

PR #642, opened 2020-03-10
Remove unused memutils submodule.

AndrejMitrovic:


AndrejMitrovic:
I have noticed it after grepping for rbtree, and I also found it has a `memutils.rbtree` module https://github.com/etcimon/memutils/blob/master/source/memutils/rbtree.d, but I'm not sure if it supports `const`.

Geod24 (approved):

Issue #587, opened 2020-02-25
Network manager needs to be reworked


AndrejMitrovic:
When we implemented it the goal was to support Full Nodes (first milestone). Full Nodes connect to a minimum number of clients via the `min_listeners` config option.

But Validators should not have this config option. The minimum number of clients it connects to is the number of nodes in its quorum configuration. They do not care about connecting to random Full Nodes, it is Full Nodes which care about connecting to Validators & other Full Nodes.

Also from #644:

There are various sets inside, and it's all a bit too complicated to reason about:

- todo_addresses
- known_addresses
- connecting_addresses

In addition to that, some of the initialization in `start()` should be moved to the constructor so the class is always in a ready-to-go state. Otherwise it's possible some incoming requests are routed to the NetworkManager while `start()` hasn't yet been called, which leaves it in an inconsistent state.

AndrejMitrovic:
Another issue: we are currently not handling connection failures. See also https://github.com/bpfkorea/agora/pull/643

AndrejMitrovic:
I will fold https://github.com/bpfkorea/agora/issues/644 into this one.

Geod24:
Also, currently the quorum set is derived before the network discovery, which needs to be reworked to allow for adjusting the qset in real time: https://github.com/bpfkorea/agora/blob/5ffc4574b99ebafc7817801ad2770ea2253cd126/source/agora/consensus/protocol/Nominator.d#L111

AndrejMitrovic:
I will add to the notes that I'm currently working on this too. But as always, the other bugs keep slashing my tires. :>

AndrejMitrovic:
This has a race condition:
 https://github.com/bpfkorea/agora/blob/9d56a9e080b9978111e9e1e5a3735881e7fb00e2/source/agora/network/NetworkManager.d#L380

There may be multiple fibers trying to connect to the very same address.

This will all be fixed once this is redesigned to make it side-effect free. Only the main fiber should handle adding / removing peers from its maps, and this fiber should spawn other worker fibers to connect to each client. And it will know not to spawn multiple fibers for the same client.

For example currently this error is sometimes visible:

> Full error: core.exception.AssertError@source/agora/test/Base.d(510): Trying to access node at address 'Validator #3 (GCAW...HKP5)' without first creating it

The error is a bit misleading actually. The validator **was** registered, but then it got unregistered. And because of the current design in NetworkManager, it seems we call `getClient` very often.

I'll assign this to myself tomorrow.

PR #643, opened 2020-03-10
Validator: Connect to all quorum set validators

AndrejMitrovic:
~Requires https://github.com/bpfkorea/agora/pull/648~

### DIFF
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -21,6 +21,7 @@ import agora.common.Config;
 import agora.common.Hash;
 import agora.common.Metadata;
 import agora.common.crypto.Key;
+import agora.common.Set;
 import agora.common.Task;
 import agora.common.Types;
 import agora.common.TransactionPool;
@@ -110,8 +111,29 @@ public class Node : API
 
         this.config = config;
         this.taskman = this.getTaskManager();
+
+        // build the list of required quorum peers to connect to if we're a Validator
+        Set!PublicKey required_peer_keys;
+        if (this.config.node.is_validator)
+        {
+            void getNodes (in QuorumConfig conf, ref Set!PublicKey nodes)
+            {
+                foreach (node; conf.nodes)
+                {
+                    if (node != config.node.key_pair.address)  // filter ourselves
+                        nodes.put(node);
+                }
+
+                foreach (sub_conf; conf.quorums)
+                    getNodes(sub_conf, nodes);
+            }
+
+            getNodes(config.quorum, required_peer_keys);
+        }
+
         this.network = this.getNetworkManager(config.node, config.banman,
-            config.network, config.dns_seeds, this.metadata, this.taskman);
+            config.network, required_peer_keys, config.quorum, config.dns_seeds,
+            this.metadata, this.taskman);
         this.storage = this.getBlockStorage(config.node.data_dir);
         this.pool = this.getPool(config.node.data_dir);
         scope (failure) this.pool.shutdown();
@@ -281,6 +303,8 @@ public class Node : API
             node_config = the node config
             banman_conf = the ban manager config
             peers = the peers to connect to
+            required_peer_keys = required peers with the given keys to connect to
+            quorum_conf = the quorum config
             dns_seeds = the DNS seeds to retrieve peers from
             metadata = metadata containing known peers and other meta info
             taskman = task manager
@@ -292,10 +316,11 @@ public class Node : API
 
     protected NetworkManager getNetworkManager (in NodeConfig node_config,
         in BanManager.Config banman_conf, in string[] peers,
+        Set!PublicKey required_peer_keys, in QuorumConfig quorum_conf,
         in string[] dns_seeds, Metadata metadata, TaskManager taskman)
     {
-        return new NetworkManager(node_config, banman_conf, peers, dns_seeds,
-            metadata, taskman);
+        return new NetworkManager(node_config, banman_conf, peers,
+            required_peer_keys, quorum_conf, dns_seeds, metadata, taskman);
     }
 
     /***************************************************************************
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -413,11 +413,13 @@ public class TestNetworkManager : NetworkManager
 
     /// Constructor
     public this (NodeConfig config, BanManager.Config ban_conf,
-        in string[] peers, in string[] dns_seeds, Metadata metadata,
-        TaskManager taskman, Registry* reg)
+        in string[] peers, Set!PublicKey required_peer_keys,
+        in QuorumConfig quorum_conf, in string[] dns_seeds,
+        Metadata metadata, TaskManager taskman, Registry* reg)
     {
         this.registry = reg;
-        super(config, ban_conf, peers, dns_seeds, metadata, taskman);
+        super(config, ban_conf, peers, required_peer_keys, quorum_conf,
+            dns_seeds, metadata, taskman);
     }
 
     /// We don't use IPs in tests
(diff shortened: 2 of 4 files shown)


Geod24:
Isn't it going against some of our goals to have validator special casing in network manager ?

AndrejMitrovic:
Right, so we could move the list of required peers to the Node class (and ValidatorNode class once that's implemented). I'll rework it.

AndrejMitrovic:
Actually, I've realized that the issue itself isn't so clear-cut.

The node doesn't **have to** be connected to all of its quorum peers in order to reach consensus. For example, let's say we have this node quorum layout:


```
A => (B, C, D)
D => (B, C)
```

A has (B, C, D) in its quorum set, while D has (B, C).

If A loses or does not establish its connection to B, it does not mean it won't receive its messages. As long as A is connected to D, then D will propagate B's messages back to A.

-----

So changing `minPeersConnected` is incorrect here, although it's safer (but works against liveness). Changing `peerLimitReached` however is correct because we should always strive towards connecting to all of our quorum peers, even if their messages can be forwarded by other nodes in the network.

AndrejMitrovic:
#648 was merged, this is now unblocked.

I've decided to keep the safe behavior by default. The validating node will attempt to connect to all of its quorum peers before `discover()` finishes. Later we can fine-tune this and make it have more liveness vs safety.

AndrejMitrovic:
Updated with better documentation. @bpfkorea/core ready for review.

source/agora/test/Base.d
```diff
@@ -744,6 +753,31 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
...
     }
 
+    // for disovery testing: full quorum, but only 1 node in the 'network' section
+    Config makeMinimalNetQuorum (size_t idx, NodeConfig self, NodeConfig[] node_confs)
+    {
+        auto prev_node = idx == 0 ? node_confs[$ - 1] : node_confs[idx - 1];
+
+        auto quorum_keys =
+            node_confs
+                .filter!(conf => conf.is_validator)
```
TrustHenry, line 783: If `conf.is_validator` is `false`, an exception may occur.

```
Module tests failed: agora.test.NetworkDiscovery
object.Exception@submodules/localrest/source/geod24/LocalRest.d-mixin-716(770): "Request timed-out"
----------------
/usr/local/opt/dmd/include/dlang/dmd/core/internal/array/equality.d:33 @safe void geod24.LocalRest.RemoteAPI!(agora.test.Base.TestAPI).RemoteAPI.start() [0x10e7876da]
/usr/local/opt/dmd/include/dlang/dmd/core/internal/array/equality.d:33 @safe void agora.test.Base.TestAPIManager.start().__lambda1!(agora.test.Base.NodePair).__lambda1(agora.test.Base.NodePair) [0x10e7965e0]
/usr/local/opt/dmd/include/dlang/dmd/core/internal/array/equality.d:33 @safe std.typecons.Flag!("each").Flag std.algorithm.iteration.each!(agora.test.Base.TestAPIManager.start().__lambda1).each!(agora.test.Base.NodePair[]).each(ref agora.test.Base.NodePair[]) [0x10e796657]
/usr/local/opt/dmd/include/dlang/dmd/core/internal/array/equality.d:33 void agora.test.Base.TestAPIManager.start() [0x10e783b7f]
/usr/local/opt/dmd/include/dlang/dmd/core/internal/array/equality.d:33 void agora.test.NetworkDiscovery.__unittest_L69_C1() [0x10e7ab21a]
source/agora/node/main.d:40 void agora.test.NetworkDiscovery.__modtest() [0x10e4edf4a]
/usr/local/opt/dmd/include/dlang/dmd/core/internal/array/equality.d:33 int agora.test.Base.customModuleUnitTester().__foreachbody3(ref agora.test.Base.customModuleUnitTester().ModTest) [0x10e7837df]
/usr/local/opt/dmd/include/dlang/dmd/core/internal/array/equality.d:33 void std.parallelism.ParallelForeach!(agora.test.Base.customModuleUnitTester().ModTest[]).ParallelForeach.opApply(scope int delegate(ref agora.test.Base.customModuleUnitTester().ModTest)).doIt() [0x10e566203]
```

source/agora/node/Node.d
```diff
@@ -110,8 +111,29 @@ public class Node : API
 
         this.config = config;
         this.taskman = this.getTaskManager();
+
+        // build the list of required quorum peers to connect to if we're a Validator
+        Set!PublicKey required_peer_keys;
+        if (this.config.node.is_validator)
```
TrustHenry, line 117: If `this.config.node.is_validator` is `false`, a unit test results in an error.

```
Module tests failed: agora.test.NetworkDiscovery
core.exception.AssertError@source/agora/test/NetworkDiscovery.d(89): Node 0 has 2 peers: ["GAYDGXP436SZ7LV6B5EXRZX27TB2D2GQRFV3SRN775RU6OQPMJT347F5", "GDD62CVWBKJE4NXYMELIV6GHZ5AKJ2TNY6MFASHUQY24RFAKJ4EZHCUF"]
```

source/agora/test/Base.d
AndrejMitrovic: if *none* of the node configs' `is_validator` is true, it will fail because `map` will try to invoke `popFront` on the empty range. But this `makeMinimalNetQuorum` should only be called for this specific test-case where all the nodes are validators.

source/agora/node/Node.d
AndrejMitrovic: Where are you setting it to false? Can you show me this test-case?

source/agora/test/Base.d
```diff
@@ -744,6 +749,31 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         return conf;
     }
 
+    // for disovery testing: full quorum, but only 1 node in the 'network' section
```
```suggestion
    // for discovery testing: full quorum, but only 1 node in the 'network' section
```

source/agora/test/Base.d
Geod24: Will it ? I would expect array on an empty range to return `null`.
Perhaps add an `assert` that `.length > 0` for clarity ?

Geod24 (changes requested):
This seems to only work for the initial connection, doesn't it ?
What happens if a required peer drops off, then a non-required peer connects, taking the "slot", and when the required peer comes back online, it wouldn't be accepted, would it ?

AndrejMitrovic:
> What happens if a required peer drops off

True. The current design of NetworkManager does not take into account dropped connections. It's not really obvious how to deal with this while using the RestInterfaceClient. Maybe after #203 is implemented it will be much clearer how to handle connections?

Geod24:
> Maybe after #203 is implemented it will be much clearer how to handle connections?

Then this is your lucky day.

source/agora/test/Base.d
AndrejMitrovic: Ah it definitely does not throw. I wonder where I got this error from.

I'll add an assert.

AndrejMitrovic:
I think reworking the NetworkManager class could be a good issue to tackle in one of the next sprints. https://github.com/bpfkorea/agora/issues/587 and https://github.com/bpfkorea/agora/issues/644

AndrejMitrovic:
> Then this is your lucky day.

Maybe today..?

AndrejMitrovic:
> This seems to only work for the initial connection, doesn't it ?
> What happens if a required peer drops off, then a non-required peer connects, taking the "slot", and when the required peer comes back online, it wouldn't be accepted, would it ?

I'll try to redesign the networking class to make it more flexible. Can we make this part of future work in #587?

source/agora/test/Base.d
```diff
@@ -767,11 +771,13 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         return conf;
     }
 
-    // for disovery testing: full quorum, but only 1 node in the 'network' section
+    // for dissovery testing: full quorum, but only 1 node in the 'network' section
```
Geod24: LOL. Also wrong commit.

Geod24 (approved):

Issue #644, opened 2020-03-10
Refactor NetworkManager's complex use of sets


AndrejMitrovic:
There are various sets inside, and it's all a bit too complicated to reason about:

- todo_addresses
- known_addresses
- connecting_addresses

In addition to that, some of the initialization in `start()` should be moved to the constructor so the class is always in a ready-to-go state. Otherwise it's possible some incoming requests are routed to the NetworkManager while `start()` hasn't yet been called, which leaves it in an inconsistent state.

I'll work on this later.

AndrejMitrovic:
Folding into #587

PR #645, opened 2020-03-10
Add a registerListener() API

AndrejMitrovic:
This was hard to pull off because of the complex design of the NetworkManager class (see https://github.com/bpfkorea/agora/issues/644).

Split off from https://github.com/bpfkorea/agora/pull/621

Part of #606 (but does not fully resolve it yet).

### DIFF
--- a/source/agora/network/NetworkClient.d
+++ b/source/agora/network/NetworkClient.d
@@ -95,19 +95,14 @@ class NetworkClient
 
     /***************************************************************************
 
-        Handshake with the node.
-
-        Currently only the node's public key is retrieved,
-        later version checks should be added here.
-
-        Note that it is currently blocking until handshake is considered complete.
+        Get the node's public key.
 
         Throws:
-            Exception if the handshake did not complete.
+            `Exception` if the request failed.
 
     ***************************************************************************/
 
-    public void handshake ()
+    public void getPublicKey ()
     {
         this.key = this.attemptRequest(this.api.getPublicKey(), this.exception);
     }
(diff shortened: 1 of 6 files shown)


AndrejMitrovic:
By the way, I know this is not an ideal API. Anyone can pass any address string here. 

It's possible to extract the address from the connection which invoked an API with vibe.d. 

But for LocalRest it would require additional work. And I'm not sure what the best design would be.

So maybe we could keep `registerListener` like this for now with the address parameter, and later remove it.

source/agora/network/NetworkClient.d
```diff
@@ -95,21 +95,34 @@ class NetworkClient
...
+            address = the address to register
 
         Throws:
-            Exception if the handshake did not complete.
+            `Exception` if the request failed.
 
     ***************************************************************************/
 
-    public void handshake ()
+    public void registerListener (string address)
```
TrustHenry:
```suggestion
    public void registerListener (Address address)
```
Can you tell why used string here?

AndrejMitrovic: Oh I made a mistake while rebasing commits. I'll fix it.

AndrejMitrovic: Nice catch!

AndrejMitrovic:
Fixed string => address.

linked0 (approved):
Nice code!
And, I'd like to know what is the worst thing when an outsider participates in the gossip protocol. I mean what happend in the worst case.

PR #646, opened 2020-03-11
CircleCI frequently fails 

MichaelKim20:
CircleCI frequently fails


Geod24:
For reference, we don't really want to re-add the SCP dependency in the integration test suite, at least not this one. See https://github.com/bpfkorea/agora/commit/87ae5d05143d758363de9ed1d3aa75b084b48a49#diff-3d9bd18baf7bf8355b12b9baad7bc07c where it was removed.

PR #648, opened 2020-03-11
Fix network discovery & start Node asynchronously

### LINKED ISSUE #276: TestAPIManager.start() is not asynchronous

### LINKED ISSUE #647: Network discovery is broken (due to lack of a proper test)
In the process of refactoring the code in the NetworkManager (which happened last year), some code was accidentally removed.

Right now in the discovery stage the network information from a node is discarded:

https://github.com/bpfkorea/agora/blob/4812c713c00a5b36a539e958552b7689397c33c5/source/agora/network/NetworkManager.d#L412-L414

There is a missing call to `addAddresses`.

AndrejMitrovic:
This fixes the network discovery issue #647. While creating a test-case I've realized that #276 is actually blocking the implementation of a test-case. So this PR includes two fixes.

Fixes #276
Fixes #647

### DIFF
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -134,16 +134,19 @@ public class Node : API
     /// The first task method, loading from disk, node discovery, etc
     public void start ()
     {
-        log.info("Doing network discovery..");
-        this.network.discover();
-
-        bool isNominating ()
+        this.taskman.runTask(
         {
-            return this.config.node.is_validator &&
-                this.nominator.isNominating();
-        }
+            log.info("Doing network discovery..");
+            this.network.discover();
 
-        this.network.startPeriodicCatchup(this.ledger, &isNominating);
+            bool isNominating ()
+            {
+                return this.config.node.is_validator &&
+                    this.nominator.isNominating();
+            }
+
+            this.network.startPeriodicCatchup(this.ledger, &isNominating);
+        });
     }
 
     /***************************************************************************
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -410,6 +410,7 @@ public class NetworkManager
             try
             {
                 auto net_info = node.getNetworkInfo();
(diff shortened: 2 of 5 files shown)


AndrejMitrovic:
The asynchronous start has managed to find a bug: https://github.com/bpfkorea/agora/pull/649/files

AndrejMitrovic:
@bpfkorea/core this is blocking other PRs. Could you please review? :)

linked0 (approved):
I reviewed all the codes yesterday.
It looks good to me.

TrustHenry (approved):
LGTM 
I'll be do merge when the CI test is over

source/agora/node/Node.d
```diff
@@ -134,16 +134,19 @@ public class Node : API
     /// The first task method, loading from disk, node discovery, etc
     public void start ()
     {
-        log.info("Doing network discovery..");
-        this.network.discover();
-
-        bool isNominating ()
+        this.taskman.runTask(
```
Geod24, line 137: https://github.com/bpfkorea/agora/blob/18c86736458a9cc3926f05be4fbf637a98fd8c8c/source/agora/node/Node.d#L544

AndrejMitrovic: I saw that, but I think I left it in out of misunderstanding. I'll remove it in a fixup PR.

TrustHenry: Oh, I didn't see this in detail.
It was wrapped twice.

PR #656, opened 2020-03-12
Remove leftover task in Node.d

AndrejMitrovic:
The start() method now already runs in its own task.

### DIFF
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -541,7 +541,7 @@ public Node runNode (Config config)
 
     auto node = new Node(config);
     router.registerRestInterface(node);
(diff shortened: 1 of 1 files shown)


source/agora/node/Node.d
```diff
@@ -541,7 +541,7 @@ public Node runNode (Config config)
 
     auto node = new Node(config);
     router.registerRestInterface(node);
-    runTask({ node.start(); });
+    node.start();  // asynchronous
```
Geod24, line 544: Or maybe, just rename the method to `startAsync` ?

Geod24 (approved):

PR #659, opened 2020-03-13
Add missing definition of operator() for hashing support

AndrejMitrovic:
std::hash is used for things like storing structs in std::set / std::map.

Required for the quorum intersection checker.

### DIFF
--- a/source/scpp/src/crypto/ByteSliceHasher.cpp
+++ b/source/scpp/src/crypto/ByteSliceHasher.cpp
@@ -1,0 +1,29 @@
+// Copyright 2018 Stellar Development Foundation and contributors. Licensed
+// under the Apache License, Version 2.0. See the COPYING file at the root
+// of this distribution or at http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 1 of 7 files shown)


Geod24:
I guess you want to import the quorum intersection check in C++ ?

source/scpp/src/crypto/ByteSliceHasher.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+namespace stellar
+{
+namespace shortHash
+{
+static unsigned char sKey[crypto_shorthash_KEYBYTES];
+void
+initialize()
+{
+    crypto_shorthash_keygen(sKey);
```
Geod24, line 16: This has to be called so you might want to stick a `shared static this` in the bindings

AndrejMitrovic:
> I guess you want to import the quorum intersection check in C++ ?

Yes. And it's a lot of code (2 Kloc or so) so I opted to bind to it rather than port it to D.

Updated to add the initialization call.

Geod24 (approved):
LGTM

PR #663, opened 2020-03-16
Add the BitSet implementation from stellar-core

### LINKED ISSUE #240: Design and implement quorum balancing

AndrejMitrovic:
This is used by the quorum intersection checker code. We won't actually bind to it.

### DIFF
--- a/source/scpp/README.md
+++ b/source/scpp/README.md
@@ -9,6 +9,12 @@ Files in `extra` are extra C++ files added to the build (e.g. to instantiate tem
 Commit used for extraction: [v11.2.0](https://github.com/stellar/stellar-core/releases/tag/v11.2.0)
 Timestamp of commit: Wed Jun 26 11:16:38 2019 -0700
 
+# Porting notes
+
+Some files have been modified after porting:
+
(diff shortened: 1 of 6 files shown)


Geod24 (approved):
Is there any diff from upstream ?

AndrejMitrovic:
The only difference is that I renamed the file to have a `.cpp` extension and treat it as a C++ file, because otherwise on the mac the compiler treats it as a C file and complains about unsupported C++ compiler flags:

> error: invalid argument '-std=c++14' not allowed with 'C'

I thought about changing `build.d`, but then there would probably need to be a special case for the Windows build as well. It seemed easier to just change the file extension.

Then again, I wonder if there could be subtle bugs if C code is compiled with a C compiler vs a C++ one. Hmm..

Geod24:
Let's assume no bugs.

PR #662, opened 2020-03-16
Add rudimentary bindings to std::unordered_map

AndrejMitrovic:
@Geod24 there seems to be a linking issue when I change `cpp_unordered_map_create` to return a proper type rather than just a `void*`. See all the `// todo` comments.

```
.dub/build/unittest-unittest-linux.posix-x86_64-dmd_2090-B745E6F9C1062E73969E8A9A1EE52EEC/agora-unittests.o: In function `_D4scpd3Cpp__T13unordered_mapTiTiZQu6createFNaNbNiNeZSQCaQBy__TQBxTiTiZQCf':
/mnt/c/dev/agora/source/scpd/Cpp.d:158: undefined reference to `std::unordered_map<int, int>* cpp_unordered_map_create<int, int>()'
```

PR #664, opened 2020-03-16
Add the quorum intersection checker code from stellar-core

AndrejMitrovic:
It also includes tests.

Requires:
- ~#662~
- ~#663~

### DIFF
--- a/source/scpd/quorum/QuorumIntersectionChecker.d
+++ b/source/scpd/quorum/QuorumIntersectionChecker.d
@@ -1,0 +1,44 @@
+/*******************************************************************************
+
+    Bindings for quorum/QuorumIntersectionChecker.h
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module scpd.quorum.QuorumIntersectionChecker;
+
+nothrow @trusted @nogc:
+
+import scpd.Cpp;
+import scpd.quorum.QuorumTracker;
+import scpd.types.Stellar_types;
+
+extern (C++, `stellar`):
+
+/// Ditto
+public abstract class QuorumIntersectionChecker
+{
+  public:
+    /// Create & initialize a QuorumIntersectionChecker with the given map
+    static shared_ptr!QuorumIntersectionChecker create (
+        ref const(QuorumTracker.QuorumMap) map);
+
+    ~this () {}
+
+    /// Returns: true if the network enjoys quorum intersection
+    abstract bool networkEnjoysQuorumIntersection ();
+
+    /// Returns: the total number of quorums found
+    abstract size_t getMaxQuorumsFound ();
+
+    /// Returns: A pair of possible quorum splits found, or empty pair if none
+    abstract pair!(vector!NodeID, vector!NodeID) getPotentialSplit ();
+}
+
+static assert(__traits(classInstanceSize, QuorumIntersectionChecker) == 8);
--- a/source/scpd/quorum/QuorumIntersectionTests.d
+++ b/source/scpd/quorum/QuorumIntersectionTests.d
@@ -1,0 +1,742 @@
+/*******************************************************************************
+
+    Contains quorum intersection tests adapted from stellar-core.
+
+    Note: some tests are only enabled when -version=PerformanceTests is enabled,
+    as these tests take a long time to finish.
+
+    Copyright:
+        Copyright 2016-2019 Stellar Development Foundation and contributors.
+
+    License:
+        Licensed under the Apache License, Version 2.0. See the COPYING file at
+        http://www.apache.org/licenses/LICENSE-2.0
+
+*******************************************************************************/
+
+module scpd.quorum.QuorumIntersectionTests;
+
+version (unittest):
+
+import scpd.quorum.QuorumIntersectionChecker;
+import scpd.quorum.QuorumTracker;
+import scpd.Cpp;
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types : uint256, NodeID;
+import scpd.types.Utils;
+import scpd.types.XDRBase;
+import agora.common.Config;
+import agora.common.crypto.Key;
+import agora.common.Types;
+import agora.utils.Log;
+
+mixin AddLogger!();
+
+// quorum intersection basic 4-node
+unittest
+{
+    auto qm = QuorumTracker.QuorumMap.create();
+
+    PublicKey pkA = KeyPair.random.address;
+    PublicKey pkB = KeyPair.random.address;
+    PublicKey pkC = KeyPair.random.address;
+    PublicKey pkD = KeyPair.random.address;
+
+    qm[pkA.toStellarKey] = makeSharedQuorumSet(2, [pkB, pkC, pkD], null);
+    qm[pkB.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkC, pkD], null);
+    qm[pkC.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkB, pkD], null);
+    qm[pkD.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkB, pkC], null);
+
+    auto qic = QuorumIntersectionChecker.create(qm);
+    assert(qic.networkEnjoysQuorumIntersection());
+}
+
+// quorum intersection 6-node with subquorums
+unittest
+{
+    auto qm = QuorumTracker.QuorumMap.create();
+
+    PublicKey pkA = KeyPair.random.address;
+    PublicKey pkB = KeyPair.random.address;
+    PublicKey pkC = KeyPair.random.address;
+    PublicKey pkD = KeyPair.random.address;
+    PublicKey pkE = KeyPair.random.address;
+    PublicKey pkF = KeyPair.random.address;
+
+    QuorumConfig qsABC = QuorumConfig(2, [pkA, pkB, pkC], null);
+    QuorumConfig qsABD = QuorumConfig(2, [pkA, pkB, pkD], null);
+    QuorumConfig qsABE = QuorumConfig(2, [pkA, pkB, pkE], null);
+    QuorumConfig qsABF = QuorumConfig(2, [pkA, pkB, pkF], null);
+
+    QuorumConfig qsACD = QuorumConfig(2, [pkA, pkC, pkD], null);
+    QuorumConfig qsACE = QuorumConfig(2, [pkA, pkC, pkE], null);
+    QuorumConfig qsACF = QuorumConfig(2, [pkA, pkC, pkF], null);
+
+    QuorumConfig qsADE = QuorumConfig(2, [pkA, pkD, pkE], null);
+    QuorumConfig qsADF = QuorumConfig(2, [pkA, pkD, pkF], null);
+
+    QuorumConfig qsBDC = QuorumConfig(2, [pkB, pkD, pkC], null);
+    QuorumConfig qsBDE = QuorumConfig(2, [pkB, pkD, pkE], null);
+    QuorumConfig qsCDE = QuorumConfig(2, [pkC, pkD, pkE], null);
+
+    qm[pkA.toStellarKey] = makeSharedQuorumSet(2, null, [qsBDC, qsBDE, qsCDE]);
+    qm[pkB.toStellarKey] = makeSharedQuorumSet(2, null, [qsACD, qsACE, qsACF]);
+    qm[pkC.toStellarKey] = makeSharedQuorumSet(2, null, [qsABD, qsABE, qsABF]);
+    qm[pkD.toStellarKey] = makeSharedQuorumSet(2, null, [qsABC, qsABE, qsABF]);
+    qm[pkE.toStellarKey] = makeSharedQuorumSet(2, null, [qsABC, qsABD, qsABF]);
+    qm[pkF.toStellarKey] = makeSharedQuorumSet(2, null, [qsABC, qsABD, qsABE]);
+
+    auto qic = QuorumIntersectionChecker.create(qm);
+    assert(qic.networkEnjoysQuorumIntersection());
+}
+
+// quorum non intersection basic 6-node
+unittest
+{
+    auto qm = QuorumTracker.QuorumMap.create();
+
+    PublicKey pkA = KeyPair.random.address;
+    PublicKey pkB = KeyPair.random.address;
+    PublicKey pkC = KeyPair.random.address;
+    PublicKey pkD = KeyPair.random.address;
+    PublicKey pkE = KeyPair.random.address;
+    PublicKey pkF = KeyPair.random.address;
+
+    qm[pkA.toStellarKey] = makeSharedQuorumSet(2, [pkB, pkC, pkD, pkE, pkF], null);
+    qm[pkB.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkC, pkD, pkE, pkF], null);
+    qm[pkC.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkB, pkD, pkE, pkF], null);
+    qm[pkD.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkB, pkC, pkE, pkF], null);
+    qm[pkE.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkB, pkC, pkD, pkF], null);
+    qm[pkF.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkB, pkC, pkD, pkE], null);
+
+    auto qic = QuorumIntersectionChecker.create(qm);
+    assert(!qic.networkEnjoysQuorumIntersection());
+}
+
+// quorum non intersection 6-node with subquorums
+unittest
+{
+    auto qm = QuorumTracker.QuorumMap.create();
+
+    PublicKey pkA = KeyPair.random.address;
+    PublicKey pkB = KeyPair.random.address;
+    PublicKey pkC = KeyPair.random.address;
+    PublicKey pkD = KeyPair.random.address;
+    PublicKey pkE = KeyPair.random.address;
+    PublicKey pkF = KeyPair.random.address;
+
+    QuorumConfig qsABC = QuorumConfig(2, [pkA, pkB, pkC], null);
+    QuorumConfig qsABD = QuorumConfig(2, [pkA, pkB, pkD], null);
+    QuorumConfig qsABE = QuorumConfig(2, [pkA, pkB, pkE], null);
+    QuorumConfig qsABF = QuorumConfig(2, [pkA, pkB, pkF], null);
+
+    QuorumConfig qsACD = QuorumConfig(2, [pkA, pkC, pkD], null);
+    QuorumConfig qsACE = QuorumConfig(2, [pkA, pkC, pkE], null);
+    QuorumConfig qsACF = QuorumConfig(2, [pkA, pkC, pkF], null);
+
+    QuorumConfig qsADE = QuorumConfig(2, [pkA, pkD, pkE], null);
+    QuorumConfig qsADF = QuorumConfig(2, [pkA, pkD, pkF], null);
+
+    QuorumConfig qsBDC = QuorumConfig(2, [pkB, pkD, pkC], null);
+    QuorumConfig qsBDE = QuorumConfig(2, [pkB, pkD, pkE], null);
+    QuorumConfig qsBDF = QuorumConfig(2, [pkB, pkD, pkF], null);
+    QuorumConfig qsCDE = QuorumConfig(2, [pkC, pkD, pkE], null);
+    QuorumConfig qsCDF = QuorumConfig(2, [pkC, pkD, pkF], null);
+
+    qm[pkA.toStellarKey] = makeSharedQuorumSet(2, null, [qsABC, qsABD, qsABE]);
+    qm[pkB.toStellarKey] = makeSharedQuorumSet(2, null, [qsBDC, qsABD, qsABF]);
+    qm[pkC.toStellarKey] = makeSharedQuorumSet(2, null, [qsACD, qsACD, qsACF]);
+
+    qm[pkD.toStellarKey] = makeSharedQuorumSet(2, null, [qsCDE, qsADE, qsBDE]);
+    qm[pkE.toStellarKey] = makeSharedQuorumSet(2, null, [qsCDE, qsADE, qsBDE]);
+    qm[pkF.toStellarKey] = makeSharedQuorumSet(2, null, [qsABF, qsADF, qsBDF]);
+
+    auto qic = QuorumIntersectionChecker.create(qm);
+    assert(!qic.networkEnjoysQuorumIntersection());
+}
+
+// quorum plausible non intersection
+unittest
+{
+    auto qm = QuorumTracker.QuorumMap.create();
+
+    PublicKey pkSDF1 = KeyPair.random.address;
+    PublicKey pkSDF2 = KeyPair.random.address;
+    PublicKey pkSDF3 = KeyPair.random.address;
+
+    PublicKey pkLOBSTR1 = KeyPair.random.address;
+    PublicKey pkLOBSTR2 = KeyPair.random.address;
+
+    PublicKey pkSatoshi1 = KeyPair.random.address;
+    PublicKey pkSatoshi2 = KeyPair.random.address;
+    PublicKey pkSatoshi3 = KeyPair.random.address;
+
+    PublicKey pkCOINQVEST1 = KeyPair.random.address;
+    PublicKey pkCOINQVEST2 = KeyPair.random.address;
+
+    // Some innersets used in quorums below.
+
+    QuorumConfig qs1of2LOBSTR = QuorumConfig(1, [pkLOBSTR1, pkLOBSTR2], null);
+    QuorumConfig qs1of2COINQVEST = QuorumConfig(1, [pkCOINQVEST1, pkCOINQVEST2], null);
+
+    QuorumConfig qs2of3SDF = QuorumConfig(1, [pkSDF1, pkSDF2, pkSDF3], null);
+
+    QuorumConfig qs2of3SatoshiPay = QuorumConfig(2, [pkSatoshi1, pkSatoshi2, pkSatoshi3],
+                                  null);
+
+    // All 3 SDF nodes get this:
+    auto qsSDF = makeSharedQuorumSet(3, [pkSDF1, pkSDF2, pkSDF3],
+                                 [qs1of2LOBSTR, qs2of3SatoshiPay]);
+    qm[pkSDF1.toStellarKey] = qsSDF;
+    qm[pkSDF2.toStellarKey] = qsSDF;
+    qm[pkSDF3.toStellarKey] = qsSDF;
+
+    // All SatoshiPay nodes get this:
+    auto qsSatoshiPay =
+        makeSharedQuorumSet(4, [pkSatoshi1, pkSatoshi2, pkSatoshi3],
+                        [qs2of3SDF, qs1of2LOBSTR, qs1of2COINQVEST]);
+    qm[pkSatoshi1.toStellarKey] = qsSatoshiPay;
+    qm[pkSatoshi2.toStellarKey] = qsSatoshiPay;
+    qm[pkSatoshi3.toStellarKey] = qsSatoshiPay;
+
+    // All LOBSTR nodes get this:
+    auto qsLOBSTR = makeSharedQuorumSet(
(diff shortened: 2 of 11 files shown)


AndrejMitrovic:
The part that assigns unassigned nodes needs to be removed. Not all nodes need to be in other nodes' quorums.

Geod24:
I rebased it on v0.x.x, let's see if it passes on Windows

Geod24:
Looks nice (and green). Will give it a review today.

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        quorum_tracker_rebuild(cast(void*)this, &call, cast(void*)&lookup);
+    }
+
+    // returns the current known quorum
+    ref const(QuorumMap) getQuorum() const;
+}
+
+private void quorum_tracker_rebuild (void* quorum_tracker, void* cb, void* dg);
```
AndrejMitrovic: Use `ref` and the proper type instead of `void*`

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,71 @@
...
+        static extern(C++) SCPQuorumSetPtr call (void* func, const ref NodeID node_id)
+        {
+            auto dg = cast(SCPQuorumSetPtr delegate(const ref NodeID)*)func;
+            return (*dg)(node_id);
+        }
+
+        quorum_tracker_rebuild(cast(void*)this, &call, cast(void*)&lookup);
+    }
+
+    // returns the current known quorum
```
AndrejMitrovic:
```suggestion
    // returns the quorum map this quorum tracker was built with
```

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,71 @@
...
+    // and add its dependencies as defined by `qset`
+    // returns true if expansion succeeded
+    //     `id` was unknown
+    //     `id` was known and didn't have a quorumset
+    // returns false on failure
+    // if expand fails, the caller should instead use `rebuild`
+    bool expand(const ref NodeID id, SCPQuorumSetPtr qSet);
+
+    // replaces 'rebuild' taking an std::function so we can use it with D delegates
+    extern (D) public final void rebuild (SCPQuorumSetPtr delegate(const ref NodeID) lookup)
```
AndrejMitrovic:
```suggestion
    extern (D) public final void rebuild (SCPQuorumSetPtr delegate(const ref NodeID) lookup) @trusted
```

And other attributes.

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,71 @@
...
+module scpd.quorum.QuorumTracker;
+
+import scpd.Cpp;
+import scpd.scp.SCP;
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types;
+
+extern (C++, `stellar`):
+
+// helper class to help track the overall quorum over time
```
AndrejMitrovic: We could improve this documentation. It was written by Stellar devs, but it's hard to parse.

Edit: do it upstream

source/scpp/src/quorum/QuorumTracker.cpp
```diff
@@ -0,0 +1,92 @@
...
+#include "quorum/QuorumTracker.h"
+#include "scp/LocalNode.h"
+
+namespace stellar
+{
+
+// workaround to be able to pass D delegates to C++
+void quorum_tracker_rebuild (void* qt, void* cb, void* dg)
+{
+    typedef SCPQuorumSetPtr (*Callback)(void*, NodeID const&);
```
AndrejMitrovic: This function body has too much whitespace.

AndrejMitrovic: Also as mentioned before, use `ref` and proper types, not pointers.

AndrejMitrovic (commented):
Completed self-review.

source/scpd/quorum/QuorumIntersectionChecker.d
```diff
@@ -0,0 +1,36 @@
...
+
+extern (C++, `stellar`):
+
+extern(C++, class) public abstract class QuorumIntersectionChecker
+{
+  public:
+    extern(D) static shared_ptr!QuorumIntersectionChecker create (
+        QuorumTracker.QuorumMap map) nothrow pure @trusted @nogc
+    {
+        return create(*cast(void**)&map);
```
AndrejMitrovic: Should be changed to take a reference.

source/scpd/quorum/QuorumIntersectionChecker.d
```diff
@@ -0,0 +1,36 @@
+/*******************************************************************************
+
+    Bindings for quorum/QuorumIntersectionChecker.h
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
```
AndrejMitrovic: 2020

source/scpd/quorum/QuorumIntersectionChecker.d
```diff
@@ -0,0 +1,36 @@
...
+import scpd.Cpp;
+import scpd.types.Stellar_types;
+import scpd.quorum.QuorumTracker;
+
+extern (C++, `stellar`):
+
+extern(C++, class) public abstract class QuorumIntersectionChecker
+{
+  public:
+    extern(D) static shared_ptr!QuorumIntersectionChecker create (
```
AndrejMitrovic: Since this is a method we introduced, we should document it.

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,71 @@
...
+    // replaces 'rebuild' taking an std::function so we can use it with D delegates
+    extern (D) public final void rebuild (SCPQuorumSetPtr delegate(const ref NodeID) lookup)
+    {
+        static extern(C++) SCPQuorumSetPtr call (void* func, const ref NodeID node_id)
+        {
+            auto dg = cast(SCPQuorumSetPtr delegate(const ref NodeID)*)func;
+            return (*dg)(node_id);
+        }
+
+        quorum_tracker_rebuild(cast(void*)this, &call, cast(void*)&lookup);
```
AndrejMitrovic: Replace pointers with references.

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,71 @@
...
+
+    // attempts to expand quorum at node `id`
+    // expansion here means adding `id` to the known quorum
+    // and add its dependencies as defined by `qset`
+    // returns true if expansion succeeded
+    //     `id` was unknown
+    //     `id` was known and didn't have a quorumset
+    // returns false on failure
+    // if expand fails, the caller should instead use `rebuild`
+    bool expand(const ref NodeID id, SCPQuorumSetPtr qSet);
```
AndrejMitrovic: Use our styling `expand (`

source/scpp/README.md
```diff
@@ -15,6 +15,8 @@ Some files have been modified after porting:
 
 - Some `private` declarations were changed to `public`. This allows us to test size & layout checks for our D glue layer, for example: https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/src/crypto/ByteSlice.h#L21 used in https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/extra/DSizeChecks.cpp and https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/extra/DLayoutChecks.cpp.
 
+- The `src/quorum` folder adapts the Quorum intersection tool which is used in Stellar's Herder implementation (not SCP). The `update.d` script does not track these files yet, so they have to be kept manually in sync when updating the bindings.
```
AndrejMitrovic:
```suggestion
- The `src/quorum` folder contains the Quorum intersection tool which is used in Stellar's Herder implementation (not SCP). The `update.d` script does not track these files yet, so they have to be kept manually in sync when updating the bindings.
```

AndrejMitrovic:
I suggest adding a in-between commit which just checks in all the intersection checker code as-is, but without bindings. And change the commit author to be the Stellar devs. Otherwise it appears like the code was forged or stolen, and we don't want to leave that kind of impression on the community.

source/scpd/quorum/QuorumIntersectionTests.d
```diff
@@ -0,0 +1,744 @@
...
+import scpd.types.XDRBase;
+
+import agora.common.Config;
+import agora.common.crypto.Key;
+
+const xvector!SCPQuorumSet nullvec;
+const shared_ptr!SCPQuorumSet nullquorum;
+
+// we replace the use of std::pair with a fixed-length array
+size_t first (size_t[2] pair) { return pair[0]; }
```
AndrejMitrovic: This is one example where the code was changed from upstream. So we should definitely have a separate commit including all the C++ code, and then the conversion step commits after that.

source/scpd/quorum/QuorumIntersectionTests.d
```diff
@@ -0,0 +1,744 @@
...
+    return makeSharedSCPQuorumSet(scp);
+}
+
+uint roundUpPct (size_t n, size_t pct)
+{
+    return cast(uint)(size_t(1) +
+        (((n * pct) - size_t(1)) / size_t(100)));
+}
+
+NodeID toStellarKey (PublicKey address)
```
AndrejMitrovic: This can be replaced soon.

source/scpd/quorum/QuorumIntersectionTests.d
```diff
@@ -0,0 +1,744 @@
...
+
+    qm[pkA.toStellarKey] = makeSharedQuorumSet(2, [pkB, pkC, pkD], null);
+    qm[pkB.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkC, pkD], null);
+    qm[pkC.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkB, pkD], null);
+    qm[pkD.toStellarKey] = makeSharedQuorumSet(2, [pkA, pkB, pkC], null);
+
+    auto qic = QuorumIntersectionChecker.create(qm);
+    assert(qic.networkEnjoysQuorumIntersection());
+
+    //CircularAppender().printConsole();
```
AndrejMitrovic: Dead code. And the new test-suite logging fixes already handle this case.

source/scpd/quorum/QuorumIntersectionTests.d
```diff
@@ -0,0 +1,744 @@
...
+}
+
+//TEST_CASE("quorum intersection 6-node with subquorums",
+//          "[herder][quorumintersection]")
```
AndrejMitrovic: This stuff in the brackets can all be removed for all of these tests.

AndrejMitrovic:
I've split this up into multiple commits. The stellar devs get the proper attribution, and the diff is much easier to understand now.

AndrejMitrovic:
I will fix up the rest of the comments.

AndrejMitrovic:
Resolved all comments.

Geod24:
> Otherwise it appears like the code was forged or stolen, and we don't want to leave that kind of impression on the community.

Just for reference: We have [this README](https://github.com/bpfkorea/agora/blob/v0.x.x/source/scpp/README.md) and each file has a `// Copyright 2019 Stellar Development Foundation and contributors. [...]` comment at the top for it. It would be pretty hard to argue we are trying to hide the origin of the code.

But I am completely fine with your approach, just wanted to give context into why I am not concerned with this myself.

Geod24:
There seem to be a POSIX specific mangling bug:
```
ld: warning: directory not found for option '-L@@HOMEBREW_CELLAR@@/libsodium/1.0.18/lib'
  "stellar::quorum_tracker_rebuild(stellar::QuorumTracker*, std::__1::shared_ptr<stellar::SCPQuorumSet> (*)(void*, stellar::PublicKey const&), void*)", referenced from:
      __D4scpd6quorum13QuorumTrackerQp7rebuildMFNeDRNbNiNeKxSQCa5types13Stellar_types9PublicKeyZSQDk3Cpp__T10shared_ptrTSQEiQCi11Stellar_SCP12SCPQuorumSetZQBwZv in agora-unittests.o
  "stellar::QuorumIntersectionChecker::create(std::unordered_map<stellar::PublicKey, std::__1::shared_ptr<stellar::SCPQuorumSet>, std::__1::hash<stellar::PublicKey>, std::__1::equal_to<stellar::PublicKey>, std::__1::allocator<std::__1::pair<stellar::PublicKey const, std::__1::shared_ptr<stellar::SCPQuorumSet> > > > const&)", referenced from:
      __D4scpd6quorum23QuorumIntersectionTests17__unittest_L20_C1FZv in agora-unittests.o
      __D4scpd6quorum23QuorumIntersectionTests17__unittest_L39_C1FZv in agora-unittests.o
      __D4scpd6quorum23QuorumIntersectionTests17__unittest_L78_C1FZv in agora-unittests.o
      __D4scpd6quorum23QuorumIntersectionTests18__unittest_L101_C1FZv in agora-unittests.o
      __D4scpd6quorum23QuorumIntersectionTests18__unittest_L143_C1FZv in agora-unittests.o
      __D4scpd6quorum23QuorumIntersectionTests18__unittest_L314_C1FZv in agora-unittests.o
      __D4scpd6quorum23QuorumIntersectionTests18__unittest_L327_C1FZv in agora-unittests.o
      ...
```

AndrejMitrovic:
Yeah that is legitimate. I also don't want to increase my stats with fake line counts.

Geod24:
Since it's OSX only, perhaps the `std::shared_ptr` mangling ?

Geod24:
> Yeah that is legitimate. I also don't want to increase my stats with fake line counts.

Ah didn't consider that. Very good point.

AndrejMitrovic:
It would be cool to have those build artifacts so we can `nm` the objects.. since I'm not on a mac right now.

TrustHenry:
Currently, the Windows CI test is marked as green because there is a bug.

Mangling bugs keep popping up in Windows CI Test.  :(

AndrejMitrovic:
I know. I will implement a workaround for the mangling issue.

AndrejMitrovic:
I thought the issue was with the return value which was `shared_ptr`, but actually it might be the `unordered_map` which in this case stores a `shared_ptr` as its value.

I see `__1` in the mangling, maybe that's what's throwing it off?

AndrejMitrovic:
ok I found the linker issue, will submit a separate PR to fix it.

AndrejMitrovic:
Completed fixes after self-review.

AndrejMitrovic:
With regards to `Make Quorum checker buildable`, note that doing this upstream isn't easy because the bindings we're targeting here is for v11.2.0, whereas the one in stellar-core master branch contains even more specialized code which we can't easily bind to. For example, they use interrupts to stop the quorum checker (because sometimes it takes too long to finish).

source/scpp/README.md
```diff
@@ -15,6 +15,8 @@ Some files have been modified after porting:
 
 - Some `private` declarations were changed to `public`. This allows us to test size & layout checks for our D glue layer, for example: https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/src/crypto/ByteSlice.h#L21 used in https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/extra/DSizeChecks.cpp and https://github.com/bpfkorea/agora/blob/76e4ca7d77b123ef867d3ce99c1c1b26afab761d/source/scpp/extra/DLayoutChecks.cpp.
 
+- The `src/quorum` folder contains the Quorum intersection code which is used in stellar-core. It was adapted to not depend on any stellar-specific type definitions. The `update.d` script does not track these files yet, so they have to be kept manually in sync when updating the bindings.
```
Geod24, line 18: Was it taken from the same commit ?

AndrejMitrovic: from 11.2.0

source/scpd/quorum/QuorumIntersectionChecker.d
```diff
@@ -0,0 +1,42 @@
...
+nothrow @trusted @nogc:
+
+import scpd.Cpp;
+import scpd.types.Stellar_types;
+import scpd.quorum.QuorumTracker;
+
+extern (C++, `stellar`):
+
+/// Ditto
+extern(C++, class) public abstract class QuorumIntersectionChecker
```
Geod24:
```suggestion
public abstract class QuorumIntersectionChecker
```

source/scpd/quorum/QuorumIntersectionChecker.d
```diff
@@ -0,0 +1,42 @@
...
+
+*******************************************************************************/
+
+module scpd.quorum.QuorumIntersectionChecker;
+
+nothrow @trusted @nogc:
+
+import scpd.Cpp;
+import scpd.types.Stellar_types;
+import scpd.quorum.QuorumTracker;
```
Geod24:
```suggestion
import scpd.quorum.QuorumTracker;
import scpd.types.Stellar_types;
```

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,67 @@
...
+    /// returns false on failure
+    /// if expand fails, the caller should instead use `rebuild`
+    bool expand(const ref NodeID id, SCPQuorumSetPtr qSet);
+
+    /// Lookup function to find a mapping from Node ID => Quorum set
+    private alias CallbackType = SCPQuorumSetPtr delegate(const ref NodeID) @trusted @nogc nothrow;
+
+    /// returns the known quorum map of the entire network
+    ref const(QuorumMap) getQuorum() const;
+}
```
Geod24, line 72:
```suggestion
}

static assert(QuorumTracker.sizeof == 48);
```

Provided this is made a `struct` instead of a `class`. But this won't pass because of #753, which this uncovered.

source/scpd/quorum/QuorumTracker.d
Geod24: This was not addressed

```suggestion
    bool expand (const ref NodeID id, SCPQuorumSetPtr qSet);
```

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,67 @@
+/*******************************************************************************
+
+    Bindings for quorum/QuorumTracker.h
+
+    Note: D code does not need to use anything in this class except the
```
Geod24:
```suggestion
    D code does not need to use anything in this class except the
```

Having `Note: ` will create a section which will look bad in the doc IIRC. If you really intended this, perhaps
```
Note:
D code does not [...]
```
is more appropriate

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,67 @@
...
+extern (C++, `stellar`):
+
+/// helper class to help track the overall quorum over time
+/// a node tracked is definitely in the transitive quorum.
+/// If its associated quorum set is empty (nullptr), it just means
+/// that another node has that node in its quorum set
+/// but could not explore the quorum further (as we're missing the quorum set)
+/// Nodes can be added one by one (calling `expand`, most efficient)
+/// or the quorum can be rebuilt from scratch by using a lookup function
+extern(C++, class) public class QuorumTracker
```
Geod24:
```suggestion
extern(C++, class) public struct QuorumTracker
```

I really need to write that C++ documentation to explain how to bind types. Usually if a type has no single virtual method, it's better to go with `struct`, unless you know it's only used as a reference type. Even then, if it's passed by `ref` instead of `ptr`, game over.

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,67 @@
...
+private:
+    SCP* mSCP;
+    QuorumMap mQuorum;
+
+public:
+    /// Initialize a new QuorumTracker
+    this (SCP* scp);
+
+    /// returns true if id is in transitive quorum for sure
+    bool isNodeDefinitelyInQuorum(const ref NodeID id);
```
Geod24: This is not correct, and neither are the following ones. In D those functions are virtual, while in C++ they are not. You can make either the affected functions `final`, or, if the class doesn't have a single `virtual` function, make the `class` itself final.

From my experience, things declared as `class` with no `virtual` method => `struct`. And even more so if the `class` is stored as a member of another type.

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,67 @@
...
+    /// and add its dependencies as defined by `qset`
+    /// returns true if expansion succeeded
+    ///     `id` was unknown
+    ///     `id` was known and didn't have a quorumset
+    /// returns false on failure
+    /// if expand fails, the caller should instead use `rebuild`
+    bool expand(const ref NodeID id, SCPQuorumSetPtr qSet);
+
+    /// Lookup function to find a mapping from Node ID => Quorum set
+    private alias CallbackType = SCPQuorumSetPtr delegate(const ref NodeID) @trusted @nogc nothrow;
```
Geod24: This is not used

source/scpd/quorum/QuorumIntersectionChecker.d
```diff
@@ -0,0 +1,42 @@
...
+
+    /// Returns: true if the network enjoys quorum intersection
+    abstract bool networkEnjoysQuorumIntersection ();
+
+    /// Returns: the total number of quorums found
+    abstract size_t getMaxQuorumsFound ();
+
+    /// Returns: A pair of possible quorum splits found, or empty pair if none
+    abstract pair!(vector!NodeID, vector!NodeID) getPotentialSplit ();
+}
```
Geod24, line 42:
```suggestion
}

static assert(__traits(classInstanceSize, QuorumIntersectionChecker) == 8);
```

source/scpd/quorum/QuorumTracker.d
```diff
@@ -0,0 +1,67 @@
...
+{
+public:
+    /// The Node ID => Quorum set map type
+    alias QuorumMap = unordered_map!(NodeID, SCPQuorumSetPtr);
+
+private:
+    SCP* mSCP;
+    QuorumMap mQuorum;
+
+public:
```
Geod24, line 50: I don't mind the usage of `:`-attributes, but can we merge the two `public:` together ?

source/scpd/quorum/QuorumIntersectionTests.d
```diff
@@ -1,315 +1,298 @@
...
+    License:
+        Licensed under the Apache License, Version 2.0. See the COPYING file at
+        http://www.apache.org/licenses/LICENSE-2.0
+
+*******************************************************************************/
+
+module scpd.quorum.QuorumIntersectionTests;
+
+import scpd.quorum.QuorumIntersectionChecker;
+import scpd.quorum.QuorumTracker;
```
Geod24, line 22: Wrong order

Geod24 (changes requested):
Woah... Sorry it took me so long to review this. So I left a few comments. The most important ones are w.r.t. `class` vs `struct`, and the need for size checks which uncovered #753 .

A few more things:
- I didn't port the SCP tests when I imported the SCP code. The reasoning was that little change would be required to SCP (if any), and none of that will be of substance. Any substantial change should be submitted upstream, where its sanity could be checked. So basically piggybacking on upstream having proper tests. I was a bit surprised when I saw the porting of the tests, and would have not originally asked for it, unless you had major plans to change the code being imported (which I guess you do). But the reasoning I just mentioned was also intended to save us time, and if the work is done, I see no reason not to pull it.
- The last two commit should be merged. I appreciate the effort to make the diff readable (and it was much more pleasant that it could have been), but since `dub` compiles *everything*, there will be a commit that doesn't build. Also, I believe `git` (/github) has greatly improved their rename tracking, so the diff might not be so bad (did you test it?). But atomic commit trumps readability in that case.
- As mentioned above, `dub` compiles everything. You might want to add the tests to the `excludedSourceFiles` so is only compiled on unittests.

Given that we use the fake wrapper you mentioned in #753 this is good to go after the fixes. Just add the commented out size check and a comment explaining why it is not currently correct.

AndrejMitrovic:
> I didn't port the SCP tests when I imported the SCP code. 

I wanted to verify the binding was correct. It wasn't a big effort tho, and it did catch bugs in my binding.

That being said I guess we don't really need it anymore. We will have separate tests for the quorums we generate, so maybe we can remove this code?

It might even be complicated to have something which is Stellar-licensed in our codebase, but I'm not lawyer.

> but since dub compiles everything, there will be a commit that doesn't build.

Yeah this was purely to make it easy to review. I'll squash them.

 > As mentioned above, dub compiles everything. You might want to add the tests to the excludedSourceFiles so is only compiled on unittests.

Oh right. Will fix.

source/scpd/quorum/QuorumTracker.d
AndrejMitrovic: > In D those functions are virtual

Oh damn. How come it worked?

> From my experience, things declared as class with no virtual method => struct.

What about classes which inherit other classes? This one specifically is defined as `class QuorumTracker : public NonMovableOrCopyable`, so it looks like it was never supposed to be used as a POD type.

Geod24: > class QuorumTracker : public NonMovableOrCopyable

Sounds like the idiomatic way to do this in D would be:
```D
mixin template NonMovableOrCopyable (T /* = typeof(this) */) // Not sure if the default arg works
{
    @disable this(this);
    @disable ref T opAssign () (auto ref T other);
}

public struct QuorumTracker
{
    mixin NonMovableOrCopyable!(QuorumTracker);
}
```

That might be a bit of an overreaching statement, but my experience is that good, principled C++ code ends up being idiomatic D code, and trying to use the same construct in D as one does in C++ might lead one astray.

> Oh damn. How come it worked?

Depends on the usage. The code called from C++ doesn't have an issue, obviously. The code called from D may or may not crash, depending on whether or not the compiler devirtualize the call. And since we dropped DMD, there is a good chance it does :)

Geod24: Note, the above snippet doesn't fully capture the C++ intent though, since in D, you *cannot* declare a `struct` are non movable. Should probably include a copy constructor too but it currently triggers a cascade of bugs.

AndrejMitrovic: > And since we dropped DMD, there is a good chance it does :)

LOL.

Yeah also I dislike virtual by default in D. Where is Manu when you need him.

source/scpd/quorum/QuorumIntersectionChecker.d
AndrejMitrovic: Sure, but why?

source/scpd/quorum/QuorumTracker.d
AndrejMitrovic: Actually I dislike `:` attributes personally. I'll fix up the mistake though.

Geod24:
> It might even be complicated to have something which is Stellar-licensed in our codebase, but I'm not lawyer.

We already do (SCP), and it's Apache licensed (and we have LICENSE-APACHE.txt included and haven't removed any authorship / license information), so it's all good.

source/scpd/quorum/QuorumIntersectionChecker.d
Geod24: Why it is `8` or why we should have this check ?

Geod24:
```
agora.obj : error LNK2001: unresolved external symbol "public: virtual __cdecl stellar::QuorumIntersectionChecker::~QuorumIntersectionChecker(void)" (??1QuorumIntersectionChecker@stellar@@UEAA@XZ)
```

AndrejMitrovic:
Also the block storage checksum thing failed on Windows again.

Like I said in https://github.com/bpfkorea/agora/pull/731, why not just version it out on Windows and fix it later?

AndrejMitrovic:
Can I just get rid of the destructor? I don't know why it's defined on the C++ side:

```C++
virtual ~QuorumIntersectionChecker(){};
```

source/scpd/quorum/QuorumIntersectionChecker.d
Geod24: @AndrejMitrovic answered directly: the later.

Because it's a rather simple check that can save you a lot of headache. For example, out of the two types introduced in this PR, one had the wrong size (`QuorumTracker`) and adding the check uncovered that and #753 . Also makes sure that in the future, we don't break this by accident (e.g. change in alignment, or adding a field in a nested type).

In my experience, those little checks have an immediate benefit.

Geod24:
It's defined so that there is an entry in the vtable for the destructor. Otherwise you end up with things like [this](https://issues.dlang.org/show_bug.cgi?id=18616).
Except in C++ it is `inline` for the base class. Just provide an empty destructor in the D binding and you should be 👍

AndrejMitrovic:
I don't understand where the submodule update is coming from..

AndrejMitrovic:
Ok I managed to get rid of it.

Geod24 (approved):
Thanks

PR #666, opened 2020-03-17
Change the mechanism of creating preimages

### LINKED ISSUE #651: Derive random seed from node's private key
We need to restore the random seed from abnormal situations of a node. One of the critical issues is losing the source of random seed, which means a validator could lose its deposit in case it can not reveal preimages. So we need to implement the way to restore the source of random seed and derive the source from a private key.

Definitions of Done
- [x] Derive the source of random seed from private key
- [x] Change the process of getting a preimage from a node in accordance of new mechanism of creating pre-images
- [x] Restore the source of random seed when a node starts or when the issue occurs
- [x] Re-create random seed if all the pre-images are revealed

The above definition could be splitted into smaller issues.

Any suggestions are welcome.

linked0:
Previously, we create only 1008 number of preimages from a random value for the enrollment process. But there is the case when the random value is lost and we can't restore pre-images in  abnormal situations like database crashes. So we need the way to restore the pre-images. we have decided to change the mechanism of creating pre-images into generating a great number of pre-images like a hundred of thousand from the private key of a node.

This PR includes the code re-creating pre-images when all the pre-images have been used in case of the block height is above 100800, or`ValidatorCycle` x 100. If the re-creation of pre-images is needed, the enrollment manager creates new bulk of pre-images with a salt hash. A salt hash is derived from a bulk index which is an order of a series of a bulk of pre-images like 0, 1, 2, ... N.

Fixes #651

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -54,11 +54,11 @@ public class EnrollmentManager
     /// Key used for enrollment which is actually an UTXO hash
     private Hash enroll_key;
 
-    /// Random seed
-    private Scalar random_seed_src;
+    /// Pre-images for each point with the interval of validator cycle
+    private Hash[ulong] preimage_rounds;
 
-    /// Preimages of hashes of random value
-    public Hash[] preimages;
+    /// Pre-images for current validator cycle
+    private Hash[ulong] cycle_preimages;
 
     /// Random key for enrollment
     private Pair signature_noise;
@@ -76,6 +76,9 @@ public class EnrollmentManager
     /// Validator set managing validators' information such as Enrollment object
     /// enrolled height, and preimages.
     private ValidatorSet validator_set;
+
+    /// The count for generating pre-images
+    private immutable uint AllCountPreimages = ValidatorSet.ValidatorCycle * 100;
 
     /***************************************************************************
 
@@ -111,13 +114,6 @@ public class EnrollmentManager
             signature_noise = deserializeFull!Pair(row.peek!(ubyte[])(0));
             break;
         }
-
-        // load preimages
-        results = this.db.execute("SELECT val FROM node_enroll_data " ~
-            "WHERE key = ?", "preimages");
-
-        if (!results.empty)
-            this.preimages = results.oneValue!(ubyte[]).deserializeFull!(Hash[]);
 
         // load enroll_key
         results = this.db.execute("SELECT val FROM node_enroll_data " ~
@@ -242,6 +238,7 @@ public class EnrollmentManager
         Params:
             frozen_utxo_hash = the hash of a frozen UTXO used to identify a validator
                         and to generate a siging key
+            height = the starting index for generating pre-images
             enroll = will contain the Enrollment if created
 
         Returns:
@@ -249,7 +246,8 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public bool createEnrollment (Hash frozen_utxo_hash, out Enrollment enroll) @trusted nothrow
+    public bool createEnrollment (Hash frozen_utxo_hash, ulong height,
+        out Enrollment enroll) @trusted nothrow
     {
         static ubyte[] buffer;
         buffer.length = 0;
@@ -262,17 +260,9 @@ public class EnrollmentManager
         // N, cycle length
         this.data.cycle_length = ValidatorSet.ValidatorCycle;
 
-        // generate random seed value
-        this.random_seed_src = Scalar.random();
-
         // X, final seed data and preimages of hashes
-        this.preimages.length = 0;
-        assumeSafeAppend(this.preimages);
-        this.preimages ~= hashFull(this.random_seed_src);
-        foreach (i; 0 .. this.data.cycle_length-1)
-            this.preimages ~= hashFull(this.preimages[i]);
-        reverse(this.preimages);
-        this.data.random_seed = this.preimages[0];
+        this.data.random_seed = this.generatePreimages(height);
+
 
         // R, signature noise
         this.signature_noise = Pair.random();
@@ -306,42 +296,6 @@ public class EnrollmentManager
                 {
                     this.db.execute("INSERT INTO node_enroll_data (key, val) VALUES (?, ?)",
                         "signature_noise", buffer);
-                }
-            }();
-        }
-        catch (Exception ex)
-        {
-            this.logMessage("Database operation error", enroll, ex);
-            return false;
-        }
-
-        // serialize preimages
-        buffer.length = 0;
-        assumeSafeAppend(buffer);
-        try
-        {
-            serializePart(this.preimages, dg);
-        }
-        catch (Exception ex)
-        {
-            this.logMessage("Serialization error of preimages", enroll, ex);
-            return false;
-        }
-
-        try
-        {
-            () @trusted {
-                auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM node_enroll_data " ~
-                    "WHERE key = ?)", "preimages");
-                if (results.oneValue!(bool))
-                {
-                    this.db.execute("UPDATE node_enroll_data SET val = ? WHERE key = ?",
-                        buffer, "preimages");
-                }
-                else
-                {
-                    this.db.execute("INSERT INTO node_enroll_data (key, val) VALUES (?, ?)",
-                        "preimages", buffer);
                 }
             }();
         }
@@ -525,9 +479,12 @@ public class EnrollmentManager
             (height - start_height) > ValidatorSet.ValidatorCycle - 1)
             return false;
 
+        if (height !in this.cycle_preimages)
+            this.generatePreimages(height);
+
         preimage.enroll_key = this.data.utxo_key;
         preimage.height = height;
-        preimage.hash = this.preimages[height - start_height];
+        preimage.hash = this.cycle_preimages[height];
         return true;
     }
 
@@ -684,6 +641,81 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Generate and store pre-images for this cycle, as well as sparsely
+        selected preimages for other cycles.
+
+        This generates all pre-images needed for a cycle in the range of less
+        than two times of a validator cycle and store them in `cycle_preimages`.
+        Additionally, It generates and stores pre-images being used for defined
+        number of cycles in `preimage_rounds` when there is no pre-image needed.
+        This function is very expensive, but should be seldom called, and might
+        take more than 30ms for generating a hundred thousand number of them.
+
+        Params:
+            height = block height used to determine which range of
+                pre-images will be generated.
+
+        Returns:
+            the pre-image value in index of `height`
+
+    ***************************************************************************/
+
+    private Hash generatePreimages (ulong height) @safe nothrow
+    {
+        // This determines which range of preimages must be generated.
+        // In order to get hash values more than one cycle, the `start_height`
+        // is to be the height of the last preimage of next cycle. The value of
+        // `height / ValidatorSet.ValidatorCycle` is the index of previous
+        // cycle, so we need to plus 2 to the value in order to get the index
+        // of the next cycle.
+        ulong start_height =
+            ((height / ValidatorSet.ValidatorCycle) + 2) *
+                ValidatorSet.ValidatorCycle;
+
+        // Clear if recreating pre-images is needed
+        if (this.preimage_rounds.byKey.maxElement(0) < start_height)
+        {
+            () @trusted {
+                this.preimage_rounds.clear();
+            }();
+        }
+
+        // if there is no pre-image, the defined number of pre-images must
+        // be generated with the interval of ValidatorCycle.
+        if (this.preimage_rounds.length == 0)
+        {
+            // The value of `bulk_index` is zero-based. so we need to plus 1 to
+            // the value to get the max height(`MaxHeight`) of this bulk.
+            ulong bulk_index = start_height / AllCountPreimages;
+            const ulong MaxHeight = (bulk_index + 1) * AllCountPreimages;
+            Hash salt = hashFull(bulk_index);
+            auto hash = hashMulti(this.key_pair.v, salt);
+            ulong idx = MaxHeight;
+            this.preimage_rounds[idx] = hash;
+            for (--idx; idx >= height; --idx)
+            {
+                hash = hashFull(hash);
+                if (idx % ValidatorSet.ValidatorCycle == 0)
+                    this.preimage_rounds[idx] = hash;
+            }
+        }
+
+        // This generates all the sequential preimages from the start index
+        () @trusted {
+            this.cycle_preimages.clear();
+        }();
+        this.cycle_preimages[start_height] = this.preimage_rounds[start_height];
+        for (; start_height > height; --start_height)
+        {
+            this.cycle_preimages[start_height - 1] =
+                hashFull(this.cycle_preimages[start_height]);
+        }
+
+        return this.cycle_preimages[height];
+    }
+
+    /***************************************************************************
+
         Logs message
 
         Params:
@@ -709,28 +741,6 @@ public class EnrollmentManager
         }
         catch (Exception ex)
         {}
-    }
-
-    /***************************************************************************
-
-        Load pre-images from the storage
-
-        Returns:
-            an array of hashes of pre-images
-
-    ***************************************************************************/
-
-    version (unittest) public Hash[] loadPreimages () @safe
-    {
-        Hash[] preimages;
-        () @trusted {
-            auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
-                "WHERE key = ?", "preimages");
-            if (!results.empty)
-                preimages = results.oneValue!(ubyte[]).deserializeFull!(Hash[]);
-        }();
-
-        return preimages;
     }
 }
 
@@ -787,7 +797,7 @@ unittest
     fail_enroll.enroll_sig = sign(fail_enroll_key_pair.v, fail_enroll_key_pair.V,
         signature_noise.V, signature_noise.v, fail_enroll);
 
-    assert(man.createEnrollment(utxo_hash, enroll));
+    assert(man.createEnrollment(utxo_hash, 1, enroll));
     assert(!man.hasEnrollment(utxo_hash));
     assert(!man.add(0, &storage.findUTXO, fail_enroll));
     assert(man.add(0, &storage.findUTXO, enroll));
@@ -797,13 +807,13 @@ unittest
 
     // create and add the second Enrollment object
     auto utxo_hash2 = utxo_hashes[1];
-    assert(man.createEnrollment(utxo_hash2, enroll2));
+    assert(man.createEnrollment(utxo_hash2, 1, enroll2));
     assert(man.add(0, &storage.findUTXO, enroll2));
     assert(man.count() == 2);
 
     auto utxo_hash3 = utxo_hashes[2];
     Enrollment enroll3;
-    assert(man.createEnrollment(utxo_hash3, enroll3));
+    assert(man.createEnrollment(utxo_hash3, 1, enroll3));
     assert(man.add(0, &storage.findUTXO, enroll3));
     assert(man.count() == 3);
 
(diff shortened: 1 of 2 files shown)


linked0:
I'm adding more test code and rebasing now.

linked0:
Ready for review

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +639,56 @@ public class EnrollmentManager
...
+            {
+                hash = hashFull(hash);
+                if (--current_idx % ValidatorSet.ValidatorCycle == 0)
+                    this.preimages[current_idx] = hash;
+            }
+        }
+
+        // This determines which range of preimages must be generated.
+        ulong start_index =
+            ((current_height / ValidatorSet.ValidatorCycle) + 2) *
```
AndrejMitrovic: Why is there a `+2` here?

AndrejMitrovic:
Why not do something like this:

```D
enum TotalRounds = 1000;
Hash[TotalRounds] preimage_rounds;
Hash[ValidatorCycle] cur_cycle_preimages;
```

Then when you start a new voting cycle, you just call `cur_cycle_preimages.clear`, you find the preimage of a given voting round in `preimage_rounds[(current_height / ValidatorSet.ValidatorCycle)]` and then generate the preimages and store them to `cur_cycle_preimages`.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +639,56 @@ public class EnrollmentManager
...
+        auto hash = this.preimages[start_index];
+        foreach (int i; 1 .. ValidatorSet.ValidatorCycle * 2)
+        {
+            hash = hashFull(hash);
+            this.preimages[--start_index] = hash;
+        }
+
+        // Old pre-images are removed to save memory
+        () @trusted {
+            foreach (ulong i; this.preimages.keys)
```
AndrejMitrovic: This allocates a new array that will store all the keys.

There is an alternative `preimages.byKey`, however it cannot be used here. This is because when doing something like this:

```D
foreach (key; this.preimages.byKey)
    this.preimages.remove(key);
```

As soon as `remove` is called, the `byKey` range is in an invalid state. So the above code should not be used.

I have suggested an alternative in the main PR comment section.

source/agora/consensus/EnrollmentManager.d
linked0: > Why is there a +2 here?

For example, if the start height is 1800, it needs the 3024th hash value of the `this.preimages`. For detail, 1800 / 1008 is 1, and the value + 2 is 3,  and 3 * 1008 is 3024. That's why. Thank you.

linked0:
> Why not do something like this:

Initially, I considered something like your suggestion like `Hash[ValidatorCycle] cur_cycle_preimages` but I didn't wanna create another member  variable. But, your suggestion could make the code more readable. I will apply your suggestion to the code.

source/agora/consensus/EnrollmentManager.d
linked0: > There is an alternative preimages.byKey, however it cannot be used here. 

If `cur_cycle_preimages` is applied to the code, the code removing old pre-images is not needed anymore. Thank you.

linked0:
Ready for review

source/agora/consensus/EnrollmentManager.d
```diff
@@ -242,14 +241,16 @@ public class EnrollmentManager
         Params:
             frozen_utxo_hash = the hash of a frozen UTXO used to identify a validator
                         and to generate a siging key
+            current_height = the starting index for generating pre-images
```
AndrejMitrovic: But this is not the starting block. It is the height of the current ledger. If the enrollment is ever added to the blockchain, it will be at a future unspecified block height.

source/agora/consensus/EnrollmentManager.d
AndrejMitrovic: In the current design if the Validator wants to start enrolling at block height `1_000_000`, it means that 1 million hashes will be discarded after being generated. Because the preimage to be used is currently tied to the block height.

source/agora/consensus/EnrollmentManager.d
linked0: > But this is not the starting block. It is the height of the current ledger. If the enrollment is ever added to the blockchain, it will be at a future unspecified block height.

It's a starting height for generating pre-images, not a starting height as validator.

source/agora/consensus/EnrollmentManager.d
linked0: > In the current design if the Validator wants to start enrolling at block height `1_000_000`, it means that 1 million hashes will be discarded after being generated. Because the preimage to be used is currently tied to the block height.

You're right. I will change the line `foreach (ulong i; 0 .. AllCountPrimages)` to a line like `foreach (ulong i; current_height-1 .. AllCountPrimages)`

linked0:
Ready for review.
I applied @AndrejMitrovic 's suggestion and added the code about recreating pre-images.
Adding unexpected code like recreating pre-images could make reviewers uncomfortable. I'm sorry for that. Thank your for your understanding.

linked0:
How about changing the number of preimages being created from a million to a hundred thousand because It could take longer than 400ms. If we decrease the count, the time spent could be reduced to less than 1/10 of the original time spent. Please leave your opinion.

linked0:
@Geod24 Ping

source/agora/consensus/EnrollmentManager.d
```diff
@@ -49,16 +49,19 @@ public class EnrollmentManager
     private Database db;
 
     /// Node's key pair
+    private KeyPair node_key_pair;
```
Geod24: Why ? We already have the key pair, in another format sure, but it's the same data. We shouldn't introduce duplication here.

Geod24:
> Previously, we create a random value and create 1008 number of preimages from the value for the enrollment process. But we have decided to change the mechanism of creating a pre-images into generating a great number of pre-images like one million from the private key of a node.

I think it's important to keep *consensus behavior* separated from *implementation*.
A question you can ask yourself to tell them apart is: "Can someone else implement a node and not do that ?".
The 1008 preimages are required by the consensus protocol. On the other hand, the high number of preimage approach was suggested in order to solve a problem with data recovery. The former is required for the consensus protocol to work, the later is just a way to make things more robust, and an implementation detail (the consensus protocol does not verify this).

Geod24:
> How about changing the number of preimages being created from a million to a hundred thousand because It could take longer than 400ms. If we decrease the count, the time spent could be reduced to less than 1/10 of the original time spent. Please leave your opinion.

400ms is not much, considering creating that many preimage should be a very rare event.
This being say, changing the number of preimage generated is not a problem, either.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -112,13 +118,6 @@ public class EnrollmentManager
             break;
         }
 
-        // load preimages
-        results = this.db.execute("SELECT val FROM node_enroll_data " ~
-            "WHERE key = ?", "preimages");
-
-        if (!results.empty)
-            this.preimages = results.oneValue!(ubyte[]).deserializeFull!(Hash[]);
```
Geod24, line 120: Oh I see, so we don't store them anymore ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -242,14 +241,16 @@ public class EnrollmentManager
...
+            current_height = the starting index for generating pre-images
             enroll = will contain the Enrollment if created
 
         Returns:
             true if the enrollment manager succeeded in creating the Enrollment
 
     ***************************************************************************/
 
-    public bool createEnrollment (Hash frozen_utxo_hash, out Enrollment enroll) @trusted nothrow
+    public bool createEnrollment (Hash frozen_utxo_hash, ulong current_height,
```
Geod24: I wouldn't use the term "current" though. It is too ambiguous. What does it mean for the height to be the "current" one ?
Does it mean it's the height the ledger is at, or the network is at ? What happens if the height is not the current one, will the function return invalid results ? Throw ?

From what I can see in the code, it is just a height at which we want to start generating preimage. We can come up with a more appropriate name (`start_name` ? But it also have some issues...), or just use "height" and document things properly.

One principle you can use, when naming, is to think about what the name tells you about the data type.
Here the data type is `ulong`. So we know it's an unsigned number, and we have a range of 2^64 possible values.
The name `height` makes it obvious what it is for, in a blockchain context at least, and tells you a lot about what *values* it can take: Usually a small value, because `height` is essentially a count and starts from 0. You also know how those values will behave (increase monotonically), e.g. there won't be any "hole" in the range of values, if 10 is "valid" and 12 is "valid", so will be 11. Another example: Imagine you're debugging Agora, and you see somewhere a height of 563934289345235. Well you *know* it is wrong immediately, because it is so far outside the range of values you expect for a height.

So with all those considerations in mind, and assuming this function does instead accept any height, even one that would happen in the future, I suggest to rename this to simply `height`.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -525,9 +482,12 @@ public class EnrollmentManager
             (height - start_height) > ValidatorSet.ValidatorCycle - 1)
             return false;
 
+        if (height !in this.cycle_preimages)
+            generatePreimages(height);
```
```suggestion
            this.generatePreimages(height);
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +642,62 @@ public class EnrollmentManager
...
+
+        Generate pre-images
+
+        Params:
+            current_height = block height used to determine which range of
+                pre-images will be generated.
+
+    ***************************************************************************/
+
+    private void generatePreimages(ulong current_height) @safe nothrow
```
```suggestion
    private void generatePreimages (ulong current_height) @safe nothrow
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +642,62 @@ public class EnrollmentManager
...
+        ulong start_height =
+            ((current_height / ValidatorSet.ValidatorCycle) + 2) *
+                ValidatorSet.ValidatorCycle;
+        ulong bulk_index = start_height / AllCountPrimages;
+        const ulong MaxHeight = (bulk_index + 1) * AllCountPrimages;
+
+        // Clear if recreating pre-images is needed
+        () @trusted {
+            if (this.preimage_rounds.length > 0 &&
+                sort(this.preimage_rounds.keys)[$-1] < start_height)
```
Geod24: `std.algorithm.max` and `byKey`

source/agora/consensus/EnrollmentManager.d
```diff
@@ -77,6 +80,9 @@ public class EnrollmentManager
     /// enrolled height, and preimages.
     private ValidatorSet validator_set;
 
+    /// The count for generating pre-images
+    private immutable uint AllCountPrimages = ValidatorSet.ValidatorCycle * 1000;
```
```suggestion
    private immutable uint AllCountPreimages = ValidatorSet.ValidatorCycle * 1000;
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +642,62 @@ public class EnrollmentManager
...
+                this.preimage_rounds.clear();
+            }
+        }();
+
+        // if there is no pre-image, the defined number of pre-images must
+        // be generated with the interval of ValidatorCycle.
+        if (this.preimage_rounds.length == 0)
+        {
+            Hash salt = hashFull(bulk_index);
+            auto hash = hashFull([node_key_pair.secret, salt]);
```
Geod24: Just use `Pair.v`

Geod24: Also we have `hashMulti` for this

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +642,62 @@ public class EnrollmentManager
         }
     }
 
+    /***************************************************************************
+
+        Generate pre-images
+
```
```suggestion
        Generate and store pre-images for this cycle, as well as <what name do we use for entries in preimage rounds?>

        This generate all pre-images in the range [...] and store them in [...].
        Additionally, [<insert when/condition>] stores [...] in `preimage_rounds`.
        This function is very expensive, but should be seldom called (<insert time estimation>).
```

Geod24: ^ Note, this is pseudo documentation, edit the [...] and <...>.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +642,62 @@ public class EnrollmentManager
...
+    ***************************************************************************/
+
+    private void generatePreimages(ulong current_height) @safe nothrow
+    {
+        // This determines which range of preimages must be generated.
+        ulong start_height =
+            ((current_height / ValidatorSet.ValidatorCycle) + 2) *
+                ValidatorSet.ValidatorCycle;
+        ulong bulk_index = start_height / AllCountPrimages;
+        const ulong MaxHeight = (bulk_index + 1) * AllCountPrimages;
```
Geod24: I have a hard time understanding this calculation here? Why the `+ 2` for example? Doesn't that mean that `start_height` will always be >= 2016 ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -936,17 +923,85 @@ unittest
...
     assert(result_image.enroll_key == utxo_hash);
-    assert(result_image.hash == man.preimages[100]);
+    assert(result_image.hash == man.cycle_preimages[100]);
     assert(result_image.height == 1100);
 }
+
+// test for generating pre-images
+unittest
+{
+    KeyPair key_pair = KeyPair.random();
```
Geod24: Using random value can be deceiving. Because when using random values, you have to program your tests, which can themselves have bugs. To test the most basic properties of a function, we favor fixed value.
In this case you could start with a fixed value for the key and the salt and just compute that the hashes are what you expect.

source/agora/consensus/EnrollmentManager.d
linked0: > Why ? We already have the key pair, in another format sure, but it's the same data. We shouldn't introduce duplication here.

Oh, I missed that the two has same data.

source/agora/consensus/EnrollmentManager.d
linked0: > Oh I see, so we don't store them anymore ?

I found that we don't need to store them. It can be created or restored in the two places where the enrollment manager is creating pre-images and getting a preimage for revealing the one.

source/agora/consensus/EnrollmentManager.d
linked0: > So with all those considerations in mind, and assuming this function does instead accept any height, even one that would happen in the future, I suggest to rename this to simply height.

Sure, `height` would be appropriate and thank your for your opinion.

source/agora/consensus/EnrollmentManager.d
linked0: > I have a hard time understanding this calculation here? Why the + 2 for example? Doesn't that mean that start_height will always be >= 2016 ?

Considering your question, the following lines could be more accurate.
```
if ((current_height - 1) / ValidatorSet.ValidatorCycle == 0) 
    start_height = ((current_height / ValidatorSet.ValidatorCycle) + 1) * ValidatorSet.ValidatorCycle;
else
    start_height = ((current_height / ValidatorSet.ValidatorCycle) + 2) * ValidatorSet.ValidatorCycle;
```
The code is just for explaining.
Based on the above code, we will get 1008, 2016, 3124, ... for `start_height`.

For detailed example,  if the start height is 1800, it needs the 3024th hash value of the this.preimages. For detail, 1800 / 1008 is 1, and the value + 2 is 3, and 3 * 1008 is 3024.

But, the original code is not bad because we don't know how many preimages need to be prepared. It's because the height at which a node requests for a registration and the enrolled height are different.

linked0:
Thank you for your review and I will modify the code and request review again.
And if there is no problem on changing the number of preimages generated, I will change it.

source/agora/consensus/EnrollmentManager.d
linked0: > `std.algorithm.max` and `byKey`

Can you show me how to use the two functions for that? I can't find an exampe or what the byKey returns.

linked0: Is the code below like what you suggested?
```
ulong max_height = 0;
foreach (key; this.preimage_rounds.byKey)
    max_height = max(key, max_height);
```

Geod24: Almost. It's actually called https://dlang.org/phobos/std_algorithm_searching.html#.maxElement
So:
```D
auto max_height = this.preimage_rounds.byKey.maxElement();
```

source/agora/consensus/EnrollmentManager.d
Geod24: That makes sense.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +638,70 @@ public class EnrollmentManager
...
+        // This determines which range of preimages must be generated.
+        ulong start_height =
+            ((height / ValidatorSet.ValidatorCycle) + 2) *
+                ValidatorSet.ValidatorCycle;
+        ulong bulk_index = start_height / AllCountPreimages;
+        const ulong MaxHeight = (bulk_index + 1) * AllCountPreimages;
+
+        // Clear if recreating pre-images is needed
+        () @trusted {
+            if (this.preimage_rounds.length > 0 &&
```
Geod24: Look at the documentation: You can remove the `.length` call by using a seed

source/agora/consensus/EnrollmentManager.d
Geod24: Why not just:
```D
start_height = current_height + (ValidatorSet.ValidatorCycle - (current_height % ValidatorSet.ValidatorCycle));
```

linked0: > Why not just:
> 
>```d
>start_height = current_height + (ValidatorSet.ValidatorCycle - (current_height
>```

I will explain an example based on the code suggested by you. If `current_height` is 1800, `start_height` will be 2016. So we will be able to get only 217 random seeds from the height of 1800 to 2016 but we need more than 1008 random seeds.

source/agora/consensus/EnrollmentManager.d
linked0: > Look at the documentation: You can remove the .length call by using a seed

The `.length` call is for determining whether there are a bulk of pre-images already calculated. I don't understand the relation between removing the `.length` call and using a seed.

source/agora/consensus/EnrollmentManager.d
Geod24: How about adding `+ ValidatorSet.ValidatorCycle` then ?

source/agora/consensus/EnrollmentManager.d
Geod24: When I say `seed`, it's a reference to the argument `maxElement` can take. If you seed it with a value, it means it will have  at least this value. However if the range is empty, you know it will have this value.
Hence just doing `if (this.preimage_rounds.byKey.maxElement(0) < start_height)` allows you to remove the `.length > 0` check because the condition will be `false` when `length == 0`.

linked0: Ah, you meant the documentation for the `maxElement`. I will read and apply what you suggested.

source/agora/consensus/EnrollmentManager.d
linked0: > How about adding + ValidatorSet.ValidatorCycle then ?

If my code and your code output same result, I'd like to choose my code because my code is simpler.

Geod24: Alright, but then you need to handle `current_height == 0` and please add a comment why we have `+ 1` and `+ 2`.

linked0: > Alright, but then you need to handle current_height == 0 and please add a comment why we have + 1 and + 2.

Sure, thank you.

linked0:
Ready for review.
If there are any mistakes in the comments in the code written in english, please let me know.

linked0:
Please review or merge this PR.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -91,7 +94,6 @@ public class EnrollmentManager
     public this (string db_path, KeyPair key_pair)
     {
         this.validator_set = new ValidatorSet(db_path);
-
```
Geod24: Unrelated

source/agora/consensus/EnrollmentManager.d
```diff
@@ -262,17 +259,9 @@ public class EnrollmentManager
...
         // X, final seed data and preimages of hashes
-        this.preimages.length = 0;
-        assumeSafeAppend(this.preimages);
-        this.preimages ~= hashFull(this.random_seed_src);
-        foreach (i; 0 .. this.data.cycle_length-1)
-            this.preimages ~= hashFull(this.preimages[i]);
-        reverse(this.preimages);
-        this.data.random_seed = this.preimages[0];
+        this.generatePreimages(height);
+        this.data.random_seed = this.cycle_preimages[height];
```
Geod24: You could improve the API of `generatePreimages` to return the seed so you don't have to do the lookup, and the code is easier to follow.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -682,6 +638,76 @@ public class EnrollmentManager
...
+        // is to be the height of the last preimage of next cycle. The value of
+        // `height / ValidatorSet.ValidatorCycle` is the index of previous
+        // cycle, so we need to plus 2 to the value in order to get the index
+        // of the next cycle.
+        ulong start_height =
+            ((height / ValidatorSet.ValidatorCycle) + 2) *
+                ValidatorSet.ValidatorCycle;
+
+        // Clear if recreating pre-images is needed
+        () @trusted {
```
Geod24: Can `@trusted` be applied only on `clear` or does something in the `if` make it `@system` ?

Geod24 (approved):
Just a couple nitpicks, after they are addressed let's merge it.

Geod24 (approved):
LGTM. Merge at will.

PR #670, opened 2020-03-18
Dub: Use source globbing for object files

Geod24:
This simplifies our build script greatly. Also add support for Windows.

Inspired by https://github.com/bpfkorea/agora/pull/624


Geod24:
The globbing happens before `preGenerateCommands` 🤦‍♂

Geod24:
Sounds good, doesn't work.

PR #672, opened 2020-03-18
Make the C++ build script cross platform

Geod24:
Use a dub receipt and run it via DUB, so we can later add packages as dependency,
and it is cross platform (works on Windows).

In support of #624

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -11,7 +11,7 @@
     "targetType": "executable",
     "targetPath": "build",
     "excludedSourceFiles": [ "source/scpp/*.d" ],
-    "preGenerateCommands": [ "source/scpp/build.d" ],
+    "preGenerateCommands": [ "$DUB source/scpp/build.d" ],
     "sourceFiles": [
         "source/scpp/build/BallotProtocol.o",
         "source/scpp/build/ByteSliceHasher.o",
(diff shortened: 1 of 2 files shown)


source/scpp/build.d
```diff
@@ -1,4 +1,10 @@
-#!/usr/bin/env rdmd
+#!/usr/bin/env dub
+/+
+ dub.json:
+ {
+     "name": "cpp_build"
```
TrustHenry, line 5: Question How does the part declared here affect?

Geod24: It does not, it's just a placeholder.

dub.json
```diff
@@ -11,7 +11,7 @@
     "targetType": "executable",
     "targetPath": "build",
     "excludedSourceFiles": [ "source/scpp/*.d" ],
-    "preGenerateCommands": [ "source/scpp/build.d" ],
+    "preGenerateCommands": [ "$DUB source/scpp/build.d" ],
```
AndrejMitrovic, line 14: Oh you can use dub this way with a single file? That's nice.

AndrejMitrovic (approved):

Geod24:
For reference, this is what dub calls a "single file package": https://dub.pm/advanced_usage.html

PR #673, opened 2020-03-18
dub.json: Add correct extern-std definition

Geod24:
To avoid any mismatch with the C++ code, this needs to be in sync with build.d

Note that upstream  compiles with C++17 already. We should probably upgrade but there's a bit of work needed on the bindings first.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -37,6 +37,8 @@
         "source/scpp/build/numeric.o",
         "source/scpp/build/uint128_t.o"
     ],
+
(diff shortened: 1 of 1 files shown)


TrustHenry (approved):
LGTM

AndrejMitrovic:
> We should probably upgrade

I hope MSVC is up to date with their C++17 support too.

AndrejMitrovic (approved):

PR #679, opened 2020-03-19
Fix 678, make vector range usable in safe code, small serializer fixes

Geod24:
Better reviewed commit-by-commit.
Those small commits are part of a larger effort to get LocalRest to use Agora's serializer.


source/scpd/types/XDRBase.d
```diff
@@ -96,7 +96,11 @@ struct opaque_array(uint32_t N = XDR_MAX_LEN)
 {
     BitBlob!(N * 8) base;
     alias base this;
-    mixin ForwardCtors!base;
+
```
AndrejMitrovic, line 98: This seems to be the last place `ForwardCtors` was still used, so we could delete this code in the `Util` module as it's no longer used.

source/scpd/Cpp.d
```diff
@@ -192,40 +192,39 @@ extern(C++, (StdNamespace)) extern(C++, class) struct vector (T, Alloc = allocat
...
-                return !(this.ptr < this.orig._end);
+                return !(this.index < this.orig.length);
             }
         }
 
-        public ref inout(T) opIndex(size_t idx) inout pure nothrow @nogc
+        public ref inout(T) opIndex(size_t idx) inout pure nothrow @nogc @trusted
         {
-            assert(idx < this.length);
+            if (idx >= this.length)
```
AndrejMitrovic, line 215: Do we want to start using `verify` from ocean?

Geod24: I don't, do you ?

source/scpd/types/XDRBase.d
Geod24: ~In fact the whole module can be nuked~
Okay, only most of it.

source/scpd/Cpp.d
AndrejMitrovic: I was just thinking it simplifies the expression. Because instead of testing "if something is out of range, throw" it tests "this must be in this specific range".

Geod24:
Updated, nuked `scpd.Util` in the process.

AndrejMitrovic (approved):

Issue #665, opened 2020-03-17
`BanManager.banFor` accepts negative values


Geod24:
https://github.com/bpfkorea/agora/blob/dbd8a5cff0758cf13e59790dcbcc5100b63ba065/source/agora/common/BanManager.d#L155-L186

Shouldn't it take `ulong` instead of `long` ?
Also we should probably check that the value fits in a `time_t` (it fails on x86).

AndrejMitrovic:
~It should also use the monotonic clock.~

Edit: this comment was outdated. Please see #1226 for the associated issue.

PR #680, opened 2020-03-20
[Fix] the delegate does not compile on X64

TrustHenry:
This is the corrected PR for errors compiled by X64 & MSVC

Change deserialize delegate to be @trusted

On Windows the call to `rawRead` / `rawWrite` is not @safe

Cast to time_t for the time being (Related to  #665 )

Related to  #264

### DIFF
--- a/source/agora/common/BanManager.d
+++ b/source/agora/common/BanManager.d
@@ -91,7 +91,7 @@ public class BanManager
             return;  // nothing to load
 
         auto ban_file = File(this.banfile_path, "r");
-        scope DeserializeDg dg = (size) @safe
+        scope DeserializeDg dg = (size) @trusted
         {
             ubyte[] res;
             res.length = size;
@@ -111,7 +111,7 @@ public class BanManager
     public void dump ()
     {
         auto ban_file = File(this.banfile_path, "w");
-        this.serialize((scope bytes) => ban_file.rawWrite(bytes));
+        this.serialize((scope bytes) @trusted => ban_file.rawWrite(bytes));
     }
 
     /***************************************************************************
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
> Fix the delegate does compile
> there are multiple overloads of map and the compiler is a bit dumb

I don't understand this commit message. What overload of `map`?

Commit messages should be descriptive. If the issue is just about calling `rawWrite`, then we should name it something like:

```
Change deserialize delegate to be @trusted

On Windows the call to `rawRead` / `rawWrite` is not @safe
```

It explains:
- What the commit is about (first line)
- Why the change was made (second line)

Also saying "compiler is dumb" doesn't help the reader at all. But I think this was probably written by @Geod24.

AndrejMitrovic:
> ##[error]fatal: reference is not a tree: 330e5b4befb6b6cca8c53840fda27670c6abae4c

I wonder why this keeps happening in the CI..

TrustHenry:
> > Fix the delegate does compile
> > there are multiple overloads of map and the compiler is a bit dumb
> 
> I don't understand this commit message. What overload of `map`?
> 
> Commit messages should be descriptive. If the issue is just about calling `rawWrite`, then we should name it something like:
> 
> ```
> Change deserialize delegate to be @trusted
> 
> On Windows the call to `rawRead` / `rawWrite` is not @safe
> ```
> 
> It explains:
> 
> * What the commit is about (first line)
> * Why the change was made (second line)
> 
> Also saying "compiler is dumb" doesn't help the reader at all. But I think this was probably written by @Geod24.

@AndrejMitrovic Thanks for commit message.
Words that are not helpful to the reader have been deleted. :)

AndrejMitrovic (approved):

PR #684, opened 2020-03-20
[Draft] Implement a quorum balancing algorithm

AndrejMitrovic:
I've had several different approaches. Initially I tried to split up the stakes into three categories, but this made the algorithm really hard to reason about.

Then I've realized that I can just use the stakes as the chances of each node being included in a quorum set. It simplifies the algorithm quite a bit.

What has been done:
- SCP verifies the quorum set is correct
- Incorporated quorum intersection checker from Stellar into the tests
- Testing of _potential_ quorum splits, this is also part of the quorum intersection checker.
- There is a toml generator, and I am using it successfully with the standalone SCP implementation [go-scp](https://github.com/bobg/scp).

What's still on the table:
- More tests need to be added with hardcoded keys to ensure no regressions.
- Integration & byzantine node tests are missing.
- Does it make sense to limit the number of nodes in a quorum? Currently the algorithm stops adding nodes to a queue as soon as `MAX_NODES_IN_QUORUM` is reached, or 67% of the economic value of the entire network is reached.

Requires:
- ~#662~
- ~#664~

Part of: #240

### DIFF
--- a/source/agora/consensus/Quorum.d
+++ b/source/agora/consensus/Quorum.d
@@ -1,0 +1,1077 @@
+/*******************************************************************************
+
+    Contains the quorum generator algorithm.
+
+    Note that if you use AscendingQuadratic for too many nodes it will
+    assert as you will quickly run out of BOA.
+
+    Note thatintersection checks take roughly ~2 minutes for a configuration
+    of 16 nodes with their quorums, by default we disable these checks for
+    many nodes unless overriden with -version=EnableAllIntersectionChecks.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.Quorum;
+
+/// Force intersection checks on all unittests
+//version = EnableAllIntersectionChecks;
+
+/// Generate new assertion blocks when calibrating the algorithm
+//version = CalibrateQuorumBalancing;
+
+import agora.common.Amount;
+import agora.common.BitField;
+import agora.common.Config;
+import agora.common.crypto.Key;
+import agora.common.Hash;
+import agora.common.Set;
+import agora.common.Types;
+import agora.consensus.data.Enrollment;
+import agora.consensus.data.UTXOSet;
+import agora.consensus.EnrollmentManager;
+import agora.utils.PrettyPrinter;
+
+import scpd.Cpp;
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types : uint256, NodeID;
+import scpd.types.Utils;
+import scpd.types.XDRBase;
+import scpd.quorum.QuorumIntersectionChecker;
+import scpd.quorum.QuorumTracker;
+
+import std.algorithm;
+import std.array;
+import std.conv;
+import std.exception;
+import std.format;
+import std.math;
+import std.random;
+import std.range;
+import std.string;
+import std.typecons;
+
+version (unittest)
+{
+    import agora.utils.Log;
+    import std.conv;
+    import std.stdio;
+}
+
+/*******************************************************************************
+
+    Minimum number of nodes to include in a quorum.
+    The range is inclusive: length <= MIN_NODES_IN_QUORUM.
+
+    This is a soft limit, if there are less total nodes in the network than
+    MIN_NODES_IN_QUORUM then all of the nodes will be included in each node's
+    quorum set.
+
+    Since some nodes may have a large amount of stake compared to the
+    rest of the network, we want to ensure that other nodes get a chance of
+    inclusion in a quorum set.
+
+    For example, with nodes with this given stake
+    (assume units are multiples of MinFreezeAmount):
+
+    n0: 10, n1: 1, n2: 1, n3: 1, n4: 1
+
+    Despite n0 having significantly more stake than the rest of the nodes,
+    we do not want to centralize the quorum configuration to only include
+    this one node. This would lead to too much political (voting) power.
+
+    The voting power of a BOA holder may be increased by spawning multiple nodes,
+    for example:
+
+    [x0: 4, x1: 3, x2: 3], n2: 1, n3: 1, n4: 1, nE: 5
+
+    However, this will lead to the BOA holder to receive less rewards than if
+    they spawned a single node with 10 units. It's their choice on how to
+    distribute their voting power vs rewards ratio.
+
+*******************************************************************************/
+
+private enum MIN_NODES_IN_QUORUM = 5;
+
+/*******************************************************************************
+
+    Maximum number of nodes to include in a quorum.
+
+    Note: this may be relaxed in the future.
+
+*******************************************************************************/
+
+private enum MAX_NODES_IN_QUORUM = 7;
+
+/*******************************************************************************
+
+    Build the quorum configuration for the entire network of the given
+    registered enrollments. The random seed is used to shuffle the quorums.
+
+    Params:
+        enrolls = the array of registered enrollments
+        finder = the delegate to find UTXOs with
+        rand_seed = the source of randomness
+
+    Returns:
+        the map of all quorum configurations
+
+*******************************************************************************/
+
+public QuorumConfig[PublicKey] buildQuorumConfigs ( Enrollment[] enrolls,
+    UTXOFinder finder, Hash rand_seed )
+{
+    Amount[PublicKey] all_stakes = buildStakes(enrolls, finder);
+
+    QuorumConfig[PublicKey] quorums;
+    foreach (node, amount; all_stakes)
+        quorums[node] = buildQuorumConfig(node, enrolls, finder, rand_seed);
+
+    return quorums;
+}
+
+/*******************************************************************************
+
+    Build the quorum configuration for the given node key and the registered
+    enrollments. The random seed is used to shuffle the quorum config.
+
+    Params:
+        node_key = the key of the node
+        enrolls = the array of registered enrollments
+        finder = the delegate to find UTXOs with
+        rand_seed = the source of randomness
+
+    Returns:
+        the map of all quorum configurations
+
+*******************************************************************************/
+
+public QuorumConfig buildQuorumConfig ( PublicKey node_key,
+    Enrollment[] enrolls, UTXOFinder finder, Hash rand_seed )
+{
+    Amount[PublicKey] all_stakes = buildStakes(enrolls, finder);
+    NodeStake[] stakes_by_price = orderStakesDescending(all_stakes);
+
+    const Amount min_quorum_amount = Amount(
+        cast(ulong)(0.67 *
+            stakes_by_price.map!(stake => stake.amount.getRaw()).sum));
+
+    auto node_stake = node_key in all_stakes;
+    assert(node_stake !is null);
+
+    return buildQuorumImpl(node_key, *node_stake, stakes_by_price,
+        min_quorum_amount, rand_seed);
+}
+
+/// 2 nodes with equal stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(2, prev => prev);
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 128, n1: 128], counts.pretty);
+}
+
+/// 3 nodes with equal stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(3, prev => prev);
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 192, n1: 192, n2: 192], counts.pretty);
+}
+
+/// 3 nodes with ascending stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(3, prev => prev.mustAdd(Amount.MinFreezeAmount));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 101, n1: 192, n2: 192], counts.pretty);
+}
+
+/// 3 nodes with descending stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(3, prev => prev.mustSub(Amount.MinFreezeAmount),
+        Amount.MinFreezeAmount.mul(3));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 192, n1: 192, n2: 102], counts.pretty);
+}
+
+/// 3 nodes with ascending quadratic stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(3, prev => prev.mustAdd(prev));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 96, n1: 134, n2: 192], counts.pretty);
+}
+
+/// 4 nodes with equal stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(4, prev => prev);
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 193, n1: 196, n2: 196, n3: 183], counts.pretty);
+}
+
+/// 4 nodes with ascending stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(4, prev => prev.mustAdd(Amount.MinFreezeAmount));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 111, n1: 139, n2: 217, n3: 256], counts.pretty);
+}
+
+/// 4 nodes with ascending quadratically increasing stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(4, prev => prev.mustAdd(prev));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 96, n1: 117, n2: 219, n3: 256], counts.pretty);
+}
+
+/// 8 nodes with equal stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(8, prev => prev);
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 331, n1: 315, n2: 309, n3: 333, n4: 321, n5: 316, n6: 319, n7: 316], counts.pretty);
+}
+
+/// 8 nodes with ascending stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(8, prev => prev.mustAdd(Amount.MinFreezeAmount));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 134, n1: 206, n2: 259, n3: 303, n4: 345, n5: 398, n6: 404, n7: 438], counts.pretty);
+}
+
+/// 8 nodes with ascending quadratically increasing stakes
+// todo: fails with the quorum split check,
+// awaiting answer on https://stellar.stackexchange.com/q/3038/2227
+version (none)
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(8, prev => prev.mustAdd(prev));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 110, n1: 134, n2: 158, n3: 167,
+                      n4: 216, n5: 253, n6: 284, n7: 326,
+                      n8: 347, n9: 405, n10: 392, n11: 430,
+                      n12: 429, n13: 480, n14: 488, n15: 501], counts.pretty);
+}
+
+/// 16 nodes with equal stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(16, prev => prev);
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 318, n1: 310, n2: 308, n3: 328, n4: 312, n5: 327, n6: 317, n7: 303, n8: 314, n9: 322, n10: 327, n11: 337, n12: 327, n13: 334, n14: 285, n15: 351], counts.pretty);
+}
+
+/// 16 nodes with ascending stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(16, prev => prev.mustAdd(Amount.MinFreezeAmount));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 90, n1: 125, n2: 185, n3: 182, n4: 222, n5: 246, n6: 299, n7: 296, n8: 345, n9: 358, n10: 397, n11: 443, n12: 467, n13: 441, n14: 503, n15: 521], counts.pretty);
+}
+
+/// 16 nodes with ascending stakes (Freeze * 2)
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(16, prev => prev.mustAdd(Amount.MinFreezeAmount.mul(2)));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 82, n1: 107, n2: 157, n3: 182, n4: 211, n5: 253, n6: 280, n7: 321, n8: 337, n9: 363, n10: 386, n11: 461, n12: 486, n13: 444, n14: 513, n15: 537], counts.pretty);
+}
+
+/// 16 nodes with ascending stakes (Freeze * 4)
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(16, prev => prev.mustAdd(Amount.MinFreezeAmount.mul(4)));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 75, n1: 106, n2: 141, n3: 184, n4: 195, n5: 263, n6: 280, n7: 313, n8: 328, n9: 376, n10: 392, n11: 468, n12: 479, n13: 456, n14: 516, n15: 548], counts.pretty);
+}
+
+/// 32 nodes with equal stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(32, prev => prev);
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 302, n1: 323, n2: 307, n3: 328, n4: 315, n5: 338, n6: 295, n7: 310, n8: 307, n9: 305, n10: 312, n11: 332, n12: 330, n13: 349, n14: 321, n15: 355, n16: 314, n17: 338, n18: 303, n19: 312, n20: 319, n21: 332, n22: 306, n23: 313, n24: 343, n25: 325, n26: 309, n27: 268, n28: 339, n29: 331, n30: 319, n31: 340], counts.pretty);
+}
+
+/// 32 nodes with ascending stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(32, prev => prev.mustAdd(Amount.MinFreezeAmount));
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
+    Assert(counts == [n0: 82, n1: 93, n2: 111, n3: 129, n4: 145, n5: 172, n6: 176, n7: 186, n8: 179, n9: 226, n10: 249, n11: 248, n12: 282, n13: 267, n14: 285, n15: 307, n16: 330, n17: 377, n18: 352, n19: 382, n20: 404, n21: 356, n22: 427, n23: 440, n24: 463, n25: 462, n26: 483, n27: 464, n28: 532, n29: 524, n30: 545, n31: 562], counts.pretty);
+}
+
+/// 64 nodes with equal stakes
+unittest
+{
+    auto enrolls = genEnrollsAndFinder(64, prev => prev);
+    auto counts = countNodeInclusions(enrolls, getRandSeeds(64));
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
I started writing better tests and realized the node would not automatically add itself in the list of quorum nodes, which was a mistake. I will update the PR with the tests & fixes in the coming days..

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+        catch (Throwable thr)
+        {
+            thr.line = line;
+            throw thr;
+        }
+    }
+}
+
+/// very simplistic way of reducing a 64-byte blob to an 8-byte seed
+private ulong hashToSeed (Hash hash)
```
AndrejMitrovic: We could probably replace this with libsodium's `crypto_shorthash`. But it needs a key, and the key should be hardcoded (we need the reduction to be reproducible given the same input).

AndrejMitrovic:
This needs to be reworked a little bit.

There is an edge-case. What if one node has 66% of the staked amount of the entire network? In the current algorithm most nodes will end up with a single node in its quorum. We should add some minimum amount of nodes that every node should include. We could make it something simple like "must include at least 5 nodes, and if those nodes do not reach a majority of stakes then add additional nodes".

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
+/*******************************************************************************
+
+    Contains the quorum generator algorithm.
+
+    We have two sets of tests, based on what's possible:
+
+    - countQuorumMatches(): used when the number of nodes is low.
```
AndrejMitrovic: If this is used rarely, then we could just remove it and test the quorum configs manually for these edge-cases.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+
+    - countNodeInclusions(): it counts the number of times a node
+        has been included in other nodes' quorum sets. Using this
+        makes it easy to determine if the algorithm behaves correctly
+        when the staken amounts are differen between nodes.
+
+    Note: If you use AscendingQuadratic for too many nodes, it will
+    assert as you will quickly run out of BOA.
+
+    Note: Intersection checks take roughly ~2 minutes for a single configuration
```
AndrejMitrovic: I wonder about why I hit this performance issue. The Stellar network has hundreds (thousands?) of nodes, it sounds like this algorithm would never finish if it attempted to do quorum intersection checking for the live system. Examine what's causing the performance issue. Maybe the bindings are wrong, the GC is kicking in, or something else.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+
+import std.stdio;
+
+version (unittest)
+{
+    import std.conv;
+}
+
+/// limits the number of nodes in a quorum set
+private enum MAX_NODES_IN_QUORUM = 7;
```
AndrejMitrovic, line 110: As mentioned in another comment, there should be a minimum too. We don't want to end up with nodes which only have 1 other nodes in its quorum set. Although tests should cover this too.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+        rand_seed = the source of randomness
+
+    Returns:
+        the map of all quorum configurations
+
+*******************************************************************************/
+
+public QuorumConfig buildQuorumConfig ( PublicKey node_key,
+    Enrollment[] enrolls, UTXOFinder finder, Hash rand_seed,
+    size_t line = __LINE__ )
```
AndrejMitrovic: `line` is unused, remove it.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+{
+    Amount[PublicKey] all_stakes = buildStakes(enrolls, finder);
+    NodeStake[] stakes_by_price = orderStakesDescending(all_stakes);
+
+    const Amount min_quorum_amount = Amount(
+        cast(ulong)(10_000_000 * (0.67 *  // todo: add multiply() support
```
AndrejMitrovic: Turn this `todo` into an issue and work on it.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+            stakes_by_price.map!(stake => stake.amount.integral)
+                .sum)));
+
+    auto node_stake = node_key in all_stakes;
+    assert(node_stake !is null);
```
AndrejMitrovic: Need to ask ourselves whether this is really a programming error or a user error.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+    return buildQuorumImpl(node_key, *node_stake, stakes_by_price,
+        min_quorum_amount, rand_seed);
+}
+
+/*******************************************************************************
+
+    Build the quorum configuration for the entire network of the given
+    registered enrollments. The random seed is used to shuffle the quorums.
+
+    Note: used with unittests, verifies the quorum intersection property for the
```
AndrejMitrovic: Actually as discovered recently, this function should be public. The reason why is because it allows us to build the `hash => quorum set` for the entire network. Therefore we won't need to ask other nodes about unknown quorum set hashes.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+        rand_seed = the source of randomness
+
+    Returns:
+        the map of all quorum configurations
+
+*******************************************************************************/
+
+version (unittest)
+private QuorumConfig[PublicKey] buildQuorumConfigs ( Enrollment[] enrolls,
+    UTXOFinder finder, Hash rand_seed, size_t line = __LINE__ )
```
AndrejMitrovic: After making it public, get rid of `line`.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+{
+    Amount[PublicKey] all_stakes = buildStakes(enrolls, finder);
```
AndrejMitrovic, line 130: Consider using a buffer for this. `buildQuorumConfigs` may be called once every N blocks, we should still try to avoid producing garbage for GC to clean.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+
+    QuorumConfig[PublicKey] quorums;
+    foreach (node, amount; all_stakes)
+        quorums[node] = buildQuorumConfig(node, enrolls, finder, rand_seed);
+
+    verifyQuorumsSanity(quorums);
```
AndrejMitrovic: Is there ever a possibility that this verification fails in production? And if it can, what is our contingency plan?

If this test is only supposed to happen in unittests then it should be moved into unittest blocks.

If this test must remain here, and it can fail in production, then we need to have a backup plan. Do we attempt to create another quorum, maybe by hashing the `rand_seed` until a good quorum is generated and `verifyQuorumsSanity` passes?

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+
+    // Intersection checks take roughly ~2 minutes for a configuration
+    // of 16 node quorums, and grows exponentially after that.
+    // By default we disable these checks unless overriden with
+    // -version=EnableAllIntersectionChecks
+    version (EnableAllIntersectionChecks)  // enable all checks
```
AndrejMitrovic: As mentioned this needs to be double-checked for the root cause of the performance bottleneck.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+                      n32: 491, n33: 495, n34: 469, n35: 458,
+                      n36: 536, n37: 520, n38: 554, n39: 530,
+                      n40: 546, n41: 572, n42: 560, n43: 573,
+                      n44: 584, n45: 618, n46: 606, n47: 591,
+                      n48: 609, n49: 662, n50: 635, n51: 679,
+                      n52: 715, n53: 690, n54: 703, n55: 768,
+                      n56: 742, n57: 765, n58: 738, n59: 775,
+                      n60: 767, n61: 774, n62: 792, n63: 798], counts.pretty);
+}
+
```
AndrejMitrovic, line 330: Add another test where a single node has a vast majority of the stake, while the rest of the nodes have less stake. We should not end up in a degenerate case where the entire rest of the network only has 1 node in its quorum set (other than themselves).

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+    assert(q0 == QuorumConfig(5, [n57, n34, n0, n46, n43, n60, n7]), q0.nice);
+
+    auto q1 = buildQuorumConfig(n1, enrolls.expand, rand_seed);
+    assert(q1 == QuorumConfig(5, [n34, n39, n1, n12, n54, n8, n48]), q1.nice);
+
+    auto q63 = buildQuorumConfig(n63, enrolls.expand, rand_seed);
+    assert(q63 == QuorumConfig(5, [n17, n63, n50, n60, n44, n56, n49]), q63.nice);
+}
+
+/// converts "GA..." => n0 / n1, etc for easier debugging
```
AndrejMitrovic: The next several declarations are support code for the usability of debugging unittests. But after that, there is a `buildQuorumImpl` which is part of the implementation code.

The support code should be moved to the end of the module so we don't mix helper functions with actual functionality.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+private string[PublicKey] key_to_simple;
```
AndrejMitrovic: Shouldn't this be within a `version (unittest)` block? Same question for most of these declarations.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+// convenience so we can use [n0, n1] when refering to public keys in the tests
+static this ()
+{
+    static foreach (node_idx; 0 .. pregen_seeds.length)
+        mixin("n%1$s = getTestKey(%1$s);".format(node_idx));
+}
+
+#line __LINE__
+
+/// Return the count of each node's inclusions in other nodes' quorums
```
AndrejMitrovic: This needs a better description.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+            foreach (node; quorum.nodes)
+                counts[node]++;
+        }
+    }
+
+    return counts;
+}
+
+version (unittest)
+private size_t[N] countQuorumMatches (size_t N, Enrolls, Range)(
```
AndrejMitrovic: This has no documentation..

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+    }
+
+    return counts;
+}
+
+/// Generate random seeds by hashing a range of numbers: [0 .. count)
+version (unittest)
+private auto getRandSeeds (size_t count)
+{
+    // test-cases already assumed 'int' by mistake, keeping it here
```
AndrejMitrovic: This shows that it's difficult to modify existing tests. If we want to change the algorithm (and we currently might do), it should be easy to fix the code.

So:
- Fix this `int` mistake
- Make the tests easier to fix if the algorithm needs modification in the future

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+
+    Params:
+        quorums = the quorum map of (node => quorum) to verify
+
+    Throws:
+        AssertError if the quorum is not considered sane by SCP.
+
+*******************************************************************************/
+
+private void verifyQuorumsSanity (QuorumConfig[PublicKey] quorums)
```
AndrejMitrovic, line 443: What is the difference between this function and `verifyQuorumsIntersect`? The documentation should say which types of checks it does.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+        CircularAppender().printConsole();
+        assert(0);
+    }
+
+    auto splits = qic.getPotentialSplit();
+
+    if (splits.first.length != 0 ||
+        splits.second.length != 0)
+    {
+        CircularAppender().printConsole();
```
AndrejMitrovic: This seems out of place.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+        auto hash = StellarHash(key);
+        auto scp_key = NodeID(hash);
+
+        qm[scp_key] = scp_quorum;
+    }
+
+    auto qic = QuorumIntersectionChecker.create(qm);
+    if (!qic.networkEnjoysQuorumIntersection())
+    {
+        CircularAppender().printConsole();
```
AndrejMitrovic: This seems out of place.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+            splits.first[].map!(node_id => key_to_simple[PublicKey(node_id)]),
+            splits.second[].map!(node_id => key_to_simple[PublicKey(node_id)]));
+
+        writefln("Quorum: %s", quorums.nice(0));
+
+        try
+        {
+            assert(0, "Splits found");  // should not happen
+        }
+        catch (Throwable thr)
```
AndrejMitrovic: Remove this entire try/catch, it has no reason to exist.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+    }
+
+    auto splits = qic.getPotentialSplit();
+
+    if (splits.first.length != 0 ||
+        splits.second.length != 0)
+    {
+        CircularAppender().printConsole();
+
+        writefln("Splits: first: %s second: %s",
```
AndrejMitrovic: Add some documentation on what "first" and "second" mean.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+private ulong hashToSeed (Hash hash)
+{
+    return (cast(ulong[])hash[]).reduce!((a, b) => a ^ b);
+}
+
+/*******************************************************************************
+
+    Create a random number generator which uses the hash of the random seed
+    and a node's public key as an initializer for the engine.
+
```
AndrejMitrovic, line 545: Whichever number generator we end up using, we'll have to commit to. So it would be good to mention why this particular one was selected.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+private auto getGenerator (PublicKey node_key, Hash rand_seed)
+{
+    Mt19937_64 gen;
+    gen.seed(hashToSeed(hashMulti(node_key, rand_seed)));
+    return gen;
+}
+
+/// The pair of (key, stake) for each node
+private struct NodeStake
```
AndrejMitrovic, line 569: Is this definition necessary, or could we just use `map` and `tuple` for example?

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+        a descending set of nodes based on their stakes
+
+*******************************************************************************/
+
+private NodeStake[] orderStakesDescending (Amount[PublicKey] stake_map)
+{
+    auto stakes = stake_map
+        .byKeyValue
+        .map!(pair => NodeStake(pair.key, pair.value))
+        .array;
```
AndrejMitrovic, line 624: This allocates. Replace with a non-allocating version.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+        Q = {t = 1, m = [{q = {t = 2, m = [{n = "bob"}, {n = "carol"}]}},
+                         {q = {t = 2, m = [{n = "dave"}, {n = "elsie"}]}}]}
+
+    Params:
+        quorums = the quorum set
+        level = indentation level
+
+*******************************************************************************/
+
+private string toToml (QuorumConfig[PublicKey] quorums)
```
AndrejMitrovic, line 659: The rest of the code from this point on is just for `.toml` generation.

However, we won't actually need this that often. And I'm not sure the golang implementation can be trusted since it's an abandoned clean-room implementation project. So it doesn't really belong here. Save it as a snippet somewhere, and remove it from the PR.

Edit: it's useful for tests, I'll leave it in.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+
+            hashToName[hash] = name;
+            return name;
+        }
+    }
+
+    return toUserReadable(input[].hashFull());
+}
+
+// wrapper around Amount with some unsafe primitives
```
AndrejMitrovic: Fix any issues with `Amount` rather than writing these sorts of wrappers.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+
+
+version (unittest)
+private PublicKey getTestKey (size_t idx)
+{
+    return KeyPair.fromSeed(Seed.fromString(pregen_seeds[idx])).address;
+}
+
+/// for debugging
+private string pretty (int[PublicKey] counts)
```
AndrejMitrovic, line 966: Some of the helper functions in this module are named `nice`, others `pretty`. We should be consistent.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1229 @@
...
+    foreach (key; keys)
+        res ~= format("%s: %s, ", key, result[key]);
+
+    return res;
+}
+
+// using pregenerated seeds to test the determenistic behavior of the algorithm
+// note: sorted for easier comparison with PublicKey[]
+version (unittest)
+private immutable string[64] pregen_seeds =
```
AndrejMitrovic, line 985: This part made me realize something about how we use seeds in the test-suite. I'll file an issue for it.

source/agora/utils/PrettyPrinter.d
```diff
@@ -45,7 +45,7 @@ unittest
 
 /// Returns:
 /// A formatting struct for a type, or the value if no such struct exists
-public auto prettify (T) (const ref T input) nothrow
+public auto prettify (T) (auto const ref T input) nothrow
```
AndrejMitrovic, line 48: This is a sneaky change and it should be moved to a separate commit, or removed entirely.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1259 @@
...
+    import scpd.scp.QuorumSetUtils;
+
+    foreach (key, quorum; quorums)
+    {
+        auto scp_quorum = toSCPQuorumSet(quorum);
+        const(char)* reason;
+
+        assert(isQuorumSetSane(scp_quorum, true, &reason),
+            format("Key %s with %s fails sanity check before normalization: %s",
+                    key.nice, quorum.toToml, reason.to!string));
```
AndrejMitrovic: This can't be used here. `nice` should only be available in unittests, using it in production we'll quickly run out of key => name maps.

source/agora/consensus/Quorum.d
AndrejMitrovic: I'm going to have to remove this piece of logic. There's two reasons:

- Floating point math could lead to inconsistent results across platforms. We need consistency here. Perhaps we could use fixed-point math, but it would require looking for a good library implementation.
- It doesn't entirely make sense that we're adding nodes into a quorum to make that quorum's monetary value 67% of the network, while at the same time having a 67% threshold value for those included nodes. That would actually turn into 44% with the threshold taken into account. Also, within that quorum set SCP treats every node as equal. This means nodes with 1/10th of the quorum stake and those with 9/10th of the value are treated the same. So monetary value != political (voting) power, as it should be. Finally, the node selection logic already makes sure that nodes with a higher stake are more frequently included in a quorum, so the `min_quorum_amount` is unnecessary.

linked0: There is a standard for double-precision that is IEEE 754. Modern 32-bit computers implement double-precision with two 32-bit memory words and 64-bit computers can do it with its 64-bit memory word. So We don't need to worry about the output of the floating-point operation.

source/agora/consensus/Quorum.d
Geod24: Given that stellar has implemented an interrupt method, it might just be that the algorithm is inherently NP-hard.
The comment mentions 2 minutes and "16 node quorums", does that mean there are 16 nodes in the network overall, or each quorums have 16 nodes ?

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1118 @@
+/*******************************************************************************
+
+    Contains the quorum generator algorithm.
+
+    Note: If you use AscendingQuadratic for too many nodes, it will
```
Geod24: Don't use `Note: `, it generates bad doc IIRC.

Also this could use more actual documentation.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1118 @@
...
+    Since some nodes may have a large amount of stake compared to the
+    rest of the network, we want to ensure that other nodes get a chance of
+    inclusion in a quorum set.
+
+    For example, with nodes with this given stake
+    (assume units are multiples of MinFreezeAmount):
+
+    nA: 10, nB: 1, nC: 1, nD: 1, nE: 1
+
+    Despite n0 having significantly more stake than the rest of the nodes,
```
Geod24, line 85: Inconsistency between `nA` and `n0`. Personally I'm all for `n0` if we stick to ASCII.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1118 @@
...
+    we do not want to centralize the quorum configuration to only include
+    this one node. This would lead to too much political (voting) power.
+
+    The voting power of a Boa holder may be increased by spawning multiple nodes,
```
Geod24:
```suggestion
    The voting power of a BOA holder may be increased by spawning multiple nodes,
```

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1118 @@
...
+    for example:
+
+    [xA: 4, xB: 3, xC: 3], nB: 1, nC: 1, nD: 1, nE: 1
+
+    However, this will lead to the Boa holder to receive less rewards than if
+    they spawned a single node with 10 units. It's their choice on how to
+    distribute their voting power vs rewards ratio.
```
Geod24, line 96: Nice explanation. We should add the general consideration in the module doc. But that can be left for later.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1118 @@
...
+
+    Note: if there are less total nodes in the network than MIN_NODES_IN_QUORUM,
+    then all of the nodes will be included in each node's quorum set.
+
+*******************************************************************************/
+
+private enum MIN_NODES_IN_QUORUM = 5;
```
Geod24, line 100: Inclusive or exclusive bound ? I suppose inclusive, and ditto for MAX, but just want to be extra clear.

source/agora/consensus/Quorum.d
Geod24: I've been wanting to add a module with "test data" for a while, I think it could be useful to have somewhere to get "valid, initialized data".

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1118 @@
...
+
+/// Generate random seeds by hashing a range of numbers: [0 .. count)
+private auto getRandSeeds (size_t count)
+{
+    // using 'ulong' to get consistent hashes
+    return iota(0, count).map!(idx => hashFull(cast(ulong)idx));
+}
+
+/// ditto
+private string pretty (PublicKey input)
```
Geod24, line 813: Noice. Probably something we want to generalize in the long run. I mean Ocean has it, and I think stellar does too.
Something to consider: For easier memorization, you can use a noun and a verb, like what docker does. E.g. "patriot pangolin" and "mesmerizing door" might stand out more than "Andrew" and "Nathan".

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1118 @@
...
+    return amount;
+}
+
+/// convenience
+private Amount subbed (in Amount lhs, in Amount rhs)
+{
+    Amount amount = lhs;
+    assert(amount.sub(rhs));
+    return amount;
+}
```
Geod24: https://github.com/bpfkorea/agora/blob/20decfd4fed2b43a55e444c284d5a88d44ef0a75/source/agora/common/Amount.d#L184-L198

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,1118 @@
...
+    return format("\nQuorumConfig(%s, %s)",
+        conf.threshold,
+        nodes);
+}
+
+// convenience so we can use [n0, n1] when refering to public keys in the tests
+static foreach (node_idx; 0 .. pregen_seeds.length)
+    mixin("PublicKey n%s;".format(node_idx));
+
+#line __LINE__
```
Geod24: Please tell me you have a bug report for this ?

Geod24 (commented):
Okay did a first pass. There's obviously A LOT to review and discuss here, will continue soon.

source/agora/consensus/Quorum.d
AndrejMitrovic: I did this out of memory because I seem to remember mixins messing with line numbers. But that was years ago. I'll try to remove it.

AndrejMitrovic:
> The comment mentions 2 minutes and "16 node quorums", does that mean there are 16 nodes in the network overall, or each quorums have 16 nodes ?

16 nodes, where each node has some number of nodes in its quorum set.

And yeah, it's unfortunate that the algorithm is inherently slow.. hence the interrupts.

source/agora/consensus/Quorum.d
AndrejMitrovic: Yeah. It would be nice to have consistent keys in our test-suite, it helps with debugging.

https://github.com/bpfkorea/agora/issues/754

source/agora/consensus/Quorum.d
AndrejMitrovic: > Also this could use more actual documentation.

Well it's a draft. If half the code gets thrown away I'd hate to throw away the documentation too. I'll add it as soon as it's finalized.

source/agora/consensus/Quorum.d
AndrejMitrovic: What if they're holding snakes huh??

source/agora/consensus/Quorum.d
AndrejMitrovic: It's more like a soft minimum. If there are less than 5 validators enrolled then this check would never pass. I should document that.

Geod24: I meant `<` or `<=`

source/agora/consensus/Quorum.d
Geod24: We send them a job offer

source/agora/consensus/Quorum.d
AndrejMitrovic: ah yes, the gifycat approach.

Or I could use https://github.com/bpfkorea/ocean/blob/dev-2020-04-02/src/ocean/io/digest/FirstName.d :D

source/agora/consensus/Quorum.d
Geod24: It does, but I don't know the details of how / when

source/agora/consensus/Quorum.d
AndrejMitrovic: Those modify the object. So I can't use them with const values.

E.g. I can't do `Amount.MinFreezeAmount.mustAdd(Amount.MinFreezeAmount)`.

Geod24: `Amount(Amount.MinFreezeAmount).mustAdd(Amount.MinFreezeAmount)` ?

AndrejMitrovic:
Windows linker error:

```
agora.obj : error LNK2019: unresolved external symbol "public: static class std::shared_ptr<class stellar::QuorumIntersectionChecker *> __cdecl stellar::QuorumIntersectionChecker::create(struct std::unordered_map<struct stellar::PublicKey,class std::shared_ptr<struct stellar::SCPQuorumSet>,struct std::hash<struct stellar::PublicKey>,struct std::equal_to<struct stellar::PublicKey>,class std::allocator<struct std::pair<struct stellar::PublicKey const ,class std::shared_ptr<struct stellar::SCPQuorumSet> > > > const &)" (?create@QuorumIntersectionChecker@stellar@@SA?AV?$shared_ptr@PEAVQuorumIntersectionChecker@stellar@@@std@@AEBU?$unordered_map@UPublicKey@stellar@@V?$shared_ptr@USCPQuorumSet@stellar@@@std@@U?$hash@UPublicKey@stellar@@@4@U?$equal_to@UPublicKey@stellar@@@4@V?$allocator@U?$pair@$$CBUPublicKey@stellar@@V?$shared_ptr@USCPQuorumSet@stellar@@@std@@@std@@@4@@4@@Z) referenced in function _D5agora9consensus6Quorum22verifyQuorumsIntersectFHSQBy6common6crypto3Key9PublicKeySQDeQBg5Types12QuorumConfigZv
```

_internal screaming_

source/agora/consensus/Quorum.d
MichaelKim20: The condition for the quorum to work normally is not the number of nodes, but **the number of validators enrolled**.
I understood well because you explained it well.

AndrejMitrovic:
For those that are curious what we talked about during our whiteboarding session:

I needed to define some constraints or properties of the algorithm. For example, knowing what the optimal quorum layout should look like.

We want several properties of the algorithm:
- It should be possible to shuffle the quorums. This can be done by using the validator's revealed preimages. The preimages are combined into a "seed", which is then fed to the quorum balancing algorithm.
- Nodes which stake a higher amount should be included more often in quorum sets.
- The amount of required messaging between nodes should be minimized.

This last part is my main focus for this sprint.

AndrejMitrovic:
This should be reworked based on what was introduced in #845. We could add the very first enhancements to the quorum generator. I'll keep this in mind as a background task.

AndrejMitrovic:
This will be replaced with a series of simpler PRs.

Issue #323, opened 2019-09-10
Add utilities which verify the vtables match on the D & C++ sides


AndrejMitrovic:
It's very easy to accidentally forget to prototype a method on the D side, causing the vtable to mismatch. This manifests in what appear to be weird memory corruption bugs, but in reality the stack trace (e.g. via LLDB) will tell us if the call stack makes sense, or if the wrong method was called.

We should have some utility code which verifies this, just like the current layout/size checks.

AndrejMitrovic:
To expand on the issue, we need to verify that the **offsets** between C++ and D match.

Here is an example test:

```D
// D code
extern(C++) class SCPDriver
{
    abstract void signEnvelope(ref SCPEnvelope envelope);
    abstract SCPQuorumSetPtr getQSet(ref const(Hash) qSetHash);
}
```

```C++
// C++ code
class SCPDriver
{
    virtual SCPQuorumSetPtr getQSet(Hash const& qSetHash) = 0;
    virtual void signEnvelope(SCPEnvelope& envelope) = 0;
};
```

Notice the mistake here. On the D side the layout of the functions is wrong. `getQSet` should be declared **before** the `signEnvelope` is declared. This means that in the virtual table, we will have:

```
vtable[0] = signEnvelope
vtable[1] = getQSet
```

But what is actually correct is:

```
vtable[0] = getQSet
vtable[1] = signEnvelope
```

So we need some mechanism of verifying that the layout in the D bindings matches the layout in the C++ code.

This is not easy to solve. It will likely require writing a new D tool which imports into all the D bindings and generates C++ code which asserts that the offsets are correct.

It would be something like this:

```D
module VTableCheckGenerator;

import scpd.scp.SCPDriver;

void main ()
{
    string[] methods;
    foreach (member; __traits(allMembers, SCPDriver))
    {
        if (__traits(isVirtualMethod, member))
            methods ~= member;
    }

    foreach (idx, name; methods)
    {
        // generate C++ code as a string, that verifies that method with this name
        // is at this given offset
    }
}
````

It's a more involved task.

**Comment by MukeunKim** [comment]
I could get the following code.
```
class SCPDriver
{
  virtual void signEnvelope (ref SCPEnvelope) = 0;
  virtual shared_ptr!(SCPQuorumSet) getQSet (ref const(opaque_array!32u)) = 0;
  virtual void emitEnvelope (ref const(SCPEnvelope)) = 0;
[...]

PR #686, opened 2020-03-20
Adds a CLI that automatically generates C++ code

MichaelKim20:
Adds a CLI that automatically generates C++ code that calculates the offset of the virtual method on the C++ side.
1. I added a CLI that automatically generates C++ code.
2. I applied the generated C++ code.

The previous PR is #689 #710

Relates to #323


source/agora/cli/gen/main.d
```diff
@@ -0,0 +1,132 @@
...
+*******************************************************************************/
+
+private int main (string[] args)
+{
+    string[] methods;
+
+    foreach (member; __traits(allMembers, scpd.scp.SCPDriver.SCPDriver))
+    {
+        if ((member == "__dtor") || (member == "__xdtor") || (member == "ValidationLevel")) continue;
+        methods ~= member;
```
AndrejMitrovic: This code is too specialized for SCPDriver, and `methods` may include fields here and not just methods, because there is no check that the member is a method.

source/scpp/extra/DVTableChecks.cpp
```diff
@@ -0,0 +1,199 @@
...
+        virtual void acceptedBallotPrepared () {};
+        virtual void confirmedBallotPrepared () {};
+        virtual void acceptedCommit () {};
+        virtual void ballotDidHearFromQuorum () {};
+};
+
+
+int checkOrderOfVM ()
+{
+    auto df0 = &SCPDriverDlang::signEnvelope;
```
AndrejMitrovic: This is just comparing two C++ classes. It doesn't check the bindings. It's not what the task was about.

PR #687, opened 2020-03-20
Add NRVO tests and make SCP types serializable with the serializer

### LINKED ISSUE #691: Ensure NRVO is done through the Deserializer to avoid memory corruption
We plan to expand the usage of the serializer to be used for network communication.
Currently we are using the less-than-ideal JSON approach, which adds a lot of overhead.
However, due to concerns about efficiency and internal pointers (e.g. `std::string`), the deserializer should not perform any unwarranted move.

DoD:
- [ ] There is an utility that test if a type is moving during deserialization;
- [ ] Tests are added so that SCPEnvelope is not moved;

Geod24:
This is a requirement for switching away from Vibe.d's REST module.

This is still a WIP, I want to add some sensible data in `SCPStatement`.
I also have an idea to systematically test NRVO on struct, I just need a bit of time.

### DIFF
--- a/source/agora/test/NRVO.d
+++ b/source/agora/test/NRVO.d
@@ -1,0 +1,521 @@
+/*******************************************************************************
+
+    Test that `Serializer` performs NRVO correctly
+
+    NRVO (named return value optimization) is the act of allocating space for
+    a return value in the call stack of the caller, and passing a pointer
+    to that memory to the callee.
+    It allows to avoid copying and moving when an aggregate is returned.
+    It is particularly useful when an aggregate is returned from deep down
+    a stack call, when an aggregate is large, or if it has elaborate semantics
+    (non-default copy constructor, postblit, destructor).
+
+    Usually, to tell if a type is being copied or moved, one can simply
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
> This is still a WIP, I want to add some sensible data in SCPStatement.

The prettify PR I made has some tests you can borrow from.

source/agora/test/NRVO.d
```diff
@@ -0,0 +1,379 @@
...
+
+    // Note: Struct needs to be > 8 bytes to ensure NRVO,
+    // otherwise registers may be used
+    T value;
+    T* self;
+
+    /// Check that `self is &this`
+    public void check (string message, int line = __LINE__) const
+    {
+        if (ptr(this.self) !is ptr(this))
```
AndrejMitrovic: would `this.self !is ptr(this)` not work?

Geod24:
Fixes #691

Geod24:
I have update this to the HEAD of my current WIP, if anyone's interested in looking at where it's heading towards.

PR #688, opened 2020-03-20
Add the ability to use `vtable` to check the order of virtual methods

MichaelKim20:
I used `__vptr` on the D side to check the order of the virtual methods.
I confirmed that there is an address of the virtual methods of C++ in `__vptr` on D side.

C++
``` C++
class TestA
{
public:
    virtual ~TestA() {}
    virtual void vfunc1() = 0;
    virtual void vfunc2() = 0;
};
```

D
``` D
// abstract class, this is in the correct order.
extern (C++) class TestA
{
public:
    ~this()
    {
    }
    abstract void vfunc1();
    abstract void vfunc2();
}

extern (C++) class BonTestA : TestA
{
public:
    override void vfunc1() {}
    override void vfunc2() {}
}

uint base = 2;
auto n = new BonTestA();
// The point of the method is the same as the expected value.
assert(n.__vptr[base + 0] == &BonTestA.vfunc1);
assert(n.__vptr[base + 1] == &BonTestA.vfunc2);
```
Also, I confirmed that `&BonTestA.vfunc1` is the address of the virtual method.

Therefore, we can expect the address of the virtual method using the order declared to D.(Use`__vptr` Array)
Since then we compare this expected value with the actual value.
Of course, the actual value is `&BonTestA.vfunc1`.

If we have the same expected value and the actual value, we have demonstrated that the virtual method is in the same order.

Relates to #323

### DIFF
--- a/source/scpd/tests/VTableTest.d
+++ b/source/scpd/tests/VTableTest.d
@@ -1,0 +1,158 @@
+/*******************************************************************************
+
+    Contains checking if the virtual methods are in the same order
+    using vtable.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module scpd.tests.VTableTest;
+
+/// In case the virtual methods are in the correct order.
+unittest
+{
+    // abstract class, this is in the correct order.
+    extern (C++) class TestA
+    {
+    public:
+        ~this()
+        {
+        }
+        abstract void vfunc1();
+        abstract void vfunc2();
+    }
+
+    extern (C++) class BonTestA : TestA
+    {
+    public:
+        override void vfunc1() {}
+        override void vfunc2() {}
+    }
+
+    uint base = 2;
+    auto n = new BonTestA();
+    // The point of the method is the same as the expected value.
+    assert(n.__vptr[base + 0] == &BonTestA.vfunc1);
+    assert(n.__vptr[base + 1] == &BonTestA.vfunc2);
+}
+
+/// In case the virtual methods are in the incorrect order.
+unittest
+{
+    // abstract class, this is in the incorrect order.
+    extern (C++) class TestB
+    {
+    public:
+        ~this()
+        {
+        }
+        abstract void vfunc2();
+        abstract void vfunc1();
+    }
+
+    extern (C++) class BonTestB : TestB
+    {
+    public:
+        override void vfunc1() {}
+        override void vfunc2() {}
+    }
+
+    uint base = 2;
+    auto n = new BonTestB();
+    // The method's point differs from the expected value.
+    assert(n.__vptr[base + 0] != &BonTestB.vfunc1);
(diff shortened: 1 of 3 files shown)


MichaelKim20:
I found that there was some problem with this. I'm sorry.

Geod24:
Something in particular ?

MichaelKim20:
> Something in particular ?

I thought it was a very good way, and I was also confident.
By the way, I checked in more detail,
I could not check the order of the virtual methods of C++ at runtime.

source/scpd/tests/VTableTest.d
```diff
@@ -0,0 +1,158 @@
...
+    import std.traits;
+
+    import core.stdc.stdint;
+    import core.time;
+
+    string[] methods;
+
+    foreach (member; __traits(allMembers, SCPDriver))
+    {
+        if ((member == "__dtor") || (member == "__xdtor")) continue;
```
Geod24, line 92: Why ? dtors are also put in the vtable

MichaelKim20: I've learned some new facts.

1. Taking the address of a destructor is forbidden
    However, the address of a destructor is the second value of the `vtable`.
2.
In C++, the virtual methods were ordered in the following ways:
However, if the destructor is located in the middle, not the first, then the number is confused.
&SCPDriver::signEnvelope
I need to find a new way to get the address of the virtual method vtable.
I need more correction to solve this problem.
I am going to modify getMethodSeqSCPDriver.

MichaelKim20: https://github.com/bpfkorea/agora/pull/689

Geod24: > In C++, the virtual methods were ordered in the following ways:


Looks like something is missing after this sentence ?

> However, the address of a destructor is the second value of the vtable.

C++ generates two destructors on most compilers. See [this SO question](https://stackoverflow.com/questions/44558119/why-do-i-have-two-destructor-implementations-in-my-assembly-output).

Geod24:
By the way, this method should *not* be specific to `SCPDriver`.
It is fine to start with `SCPDriver` and progressively abstract things, but we should not depend explicitly on `SCPDriver`. Which means:
- It should work with classes that are not abstract;
- It should work with virtual methods that have been overriden;
- It should work with virtual methods that have been overriden and are `final`;
- It should not depend on the parameter types;

Geod24:
One thing you can rely on: the mangling.

MichaelKim20:
@Geod24 This PR is Closed. use PR #689

MichaelKim20:
This method was determined to solve the problem, and the PR was closed. 
And when I pushed the branch 'vtable_check' by force, I couldn't open it again.
So I restart at PR #689

source/scpd/tests/VTableTest.d
MichaelKim20: I saw two of them with the same address. The two addresses were `__dtor`.

MichaelKim20: And I find that the addresses of `__dtor` are always `__vptr[0]` and `__vptr[1]`.

PR #690, opened 2020-03-22
Re-enable LDC testing + small improvement to the docs action

Geod24:


Geod24:
Note: the end goal is to drop DMD. But we better ensure we have at least 2 compilers that can compile and run our code successfully.

.github/workflows/github_push.yaml
```diff
@@ -6,7 +6,7 @@ jobs:
     name: MacOS
     strategy:
       matrix:
-        os: [macOS-latest]
+        os: [macOS-15.04]
```
AndrejMitrovic, line 9: Why 15.04?

Geod24: That's the latest version (aka 10.15.04)

AndrejMitrovic: Oh this is Mac, not Ubuntu. My bad.

Geod24:
So are we good to go ?

Geod24:
Needed by #691

TrustHenry:
looks good to me. Debugging is also good.  However, I haven't tested it on the Windows version.

AndrejMitrovic (approved):

PR #692, opened 2020-03-23
Remove unused protobuf-d submodules

Geod24:
The idea was to use it for the TCP communication,
but in order to deal with C++ data type we'll have to take another route.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -43,9 +43,6 @@
 [submodule "d2sqlite3"]
 	path = submodules/d2sqlite3
 	url = https://github.com/biozic/d2sqlite3.git
-[submodule "protobuf-d"]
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
Is the issue the `union` or just protobuf doesn't support pointers..?

AndrejMitrovic (approved):

Geod24:
> Is the issue the `union` or just protobuf doesn't support pointers..?

It does, but:
- We don't control the deserialization (moves);
- One of the upside of having the same serialization protocol as on disk is that we can just dump to the network what's on disk;
- We don't really need all the features protobuf offer, because we still rely on the hash and so auto upgrade/downgrade of blocks is out of the question

PR #697, opened 2020-03-25
Add bindings to updateLocalQuorumSet()

AndrejMitrovic:
I was surprised to see this exists in SCP.

However, it is actually unused in stellar-core. So whether it actually works as designed I am not completely sure yet.

In theory all that's needed to hot-swap a node's quorum settings is to cancel all SCP timers and then call this function with the new quorum set.

Needed for: #240


AndrejMitrovic:
Damn it's in the wrong location..

AndrejMitrovic:
Wrong PR, it's already there but in a different module.

PR #698, opened 2020-03-25
Rework Github actions to be more reliable and less copy-pasted

Geod24:
After doing work for upstream Github actions, I realized there were many ways we could improve ours.
This is a good first step in removing duplication.
I haven't found a satisfactory way to remove the duplication between pull requests and push actions yet, but at least we only have 2 sections instead of 4 now.


.github/workflows/github_push.yaml
```diff
@@ -1,68 +1,79 @@
...
-on: [push]
+# Github worflow to test Agora on a variety of platforms
+#
+# The CI triggers on 'push' to upstream.
+#
+# Note:
+# - Try to use the native Github action syntax (${{ expression }}) when possible,
+#   as they are substituted with their value in the logs, unlike env variable.
+#   For example use `${{ github.workspace }}` over `${GITHUB_WORKSPACE}`
+name: CD pipeline
```
AndrejMitrovic, line 9: Could add a comment that the D in CD refers to the documentation being auto-deployed.

AndrejMitrovic (approved):

.github/workflows/github_push.yaml
```diff
@@ -1,68 +1,112 @@
...
+        ${DC} --version
+        dub --version
+
+    # Build and run the tests
+    - name: Build & test Agora
+      run: ./ci/run.sh
+
+    # Documentation: Note that this will push documentation to your fork, too!
+    - name: Build and deploy documentation
+      if: matrix.dc == 'dmd-latest'
```
TrustHenry, line 87: This is likely to be done twice on Mac and Linux now.
As an additional condition, Runner.os == 'Linux' will be required.

.github/workflows/github_push.yaml
```diff
@@ -1,68 +1,112 @@
...
+      fail-fast: false
       matrix:
-        os: [macOS-15.04]
-        dc: [dmd-master, ldc-master]
+        os: [ ubuntu-18.04, macOS-10.15, windows-2019 ]
+        dc: [ dmd-master, ldc-master, dmd-latest ]
+        exclude:
+          # Not supported yet
+          - os: windows-2019
+          # We only build documentation on Linux
```
TrustHenry, line 35: Is this the right place?

.github/workflows/github_push.yaml
```diff
@@ -1,68 +1,112 @@
...
+          # Just use the latest release of DMD here, don't pin the version
+          - { os: macOS-10.15,  dc: dmd-latest }
+          - { os: windows-2019, dc: dmd-latest }
```
TrustHenry, line 38: If you only test the latest version here, you might want to use `Include:`

AndrejMitrovic:
Should we phase out testing with Travis? Just let it run on master branch pushes?

.github/workflows/github_push.yaml
Geod24: It refers to "deployment" as in "Continuous deployment". Couldn't find a cool name. Suggestions welcome.

.github/workflows/github_push.yaml
Geod24: `include` does not allow to add rows in a matrix

.github/workflows/github_push.yaml
Geod24: If we want those two files to be merged, yes.
Note that `include` does not allow to add rows in a matrix, you can only `exclude` them. `include` only adds job-specific data.

.github/workflows/github_push.yaml
Geod24: As mentioned above, the `exclude` takes care of that.

Geod24:
> Should we phase out testing with Travis? Just let it run on master branch pushes?

I guess your concern is that it's too slow, right ?

AndrejMitrovic:
Yes the team mentioned CI testing was slowing down development. And also we need to maintain two different CI system configurations.

.github/workflows/github_push.yaml
AndrejMitrovic: oh I know, but mentioning what is actually deployed would be nice.

.github/workflows/github_push.yaml
TrustHenry: https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-configurations-in-a-matrix-build
I think it's used like this.

Geod24: > You cannot use `include` to add new jobs to a build matrix.

TrustHenry (approved):
It is correct to add exclude here because the purpose is to eliminate duplication.

TrustHenry:
I will merge. wait for this and check the test results of Github actions push.

Issue #694, opened 2020-03-24
Add validation checking code for pre-images being revealed


linked0:
In current implementation of managing enrollments and pre-images, there is no code checking validity of pre-images. The validator set(`ValidatorSet`) must check validity of pre-images being received from several validators.

Definition of Done:
- [x] The validator set checks incoming pre-images with enrollment data

How to implement:
* The validator set hashes a random seed of an enrollment by the difference in the height between the random seed and the incoming pre-image, and compares the hash value and one of the pre-image.
* The `Enrollment` structure must have the height of its random seed for checking the validity.

Geod24:
Fixed by https://github.com/bpfkorea/agora/pull/851

PR #703, opened 2020-03-30
Check the validity of a preimage

linked0:
When new pre-image is revealed, it must be checked if it's the right pre-image for a current pre-image stored in the validator set. This PR adds the `isInvalidPreimageReason` function checking the validity of a preimage and makes the validator set use the function in the `addPreimage` function.

Relates #694

### DIFF
--- a/source/agora/consensus/Validation.d
+++ b/source/agora/consensus/Validation.d
@@ -20,6 +20,7 @@ import agora.common.crypto.Schnorr;
 import agora.common.Hash;
 import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
+import agora.consensus.data.PreimageInfo;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSet;
 import agora.consensus.Genesis;
@@ -956,6 +957,58 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
     return isInvalidEnrollmentReason(enrollment, block_height, findUTXO) is null;
 }
 
+/*******************************************************************************
+
+    Check the validity of a new pre-image information
+
+    Pre-image infomation is considered valid if:
+        - The keys for enrollment in two pieces of pre-image information are
+            same
+        - Height for a pre-image is greater than height for a current image
+        - A current image is same as n times hashed value of a pre-image
+            (n = difference of two heights).
+
+    Params:
+        new_image = The pre-image information to check
+        prev_image = The previous pre-image information
+        validator_cycle = The defined period as a validator
+
+    Returns:
+        `null` if the pre-image is valid, otherwise a string
+        explaining the reason it is invalid.
+
+*******************************************************************************/
+
+public string isInvalidPreimageReason (const ref PreimageInfo new_image,
+    const ref PreimageInfo prev_image, uint validator_cycle) nothrow @safe
+{
+    if (new_image.enroll_key != prev_image.enroll_key)
+        return "The pre-image's enrollment key differs from its descendant";
+
+    if (new_image.height < prev_image.height)
+        return "The height of new pre-image is smaller than that of previous one";
+
+    if (new_image.height - prev_image.height >= validator_cycle)
+        return "The hashing count of two pre-images is above the validator cycle";
+
+    Hash temp_hash = new_image.hash;
+    for (uint i = 0; i < new_image.height - prev_image.height; i++)
+        temp_hash = hashFull(temp_hash);
+    if (temp_hash != prev_image.hash)
+        return "The pre-image has a invalid hash value";
+
+    return null;
+}
+
+/// Ditto but returns `bool`, only usable in unittests
+version (unittest)
+public bool isValidPreimage (const ref PreimageInfo prev_image,
+    const ref PreimageInfo new_image, uint validator_cycle) nothrow @safe
+{
+    return isInvalidPreimageReason(prev_image, new_image, validator_cycle)
+        is null;
+}
+
 /// Ditto but returns `bool`, only usable in unittests
 version (unittest)
 public bool isValid (const ref Block block, ulong prev_height,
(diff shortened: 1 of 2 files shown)


source/agora/consensus/Validation.d
```diff
@@ -956,6 +957,53 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
...
+    Pre-image infomation is considered valid if:
+        - The keys for enrollment in two pieces of pre-image information are
+            same
+        - Height for a pre-image is greater than height for a current image
+        - A current image is same as n times hashed value of a pre-image
+            (n = difference of two heights).
+
+    Params:
+        pre_image = The pre-image information
+        cur_image = The current image information
```
AndrejMitrovic: How about naming them `prev_image` and then `next_image` or `new_image`, just to make it obvious which preimage was already in the EnrollmentManager?

source/agora/consensus/Validation.d
```diff
@@ -956,6 +957,53 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
...
+    const ref PreimageInfo cur_image) nothrow @safe
+{
+    if (pre_image.enroll_key != cur_image.enroll_key)
+        return "The pre-image has a wrong enrollment key";
+
+    if (pre_image.height <= cur_image.height)
+        return "the pre-image has a invalid height";
+
+    Hash temp_hash = pre_image.hash;
+    for (uint i = 0; i < pre_image.height - cur_image.height; i++)
```
AndrejMitrovic: What happens when a byzantine node sends a `PreimageInfo` with `pre_image.height` set to `ulong.max`?

source/agora/consensus/ValidatorSet.d
```diff
@@ -526,6 +526,19 @@ public class ValidatorSet
             return false;
         }
 
+        // check the validity of new pre-image based on the stored pre-image
+        PreimageInfo cur_image;
+        if (this.getPreimage(preimage.enroll_key, cur_image) &&
+            cur_image != PreimageInfo.init)
```
AndrejMitrovic: Why would the preimage be default-initialized if `getPreimage` returns `true`? It seems to me that should never happen.

source/agora/consensus/Validation.d
linked0: > How about naming them prev_image and then next_image or new_image, just to make it obvious which preimage was already in the EnrollmentManager?

Good suggestion!

source/agora/consensus/ValidatorSet.d
linked0: > Why would the preimage be default-initialized if getPreimage returns true? It seems to me that should never happen.

It's possible there is no pre-image. The return value of `getPreimage` is about whether the operation is processed succesfully, not about whether a pre-image exists.

linked0:
Ready for review.
I applied the suggestion to the code and dealt with the problem related to the height of an preimage.

source/agora/consensus/Validation.d
```diff
@@ -956,6 +957,55 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
...
+        `null` if the pre-image is valid, otherwise a string
+        explaining the reason it is invalid.
+
+*******************************************************************************/
+
+public string isInvalidPreimageReason (const ref PreimageInfo new_image,
+    const ref PreimageInfo prev_image, uint validator_cycle) nothrow @safe
+{
+    if (new_image.enroll_key != prev_image.enroll_key)
+        return "The pre-image has a wrong enrollment key";
```
```suggestion
        return "The pre-image's enrollment key differ from it's descendant";
```

Rationale for the wording: "wrong" doesn't mean much. It doesn't say what was expected, or how that expectation wasn't met.

source/agora/consensus/Validation.d
```diff
@@ -956,6 +957,55 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
...
+
+    if (new_image.height <= prev_image.height ||
```
Geod24: Why the `<=` ? It could be just `<`, couldn't it ?

source/agora/consensus/Validation.d
```diff
@@ -956,6 +957,55 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
...
+        new_image.height - prev_image.height >= validator_cycle)
+        return "the pre-image has a invalid height";
```
Geod24: II would actually split this into two returns. One with `<` and one with the check against `validator_cycle` so you can provide a more useful error message.

source/agora/consensus/Validation.d
```diff
@@ -956,6 +957,55 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
...
+
+    Hash temp_hash = new_image.hash;
+    for (uint i = 0; i < new_image.height - prev_image.height; i++)
```
Geod24, line 995:
```suggestion
    foreach (_; prev_image.height .. new_image.height)
```

Not tested, but it *should* work, and it's much easier to see what is happening

source/agora/consensus/Validation.d
```diff
@@ -956,6 +957,55 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
...
+    Pre-image infomation is considered valid if:
+        - The keys for enrollment in two pieces of pre-image information are
+            same
+        - Height for a pre-image is greater than height for a current image
+        - A current image is same as n times hashed value of a pre-image
+            (n = difference of two heights).
+
+    Params:
+        new_image = The pre-image information to check
+        prev_image = The previous pre-image information
```
Geod24, line 973: Missing `validator_cycle` documentation

source/agora/consensus/Validation.d
```diff
@@ -956,6 +957,55 @@ public bool isValidEnrollment (const ref Enrollment enrollment,
     return isInvalidEnrollmentReason(enrollment, block_height, findUTXO) is null;
 }
 
+/*******************************************************************************
+
+    Check the validity of a pre-image information
```
Geod24: You might want to express your documentation in terms of "parent" and "descendant" / "child".
Because there's a clear hierarchy in pre-image: we successively reveal "parent" or "ancestor" of a hash.

Documenting "pre-image" as invalid is a bit confusing to me, because a pre-image is not a self-contained entity, it's defined in terms of its image. What we can test is if a `PreimageInfo` is a parent for another `PreimageInfo`.

Consider the following scenario: You have a `PreimageInfo` at height 98, and you get a `PreimageInfo` for height 90. Except the hash actually matches the pre-image for height 92. The `PreimageInfo` is not a parent for the `PreimageInfo` we have, but the hash itself is a pre-image - just at a different height.

Geod24 (changes requested):
This need a test in `agora.test`

source/agora/consensus/Validation.d
linked0: The `PreimageInfo` is a model for "pre-image" we described in the white paper. So we can use `PreimageInfo` and "pre-image" interchangingly. Because we are developing a "pre-image" in our product, we can use the word, "pre-image" as `PreimageInfo`. So the "pre-image" gets to have an actual preimage and a height. It's not the pre-image we use in Mathematics or Cryptography. We need to use an word like "pre-image" in our context.
I hope I understood what you mean exactly.

Geod24: > It's not the pre-image we use in Mathematics or Cryptography.

My point was that, since we are borrowing vocabulary from those field, we shouldn't conflate the two. Whenever talking about details, we could use `PreImageInfo`, but in the general sense we refer to them as "preimage".
But the documentation, as it stands, is good enough.

Geod24 (approved):
LGTM

source/agora/consensus/Validation.d
linked0: > My point was that, since we are borrowing vocabulary from those field, we shouldn't conflate the two. Whenever talking about details, we could use PreImageInfo, but in the general sense we refer to them as "preimage".

Thank your for the additional comment and the merge.

Issue #605, opened 2020-02-28
Make the validator set restore validators' data from the chain on startup


linked0:
In current implementation of the enrollment manager, there are a problem to solve. All the problems occurs when the database crashes and a node cannot restore the data for validators

The information of validator should be stored safely so that it can be restored any time like UTXO set.

Definition of Done:

- [x] Restore data of validator set such as enrollment information and the enrolled height from the chain into the validator set table.

How to implement:
- When a`ValidatorSet`(which is being split from `EnrollmentManager`) is created, it can enrollment information and enrolled height from the chain.
- The previously broadcast from a validator cannot restore from the chain in currently implemented code, so we must find the way to get preimages.

Geod24:
A shutdown is not going to solve the problem though. We should have a recovery strategy in place.

What's causing the problem ? If it's a database error, is it because the table is corrupted ? As long as we don't loose *critical* data (such as the random value that is the root of the preimage), we can still emit those preimage.

Shutting down will *not* be expected, and while people should have a restart mechanism in place, we should not abuse it.

AndrejMitrovic:
I suggested shutting down if the error is unrecoverable.

Now, if sqlite fails you may try to regenerate the preimage and then try to use the database again. But what guarantees do you have that you will not run into the same error as before? How many times should the node attempt to store the preimage before it gives up? If it never gives up, you risk running into an infinite loop of failed retries.

(There is an option of using error stats to trigger an alarm of sorts, in place of just shutting down. But those are unreliable too, because they use disk I/O which can fail. )

So we would need to analyze the d2sqlite bindings and see how the Exceptions are used. To find out why Exceptions can be thrown, and which errors are recoverable. In addition, the SQLite documentation needs to be consulted too.

Geod24:
> How many times should the node attempt to store the preimage before it gives up?

[There are common answers to this problem](https://en.wikipedia.org/wiki/Exponential_backoff).

But even if the node cannot save the pre-image, the error isn't so critical that it would prevent it from making progression on the blockchain, is it ?

AndrejMitrovic:
> There are common answers to this problem.

It was more of a rhetorical question. There is complexity when you write code which tries to recover from errors, and that code too could have bugs and error out.

But yeah in the general case if the error is recoverable, we should try to recover.

> But even if the node cannot save the pre-image, the error isn't so critical that it would prevent it from making progression on the blockchain, is it ?

[...]

PR #705, opened 2020-03-30
Create signature noise from a private key

linked0:
Currently, the enrollment manager creates a signature noise from random value. But, if abnormal situations occur, there is no way to restore it. So, we make the signature noise recoverable by making it from a private key.

Relates #605

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -105,18 +105,8 @@ public class EnrollmentManager
         this.key_pair.v = secretKeyToCurveScalar(key_pair.secret);
         this.key_pair.V = this.key_pair.v.toPoint();
 
-        // load signature noise
-        auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
-            "WHERE key = ?", "signature_noise");
-
-        foreach (row; results)
-        {
-            signature_noise = deserializeFull!Pair(row.peek!(ubyte[])(0));
-            break;
-        }
-
         // load enroll_key
-        results = this.db.execute("SELECT val FROM node_enroll_data " ~
+        auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
             "WHERE key = ?", "enroll_key");
 
         if (!results.empty)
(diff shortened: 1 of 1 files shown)


Geod24:
Typo: signautre

linked0:
Ready for review.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +668,23 @@ public class EnrollmentManager
...
+    /***************************************************************************
+
+        Create signature noise for enrollment data
+
+        Returns:
+            the signature noise created using the private key
+
+    ***************************************************************************/
+
+    private Pair createSingatureNoise ()
```
AndrejMitrovic: Typo:

```suggestion
    private Pair createSignatureNoise ()
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +668,23 @@ public class EnrollmentManager
...
+    {
+        Pair key_pair;
+        key_pair.v = Scalar(hashMulti(this.key_pair.v, "signature_noise"));
```
AndrejMitrovic: I think this might be problematic. This means the node will always use the same starting R for every new enrollment.

@Geod24 can you offer some advice here? I think you suggested hashing the private key with the current block height, which I think would work alright.

AndrejMitrovic (dismissed):

source/agora/consensus/EnrollmentManager.d
linked0: > I think you suggested hashing the private key with the current block height, which I think would work alright.

I agree with you. To restore the signature noise with the height you suggested, I should add a height variable where a enrollment process occurs, into the `Enrollment` structure.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +668,26 @@ public class EnrollmentManager
...
+
+        Returns:
+            the signature noise created using the private key
+
+    ***************************************************************************/
+
+    private Pair createSignatureNoise (ulong height) nothrow
+    {
+        Pair key_pair;
+        key_pair.v = Scalar(hashMulti(this.key_pair.v, height));
```
Geod24: Whenever we derive something from something else, we usually need three component:
- The secret we derive from (here, private key);
- A constant value which state the "purpose" of the derivation;
- A non-constant value to actually make the derivation;

The first element goes without saying. Here, the third element is `height`, but it can also be a combination of other values. However we're missing the second element. Why is it important ? Because we might do the derivation in a few unrelated places, using potentially the same secret, and a data-type that might overlap. For example we could have a derivation involving the secret and a number of round. If that happens, an attacker could use the result of a derivation in the signature noise to predict the values generated for the second derivation.
Hence something like this:
```suggestion
        key_pair.v = Scalar(hashMulti(this.key_pair.v, "consensus.signature.noise", height));
```

Where we make sure to never reuse `consensus.signature.noise` would be better.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +668,26 @@ public class EnrollmentManager
         return this.cycle_preimages[height];
     }
 
+    /***************************************************************************
+
+        Create signature noise for enrollment data
+
+        Parameters:
```
```suggestion
        Params:
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +668,26 @@ public class EnrollmentManager
         return this.cycle_preimages[height];
     }
 
+    /***************************************************************************
+
+        Create signature noise for enrollment data
+
```
Geod24, line 679: Please add a description explaining what it is, why we need it, why we derive it, how, etc...

Geod24 (commented):
Nice! Just need documentation and the minor change on the derivation, and we'll be good to go.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +668,27 @@ public class EnrollmentManager
...
+        Params:
+            height = the height used for a salt value
+
+        Returns:
+            the signature noise created using the private key
+
+    ***************************************************************************/
+
+    private Pair createSignatureNoise (ulong height) nothrow
```
```suggestion
    private Pair createSignatureNoise (ulong height) nothrow @safe
```

Might be able to mark it `@nogc` too, although not sure

Geod24: Hu ?

linked0: > Hu ?

I'm not ready for review yet. :)

linked0: Marking the function `@safe` emits error like 

> Error: @safe function agora.consensus.EnrollmentManager.EnrollmentManager.createSignatureNoise cannot call @system constructor agora.common.crypto.ECC.Scalar.this

Do you mean I should modify the code in order to mark the function `@safe`?

Geod24 (approved):
Merge at will after rebasing on `v0.x.x`

PR #706, opened 2020-03-30
Add the function restoring validators' information

linked0:
In abnormal situations, the validator set must restore the information from blocks which have Enrollment objects and enrolled heights in their headers. This PR adds the `restoreValidators` function for that.

### DIFF
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -20,6 +20,7 @@ import agora.common.crypto.Key;
 import agora.common.crypto.Schnorr;
 import agora.common.Hash;
 import agora.common.Serializer;
+import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.PreimageInfo;
 import agora.consensus.data.UTXOSet;
@@ -211,7 +212,7 @@ public class ValidatorSet
     ***************************************************************************/
 
     public bool updateEnrolledHeight (const ref Hash enroll_hash,
-        const size_t block_height) @safe
+        const size_t block_height) @safe nothrow
     {
         try
         {
@@ -560,6 +561,41 @@ public class ValidatorSet
 
         return true;
     }
+
+    /***************************************************************************
+
+        Restore validators' information from block
+
+        Params:
+            last_height = the latest block height
+            block = the block to update the validator set with
+            finder = the delegate to find UTXOs with
+
+    ***************************************************************************/
+
+    public void restoreValidators (ulong last_height, const ref Block block,
+        scope UTXOFinder finder) @safe nothrow
+    {
+        assert(last_height >= block.header.height);
+        if (last_height - block.header.height < ValidatorCycle)
+        {
+            foreach (const ref enroll; block.header.enrollments)
+            {
+                if (!this.hasEnrollment(enroll.utxo_key))
+                {
+                    if (!this.add(block.header.height, finder, enroll))
+                    {
+                        assert(0);
+                    }
+                }
+                if (!this.updateEnrolledHeight(enroll.utxo_key,
+                        block.header.height))
+                {
+                        assert(0);
+                }
+            }
+        }
+    }
 }
 
 version (unittest)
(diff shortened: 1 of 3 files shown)


source/agora/consensus/ValidatorSet.d
```diff
@@ -562,6 +563,46 @@ public class ValidatorSet
...
+    {
+        foreach (const ref block; blocks)
+        {
+            if (block.header.enrollments.length > 0)
+            {
+                foreach (const ref enroll; block.header.enrollments)
+                {
+                    try
+                    {
+                        this.add(block.header.height, finder, enroll);
```
AndrejMitrovic: You are wrapping these functions in a try/catch even though they never throw. And you are **ignoring** the return values which are booleans.

And now you introduced yet another `bool` function `restoreValidators`.

Do you understand the problem with bool return functions now?

linked0: > Do you understand the problem with bool return functions now?

Yes, I do. Sorry for that. I will deal with it.

linked0:
Ready for review.

source/agora/consensus/ValidatorSet.d
```diff
@@ -546,6 +547,44 @@ public class ValidatorSet
...
+            if (block.header.enrollments.length > 0)
+            {
+                foreach (const ref enroll; block.header.enrollments)
+                {
+                    if (!this.hasEnrollment(enroll.utxo_key))
+                    {
+                        if (!this.add(block.header.height, finder, enroll))
+                            assert(0);
+                    }
+                    try
```
Geod24: Why the `try` if `updateEnrolledHeight` is `nothrow` ?

source/agora/consensus/ValidatorSet.d
```diff
@@ -546,6 +547,44 @@ public class ValidatorSet
...
+        Restore validators' information from blocks
+
+        Params:
+            blocks = blocks to get validators' information
+            finder = the delegate to find UTXOs with
+
+    ***************************************************************************/
+
+    public void restoreValidators (const(Block)[] blocks,
+        scope UTXOFinder finder) @safe nothrow
```
Geod24, line 577: This function is not called from anywhere, how do you plan to have it used ?
Also passing a full array seems quite wasteful. A range might be more appropriate, but it's hard to say without seeing the usage. Do you have a WIP for it ? I understand if you want to keep it separate from this PR.

source/agora/consensus/ValidatorSet.d
linked0: > Why the try if updateEnrolledHeight is nothrow ?

It's my mistake. It will correct it.

source/agora/consensus/ValidatorSet.d
linked0: > This function is not called from anywhere, how do you plan to have it used ?

It was for smaller PR. But, It would be good to add the code calling this function. I will add more commits for that.

> Also passing a full array seems quite wasteful. A range might be more appropriate,

Please explain more about how I can pass a range. In current plan, this function will be called from Ledger class through the EnrollmentManger.

Geod24: In order to receive a range:
```D
    public void restoreValidators (R) (R blocks, scope UTXOFinder finder) @safe nothrow
        if (isInputRange!R)
```

If you want to ensure that only blocks are passed, you can do:
```D
        if (isInputRange!R && is(ElementType!R : const(Block)))
```

linked0: > public void restoreValidators (R) (R blocks, scope UTXOFinder finder) @safe nothrow
>         if (isInputRange!R)

Thank you for the information. I have read the doc you let me know yesterday and it really helped me understand the concept of the Range. But, I have decided to use `IBlockStorage` instead of `InputRange` because only `ValidatorSet` knows how many blocks are needed in order to restore validators' information in the headers like as follows.
```
public void restoreValidators (ulong last_height, IBlockStorage storage,
        scope UTXOFinder finder) @safe
    {
        Block block;
        foreach (height; last_height - ValidatorCycle .. last_height)
```

Geod24: I'm not sure I follow. Ranges are an higher order data-structure, so they should work for that use case too.
For example, if you want to repeat an action X times:

```D
iota(X).map!(_ => my_action_here);
```

So in pseudo code:
```D
iota(X).map!(o => storage.getHeight(last_height - ValidatorCycle + o)).map!(block => block.header.enrollments)
    .filter!(enroll => !this.hasEnrollment(enroll.utxo_key)).each!(enroll => this.add(block.header.height, finder, enroll));
```

In reality it's a tiny bit more complicated because you need to flatten the `enrollments` range and keep `block` around for the last `each` call, but there's hardly a concept you can't represent with `Range`. Let me know when you have implemented the `IBlockStorage` approach and I'll give it another review.

linked0: Sorry for my incomplete example. I'll request another review with `IBlockStorage` and get your code above understood.

linked0:
Ready for review.
I applied all the suggestions to the code.

source/agora/consensus/ValidatorSet.d
MichaelKim20: I think it'll help if you watch this next time.
Ledger.getBlocksFrom (size_t start_height)

linked0: > I think it'll help if you watch this next time.
Ledger.getBlocksFrom (size_t start_height)

Thank you for the information.

MichaelKim20 (approved):

linked0:
I'm modifying the code now. I will get things done by 4.

linked0:
Ready for review.
I have tried to apply the Range, but I failed to apply that to the code because there are many things to know. After studying more about that, I will apply it in similar tasks. Aside from that, I have one more thing to do about checking validity of the preimage for this sprint. So if this PR has no problems, Please approve it.

source/agora/node/Ledger.d
```diff
@@ -99,6 +100,18 @@ public class Ledger
                 this.updateUTXOSet(block);
             }
         }
+
+        // restore validator set from lastest blocks
+        foreach (i; 0 .. ValidatorSet.ValidatorCycle)
+        {
+            if (i >= this.last_block.header.height)
```
AndrejMitrovic: Shouldn't this be `>`?

If there are enrollment in the genesis block they will be skipped. Because the genesis block height is 0, and in the first loop `i` will be 0 so the `>=` condition is true and then it breaks.

linked0: > Shouldn't this be >?

Yes, that should. I will fix the bug by changing the start value of the `foreach` into 1. Thanks.

Geod24 (approved):
There's still a few things needed, but let's go with this as a first step.

PR #708, opened 2020-03-31
Print logs of worker threads when unittests fail & fix interleaved output

AndrejMitrovic:
For example if a test like `assert(ledger.acceptTransaction(tx));` in `agora.node.Ledger` fails then we currently can't tell why it fails. All of the diagnostics are logged to the logger.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -137,6 +137,7 @@ private UnitTestResult customModuleUnitTester ()
         {
             writefln("Module tests failed: %s", mod.name);
             writeln(ex);
+            CircularAppender().printConsole();  // print logs of the work thread
         }
     }
 
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -123,10 +123,10 @@ private UnitTestResult customModuleUnitTester ()
 
     shared size_t executed;
     shared size_t passed;
-    foreach (mod; parallel(mod_tests))
+
+    void runTest (ModTest mod)
     {
         atomicOp!"+="(executed, 1);
-
         try
         {
             //writefln("Unittesting %s..", mod.name);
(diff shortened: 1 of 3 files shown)


source/agora/test/Base.d
```diff
@@ -135,28 +135,21 @@ private UnitTestResult customModuleUnitTester ()
         }
         catch (Throwable ex)
         {
-            writefln("Module tests failed: %s", mod.name);
-            writeln(ex);
+            synchronized
```
Geod24: Can we use an explicit mutex here ?

source/agora/test/Base.d
```diff
@@ -135,28 +135,21 @@ private UnitTestResult customModuleUnitTester ()
...
-            atomicOp!"+="(passed, 1);
-        }
-        catch (Throwable ex)
-        {
-            writefln("Module tests failed: %s", mod.name);
-            writeln(ex);
-        }
-    }
-
+        runTest(mod);
```
Geod24, line 154: 👍

source/agora/test/Base.d
AndrejMitrovic: I guess so. Is there any reason why we should avoid using the one that's autogenerated for a code block?

Geod24: Explicit lifetime control ?
I have an innate doubt towards any D-generated magic. If you feel strongly about it, I can drop it. But it's one of those things I wish would go away.

AndrejMitrovic: No it's alright to be explicit about it. I've updated the PR.

I've also added another commit (hopefully not controversial).

AndrejMitrovic: Actually I remember now I had bugs in my code where I assumed multiple `synchronized {}` blocks used the same lock, but in fact each statement uses its own lock. So you're right, we should be explicit.

Geod24 (approved):
LGTM, merge at will

AndrejMitrovic:
Github CI seems to fail:

> Failed to execute program (No such file or directory)

It's really odd..

Geod24 (approved):

PR #689, opened 2020-03-21
Check the offset of virtual method in `vtable`

MichaelKim20:
I checked the offset change in `vtable` according to the virtual method declaration order of C++ side and D side
1. I added the ability to check the offset of the virtual method.
 
The next PR is #710 #711 

Relates to #323

PR #710, opened 2020-03-31
Adds the ability to check if the offset in the `vtable` on C++ side and D side is the same.

MichaelKim20:
It checks the offset of the virtual method for any class.

The previous PR is #689 

The next PR is #711 

Relates to #323


Geod24:
So do we still need this one ? I see that #711 includes those commits. And as I mentioned before, I don't think those steps are necessary.

MichaelKim20:
> So do we still need this one ? I see that #711 includes those commits. And as I mentioned before, I don't think those steps are necessary.

Okay then I'd better close this PR.

Issue #560, opened 2020-02-19
[C++] Document usage of C++ types, limitations, and patterns


Geod24:
Currently we don't have a good documentation about what can, should, and shouldn't be done with `extern(C++)` types. Given there are a lot of technicalities involved, it would be good to have a rather extensive documentation.

Such a documentation would introduce knowledge of constructs in D that are safe to use, as well as C++-specific knowledge, for example copy constructors, default constructors, constructors deletion, move constructors, non-representable types, etc...

That documentation is also very likely to be specific to the version of C++ we're using (there are some subtle differences between C++11, C++14 and C++17, and we're currently using the later).

Another note: The D documentation for C++ interfacing is wildly outdated...

AndrejMitrovic:
Hmm yeah. Looking at it now again, we return vectors to the C++ side - for example in `combineCandidates`. But when we return it by value we don't do a deep copy (to the C heap), the vectors still contain references to GC allocated memory (which then goes out of scope).

Geod24:
Return by value is actually fine, it uses NRVO so dtor / copy ctors are not involved

Geod24:
Re unwanted moves, on the C++ side we have https://reviews.llvm.org/D43322 (flag `-Wreturn-std-move`).

PR #711, opened 2020-03-31
Add utilities which verify the offset of the virtual method on the D & C++ sides

MichaelKim20:
Add utilities which verify the offset of the virtual method on the D & C++ sides

1. I added the ability to check the offset of the virtual method.
2. It checks the offset of the virtual method for any class.
3. I added a CLI that automatically generates C++ code.

Relates to #323


source/scpp/extra/DVMChecks.cpp
```diff
@@ -0,0 +1,152 @@
+/*******************************************************************************
```
Geod24: This file is generated by `cli/gen/main.d` right ?

source/scpd/tests/VTableTest.d
```diff
@@ -0,0 +1,286 @@
...
+                            val_cpp = getVMOffset%1$s(`%2$s`);
+                            offset_cpp[`%2$s`] = val_cpp;
+                            values[val_cpp] = true;
+                        }
+                    }
+                }.format(`%1$s`, member));
+            }
+
+            //  If the destructor is a virtual method,
+            //  Finds two consecutive empty indexes.
```
Geod24: Aren't they guaranteed to be at the beginning ?

Geod24:
Side note: When comparing pointers, use `is` or `!is`

source/scpp/extra/DVMChecks.cpp
MichaelKim20: I recorded that this file was created automatically. And I described how to make modifications.

source/scpd/tests/VTableTest.d
MichaelKim20: There is no guarantee of where the order will be.

MichaelKim20: Then, do you think it is good to remove the contents related to the destructor?

MichaelKim20: I think there is no difference depending on the destructor when comparing the offset of the virtual method on C++ and D.
What do you think?

Geod24: > There is no guarantee of where the order will be.

> The order of the virtual function pointers in a virtual table is the order of declaration of the corresponding member functions in the class. If an implicitly-declared copy assignment operator, move assignment operator, or destructor is virtual, it is treated as if it were declared at the end of the class, in that order. (Implicitly-declared assignment operators may be virtual if a base class declares a virtual assignment operator taking a reference to a derived class type.)

https://itanium-cxx-abi.github.io/cxx-abi/abi.html#vtable-components

source/scpp/extra/DVMChecks.cpp
Geod24: Great. So I don't think we should check it in the repository.

MukeunKim: /Users/mukeunkim/Work/WorkSpace_BPFK/agora-node/agora/source/scpp/extra/DVMChecks_Generated.cpp

MichaelKim20: This is a video capture of running agora-gen.
[agora-gen.mov.zip](https://github.com/bpfkorea/agora/files/4418536/agora-gen.mov.zip)

MichaelKim20:
@Geod24 Ready for review.

source/scpd/tests/VTableTypes.d
```diff
@@ -0,0 +1,35 @@
...
+{
+public:
+    ~this()
+    {
+    }
+    abstract void vfunc1();
+    abstract void vfunc2();
+}
+
+public const string[] VTableCheckClasses =
```
Geod24:
```suggestion
public immutable VTableCheckClasses =
```

source/scpd/tests/VTableTest.d
```diff
@@ -16,6 +16,13 @@ module scpd.tests.VTableTest;
 
 version (Windows) {} else:
 
+import std.format;
+
+version (unittest)
+{
+    extern (C++) long getVMOffsetTestA(const char* name);
```
Geod24:
```suggestion
    extern(C++) long getVMOffsetTestA (const char* name);
```

Geod24 (commented):
I still don't understand the point of the first few commits.
It starts with all the checks being hardcoded, and gradually hardcode less and less. But the point of the issue is to have the generator, so those intermediate steps, while valuable for you in your search for a solution, shouldn't be in the final version.

source/scpp/extra/DVMChecks.cpp
MichaelKim20: log : https://gist.github.com/MukeunKim/fe9c1b848463b834260e6e5fa2453dab

source/agora/cli/gen/main.d
```diff
@@ -0,0 +1,200 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.cli.gen.main;
```
Geod24: `gen` is a bit too generic of a name. Something like `check-vtable` would be more accurate.

source/agora/cli/gen/main.d
```diff
@@ -0,0 +1,200 @@
...
+            file.writeln(`};`);
+            file.writeln(``);
+
+
+            file.writeln(`///  Checks the offset of virtual methods inside the class %1$s.`);
+            file.writeln(`int checkVMOffset%1$s ()`);
+            file.writeln(`{`);
+
+            method_idx = 1;
+            static foreach (member; __traits(allMembers, %1$s))
```
Geod24: How are overloads handled ?

source/scpp/extra/DVMChecks.cpp
```diff
@@ -0,0 +1,320 @@
...
+        return ((long)reinterpret_cast<void *&>(pf)) / sizeof(long);
+    }
+    else
+    {
+        return -1;
+    }
+}
+
+///  Uses to test SCPDriver.
+class SCPDriverForTest
```
Geod24: We shouldn't need to declare this extra type on the C++ side. Can't we just get the offset on the D side ?

Geod24 (commented):
Comments:
- First few commits are steps in development and shouldn't be here;
- Last commit shouldn't be in either: let's not checkout the generated file. Instead I'll add a check to the CI.
- Overloads are completely ignored, so if upstream introduces an overload, this will break;
- The method of declaring the class on the C++ side is quite brittle too. Why can't we get that offset from the D side and pass it to `checkVMOffsetSCPDriver` ?

source/agora/cli/gen/main.d
MichaelKim20: I don't know why this place needs overloads.
Why do you think it is necessary?
The parameter and return values are not used here.
I knew it was decided not to take care of the override.

source/scpp/extra/DVMChecks.cpp
MichaelKim20: > We shouldn't need to declare this extra type on the C++ side. Can't we just get the offset on the D side ?

I wonder if this type of data is possible.
map<string, int>

source/agora/cli/gen/main.d
MichaelKim20: There is also `getVMOffset...`, what about this?

MichaelKim20:
``` D
/// It checks the offset of virtual on C++ and D for any class.
unittest
{
    //  Offset of  C++ side
    long[string] offset_cpp;

    //  Offset of D side
    long[string] offset_d;
    long val_d;

    static foreach (e; VTableCheckClasses)
    {
        offset_cpp.clear();
        offset_d.clear();
        val_d = 0;

        mixin(
        q{
            static foreach (member; __traits(allMembers, %1$s))
            {
                mixin(
                q{
                    static if (__traits(isVirtualMethod, %1$s.%2$s) && (`%2$s` != `__xdtor`))
                    {
                        static if (`%2$s` == `__dtor`)
                        {
                            // for D side
                            val_d += 2;
                        }
                        else
                        {
                            // for D side
                            offset_d[`%2$s`] = val_d++;

                            // for C++ side
                            offset_cpp[`%2$s`] = getVMOffset%1$s(`%2$s`);
                        }
                    }
                }.format(`%1$s`, member));
            }

            assert(offset_d == offset_cpp, format(`%%s : %%s`, offset_d, offset_cpp));

        }.format(e, `%2$s`));
    }
}
```

MichaelKim20:
@Geod24 
I think I know a little about what you want.
I removed most of the C++ side code, leaving only part of it.


I'm done with the modifications.
The automatically generated code is here.
https://gist.github.com/MukeunKim/796b82581a76b16ce9e3929844b76406

source/agora/cli/checkvtable/main.d
```diff
@@ -0,0 +1,193 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.checkvtable.gen.main;
```
Geod24: That should be `agora.cli.checkvtable.main`

source/agora/cli/checkvtable/main.d
```diff
@@ -0,0 +1,193 @@
+/*******************************************************************************
+
+    Entry point for the agora-gen
```
Geod24: This also. And since it's a single-file program, please add a description of what it does and why.

MichaelKim20:
@Geod24 
Ready for review

I left only the code below on C++.

``` C++
#include "DVMChecks.h"

int checkVMOffset (const char * classname, const char* offsets)
{
    return 0;
}
```
I added one feature to the agora-checkvtable

The automatically generated code is here.
https://gist.github.com/MukeunKim/53f0e9d9183c3dde92df4b99d74e22bd

MichaelKim20:
In C++
``` C++
class TestA
{
public:
    virtual void foo1(int a) {};
    virtual void foo2(int a) {};
};
```

In D
``` D

extern(C++) abstract class TestA
{
    abstract void foo2 (int a);
    abstract void foo1 (int a);
}

extern(C++) class BonTestA : TestA
{
    public override void foo2 (int a) { writeln("Hello World 2"); }
    public override void foo1 (int a) { writeln("Hello World 1"); }
}

pragma(msg, BonTestA.foo1.mangleof);
pragma(msg, BonTestA.foo2.mangleof);

mixin("extern(C) void ", BonTestA.foo1.mangleof, "(void* _this, int);");
mixin("alias callMe1 = ", BonTestA.foo1.mangleof, ";");

mixin("extern(C) void ", BonTestA.foo2.mangleof, "(void* _this, int);");
mixin("alias callMe2 = ", BonTestA.foo2.mangleof, ";");

void main ()
{
    writefln("&callMe1 =       %s", &callMe1);
    writefln("&callMe2 =       %s", &callMe2);
    writefln("&BonTestA.foo1 = %s", &BonTestA.foo1);
    writefln("&BonTestA.foo2 = %s", &BonTestA.foo2);

    auto m = new BonTestA();

    writefln("m.__vptr[0] = %s", m.__vptr[0]);
    writefln("m.__vptr[1] = %s", m.__vptr[1]);
}
```
```
&callMe1 =       10143648C
&callMe2 =       101436464
&BonTestA.foo1 = 10143648C
&BonTestA.foo2 = 101436464
m.__vptr[0] = 101436464
m.__vptr[1] = 10143648C
```

Both callMe1 and callMe2 have the addresses of virtual methods on the D side.

source/scpd/tests/VTableTest.d
```diff
@@ -0,0 +1,75 @@
...
+    auto n = new BonTestA();
+    // The point of the method is the same as the expected value.
+    assert(n.__vptr[2] is &BonTestA.vfunc1);
+    assert(n.__vptr[3] is &BonTestA.vfunc2);
+}
+
+/// In case the virtual methods are in the incorrect order.
+unittest
+{
+    // abstract class, this is in the incorrect order.
```
AndrejMitrovic: What do you mean, incorrect order? The vtable will be based on the order of the methods as defined in the **base** class, not the inherited class. It doesn't matter what the order is in the derived class.

In your `BonTestB` derived class you can reorder the methods and your tests will still pass. But these are the sort of tests that belong in the compiler's test-suite, not Agora.

source/scpp/extra/DVMTestTypes.cpp
```diff
@@ -0,0 +1,20 @@
+/*******************************************************************************
```
AndrejMitrovic: This seems to belong in the next commit, it's not used in this commit.

source/scpp/extra/DVMTestTypes.h
```diff
@@ -0,0 +1,30 @@
+/*******************************************************************************
```
AndrejMitrovic: It belongs in the next commit.

source/scpd/tests/VTableTest.d
```diff
@@ -73,3 +78,44 @@ unittest
...
 }
+
+/// It checks the offset of virtual on C++ and D for any class.
+unittest
+{
+    import std.format;
+    import std.string;
+
+    long[string] offset_d;
+    long val_d;
```
AndrejMitrovic: What is "val_d"? Can you add a comment or rename it to something more obvious?

source/scpd/tests/VTableTest.d
```diff
@@ -73,3 +78,44 @@ unittest
...
+        mixin(
+        q{
+            static foreach (member; __traits(allMembers, %1$s))
+            {
+                mixin(
+                q{
+                    static if (__traits(isVirtualMethod, %1$s.%2$s) && (`%2$s` != `__xdtor`))
+                    {
+                        static if (`%2$s` == `__dtor`)
+                            val_d += 2;
```
AndrejMitrovic: Why += 2?

source/scpp/extra/DVMChecks.cpp
```diff
@@ -0,0 +1,19 @@
...
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+#include "DVMChecks.h"
+
+int checkVMOffset (const char* classname, const char* offsets)
+{
+    return 0;
```
AndrejMitrovic: What is the purpose of this function? It just returns 0.

source/agora/cli/checkvtable/main.d
```diff
@@ -0,0 +1,222 @@
+/*******************************************************************************
+
+    Entry point for the agora-checkvtable
+        It generate C++ code for chacking the offset of the vtable.
+        The path to the file it creates is `source/scpp/extra/DVMChecks.cpp`.
+
+        The use of this is as follows.
+        agora-checkvtable source/scpp/extra/DVMChecks.cpp
```
AndrejMitrovic:
```suggestion
    It generates C++ code for checking the offset of the vtable.
    The path to the file it creates is `source/scpp/extra/DVMChecks.cpp`.

    The use of this is as follows.
    agora-checkvtable source/scpp/extra/DVMChecks.cpp
```

AndrejMitrovic: It should not generate temporary files in the source tree. It should generate them in a temporary directory, verify the offsets work, and then exit with 0. We should not commit these autogenerated files into the repository. 

Because any time any of the SCP classes change, we will have to remember to commit new files into the repository. We want an **automated** testing process.

source/agora/cli/checkvtable/main.d
```diff
@@ -0,0 +1,222 @@
...
+/// expects a main() function and invokes it after unittesting.
+version (unittest) void main () { } else:
+
+/*******************************************************************************
+
+    Application entry point
+
+*******************************************************************************/
+
+int main (string[] args)
```
AndrejMitrovic: The entire body of this function should be another function which either returns a `string` or writes to a `buffer`. That way, you can unittest the function and see that it generates the right code.

AndrejMitrovic: And then in `main()` you would just write that buffer to a file.

source/agora/cli/checkvtable/main.d
```diff
@@ -0,0 +1,222 @@
...
+
+    file.close();
+
+    writefln("Source code was generated with the file name %s.", filename);
+
+    return 0;
+}
+
+/// Writes the header of the source file
+void writeHeader(File file)
```
AndrejMitrovic: It should write to a buffer or an output sink (just like the Formatter and the Serializer)

AndrejMitrovic: Also I have the same comment for all the other functions which write to a file.

source/agora/cli/checkvtable/main.d
```diff
@@ -0,0 +1,222 @@
...
+    }
+}
+`
+    );
+}
+
+/// Writes the contents of the method
+void writeMethodContents(File file, bool first, string entry, string method)
+{
+    if (first)
```
AndrejMitrovic: You could simplify this by removing the need to use `first`. So instead of generating this:

```C++
if (strcmp(name, "foo") == 0)
{
    auto pf = class::foo;
    return  ((long)reinterpret_cast<void *&>(pf)) / sizeof(long);
}
else if (strcmp(name, "bar") == 0)
{
    auto pf = class::bar;
    return  ((long)reinterpret_cast<void *&>(pf)) / sizeof(long);
}
```

You could remove the `first` parameter and generate this:


```C++
if (strcmp(name, "foo") == 0)
{
    auto pf = class::foo;
    return ((long)reinterpret_cast<void *&>(pf)) / sizeof(long);
}

if (strcmp(name, "bar") == 0)
{
    auto pf = class::bar;
    return ((long)reinterpret_cast<void *&>(pf)) / sizeof(long);
}
```

The end result is the same (short-circuiting evaluation) as using `if/else if`, because there is a `return` statement in each block.

AndrejMitrovic (changes requested):

Geod24:
Ping on the review and the commits: It looks like everything should be squashed to one commit ?

MichaelKim20:
> Ping on the review and the commits: It looks like everything should be squashed to one commit ?

@Geod24 Ready for review

Geod24:
Has the rest of the review been addressed ? It doesn't seem so.

source/agora/cli/checkvtable/main.d
MichaelKim20: @AndrejMitrovic That's a good idea. Thank you.

source/scpd/tests/VTableTest.d
MichaelKim20: Yes, I agree with you.
This is not to solve the issue.
But this will allow us to understand the issue correctly.
In conclusion, I don't think it's necessary. I'll remove it.

AndrejMitrovic: > But this will allow us to understand the issue correctly.

We could write some documentation about it, if it's unclear. I think we have an issue for that. Here: #560

source/scpd/tests/VTableTest.d
MichaelKim20: The destructor always occupies two pointers on the vtable.
So I added two.

source/scpp/extra/DVMChecks.cpp
MichaelKim20: @Geod24 gave me this comment.
https://github.com/bpfkorea/agora/pull/711#pullrequestreview-386915944

I just want to keep that pattern at unittest.
And I thought that A would be automatically generated in CI and the test would proceed.
I couldn't implement overloaded methods with C++ source.
So I had no choice but to make this decision.

source/scpd/tests/VTableTest.d
AndrejMitrovic: Ah yes, I think on the D side there are two destructors. One is `__dtor` and one is `__xdtor`.

MichaelKim20: The offsets of TestA are 
TestA._dtor : 0
TestA.vfunc1 : 1
TestA.vfunc2 : 2
TestA.__xdtor : 3
If `__dtor` and `__xdtor` come out one after another, we can add 1 more, but since that's not the case, we have to add 2.

source/agora/cli/checkvtable/main.d
MichaelKim20: It's just an example.
The fIle `source/scpp/extra/DVMChecks.cpp` is not an automatically generated file.

MichaelKim20: For automation, the source must be generated before the pre-builder.
I think the processing of this will have to make another CI.

AndrejMitrovic: Ok so is your plan to add this CI step in this PR, or in another PR?

AndrejMitrovic:
> @Geod24: Last commit shouldn't be in either: let's not checkout the generated file. Instead I'll add a check to the CI.

So then this PR here needs some support work from @Geod24?

@MukeunKim is there anything else blocking you?

source/scpd/tests/VTableTest.d
MichaelKim20:
``` C++
class A
{
public:
    virtual ~A() {};
    virtual void vfunc1(){};
    virtual void vfunc2(){};
};

class B
{
public:
    virtual void vfunc1(){};
    virtual void vfunc2(){};
    virtual ~B() {};
};

int CPPTest1 ()
{
    auto pAf1 = &A::vfunc1;
    printf("A.vfunc1 : %d\n", (long)reinterpret_cast<void *&>(pAf1) / sizeof(long));
    auto pAf2 = &A::vfunc2;
    printf("A.vfunc2 : %d\n", (long)reinterpret_cast<void *&>(pAf2) / sizeof(long));
    auto pBf1 = &B::vfunc1;
    printf("B.vfunc1 : %d\n", (long)reinterpret_cast<void *&>(pBf1) / sizeof(long));
    auto pBf2 = &B::vfunc2;
    printf("B.vfunc2 : %d\n", (long)reinterpret_cast<void *&>(pBf2) / sizeof(long));
}
```
Result is 
A.vfunc1 : 2
A.vfunc2 : 3
B.vfunc1 : 0
B.vfunc2 : 1

In A, 0 and 1 are destructors, and in B, 2 and 3 are destructors.

source/agora/cli/checkvtable/main.d
MichaelKim20: I didn't have any specific plans for CI, but with your help, I thought a lot about it.

I will remove the file below from the current PR.
source/scpd/tests/VTableTest.d
source/scpp/extra/DMMCheks.cpp

I think CI needs the following functions.
1. C++ Source Generator (D): I will use CLI checkvtable
2. Prebuilder (C++): I will use source/scpp/build.d.
3. A small project for small checks.
This will include source/scpd/tests/VTableTest.d.

And I will make CI in another PR.

source/agora/cli/checkvtable/main.d
MichaelKim20: I tried this.
There is an error message that the computer time cannot be accessed. 
I could not change to another string array instead of VTableCheckClasses.

MukeunKim:
> > @Geod24: Last commit shouldn't be in either: let's not checkout the generated file. Instead I'll add a check to the CI.
> 
> So then this PR here needs some support work from @Geod24?
> 
> @MukeunKim is there anything else blocking you?

If we've confirmed some of the current structure, I'll have to make CI next time.
I think the current structure needs to be changed a little bit when we create CI.
Maybe we don't need a CLI, maybe we need a small project.
Therefore, we should first listen to @Geod24's plan and proceed with the development accordingly.

MichaelKim20:
> > @Geod24: Last commit shouldn't be in either: let's not checkout the generated file. Instead I'll add a check to the CI.
> 
> So then this PR here needs some support work from @Geod24?
> 
> @MukeunKim is there anything else blocking you?

@AndrejMitrovic 
If we've confirmed some of the current structure, I'll have to make CI next time.
I think the current structure needs to be changed a little bit when we create CI.
Maybe we don't need a CLI, maybe we need a small project.
Therefore, we should first listen to @Geod24's plan and proceed with the development accordingly.

MichaelKim20:
@AndrejMitrovic I think I applied everything I could to your proposal.
I couldn't find a way for the next one.
Is there a good way?
https://github.com/bpfkorea/agora/pull/711#discussion_r407955740

source/agora/cli/checkvtable/main.d
AndrejMitrovic: > There is an error message that the computer time cannot be accessed.

Could you show the full error message and also the code which is causing the issue?

source/agora/cli/checkvtable/main.d
MichaelKim20: I implemented this as follows.
A new feature has been added to the CLI agora-checkvtable.
Previously, there was only the ability to create C++source.
This can be done with the command:
agora-checkvtable generation outfilename
And I added the function to check the offset of the virtual method.
This can be done with the command:
agora-checkvtable check
And I added a test code that can be used by CI
ci/check_vtable_test.d

source/agora/cli/checkvtable/main.d
MichaelKim20: I fixed the source as below, and when I tested it, an error occurred.

``` D
/// This is agora-checkvtable sub-function for generating C++ source code
int generate (string[] args, ref string[] outputs)
{
    if (args.length < 3)
    {
        writeln("usage: agora-gen command [output source file]");
        return 1;
    }

    auto filename = args[2];
    writecode(VTableCheckClasses, outputs);
    File file = File(filename, "w+");

    foreach(ref line; outputs)
        file.writeln(line);

    file.close();

    writefln("Source code was generated with the file name %s.", filename);
    return 0;
}

/// Writes C++ source code to outputs
void writecode (immutable(string[]) types, ref string[] outputs)
{
    writeHeader(outputs);

    static foreach (e; types)
    {
        mixin(
        q{
            writeMethodHeader(outputs, `%1$s`, `getVMOffset%1$s`);
            static foreach (idx, member; __traits(allMembers, %1$s))
            {
                mixin(
                q{
                    static if (__traits(isVirtualMethod, %1$s.%2$s) && (`%2$s` != `__dtor`) && (`%2$s` != `__xdtor`))
                    {
                        writeMethodContents(outputs, `%1$s`, `%2$s`);
                    }
                }.format(`%1$s`, member));
            }

            writeMethodFooter(outputs);
            writeCheckFunction(outputs, `%1$s`);
            outputs ~= ``;

        }.format(e, `%2$s`));
    }

    outputs ~= `int checkVMOffset (const char* classname, const char* offsets)`;
    outputs ~= `{`;

    static foreach (e; types)
    {
        mixin(
        q{
            outputs ~= `    if (strcmp(classname, "%1$s") == 0)`;
            outputs ~= `    {`;
            outputs ~= `        return checkVMOffset%1$s(offsets);`;
            outputs ~= `    }`;
        }.format(e));
    }

    outputs ~= `    return -1;`;
    outputs ~= `}`;
}
```

source/agora/cli/checkvtable/generate.d(55,24): Error: variable types cannot be read at compile time
source/agora/cli/checkvtable/generate.d(81,24): Error: variable types cannot be read at compile time

dub.json
```diff
@@ -119,6 +124,29 @@
...
+                "source/agora/node/main.d"
+            ]
+        },
+        {
+            "name": "checkvtable",
+            "targetName": "agora-checkvtable",
+            "mainSourceFile": "source/agora/cli/checkvtable/main.d",
+            "sourceFiles-posix": [
+                "source/scpp/build/DSizeChecks.o",
+                "source/scpp/build/DLayoutChecks.o",
```
Geod24: Is it needed ?

dub.json
```diff
@@ -106,11 +107,15 @@
             "excludedSourceFiles": ["source/agora/cli/*/main.d"],
             "sourceFiles-posix": [
                 "source/scpp/build/DSizeChecks.o",
-                "source/scpp/build/DLayoutChecks.o"
+                "source/scpp/build/DLayoutChecks.o",
+                "source/scpp/build/DVMTestTypes.o",
+                "source/scpp/build/DVMChecks.o"
```
Geod24: Is it needed ?

dub.json
```diff
@@ -106,11 +107,15 @@
...
             ],
             "sourceFiles-windows": [
                 "source/scpp/build/DSizeChecks.obj",
-                "source/scpp/build/DLayoutChecks.obj"
+                "source/scpp/build/DLayoutChecks.obj",
+                "source/scpp/build/DVMTestTypes.o",
+                "source/scpp/build/DVMChecks.o"
```
Geod24: That's obviously wrong, so probably not needed

dub.json
```diff
@@ -119,6 +124,29 @@
...
+                "source/scpp/build/DSizeChecks.o",
+                "source/scpp/build/DLayoutChecks.o",
+                "source/scpp/build/DVMTestTypes.o",
+                "source/scpp/build/DVMChecks.o"
+            ],
+            "sourceFiles-windows": [
+                "source/scpp/build/DSizeChecks.obj",
+                "source/scpp/build/DLayoutChecks.obj",
+                "source/scpp/build/DVMTestTypes.o",
+                "source/scpp/build/DVMChecks.o"
```
Geod24: Not needed

source/agora/cli/checkvtable/main.d
```diff
@@ -0,0 +1,58 @@
...
+module agora.cli.checkvtable.main;
+
+import agora.cli.checkvtable.generate;
+import agora.cli.checkvtable.check;
+
+import std.stdio;
+
+/// Workaround for issue likely related to dub #225,
+/// expects a main() function and invokes it after unittesting.
+version (unittest) void main () { } else:
```
Geod24: This code should never be compiled with `-unittest` so this is not needed.

ci/check_vtable_test.d
```diff
@@ -0,0 +1,54 @@
...
+immutable BuildBinary = [ "dub", "build", "-c", "checkvtable", ];
+immutable GenerateCode = [ "build/agora-checkvtable", "generate", "source/scpp/extra/DVMChecks.cpp"];
+immutable CheckVTable = [ "build/agora-checkvtable", "check"];
+
+private int main (string[] args)
+{
+    // If the user pass `nobuild` as first argument, skip build,
+    // Build `agora-checkvtable`
+    if (args.length < 2 || args[1] != "nobuild")
+        runCmd(BuildBinary);
```
Geod24: Let's remove this

dub.json
```diff
@@ -119,6 +122,27 @@
             "mainSourceFile": "source/agora/cli/multi/main.d",
             "excludedSourceFiles": [
                 "source/agora/cli/client/*",
+                "source/agora/cli/checkvtable/*",
+                "source/agora/node/main.d"
+            ]
+        },
+        {
+            "name": "checkvtable",
```
Geod24: As discussed this needs to be split in 2 configurations

source/agora/cli/checkvtable/check.d
```diff
@@ -0,0 +1,74 @@
...
+
+module agora.cli.checkvtable.check;
+
+import agora.cli.checkvtable.types;
+
+import std.format;
+import std.stdio;
+import std.string;
+
+version (Windows)
```
Geod24: Add a message here

MichaelKim20: I fixed it as follows.
```
version (Windows) { pragma(msg, "This does not support Windows."); } else:
```

source/agora/cli/checkvtable/main.d
MichaelKim20: I solved this by using CTFE.

ci/check_vtable_test.d
MichaelKim20: Okay

source/agora/cli/checkvtable/main.d
MichaelKim20: Yes

dub.json
MichaelKim20: I fixed it

MichaelKim20:
@Geod24 Ready for review.

source/agora/cli/checkvtable/generate.d
```diff
@@ -0,0 +1,318 @@
...
+{
+    if (args.length < 2)
+    {
+        writeln("usage: agora-checkvtable-gen output-file-name");
+        return 1;
+    }
+    string[] outputs;
+    auto filename = args[1];
+    writecode(outputs);
+    File file = File(filename, "w+");
```
AndrejMitrovic: This can be just `"w"` I think?

source/agora/cli/checkvtable/generate.d
```diff
@@ -0,0 +1,318 @@
...
+
+module agora.cli.checkvtable.generate;
+
+import agora.cli.checkvtable.types;
+
+import std.file;
+import std.format;
+import std.stdio;
+
+version (Windows) { pragma(msg, "This does not support Windows."); } else:
```
AndrejMitrovic, line 22: What was wrong with checking the vtable on Windows?

source/agora/cli/checkvtable/types.d
```diff
@@ -0,0 +1,34 @@
...
+
+*******************************************************************************/
+
+module agora.cli.checkvtable.types;
+
+public import scpd.scp.SCPDriver;
+
+import std.meta;
+
+extern (C++) public class TestA
```
AndrejMitrovic: We don't need this class anymore right? It was only used during development. We only care about the SCPDriver class.

AndrejMitrovic (commented):
Just two small comments.

source/agora/cli/checkvtable/types.d
MichaelKim20: > We don't need this class anymore right? It was only used during development. We only care about the SCPDriver class.

I needed a function that could handle several things using 'foreach statement'. I needed more than one class to test this.

source/agora/cli/checkvtable/generate.d
MichaelKim20: > What was wrong with checking the vtable on Windows?

I tested 'MS Visual C++' on the following site.
https://rextester.com/l/cpp_online_compiler_visual

``` C++
#include <iostream>

class A
{
public:
    virtual ~A() {};
    virtual void vfunc1(){};
    virtual void vfunc2(){};
};

int main ()
{
    auto pAf1 = &A::vfunc1;
    std::cout << pAf1;
    
    return 0;
}

```
It worked abnormally in MS Visual C++.

source/agora/cli/checkvtable/types.d
MichaelKim20: I removed TestA. In the third unittest(source/agora/cli/checkvtable/generate.d), I tested one or more classes.

source/agora/cli/checkvtable/generate.d
MichaelKim20: Good catch.

MichaelKim20:
Ping @AndrejMitrovic @Geod24

source/agora/cli/checkvtable/generate.d
MichaelKim20: If the offset of vtable's virtual methods is proven to be the same in `gcc`, then of course in Windows, we no longer have to make unnecessary efforts.
The test will also be run only on CI.

AndrejMitrovic:
This PR adds the vtable check generator, but it never uses it? Where is the call to `./ci/check_vtable_test.d`?

MichaelKim20:
> This PR adds the vtable check generator, but it never uses it? Where is the call to `./ci/check_vtable_test.d`?

@AndrejMitrovic I added `./ci/check_vtable_test.d` to `.circleci/config.yml`.

AndrejMitrovic:
I tested it locally and it works.

AndrejMitrovic (approved):

.circleci/config.yml
```diff
@@ -41,6 +41,13 @@ jobs:
             sudo make install
             sudo ldconfig # Refresh cache
             popd
+      - run:
+          name: Check virtual method offset
+          command: |
+            source $HOME/dlang/ldc-1.20.1/activate
+            ci/check_vtable_test.d
```
AndrejMitrovic, line 48: By the way, we could have added this check into `run.sh` instead? Is there a reason why it was moved to CircleCI?

MichaelKim20: https://github.com/bpfkorea/agora/pull/711#pullrequestreview-386915944
I just didn't want to have a big impact on Agora's development environment.

.circleci/config.yml
```diff
@@ -41,6 +41,13 @@ jobs:
             sudo make install
             sudo ldconfig # Refresh cache
             popd
+      - run:
+          name: Check virtual method offset
+          command: |
+            source $HOME/dlang/ldc-1.20.1/activate
```
Geod24:
```suggestion
            source $HOME/dlang/ldc-1.21.0/activate
```

.circleci/config.yml
```diff
@@ -41,6 +41,13 @@ jobs:
...
+            ci/check_vtable_test.d
+            # Work around druntime setting the permissions to 600..
```
Geod24:
```suggestion
```

Geod24 (approved):
Just a minor fix, otherwise LGTM

Geod24:
You also need to rebase on `v0.x.x` it seems

.circleci/config.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           command: |
             mkdir -p $HOME/dlang && wget https://dlang.org/install.sh -O $HOME/dlang/install.sh
             chmod +x $HOME/dlang/install.sh
-            $HOME/dlang/install.sh install ldc-1.20.1
```
Geod24: https://github.com/bpfkorea/agora/commit/c79a5103c995c969af0a6c0d792d2799fd8497d3

PR #712, opened 2020-03-31
Add ability to provide a custom Genesis block

### LINKED ISSUE #709: Add ability to define a custom Genesis block for tests
The `GenesisBlock` is currently used by unittests and integration tests everywhere.

This is a problem because the unittests are using the private key from `getGenesisKeyPair` in order to sign transactions. Before CoinNet is released, the `GenesisBlock` will be changed to use the actual public key of the foundation. The tests will fail, unless `getGenesisKeyPair` returns the private key of the foundation nodes. And obviously we don't want to release our private key to the world.

We also currently cannot use signature validation routines which use enrollments, because blocks after the genesis block don't have a set of validators with which to validate those blocks. Block #1 (after genesis) would need to be validated with an enrolled validator, but a validator cannot enroll in the Genesis block (it's hardcoded right now).

What needs to be done:
- [ ] In `version(unittest)` the genesis block should equal to whatever the tests define it to be. This makes unittests flexible as they can define their own genesis block and not have to depend on a hardcoded one
- [ ] Handle system integration tests. They do not use `version(unittest)`. We might have to use some kind of form: `version (SystemIntegration ) { genesis = ... } else version (unittest) { genesis = ... } else { genesis = <real-genesis> }`

Related issue about freeze in GenesisBlock: https://github.com/bpfkorea/agora/issues/241

I'm assigning this issue now because it's starting to block #240 which requires enrollments and #208 which also requires enrollments.

AndrejMitrovic:
This introduces a getter function for the genesis block and a config option allowing the genesis block to be overriden.

In unittest mode we use a module constructor to initialize the Genesis block to the one most unittests expect. Otherwise the block is initialized to the CoinNet genesis block (which is currently just a placeholder!).

If a unittest wishes to test with a custom genesis block they can do this:

```D
unittest
{
    immutable old_gen = GenesisBlock;
    scope (exit) GenesisBlock = old_gen;

    setGenesisBlock(new immutable(Block){ ... });
    // ...
}
```

This will also work with multithreaded unittests.

The Node config option `genesis_block` allows overriding the genesis block when the Node is built in non-unittest mode. This is necessary for system integration tests, and will also be useful for TestNet or any other private network the user wishes to use.

Fixes https://github.com/bpfkorea/agora/issues/709
Unblocks: #507, #684


AndrejMitrovic:
Forgot a comma in my json. Of course.

source/agora/common/Serializer.d
```diff
@@ -762,7 +762,8 @@ unittest
     }
 
     testSymmetry!OptionalHash();
-    testSymmetry(OptionalHash(42, &GenesisBlock, "Baguettes are good"));
+    auto block = GenesisBlock;
+    testSymmetry(OptionalHash(42, &block, "Baguettes are good"));
```
Geod24: <3

AndrejMitrovic:
I had a realization. There's a better way to provide a genesis block in integration tests. Add a `--genesis` flag taking the serialized hexdump of the genesis block to use. This ensures we use an unmodified binary for integration testing, and can also allow easy testnet configuration (and also any private test/coinnet)

AndrejMitrovic:
@Geod24 I've reworked this PR. Previously it used `version (IntegrationTests)` blocks, but as mentioned in https://github.com/bpfkorea/agora/pull/712#issuecomment-606961852 I found a better solution. 

I've updated the PR description with an overview.

AndrejMitrovic:
With regards to the pointer dereference and performance I think we can add something like this later:

```D
version (deployed) 
{ 
    alias GenesisBlock = CoinNetGenesisBlock; 
} 
else 
{
    ref immutable(Block) GenesisBlock() { ... }
}
```

AndrejMitrovic:
There seems to be a flaky test in https://github.com/bpfkorea/agora/blob/c7811c1949f7125e2dfe18eb783389b0af27fd1a/source/agora/test/GossipProtocol.d#L73 which fails frequently on Travis. I also reproduced it once locally, but not again.

So it seems there is some timing issue somewhere. Those are always fun..

source/agora/consensus/Genesis.d
```diff
@@ -20,57 +20,110 @@ import agora.common.crypto.Key;
...
+    return gen_block.txs[0];
+}
+
+/// initialize the genesis block
+static this ()
+{
+    version (unittest)
+        setGenesisBlock(UnitTestGenesisBlock);
+    else
+        setGenesisBlock(CoinNetGenesisBlock);
```
Geod24: Does `= &CoinNetGenesisBlock;` in the declaration work ? `&CoinNetGenesisBlock` has a constant value

source/agora/consensus/Genesis.d
```diff
@@ -80,27 +133,33 @@ private immutable ubyte[] GenesisAddressUbyte =
...
+    testSymmetry(UnitTestGenesisBlock);
+}
 
 version (unittest)
 {
     /***************************************************************************
 
-        Get the key-pair which can spend the UTXO in the genesis transaction.
+    Get the key-pair which can spend the UTXO in the unittests
+    genesis transaction.
```
Geod24: ??

Geod24 (commented):
Not a fan of the first commit. It's already in `version (unittest)` so I think that's enough. Couple of comments. It's not great to loose the ability to refer to Genesis as a manifest constant, either (before it was just a data read, not it involves an indirection and a TLS lookup). But if we need it...

source/agora/utils/Test.d
```diff
@@ -152,7 +152,8 @@ unittest
 *******************************************************************************/
 
 public Transaction[] makeChainedTransactions (KeyPair key_pair,
-    Transaction[] prev_txs, size_t block_count, ulong spend_amount = 40_000_000)
+    Transaction[] prev_txs, size_t block_count,
+    ulong spend_amount = 40_000_000, in Transaction gen_tx = Transaction.init)
```
AndrejMitrovic: Missing documentation.

source/agora/common/Config.d
```diff
@@ -95,6 +95,10 @@ public struct NodeConfig
     static assert(!hasUnsharedAliasing!(typeof(this)),
         "Type must be shareable accross threads");
 
+    /// If set, a hexdump serialized representation of the genesis block to use
+    /// in place of the built-in genesis block as used by CoinNet
```
AndrejMitrovic:
```suggestion
    /// in place of the built-in genesis block as defined by CoinNet
```

source/agora/common/Serializer.d
```diff
@@ -762,7 +762,8 @@ unittest
     }
 
     testSymmetry!OptionalHash();
-    testSymmetry(OptionalHash(42, &GenesisBlock, "Baguettes are good"));
+    auto block = GenesisBlock;
```
AndrejMitrovic: Add a small comment why this change was necessary.

source/agora/consensus/Genesis.d
AndrejMitrovic: Right!!

source/agora/consensus/Genesis.d
AndrejMitrovic: This `+unittests` wording change should be moved to a previous commit.

There will be no key-pair for the coinnet transactions, because only the public key will be visible and not the private key.

Or is something else not clear?

Geod24: Sorry, my incredulity was towards the broken indentation ;)

AndrejMitrovic:
Resolved conflicts and fixed all comments.

source/agora/utils/Test.d
```diff
@@ -148,11 +148,15 @@ unittest
...
+        gen_tx = the genesis transaction to refer to for the first set of
+                 transactions. If none set, the one returned by
+                 GenesisTransaction() is used.
 
 *******************************************************************************/
 
 public Transaction[] makeChainedTransactions (KeyPair key_pair,
-    Transaction[] prev_txs, size_t block_count, ulong spend_amount = 40_000_000)
+    Transaction[] prev_txs, size_t block_count,
+    ulong spend_amount = 40_000_000, in Transaction gen_tx = Transaction.init)
```
Geod24, line 159: That's when you need an `Optional` type

source/agora/utils/Test.d
```diff
@@ -189,9 +193,15 @@ public Transaction[] makeChainedTransactions (KeyPair key_pair,
     {
         Input input;
         if (prev_txs.length == 0)  // refering to genesis tx's outputs
-            input = Input(hashFull(GenesisTransaction), idx.to!uint);
+        {
+            const tx_hash = hashFull(gen_tx == Transaction.init
+                ? GenesisTransaction : gen_tx);
```
Geod24: Actually instead of `Transaction.init` as input, why not `GenesisTransaction` ?

source/agora/node/Node.d
```diff
@@ -104,9 +106,25 @@ public class Node : API
...
+        if (config.node.genesis_block.length > 0)
+        {
+            import std.array;
+            import std.conv;
+
+            // hex => bin
+            auto block_bytes = config.node.genesis_block.chunks(2).map!(
+                twoDigits => twoDigits.parse!ubyte(16)).array();
+            this.genesis_block = block_bytes.deserializeFull!(immutable(Block));
+            setGenesisBlock(this.genesis_block);
```
Geod24, line 125: The block needs to be validated. What if the merkle root is invalid ?

source/agora/common/Config.d
```diff
@@ -96,6 +96,10 @@ public struct NodeConfig
     static assert(!hasUnsharedAliasing!(typeof(this)),
         "Type must be shareable accross threads");
 
+    /// If set, a hexdump serialized representation of the genesis block to use
+    /// in place of the built-in genesis block as defined by CoinNet
+    public string genesis_block;
```
Geod24, line 101: No default ?

Geod24 (changes requested):
Needs validation of the block (besides that of deserializeFull), and could use a default to avoid branching.

AndrejMitrovic:
> It's not great to lose the ability to refer to Genesis as a manifest constant, either (before it was just a data read, not it involves an indirection and a TLS lookup). But if we need it...

Well I see several options:

A) Getter method like in this PR (most flexible, but it can be error-prone)
B) `version` blocks, where in production we use the CoinNet Genesis but in system integration tests we use another version block for which the private key is known (to be able to write new tests with). That would mean system integration tests do not use the production agora binary to test with, but a specially built `version (SystemIntegrationTests)` binary (also used for unittests). This is what the PR initially looked like btw. But I didn't like it that we built a special binary of agora when testing it with system integration tests. It felt "wrong"? :)
C) We delay solving this problem for some time in the future and temporarily introduce a set of hard-coded Enrollments in the Genesis block - as we need them for testing soon.

Can you think of any other ideas?

Geod24:
My previous comment was made by a dude that forgot our previous conversation. After a good night of sleep, I'm fine with this PR concept wise.

source/agora/consensus/Genesis.d
```diff
@@ -111,14 +165,77 @@ version (unittest)
...
 
+/// The genesis block as defined by CoinNet
+private immutable Block CoinNetGenesisBlock =
+{
+    header:
+    {
+        prev_block:  Hash.init,
+        height:      0,
+        merkle_root: CoinNetGenesisMerkleRoot,
+    },
```
TrustHenry, line 183: This comment may not be related to this PR.
Now we need to think about the validators who will make the next block from Genesis.

AndrejMitrovic: That's why this PR was made, so the tests can define who the validators are so we can define various testing scenarios.

Or did you wonder about something else?

TrustHenry: There are no validators in the Genesis Block at this time.
Then there is no node to validator or nominator the next block.
I thought we should define a validator in Genesis Block.

source/agora/node/Node.d
AndrejMitrovic: Should I add the validation here, or better yet into the `setGenesisBlock` routine?

source/agora/common/Config.d
AndrejMitrovic: If there's a default here then it will always override the initialization that's set in `Genesis.d`:

```
/// Contains a pointer to the genesis block.
version (unittest)
    private immutable(Block)* gen_block = &UnitTestGenesisBlock;
else
    private immutable(Block)* gen_block = &CoinNetGenesisBlock;
```

On the other hand, if we only provide here in the Config struct then we could remove all the definitions in `Genesis.d`. But I'm not sure if that's the right approach. 🤔

source/agora/node/Node.d
AndrejMitrovic: oh lol we don't have validation for the genesis block:

```
// special case for the genesis block
if (block.header.height == 0)
    return block == GenesisBlock ?
        null : "Block: Height 0 but not Genesis block";
```

That's bad.

source/agora/common/Config.d
TrustHenry: I think I can do new blockchain(fork?) easily if I just change the config file.

AndrejMitrovic:
I resolved the comments. The validation is not added because it would fail due to #747, but I added a comment for it.

source/agora/common/Config.d
AndrejMitrovic: Forking refers to splitting the network from an origin point. If networks differ in their genesis block then they do not fork, because they don't have a common ancestor (the genesis block)

AndrejMitrovic:
@Geod24 it's green. I'll add Genesis block validation some time soon, unless someone else takes care of it.

Geod24 (approved):

PR #714, opened 2020-04-01
Emit only error logs when tests fail

AndrejMitrovic:
Small annoyance of the logging I've added in https://github.com/bpfkorea/agora/pull/708


Geod24 (approved):
LGTM, merge at will.

PR #717, opened 2020-04-01
Minor modifications for MSVC mangling

TrustHenry:
This problem was found in the WindowsX64 MSVC test.

@Geod24 description
POSIX the type isn’t mangled
static and __gshared are redundant. 
Because static means “TLS global” when in an aggregate, while __gshared means “non-TLS global”
So they are both global, and __gshared wins over static (because TLS is the default)


Geod24 (approved):
LGTM, merge at will

PR #721, opened 2020-04-02
Make a constructor of Scalar trusted

Geod24:
We know the call is safe because the C library under the hood accept fixed-size values.
Two static assert were added both to verify this and to make it clearer to the reader.


linked0 (approved):
LGTM

Issue #650, opened 2020-03-11
Re-enable BlockStorageChecksum test


Geod24:
https://github.com/bpfkorea/agora/blob/0120d6832a6112338d8588480400beb240ea60be/tests/unit/BlockStorageChecksum.d#L40-L41

TrustHenry:
I think this is applied.

PR #731, opened 2020-04-06
Correct the Github actions CI Test to return the error level correctly

TrustHenry:
A build test error occurred in the window cmd environment, but the exit code is not returned.
Returns only the last run result.

### DIFF
--- a/ci/run.bat
+++ b/ci/run.bat
@@ -1,4 +1,8 @@
 call dub test --skip-registry=all --compiler=%DC%
+::if %errorlevel% neq 0 exit /b %errorlevel%
 call rdmd --compiler=%DC% ./tests/runner.d --compiler=%DC% -cov
+::if %errorlevel% neq 0 exit /b %errorlevel%
 call dub build --skip-registry=all --compiler=%DC%
+if %errorlevel% neq 0 exit /b %errorlevel%
 call dub build -c client --skip-registry=all --compiler=%DC%
+if %errorlevel% neq 0 exit /b %errorlevel%


ci/run.bat
```diff
@@ -1,4 +1,8 @@
 call dub test --skip-registry=all --compiler=%DC%
+if not %errorlevel% 0 exit /b %errorlevel%
```
Geod24: Shouldn't it be `if %errorlevel% neq 0 exit /b %errorlevel%` ?

TrustHenry: Maybe that's more accurate.

Geod24 (approved):
LGTM

Geod24:
@AndrejMitrovic and me were discussing whether `call` was needed or not on the `.bat` files. I gave it a shot.
Turns out, without `call`, cmd will work as an `execve`, that is, it will never execute the commands following the `.bat`. So we very much need `call` on those `.bat`.

Geod24:
@TrustHenry : Can any of those be enabled, or do every single line fail ?

TrustHenry:
https://github.com/bpfkorea/agora/issues/739 No errors will occur except for this issue.
I only activated build test and cli build.

AndrejMitrovic:
@TrustHenry you could add a `version (Windows) { } else:` in the `\tests\unit\BlockStorageChecksum.d` module and keep all of those `if %errorlevel% neq 0 exit /b %errorlevel%` checks.

That way we continue having proper Windows build tests and we can worry about fixing the linker issue later.

Geod24:
That test is broken anyway, so merging this as is, and let's revisit when convenient: https://github.com/bpfkorea/agora/issues/650

TrustHenry:
The annotated part leaves issues with unittest link error and Re-BlockStorageChecksum test error.
Related issues are #650, #739

PR #736, opened 2020-04-07
Remove SCP's XDR files

AndrejMitrovic:
We can just look at the C++ headers, they're not hard to follow and they show any additional generated code we may have to look for.


Geod24:
I suppose this is because you want to remove some stuff from the header files ? I originally kept them as they make comparing with upstream easier.

AndrejMitrovic:
Ah if it helps with upstream diff, then we could keep them. I thought they were here by mistake.

Issue #702, opened 2020-03-28
Use our own hashing in `getQSet`  instead of relying on Stellar's


Geod24:
Currently we have this code:
https://github.com/bpfkorea/agora/blob/b20a58d39bc3ae3d9993fa9a69763bb85ea1a62b/source/agora/consensus/protocol/Nominator.d#L117-L119

This is used mostly here  (fetching it):
https://github.com/bpfkorea/agora/blob/e35e09a740e335f1d0695d7ddaa2ca0e9c4db966/source/agora/consensus/protocol/Nominator.d#L403-L409
(BTW, nice typo)

However, we don't need to use ByteSliceHasher and the likes, we can just hash things using our own `hash` function. This would avoid the underlying vector being copied also.

AndrejMitrovic:
Can we? StellarHash is a 32-byte type, and our own hash is 64-byte.

AndrejMitrovic:
> Can we? StellarHash is a 32-byte type, and our own hash is 64-byte.

The answer is yes we can, but we need to change the Hash definition in SCPP:

```diff
diff --git a/source/scpp/src/xdr/Stellar-types.h b/source/scpp/src/xdr/Stellar-types.h
index 28ecc823..5d989d3f 100644
--- a/source/scpp/src/xdr/Stellar-types.h
+++ b/source/scpp/src/xdr/Stellar-types.h
@@ -9,7 +9,7 @@

 namespace stellar {

-using Hash = xdr::opaque_array<32>;
+using Hash = xdr::opaque_array<64>;
 using uint256 = xdr::opaque_array<32>;
 using uint32 = std::uint32_t;
 using int32 = std::int32_t;
```

I need to do this as part of https://github.com/bpfkorea/agora/pull/621, so I'll give this a go.

AndrejMitrovic:
Hmm there's another problem. SCP sets its own quorum set hash:

https://github.com/bpfkorea/agora/blob/05bfe591aec457544bfe9fc8b799f592d4d6b303/source/scpp/src/scp/LocalNode.cpp#L27

and https://github.com/bpfkorea/agora/blob/05bfe591aec457544bfe9fc8b799f592d4d6b303/source/scpp/src/scp/LocalNode.cpp#L49

The hash passed to `Nominator.getQSet` is the one generated by SCP (which uses `xdr_to_opaque`). So what `Nominator` does has to match what `LocalNode` does.

So there's two problems here:
[...]

PR #737, opened 2020-04-07
Port SCP's hashes into 64-bytes

AndrejMitrovic:
This is the first part of resolving https://github.com/bpfkorea/agora/issues/702

### DIFF
--- a/source/scpp/src/crypto/SHA.cpp
+++ b/source/scpp/src/crypto/SHA.cpp
@@ -10,76 +10,76 @@
 namespace stellar
 {
 
-// Plain SHA256
-uint256
-sha256(ByteSlice const& bin)
+// Plain SHA512
+uint512
+sha512(ByteSlice const& bin)
 {
-    uint256 out;
-    if (crypto_hash_sha256(out.data(), bin.data(), bin.size()) != 0)
+    uint512 out;
+    if (crypto_hash_sha512(out.data(), bin.data(), bin.size()) != 0)
     {
-        throw std::runtime_error("error from crypto_hash_sha256");
+        throw std::runtime_error("error from crypto_hash_sha512");
     }
     return out;
 }
 
-class SHA256Impl : public SHA256, NonCopyable
+class SHA512Impl : public SHA512, NonCopyable
 {
-    crypto_hash_sha256_state mState;
+    crypto_hash_sha512_state mState;
     bool mFinished;
 
   public:
-    SHA256Impl();
+    SHA512Impl();
     void reset() override;
     void add(ByteSlice const& bin) override;
-    uint256 finish() override;
+    uint512 finish() override;
 };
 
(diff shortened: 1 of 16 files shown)


AndrejMitrovic:
StellarHashFmt needs to be removed in SCPPrettyPrinter.

AndrejMitrovic:
> StellarHashFmt needs to be removed in SCPPrettyPrinter.

Done.

AndrejMitrovic:
Ah I think the tests will fail now, the hashes probably changed..

AndrejMitrovic:
Hm nope, I was wrong. :)

source/scpp/src/scp/SCPDriver.cpp
```diff
@@ -42,26 +42,27 @@ static const uint32 hash_K = 3;
...
     extra(h.get());
-    uint256 t = h->finish();
-    uint64 res = 0;
-    for (size_t i = 0; i < sizeof(res); i++)
-    {
-        res = (res << 8) | t[i];
-    }
+    uint512 t = h->finish();
+
+    // WARNING: Do not replace this back with SCP's original implementation.
```
AndrejMitrovic: Let's revert this change.

source/scpp/src/xdr/Stellar-types.h
```diff
@@ -9,8 +9,9 @@
 
 namespace stellar {
 
-using Hash = xdr::opaque_array<32>;
+using Hash = xdr::opaque_array<64>;
 using uint256 = xdr::opaque_array<32>;
+using uint512 = xdr::opaque_array<64>;
```
Geod24, line 16: Can you add a comment above those types, explaining that they were modified on purpose ?

source/scpp/src/util/HashOfHash.cpp
```diff
@@ -3,7 +3,6 @@
 
 namespace std
 {
-
```
Geod24: Unrelated

source/scpp/src/crypto/SHA.cpp
```diff
@@ -10,76 +10,76 @@
 namespace stellar
 {
 
-// Plain SHA256
-uint256
-sha256(ByteSlice const& bin)
+// Plain SHA512
```
Geod24, line 13: So I guess you want to do the migration in two steps ?

Geod24 (commented):
Few comments, looks nice

source/scpp/src/crypto/SHA.cpp
AndrejMitrovic: I just realized this could have been all part of the same PR. But I'll leave it as is then, and fix up the comments.

AndrejMitrovic:
Fixed up, green.

Geod24 (approved):

PR #740, opened 2020-04-07
Add @safe to quorum conversion functions

AndrejMitrovic:
When QuorumConfig was moved out, I forgot to move these additional routines.


TrustHenry (approved):

AndrejMitrovic:
Aha.. I can't seem to do this without breaking the system integration tests. Because then the C++ runtime is pulled in. argh

AndrejMitrovic:
I removed the first commit then. I'll just mark them @safe.

Geod24:
So much for that

PR #741, opened 2020-04-07
Use our own hashing routines instead of SCPs

### LINKED ISSUE #702: Use our own hashing in `getQSet`  instead of relying on Stellar's

AndrejMitrovic:
Requires:
- ~https://github.com/bpfkorea/agora/pull/737~
- ~https://github.com/bpfkorea/agora/pull/740~

Fixes #702


AndrejMitrovic:
Note: this is only for `SCPQuorumSet`. I haven't changed the routines for hashing the `Value`.

source/scpp/src/scp/LocalNode.cpp
```diff
@@ -24,14 +24,14 @@ LocalNode::LocalNode(NodeID const& nodeID, bool isValidator,
     : mNodeID(nodeID), mIsValidator(isValidator), mQSet(qSet), mSCP(scp)
 {
     normalizeQSet(mQSet);
-    mQSetHash = sha512(xdr::xdr_to_opaque(mQSet));
+    mQSetHash = getHashOf(mQSet);
```
AndrejMitrovic, line 27: This commit should replace **all** uses of sha512, not just the ones for qset.

source/agora/common/SCPHash.d
```diff
@@ -0,0 +1,60 @@
...
+        the 64-byte hash
+
+*******************************************************************************/
+
+public uint512 getHashOf (ref const(SCPQuorumSet) qset)
+{
+    return uint512(hashFull(qset));
+}
+
+/// ditto
```
TrustHenry: It would be nice if D is uppercase.

AndrejMitrovic: Okay! It doesn't make a difference to the compiler, but sure.

AndrejMitrovic: Ah for consistency, sure.

AndrejMitrovic:
I'll fix the linking issue in a bit.

AndrejMitrovic:
I get linker errors with 1.20.1 but not 1.20.0. Sigh.

AndrejMitrovic:
Fixed the linker issue.

Geod24 (approved):

PR #746, opened 2020-04-09
Fix an annotation for Fullnode.d

TrustHenry:
Modify the meaning of the feature implementation and other annotations.


linked0 (approved):
Right correction!

Issue #735, opened 2020-04-07
Change the way we store Enrollments


Geod24:
Currently we add `Enrollment` to the `EnrollmentManager` when they are received but before they are validated:
https://github.com/bpfkorea/agora/blob/05bfe591aec457544bfe9fc8b799f592d4d6b303/source/agora/node/Node.d#L507-L518

Then consensus is made on them. This has a few issues:
- It is impossible to replace an enrollment, lest the user change their key (so there's a 2 week waiting period);
- This table is never cleaned up;
- We mix the data we use before and after consensus, which is error-prone.
- `addValidatedBlock` does not add Enrollments, only update their height, which mean catch-up is broken.

It would be best to have those two data set apart, and only have active enrollments in the EnrollmentManager. The separation should probably be similar to transaction pool / UTXO set.

Definition of Done:
- [x] Make the way to sotre Enrollments like the way similar to to transaction pool / UTXO set.
- [x] Clean up the table when a enrollment is expired.

linked0:
Among the issues to be resovled, some issues as follows can be split into another issue.
- It is impossible to replace an enrollment, lest the user change their key (so there's a 2 week waiting period);
- This table is never cleaned up;

Those can be handled after chaning the way we store the Enrollments. So I will create new issue named like "Add the feature to change or delete enrollments stored".

linked0:
After the PR #748 , there will be more PRs for tasks as follows.

- [x] Refactor the relations between classes related to the enrollment process and other classes that use the classes(including the change of the name and function modifier of member functions.
- [x] Refine the API endpoints related to the enrollment process, for example, `getPreimage`.
- [ ] Remove table in the `EnrollmentManager`.
- [x] Refine test code in the `test.EnrollmentManager` package.

Geod24:
@linked0 : I think we can reasonably consider this closed. Could you open issues for any remaining tasks so they can be addressed in the next sprint ?

linked0:
> @linked0 : I think we can reasonably consider this closed. Could you open issues for any remaining tasks so they can be addressed in the next sprint ?

There is one remaining task which is to remove the table from the `EnrollmentManager`. I will create an issue and address it in the next sprint.

PR #748, opened 2020-04-10
Split enrollments of the ValidatorSet into candidates and validators

linked0:
The old `ValidatorSet` had an issue that it manages two kinds of enrollments which consist of the enrollments waiting to be a validator and the enrollments having already been a validator. This could cause an unexpected error, so we are splitting the class into the `EnrollmentPoo`l for the candidates of a validator and the `ValidatorSet` for the confirmed validators.

**In this PR, I have just split the ValidatorSet class into two classes, EnrollmentPool and ValidatorSet for smaller PR.** 

There will be other PRs having the following modifications.
1. Change unnecessary public into private in the `ValidatorSet`
2. Change function names in the `ValidatorSet` into proper names like `getEnrollment` into `getValidator`
3. Check abnormal situations in `add` of the `ValidatorSet` and `updateEnrolledHeight` of the EnrollmentManager.

Relates #735

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -22,6 +22,7 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.UTXOSet;
+import agora.consensus.EnrollmentPool;
 import agora.consensus.Validation;
 import agora.consensus.ValidatorSet;
 import agora.utils.Log;
@@ -78,6 +79,9 @@ public class EnrollmentManager
     /// enrolled height, and preimages.
     private ValidatorSet validator_set;
 
+    /// Enrollment pool managing enrollments waiting to be a validator
+    private EnrollmentPool enroll_pool;
+
     /// The count for generating pre-images
     private immutable uint AllCountPreimages = ValidatorSet.ValidatorCycle * 100;
 
@@ -95,6 +99,7 @@ public class EnrollmentManager
     public this (string db_path, KeyPair key_pair)
     {
         this.validator_set = new ValidatorSet(db_path);
+        this.enroll_pool = new EnrollmentPool(db_path);
 
         this.db = Database(db_path);
 
@@ -130,25 +135,26 @@ public class EnrollmentManager
     {
         this.db.close();
         this.validator_set.shutdown();
+        this.enroll_pool.shutdown();
     }
 
     /***************************************************************************
 
-        Add a enrollment data to the validators set
+        Add a enrollment data to the enrollment pool
 
         Params:
-            finder = the delegate to find UTXOs with
             enroll = the enrollment data to add
+            finder = the delegate to find UTXOs with
 
         Returns:
-            true if the enrollment data has been added to the validator set
+            true if the enrollment data has been added to the enrollment pool
 
     ***************************************************************************/
 
-    public bool add (scope UTXOFinder finder, const ref Enrollment enroll)
+    public bool add (const ref Enrollment enroll, scope UTXOFinder finder)
         @safe nothrow
     {
-        return this.validator_set.add(0, finder, enroll);
+        return this.enroll_pool.add(enroll, finder);
     }
 
     /***************************************************************************
@@ -162,7 +168,7 @@ public class EnrollmentManager
 
     public void remove (const ref Hash enroll_hash) @trusted
     {
-        this.validator_set.remove(enroll_hash);
+        this.enroll_pool.remove(enroll_hash);
     }
 
     /***************************************************************************
@@ -188,8 +194,8 @@ public class EnrollmentManager
 
         Params:
             enroll = Enrollment structure to add to the validator set
-            finder = the delegate to find UTXOs with
             block_height = enrolled blockheight
+            finder = the delegate to find UTXOs with
 
         Returns:
             true if the update operation was successful, false otherwise
@@ -197,14 +203,12 @@ public class EnrollmentManager
     ***************************************************************************/
 
     public bool addValidator (const ref Enrollment enroll,
-        scope UTXOFinder finder, size_t block_height) @safe
+        size_t block_height, scope UTXOFinder finder) @safe
     {
+        this.enroll_pool.remove(enroll.utxo_key);
+
         if (!this.validator_set.add(block_height, finder, enroll))
-        {
-            if (!this.validator_set.updateEnrolledHeight(enroll.utxo_key,
-                block_height))
-                return false;
-        }
+            return false;
 
         // set next height for revealing a pre-image
         if (enroll.utxo_key == this.enroll_key)
@@ -286,19 +290,19 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Check if a enrollment data exists in the validator set.
+        Check if a enrollment data exists in the enrollment pool.
 
         Params:
             enroll_hash = key for an enrollment data which is hash of frozen UTXO
 
         Returns:
-            true if the validator set has the enrollment data
+            true if the enrollment pool has the enrollment data
 
     ***************************************************************************/
 
     private bool hasEnrollment (const ref Hash enroll_hash) @trusted
     {
-        return this.validator_set.hasEnrollment(enroll_hash);
+        return this.enroll_pool.hasEnrollment(enroll_hash);
     }
 
     /***************************************************************************
@@ -318,7 +322,7 @@ public class EnrollmentManager
     public bool getEnrollment (const ref Hash enroll_hash,
         out Enrollment enroll) @trusted
     {
-        return this.validator_set.getEnrollment(enroll_hash, enroll);
+        return this.enroll_pool.getEnrollment(enroll_hash, enroll);
     }
 
     /***************************************************************************
@@ -372,7 +376,7 @@ public class EnrollmentManager
     public Enrollment[] getUnregistered (ref Enrollment[] enrolls)
         @trusted
     {
-        return this.validator_set.getUnregistered(enrolls);
+        return this.enroll_pool.getEnrollments(enrolls);
     }
 
     /***************************************************************************
@@ -795,23 +799,23 @@ unittest
 
     assert(man.createEnrollment(utxo_hash, 1, enroll));
     assert(!man.hasEnrollment(utxo_hash));
-    assert(!man.add(&storage.findUTXO, fail_enroll));
-    assert(man.add(&storage.findUTXO, enroll));
-    assert(man.validator_set.count() == 1);
+    assert(!man.add(fail_enroll, &storage.findUTXO));
+    assert(man.add(enroll, &storage.findUTXO));
+    assert(man.enroll_pool.count() == 1);
     assert(man.hasEnrollment(utxo_hash));
-    assert(!man.add(&storage.findUTXO, enroll));
+    assert(!man.add(enroll, &storage.findUTXO));
 
     // create and add the second Enrollment object
     auto utxo_hash2 = utxo_hashes[1];
     assert(man.createEnrollment(utxo_hash2, 1, enroll2));
-    assert(man.add(&storage.findUTXO, enroll2));
-    assert(man.validator_set.count() == 2);
+    assert(man.add(enroll2, &storage.findUTXO));
+    assert(man.enroll_pool.count() == 2);
 
     auto utxo_hash3 = utxo_hashes[2];
     Enrollment enroll3;
     assert(man.createEnrollment(utxo_hash3, 1, enroll3));
-    assert(man.add(&storage.findUTXO, enroll3));
-    assert(man.validator_set.count() == 3);
+    assert(man.add(enroll3, &storage.findUTXO));
+    assert(man.enroll_pool.count() == 3);
 
     Enrollment[] enrolls;
     man.getUnregistered(enrolls);
@@ -825,20 +829,22 @@ unittest
 
     // remove an Enrollment object
     man.remove(utxo_hash2);
-    assert(man.validator_set.count() == 2);
+    assert(man.enroll_pool.count() == 2);
 
     // test for getEnrollment with removed enrollment
     assert(!man.getEnrollment(utxo_hash2, stored_enroll));
 
     // test for enrollment block height update
     assert(!man.getEnrolledHeight(utxo_hash));
-    assert(man.addValidator(enroll, &storage.findUTXO, 9));
+    assert(man.addValidator(enroll, 9, &storage.findUTXO));
     assert(man.getEnrolledHeight(enroll.utxo_key) == 9);
-    assert(!man.addValidator(enroll, &storage.findUTXO, 9));
-    assert(man.getEnrolledHeight(utxo_hash2) == 0);
+    assert(!man.addValidator(enroll, 9, &storage.findUTXO));
+    assert(man.getEnrolledHeight(enroll2.utxo_key) == 0);
     man.getUnregistered(enrolls);
     assert(enrolls.length == 1);
-    assert(man.validator_set.count() == 2);  // has not changed
+    // One Enrollment was moved to validator set
+    assert(man.validator_set.count() == 1);
+    assert(man.enroll_pool.count() == 1);
 
     man.remove(utxo_hash);
     man.remove(utxo_hash2);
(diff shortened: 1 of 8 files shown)


linked0:
Ready for review.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,376 @@
...
+            return false;
+        }
+
+        try
+        {
+            () @trusted {
+                this.db.execute("INSERT INTO enrollment_pool (key, val) VALUES (?, ?)",
+                    enroll.utxo_key.toString(), buffer);
+            }();
+
```
AndrejMitrovic:
```suggestion
```

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        return true;
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the number of enrollments being managed by the enrollment pool,
+            which includes both registered and un-registered enrollments.
```
AndrejMitrovic: In `EnrollmentManager.updateEnrolledHeight`, the enrollment is removed from the pool when the height is updated. I assume that when the height is updated it means the enrollment has been registered. So then the pool only contains unregistered enrollments, right?

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,376 @@
...
+
+    ***************************************************************************/
+
+    public this (string db_path)
+    {
+        this.db = Database(db_path);
+
+        // create the table for enrollment pool if it doesn't exist yet
+        this.db.execute("CREATE TABLE IF NOT EXISTS enrollment_pool " ~
+            "(key TEXT PRIMARY KEY, val BLOC NOT NULL)");
```
AndrejMitrovic:
```suggestion
            "(key TEXT PRIMARY KEY, val BLOB NOT NULL)");
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -22,6 +22,7 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
```
AndrejMitrovic, line 22: Typo in commit message: candicates => candidates

source/agora/consensus/EnrollmentManager.d
```diff
@@ -212,7 +218,13 @@ public class EnrollmentManager
     public bool updateEnrolledHeight (const ref Hash enroll_hash,
         const size_t block_height) @safe
     {
-        if (!this.validator_set.updateEnrolledHeight(enroll_hash, block_height))
+        Enrollment enroll;
+        if (!this.enroll_pool.getEnrollment(enroll_hash, enroll))
+            return false;
+
+        this.enroll_pool.remove(enroll_hash);
```
AndrejMitrovic: There is a problem here:

- We take one enrollment from the pool
- We remove it from the pool
- We try to add it to the validator set

If the last step fails the enrollment no longer exists anywhere because it was removed from the pool. We should only call `this.enroll_pool.remove(enroll_hash);` if updating the enrolled height (adding it to the validator set) was successful.

AndrejMitrovic (changes requested):

source/agora/consensus/EnrollmentManager.d
linked0: Thanks for your comment. In the case, the function returns `false` and the ledger will terminate a node with the `assert(0)` call. The validator set can be restored with the ledger when a node starts.

AndrejMitrovic: Then the assert should be moved inside of here. Otherwise this is what's called "leaky abstractions".

linked0: It could be better for the ledger to have the role of determining whether a node terminates or not. But, I'll think about what you suggested.

AndrejMitrovic: If you want the Ledger to have the option to recover, then you need to make sure the EnrollmentManager is in a valid state.

That means to not remove the enrollment from the pool unless it was successfully added to the validator set.

linked0: The information for validators is stored in a block and the ledger restores validators from the block. At the time when the `updateEnrolledHeight` is being called, the validators' information has already been stored in a block.

source/agora/consensus/ValidatorSet.d
```diff
@@ -83,56 +80,40 @@ public class ValidatorSet
...
-                this.db.execute("INSERT INTO validator_set (key, val) VALUES (?, ?)",
-                    enroll.utxo_key.toString(), buffer);
-            }();
+            if (this.hasEnrollment(enroll.utxo_key))
+            {
+                log.info("Rejected already existing validator, Data was: {}",
+                    enroll);
+                return false;
+            }
+            else
```
Geod24: No need for `else` if you `return`.

source/agora/test/EnrollmentManager.d
```diff
@@ -26,71 +26,7 @@ import agora.consensus.EnrollmentManager;
 import agora.consensus.Genesis;
 import agora.test.Base;
 
-/// test enrollment process
```
Geod24, line 33: Why was this removed ?

source/agora/consensus/ValidatorSet.d
```diff
@@ -788,7 +681,7 @@ unittest
 
     // make test blocks used for restoring validator set
     Block[] blocks;
-    ulong last_height = ValidatorSet.ValidatorCycle;
+    ulong last_height = Enrollment.ValidatorCycle;
```
Geod24, line 702: There's a lot of diff coming from simply moving this constant. I think it should be its own commit.

source/agora/consensus/EnrollmentManager.d
Geod24: > It could be better for the ledger to have the role of determining whether a node terminates or not.

Ideally, it would be better if the node would not terminate ;)

> At the time when the `updateEnrolledHeight` is being called, the validators' information has already been stored in a block.

Right, but we don't read the block here, we read the `enroll_pool`. And if the `enroll_pool` does not contain the enrollment, we bail out. No checking the block, no requesting other nodes.

Also, I didn't see where it was called from ?

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,375 @@
...
+                        :memory: was passed
+
+    ***************************************************************************/
+
+    public this (string db_path)
+    {
+        this.db = Database(db_path);
+
+        // create the table for enrollment pool if it doesn't exist yet
+        this.db.execute("CREATE TABLE IF NOT EXISTS enrollment_pool " ~
```
Geod24, line 59: Do we really need a table here ?

Geod24 (changes requested):
A few comments, let's discuss IRL the rest for simplicity.

source/agora/test/EnrollmentManager.d
linked0: The unit test code has been merged into the code below the comment `/// test for  enrollment process & revealing a pre-image periodically`. Changing the way the EnrollmentManager stores the data make the test code need to be changed and I found there was redundant test code, I removed the redundant code and merged the part of it into the other test code.

source/agora/consensus/ValidatorSet.d
linked0: I will do it. but in some commit of this PR, there will be two `ValidatorCycle` constants because the `EnrollmentPool` must reference it. In the end, the ValidatorCycle will only have existed in the `Enrollment` class.

source/agora/consensus/ValidatorSet.d
linked0: Please share your idea in more detail.

source/agora/consensus/EnrollmentPool.d
linked0: If we don't need this new table, I didn't need to start to resolve this issue. This PR and the issue related to this is all about creating a new table like enrollment_pool.

linked0: > No checking the block, no requesting other nodes.

You're right. We need some recovery code like requesting enrollments from other nodes. We need another issue for that and I'm sorry for the late catch.

source/agora/consensus/ValidatorSet.d
AndrejMitrovic:
```D
if (condition)
{
    do();
    return;
}
else
{
    something();
}
```

Is the same as:

```D
if (condition)
{
    do();
    return;
}

something();
```

source/agora/consensus/EnrollmentPool.d
linked0: > No checking the block, no requesting other nodes

I have created an issue for the problem. #755

linked0:
Ready for review.
I resolved the comments and created a new issue for an issue that can't be handled in this PR.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -847,7 +859,6 @@ unittest
     assert(man.getEnrolledHeight(utxo_hash2) == 0);
     man.getUnregistered(enrolls);
     assert(enrolls.length == 1);
-    assert(man.count() == 2);  // has not changed
```
AndrejMitrovic: Why was the unittest removed instead of just changing the number that's tested?

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,374 @@
...
+        }
+
+        // check if already exists
+        if (this.hasEnrollment(enroll.utxo_key))
+        {
+            log.info("Rejected already existing enrollment, Data was: {}",
+                enroll);
+            return false;
+        }
+
```
AndrejMitrovic, line 107: The `buffer` should be declared here where it's initially used, not at the start of the function. We always try to keep declarations close to where they are initially used, it prevents accidental usage of uninitialized variables.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,374 @@
...
+import d2sqlite3.database;
+import d2sqlite3.library;
+import d2sqlite3.results;
+import d2sqlite3.sqlite3;
+
+mixin AddLogger!();
+
+/// Ditto
+public class EnrollmentPool
```
AndrejMitrovic, line 39: There is an inconsistency in the API.

`hasEnrollment` doesn't throw, but `count` throws, `add` doesn't throw, but `remove` throws, and then `getEnrollment` throws and **also** returns a status flag.

These methods need to be consistent.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -130,44 +135,45 @@ public class EnrollmentManager
...
-        return this.validator_set.add(block_height, finder, enroll);
+        return this.enroll_pool.add(block_height, finder, enroll);
     }
 
     /***************************************************************************
 
         Returns:
-            the number of enrollments being managed by this EnrollmentManager.
+            the number of enrollments being managed by the enrollment pool.
             Note: this includes both registered and un-registered enrollments.
```
AndrejMitrovic: I think the Note: should be removed, because the pool only tracks un-registered enrollments.

AndrejMitrovic (dismissed):

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,374 @@
...
+    ***************************************************************************/
+
+    public bool add (ulong block_height, scope UTXOFinder finder,
+        const ref Enrollment enroll) @safe nothrow
+    {
+        static ubyte[] buffer;
+
+        // check validity of the enrollment data
+        if (auto reason = isInvalidEnrollmentReason(enroll, block_height + 1,
+            finder))
```
linked0: The `block_height` is not used properly, the function of `isInvalidEnrollmentReason` should be changed.

source/agora/consensus/ValidatorSet.d
```diff
@@ -83,56 +84,38 @@ public class ValidatorSet
...
-        }
-
         try
         {
+            if (this.hasEnrollment(enroll.utxo_key))
+            {
+                log.info("Rejected already existing validator, Data was: {}",
+                    enroll);
+                return false;
+            }
```
linked0: This code doesn't need to be here in this `try/catch` statement.

source/agora/consensus/EnrollmentPool.d
linked0: > These methods need to be consistent.

Right, those functions should be consistent. I'll make it so.

source/agora/consensus/EnrollmentManager.d
linked0: > Why was the unittest removed instead of just changing the number that's tested?

This kind of deletion of code could be confusing. I'll restore it.

source/agora/consensus/ValidatorSet.d
```diff
@@ -83,56 +84,38 @@ public class ValidatorSet
...
     {
         static ubyte[] buffer;
-
-        // check validity of the enrollment data
-        if (auto reason = isInvalidEnrollmentReason(enroll, block_height + 1,
-            finder))
-        {
-            log.info("Invalid enrollment data: {}, Data was: {}", reason, enroll);
-            return false;
-        }
```
linked0: Revert it

linked0:
It needs some more modification. I'll request a review after the modifications are done.

linked0:
Ready for review about the code that Mathias and I pair programmed.

Geod24:
Could you put the first commit in its own PR ?

linked0:
> Could you put the first commit in its own PR ?

OK. I will do it.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,404 @@
...
+        Params:
+            finder = the delegate to find UTXOs with
+            enroll = the enrollment data to add
+
+        Returns:
+            true if the enrollment data has been added to the enrollment pool
+
+    ***************************************************************************/
+
+    public bool add (scope UTXOFinder finder, const ref Enrollment enroll)
```
linked0: Invert params

linked0 (commented):
Some suggestions.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -130,25 +135,26 @@ public class EnrollmentManager
...
+        Add a enrollment data to the enrollment pool
 
         Params:
-            finder = the delegate to find UTXOs with
             enroll = the enrollment data to add
+            finder = the delegate to find UTXOs with
 
         Returns:
-            true if the enrollment data has been added to the validator set
+            true if the enrollment data has been added to the the enrollment pool
```
TrustHenry:
```suggestion
            true if the enrollment data has been added to the enrollment pool
```

source/agora/consensus/ValidatorSet.d
```diff
@@ -310,7 +262,7 @@ public class ValidatorSet
         {
             () @trusted {
                 auto results = this.db.execute("SELECT val FROM validator_set" ~
-                    " WHERE enrolled_height is not null");
+                    " ORDER BY key ASC");
```
TrustHenry, line 265: > Get all the current validators

This change requires the document to be changed.
The meaning of sorted should be added.

TrustHenry (dismissed):

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,404 @@
...
+        size_t cnt;
+        try
+        {
+            cnt = this.db.execute("SELECT count(*) FROM enrollment_pool").
+                oneValue!size_t;
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error on count");
+        }
```
AndrejMitrovic:
```suggestion
        try
        {
            return this.db.execute("SELECT count(*) FROM enrollment_pool").
                oneValue!size_t;
        }
        catch (Exception ex)
        {
            log.error("Database operation error on count");
            return 0;
        }
```

And the documentation needs to be updated to say 0 is returned in case of errors.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -0,0 +1,404 @@
...
+    public void remove (const ref Hash enroll_hash) @trusted nothrow
+    {
+        try
+        {
+            this.db.execute("DELETE FROM enrollment_pool WHERE key = ?",
+                enroll_hash.toString());
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error on remove");
```
AndrejMitrovic, line 175: When I said before that the API needs to be consistent I meant the code should do either one of the following:

A) Return a status code for every function which can fail. So the calling code can handle it.
B) Make all functions throw if they can fail. The calling code then must use a try/catch to handle it, or it can let the exception propagate upwards so other calling code handles it.
C) Catch any exceptions and halt the program with `assert(0)`.

Right now the code doesn't do any of these. Instead it catches the exception, logs it, and continues execution.

But what is the purpose of continuing execution like nothing bad happened? Will the database automatically fix itself on the next call? What is the expectation here?

linked0: The failure to remove an enrollment data from the enrollment pool is not so critical and so execution can continue. In this PR, if adding the enrollment to the validator set fails, the ledger makes the system terminate as follows.

```
foreach (enrollment; block.header.enrollments)
    if (!this.enroll_man.addValidator(enrollment,
        assert(0);
```

Mathias and I discussed that in the pair programming session.
The recovery from the database failure can be executed when a node start, which is not implemented yet and could be another issue.

linked0:
Ready for review.
Additionally, I will make another PR about deleting the table named `node_enroll_data` used in the `EnrollmentManager` class soon like next Monday. The table manages only the `next_reveal_height` value which can be just set to the block height after the validator set is restored through the `restoreValidators` function.

Geod24:
Rebased on v0.x.x

Geod24 (approved):

Issue #586, opened 2020-02-25
Split up the Full Node and Validator functionality into two separate classes


AndrejMitrovic:
Currently the `Node` class contains functionality for both a Full Node and Validator. 

The `nominator` instance should not be visible for Full Nodes, because they are not part of consensus.

Ideally we would have something like this:

```D
module agora.node.FullNode; 

import agora.api.FullNode;
class FullNode : API { ... }
```

```D
module agora.node.Validator; 

import agora.api.Validator;
import agora.node.FullNode;

class Validator : FullNode, API { ... }
```

Notice how Validator **is a** FullNode, and provides additional functionality through the agora.api.validator API interface. Also notice that `: API` in the `agora.node.FullNode` and `agora.node.Validator` modules are **different**. One refers to the one in `agora.api.FullNode`, and another refers to `agora.api.Validator`.

Definition of done:

- [x] Split the classes into two
- [x] Each class should ideally live in its own module, e.g. FullNode.d, Validator.d
- [x] Based on the configuration of the node, either `FullNode` or `Validator` should be instantiated when we want to start a node.
- [x] Fields which only belong in the Validator class should not be present in the FullNode class. For example, the `Nominator` instance should only be visible in the Validator class.
- [x] The test-suite has tests for both FullNode and Validator. It should be possible to instantiate a mix set of nodes (some are FullNode, some are Validator) and make the tests still work.

TrustHenry:
> Based on the configuration of the node, either FullNode or Validator should be instantiated when we want to start a node.

The configuration part determines whether it operates as FullNode or Validator based on the `node.is_validator` : boolean value.

AndrejMitrovic:
Some things that need to be thought about with regards to the test-suite:

- `TestNode` in the test-suite. Currently it's defined as `public class TestNode : Node, TestAPI`. But you will probably have to change this. Maybe you will have `TestFullNode` and `TestValidatorNode`, or maybe you could add a template parameter to TestNode like so: `public class TestNode (BaseAPI) : BaseAPI, TestAPI`. Experiment and see what is the best option.
- `TestAPIManager.createNewNode`.  It currently instantiates a `TestNode` and then appends it to the `this.nodes` array. But If we have two different node types we might have to split them into two. Maybe `this.full_nodes` and `this.validators`. Tests in the test-suite will have to be fixed too.

TrustHenry:
We thought publickey was not required in Fullnode.
[...]

PR #750, opened 2020-04-14
Change the name from NetworkInfo to NodeInfo

TrustHenry:
The existing struct `networkInfo` had only network information.
We need to know whether the node is `Validator` or not in order to distinguish between `FullNode` and `Validator`.
This is the first work of the work.

Relates #586

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -33,8 +33,8 @@ public enum NetworkState : ubyte
     Complete
 }
 
-/// Contains the network info (state & addresses)
-public struct NetworkInfo
+/// Contains the node info (state & addresses & isValidator)
+public struct NodeInfo
 {
     /// Whether the node knows about the IPs of all its quorum set nodes
     public NetworkState state;
@@ -88,14 +88,14 @@ public interface API
     /***************************************************************************
 
         Returns:
-            The peer network of this node
+            The peer information on this node
 
         API:
-            GET /network_info
+            GET /node_info
 
     ***************************************************************************/
 
-    public NetworkInfo getNetworkInfo ();
+    public NodeInfo getNodeInfo ();
 
     /***************************************************************************
 
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -223,8 +223,8 @@ public class Node : API
         return this.config.node.key_pair.address;
     }
 
-    /// GET: /network_info
-    public override NetworkInfo getNetworkInfo () pure nothrow @safe @nogc
+    /// GET: /node_info
+    public override NodeInfo getNodeInfo () pure nothrow @safe @nogc
     {
         return this.network.getNetworkInfo();
     }
--- a/source/agora/network/NetworkClient.d
+++ b/source/agora/network/NetworkClient.d
@@ -128,19 +128,19 @@ class NetworkClient
     /***************************************************************************
 
         Get the network info of the node, stored in the
-        `net_info` parameter if the request succeeded.
+        `node_info` parameter if the request succeeded.
 
         Returns:
-            `NetworkInfo` if successful
+            `NodeInfo` if successful
 
         Throws:
             `Exception` if the request failed.
 
     ***************************************************************************/
 
-    public NetworkInfo getNetworkInfo ()
+    public NodeInfo getNodeInfo ()
     {
-        return this.attemptRequest(this.api.getNetworkInfo(), this.exception);
+        return this.attemptRequest(this.api.getNodeInfo(), this.exception);
     }
 
     /***************************************************************************
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -417,9 +417,9 @@ public class NetworkManager
         {
             try
             {
-                auto net_info = node.getNetworkInfo();
-                this.addAddresses(net_info.addresses);
-                if (net_info.state == NetworkState.Complete)
+                auto node_info = node.getNodeInfo();
+                this.addAddresses(node_info.addresses);
+                if (node_info.state == NetworkState.Complete)
                     return;  // done
 
                 // if it's incomplete give the client some time to connect
@@ -502,9 +502,9 @@ public class NetworkManager
     }
 
     /// Returns: the list of node IPs this node is connected to
-    public NetworkInfo getNetworkInfo () pure nothrow @safe @nogc
+    public NodeInfo getNetworkInfo () pure nothrow @safe @nogc
     {
-        return NetworkInfo(
+        return NodeInfo(
             this.minPeersConnected()
                 ? NetworkState.Complete : NetworkState.Incomplete,
             this.known_addresses);
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -378,7 +378,7 @@ public class TestAPIManager
         {
             const timeout = 5.seconds;
             this.nodes.each!(node =>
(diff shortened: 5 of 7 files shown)


TrustHenry:
It's a self-preview.
There is no problem with the unit test, 
but the node address is initialized in the docker test(circle ci)

source/agora/node/Node.d
```diff
@@ -223,10 +223,13 @@ public class Node : API
...
 
-    /// GET: /network_info
-    public override NetworkInfo getNetworkInfo () pure nothrow @safe @nogc
+    /// GET: /node_info
+    public override NodeInfo getNodeInfo () pure nothrow @safe @nogc
     {
-        return this.network.getNetworkInfo();
+        // NetworkInfo only brings network-related information.
+        NodeInfo node_info = this.network.getNetworkInfo();
+        node_info.is_validator = this.config.node.is_validator;
```
TrustHenry: I wanted to separate from `network.NetworkInfo`
~~Is this where the initialization takes place?~~

source/agora/node/Node.d
```diff
@@ -226,7 +226,10 @@ public class Node : API
     /// GET: /node_info
     public override NodeInfo getNodeInfo () pure nothrow @safe @nogc
     {
-        return this.network.getNetworkInfo();
+        // NetworkInfo only brings network-related information.
+        NodeInfo node_info = this.network.getNetworkInfo();
```
linked0: If we get a NodeInfo object through the `getNetworkInfo` function, why don't you change the name of the function into `getNodeInfo` also?

TrustHenry: Like the annotation written on the 229th line, getNetworkInfo only brings network information, and NodeInfo is a higher concept than NetworkInfo, so getNetNetworkInfo in the network uses its function as it is.

TrustHenry: And since the assigned struct is `NodeInfo`, I assigned the variable name to `node_info`

MichaelKim20:
tests/system/source/main.d(40,21): no property `getNetworkInfo` for type `agora.api.FullNode.API`

TrustHenry:
There was an uncorrected part of the circleCI error in the agora cli part. Modification completed

source/agora/node/Node.d
AndrejMitrovic: It's still misleading to have a function named `getNetworkInfo` while returning a different type.

I would suggest doing something like this:

```D
NodeInfo node_info;
this.network.fillNetworkInfo(node_info.state, node_info.addresses);
node_info.is_validator = this.config.node.is_validator;
```

That means changing the function name and its code.

TrustHenry:
Ready for review. @linked0 @MukeunKim @AndrejMitrovic @Geod24 
I'm happy to receive everyone's review.

linked0 (approved):
Nice work.
A minor change is needed like "The peer infomation of this node" into "The information of this node".

MichaelKim20 (approved):
LGTM

Geod24:
Yes to the first commit, but not sure about the second one. First, it doesn't sound like we should rely on our peers to tell us if they are validators or not. Second, that boolean is likely to not be enough. Do you have more commits where you build up on this ?

TrustHenry:
> First, it doesn't sound like we should rely on our peers to tell us if they are validators or not. Second, that boolean is likely to not be enough. Do you have more commits where you build up on this ?

Clearly, the boolean value of the is_valider is running as a Validator for Node.
It is not clear whether this value is a real Validator.
This is simply because it is the set config value of the node.
However, the purpose of this PR should be seen as part of the action to address issue #586.

@Geod24 If you provide more information than boolean value, what additional value would you like to provide?

AndrejMitrovic:
@TrustHenry's original problem as he described it to me:

#512 splits the Node into two. But there's also the issue of the NetworkManager. It calls `getPublicKey` on all of the nodes. But only the Validator nodes will implement the `getPublicKey` API, not FullNode, so this call will fail (probably throw?) when called on a FullNode. To work around that we could add the `is_validator` flag like Henry did, but there's also the issue of trusting what the node says. The node could lie about `is_validator`, and the next call to `getPublicKey` would fail.

So the way things work now:
- Node has a quorum set config
- Node has a DNS seed or a list of initial nodes to connect to
- Node starts, discovers the network, calls getPublicKey on each node. It tries to connect to all the public keys as listed in the quorum set. Then it starts nominating.

There's also a problem with this implementation because `getPublicKey` can't be trusted. A node could lie about its public key. It would work if we required the node to sign all messages. But to ensure all messages are signed it means /all/ nodes must have a public key, not just the Validators. Do you see the problem?

Maybe a FullNode should also have a PublicKey.

TrustHenry:
To sign all messages on a node, both Fullnode and Validator require a publickey.
If node sign only the messages that need to be signed, node may not need publickey all of them.
What do you think about signing API messages that only Validaotor provides?

Geod24:
What message need to be signed ?

TrustHenry:
In an unreliable state, getPublickey is the most basic information for creating a trusted network.
The most important thing is to send a response message from getPublickey along with the message of the signature & publickey.

Geod24:
Public keys don't provide more safety than an IP address. It is trivial and cheap to generate a public key, so nodes can do so at will if they wish to flood the network.

TrustHenry:
Ah, it would be better to sign with IP address and port. &  include a public key.

TrustHenry:
If another node ~~intercepts~~ copy the signature information, it is more difficult because it also has to intercept the IP address.

Geod24:
What do you mean by "intercept" ? Even if someone MITM the connection and change the returned `PublicKey` sent by a validator to its client, they won't be able to generate signed data. But for that, we should just encrypt connections so they can't be intercepted, and that happens at a lower layer.

TrustHenry:
Ah sorry, I say it's a signature copy, not an intercept.
I am talking about how a node that's not the original node uses the signature by replicating the signature value.

Geod24:
I think we are getting lost here.

Let me recap (the following is a list of successive reasoning, not a list):
- Currently, the `NetworkInfo` relies on all nodes having `PublicKey`;
- This is an issue for non-Validator nodes, because they do not need a `PublicKey`;
- We have two options: Make non-Validators have `PublicKey`, or change `NetworkInfo`;
- What would be the benefit of having non-Validators have a `PublicKey` ?
- Well, what is the benefit of having Validators have a `PublicKey`, and does that reasoning apply to non-Validators ?
- Validators need a public key because it is their identity. This identity is also stored in the chain;
- However, this does not apply to non-Validator: their identity is *not* stored in the chain, and *not* tied to the 40k+ stake of BOA;
- A `PublicKey` is easy to generate (it's just a derivation from a random number), so if we don't have Sybil prevention, it's worthless as an identity check;
- Hence, non-Validators will not gain anything from having a `PublicKey`.

Does this make sense ?

Geod24:
@TrustHenry : If you remove the second commit, this will just be a rename and can be merged right away.

TrustHenry:
I'm ready. removed the second commit

source/agora/api/FullNode.d
```diff
@@ -88,14 +88,14 @@ public interface API
     /***************************************************************************
 
         Returns:
-            The peer network of this node
+            The peer infomation of this node
```
Geod24:
```suggestion
            The peer information on this node
```

Geod24 (approved):
LGTM, merge when green

PR #751, opened 2020-04-14
Make the validator set to clear up expired validators

linked0:
This PR changes the name of the function which clears up an expired validator to a proper name and makes the ledger to call the function in adding a block.

The test code on the module of agora.test.EnrollmentManager will be added in other PR because it's a kind of big task for me. I'll make a PR for that on Thursday or Friday.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -236,6 +236,8 @@ public class Ledger
         // read back and cache the last block
         if (!this.storage.readLastBlock(this.last_block))
             assert(0);
+
+        this.enroll_man.clearExpiredValidators(this.last_block.header.height);
     }
 
     /***************************************************************************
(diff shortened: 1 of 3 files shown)


source/agora/node/Ledger.d
```diff
@@ -236,6 +236,8 @@ public class Ledger
         // read back and cache the last block
         if (!this.storage.readLastBlock(this.last_block))
             assert(0);
+
```
AndrejMitrovic, line 239: The commit message is unclear. "call the in"?

Maybe just write: "Clean expired enrollments after the block was finalized"

linked0: The `git commit -m` command removes the text between two grave accents(`) in my commit message. I will change it. Thanks.

Geod24 (approved):
Nice!

PR #756, opened 2020-04-16
Split up the Full Node and Validator functionality into two separate classes

TrustHenry:
We want to be a mixed network configuration of Fullnode and Validator.
The function of the Validator is not required when operating as a fullnode.
Each function can be separated to remove or add unnecessary functions.

~~Currently, this PR has two errors in tests~~
~~Error occurs in unittest in agora.test.Failures.~~
~~Also in the System Integration Test.~~
~~There seems to be a problem when a network test occurs.~~

Relates #586


AndrejMitrovic:
This is still work in progress. Do you want a review, or should we wait until it's finished?

TrustHenry:
> This is still work in progress. Do you want a review, or should we wait until it's finished?


I really want a review.  But there is a problem and I want to know what the problem is.

source/agora/node/FullNode.d
```diff
@@ -181,7 +174,7 @@ public class Node : API
 
             bool isNominating ()
             {
-                return this.config.node.is_validator &&
+                return (this.nominator !is null) &&
```
AndrejMitrovic: A FullNode is not a Validator, right? Therefore it will never have a Nominator instance. So what is the purpose of this check here if it will never return true?

One option is to replace it with an inline delegate like so:

```D
// false: full node is never in the middle of nominating
this.network.startPeriodicCatchup(this.ledger, () => false);
```

But a better solution is to change `startPeriodicCatchup` to make the delegate optional.

source/agora/network/NetworkClient.d
```diff
@@ -186,6 +187,9 @@ class NetworkClient
     {
         try
         {
+            if (!this.api.getNodeInfo().is_validator)
```
AndrejMitrovic: This change has several problems:
- sendEnvelope() was previously an asynchronous task. You have now turned it into a synchronous (blocking) task. Naturally this will break all the unittest code that uses timeouts.
- You should query the node whether it is a validator **once** after you connect to it. What is the purpose of asking it if it's still a validator every time we gossip an envelope? The node won't magically turn into a FullValidator while the connection is still active.

source/agora/network/NetworkManager.d
```diff
@@ -375,7 +375,7 @@ public class NetworkManager
         {
             try
             {
-                node.getPublicKey();
+                //node.getPublicKey();
```
AndrejMitrovic: This should be fixed.

source/agora/node/FullNode.d
```diff
@@ -255,7 +242,7 @@ public class Node : API
             this.network.gossipTransaction(tx);
 
             // then nominate
-            if (this.config.node.is_validator)
+            if (this.nominator !is null)
```
AndrejMitrovic: This belongs in the Validator class. The purpose of splitting the classes was to not have to do these checks.

https://github.com/bpfkorea/agora/issues/586

In the definition of done, the 4th item in the list is: 

> Fields which only belong in the Validator class should not be present in the FullNode class. For example, the Nominator instance should only be visible in the Validator class.

So remove the `nominator` instance from the FullNode class.

source/agora/node/FullNode.d
```diff
@@ -572,7 +531,7 @@ public class Node : API
 
 *******************************************************************************/
 
-public Node runNode (Config config)
+public Node runNode (Node) (Config config)
```
AndrejMitrovic: Good change, but the documentation needs to be updated.

source/agora/node/Validator.d
```diff
@@ -0,0 +1,95 @@
...
+
+*******************************************************************************/
+
+public class Validator : FullNode, API
+{
+    /// Ctor
+    public this (const Config config)
+    {
+        super(config);
+        if (this.config.node.is_validator)
```
AndrejMitrovic: Why? This class **is** a validator. Remove these `if` checks.

source/agora/node/Validator.d
```diff
@@ -0,0 +1,95 @@
...
+
+        Params:
+            envelope = the SCP envelope
+
+    ***************************************************************************/
+
+    public void receiveEnvelope (SCPEnvelope envelope)
+    {
+        // we should not receive SCP messages unless we're a validator node
+        if (!this.config.node.is_validator)
```
AndrejMitrovic: Remove this.

source/agora/node/Validator.d
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /// GET: /get_preimage
+    public PreimageInfo getPreimage (Hash enroll_key)
+    {
+        PreimageInfo preimage;
+        this.enroll_man.getValidatorPreimage(enroll_key, preimage);
+        return preimage;
+    }
+
```
```suggestion
```

source/agora/node/Validator.d
```diff
@@ -0,0 +1,95 @@
...
+    {
+        // we should not receive SCP messages unless we're a validator node
+        if (!this.config.node.is_validator)
+            return;
+
+        this.nominator.receiveEnvelope(envelope);
+    }
+
+    /// GET: /get_preimage
+    public PreimageInfo getPreimage (Hash enroll_key)
```
AndrejMitrovic: `override`

source/agora/node/Validator.d
```diff
@@ -0,0 +1,95 @@
...
+
+        API:
+            GET /envelope
+
+        Params:
+            envelope = the SCP envelope
+
+    ***************************************************************************/
+
+    public void receiveEnvelope (SCPEnvelope envelope)
```
AndrejMitrovic: `override`

source/agora/node/Validator.d
```diff
@@ -0,0 +1,95 @@
...
+        if (this.config.node.is_validator)
+        {
+            this.nominator = this.getNominator(this.network,
+            this.config.node.key_pair, this.ledger, this.taskman,
+            this.config.quorum);
+        }
+    }
+
+    /// GET /public_key
+    public PublicKey getPublicKey () pure nothrow @safe @nogc
```
AndrejMitrovic: `override`

AndrejMitrovic (changes requested):

source/agora/cli/multi/main.d
```diff
@@ -13,9 +13,9 @@
 
 module agora.cli.multi.main;
 
-import agora.api.FullNode;
+import agora.api.Validator;
```
MichaelKim20: I think the following source needs to be fixed in order to run normally.

``` D
import agora.api.Validator;
...
import agora.node.FullNode;
import agora.node.Validator;
...
    Validator[] nodes;
        nodes ~= runNode!Validator(config);
...
}
```

MichaelKim20 (changes requested):
There is one modification in the CLI.

MichaelKim20:
To compile agora-client, `dub.json` must modify it as follows.

```
        {
            "name": "client",
            "targetName": "agora-client",
            "mainSourceFile": "source/agora/cli/client/main.d",
            "excludedSourceFiles": [
                "source/agora/node/main.d",
                "source/agora/cli/multi/*"
            ]
        },
```

dub.json
```diff
@@ -87,16 +87,7 @@
...
-                "source/agora/api/Validator.d",
-                "source/agora/common/Config.d",
-                "source/agora/common/SCPHash.d",
-                "source/agora/network/*",
-                "source/agora/node/*",
-                "source/agora/utils/PrettyPrinter.d",
-                "source/agora/utils/SCPPrettyPrinter.d",
-                "source/scpp/*",
-                "source/scpd/*",
-                "source/agora/consensus/protocol/*",
```
Geod24: That doesn't look right.

source/agora/api/FullNode.d
```diff
@@ -201,23 +204,4 @@ public interface API
...
-            enroll_key = The key for the enrollment in which the pre-image is
-                contained.
-
-        Returns:
-            preimage information of the validator if exists, otherwise
-                PreImageInfo.init
-
-    ***************************************************************************/
-
-    public PreImageInfo getPreimage (Hash enroll_key);
```
Geod24, line 225: This doesn't look right either. The full nodes should be able to propagate preimages too.

source/agora/cli/multi/main.d
```diff
@@ -13,9 +13,10 @@
...
 import agora.common.Config;
-import agora.node.Node;
+import agora.node.FullNode;
```
Geod24, line 17: This shouldn't be needed, is it ?

source/agora/cli/multi/main.d
```diff
@@ -13,9 +13,10 @@
 
 module agora.cli.multi.main;
 
-import agora.api.FullNode;
+import agora.api.Validator;
```
Geod24: I don't think the import is necessary, actually. Where is it used ?

source/agora/network/NetworkManager.d
```diff
@@ -502,12 +502,11 @@ public class NetworkManager
...
-    public NodeInfo getNetworkInfo () pure nothrow @safe @nogc
+    public void fillNetworkInfo (ref NetworkState state, ref Set!string addresses)
+        pure nothrow @safe @nogc
     {
-        return NodeInfo(
-            this.minPeersConnected()
-                ? NetworkState.Complete : NetworkState.Incomplete,
-            this.known_addresses);
+        state = this.minPeersConnected() ? NetworkState.Complete : NetworkState.Incomplete;
+        addresses = this.known_addresses;
```
Geod24: Looks unrelated (bad rebase?)

source/agora/node/FullNode.d
```diff
@@ -568,11 +495,12 @@ public class Node : API
     The initialization process of the node is then completed.
 
     Params:
+      Node = FullNode, Validator Template is supported
```
Geod24:
```suggestion
      Node = Type of node to spawn (`FullNode` or `Validator`)
```

source/agora/node/FullNode.d
```diff
@@ -541,21 +475,14 @@ public class Node : API
...
 
-    /// GET: /get_preimage
-    public PreImageInfo getPreimage (Hash enroll_key)
+    ///
+    protected void callTryNominate () @safe
     {
-        PreImageInfo preimage;
-        this.enroll_man.getValidatorPreimage(enroll_key, preimage);
-        return preimage;
     }
```
Geod24: This `callTryNominate` looks suspicious.

Geod24 (changes requested):

dub.json
TrustHenry: Re-add only that part to the ExcludedSourceFiles

source/agora/api/FullNode.d
TrustHenry: They're back in place.

source/agora/cli/multi/main.d
TrustHenry: Yes,

source/agora/network/NetworkManager.d
TrustHenry: This is the first commit.
I excluded it because it is not relevant.

source/agora/node/FullNode.d
TrustHenry: Thank you.

MichaelKim20:
Please look at the source code below.
https://github.com/MukeunKim/agora-test/blob/master/source/app.d
Error occurs when invoking function of ValidatorAPI.

```
core.exception.AssertError@../../../../.dub/packages/vibe-core-1.9.1/vibe-core/source/vibe/core/log.d(895): Not Found (404)
```

Is it possible to use two API interfaces in vibe.d?

AndrejMitrovic:
> This doesn't look right either. The full nodes should be able to propagate preimages too.

Oh right, because they will need it to be able to validate blocks.

AndrejMitrovic:
> Please look at the source code below.

@MukeunKim your link is also a 404. :D

MichaelKim20:
> > Please look at the source code below.
> 
> @MukeunKim your link is also a 404. :D

@AndrejMitrovic I changed it to the public.

AndrejMitrovic:
@MukeunKim: This fixes it:

```diff
diff --git a/source/agora/node/Node.d b/source/agora/node/Node.d
index c04f291a..eb6014c6 100644
--- a/source/agora/node/Node.d
+++ b/source/agora/node/Node.d
@@ -638,7 +638,7 @@ class ValidatorNode : FullNode, ValidatorAPI
 unittest
 {
     auto router2 = new URLRouter;
-    router2.registerRestInterface(new ValidatorNode);
+    router2.registerRestInterface!ValidatorAPI(new ValidatorNode);
     auto settings2 = new HTTPServerSettings;
     settings2.port = 8002;
     settings2.bindAddresses = ["::1", "127.0.0.1"];
```

I think the issue is that `registerRestInterface` doesn't look into the base class for all the APIs the class implements. Maybe this is a vibe.d bug. However, you can work around it with the change I showed here.

Geod24:
> Is it possible to use two API interfaces in vibe.d?

Do you get a warning while building ? [This kind](https://github.com/vibe-d/vibe.d/issues/2414) ?

MichaelKim20:
> > Is it possible to use two API interfaces in vibe.d?
> 
> Do you get a warning while building ? [This kind](https://github.com/vibe-d/vibe.d/issues/2414) ?

``` D
interface FullNodeAPI
{
    int getValue1();
}

interface ValidatorAPI : FullNodeAPI
{
    int getValue2();
}

class FullNode : FullNodeAPI
{
    public int getValue1()
    {
        return 1;
    }
}

class ValidatorNode : FullNode, ValidatorAPI
{
    public int getValue2()
    {
        return 2;
    }
}
```
Create a server using `ValidatorNode`.
HTTP404 error occurs when invoking the method `getValue2` that is only in interface `ValidatorAPI`.

source/agora/test/Failures.d
```diff
@@ -30,6 +30,7 @@ unittest
 
     TestConf conf = { threshold : 100, timeout : 1_000 };
     auto network = makeTestNetwork(conf);
+    scope (exit) network.printLogs();
```
AndrejMitrovic: @TrustHenry we should remove this, because the nodes will be `null` after the constructor fails, so we cannot call their `printLog` method.

AndrejMitrovic:
This is currently WIP, don't review yet.

AndrejMitrovic:
So here's what we need to do:

- Change `getClient` to return a `ValidatorAPI` interface. This means we don't have to use templates in `Base.d`.
- Change `NetworkManager` to contain a map of PubKey => NetworkClient
- When we connect to each node we ask for its public key. If we get a 404 error it means this is a FullNode, not a validator. If we get a public key it means we can put the Key/ValidatorAPI into the map.
- Change `NetworkManager` to propagate validator messages to only validator nodes, and only to the ones in our quorum configuration. This means using the map that we just talked about. So for example, `gossipEnvelope` would only emit `SCPEnvelope` to nodes which are in that map

Personally I think we would need two maps:

```
Address[PubKey] validator_to_addr;  // validator to address
NetworkClient[Address] addr_to_client;  // address to NetworkClient
```

We would always add nodes to the `addr_to_client` map (because all nodes have an address), and we would add only Validator nodes to the `validator_to_addr` map. Then for each Validator in `validator_to_addr` you can find its NetworkClient in the `addr_to_client` map.

Maybe @Geod24 has a better idea than me though.

TrustHenry:
> When we connect to each node we ask for its public key. If we get a 404 error it means this is a FullNode, not a validator. If we get a public key it means we can put the Key/ValidatorAPI into the map.

If It request a public key when connecting to each node and distinguish the Validator by 404 error only,
Even if It are really a validator, you may be excluded from the 404 error code as a instant network sequence. Is it right to be excluded until the next request?

AndrejMitrovic:
404 means the node does not implement the API method. If there was just a network connection issue it would cause a request timeout and that's different.

Is this what you were asking?

TrustHenry:
And wait for the timeout until the error code for 404 appears.
What happens when this accumulates?

TrustHenry:
> 404 means the node does not implement the API method. If there was just a network connection issue it would cause a request timeout and that's different.
> 
> Is this what you were asking?

Yes, How about `403` or `405` `Method Not Allowed` with HTTP response code?

AndrejMitrovic:
> Yes, How about 403 or 405 Method Not Allowed with HTTP response code?

https://github.com/bpfkorea/vibe.d/blob/4076c01d438f39c358f976af69241243177c0d14/http/vibe/http/router.d#L27-L30

TrustHenry:
> > Yes, How about 403 or 405 Method Not Allowed with HTTP response code?
> 
> https://github.com/bpfkorea/vibe.d/blob/4076c01d438f39c358f976af69241243177c0d14/http/vibe/http/router.d#L27-L30

Oh, the default value is 404 which seems to be too common.

AndrejMitrovic:
You could do something like this:

```D
import vibe.http.common;

try
{
    pub_key = api.getPublicKey();
}
catch (HTTPStatusException ex)
{
    if (ex.status == 404)
    {
        // not a validator
    }
}
```

However, this will only work for system integration tests because they use vibe.d. When using LocalRest (in unittests) you will not get this exception thrown.

So we might have to change LocalRest. This code here: https://github.com/Geod24/localrest/blob/6da1455bb6793e76311acf4536be9506772e7372/source/geod24/LocalRest.d#L387

We could make it throw a custom exception type. For example we define `UnmatchedMethodException : Exception` in LocalRest, throw it, and then in Agora we could have:

```D
import vibe.http.common;

try
{
    pub_key = api.getPublicKey();
}
catch (HTTPStatusException ex)
{
    if (ex.status == 404)
    {
        // not a validator
    }
}
catch (UnhandledMethodException ex)
{
    // not a validator
}
```

Also seek input from @Geod24 because he's the LocalRest chief.

TrustHenry:
> We could make it throw a custom exception type. For example we define `UnmatchedMethodException : Exception` in LocalRest, throw it, and then in Agora we could have:
> 
> ```d
> import vibe.http.common;
> 
> try
> {
>     pub_key = api.getPublicKey();
> }
> catch (HTTPStatusException ex)
> {
>     if (ex.status == 404)
>     {
>         // not a validator
>     }
> }
> catch (UnhandledMethodException ex)
> {
>     // not a validator
> }
> ```
> 
> Also seek input from @Geod24 because he's the LocalRest chief.

@Geod24 ping

Geod24:
Yes I think that makes sense.

TrustHenry:
Is there a solution for `agora.cli` link error?
Do I have to delete the `extludedSourceFiles` of the `dub.json` in `client`?

```
Linking...
  "__D4scpd5types11Stellar_SCP12SCPQuorumSet11__xopEqualsFKxSQCdQCbQByQBoKxQpZb", referenced from:
      __D48TypeInfo_S4scpd5types11Stellar_SCP12SCPQuorumSet6__initZ in agora-client.o
  "__D4scpd5types11Stellar_SCP12SCPQuorumSet9__xtoHashFNbNeKxSQCeQCcQBzQBpZm", referenced from:
      __D48TypeInfo_S4scpd5types11Stellar_SCP12SCPQuorumSet6__initZ in agora-client.o
  "__D4scpd5types11Stellar_SCP12__ModuleInfoZ", referenced from:
      __D5agora3api9Validator12__ModuleInfoZ in agora-client.o
  "__D4scpd5types13Stellar_types9PublicKey11__xopEqualsFKxSQCbQBzQBwQBkKxQpZb", referenced from:
      __D46TypeInfo_S4scpd5types13Stellar_types9PublicKey6__initZ in agora-client.o
  "__D4scpd5types13Stellar_types9PublicKey9__xtoHashFNbNeKxSQCcQCaQBxQBlZm", referenced from:
      __D46TypeInfo_S4scpd5types13Stellar_types9PublicKey6__initZ in agora-client.o
  "__D5agora3api8FullNode12__ModuleInfoZ", referenced from:
      __D5agora6client13SendTxProcess12__ModuleInfoZ in agora-client.o
      __D5agora6client4main12__ModuleInfoZ in agora-client.o
      __D5agora3api9Validator12__ModuleInfoZ in agora-client.o
  "__D5agora3api8FullNode3API11__InterfaceZ", referenced from:
      __D5agora3api9Validator3API16__interfaceInfosZ in agora-client.o
      __D4vibe3web4rest__T19RestInterfaceClientTC5agora3api8FullNode3APIZQBv16__interfaceInfosZ in agora-client.o
  "void push_back<xdr::xvector<unsigned char, 4294967292u>, std::__1::vector<xdr::xvector<unsigned char, 4294967292u>, std::__1::allocator<xdr::xvector<unsigned char, 4294967292u> > > >(std::__1::vector<xdr::xvector<unsigned char, 4294967292u>, std::__1::allocator<xdr::xvector<unsigned char, 4294967292u> > >&, xdr::xvector<unsigned char, 4294967292u>&)", referenced from:
     
     __D4scpd3Cpp__T6vectorTSQv5types7XDRBase__T7xvectorThVki4294967292ZQyTSQCqQCo__T9allocatorTQCqZQpZQDf10fromStringFNfAyaZSQEoQEm__TQElTQEhTQCqZQEx in agora-client.o
  "void push_back<stellar::SCPQuorumSet, std::__1::vector<stellar::SCPQuorumSet, std::__1::allocator<stellar::SCPQuorumSet> > >(std::__1::vector<stellar::SCPQuorumSet, std::__1::allocator<stellar::SCPQuorumSet> >&, stellar::SCPQuorumSet&)", referenced from:
      __D4scpd3Cpp__T6vectorTSQv5types11Stellar_SCP12SCPQuorumSetTSQCgQCe__T9allocatorTQCgZQpZQCv10fromStringFNfAyaZSQEeQEc__TQEbTQDxTQCqZQEn in agora-client.o
  "void push_back<stellar::PublicKey, std::__1::vector<stellar::PublicKey, std::__1::allocator<stellar::PublicKey> > >(std::__1::vector<stellar::PublicKey, std::__1::allocator<stellar::PublicKey> >&, stellar::PublicKey&)", referenced from:
      __D4scpd3Cpp__T6vectorTSQv5types13Stellar_types9PublicKeyTSQCeQCc__T9allocatorTQCeZQpZQCt10fromStringFNfAyaZSQEcQEa__TQDzTQDvTQCqZQEl in agora-client.o
  "void push_back<unsigned char, std::__1::vector<unsigned char, std::__1::allocator<unsigned char> > >(std::__1::vector<unsigned char, std::__1::allocator<unsigned char> >&, unsigned char&)", referenced from:
      __D4scpd3Cpp__T6vectorThTSQxQu__T9allocatorThZQnZQBi10fromStringFNfAyaZSQCrQCp__TQCoThTQCkZQCy in agora-client.o
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

TrustHenry:
> Is there a solution for `agora.cli` link error?
> Do I have to delete the `extludedSourceFiles` of the `dub.json` in `client`?

This problem has been solved.
I deleted it from ExcludedSourceFiles. "source/agora/api/FullNode.d"

MukeunKim:
> > Is there a solution for `agora.cli` link error?
> > Do I have to delete the `extludedSourceFiles` of the `dub.json` in `client`?
> 
> This problem has been solved.
> I deleted it from ExcludedSourceFiles. "source/agora/api/FullNode.d"

MichaelKim20:
> Is there a solution for `agora.cli` link error?
> Do I have to delete the `extludedSourceFiles` of the `dub.json` in `client`?
> 
> ```
> Linking...
> Undefined symbols for architecture x86_64:
>   "__D4scpd5types11Stellar_SCP12SCPQuorumSet11__xopEqualsFKxSQCdQCbQByQBoKxQpZb", referenced from:
>       __D48TypeInfo_S4scpd5types11Stellar_SCP12SCPQuorumSet6__initZ in agora-client.o
>   "__D4scpd5types11Stellar_SCP12SCPQuorumSet9__xtoHashFNbNeKxSQCeQCcQBzQBpZm", referenced from:
>       __D48TypeInfo_S4scpd5types11Stellar_SCP12SCPQuorumSet6__initZ in agora-client.o
>   "__D4scpd5types11Stellar_SCP12__ModuleInfoZ", referenced from:
>       __D5agora3api9Validator12__ModuleInfoZ in agora-client.o
>   "__D4scpd5types13Stellar_types9PublicKey11__xopEqualsFKxSQCbQBzQBwQBkKxQpZb", referenced from:
>       __D46TypeInfo_S4scpd5types13Stellar_types9PublicKey6__initZ in agora-client.o
>   "__D4scpd5types13Stellar_types9PublicKey9__xtoHashFNbNeKxSQCcQCaQBxQBlZm", referenced from:
>       __D46TypeInfo_S4scpd5types13Stellar_types9PublicKey6__initZ in agora-client.o
>   "__D5agora3api8FullNode12__ModuleInfoZ", referenced from:
>       __D5agora6client13SendTxProcess12__ModuleInfoZ in agora-client.o
>       __D5agora6client4main12__ModuleInfoZ in agora-client.o
>       __D5agora3api9Validator12__ModuleInfoZ in agora-client.o
>   "__D5agora3api8FullNode3API11__InterfaceZ", referenced from:
>       __D5agora3api9Validator3API16__interfaceInfosZ in agora-client.o
>       __D4vibe3web4rest__T19RestInterfaceClientTC5agora3api8FullNode3APIZQBv16__interfaceInfosZ in agora-client.o
>   "void push_back<xdr::xvector<unsigned char, 4294967292u>, std::__1::vector<xdr::xvector<unsigned char, 4294967292u>, std::__1::allocator<xdr::xvector<unsigned char, 4294967292u> > > >(std::__1::vector<xdr::xvector<unsigned char, 4294967292u>, std::__1::allocator<xdr::xvector<unsigned char, 4294967292u> > >&, xdr::xvector<unsigned char, 4294967292u>&)", referenced from:
>      
>      __D4scpd3Cpp__T6vectorTSQv5types7XDRBase__T7xvectorThVki4294967292ZQyTSQCqQCo__T9allocatorTQCqZQpZQDf10fromStringFNfAyaZSQEoQEm__TQElTQEhTQCqZQEx in agora-client.o
>   "void push_back<stellar::SCPQuorumSet, std::__1::vector<stellar::SCPQuorumSet, std::__1::allocator<stellar::SCPQuorumSet> > >(std::__1::vector<stellar::SCPQuorumSet, std::__1::allocator<stellar::SCPQuorumSet> >&, stellar::SCPQuorumSet&)", referenced from:
>       __D4scpd3Cpp__T6vectorTSQv5types11Stellar_SCP12SCPQuorumSetTSQCgQCe__T9allocatorTQCgZQpZQCv10fromStringFNfAyaZSQEeQEc__TQEbTQDxTQCqZQEn in agora-client.o
>   "void push_back<stellar::PublicKey, std::__1::vector<stellar::PublicKey, std::__1::allocator<stellar::PublicKey> > >(std::__1::vector<stellar::PublicKey, std::__1::allocator<stellar::PublicKey> >&, stellar::PublicKey&)", referenced from:
>       __D4scpd3Cpp__T6vectorTSQv5types13Stellar_types9PublicKeyTSQCeQCc__T9allocatorTQCeZQpZQCt10fromStringFNfAyaZSQEcQEa__TQDzTQDvTQCqZQEl in agora-client.o
>   "void push_back<unsigned char, std::__1::vector<unsigned char, std::__1::allocator<unsigned char> > >(std::__1::vector<unsigned char, std::__1::allocator<unsigned char> >&, unsigned char&)", referenced from:
>       __D4scpd3Cpp__T6vectorThTSQxQu__T9allocatorThZQnZQBi10fromStringFNfAyaZSQCrQCp__TQCoThTQCkZQCy in agora-client.o
> ld: symbol(s) not found for architecture x86_64
> clang: error: linker command failed with exit code 1 (use -v to see invocation)
> ```

The cause of this error is when the SCP C++ library cannot be found.
SCP C++ library is required when referring to agora.api.Validator .

source/agora/node/FullNode.d
```diff
@@ -100,17 +98,25 @@ public class Node : API
...
+
+        Params:
+            config = Config instance
+            required_peer_keys = if not empty, the set of public keys for which
+                to find the associated nodes in the network and to connect to
+
+    ***************************************************************************/
+
+    public this (const Config config,
+        Set!PublicKey required_peer_keys = Set!PublicKey.init)
```
AndrejMitrovic, line 119: This is something I'd like to move out of the `FullNode` constructor, but I couldn't find a good solution yet.

AndrejMitrovic:
> node-3_1_8aa41cd509c7 | 2020-04-23T08:19:08.671319883Z Failed to parse config file 'config.yaml'. Error: Unable to open file config.yaml for YAML loading: config.yaml: No such file or directory

@TrustHenry the config file wasn't commited. I will push to your branch with the config file, ok?

AndrejMitrovic:
@TrustHenry you can `git fetch henry` or whatever your remote's name is. The file should be there now.

source/agora/api/FullNode.d
TrustHenry: This PR now spreads preimages, including full nodes.

TrustHenry:
It's working now
@bpfkorea/core Ready for review.

source/agora/cli/multi/main.d
```diff
@@ -104,9 +104,9 @@ private int main (string[] args)
         }
     }
 
-    Node[] nodes;
+    Validator[] nodes;
     foreach (const ref config; configs)
-        nodes ~= runNode(config);
+        nodes ~= runNode!Validator(config);
```
AndrejMitrovic: I think this is wrong. The configs that are passed can be both Validators and FullNodes. So the code will have to check each config and spawn each type. accordingly.

source/agora/network/NetworkManager.d
```diff
@@ -230,8 +236,9 @@ public class NetworkManager
                     ledger.getBlockHeight() + 1,
                     blocks => blocks.all!(block =>
```
AndrejMitrovic, line 237: We should think about fixing this in the future. It does not short-circuit.

Also there will be many `isNominating is null` checks in a tight loop here.

source/agora/node/Validator.d
```diff
@@ -0,0 +1,164 @@
...
+    protected Nominator nominator;
+
+    /// Ctor
+    public this (const Config config)
+    {
+        assert(config.node.is_validator);
+
+        // build the list of required quorum peers to connect to
+        Set!PublicKey required_peer_keys;
+        void getNodes (in QuorumConfig conf, ref Set!PublicKey nodes)
```
AndrejMitrovic: I recommend moving this into a private static method called `buildRequiredKeys`. The constructor code is hard to follow when it has a big nested function like this one inside.

Edit: It's also ok to leave this in a future PR to fix it up.

source/agora/node/Validator.d
```diff
@@ -0,0 +1,164 @@
...
+    {
+        return this.config.node.key_pair.address;
+    }
+
+    /***************************************************************************
+
+        Receive an SCP envelope.
+
+        API:
+            GET /envelope
```
AndrejMitrovic: Shouldn't this be `PUT`?

source/agora/node/main.d
```diff
@@ -70,7 +72,10 @@ private int main (string[] args)
             writefln("Config file '%s' succesfully parsed.", cmdln.config_path);
             return 0;
         }
-        runTask(() => node = runNode(config));
+        if (config.node.is_validator)
```
AndrejMitrovic, line 76:
```suggestion

        if (config.node.is_validator)
```

source/agora/node/main.d
```diff
@@ -61,7 +63,7 @@ private int main (string[] args)
             ex.message);
     }
 
-    Node node;
+    FullNode node;
```
AndrejMitrovic, line 66: I would rename this to the `API` from `agora.api.FullNode`. Because technically it can be a Validator, and using the interface makes it more explicit that this is not necessarily always a `FullNode` object.

source/agora/test/Base.d
```diff
@@ -43,9 +41,14 @@ import agora.consensus.EnrollmentManager;
 import agora.consensus.Genesis;
 import agora.network.NetworkManager;
 import agora.node.Ledger;
-import agora.node.Node;
+import agora.node.FullNode;
```
AndrejMitrovic, line 43: FullNode import should be moved before Ledger for alphabetic order.

source/agora/test/Base.d
```diff
@@ -560,6 +571,64 @@ public class TestNode : Node, TestAPI
...
+    }
+
+    public override PublicKey getPublicKey () @safe
+    {
+        // NetworkManager assumes that if key == PublicKey.init,
+        // we are *not* a Validator node, treated as a FullNode instead.
+        return PublicKey.init;
+    }
+
+    public override void receiveEnvelope(SCPEnvelope envelope) @safe
```
```suggestion
    public override void receiveEnvelope (SCPEnvelope envelope) @safe
```

And these last two methods need `///` comments.

source/agora/test/Base.d
```diff
@@ -560,6 +571,64 @@ public class TestNode : Node, TestAPI
...
+    {
+        assert(0);
+    }
+}
+
+/// A Validator which also implements test routines in TestAPI
+public class TestValidatorNode : Validator, TestAPI
+{
+    mixin TestNodeMixin!();
```
AndrejMitrovic, line 618:
```suggestion
    ///
    mixin TestNodeMixin!();
```


This will expand the mixin in the autogenerated documentation.

source/agora/test/Base.d
```diff
@@ -560,6 +571,64 @@ public class TestNode : Node, TestAPI
     {
         return new EnrollmentManager(":memory:", node_config.key_pair);
     }
+}
+
+/// A FullNode which also implements test routines in TestAPI
+public class TestFullNode : FullNode, TestAPI
+{
+    mixin TestNodeMixin!();
```
AndrejMitrovic, line 577:
```suggestion
    ///
    mixin TestNodeMixin!();
```

ci/system_integration_test.d
```diff
@@ -28,7 +28,8 @@ immutable DockerComposeLogs = [ "docker-compose", "-f", ComposeFile, "logs", "-t
 immutable RunIntegrationTests = [ "dub", "--root", IntegrationPath, "--",
                                     "http://127.0.0.1:4000",
                                     "http://127.0.0.1:4001",
-                                    "http://127.0.0.1:4002"];
+                                    "http://127.0.0.1:4002",
+                                    "http://127.0.0.1:4003"];
 immutable Cleanup = [ "rm", "-rf", IntegrationPath.buildPath("node/0/.cache/"),
                       IntegrationPath.buildPath("node/1/.cache/"),
                       IntegrationPath.buildPath("node/2/.cache/")];
```
MichaelKim20:
```suggestion
                      IntegrationPath.buildPath("node/2/.cache/"),                                       
                      IntegrationPath.buildPath("node/3/.cache/")];
```

MichaelKim20 (changes requested):
Good. It needs one change.

source/agora/cli/multi/main.d
MichaelKim20: The following files are also required when testing modifications.
ci/single_binary_test.d.
I think it needs to be corrected.

TrustHenry: applied the code checks each config and spawn each type.

source/agora/node/main.d
TrustHenry: This seems to require sufficient discussion.

TrustHenry: Currently, this the 66line refers to class `agora.node.FullNode`

source/agora/node/Validator.d
TrustHenry: Applied through additional commit.

TrustHenry:
I've applied suggestions up to here.
@bpfkorea/core Ready for review.

source/agora/node/FullNode.d
TrustHenry: Minor changes and names have been changed.

source/agora/node/main.d
AndrejMitrovic: Ok it was just a suggestion, feel free to overrule. :)

source/agora/node/Validator.d
```diff
@@ -0,0 +1,178 @@
...
+    /***************************************************************************
+
+        Build the list of required quorum peers to connect to
+
+        Supports recalling `sub_quorums` config structures.
+
+        Params:
+            conf = the node configuration
+            quorum_conf = the SCP quorum set configuration
+            nodes = Instance of Set!PublicKey The publickey of nodes
```
```suggestion
            nodes = will contain the set of public keys to connect to
```

AndrejMitrovic:
Just two comments. And the PR will need a rebase due to conflicts.

source/agora/node/Validator.d
TrustHenry: apply the suggestion.

source/agora/network/NetworkManager.d
AndrejMitrovic: We can think about fixing this in the future. It's somewhat unrelated to this PR.

AndrejMitrovic (approved):

tests/system/node/3/config.yaml
```diff
@@ -0,0 +1,63 @@
...
+  address: 0.0.0.0
+  port:    5826
+  retry_delay: 1.5
+  max_retries: 5
+  timeout: 1000
+  # This is a randomly generated keypair
+  # If this node is not a validator, this will be ignored
+  #
+  # DO NOT USE THOSE VALUES ANYWHERE
+  seed:    SAQEQ475K4BTKTW5RRZGX6NGRM37SOPIEM3OGVD2WNZMGZHIFDE65KQQ
```
Geod24: This can be removed

tests/system/node/3/config.yaml
```diff
@@ -0,0 +1,63 @@
...
+  max_failed_requests: 10
+  ban_duration: 86400
+
+################################################################################
+##                          Administrative interface                          ##
+################################################################################
+admin:
+  enabled: false
+  address: 127.0.0.1
+  port:    2827
```
Geod24: Remove

tests/system/node/3/config.yaml
```diff
@@ -0,0 +1,63 @@
...
+  - http://node-2:4826
+
+################################################################################
+##                               Quorum slices                                ##
+################################################################################
+quorum:
+  threshold: 66%
+  nodes:
+    - GBUVRIIBMHKC4PE6BK7MO2O26U2NJLW4WGGWKLAVLAA2DLFZTBHHKOEK
+
```
Geod24, line 50: That should be removed as well

source/agora/test/NetworkManager.d
```diff
@@ -110,15 +110,24 @@ unittest
         public override void createNewNode (PublicKey address, Config conf)
         {
             RemoteAPI!TestAPI api;
-            if (this.nodes.length == 2)
+
+            // the test has 3 nodes:
+            // 1 validator => used for creating blocks
+            // 1 byzantine FullNode => lies about the blockchain
```
Geod24, line 116:
```suggestion
            // 1 byzantine FullNode => lies about the blockchain (returns syntactically invalid data)
```

This is changing the meaning of this test a bit. Originally there were 2 validators and 1 byzantine node.
I don't mind though, but that might create issues with quorum sets.

source/agora/test/Base.d
```diff
@@ -560,6 +571,68 @@ public class TestNode : Node, TestAPI
...
+    public override Enrollment createEnrollmentData ()
+    {
+        assert(0);
+    }
+
+    /// FullNode does not implement this
+    public override void updateEnrolledHeight (Hash enroll_key, ulong height)
+    {
+        assert(0);
+    }
```
Geod24: Those methods shouldn't be available if they aren't implemented then

source/agora/node/FullNode.d
```diff
@@ -533,37 +450,15 @@ public class Node : API
...
-    node.start();  // asynchronous
+    ***************************************************************************/
 
-    log.info("About to listen to HTTP: {}", settings.port);
-    listenHTTP(settings, router);
-    return node;
+    protected void onAcceptedTransaction () @safe
+    {
+    }
 }
```
Geod24, line 466: Moving this could be its own commit

Geod24:
Looks pretty nice overall, just a few comments

source/agora/test/NetworkManager.d
AndrejMitrovic: Actually I think the documentation is correct, it's just that it was undocumented before. Maybe this change should go into its own commit.

tests/system/node/3/config.yaml
AndrejMitrovic: This will cause failures because the config parser always tries to parse quorum slices.

There is another issue about removing the quorum config from the config though.

TrustHenry: If it remove this, Clearly, quorum configuration analysis error occurs.

source/agora/node/FullNode.d
TrustHenry: The existing code use `this.nominator.tryNominate()` in `FullNode.putTransaction`
Because the Nominator has been moved from the corresponding refectoring, it must be handled by the same commit.

source/agora/test/Base.d
TrustHenry: `public class TestFullNode : FullNode, TestAPI`
Here, `TestAPI` is inheritance interface `TestAPI: ValidatorAPI`
I'd like to delete that method.
But, I think there is nothing we can do about the current TestAPI inheritance structure.
Do you have a better idea?

source/agora/test/NetworkManager.d
TrustHenry: Add it as own commit.

TrustHenry:
I applied Suggestion & replied
@Geod24 ping

MichaelKim20 (approved):

TrustHenry:
Rebase with the latest commit.

TrustHenry:
Rebased with the latest commit.

source/agora/node/FullNode.d
TrustHenry: Oh, I understood the point.
I separated own commit.

TrustHenry:
@Geod24 Ready for review

Geod24 (approved):

PR #758, opened 2020-04-17
Bug fix: get pre-image based on height

linked0:
The recent commits related to getting a preimage, the height had been supposed to be used for getting a preimage, but the code using the old way using a round of cycle as an index for getting a preimage had not changed. So I have corrected the wrong code.

This bug was found in doing peer programming with Mathias.


Geod24:
Could you add a test that would not pass before the fix but passes now ?

AndrejMitrovic (approved):

PR #760, opened 2020-04-17
Change derivation method for preimage generation

Geod24:
```
Make the derivation method match the signature noise approach:
the hash of a private piece of data, a unique string as salt, and an increasing index.
This makes the whole process easy to reproduce.
```

The tests currently have a lot of overlap with the code, and are quite rigid. I plan to refactor them in the future once other improvements make it in.


linked0 (approved):
Reasonable change!

PR #761, opened 2020-04-17
Add PreImageCache to handle sparsely populate preimage range

Geod24:
Super simple stuff, I plan to use this in place of the associative arrays for pre-image handling (both short and long range).

### DIFF
--- a/source/agora/common/Hash.d
+++ b/source/agora/common/Hash.d
@@ -289,3 +289,185 @@ nothrow @nogc @safe unittest
     auto hash_2 = hashMulti(420, "bpfk", S('a', 1, 'b', 2, 'c', 3));
     assert(hash_1 == hash_2);
 }
+
+/*******************************************************************************
+
+    Helper struct to manage a range of pre-images
+
+    When dealing with pre-images, the only unrecoverable data is the seed.
+    Everything else is, by definition, derived from the seed.
+    As a result, we can compress the data as much as we want,
+    to reduce memory usage. However, this comes at the cost of more computation.
+
+    For example, if we have cycles of 10 pre-images, we can first compute the
+    seed (H0), then reveal our initial commitment (H9). This will cost us
+    9 hash operations. Then on the next reveal, in order to get H8, we need to
+    perform 8 hash operations, then 7 for H7, etc...
+
+    If we want to improve this, we can store intermediate results:
+    when generating H9, we do perform 9 hash operations, but store H5.
+    Hence, when generating H8 and H7, we only need to perform
+    3 and 2 hash operations, respectively, and will never need to perform more
+    than 4 operations (save the initial commitment).
+
+    This structure is a mean to generalize this approach, with arbitrary count
+    and arbitrary sample size (interval, or distance, between saved pre-images).
+
+*******************************************************************************/
+
+public struct PreImageCache
+{
+    @safe nothrow:
+
+    /// Store the actual preimages
+    private Hash[] data;
+
+    /// Interval between two preimages in `data`
+    private const ulong interval;
+
+    /// Construct an instance using already existing data
+    public this (inout(Hash)[] data_, ulong sample_size) inout @nogc pure
+    {
+        this.data = data_;
+        this.interval = sample_size;
+    }
+
+    /***************************************************************************
+
+        Construct an instance and allocate memory
+
+        This takes a `count` of pre-image and a sample size, or distance.
+        For example, if one wish to represent a range of 1000 pre-images with
+        this cache, and perform no more than 5 hash operations each time,
+        the `sample_size` should be `5` and the `count` should be `200`
+        (1000 / 5).
+
+        Params:
+          count = Number of samples to store (number of entries in the array)
+          sample_size = Distance between two pre-images
+
+    ***************************************************************************/
+
+    public this (ulong count, ulong sample_size) pure
+    {
+        assert(count > 1, "A count of less than 2 does not make sense");
+        assert(sample_size > 1, "A distance of less than 2 does not make sense");
+
+        this.interval = sample_size;
+        this.data = new Hash[](count);
+    }
+
+    /***************************************************************************
+
+        Populate this cache from the `seed`
+
+        Params:
+          seed = Initial value to derive preimage from. This will always be the
+                 first value of the array.
+          length = The number of entries to populate.
+                   This can be used when large sample size are used,
+                   and one wish to stop initialization past a certain threshold.
+
+        Returns:
+          The result of hashing seed `sample_size * (count + 1) - 1` times
+          if `length == 0`, or `sample_size * (length + 1) - 1` otherwise,
+          provided `length <= count`.
+
+    ***************************************************************************/
+
+    public Hash reset (Hash seed, size_t length)
+    {
+        // Set the length, so that extra entries are not accessible through
+        // `opIndex`
+        assert(length > 0, "The length of the array should be at least 1");
+        this.data.length = length;
+        () @trusted { assumeSafeAppend(this.data); }();
+        return this.reset(seed);
+    }
+
+    /// Ditto
+    public Hash reset (Hash seed) @nogc
+    {
+        size_t count = 0;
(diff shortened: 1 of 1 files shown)


source/agora/common/Hash.d
```diff
@@ -289,3 +289,182 @@ nothrow @nogc @safe unittest
...
+
+    /***************************************************************************
+
+        Construct an instance and allocate memory
+
+        This takes a `count` of pre-image and a sample size, or distance.
+        For example, if one wish to store 1000 pre-images in this cache,
+        and perform no more than 5 hash operations each time,
+        the `sample_size` should be `5` and the `count` should be `200`
+        (1000 / 5).
```
AndrejMitrovic, line 343: > if one wish to store 1000 pre-images in this cache

The wording threw me off a bit. I think you meant if one wishes to _use_ 1000 preimages. It only physically stores `count` number of images.

source/agora/common/Hash.d
```diff
@@ -289,3 +289,182 @@ nothrow @nogc @safe unittest
...
+    }
+
+    /***************************************************************************
+
+        Construct an instance and allocate memory
+
+        This takes a `count` of pre-image and a sample size, or distance.
+        For example, if one wish to store 1000 pre-images in this cache,
+        and perform no more than 5 hash operations each time,
+        the `sample_size` should be `5` and the `count` should be `200`
```
AndrejMitrovic, line 342: What happens when `sample_size % count != 0`? I see the unittests are only testing with powers of 2. I'm wondering about edge-cases.

source/agora/common/Hash.d
```diff
@@ -289,3 +289,182 @@ nothrow @nogc @safe unittest
...
+
+    ***************************************************************************/
+
+    public this (ulong count, ulong sample_size) pure
+    {
+        assert(count > 1, "A count of less than 2 does not make sense");
+        assert(sample_size > 1, "A distance of less than 2 does not make sense");
+
+        this.interval = sample_size;
+        this.data = new Hash[](count);
```
AndrejMitrovic, line 357: We could make this `@nogc` in the future.

`PreimageCache` is a struct, and it's not exposing the internal array to the client code,  `opIndex` returns a value (a copy of a POD).

But it's more work, and would require reference counting and a destructor to deallocate the array.

source/agora/common/Hash.d
```diff
@@ -289,3 +289,182 @@ nothrow @nogc @safe unittest
...
+
+        Populate this cache from the `seed`
+
+        Params:
+          seed = Initial value to derive preimage from. This will always be the
+                 first value of the array.
+          length = The number of entries to populate.
+                   This can be used when large sample size are used,
+                   and one wish to stop initialization past a certain threshold.
+                   The special value `0` means "everything", and is the default.
```
AndrejMitrovic: I wonder if overloads would be better fit for this? For example:

```D
/// Returns: The result of hashing seed `sample_size * (count + 1) - 1`
public Hash reset (Hash seed)
{
    return this.reset(seed, this.data.length);
}

/// Returns: The result of hashing seed `sample_size * (length + 1) - 1`
public Hash reset (Hash seed, size_t length)
{
    ...
}
```

It's a little easier to reason about having two different functions rather than 1 parameter that has a different meaning based on its value.

source/agora/common/Hash.d
```diff
@@ -289,3 +289,182 @@ nothrow @nogc @safe unittest
...
+            () @trusted { assumeSafeAppend(this.data); }();
+        }
+        size_t count = 0;
+        immutable end = this.data.length * this.interval - 1;
+        foreach (idx, ref entry; this.data)
+        {
+            entry = seed;
+            do
+                seed = hashFull(seed);
+            while (++count % end % this.interval);
```
AndrejMitrovic: The association is always left to right in D, correct? `A % B % C` => `(A % B) % C`?

You can leave it as-is if it should be obvious.

source/agora/common/Hash.d
```diff
@@ -289,3 +289,182 @@ nothrow @nogc @safe unittest
...
+        // array by just taking the computational hit, since indexing past the
+        // end means we are generating new pre-images, but it sounds like abuse
+        Hash value = this.data[startIndex];
+        foreach (_; (startIndex * this.interval) .. index)
+            value = hashFull(value);
+        return value;
+    }
+}
+
+///
```
AndrejMitrovic, line 425: I was just thinking about this. The documented examples would ideally show a use-case. Here they're just elaborate tests that try to do full coverage.

But this is true right now for most of our tests. In the future we might want to provide better _documented_ examples.

AndrejMitrovic (commented):
Just a few comments.

source/agora/common/Hash.d
Geod24: "represent a range of 100 pre-images with this cache" ?

AndrejMitrovic: Good!

source/agora/common/Hash.d
Geod24: I originally wrote it to handle every edge case one could throw at it, then realized it was stupid and unnecessary. Originally the code did a division, now it's only doing a multiplication, so the sample_size is not exposed in the API, only the count and the interval, hence the case you mentioned is impossible. The reason to not support this is simply that the intent of this cache is to provide a predictable behavior (maximum number of hash operations), and silently ignoring what is likely to be user error doesn't sound like a feature.

I should probably add a test with a non-power of 2 test (e.g. sample=30 interval=3) tho.

source/agora/common/Hash.d
Geod24: We could, but I think we are going to use those caches in a long-lived manner. E.g. we'll allocate once at the beginning, and progressively shrink them as the pre-images are exhausted, and suddenly resize them to their max value after 19ish years.

source/agora/common/Hash.d
Geod24: Correct

source/agora/common/Hash.d
Geod24: Good point. I wasn't very happy with this API, and your suggestion looks very good. One point though: The `reset(Hash seed)` should be called by `reset(Hash, size_t)`, this way the former can be made `@nogc`.

source/agora/common/Hash.d
Geod24: Oh right. Yeah definitely. Should we open an issue to "curate" the documentation ? TestNet milestone plz.

Geod24:
Rebased on v0.x.x, addressed review.

source/agora/common/Hash.d
```diff
@@ -289,3 +289,184 @@ nothrow @nogc @safe unittest
...
+    {
+        // Set the length, so that extra entries are not accessible through
+        // `opIndex`
+        assert(length > 0, "The length of the array should be at least 1");
+        this.data.length = length;
+        () @trusted { assumeSafeAppend(this.data); }();
+        return this.reset(seed);
+    }
+
+    public Hash reset (Hash seed) @nogc
```
AndrejMitrovic, line 389: No `/// Ditto` on purpose?

Geod24: Oops, fixed

AndrejMitrovic (approved):

source/agora/common/Hash.d
```diff
@@ -289,3 +289,185 @@ nothrow @nogc @safe unittest
...
+    9 hash operations. Then on the next reveal, in order to get H8, we need to
+    perform 8 hash operations, then 7 for H7, etc...
+
+    If we want to improve this, we can store intermediate results:
+    when generating H9, we do perform 9 hash operations, but store H5.
+    Hence, when generating H8 and H7, we only need to perform
+    3 and 2 hash operations, respectively, and will never need to perform more
+    than 4 operations (save the initial commitment).
+
+    This structure is a mean to generalize this approach, with arbitrary count
```
TrustHenry, line 313:
```suggestion
    This structure is a means to generalize this approach, with arbitrary count
```

PR #765, opened 2020-04-17
Change PreImageInfo.height to distance

Geod24:
```
We should not use absolute height values for pre-image,
instead we should use the 'distance' from the commitment.
```


linked0 (approved):
LGTM.
From this commit, we should start to apply the concept of a sliding window of pre-images.

PR #770, opened 2020-04-22
Remove unneeded `block_height` from `Enrollment` validation

linked0:
There is no need to use `block_height` because the code about comparing with the `unlock_height` is always true.
If an UTXO is frozen, its `unlock_ height` is always 0, hence the check `block_height < unlock_height` will always be true.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -137,7 +137,6 @@ public class EnrollmentManager
         Add a enrollment data to the validators set
 
         Params:
-            block_height = the current block height in the ledger
             finder = the delegate to find UTXOs with
             enroll = the enrollment data to add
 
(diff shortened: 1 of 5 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -146,10 +145,10 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public bool add (ulong block_height, scope UTXOFinder finder,
-        const ref Enrollment enroll) @safe nothrow
+    public bool add (scope UTXOFinder finder, const ref Enrollment enroll)
+        @safe nothrow
     {
-        return this.validator_set.add(block_height, finder, enroll);
+        return this.validator_set.add(0, finder, enroll);
```
AndrejMitrovic, line 151: The `block_height` parameter should also be removed from `ValidatorSet.add`, it's unused now.

AndrejMitrovic (dismissed):

source/agora/consensus/EnrollmentManager.d
Geod24: Discussed IRL, this will be addressed here: https://github.com/bpfkorea/agora/pull/748/files#diff-a1aa02c9e0b910c82366388fb0900864R154-R158

Geod24 (approved):

PR #776, opened 2020-04-28
Rename updateEnrolledHeight to addValidator

linked0:
As we are splitting the EnrollmentManager into EnrollmentPool and ValidatorSet, this function had to be renamed to fit its new purpose.
Before, enrollments that were not in the chain would have an enrolled height of 0, and when they appeared in the chain, the actual enrolled height would be set through this function.
Soon the storage will be separate, and instead, this function will just add new enrollments and clear up the pool.
Additionally, this fixes a bug, where catchup didn't work when an Enrollment seen in a block was not found in the Enrollment pool.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -197,10 +197,11 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Update the enrolled height of the validatorSet DB.
+        Add a validator to the validator set or update the enrolled height.
 
         Params:
             enroll_hash = enrollment blockheight to update enroll hash
+            finder = the delegate to find UTXOs with
             block_height = enrolled blockheight
 
         Returns:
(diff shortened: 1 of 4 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -201,21 +201,26 @@ public class EnrollmentManager
 
         Params:
             enroll_hash = enrollment blockheight to update enroll hash
```
TrustHenry, line 203: Need to modify the description.

source/agora/node/Ledger.d
```diff
@@ -229,8 +229,8 @@ public class Ledger
             assert(0);
 
         foreach (enrollment; block.header.enrollments)
-            if (!this.enroll_man.updateEnrolledHeight(enrollment.utxo_key,
-                block.header.height))
+            if (!this.enroll_man.addValidator(enrollment,
+                this.utxo_set.getUTXOFinder(), block.header.height))
```
TrustHenry, line 232: This change also requires you to change the description of the function.

linked0:
I applied @TrustHenry 's comments to the code. Thank you.

Geod24 (approved):

source/agora/consensus/EnrollmentManager.d
```diff
@@ -197,25 +197,30 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Update the enrolled height of the validatorSet DB.
+        Add a validator to the validator set or update the enrolled height.
 
         Params:
             enroll_hash = enrollment blockheight to update enroll hash
```
TrustHenry, line 203: `enroll_hash = enrollment blockheight to update enroll hash`
 It has been changed to `Enrollment enroll`
It has not been changed but became Merge.

source/agora/consensus/EnrollmentManager.d
Geod24: Sorry didn't see this!

source/agora/consensus/EnrollmentManager.d
Geod24: https://github.com/bpfkorea/agora/pull/781

PR #777, opened 2020-04-28
Refactor TransactionPool a bit

AndrejMitrovic:


Geod24:
A bit??

AndrejMitrovic:
Is there a way to wrap a range into another range and make the dtor @safe/@trusted?

It seems the `ResultRange` from the sqlite bindings has an unsafe destructor.

AndrejMitrovic:
Added another commit. `Use the tx pool range in its constructor`

source/agora/common/TransactionPool.d
```diff
@@ -193,32 +194,17 @@ public class TransactionPool
...
-            dg = the delegate to call
-
         Returns:
-            the status code of the delegate, or zero
+            an input range of (hash, transaction) tuples
 
     ***************************************************************************/
 
-    public int opApply (scope int delegate(Hash hash, Transaction tx) dg) @trusted
+    public auto opSlice () @trusted
```
Geod24: If the resulting range is not safe, isn't it better to not marked this as `@trusted` ?

source/agora/node/Ledger.d
```diff
@@ -283,19 +283,19 @@ public class Ledger
 
         this.enroll_man.getUnregistered(data.enrolls);
 
-        foreach (hash, tx; this.pool)
-        {
-            if (auto reason = tx.isInvalidReason(utxo_finder, next_height))
-                log.trace("Rejected invalid ('{}') tx: {}", reason, tx);
-            else
-                data.tx_set.put(tx);
+        () @trusted {
```
Geod24: I'm not sure that's an enhancement to be fair...

Geod24 (commented):
Not so convince about the range change. In spirit it's nice, but the result is, the calling code looks way worse.

source/agora/common/TransactionPool.d
AndrejMitrovic: Well the method itself is trusted, but the returned type is not. I wish I could do `@trusted(auto)` hah.

AndrejMitrovic:
> Not so convince about the range change. In spirit it's nice, but the result is, the calling code looks way worse.

Yep it's because of the dtor.

Actually I wanted to try to do copy ellision. Right now the opApply copies hashes / txes.

AndrejMitrovic:
I'll revert the range change for now.

MichaelKim20 (approved):
LGTM

MichaelKim20:
I changed the order of the source code, rearranged it, and looked at it one by one. 
LGTM

Geod24 (approved):

PR #782, opened 2020-04-28
Fix incorrectly described API documentation

TrustHenry:
the method and URI parts will be inferred from the method name by looking for a known prefix
No get is required before the function name of the interface endpoint.


linked0 (approved):
Nice catch!
LGTM.

Geod24 (approved):

PR #783, opened 2020-04-29
Remove `hasEnrollment` API in the FullNode

linked0:
The `hasEnrollment` function is kind of useless if it's used to check if an enrollment data exists before sending an enrollment. It makes unnecessary communication between nodes. So it's simple to just send enrollment information. And we add the `getEnrollment` API to replace the use of the hasEnrollment.

Additionally, the `hasEnrollment` function became private because the `hasEnrollment` API doesn't exist anymore.

Mathias and I discussed removing the API in the pair programming session.

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -189,21 +189,21 @@ public interface API
 
     /***************************************************************************
 
-        Check if an enrollment data exists in the validator set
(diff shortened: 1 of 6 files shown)


AndrejMitrovic (approved):

Geod24 (approved):
Nice!

PR #784, opened 2020-04-29
Fix SCPEnvelope serialization to produce valid JSON

Geod24:
Before it was producing a weird JSON hybrid that was not parseable
by anything but the fromString method.

### DIFF
--- a/source/scpd/types/Stellar_SCP.d
+++ b/source/scpd/types/Stellar_SCP.d
@@ -104,45 +104,54 @@ struct SCPStatement {
         /// Support (de)serialization from Vibe.d
         extern(D) string toString () const @trusted
         {
+            Json json = Json.emptyObject;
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
LGTM. `(*obj)` can just be `obj` when accessing properties, but I guess you left it on purpose?

AndrejMitrovic (approved):

Geod24:
`deserializeJson` is not a property, it's using UFCS. Currently it accepts a `Json` so it would work, but I'm a bit wary of templates (as soon as you have templates, the auto-dereferencing just falls appart).

TrustHenry:
Github actions CI can't check ??

Geod24:
Yeah IDK why it didn't show up. Rebased and now it's there.

Issue #791, opened 2020-05-06
buildMerkleTree should assert on empty txs


AndrejMitrovic:
```
auto merkle_root = Block.buildMerkleTree(txs, merkle_tree);
```

If you pass a null array to this, you get:

```
core.exception.OutOfMemoryError@core/exception.d(647): Memory allocation failed
```

This should be caught before.

PR #794, opened 2020-05-06
Add the ability to check if `txs` is null, when calling `buildMerkleTree`

MichaelKim20:
Use 'Function contacts' to verify the parameter `txs`

Relates to #791


AndrejMitrovic (approved):

AndrejMitrovic:
In the future it's better to write `Fixes #791` rather than `Relates to #791`, because otherwise #791 will not be automatically closed.

Issue #739, opened 2020-04-07
Fix link errors when running unit test & integrated test in Windows OS


TrustHenry:
An error occurs when running the unit test and integrated test in the current windows version.

In the unit test, one link error is occurring due to MSVC mangling problems.

```
agora-unittests.obj : error LNK2019: unresolved external symbol 
"void __cdecl push_back<struct xdr::xvector<unsigned char,4294967292>,
struct xdr::xvector<struct xdr::xvector<unsigned char,4294967292>,4294967292> >
(struct xdr::xvector<struct xdr::xvector<unsigned char,4294967292>,4294967292> 
&,struct xvector<unsigned char,4294967292>::xvector<unsigned char,4294967292> &)" 
(??$push_back@U?$xvector@E$0PPPPPPPM@@xdr@@U?$xvector@U?$xvector@E$0PPPPPPPM@@xdr@@$0PPPPPPPM@@2@@@YAXAEAU?$xvector@U?$xvector@E$0PPPPPPPM@@xdr@@$0PPPPPPPM@@xdr@@AEAU?$xvector@E$0PPPPPPPM@@2@@Z) 
referenced in function _D5agora5utils16SCPPrettyPrinter18__unittest_L415_C1FZv
```

Errors when running the integration test.

```
Running test on d:\a\agora\agora\.\tests\unit\BlockStorageChecksum.d
[stderr]	

[stderr]	std.exception.ErrnoException@C:\hostedtoolcache\windows\dc\ldc2-ff36be2c\x64\ldc2-ff36be2c-windows-multilib\bin\..\import\std\stdio.d(1177): Wrote 0 instead of 1 objects of type ubyte to file `C:\Users\RUNNER~1\AppData\Local\Temp\agora_testing_framework\unit.BlockStorageChecksum\B000000000000.dat' (No error)
```

This causes problems in the CI test.

Definition done

- [x] The Test CI should end normally without errors.
- [x] Windows build should be built and terminated without errors.

TrustHenry:
windows
```
void __cdecl push_back<struct xdr::xvector<unsigned char,-4>,struct xdr::xvector<struct xdr::xvector<unsigned char,-4>,-4> >(struct xdr::xvector<struct xdr::xvector<unsigned char,-4>,-4> & __ptr64,struct xvector<unsigned char,-4>::xvector<unsigned char,-4> & __ptr64)
```
mac
```
[...]

PR #799, opened 2020-05-07
Fix rawWrite of checksum test in Windows

TrustHenry:
The setmode of the file is r+b.
However, when you switch between reading and writing, there must be an intervening fflush, fsetpos, fseek, or rewind operation
Only for windows.
But it works normally on Posix, too.
Relates to #739

### DIFF
--- a/tests/unit/BlockStorageChecksum.d
+++ b/tests/unit/BlockStorageChecksum.d
@@ -79,6 +79,7 @@ private void corruptBlocks (string dir_path)
 
         // write a modified byte
         bytes[0]++;
+        block_file.seek(5, SEEK_SET);
         block_file.rawWrite(bytes);
     }
 }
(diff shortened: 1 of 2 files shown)


tests/unit/BlockStorageChecksum.d
```diff
@@ -79,6 +79,7 @@ private void corruptBlocks (string dir_path)
 
         // write a modified byte
         bytes[0]++;
+        block_file.seek(5, SEEK_SET);
```
MichaelKim20:
```suggestion
        block_file.seek(ChecksumSize+1, SEEK_SET);
```

MichaelKim20 (changes requested):
I request one modification.

tests/unit/BlockStorageChecksum.d
TrustHenry: Applied suggestion. make a new commit.

Geod24:
Would it work with just a `fflush` ?

MichaelKim20 (approved):
LGTM

TrustHenry:
> Would it work with just a `fflush` ?

No, it doesn't work.

Issue #785, opened 2020-04-29
Implement shell quorum balancing


Geod24:
As discussed in today's white boarding session, our best approach to the quorum balancing issue (#240 ) is to start with a "shell" algorithm which simply output a single quorum.

Definition of done:
- [x] There is no user-visible quorum configuration;
- [x] The quorum configuration is derived from the state of the chain;
- [x] The output of the shell QB algorithm is a single quorum with all enrolled validators;

PR #802, opened 2020-05-07
Simplify createNewNode()

AndrejMitrovic:
The address was already accessible through the public key.

Simplification as part of #785

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -44,6 +44,7 @@ import agora.node.FullNode;
 import agora.node.Ledger;
 import agora.node.Validator;
 import agora.utils.Log;
+import agora.utils.PrettyPrinter;
 public import agora.utils.Test;  // frequently needed in tests
 import agora.api.FullNode : NodeInfo, NetworkState;
 import agora.api.Validator : ValidatorAPI = API;
@@ -241,7 +242,7 @@ public class FakeClockBanManager : BanManager
 public struct NodePair
 {
     ///
-    public PublicKey key;
+    public Address address;
 
     ///
     public RemoteAPI!TestAPI client;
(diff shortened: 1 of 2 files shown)


Geod24:
What about full nodes ?

AndrejMitrovic:
Good catch. This means currently we still generate a keypair for full nodes. This will need to be fixed.

AndrejMitrovic:
I'd even say that we should remove `is_validator`. If there is a `key_pair`, the node is a validator.

TrustHenry:
> I'd even say that we should remove `is_validator`. If there is a `key_pair`, the node is a validator.

if there is a key_pair, I think it can be a validator.
But What if a node wants to stop the consensus when it's a validator?
Some reasons (upgrade, maintenance, withdrawal of funds, etc.)

AndrejMitrovic:
Good thinking @TrustHenry.

AndrejMitrovic:
Updated. I've simplified it again, improved the addressing and refactored some code. It should be much better and more correct.

The logs are much nicer now. For example:

```
Log for node Validator #2 (GA44...6FRQ):
======================================================================
2020-05-08 10:55:17,500 Info [agora.node.FullNode] - Doing network discovery..
2020-05-08 10:55:17,500 Trace [agora.common.BanManager] - BanManager: Address Validator #2 (GA44...6FRQ) banned until 9223372036854775807
2020-05-08 10:55:17,500 Info [agora.network.NetworkManager] - Discovering from { r: { impl: 0X000000011106C8C0, idx: 1 } }
2020-05-08 10:55:17,500 Info [agora.network.NetworkManager] - Connecting to next set of addresses: ["Validator #1 (GBZX...O3WC)", "Validator #0 (GAQJ...GBDP)"]
```

And for full nodes I just use a sequence without the public key:

```
Log for node FullNode #0:
======================================================================
2020-05-08 10:59:21,78 [32mInfo[0m [agora.node.FullNode] - Doing network discovery..
```

This should make it much easier to debug test-suite failures.

Geod24 (approved):
Sweeeeeeeeet

AndrejMitrovic:
There was a bug in my code. I added another commit which makes it easier to see which logs are being printed out. It helped me find the bug.

TrustHenry (approved):
Nice! It has become easier to distinguish nodes.

PR #804, opened 2020-05-08
Small fix for logging & usage of a pointer

AndrejMitrovic:
I had a bug where I used a variable on the stack for the genesis block. Woops.

### DIFF
--- a/source/agora/consensus/Genesis.d
+++ b/source/agora/consensus/Genesis.d
@@ -49,14 +49,16 @@ public ref immutable(Block) GenesisBlock () nothrow @safe @nogc
 *******************************************************************************/
 
 pragma(inline, true)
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
Maybe in the future we could find a better design for setting that genesis block.

source/agora/consensus/Genesis.d
```diff
@@ -49,14 +49,14 @@ public ref immutable(Block) GenesisBlock () nothrow @safe @nogc
 *******************************************************************************/
 
 pragma(inline, true)
-public void setGenesisBlock (ref immutable Block block)
+public void setGenesisBlock (immutable Block* block)
```
Geod24, line 52: Should we assert for non-null ?

Geod24 (approved):

Issue #744, opened 2020-04-08
Change SCP's Signature definition to be a fixed-size type


AndrejMitrovic:
Currently it's implemented as an std::vector: `opaque_vec!64`. It seems rather pointless to have to allocate memory just in order to be able to pass pointers to it, when the referenced memory should always be 64 bytes (in its current implementation).

We should replace this definition with a fixed-length array, matching the one in `agora.common.Types`, which is a `BitBlob!512`, in other words fixed-length array of 64 bytes.

This is blocking #548.

Geod24:
Do you know why it's this way ? Did upstream want to abstract the signature size, originally ? That's my best guess.

AndrejMitrovic:
I wish I knew why. It really doesn't seem efficient to do it this way. But maybe they are more explicit with acquire/release of memory and they don't want to increase the size of the SCPStatement structure?

MichaelKim20:
I have confirmed that the declaration was made as follows.
``` C++
using Signature = xdr::opaque_vec<64>;
```
I will change it as follows.
``` C++
using Signature = xdr::opaque_array<64>;
```

MichaelKim20:
@AndrejMitrovic Do you think this is completely resolved now?

AndrejMitrovic:
Yes, feel free to close it.

PR #806, opened 2020-05-08
Change SCP's Signature definition to be a fixed-size type

MichaelKim20:
I changed `Signature` to fixed length array.

``` C++
using Signature = xdr::opaque_vec<64>;
```
I changed it as follows.
``` C++
using Signature = xdr::opaque_array<64>;
```

`Signature` is the data structure used by SCP.
This is declared vector in SCP because of compatibility between versions.
SCP has several versions. In some of the previous versions, Signature was 32 Byte and is now 64 Byte.

For this reason, I think there is no problem if we change to 64Byte fixed length.

And I tried to compile it after changing it in `stellar-core`.
There were some errors in the test codes, but there were no problems with the rest.
[

https://github.com/stellar/stellar-core/blob/69c75632ae903191a82c7b2b010cce4022a60870/src/crypto/SecretKey.cpp#L128

https://github.com/stellar/stellar-core/blob/69c75632ae903191a82c7b2b010cce4022a60870/src/transactions/test/TxEnvelopeTests.cpp#L83-L101

https://github.com/stellar/stellar-core/blob/69c75632ae903191a82c7b2b010cce4022a60870/src/herder/test/HerderTests.cpp#L1262-L1266
]
Relates to #744

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -772,15 +772,12 @@ private struct SCPEnvelopeHash
         "4039907a44d671dbffe55ce9ae21f8eca7d218e6c87573c381ae20d96bf4a56"),
     getEnvHash().hashFull().to!string);
 
-    alias Sig = agora.common.Types.Signature;
-    Sig sig;
     () @trusted
     {
         auto seed = "SAI4SRN2U6UQ32FXNYZSXA5OIO6BYTJMBFHJKX774IGS2RHQ7DOEW5SJ";
         auto pair = KeyPair.fromSeed(Seed.fromString(seed));
         auto msg = getStHash().hashFull();
-        sig = pair.secret.sign(msg[]);
-        env.signature = sig;
+        env.signature = pair.secret.sign(msg[]);
     }();
 
     // with a signature
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -761,26 +761,26 @@ private struct SCPEnvelopeHash
     // empty envelope
     import std.conv;
     assert(getEnvHash().hashFull() == Hash.fromString(
-        "0xd3dc2318365e55ea3a62b403fcbe22d447402741a5151a703326dbf852350dcd0" ~
-        "e020a762ae12a871473aed80d82f51c1cd3942c0b2360c2d609279a2867fe68"),
+        "0xfd2ea2b85d2a315a9817e6661bc3c4378637de37649d2fdb9ca82d6e4172e9e46" ~
+        "af5a57113cfb7cb09d25eb7b4518eca9930c57231a29ffa396661822603c509"),
     getEnvHash().hashFull().to!string);
 
     // with a statement
     env.statement = st;
     assert(getEnvHash().hashFull() == Hash.fromString(
-        "0xbba4bdee0e083e6e5f56ddc2815afcd509f597f45d6ae5c83af747de2d568a26d" ~
-        "bc1f0792c8c6f990816bf9f2fc913ccc700c0a022644f8bd25835a6b439944c"),
+        "0x45d6b9adba8da9f2763f33960d7cd77b6c7e844fc11b0c7d793dfa47c99bc4377" ~
+        "4039907a44d671dbffe55ce9ae21f8eca7d218e6c87573c381ae20d96bf4a56"),
     getEnvHash().hashFull().to!string);
 
     alias Sig = agora.common.Types.Signature;
-    Sig sig;  // be warned: toVec() will point to stack-allocated memory
+    Sig sig;
     () @trusted
     {
         auto seed = "SAI4SRN2U6UQ32FXNYZSXA5OIO6BYTJMBFHJKX774IGS2RHQ7DOEW5SJ";
         auto pair = KeyPair.fromSeed(Seed.fromString(seed));
         auto msg = getStHash().hashFull();
         sig = pair.secret.sign(msg[]);
-        env.signature = sig[].toVec();
+        env.signature = sig;
     }();
 
     // with a signature
--- a/source/agora/utils/SCPPrettyPrinter.d
+++ b/source/agora/utils/SCPPrettyPrinter.d
@@ -354,14 +354,10 @@ private struct SCPEnvelopeFmt
     {
         try
         {
-            Signature sig;
-            if (this.envelope.signature[].length == Signature.Width)
-                sig = Signature(this.envelope.signature[]);
-
             formattedWrite(sink,
                 "{ statement: %s, sig: %s }",
                 SCPStatementFmt(this.envelope.statement, this.getQSet),
-                prettify(sig));
+                prettify(this.envelope.signature));
         }
         catch (Exception ex)
         {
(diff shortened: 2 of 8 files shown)


source/agora/utils/SCPPrettyPrinter.d
```diff
@@ -459,7 +459,7 @@ GCOQ...LRIJ(62,500,000), GCOQ...LRIJ(62,500,000)], enrolls: [{ utxo: 0x0000...e2
     assert(MissingSig == format("%s", scpPrettify(&env)),
```
AndrejMitrovic, line 455: In this module there is a warning:

> // be warned: toVec() will point to stack-allocated memory

This warning should now be removed. Actually there is a call to `toVec()` in `env.signature = sig[].toVec();`, so the `toVec()` call should be removed too. It should be just `env.signature = pair.secret.sign(msg[]);`

source/agora/consensus/protocol/Nominator.d
```diff
@@ -761,26 +761,23 @@ private struct SCPEnvelopeHash
     // empty envelope
     import std.conv;
     assert(getEnvHash().hashFull() == Hash.fromString(
-        "0xd3dc2318365e55ea3a62b403fcbe22d447402741a5151a703326dbf852350dcd0" ~
-        "e020a762ae12a871473aed80d82f51c1cd3942c0b2360c2d609279a2867fe68"),
+        "0xfd2ea2b85d2a315a9817e6661bc3c4378637de37649d2fdb9ca82d6e4172e9e46" ~
```
AndrejMitrovic, line 764: The reason why this changed is because the signature can no longer be empty when it's a fixed-length array. It wasn't obvious at first.

source/scpp/src/crypto/SecretKey.cpp
```diff
@@ -91,7 +93,7 @@ SecretKey::sign(ByteSlice const& bin) const
 {
     assert(mKeyType == PUBLIC_KEY_TYPE_ED25519);
 
-    Signature out(crypto_sign_BYTES, 0);
+    Signature out;
```
AndrejMitrovic, line 96: How about using `Signature out(0);`. Does it work?

MichaelKim20: I will try it.

source/scpp/src/xdr/Stellar-types.x
```diff
@@ -52,7 +52,7 @@ case SIGNER_KEY_TYPE_HASH_X:
 };
 
 // variable size as the size depends on the signature scheme used
```
AndrejMitrovic: This comment can be replaced:

```
// fixed size as we use a 64-byte Signature in Agora
```

AndrejMitrovic (approved):
Just small comments, it LGTM.

source/agora/consensus/protocol/Nominator.d
MichaelKim20: I think the length of 'xdr:::opaque_vec' is recorded in front  when it is serialized.

AndrejMitrovic: It's not about serialization, just about hashing. In this specific case the unittest was testing what happens when the signature is empty, because a dynamic array can be empty. But now that the signature is a fixed length array, it's not possible to test for an "empty" signature.

Anyway your code is correct, I just made a comment in case anyone else is confused why the hash changed.

source/scpp/src/crypto/SecretKey.cpp
MichaelKim20: I tried this, and this is what I got.
```
SecretKey.cpp:96:15: error: no matching constructor for initialization of 'stellar::Signature' (aka 'opaque_array<64>')
    Signature out(0);
```

source/agora/consensus/protocol/Nominator.d
MichaelKim20: Yeah, I checked. Thank you.

source/agora/utils/SCPPrettyPrinter.d
Geod24: Looks like it's been addressed ? I don't see the warning in this module.

source/scpp/extra/DSizeChecks.cpp
```diff
@@ -46,7 +46,7 @@ CPPSIZEOF(CryptoKeyType)
 CPPSIZEOF(PublicKeyType)
 CPPSIZEOF(SignerKeyType)
 CPPSIZEOF(PublicKey)
-CPPSIZEOF(Signature)
+// Signature is removed because it is the same as Hash.
```
Geod24, line 50: Does the test cause problem ?

Geod24 (commented):
Just a couple questions, looks good

source/scpp/extra/DSizeChecks.cpp
MichaelKim20: I could see the following.
```
Build failed: /agora-node/agora/source/scpp/extra/DSizeChecks.cpp:49:1: error: redefinition of 'cppSizeOf'
CPPSIZEOF(Signature)
^
```

MichaelKim20: I changed the pointer of SCPEnvelope to non-pointer.

MichaelKim20: Could you check how it is?

MichaelKim20: There was no warning. But I think it's better to change the same as other structs

MichaelKim20:
@AndrejMitrovic After your approval, I have modified the code, so please review it again.

AndrejMitrovic:
What did you change?

source/agora/utils/SCPPrettyPrinter.d
```diff
@@ -317,30 +301,12 @@ private struct SCPStatementFmt
 /// Formatting struct for `SCPEnvelope`
 private struct SCPEnvelopeFmt
 {
-    /// Pointer to SCPEnvelope
-    private const SCPEnvelope* envelope;
```
AndrejMitrovic: The reason why this was a pointer was to avoid memory copies.

This is even more important now that the signature is a static array.

MichaelKim20: @Geod24, @AndrejMitrovic  Okay. I'll remove the last commit.

AndrejMitrovic (approved):

linked0 (approved):
I checked all the code and LGTM.

source/scpp/extra/DSizeChecks.cpp
AndrejMitrovic: Could we add something like a static assert in the C++ code?

```C++
#include <type_traits> 

static_assert(std::is_same<Signature, Hash>::value, "Signature and Hash must be the same type");
```

Geod24 (approved):
Just include https://github.com/bpfkorea/agora/pull/806/files#r425007595 and LGTM

MichaelKim20:
> Just include https://github.com/bpfkorea/agora/pull/806/files#r425007595 and LGTM

It's just been completed

AndrejMitrovic (approved):

PR #809, opened 2020-05-11
Implement validatorCount() and clear enrollments before adding them

AndrejMitrovic:
Enhancement & refactoring, but it simplifies #785

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -345,6 +345,23 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Get the number of active validators.
+
+        Client code can use this between clearExpiredValidators() calls to
+        determine if the number of validators has changed, and act accordingly.
+
+        Returns:
+            the number of active validators
+
+    ***************************************************************************/
+
+    public size_t validatorCount () @safe
+    {
+        return this.validator_set.count();
+    }
+
+    /***************************************************************************
+
         Clear up expired validators whose cycle for a validator ends
 
         The enrollment manager clears up expired validators from the set based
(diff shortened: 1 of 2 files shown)


linked0 (approved):
It's a simple and nice change.
LGTM.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -343,6 +343,22 @@ public class EnrollmentManager
         return this.validator_set.getValidators(validators);
     }
 
+    /***************************************************************************
+
+        Get the number of active validators. Client code can use this between
+        clearExpiredValidators() calls to determine if the number of validators
+        has changed, and act accordingly.
```
```suggestion
        Get the number of active validators
        
        Client code can use this between clearExpiredValidators() calls to determine
        if the number of validators has changed, and act accordingly.
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -343,6 +343,22 @@ public class EnrollmentManager
...
+
+        Returns:
+            the number of active validators
+
+    ***************************************************************************/
+
+    public size_t validatorCount () @safe
```
Geod24, line 358: Can this be `const` ?

Geod24 (approved):
LGTM just nits

source/agora/consensus/EnrollmentManager.d
AndrejMitrovic: Unfortunately not, because ValidatorSet.count() can't be const because it calls `db.execute` which cannot know in advance if the operation is a read or a modify.

AndrejMitrovic:
Fixed docs.

source/agora/consensus/EnrollmentManager.d
Geod24: That's what I feared :(

Geod24 (approved):
LGTM merge at will

PR #810, opened 2020-05-11
Store initial blockchain state in TestAPIManager

AndrejMitrovic:
This will enable dependency injection for #785.

### DIFF
--- a/source/agora/node/BlockStorage.d
+++ b/source/agora/node/BlockStorage.d
@@ -908,6 +908,11 @@ public class BlockStorage : IBlockStorage
 
 public class MemBlockStorage : IBlockStorage
 {
+    // we already know the blocks in the ctor,
+    // but we should only load them on the call to load()
+    // see also: #599
+    private const(Block)[] _to_load;
+
     /// Storage for all the blocks
     private ubyte[][] blocks;
 
@@ -918,8 +923,9 @@ public class MemBlockStorage : IBlockStorage
     private IndexHash hash_idx;
 
     /// Ctor
-    public this ()
+    public this (const(Block)[] blocks = null)
     {
+        this._to_load = blocks;
         this.height_idx = new IndexHeight();
         this.hash_idx = new IndexHash();
     }
@@ -927,8 +933,14 @@ public class MemBlockStorage : IBlockStorage
     /// No-op: MemBlockStorage does no I/O
     public override bool load ()
     {
-        if (this.blocks.length == 0)
+        if (this._to_load.length == 0)
             this.saveBlock(GenesisBlock);
+        else
+        {
+            foreach (const ref block; this._to_load)
+                this.saveBlock(block);
+        }
+
         return true;
     }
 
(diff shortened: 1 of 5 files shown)


source/agora/test/Base.d
```diff
@@ -492,6 +498,34 @@ public interface TestAPI : ValidatorAPI
...
+
+    /// Load the blockchain
+    public override bool load ()
+    {
+        foreach (const ref block; this.blocks)
+            this.saveBlock(block);
+
+        return true;
+    }
+}
```
Geod24: Does it makes sense to have it here ? That could just replace the MemBlockStorage, couldn't it ? Do you remember why `MemBlockStorage` has serialized Blocks ?

AndrejMitrovic: > Does it makes sense to have it here ? 

Ah it should probably not be `public`. I can move it inside of `TestNodeMixin` I suppose. I only have one use-case for it and it's here in the tests.

> That could just replace the MemBlockStorage, couldn't it ? Do you remember why MemBlockStorage has serialized Blocks ?

Not sure. I think we can just remove that part? Then I could just add a constructor to `MemBlockStorage` and get rid of `PreLoadedMemBlockStorage` I suppose.

AndrejMitrovic:
Ok I see why it's using serialization.

Because it needs to store `const(Block)` in `saveBlock`, but it needs to load mutable `Block` in e.g. `tryReadBlock`.

AndrejMitrovic:
I've removed the extra class and just moved this stuff into the `MemBlockStorage` constructor.

source/agora/node/BlockStorage.d
```diff
@@ -918,17 +918,23 @@ public class MemBlockStorage : IBlockStorage
...
+            foreach (const ref block; blocks)
+                this.saveBlock(block);
+        }
     }
 
     /// No-op: MemBlockStorage does no I/O
     public override bool load ()
     {
-        if (this.blocks.length == 0)
-            this.saveBlock(GenesisBlock);
```
Geod24, line 931: I would add a boolean in the new class, or at least something that checks we don't accidentally forget to initialize the BlockStorage from our code. [It happened before](https://github.com/bpfkorea/agora/commit/c7b90eb9483607d900fef63c3a7b63b2cae2b5dc)

Geod24:
[This one too](https://github.com/bpfkorea/agora/pull/840/files#diff-b11d74d50f129e77728ec32bb7f9b4d9R1367) ?

source/agora/node/BlockStorage.d
AndrejMitrovic: Ah then I'll move the loading into this method.

AndrejMitrovic:
Split it into two commits, added a test-case for https://github.com/bpfkorea/agora/commit/c7b90eb9483607d900fef63c3a7b63b2cae2b5dc, and removed `PreloadedMemBlockStorage`.

source/agora/node/BlockStorage.d
```diff
@@ -937,10 +949,6 @@ public class MemBlockStorage : IBlockStorage
         // Basic consistenty checks
         assert(this.height_idx.length == this.hash_idx.length);
         assert(this.height_idx.length == this.blocks.length);
-
-        // Make sure we have no empty block
-        foreach (blk; this.blocks)
-            assert(blk.length > 0);
```
Geod24: Is this test failing now ?

AndrejMitrovic: Ah I'll revert this, I initially replaced the serialized form with full blocks but that didn't work.

source/agora/test/Base.d
```diff
@@ -543,6 +549,17 @@ private mixin template TestNodeMixin ()
...
+    private immutable(Block)[] blocks;
+
+    ///
+    public this (Config config, Registry* reg, immutable(Block)[] blocks)
+    {
+        this.registry = reg;
+        this.blocks = blocks;
+        super(config);
+    }
+
```
Geod24, line 562: Why aren't the ctor grouped together ?

AndrejMitrovic: Can you explain what you mean?

Geod24:
Looks roughly good to me. Second commit title looks like a copy-pasta mistake though

AndrejMitrovic:
Last commit message is wrong, I'll fix it.

source/agora/test/Base.d
Geod24: I need some sleep

AndrejMitrovic:
Fixed commits and reverted other change.

Geod24 (approved):

PR #811, opened 2020-05-11
Use Serializer instead of JSON for LocalRest tests

Geod24:
This should allow us to test whether a real binary protocol would work, serialization wise.
It should also make our unittests faster.

Related to #203

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -53,7 +53,7 @@ import scpd.types.Stellar_SCP;
 
 import ocean.util.log.Logger;
 
-import geod24.LocalRest;
+static import geod24.LocalRest;
 import geod24.Registry;
 
 import core.stdc.time;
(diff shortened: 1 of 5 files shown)


source/agora/test/Base.d
```diff
@@ -172,6 +172,49 @@ private UnitTestResult customModuleUnitTester ()
...
+                writeln(t);
+            }
+            catch (Exception ignored) {}
+            assert(0);
+        }
+    }
+}
+
+/// A different default serializer from `LocalRest` for `RemoteAPI`
+alias RemoteAPI (APIType) = geod24.LocalRest.RemoteAPI!(APIType, Serializer);
```
AndrejMitrovic: I think the code in `agora.test.NetworkManager` will still use the default one that uses Json serialization, because it uses a local import into LocalRest.. which will probably hide this symbol here.

Geod24: The PR also removes that import

AndrejMitrovic: Oh didn't see. What about also making the import in this module static? I don't think any other symbols are used from it except here?

Geod24:
Applied suggestion (`static import`) and made the `alias` `public`. Ready for review.

source/scpd/Cpp.d
```diff
@@ -348,6 +352,28 @@ extern(C++, (StdNamespace)) extern(C++, class) struct vector (T, Alloc = allocat
...
+                serializePart(entry, dg);
+        }
+
+        static QT fromBinary (QT) (scope DeserializeDg data,
+            scope const ref DeserializerOptions opts) @safe
+        {
+            import std.traits;
+            import scpd.types.Utils;
+
+            auto arr = deserializeFull!(QT.ElementType[])(data, opts);
```
AndrejMitrovic: Hmm.. Why is it allocating an array here? Instead of manually walking over the serialized data?

Geod24: Right, could do that. I think the original concern was to make sure the data would stay alive after this returns. I'll see if I can improve this.

Honestly this is the worst part of the code, because dealing with vectors is terrible. We really need a way to build a `const` vector from an array.

AndrejMitrovic: Ah right because the vector only stores the pointers. Then if there's no better way, I'll just approve this.

AndrejMitrovic (approved):

AndrejMitrovic:
Merge at-will.

Geod24:
Changed according to @AndrejMitrovic 's comment:
```diff
diff --git a/source/scpd/Cpp.d b/source/scpd/Cpp.d
index 06aeee3a..6a084285 100644
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -366,12 +366,15 @@ extern(C++, (StdNamespace)) extern(C++, class) struct vector (T, Alloc = allocat
             import std.traits;
             import scpd.types.Utils;

-            auto arr = deserializeFull!(QT.ElementType[])(data, opts);
             // Note: Unqual necessary because we can't construct an
             // `immutable` vector yet
             Unqual!(vector!(Unqual!(QT.ElementType))) ret;
-            foreach (ref entry; arr)
+            immutable len = deserializeLength(data, opts.maxLength);
+            foreach (idx; 0 .. len)
+            {
+                auto entry = deserializeFull!(QT.ElementType)(data, opts);
                 push_back(cast() ret, entry);
+            }
             return () @trusted { return cast(QT) ret; }();
         }
     }
```

PR #813, opened 2020-05-11
Throw if the genesis block in the storage doesn't match the configured genesis block

AndrejMitrovic:
@Geod24 in general I don't like the way the code works. The `setGenesis(new) scope (exit) setGenesis(old)` reminds me a lot about `enableGagging` in DMD. Just the use of globals seems bad..

And in retrospect having the genesis block be defined in a config file is error-prone. If the file is corrupt, or if a user copy-pastes a configuration from some malicious website, they may end up on the wrong chain.

I'll have to think about a better alternative to `setGenesisBlock`..


source/agora/node/Ledger.d
```diff
@@ -1346,3 +1352,50 @@ unittest
...
+
+    // hex => bin
+    auto block_bytes = gen_block_hex.chunks(2).map!(
+        twoDigits => twoDigits.parse!ubyte(16)).array();
+    immutable new_gen_block = block_bytes.deserializeFull!(immutable(Block));
```
MichaelKim20, line 1359: It's okay as it is, but if you add it, others will understand it better.

```suggestion
    immutable new_gen_block = block_bytes.deserializeFull!(immutable(Block));
    assert(new_gen_block != GenesisBlock);
```

MichaelKim20 (approved):
It's all good. I made a suggestion that is not serious.

Geod24:
So we discussed this IRL with @AndrejMitrovic and talked about providing CLI switches instead. Where are we on this ?

AndrejMitrovic:
See the comments in #821

source/agora/node/Ledger.d
```diff
@@ -1346,3 +1352,50 @@ unittest
...
+    auto utxo_set = new UTXOSet(":memory:");
+    scope (exit) utxo_set.shutdown();
+    auto config = new Config();
+    auto enroll_man = new EnrollmentManager(":memory:", gen_key_pair);
+    scope (exit) enroll_man.shutdown();
+    config.node.is_validator = true;
+
+    try
+    {
+        scope ledger = new Ledger(pool, utxo_set, storage, enroll_man, config.node);
```
Geod24: Can you replace this with a `TestLedger` ?

source/agora/node/Ledger.d
```diff
@@ -1346,3 +1352,50 @@ unittest
...
+    {
+        assert(ex.msg == "Genesis block loaded from disk is different to the one in the config file");
+    }
+
+    auto old_gen = &GenesisBlock();
+    setGenesisBlock(&new_gen_block);
+    scope (exit) setGenesisBlock(old_gen);  // must reset it for other tests
+
+    // will no longer fail
+    scope ledger = new Ledger(pool, utxo_set, storage, enroll_man, config.node);
```
Geod24: Ditto `TestLedger`

source/agora/node/Ledger.d
```diff
@@ -1346,3 +1352,50 @@ unittest
     genBlockTransactionsFreeze(1, TxType.Payment);
     assert(ledger.getBlockHeight() == 11 + 2016);
 }
+
+/// test that loading wrong genesis block throws
+unittest
+{
+    import agora.common.crypto.Key;
+    import agora.common.Types;
+    import agora.common.Hash;
```
Geod24: Those three imports are already present at module level

source/agora/node/Ledger.d
```diff
@@ -1346,3 +1352,50 @@ unittest
...
+    import agora.common.Serializer;
+    import std.conv;
+
+    auto gen_key_pair = getGenesisKeyPair();
```
Geod24: This is gone

source/agora/node/Ledger.d
```diff
@@ -1346,3 +1352,50 @@ unittest
...
+
+    // hex => bin
+    auto block_bytes = gen_block_hex.chunks(2).map!(
+        twoDigits => twoDigits.parse!ubyte(16)).array();
+    immutable new_gen_block = block_bytes.deserializeFull!(immutable(Block));
+    auto storage = new MemBlockStorage();
+    storage.saveBlock(new_gen_block);  // stores it as the genesis block
+
+    auto pool = new TransactionPool(":memory:");
+    scope(exit) pool.shutdown();
```
```suggestion
    scope (exit) pool.shutdown();
```

For consistency

Geod24 (commented):
So, even with the approach of splitting the Genesis block, we still need this. Sorry for not seeing this earlier. Can you rebase / address review ?

source/agora/node/Ledger.d
AndrejMitrovic: I don't think so, I need to pass the MemBlockStorage loaded with the wrong block to the constructor of Ledger.

Geod24: `TestLedger` accepts an array of `Block` as second argument which will then be loaded by the `MemBlockStorage`.

AndrejMitrovic:
Updated, using TestLedger now.

Geod24 (approved):
Noice

PR #817, opened 2020-05-12
Sort enrollments in the call to makeNewBlock

AndrejMitrovic:
While adding enrollments in the tests this code was failing for me. But the function already does sorting on transactions.

I've left in the assert because it also checks for duplicate enrollments.

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -409,8 +409,9 @@ public Block makeNewBlock (Transactions)(const ref Block prev_block,
 
     block.header.merkle_root = block.buildMerkleTree();
     block.header.enrollments = enrollments;
+    block.header.enrollments.sort!((a, b) => a.utxo_key < b.utxo_key);
     assert(block.header.enrollments.isStrictlyMonotonic!
-        ("a.utxo_key < b.utxo_key"));
+        ("a.utxo_key < b.utxo_key"));  // there cannot be duplicates either
     return block;
 }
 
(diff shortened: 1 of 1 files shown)


source/agora/consensus/data/Block.d
```diff
@@ -410,8 +411,9 @@ public Block makeNewBlock (Transactions)(const ref Block prev_block,
 
     block.header.merkle_root = block.buildMerkleTree();
     block.header.enrollments = enrollments;
+    block.header.enrollments.sort!((a, b) => a.utxo_key < b.utxo_key);
```
AndrejMitrovic, line 412: Ah maybe I can tell `sort` it should be strictly monotonic. Let me read the docs.

AndrejMitrovic: It doesn't seem to have this feature.

Geod24:
Needs rebase. Also, test ?

AndrejMitrovic:
Added a test-case.

Geod24 (approved):

source/agora/consensus/data/Block.d
TrustHenry: Looking back on my old memories, I know that the enrollments are missing here because they are sorted from the DB.

at present here
https://github.com/bpfkorea/agora/blob/e72c0fbd60013b02303a705ce3d3718372081440/source/agora/consensus/EnrollmentPool.d#L266-L267

In the future
https://github.com/bpfkorea/agora/blob/e72c0fbd60013b02303a705ce3d3718372081440/source/agora/consensus/ValidatorSet.d#L259

TrustHenry: Is it because of the data change in the candidate process?

AndrejMitrovic: It reduces the chance of bugs creeping in.

There's a much bigger cost in implicitly sorting the transactions. There will be just a few enrollments and only periodically and not in every block, but there will be many transactions in every block.

PR #818, opened 2020-05-12
Take const transaction in makeChainedTransactions

AndrejMitrovic:
Also fixed up some docs.


linked0 (approved):
LGTM

TrustHenry:
`MakeChainedTransactions` is used in many unit tests.
 Most of them are not `const`
Is that all right with you?

AndrejMitrovic:
Mutable implicitly converts to const. Const does not implicitly convert to mutable. This doesn't break any code.

MichaelKim20 (approved):

PR #819, opened 2020-05-12
Print logs when constructor / start() fails

AndrejMitrovic:
Currently when a single node fails in either its constructor call or when start() is called, its logs will not be emitted.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -61,6 +61,7 @@ import std.array;
 import std.algorithm;
 import std.exception;
 import std.format;
+import std.stdio;
 
 import core.runtime;
 import core.time;
@@ -81,7 +82,6 @@ private UnitTestResult customModuleUnitTester ()
     import std.algorithm;
     import std.parallelism;
     import std.process;
-    import std.stdio;
     import std.string;
     import std.uni;
     import core.atomic;
@@ -353,15 +353,9 @@ public class TestAPIManager
     {
         synchronized  // make sure logging output is not interleaved
         {
-            import std.stdio;
             writefln("%s(%s): Node logs:\n", file, line);
             foreach (node; this.nodes)
-            {
-                writefln("Log for node %s:", node.address);
-                writeln("======================================================================");
                 node.client.printLog();
-                writeln("======================================================================\n");
-            }
         }
     }
 
@@ -515,7 +509,10 @@ private mixin template TestNodeMixin ()
     /// Prints out the log contents for this node
     public void printLog ()
     {
+        writefln("Log for node: %s", this.config.node.address);
+        writeln("======================================================================");
         CircularAppender().printConsole();
+        writeln("======================================================================\n");
     }
 
     /// Used by the node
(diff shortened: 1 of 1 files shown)


source/agora/test/Base.d
```diff
@@ -515,7 +511,11 @@ private mixin template TestNodeMixin ()
     /// Prints out the log contents for this node
     public void printLog ()
     {
+        import std.stdio;
+        writefln("Log for node %s:", this.config.node.address);
```
TrustHenry: `writefln("Log for node :%s", this.config.node.address);`

TrustHenry (approved):
LGTM , a nick pick

AndrejMitrovic:
@TrustHenry fixed.

MichaelKim20:
Everything is good. But I wanted to check one thing.

I intentionally added `assert(0)` to the constructor and start() and checked the log.
I checked the logs that occur when the constructor fails.
I also checked the log at `start()`.

But when I added `assert(0)` to `this.tasman.runTask` in the `start()`, I couldn't see the log.
Is this what you wanted normally?

AndrejMitrovic:
You're right, the issue is `runTask` runs it in another fiber, so `start()` will not catch it. I will come up with a solution.

AndrejMitrovic:
It doesn't seem possible to print out failures in `start()`. So I removed that change.

We cannot catch the exception in the `start()` call since the code runs in another fiber.

I tried a different solution:

```diff
diff --git a/source/agora/test/NetworkManager.d b/source/agora/test/NetworkManager.d
index 40f153a2..f1618df7 100644
--- a/source/agora/test/NetworkManager.d
+++ b/source/agora/test/NetworkManager.d
@@ -139,9 +139,9 @@ unittest

     TestConf conf = { topology : NetworkTopology.OneValidator, nodes : 3 };
     auto network = makeTestNetwork!BadAPIManager(conf);
+    scope(failure) network.printLogs();
     network.start();
     scope(exit) network.shutdown();
-    scope(failure) network.printLogs();
     network.waitForDiscovery();

     auto nodes = network.clients;
```

However this will still not work. The start() call will silently fail, LocalRest will destroy the threads. Then when a unittest fails, the `scope(exit) network.shutdown();` call will finally be reached, but the threads are dead so we can't access their logs anymore.

Maybe LocalRest could have some kind of optional thread cleanup delegate parameter. So when a thread is about to unexpectedly die it could call a user-defined function. In our case, that would be `CircularAppender.printConsole`.

Or maybe that's too much work for too little benefit..

Anyway this PR is ready.

MichaelKim20 (approved):
Yes, I fully understand. All right.

Geod24:
> Or maybe that's too much work for too little benefit..

Yeah I'm not so sure about providing such delegate. Because it should be so rare that it won't be properly tested. Instead just make sure your constructor call doesn't crash, it ain't that hard peeps :D

source/agora/test/Base.d
```diff
@@ -515,7 +510,11 @@ private mixin template TestNodeMixin ()
     /// Prints out the log contents for this node
     public void printLog ()
     {
+        import std.stdio;
```
Geod24: Might as well move it at module scope ? It's a unittest module, we don't really shy on the dependencies.

Geod24 (approved):

PR #798, opened 2020-05-07
Restore a node's enrollment data from the Genesis block

linked0:
This PR is for restoring the enrollment data of a node from the `enrollments` field of the header of blocks. This is the first step to remove the `node_enroll_data` table in the `EnrollmentManager` class, which is to remove the `Enrollment` object from the table. The next step is to remove `next_reveal_height` from the table. Finally, the table will be removed.

(EDIT)
This PR has added two more items as follows
- Restoring the enrollment data when the `EnrollmentManager` adds an validator to the `ValidatorSet`.
- Restoring the pre-image data from blocks of the chain

Fixes #789

PR #823, opened 2020-05-13
Add `opSlice` to the `Point` structure

linked0:
There is a need to convert a `Point` value into a `PublicKey` value. The need for the function is related to PR #798

### DIFF
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -268,6 +268,12 @@ public struct Point
         return result;
     }
 
+    /// Convenience overload to allow this to be converted to a `PublicKey`
+    public const(ubyte)[] opSlice () const @safe pure nothrow @nogc
+    {
+        return this.data[];
+    }
+
     /// Support for comparison
     public int opCmp (ref const typeof(this) s) const
     {
--- a/source/agora/common/crypto/Key.d
+++ b/source/agora/common/crypto/Key.d
@@ -430,3 +430,13 @@ unittest
     Point point_Address = Point(kp.address);
     assert(verify(point_Address, enroll_sig, "BOSAGORA"));
 }
+
+/// Test for converting from `Point` to `PublicKey`
+unittest
+{
+    import agora.common.crypto.Schnorr;
+
+    Pair pair = Pair.random();
+    auto pubkey = PublicKey(pair.V[]);
+    assert(pubkey.data == pair.V.data);
+}


linked0:
This is a PR for getting your help. I'm blocked in writing an unit test code like as follows.
```
Pair pair = Pair.random();
auto width = crypto_core_ed25519_BYTES * 8;
auto pubkey = PublicKey(pair.V.data[0..width]);
assert(pubkey.data == pair.V.data);
```
How can I found the width of the `pair.V.data`?

source/agora/common/crypto/ECC.d
```diff
@@ -268,6 +268,12 @@ public struct Point
         return result;
     }
 
+    /// Convenience overload
+    public inout(ubyte)[] opSlice (size_t from, size_t to) inout
+    {
+        return this.data[from .. to];
+    }
```
```suggestion
    /// Convenience overload to allow this to be converted to a `PublicKey`
    public inout(ubyte)[] opSlice () inout
    {
        return this.data[];
    }
```

That should work.

linked0: Thanks for your help. @Geod24

linked0:
Ready for review.

source/agora/common/crypto/ECC.d
```diff
@@ -268,6 +268,12 @@ public struct Point
         return result;
     }
 
+    /// Convenience overload to allow this to be converted to a `PublicKey`
+    public inout(ubyte)[] opSlice ( ) inout nothrow @nogc
```
AndrejMitrovic: Why `inout` though? I think this should be `const`. Is there a use-case for modifying the internal data?

linked0: If I apply the `const` to the function, I get an error like this.
```
source/agora/common/Hash.d(133,21): Error: mutable method agora.common.crypto.ECC.Point.opSlice is not callable using a const object
```

AndrejMitrovic: `const` should be on both sides. Try that.

linked0: It works. Thanks.

source/agora/common/crypto/ECC.d
```diff
@@ -268,6 +268,12 @@ public struct Point
         return result;
     }
 
+    /// Convenience overload to allow this to be converted to a `PublicKey`
+    public const(ubyte)[] opSlice ( ) const nothrow @nogc
```
```suggestion
    public const(ubyte)[] opSlice () const @safe pure nothrow @nogc
```
Or am I missing something here ?

source/agora/common/crypto/Key.d
```diff
@@ -430,3 +430,13 @@ unittest
...
     assert(verify(point_Address, enroll_sig, "BOSAGORA"));
 }
+
+/// Test for converting from `Point` sturcture to `PublicKey` structure
+unittest
+{
+    import agora.common.crypto.Schnorr;
+
+    Pair pair = Pair.random();
+    auto pubkey = PublicKey(pair.V.data[]);
```
MichaelKim20: Can't we do this?
```suggestion
    auto pubkey = PublicKey(pair.V[]);
```

linked0: > Can't we do this?

Yes. we can. It works. Thanks.

MichaelKim20: > > Can't we do this?
> 
> Yes. we can. It works. Thanks.

I checked a little more
I think you need to fix this to test Point.opSlice.

linked0: > I think you need to fix this to test Point.opSlice.

Do you mean I should change to calling the Point.opSlice instead of [] operator? It would be more helpful if you suggest a code.

MichaelKim20 (approved):
LGTM

source/agora/common/crypto/ECC.d
AndrejMitrovic: @linked0 can you add `@safe pure` here?

linked0:
Could anyone review or approve this PR?

AndrejMitrovic:
> Could anyone review or approve this PR?

https://github.com/bpfkorea/agora/pull/823/files#r424190337

linked0:
> > Could anyone review or approve this PR?
> 
> https://github.com/bpfkorea/agora/pull/823/files#r424190337

Yes, I will. Well, I have forgotten this.

Geod24 (approved):

PR #831, opened 2020-05-13
 Test & fix BanManager's serialization

Geod24:
```
Before this change, the serialization symmetry wasn't tested at all,
since we don't have any test that restart a node yet,
and this type isn't passed around.
Being untested, it was, obviously, buggy: while the serialization
was serializing `{ length, keys[], values[] }`, the deserialization
expected `{ length, { keys[idx], values[idx] }[] }`.
The serialization part was changed to meet the deserialization's
reasonable expectations.
```

### DIFF
--- a/source/agora/common/BanManager.d
+++ b/source/agora/common/BanManager.d
@@ -127,7 +127,7 @@ public class BanManager
         if (!this.banfile_path.exists())
             return;  // nothing to load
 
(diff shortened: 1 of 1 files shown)


AndrejMitrovic (approved):

AndrejMitrovic:
Sorry for not fixing this back when it was first discovered.

PR #832, opened 2020-05-14
Tests: Use more public import, hoist imports

Geod24:
Some imports, such as std.algorithm, are ubiquitous in the test code.
Instead of copy-pasting a large list of import everywhere, we should make it easier to write test cases.
Simplifying the dependency is a first step in that direction.


AndrejMitrovic (approved):

AndrejMitrovic:
That makes a lot of sense.

PR #837, opened 2020-05-14
Improved messages from unit test build in Windows

TrustHenry:
Currently, This is the message that comes from the unit test build in the windows.

```
This does not support Windows.
This does not support Windows.
```

It makes it clearer where it's being printed.

### DIFF
--- a/source/agora/cli/checkvtable/check.d
+++ b/source/agora/cli/checkvtable/check.d
@@ -19,7 +19,11 @@ import std.format;
 import std.stdio;
 import std.string;
 
-version (Windows) { pragma(msg, "This does not support Windows."); } else:
+version (Windows)
+{
+    pragma(msg, __MODULE__, " : This does not support Windows.");
+}
+else:
 
 version (unittest) { } else:
 
--- a/source/agora/cli/checkvtable/generate.d
+++ b/source/agora/cli/checkvtable/generate.d
@@ -19,7 +19,11 @@ import std.file;
 import std.format;
 import std.stdio;
 
-version (Windows) { pragma(msg, "This does not support Windows."); } else:
+version (Windows)
+{
+    pragma(msg, __MODULE__, " : This does not support Windows.");
+}
+else:
 
 version (unittest) { } else
 


source/agora/cli/checkvtable/check.d
```diff
@@ -19,7 +19,10 @@ import std.format;
 import std.stdio;
 import std.string;
 
-version (Windows) { pragma(msg, "This does not support Windows."); } else:
+version (Windows)
+{
+    pragma(msg, "Checking virtual method offset : This does not support Windows.");
```
linked0: How about inserting the module name instead of the text of "Checking virtual method offset"?

source/agora/cli/checkvtable/generate.d
```diff
@@ -19,7 +19,10 @@ import std.file;
 import std.format;
 import std.stdio;
 
-version (Windows) { pragma(msg, "This does not support Windows."); } else:
+version (Windows)
+{
+    pragma(msg, "The agora-checkvtable-gen : This does not support Windows.");
```
linked0: How about inserting the module name instead of the text of "The agora-checkvtable-gen"?

linked0 (commented):
Simple and nice change!
But, some suggestions.

source/agora/cli/checkvtable/check.d
Geod24: To add to @linked0 's suggestion:
```suggestion
    pragma(msg, __MODULE__, " : This does not support Windows.");
```

TrustHenry:
Actually, I was thinking about that.
The suggestions have been applied. 
@linked0 @Geod24 Thank you.

MichaelKim20 (approved):
It was a good work.

source/agora/cli/checkvtable/check.d
```diff
@@ -19,7 +19,10 @@ import std.format;
 import std.stdio;
 import std.string;
 
-version (Windows) { pragma(msg, "This does not support Windows."); } else:
+version (Windows)
+{
+    pragma(msg, __MODULE__, " : This does not support Windows.");
+} else:
```
Geod24:
```suggestion
}
else:
```

Geod24 (approved):
Merge at will

source/agora/cli/checkvtable/check.d
TrustHenry: a sharp point.

PR #838, opened 2020-05-14
Group cycle data together and use PreImageCache

Geod24:
See each commit individually. There are still a few things to fix (e.g. `nonce` is not used) but I'll leave it for a later PR.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -115,6 +115,14 @@ public class EnrollmentManager
     /// This struct hold all the cycle data together for better readability
     private static struct PreImageCycle
     {
+        /// Make sure we get initialized by disabling the default ctor
+        @disable public this();
+        /// Ditto
+        public this (typeof(PreImageCycle.tupleof) args)
+        {
+            this.tupleof = args;
+        }
+
         /***********************************************************************
 
             The number of the current cycle
@@ -158,7 +166,7 @@ public class EnrollmentManager
 
         ***********************************************************************/
 
-        private Hash[ulong] preimages;
+        private PreImageCache preimages;
     }
 
     /// Ditto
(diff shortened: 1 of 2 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -278,6 +287,8 @@ public class EnrollmentManager
         this.enroll_key = frozen_utxo_hash;
 
         // N, cycle length
+        // Note, this name iis misleading, as a cycle here is the length
```
```suggestion
        // Note, this name is misleading, as a cycle here is the length
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -118,8 +118,17 @@ public class EnrollmentManager
...
+
+        The number of enrollments we can emit in a cycle
+
+        With the current setting of ValidatorCycle being 1008 and blocks
+        having 10 minutes of interval, this gives us roughly 700 days,
+        or two years, per cycle.
+
+    ***************************************************************************/
+
+    private immutable uint EnrollPerCycle = 100;
```
AndrejMitrovic: I'm not sure this name is that much better. The name seems to suggest there will be 100 enrollments per cycle, and a cycle is 1008 blocks? But that doesn't make sense.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -287,12 +330,12 @@ public class EnrollmentManager
         this.enroll_key = frozen_utxo_hash;
 
         // N, cycle length
-        // Note, this name iis misleading, as a cycle here is the length
+        // Note, this name is misleading, as a cycle here is the length
```
AndrejMitrovic: It looks like you introduced the spelling mistake in the previous commit, and fixed it in this commit.

source/agora/consensus/EnrollmentManager.d
Geod24: A cycle is 100 enrollments, and enrollment is 1008 blocks.

linked0: We use a cycle in white paper as the period where a node is validator. So how about `CyclesPerNonce`?

Geod24:
Removed the first commit, and the typo by extension

linked0 (approved):
LGTM.
I checked the code throughly. It's very structural change.

AndrejMitrovic (approved):

PR #839, opened 2020-05-14
Move pre-image cache handling in inner struct, use PreImageCache for seeds

Geod24:
Base on #838, only last 2 commits are part of this PR.

```
Like we did for `preimages`, switch `seeds` to use `PreImageCache`.
This is way more memory efficient and reduce the amount of duplication.
It allows to reduce `generatePreimage` to a very simple, 5 lines functions,
which can in turn be moved inside the inner struct,
to which we added proper handling of `nonce`.

The last thing left to do is to move the consume call to `addValidator`,
so that we only consume a round of pre-image when we are confirmed as
a validator. The `createEnrollment` code will have to generate the
next round's pre-image manually, which can be expensive if the next round
is in another cycle, but is another step towards decouple generating
an Enrollment and actually acting on this Enrollment.
```

Note that this is still lacking tests, but I need to eliminate the constant first.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -155,7 +155,7 @@ public class EnrollmentManager
 
         ***********************************************************************/
 
-        private Hash[ulong] seeds;
+        private PreImageCache seeds;
 
         /***********************************************************************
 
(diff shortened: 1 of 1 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -167,6 +167,46 @@ public class EnrollmentManager
...
+        {
+            // Populate the nonce cache if necessary
+            if (this.index == 0)
+            {
+                const cycle_seed = hashMulti(
+                    secret, "consensus.preimages", this.nonce);
+                this.seeds.reset(cycle_seed);
+            }
+            // Populare the current enrollment round cache
+            const ret = this.preimages.reset(this.seeds[index]);
```
```suggestion
            const ret = this.preimages.reset(this.seeds[this.index]);
```

AndrejMitrovic (changes requested):

source/agora/consensus/EnrollmentManager.d
Geod24: Good catch

Geod24:
Rebased on master, addressed @AndrejMitrovic 's review.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -167,6 +167,46 @@ public class EnrollmentManager
...
+
+            Consume the current cycle
+
+            This will first populate the caches (seeds and preimages)
+            as necessary, then increase the `index` by one, or reset it to 0
+            and increase `nonce` if necessary.
+            Note that the increment is done after the caches are populated,
+            so `consume` needs to be called once this node has confirmed it
+            is part of consensus.
+
```
linked0, line 181: Need to add documentation for the parameter.

Geod24: Good catch, fixed

AndrejMitrovic (approved):

PR #800, opened 2020-05-07
Add ability to create a merkle tree of arbitrary number of transactions

AndrejMitrovic:
As mentioned in https://github.com/bpfkorea/agora/issues/797#issuecomment-625054477, it's not blocking anything yet. But it probably will once we remove the limitation of 8 transactions per block.

Fixes #797

PR #840, opened 2020-05-14
Fix Enrollments not being loaded from the Genesis block

AndrejMitrovic:


source/agora/node/Ledger.d
```diff
@@ -105,19 +105,20 @@ public class Ledger
             }
         }
 
-        // restore validator set from lastest blocks
-        if (this.last_block.header.height > 0)
+        // max: there may be less than 1008 blocks stored in the ledger
+        const min_height = max(long(0), long(this.last_block.header.height)
+            - Enrollment.ValidatorCycle);
```
```suggestion
        const min_height = max(long(0), long(this.last_block.header.height)
            - Enrollment.ValidatorCycle + 1);
```
I've checked it over and over again, and I think I need to add one more, so please check it once.

AndrejMitrovic: I'll add a test-case to be sure.

Geod24: Not a big fan of the `long(...) - X` thing, I'm more a `x >= y ? 0 : x - y` kind of person, but to each their own.

source/agora/node/Ledger.d
```diff
@@ -1352,3 +1353,88 @@ unittest
...
     assert(ledger.getBlockHeight() == 11 + 2016);
 }
+
+/// test enrollments in the genesis block
+unittest
+{
+    import agora.common.BitField;
+    import std.exception;
+
+    class PreLoadedMemBlockStorage : MemBlockStorage
```
```suggestion
    static class PreLoadedMemBlockStorage : MemBlockStorage
```

Nitpick

source/agora/node/Ledger.d
```diff
@@ -1352,3 +1353,88 @@ unittest
...
+    Transaction tx =
+    {
+        type : TxType.Freeze,
+        inputs : [ Input.init ],
+        outputs : [Output(Amount.MinFreezeAmount, key_pair.address)]
+    };
+
+    txs ~= tx;
+    Hash txhash = hashFull(tx);
+    Hash utxo = UTXOSet.getHash(txhash, 0);
```
Geod24: Is it intentional that there are only 2 transactions ?

Note that we recently had a similar problem with Jay, `makeChainedTransactions` doesn't allow to create frozen tx. That's when I started to poke around the test suite. What I'd do, if we wanted 8 txs, would be to go with `makeChainedTxs` but then change the `type` and re-sign the txs (should be a good 5 lines).
Of course after writing this I realize it doesn't apply because you need to use `Input.init`...

source/agora/node/Ledger.d
```diff
@@ -1352,3 +1353,88 @@ unittest
...
+    scope enroll_man = new EnrollmentManager(":memory:", key_pair);
+    scope (exit) enroll_man.shutdown();
```
```suggestion
    scope(exit) enroll_man.shutdown();
```

source/agora/node/Ledger.d
```diff
@@ -1352,3 +1353,88 @@ unittest
...
+        ),
+        txs.assumeUnique,
+        merkle_tree.assumeUnique
+    );
+
+    scope storage = new PreLoadedMemBlockStorage([genesis]);
+    scope pool = new TransactionPool(":memory:");
+    scope(exit) pool.shutdown();
+    scope utxo_set = new UTXOSet(":memory:");
+    scope (exit) utxo_set.shutdown();
```
```suggestion
    scope(exit) utxo_set.shutdown();
```

Geod24 (approved):
A few nits but LGTM, thanks!

AndrejMitrovic:
I've added an additional test-case.

Just to make sure we're on the same page here:

If a Validator's enrollment is in the Genesis block, it means it may validate blocks [1 .. 1028]. However, once Block 1028 is externalized that Validator's enrollment is over. Only externalized blocks are loaded from disk.

source/agora/node/Ledger.d
AndrejMitrovic: > Is it intentional that there are only 2 transactions ?

Yes. #800

Actually I just added a comment about it after I realized it was implicit here.

source/agora/node/Ledger.d
AndrejMitrovic: Ha I fixed that too before reading your review. :D

source/agora/node/Ledger.d
AndrejMitrovic: > I'm more a `x >= y ? 0 : x - y` kind of person

Sure, but you wrote the wrong comparison there.

source/agora/node/Ledger.d
AndrejMitrovic: And also yes we need to refactor the transaction creation and make it much more flexible in the future.

AndrejMitrovic:
I replaced the `max` call.

source/agora/node/Ledger.d
```diff
@@ -1352,3 +1354,143 @@ unittest
...
+            const NoEnrollments = null;
+            blocks ~= makeNewBlock(blocks[$ - 1], txs, NoEnrollments);
+            prev_txs = txs;
+        }
+
+        return blocks.assumeUnique;
+    }
+
+    // only genesis loaded: validator is active
+    {
```
AndrejMitrovic, line 1446: @Geod24 if you don't like these `{ }` scopes I'll move them to separate tests. I just wanted to avoid accidentally re-using existing variables. Re-using the enrollment manager for example would cause a bug here.

source/agora/node/Ledger.d
Geod24: I never said I was the right kind of person tho

AndrejMitrovic: :P

MichaelKim20: The code is very simple and I love it. I also checked the added test code.
What is the height of the block at 1009?

MichaelKim20: The code is very simple and I love it. I also checked the added test code.
What is the result of block height 1009?

AndrejMitrovic: Well if the validator set is empty at block 1008, it's also empty at block 1009.

MichaelKim20: Okay~

MichaelKim20 (approved):
LGTM

source/agora/node/Ledger.d
Geod24: No that works great to split tests

source/agora/node/Ledger.d
Geod24: >     // 1 payment tx, the rest are freeze txs

This comment kinda threw me off, I expect "the rest" to be more than one (other it's "the other").

Geod24:
The `scope (` things crept back

AndrejMitrovic:
Thanks @MichaelKim20. There is an off-by-1 error still where it tries to load 1009 blocks rather than 1008. But it uncovered another bug in my tests.

AndrejMitrovic:
There is a check in `ValidatorSet` with:

```
if (last_height - block.header.height < Enrollment.ValidatorCycle)
```

That's why the bug of loading 1009 blocks instead of 1008 hasn't appeared. But in any case, I'll add the +1 to the offset so we don't load that 1 extra block.

AndrejMitrovic:
Fixed the check and the styling.

source/agora/node/Ledger.d
```diff
@@ -105,19 +105,23 @@ public class Ledger
...
+        const ulong block_count = this.last_block.header.height + 1;
+
+        // we are only interested in the last 1008 blocks,
+        // because that is the maximum length of an enrollment.
+        const ulong min_height =
+            block_count >= Enrollment.ValidatorCycle
+            ? block_count - Enrollment.ValidatorCycle : 0;
+
+        // restore validator set from the blockchain.
+        // using block_count, as the range is inclusive
```
Geod24, line 118: Well the range is not inclusive: the right bound is exclusive, but since you used `+ 1` it works.

Geod24 (approved):

PR #842, opened 2020-05-14
Fix 704: Split validation into a package

### LINKED ISSUE #704: Split the Validation module into several modules + a package module
Since the Validation module is becoming quite big, it would be useful to split its validation routines into separate modules. For example:

```
src/agora/consensus/validation/Block.d
src/agora/consensus/validation/Transaction.d
src/agora/consensus/validation/Enrollment.d
```

In addition to that, we should add a _package module_:

```
src/agora/consensus/validation/package.d
```

And essentially do:

```D
import agora.consensus.validation.Block;
import agora.consensus.validation.Transaction;

/// merge overload sets
public alias isValid = agora.consensus.validation.Block.isValid;
public alias isValid = agora.consensus.validation.Transaction.isValid;
```

@bpfkorea/core:

- this is what a package module is https://dlang.org/spec/module.html#package-module
- The need to "merge overload sets" is described in bullet point 5 here: https://dlang.org/spec/function.html#overload-sets

Geod24:
```
Split validation into a package so that it is easier to see which tests
apply to what method and to see dependencies.
```

Fixes #704


source/agora/consensus/validation/Enrollment.d
```diff
@@ -0,0 +1,217 @@
...
+    {
+        address = Point(utxo_set_value.output.address);
+    }
+    catch (Exception ex)
+    {
+        return "Error converting address to point";
+    }
+
+    if (!verify(address, enrollment.enroll_sig, enrollment))
+        return "Enrollment signature verification has an error.";
```
AndrejMitrovic, line 72: I know this PR is just moving things around.

Anyway we could reword this a bit, because it sounds like the verification process itself has an error rather than the signature being invalid.

```suggestion
        return "Enrollment signature is invalid";
```

source/agora/consensus/validation/Transaction.d
```diff
@@ -1,35 +1,26 @@
 /*******************************************************************************
 
-    Contains validation routines for all data types required for consensus.
+    Contains validation rountines for transactions
```
AndrejMitrovic:
```suggestion
    Contains validation routines for transactions
```

source/agora/consensus/validation/package.d
```diff
@@ -0,0 +1,48 @@
+/*******************************************************************************
+
+    Contains validation routines for all data types required for consensus.
+
+    We split validation in two separate phases:
+    - Validation that does not require a 'context';
+    - Validation that does require a context;
+
+    We refer to context-less checks as 'syntax' checks, and context-using
+    checks as 'semantic' checks (even though those terms are misnommers).
```
AndrejMitrovic:
```suggestion
    checks as 'semantic' checks (even though those terms are misnomers).
```

AndrejMitrovic:
There's a missing import to `import agora.common.crypto.Schnorr;` in `Block.d`. Otherwise just a few typos, LGTM.

source/agora/consensus/validation/Enrollment.d
Geod24: There's a couple things I actually want to do after this split. Fixing the wording is one of them, and tiny things like removing full stop. The other is to add some `scope`, etc..
So I think we can leave this for the next PR where we revisit all messages to be consistent, what do you think ?

Geod24:
Fixed

AndrejMitrovic (approved):

PR #843, opened 2020-05-15
Check validity of pre-image with random seed

### LINKED ISSUE #694: Add validation checking code for pre-images being revealed

linked0:
If there is no pre-image revealed previously, the validity of a newly received pre-image should be checked with the random seed of an enrollment. All the operations for checking the validity are based on the distance of two pre-image hashes.

Last commit is about fixing an bug related to calculating distance in order to get pre-image, but the bug had been found when applying the new validity checking code. So I include the commit in this PR. If it's a problem, I will make another PR.

Fixes #694

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -588,7 +588,7 @@ public class EnrollmentManager
 
         preimage.enroll_key = this.data.utxo_key;
         preimage.distance = index;
-        preimage.hash = this.cycle.preimages[index];
+        preimage.hash = this.cycle.preimages[$ - index - 1];
         return true;
     }
 
@@ -943,7 +943,9 @@ unittest
     assert(man.addValidator(enroll, 10, &storage.findUTXO));
     assert(!man.getPreimage(10, preimage));
     assert(man.getPreimage(11, preimage));
+    assert(preimage.hash == man.cycle.preimages[$ - 1]);
     assert(man.getPreimage(10 + Enrollment.ValidatorCycle, preimage));
+    assert(preimage.hash == man.cycle.preimages[0]);
     assert(!man.getPreimage(11 + Enrollment.ValidatorCycle, preimage));
 
     /// test for the functions about periodic revelation of a pre-image


source/agora/consensus/validation/PreImage.d
```diff
@@ -47,21 +47,35 @@ version (unittest)
 *******************************************************************************/
 
 public string isInvalidReason (const ref PreImageInfo new_image,
-    const ref PreImageInfo prev_image, uint validator_cycle) nothrow @safe
+    const ref PreImageInfo prev_image, uint validator_cycle, Hash random_seed)
```
AndrejMitrovic: Needs documentation for `random_seed`

linked0:
Ready for review.

source/agora/consensus/validation/PreImage.d
AndrejMitrovic: Ah you just added it.

Geod24:
Instead of the first two commit, why didn't you go with:
```diff
diff --git a/source/agora/consensus/ValidatorSet.d b/source/agora/consensus/ValidatorSet.d
index 9a6b3a70..07dea8ac 100644
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -113,8 +113,9 @@ public class ValidatorSet

             () @trusted {
                 this.db.execute("INSERT INTO validator_set " ~
-                    "(key, val, enrolled_height) VALUES (?, ?, ?)",
-                    enroll.utxo_key.toString(), buffer, block_height);
+                    "(key, val, enrolled_height, preimage) VALUES (?, ?, ?, ?)",
+                    enroll.utxo_key.toString(), buffer, block_height,
+                    PreImageInfo(enroll.random_seed, block_height));
             }();
         }
         catch (Exception ex)
```

linked0:
> Instead of the first two commit, why didn't you go with:

The `PreImageInfo` has `distance` not `height`, so there is no way to represent the random seed in `PreImageInfo` because `cycle.preimages[0]` is the pre-image of the random seed.

Geod24:
Ah, then use `0`, not block height. 

> `cycle.preimages[0]` is the pre-image of the random seed

But we're talking about `cycle.preimages[$]` here, no ? In any case, there is definitely a better way to handle this than to leak this into the validation routine.

linked0:
> Ah, then use `0`, not block height.
> 
> > `cycle.preimages[0]` is the pre-image of the random seed
> 
> But we're talking about `cycle.preimages[$]` here, no ? In any case, there is definitely a better way to handle this than to leak this into the validation routine.

Right, I meant cycle.preimages[$]. Sorry for the confusion.
`cycle.preiamges[$]` is the pre-image of the random seed.

source/agora/consensus/validation/PreImage.d
```diff
@@ -47,21 +51,39 @@ version (unittest)
...
+    else if (random_seed != Hash.init)
+    {
+        distance = new_image.distance + 1;
+        prev_hash = random_seed;
+    }
+    else
+    {
+        return "The random seed is invalid";
+    }
+
```
Geod24: We discussed this IRL, but I think it's important to re-state it: We should not change the algorithm to adapt to the data. Or at least, not this much. We're essentially duplicating a part of the validation routine to adapt for the special case of the `random_seed`. Except that it's not a special case. One can just think of the `random_seed` as the *first* pre-image, and store it in the table. This way, we always have a pre-image, and we don't need all this complication.

linked0:
I will apply the code of the PR #848  into this PR today.

linked0:
Ready to review.
The main part of this PR has been handled in the PR #848 and #851  by @Geod24 . So only the bug fix about getting a pre-image of a Node remains in this PR.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -588,7 +588,8 @@ public class EnrollmentManager
 
         preimage.enroll_key = this.data.utxo_key;
         preimage.distance = index;
-        preimage.hash = this.cycle.preimages[index];
+        preimage.hash =
+            this.cycle.preimages[Enrollment.ValidatorCycle - index - 1];
```
```suggestion
            this.cycle.preimages[$ - index - 1];
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -943,7 +944,9 @@ unittest
     assert(man.addValidator(enroll, 10, &storage.findUTXO));
     assert(!man.getPreimage(10, preimage));
     assert(man.getPreimage(11, preimage));
+    assert(preimage.hash == man.cycle.preimages[$ - 1]);
```
Geod24, line 946: It would be better if we had tests about the interface of this function rather than the implementation.
The previous behavior was completely broken, and the code introduced should catch that without a doubt, but at the moment I'm not sure it does that. We really need more network test for pre-image and Enrollments.

linked0: I agree that the test code is incomplete. I will add more test code for pre-image and enrollments in the agora.test.EnrollmentManager module.

Geod24 (approved):

PR #841, opened 2020-05-14
Remove unused quorum topologies from the tests

AndrejMitrovic:
Part of #785. I wanted to split it off from the main PR to simplify it.

Issue #836, opened 2020-05-14
Improve usage of SQLite for UTXO & validator DBs


Geod24:
We are currently using SQLite for a bunch of node metadata. The intent was to have something that is extremely reliable. However it lead to less-than-pleasant code.

There are a few issues:
- The code is neither `@safe` nor `nothrow` which forces us to liter the code with `try / catch`, `@trusted`, and `scope (failure) assert(0);`, e.g. [here](https://github.com/bpfkorea/agora/blob/d396373156c94621b3d41cbe124ce8bc2dfca873/source/agora/consensus/data/UTXOSet.d#L409-L414), [here](https://github.com/bpfkorea/agora/blob/d396373156c94621b3d41cbe124ce8bc2dfca873/source/agora/consensus/EnrollmentPool.d#L119-L130), or [here](https://github.com/bpfkorea/agora/blob/d396373156c94621b3d41cbe124ce8bc2dfca873/source/agora/consensus/ValidatorSet.d#L103-L123).
- There is no usage of transactions, so the underlying problem of consistency is not fixed. For example, if we were to crash during `addValidatedBlock` (think power outage), it is very likely our node wouldn't recover from this, nor be able to detect the failure.
- There are some blob being stored in the DB. I don't know if it's the most sensible choice. Should we take the storage overhead and store binary data as hex string ? For hash it would make sense.
- [In the ValidatorSet](https://github.com/bpfkorea/agora/blob/d396373156c94621b3d41cbe124ce8bc2dfca873/source/agora/consensus/ValidatorSet.d#L103-L116), the pre-image info is duplicated. The `random_seed` is actually the first pre-image and could be treated as such. There is an odd mix of relational and non-relational data, were we store some of the data from the enrollment non-serialized, and the whole enrollment serialized as well.

My original thought was to simplify all of this and use memory, and just commit from time to time. @AndrejMitrovic pointed out that SQLite in fact has many advantages, which solve most of our problems. So the fault probably lies in our binding / usage.

Old description:
> At the moment the UTXO DB relies on a SQLite database, which means that the node can fail if there are disk error. That shouldn't happen, instead we should have this data in memory, with the SQLite being a backing store.

Geod24:
This should also apply to the validator DB.

Geod24:
Things to note:
[...]

Issue #824, opened 2020-05-13
It should be possible to re-enroll a Validator while it is in the validator set


AndrejMitrovic:
As discussed with @Geod24. This is similar to https://github.com/bpfkorea/agora/issues/822, but is a different feature.

Consider this case:
- Genesis block begins with 2 validators
- Their expiration block is at height 1008
- At block height 1008 the validators want to re-enroll so they can begin validating blocks [1009 .. 2017]

Problem:
- A Validator's Enrollment will be rejected at block height 1008 because the Validator's enrollment is already active at that block height.
- The Validator must wait until block height 1009 to re-enroll.
- In the context of #822 in this specific case this means the blockchain will halt. Even if #822 was fixed the blockchain will still halt because the Validators cannot re-enroll while being an active Validator.

PR #845, opened 2020-05-15
Implement shell quorum balancing

AndrejMitrovic:
Requires:
~#800~, ~#840~, ~#810~, ~#841~, ~#881~

### DIFF
--- a/source/agora/consensus/Quorum.d
+++ b/source/agora/consensus/Quorum.d
@@ -1,0 +1,192 @@
+/*******************************************************************************
+
+    Contains the quorum generator algorithm.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.Quorum;
+
+import agora.common.Config;
+import agora.common.crypto.Key;
+import agora.common.Set;
+import agora.common.Types;
+import agora.consensus.data.Enrollment;
+import agora.consensus.UTXOSet;
+
+import std.algorithm;
+import std.array;
+import std.exception;
+import std.format;
+import std.math;
+import std.range;
+import std.string;
+
+/*******************************************************************************
+
+    Build the quorum configuration for the given public key and the staked
+    enrollments.
+
+    Params:
+        node_key = the key of the node for which to generate the quorum
+        enrolls = the array of registered enrollments
+        finder = UTXO finder delegate
+
+    Returns:
+        the generated quorum config
+
+*******************************************************************************/
+
+public QuorumConfig buildQuorumConfig ( const ref PublicKey node_key,
+    in Enrollment[] enrolls, in UTXOFinder finder )
+{
+    Set!PublicKey all_nodes = getPublicKeys(enrolls, finder);
+
+    QuorumConfig quorum;
+    all_nodes.each!(node => quorum.nodes ~= node);
+    quorum.nodes.sort();
+    quorum.threshold = quorum.nodes.length;
+
+    return quorum;
+}
+
+///
+unittest
+{
+    // test up to 1024 nodes
+    foreach (num_nodes; iota(1, 10).map!(n => 2 ^^ n))
+    {
+        auto data = genTestEnrolls(num_nodes);
+        foreach (key; data.keys)
+        {
+            auto quorum = buildQuorumConfig(key, data.enrolls, data.finder);
+            verifyQuorumSanity(quorum);
+            const expected = QuorumConfig(num_nodes, data.keys);
+            assert(quorum == expected,
+                format("Expected: %s. Got: %s", expected, quorum));
+        }
+    }
+}
+
+/*******************************************************************************
+
+    Verify that the provided quorum configuration is considered sane by SCP.
+
+    Params:
+        quorum = the quorum config
+
+    Throws:
+        an Exception if the quorum is not considered sane by SCP.
+
+*******************************************************************************/
+
+private void verifyQuorumSanity (const ref QuorumConfig quorum)
+{
+    import scpd.scp.QuorumSetUtils;
+    auto scp_quorum = toSCPQuorumSet(quorum);
+
+    const bool ExtraChecks = true;
+    const(char)* fail_reason;
+    enforce(isQuorumSetSane(scp_quorum, ExtraChecks, &fail_reason),
+        format("Quorum %s fails sanity check before normalization: %s",
+                quorum, fail_reason.fromStringz));
+
+    normalizeQSet(scp_quorum);
+    enforce(isQuorumSetSane(scp_quorum, ExtraChecks, &fail_reason),
+        format("Quorum %s fails sanity check after normalization: %s",
+                quorum, fail_reason.fromStringz));
+}
+
+/*******************************************************************************
+
+    For each enrollment find the staked amount from the associated UTXO
+    in the Enrollment, and build a key => amount map.
+
+    Params
+        enrolls = the list of enrollments
+        finder = UTXO finder delegate
+
+    Returns:
+        a mapping of all keys => staken amount
+
+*******************************************************************************/
+
+private Set!PublicKey getPublicKeys (in Enrollment[] enrolls,
+    in UTXOFinder finder)
+{
+    Set!PublicKey stakes;
+    foreach (enroll; enrolls)
+    {
+        UTXOSetValue value;
+        if (!finder(enroll.utxo_key, size_t.max, value))
+            assert(0, "UTXO for validator not found!");  // should never happen
+
+        if (value.output.address in stakes)  // should never happen
+            assert(0, "Cannot have multiple enrollments for one validator!");
+
+        stakes.put(value.output.address);
+    }
+
+    return stakes;
+}
+
+/*******************************************************************************
+
+    Params:
+        num_enrollments = the number of enrollments to create
+
+    Returns:
+        a tuple of (Enrollment[], UTXOFinder)
+
+*******************************************************************************/
+
+version (unittest)
+private auto genTestEnrolls (size_t num_enrollments)
+{
+    struct Result
+    {
+        PublicKey[] keys;
+        Enrollment[] enrolls;
+        UTXOFinder finder;
+    }
+
+    import agora.common.Amount;
+    import agora.consensus.data.Transaction;
+
+    TestUTXOSet storage = new TestUTXOSet;
+    KeyPair[] key_pairs = num_enrollments.iota.map!(x => KeyPair.random()).array;
+
+    foreach (idx; 0 .. num_enrollments)
+    {
+        Transaction tx =
+        {
+            type : TxType.Freeze,
+            outputs: [Output(Amount.MinFreezeAmount, key_pairs[idx].address)]
+        };
+
+        storage.put(tx);
+    }
+
+    Enrollment[] enrolls;
+    foreach (utxo; storage.keys)
+    {
+        const Enrollment enroll =
+        {
+            utxo_key : utxo,
+            cycle_length : 1008
+        };
+
+        enrolls ~= enroll;
+    }
+
+    // cast: we need a mutable so we can sort
+    PublicKey[] keys = key_pairs.map!(k => cast()k.address).array;
+    keys.sort();
+    return Result(keys, enrolls, &storage.findUTXO);
+}
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -63,6 +63,10 @@ public class Ledger
     /// Node config
     private NodeConfig node_config;
 
+    /// If not null call this delegate if the validator set changed after
+    /// a block was externalized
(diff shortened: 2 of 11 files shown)


AndrejMitrovic:
Update: Added #810 to the list of required PRs. I marked that commit as `[remove]` in the PR.

source/agora/node/Validator.d
```diff
@@ -112,7 +174,8 @@ public class Validator : FullNode, API
 
             log.info("Doing network discovery..");
             this.network.discover(required_peer_keys);
-            this.network.startPeriodicCatchup(this.ledger, &this.nominator.isNominating);
+            this.network.startPeriodicCatchup(this.ledger,
```
AndrejMitrovic: Unrelated change.

AndrejMitrovic:
Need to fix the integration test.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,192 @@
...
+    quorum.threshold = quorum.nodes.length;
+
+    return quorum;
+}
+
+///
+unittest
+{
+    // test up to 1024 nodes
+    foreach (num_nodes; iota(1, 10).map!(n => 2 ^^ n))
```
Geod24, line 63: That's how I felt

AndrejMitrovic: Yeah sometimes I find myself abusing ranges just to write something clever. I could have just written 2, 4, 8, 16, 32, 64, 128.. etc I guess.

AndrejMitrovic:
It seems the test-suite times out (30 minutes). I'm trying to bisect it.

source/agora/test/Base.d
```diff
@@ -939,23 +949,23 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
...
-        node_configs = the list of node configurations
+        key_pairs = key pairs for signing enrollments with
 
     Returns:
         The genesis block
 
 *******************************************************************************/
 
-private immutable(Block) makeGenesisBlock (in NodeConfig[] node_configs)
+private immutable(Block) makeGenesisBlock (in KeyPair[] key_pairs)
```
AndrejMitrovic: This change should be moved to the previous commit where the function was introduced.

AndrejMitrovic:
I've discovered that what is causing these failures is actually the code coverage mode. The test-suite works fine without it. But with it there are timeouts. Either the code coverage is significantly slowing down the tests, or there is a bug in the coverage generator code in DMD front-end.

AndrejMitrovic:
So far I can't reproduce the segfault when running github CI locally, but I can reproduce the timeout issue:

```
(lldb) bt
* thread #1, queue = 'com.apple.main-thread', stop reason = signal SIGSTOP
  * frame #0: 0x00007fff6099e9de libsystem_kernel.dylib`__ulock_wait + 10
    frame #1: 0x00007fff60a5f6de libsystem_pthread.dylib`_pthread_join + 358
    frame #2: 0x000000010e81e485 agora-unittests`thread_joinAll + 197
    frame #3: 0x000000010e839225 agora-unittests`rt_term + 69
    frame #4: 0x000000010e83977c agora-unittests`_D2rt6dmain212_d_run_main2UAAamPUQgZiZ6runAllMFZv + 172
    frame #5: 0x000000010e839537 agora-unittests`_d_run_main2 + 391
    frame #6: 0x000000010e83939d agora-unittests`_d_run_main + 141
    frame #7: 0x000000010deaecd5 agora-unittests`main(argc=1, argv=0x00007ffee1ea4320) + 37
    frame #8: 0x00007fff608673d5 libdyld.dylib`start + 1
```

It seems the process is stuck on `thread_joinAll`

AndrejMitrovic:
Ok I've added a Thread.sleep call to work around #869, but now there is a consistent segfault in the NetworkClient. I've seen this error before too:

```
Thread 40 "agora-unittests" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffe836b0700 (LWP 19915)]
0x00000000081a4493 in _D5agora7network13NetworkClientQp12sendPreimageMFNeKSQBz9consensus4data12PreImageInfoQoZ9__lambda2MFZ12__dgliteral1MFNaNiNfZv ()
    at /mnt/c/dev/agora/source/agora/network/NetworkClient.d:327
327                 this.attemptRequest(this.api.receivePreimage(preimage), null);
(gdb)

#2  0x00000000081a444d in _D5agora7network13NetworkClientQp12sendPreimageMFNeKSQBz9consensus4data12PreImageInfoQoZ9__lambda2MFZv ()
    at /mnt/c/dev/agora/source/agora/network/NetworkClient.d:327
327                 this.attemptRequest(this.api.receivePreimage(preimage), null);
(gdb) p api
$1 = (agora.api.Validator.API *) 0x0
```

So the `api` is null.

AndrejMitrovic:
It seems that disabling parallelism fixes the issue. Unfortunately I cannot reproduce the issue locally. And also, when running the CI test-suite under LLDB it also cannot reproduce the issue.

AndrejMitrovic:
There is something wrong with the implementation of `dispatch` in `FiberScheduler`. We lifted it from `std.concurrency`, and it was documented as an "example scheduler". 

This is my workaround which gets rid of segfaults: https://github.com/AndrejMitrovic/localrest/compare/2cf119e759cdbbe2879bf65435efd18084620498...6f35619393bfdfbb4b9f84379da68d8f40c3675a#diff-7e552a35c6e12cbbfb124195a3e45cd2

I would guess that it has something to do with manipulating the `m_fibers` array both in the `dispatch` routine and in `create()`: https://github.com/AndrejMitrovic/localrest/compare/AndrejMitrovic:2cf119e...AndrejMitrovic:6f35619#diff-7e552a35c6e12cbbfb124195a3e45cd2R936

The end result before this patch is that the fiber somehow ends up jumping to a random source code location. So it seems like the fiber gets corrupt somehow.

I still need to work on reducing this further down to a test-case..

AndrejMitrovic:
It turned green. wow

AndrejMitrovic:
Blocked on https://github.com/bpfkorea/agora/issues/824

And https://github.com/bpfkorea/agora/pull/889 should improve the way networking is done after the re-shuffle.

AndrejMitrovic:
Rebased. Still blocked, and I had to revert the change Mathias made in order to fix the tests.

AndrejMitrovic:
I'll try to unblock this by changing the test to use new validator sets instead of re-enrolling existing validators.

AndrejMitrovic:
I should split off the first 4 commits so they can be merged right away.

AndrejMitrovic:
Test-suite passes, but it's flakey unfortunately.

There still seems to be an issue where a task (periodic preimage reveal) starts after a sleep call and ends up segfaulting. I haven't tracked it down yet.

AndrejMitrovic:
The feature I discovered in https://github.com/Geod24/localrest/pull/81 came in useful. I rewrote one test.

AndrejMitrovic:
I further simplified the changeset. The diff is much smaller now.

AndrejMitrovic:
@bpfkorea/core please review when you have the time.

source/agora/node/Validator.d
```diff
@@ -51,24 +58,75 @@ public class Validator : FullNode, API
...
+        Enrollment[] enrollments;
+        if (!this.enroll_man.getValidators(enrollments) ||
+            enrollments.length == 0)
+        {
+            log.fatal("Could not retrieve enrollments / no enrollments found");
+            assert(0);
+        }
+
+        return .buildQuorumConfig(this.config.node.key_pair.address,
+            enrollments, this.utxo_set.getUTXOFinder());
```
TrustHenry, line 129: agora.consensus.Quorum.buildQuorumConfig ( const ref PublicKey node_key,
    in Enrollment[] enrolls, in UTXOFinder finder )
I think it's pointing here.
There's no problem. But The code is difficult to read.

source/agora/node/Ledger.d
```diff
@@ -228,6 +228,9 @@ public class Ledger
         this.enroll_man.clearExpiredValidators(block.header.height);
 
         foreach (idx, ref enrollment; block.header.enrollments)
+        {
+            this.enroll_man.pool.remove(enrollment.utxo_key);
+
             if (auto r = this.enroll_man.addValidator(enrollment,
                 block.header.height, this.utxo_set.getUTXOFinder()))
```
TrustHenry, line 248: this.enroll_man.pool.remove
The `remove` is run twice here.

source/agora/node/Ledger.d
```diff
@@ -228,6 +228,9 @@ public class Ledger
         this.enroll_man.clearExpiredValidators(block.header.height);
 
         foreach (idx, ref enrollment; block.header.enrollments)
+        {
+            this.enroll_man.pool.remove(enrollment.utxo_key);
```
Geod24, line 245: https://github.com/bpfkorea/agora/blob/d2931dd0d0f552997bc80e799a5275f5c3cecff5/source/agora/consensus/EnrollmentManager.d#L214

TrustHenry (commented):
Overall, it looks Great.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,192 @@
...
+*******************************************************************************/
+
+private void verifyQuorumSanity (const ref QuorumConfig quorum)
+{
+    import scpd.scp.QuorumSetUtils;
+    auto scp_quorum = toSCPQuorumSet(quorum);
+
+    const bool ExtraChecks = true;
+    const(char)* fail_reason;
+    enforce(isQuorumSetSane(scp_quorum, ExtraChecks, &fail_reason),
```
MichaelKim20, line 96: I think it would be nice to have a test code for `isQuorumSetSane`.

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,192 @@
...
+        format("Quorum %s fails sanity check before normalization: %s",
+                quorum, fail_reason.fromStringz));
+
+    normalizeQSet(scp_quorum);
```
MichaelKim20, line 100: I think it would be nice to have a test code for `normalizeQSet `.

source/agora/test/Quorum.d
```diff
@@ -0,0 +1,157 @@
...
+    nodes.enumerate.each!((idx, node) =>
+        retryFor(node.getBlockHeight() == 1007, 5.seconds,
+            format("Node %s has block height %s. Expected: %s",
+                idx, node.getBlockHeight(), 1007)));
+
+    // now we can create enrollments
+    Enrollment enroll_0 = nodes[$ - 2].createEnrollmentData();
+    Enrollment enroll_1 = nodes[$ - 1].createEnrollmentData();
+    nodes[2].enrollValidator(enroll_0);
+    nodes[3].enrollValidator(enroll_1);
```
Geod24, line 116: Question about this: Should we add the ability to enroll as soon as we freeze ?
Pro:
- Genesis does it;
- It allows to potentially solve the censorship issue if paired with a rule that more validators => Higher priority;
Cons:
- Attack vector ?

Geod24 (approved):

Geod24:
I don't think the double clear should block the merge.

source/agora/test/Quorum.d
AndrejMitrovic: Let's file an issue for this. And yes, I wanted this feature too.

source/agora/node/Ledger.d
AndrejMitrovic: Where do you see it?

Geod24: https://github.com/bpfkorea/agora/pull/845#discussion_r435748483

source/agora/consensus/Quorum.d
```diff
@@ -0,0 +1,192 @@
...
+    quorum.threshold = quorum.nodes.length;
+
+    return quorum;
+}
+
+///
+unittest
+{
+    // test up to 1024 nodes
+    foreach (num_nodes; iota(1, 10).map!(n => 2 ^^ n))
```
TrustHenry, line 63: // test up to 1022 nodes

```
unittest
{
    import std.algorithm;
    import std.stdio;
    import std.range;

    uint sum;
    foreach (num_nodes; iota(1, 10).map!(n => 2 ^^ n))
       sum+=num_nodes;
    writeln(sum);
}
```
Result  : 1022

AndrejMitrovic: It's not testing all node count configurations. That would be too expensive (especially once the algorithm is changed).

What I am testing here is this:

```
num_nodes : 2
num_nodes : 4
num_nodes : 8
num_nodes : 16
num_nodes : 32
num_nodes : 64
num_nodes : 128
num_nodes : 256
num_nodes : 512
num_nodes : 1024
```

See?

AndrejMitrovic: Well actually it's not inclusive as I forgot to use `11` in the iota call. +1 points to you.

source/agora/node/Ledger.d
AndrejMitrovic: Ah, right. I think in my fixes for #836 I will remove the call within `addValidator` because that function can fail.

Issue #820, opened 2020-05-12
Error code is returned when unittest is performed on Windows


TrustHenry:
`Running .\build\agora-unittests.exe Program exited with code -1073740791`

When debugging through the windbg, the following message is printed.

```
The code execution cannot proceed because libssl-1_1-x64.dll
Was not found. Reinstalling the program may fix this problem.
```

```
The code execution cannot proceed because libcrypto-1_1-x64.dll
Was not found. Reinstalling the program may fix this problem.
```

Definition done

- [ ] Unit testing in the window environment should normally be returned without error codes.

AndrejMitrovic:
> libel-1_1-x64.dll 
> Libctypto-1_1-x64.dll

You mean `libssl-1_1-x64.dll` and `libcrypto-1_1-x64.dll`? Because these look like typos.

TrustHenry:
Oh, that's right. It's a typo:

AndrejMitrovic:
It looks like you need to install OpenSSL.

TrustHenry:
Yes, I think it should be installed,
but I'm trying to see if it's included in Visual Studio 2019 runtime.
https://wiki.openssl.org/index.php/Binaries

TrustHenry:
The library was being used by vive-d.
I've figured out the location.
`..\submodules\vibe.d\lib\win-amd64\libcrypto-1.1-x64.dll`

TrustHenry:
```
CommandLine: C:\dev\agora\build\agora-unittests.exe
Symbol search path is: srv*
Executable search path is: 
ModLoad: 00007ff6`2af40000 00007ff6`2b85b000   agora-unittests.exe
ModLoad: 00007ffa`e3560000 00007ffa`e3750000   ntdll.dll
ModLoad: 00007ffa`e3230000 00007ffa`e32e2000   C:\Windows\System32\KERNEL32.DLL
ModLoad: 00007ffa`e1100000 00007ffa`e13a3000   C:\Windows\System32\KERNELBASE.dll
ModLoad: 00007ffa`e1f90000 00007ffa`e2033000   C:\Windows\System32\ADVAPI32.dll
ModLoad: 00007ffa`e3480000 00007ffa`e351e000   C:\Windows\System32\msvcrt.dll
ModLoad: 00007ffa`e1ec0000 00007ffa`e1f57000   C:\Windows\System32\sechost.dll
[...]

PR #846, opened 2020-05-15
Fix the Enable Windows version unit test

TrustHenry:
This runs a unit test of the Windows version.
In the Windows version, due to problems with the parallelism test,
the test is performed with single threads.
And Fix the can't allocate of push_back

Related to #820

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -150,7 +150,12 @@ private UnitTestResult customModuleUnitTester ()
             else if (mod.name == "agora.test.ManyValidators")
                 heavy_tests ~= ModTest(mod.name, fp);
             else
-                parallel_tests ~= ModTest(mod.name, fp);
+                // due to problems with the parallelism test,
+                // the test is performed with single threads
+                version (Windows)
+                    single_threaded ~= ModTest(mod.name, fp);
+                else
+                    parallel_tests ~= ModTest(mod.name, fp);
         }
     }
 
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
I recommended opening the PR with this because the "cannot allocate" issues Henry is seeing might be because of running out of RAM when running the VM. Let's see..

source/agora/test/Base.d
```diff
@@ -185,8 +187,15 @@ private UnitTestResult customModuleUnitTester ()
...
-        runTest(mod);
+    version (Windows)
+    {
+        // TODO
+    }
+    else
+    {
+        foreach (mod; parallel(parallel_tests))
+            runTest(mod);
+    }
```
Geod24: No! The code I gave you yesterday would run the test single-threaded. This one just excludes the test, which is absolutely not what we want.

TrustHenry: Yes, I proceeded with the missing test. 
In the Windows version, modify to test only single threads.

TrustHenry:
It feels good. It's all green. without (ci/dockercloud)
All of Fiber's stackSize=512KB were successful in CI Tests on Ubuntu, macOS, and Windows-2019 OS.

source/agora/test/Base.d
```diff
@@ -183,13 +188,22 @@ private UnitTestResult customModuleUnitTester ()
 
     // Run single-threaded tests
     foreach (mod; single_threaded)
+    {
+        writefln("single_thread_tests/%s", mod.name);
```
TrustHenry: I added a print for the test. I'm going to remove it.

TrustHenry:
What do we think about changing the `Fiber` stack size of the `localrest`?

TrustHenry:
Ready for review. All Green!

AndrejMitrovic (approved):

Issue #847, opened 2020-05-15
`PreImageCache` should contain the random seed of an enrollment


linked0:
The `PraImageCache` is used to manage all the pre-images for a cycle of an enrollment. We get a pre-image from the class based on distance if needed. The `EnrollmentManager` or `ValidatorSet` need to get the random seed of an enrollment from the class. But, it does not have the random seed, there is no way to get the random seed from the class. So the 'PreImageCache` should return the last hash of it as a random seed or save the random seed.

linked0:
All the work needed was done.

PR #848, opened 2020-05-17
Fix pre-image handling & issue 847

Geod24:
This is quite embarrassing... The way the `Cycle` structure work is that it contains a two-layers cache, and the first layer is the seed of every enrollment, while the second layer contains the pre-images for the current cycle.

```
L1: [0, 1008, 2016, 3024, 4032, ..., 10800]
L2: [0 .. 1007] or [1008 .. 2015] or [2016 .. 3023], etc...
```

The were multiple problems with the code:
- First and foremost, while cycling through the L1, we would first reveal [0 .. 1007], then [1008 .. 2015], etc... Those are the last that should be revealed, we should start with [9798 .. 10800] !
- When resetting the cycle (once all enrollments are exhausted) there was an OB1 error leading to a `RangeError`;
- The way things were stored was overly complicated and not very usable (see #847). This has been refactored.
- There was no proper test of the behavior;

So all of those are fixed in those commits. In addition, I added a `toString`, which proved to be a valuable debugging utility. It should unblock some of @linked0 's PRs too.

### DIFF
--- a/source/agora/common/Hash.d
+++ b/source/agora/common/Hash.d
@@ -478,10 +478,6 @@ unittest
             assert(cache.data.length == 1);
             assert(cache.data[0] == data[0]);
             break;
-        case 0:
-            assert(cache.data.length == 32);
-            assert(cache.data[] == data[]);
-            break;
 
         default:
             assert(0);
--- a/source/agora/common/Hash.d
+++ b/source/agora/common/Hash.d
@@ -426,6 +426,9 @@ public struct PreImageCache
         return value;
     }
 
+    /// Alias to the underlying data, useful when dealing with multiple levels
+    public const(Hash)[] byStride () const pure @nogc { return this.data; }
+
     /// Returns: The number of preimages this cache can represent
     public size_t length () const pure nothrow @nogc @safe
     {
--- a/source/agora/common/Hash.d
+++ b/source/agora/common/Hash.d
@@ -317,7 +317,72 @@ nothrow @nogc @safe unittest
 
 public struct PreImageCache
 {
-    @safe nothrow:
+    @safe:
+
+    /// Describe print modes for `toString`
+    public enum PrintMode
+    {
+        /// Default value, just print the two extreme values
+        Bounds,
+        /// Print all stored intermediate value and the last bound
+        Stride,
+        /// Print all values, performing intermediate hashing
+        All,
+    }
+
+    /***************************************************************************
+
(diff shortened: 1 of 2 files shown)


Geod24:
Rebased

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1024,3 +1028,67 @@ unittest
...
+    auto cycle = EnrollmentManager.PreImageCycle(
+        /* nounce: */ 0,
+        /* index:  */ 0,
+        /* seeds:  */ PreImageCache(EnrollmentManager.NumberOfCycles, Enrollment.ValidatorCycle),
+        /* preimages: */ PreImageCache(Enrollment.ValidatorCycle, 1)
+    );
+
+    auto secret = Scalar.random();
+    Scalar fake_secret; // Used whenever `secret` *shouldn't* be used
+    foreach (uint cycleCount; 0 .. 10)
```
linked0: The `cycle` in the `cycleCount` is confusing with the `Cycle` in the `NumberOfCycles`. I'd like you to change `cycleCount` into another word like `cycleGroupCount`

linked0 (commented):
Great job and thank your for your support about handling pre-images.
I have one small comment about this PR.

source/agora/consensus/EnrollmentManager.d
Geod24: Done

Geod24:
Green, anything else ?

linked0:
Nothing. I guess it could be merged. I will approve.

linked0 (approved):
LGTM

source/agora/common/Hash.d
```diff
@@ -317,7 +317,72 @@ nothrow @nogc @safe unittest
...
+
+        Params:
+          sink = The sink to write the piecemeal string data to
+          mode = The `PrintMode` to use for printing the content.
+                 By default, only the bounds (0 and $) are printed.
+
+    ***************************************************************************/
+
+    public void toString (scope void delegate(scope const(char)[]) @safe sink,
+                          PrintMode mode = PrintMode.Bounds) const
```
AndrejMitrovic, line 352: Can you use this with `format` or you have to call it explicitly?

Geod24: You have to call it explicitly

AndrejMitrovic (approved):

PR #850, opened 2020-05-17
Improve and simplify ValidatorSet tests

Geod24:
As I was trying to adapt them to check for the pre-image when it's there from the beginning, I hit an issue with some of the assumptions being made, which this is removing.


linked0 (approved):
All the refactoring on the PR makes the code simpler and clearer.
LGTM.

PR #851, opened 2020-05-18
Store the commitment as pre-image in the validator table

Geod24:
Since the random_seed is actually a pre-image (the first one),
store it as such in the ValidatorSet.
This has the nice effect of automatically fixing issue 694,
since `addPreimage` now checks the distance when a new
`PreImageInfo` is added.

### DIFF
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -110,11 +110,14 @@ public class ValidatorSet
         {
             static ubyte[] buffer;
             serializeToBuffer(enroll, buffer);
+            const img = PreImageInfo(enroll.utxo_key, enroll.random_seed, 0);
+            static ubyte[] pbuffer;
+            serializeToBuffer(img, pbuffer);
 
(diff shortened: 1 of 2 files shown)


source/agora/consensus/ValidatorSet.d
```diff
@@ -108,13 +108,15 @@ public class ValidatorSet
 
         try
         {
-            static ubyte[] buffer;
+            static ubyte[] buffer, pbuffer;
```
AndrejMitrovic: I prefer moving variable declarations as close to where they're initialized as possible. So just above the `serializeToBuffer(img, pbuffer);` call. 😇

linked0 (approved):
LGTM

source/agora/consensus/ValidatorSet.d
Geod24: Variables reorganized my good sir!

AndrejMitrovic (approved):

AndrejMitrovic:
merge at will

PR #859, opened 2020-05-20
Run Serializer test first

AndrejMitrovic:
There was still an issue where node threads can cause the GC allocation stats to change
while this test was running.

Just happened to me locally now.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -135,6 +135,8 @@ private UnitTestResult customModuleUnitTester ()
                 continue;
             }
 
+            // this test checks GC usage stats before / after tests,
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
Just to be a little bit more specific:

The `shutdown` calls are asynchronous, so it's possible the `parallel` loop there returns while some nodes are still active and periodically allocating.

Geod24 (approved):
Merge at will

AndrejMitrovic:
There is some build artifact failure, but the PR is still green? hmm

Geod24:
Maybe Github doesn't trust their own infra ?

Issue #582, opened 2020-02-24
Reveal pre-images on a timer


Geod24:
Instead of having pre-image reveal happening on `putTransaction`, it should be done via a timer.
Currently this is done here: https://github.com/bpfkorea/agora/blob/0cdc928b09eddec41f78a2c1a36459dac532a9c6/source/agora/node/Node.d#L252-L260

This can be trivially done via Vibe.d's [setTimer](https://vibed.org/api/vibe.core.core/setTimer) function.
It is currently not implemented in LocalRest, but since we have non-blocking `sleep` and `runTask` as primitives, it should be trivial to implement it.

Issue #520, opened 2020-02-03
Make the meta data for enrollment configurable


linked0:
There are some meta data to be configurable for enrollment and pre-image. So we will select the meta data for being configurable and make the `EnrollmentManage` restore the data.

Enrolled validators should always make sure that their preimage are available to other nodes in a timely manner. The issue is what is the best period for revealing the preimage. So It should be cofigurable as an item like  in the configuration file.

 Definition of Done:

- [x] Include the period of revealing pre-image, or `reveal_preimage_period` is in the config file 
- [x] Include the an aging period after which an enrollment is deleted from the pool
- [x] Include the period of revelation of pre-image
- [x] Pass the meta value to the enrollment manager.

This is related to #492

Geod24:
This has been addressed.

1) https://github.com/bosagora/agora/blob/2558cd782c4cbcf82aed2d9fe684f85e85511842/source/agora/common/Config.d#L251-L252

2) Enrollments are only valid for one height since #1902 ;

3, 4) See 1

PR #860, opened 2020-05-20
Reveal pre-images on a timer 

TrustHenry:
Described in each commit.

Relates to #582

### DIFF
--- a/source/agora/common/Task.d
+++ b/source/agora/common/Task.d
@@ -19,6 +19,13 @@ module agora.common.Task;
 
 import core.time;
 
+/// Whether the timer periodic type
+public enum Periodic : bool
+{
+    No,
+    Yes
+}
+
 /// Exposes primitives to run tasks through Vibe.d
 public class TaskManager
 {
@@ -50,5 +57,30 @@ public class TaskManager
     {
         static import vibe.core.core;
         vibe.core.core.sleep(dur);
+    }
+
+    /***************************************************************************
+
+        Run an asynchronous task after a given time.
+
+        The task will first run after the given `timeout`, and
+        can either repeat or run only once (the default).
+        See_Also: https://vibed.org/api/vibe.core.core/setTimer
+
+        Params:
+            timeout = Determines the minimum amount of time that elapses before
+                the timer fires.
+            dg = This delegate will be called when the timer fires.
+            periodic = Specifies if the timer fires repeatedly or only once.
+
+    ***************************************************************************/
+
+    public void setTimer (Duration timeout, void delegate() dg,
+        Periodic periodic = Periodic.No)
+    {
+        assert(dg !is null, "Cannot call this delegate if null");
+
+        static import vibe.core.core;
+        vibe.core.core.setTimer(timeout, dg, periodic);
     }
 }
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -33,6 +33,8 @@ import scpd.types.Stellar_SCP;
 
 import ocean.util.log.Logger;
 
+import core.time;
+
 mixin AddLogger!();
 
 /*******************************************************************************
@@ -81,6 +83,8 @@ public class Validator : FullNode, API
 
             log.info("Doing network discovery..");
             this.network.discover(required_peer_keys);
+            // call checkRevealPreimage periodically every 10 seconds
+            this.taskman.setTimer(10.seconds, &this.checkRevealPreimage, Periodic.Yes);
             this.network.startPeriodicCatchup(this.ledger, &this.nominator.isNominating);
         });
     }
@@ -170,5 +174,25 @@ public class Validator : FullNode, API
 
         foreach (sub_conf; quorum_conf.quorums)
             buildRequiredKeys(filter, sub_conf, nodes);
+    }
+
+    /***************************************************************************
+
+        Periodically check for pre-images revelation
+        Increase the next reveal height by revelation period if necessary.
+
+    ***************************************************************************/
+
+    private void checkRevealPreimage () @safe
+    {
+        if (!this.enroll_man.needRevealPreimage(this.ledger.getBlockHeight()))
+            return;
+
+        PreImageInfo preimage;
+        if (this.enroll_man.getNextPreimage(preimage))
+        {
+            this.receivePreimage(preimage);
+            this.enroll_man.increaseNextRevealHeight();
+        }
     }
 }
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -264,6 +264,28 @@ public class LocalRestTaskManager : TaskManager
     {
         geod24.LocalRest.sleep(dur);
     }
+
+    /***************************************************************************
+
+        Run an asynchronous task after a given time in LocalRest's
+
+        The task will first run after the given `timeout`, and
+        can either repeat or run only once (the default).
+        Works similarly to Vibe.d's `setTimer`.
+
+        Params:
+            timeout = Determines the minimum amount of time that elapses before
+                the timer fires.
+            dg = This delegate will be called when the timer fires.
+            periodic = Specifies if the timer fires repeatedly or only once.
+
+    ***************************************************************************/
+
+    public override void setTimer (Duration timeout, void delegate() dg,
+        Periodic periodic = Periodic.No)
+    {
+        geod24.LocalRest.setTimer(timeout, dg, periodic);
+    }
 }
 
 /// A ban manager with a fake clock for reliable unittesting
--- a/source/agora/test/EnrollmentManager.d
+++ b/source/agora/test/EnrollmentManager.d
@@ -89,22 +89,9 @@ unittest
     txs2.each!(tx => node_1.putTransaction(tx));
     containSameBlocks(nodes, 2).retryFor(8.seconds);
 
-    // Currently a new transaction needs to be sent to trigger
-    // the reveal of a pre-image
-    // See https://github.com/bpfkorea/agora/issues/582
-    Transaction tx =
-    {
-        TxType.Payment,
-        [Input(hashFull(txs2[0]), 0)],
-        [Output(Amount(100), pubkey_1)]
-    };
-    auto signature = gen_key_pair.secret.sign(hashFull(tx)[]);
-    tx.inputs[0].signature = signature;
-    node_1.putTransaction(tx);
-
-    // check if nodes have a pre-image newly sent
-    // during creating transactions for the new block
+    // Check if nodes have a pre-image newly sent
+    // While the timer is running on the taskmanager
     nodes.each!(node =>
         retryFor(node.getPreimage(enroll.utxo_key) != PreImageInfo.init,
-            5.seconds));
+            10.seconds));
 }
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -217,16 +217,6 @@ public class FullNode : API
             this.network.gossipTransaction(tx);
             this.onAcceptedTransaction();
         }
-
-        if (this.enroll_man.needRevealPreimage(this.ledger.getBlockHeight()))
-        {
-            PreImageInfo preimage;
-            if (this.enroll_man.getNextPreimage(preimage))
-            {
-                this.receivePreimage(preimage);
-                this.enroll_man.increaseNextRevealHeight();
-            }
-        }
     }
 
     /// GET: /has_transaction_hash


TrustHenry:
I should have pressed Draft PR, but I couldn't. :(

Geod24:
You can change it :)

source/agora/node/Validator.d
```diff
@@ -33,6 +33,9 @@ import scpd.types.Stellar_SCP;
 
 import ocean.util.log.Logger;
 
+import std.datetime;
```
```suggestion
import core.time;
```

source/agora/node/Validator.d
```diff
@@ -83,6 +86,9 @@ public class Validator : FullNode, API
             this.network.discover(required_peer_keys);
             this.network.startPeriodicCatchup(this.ledger, &this.nominator.isNominating);
         });
+
+        // start a periodic timer that `checkRevealPreimage` the time 10 seconds
+        this.taskman.timer(10.seconds, toDelegate(&checkRevealPreimage));
```
AndrejMitrovic: Why do you need `toDelegate`? `&checkRevealPreimage` is already a delegate.

AndrejMitrovic:
With regards to the timer implementation, you will need to first submit a PR to LocalRest with a test-case.

source/agora/common/Task.d
```diff
@@ -51,4 +51,26 @@ public class TaskManager
...
+            timeout = Determines the minimum amount of time that elapses before
+                the timer fires.
+            callback = If non-null, this delegate will be called when the timer
+                fires
+            periodic = Speficies if the timer fires repeatedly or only once
+
+    ***************************************************************************/
+
+    public void timer (Duration timeout, scope void delegate() dg,
+        bool periodic = true)
```
AndrejMitrovic: How about we make this an `enum`. `enum Periodic { No, Yes }` so then at the call site we will have:

```D
taskman.timer(10.seconds, &dg, Periodic.No);
```

Because if you have this:


```D
taskman.timer(10.seconds, &dg, false);
```

Then it's not obvious what this means without looking at the documentation.

source/agora/test/Base.d
```diff
@@ -262,6 +262,26 @@ public class LocalRestTaskManager : TaskManager
...
+
+    /***************************************************************************
+
+        Run an asynchronous task after the set time in LocalRest's
+        can set it to repeat or run it only once.
+
+        Params:
+        timeout = Determines the minimum amount of time that elapses before
+            the timer fires.
+        callback = If non-null, this delegate will be called when the timer
```
AndrejMitrovic: You are not checking for `null` in the LocalRest change in https://github.com/TrustHenry/localrest/compare/89f3a04e1d87a2c3c45e81659598e0d72bcb7c54...541d4db05ccc68dc423d93d93128502739a36325#diff-b5229c55d46fb0a852c49674585a3595.

So the code will segfault with a null callback here.

TrustHenry: Yes, Good catch!

source/agora/node/Validator.d
TrustHenry: I remove it.

MichaelKim20:
I'd like to PR `https://github.com/Geod24/localrest/`. It would be convenient to review and fix.

source/agora/common/Task.d
```diff
@@ -51,4 +51,26 @@ public class TaskManager
...
+    {
+        import vibe.core.core;
+        auto tm = setTimer(timeout, dg, periodic);
+        scope (exit) tm.stop();
```
AndrejMitrovic: @TrustHenry could you explain the logic to me here? You are setting a timer in the statement before, and then you are immediately stoping it.

AndrejMitrovic: Also having a `scope (exit)` at the end of a function does not make any sense, it is the same thing as writing the statement without the `scope (exit)`.

TrustHenry: You're right.

source/agora/common/Task.d
```diff
@@ -51,4 +51,25 @@ public class TaskManager
...
+        Params:
+            timeout = Determines the minimum amount of time that elapses before
+                the timer fires.
+            dg = If non-null, this delegate will be called when the timer
+                fires
+            periodic = Speficies if the timer fires repeatedly or only once
+
+    ***************************************************************************/
+
+    public void timer (Duration timeout, scope void delegate() dg,
```
Geod24: It cannot be `scope`.

source/agora/node/Validator.d
```diff
@@ -171,4 +176,24 @@ public class Validator : FullNode, API
...
+    {
+        if (this.enroll_man.needRevealPreimage(this.ledger.getBlockHeight()))
+        {
+            PreImageInfo preimage;
+            if (this.enroll_man.getNextPreimage(preimage))
+            {
+                this.receivePreimage(preimage);
+                this.enroll_man.increaseNextRevealHeight();
+            }
+        }
```
Geod24: Prefer early return when possible.

TrustHenry: I just applied it.

source/agora/common/Task.d
TrustHenry: I apply the suggestion.

source/agora/common/Task.d
```diff
@@ -51,4 +58,46 @@ public class TaskManager
...
+
+    public void timer (Duration timeout, void delegate() dg,
+        Periodic periodic = Periodic.No)
+    {
+        static import vibe.core.core;
+        vibe.core.core.setTimer(timeout, dg, periodic);
+    }
+}
+
+/// Test timer
```
AndrejMitrovic: Unittests should not do I/O, and vibe.d does I/O.

This test might belong to `LocalRestTaskManager`.

source/agora/node/Validator.d
```diff
@@ -83,6 +85,9 @@ public class Validator : FullNode, API
             this.network.discover(required_peer_keys);
             this.network.startPeriodicCatchup(this.ledger, &this.nominator.isNominating);
         });
+
+        // start a periodic timer that `checkRevealPreimage` the time 10 seconds
```
```suggestion
        // call checkRevealPreimage periodically every 10 seconds
```

source/agora/common/Task.d
```diff
@@ -51,4 +58,28 @@ public class TaskManager
...
+        Run an asynchronous task after a given time.
+
+        The task will first run after the given `timeout`, and
+        can either repeat or run only once (the default).
+        Use the `setTimer` vibe.d's
+
+        Params:
+            timeout = Determines the minimum amount of time that elapses before
+                the timer fires.
+            dg = If non-null, this delegate will be called when the timer fires
```
AndrejMitrovic: It cannot be `null`. vibe.d doesn't check for null.

source/agora/node/Validator.d
```diff
@@ -83,6 +85,9 @@ public class Validator : FullNode, API
             this.network.discover(required_peer_keys);
             this.network.startPeriodicCatchup(this.ledger, &this.nominator.isNominating);
         });
+
+        // call checkRevealPreimage periodically every 10 seconds
+        this.taskman.timer(10.seconds, &checkRevealPreimage, Periodic.Yes);
```
```suggestion
        this.taskman.timer(10.seconds, &this.checkRevealPreimage, Periodic.Yes);
```

source/agora/test/Base.d
```diff
@@ -264,6 +264,28 @@ public class LocalRestTaskManager : TaskManager
...
+        Run an asynchronous task after a given time in LocalRest's
+
+        The task will first run after the given `timeout`, and
+        can either repeat or run only once (the default).
+        Works similarly to Vibe.d's `setTimer`.
+
+        Params:
+            timeout = Determines the minimum amount of time that elapses before
+                the timer fires.
+            dg = If non-null, this delegate will be called when the timer fires
```
```suggestion
            dg = This delegate will be called when the timer fires
```

Again I think we don't check for null before calling the delegate.

source/agora/node/Validator.d
TrustHenry: Oh! Good catch.

AndrejMitrovic:
Actually I am a little bit confused about vibe.d. It claims it supports null delegates, but it has two overloads of `setTimer`. One which takes a `nothrow` and which can be null, and another which takes a throwable delegate which it calls without checking for null.

It's probably an oversight? Hmm.

AndrejMitrovic:
I think maybe we should just reject null delegates? I don't know of any use-case for it.

TrustHenry:
https://github.com/vibe-d/vibe-core/blob/872bff8472cc6755e4d0424b347a19a4f5d86ec3/source/vibe/core/core.d#L873-L883
I've referred here.
But as @AndrejMitrovic  mentioned, we don't have to allow null delegate yet.

AndrejMitrovic:
> I've referred here.

Yes but your code doesn't call that overload. It calls this one:

https://github.com/vibe-d/vibe-core/blob/872bff8472cc6755e4d0424b347a19a4f5d86ec3/source/vibe/core/core.d#L906

TrustHenry:
That's right. It's overloaded.
We'd better not allow null delegate.

source/agora/node/Validator.d
TrustHenry: Thanks, the suggestion.

source/agora/common/Task.d
TrustHenry: https://github.com/Geod24/localrest/blob/2cf119e759cdbbe2879bf65435efd18084620498/source/geod24/LocalRest.d#L1878-L1919
The test is being performed in the LocalRest, so I delete it.

source/agora/node/Validator.d
```diff
@@ -83,6 +85,9 @@ public class Validator : FullNode, API
             this.network.discover(required_peer_keys);
             this.network.startPeriodicCatchup(this.ledger, &this.nominator.isNominating);
         });
+
+        // call checkRevealPreimage periodically every 10 seconds
+        this.taskman.timer(10.seconds, &this.checkRevealPreimage, Periodic.Yes);
```
AndrejMitrovic: I think this routine should not be here. It should be after `this.network.discover`. Otherwise we will try to send preimages before the network has been discovered, which isn't really useful at that point. There will be no nodes to send the preimages to.

AndrejMitrovic: Note that it should be completely fine to start a timer within a task.

TrustHenry: I agree with @AndrejMitrovic 
I don't think there's a suitable call point yet.
The timer should only be called once in the node life cycle.

Oh, does our node have a restart function after starting?

AndrejMitrovic: I mean, you can move it below

```
this.network.discover(required_peer_keys);
>here<
```

That should work fine?

TrustHenry: I misunderstood.  work fine :)

AndrejMitrovic:
There is an existing networking test which tests that preimages are revealed when a transaction is sent. That test should now be rewritten.

TrustHenry:
The existing issue comments were deleted and the test was simplified

source/agora/common/Task.d
```diff
@@ -51,4 +58,30 @@ public class TaskManager
...
         vibe.core.core.sleep(dur);
     }
+
+    /***************************************************************************
+
+        Run an asynchronous task after a given time.
+
+        The task will first run after the given `timeout`, and
+        can either repeat or run only once (the default).
+        Use the `setTimer` vibe.d's
```
Geod24: I would remove this last (incomplete) sentence and instead add, at the end (after `Params`):

```
        See_Also: https://vibed.org/api/vibe.core.core/setTimer
```

source/agora/common/Task.d
```diff
@@ -51,4 +58,30 @@ public class TaskManager
...
+
+        Params:
+            timeout = Determines the minimum amount of time that elapses before
+                the timer fires.
+            dg = This delegate will be called when the timer fires.
+            periodic = Speficies if the timer fires repeatedly or only once.
```
Geod24:
```suggestion
            periodic = Specifies if the timer fires repeatedly or only once.
```

source/agora/test/Base.d
```diff
@@ -264,6 +264,28 @@ public class LocalRestTaskManager : TaskManager
...
+
+        The task will first run after the given `timeout`, and
+        can either repeat or run only once (the default).
+        Works similarly to Vibe.d's `setTimer`.
+
+        Params:
+            timeout = Determines the minimum amount of time that elapses before
+                the timer fires.
+            dg = This delegate will be called when the timer fires.
+            periodic = Speficies if the timer fires repeatedly or only once.
```
```suggestion
            periodic = Specifies if the timer fires repeatedly or only once.
```

source/agora/common/Task.d
```diff
@@ -51,4 +58,30 @@ public class TaskManager
...
+
+    ***************************************************************************/
+
+    public void timer (Duration timeout, void delegate() dg,
```
Geod24: Since both functions are called `setTimer`, why change the name ?

source/agora/test/EnrollmentManager.d
```diff
@@ -89,21 +89,8 @@ unittest
...
-        TxType.Payment,
-        [Input(hashFull(txs2[0]), 0)],
-        [Output(Amount(100), pubkey_1)]
-    };
-    auto signature = gen_key_pair.secret.sign(hashFull(tx)[]);
-    tx.inputs[0].signature = signature;
-    node_1.putTransaction(tx);
-
-    // check if nodes have a pre-image newly sent
-    // during creating transactions for the new block
```
Geod24, line 106: Isn't it great to remove workaround code ? :D

However the `retryFor` is for 5 seconds only, isn't there a risk that the image isn't propagated ?

Geod24 (approved):
Broadly LGTM

source/agora/test/EnrollmentManager.d
```diff
@@ -89,21 +89,8 @@ unittest
...
+    // Check if nodes have a pre-image newly sent
+    // While the timer is running on the taskmanager
     nodes.each!(node =>
         retryFor(node.getPreimage(enroll.utxo_key) != PreImageInfo.init,
             5.seconds));
```
AndrejMitrovic: @TrustHenry I think you will need a bigger timeout. You configured the timer for 10 seconds, and this timeout waits for 5.

source/agora/common/Task.d
TrustHenry: https://github.com/vibe-d/vibe-core/blob/872bff8472cc6755e4d0424b347a19a4f5d86ec3/source/vibe/core/core.d#L876

Copy and paste problems :(

source/agora/common/Task.d
TrustHenry: I wanted to avoid confusion with the core of vibe.d.

source/agora/common/Task.d
TrustHenry: I apply suggestions

source/agora/test/EnrollmentManager.d
TrustHenry: Yes, that's right.

source/agora/common/Task.d
Geod24: But we already have this for `runTask` and `sleep`

TrustHenry:
@bpfkorea/core Is 10 seconds appropriate for my suggested interval of Reveal pre-images?

Geod24:
I think it's a good start

source/agora/common/Task.d
TrustHenry: Yes, I apply your opinion.

linked0:
I'll add a task item about making the period for revalation of pre-image configurable in the issue #520 .

Geod24 (approved):
LGTM. Merge at will.

Issue #822, opened 2020-05-13
We should not accept a block with no enrollments if the validator set will become empty


AndrejMitrovic:
As discussed with @Geod24. 

Consider this **example** case:

- Genesis block begins with 2 validators
- Their expiration block is at height 1008
- Block height 1008 is reached, and the validators are now expired.

Now the blockchain has permanently halted. Block 1009 will never be created because there are no active validators which can reach consensus and sign a new block. Re-enrolling is not possible at this point either, because enrollments need to be written to a block to be considered active.

We need to come up with some kind of rule. For example, if all of the validators are about to expire, then do not create a block unless there are new enrollments in the block.

Additionally we may want to make another protocol rule: Do not allow less than X validators to be enrolled at all times. For example: If there are only 2 enrolled validators left in the validator set and those two validators go permanently offline (or lose their key, etc), the blockchain permanently halts.

MichaelKim20:
How can I add enrollment data in genesis block when there is no frozen UTXO?

AndrejMitrovic:
By adding the frozen UTXO to the genesis block.

AndrejMitrovic:
For example, for my shell quorum PR I made this function:

```D
private immutable(Block) makeGenesisBlock (in KeyPair[] key_pairs)
{
    // 1 payment tx, the rest are freeze txs
    auto gen_addr = getGenesisKeyPair().address;
    Transaction FirstTx =
    {
        TxType.Payment,
        inputs: [ Input.init ],
        outputs: [
            Output(Amount(62_500_000L * 10_000_000L), gen_addr),
            Output(Amount(62_500_000L * 10_000_000L), gen_addr),
            Output(Amount(62_500_000L * 10_000_000L), gen_addr),
            Output(Amount(62_500_000L * 10_000_000L), gen_addr),
            Output(Amount(62_500_000L * 10_000_000L), gen_addr),
            Output(Amount(62_500_000L * 10_000_000L), gen_addr),
            Output(Amount(62_500_000L * 10_000_000L), gen_addr),
            Output(Amount(62_500_000L * 10_000_000L), gen_addr),
        ],
    };

    Transaction[] txs;
    txs ~= FirstTx;
    Enrollment[] enrolls;

    foreach (const ref key_pair; key_pairs)
    {
        Transaction tx =
        {
            type : TxType.Freeze,
            inputs : [ Input.init ],
            outputs : [Output(Amount.MinFreezeAmount, key_pair.address)]
        };

[...]

PR #862, opened 2020-05-21
Add the ability to obtain the number of valid validator at the next block height

MichaelKim20:
This is the first PR to solve issue 822.
Minimum valid validators must exist at all block heights. This ensures that the following blocks are continuously created:
However, this should not be checked after the block is made. It is necessary to check this during the consensus process.
So we need to get the number of valid validators at the next block height, when the blocks that are now being agreed upon are applied.

I implemented this.

Relates to Issue #822

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -811,6 +811,29 @@ public class EnrollmentManager
         catch (Exception ex)
         {}
     }
+
+    /***************************************************************************
+
+        Gets the number of active validators at the block height.
+
+        `block_height` is the height of the newly created block.
+        If the active validators are less than the specified value,
+        new blocks cannot be created.
+
+        Params:
+            block_height = the height of proposed block
+
+        Returns:
+            Returns the number of active validators when the block height is
+            `block_height`.
+            Returns 0 in case of error.
+
+    ***************************************************************************/
+
+    public ulong getValidatorCount (ulong block_height) @safe nothrow
+    {
+        return this.validator_set.getValidatorCount(block_height);
+    }
 }
 
 /// tests for member functions of EnrollmentManager
@@ -1092,4 +1115,100 @@ unittest
             }
         }
     }
+}
+
+/// tests for `EnrollmentManager.getValidatorCount
+unittest
+{
+    import agora.common.Amount;
+    import agora.consensus.data.Transaction;
+    import agora.consensus.Genesis;
+    import std.format;
+    import std.conv;
+
+    scope storage = new TestUTXOSet;
+
+    auto gen_key_pair = getGenesisKeyPair();
+    KeyPair key_pair = KeyPair.random();
+
+    foreach (idx; 0 .. 8)
+    {
+        auto input = Input(hashFull(GenesisTransaction), idx.to!uint);
+
+        Transaction tx =
+        {
+            TxType.Freeze,
+            [input],
+            [Output(Amount.MinFreezeAmount, key_pair.address)]
+        };
+
+        auto signature = gen_key_pair.secret.sign(hashFull(tx)[]);
+        tx.inputs[0].signature = signature;
+        storage.put(tx);
+    }
+
+    // create an EnrollmentManager object
+    auto man = new EnrollmentManager(":memory:", key_pair);
+    scope (exit) man.shutdown();
+    Hash[] utxo_hashes = storage.keys;
+
+    Enrollment enrollment;
+    ulong block_height = 2;
+
+    // create and add the first Enrollment object
+    auto utxo_hash1 = utxo_hashes[0];
+    assert(man.createEnrollment(utxo_hash1, block_height, enrollment));
+    assert(!man.hasEnrollment(utxo_hash1));
+    assert(man.addEnrollment(enrollment, &storage.findUTXO));
+    assert(man.hasEnrollment(utxo_hash1));
+    assert(man.getValidatorCount(block_height) + 1 == 1);
+
+    man.clearExpiredValidators(block_height);
+    man.addValidator(enrollment, block_height, &storage.findUTXO);
+
+    block_height = 3;
+
+    // create and add the second Enrollment object
+    auto utxo_hash2 = utxo_hashes[1];
+    assert(man.createEnrollment(utxo_hash2, block_height, enrollment));
+    assert(!man.hasEnrollment(utxo_hash2));
+    assert(man.addEnrollment(enrollment, &storage.findUTXO));
+    assert(man.hasEnrollment(utxo_hash2));
+    assert(man.getValidatorCount(block_height) + 1 == 2);
+
+    man.clearExpiredValidators(block_height);
+    man.addValidator(enrollment, block_height, &storage.findUTXO);
+
+    block_height = 4;
+
+    // create and add the third Enrollment object
+    auto utxo_hash3 = utxo_hashes[2];
+    assert(man.createEnrollment(utxo_hash3, block_height, enrollment));
+    assert(!man.hasEnrollment(utxo_hash3));
+    assert(man.addEnrollment(enrollment, &storage.findUTXO));
+    assert(man.hasEnrollment(utxo_hash3));
+    assert(man.getValidatorCount(block_height) + 1 == 3);
+
+    man.clearExpiredValidators(block_height);
+    man.addValidator(enrollment, block_height, &storage.findUTXO);
+
+    block_height = 5;    // valid block height : 0 <= H < 1008
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 3);
+
+    block_height = 1009; // valid block height : 2 <= H < 1010
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 3);
+
+    block_height = 1010; // valid block height : 3 <= H < 1011
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 2);
+
+    block_height = 1011; // valid block height : 4 <= H < 1012
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 1);
+
+    block_height = 1012; // valid block height : 5 <= H < 1013
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 0);
 }
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -321,6 +321,42 @@ public class ValidatorSet
 
     /***************************************************************************
 
+        Gets the number of active validators at the block height.
+
+        `block_height` is the height of the newly created block.
+        If the active validators are less than the specified value,
+        new blocks cannot be created.
+
+        Params:
+            block_height = the height of proposed block
+
+        Returns:
+            Returns the number of active validators when the block height is
+            `block_height`.
+            Returns 0 in case of error.
+
+    ***************************************************************************/
+
+    public ulong getValidatorCount (ulong block_height) @safe nothrow
+    {
+        ulong validator_count = 0;
+        try
+        {
+            ulong height = (block_height >= Enrollment.ValidatorCycle) ?
+                block_height - Enrollment.ValidatorCycle + 1 : 0;
+            () @trusted {
+                validator_count = this.db.execute("SELECT count(*) FROM validator_set WHERE enrolled_height >= ?", height).oneValue!ulong;
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error {}", ex);
+        }
+        return validator_count;
+    }
+
+    /***************************************************************************
+
         Check if a pre-image exists
 
         Params:


source/agora/consensus/ValidatorSet.d
```diff
@@ -319,6 +320,35 @@ public class ValidatorSet
...
+        Extracts the UTXO hash value of a valid validators at a next
+        block height.
+
+        Params:
+            block_height = the next block height
+            utxo_key_set = Store the validators's UTXO hash value.
+
+    ***************************************************************************/
+
+    public void getAvailableKeys (ulong block_height, ref Set!Hash utxo_key_set) @safe nothrow
```
AndrejMitrovic: The `utxo_key_set` seems unnecessary. This function could be renamed `ulong getValidatorCount (ulong block_height)`. Then all you have to do is call:

```D
this.db.execute("SELECT count(*) FROM validator_set WHERE enrolled_height >= ?", height).oneValue!ulong;
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -811,6 +812,28 @@ public class EnrollmentManager
...
+
+        Gets the number of a valid validators at a next block height.
+
+        Params:
+            block_height = the next block height
+            enrolls = the enrollment data included when creating blocks
+
+    ***************************************************************************/
+
+    public ulong getAvailableCount (ulong block_height, ref Enrollment[] enrolls) @safe nothrow
```
AndrejMitrovic: `getValidatorCount` would be a better name, and it should just forward to the implementation in `ValidatorSet`.

You don't need to pass the enrollments here. Think about it, if a block will contain enrollments it means the validator set will not be empty.

There is only a single protocol check that we need to do:

- If the enrollments in the proposed `ConsensusData` are empty, then check the ValidatorSet and make sure that there will still be valid enrollments for the proposed new block height.

AndrejMitrovic (changes requested):

source/agora/consensus/ValidatorSet.d
MichaelKim20: Shouldn't there be more than X validators?
If only one validator is needed, I agree with you. 
However, When we need several, shouldn't we remove the same one and check the number?

AndrejMitrovic: `SELECT count(*)` returns the count.

source/agora/consensus/EnrollmentManager.d
MichaelKim20: Shouldn't there be more than X validators?
If only one validator is needed, I agree with you.
However, When we need several, shouldn't we remove the same one and check the number?

AndrejMitrovic: What I mean is, is that the calling function can already do this. It doesn't need to pass the `Enrollment[]` here. It could either pass the length, or it could calculate it at the call site.

So the code in Nominator could do something like this:

```D
if (consensus_data.enrollments.length + enrollment_man.getValidatorCount(proposed_block_height) < MIN_VALIDATOR_COUNT)
    return false;
```

MichaelKim20: I understood your suggestion accurately. Thank you very much for your good opinion.

MichaelKim20: I realized late that `ConsensusData.enrollments` and `ValidatorSet` cannot have the same data. @AndrejMitrovic , I'm sorry. It's all fixed.

AndrejMitrovic:
I will review it again soon.

source/agora/consensus/ValidatorSet.d
```diff
@@ -319,6 +319,33 @@ public class ValidatorSet
...
+            block_height = the height of proposed block
+
+    ***************************************************************************/
+
+    public ulong getValidatorCount (ulong block_height) @safe nothrow
+    {
+        ulong validator_count = 0;
+        try
+        {
+            ulong height = (block_height >= Enrollment.ValidatorCycle) ?
```
linked0, line 345: The `height` could be better with the name of `start_height` because the `enrolled_height` column of the `validator_set` table is the height when the enrollment information is stored in a block and the `start_height` is actually the `enrolled_height` plus 1.

source/agora/consensus/ValidatorSet.d
```diff
@@ -319,6 +319,33 @@ public class ValidatorSet
...
+                block_height - Enrollment.ValidatorCycle + 1 : 0;
+            () @trusted {
+                validator_count = this.db.execute("SELECT count(*) FROM validator_set WHERE enrolled_height >= ?", height).oneValue!ulong;
```
linked0, line 348: I think the code below is right because the `height` value that you have calculated is actually the `enrolled_height` plus 1.

```
validator_count = this.db.execute("SELECT count(*) FROM validator_set WHERE enrolled_height >= ?", height-1).oneValue!ulong;
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1091,3 +1105,115 @@ unittest
...
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 3);
+
+    block_height = 1009; // valid block height : [2 .. 1010]
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 3);
+
+    block_height = 1010; // valid block height : [3 .. 1011]
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 2);
```
linked0, line 1205: I think this should be 3 if because the first enrollment has the `enrolled_height` of 2 so the start height as a validator is 3 as you described. I have been thinking the `block_height` you passed in as parameter is the height where the new block is being made.
(이 assert 문에서 체크해야할 값은 3이 되어야한다고 생각합니다. 왜냐하면, 첫번째 enrollment의 enrolled_height는 2이고, validator로서 기능을 시작하는 것은 height가 3이 됩니다. 따라서, 첫번째 enrollment는 3부터 1010까지 validator로서 유효하기 때문에 위 assert 체크값은 3이 되어야 합니다.)

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1091,3 +1105,115 @@ unittest
...
+    assert(man.createEnrollment(utxo_hash2, block_height, enroll2));
+    assert(!man.hasEnrollment(utxo_hash2));
+    assert(man.add(enroll2, &storage.findUTXO));
+    assert(man.enroll_pool.count() == 1);
+    assert(man.hasEnrollment(utxo_hash2));
+    enrollments ~= enroll2;
+    assert(man.getValidatorCount(block_height) + enrollments.length == 2);
+
+    man.clearExpiredValidators(block_height);
+    foreach (ref enrollment; enrollments)
```
linked0: The `enrollments` in all the `foreach` statement always has just one element, so why don't you use just the variable of `Enrollment` type and not to use `foreach` statement?

linked0 (commented):
Good job but there are some codes to be fixed.
This comment should have posted earlier though.

source/agora/consensus/ValidatorSet.d
MichaelKim20: I thought after watching `ClearExpiredValidators`.
Please explain in detail what the problem is.

MichaelKim20: ClearExpiredValidators : 
enrolled_height <= block_height - Enrollment.ValidatorCycle
getValidatorCount:
enrolled_height >= block_height - Enrollment.ValidatorCycle + 1

source/agora/consensus/ValidatorSet.d
MichaelKim20: I thought after watching ClearExpiredValidators.
Please explain in detail what the problem is.
ClearExpiredValidators :
enrolled_height <= block_height - Enrollment.ValidatorCycle
getValidatorCount:
enrolled_height >= block_height - Enrollment.ValidatorCycle + 1

source/agora/consensus/EnrollmentManager.d
MichaelKim20: I thought after watching ClearExpiredValidators.
Please explain in detail what the problem is.
ClearExpiredValidators :
enrolled_height <= block_height - Enrollment.ValidatorCycle
getValidatorCount:
enrolled_height >= block_height - Enrollment.ValidatorCycle + 1

source/agora/consensus/EnrollmentManager.d
MichaelKim20: I thought after watching ClearExpiredValidators.
Please explain in detail what the problem is.
ClearExpiredValidators :
enrolled_height <= block_height - Enrollment.ValidatorCycle
getValidatorCount:
enrolled_height >= block_height - Enrollment.ValidatorCycle + 1

source/agora/consensus/ValidatorSet.d
AndrejMitrovic: I think Michael's code is correct. Consider:

- Validator V is enrolled at block height: `#1`
- V validates and signs blocks `[#2 .. #1009]` inclusively.
- The Ledger is currently at block `#1008`.
- We are nominating block `#1009`.
- We need to ensure there are still active validators which can nominate and sign blocks `#1010+`.
- We call `getValidatorCount(1009)`.
- V should not be counted in this call. Because after block `#1009` is externalized V cannot validate or sign blocks #1010+. Its range was `[#2 .. #1009]`.

So: `getValidatorCount(1009)`:
- => 1009 - 1008 + 1 = 2
- => `SELECT count(*) FROM validator_set WHERE enrolled_height >= 2` will return 0, which is correct. This validator will not be active after block 1009 is externalized.

linked0: I thought the function finds the  count of validators for the the height that is passed in. I mean, I thought the function finds the count of validators for nominating block you described. I guess I misunderstand the meaning of the block height.

AndrejMitrovic: Right, it's a misnomer. It's not about who can validate a given block, but rather who will be active at that block height so they can validate height + 1 and so on.

Maybe we can find a better name for the parameter.. or document the function better.

source/agora/consensus/EnrollmentManager.d
MichaelKim20: I changed it.
Thank you for your good suggestion.

source/agora/consensus/ValidatorSet.d
```diff
@@ -319,6 +319,33 @@ public class ValidatorSet
...
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Database operation error {}", ex);
```
AndrejMitrovic, line 353: We will need to handle this better in the future. But it might be left for a future PR.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1093,3 +1107,103 @@ unittest
...
+
+    block_height = 1009; // valid block height : [2 .. 1010]
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 3);
+
+    block_height = 1010; // valid block height : [3 .. 1011]
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 2);
+
+    block_height = 1011; // valid block height : [4 .. 1012]
```
```suggestion
    block_height = 1011; // valid block height : [4 .. 1011]
```

Because at block 1012 there will be zero validators.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1093,3 +1107,103 @@ unittest
...
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 1);
+
+    block_height = 1012; // valid block height : [5 .. 1013]
```
```suggestion
    block_height = 1012; // valid block height : [4 .. 1011]
```

Because at block 1012 there will be zero validators.

AndrejMitrovic (changes requested):

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1093,3 +1107,103 @@ unittest
...
     }
 }
+
+/// tests for `EnrollmentManager.getValidatorCount
+unittest
+{
+    import agora.common.Amount;
+    import agora.consensus.data.Transaction;
+    import agora.consensus.Genesis;
+    import std.algorithm;
```
Geod24: Already imported at module scope

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1093,3 +1107,103 @@ unittest
...
+
+    Enrollment enrollment;
+    ulong block_height = 2;
+
+    // create and add the first Enrollment object
+    auto utxo_hash1 = utxo_hashes[0];
+    assert(man.createEnrollment(utxo_hash1, block_height, enrollment));
+    assert(!man.hasEnrollment(utxo_hash1));
+    assert(man.add(enrollment, &storage.findUTXO));
+    assert(man.enroll_pool.count() == 1);
```
Geod24: I don't think we should check this. The enrollment pool is an internal detail and we don't really care if it's 1, 10 or 100 here.

AndrejMitrovic (approved):

source/agora/consensus/ValidatorSet.d
```diff
@@ -319,6 +319,33 @@ public class ValidatorSet
...
+        }
+        return validator_count;
```
Geod24, line 355: 0 is a valid value to return, but how is this function going to be used ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1093,3 +1107,103 @@ unittest
...
+    assert(man.getValidatorCount(block_height) + 1 == 3);
+
+    man.clearExpiredValidators(block_height);
+    man.addValidator(enrollment, block_height, &storage.findUTXO);
+
+    block_height = 5;    // valid block height : [0 .. 1008]
+    man.clearExpiredValidators(block_height);
+    assert(man.getValidatorCount(block_height) == 3);
+
+    block_height = 1009; // valid block height : [2 .. 1009]
```
Geod24: I can see two options here:
```suggestion
    block_height = 1009; // valid block height : [2, 1009]
```
OR:
```suggestion
    block_height = 1009; // valid block height : [2 .. 1010]
```

I have a slight preference for the second one, but can see how it could be confusing.
But if we are using the D syntax for array subscription, we have to make it non-inclusive at the end.
In mathematical notation, we would use `[)`, not `[]`, for this.

Geod24 (commented):
Broad LGTM.
Remove the tests which touch the enrollment pool (the feature is about the *validator set*), and a few other comments, and we're good to go.

MichaelKim20:
@Geod24 Ready for review

MichaelKim20:
@Geod24 Ready for review

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1093,3 +1116,99 @@ unittest
...
+    assert(man.createEnrollment(utxo_hash3, block_height, enrollment));
+    assert(!man.hasEnrollment(utxo_hash3));
+    assert(man.addEnrollment(enrollment, &storage.findUTXO));
+    assert(man.hasEnrollment(utxo_hash3));
+    assert(man.getValidatorCount(block_height) + 1 == 3);
+
+    man.clearExpiredValidators(block_height);
+    man.addValidator(enrollment, block_height, &storage.findUTXO);
+
+    block_height = 5;    // valid block height : 0 <= H < 1008
```
Geod24, line 1195: Nice, that's even better than what I suggested!

Geod24 (approved):

PR #866, opened 2020-05-21
build.d: Use g++ instead of gcc

Geod24:
It has been known (and experienced) to cause issues when using gcc to link C++ code.


AndrejMitrovic (approved):

MichaelKim20:
I also confirmed that `ci/check_vtable_test.d` is fine.

MichaelKim20 (approved):

Issue #849, opened 2020-05-17
Discussion: Should a node be able to enroll multiple times with the same address ?


Geod24:
Currently, our pre-image scheme relies only on the private key, a nonce, and a fixed string.
This leads to a situation where a user would get the same pre-image if they enrolled with two different frozen txs controlled by the same private key. This might not be wanted, or this could be a feature. It's not an issue with the protocol, but rather our implementation. Thoughts ?

CC @bpfkorea/core

linked0:
How about using the HD Key generation concept?
![HD-wallets-hierarchical-tree-1024x600](https://user-images.githubusercontent.com/2503139/85526949-62836b80-b645-11ea-83af-c3fbd01b701b.jpg)

In the above picture, we can use the child keys of the first child as the seeds of random seeds. One child has about 2^31 number(2,147,483,648) of child keys so each height can have the seed or private key for initial pre-image which is described as the random seed in the white paper.

It also makes us not need to look through all the blocks in order to restore the cycle index for the next random seed.

linked0:
The idea about how to apply HDKD(Hierarchical Deterministic Key Derivation) to AGORA
- We could use the public keys of children of the first child of master keys as the key to an enrollment.
- We could use the private keys of children of the first child of master keys as a component for making a seed of pre-images. Currently, we use the private key of nodes for it.

**Docs about HD wallet**
* [What the HD wallet is (part of the book, "Mastering Blockchain")](https://github.com/bitcoinbook/bitcoinbook/blob/develop/ch05.asciidoc)
* [Why the hardened derivation is needed](https://bitcoin.stackexchange.com/questions/56916/derivation-of-parent-private-key-from-non-hardened-child)

**Common industry standards**
* Mnemonic code words, based on [BIP-39](https://github.com/bitcoin/bips/blob/master/bip-0039.mediawiki)
* HD wallets, based on [BIP-32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki)
* Multipurpose HD wallet structure, based on [BIP-43](https://github.com/bitcoin/bips/blob/master/bip-0043.mediawiki)
* Multicurrency and multiaccount wallets, based on [BIP-44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki)

**Best Practice**
As a best practice, the level-1 children of the master keys are always derived through the hardened derivation, to prevent compromise of the master keys.

linked0:
[...]

PR #873, opened 2020-05-26
`EnrollmentManager` add an enrollment before the cycle

### LINKED ISSUE #824: It should be possible to re-enroll a Validator while it is in the validator set

linked0:
The validator set should allow a node to re-enroll before the validator cycle ends. It's not for sure that an enrollment will be contained in the next new block, so the validator set must add the enrollment earlier before the cycle ends.

I also added the constant about the round where a node can re-enroll because unnecessary re-enrollments should be avoid. The value can be configurable in the issue #520 .

Relates #824

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -184,9 +184,44 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public Height getEnrolledHeight (const ref Hash enroll_hash) @trusted
+    public Height getEnrolledHeight (const ref Hash enroll_hash)
+        @trusted nothrow
     {
         return this.validator_set.getEnrolledHeight(enroll_hash);
+    }
+
+    /***************************************************************************
+
+        Get the unregistered enrollments that can be validator in the next
+        block based on the current block height.
+
+        Params:
+            enrolls = will contain the unregistered enrollments data if found
+            height = current block height
+
+        Returns:
+            The unregistered enrollments data
+
+    ***************************************************************************/
+
+    public Enrollment[] getEnrollments (ref Enrollment[] enrolls, Height height)
+        @trusted nothrow
+    {
+        enrolls.length = 0;
+        assumeSafeAppend(enrolls);
+
+        static Enrollment[] pool_enrolls;
+        this.enroll_pool.getEnrollments(pool_enrolls);
+        foreach (enroll; pool_enrolls)
+        {
+            const enroll_height = this.getEnrolledHeight(enroll.utxo_key);
+            if (enroll_height == ulong.max ||
+                height >= enroll_height + this.params.ValidatorCycle - 1)
+            {
+                enrolls ~= enroll;
+            }
+        }
+        return enrolls;
     }
 
     /***************************************************************************
@@ -750,7 +785,7 @@ unittest
     assert(man.pool.count() == 3);
 
     Enrollment[] enrolls;
-    man.pool.getEnrollments(enrolls);
+    man.getEnrollments(enrolls, Height(1));
     assert(enrolls.length == 3);
     assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
 
@@ -772,7 +807,7 @@ unittest
     assert(man.getEnrolledHeight(enroll.utxo_key) == 9);
     assert(man.addValidator(enroll, Height(9), &storage.findUTXO) !is null);
     assert(man.getEnrolledHeight(enroll2.utxo_key) == ulong.max);
-    man.pool.getEnrollments(enrolls);
+    man.getEnrollments(enrolls, Height(9));
     assert(enrolls.length == 1);
     // One Enrollment was moved to validator set
     assert(man.validator_set.count() == 1);
@@ -781,7 +816,7 @@ unittest
     man.pool.remove(utxo_hash);
     man.pool.remove(utxo_hash2);
     man.pool.remove(utxo_hash3);
-    assert(man.pool.getEnrollments(enrolls).length == 0);
+    assert(man.getEnrollments(enrolls, Height(9)).length == 0);
 
     Enrollment[] ordered_enrollments;
     ordered_enrollments ~= enroll;
@@ -791,7 +826,7 @@ unittest
     ordered_enrollments.sort!("a.utxo_key > b.utxo_key");
     foreach (ordered_enroll; ordered_enrollments)
         assert(man.pool.add(ordered_enroll, &storage.findUTXO));
-    man.pool.getEnrollments(enrolls);
+    man.getEnrollments(enrolls, Height(man.params.ValidatorCycle + 8));
     assert(enrolls.length == 3);
     assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
 
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -336,7 +336,8 @@ public class Ledger
         const next_height = Height(this.getBlockHeight() + 1);
         auto utxo_finder = this.utxo_set.getUTXOFinder();
 
-        this.enroll_man.pool.getEnrollments(data.enrolls);
+        this.enroll_man.getEnrollments(data.enrolls,
+            Height(this.getBlockHeight()));
         foreach (hash, tx; this.pool)
         {
             if (auto reason = tx.isInvalidReason(utxo_finder, next_height))
@@ -1046,7 +1047,8 @@ unittest
 
     // Check if there are any unregistered enrollments
     Enrollment[] unreg_enrollments;
-    assert(ledger.enroll_man.pool.getEnrollments(unreg_enrollments) is null);
+    assert(ledger.enroll_man.getEnrollments(unreg_enrollments,
+        ledger.getBlockHeight()) is null);
     auto block_4 = ledger.getBlocksFrom(Height(4));
     enrollments.sort!("a.utxo_key < b.utxo_key");
     assert(block_4[0].header.enrollments == enrollments);
--- a/source/agora/test/EnrollmentManager.d
+++ b/source/agora/test/EnrollmentManager.d
@@ -96,3 +96,46 @@ unittest
                  node.getPreimage(enroll_3.utxo_key) != PreImageInfo.init,
             5.seconds));
 }
+
+// Test for re-enroll before the validator cycle ends
+unittest
+{
+    // Boilerplate
+    const validator_cycle = 20;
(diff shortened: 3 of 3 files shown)


source/agora/consensus/ValidatorSet.d
```diff
@@ -40,6 +40,10 @@ public class ValidatorSet
     /// SQLite db instance
     private Database db;
 
+    /// The available round for re-enrollment that is configurable
+    static private const uint ReenrollAvailableRound =
```
TrustHenry: `static private const uint ReEnrollAvailableRound =`

How about this?

AndrejMitrovic:
But will this mean that there can be two enrollments co-existing with the same public key but a different UTXO? Is there a test-case covering this now?

AndrejMitrovic:
> But will this mean that there can be two enrollments co-existing with the same public key but a different UTXO? Is there a test-case covering this now?

Sorry the question isn't so much related to this specific PR, it just came to mind.

source/agora/consensus/ValidatorSet.d
AndrejMitrovic:
```suggestion
    static private const uint MinimumValidateBlocks =
```

Maybe something like this. The Validator must validate at least this amount of blocks before the enrollment can be replaced.

I am not sure how well this fits in with the whitepaper though.

@Geod24? Would this break any assumptions about how enrollments work?

AndrejMitrovic: The name I suggested isn't perfect. Feel free to come up with something better if you can, or just leave it as-is.

source/agora/consensus/ValidatorSet.d
```diff
@@ -98,27 +102,43 @@ public class ValidatorSet
...
-        }
-
         try
         {
+            const bool exist = this.hasEnrollment(enroll.utxo_key);
+            if (exist)
+            {
+                // this checks that the re-enrollment is requested past the half
+                // of the validator cycle.
+                if (block_height - this.getEnrolledHeight(enroll.utxo_key) <
```
AndrejMitrovic: Shouldn't this check be in the validation part, not here?

source/agora/consensus/ValidatorSet.d
linked0: > I am not sure how well this fits in with the whitepaper though.
> 
> @Geod24? Would this break any assumptions about how enrollments work?
> 

Your question is reasonable. I just wanted some constraint against re-enrolling so often. The idea about that is open.

linked0:
> But will this mean that there can be two enrollments co-existing with the same public key but a different UTXO? Is there a test-case covering this now?

Your question is kind of related to #849 . Anyway I will think about the above question.

AndrejMitrovic:
> Your question is kind of related to #849 . Anyway I will think about above question.

Aha so it's already filed. Thanks for linking to it.

Geod24:
This won't work. See #900 and its workaround in #899.
I think we need to address #900, at least for the Enrollment part, for this to be viable.
Also, having hard-set limit seems rather... arbitrary. What's the reasoning behind it ?

linked0:
> Also, having hard-set limit seems rather... arbitrary. What's the reasoning behind it ?

The reasoning behind it is re-enrollment should be avoided during half of the validation cycle. Whatever the validation cycle is set, the limit will be half of the validation cycle.

source/agora/consensus/ValidatorSet.d
linked0: Sorry for the late answer.
I think the validation part is for validating the enrollment itself. The checking based on the relation with other objects/values like `block_height` or  `enroll_height` can be here because the enrollment has no height information and the validation part is also used in adding the enrollment in the enrollment pool.

source/agora/consensus/ValidatorSet.d
```diff
@@ -95,23 +99,43 @@ public class ValidatorSet
...
-            () @trusted {
-                this.db.execute("INSERT INTO validator_set " ~
-                    "(key, val, enrolled_height, preimage) VALUES (?, ?, ?, ?)",
-                    enroll.utxo_key.toString(), buffer, block_height, pbuffer);
-            }();
+            if (exist)
+            {
+                () @trusted {
+                    this.db.execute("UPDATE validator_set SET " ~
+                        "val = ?, enrolled_height = ? , preimage = ? WHERE key = ?",
```
AndrejMitrovic:
```suggestion
                        "val = ?, enrolled_height = ?, preimage = ? WHERE key = ?",
```

linked0:
Ready for review.

AndrejMitrovic:
The lookup code can be optimized. Right now it does three lookups (exists, height check, and then insert/update). Here is one way to do it:

```D
public string add (ulong block_height, scope UTXOFinder finder,
    const ref Enrollment enroll) @trusted nothrow
{
    // check validaty of the enrollment data
    if (auto reason = isInvalidReason(enroll, finder))
        return reason;

    try
    {
        const min_height = long(block_height) - ReEnrollAvailableRound;

        if (this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
            "validator_set WHERE key = ? AND enrolled_height > ?)",
            enroll.utxo_key.toString(), min_height).front.peek!bool(0))
        {
            log.info("Rejected re-enrollment is too early, " ~
                     "Data was: {}", enroll);
            return "The re-enrollment is too early";
        }

        const ZeroDistance = 0;  // initial distance
        this.db.execute("INSERT OR REPLACE INTO validator_set (" ~
            "key, cycle_length, enrolled_height, distance, preimage) " ~
            "VALUES(?, ?, ?, ?, ?) ",
            enroll.utxo_key.toString(),
            enroll.cycle_length,
            block_height, ZeroDistance,
            enroll.random_seed.toString());
    }
    catch (Exception ex)
    {
        // This should never happen, hence why we log it as well
        log.error("Unexpected error while adding a validator: {} ({})",
                  ex, enroll);
        return "Internal error in `ValidatorSet.add`";
    }

    return null;
}
```

Also:

```diff
diff --git a/source/agora/consensus/ValidatorSet.d b/source/agora/consensus/ValidatorSet.d
index 2ed8b2c9e..7b868b598 100644
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -42,7 +42,7 @@ public class ValidatorSet
     private ManagedDatabase db;

     /// The available round for re-enrollment that is configurable
-    static private const uint ReEnrollAvailableRound =
+    static private const long ReEnrollAvailableRound =
         Enrollment.ValidatorCycle / 2;
```

source/agora/consensus/ValidatorSet.d
```diff
@@ -91,21 +95,40 @@ public class ValidatorSet
...
+                        "Data was: {}", enroll);
+                    return "The re-enrollment is too early";
+                }
+
+                () @trusted {
+                    this.db.execute("UPDATE validator_set SET " ~
+                        "enrolled_height = ?, distance = ?, preimage = ? WHERE key = ?",
+                        block_height.value, ZeroDistance, enroll.random_seed.toString(),
+                        enroll.utxo_key.toString());
+                }();
```
Geod24: This is overriding an Enrollment before the Enrollment is over. That's something we should absolutely never do, and that is not what the issue is about. The issue is about the ability to re-enroll on the last block where a node is a validator, to avoid a 2-blocks downtime in validation. It does not **override** being a validator, it extends it.

Geod24 (changes requested):
- This does not solve the issue;
- It needs a network test;
- You will need to store when an Enrollment is available in the EnrollmentPool;
- You will also need to modify [this function](https://github.com/bpfkorea/agora/blob/913bee17afb8e2e467f0cc596aac3e7ccc37e2b0/source/agora/consensus/EnrollmentPool.d#L232-L266).

linked0:
Ready for review.
I have totally re-written the code as discussed with @Geod24 and added the network test.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -184,11 +184,46 @@ public class EnrollmentManager
...
+
+    ***************************************************************************/
+
+    public Enrollment[] getEnrollments (ref Enrollment[] enrolls, Height height)
+        @trusted nothrow
+    {
+        enrolls.length = 0;
+        assumeSafeAppend(enrolls);
+
+        Enrollment[] pool_enrolls;
```
```suggestion
        static Enrollment[] pool_enrolls;
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -184,11 +184,46 @@ public class EnrollmentManager
...
+        {
+            const enroll_height = this.getEnrolledHeight(enroll.utxo_key);
+            if (enroll_height == ulong.max ||
+                height == enroll_height + this.params.ValidatorCycle - 1)
+            {
+                enrolls ~= enroll;
+            }
+        }
+        return enrolls;
+    }
```
Geod24: Why isn't this in the pool ?

linked0: > Why isn't this in the pool ?

You have suggested the code should be in the pool. But, the code needs to check the enrollment is already a validator. The `EnrollmentManager` manages the pool and the validator set, it's easy to execute the process in the manager.

linked0:
Ready for review.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -184,11 +184,46 @@ public class EnrollmentManager
...
+        enrolls.length = 0;
+        assumeSafeAppend(enrolls);
+
+        static Enrollment[] pool_enrolls;
+        this.enroll_pool.getEnrollments(pool_enrolls);
+        foreach (enroll; pool_enrolls)
+        {
+            const enroll_height = this.getEnrolledHeight(enroll.utxo_key);
+            if (enroll_height == ulong.max ||
+                height == enroll_height + this.params.ValidatorCycle - 1)
```
```suggestion
            if (enroll_height == ulong.max ||
                height >= enroll_height + this.params.ValidatorCycle - 1)
```

If you check for an exact match, and for whatever reason the enrollment can not be added on the exact height, it will not be proposed again.

Geod24 (approved):
Just the one change, the rest LGTM

PR #874, opened 2020-05-26
Rename the member functions of the `EnrollmentManager`

linked0:
The names of `add` and `remove` functions were really confusing because the name didn't tell whether it's for the `EnrollmentPool` or the `ValidatorSet`. So I make the name clearer. As a result, we have three functions for the `EnrollmentPool`; `hasEnrollment`, `addEnrollment`, and `removeEnrollment`.


linked0:
Ready for review.
It's a simple change for naming functions more properly.

AndrejMitrovic (approved):

PR #875, opened 2020-05-26
Apply the ability to check the number of active validators

MichaelKim20:
Consider this case:

- Genesis block begins with 2 validators
- Their expiration block is at height 1008
- Block height 1008 is reached, and the validators are now expired.

When the height 1008 block is being created, I developed so that the block is not created without any active validators.

The creation of a block must stop immediately just before all the enrollment data expire.
This is to allow additional enrollment of validators.
If enrollment's cycle is 1008,
If none of the active validators exist at height 1008, block generation must stop at height 1007.
If the number of additionally enrolled validators at height 1007 is satisfied, a block of 1008 can be created.

The optimal number of active validators is 3 or 4.
I will increase the value of the minimum active validator from 1 to 3.
If I make a lot of changes at once, suddenly there are more test codes to modify, and the solutions are complicated.
So I will change it little by little and reach the goal.

Related to #822

### DIFF
--- a/source/agora/consensus/data/Enrollment.d
+++ b/source/agora/consensus/data/Enrollment.d
@@ -48,6 +48,9 @@ public struct Enrollment
     /// S: A signature for the message H(K, X, n, R) and the key K, using R
     public Signature enroll_sig;
 
+    /// The minimum number of validators required to create a block
+    public static immutable uint MinValidatorCount = 1;
+
     /***************************************************************************
 
         Implements hashing support
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -18,6 +18,7 @@ import agora.common.Hash;
 import agora.common.Set;
 import agora.common.Types;
 import agora.consensus.data.Block;
+import agora.consensus.data.Enrollment;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
 import VEn = agora.consensus.validation.Enrollment;
@@ -68,6 +69,8 @@ version (unittest)
         prev_height = the height of the direct ancestor of this block
         prev_hash = the hash of the direct ancestor of this block
         findUTXO = delegate to find the referenced unspent UTXOs with
+        active_enrollments = the number of enrollments that do not expire
+            at the next block height (prev_height + 1).
 
     Returns:
         `null` if the block is valid, a string explaining the reason it
@@ -76,7 +79,8 @@ version (unittest)
 *******************************************************************************/
 
 public string isInvalidReason (const ref Block block, Height prev_height,
-    in Hash prev_hash, UTXOFinder findUTXO) nothrow @safe
+    in Hash prev_hash, UTXOFinder findUTXO,
+    size_t active_enrollments) nothrow @safe
 {
     import std.algorithm;
 
@@ -106,6 +110,9 @@ public string isInvalidReason (const ref Block block, Height prev_height,
 
     if (!isStrictlyMonotonic!"a.utxo_key < b.utxo_key"(block.header.enrollments))
         return "Block: The enrollments are not sorted in ascending order";
+
+    if (block.header.enrollments.length + active_enrollments < Enrollment.MinValidatorCount)
+            return "Enrollment: Insufficient number of active validators";
 
     foreach (const ref enrollment; block.header.enrollments)
     {
@@ -239,7 +246,7 @@ unittest
     assert(block.isGenesisBlockValid());
 
     // don't accept block height 0 from the network
-    assert(!block.isValid(Height(0), Hash.init, null));
+    assert(!block.isValid(Height(0), Hash.init, null, Enrollment.MinValidatorCount));
 
     // height check
     block.header.height = 1;
@@ -389,9 +396,11 @@ public bool isGenesisBlockValid (const ref Block genesis_block)
 /// Ditto but returns `bool`, only usable in unittests
 version (unittest)
 public bool isValid (const ref Block block, Height prev_height,
-    Hash prev_hash, UTXOFinder findUTXO) nothrow @safe
+    Hash prev_hash, UTXOFinder findUTXO,
+    size_t active_enrollments) nothrow @safe
 {
-    return isInvalidReason(block, prev_height, prev_hash, findUTXO) is null;
+    return isInvalidReason(block, prev_height, prev_hash, findUTXO,
+        active_enrollments) is null;
 }
 
 ///
@@ -411,48 +420,60 @@ unittest
     auto block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, null, 1));
 
     // height check
-    assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+    assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+        Enrollment.MinValidatorCount));
 
     block.header.height = 100;
-    assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+    assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+        Enrollment.MinValidatorCount));
 
     block.header.height = GenesisBlock.header.height + 1;
-    assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+    assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+        Enrollment.MinValidatorCount));
 
     /// .prev_block check
     block.header.prev_block = block.header.hashFull();
-    assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+    assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+        Enrollment.MinValidatorCount));
 
     block.header.prev_block = gen_hash;
-    assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+    assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+        Enrollment.MinValidatorCount));
 
     /// Check consistency of `txs` field
     {
         auto saved_txs = block.txs;
 
         block.txs = saved_txs[0 .. $ - 1];
-        assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+        assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+            Enrollment.MinValidatorCount));
 
         block.txs = (saved_txs ~ saved_txs).sort.array;
-        assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+        assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+            Enrollment.MinValidatorCount));
 
         block.txs = saved_txs;
-        assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+        assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+            Enrollment.MinValidatorCount));
 
         /// Txs sorting check
         block.txs.reverse;
-        assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+        assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+            Enrollment.MinValidatorCount));
 
         block.txs.reverse;
-        assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+        assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+            Enrollment.MinValidatorCount));
     }
 
     /// no matching utxo => fail
     utxos.clear();
-    assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+    assert(!block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+        Enrollment.MinValidatorCount));
 
     GenesisBlock.txs.each!(tx => utxos.put(tx));
-    assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+    assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+        Enrollment.MinValidatorCount));
 
     utxos.clear();  // genesis is spent
     auto prev_txs = block.txs;
@@ -461,7 +482,7 @@ unittest
     auto prev_block = block;
     block = block.makeNewBlock(makeChainedTransactions(gen_key, prev_txs, 1));
     assert(block.isValid(prev_block.header.height, prev_block.header.hashFull(),
-        findUTXO));
+        findUTXO, Enrollment.MinValidatorCount));
 
     assert(prev_txs.length > 0);  // sanity check
     foreach (tx; prev_txs)
@@ -469,11 +490,11 @@ unittest
         // one utxo missing from the set => fail
         utxos.storage.remove(UTXOSetValue.getHash(tx.hashFull(), 0));
         assert(!block.isValid(prev_block.header.height, prev_block.header.hashFull(),
-            findUTXO));
+            findUTXO, Enrollment.MinValidatorCount));
 
         utxos.put(tx);
         assert(block.isValid(prev_block.header.height, prev_block.header.hashFull(),
-            findUTXO));
+            findUTXO, Enrollment.MinValidatorCount));
     }
 
     // the key is hashMulti(hash(prev_tx), index)
@@ -510,7 +531,7 @@ unittest
     // consumed all utxo => fail
     block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, null, 1));
     assert(block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
-            findNonSpent));
+            findNonSpent, Enrollment.MinValidatorCount));
 
     // All `payment` utxos have been consumed
     assert(GenesisBlock.txs[0].type == TxType.Freeze);
@@ -524,37 +545,37 @@ unittest
     double_spend[$ - 1] = double_spend[$ - 2];
     block = makeNewBlock(GenesisBlock, double_spend);
     assert(!block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
-            findNonSpent));
+            findNonSpent, Enrollment.MinValidatorCount));
 
     // we stopped validation due to a double-spend
     assert(used_set.length == double_spend.length - 1);
 
     block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, prev_txs, 1));
     assert(block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
-        findUTXO));
+        findUTXO, Enrollment.MinValidatorCount));
 
     // modify the last hex byte of the merkle root
     block.header.merkle_root[][$ - 1]++;
 
     assert(!block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
-        findUTXO));
+        findUTXO, Enrollment.MinValidatorCount));
 
     // now restore it back to what it was
     block.header.merkle_root[][$ - 1]--;
     assert(block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
-        findUTXO));
+        findUTXO, Enrollment.MinValidatorCount));
     const last_root = block.header.merkle_root;
 
     // txs with a different amount
     block = GenesisBlock.makeNewBlock(
         makeChainedTransactions(gen_key, prev_txs, 1, 20_000_000));
     assert(block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
-        findUTXO));
+        findUTXO, Enrollment.MinValidatorCount));
 
     // the previous merkle root should not match the new txs
     block.header.merkle_root = last_root;
     assert(!block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
-        findUTXO));
+        findUTXO, Enrollment.MinValidatorCount));
 }
 
 ///
@@ -580,7 +601,8 @@ unittest
         400_000_000_000 * Block.TxsInBlock).sort.array;
 
     auto block1 = makeNewBlock(GenesisBlock, txs_1);
-    assert(block1.isValid(GenesisBlock.header.height, gen_hash, findUTXO));
+    assert(block1.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
+        Enrollment.MinValidatorCount));
 
     foreach (ref tx; txs_1)
         utxo_set.put(tx);
@@ -622,7 +644,8 @@ unittest
     }
 
     auto block2 = makeNewBlock(block1, txs_2);
-    assert(block2.isValid(block1.header.height, hashFull(block1.header), findUTXO));
+    assert(block2.isValid(block1.header.height, hashFull(block1.header), findUTXO,
+        Enrollment.MinValidatorCount));
     foreach (ref tx; txs_2)
         utxo_set.put(tx);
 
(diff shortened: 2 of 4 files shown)


MichaelKim20:
~~I added new CLI `agora-checkval`~~
~~I don't know why `agora-checkval` doesn't work.~~

source/agora/common/Config.d
```diff
@@ -130,6 +130,9 @@ public struct NodeConfig
 
     /// Path to the data directory to store metadata and blockchain data
     public string data_dir = "/var/lib/agora/";
+
+    /// The minimum number of validators
+    public size_t min_validators = 0;
```
AndrejMitrovic: This would mean that nodes which have a different minimum set in the config would fork.

The minimum number of validators needs to be a protocol constant.

ci/check_validator_test.d
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env rdmd
+module check_vtable_test;
```
AndrejMitrovic:
```suggestion
module check_validator_test;
```

source/agora/common/Config.d
MichaelKim20: Okay, I'll try to use `agora.test.GenesisBlock.` I found this a little late.

ci/check_validator_test.d
MichaelKim20: That's a good catch.

AndrejMitrovic:
The reason your `main` function in `agora.cli.checkvalidator.main` is not called is because you're importing into `Base.d`, which defines a custom test runner function. See here:

https://github.com/bpfkorea/agora/blob/7e935bf4bdd3847ee94417f3c98775799cfca183/source/agora/test/Base.d#L184

This is why I asked you, why do you need to import into Base.d?

MichaelKim20:
> The reason your `main` function in `agora.cli.checkvalidator.main` is not called is because you're importing into `Base.d`, which defines a custom test runner function. See here:
> 
> https://github.com/bpfkorea/agora/blob/7e935bf4bdd3847ee94417f3c98775799cfca183/source/agora/test/Base.d#L184
> 
> This is why I asked you, why do you need to import into Base.d?

It needs to make 1008 blocks. Using LocalRest can save time.
I want to use this because Base.d has a lot of resources.

AndrejMitrovic:
If it's supposed to use LocalRest, why was it implemented as a CLI app? It can just be another unittest module in `source/agora/test`

MichaelKim20:
> If it's supposed to use LocalRest, why was it implemented as a CLI app? It can just be another unittest module in `source/agora/test`

The test to make 1008 blocks takes about 5 minutes. It's taking too long.

MichaelKim20:
The test code below takes too long to complete.
So I blocked it.
agora.test.ValidatorCount

AndrejMitrovic:
So how does moving it to a CLI solve the problem? It's still slow, no?

You can just store the blocks in the `TestAPIManager`. That's what was implemented in https://github.com/bpfkorea/agora/pull/810/files

You just have to create your blocks and then store it in the call to `auto net = new APIManager(your_blocks);`

Geod24:
> I added new CLI `agora-checkval`

Could you explain why you couldn't add a test in `source/agora/test/` ?

MichaelKim20:
> > I added new CLI `agora-checkval`
> 
> Could you explain why you couldn't add a test in `source/agora/test/` ?

The test I made uses LocalRest. It takes about 4 minutes and 10 seconds to make 1008 blocks. 
Isn't it too long to add as a test?

So I blocked it.
agora.test.ValidatorCount
https://github.com/bpfkorea/agora/pull/875/files#diff-c9ef8b448edfa7bba9edb775916c4a2c


Result is 
Start ...
Stop ... 4 minutes, 10 secs, 862 ms, 719 μs, and 9 hnsecs

Geod24:
We discussed it on Slack. The proper way to handle this is to reduce the amount of data needed to be generated in the tests (read: the length of cycles) instead of generating it, as this problem will show up countless times.

source/agora/test/Base.d
```diff
@@ -1028,3 +1034,79 @@ public bool containSameBlocks (APIS)(APIS nodes, size_t height)
...
+    // 1 payment tx, the rest are freeze txs
+    Transaction[] txs;
+    txs ~= UnitTestGenesisTransaction.serializeFull.deserializeFull!Transaction;
+    Enrollment[] enrolls;
+
+    foreach (idx; 0 .. max(Enrollment.MinValidatorCount, key_pairs.length))
+    {
+        auto key_pair = (idx < key_pairs.length)
+                        ? key_pairs[idx]
+                        : KeyPair.random;
```
AndrejMitrovic: Like I mentioned in the other PR, this code is a workaround for the tests which would break because of the new protocol change.

We could do the following:

- Implement validation and set the MinValidatorCount to 1. This means existing tests will still be correct.
- Fix the tests to make them more generic and so they always spawn at least `MinValidatorCount` Validators. That means `key_pairs` here will always be at least `>= MinValidatorCount `. We will probably rename `NetworkTopology.OneValidator` to `NetworkTopology.MinimumValidators`, so for example if we want 10 nodes, 2 would automatically become Validators and the rest FullNodes.
- Increase `MinValidatorCount` to a number that we define later. Maybe we will decide that 3 is a good number, or 4.

The work that I listed here can be done independently, one after the other.

MichaelKim20: I saw your review on #877 and moved the code here.
I understand all your suggestions. Thank you for your good suggestion.
I will proceed with the three tasks in order.

MichaelKim20:
The creation of a block must stop immediately just before all the active validators is expired.
This is to allow additional enrollment of validators.
Enrollment's cycle is defined as `Enrollment.ValidatorCycle`,
If none of the active validators exist at height `Enrollment.ValidatorCycle`,
block generation must stop at height `Enrollment.ValidatorCycle`-1.

The optimal minimum number of active validators is 3 or 4.
I proceeded by increasing the value of the minimum number of active validators by 1 to 3.
I corrected all the problematic unit tests.
Finally I set the minimum number of active validators to 3.
Even if the minimum number of active validators was set to 4, all tests were run normally.

From now on, I will check 'System Integration Test'.`

MichaelKim20:
'System Integration Test' has been resolved using the config files included in @AndrejMitrovic 's PR #845
https://github.com/bpfkorea/agora/pull/845/commits/e5afaa37aafe7f6fe3527829bd9ae82bf8fe8e74

MichaelKim20:
> We discussed it on Slack. The proper way to handle this is to reduce the amount of data needed to be generated in the tests (read: the length of cycles) instead of generating it, as this problem will show up countless times.

**1. Prepare**

I modified some agora.test.Base. It was written based on the source of @AndrejMitrovic .
In `NetworkTopology`, if a node is a validator, I have written code that generates freeze UTXO with the node's KeyPair and adds the enrollment to the Genesis block.
It takes too much time to make 1008 blocks in the unit test. Therefore, I set `Enrollment.ValidatorCycle` to 20 in the test. I have corrected all the test codes that cause problems.

**2. Sending transaction**

8 transactions were delivered to node 1.
Then wait for 8 seconds for one block to be made.
I coded this to repeat 19 times.
So far the final block height is 19.

**3. Check**

The next eight transactions are delivered to node 1.
But the height of the block does not increase.
When it reaches 20, all the enrollments expire and there is no validator enabled, so the node can no longer create blocks.

**4. Conclusion**

I got the expected result.

Geod24:
A reminder: Commits should be atomic. Here we have multiple "progression" commits that don't make sense, e.g. changing from 1 to 2 then 2 to 3. Also commits are out of order. I'll review the rest though.

Geod24:
Review in progress. I asked @AndrejMitrovic to submit #892 to simplify this PR. Thank you.

MichaelKim20:
> Review in progress. I asked @AndrejMitrovic to submit #892 to simplify this PR. Thank you.

After it was merged, I rebased my PR #875, and arranged the order.
Thank you.

MichaelKim20:
@AndrejMitrovic  made the following suggestion in my previous PR #877  
I accepted the suggestion and implemented it.
https://github.com/bpfkorea/agora/pull/877#issuecomment-634582980

source/agora/test/BanManager.d
```diff
@@ -43,13 +45,16 @@ unittest
...
-    // three nodes, two validators, and 1 non-validator
-    auto node_1 = network.nodes[0].client;
-    auto node_2 = network.nodes[1].client;
-    auto node_3 = network.nodes[2].client;  // non-validator
-    auto nodes = [node_1, node_2, node_3];
+    auto nodes = network.clients;
+    assert(nodes.length >= 3);
+    // The last node is non-validator, and the others are validators.
+    auto node_1 = nodes[0];     // validators
+    auto node_2 = nodes[1];     // validators
```
AndrejMitrovic:
```suggestion
```

You don't need these anymore because you're using `validators` below.

source/agora/test/BanManager.d
```diff
@@ -43,13 +45,16 @@ unittest
...
+    auto node_3 = nodes[$-1];   // non-validator
```
AndrejMitrovic:
```suggestion
    auto full_node = nodes[$-1];
```

That way it's explicit in the test, and also it doesn't matter which number it is anymore (node 3, 4, and so on).

source/agora/test/Ledger.d
```diff
@@ -89,7 +90,10 @@ unittest
 /// test catch-up phase after initial booting (periodic catch-up)
 unittest
 {
-    TestConf conf = { topology : NetworkTopology.OneValidator };
+    TestConf conf = {
+        topology : NetworkTopology.MinimumValidators,
+        nodes : NetworkTopology.MinimumValidators + 1
```
AndrejMitrovic: I wonder if it would make sense to split this into two:

```
extra_full_nodes
extra_validators
```

And then remove the `nodes` field. This way we can remove the need for `OneNonValidator` and also `MinimumValidators`, because `Simple` can be changed to do the same thing as `MinimumValidators` plus instantiate any extra nodes set in `extra_full_nodes` and `extra_validators`. 

This way we would always have at least `Enrollment.MinValidatorCount` validators in the genesis block.  Does it make sense to you?

AndrejMitrovic (changes requested):

source/agora/test/Ledger.d
MichaelKim20: Oh, that's my mistake. I'll fix it.

MichaelKim20:
Ready for review

MichaelKim20:
A condition has been added that requires at least three active validators.
Therefore, I changed it as below.

NetworkTopology.OneValidator => NetworkTopology.MinimumValidators
NetworkTopology.TwoOutsiderValidators => NetworkTopology.ThreeOutsiderValidators

source/agora/consensus/protocol/Nominator.d
```diff
@@ -353,15 +353,28 @@ extern(D):
                     fail_reason, data);
                 return ValidationLevel.kInvalidValue;
             }
+
+            if (data.enrolls.length >= Enrollment.MinValidatorCount)
```
AndrejMitrovic: There is already `validateConsensusData()`, these checks belong there.

source/agora/test/Base.d
```diff
@@ -770,13 +770,13 @@ public enum NetworkTopology
     OneNonValidator,
 
     /// Only one of the nodes is a validator, the rest are full nodes
-    OneValidator,
+    MinimumValidators,
 
     /// One FullNode is not part of the network for any other nodes
     OneFullNodeOutsider,
 
     /// Two Validators are not part of the network for any other nodes
```
AndrejMitrovic: Documentation needs to be updated.

source/agora/test/Base.d
```diff
@@ -998,12 +999,14 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         .toHexString();
 
     // add two additional validators whose enrollments are not in genesis block
```
AndrejMitrovic: Documentation needs to be updated.

source/agora/consensus/protocol/Nominator.d
MichaelKim20: I didn't know the height of the block there. `uint64_t slot_idx`

AndrejMitrovic:
```D
auto slot_idx = this.ledger.getBlockHeight() + 1;
```

This is already inside of `validateConsensusData`:

```D
const ulong expect_height = this.getBlockHeight() + 1;
```

MichaelKim20: If the value is from another node, I don't think it should be compared to the block height of the current node.

AndrejMitrovic: If `slot_idx != this.ledger.getBlockHeight() + 1;` then we should return `ValidationLevel.kInvalidValue`. We cannot proceed with nominating or voting on a value unless we're voting for the next block height.

MichaelKim20: I will change this part to your opinion, please check it next time.

source/agora/node/Ledger.d
```diff
@@ -350,7 +350,20 @@ public class Ledger
             }
         }
 
-        return null;
+        if (data.enrolls.length >= Enrollment.MinValidatorCount)
```
AndrejMitrovic: I think we also need to add this check to block validation. But we don't want to duplicate the logic.

So I have a suggestion:

Write an overload of `isInvalidReason` in `agora.consensus.validation.Enrollment` which takes an array of Enrollments, the utxo finder, and a count parameter for the number of existing enrollments.

Something like this:

```D
public string isInvalidReason (in Enrollment[] enrolls, UTXOFinder finder, size_t active_enrollments)
{
    if (enrolls.length + active_enrollments < Enrollment.MinValidatorCount)

    foreach (enroll; enrolls)
        if (auto fail_reason = enroll.isInvalidReason(utxo_finder))
            return fail_reason;

    return null;
}
```

Then, use this function from within `validateConsensusData` as well as `isInvalidReason` in `agora.consensus.validation.Block`. Of course this means that `isInvalidReason` in `agora.consensus.validation.Block` will have to be changed to accept another parameter `active_enrollments`. And `validateBlock` in `Ledger` will also have to be changed to provide this argument.

AndrejMitrovic: > and a count parameter for the number of existing enrollments.

This would be the count of active validators for this next block height that we're about to accept or reject.

MichaelKim20: That's a very good point.
I applied it all.

MichaelKim20:
Ready for review

source/agora/node/Ledger.d
AndrejMitrovic: > I applied it all.

There is still no check inside of `agora.consensus.validation.Block`. This was my comment:

> Then, use this function from within validateConsensusData as well as isInvalidReason in `agora.consensus.validation.Block`

MichaelKim20: I knew I had to modify the code for the unit test that wasn't using the network.
It's gonna take a while to finish.

MichaelKim20: I solved this problem using the delegate to get the number of active validators at the block height.

source/agora/common/Types.d
```diff
@@ -58,3 +58,6 @@ public struct QuorumConfig
     /// List of any sub-quorums
     public QuorumConfig[] quorums;
 }
+
+/// Delegate to get the number of active validator at the block height
+public alias ValidatorCountGetter = scope ulong delegate (ulong block_height) @safe nothrow;
```
MichaelKim20: I added a delegate.
I couldn't find a good location. So I added it here.

AndrejMitrovic: I don't see why this is necessary at all. Just pass it as a count?

MichaelKim20: You're right. That's a much better way.

MichaelKim20: @AndrejMitrovic: I fixed it. thank you for review.

source/agora/node/Ledger.d
```diff
@@ -163,15 +167,24 @@ public class Ledger
...
+                `ValidatorSet`.
 
         Returns:
             true if the block was accepted
 
     ***************************************************************************/
 
-    public bool acceptBlock (const ref Block block) @safe
+    public bool acceptBlock (const ref Block block,
+        size_t active_enrollments = size_t.max) @safe
```
AndrejMitrovic: Why does this need to be here? It looks like a workaround for something. We should just get the validator count from the enrollment manager, and that's it. No `active_enrollments` parameter in `acceptBlock` and also `onExternalized` should not have it either.

AndrejMitrovic: In fact `validateBlock` should internally call `getValidatorCount()`

MichaelKim20: I found in the unittest that there was a place to call `acceptBlock` directly.
I did this for the unittest.

https://github.com/bpfkorea/agora/blob/dffeab65a57dafae6c267f7d673b957a42f66c62/source/agora/node/Ledger.d#L645

MichaelKim20: Ah! I know what you want.

MichaelKim20: That's a good idea, but we're going to have to put enrollment in the previous test, so it'll take some time.

MichaelKim20: I looked back and there were quite a few things that needed to be fixed.
I also fixed the code related to freezing period.
I also fixed the test code of agora.node.Ledger.

AndrejMitrovic:
This is missing tests. The `validation.d` update changes protocol rules. We have to test all the parameters in `isInvalidReason`. A few examples that I can think of:

- Blocks with insufficient enrollments should fail validation
- Block with 1 enrollments but where `active_enrollments` is set to 1 should fail validation
- Block with 1 enrollments but where `active_enrollments` is set to 2 should pass validation

MichaelKim20:
@AndrejMitrovic: There have been many conflicts recently due to the addition of ConsensusParams.
I resolved this conflict. 
And added the test code you requested (agora.consensus.validation.Block).

source/agora/node/Ledger.d
```diff
@@ -1130,11 +1236,15 @@ unittest
 
     auto gen_key_pair = getGenesisKeyPair();
 
-    auto storage = new MemBlockStorage();
+    auto FreezingPeriod = 2016;
+    auto validator_cycle = FreezingPeriod / 2;
+    auto params = new immutable(ConsensusParams)(validator_cycle);
+    // Generates a genesis block containing the validator's enrollment data.
+    const genesis = makeGenesisBlock(params);
+    scope storage = new MemBlockStorage([genesis]);
```
AndrejMitrovic: Shouldn't this have been resolved by https://github.com/bpfkorea/agora/issues/907? Then we wouldn't have to fix any of this code, because it could just use the existing genesis block enrollments.

AndrejMitrovic: My comment was on the wrong line. I meant the previous unittests in this module.

However since #907 will not be fixed soon, we can keep it like this for now.

source/agora/node/Ledger.d
```diff
@@ -475,6 +477,95 @@ version (unittest)
     }
 }
 
+
+// Generates a genesis block containing the validator's enrollment data.
+version (unittest)
+private const(Block) makeGenesisBlock (immutable(ConsensusParams) params)
```
AndrejMitrovic, line 539: Is this version of `makeGenesisBlock` the same as the one in `Base.d`? If we need to reuse this code, we should move it to a common module like `agora.utils.Test`.

source/agora/node/Ledger.d
```diff
@@ -1311,6 +1421,34 @@ unittest
...
+        genNormalBlockTransactions((FreezingPeriod / 2) - 1 - ledger.getBlockHeight());
+        assert(ledger.getBlockHeight() == FreezingPeriod / 2 - 1);
+
+        auto enrolls = remakeEnrollment(genesis, params);
+        auto findUTXO = utxo_set.getUTXOFinder();
+        foreach (ref enroll; enrolls)
+            assert(enroll_man.pool.add(enroll, findUTXO));
+
+        genNormalBlockTransactions((FreezingPeriod / 2) - ledger.getBlockHeight());
+        assert(ledger.getBlockHeight() == FreezingPeriod / 2);
```
AndrejMitrovic: Does this code actually re-enroll the validator? Because the test is only testing the block height.

source/agora/test/NetworkManager.d
```diff
@@ -123,7 +124,7 @@ unittest
             }
```
AndrejMitrovic: There is a comment that needs to be updated. It says `the test has 3 nodes`.

source/agora/test/NetworkManager.d
```diff
@@ -136,20 +137,29 @@ unittest
...
     scope(failure) network.printLogs();
     network.waitForDiscovery();
 
     auto nodes = network.clients;
-    auto node_validator = nodes[0];  // validator, creates blocks
-    auto node_test = nodes[1];  // full node, does not create blocks
-    auto node_bad = nodes[2];  // full node, returns bad blocks in getBlocksFrom()
+    assert(nodes.length >= Enrollment.MinValidatorCount+2);
+    // validator, creates blocks
+    auto node_validator = nodes[0];
```
AndrejMitrovic: Is `node_validator` now unused? Then it can be removed.

AndrejMitrovic: Ah I see it's used below in:

```
assert(node_validator.getBlockHeight() == 10,
```

But maybe this should test for all validators instead?

source/agora/node/Ledger.d
MichaelKim20: When the height of the block is 1007 and 2015, re-enroll the validator.
Re-enrollment is required because validation is checked in the process of making blocks.

source/agora/node/Ledger.d
MichaelKim20: The following error occurred and could not be compiled: I think 907 will solve this part.
So we'd better put makeGenesisBlock here and fix it when 907 is far away. What do you think?
```
Cyclic dependency between module constructors/destructors of agora.consensus.ValidatorSet and agora.consensus.EnrollmentManager
```

AndrejMitrovic: Alright.

source/agora/consensus/validation/Block.d
```diff
@@ -665,3 +668,238 @@ unittest
...
+        {
+            tx.outputs ~= Output(Amount.MinFreezeAmount, keypair.address);
+            tx.outputs ~= Output(Amount.MinFreezeAmount, keypair.address);
+            tx.outputs ~= Output(Amount.MinFreezeAmount, keypair.address);
+        }
+        else
+        {
+            tx.type = TxType.Payment;
+            foreach (_; 0 .. Block.TxsInBlock)
+                tx.outputs ~= Output(Amount(100), keypair.address);;
```
AndrejMitrovic, line 752:
```suggestion
                tx.outputs ~= Output(Amount(100), keypair.address);
```

source/agora/node/Ledger.d
AndrejMitrovic: Ok. Well these tests will probably be rewritten some time in the future, because they have become very complex to follow.

AndrejMitrovic:
I'll let @Geod24 do his review as well. I don't have any more comments.

source/agora/consensus/data/Enrollment.d
```diff
@@ -48,6 +48,9 @@ public struct Enrollment
     /// S: A signature for the message H(K, X, n, R) and the key K, using R
     public Signature enroll_sig;
 
+    /// The minimum number of validators required to create a block
+    public static immutable uint MinValidatorCount = 3;
```
Geod24: This should be 1

source/agora/consensus/validation/Block.d
```diff
@@ -107,11 +110,9 @@ public string isInvalidReason (const ref Block block, Height prev_height,
     if (!isStrictlyMonotonic!"a.utxo_key < b.utxo_key"(block.header.enrollments))
         return "Block: The enrollments are not sorted in ascending order";
```
Geod24, line 113:
```D
	if (enrolls.length + active_enrollments < Enrollment.MinValidatorCount)
```

source/agora/node/Ledger.d
```diff
@@ -348,13 +348,10 @@ public class Ledger
...
         }
 
-        foreach (enroll; data.enrolls)
-        {
-            if (auto fail_reason = enroll.isInvalidReason(utxo_finder))
-            {
-                return fail_reason;
-            }
-        }
+        size_t active_enrollments = enroll_man.getValidatorCount(expect_height);
```
Geod24, line 381:
```D
    if (enrolls.length + active_enrollments < Enrollment.MinValidatorCount)
```

source/agora/node/Ledger.d
```diff
@@ -475,6 +477,95 @@ version (unittest)
     }
 }
 
+
+// Generates a genesis block containing the validator's enrollment data.
```
Geod24, line 533:
```suggestion
// Generates a genesis block containing the validator's enrollment data.
//
// TODO: Once issue XXX and issue XXX are fixed, this can be replaced with `agora.utils.Test.XXX`.
// Currently we need it because XXX and XXX.
```

Geod24 (changes requested):

MichaelKim20:
The test code below must now be re-enrollable before it can work. And this has nothing to do with the issue.
Therefore, I would like to remove it first and add it again when the re-enrollment is possible.
https://github.com/bpfkorea/agora/blob/v0.x.x/source/agora/node/Ledger.d#L1081-L1329

There was also something @AndrejMitrovic  mentioned about this.
https://github.com/bpfkorea/agora/pull/875#discussion_r439943082

@AndrejMitrovic @Geod24 Ready for review

source/agora/consensus/validation/Block.d
```diff
@@ -68,6 +69,8 @@ version (unittest)
         prev_height = the height of the direct ancestor of this block
         prev_hash = the hash of the direct ancestor of this block
         findUTXO = delegate to find the referenced unspent UTXOs with
+        active_enrollments = the number of enrollments that do not expire
+            at the block height.
```
```suggestion
            at the next block height (prev_height + 1).
```

source/agora/consensus/validation/Block.d
```diff
@@ -665,3 +674,156 @@ unittest
     // Block: The enrollments are not sorted in ascending order
     assert(!block3.isValid(block2.header.height, hashFull(block2.header), findUTXO));
 }
+
+///
```
```suggestion
/// test that there must always exist active validators
```

source/agora/consensus/validation/Block.d
```diff
@@ -383,9 +390,11 @@ public bool isGenesisBlockValid (const ref Block genesis_block)
 /// Ditto but returns `bool`, only usable in unittests
 version (unittest)
 public bool isValid (const ref Block block, Height prev_height,
-    Hash prev_hash, UTXOFinder findUTXO) nothrow @safe
+    Hash prev_hash, UTXOFinder findUTXO,
+    size_t active_enrollments = Enrollment.MinValidatorCount) nothrow @safe
```
AndrejMitrovic: This should not be a default parameter.

source/agora/consensus/validation/Block.d
```diff
@@ -665,3 +674,156 @@ unittest
...
+    }
+
+    auto block2 = makeNewBlock(block1, txs_2);
+    assert(block2.isValid(block1.header.height, hashFull(block1.header), findUTXO));
+    foreach (ref tx; txs_2)
+        utxo_set.put(tx);
+
+    // When the number of active validator is 0 and the number of enrollments
+    // data in the block is 0,
+    // It isn't approved the validity of the block
```
```suggestion
    // When all existing validators expire at the new block height and the number of enrollments
    // in the new block is 0, the block is considered invalid.
```

This distinction is very important!
```

File: source/agora/consensus/validation/Block.d [hunk 27]
```diff
@@ -665,3 +674,156 @@ unittest
...
+        node_key_pair.V = node_key_pair.v.toPoint();
+
+        auto block3 = makeNewBlock(block2, txs_3);
+        assert(block3.header.enrollments.length < Enrollment.MinValidatorCount);
+        assert(!block3.isValid(block2.header.height, hashFull(block2.header), findUTXO, 0));
+    }
+
+    // When the number of active validator is 0 and the number of enrollments
+    // data in the block is one,
+    // It is approved the validity of the block
```
```suggestion
    // When all existing validators expire at the new block height but the number of enrollments
    // in the new block is at least 1, the block may be considered valid.
```

source/agora/consensus/validation/Block.d
```diff
@@ -665,3 +674,156 @@ unittest
...
+            txs_3 ~= tx;
+        }
+
+        Pair signature_noise = Pair.random;
+        Pair node_key_pair;
+        node_key_pair.v = secretKeyToCurveScalar(keypair.secret);
+        node_key_pair.V = node_key_pair.v.toPoint();
+
+        auto block3 = makeNewBlock(block2, txs_3);
+        assert(block3.header.enrollments.length < Enrollment.MinValidatorCount);
```
```suggestion
        assert(block3.header.enrollments.length == 0);
```

source/agora/consensus/validation/Block.d
```diff
@@ -665,3 +674,156 @@ unittest
...
+
+        Enrollment[] enrollments;
+        enrollments ~= enroll1;
+        enrollments.sort!("a.utxo_key < b.utxo_key");
+        auto block3 = makeNewBlock(block2, txs_3, enrollments);
+        assert(block3.header.enrollments.length == Enrollment.MinValidatorCount);
+        assert(block3.isValid(block2.header.height, hashFull(block2.header), findUTXO, 0));
+    }
+
+    // When the number of enrollments data in the block is 0
```
```suggestion
    // When there are still active validators at the new block height,
    // then new block does not need to contain new enrollments to be considered valid
```

And also please explicitly pass the count of active validators in the call to `isValid`.

source/agora/test/ValidatorCount.d
```diff
@@ -0,0 +1,75 @@
...
+        // create enough tx's for a single block
+        txs = makeChainedTransactions(gen_key_pair, txs, 1);
+
+        // send it to one node
+        txs.each!(tx => node_1.putTransaction(tx));
+
+        containSameBlocks(nodes, block_idx).retryFor(5.seconds);
+    }
+
+    containSameBlocks(nodes, validator_cycle - 1).retryFor(5.seconds);
```
AndrejMitrovic:
```suggestion
```

Unnecessary, the foreach loop already handles this.

source/agora/test/ValidatorCount.d
```diff
@@ -0,0 +1,75 @@
...
+
+    // All active validators is expired on height validator_cycle.
```
AndrejMitrovic:
```suggestion
    // Block will not be created because otherwise there would be no active validators
```

source/agora/test/ValidatorCount.d
```diff
@@ -0,0 +1,75 @@
...
+    {
+        txs = makeChainedTransactions(gen_key_pair, txs, 1);
+        txs.each!(tx => node_1.putTransaction(tx));
+    }
+
+    Thread.sleep(2.seconds);  // wait for propagation
+
+    // Blocks with height validator_cycle can't be created.
```
AndrejMitrovic:
```suggestion
    // New block was not created because all validators would expire
```

AndrejMitrovic (changes requested):
Found a few issues.

I hope #907 will be resolved soon so we can simplify the rest of the unittest code which I didn't comment on.

source/agora/test/ValidatorCount.d
MichaelKim20: Thanks for your review. I fixed all

AndrejMitrovic (approved):

AndrejMitrovic (approved):

PR #877, opened 2020-05-27
Add the ability to create a genesis block with the enrollment data.

MichaelKim20:
Genesis block has frozen UTXOs of the validator's address.
It has the enrollment data using frozen UTXOs.

I extracted code from PR #845 written by @AndrejMitrovic .

Related to #822

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -24,6 +24,7 @@ version (unittest):
 
 import agora.common.Amount;
 import agora.common.BanManager;
+import agora.common.BitField;
 import agora.common.Config;
 import agora.common.Types;
 import agora.common.Hash;
@@ -821,10 +822,6 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
 
     assert(test_conf.nodes >= 2, "Creating a network require at least 2 nodes");
 
-    // custom genesis block
-    immutable string gen_block_hex = test_conf.gen_block != immutable(Block).init
-        ? test_conf.gen_block.serializeFull.toHexString : null;
-
     size_t full_node_idx;
     size_t validator_idx;
 
@@ -851,7 +848,6 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
                 ? test_conf.nodes - 1 : test_conf.min_listeners,
             max_listeners : (test_conf.max_listeners == 0)
                 ? test_conf.nodes - 1 : test_conf.max_listeners,
-            genesis_block : gen_block_hex
         };
 
         return conf;
@@ -986,10 +982,20 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         break;
     }
 
-    immutable GenBlock = test_conf.gen_block != immutable(Block).init
-        ? test_conf.gen_block : GenesisBlock;
+    auto gen_block = makeGenesisBlock(
+        node_configs
+            .filter!(conf => conf.is_validator)
+            .map!(conf => conf.key_pair)
+            .array);
 
-    auto net = new APIManager([GenBlock]);
+    immutable string gen_block_hex = gen_block
+        .serializeFull()
+        .toHexString();
+
+    foreach (ref conf; configs)
+        conf.node.genesis_block = gen_block_hex;
+
+    auto net = new APIManager([gen_block]);
     foreach (ref conf; configs)
         net.createNewNode(conf);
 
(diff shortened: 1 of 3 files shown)


source/agora/test/GenesisBlock.d
```diff
@@ -16,83 +17,37 @@ module agora.test.GenesisBlock;
...
-
-    TestConf conf = { gen_block : genesis[0] };
-    auto network = makeTestNetwork(conf);
+    auto network = makeTestNetwork(TestConf.init);
     network.start();
     scope(exit) network.shutdown();
     scope(failure) network.printLogs();
     network.waitForDiscovery();
 
+    auto gen_tx = GenesisTransaction();
```
AndrejMitrovic: This is not the right genesis transaction. This is the main thread, and it doesn't have the custom genesis block that's set in `Base.d`. You should use `network.blocks`.

In fact all you have to do here is compare `node.getBlocksFrom(0, 1)[0]` with `network.blocks[0]`.

source/agora/test/Base.d
```diff
@@ -1028,3 +1034,70 @@ public bool containSameBlocks (APIS)(APIS nodes, size_t height)
...
+    Returns:
+        The genesis block
+
+*******************************************************************************/
+
+private immutable(Block) makeGenesisBlock (in KeyPair[] key_pairs)
+{
+    // 1 payment tx, the rest are freeze txs
+    Transaction[] txs;
+    txs ~= cast(Transaction)UnitTestGenesisTransaction;
```
AndrejMitrovic: Casting away immutable is bad practice. If we really need mutability we can do `UnitTestGenesisTransaction.serializeFull.deserializeFull!Transaction`.

source/agora/test/GenesisBlock.d
AndrejMitrovic: So the `auto index = b.txs.enumerate.assumeSorted.find!(res => res[1] == gen_tx);` part should be replaced with just checking if the block is the same as the one in the APIManager (`network` variable).

AndrejMitrovic (changes requested):
Small changes requested, the rest looks good.

source/agora/test/GenesisBlock.d
MichaelKim20: That's a good suggestion.

source/agora/test/Base.d
MichaelKim20: That's a good suggestion. `serializeFull.deserializeFull`

source/agora/test/Base.d
```diff
@@ -1028,3 +1034,77 @@ public bool containSameBlocks (APIS)(APIS nodes, size_t height)
...
+        Hash txhash = hashFull(tx);
+        Hash utxo = UTXOSet.getHash(txhash, 0);
+        scope enr = new EnrollmentManager(":memory:", key_pair);
+        scope (exit) enr.shutdown();
+
+        Enrollment enroll;
+        const StartHeight = 1;
+        assert(enr.createEnrollment(utxo, StartHeight, enroll));
+        enrolls ~= enroll;
+    }
```
TrustHenry, line 1091: Genesis block begins with 2 validators.
So, Two enrollments have to be made.

source/agora/test/Base.d
```diff
@@ -1028,3 +1034,77 @@ public bool containSameBlocks (APIS)(APIS nodes, size_t height)
...
+
+    immutable(BlockHeader) makeHeader ()
+    {
+        return immutable(BlockHeader)(
+            Hash.init,   // prev
+            0,           // height
+            merkle_root,
+            BitField!uint.init,
+            Signature.init,
+            enrolls.assumeUnique,
```
TrustHenry, line 1107: enrolls.sort!("a.utxo_key < b.utxo_key");
The enrollments should be aligned like this.

TrustHenry (commented):
Overall, it's good.
In addition, unitTest should check the sorting of the enrolment.

source/agora/test/Base.d
AndrejMitrovic: What? Where do you see this definition?

AndrejMitrovic: The genesis block starts with whatever we define it to start with.

TrustHenry: This is a test code. Consider this case:
https://github.com/bpfkorea/agora/issues/822#issue-617090208

AndrejMitrovic: It's just an example. The code Michael wrote (which was taken from my PR) is generic and can create any number of enrollments.

> Consider this case:

This means "consider this example".

TrustHenry: I think it would be nice to start with more than two enrollments.
Then we can test it more precisely.

MichaelKim20: Thank you for your good suggestion.

source/agora/test/Base.d
MichaelKim20: Thank you for your good suggestion.

source/agora/test/Base.d
```diff
@@ -1028,3 +1034,79 @@ public bool containSameBlocks (APIS)(APIS nodes, size_t height)
...
+private immutable(Block) makeGenesisBlock (in KeyPair[] key_pairs)
+{
+    import agora.common.Serializer;
+
+    // 1 payment tx, the rest are freeze txs
+    Transaction[] txs;
+    txs ~= UnitTestGenesisTransaction.serializeFull.deserializeFull!Transaction;
+    Enrollment[] enrolls;
+
+    foreach (idx; 0 .. max(Enrollment.MinValidatorCount, key_pairs.length))
```
AndrejMitrovic: It would be much better to add an assertion that `key_pairs.length >= Enrollment.MinValidatorCount`.

If you generate random keypairs here, how will they be used by the Validators? Who will validate blocks [1..$]?

source/agora/consensus/data/Enrollment.d
```diff
@@ -51,6 +51,9 @@ public struct Enrollment
     /// The cycle length for a validator
     public static immutable uint ValidatorCycle = 1008; // freezing period / 2
 
+    /// The minimum number of validators required to a create block
+    public static immutable uint MinValidatorCount = 2;
```
AndrejMitrovic: Why? This is introducing an arbitrary protocol change.

source/agora/test/Base.d
MichaelKim20: Only one validator is created when the value of TestConf.topology is NetworkTopology.OneValidator.
This is because the existing test code does not work.
So I created a dummy enrollment data. Is there any other good way to solve this?

AndrejMitrovic: Just let it be one validator. There don't have to be two validators.

MichaelKim20: Thank you for the good suggestion.

source/agora/consensus/data/Enrollment.d
MichaelKim20: Yes, you're right. I'll remove this.

AndrejMitrovic:
I'll look at this PR more tomorrow.

But just to be more clear about this: If the test is wrong (it only has one validator), then the test should be fixed. We shouldn't add workaround code if our tests are broken. We should fix the tests instead. 

Because otherwise adding workarounds like this just means somebody else will have to fix the tests later on when we add block signature validation. 

So if we introduce a protocol rule that says "minimum 2 validators are enrolled at all times", it means we have to fix `NetworkTopology.OneValidator` and change it to  `NetworkTopology.TwoValidators`.

source/agora/test/GenesisBlock.d
```diff
@@ -16,83 +17,36 @@ module agora.test.GenesisBlock;
...
     auto node_1 = nodes[0];
 
-    // create spending txs refering to our own genesis block txs
-    auto txes = makeChainedTransactions(kp, null, 1, 40_000_000, genesis[1]);
+    assert(network.blocks[0].header.enrollments.length == 4);
+
+    auto index = network.blocks[0].txs
+        .enumerate.assumeSorted
+        .find!(res => res[1] == gen_tx);
+    assert(!index.empty);
```
AndrejMitrovic: We want to test that `NodeConfig.genesis_block` is read properly. 

```suggestion
    nodes.all!(node => node.getBlocksFrom(0, 1)[0] == network.blocks[0])
        .retryFor(2.seconds);
```

AndrejMitrovic (changes requested):

source/agora/test/GenesisBlock.d
MichaelKim20: Yes, that's the more necessary test. 
I applied your suggestion.

AndrejMitrovic (approved):

Issue #789, opened 2020-05-06
Restore the enrollment information of a node


linked0:
The table managing the enrollment information about the node can be removed because all the data it manages can be restored in many ways. The use of a database table means that there are many complex data so it should be managed efficiently but the table manages very small data now. This task contains a few things to do as follows in the Definition of Done.

Definition of Done:

- [x] Restoring enrollment information used in registering an enrollment by a node
- [x] Restoring a block height in which a validator should reveal its pre-image
- [x] Remove`remove` from EnrollmentManager
- [x] Change the modifier of `remove` of `ValidatorSet` into private.

How to implement:

- The enrollment manager restores the information from the `enrollments` field of a block header.
- The block height for the next revelation of a pre-image is set to the current block height when a node starts.

PR #880, opened 2020-05-27
Get a public key used for enrollment process

linked0:
The commit in this PR will be used for other branches so I cherry-picked the "Get a public key used for enrollment process" commit which has been already contained in the PR #789 and will be removed from the PR after this PR is merged.


linked0:
Ready for review.
This is simple code and I will have 2 PRs that use this commit.
So, If you have a quick review about this or merge it, It would be very appreciated.

Geod24 (approved):

PR #884, opened 2020-05-28
Create a function for the validation of the Genesis block

### LINKED ISSUE #747: Block with height 0 also needs to be verified
This code verifies that the first block equals the genesis block:

https://github.com/bpfkorea/agora/blob/1dcc4de9cfb7adfe317c2571c2b5cd3777683156/source/agora/consensus/Validation.d#L852-L855

But it also assumes that the definition of `GenesisBlock` is correct. It should verify all the rules we have in place: merkle root, transaction limit, etc..

TrustHenry:
Block with height 0 also needs to be verified

This validation the definition of our Genesis block.

Related to #747

### DIFF
--- a/source/agora/consensus/Genesis.d
+++ b/source/agora/consensus/Genesis.d
@@ -18,6 +18,7 @@ import agora.common.Hash;
 import agora.common.crypto.Key;
 import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
+import agora.consensus.validation.Block;
 
 /// Contains a pointer to the genesis block.
 version (unittest)
@@ -50,13 +51,12 @@ public ref immutable(Block) GenesisBlock () nothrow @safe @nogc
 
 pragma(inline, true)
 public void setGenesisBlock (immutable Block* block)
-    nothrow @trusted @nogc
+    @safe
 {
     assert(block !is null);
 
-    // note: see #747
-    //if (auto reason = isInvalidReason(block, 0, Hash.init, null))
-    //    throw new Exception(reason);
+    if (auto reason = isGenesisBlockInvalidReason(*gen_block))
+        throw new Exception(reason);
 
     gen_block = block;
 }
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -14,12 +14,17 @@
 module agora.consensus.validation.Block;
 
 import agora.common.Amount;
+import agora.common.Hash;
+import agora.common.Set;
 import agora.common.Types;
 import agora.consensus.data.Block;
+import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
 import VEn = agora.consensus.validation.Enrollment;
 import VTx = agora.consensus.validation.Transaction;
 import agora.consensus.Genesis;
+
+import std.algorithm;
 
 version (unittest)
 {
@@ -54,6 +59,10 @@ version (unittest)
     it as a spent output. See the `findNonSpent` function in the
     unittest for an example.
 
+    As a special case, the genesis block is rejected by this function.
+    Validation of a genesis block should be done through the
+    `isGenesisBlockInvalidReason` function.
+
     Params:
         block = the block to check
         prev_height = the height of the direct ancestor of this block
@@ -70,11 +79,6 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
     in Hash prev_hash, UTXOFinder findUTXO) nothrow @safe
 {
     import std.algorithm;
-
-    // special case for the genesis block
-    if (block.header.height == 0)
-        return block == GenesisBlock ?
-            null : "Block: Height 0 but not Genesis block";
 
     if (block.header.height > prev_height + 1)
         return "Block: Height is above expected height";
@@ -112,6 +116,270 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
     return null;
 }
 
+/*******************************************************************************
+
+    Check the validity of a genesis block
+
+    Follow the same rules as for `Block` except for the following:
+        - Block height must be 0
+        - The previous block hash of the block must be empty
+        - The number of transactions in the block must be in the range `(0;
+          Block.TxsInBlock]`.
+        - Transactions must have no input
+        - Transactions must have at least one output
+        - All the enrollments pass validation, which implies:
+            - The enrollments refer to freeze tx's in this block
+            - The signature for the Enrollment is valid
+
+    Params:
+        block = The genesis block to check
+
+    Returns:
+        `null` if the genesis block is valid, otherwise a string explaining
+        the reason it is invalid.
+
+*******************************************************************************/
+
+public string isGenesisBlockInvalidReason (const ref Block block) nothrow @safe
+{
+    if (block.header.height != 0)
+        return "GenesisBlock: The height of the block is not 0";
+
+    if (block.header.prev_block != Hash.init)
+        return "GenesisBlock: Header.prev_block is not empty";
+
+    if (block.txs.length == 0)
+        return "GenesisBlock: Transaction(s) are empty";
+
+    if (block.txs.length > Block.TxsInBlock)
+        return "GenesisBlock: The number of transactions is out of bounds";
+
+    if (!block.txs.isSorted())
+        return "GenesisBlock: Transactions are not sorted";
+
+    UTXOSetValue[Hash] utxo_set;
+    foreach (const ref tx; block.txs)
+    {
+        if (!(tx.type == TxType.Payment || tx.type == TxType.Freeze))
+            return "GenesisBlock: Invalid enum value for type field";
+
+        if (tx.inputs.length != 0)
+             return "GenesisBlock: Transactions must not have input";
+
+        if (tx.outputs.length == 0)
+            return "GenesisBlock: No output(s) in the transaction";
+
+        Hash tx_hash = tx.hashFull();
+        foreach (idx, const ref output; tx.outputs)
+        {
+            // disallow negative amounts
+            if (!output.value.isValid())
+                return "GenesisBlock: Output(s) overflow or underflow"
+                    ~ "in the transaction";
+
+            // disallow 0 amount
+            if (output.value == Amount(0))
+                return "GenesisBlock: Value of output is 0"
+                    ~ "in the transaction";
+
+            const UTXOSetValue utxo_value = {
+                unlock_height: 0,
+                type: tx.type,
+                output: output
+            };
+            utxo_set[UTXOSetValue.getHash(tx_hash, idx)] = utxo_value;
+        }
+    }
+
+    Hash[] merkle_tree;
+    if (block.header.merkle_root !=
+        Block.buildMerkleTree(block.txs, merkle_tree))
+        return "GenesisBlock: Merkle root does not match header's";
+
+    // If there are no enrollments, return them here early.
+    if (block.header.enrollments.length == 0)
+        return null;
+
+    if (!isStrictlyMonotonic!"a.utxo_key < b.utxo_key"
+        (block.header.enrollments))
+        return "GenesisBlock: The enrollments should be arranged in "
+            ~ "ascending order by the utxo_key";
+
+    Set!Hash used_utxos;
+    bool findUTXO (Hash utxo_hash, size_t index, out UTXOSetValue value)
+        nothrow @safe
+    {
+        if (utxo_hash in used_utxos)
+            return false;  // double-spend
+
+        if (auto ptr = utxo_hash in utxo_set)
+        {
+            value = *ptr;
+            used_utxos.put(utxo_hash);
+            return true;
+        }
+        return false;
+    }
+
+    foreach (const ref enrollment; block.header.enrollments)
+    {
+        if (auto fail_reason = VEn.isInvalidReason(enrollment, &findUTXO))
+            return fail_reason;
+    }
+
+    return null;
+}
+
+/// Genesis block validation fail test
+unittest
+{
+    import agora.common.Serializer;
+
+    Block block = GenesisBlock.serializeFull.deserializeFull!Block;
+    assert(block.isGenesisBlockValid());
+
+    // don't accept block height 0 from the network
+    assert(!block.isValid(0, Hash.init, null));
+
+    // height check
+    block.header.height = 1;
+    assert(!block.isGenesisBlockValid());
+
+    block.header.height = 0;
+    assert(block.isGenesisBlockValid());
+
+    // .prev_block check
+    block.header.prev_block = block.header.hashFull();
+    assert(!block.isGenesisBlockValid());
+
+    block.header.prev_block = Hash.init;
+    assert(block.isGenesisBlockValid());
+
+    Transaction[] txs =
+        GenesisBlock.txs.serializeFull.deserializeFull!(Transaction[]);
+
+    void buildMerkleTree (ref Block block)
+    {
+        Hash[] merkle_tree;
+        block.header.merkle_root =
+            Block.buildMerkleTree(block.txs, merkle_tree);
+    }
+
+    Transaction makeNewTx ()
+    {
+        Transaction new_tx =
+        {
+            TxType.Payment,
+            inputs: [],
+            outputs: [Output(Amount(100), KeyPair.random().address)]
+        };
+        return new_tx;
+    }
+
+    // Check consistency of `txs` field
+    {
+        // Txs length check
+        Transaction[] null_txs;
+        block.txs = null_txs;
+        assert(!block.isGenesisBlockValid());
+
+        foreach (_; 0 .. Block.TxsInBlock)
+            block.txs ~= makeNewTx();
+        block.txs.sort;
+        assert(block.txs.length == Block.TxsInBlock);
+        buildMerkleTree(block);
+        assert(block.isGenesisBlockValid());
+
+        // Txs sorting check
+        block.txs.reverse;
+        buildMerkleTree(block);
+        assert(!block.isGenesisBlockValid());
+
+        block.txs.reverse;
+        buildMerkleTree(block);
+        assert(block.isGenesisBlockValid());
+
+        // Txs length out of bounds check
+        block.txs ~= makeNewTx();
+        block.txs.sort;
+        buildMerkleTree(block);
+        assert(block.txs.length == Block.TxsInBlock + 1);
+        assert(!block.isGenesisBlockValid());
+
+        block = GenesisBlock.serializeFull.deserializeFull!Block;
+
+        // Txs type check
+        block.txs[0].type = cast(TxType)2;
+        buildMerkleTree(block);
+        assert(!block.isGenesisBlockValid());
+
+        block.txs[0].type = TxType.Payment;
+        buildMerkleTree(block);
+        assert(block.isGenesisBlockValid());
+
+        block.txs[0].type = TxType.Freeze;
+        buildMerkleTree(block);
+        assert(block.isGenesisBlockValid());
+
+        // Input empty check
+        block.txs[0].inputs ~= Input.init;
+        buildMerkleTree(block);
+        assert(!block.isGenesisBlockValid());
+
+        block.txs = txs;
+        buildMerkleTree(block);
+        assert(block.isGenesisBlockValid());
+
+        // Output not empty check
+        block.txs[0].outputs = null;
+        buildMerkleTree(block);
+        assert(!block.isGenesisBlockValid());
+
+        // disallow 0 amount
+        Output zeroOutput =
+            Output(Amount.invalid(0), WK.Keys[0].address);
+        block.txs[0].outputs ~= zeroOutput;
+        buildMerkleTree(block);
+        assert(!block.isGenesisBlockValid());
+    }
+
+    block = GenesisBlock.serializeFull.deserializeFull!Block;
+
+    // enrollments validation test
+    Enrollment[] enrolls;
+    enrolls ~= Enrollment.init;
+    block.header.enrollments = enrolls;
+    assert(!block.isGenesisBlockValid());
+
+    block.header.enrollments.length = 0;
+    assert(block.isGenesisBlockValid());
+
+    block = GenesisBlock.serializeFull.deserializeFull!Block;
+
+    // modify the last hex byte of the merkle root
+    block.header.merkle_root[][$ - 1]++;
+    assert(!block.isGenesisBlockValid());
+
+    // now restore it back to what it was
+    block.header.merkle_root[][$ - 1]--;
+    assert(block.isGenesisBlockValid());
+    const last_root = block.header.merkle_root;
+
+    // the previous merkle root should not match the new txs
+    block.txs ~= makeNewTx();
+    block.header.merkle_root = last_root;
+    assert(!block.isGenesisBlockValid());
+}
+
+/// Ditto but returns `bool`, only usable in unittests
+/// Only the genesis block Validation
+version (unittest)
+public bool isGenesisBlockValid (const ref Block genesis_block)
+    nothrow @safe
+{
+    return isGenesisBlockInvalidReason(genesis_block) is null;
+}
+
 /// Ditto but returns `bool`, only usable in unittests
 version (unittest)
 public bool isValid (const ref Block block, ulong prev_height,
@@ -130,7 +398,7 @@ unittest
     scope findUTXO = &utxos.findUTXO;
 
     auto gen_key = getGenesisKeyPair();
-    assert(GenesisBlock.isValid(GenesisBlock.header.height, Hash.init, null));
+    assert(GenesisBlock.isGenesisBlockValid());
     auto gen_hash = GenesisBlock.header.hashFull();
 
     utxos.put(GenesisTransaction);
@@ -294,7 +562,7 @@ unittest
     UTXOFinder findUTXO = utxo_set.getUTXOFinder();
 
     auto gen_key = getGenesisKeyPair();
-    assert(GenesisBlock.isValid(GenesisBlock.header.height, Hash.init, null));
+    assert(GenesisBlock.isGenesisBlockValid());
     auto gen_hash = GenesisBlock.header.hashFull();
     foreach (ref tx; GenesisBlock.txs)
         utxo_set.put(tx);
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -17,7 +17,6 @@ import agora.common.Amount;
 import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import VEn = agora.consensus.validation.Enrollment;
 import VTx = agora.consensus.validation.Transaction;
 import agora.consensus.Genesis;
@@ -287,7 +286,6 @@ unittest
     import agora.common.Amount;
     import agora.consensus.data.Enrollment;
     import agora.consensus.data.Transaction;
-    import agora.consensus.UTXOSet;
 
     import std.algorithm;
     import std.range;
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -55,7 +55,6 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import agora.consensus.EnrollmentPool;
 import agora.consensus.PreImage;
 import agora.consensus.validation;
--- a/source/agora/consensus/EnrollmentPool.d
+++ b/source/agora/consensus/EnrollmentPool.d
@@ -25,7 +25,6 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import agora.consensus.validation;
 import agora.utils.Log;
 
--- a/source/agora/consensus/Quorum.d
+++ b/source/agora/consensus/Quorum.d
@@ -19,7 +19,6 @@ import agora.common.Set;
 import agora.common.Types;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 
 import std.algorithm;
 import std.array;
--- a/source/agora/consensus/UTXOSet.d
+++ b/source/agora/consensus/UTXOSet.d
@@ -388,7 +388,6 @@ unittest
     import agora.common.Amount;
     import agora.consensus.data.Transaction;
     import agora.consensus.data.UTXOSetValue;
-    import agora.consensus.UTXOSet;
 
     KeyPair[] key_pairs = [KeyPair.random, KeyPair.random];
 
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -24,7 +24,6 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import agora.consensus.PreImage;
 import agora.consensus.validation;
 import agora.utils.Log;
--- a/source/agora/consensus/validation/Enrollment.d
+++ b/source/agora/consensus/validation/Enrollment.d
@@ -20,7 +20,6 @@ import agora.common.crypto.Key;
 import agora.common.crypto.Schnorr;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 
 import std.conv;
 
--- a/source/agora/consensus/validation/Transaction.d
+++ b/source/agora/consensus/validation/Transaction.d
@@ -17,7 +17,6 @@ import agora.common.Amount;
 import agora.common.Hash;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 
 version (unittest)
 {
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -26,7 +26,6 @@ import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.Transaction;
 import agora.consensus.EnrollmentManager;
 import agora.consensus.protocol.Nominator;
-import agora.consensus.UTXOSet;
 import agora.consensus.Quorum;
 import agora.network.NetworkManager;
 import agora.node.BlockStorage;
--- a/source/agora/test/Ledger.d
+++ b/source/agora/test/Ledger.d
@@ -24,7 +24,6 @@ import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import agora.consensus.Genesis;
 import agora.consensus.validation;
 import agora.test.Base;


source/agora/consensus/validation/GenesisBlock.d
```diff
@@ -0,0 +1,181 @@
...
+    {
+        if (!(tx.type == TxType.Payment || tx.type == TxType.Freeze))
+            return "GenesisBlock: Transaction type allows only "
+                ~ "Payment and Freeze";
+
+        if (tx.inputs.length == 0)
+            return "GenesisBlock: No a input in the transaction";
+
+        if (tx.inputs.length != 1)
+            return "GenesisBlock: Transaction must have only one input";
```
MichaelKim20: I think I don't need `if (tx.inputs.length == 0)`.

MichaelKim20 (commented):
The basic structure is good. I'd like to supplement a few more.
I think we need to add a test code.
Especially, I think more testing is needed in the verification of UTXO, enrolment.

source/agora/consensus/validation/GenesisBlock.d
```diff
@@ -0,0 +1,216 @@
...
```
AndrejMitrovic: You just need one check, not both.

And, you can decide to change the rule here. Make inputs equal zero rather than one. There is no real input.

source/agora/consensus/validation/GenesisBlock.d
```diff
@@ -0,0 +1,216 @@
...
+    Params:
+        block = The genesisBlock to check
+
+    Returns:
+        `null` if the genesisBlock is valid, a string explaining the reason it
+        is invalid otherwise.
+
+*******************************************************************************/
+
+public string isInvalidReason (const ref Block genesisBlock) nothrow @safe
```
AndrejMitrovic: This should not be a public function, and it should not be an overload.

AndrejMitrovic:
Wouldn't it be better to add genesis block special code path within the existing `isInvalidReason (block)` function? Otherwise all of this code is duplicated. And if we update the code in one location we could forget updating it in another location.

source/agora/consensus/validation/GenesisBlock.d
```diff
@@ -0,0 +1,216 @@
...
+/// Ditto but returns `bool`, only usable in unittests
+version (unittest)
+public bool isValid (const ref Block genesisBlock) nothrow @safe
+{
+    return isInvalidReason(genesisBlock) is null;
+}
+
+unittest
+{
+    assert(isValid(UnitTestGenesisBlock));
```
AndrejMitrovic: These tests belongs in the `agora.consensus.Genesis` module. Then you don't have to make those two symbols public.

source/agora/test/Base.d
```diff
@@ -40,6 +40,7 @@ import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSet;
 import agora.consensus.EnrollmentManager;
 import agora.consensus.Genesis;
+import VGb = agora.consensus.validation.GenesisBlock;
```
AndrejMitrovic: Why? Just import `agora.consensus.validation` and call `isValid` on the block.

AndrejMitrovic (changes requested):

AndrejMitrovic:
There is not a single test where the genesis block fails validation in this PR.

TrustHenry:
> Wouldn't it be better to add genesis block special code path within the existing isInvalidReason (block) function? Otherwise all of this code is duplicated. And if we update the code in one location we could forget updating it in another location.

@AndrejMitrovic Thank you for your reviews
I'm asking because I don't understand your opinion.

Does this mean that its best position is to make it in the `agora.consensus.Genesis` ?

I made it according to the function classification of the package name.
Well, if this is okay with you, Andrej's opinion is good.

AndrejMitrovic:
> I made it according to the function classification of the package name.

But those are each testing a different type of structure. That is the entire purpose of having them named the same, so the right function overload is selected based on the type of the argument.

@Geod24 please share your thoughts here.

Geod24:
To add to what @AndrejMitrovic is saying:

- Overloads are based on *type*. `Transaction`, `Input`, `Block`, etc... Here you are trying to add an overload for the type `Block` but we already have one!
- What you want to do is call one function if the type have a certain value (it's a genesis block), and the other function otherwise.
- However, since we have this overload model in place, and calling one function or the other would rely on the programmer doing things correctly, this is not a good approach.
- Instead, we want to have a *single* function to call (an overload on `Block`) and that overload selectively call the other function when it sees it is a genesis block.

In essence, you could have the `private` function `isValidGenesisBlock` (or whatever name, but *not* `isValid`) in `Block` validation (because it does validate a `Block`, just a special one) and have it called from `isValid` (like you do currently anyway).

Think of it this way: Imagine you have an algorithm that multiplies two values together.
The type of your operands are `int`, so the prototype is: `int mul (int a, int b)`.
However, in some case (e.g. when one of the operand is a power of 2) you want to call a special, optimized function: `int fast_mul(int x, int log);` where `sqr` is the log2 of the value.
You could do this by having the *caller* check if one of the operands is a power of two. So one would write:
```D
int a, b;
if (isSquareRootOfTwo(a))
    fast_mul(b, log2(a);
else if (isSquareRootOfTwo(b))
    fast_mul(a, log2(b));
else
    mul(a, b);
```
That's a bit... Complex, and easy to forget / get wrong. What if the caller invert the parameter ? Or use log10 ? What you want to do is move this logic inside `mul` so  the caller only has to do `mul(a, b)` and things work.

The reason we have overload is to allow this kind of logic to be built into the language for *disjoint types*. So for example you can have `int mul(int, int)` and `double mul(double, double)` and it just works. But if you need to call a different function based on a value rather than a type, don't use overloads.

source/agora/consensus/validation/GenesisBlock.d
TrustHenry: Apply the suggestion.

source/agora/consensus/validation/GenesisBlock.d
TrustHenry: Apply the suggestion.

source/agora/consensus/validation/GenesisBlock.d
TrustHenry: Apply the suggestion.

source/agora/test/Base.d
TrustHenry: I remove it.

TrustHenry:
> There is not a single test where the genesis block fails validation in this PR.

I'm going to add it :)

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +115,151 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    Returns:
+        `null` if the genesisBlock is valid, a string explaining the reason it
+        is invalid otherwise.
+
+*******************************************************************************/
+
+private string isValidGenesisBlock (const ref Block genesisBlock) nothrow @safe
+{
+    if (genesisBlock == Block.init)
+        return "GenesisBlock: The block is empty";
```
Geod24: This is implicitly checked after, you can remove it.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +115,151 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    UTXOSetValue[Hash] utxo_set;
+    foreach (const ref tx; genesisBlock.txs)
+    {
+        if (!(tx.type == TxType.Payment || tx.type == TxType.Freeze))
+            return "GenesisBlock: Transaction type allows only "
+                ~ "Payment and Freeze";
+
+        if (tx.inputs.length != 1)
+            return "GenesisBlock: Transaction must have only one input";
```
Geod24: I think we should remove that empty transaction ?

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +115,151 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+        if (tx.outputs.length == 0)
+            return "GenesisBlock: No output(s) in the transaction";
+
+        Hash tx_hash = tx.hashFull();
+        foreach (idx, const ref output; tx.outputs)
+        {
+            if (output.address == PublicKey.init)
+                return "GenesisBlock: The address of the output(s) is empty"
+                    ~ "in the transaction";
```
Geod24: Remove. There is nothing wrong with an output address of `0` (it's a burn)

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +115,151 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+                return "GenesisBlock: Value of output is 0"
+                    ~ "in the transaction";
+
+            if (tx.inputs[0].signature != Signature.init
+                && output.address.verify(tx.inputs[0].signature, tx_hash[]))
+                return "GenesisBlock: Input has invalid signature "
+                    ~ "in the transaction";
+
+            const utxo_value = UTXOSetValue(genesisBlock.header.height,
+                tx.type, output);
```
Geod24:
```suggestion
            const utxo_value = UTXOSetValue(/* height: */ 0, tx.type, output);
```

In general your approach is good, but if it forces you to break down your statement on multiple lines, look for alternatives.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +115,151 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+            // disallow 0 amount
+            if (output.value == Amount(0))
+                return "GenesisBlock: Value of output is 0"
+                    ~ "in the transaction";
+
+            if (tx.inputs[0].signature != Signature.init
+                && output.address.verify(tx.inputs[0].signature, tx_hash[]))
+                return "GenesisBlock: Input has invalid signature "
+                    ~ "in the transaction";
```
Geod24: What is this ? It will never pass.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +115,151 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    if (!isStrictlyMonotonic!"a.utxo_key < b.utxo_key"
+        (genesisBlock.header.enrollments))
+        return "GenesisBlock: The enrollments should be arranged in "
+            ~ "ascending order by the utxo_key";
+
+    foreach (const ref enrollment; genesisBlock.header.enrollments)
+    {
+        if (auto fail_reason = VEn.isInvalidReason(enrollment, &findUTXO))
+            return "GenesisBlock: " ~ fail_reason;
```
Geod24: Return `fail_reason`. Those routines should not allocate.

Geod24 (changes requested):
A few things, but nice work.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +116,138 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+              with the `Amount` rule
+        - The merkle root in the header matches the re-built merkle tree root
+        - All the enrollments pass validation, which implies:
+            - The Enrolments should be arranged in ascending order by utxo_key
+            - Signature are authentic
+            - The enrollments refer to freeze tx's in this block
+            - The freeze tx's spend an output which must be found with
+              the findUTXO() delegate
+            - Freeze transactions referenced in registration should not be
+              double-spend
```
Geod24: This is repeating much of the `Block` part, right ? I'd rather have this say "Follow the Block rules except for the following:" so it's easier on the reader.

Geod24:
Can you squash the commits together (`fix` in `git`) and rebase on `v0.x.x` ?

Geod24:
Note, this issue is pretty important for us to move forward

TrustHenry:
PR that was closed by mistake has been reopened.

source/agora/consensus/validation/Block.d
TrustHenry: I fixed it.

TrustHenry:
Ready for review.

source/agora/consensus/validation/Block.d
TrustHenry: I think the Transaction.inputs in Genesis Block can be empty.

AndrejMitrovic:
PR should be renamed. :)

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    Params:
+        block = The genesis block to check
+
+    Returns:
+        `null` if the genesis block is valid, a string explaining the reason it
+        is invalid otherwise.
+
+*******************************************************************************/
+
+private string isValidGenesisBlock (const ref Block genesisBlock) nothrow @safe
```
```suggestion
private string isValidGenesisBlock (const ref Block genesis_block) nothrow @safe
```

Or alternatively:

```suggestion
private string isValidGenesisBlock (const ref Block block) nothrow @safe
```

We need to follow our style guide.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
 
+/*******************************************************************************
+
+    Check the validity of a genesis block
+
+    Follow the Block rules except for the following
+        - Block height is 0
+        - The previous block hash of the block must be empty
+        - The number of transactions in the block is above zero and less than
+          Block.TxsInBlock
```
```suggestion
          or equal to Block.TxsInBlock
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+        - All the transactions pass validation, which implies:
+            - Transaction type in block should only be Payment or Freeze
+            - There is empty a input for the transaction
```
```suggestion
            - The inputs of the transaction(s) are empty
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+            - There is one or more outputs of the transaction
+        - All the enrollments pass validation, which implies:
+            - The enrollments refer to freeze tx's in this block
```
AndrejMitrovic, line 135:
```suggestion
            - The enrollments refer to the freeze tx's in this block
```

AndrejMitrovic, line 131: Sorry I accidentally suggested "the" here, but I was wrong. :)

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    Params:
+        block = The genesis block to check
```
AndrejMitrovic, line 135: When you change `genesisBlock` to follow the styling convention, you will also have to change this param here (it's currently wrong).

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    Returns:
+        `null` if the genesis block is valid, a string explaining the reason it
+        is invalid otherwise.
```
AndrejMitrovic: I know you probably didn't write this, there is another one of these somewhere else, but it's not proper English. This is the right way:

```suggestion
        `null` if the genesis block is valid, otherwise a string explaining the reason it
        is invalid.
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+private string isValidGenesisBlock (const ref Block genesisBlock) nothrow @safe
+{
+    if (genesisBlock.header.height != 0)
+        return "GenesisBlock: The height of the block is not 0";
+
+    if (genesisBlock.header.prev_block != Hash.init)
+        return "GenesisBlock: Header.prev_block is not empty";
+
+    if (genesisBlock.txs.length == 0)
+        return "GenesisBlock: Transaction(s) is empty";
```
```suggestion
        return "GenesisBlock: Transaction(s) are empty";
```

source/agora/consensus/validation/Block.d
```diff
@@ -73,8 +80,7 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
 
     // special case for the genesis block
     if (block.header.height == 0)
-        return block == GenesisBlock ?
-            null : "Block: Height 0 but not Genesis block";
+        return isValidGenesisBlock(block);
```
AndrejMitrovic: Alright, I see there is actually a problem with this now

There is a possibility of a DDoS here. We should only accept a block with height zero if none was stored yet. If we already have a block with height 0 in our ledger, then calling `isValidGenesisBlock` on blocks with height zero received from the network becomes a form of DDoS. We should outright reject such blocks.

We only set a custom genesis block through the Config file. Therefore we never have to verify blocks with height 0 coming from the network. They should be rejected.

Here is what I suggest:

Inside here simply do:

```
if (block.header.height == 0)
    return "Block: Rejected Genesis block"
```

And you will indeed have to make `isValidGenesisBlock` `public`.

Furthermore, there is a critical piece of code missing from your PR. `setGenesisBlock` is not validating the genesis block which was set. This will have to be fixed to call this new `public` function.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+        return "GenesisBlock: Header.prev_block is not empty";
+
+    if (genesisBlock.txs.length == 0)
+        return "GenesisBlock: Transaction(s) is empty";
+
+    if (!genesisBlock.txs.isSorted())
+        return "GenesisBlock: Transactions are not sorted";
+
+    if (genesisBlock.txs.length > Block.TxsInBlock)
+        return "GenesisBlock: The number of transactions is out of bounds";
```
AndrejMitrovic: This right here is your opportunity for a DDoS. Consider that you have three checks here:

- Length check against zero
- Sortedness check
- Out of bounds check

If a byzantine node sends a block with height 0 and 1,000,000 transactions, this code will attempt to determine if the 1 million transactions are sorted, and then it will check the out of bounds later.

We should always check out of bounds conditions **first** before any expensive operation.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    UTXOSetValue[Hash] utxo_set;
```
AndrejMitrovic, line 160: Please move this to line 178, closest to where it's actually used.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+            return "GenesisBlock: Transaction must have only one input";
+
+        if (tx.inputs[0] != Input.init)
+             return "GenesisBlock: Not empty a input in the transaction";
+
+        if (tx.outputs.length == 0)
+            return "GenesisBlock: No output(s) in the transaction";
+
+        Hash tx_hash = tx.hashFull();
+        foreach (idx, const ref output; tx.outputs)
```
AndrejMitrovic, line 173: @Geod24 it occurs to me that we don't have limits in place for the number of Outputs (both here and for regular blocks). Sounds like a DDoS vector that we might want to fix?

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+        if (auto ptr = utxo_hash in utxo_set)
+        {
+            value = *ptr;
+            used_utxos.put(utxo_hash);
+            return true;
+        }
+        return false;
+    }
+
+    // If there are no enrollments, return them here early.
```
AndrejMitrovic: Move this check above where `used_utxos` is defined

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +118,268 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    if (genesisBlock.header.enrollments.length == 0)
+        return null;
+
+    if (!isStrictlyMonotonic!"a.utxo_key < b.utxo_key"
```
AndrejMitrovic: Also this one should be moved. So basically, just move `findUTXO` and `used_utxos` below this one.

AndrejMitrovic:
Ok done with review. :)

source/agora/consensus/validation/Block.d
TrustHenry: That's right. It's an important point.

And we can setGenesisBlock only in FullNode.

source/agora/consensus/validation/Block.d
TrustHenry: The location is closest. It located outside the transaction loop.

AndrejMitrovic: Ah sorry I misread it.

source/agora/consensus/validation/Block.d
TrustHenry: I moved it.

source/agora/consensus/validation/Block.d
TrustHenry: ^ I moved it.

TrustHenry:
@AndrejMitrovic Thank you for your reviews.
Most suggestions have been applied.

TrustHenry:
An error occurred in the integration test. I'm looking at it.

https://github.com/bpfkorea/agora/pull/884/checks?check_run_id=749058899#step:8:64

AndrejMitrovic:
The problem is caused by `Genesis` importing into `validation.Block`, which itself imports into `UTXOSet` which itself imports `d2sqlite`. 

`validation.Block` only needs a few symbols from the `UTXOSet` module: `UTXOFinder`, `UTXOSetValue`, and `TestUTXOSet` for its unittests.

Maybe we can fix this by splitting this stuff into multiple modules. These 3 symbols which could live in another module which `validation.Block` imports.

TrustHenry:
It was unblocked by the merged PR #947
Ready for review.

source/agora/consensus/Genesis.d
```diff
@@ -50,14 +50,9 @@ public ref immutable(Block) GenesisBlock () nothrow @safe @nogc
 
 pragma(inline, true)
 public void setGenesisBlock (immutable Block* block)
-    nothrow @trusted @nogc
+    @safe
 {
     assert(block !is null);
```
AndrejMitrovic, line 56: If you were unblocked, it means you can put the validation back here again.

When I suggested to move this into `FullNode`, it was a suggestion in case we didn't have time for #947.

source/agora/consensus/validation/Block.d
```diff
@@ -113,6 +119,282 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    return null;
+}
+
+/// Genesis block validation fail test
+unittest
+{
+    import agora.common.Serializer;
+
+    Block block = GenesisBlock.serializeFull.deserializeFull!Block;
+    assert(block.testIsValidGenesisBlock());
```
AndrejMitrovic: Also add:

```
assert(!block.isValid(0, 0, null));  // don't accept block height 0 from the network
```

source/agora/consensus/validation/Block.d
```diff
@@ -55,6 +60,8 @@ version (unittest)
     it as a spent output. See the `findNonSpent` function in the
     unittest for an example.
 
+    As a special case, the genesis block follows its own validation logic.
```
AndrejMitrovic: This comment is outdated. It should say genesis is rejected in this function, and only accepted in the other function that was introduced.

source/agora/consensus/Genesis.d
TrustHenry: It causes the same problem again.
It is being used in an integrated test.

source/agora/consensus/validation/Block.d
```diff
@@ -113,6 +119,282 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    if (block.header.height != 0)
+        return "GenesisBlock: The height of the block is not 0";
+
+    if (block.header.prev_block != Hash.init)
+        return "GenesisBlock: Header.prev_block is not empty";
+
+    if (block.txs.length == 0)
+        return "GenesisBlock: Transaction(s) are empty";
+
+    if (!block.txs.isSorted())
```
AndrejMitrovic: I asked to move this below the length check. But it seems my comment is gone?

source/agora/consensus/validation/Block.d
```diff
@@ -113,6 +119,282 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+            utxo_set[UTXOSetValue.getHash(tx_hash, idx)] = utxo_value;
+        }
+    }
+
+    Hash[] merkle_tree;
+    if (block.header.merkle_root !=
+        Block.buildMerkleTree(block.txs, merkle_tree))
+        return "GenesisBlock: Merkle root does not match header's";
+
+    // If there are no enrollments, return them here early.
```
AndrejMitrovic, line 199: This conflicts with https://github.com/bpfkorea/agora/pull/875. Depending on which PR gets merged first, this code will have to be fixed somewhere.

source/agora/consensus/Genesis.d
AndrejMitrovic: That's why I asked you if your PR fixes the issue..

source/agora/consensus/validation/Block.d
TrustHenry: It's good test!

source/agora/consensus/validation/Block.d
TrustHenry: Could you give me your opinion one more time?

AndrejMitrovic: https://github.com/bpfkorea/agora/pull/884#discussion_r435957214

You check the length, then the is sorted call, then the length again. The sorted call should be after all the length calls.

source/agora/consensus/validation/Block.d
```diff
@@ -113,6 +119,282 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    Params:
+        block = The genesis block to check
+
+    Returns:
+        `null` if the genesis block is valid, otherwise a string explaining
+        the reason it is invalid.
+
+*******************************************************************************/
+
+public string isValidGenesisBlock (immutable Block* block) nothrow @safe
```
AndrejMitrovic: The naming convention does not match. All other string returning validation function have `isInvalid` in their name, not `isValid`.

So maybe it should be named `isGenesisBlockInvalidReason`

AndrejMitrovic: This should take `const ref`, not `immutable *`

AndrejMitrovic: Then you wouldn't need the `serializeFull.deserialize` workaround in `testIsValidGenesisBlock`.

AndrejMitrovic: Alternatively just take it by value if it's only used in one place.

AndrejMitrovic:
This diff works fine for me, no linker errors:

```diff
diff --git a/source/agora/consensus/EnrollmentManager.d b/source/agora/consensus/EnrollmentManager.d
index ae16fa12c..8aae7b8f0 100644
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -55,7 +55,6 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import agora.consensus.EnrollmentPool;
 import agora.consensus.PreImage;
 import agora.consensus.validation;
diff --git a/source/agora/consensus/EnrollmentPool.d b/source/agora/consensus/EnrollmentPool.d
index e6301b149..a295af496 100644
--- a/source/agora/consensus/EnrollmentPool.d
+++ b/source/agora/consensus/EnrollmentPool.d
@@ -25,7 +25,6 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import agora.consensus.validation;
 import agora.utils.Log;
 
diff --git a/source/agora/consensus/Genesis.d b/source/agora/consensus/Genesis.d
index 7a3cce4b0..251ededde 100644
--- a/source/agora/consensus/Genesis.d
+++ b/source/agora/consensus/Genesis.d
@@ -18,6 +18,7 @@ import agora.common.Hash;
 import agora.common.crypto.Key;
 import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
+import agora.consensus.validation.Block;
 
 /// Contains a pointer to the genesis block.
 version (unittest)
@@ -54,6 +55,9 @@ public void setGenesisBlock (immutable Block* block)
 {
     assert(block !is null);
     gen_block = block;
+
+    if (auto reason = isValidGenesisBlock(*gen_block))
+        throw new Exception(reason);
 }
 
 /*******************************************************************************
diff --git a/source/agora/consensus/Quorum.d b/source/agora/consensus/Quorum.d
index c1e446e5d..dfb6270c7 100644
--- a/source/agora/consensus/Quorum.d
+++ b/source/agora/consensus/Quorum.d
@@ -19,7 +19,6 @@ import agora.common.Set;
 import agora.common.Types;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 
 import std.algorithm;
 import std.array;
diff --git a/source/agora/consensus/UTXOSet.d b/source/agora/consensus/UTXOSet.d
index ab8999171..0e9b484d9 100644
--- a/source/agora/consensus/UTXOSet.d
+++ b/source/agora/consensus/UTXOSet.d
@@ -388,7 +388,6 @@ unittest
     import agora.common.Amount;
     import agora.consensus.data.Transaction;
     import agora.consensus.data.UTXOSetValue;
-    import agora.consensus.UTXOSet;
 
     KeyPair[] key_pairs = [KeyPair.random, KeyPair.random];
 
diff --git a/source/agora/consensus/ValidatorSet.d b/source/agora/consensus/ValidatorSet.d
index 5808c06a7..1bbda6592 100644
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -24,7 +24,6 @@ import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import agora.consensus.PreImage;
 import agora.consensus.validation;
 import agora.utils.Log;
diff --git a/source/agora/consensus/validation/Block.d b/source/agora/consensus/validation/Block.d
index 5cf085979..d2111af42 100644
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -20,7 +20,6 @@ import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import VEn = agora.consensus.validation.Enrollment;
 import VTx = agora.consensus.validation.Transaction;
 import agora.consensus.Genesis;
@@ -144,7 +143,7 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
 
 *******************************************************************************/
 
-public string isValidGenesisBlock (immutable Block* block) nothrow @safe
+public string isValidGenesisBlock (const ref Block block) nothrow @safe
 {
     if (block.header.height != 0)
         return "GenesisBlock: The height of the block is not 0";
@@ -389,10 +388,7 @@ public bool testIsValidGenesisBlock (const ref Block genesis_block)
     @trusted
 {
     import agora.common.Serializer;
-
-    immutable Block block =
-        genesis_block.serializeFull.deserializeFull!(immutable(Block));
-    return isValidGenesisBlock(&block) is null;
+    return isValidGenesisBlock(genesis_block) is null;
 }
 
 /// Ditto but returns `bool`, only usable in unittests
@@ -573,7 +569,7 @@ unittest
     import agora.common.Serializer;
     import agora.consensus.data.Enrollment;
     import agora.consensus.data.Transaction;
-    import agora.consensus.UTXOSet;
+    import agora.consensus.data.UTXOSetValue;
 
     import std.algorithm;
     import std.range;
diff --git a/source/agora/consensus/validation/Enrollment.d b/source/agora/consensus/validation/Enrollment.d
index 091574419..efed2c61f 100644
--- a/source/agora/consensus/validation/Enrollment.d
+++ b/source/agora/consensus/validation/Enrollment.d
@@ -20,7 +20,6 @@ import agora.common.crypto.Key;
 import agora.common.crypto.Schnorr;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 
 import std.conv;
 
diff --git a/source/agora/consensus/validation/Transaction.d b/source/agora/consensus/validation/Transaction.d
index 6f44c0edd..9d021334c 100644
--- a/source/agora/consensus/validation/Transaction.d
+++ b/source/agora/consensus/validation/Transaction.d
@@ -17,7 +17,6 @@ import agora.common.Amount;
 import agora.common.Hash;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 
 version (unittest)
 {
diff --git a/source/agora/node/FullNode.d b/source/agora/node/FullNode.d
index a3aaee40d..884d6a24d 100644
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -129,13 +129,6 @@ public class FullNode : API
             auto block_bytes = config.node.genesis_block.chunks(2).map!(
                 twoDigits => twoDigits.parse!ubyte(16)).array();
             this.genesis_block = block_bytes.deserializeFull!(immutable(Block));
-
-            // genesis block validation
-            if (auto reason = isValidGenesisBlock(&this.genesis_block))
-            {
-                log.error(reason);
-                abort();
-            }
             setGenesisBlock(&this.genesis_block);
         }
 
diff --git a/source/agora/node/Validator.d b/source/agora/node/Validator.d
index 82130b783..d7c6b5cbc 100644
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -26,7 +26,6 @@ import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.Transaction;
 import agora.consensus.EnrollmentManager;
 import agora.consensus.protocol.Nominator;
-import agora.consensus.UTXOSet;
 import agora.consensus.Quorum;
 import agora.network.NetworkManager;
 import agora.node.BlockStorage;
diff --git a/source/agora/test/Ledger.d b/source/agora/test/Ledger.d
index 870dcc528..caf07310f 100644
--- a/source/agora/test/Ledger.d
+++ b/source/agora/test/Ledger.d
@@ -24,7 +24,6 @@ import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
-import agora.consensus.UTXOSet;
 import agora.consensus.Genesis;
 import agora.consensus.validation;
 import agora.test.Base;
```

The issue with the previous PR is that not all of the imports were removed.

source/agora/consensus/validation/Block.d
TrustHenry: https://github.com/bpfkorea/agora/blob/72542f5cc9e0c624176f3b809670542ce4abbc87/source/agora/consensus/Genesis.d#L23-L26
Used as declared.
If I want to cast as `immutable` to `const` shouldn't I have to do `GenesisBlock.serializeFull.deserializeFull!Block`?

AndrejMitrovic: You don't have to cast anything. Just assign the genesis block, and then pass `*gen_block` to the validation function

TrustHenry: I renamed `isGenesisBlockInvalidReason`

source/agora/consensus/validation/Block.d
TrustHenry: I apply the suggestion.

source/agora/consensus/Genesis.d
TrustHenry: I did a test with the existing code.
I have added a new commit to remove unnecessary import.

TrustHenry:
All suggestions have been applied.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    assert(!block.isGenesisBlockInvalid());
+}
+
+/// Ditto but returns `bool`, only usable in unittests
+/// Only the genesis block Validation
+version (unittest)
+public bool isGenesisBlockInvalid (const ref Block genesis_block)
+    nothrow @safe
+{
+    return isGenesisBlockInvalidReason(genesis_block) is null;
```
AndrejMitrovic, line 380: This should either be `!is null` or renamed to `isGensisBlockValid`. Look at your call below:

```D
assert(GenesisBlock.isGenesisBlockInvalid());
```

It's asserting that the default genesis block is invalid.

TrustHenry: The meaning of the test function was the opposite.
It's a good point.

I changed its name to `isGenesisBlockValid`
I think there is a typo in the name you recommended :)

TrustHenry:
Renamed second commit Title.

AndrejMitrovic (approved):
Very good work.

AndrejMitrovic:
@Geod24 can you also take a 👀?

source/agora/consensus/Genesis.d
```diff
@@ -102,7 +102,7 @@ private immutable Hash UnitTestGenesisMerkleRoot =
          `3dc369ef2a44b62ba113814a9d819a276ff61582874c9aee9c98efa2aa1f10d73`);
 
 /// The single transaction that are part of the genesis block
-private immutable Transaction UnitTestGenesisTransaction =
+public immutable Transaction UnitTestGenesisTransaction =
```
Geod24: That reverts a change that was made recently.

source/agora/consensus/validation/Block.d
```diff
@@ -73,8 +81,7 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
 
     // special case for the genesis block
     if (block.header.height == 0)
-        return block == GenesisBlock ?
-            null : "Block: Height 0 but not Genesis block";
+        return "Block: Rejected Genesis block";
```
Geod24: Do we really need this then ? The condition at L88 takes care of this (since `prev_height` can be 0 at least)

source/agora/consensus/validation/Block.d
```diff
@@ -54,6 +59,9 @@ version (unittest)
     it as a spent output. See the `findNonSpent` function in the
     unittest for an example.
 
+    As a special case, the genesis is rejected in this function,
+    and only accepted in the `isGenesisBlockInvalidReason` function
```
Geod24:
```suggestion
    As a special case, the genesis block is rejected by this function.
    Validation of a genesis block should be done through the
    `isGenesisBlockInvalidReason` function.
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
     return null;
 }
 
+/*******************************************************************************
+
+    Check the validity of a genesis block
+
+    Follow the Block rules except for the following
```
Geod24:
```suggestion
    Follow the same rules as for `Block` except for the following:
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+        - Block height is 0
```
Geod24:
```suggestion
        - Block height must be 0
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+        - The previous block hash of the block must be empty
+        - The number of transactions in the block is above zero and less than
+          or equal to Block.TxsInBlock
```
Geod24:
```suggestion
        - The number of transactions in the block must be in the range `(0; Block.TxsInBlock]`.
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+        - All the transactions pass validation, which implies:
+            - Transaction type in block should only be Payment or Freeze
+            - The inputs of the transaction(s) are empty
+            - There is one or more outputs of the transaction
```
Geod24:
```suggestion
        - Transactions must have no input
        - Transactions must have at least one output
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+        - All the enrollments pass validation, which implies:
+            - The enrollments refer to freeze tx's in this block
```
Geod24, line 131:
```suggestion
        - All the enrollments pass validation, which implies:
            - The enrollments refer to freeze tx's in this block
            - The signature for the Enrollment is valid
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    UTXOSetValue[Hash] utxo_set;
+    foreach (const ref tx; block.txs)
+    {
+        if (!(tx.type == TxType.Payment || tx.type == TxType.Freeze))
+            return "GenesisBlock: Transaction type allows only "
+                ~ "Payment and Freeze";
+
+        if (tx.inputs.length != 0)
+             return "GenesisBlock: Not empty input(s) in the transaction";
```
Geod24:
```suggestion
             return "GenesisBlock: Transactions must not have input";
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    if (!block.txs.isSorted())
+        return "GenesisBlock: Transactions are not sorted";
+
+    UTXOSetValue[Hash] utxo_set;
+    foreach (const ref tx; block.txs)
+    {
+        if (!(tx.type == TxType.Payment || tx.type == TxType.Freeze))
+            return "GenesisBlock: Transaction type allows only "
+                ~ "Payment and Freeze";
```
Geod24:
```suggestion
            return "GenesisBlock: Invalid enum value for type field";
```

Since we only have two possible values.

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+    Set!Hash used_utxos;
+    bool findUTXO (Hash hash, size_t index, out UTXOSetValue value)
+        nothrow @safe
+    {
+        Hash utxo_hash;
+
+        if (index == size_t.max)
+            utxo_hash = hash;
+        else
+            utxo_hash = UTXOSetValue.getHash(hash, index);
```
Geod24: I think we only need the `size_t.max` case here, no ?

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+{
+    import agora.common.Serializer;
+
+    Block block = GenesisBlock.serializeFull.deserializeFull!Block;
+    assert(block.isGenesisBlockValid());
+
+    // don't accept block height 0 from the network
+    assert(!block.isValid(0, Hash.init, null));
+
+    KeyPair keypair;
```
Geod24: Remove and use `KeyPair` instead (`random` is `static`).

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+
+    // .prev_block check
+    block.header.prev_block = block.header.hashFull();
+    assert(!block.isGenesisBlockValid());
+
+    block.header.prev_block = Hash.init;
+    assert(block.isGenesisBlockValid());
+
+    Transaction tx =
+        UnitTestGenesisTransaction.serializeFull.deserializeFull!Transaction;
```
Geod24:
```suggestion
    Transaction[] txs = GenesisBlock.txs.serializeFull.deserializeFull!(Transaction[]);
```

source/agora/consensus/validation/Block.d
```diff
@@ -112,6 +119,281 @@ public string isInvalidReason (const ref Block block, in ulong prev_height,
...
+        buildMerkleTree(block);
+        assert(!block.isGenesisBlockValid());
+
+        block.txs[0] = tx;
+        buildMerkleTree(block);
+        assert(block.isGenesisBlockValid());
+
+        // Output not empty check
+        Output[] outputs;
+        block.txs[0].outputs = outputs;
```
Geod24:
```suggestion
        block.txs[0].outputs = null;
```

Geod24 (approved):
Just a few nits, otherwise LGTM. Good job!

source/agora/consensus/validation/Block.d
AndrejMitrovic: With this change you can go back to making `UnitTestGenesisTransaction` private again.

source/agora/consensus/Genesis.d
TrustHenry: I revert it and applied the suggestions below.

source/agora/consensus/validation/Block.d
TrustHenry: I checked the logics below.

source/agora/consensus/validation/Block.d
TrustHenry: That's right.

source/agora/consensus/validation/Block.d
TrustHenry: That's right.

TrustHenry:
@Geod24 @AndrejMitrovic Thank you for your reviews.
All suggestions have been applied.

Geod24 (approved):

linked0 (approved):
Nice work.

PR #885, opened 2020-05-28
[trivial] Privatize everything in NetworkManager

AndrejMitrovic:


Geod24:
Why ? We did the exact opposite change in the past.

Geod24:
Namely: https://github.com/bpfkorea/agora/commit/dfc892a4c6a68a1465012788e4a580b77acfde5d

AndrejMitrovic:
Ah but why did you expose this? Did we need to use any of these fields?

I didn't remember the change so I thought it was just a mistake.

PR #886, opened 2020-05-29
[WIP] Uses the enrollments contained in the genesis block to test `pre-images revelation`

MichaelKim20:
I previously made a PR to add the validator's enrollment to Genesis, and it's merged.
Since then, there have been occasional errors in the unit test.
I have identified and corrected the cause of the unit test error.
I hope this will be approved as soon as possible and our work will proceed smoothly.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -228,6 +228,9 @@ public struct Serializer
 /// A different default serializer from `LocalRest` for `RemoteAPI`
 public alias RemoteAPI (APIType) = geod24.LocalRest.RemoteAPI!(APIType, Serializer);
 
+/// The number of validators in the genesis block
+public static immutable uint ValidateCountInGenesis = 2;
+
 /*******************************************************************************
 
     Task manager backed by LocalRest's event loop.
@@ -937,22 +940,32 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
     NodeConfig[] node_configs;
     Config[] configs;
 
+    void addAdditionalValidator()
+    {
+        auto validator_count = node_configs.filter!(conf => conf.is_validator).array.length;
+        foreach (idx; validator_count .. ValidateCountInGenesis)
+            node_configs ~= makeNodeConfig(true);
+    }
+
     final switch (test_conf.topology)
     {
     case NetworkTopology.Simple:
         node_configs = iota(test_conf.nodes).map!(_ => makeNodeConfig(true)).array;
+        addAdditionalValidator();
         configs = iota(test_conf.nodes)
             .map!(idx => makeConfig(node_configs[idx], node_configs)).array;
         break;
 
     case NetworkTopology.FindNetwork:
         node_configs = iota(test_conf.nodes).map!(_ => makeNodeConfig(false)).array;
+        addAdditionalValidator();
         configs = iota(test_conf.nodes)
             .map!(idx => makeMinimalNetwork(idx, node_configs[idx], node_configs)).array;
         break;
 
     case NetworkTopology.FindQuorums:
         node_configs = iota(test_conf.nodes).map!(_ => makeNodeConfig(true)).array;
+        addAdditionalValidator();
         configs = iota(test_conf.nodes)
             .map!(idx => makeMinimalNetQuorum(idx, node_configs[idx], node_configs)).array;
         break;
@@ -960,12 +973,14 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
     case NetworkTopology.OneNonValidator:
         node_configs ~= iota(test_conf.nodes - 1).map!(_ => makeNodeConfig(true)).array;
         node_configs ~= makeNodeConfig(false);
+        addAdditionalValidator();
         configs = iota(test_conf.nodes)
             .map!(idx => makeConfig(node_configs[idx], node_configs)).array;
         break;
 
     case NetworkTopology.OneValidator:
         node_configs ~= makeNodeConfig(true);
+        addAdditionalValidator();
         node_configs ~= iota(test_conf.nodes - 1).map!(_ => makeNodeConfig(false)).array;
         configs = iota(test_conf.nodes)
             .map!(idx => makeConfig(node_configs[idx], node_configs)).array;
@@ -978,6 +993,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
 
         // add one non-validator outside the network
         node_configs ~= makeNodeConfig(false);
+        addAdditionalValidator();
         configs ~= makeConfig(node_configs[$ - 1], node_configs);
         break;
     }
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
It's not really the correct fix. It's just a workaround.

Because with this PR we no longer have networking tests for the `enrollValidator` API endpoint.

But if it's blocking work, we can temporarily accept it. My tests in #845 will include re-enrolling, so that endpoint will be tested again.

Geod24:
> Since then, there have been occasional errors in the unit test.

Could you post a link to the failure ?

MichaelKim20:
> It's not really the correct fix. It's just a workaround.
> 
> Because with this PR we no longer have networking tests for the `enrollValidator` API endpoint.
> 
> But if it's blocking work, we can temporarily accept it. My tests in #845 will include re-enrolling, so that endpoint will be tested again.

Yes, that's a very good point.
All of the validators' enrollment data are included in the Genesis block.
We cannot change a non-validator to a validator in a unit test.
So I didn't call the enrollValidator.

Therefore, in order to register further, we will need to allow re-enrollment 
This is a departure from my issue.

I'll think about a better way.

MichaelKim20:
> > Since then, there have been occasional errors in the unit test.
> 
> Could you post a link to the failure ?

@Geod24 @AndrejMitrovic I found errors in the unit test and recorded them here.
https://github.com/bpfkorea/agora/pull/887/checks?check_run_id=718978489
https://github.com/bpfkorea/agora/pull/887/checks?check_run_id=718978503
https://travis-ci.com/github/bpfkorea/agora/jobs/340796300

AndrejMitrovic:
Ok, but this means the test is causing the Node to enter into an unexpected state.

> Full error: core.exception.AssertError@source/agora/consensus/protocol/Nominator.d(372): Assertion failure

This assertion should never happen. It means there is a bug. We shouldn't just hide the bug, but we should find the root cause of the issue and fix that.

MichaelKim20:
@Geod24 @AndrejMitrovic The genesis block contains the enrollment data of all validators.
Additional enrollment is not allowed before maturity.
I created a new network topology. `NetworkTopology.AllValidatorOneNotEnroll`
All nodes are validators and only the last one node doesn't have enrollment data in the genesis block.

source/agora/common/Config.d
```diff
@@ -103,6 +103,9 @@ public struct NodeConfig
     /// Is this a validator node
     public bool is_validator;
 
+    /// There is enrollment data in the genesis block
+    public bool has_enroll = true;
```
AndrejMitrovic: This does not belong here. This is the Node's public configuration. Why is there a workaround for a unittest in the Node configuration?

MichaelKim20: Okay

Geod24 (changes requested):
As @AndrejMitrovic pointed out using a config option is not going to be good.
I would recommend the following instead:
1) Have a fixed number of Enrollments in the Genesis block;
2) If the test configuration has more validators than the number of Enrollments in the Genesis blocks, those Validators should be created.
3) When extra validators are created, they will end up Enrolling themselves;

Does this make sense ?
Note that we still need to provide them with a frozen UTXO, or build a test node that creates one frozen UTXO.

MichaelKim20:
> Note that we still need to provide them with a frozen UTXO, or build a test node that creates one frozen UTXO.

I changed it to the way you suggested.
Most unit tests required the enrollment of a validator.
There were some tricky problems.
1. Information on which node is the validator is unclear.
2. Transactions from genesis block are consumed, and for further testing, a test transaction must be prepared. These may cause confusion in other tests.
Do you mind if we continue this way?

source/agora/test/Base.d
```diff
@@ -228,6 +228,9 @@ public struct Serializer
 /// A different default serializer from `LocalRest` for `RemoteAPI`
 public alias RemoteAPI (APIType) = geod24.LocalRest.RemoteAPI!(APIType, Serializer);
 
+/// The number of validators in the genesis block
+public static immutable uint ValidateCountInGenesis = 2;
```
Geod24, line 232: Where does this come from ? And why 2 instead of 1 ?

source/agora/test/Base.d
```diff
@@ -937,35 +940,47 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
     NodeConfig[] node_configs;
     Config[] configs;
 
+    void addAdditionalValidator()
+    {
+        auto validator_count = node_configs.filter!(conf => conf.is_validator).array.length;
+        foreach (idx; validator_count .. ValidateCountInGenesis)
+            node_configs ~= makeNodeConfig(true);
+    }
+
```
Geod24, line 949: I'm not sure I follow... Doesn't this add extra validators to the topology ? So if the user specifies e.g. 4 validators, he'll get 4 validators + whatever are in the genesis block ?

source/agora/test/EnrollmentManager.d
```diff
@@ -26,48 +26,80 @@ import agora.consensus.EnrollmentManager;
...
-    auto node_2 = nodes[1];
-
-    // make transactions which have UTXOs for the node.
     auto gen_key_pair = getGenesisKeyPair();
-    auto pubkey_1 = node_1.getPublicKey();
 
     Transaction[] txs;
-    foreach (idx; 0 .. Block.TxsInBlock)
+    int tx_idx = 0;
+    foreach (idx; 0 .. not_enrolled_validators.length)
```
Geod24, line 41: Style note, this would be much better expressed as:
```suggestion
    foreach (idx, not_enrolled; not_enrolled_validators)
```

But remember: Double negatives are bad. Instead of "!not_enrolled", we should make this `enrolled`.
But that is just for an array of `bool`. In practice, a better way to do this would be to pass a range that has been `filter`ed out of the values we want to exclude.

Geod24 (changes requested):
It doesn't look like this is re-usable for other tests, is it ? Will new tests get a valid Genesis ?

source/agora/test/Base.d
MichaelKim20: It was recorded as 2 in Issue 822, and in the article above you only said it was a fixed value, but you didn't specify the exact value.

Geod24: It was mentioned as a possibility, but it's up  to the person implementing the feature what makes sense and to explain why. I'll answer on the issue itself.

source/agora/test/Base.d
MichaelKim20: https://github.com/bpfkorea/agora/pull/886#pullrequestreview-420698403
I am implementing what you have proposed, but I think I understand it differently.
I understood like this, can you check if it's right?
1. There is the enrollment of a fixed number of validators on the genesis block.
2. If the number of validators in the test configuration is greater than the fixed number, create additional validators.
3. When additional validators are created, they will eventually register themselves.

MichaelKim20: And I only worked on 'agora.test.EnrollmentManager' and I haven't finished another unit test yet.
I wanted to know if what I've done so far is the same as what you think.

MichaelKim20:
What this PR was trying to resolve has already been resolved.
If you have any other opinions or issues, please proceed with new issues or related PR.

MichaelKim20:
I'm working on what @Geod24 suggested.
I need to check if this is related to my current issue or if it is something to proceed with a new one.
In order to solve this problem, I think @Geod24 and @AndrejMitrovic need to agree.
Therefore, I think it would be better to proceed with this part if you create a new issue.

PR #889, opened 2020-05-29
Rework NetworkManager using tasks

### LINKED ISSUE #587: Network manager needs to be reworked

AndrejMitrovic:
Works for me locally, let's see what the CI says.

There might be some missing `scope` / `@safe` and so on, I'll add those if neccessary.

Fixes https://github.com/bpfkorea/agora/issues/587

### DIFF
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -385,9 +385,10 @@ public class NetworkManager
         {
             try
             {
+                PublicKey key;
                 try
                 {
-                    node.getPublicKey();
+                    key = node.getPublicKey();
                 }
                 catch (HTTPStatusException ex)  // vibe.d (non-unittests)
                 {
@@ -396,17 +397,17 @@ public class NetworkManager
                         throw ex;
                 }
 
-                const is_validator = node.key != PublicKey.init;
+                const is_validator = key != PublicKey.init;
                 this.connecting_addresses.remove(node.address);
 
                 if (is_validator)
-                    this.required_peer_keys.remove(node.key);
+                    this.required_peer_keys.remove(key);
 
                 if (this.peerLimitReached())
                     return;
 
                 if (is_validator)
-                    log.info("Established new connection with validator: {} (key: {})", node.address, node.key);
+                    log.info("Established new connection with validator: {} (key: {})", node.address, key);
                 else
                     log.info("Established new connection with full node: {}", node.address);
 
@@ -416,7 +417,7 @@ public class NetworkManager
 
                 this.peers[node.address] = node;
                 if (is_validator)  // Add to Validator node map
-                    this.validator_to_addr[node.key] = node.address;
+                    this.validator_to_addr[key] = node.address;
 
                 this.metadata.peers.put(node.address);
                 break;
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -57,11 +57,252 @@ mixin AddLogger!();
 /// Ditto
 public class NetworkManager
 {
+    /// Node information
+    private static struct NodeConnInfo
+    {
+        /// Address of the node
+        Address address;
+
+        /// Is this node a Validator
+        bool is_validator;
+
+        /// Public key of the node, only set if the node is a Validator
+        PublicKey key;
+
+        /// Client
+        NetworkClient client;
+    }
+
+    /***************************************************************************
+
+        Establishes a connection with a Node address in a new task.
+
+    ***************************************************************************/
+
+    private class ConnectionTask
+    {
+        /// Address to connect to
+        private const Address address;
+
+        /// Called when we've connected and determined if this is
+        /// a FullNode / Validator
+        private void delegate (scope ref NodeConnInfo) onHandshakeComplete;
+
+        /// Called when a request to a node fails.
+        /// The delegate should return true if we should continue trying
+        /// to send requests to this node, or false (e.g. if the node is banned)
+        private bool delegate (in Address) onFailedRequest;
+
+
+        /***********************************************************************
+
+            Ctor
+
+            Params:
+                address = the adddress of the node
+                onHandshakeComplete = called when we've successfully connected
+                                      and determined the type of the Node
+                onFailedRequest = called when a request fails, returns true
+                                  if we should keep trying to send requests
+
+        ***********************************************************************/
+
+        public this (in Address address,
+            void delegate (scope ref NodeConnInfo node) onHandshakeComplete,
+            bool delegate (in Address address) onFailedRequest)
+        {
+            this.address = address;
+            this.onHandshakeComplete = onHandshakeComplete;
+            this.onFailedRequest = onFailedRequest;
+            this.outer.taskman.runTask(&this.connect);
+        }
+
+        /***********************************************************************
+
+            Repeatedly attempt connecting to an address, and try to determine
+            if this is a FullNode or a Validator.
+
+            After each connection failure / request failure the ban manager is
+            queried, and if this address was banned then 'onFailedRequest()'
+            will be called and the task will be killed.
+
+            If we've successfully connected and determined the Node type,
+            'onHandshakeComplete' is called and the task is killed.
+
+        ***********************************************************************/
+
+        private void connect ()
+        {
+            auto client = new NetworkClient(this.outer.taskman,
+                this.outer.banman, this.address,
+                this.outer.getClient(this.address,
+                    this.outer.node_config.timeout.msecs),
+                this.outer.node_config.retry_delay.msecs,
+                this.outer.node_config.max_retries);
+
+            PublicKey key;
+            while (1)
+            {
+                try
+                {
+                    // LocalRest will return PublicKey.init,
+                    // vibe.d will throw HTTPStatusException with status == 404
+                    key = client.getPublicKey();
+                    break;
+                }
+                catch (HTTPStatusException ex)
+                {
+                    // 404 => API not implemented, it's a FullNode
+                    if (ex.status == 404)
+                        break;
+
+                    if (!this.onFailedRequest(this.address))
+                        return;
+
+                    // else try again
+                    this.outer.taskman.wait(
+                        this.outer.node_config.retry_delay.msecs);
+                }
+            }
+
+            const is_validator = key != PublicKey.init;
+            if (is_validator)
+                log.info("Found new Validator: {} (key: {})", address, key);
+            else
+                log.info("Found new FullNode: {}", address);
+
+            NodeConnInfo node = {
+                address : this.address,
+                is_validator : is_validator,
+                key : key,
+                client : client
+            };
+
+            this.onHandshakeComplete(node);
+        }
+    }
+
+    /***************************************************************************
+
+        Queries a Node's getNodeInfo() API endpoint to discover new sets
+        of addresses we may want to connect to.
+
+        The provided delegate will be called each time a Node's known
+        addresses were retrieved via the getNodeInfo() API endpoint.
+
+        Note that this is a never-ending task. A node's NodeInfo may signal
+        that it's complete, however the list of its peers may grow as new
+        incoming connections are established as well as when any quorum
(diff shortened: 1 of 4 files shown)


source/agora/common/Config.d
```diff
@@ -110,6 +110,9 @@ public struct NodeConfig
     /// Maximum number of listeners to connect to
     public size_t max_listeners = 10;
 
+    /// The maximum number of connection tasks to use during network discovery
+    public size_t connection_tasks = 10;
+
```
Geod24: It's a bit eerie to expose this in a config file that might be touched by not-so-technical ppl.

source/agora/node/FullNode.d
```diff
@@ -146,14 +146,38 @@ public class FullNode : API
...
         this.taskman.runTask(
+        ()
         {
-            log.info("Doing network discovery..");
-            this.network.discover();
-            this.network.startPeriodicCatchup(this.ledger);
+            while (1)
+            {
+                scope (success)
+                    this.taskman.wait(5.seconds);
```
Geod24: Why the `scope (success)` ?

source/agora/node/Validator.d
```diff
@@ -71,21 +71,42 @@ public class Validator : FullNode, API
...
+    /***************************************************************************
+
+        Starts the periodic network discovery task.
+
+    ***************************************************************************/
+
+    private void startPeriodicDiscovery ()
     {
         this.taskman.runTask(
+        ()
```
Geod24, line 96: Fun fact: This is not required, but it also causes an ambiguity in the parser which needs to be resolved (using heuristics and lookahead).

source/agora/node/Validator.d
```diff
@@ -71,21 +71,42 @@ public class Validator : FullNode, API
...
         {
             // build the list of required quorum peers to connect to
             Set!PublicKey required_peer_keys;
             buildRequiredKeys(this.config.node.key_pair.address,
                 this.config.quorum, required_peer_keys);
 
-            log.info("Doing network discovery..");
```
Geod24, line 84: With this message gone, do we have an indication of when a node is "ready" in the logs ?

source/agora/network/NetworkManager.d
```diff
@@ -57,31 +57,288 @@ mixin AddLogger!();
 /// Ditto
 public class NetworkManager
 {
+    /// Node information
+    private static struct Node
```
Geod24: Should this be something less ambiguous, like `NodeInfo` ?

source/agora/network/NetworkManager.d
```diff
@@ -57,31 +57,288 @@ mixin AddLogger!();
...
+
+            PublicKey key;
+            while (1)
+            {
+                try
+                {
+                    key = client.getPublicKey();
+                    break;
+                }
+                catch (HTTPStatusException ex)  // vibe.d (non-unittests)
```
Geod24: What happens in LocalRest ?

source/agora/network/NetworkManager.d
```diff
@@ -57,31 +57,288 @@ mixin AddLogger!();
...
+
+            Ctor
+
+            Params:
+                onNewAddresses = called when a set of addresses were retrieved
+                                 from a node
+
+        ***********************************************************************/
+
+        public this (void delegate (Set!Address addresses) onNewAddresses )
```
Geod24:
```suggestion
        public this (void delegate (Set!Address addresses) onNewAddresses)
```

source/agora/network/NetworkManager.d
```diff
@@ -57,31 +57,288 @@ mixin AddLogger!();
...
+            clients list.
+
+        ***********************************************************************/
+
+        private void getNewAddresses ()
+        {
+            while (1)
+            {
+                scope (success)
+                    this.outer.taskman.wait(1.seconds);
```
Geod24, line 273: Ditto why scope success ?

AndrejMitrovic:
Oh and `taskManager.schedule` will be very useful to have once we have that. It should make it more efficient. But it's not a strict requirement.

Geod24 (commented):
That's a big diff

source/agora/common/Config.d
AndrejMitrovic: Should we just make it a constant then? I'm not sure which number to use. But I guess we can start with a low number like 10 and then increase it as we see fit?

source/agora/node/FullNode.d
AndrejMitrovic: Because I don't want to suspend the fiber if an exception is thrown. I want it to kill the fiber instead.

And now that I said that out loud, I realize I should probably make `discover` nothrow.

source/agora/node/Validator.d
AndrejMitrovic: Oh right. Maybe it's muscle memory? I think `{ }` used to be interpreted as a struct initializer.

source/agora/node/Validator.d
AndrejMitrovic: It was moved inside the `discover()` method instead. The reasoning being, it will be run multiple times (once quorum reshuffling is implemented).

source/agora/network/NetworkManager.d
AndrejMitrovic: I had that exact same thought. But then I realized we already have that struct, returned in `getNodeInfo`. I need to think of another name..

source/agora/network/NetworkManager.d
AndrejMitrovic: it returns `PublicKey.init`. This was implemented this way as a workaround when splitting up full node vs validator node, Henry's PR.

source/agora/node/Validator.d
Geod24: It only happens if you forgot the semicolon IIRC. But I don't mind the explicitness.

AndrejMitrovic:
- Fixed styling
- Added comment about the special-case in LocalRest
- scope (success) was used because if the Fiber shuts down, the sleep call is unnecessary
- Renamed Node => NodeConnInfo (didn't call it NodeInfo to avoid clashes with the existing struct).
- Removed the config option.
- Sorry for the big diff.

source/agora/network/NetworkManager.d
```diff
@@ -57,31 +57,290 @@ mixin AddLogger!();
...
+            will be called and the task will be killed.
+
+            If we've successfully connected and determined the Node type,
+            'onHandshakeComplete' is called and the task is killed.
+
+        ***********************************************************************/
+
+        private void connect ()
+        {
+            API api;
```
Geod24: Side note: This is an issue that came up when trying to improve the REST module / implementing the TCP protocol.
It gives a pretty bad API to expose disconnections to the client by forcing the user to re-create a client object, so I ended up handling it internally and have auto-reconnect (and potential delegates for failures).

In other words: A client is tied to an address, not a connection.

source/agora/network/NetworkManager.d
```diff
@@ -57,31 +57,290 @@ mixin AddLogger!();
...
+                this.outer.node_config.retry_delay.msecs,
+                this.outer.node_config.max_retries);
+
+            PublicKey key;
+            while (1)
+            {
+                try
+                {
+                    // LocalRest will return PublicKey.init,
+                    // vibe.d will throw HTTPStatusException with status == 404
```
Geod24, line 149: We should file an issue against LocalRest

source/agora/node/FullNode.d
Geod24: Buuuuuuuttttttttt... Why not simply:
```D
while (true)
{
    this.network.discover();
    this.taskman.wait(5.seconds);
}
```

The wait is in a `scope (success)` in a loop with a single statement. Either the statement throws and the code is never executed, or the statement doesn't throw and the code is executed. What am I missing ?

source/agora/network/NetworkManager.d
```diff
@@ -57,31 +57,290 @@ mixin AddLogger!();
...
+                {
+                    // 404 => API not implemented, it's a FullNode
+                    if (ex.status == 404)
+                        break;
+
+                    if (this.outer.banman.isBanned(this.address))
+                    {
+                        this.onFailedConnection(this.address);
+                        return;
+                    }
```
Geod24: My inner developer suggests me that we might want to make `onFailedConnection` returns a boolean telling whether we should retry or abort the request, and that we *could* move the `isBanned` check in there.
I don't know how sensible this would be, and I'm not asking you to do it, just raising possibilities.

Geod24 (commented):
Nice. Definitely the right directions. I'm not sure I can give this the review it deserves in the time remaining in the sprint, but I really, really want to get this through the door.

Typo in the commit message: "does not [missing 'need' I guess] to be re-triggered"

source/agora/network/NetworkManager.d
AndrejMitrovic: It's not an issue. It's implemented this way in Agora. I should have been more specific.

source/agora/network/NetworkManager.d
AndrejMitrovic: Oh this is a good idea.

source/agora/network/NetworkManager.d
AndrejMitrovic: In this specific case it's because of the `continue` below. I didn't want to duplicate the wait call.

source/agora/network/NetworkManager.d
Geod24: Yep, but arguably crashing the node on a client mismatch is not quite sensible

source/agora/node/FullNode.d
AndrejMitrovic: Ahh right, my bad. It was copied over from the other example usage where I have a `continue`. I'll fix it.

AndrejMitrovic:
> Nice. Definitely the right directions. I'm not sure I can give this the review it deserves in the time remaining in the sprint, but I really, really want to get this through the door.

If there's anything missing (I'm sure there is), we can continue work on it. Filing another issue with the rest of the problems that have to be fixed.

source/agora/network/NetworkManager.d
AndrejMitrovic: Yeah. I really hate this part of vibe.d.

source/agora/network/NetworkManager.d
Geod24: Filed as geod24/localrest#79

AndrejMitrovic:
There are random failures due to #887 and also the occasional failure by dub not finding a file. But otherwise it's ready to go.

source/agora/network/NetworkManager.d
```diff
@@ -57,159 +57,438 @@ mixin AddLogger!();
...
+    private class ConnectionTask
+    {
+        /// Address to connect to
+        private const Address address;
+
+        /// Called when we've connected and determined if this is
+        /// a FullNode / Validator
+        private void delegate (scope ref NodeConnInfo) onHandshakeComplete;
+
+        /// Called when we could not connect.
```
Geod24: Wrong doc

AndrejMitrovic:
Fixed wrong docs.

Geod24 (approved):

Geod24:
> Full error: core.exception.AssertError@source/agora/consensus/protocol/Nominator.d(372): Assertion failure


This triggers quite a bit...

Geod24:
Rebased on v0.x.x

PR #893, opened 2020-05-29
[WIP] Hackathon test code for recovering a enrollment information

### LINKED ISSUE #789: Restore the enrollment information of a node

linked0:
The current implementation of the enrollment process doesn't have enough network tests. So, a part of the hackathon, I added a network test code for the recovery of enrollment information of a node. 

This code is a network test code for the PR #798 .

### DIFF
--- a/source/agora/test/EnrollmentManager.d
+++ b/source/agora/test/EnrollmentManager.d
@@ -95,3 +95,112 @@ unittest
         retryFor(node.getPreimage(enroll.utxo_key) != PreImageInfo.init,
             10.seconds));
 }
+
+/// Situation: A node in the network crashes after enrolling as an validator.
+///     And the node restarts before a new block containing the enrollment
+///     information is created. And the block is created, and the enrollment
+///     information of the new block is adding to the validator set.
+/// Expectation: The node restores its own enrollment data when the block
+///     containing the enrollment information arrives.
+unittest
+{
+    import agora.consensus.validation.PreImage;
+    import core.thread;
+    import std.conv;
+
+writeln("############################## start ##############################");
+    auto network = makeTestNetwork(TestConf.init);
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.start();
+    network.waitForDiscovery();
+
+    // Boilerplate
+    auto nodes = network.clients;
+    auto gen_key_pair = getGenesisKeyPair();
+    auto node_1 = nodes[0];
+    auto pubkey_1 = node_1.getPublicKey();
+
+    // Create a new block with frozen transactions so we can Enroll
+    Transaction[] txs;
+    {
+        foreach (idx; 0 .. Block.TxsInBlock)
+        {
+            auto input = Input(hashFull(GenesisTransaction), idx.to!uint);
+            Transaction tx =
+                {
+                    TxType.Freeze,
+                    [input],
+                    [Output(Amount.MinFreezeAmount, pubkey_1),
+                        Output(Amount(10000), gen_key_pair.address)]
+                };
+            auto signature = gen_key_pair.secret.sign(hashFull(tx)[]);
+            tx.inputs[0].signature = signature;
+            txs ~= tx;
+        }
+        // Send all transactions to node 1 and rely on gossip
+        txs.each!(tx => node_1.putTransaction(tx));
+        // Make sure the block is created
+        containSameBlocks(nodes, 1).retryFor(5.seconds);
+    }
+
+    // The nodes[0] register as validator
+    auto enroll = node_1.createEnrollmentData();
+    node_1.enrollValidator(enroll);
+    nodes.each!(node =>
+        retryFor(node.getEnrollment(enroll.utxo_key) == enroll, 5.seconds));
+
+    // The `enroll` hasn't been validator yet so there must not be a pre-image
+    assert(node_1.getPreimage(enroll.utxo_key) == PreImageInfo.init);
+
+    // Check if the `node_1` restores block from the chain after restarting
+    network.restart(node_1);
+    network.waitForDiscovery();
+    containSameBlocks(nodes, 1).retryFor(5.seconds);
+
+    // Now make a second block, and check what's inside
+    Transaction[] txs2;
+    {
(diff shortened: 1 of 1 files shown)


source/agora/test/EnrollmentManager.d
```diff
@@ -95,3 +95,135 @@ unittest
         retryFor(node.getPreimage(enroll.utxo_key) != PreImageInfo.init,
             10.seconds));
 }
+
+/// Situation: A node in the network crashes after enrolling as an validator.
+///     And the node restarts before a new block block containing the enrollment
+///     information is created. And the block is created, and the enrollment
+///     information of the new block is adding to the validator set.
+/// Expectation: The node restores its own enrollment data when the block
+///     containing the enrollment information arrives.
```
Geod24: Very nice!

source/agora/test/EnrollmentManager.d
```diff
@@ -95,3 +95,135 @@ unittest
...
+        txs.each!(tx => node_1.putTransaction(tx));
+        // Make sure the block is created
+        containSameBlocks(nodes, 1).retryFor(5.seconds);
+    }
+
+    // The nodes[0] register as validator
+    auto enroll = node_1.createEnrollmentData();
+    node_1.enrollValidator(enroll);
+    nodes.each!(node =>
+        retryFor(node.getEnrollment(enroll.utxo_key) == enroll, 5.seconds));
```
Geod24, line 151: I think we should send that enrollment to another node directly instead of sending it to node 1.
Then we don't have to check if the other nodes have it (just rely on gossipping).
However, I think we should render `node1` irresponsive before doing so. Restart is an single operation so we cannot stop the node, do something, start the node. Instead we should:
- filter everything / make it sleep with drop message
- Do something
- Call restart

Last but not least, a node has a cleanup procedure that it needs to go through, otherwise we get errors in the testsuite. The best way to handle this would be to add a function in `network` IMO.

source/agora/test/EnrollmentManager.d
```diff
@@ -95,3 +95,135 @@ unittest
...
+
+    // The `enroll` hasn't been validator yet so there must not be a pre-image
+    assert(node_1.getPreimage(enroll.utxo_key) == PreImageInfo.init);
+
+    // Check if the `node_1` restores block from the chain after restarting
+    node_1.restart();
+    RemoteAPI!TestAPI[] new_nodes;
```
Geod24: We don't need to create a new array, just use `nodes` as before.
That was the point of adding this feature to LocalRest directly instead of putting it here: If we needed a new client, then it would mean we'd have to handle this for *every running nodes*,  not only the main thread.

source/agora/test/EnrollmentManager.d
```diff
@@ -95,3 +95,135 @@ unittest
...
+                [input],
+                [Output(Amount(100), gen_key_pair.address)]
+            };
+
+            auto signature = gen_key_pair.secret.sign(hashFull(tx)[]);
+            tx.inputs[0].signature = signature;
+            txs2 ~= tx;
+        }
+        txs2.each!(tx => node_1.putTransaction(tx));
+        containSameBlocks(nodes, 2).retryFor(20.seconds);
```
Geod24, line 180: This doesn't quite test what we discussed. It's a good test, for sure, but what we discussed was to create a test where the validator gets its Enrollment during *catchup*. Here it gets it through regular blockchain update.

source/agora/test/EnrollmentManager.d
```diff
@@ -95,3 +95,135 @@ unittest
...
+    }
+
+    // Check if the validator information had been stored from the second block.
+    PreImageInfo preimage_1 = node_1.getPreimage(enroll.utxo_key);
+    assert(preimage_1 != PreImageInfo.init);
```
Geod24: This does not check the block though, it checks the preimage ?

source/agora/test/EnrollmentManager.d
```diff
@@ -95,3 +95,135 @@ unittest
...
+
+    // Check if the node had restored enrollment information
+    // from the last `restart` of `node_1`
+    const b2 = node_1.getBlocksFrom(2, 2)[0];
+    assert(b2.header.enrollments.length == 1);
+    assert(enroll == b2.header.enrollments[0]);
+
+    // Wait for the timer for revelation of a pre-image to elaspe
+    Thread.sleep(10.msecs);
```
Geod24: That's why we have `retryFor` ;)

source/agora/test/EnrollmentManager.d
```diff
@@ -95,3 +95,135 @@ unittest
...
+    // Check if the `node_1` restores block from the chain after restarting
+    node_1.restart();
+    RemoteAPI!TestAPI[] new_nodes;
+    new_nodes ~= node_1;
+    foreach (node; nodes)
+    {
+        writeln("getHeight of node:", node.getBlockHeight());
+        new_nodes ~= node;
+    }
+    containSameBlocks(new_nodes, 1).retryFor(5.seconds);
```
Geod24: So the CI currently fails here.

linked0: I changed the code as follows. But do you think the code will succeeded?
```
    node_1.restart();
    foreach (node; nodes)
    {
        writeln("getHeight of node: ", node.getBlockHeight());
    }
    containSameBlocks(nodes, 1).retryFor(5.seconds);
```

It fails. Is there any code I should add?

linked0:
The commit in this PR is merged into the PR #798 . So I am closing this PR.

Issue #888, opened 2020-05-29
Remove `scope (failure) assert(0)` from Nominator.valueExternalized


Geod24:
https://github.com/bpfkorea/agora/blob/35cd603ba7e4e0db798c4b61e31d5ef689041f87/source/agora/consensus/protocol/Nominator.d#L372

Since deserialization can fail, the method has legitimate reasons to error. And it's currently hiding the underlying error, see #886

AndrejMitrovic:
https://github.com/bpfkorea/agora/issues/625

PR #895, opened 2020-05-29
Change exception handling when saving the enrollment data of the validator

MichaelKim20:
**This is PR to solve issue #888.**
When Nominator.valueExternalized occurs, the agreed data is recorded in the ledger.
There were occasional node shutdowns in this process.
I tried to find the cause through the log and code.
This is the cause of it. This is because when the enrollment data is abnormal, asset (0) is executed.

**Before changes**
When abnormal data is received, 'EnrollmentManager.addValidator' returns false then the node is terminated abnormally.

After changes
Abnormal data were filtered, and true returned.
Returns false only when it fails in the process of storing normal data.

Relates #888

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -351,14 +351,39 @@ public class EnrollmentManager
     {
         this.enroll_pool.remove(enroll.utxo_key);
 
-        if (!this.validator_set.add(block_height, finder, enroll))
-            return false;
+        if (this.validator_set.canAdd(block_height, finder, enroll))
+        {
+            if (!this.validator_set.add(block_height, finder, enroll))
+                return false;
 
-        // set next height for revealing a pre-image
-        if (enroll.utxo_key == this.enroll_key)
-            this.setNextRevealHeight(block_height);
-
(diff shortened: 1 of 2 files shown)


MichaelKim20:
I still need to make some more changes.

source/agora/consensus/data/Enrollment.d
```diff
@@ -68,6 +68,14 @@ public struct Enrollment
...
 }
 
+/// Results after adding the validator's enrollment data
+public enum ValidatorAddResult : ubyte
+{
+    Failure,    //  if an error occurs,
+    Invalid,    //  if the enrollment is abnormal,
+    Success     //  if the enrollment is added normally
+}
+
```
Geod24: That's not a way we want to go. Having `Failure` and `Invalid` is quite confusing for the caller, and probably won't be handled correctly. It is also a [leaky abstraction](https://www.joelonsoftware.com/2002/11/11/the-law-of-leaky-abstractions/).

MichaelKim20: I think so too, I should change it again in a first way. it was complicated to modify the unit test.

PR #899, opened 2020-05-31
Clean up EnrollmentManager, simplify logging & fix spurious failure

Geod24:
~Still a WIP...~

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -227,10 +227,15 @@ public class Ledger
 
         this.enroll_man.clearExpiredValidators(block.header.height);
 
-        foreach (enrollment; block.header.enrollments)
-            if (!this.enroll_man.addValidator(enrollment,
+        foreach (idx, ref enrollment; block.header.enrollments)
+            if (auto r = this.enroll_man.addValidator(enrollment,
                 block.header.height, this.utxo_set.getUTXOFinder()))
+            {
+                log.fatal("Error while adding a new validator: {}", r);
+                log.fatal("Enrollement #{}: {}", idx, enrollment);
+                log.fatal("Validated block: {}", block);
                 assert(0);
+            }
 
         // read back and cache the last block
         if (!this.storage.readLastBlock(this.last_block))
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -276,7 +276,7 @@ public class Ledger
         const ulong next_height = this.getBlockHeight() + 1;
         auto utxo_finder = this.utxo_set.getUTXOFinder();
 
-        this.enroll_man.getUnregistered(data.enrolls);
+        this.enroll_man.pool.getEnrollments(data.enrolls);
 
         foreach (hash, tx; this.pool)
         {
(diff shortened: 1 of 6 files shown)


Geod24:
Like all good heisenbugs, this one disappears as I try to add more logging...

source/agora/node/Ledger.d
```diff
@@ -227,10 +227,15 @@ public class Ledger
...
         this.enroll_man.clearExpiredValidators(block.header.height);
 
-        foreach (enrollment; block.header.enrollments)
-            if (!this.enroll_man.addValidator(enrollment,
+        foreach (idx, ref enrollment; block.header.enrollments)
+            if (auto r = this.enroll_man.addValidator(enrollment,
                 block.header.height, this.utxo_set.getUTXOFinder()))
+            {
+                log.fatal("Error while adding a new validator: {}", r);
+                log.fatal("Enrollement #{}: {}", idx, enrollment);
```
AndrejMitrovic, line 235: Can you fix up the typo `Enrollement` in a future PR?

Geod24: Also

source/agora/node/Ledger.d
```diff
@@ -281,7 +281,21 @@ public class Ledger
...
+        // TODO: Put a check higher-up that a block generated from the
+        // nomination data can be added.
+        Enrollment[] candidates;
+        this.enroll_man.pool.getEnrollments(candidates);
+        foreach (ref c; candidates)
+        {
+            if (this.enroll_man.getEnrolledHeight(c.utxo_key) == ulong.max)
+                data.enrolls ~= c;
+            else
+                this.enroll_man.pool.remove(c.utxo_key);
```
AndrejMitrovic, line 296: Right, now I see what you meant by removing from the pool.

I guess we also need some mechanism of putting the enrollments back into the pool if the candidate set here is not the same set that becomes externalized.

AndrejMitrovic: Should we file an issue for this?

Geod24: #900

Geod24: Note: With a periodic cleanup, we could also simply remove this branch

AndrejMitrovic: I don't think this was correct. The items should be removed from the pool after a block is externalized, not during the nomination process..

My Quorum PR failing. I'm still investigating why though.

Geod24: Well the idea with the pool is that we shouldn't *need* the items in the pool. Evicting items from the pool too early simply means we'll have less validators to propose, and as a results, the blocks we nominate are less likely to be chosen (once the consensus rules are implemented).

AndrejMitrovic: I can see why it's an issue for my PR. It's because I'm reusing the same UTXO key. But this should be supported as mentioned in #824.

Also I don't quite understand your comment.

linked0: I don't understand the reason why the candidates should be removed from the pool before the consensus is made. What I'm concerned about is the case a consensus fails with enrollments removed.

PR #902, opened 2020-05-31
Use lockingTextWriter instead of synchronized for logs

Geod24:
Using lockingTextWriter ensures that inter-dependent loggings
will be next to each others, and is more fine-grained.

I still need to test this


AndrejMitrovic:
> // Note: This was documented until 2013/08

Wow that is some useful documentation in Phobos.

Geod24:
Tested, works

AndrejMitrovic (approved):

PR #905, opened 2020-06-01
Move PreImage utilities into a common module

Geod24:
This way they can be re-used when a node is simulating another node,
or want to do speculative work.


linked0 (approved):
LGTM

Geod24:
Should be fixed now

linked0:
> Should be fixed now

Do you mean "merge at green"?

Geod24:
Related to #907 as we need to add pre-images to the genesis block as part of the Enrollment.

PR #909, opened 2020-06-02
Properly silence SCP trace logs

AndrejMitrovic:
Only some of the SCPP code checks if (mLogTrace(...)),
in many routines this is not checked. For example in
Slot::recordStatement which unconditionally calls
CLOG(DEBUG, "SCP") << "new statement: " ..

### DIFF
--- a/source/scpp/src/util/Logging.cpp
+++ b/source/scpp/src/util/Logging.cpp
@@ -19,6 +19,9 @@ DLogger::DLogger(int level, std::string const& loggerName)
 
 DLogger::~DLogger()
 {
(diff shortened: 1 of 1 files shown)


Geod24:
Alternatively, maybe we should just configure the logger on the D side ? Although that would still lead to overhead.

AndrejMitrovic:
Didn't you make a commit to silence SCP first? it was done on the C++ side.

Geod24:
Yep, but as you discovered, it was far less from ideal.
Anyway we can go ahead with it.

Geod24 (approved):

PR #910, opened 2020-06-02
Replace unnecessary scope(success)

AndrejMitrovic:
@TrustHenry this should make things clearer.

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -102,10 +102,8 @@ public class Validator : FullNode, API
 
             while (1)
             {
-                scope (success)
-                    this.taskman.wait(5.seconds);
-
                 this.network.discover(required_peer_keys);
+                this.taskman.wait(5.seconds);
             }
         });
     }


TrustHenry (approved):
LGTM

source/agora/network/NetworkManager.d
```diff
@@ -410,9 +409,6 @@ public class NetworkManager
 
         while (!this.peerLimitReached())
         {
-            scope (success)
-                this.taskman.wait(this.node_config.retry_delay.msecs);
```
Geod24: There's a small difference here: It won't sleep after the `continue` on L412. But I'm wondering, isn't there an infinite loop there ?

source/agora/network/NetworkManager.d
```diff
@@ -269,9 +269,6 @@ public class NetworkManager
         {
             while (1)
             {
-                scope (success)
-                    this.outer.taskman.wait(1.seconds);
```
Geod24: Same here

source/agora/network/NetworkManager.d
AndrejMitrovic: Yes until `peerLimitReached()` is true.

Geod24: But it's a busy wait, before it wasn't

AndrejMitrovic: It's the same as before?

https://github.com/bpfkorea/agora/blob/aef86ff0151c36ff83e9af9d05ec77fc7b75fd65/source/agora/network/NetworkManager.d#L208-L212

AndrejMitrovic:
I had to revert two changes. I missed there was still a `continue` in there. Maybe this can be simplified in the future..

Geod24 (approved):

PR #911, opened 2020-06-02
Add an overload of 'sign' which accepts a signature noise

Geod24:
Using a signature noise might be frequent enough to warrant an overload.
This needs to be used with care, as reusing `R` would be deadly to the node.


linked0 (approved):
This makes the code simpler so LGTM.

PR #912, opened 2020-06-02
Improve Enrollment validation messages and tests

Geod24:
Pretty simple, see each commit for rationale.

### DIFF
--- a/source/agora/consensus/validation/Enrollment.d
+++ b/source/agora/consensus/validation/Enrollment.d
@@ -53,10 +53,10 @@ public string isInvalidReason (const ref Enrollment enrollment,
 {
     UTXOSetValue utxo_set_value;
     if (!findUTXO(enrollment.utxo_key, size_t.max, utxo_set_value))
-        return "Unspent frozen UTXO not found for the validator.";
+        return "Enrollment: UTXO not found";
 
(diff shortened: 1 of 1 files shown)


source/agora/consensus/validation/Enrollment.d
```diff
@@ -53,10 +53,10 @@ public string isInvalidReason (const ref Enrollment enrollment,
 {
     UTXOSetValue utxo_set_value;
     if (!findUTXO(enrollment.utxo_key, size_t.max, utxo_set_value))
-        return "Unspent frozen UTXO not found for the validator.";
+        return "Enrollment: UTXO not found";
 
     if (utxo_set_value.type != typeof(utxo_set_value.type).Freeze)
-        return "UTXO is not frozen.";
+        return "Enrollment: UTXO is not frozen";
```
linked0, line 59: The line, 202 as follows should be changed based on the code `return "Enrollment: UTXO is not frozen";`, which means that it is needed to omit the period(.).
```
assert(canFind(enroll2.isInvalidReason(utxoFinder), "UTXO is not frozen."));
```

Geod24: Thanks, fixed

linked0 (approved):
LGTM

PR #916, opened 2020-06-02
Amount: Add support for division

Geod24:
@AndrejMitrovic : I remember you wanted this ?

### DIFF
--- a/source/agora/common/Amount.d
+++ b/source/agora/common/Amount.d
@@ -169,6 +169,35 @@ public struct Amount
         return true;
     }
 
+    /***************************************************************************
+
+        Divide this `Amount` into `denominator` equal parts and
+        returns the remainder
+
+        This function allows to losslessly partition an `Amount`.
+        Usage example include distributing an `Amount` evenly among participant,
+        or computing an absolute majority of 2/3rd.
+
+        Params:
+            denominator = The value to divide by.
+                          Dividing by `0` is the same as dividing by `1`.
+
+        Returns:
+            The remainder of the division
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public Amount div (uint denominator)
+    {
+        // No-op
+        if (denominator < 2) return Amount(0);
+
+        const Amount remainder = this.value % denominator;
+        this.value = (this.value / denominator);
+        return remainder;
+    }
+
     /// Returns: The integral part of the amount (value / 1 BOA)
     public ulong integral () const
     {
@@ -240,6 +269,29 @@ nothrow pure @nogc @safe unittest
     assert(two == Amount(1));
     two.mustSub(Amount(1));
     assert(two == Amount(0));
+
+    // Tests for division with remainder
+    Amount val = Amount.MinFreezeAmount;
+    assert(val.div(3) == Amount(1));
+    assert(val.integral() == 13_333);
+    assert(val.decimal() == 3_333_333);
+
+    assert(val.div(100) == Amount(33));
+    assert(val.integral() == 133);
+    assert(val.decimal() == 3_333_333);
+
+    // Division without remainder
+    val = Amount.MinFreezeAmount;
+    assert(val.div(1000) == Amount(0));
+    assert(val.integral() == 40);
+    assert(val.decimal() == 0);
+
+    // Division edge cases
+    val = Amount.UnitPerCoin;
+    assert(val.div(0) == Amount(0));
+    assert(val == Amount.UnitPerCoin);
+    assert(val.div(0) == Amount(0));
+    assert(val == Amount.UnitPerCoin);
 }
 
 pure @safe unittest


AndrejMitrovic:
Yes. Also multiplication would be nice too!

As long as it's explicit and not an operator overload, it should be fine right?

source/agora/common/Amount.d
```diff
@@ -169,6 +169,34 @@ public struct Amount
...
+
+        Divide this `Amount` into `Q` equal parts and returns the remainder
+
+        This function allows to losslessly partition an `Amount`.
+        Usage example include distributing an `Amount` evenly among participant,
+        or computing an absolute majority of 2/3rd.
+
+        Params:
+            value = The value to divide by.
+                    Dividing by `0` is the same as dividing by `1`.
```
AndrejMitrovic:
```suggestion
                    Dividing by `0` is the same as dividing by `1`, and will return Amount(0)
```

Maybe it should be named `split` if it **must** split a value into different parts?

I'm not sure. I've never seen a math function fail to divide by 1, but I'm sure you have a good reason for it. :)

AndrejMitrovic: Ah it returns the remainder, I'm sorry I misread it!

AndrejMitrovic: Is there a particular reason why all the operations have side-effects rather than returning a new value?

Geod24: Because if we wanted them to create new values, it would be:
```D
bool div (ref Amount result, uint value) const;
```

Because we don't want to `return` the value (to avoid misuse of the API). And it felt like, if you need a new value, you can just copy it. But I'm open to suggestions.

AndrejMitrovic: > to avoid misuse of the API

That is what I'm wondering about. What are the examples of misusing it?

My only concern is that these operators do not compose well with range routines like `map` and I can't use them with `const` values, like `const amount = left + right`. In general I prefer functions with no side-effects.

Geod24:
```D
bool b;
auto result = myAmount.add( Amount(84), b);
assert(b);
// ...
auto res = result.add(Amount(42), b);
// No check for `b` here
```

In general it's easier to miss checking a parameter than a return value.
Regarding `map`, is it for unittests or production code ? For the former you can just use `Amount(cost_val).mustAdd(another_amount)`.

AndrejMitrovic: Right, this is good.

I think we really need to propose a `@noignore` attribute to dlang though. I suggested it years ago but there was no DIP process back then.

AndrejMitrovic: https://forum.dlang.org/post/mailman.4463.1420983153.9932.digitalmars-d@puremagic.com

TrustHenry:
```
unittest
{
    import std.stdio;
    Amount val1 = Amount(10);
    writeln("val1 : ", val1);
    writeln("div : ", val1.div(3));
    writeln("integral : ", val1.integral());
    writeln("decimal : ", val1.decimal());
}
```

run dub test

```
val1 : 10
div : 1
integral : 0
decimal : 3
```


Is this the intended result?

MichaelKim20:
> ```
> unittest
> {
>     import std.stdio;
>     Amount val1 = Amount(10);
>     writeln("val1 : ", val1);
>     writeln("div : ", val1.div(3));
>     writeln("integral : ", val1.integral());
>     writeln("decimal : ", val1.decimal());
> }
> ```
> 
> run dub test
> 
> ```
> val1 : 10
> div : 1
> integral : 0
> decimal : 3
> ```
> 
> Is this the intended result?

I think it's working fine.

AndrejMitrovic:
> Is this the intended result?

Yes. See `UnitPerCoin`. If you expected `integral` to change you would need to use a bigger value. For example:

```D
unittest
{
    import std.stdio;
    Amount val1 = Amount(100_000_000);
    writeln("div : ", val1.div(3));
    writeln("integral : ", val1.integral());
    writeln("decimal : ", val1.decimal());
}
```

```
div : 1
integral : 3
decimal : 3333333
```

Geod24:
Yep, that's why the test use `MinFreezeAmount`.

source/agora/common/Amount.d
```diff
@@ -169,6 +169,34 @@ public struct Amount
...
+        Returns:
+            The remainder of the division
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public Amount div (uint value)
+    {
+        // No-op
+        if (value < 2) return Amount(0);
```
MichaelKim20: Does the compiler recognize the `value` as the `value` of the parameter? Or should we recognize it as `this.value`?

MichaelKim20: I see that the test code is working properly, so I know exactly what the `value` of the parameter is.
The compiler is smart.

Geod24: I should probably use a different name though, good point.

Geod24:
Changed parameter name

source/agora/common/Amount.d
```diff
@@ -169,6 +169,34 @@ public struct Amount
         return true;
     }
 
+    /***************************************************************************
+
+        Divide this `Amount` into `Q` equal parts and returns the remainder
```
AndrejMitrovic: Is `Q` a math symbol or what?

Geod24: I have no idea what happened here

Geod24: Fixed

AndrejMitrovic (approved):

Issue #894, opened 2020-05-29
Turn necessary consensus-critical constant into argument


Geod24:
At the moment, we have issues with consensus constants being constant and accessed through the code source. E.g. we have the various genesis values, but also the cycle length.

The problem with such constants is that they might need to be modified to ensure things work correctly at a low cost. E.g. our cycles are 1008 blocks at the moment, which is expensive to generate in a test. Being able to reduce this to a few blocks would make things more scalable.

There's three approaches I can think of:
- Having them with `version` => This is what we're doing currently. It works well when one need not to change those values, or when there's a "common default". E.g. we have a common genesis block for all unittests. However it creates problems for integration testing as we need a special build.

- Having them global with selected with a runtime parameter. This is what we started to do with the genesis block, e.g. #712 , but it also creates some issues, e.g. #813. This issue recently showed again with `ValidatorCycle`.

- Having a global context that is passed around. This would be similar to the idea of "Consensus params" in the BTC implementation: https://github.com/bitcoin/bitcoin/blob/master/src/consensus/params.h but might lead to a lot of context being passed around if overused.

Note that BTC does have a mixture of both 2 and 3: https://github.com/bitcoin/bitcoin/blob/master/src/consensus/consensus.h

linked0:
I think the best way to deal with this issue is to make a configuration for the constants like `ConsensusConfig`. How do you think of it?
If my idea is right, it's related to the issue #520, isn't it?

AndrejMitrovic:
I'm not sure it's related to #520. #520 is unrelated to the protocol (you can change any settings and it would not break the protocol), whereas this issue #894 is about protocol changes.

AndrejMitrovic:
Should the genesis block then also be part of the Consensus parameters?

linked0:
> Should the genesis block then also be part of the Consensus parameters?

Sure, it is a good candidate for the class because it never changes on the network and it’s related to consensus. I can add the genesis block to the class after the related PR is merged and @Geod24  agrees to that.

Geod24:
What is left to do ?
I can think of:
- Moving the Genesis block there: This can be a separate issue for the next sprint;
- Adapting the CLI to parse the argument: That's the only thing I can think of that is required to close this issue.

PR #920, opened 2020-06-03
Create a class managing consensus constants

linked0:
In order to set the consensus-critical constants in runtime and share them between every object, we must provide only one parameter set. So we worked on this with the concept of the Singleton pattern in order to meet the requirements.

The only `Params` object exists after being created with the `createParams` function that calls the constructor of the class.

Relates #894

### DIFF
--- a/source/agora/consensus/data/Params.d
+++ b/source/agora/consensus/data/Params.d
@@ -1,0 +1,107 @@
+/*******************************************************************************
+
+    Parameter set for consensus-critcal constants
+
+    In order to set the consensus-critical constants in runtime and share them
+    between every objects, we must provide only one parameter set. So we worked
+    on this with the concept of the Singleton pattern in order to meet the
+    requirements.
+
+    The only `Params` object exists after being created with the `createParams`
+    function that calls the constructor of the `Params` class.
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.Params;
+
+class Params
+{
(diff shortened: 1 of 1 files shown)


source/agora/consensus/data/Params.d
```diff
@@ -0,0 +1,91 @@
...
+        Params:
+            validator_cycle = cycle length for a validator
+            preimage_reveal_period = period for revealing a preimage
+
+    ***************************************************************************/
+
+    static void createParams (uint validator_cycle,
+        uint preimage_reveal_period)
+    {
+        if (!instantiated_)
```
AndrejMitrovic, line 69: You still have a data race here.

linked0: This might not right for the function because the function is not called often. Let me check again.

I referred to the Singleton pattern.
https://wiki.dlang.org/Low-Lock_Singleton_Pattern

AndrejMitrovic: Aha it is thread local, I thought it was global but I didn't notice that.

AndrejMitrovic:
I'm not entirely sure if this is what @Geod24 had in mind. And since it's not used yet, how would we know this is the right approach?

@Geod24 can you share your thoughts if this is what you wanted?

linked0:
Ready for review.

linked0:
> I'm not entirely sure if this is what @Geod24 had in mind. And since it's not used yet, how would we know this is the right approach?
> 
> @Geod24 can you share your thoughts if this is what you wanted?

How about having a white-boarding session about this today?

linked0:
I have realized that using the Singleton pattern is a totally wrong solution. So I'm closing it.
Sorry for the confusion. 

I guess the best solution is to create a class/structure like `Params`, which is created and managed by a Node. A Node should have a function for setting values for the parameters.

Issue #872, opened 2020-05-26
Use the new TaskManager.setTimer() function


AndrejMitrovic:
These need to be replaced to use setTimer once TaskManager is fixed to use it.

https://github.com/bpfkorea/agora/blob/bebe6555052b830bff1040847edd035046d24cd4/source/agora/consensus/protocol/Nominator.d#L507-L509

https://github.com/bpfkorea/agora/blob/bebe6555052b830bff1040847edd035046d24cd4/source/agora/network/NetworkManager.d#L203-L210

https://github.com/bpfkorea/agora/blob/bebe6555052b830bff1040847edd035046d24cd4/source/agora/network/NetworkManager.d#L235-L246

_Originally posted by @AndrejMitrovic in https://github.com/bpfkorea/agora/pull/871#issuecomment-633762479_

Geod24:
`setTimer` available as of #860

TrustHenry:
https://github.com/bpfkorea/agora/blob/bebe6555052b830bff1040847edd035046d24cd4/source/agora/network/NetworkManager.d#L203-L210
https://github.com/bpfkorea/agora/blob/bebe6555052b830bff1040847edd035046d24cd4/source/agora/network/NetworkManager.d#L235-L246

Here's the question.
If it look at the existing code, wait after the function is done.
For the new function localrest, first wait and then call delegate.
Would it be okay to do the two things mentioned here after wait and call?

This creates a subtle time difference at the beginning.

AndrejMitrovic:
Please be aware you're pointing to outdated commits.

Also yes, there's a difference. We probably don't want to delay every request. It would be nice to be able to do `start + wait` rather than just `wait + start`.

TrustHenry:
The setTimer() also requires the stop function.

TrustHenry:
Applied code line
2. https://github.com/bpfkorea/agora/blob/07dfc407a31828ee3d0596da29696235c5de5c81/source/agora/node/FullNode.d#L178
3.
https://github.com/bpfkorea/agora/blob/4c6b990955645003f6736d231fc781f366b7588b/source/agora/network/NetworkManager.d#L532

PR #921, opened 2020-06-03
Use setTimer in NetworkManager.startPeriodicCatchup 

TrustHenry:
These need to be replaced to use setTimer once TaskManager is fixed to use it.

Relates #872

### DIFF
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -515,14 +515,10 @@ public class NetworkManager
         this.taskman.runTask(
         ()
         {
-            // periodic task
-            while (1)
+            void catchup ()
             {
-                scope (success)
-                    this.taskman.wait(2.seconds);
-
                 if (this.peers.length == 0)  // no clients yet (discovery)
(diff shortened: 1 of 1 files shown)


source/agora/network/NetworkManager.d
```diff
@@ -430,6 +430,10 @@ public class NetworkManager
                     address, &onHandshakeComplete, &onFailedRequest);
             }
         }
+
+        connnectionTask(); // avoid delay
+        this.taskman.setTimer(this.node_config.retry_delay.msecs,
```
AndrejMitrovic: This is not correct. Previously the calling fiber was yielded, but here you are spawning a new one. You're going to end up with multiple fibers running within this class. This part should not be changed.

TrustHenry: That's what I was curious about. Thank you for your clear answer.

TrustHenry: The commit has been deleted.

AndrejMitrovic (approved):

PR #924, opened 2020-06-03
Add a function to restart a given node (and a test for it)

Geod24:
```
Using `ctrl.restart` directly is not advised as the teardown / startup
code won't be run. Instead, this method is provided.
```

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -444,6 +444,24 @@ public class TestAPIManager
 
     /***************************************************************************
 
+        Restart a specific node
+
+        This routine restarts the given `client`, making sure it gracefully
+        shuts down then restart properly.
+
+        Params:
+          client = Reference to the client to restart
+
+    ***************************************************************************/
+
+    public void restart (scope RemoteAPI!TestAPI client)
+    {
+        client.ctrl.restart((TestAPI node) { (cast(FullNode)node).shutdown(); });
+        client.ctrl.withTimeout(0.msecs, (scope TestAPI api) { api.start(); });
+    }
+
+    /***************************************************************************
+
         Print out the logs for each node
 
     ***************************************************************************/
--- a/source/agora/test/Simple.d
+++ b/source/agora/test/Simple.d
@@ -44,3 +44,29 @@ unittest
     nodes.all!(node => node.getBlockHeight() == 1)
         .retryFor(2.seconds);
 }
+
+/// A test that stops and restarts a node
+unittest
+{
+    auto network = makeTestNetwork(TestConf.init);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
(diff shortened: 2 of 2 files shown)


source/agora/test/Base.d
```diff
@@ -442,6 +442,29 @@ public class TestAPIManager
...
+
+        This routine restarts the node at the given `index`, making sure it
+        gracefully shuts down then restart properly.
+
+        Params:
+          index = Index of the node to restart
+
+    ***************************************************************************/
+
+    public void restart (size_t index)
```
AndrejMitrovic: Could this take a client as a parameter instead?

```D
auto node = network.clients[4];
network.restart(node);
```

I think using an index could be a little awkward to use?

Geod24: Doh. So obvious now you mention it.

Geod24: Fixed

source/agora/test/Simple.d
```diff
@@ -44,3 +44,28 @@ unittest
...
+    txes.each!(tx => node_1.putTransaction(tx));
+
+    nodes.all!(node => node.getBlockHeight() == 1)
+        .retryFor(2.seconds);
+
+    // Now shut down & restart one node
+    auto restartMe = nodes[$-1];
+    network.restart(restartMe);
+    network.waitForDiscovery();
+    retryFor(restartMe.getBlockHeight() == 1, 5.seconds);
```
linked0: How about changing the line as follows or adding this?
```
nodes.all!(node => node.getBlockHeight() == 1)
        .retryFor(2.seconds);
```
It depends on you.

linked0 (approved):
Nice work.

source/agora/test/Simple.d
Geod24: Makes sense, done

Geod24:
Apparently the timeout wasn't enough so I put it back to 5 seconds

AndrejMitrovic (approved):

source/agora/test/Simple.d
```diff
@@ -44,3 +44,29 @@ unittest
...
+    nodes.all!(node => node.getBlockHeight() == 1)
+        .retryFor(2.seconds);
+
+    // Now shut down & restart one node
+    auto restartMe = nodes[$-1];
+    network.restart(restartMe);
+    network.waitForDiscovery();
+    nodes.all!(node => node.getBlockHeight() == 1)
+        .retryFor(5.seconds);
+}
```
MichaelKim20, line 72: I tested it by adding the following code.
Everything is good.

``` D
    auto txes2 = makeChainedTransactions(getGenesisKeyPair(), txes, 1);
    txes2.each!(tx => node_1.putTransaction(tx));

    nodes.all!(node => node.getBlockHeight() == 2)
        .retryFor(5.seconds);
```

MichaelKim20 (approved):

PR #927, opened 2020-06-04
[WIP]Implement a universal function for enrollment the validators

MichaelKim20:
I implemented a universal function for enrollment the validators

Create the first block by creating frozen UTXOs to enroll the validator, and create a second block to enroll the validator.

Relates to #822


Geod24:
Is this something that is still needed ? Between `EnrollmentManager.makeEnrollment` and the changes to the `UnittestGenesisBlock`, I think we are covered, aren't we ?

PR #928, opened 2020-06-04
Add a 'split' function to simplify writing tests

MichaelKim20:
This function is aimed to partially replace `makeChainedTransactions`.
This function transfers exactly all amounts from the original transaction to the target transaction.
At this time we can set the number and address of the outputs.

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -29,10 +29,13 @@
 
 module agora.utils.Test;
 
+import agora.common.Amount;
 import agora.common.crypto.Key;
+import agora.common.Hash;
 import agora.consensus.data.Transaction;
 import agora.consensus.Genesis;
 
+import std.algorithm;
 import std.file;
 import std.path;
 
@@ -286,7 +289,6 @@ unittest
 /// custom genesis tx
 unittest
 {
-    import std.algorithm;
     import std.exception : assumeUnique;
     import std.range;
     import core.thread;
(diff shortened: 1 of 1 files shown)


source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
     }
 }
+
+/*******************************************************************************
+
+    Generate a new transaction that evenly splits the input accross parties
+
+*******************************************************************************/
+
+public Transaction split (TxType type = TxType.Payment)
```
Geod24, line 618: Note to reviewers: This does not need to be a template parameter, but it simplifies the API IMHO.

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
         assert(!verify(pz.V, ssz, "WK.Keys.z".representation));
     }
 }
+
+/*******************************************************************************
+
+    Generate a new transaction that evenly splits the input accross parties
+
+*******************************************************************************/
```
```suggestion
/*******************************************************************************

    Generate a new transaction that evenly splits the input accross parties

    The `input` transaction will be split evenly in `toward.length` outputs,
    each of which will be controlled by a key in `toward`.
    If the sum of outputs in the transaction is not a multiple of
    `toward.length`, the leftover `Amount` will be added to
    the output for `toward[0]`.
    Like other testing utilities, if an error happens (e.g. `input` is invalid,
    or `from` is missing some keys), an `assert` will be triggered.
    
    Params:
        type = Type of transaction to generate (frozen or simple payment)
        input = Transaction to spend completely
        from = Array of keys controlling all the outputs in `input`
        toward = Receivers for the newly-created output
        
    Returns:
        A newly created, valid, and signed `Transaction` spending `input`.

*******************************************************************************/
```

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+
+    // Add the remainder to the first output.
+    result.outputs[0].value.mustAdd(remainder);
+    const inputHash = input.hashFull();
+
+    // Add support for Transactions with multiple recipients
+    foreach (idx, const ref _; input.outputs)
+        result.inputs ~= Input(inputHash, cast(uint) idx);
+
+    auto sign = (KeyPair kp, Hash h) @trusted { return kp.secret.sign(h[]); };
```
```suggestion
    scope sign = (KeyPair kp, Hash h) @trusted { return kp.secret.sign(h[]); };
```

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+public Transaction split (TxType type = TxType.Payment)
+    (const ref Transaction input, scope const KeyPair[] from,
+     scope const PublicKey[] toward...)
+    @safe
+{
+    Amount amount;
+    if (!input.getSumOutput(amount))
+        assert(0, "Invalid transaction passed to `split`");
+
+    auto remainder = amount.div(cast(uint) toward.length);
```
```suggestion
    auto remainder = amount.div(toward.length);
```

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+/// Test with twice as many outputs as inputs
+unittest
+{
+    import std.range;
+
+    KeyPair[] keys = iota(16).map!(_ => KeyPair.random()).array;
+    // Use Genesis
+    KeyPair genesisKP = getGenesisKeyPair();
+    const first = GenesisBlock.txs[0];
+    const equalTx = first.split([genesisKP], keys.map!(k => k.address).array);
```
```suggestion
    const result = first.split([genesisKP], keys.map!(k => k.address).array);
```

I used `equalTx` because it was an "equality" transaction, but here it doesn't make sense.
For consistency, you can change all `equalTx` to `result` (even the first one).

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+    assert(equalTx.inputs.length == 8);
+    assert(equalTx.outputs.length == 3);
+
+    Amount ExpectedAmount;
+    first.getSumOutput(ExpectedAmount);
+    auto remainder = ExpectedAmount.div(3);
+    assert(remainder != Amount(0));
+
+    // The first output includes the remainder.
+    assert(equalTx.outputs[0].value >= ExpectedAmount);
```
Geod24: This is not a good test!! We want to be sure that the result match our expectation!

```suggestion
    assert(equalTx.outputs[0].value == (ExpectedAmount + 2));
```

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+    const first = GenesisBlock.txs[0];
+    const equalTx = first.split([genesisKP], keys.map!(k => k.address).array);
+    // This transaction has 3 txs
+    assert(equalTx.inputs.length == 8);
+    assert(equalTx.outputs.length == 3);
+
+    Amount ExpectedAmount;
+    first.getSumOutput(ExpectedAmount);
+    auto remainder = ExpectedAmount.div(3);
+    assert(remainder != Amount(0));
```
Geod24: Using `getSumOutput` isn't great here. I'd recommend to hardcode the value (even if it means changing it later on).  One thing we don't want is having tests that replicate the code in the function. We should test the behavior instead, and having hard-coded value makes this simpler.

For example:
```suggestion
    // 500M / 3
    const Amount ExpectedAmount        = Amount(166_666_666_6666_666);
    const Amount ExpectedFirstAmount = Amount(166_666_666_6666_668);
```

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+    // Use Genesis
+    KeyPair genesisKP = getGenesisKeyPair();
+    const first = GenesisBlock.txs[0];
+    const equalTx = first.split([genesisKP], [key.address]);
+    // This transaction has 3 txs
+    assert(equalTx.inputs.length == 8);
+    assert(equalTx.outputs.length == 1);
+
+    Amount ExpectedAmount;
+    first.getSumOutput(ExpectedAmount);
```
Geod24: This usage of `getSumOutput` is okay-ish, but the return value needs to be checked!

```suggestion
    Amount ExpectedAmount;
    assert(first.getSumOutput(ExpectedAmount));
```

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+    KeyPair genesisKP = getGenesisKeyPair();
+    const first = GenesisBlock.txs[0];
+    const resTx1 = first.split([genesisKP], keys8.map!(k => k.address).array);
+    // This transaction has 8 txs, hence it's just equality
+    assert(resTx1.inputs.length == 8);
+    assert(resTx1.outputs.length == 8);
+    // Since the amount is evenly distributed in Genesis,
+    // they all have the same value
+    const ExpectedAmount1 = first.outputs[0].value;
+    assert(resTx1.outputs.all!(val => val.value == ExpectedAmount1));
```
Geod24: Since this is essentially the first unittest, you could just merge both. We shouldn't copy-paste tests around or it leads to a lot of duplication / resistance when changing things.

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+
+    // Split into 16 outputs
+    KeyPair[] keys16 = iota(16).map!(_ => KeyPair.random()).array;
+    const resTx2 = resTx1.split(keys8, keys16.map!(k => k.address).array);
+    // This transaction has 16 txs, hence it's just equality
+    assert(resTx2.inputs.length == 8);
+    assert(resTx2.outputs.length == 16);
+
+    Amount ExpectedAmount2;
+    first.getSumOutput(ExpectedAmount2);
```
Geod24: Please use a fixed (hardcoded value).

source/agora/utils/Test.d
```diff
@@ -589,3 +591,174 @@ unittest
...
+    KeyPair[] keys32 = iota(32).map!(_ => KeyPair.random()).array;
+    const resTx3 = resTx2.split(keys16, keys32.map!(k => k.address).array);
+    // This transaction has 32 txs, hence it's just equality
+    assert(resTx3.inputs.length == 16);
+    assert(resTx3.outputs.length == 32);
+
+    Amount ExpectedAmount3;
+    first.getSumOutput(ExpectedAmount3);
+    auto remainder3 = ExpectedAmount3.div(32);
+    assert(remainder3 == Amount(0));
```
Geod24: Please use a fixed (hardcoded value).

Geod24 (changes requested):
A few suggestions

MichaelKim20:
@Geod24 Ready for Review

Geod24 (approved):

PR #932, opened 2020-06-04
getBlocksFrom() should return a mutable array

AndrejMitrovic:
Commit explains it.

@Geod24 is there a way I can avoid doing the `cast()` here? Otherwise I'm getting the mutable/immutable array compare mismatch error:

> source/agora/test/Ledger.d(69,12): Error: array equality comparison type mismatch, Transaction[] vs immutable(Transaction[])

Maybe we just need to implement opEquals manually?

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -152,7 +152,7 @@ public interface API
 
     ***************************************************************************/
 
-    public const(Block)[] getBlocksFrom (ulong block_height, uint max_blocks);
+    public Block[] getBlocksFrom (ulong block_height, uint max_blocks);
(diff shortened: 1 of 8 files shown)


AndrejMitrovic:
ah this change is bad, I should not pass around mutable. I realized it shortly after making the pr. :)

AndrejMitrovic:
I'm gonna close this for now. My original intention was to remove the need for a write + read (+ deserialization) here: https://github.com/bpfkorea/agora/blob/d2931dd0d0f552997bc80e799a5275f5c3cecff5/source/agora/node/Ledger.d#L241

PR #933, opened 2020-06-05
Introduce a ManagedDatabase class

AndrejMitrovic:
This finally resolves all of our issues with the D2SQLite destructor throwing Errors, and it removes the need for all of the .shutdown() routines and `scope (exit)` calls.

These benefits were actually incidental. I was designing a database manager class which exposed `batch / commit / rollback` SQLite routines on all its managed databases. But I soon ran into many issues with managing the `Database` D2SQLite struct instance, its internal RefCounted struct was a pain and kept causing issues.

Then I've realized that since we only have a small set of databases and we only want to destroy them at the very end, why not just destroy them when the thread is destroyed? This way even if an assert in a unittest fails, the static module destructor will *not* be detected as being ran in a "regular" destructor.

Part of #836.

### DIFF
--- a/source/agora/common/ManagedDatabase.d
+++ b/source/agora/common/ManagedDatabase.d
@@ -1,0 +1,81 @@
+/*******************************************************************************
+
+    Contains a managed wrapper around SQLite databases.
+
+    This class ensures that all D2SQLite databases handled by the thread
+    are cleanly destroyed after the thread shuts down.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.ManagedDatabase;
+
+import agora.utils.Log;
+
+import d2sqlite3.database;
+import d2sqlite3.sqlite3;
+
+mixin AddLogger!();
+
+/// Ditto
+public class ManagedDatabase
+{
+    /// ManagedDatabase managed by the current thread
+    private static Database*[] thread_dbs;
+
+    /// Close all database handles
+    static ~this ()
+    {
+        try
+        {
+            foreach (db; thread_dbs)
+                db.close();
+        }
+        catch (Exception ex)
+        {
+            log.error("Error closing database handles: {}", ex.message);
+            throw ex;
+        }
+    }
+
+    /// Pointer to the database handle
+    private Database* database;
+
+    /// Subtype
+    public alias getDatabase this;
+
+    /***************************************************************************
+
+        Constructs a managed database
+
+        Params:
+            path = path to the db on disk, or :memory: for an in-memory database
+            flags = SQLite read / write flags
+
+    ***************************************************************************/
+
+    public this (string path,
+        int flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE)
+    {
+        this.database = new Database(path, flags);
+        thread_dbs ~= this.database;
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the database handle (used via alias this)
+
+    ***************************************************************************/
+
+    public Database* getDatabase () pure nothrow @safe @nogc
+    {
+        return this.database;
+    }
+}
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -14,6 +14,7 @@
 
 module agora.common.TransactionPool;
 
+import agora.common.ManagedDatabase;
 import agora.common.Types;
 import agora.common.Hash;
 import agora.common.Serializer;
@@ -21,7 +22,6 @@ import agora.common.Set;
 import agora.consensus.data.Transaction;
 import agora.utils.Log;
 
-import d2sqlite3.database;
 import d2sqlite3.library;
 import d2sqlite3.results;
 import d2sqlite3.sqlite3;
@@ -86,7 +86,7 @@ public class TransactionPool
     }
 
     /// SQLite db instance
-    private Database db;
+    private ManagedDatabase db;
 
     /// Map for hashes for Input objects of each transaction in TX pool
     private Set!Hash input_set;
@@ -106,7 +106,7 @@ public class TransactionPool
             log.info("Loading database from: {}", db_path);
 
         // note: can fail. we may want to just recover txes from the network instead.
-        this.db = Database(db_path);
+        this.db = new ManagedDatabase(db_path);
 
         if (db_exists)
             log.info("Loaded database from: {}", db_path);
@@ -241,20 +241,6 @@ public class TransactionPool
 
     /***************************************************************************
 
-        Shut down the database
-
-        Note: this method must be called explicitly, and not inside of
-        a destructor.
-
-    ***************************************************************************/
-
-    public void shutdown ()
-    {
-        this.db.close();
-    }
-
-    /***************************************************************************
-
         Check if the input transaction has any double spending input.
 
         Params:
@@ -336,8 +322,6 @@ unittest
     import agora.consensus.Genesis;
 
     auto pool = new TransactionPool(":memory:");
-    scope(exit) pool.shutdown();  // note: must call outside destructor
-
     auto gen_key = getGenesisKeyPair();
     auto txs = makeChainedTransactions(gen_key, null, 1);
 
@@ -376,8 +360,6 @@ unittest
     import std.exception;
 
     auto pool = new TransactionPool(":memory:");
-    scope(exit) pool.shutdown();  // note: must call outside destructor
-
     auto gen_key = getGenesisKeyPair();
     auto txs = makeChainedTransactions(gen_key, null, 1);
 
@@ -409,8 +391,6 @@ unittest
     import core.memory;
 
     auto pool = new TransactionPool(":memory:");
-    scope(exit) pool.shutdown();
-
     auto gen_key = getGenesisKeyPair();
     auto txs = makeChainedTransactions(gen_key, null, 1);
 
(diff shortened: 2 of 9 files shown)


source/agora/common/Database.d
```diff
@@ -0,0 +1,54 @@
...
+
+module agora.common.Database;
+
+import SQLite = d2sqlite3.database;
+import d2sqlite3.sqlite3;
+
+/// Ditto
+public class Database
+{
+    /// Databases managed by either the current Agora instance or the current
```
AndrejMitrovic: outdated comment

source/agora/common/Database.d
```diff
@@ -0,0 +1,54 @@
...
+        all_dbs.length = 0;
+    }
+
+    ///
+    public SQLite.Database* database;
+
+    /// Subtype
+    public alias database this;
+
+    /// Get a managed database
```
AndrejMitrovic: outdated comment

source/agora/common/Database.d
```diff
@@ -0,0 +1,54 @@
...
+            db.close();
+
+        foreach (db; all_dbs)
+            destroy(*db);
+
+        all_dbs.length = 0;
+    }
+
+    ///
+    public SQLite.Database* database;
```
AndrejMitrovic: missing docs

source/agora/common/Database.d
```diff
@@ -0,0 +1,54 @@
...
+    /// Close and destroy all database handles
+    static ~this ()
+    {
+        foreach (db; all_dbs)
+            db.close();
+
+        foreach (db; all_dbs)
+            destroy(*db);
+
+        all_dbs.length = 0;
```
AndrejMitrovic: seems unnecessary

source/agora/common/Database.d
```diff
@@ -0,0 +1,54 @@
...
+    private static SQLite.Database*[] all_dbs;
+
+    /// Close and destroy all database handles
+    static ~this ()
+    {
+        foreach (db; all_dbs)
+            db.close();
+
+        foreach (db; all_dbs)
+            destroy(*db);
```
AndrejMitrovic: would be good to check if this is really necessary or if `close()` is enough.

source/agora/common/Database.d
AndrejMitrovic: Missing param docs too.

source/agora/common/TransactionPool.d
```diff
@@ -102,7 +102,7 @@ public class TransactionPool
             log.info("Loading database from: {}", db_path);
 
         // note: can fail. we may want to just recover txes from the network instead.
```
AndrejMitrovic, line 108: Oh. I think we should examine this again. Getting the tx's other nodes' pools would be a useful feature.

source/agora/common/Database.d
```diff
@@ -0,0 +1,54 @@
...
+
+*******************************************************************************/
+
+module agora.common.Database;
+
+import SQLite = d2sqlite3.database;
+import d2sqlite3.sqlite3;
+
+/// Ditto
+public class Database
```
AndrejMitrovic: Consider renaming this to something like `ManagedDatabase`. We don't want to accidentally instantiate the wrong type of database because of a wrong import (d2sqlite vs agora import).

source/agora/consensus/UTXOSet.d
```diff
@@ -279,7 +266,7 @@ private class UTXODB
         // todo: can fail. we would have to recover by either:
```
AndrejMitrovic, line 234: same comment as before, turn into an issue pls

source/agora/consensus/validation/Block.d
```diff
@@ -287,12 +287,10 @@ unittest
     import agora.consensus.data.Enrollment;
     import agora.consensus.data.Transaction;
     import agora.consensus.UTXOSet;
-
     import std.algorithm;
     import std.range;
 
-    auto utxo_set = new UTXOSet(":memory:");
-    scope (exit) utxo_set.shutdown();
+    auto utxo_set = new UTXOSet(":memory:");  // todo: replace
```
AndrejMitrovic: Oh, we can use `TestUTXO` here now thanks to Mathias. Let's fix this up.

AndrejMitrovic:
Fixed comments.

AndrejMitrovic:
I'll resolve conflicts, and also I'll add another commit adding a `@trusted` wrapper around `execute` so we can finally stop writing `@trusted` delegates everywhere.

AndrejMitrovic:
Updated.

All the call sites should now be `@safe`.

AndrejMitrovic:
Sorry, pushed a few more @trusted => @safe just now.

AndrejMitrovic:
Actually I'm not so sure, I think this might hide if the delegate itself is not safe.

source/agora/common/ManagedDatabase.d
```diff
@@ -0,0 +1,106 @@
...
+            T = inferred from the optionally provided delegate, its return type
+            sql = the sql statement to execute
+            dg = if provided, a delegate which will be called on the ResultRange
+                 of the SQLite execute() result. If the delegate has a return
+                 value, it will be returned in the execute() call.
+            args = any arguments to bind to
+
+    ***************************************************************************/
+
+    public void execute (Args...)(string sql, auto ref Args args) @trusted
```
Geod24: Yeah that's a `@safe` violation alright.

source/agora/common/ManagedDatabase.d
```diff
@@ -0,0 +1,106 @@
...
+
+    /// Ditto
+    public auto execute (T, Args...)(string sql, auto ref Args args,
+        T delegate(ResultRange) dg) @trusted
+    {
+        return dg(this.database.execute(sql, args));
+    }
+
+    /// Ditto
+    public auto execute (T)(string sql, T delegate(ResultRange) dg) @trusted
```
Geod24: So is this

source/agora/common/ManagedDatabase.d
AndrejMitrovic: Yeah I realized this as soon as I made the PR.

AndrejMitrovic: s/ made the PR / added the commit

AndrejMitrovic:
Reverted last commit. Even with workarounds it is not possible to make the calling delegate safe. The issue is that every public function in `ResultRange` is `@system`. tl;dr we should probably not use that binding in the future..

AndrejMitrovic:
On the plus side I grew a few gray hairs and filed a new bug https://issues.dlang.org/show_bug.cgi?id=20912

AndrejMitrovic:
Rebased.

source/agora/common/ManagedDatabase.d
```diff
@@ -0,0 +1,69 @@
...
+    {
+        foreach (db; thread_dbs)
+            db.close();
+    }
+
+    /// Pointer to the database handle
+    private Database* database;
+
+    /// Subtype
+    public alias getDatabase this;
```
AndrejMitrovic, line 51: I intend to replace this with just an `execute` wrapper which handles all the various SQLite error codes, and maybe calls an error delegate or something like that to make the calling code `nothrow`. But I'll add this to another PR.

source/agora/common/ManagedDatabase.d
```diff
@@ -0,0 +1,69 @@
...
+public class ManagedDatabase
+{
+    /// ManagedDatabase managed by the current thread
+    private static Database*[] thread_dbs;
+
+    /// Close all database handles
+    static ~this ()
+    {
+        foreach (db; thread_dbs)
+            db.close();
```
TrustHenry: If a SqliteException occurs here, how will it be handled?

AndrejMitrovic: The application will exit with an error code and a stack trace will be emitted to the console.

It might not be a bad idea to log this. I'll update.

AndrejMitrovic: But in practice there's nothing much more we can do other than to log the failure.

source/agora/common/ManagedDatabase.d
```diff
@@ -0,0 +1,81 @@
...
+        Params:
+            path = path to the db on disk, or :memory: for an in-memory database
+            flags = SQLite read / write flags
+
+    ***************************************************************************/
+
+    public this (string path,
+        int flags = SQLITE_OPEN_READWRITE | SQLITE_OPEN_CREATE)
+    {
+        this.database = new Database(path, flags);
```
TrustHenry, line 66: I think it's good to catch SqliteException here, too
"Error opening the database"

TrustHenry (approved):
Looks good to me.
We have expectations that a safe database will be created for us. :)

source/agora/common/ManagedDatabase.d
AndrejMitrovic: And do what?

TrustHenry: Done!

AndrejMitrovic:
> We have expectations that a safe database will be created for us. :)

Handling error codes will be added later. This class is just a first step. What we need is:
- Better handling of shutdown (we no longer need scope(exit)), this is what this PR does.
- Atomic commits of everything in addValidatedBlock (see https://github.com/bpfkorea/agora/pull/938 how this class is modified)
- Better error handling so we can use `nothrow`. This will be added later. This also requires adding some more error code support in d2sqlite (see also https://github.com/biozic/d2sqlite3/pull/58)

Geod24 (approved):
I think the static part and the actual class should be more cleanly separated, but seeing you have other plan, this is already a big improvement in itself.

AndrejMitrovic:
Thanks, appreciate it!

PR #938, opened 2020-06-08
Use atomic transactions when altering databases

AndrejMitrovic:
Although it includes unittests it doesn't include an integration test. I would like to test that crashing the node while it's busy in `addValidatedBlock` would ensure the Node's databases don't get corrupted. In theory this should just work, but a test would be nice..

### DIFF
--- a/source/agora/common/ManagedDatabase.d
+++ b/source/agora/common/ManagedDatabase.d
@@ -21,6 +21,8 @@ import agora.utils.Log;
 import d2sqlite3.database;
 import d2sqlite3.sqlite3;
 
+import core.stdc.stdlib : abort;
+
 mixin AddLogger!();
 
 /// Ditto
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
I should mention: I had an alternative implementation where each class would have `begin / commit / rollback`, but then I realized there is a lot of code duplication everywhere. It's easier if we just used a managed database class that keeps track of all database handles and then use them all atomically via SQL transactions.

TrustHenry:
I saw it in Forum.dlang. How about this?
Should we always consider `commit` timing?
```
void transaction(Callable)(Callable c) if(isSomeFunction!Callable) {
  db.begin();
  scope(success) db.commit();
  scope(failure) db.rollback();
  c();
 }
```

AndrejMitrovic:
> Should we always consider commit timing?

Commit should only be issued once we're done with all the databases. Additionally if you look closer, I commit as soon as all the databases were modified and before `readLastBlock` is called or `onValidatorsChanged` is called. In case either of these two calls fail I don't want to roll back the UTXO set. So here `scope (success)` would be inappropriate.

source/agora/common/ManagedDatabase.d
```diff
@@ -78,4 +80,81 @@ public class ManagedDatabase
...
+    {
+        foreach (db; thread_dbs)
+        {
+            try
+            {
+                db.begin();
+            }
+            catch (Exception exc)
+            {
+                // in most cases this would only trigger if there was a
```
AndrejMitrovic, line 112: I will fix this up with the work that's going on in https://github.com/bpfkorea/agora/issues/836. I need all the SQLite errors to properly be handled in the binding, and then consider which cases here are fatal /  could be retried.

source/agora/common/ManagedDatabase.d
```diff
@@ -78,4 +80,81 @@ public class ManagedDatabase
...
+        Begins a batch operation on all the open databases.
+
+        A beginBatch() call must be followed by either a commitBatch() or
+        a rollback() call (in case of exceptions).
+
+        The rollback() call will cancel any transactions started after the
+        beginBatch() call.
+
+        In case of crashes it is not necessary to call rollback() because
+        none of the transactions were COMMITED, and since the application
```
AndrejMitrovic:
```suggestion
        none of the transactions were commited, and since the application
```

Geod24:
Regarding commit 1: Working on eliminating this function
Regarding commit 2: Doesn't that just switch the problem ? Those two operations shouldn't affect each other.

The rest, needs a bit more review (today).

AndrejMitrovic:
> Regarding commit 2: Doesn't that just switch the problem ? Those two operations shouldn't affect each other.

You explained to me IRL that we should always store a block regardless if adding stuff to the UTXO set fails?

Geod24 (approved):

Geod24:
Needs a rebase tho

PR #941, opened 2020-06-08
Share consensus-critical constants through parameter

linked0:
For unit tests, we need to modify the protocol-level constants because running the network based on the defined constants like`ValidatorCycle`, or 1008 make unit tests very time-consuming and error-prone, so we need to make the contents variable in unit tests. 

I applied the concept of passing parameters to the code, so we can set the constants when calling the `runNode` function for the real environment or calling the `makeTestNetwork` function for test environments.

Relates to #894

### DIFF
--- a/source/agora/consensus/EnrollmentPool.d
+++ b/source/agora/consensus/EnrollmentPool.d
@@ -266,7 +266,8 @@ public class EnrollmentPool
 
 version (unittest)
 private Enrollment createEnrollment(const ref Hash utxo_key,
-    const ref KeyPair key_pair, ref Scalar random_seed_src)
+    const ref KeyPair key_pair, ref Scalar random_seed_src,
+    uint validator_cycle)
 {
     import std.algorithm;
 
@@ -280,7 +281,7 @@ private Enrollment createEnrollment(const ref Hash utxo_key,
     auto signature_noise = Pair.random();
 
     enroll.utxo_key = utxo_key;
-    enroll.cycle_length = Enrollment.ValidatorCycle;
+    enroll.cycle_length = validator_cycle;
     preimages ~= hashFull(random_seed_src);
     foreach (i; 0 ..  enroll.cycle_length-1)
         preimages ~= hashFull(preimages[i]);
@@ -294,6 +295,7 @@ private Enrollment createEnrollment(const ref Hash utxo_key,
 unittest
 {
     import agora.common.Amount;
+    import agora.consensus.data.ConsensusParams;
     import agora.consensus.data.Transaction;
     import agora.consensus.Genesis;
     import std.algorithm;
@@ -324,10 +326,12 @@ unittest
     Hash[] utxo_hashes = storage.keys;
 
     // add enrollments
+    auto params = new immutable(ConsensusParams)();
     Scalar[Hash] seed_sources;
     auto utxo_hash = utxo_hashes[0];
     seed_sources[utxo_hash] = Scalar.random();
-    auto enroll = createEnrollment(utxo_hash, key_pair, seed_sources[utxo_hash]);
+    auto enroll = createEnrollment(utxo_hash, key_pair, seed_sources[utxo_hash],
+        params.ValidatorCycle);
     assert(pool.add(enroll, &storage.findUTXO));
     assert(pool.count() == 1);
     assert(pool.hasEnrollment(utxo_hash));
@@ -335,13 +339,15 @@ unittest
 
     auto utxo_hash2 = utxo_hashes[1];
     seed_sources[utxo_hash2] = Scalar.random();
-    auto enroll2 = createEnrollment(utxo_hash2, key_pair, seed_sources[utxo_hash2]);
+    auto enroll2 = createEnrollment(utxo_hash2, key_pair, seed_sources[utxo_hash2],
+        params.ValidatorCycle);
     assert(pool.add(enroll2, &storage.findUTXO));
     assert(pool.count() == 2);
 
     auto utxo_hash3 = utxo_hashes[2];
     seed_sources[utxo_hash3] = Scalar.random();
-    auto enroll3 = createEnrollment(utxo_hash3, key_pair, seed_sources[utxo_hash3]);
+    auto enroll3 = createEnrollment(utxo_hash3, key_pair, seed_sources[utxo_hash3],
+        params.ValidatorCycle);
     assert(pool.add(enroll3, &storage.findUTXO));
     assert(pool.count() == 3);
 
--- a/source/agora/consensus/EnrollmentPool.d
+++ b/source/agora/consensus/EnrollmentPool.d
@@ -320,7 +320,7 @@ unittest
         tx.inputs[0].signature = signature;
         storage.put(tx);
     }
-    EnrollmentPool pool = new EnrollmentPool(":memory:");
+    auto pool = new EnrollmentPool(":memory:");
     Hash[] utxo_hashes = storage.keys;
 
     // add enrollments
--- a/source/agora/consensus/data/ConsensusParams.d
+++ b/source/agora/consensus/data/ConsensusParams.d
@@ -1,0 +1,50 @@
+/*******************************************************************************
+
+    The set for consensus-critical constants
+
+    This defines the class for the consensus-critical constants. Only one
+    object should exist for a single node. The `class` is `immutable`, hence
+    the constants need to be set at the start of the process. The
+    consensus-critical constants are the protocol-level constants, so they
+    shouldn't be modified outside of test environments.
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.ConsensusParams;
+
+/// Ditto
+public immutable class ConsensusParams
+{
+    /// The cycle length for a validator
+    public uint ValidatorCycle;
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            validator_cycle = cycle length for a validator
+
+    ***************************************************************************/
+
+    public this (uint validator_cycle = 1008)
+    {
+        this.ValidatorCycle = validator_cycle;
+    }
+}
+
+unittest
+{
+    auto params = new immutable(ConsensusParams)(100);
+    assert(params.ValidatorCycle == 100);
+
+    auto params2 = new immutable(ConsensusParams)();
+    assert(params2.ValidatorCycle == 1008);
+}
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -23,6 +23,7 @@ import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.ConsensusData;
 import agora.consensus.data.Enrollment;
+import agora.consensus.data.ConsensusParams;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
 import agora.consensus.UTXOSet;
@@ -68,6 +69,9 @@ public class Ledger
     /// a block was externalized
     private void delegate () nothrow @safe onValidatorsChanged;
 
+    /// Parameters for consensus-critical constants
+    private immutable(ConsensusParams) params;
+
     /***************************************************************************
 
         Constructor
@@ -78,6 +82,7 @@ public class Ledger
             storage = the block storage
             enroll_man = the enrollmentManager
             node_config = the node config
+            params = the consensus-critical constants
             onValidatorsChanged = optional delegate to call after the validator
                                   set changes when a block was externalized
 
@@ -88,6 +93,7 @@ public class Ledger
         IBlockStorage storage,
         EnrollmentManager enroll_man,
         NodeConfig node_config,
+        immutable(ConsensusParams) params,
         void delegate () nothrow @safe onValidatorsChanged = null)
     {
         this.pool = pool;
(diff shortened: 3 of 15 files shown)


source/agora/consensus/data/Params.d
```diff
@@ -0,0 +1,66 @@
...
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            validator_cycle = cycle length for a validator
+
+    ***************************************************************************/
+
+    public this (uint validator_cycle)
```
linked0:
```suggestion
    public this (uint validator_cycle = 1008)
```

source/agora/consensus/data/Params.d
```diff
@@ -0,0 +1,66 @@
...
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.Params;
+
+class Params
+{
+    /// The cycle length for a validator
+    public immutable uint ValidatorCycle;
```
linked0:
```suggestion
    public immutable uint ValidatorCycle = 1008;
```

source/agora/consensus/data/Params.d
```diff
@@ -0,0 +1,66 @@
...
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.Params;
+
+class Params
```
linked0:
```suggestion
public immutable class Params
```

source/agora/node/Validator.d
```diff
@@ -52,10 +53,10 @@ public class Validator : FullNode, API
     protected Nominator nominator;
 
     /// Ctor
-    public this (const Config config)
+    public this (const Config config, const Params params = null)
```
linked0:
```suggestion
    public this (const Config config, const ConsensusParams cparams)
```

source/agora/test/Base.d
```diff
@@ -821,14 +826,15 @@ public struct TestConf
...
+        params = the consensus-critical constants
 
     Returns:
         The set of public key added to the node
 
 *******************************************************************************/
 
 public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)(
-    in TestConf test_conf)
+    in TestConf test_conf, in Params params = null)
```
AndrejMitrovic: `TestConf` is already supposed to carry all test-related configuration. So I suggest moving `Params` inside `TestConf`.

AndrejMitrovic:
If you make it a `struct`, then you won't have to do `cast(immutable)` when passing it across threads, and you won't have to do `assert(params !is null)` if you pass the struct by `const ref`.

That's just one idea though.

source/agora/test/Base.d
linked0: > `TestConf` is already supposed to carry all test-related configuration. So I suggest moving `Params` inside `TestConf`.

I'll check your suggestion. Thanks.

linked0:
> If you make it a `struct`, then you won't have to do `cast(immutable)` when passing it across threads, and you won't have to do `assert(params !is null)` if you pass the struct by `const ref`.

The original idea is to have only one object over a node. But as you suggested, `struct` could be used for that with `const ref`. I will think about it.

linked0:
After a self-review, I will request reviews.

AndrejMitrovic:
If you make it a struct it becomes easy to construct:

```
struct ConsensusParams
{
    const uint cycle_length = 1008;
    immutable Block gen_block = UnitTestGenesisBlock;
}

ConsensusParams params = { cycle_length : 10 }  // just override cycle
ConsensusParams params = { gen_block : makeGenBlock(...) }  // just override genesis block
```

This is better than `auto params = new ConsensusParams(10, 300, 12345)` because it makes it easy to determine what we're initializing, and it also makes it possible to leave all other parameters be the default except the ones we want to manipulate in the tests (like shown in the example above).

This also makes it easy to pass this type to the Node's constructor even if it's running in a different thread. 

And also we should never cast objects to immutable.

AndrejMitrovic:
Also you could make it a default parameter, leading to less code changed. `ConsensusParams cparams = ConsensusParams.init` for example.

AndrejMitrovic:
And it's ok to have it as a default parameter like this because the defaults will be used in production for CoinNet.

AndrejMitrovic:
@Geod24 I'd like to hear your opinion here too before anything is changed.

linked0:
@AndrejMitrovic Thank you for your suggestions. I'll modify as you suggested after getting @Geod24 's opinion.

> This also makes it easy to pass this type to the Node's constructor even if it's running in a different thread.

I'm not sure I follow you. I know there can be several Nodes separately in each thread and `ConsensusParams` objects exist separately in each thread. Can you explain why your suggested code makes the passing the parameter easy even if a node is running in a different thread?

AndrejMitrovic:
> I'm not sure I follow you. I know there can be several Nodes separately in each thread and ConsensusParams objects exist separately in each thread.

No they don't. You are casting the object to immutable, and passing a pointer to it to all other threads.

AndrejMitrovic:
You could do `new immutable(ConsensusParams)(...)` and that would be safe.

Geod24:
Or use an `immutable class`.

Geod24:
I think either way (class or struct) makes sense, depending on how you approach the problem.
However I think what is important is to have fields (or the `class` instance if we use a `class`) be at least `const` (needs to be `immutable` for the `class`).

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -0,0 +1,62 @@
+/*******************************************************************************
+
+    Parameter set for consensus-critcal constants
+
+    FIXME: Modfiy the description
```
Geod24: 👋

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -0,0 +1,62 @@
+/*******************************************************************************
+
+    Parameter set for consensus-critcal constants
```
Geod24:
```suggestion
    Parameter set for consensus-critical constants
```

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -0,0 +1,62 @@
...
+
+        Params:
+            validator_cycle = cycle length for a validator
+
+    ***************************************************************************/
+
+    public this (uint validator_cycle)
+    {
+        this.ValidatorCycle = validator_cycle;
+    }
```
Geod24: As discussed, if we use a class, we can achieve `struct`-like default constructor by doing:
```suggestion
    /***************************************************************************

        Constructor

        Params:
            validator_cycle = cycle length for a validator

    ***************************************************************************/

    public this (uint validator_cycle = 1008)
    {
        this.ValidatorCycle = validator_cycle;
    }
```

linked0:
Applying class for the `ConsensusParams` would be better considering the concept of the consensus-critical constants because we intend to have only one object and share that in the network although we can do that with `struct`.

I will apply `immutable` for the class.

linked0:
Ready for review.

source/agora/test/NetworkManager.d
```diff
@@ -98,9 +100,10 @@ unittest
     static class BadAPIManager : TestAPIManager
     {
         ///
-        public this (immutable(Block)[] blocks)
+        public this (immutable(Block)[] blocks,
+            immutable(ConsensusParams) params = null)
```
AndrejMitrovic:
```suggestion
            immutable(ConsensusParams) params)
```

source/agora/test/Base.d
```diff
@@ -986,11 +993,14 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         break;
     }
 
+    immutable const_params =
+        (params !is null) ? params : new immutable(ConsensusParams)();
```
AndrejMitrovic: Technically this is not const anymore. Maybe:

```suggestion
    immutable cons_params =
        (params !is null) ? params : new immutable(ConsensusParams)();
```

So it refers to consensus, rather than "constant".

linked0: Right, it should have the meaning of consensus.

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -0,0 +1,49 @@
...
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.ConsensusParams;
+
+public immutable class ConsensusParams
```
AndrejMitrovic, line 23:
```suggestion
/// Ditto
public immutable class ConsensusParams
```

So it appears in the documentation.

linked0:
I applied all the suggestions, so ready for review.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -41,19 +42,24 @@ public class EnrollmentPool
     /// SQLite DB instance
     private ManagedDatabase db;
 
+    /// Parameters for consensus-critical constants
+    private immutable(ConsensusParams) params;
```
AndrejMitrovic: I think it's not needed in this class. I can't think of any consensus parameters that would affect how the pool behaves.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -320,28 +327,31 @@ unittest
...
+    auto pool = new EnrollmentPool(":memory:", new immutable(ConsensusParams)());
     Hash[] utxo_hashes = storage.keys;
 
     // add enrollments
     Scalar[Hash] seed_sources;
     auto utxo_hash = utxo_hashes[0];
     seed_sources[utxo_hash] = Scalar.random();
-    auto enroll = createEnrollment(utxo_hash, key_pair, seed_sources[utxo_hash]);
+    auto enroll = createEnrollment(utxo_hash, key_pair, seed_sources[utxo_hash],
+        pool.params.ValidatorCycle);
```
AndrejMitrovic: Yeah this is unrelated to the pool. The longevity of a validator's cycle is unrelated to how long an enrollment stays in the pool.

source/agora/node/Ledger.d
```diff
@@ -120,8 +127,8 @@ public class Ledger
         // we are only interested in the last 1008 blocks,
```
AndrejMitrovic, line 127:
```suggestion
        // we are only interested in the last `ValidatorCycle` blocks,
```

AndrejMitrovic: Actually there are many places in the documentation which still reference `1008` as a number. We might want to fix this in another PR though, don't worry.

source/agora/consensus/EnrollmentPool.d
linked0: The consensus-critical constants are protocol-level values, so it must affect all the objects in the network. But the pool does not use this value now. so it can be deleted. I will delete it.

source/agora/consensus/EnrollmentPool.d
linked0: I can change it with a `ConsensusParams' object. Thanks.

source/agora/node/Ledger.d
linked0: I'll make a PR for it later.

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -0,0 +1,50 @@
+/*******************************************************************************
+
+    The set for consensus-critical constants
+
+    This defines the class for the consensus-critical constants. Only one
+    object is being created and being shared in the network. It can be very
+    useful to set the constanst from the start of the network. The
+    consconsensus-critcal constants are the protocol-level contants, so it
+    is only modified for test environments.
```
Geod24:
```suggestion
    object should exists for a single node. The `class` is `immutable`, hence
    the constants needs to be set at the start of the process. The
    consensus-critical constants are the protocol-level constants, so they
    shouldn't be modified outside of for test environments.
```

Please re-read documentation. Any simple spell-checker would catch typos such as "constanst", "consconsensus", "critcal" and "contents". A couple typos here and there happens, but that's a lot.

Geod24 (changes requested):
Nice job. Just a few nitpicks and we're good, + what @AndrejMitrovic said. Additionally regarding the commit, the second should be named "Propagate ConsensusParams through the code" because that's what it does (it doesn't actually *use* it) and the third one could be named "Switch ValidatorCycle from constant to ConsensusParams".

Geod24:
Note: I had some outdated comments from a previously uncompleted review. Ignore those.

source/agora/consensus/data/ConsensusParams.d
linked0: > Please re-read documentation. Any simple spell-checker would catch typos such as "constanst", "consconsensus", "critcal" and "contents". A couple typos here and there happens, but that's a lot.

Sorry for the many typos. You had already suggested a plug-in for VS Code but it doesn't work. I will find other plug-in or the way I don't make typos. Thanks.

linked0:
Ready for review.

source/agora/consensus/data/ConsensusParams.d
AndrejMitrovic: > struct-like default constructor

I think the class is only going to become awkward to use as we add more parameters to it.

For example let's say we have:

```
this (uint validator_cycle = 1008, uint tx_in_blocks = 8, uint min_validators = 3)
```

What if I want to instantiate `ConsensusParams` and only change `min_validators` and leave the rest to their defaults? I would have to write code like this:

```D
const def = new ConsensusParams();
auto actual_cparams = new immutable ConsensusParams(def.ValidatorCycle, def.TxInBlocks, 10);
```

If it was a struct all I would have to do is:

```D
ConsensusParams params = { min_validators : 10 }
```

It's more readable and composable.

Geod24 (approved):
LGTM. @AndrejMitrovic do you want to have a last look ? Otherwise ready for merging.

PR #944, opened 2020-06-08
Slightly improve well-known keys utilities

Geod24:
See each commit.

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -364,7 +364,7 @@ public struct WK
         @disable public this ();
 
         /// Given a keypair, return a name
-        public static string opIndex (const KeyPair kp)
+        public static string opIndex (const KeyPair kp) @safe nothrow @nogc
         {
             if (auto name = kp in nameMap)
                 return *name;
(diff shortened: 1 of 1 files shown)


source/agora/utils/Test.d
```diff
@@ -363,6 +363,31 @@ public struct WK
...
+            }
+
+            return Range(0);
+        }
+
+        /// Returns: The `KeyPair` matching this `pubkey`, or `KeyPair.init`
+        public static KeyPair opIndex (PublicKey pubkey) @safe nothrow @nogc
+        {
+            auto result = this.byRange.find!(k => k.address == pubkey);
+            return result.empty ? KeyPair.init : result.front();
```
AndrejMitrovic, line 388: Should this be an assert? or is it expected that some keys will be missing?

Geod24: Yeah we need to have a way to make it optional. I was thinking about `opIn` but it's a bit painful, because it returns a pointer, so no `immutable` => mutable.
So this was the compromise I found. It slightly violates what `opIndex` is supposed to do, but 🤷

AndrejMitrovic (approved):

PR #948, opened 2020-06-09
Refactor ValidatorSet

AndrejMitrovic:
See each commit. I put a lot of effort into making the commits as atomic as possible.

Part of #836

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -398,22 +398,22 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Get validator's pre-image from the validator set
+        Get validator's pre-image from the validator set.
 
         Params:
             enroll_key = The key for the enrollment in which the pre-image is
                 contained.
-            result_image = will contain the PreImageInfo if exists
 
         Returns:
-            true if getting pre-image is successfully processed
+            the PreImageInfo of the enrolled key if it exists,
+            otherwise PreImageInfo.init
 
     ***************************************************************************/
 
-    public bool getValidatorPreimage (const ref Hash enroll_key,
-        out PreImageInfo result_image) @trusted nothrow
+    public PreImageInfo getValidatorPreimage (const ref Hash enroll_key)
+        @trusted nothrow
     {
-        return this.validator_set.getPreimage(enroll_key, result_image);
+        return this.validator_set.getPreimage(enroll_key);
     }
 
     /***************************************************************************
@@ -888,15 +888,12 @@ unittest
     assert(man.createEnrollment(utxo_hash, 1, enroll));
     assert(man.pool.add(enroll, &storage.findUTXO));
 
-    PreImageInfo result_image;
     assert(Enrollment.ValidatorCycle - 101 == 907); // Sanity check
-    assert(man.getValidatorPreimage(utxo_hash, result_image));
-    assert(result_image == PreImageInfo.init);
+    assert(man.getValidatorPreimage(utxo_hash) == PreImageInfo.init);
     auto preimage = PreImageInfo(utxo_hash, man.cycle.preimages[100], 907);
     assert(man.addValidator(enroll, 2, &storage.findUTXO) is null);
     assert(man.addPreimage(preimage));
-    assert(man.getValidatorPreimage(utxo_hash, result_image));
-    assert(result_image == preimage);
+    assert(man.getValidatorPreimage(utxo_hash) == preimage);
 }
 
 /// Test `PreImageCycle` consistency between seeds and preimages
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -275,20 +275,19 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Get all the current validators
+        Get all the enrolled validator's UTXO keys.
 
         Params:
-            validators = will be filled with all the validators during
-                their validation cycles
+            utxo_keys = will contain the set of UTXO keys
 
         Returns:
-            Return true if there is no error in getting validators
+            Return true if there was no error in getting the UTXO keys
 
     ***************************************************************************/
 
-    public bool getValidators (out Enrollment[] validators) @safe nothrow
+    public bool getEnrolledUTXOs (out Hash[] keys) @safe nothrow
     {
-        return this.validator_set.getValidators(validators);
+        return this.validator_set.getEnrolledUTXOs(keys);
     }
 
     /***************************************************************************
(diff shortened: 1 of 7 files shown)


source/agora/consensus/ValidatorSet.d
```diff
@@ -436,86 +406,88 @@ public class ValidatorSet
...
 
         Returns:
-            true if the pre-image information has been added to the validator
+            Return true if the preimage was found
 
     ***************************************************************************/
 
-    public bool addPreimage (const ref PreImageInfo preimage) @safe nothrow
+    private bool getOlderPreimageDistance (Hash enroll_hash, ushort new_distance,
+        out Hash preimage, out ushort distance) @trusted nothrow
```
Geod24: You can merge this with `getPreimage` if you change the signature of `getPreimage` to:
```D
public PreImageInfo getPreimage (const ref Hash enroll_key, ushort min_distance = 0) @trusted nothrow
```

source/agora/consensus/ValidatorSet.d
```diff
@@ -436,86 +406,88 @@ public class ValidatorSet
...
+        Returns:
+            true if the pre-image information has been added to the validator
+
+    ***************************************************************************/
+
+    public bool addPreimage (const ref PreImageInfo preimage) @trusted nothrow
+    {
+        Hash prev_preimage;
+        ushort prev_distance;
+        if (!this.getOlderPreimageDistance(preimage.enroll_key,
```
AndrejMitrovic: Maybe we don't need this. Maybe we could just add a `new_distance` parameter to `getPreimage`. It would simplify the implementation.

source/agora/consensus/ValidatorSet.d
```diff
@@ -436,86 +406,88 @@ public class ValidatorSet
...
         {
             () @trusted {
-                this.db.execute("UPDATE validator_set SET preimage = ? " ~
-                    "WHERE key = ?", buffer, preimage.enroll_key.toString());
+                this.db.execute("UPDATE validator_set SET preimage = ?, " ~
+                    "distance = ? WHERE key = ?",
+                    preimage.hash.toString(), preimage.distance,
+                    preimage.enroll_key.toString());
             }();
         }
```
Geod24: Instead of doing a `SELECT` then `UPDATE`, do `UPDATE` + `WHERE` ?

source/agora/consensus/ValidatorSet.d
AndrejMitrovic: Haha I made the same comment just a minute before you. :)

source/agora/consensus/ValidatorSet.d
AndrejMitrovic: Not a bad idea.

AndrejMitrovic: Although, how would I verify the preimage hash matches? I still need to read the old distance and then call `isInvalidReason`.

Geod24: Damn, right.

source/agora/node/FullNode.d
```diff
@@ -453,9 +453,7 @@ public class FullNode : API
     /// GET: /preimage
     public override PreImageInfo getPreimage (Hash enroll_key)
     {
-        PreImageInfo preimage;
-        this.enroll_man.getValidatorPreimage(enroll_key, preimage);
-        return preimage;
+        return this.enroll_man.getValidatorPreimage(enroll_key);
```
Geod24, line 456: Technically here we should throw an Exception for the generator to pick up a non-200 is needed. But that works too.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -398,22 +398,22 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Get validator's pre-image from the validator set
+        Get validator's pre-image from the validator set.
```
Geod24, line 400: Commit message gas a typo (`PreImangeInfo`)

source/agora/consensus/ValidatorSet.d
```diff
@@ -103,16 +104,16 @@ public class ValidatorSet
...
         {
-            static ubyte[] buffer;
-            serializeToBuffer(enroll, buffer);
-            const img = PreImageInfo(enroll.utxo_key, enroll.random_seed, 0);
-            static ubyte[] pbuffer;
-            serializeToBuffer(img, pbuffer);
-
             () @trusted {
+                // todo: we probably don't need to store enrolled_height here
+                // as it's only used for *this* validator
```
Geod24: Can you expand on this ?  Enrolled height is used to tell when we need to remove a validator, isn't it ?

source/agora/consensus/ValidatorSet.d
```diff
@@ -34,6 +34,8 @@ import d2sqlite3.library;
 import d2sqlite3.results;
 import d2sqlite3.sqlite3;
 
+import ocean.text.convert.Formatter;
```
Geod24: Can we just fix the bindings instead ?

Geod24 (commented):
Broadly LGTM, except for the "Remove utxo key comparison in PreImage validation" commit. I think our validation routines should be self-consistent, and assuming a prior check is a bit odd to me.

source/agora/consensus/ValidatorSet.d
AndrejMitrovic: Ah right, brain slip. I'll remove this.

source/agora/consensus/ValidatorSet.d
AndrejMitrovic: Alright. I'll remove the last commit for now.

source/agora/node/FullNode.d
AndrejMitrovic: Hmm. It seems then we need to be more consistent. `getEnrollment` / `getMerklePath` also return `.init`. Should we file an issue for this?

AndrejMitrovic:
I'll push the changes tomorrow, I have a commit that's still on my laptop.

AndrejMitrovic:
Updated.

I haven't added `min_distance` to `getPreimage` because what we actually want is `max_distance` (we're specifically looking for an older preimage). And if I used `max_distance` it would be an awkward API because we already have `hasPreimage` that asks for a minimum distance. The two APIs would have opposite meanings of the second parameter.


However it was a micro-optimization anyway, we don't really need it.

AndrejMitrovic:
Oh but I did remove `getOlderPreimageDistance`, since I also removed the `Remove utxo key comparison in PreImage validation` commit. Also removed the string buffer optimization.

source/agora/node/FullNode.d
Geod24: Yes

Geod24 (approved):

PR #951, opened 2020-06-10
Make 'height' parameter strongly typed

Geod24:
By using a `struct` which `alias this` to the underlying (ulong) type,
we can guarantee that callers of the functions accepting `Height`
will need to provide a `Height` (or build one), while allowing
most manipulation of `Height` as a regular `ulong`.

### DIFF
--- a/source/agora/common/Types.d
+++ b/source/agora/common/Types.d
@@ -58,3 +58,13 @@ public struct QuorumConfig
     /// List of any sub-quorums
     public QuorumConfig[] quorums;
 }
+
+/// A type to ensure that height and other integer values aren't mixed
+public struct Height
+{
+    ///
+    public ulong value;
+
+    /// Provides implicit conversion to `ulong`
+    public alias value this;
+}
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -52,6 +52,7 @@ import agora.common.crypto.Schnorr;
 import agora.common.Hash;
 import agora.common.ManagedDatabase;
 import agora.common.Serializer;
+import agora.common.Types;
 import agora.consensus.data.Block;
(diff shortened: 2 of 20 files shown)


TrustHenry (approved):
It's excluded the network API and checked its applied.

source/agora/common/Types.d
```diff
@@ -58,3 +58,13 @@ public struct QuorumConfig
...
     public QuorumConfig[] quorums;
 }
+
+/// A type to ensure that height and other integer values aren't mixed
+public struct Height
+{
+    ///
+    public ulong value;
+
+    /// Provides implicit convertion to `ulong`
```
AndrejMitrovic:
```suggestion
    /// Provides implicit conversion to `ulong`
```

Geod24: Done

AndrejMitrovic:
There's a bunch of `size_t block_height` I see in `BlockStorage.d` and `NetworkManager.d`

linked0:
Could you explain the meaning of providing a `Height` instead of `ulong`?

Geod24:
It avoids issues such as mistakenly passing a distance when a height is expected, for example.

Geod24:
> There's a bunch of `size_t block_height` I see in `BlockStorage.d` and `NetworkManager.d`

Fixed.

Geod24:
This PR seems to have uncovered a bug with the Serializer. I found it, and will address it soon. If you ACK the concept (and most of the change, bar the failures / conflict), I will fix and self-merge over the weekend.

AndrejMitrovic:
ACK.

AndrejMitrovic (approved):

Geod24:
Found the bug: https://github.com/bpfkorea/agora/pull/958

Geod24:
Rebased on v0.x.x, will self merge when green.

Issue #907, opened 2020-06-02
Add Enrollments to the hardcoded genesis blocks


Geod24:
Depends on #821

Our genesis blocks need to have Enrollments in themselves in order to be valid and usable.
Currently the lack of Enrollment means we have to come up with weird workaround for our tests.
Issues such as #822 ended up being slowed down by this.

Geod24:
Not yet

AndrejMitrovic:
Solved in https://github.com/bpfkorea/agora/pull/984

AndrejMitrovic:
Not really solved.

PR #952, opened 2020-06-11
Modify `makeChainedTransactions` using `TxBuilder`

MichaelKim20:
This utility uses well-known keys (`WK.Keys`) for signature and expect
output created (via `split`, `merge`, etc...) to use them as addresses.

We used this to improve 'makeChainedTransactions'.
We haven't improved makeTransactionForFreezing yet. We're going to continue with this, too.

And we couldn't apply the new signature of the transaction. It will be applied to the next step.

Relates to #907

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -148,83 +148,104 @@ unittest
     spends the entire sum of each provided transaction's output as
     set in the parameters.
 
-    If prev_txs is null, the first set of transactions that fill a block will
-    spend the genesis transaction's outputs.
 
     Params:
+        toward = address of destination to be transferred.
         prev_txs = the previous transactions to refer to
-        key_pair = the key pair used to sign transactions and to send
-                   the output to
         block_count = the number of blocks that will be created if the
-                      returned transactions are added to the ledger
-        spend_amount = the total amount to spend (evenly distributed)
-        gen_tx = the genesis transaction to refer to for the first set of
-                 transactions. If none set, the one returned by
-                 GenesisTransaction() is used.
+            returned transactions are added to the ledger
+        share = amount to be distributed to one address
+        use_remainder = whether to use the remaining amount after distribution.
+        type = type of `Transaction`
+        unknowns = unknown `KeyPair`s.
+            finds in `unknowns` first. If not, finds it in 'WK'.
+
+    Returns:
+        returns the `Transaction`.
 
 *******************************************************************************/
 
-public Transaction[] makeChainedTransactions (KeyPair key_pair,
-    const(Transaction)[] prev_txs, size_t block_count,
-    ulong spend_amount = 40_000_000, in Transaction gen_tx = GenesisTransaction)
-    @safe
+public Transaction[] makeChainedTransactions (TxRange) (
+    scope const(PublicKey)[] toward,
+    TxRange prev_txs_range,
+    scope const size_t block_count,
+    scope const Amount share = Amount.MinFreezeAmount,
+    TxType type = TxType.Payment) @safe
+    if (isInputRange!(TxRange))
 {
     import agora.consensus.data.Block;
-    import agora.consensus.Genesis;
-    import std.conv;
 
-    assert(prev_txs.length == 0 || prev_txs.length == Block.TxsInBlock);
-    const TxCount = block_count * Block.TxsInBlock;
+    // https://issues.dlang.org/show_bug.cgi?id=20937
+    const(Transaction)[] prev_txs = () @trusted { return prev_txs_range.array; }();
 
-    // in unittests we use the following blockchain layout:
-    //
-    // genesis => 8 outputs
-    // txs[0] => spend gen_tx.outputs[0]
-    // txs[1] => spend gen_tx.outputs[1]...
-    // ..
-    // tx[9] => spend tx[0].outputs[0]
-    // tx[10] => spend tx[1].outputs[0]
-    // ..
-    // tx[17] => spend tx[9].outputs[0]
-    // tx[18] => spend tx[10].outputs[0]
-    // ..
-    // therefore the genesis block and the 1st block are unique here,
-    // as the 1st block spends all the genesis outputs via separate
-    // transactions, and subsequent blocks have transactions which
-    // spend the only outputs in the transaction from the previous block
+    // The number of prev_txs must be greater than zero and
+    // less than or equal to Block.TxsInBlock.
+    assert((prev_txs.length > 0) && (prev_txs.length <= Block.TxsInBlock));
+
+    // If the number of prev_txs is less than Block.TxsInBlock,
+    // then the number of all outputs are must be greater than or equal to Block.TxsInBlock.
+    assert(
+        (prev_txs.length == Block.TxsInBlock) ||
+        (
+            (prev_txs.length < Block.TxsInBlock) &&
+            (prev_txs.map!(tx => tx.outputs.length).sum() >= Block.TxsInBlock)
+        )
+    );
 
     Transaction[] transactions;
 
-    // always use the same amount, for simplicity
-    const Amount AmountPerTx = spend_amount / Block.TxsInBlock;
-
-    foreach (idx; 0 .. TxCount)
+    if (prev_txs.length == Block.TxsInBlock)
     {
-        Input input;
-        if (prev_txs.length == 0)  // refering to genesis tx's outputs
+        foreach (block_idx; 0 .. block_count)
         {
-            input = Input(hashFull(gen_tx), idx.to!uint);
+            foreach (tx_idx; 0 .. Block.TxsInBlock)
+            {
+                transactions ~= TxBuilder(
+                    (block_idx == 0)
+                        ? prev_txs[tx_idx]
+                        : transactions[(block_idx - 1) * Block.TxsInBlock + tx_idx]
+                    )
+                    .draw(share, toward)
+                    .sign(type);
+            }
         }
-        else  // refering to tx's in the previous block
+    }
+    else
+    {
+        TxBuilder[] builders;
+        foreach (block_idx; 0 .. block_count)
         {
-            input = Input(hashFull(prev_txs[idx % Block.TxsInBlock]), 0);
-        }
+            uint[] offsets = new uint[](Block.TxsInBlock);
 
-        Transaction tx =
-        {
-            TxType.Payment,
-            [input],
-            [Output(AmountPerTx, key_pair.address)]  // send to the same address
-        };
+            int getAvableIndex (uint last)
+            {
+                foreach_reverse (idx; 0 .. last+1)
+                    if ((idx < prev_txs.length) && (offsets[idx] < prev_txs[idx].outputs.length))
+                        return idx;
+                return -1;
+            }
 
-        auto signature = () @trusted { return key_pair.secret.sign(hashFull(tx)[]); }();
-        tx.inputs[0].signature = signature;
-        transactions ~= tx;
+            // attach
+            uint loop = 0;
+            while (true)
+            {
+                auto tx_idx = loop++ % Block.TxsInBlock;
+                auto avable_idx = getAvableIndex(tx_idx);
 
-        // new transactions will refer to the just created transactions
-        // which will be part of the previous block after the block is created
-        if ((idx > 0 && ((idx + 1) % Block.TxsInBlock == 0)))
-        {
+                if (avable_idx < 0)
+                    break;
+
+                if (builders.length <= tx_idx)
+                    builders ~= TxBuilder(prev_txs[avable_idx], offsets[avable_idx]);
+                else
+                    builders[tx_idx].attach(prev_txs[avable_idx], offsets[avable_idx]);
+                offsets[avable_idx]++;
+            }
+
+            assert(builders.length == Block.TxsInBlock);
+            builders.each!(b => transactions ~= b.split(toward).sign(type));
+            builders = null;
+
             // refer to tx'es which will be in the previous block
             prev_txs = transactions[$ - Block.TxsInBlock .. $];
         }
@@ -241,7 +262,8 @@ unittest
     auto gen_key = WK.Keys.Genesis;
 
     /// should spend genesis block's outputs
-    auto txes = makeChainedTransactions(gen_key, null, 1);
+    auto txes = makeChainedTransactions([WK.Keys.A.address],
+        genesisSpendable(), 1);
     foreach (idx; 0 .. Block.TxsInBlock)
     {
         assert(txes[idx].inputs.length == 1);
@@ -251,7 +273,7 @@ unittest
 
     auto prev_txs = txes;
     // should spend the previous tx'es outputs
-    txes = makeChainedTransactions(gen_key, txes, 1);
+    txes = makeChainedTransactions([WK.Keys.A.address], txes, 1);
 
     foreach (idx; 0 .. Block.TxsInBlock)
     {
@@ -260,15 +282,15 @@ unittest
         assert(txes[idx].inputs[0].previous == hashFull(prev_txs[idx]));
     }
 
-    const TotalSpend = 20_000_000;
-    txes = makeChainedTransactions(gen_key, prev_txs, 1, TotalSpend);
-    auto SpendPerTx = TotalSpend / Block.TxsInBlock;
+    const share = Amount(20_000_000);
+    txes = makeChainedTransactions([WK.Keys.A.address], prev_txs, 1, share);
+
     foreach (idx; 0 .. Block.TxsInBlock)
     {
         assert(txes[idx].inputs.length == 1);
         assert(txes[idx].inputs[0].index == 0);
         assert(txes[idx].inputs[0].previous == hashFull(prev_txs[idx]));
-        assert(txes[idx].outputs[0].value == Amount(SpendPerTx));
+        assert(txes[idx].outputs[0].value == share);
     }
 }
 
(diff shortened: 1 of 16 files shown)


Geod24:
The first commit seems rather independent and should be its own PR.

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
+}
 
-    Returns:
-        A newly created, valid, and signed `Transaction` spending `input`.
+/// Get an `OutputRef` from a `Transaction`
+public inout(OutputRef) oref (inout Transaction tx, uint index)
+{
+    assert(tx.outputs.length > index);
+    return inout(OutputRef)(tx, index);
+}
```
Geod24: This can be removed and replaced by just a constructor call.

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
+        const txHash = this.data.hashFull();
+        // Sign all inputs using WK keys
+        foreach (idx, ref in_; this.inputs)
+        {
+            auto ownerKP = getOwnerKeyPair(in_.output.address, unknowns);
+            assert(ownerKP !is KeyPair.init,
+                    "Address not found in Well-Known keypairs: "
+                    ~ in_.output.address.toString());
+            auto signature = () @trusted { return ownerKP.secret.sign(txHash[]); }();
+            this.data.inputs[idx].signature = signature;
```
Geod24: Merge those two statements

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
-    const resultHash = result.hashFull();
-    foreach (idx, ref in_; result.inputs)
+    private KeyPair getOwnerKeyPair (scope const PublicKey address,
+        scope const(KeyPair)[] unknowns = null) @safe
     {
-        auto rng = from.find!(a => a.address == input.outputs[idx].address);
-        assert(rng.length);
-        const owner = rng[0];
-        in_.signature = sign(owner, resultHash);
+        version(unittest)
```
Geod24: This whole module is made to be used from tests so having this special case is a bit surprising

Geod24: I see why it was introduced but I think there are better ways. Let's discuss a bit later.

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
+
+    /***************************************************************************
+
+        Attaches one output of a transaction
+
+    ***************************************************************************/
+
+    public void attach (scope const Transaction tx, uint index) @trusted
+    {
+        this.inputs ~= cast(OutputRef)oref(tx, index);
```
Geod24: Why the `cast` ?

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
+
+    /***************************************************************************
+
+        Attaches all outputs of a transaction
+
+    ***************************************************************************/
+
+    public void attach (scope const Transaction tx) @trusted
+    {
+        this.inputs ~= cast(OutputRef[])oref(tx).array;
```
Geod24: Why the `cast` ?

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
+            unknowns = unknown `KeyPair`s.
+                finds in `unknowns` first. If not, finds it in 'WK'.
+
+        Returns:
+            returns the final built `Transaction`.
+
+    ***************************************************************************/
+
+    public Transaction sign (TxType type = TxType.Payment,
+        scope const(KeyPair)[] unknowns = null) @safe
```
Geod24: Why is this needed ?

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
+        Params:
+            share = amount to be distributed to one address
+            use_remainder = whether to use the remaining amount
+                after distribution.
+            toward = address of destination to be transferred.
+
+    ***************************************************************************/
+
+    public void distribute (scope const Amount share, bool use_remainder,
+        scope const PublicKey[] toward...) @safe
```
Geod24: This offers a rather confusing API:
- In general, we want to offer a simple API that covers most needs. Users can always manually modify the tx being built in order to satisfy their specific needs;
- As a result, `use_remainder` is the first thing that stands out. Remove it;
- Likewise, why is `share` needed ? The idea is that if you want to build complex scenarii, you will chain calls to `merge` and `split`.

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
+}
+
+/// Test of TxBuilder.getOwnerKeyPair
+unittest
+{
+    TxBuilder builder;
+    KeyPair unknown1 = KeyPair.fromSeed(
+        Seed.fromString("SCLEH2TOYRNMH6NS2POFFMINKY6DYBWKIIW7A7BAZ3H5BE7KZIWUKF42"));
+    KeyPair unknown2 = KeyPair.fromSeed(
+        Seed.fromString("SDDQOCBBREPZZJDUWMZUGEUZVNTYBCF4AJPAQYIHU6TINNV7SXMKBUAJ"));
```
Geod24: Why not `random` ?

source/agora/utils/Test.d
```diff
@@ -624,62 +627,274 @@ unittest
...
+    /***************************************************************************
+
+        Splits the attached input into multiple outputs.
+
+        Params:
+            toward = address of destination to be transferred.
+
+    ***************************************************************************/
+
+    public void split (scope const PublicKey[] toward...) @safe
```
```suggestion
    public ref typeof(this) split (scope const PublicKey[] toward...) @safe return
```

This will allow us to do `auto tx = TxBuilder(...).split(...).merge(...).split(...).sign();`

Geod24 (commented):
Haven't reviewed the last commit yet.

source/agora/utils/Test.d
MichaelKim20: The following message came out.
`cannot append type const(OutputRef) to type OutputRef[]`

source/agora/utils/Test.d
MichaelKim20: The parameters unknowns are added for `system integration test` and
and `source/tests/unit/*.d`.

source/agora/utils/Test.d
MichaelKim20: I will remove

source/agora/utils/Test.d
MichaelKim20: They need a genesis KeyPair of coin-net.

source/agora/utils/Test.d
MichaelKim20: That's a good idea.

source/agora/utils/Test.d
MichaelKim20: I'll change the member like this and remove `cast`.
``` D
private const(OutputRef)[] inputs;
```

Geod24:
Could you raise a PR with the first commit only ?

source/agora/utils/Test.d
```diff
@@ -624,62 +627,309 @@ unittest
...
+    /// The index of the `Output` being referenced
+    public uint        index;
+}
+
+/// Get a range of all `OutputRef` from a `Transaction`
+public auto oref (const Transaction tx)
+{
+    import std.range;
+    return iota(tx.outputs.length)
+        .map!(idx => OutputRef(cast(Transaction)tx, cast(uint)idx));
```
Geod24: This is casting away `const` and is a big no-no.

source/agora/utils/Test.d
```diff
@@ -624,62 +627,309 @@ unittest
...
+/// A reference to an output
+/// It has a `Transaction` and an index of `Outputs` that it references.
+public struct OutputRef
+{
+    this (Transaction tx, uint index)
+    {
+        assert(tx.outputs.length > index);
+        this.tx = tx;
+        this.index = index;
+    }
```
Geod24: Let's just remove it. The only advantage of this constructor is to do bounds checking, but it will blow up when accessing `output` anyway. And if we wanted to keep it, we would need to add the attributes (`@safe pure nothrow @nogc`).

source/agora/utils/Test.d
```diff
@@ -624,62 +627,309 @@ unittest
...
+        // Finalize the transaction by adding inputs
+        foreach (ref in_; this.inputs)
+            this.data.inputs ~= Input(in_.hash(), in_.index);
+
+        // Get the hash to sign
+        const txHash = this.data.hashFull();
+        // Sign all inputs using WK keys
+        foreach (idx, ref in_; this.inputs)
+        {
+            auto ownerKP = getOwnerKeyPair(in_.output.address);
```
Geod24: Missing `this.`

source/agora/utils/Test.d
```diff
@@ -624,62 +627,309 @@ unittest
...
+    ***************************************************************************/
+
+    public ref typeof(this) merge (scope const PublicKey toward)
+        @safe return
+    {
+        if (this.inputs.length == 0)
+            return this;
+
+        Amount amount;
+        if (this.inputs.all!(val => amount.add(val.output.value)))
```
Geod24: Since those are test utilities, we shouldn't ignore failures like this.

source/agora/utils/Test.d
```diff
@@ -624,62 +627,309 @@ unittest
...
+
+        Params:
+            tx = transaction
+
+        Returns:
+            reference of this
+
+    ***************************************************************************/
+
+    public ref typeof(this) attach (scope const Transaction tx) @trusted return
```
Geod24: Why `scope` ? Why `@trusted` ?

source/agora/utils/Test.d
```diff
@@ -624,62 +627,309 @@ unittest
...
+
+        Returns:
+            reference of this
+
+    ***************************************************************************/
+
+    public ref typeof(this) attach (scope const Transaction tx, uint index)
+        @trusted return
+    {
+        this.inputs ~= OutputRef(cast(Transaction)tx, index);
```
Geod24: We should almost never need to `cast` away `const`.

Geod24 (changes requested):
Haven't finished reviewing, but looks like it's going in the right direction.
There is a misunderstanding for the transaction builder though, it should modify its current state, so `attach` should just append an identity transformation and `split` and `merge` should act on the `output`.

Geod24:
Rebased on master and applied a few of the improvement, will apply the `output` bit soon.

source/agora/utils/Test.d
```diff
@@ -146,89 +152,110 @@ unittest
...
+    else
+    {
+        auto builders = iota(Block.TxsInBlock).map!(_ => new TxBuilder(unknowns)).array;
+        foreach (block_idx; 0 .. block_count)
         {
-            input = Input(hashFull(prev_txs[idx % Block.TxsInBlock]), 0);
-        }
+            uint[] offsets;
+
+            iota(Block.TxsInBlock).each!(_ => offsets ~= 0);
```
Geod24: That is a very complicated way to write:
```suggestion
            uint[] offsets = new uint[](Block.TxsInBlock);
```

source/agora/utils/Test.d
```diff
@@ -146,89 +152,110 @@ unittest
...
+
+            int getAvableIndex (uint last)
```
Geod24, line 220:
```suggestion
            int getAvailableIndex (uint last)
```

I suppose that's what you meant ?

Geod24:
@MichaelKim20 : I have updated this with my WIP. I removed the `amount` function and added `draw(amount, PublicKey[]...)` instead, as `amount` was mostly used after `split`.
I also changed the initial implementation to use a `refund` transaction instead of always iterating on the input, so we can do things like `TxBuilder(addr).split(...).attach(...).split(...)`, etc...
However, the `split` implementation hasn't been fixed yet (will do that soon).
There is only one test which doesn't pass, because now we have more diversity in keys and we're not always using the genesis keypair. I will fix it soon too.

Geod24:
Re-updated. Changes:
- Renamed `GenesisMainTransactions` to `genesisSpendable`. Made it return a range the filters the transactions for future-proofness;
- Added an example on how one could replace `makeChainedTransactions` in `source/agora/common/TransactionPool.d`. After much thought I will just the content of `makeChainedTransaction` again, although that might take a bit of time.

Geod24:
Just rebased on top of #965 after adding range interface and there was only one usage of the VLA, yay

Geod24:
Rebased on v0.x.x, off to the next round of splitting off

source/agora/utils/Test.d
```diff
@@ -148,83 +148,104 @@ unittest
     spends the entire sum of each provided transaction's output as
     set in the parameters.
 
-    If prev_txs is null, the first set of transactions that fill a block will
-    spend the genesis transaction's outputs.
 
     Params:
+        toward = address of destination to be transferred.
```
AndrejMitrovic, line 153:
```suggestion
        toward = addresses which will receive the payments
```

source/agora/utils/Test.d
```diff
@@ -148,83 +148,104 @@ unittest
...
-        key_pair = the key pair used to sign transactions and to send
-                   the output to
         block_count = the number of blocks that will be created if the
-                      returned transactions are added to the ledger
-        spend_amount = the total amount to spend (evenly distributed)
-        gen_tx = the genesis transaction to refer to for the first set of
-                 transactions. If none set, the one returned by
-                 GenesisTransaction() is used.
+            returned transactions are added to the ledger
+        share = amount to be distributed to one address
```
AndrejMitrovic, line 157:
```suggestion
        share = amount to be distributed to each address
```

source/agora/utils/Test.d
```diff
@@ -148,83 +148,104 @@ unittest
...
+        use_remainder = whether to use the remaining amount after distribution.
```
AndrejMitrovic, line 158: What does "to use" mean? Use how?

source/agora/utils/Test.d
```diff
@@ -148,83 +148,104 @@ unittest
...
+        type = type of `Transaction`
+        unknowns = unknown `KeyPair`s.
+            finds in `unknowns` first. If not, finds it in 'WK'.
```
AndrejMitrovic, line 161: ??

source/agora/utils/Test.d
```diff
@@ -148,83 +148,104 @@ unittest
...
 *******************************************************************************/
 
-public Transaction[] makeChainedTransactions (KeyPair key_pair,
-    const(Transaction)[] prev_txs, size_t block_count,
-    ulong spend_amount = 40_000_000, in Transaction gen_tx = GenesisTransaction)
-    @safe
+public Transaction[] makeChainedTransactions (TxRange) (
+    scope const(PublicKey)[] toward,
+    TxRange prev_txs_range,
+    scope const size_t block_count,
```
AndrejMitrovic, line 171:
```suggestion
    const size_t block_count,
```

`scope` on a `size_t` does nothing.

source/agora/utils/Test.d
```diff
@@ -148,83 +148,104 @@ unittest
...
+    scope const Amount share = Amount.MinFreezeAmount,
```
AndrejMitrovic, line 172:
```suggestion
   const Amount share = Amount.MinFreezeAmount,
```

`scope` here is unnecessary.

source/agora/utils/Test.d
AndrejMitrovic: Why did you mark this comment as resolved but didn't fix the typo?

AndrejMitrovic (changes requested):
I'm not sure about this code. Somehow `makeChainedTransactions` is even more complicated than it was before. I thought the idea was that we would make code easier to understand?

AndrejMitrovic:
I've tried to simplify it a bit. There's still more work that should be done however.

```D
public Transaction[] makeChainedTransactions (TxRange) (
    scope const(PublicKey)[] toward,
    TxRange prev_txs_range,
    scope const size_t block_count,
    scope const Amount share = Amount.MinFreezeAmount,
    TxType type = TxType.Payment) @safe
    if (isInputRange!(TxRange))
{

    // https://issues.dlang.org/show_bug.cgi?id=20937
    const(Transaction)[] prev_txs = () @trusted { return prev_txs_range.array; }();

    // sanity checks
    assert(prev_txs.length > 0 && prev_txs.length <= Block.TxsInBlock);
    assert(
        (prev_txs.length == Block.TxsInBlock) ||
        (
            (prev_txs.length < Block.TxsInBlock) &&
            (prev_txs.map!(tx => tx.outputs.length).sum() >= Block.TxsInBlock)
        )
    );

    Transaction[] transactions;

    if (prev_txs.length == Block.TxsInBlock)
    {
        foreach (block_idx; 0 .. block_count)
        {
            foreach (tx_idx; 0 .. Block.TxsInBlock)
                transactions ~= TxBuilder(prev_txs[tx_idx])
                    .draw(share, toward).sign(type);

            prev_txs = transactions[$ - Block.TxsInBlock .. $];
        }
    }
    else
    {
        foreach (block_idx; 0 .. block_count)
        {
            TxBuilder[] builders;
            uint[Block.TxsInBlock] offsets;
            int getAvailableIndex (uint last)
            {
                foreach_reverse (idx; 0 .. last+1)
                    if ((idx < prev_txs.length) && (offsets[idx] < prev_txs[idx].outputs.length))
                        return idx;
                return -1;
            }

            // attach
            uint loop = 0;
            while (true)
            {
                auto tx_idx = loop++ % Block.TxsInBlock;
                auto available_idx = getAvailableIndex(tx_idx);
                if (available_idx < 0)
                    break;

                if (builders.length <= tx_idx)
                    builders ~= TxBuilder(prev_txs[available_idx], offsets[available_idx]);
                else
                    builders[tx_idx].attach(prev_txs[available_idx], offsets[available_idx]);

                offsets[available_idx]++;
            }

            assert(builders.length == Block.TxsInBlock);
            builders.each!(b => transactions ~= b.split(toward).sign(type));

            // refer to tx'es which will be in the previous block
            prev_txs = transactions[$ - Block.TxsInBlock .. $];
        }
    }

    return transactions;
}
```

Tests still pass. 

I still have to reverse-engineer the `getAvailableIndex` stuff.

Geod24:
This has been split in quite a few commits and changes, and the core that was blocking other issues have been resolved, so let's close this. Thanks a lot for your work @MichaelKim20 !

PR #953, opened 2020-06-12
Use setTimer on Nominator.setupTimer

TrustHenry:
`ITimer` interface has been newly created and this is implemented as `VibeTimer` and `LocalRestTimer`.
Currently, it only defined a stop function in the `ITimer`, which can be expanded.
Also, the `Nominator.setupTimer` uses `TaskManager`'s `setTimer` new task which waits until a timer expires and call

Relates #872

### DIFF
--- a/source/agora/common/Task.d
+++ b/source/agora/common/Task.d
@@ -73,14 +73,57 @@ public class TaskManager
             dg = This delegate will be called when the timer fires.
             periodic = Specifies if the timer fires repeatedly or only once.
 
+        Returns:
+            An `ITimer` interface with the ability to control the timer
+
     ***************************************************************************/
 
-    public void setTimer (Duration timeout, void delegate() dg,
+    public ITimer setTimer (Duration timeout, void delegate() dg,
         Periodic periodic = Periodic.No)
     {
         assert(dg !is null, "Cannot call this delegate if null");
-
         static import vibe.core.core;
-        vibe.core.core.setTimer(timeout, dg, periodic);
+        return new VibedTimer(vibe.core.core.setTimer(timeout, dg, periodic));
+    }
+}
+
+/*******************************************************************************
+
+    Defines an abstraction over the timer implementation
+
+*******************************************************************************/
+
+public interface ITimer
+{
+    /***************************************************************************
+
+        Stop or cancel this timer
+
+    ***************************************************************************/
+
+    public void stop ();
+}
+
+/*******************************************************************************
+
+    Vibe.d timer
+
+*******************************************************************************/
+
+private final class VibedTimer : ITimer
+{
+    static import Vibe = vibe.core.core;
+
+    private Vibe.Timer timer;
+
+    public this (Vibe.Timer timer) @safe nothrow
+    {
+        this.timer = timer;
+    }
+
+    /// Ditto
+    public override void stop () @safe nothrow
+    {
+        this.timer.stop();
     }
 }
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -74,11 +74,8 @@ public extern (C++) class Nominator : SCPDriver
     private alias TimerType = Slot.timerIDs;
     static assert(TimerType.max == 1);
 
-    /// Tracks unique incremental timer IDs
-    private ulong[TimerType.max + 1] last_timer_id;
-
-    /// Timer IDs with >= of the active timer will be allowed to run
-    private ulong[TimerType.max + 1] active_timer_ids;
+    /// Currently active timers grouped by type
+    private ITimer[TimerType.max + 1] active_timers;
 
     /// Whether we're in the process of nominating
     private bool is_nominating;
@@ -518,24 +515,16 @@ extern(D):
 
         const type = cast(TimerType) timer_type;
         assert(type >= TimerType.min && type <= TimerType.max);
-        if (callback is null || timeout == 0)
+        if (auto timer = this.active_timers[type])
         {
-            // signal deactivation of all current timers with this timer type
-            this.active_timer_ids[type] = this.last_timer_id[type] + 1;
-            return;
+            timer.stop();
+            this.active_timers[type] = null;
         }
 
-        const timer_id = ++this.last_timer_id[type];
-        this.taskman.runTask(
-        {
-            this.taskman.wait(timeout.msecs);
-
-            // timer was cancelled
-            if (timer_id < this.active_timer_ids[type])
-                return;
-
-            callCPPDelegate(callback);
-        });
+        if (callback is null || timeout == 0)
+            return;
+        this.active_timers[type] = this.taskman.setTimer(
+            timeout.msecs, { callCPPDelegate(callback); });
     }
 }
 
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -277,12 +277,40 @@ public class LocalRestTaskManager : TaskManager
             dg = This delegate will be called when the timer fires.
             periodic = Specifies if the timer fires repeatedly or only once.
 
+        Return:
+           An `ITimer` interface with the ability to control the timer
+
     ***************************************************************************/
 
-    public override void setTimer (Duration timeout, void delegate() dg,
+    public override ITimer setTimer (Duration timeout, void delegate() dg,
         Periodic periodic = Periodic.No)
     {
-        geod24.LocalRest.setTimer(timeout, dg, periodic);
+        return new LocalRestTimer(geod24.LocalRest.setTimer(timeout, dg,
+            periodic));
+    }
+}
+
+/*******************************************************************************
+
+    LocalRest only timer (for unittests)
+
+*******************************************************************************/
+
+private final class LocalRestTimer : ITimer
+{
+    import LocalRest = geod24.LocalRest;
+
+    private LocalRest.Timer timer;
+
+    public this (LocalRest.Timer timer)
+    {
+        this.timer = timer;
+    }
+
+    /// Ditto
+    public override void stop ()
+    {
+        this.timer.stop();
     }
 }
 


source/agora/consensus/protocol/Nominator.d
```diff
@@ -526,15 +526,36 @@ extern(D):
...
+                    timer.stop();
+                    return;
+                }
+                callCPPDelegate(callback);
+            }
 
-            callCPPDelegate(callback);
+            callDelegate(); // avoid delay
+            timer = this.taskman.setTimer(timeout.msecs,
+                &callDelegate, Periodic.Yes);
```
AndrejMitrovic: This is wrong, it was not a periodic timer. There was no infinite loop.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -526,15 +526,36 @@ extern(D):
...
+                if (timer_id < this.active_timer_ids[type])
+                {
+                    timer.stop();
+                    return;
+                }
+                callCPPDelegate(callback);
+            }
 
-            callCPPDelegate(callback);
+            callDelegate(); // avoid delay
```
AndrejMitrovic: This is also wrong. Notice that there was a `wait` and then it was called. There is a specific reason for this. These are timeout routines.

AndrejMitrovic: I'll show you exactly how to do it.

AndrejMitrovic (changes requested):

AndrejMitrovic:
This is what `setupTimer` should look like:

```D
public override void setupTimer (ulong slot_idx, int timer_type,
    milliseconds timeout, CPPDelegate!SCPCallback* callback)
{
    scope (failure) assert(0);

    const type = cast(TimerType) timer_type;
    assert(type >= TimerType.min && type <= TimerType.max);
    if (callback is null || timeout == 0)
    {
        auto timer = this.active_timers[type];
        if (timer !is null)
        {
            timer.stop();
            this.active_timers[type] = null;
        }

        return;
    }

    this.active_timers[type] = this.taskman.setTimer(
        timeout.msecs, { callCPPDelegate(callback); });
}
```

AndrejMitrovic:
And you should change `active_timers` variable to this:

```D
private ITimer[TimerType.max + 1] active_timers;
```

AndrejMitrovic:
Also remove the `last_timer_id` variable.

AndrejMitrovic:
@TrustHenry actually you will have to fix the implementation in your LocalRest PR. We need an ability to stop a timer after `sleep(timeout)` is called but before `dg()` is called.

AndrejMitrovic:
The current tests will pass with the changes I proposed, however that is only because non of the nodes every time out during consensus (there is a lack of tests).

TrustHenry:
Very sharp.
I also thought there was a problem with the timing.

TrustHenry:
> The current tests will pass with the changes I proposed, however that is only because non of the nodes every time out during consensus (there is a lack of tests).

@AndrejMitrovic could you tell me more details about this?

AndrejMitrovic:
> @AndrejMitrovic could you tell me more details about this?

This code might be a little hard to follow but let me explain:

https://github.com/bpfkorea/agora/blob/52e8aa0cfe7852954ce6430ab57cde0c153b2911/source/scpp/src/scp/BallotProtocol.cpp#L512-L515

The C++ SCP code calls the D code and asks it to set up a timer. If the timer has the chance to execute (the timeout expired and run() is called), then that run() function will call this:

https://github.com/bpfkorea/agora/blob/52e8aa0cfe7852954ce6430ab57cde0c153b2911/source/scpp/src/scp/BallotProtocol.cpp#L527-L531

But I'm not sure if this has been tested in Agora yet. It's worth looking into at some point.

TrustHenry:
This requires a localRest update.
https://github.com/Geod24/localrest/pulls

source/agora/common/Task.d
```diff
@@ -73,14 +73,80 @@ public class TaskManager
             dg = This delegate will be called when the timer fires.
             periodic = Specifies if the timer fires repeatedly or only once.
 
+        Returns:
+            A `ITimer` instance with the ability to control the VibeTimer
```
AndrejMitrovic:
```suggestion
            An `ITimer` interface with the ability to control the timer
```

source/agora/common/Task.d
```diff
@@ -73,14 +73,80 @@ public class TaskManager
...
-
         static import vibe.core.core;
-        vibe.core.core.setTimer(timeout, dg, periodic);
+        return new VibeTimer(vibe.core.core.setTimer(timeout, dg, periodic));
+    }
+}
+
+/*******************************************************************************
+
+    Define the ability to control the timer
```
AndrejMitrovic:
```suggestion
    Defines an abstraction over the timer implementation
```

source/agora/common/Task.d
```diff
@@ -73,14 +73,80 @@ public class TaskManager
...
+
+*******************************************************************************/
+
+public interface ITimer
+{
+    /***************************************************************************
+
+        Stop the timer executed by the setTimer
```
AndrejMitrovic:
```suggestion
        Stop or cancel this timer
```

source/agora/common/Task.d
```diff
@@ -73,14 +73,80 @@ public class TaskManager
...
+
+    ***************************************************************************/
+
+    public void stop();
+}
+
+/*******************************************************************************
+
+    Vibe only timer
```
AndrejMitrovic:
```suggestion
    Vibe.d timer
```

source/agora/common/Task.d
```diff
@@ -73,14 +73,80 @@ public class TaskManager
...
+
+*******************************************************************************/
+
+final class VibeTimer : ITimer
```
AndrejMitrovic:
```suggestion
private final class VibedTimer : ITimer
```

source/agora/common/Task.d
```diff
@@ -73,14 +73,80 @@ public class TaskManager
...
+    }
+}
+
+/*******************************************************************************
+
+    LocalRest only timer (for unittests)
+
+*******************************************************************************/
+
+final class LocalRestTimer : ITimer
```
AndrejMitrovic: This doesn't belong here, but in `Base.d` where it's used. And it should be `private` too.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -74,11 +74,8 @@ public extern (C++) class Nominator : SCPDriver
     private alias TimerType = Slot.timerIDs;
     static assert(TimerType.max == 1);
 
-    /// Tracks unique incremental timer IDs
-    private ulong[TimerType.max + 1] last_timer_id;
-
-    /// Timer IDs with >= of the active timer will be allowed to run
-    private ulong[TimerType.max + 1] active_timer_ids;
+    /// The active timer will be allowed to run
```
```suggestion
    /// Currently active timers grouped by type
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -520,24 +517,20 @@ extern(D):
...
-
-            // timer was cancelled
-            if (timer_id < this.active_timer_ids[type])
-                return;
-
-            callCPPDelegate(callback);
-        });
+        this.active_timers[type] = this.taskman.setTimer(
+            timeout.msecs, { callCPPDelegate(callback); });
+        }
```
```suggestion
```

was this a typo? It seems out of place.

source/agora/test/Base.d
```diff
@@ -276,12 +276,16 @@ public class LocalRestTaskManager : TaskManager
             dg = This delegate will be called when the timer fires.
             periodic = Specifies if the timer fires repeatedly or only once.
 
+        Return:
+            A `LocalRsetTimer` instance with the ability to control the timer
```
```suggestion
            An `ITimer` interface with the ability to control the timer
```

Also `LocalRsetTimer` was a typo. It would be good to catch these typos before requesting a review. An initial self-review is always good.

AndrejMitrovic (dismissed):
Just small nits.

TrustHenry:
@AndrejMitrovic Thank you for your review.
All suggestions have been applied.

source/agora/common/Task.d
```diff
@@ -73,14 +73,56 @@ public class TaskManager
             dg = This delegate will be called when the timer fires.
             periodic = Specifies if the timer fires repeatedly or only once.
 
+        Returns:
+            An `ITimer` interface with the ability to control the timer
```
Geod24, line 77:
```suggestion
            An `ITimer` interface with the ability to control the timer

```

source/agora/common/Task.d
```diff
@@ -73,14 +73,56 @@ public class TaskManager
...
+
+public interface ITimer
+{
+    /***************************************************************************
+
+        Stop or cancel this timer
+
+    ***************************************************************************/
+
+    public void stop();
```
Geod24: Vibe's Timer is `nothrow @safe`, but localrest is not annotated.
Also:
```suggestion
    public void stop ();
```

source/agora/common/Task.d
```diff
@@ -73,14 +73,56 @@ public class TaskManager
...
+
+    private Vibe.Timer timer;
+
+    public this (Vibe.Timer timer)
+    {
+        this.timer = timer;
+    }
+
+    /// Ditto
+    public override void stop()
```
Geod24:
```suggestion
    public override void stop () @safe nothrow
```

source/agora/common/Task.d
```diff
@@ -73,14 +73,56 @@ public class TaskManager
...
+
+*******************************************************************************/
+
+private final class VibedTimer : ITimer
+{
+    static import Vibe = vibe.core.core;
+
+    private Vibe.Timer timer;
+
+    public this (Vibe.Timer timer)
```
Geod24:
```suggestion
    public this (Vibe.Timer timer) @safe pure nothrow @nogc
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -520,22 +517,18 @@ extern(D):
         assert(type >= TimerType.min && type <= TimerType.max);
         if (callback is null || timeout == 0)
         {
-            // signal deactivation of all current timers with this timer type
-            this.active_timer_ids[type] = this.last_timer_id[type] + 1;
+            auto timer = this.active_timers[type];
+            if (timer !is null)
```
Geod24: You can put those two statements in one:

```suggestion
            if (auto timer = this.active_timers[type])
```

source/agora/test/Base.d
```diff
@@ -277,12 +277,40 @@ public class LocalRestTaskManager : TaskManager
...
+
+    public this (LocalRest.Timer timer)
+    {
+        this.timer = timer;
+    }
+
+    /// Ditto
+    public override void stop()
+    {
+        this.timer.stop();
```
Geod24: Same things about attributes and style

Geod24 (changes requested):

source/agora/common/Task.d
```diff
@@ -73,14 +73,57 @@ public class TaskManager
...
+
+*******************************************************************************/
+
+private final class VibedTimer : ITimer
+{
+    static import Vibe = vibe.core.core;
+
+    private Vibe.Timer timer;
+
+    public this (Vibe.Timer timer) @safe nothrow
```
Geod24, line 119: Missing `pure @nogc`

source/agora/test/Base.d
```diff
@@ -277,12 +277,40 @@ public class LocalRestTaskManager : TaskManager
...
+
+*******************************************************************************/
+
+private final class LocalRestTimer : ITimer
+{
+    import LocalRest = geod24.LocalRest;
+
+    private LocalRest.Timer timer;
+
+    public this (LocalRest.Timer timer)
```
Geod24, line 305: Missing attributes

source/agora/test/Base.d
```diff
@@ -277,12 +277,40 @@ public class LocalRestTaskManager : TaskManager
...
+    {
+        this.timer = timer;
+    }
+
+    /// Ditto
+    public override void stop ()
```
Geod24, line 311: We can proceed as is, or you can make a PR to LocalRest to fix the attributes. I'll merge it fast.

source/agora/common/Task.d
TrustHenry:
```
source/agora/common/Task.d(119,12): Error: pure constructor agora.common.Task.VibedTimer.this cannot call impure destructor vibe.core.core.Timer.~this
source/agora/common/Task.d(119,12): Error: @nogc constructor agora.common.Task.VibedTimer.this cannot call non-@nogc destructor vibe.core.core.Timer.~this
source/agora/common/Task.d(121,20): Error: pure constructor agora.common.Task.VibedTimer.this cannot call impure function vibe.core.core.Timer.__postblit
source/agora/common/Task.d(121,20): Error: @nogc constructor agora.common.Task.VibedTimer.this cannot call non-@nogc function vibe.core.core.Timer.__postblit
```

Geod24: Oh man. LGTM then.

source/agora/test/Base.d
TrustHenry: https://github.com/Geod24/localrest/blob/2260b8eae7daec3d1991a1859bec759cc729a653/source/geod24/concurrency.d#L945
The attributes are not safe. Is it intended?

Geod24 (approved):

source/agora/consensus/protocol/Nominator.d
```diff
@@ -520,22 +517,16 @@ extern(D):
...
-            this.taskman.wait(timeout.msecs);
-
-            // timer was cancelled
-            if (timer_id < this.active_timer_ids[type])
-                return;
-
-            callCPPDelegate(callback);
-        });
+        this.active_timers[type] = this.taskman.setTimer(
+            timeout.msecs, { callCPPDelegate(callback); });
```
Geod24, line 527: Looks like previously set timers aren't stopped ?

AndrejMitrovic: @TrustHenry could you also check if `this.active_timers[type]` exists here, and if it does call stop on it?

TrustHenry: In the unit test, Timer.stops confirmed a total of 106 occurrences.
This is also `if (auto timer = this.active_timers[type])`

AndrejMitrovic: I meant that we should probably do something like this:

```diff
diff --git a/source/agora/consensus/protocol/Nominator.d b/source/agora/consensus/protocol/Nominator.d
index 4a0f115e2..381c6faca 100644
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -515,16 +515,15 @@ extern(D):

         const type = cast(TimerType) timer_type;
         assert(type >= TimerType.min && type <= TimerType.max);
-        if (callback is null || timeout == 0)
+        if (auto timer = this.active_timers[type])
         {
-            if (auto timer = this.active_timers[type])
-            {
-                timer.stop();
-                this.active_timers[type] = null;
-            }
-            return;
+            timer.stop();
+            this.active_timers[type] = null;
         }

+        if (callback is null || timeout == 0)
+            return;
+
         this.active_timers[type] = this.taskman.setTimer(
             timeout.msecs, { callCPPDelegate(callback); });
     }
```

So always stop any existing timers. And then if callback is null we just return, otherwise we set a timer.

TrustHenry: Was there a leak here?
The Code of @AndrejMitrovic  applied successfully.

TrustHenry: If this callback is not null?

AndrejMitrovic: It's possible.

Geod24 (approved):

PR #954, opened 2020-06-12
Adapting CLI to parse consensus parameters

### LINKED ISSUE #894: Turn necessary consensus-critical constant into argument

linked0:
After making the consensus-critical constants to be propagated through parameters, we adapt the CLI to parse the constants.

Fixes #894


source/agora/cli/multi/main.d
```diff
@@ -106,7 +106,11 @@ private int main (string[] args)
 
     FullNode[] nodes;
     foreach (const ref config; configs)
-        nodes ~= runNode(config, new immutable(ConsensusParams)());
+    {
+        auto params = new immutable(ConsensusParams)(
+            config.node.validator_cycle);
+        nodes ~= runNode(config, params);
```
Geod24, line 112: Why the separation though ?

Geod24:
I have a 99.99% certitude this doesn't work.

source/agora/cli/multi/main.d
linked0: Not much meaning. It's just for more readable code.

linked0:
> I have a 99.99% certitude this doesn't work.

I found the missing part. Sorry for that.

linked0:
Ready for review.

source/agora/common/Config.d
```diff
@@ -291,6 +294,7 @@ private NodeConfig parseNodeConfig (Node* node, const ref CommandLine cmdln)
 
     string data_dir = get!(string, "node", "data_dir")(cmdln, node);
     auto port = get!(ushort, "node", "port")(cmdln, node);
+    auto validator_cycle = get!(uint, "node", "validator_cycle")(cmdln, node);
```
Geod24, line 297: This needs to be optional

linked0: > This needs to be optional

How about this?
```
auto value = get!(uint, "node", "validator_cycle")(cmdln, node);
auto validator_cycle = (value != 0) ? value : NodeConfig.init.validator_cycle;
```

Geod24: There is `opt` for this

Geod24: Ah, my bad, I just realized that `get` also uses the default value.

Geod24:
Reverted to a previous version (my feedback was wrong), will merge when green.

Geod24 (approved):

PR #965, opened 2020-06-16
Implement a helper struct to build transactions

Geod24:
```
This utility uses well-known keys (`WK.Keys`) for signature and expect
output created (via `split`, `merge`, etc...) to use them as addresses.
It works by adding all inputs to a "refund" address and having operations,
such as `draw` or `split`, work on this refund address.
```

Split off #952 for easier reviewing.

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -31,14 +31,19 @@ module agora.utils.Test;
 
 import agora.common.Amount;
 import agora.common.crypto.Key;
+import agora.common.crypto.Schnorr;
 import agora.common.Hash;
+import agora.common.Serializer;
 import agora.common.Types;
 import agora.consensus.data.Transaction;
 import agora.consensus.Genesis;
 
 import std.algorithm;
+import std.array;
 import std.file;
+import std.format;
 import std.path;
+import std.range;
 
 import core.exception;
 import core.time;
@@ -101,9 +106,7 @@ public void retryFor (Exc : Throwable = AssertError) (lazy bool check,
     Duration timeout, lazy string msg = "",
     string file = __FILE__, size_t line = __LINE__)
 {
-    import core.exception;
     import core.thread;
-    import std.format;
 
     // wait 100 msecs between attempts
     const SleepTime = 100;
@@ -130,7 +133,6 @@ public void retryFor (Exc : Throwable = AssertError) (lazy bool check,
 ///
 unittest
 {
-    import core.exception;
     import std.exception;
 
     static bool willSucceed () { static int x; return ++x == 2; }
@@ -167,8 +169,6 @@ public Transaction[] makeChainedTransactions (KeyPair key_pair,
     ulong spend_amount = 40_000_000, in Transaction gen_tx = GenesisTransaction)
     @safe
 {
-    import agora.common.Amount;
-    import agora.common.Hash;
     import agora.consensus.data.Block;
     import agora.consensus.Genesis;
     import std.conv;
@@ -237,9 +237,7 @@ unittest
 {
     import agora.consensus.data.Block;
     import agora.consensus.Genesis;
-    import agora.common.Amount;
-    import agora.common.Hash;
-    import std.format;
+
     auto gen_key = WK.Keys.Genesis;
 
     /// should spend genesis block's outputs
@@ -279,8 +277,6 @@ unittest
 {
     import agora.consensus.data.Block;
     import agora.consensus.Genesis;
-    import agora.common.Amount;
-    import agora.common.Hash;
 
     auto gen_key = WK.Keys.Genesis;
     const(Transaction)[] txes = makeChainedTransactions(gen_key, null, 1);
@@ -291,11 +287,8 @@ unittest
 unittest
 {
     import std.exception : assumeUnique;
-    import std.range;
     import core.thread;
-    import agora.common.Amount;
     import agora.common.BitField;
-    import agora.common.Hash;
     import agora.consensus.data.Block;
     import agora.consensus.Genesis;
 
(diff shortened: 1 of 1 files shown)


source/agora/utils/Test.d
```diff
@@ -627,60 +632,360 @@ unittest
...
+{
+    // Workaround DMD not understanding `inout` in the context of a delegate
+    // TODO: Report bug / find it on `issues.dlang.org`
+    auto retInOut (inout typeof(OutputRef.tupleof) args)
+    {
+        return inout(OutputRef)(args);
+    }
+    return iota(tx.outputs.length)
+        .map!(idx => retInOut(tx, cast(uint)idx));
+}
```
Geod24: Since this is only used once, perhaps I should just remove it

source/agora/utils/Test.d
```diff
@@ -627,60 +632,360 @@ unittest
...
-    result.outputs[0].value.mustAdd(remainder);
-    const inputHash = input.hashFull();
+    private KeyPair getOwnerKeyPair (scope const PublicKey address) @safe
+    {
+        // FIXME: Using the binary version didn't work (?)
+        if (address.toString() == "GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ")
+            return  () @trusted {
+                return KeyPair.fromSeed(Seed.fromString(
+                    "SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4"));
+            }();
```
Geod24: Perhaps we can just have the genesis key be in the well-known key (not as a `A`, `B`, `C`, etc... but as `Genesis`).

source/agora/utils/Test.d
```diff
@@ -627,60 +632,360 @@ unittest
...
 
-    scope sign = (KeyPair kp, Hash h) @trusted { return kp.secret.sign(h[]); };
+        Returns:
+            Reference to `this` for easy chaining
 
-    const resultHash = result.hashFull();
-    foreach (idx, ref in_; result.inputs)
+    ***************************************************************************/
+
+    public ref typeof(this) draw (Amount amount, in PublicKey[] toward... )
```
Geod24: Possible future improvement: An overload that accepts a `const Amount[PublicKey]`

source/agora/utils/Test.d
```diff
@@ -627,60 +632,360 @@ unittest
...
+
+    /// Transactions to be built and returned
+    private Transaction data;
+}
+
+/// Returns: A range ofTransactions which are spendable in the Genesis block
+public auto genesisSpendable () @safe
+{
+    return GenesisBlock.txs.filter!(tx => tx.type == TxType.Payment);
+}
```
Geod24: Since in the future we want genesis to have both `Freezing` and `Payment` transactions, this is needed

source/agora/utils/Test.d
```diff
@@ -689,10 +994,13 @@ unittest
 {
     import std.range;
 
-    KeyPair[] keys = iota(8).map!(_ => KeyPair.random()).array;
-    KeyPair genesisKP = getGenesisKeyPair();
-    const first = GenesisBlock.txs[0];
-    const equalTx = first.split([genesisKP], keys.map!(k => k.address).array);
+    KeyPair[] keys = iota(8).map!(idx => WK.Keys[idx]).array;
```
```suggestion
    KeyPair[] keys = WK.Keys.byRange.take(8).array;
```

source/agora/utils/Test.d
```diff
@@ -689,10 +994,13 @@ unittest
...
+    auto first = genesisSpendable.front;
+
+    const equalTx = TxBuilder(first)
```
Geod24: TODO: Change name

source/agora/utils/Test.d
```diff
@@ -627,60 +632,360 @@ unittest
...
+
+        Params:
+            toward = Beneficiary to split the currently available amount between
+
+        Returns:
+            Reference to `this` for easy chaining
+
+    ***************************************************************************/
+
+    public ref typeof(this) split (scope const PublicKey[] toward...)
```
Geod24: Should there be an overload that accepts a range ?

source/agora/utils/Test.d
Geod24: Also range overload

source/agora/utils/Test.d
```diff
@@ -707,11 +1015,13 @@ unittest
 {
     import std.range;
 
-    KeyPair[] keys16 = iota(16).map!(_ => KeyPair.random()).array;
-    // Use Genesis
-    KeyPair genesisKP = getGenesisKeyPair();
-    const first = GenesisBlock.txs[0];
-    const resTx1 = first.split([genesisKP], keys16.map!(k => k.address).array);
+    KeyPair[] keys16 = iota(16).map!(idx => WK.Keys[idx]).array;
```
```suggestion
    KeyPair[] keys = WK.Keys.byRange.take(16).array;
```

source/agora/utils/Test.d
```diff
@@ -707,11 +1015,13 @@ unittest
...
+    auto first = genesisSpendable.front;
+
+    const resTx1 = TxBuilder(first)
+        .split(keys16.map!(k => k.address).array)
```
Geod24: Since all those unittests do `map` and `array` again, the `map` could be done before.
Alternatively, range overload.

source/agora/utils/Test.d
```diff
@@ -775,3 +1092,20 @@ unittest
     const Amount ExpectedAmount = Amount(500_000_000L * 10_000_000L);
     assert(result.outputs[0].value == ExpectedAmount);
 }
+
+/// Test merging all inputs into one output
```
Geod24: Fix comment plz

source/agora/utils/Test.d
```diff
@@ -775,3 +1092,20 @@ unittest
...
+
+    const result = TxBuilder(first)
+        .refund(WK.Keys.A.address)
+        .sign();
+
+    assert(result.inputs.length == 8);
+    assert(result.outputs.length == 1);
+
+    // 500M
+    const Amount ExpectedAmount = Amount(500_000_000L * 10_000_000L);
```
Geod24: There's a constant for that.

source/agora/utils/Test.d
```diff
@@ -31,14 +31,19 @@ module agora.utils.Test;
...
 import agora.common.Types;
 import agora.consensus.data.Transaction;
 import agora.consensus.Genesis;
 
 import std.algorithm;
+import std.array;
 import std.file;
+import std.format;
 import std.path;
+import std.range;
```
Geod24, line 47: Some imports in this module can be removed as a consequence, so they should

Geod24 (commented):
A few things to fix.

source/agora/utils/Test.d
Geod24: Actually scratch this, it will change soon

Geod24:
Updated: Moved to a range overload. It looks pretty damn rad. The comments which are left are more notes for the reviewers, not things I plan to address in this PR.

source/agora/utils/Test.d
```diff
@@ -627,106 +620,413 @@ unittest
...
+        Returns:
+            The finalized & signed `Transaction`.
+
+    ***************************************************************************/
+
+    public Transaction sign (TxType type = TxType.Payment) @safe
+    {
+        assert(this.inputs.length, "Cannot sign input-less transaction");
+        assert(this.data.outputs.length || this.leftover.value > Amount(0),
+               "Onput-less transactions are not valid");
```
TrustHenry:
```suggestion
               "Output-less transactions are not valid");
```

source/agora/utils/Test.d
```diff
@@ -627,106 +620,413 @@ unittest
...
+            KeyRange = A range of `PublicKey`
+            amount = `Amount` to distribute to each `PublicKey`
+            toward = Array of `PublicKey` to give `amount` to
+
+        Returns:
+            Reference to `this` for easy chaining
+
+    ***************************************************************************/
+
+    public ref typeof(this) draw (KeyRange) (Amount amount, scope KeyRange toward)
```
AndrejMitrovic, line 918: Is `draw` supposed to be short for `withdraw`? It's ok to keep the name, I just wonder.

Geod24: Indeed. I think it's proper English though ? [Source](https://dictionary.reverso.net/english-cobuild/to+draw+money+from+an+account). CC @bpalaggi

source/agora/utils/Test.d
Geod24: Thanks, fixed

source/agora/utils/Test.d
MichaelKim20: I think this is not divided into fixed value, but filled with the user's chosen values.

source/agora/utils/Test.d
```diff
@@ -627,106 +620,413 @@ unittest
...
+
+    An helper utility to build transactions
+
+    This utility exposes a mutable transaction builder, used to simplify
+    generating complex `Transaction`s, as well as ensuring that generated
+    transactions are valid. Generating invalid `Transaction`s is not supported,
+    however one can generate a valid `Transaction` and mutate it afterwards.
+
+    Basics:
+    When building a transaction, one must first attach an `Output`,
```
AndrejMitrovic, line 706: I don't see `Output` being supported in the constructors. And also the ctor taking in a public key is not mentioned. Or is that what you meant by Output? Hmm

source/agora/utils/Test.d
```diff
@@ -627,106 +620,413 @@ unittest
...
+    ---
+    This will create 4 or 5 outputs, depending on the amounts. The first two
+    will split `myTx` evenly towards `addr1` and `addr2`, the next two will
+    split `otherTx` evenly between `addr3` and `addr4`. If either transaction
+    has an uneven amount, a refund transaction towards the owner of the first
+    output of `myTx` will be created.
+
+    Refund_Address:
+    When making a `TxBuilder`, the minimum requirement is to provide a refund
+    address. If an output is provided, the address which owns this output
```
AndrejMitrovic, line 734: Aha it's mentioned here, a bit later.

source/agora/utils/Test.d
Geod24: "Attaching an Output" doesn't mean that it is done via ctor. You could do `TxBuilder(key).attach(tx)`.
Also `Output` is not directly supported because we need a reference to the transaction, hence to give an `Output` you actually give a `Transaction` + index.

source/agora/utils/Test.d
bpalaggi: > Indeed. I think it's proper English though ? [Source](https://dictionary.reverso.net/english-cobuild/to+draw+money+from+an+account). CC @bpalaggi

Yes, that works.

source/agora/utils/Test.d
```diff
@@ -627,106 +620,413 @@ unittest
...
+    or inability to perform an operation will result in an assertion failure.
+
+    Chaining:
+    As can be seen in the example, operations which modify the state will
+    return a reference to the `TxBuilder` to allow for easy chaining.
+
+    Note:
+    This `struct` is currently not reusable as there is not yet a use case for
+    it, but could be made so in the future. Currently, calling `sign` will
+    invalidate the internal state.
```
AndrejMitrovic, line 753: Have you ever accidentally used it after calling `sign`? Is it worth adding a check like `this.used = true` in `sign()` and adding an invariant?

source/agora/utils/Test.d
AndrejMitrovic: Aha so it's not in the well known key list. Could you add a comment for this in the fixme and replace the `'?'` so it's more obvious?

source/agora/utils/Test.d
Geod24: So far no, as you can see in #952 it's mostly used in a pipeline: `TxBuilder(tx).refund(key).split(WK.Keys.byRange.take(Block.TxsInBlock)).sign()` for example.

source/agora/utils/Test.d
AndrejMitrovic: Right maybe this should be named `distribute` or just `dist` for short?

AndrejMitrovic: But if it's an already existing terminology, that's ok to keep it this way.

source/agora/utils/Test.d
```diff
@@ -627,106 +620,413 @@ unittest
...
+    private Output leftover;
+
+    /// Stores the inputs to consume until `sign` is called
+    private const(OutputRef)[] inputs;
+
+    /// Transactions to be built and returned
+    private Transaction data;
+}
+
+/// Returns: A range ofTransactions which are spendable in the Genesis block
```
AndrejMitrovic:
```suggestion
/// Returns: A range of Transactions which are spendable in the Genesis block
```

AndrejMitrovic:
It would help this PR a lot if the new routines tried to be used in our test-suite. Then we could see if anything needs improving..

AndrejMitrovic:
Ah the usage code was moved to https://github.com/bpfkorea/agora/pull/952?

source/agora/utils/Test.d
```diff
@@ -627,106 +620,413 @@ unittest
...
+    A reference to an `Output` within a `Transaction`
 
-    The `input` transaction will be split evenly in `toward.length` outputs,
-    each of which will be controlled by a key in `toward`.
-    If the sum of outputs in the transaction is not a multiple of
-    `toward.length`, the leftover `Amount` will be added to
-    the output for `toward[0]`.
-    Like other testing utilities, if an error happens (e.g. `input` is invalid,
-    or `from` is missing some keys), an `assert` will be triggered.
+    In order to reference an `Output` in an UTXO, one need to know the hash
```
linked0, line 625:
```suggestion
    In order to refer to an `Output` in an UTXO, one needs to know the hash
```

source/agora/utils/Test.d
```diff
@@ -775,3 +1073,20 @@ unittest
...
     assert(result.outputs[0].value == ExpectedAmount);
 }
+
+/// Test changing the refund address (and merging outputs by extension)
+unittest
+{
+    auto first = genesisSpendable.front;
+
+    const result = TxBuilder(first)
+        .refund(WK.Keys.A.address)
```
linked0: It's too simple. The test should show the case that the function or `refund` will be used in a real situation.

linked0 (approved):
Great job.
I left just one suggestion.

source/agora/utils/Test.d
Geod24: Fixed

source/agora/utils/Test.d
Geod24: Okay let's leave it at `draw` for the time being

source/agora/utils/Test.d
Geod24: I update the UT, LMK if that works for you.

source/agora/utils/Test.d
Geod24: Fixed

Geod24:
Pushed a rebased (removing the need to special case Genesis) & updated from feedback comments.

source/agora/utils/Test.d
linked0: > I update the UT, LMK if that works for you.

It does. Thanks.

AndrejMitrovic (approved):

PR #966, opened 2020-06-16
MemBlockStorage: Detect when blocks failed to load

Geod24:
saveBlock can return false, and if ignored, the error might not be detected until much later.
This can easily happen when one provides a block at the wrong height
(builds a block on top of Genesis and forget to pass Genesis in blocks).


Geod24:
Ah well, apparently an unittest relied on this bug. So blocked on https://github.com/bpfkorea/agora/pull/967

Geod24:
Rebased on v0.x.x

AndrejMitrovic (approved):

PR #967, opened 2020-06-16
Improve some unittests in Ledger

Geod24:
See each commits.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -878,7 +878,7 @@ unittest
     // and one transaction has eight Outputs with a value of 40,000 values.
     void splitGenesis ()
     {
-        splited_txex = splitGenesisTransaction(getGenKeyPairs(), splited_keys);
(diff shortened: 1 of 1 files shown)


source/agora/node/Ledger.d
```diff
@@ -878,7 +878,7 @@ unittest
     // and one transaction has eight Outputs with a value of 40,000 values.
     void splitGenesis ()
     {
-        splited_txex = splitGenesisTransaction(getGenKeyPairs(), splited_keys);
+        splited_txex = splitGenesisTransaction(splited_keys);
```
AndrejMitrovic, line 873: Can we also fix this typo? it's `splitted`, everywhere.

AndrejMitrovic: Actually it should be `split`, not splitted maybe? hmm

Geod24: I plan to nuke those functions, but can fix the typo now if you'd like

AndrejMitrovic (approved):

PR #972, opened 2020-06-18
Add agora-history-push to allow for Stoa development

Geod24:
@MichaelKim20 @TrustHenry Let me know when you start to use it so I can make sure it does everything you need.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -107,6 +107,7 @@
                 "source/agora/consensus/protocol/*",
                 "source/agora/consensus/SCPEnvelopeStore.d",
                 "source/agora/cli/checkvtable/*",
+                "source/agora/cli/history/*",
                 "source/agora/cli/multi/*",
                 "source/agora/cli/vanity/*",
                 "source/agora/registry/main.d"
@@ -131,8 +132,9 @@
             "targetName": "agora-multi",
             "mainSourceFile": "source/agora/cli/multi/main.d",
             "excludedSourceFiles": [
-                "source/agora/cli/checkvtable/*",
                 "source/agora/cli/client/*",
+                "source/agora/cli/history/*",
+                "source/agora/cli/checkvtable/*",
                 "source/agora/cli/vanity/*",
                 "source/agora/node/main.d",
                 "source/agora/registry/main.d"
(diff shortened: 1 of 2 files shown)


Geod24:
TODO: Check compilation in the CI

TrustHenry:
It's simple and very good.
It is pushed to JSON by reading from the ledger.

In order to find the address of Validators among the APIs that Stoa will develop first, we need to read BlockHeader's enrollments in the ledger and match the utxo_key of the enrollment with the transaction output.
In this process, transaction hashes are not stored in the ledger.
The transaction hash and the utxo_key contain index.

AndrejMitrovic:
I'm not sure how this is supposed to be used? It's going to read the blockchain from the same drive as a currently running node? What if the node is writing to the blockstorage on disk while this app is reading from it?

Geod24:
> I'm not sure how this is supposed to be used? It's going to read the blockchain from the same drive as a currently running node? What if the node is writing to the blockstorage on disk while this app is reading from it?

Yes it's a utility tool for the moment, so better run it with a storage that is shut down or blocks that have already been written. Although ideally we should update the storage atomically. Hence why I left it as a WIP, it is solely there for @MichaelKim20 and @TrustHenry to test.

AndrejMitrovic:
Oh it's just a utility, I get it.

AndrejMitrovic (approved):

TrustHenry:
@Geod24 'd better define the key in POST and send the data in the value.

TrustHenry:
^ It works well.

TrustHenry:
I think we need to logically process the HTTP response code.
200: OK
400: Not a valid JSON format, Missing 'block' object in body
500: An error occurred while saving

MichaelKim20 (approved):
LGTM

Geod24:
I'll rebased and git it green soon

TrustHenry:
@Geod24 Are you thinking of a trigger or event to run this periodically later?

Geod24:
With the current state of Stoa, and the new features in Agora, I don't think we need this anymore. Closing.

PR #973, opened 2020-06-18
Set the limit on the validator cycle

linked0:
We have made the consensus-critical constants set through a parameter in the last sprint. By the way, developers could set the validator cycle to a value smaller than the `RevealPreimagePeriod` value inadvertently, which make the developers not to be able to find the cause of failures about tests of revealing pre-images.

So I set the restriction that the value must be equal to or greater than the validator cycle.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -102,10 +102,6 @@ public class EnrollmentManager
     /// Next height for pre-image revelation
     private Height next_reveal_height;
 
-    /// The period for revealing a preimage
-    /// It is an hour interval if a block is made in every 10 minutes
-    public static immutable uint PreimageRevealPeriod = 6;
-
     /// Validator set managing validators' information such as Enrollment object
     /// enrolled height, and preimages.
     private ValidatorSet validator_set;
@@ -332,7 +328,8 @@ public class EnrollmentManager
 
     public bool getNextPreimage (out PreImageInfo preimage) @safe
     {
-        auto height = Height(this.next_reveal_height + PreimageRevealPeriod * 2);
+        auto height = Height(this.next_reveal_height +
+            ConsensusParams.PreimageRevealPeriod);
         return getPreimage(height, preimage);
     }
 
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
But this is strictly not a consensus parameter.. How often a node wants to reveal preimages doesn't change consensus rules.

linked0:
> But this is strictly not a consensus parameter.. How often a node wants to reveal preimages doesn't change consensus rules.

I know that. But the code is only relevant when unit tests are running and making a new file for that is over-kill.

AndrejMitrovic:
Isn't it supposed to be configurable by the users too? CC @Geod24 

Then I would suggest moving it to the node config.

linked0:
> Isn't it supposed to be configurable by the users too? CC @Geod24
> 
> Then I would suggest moving it to the node config.

Do we have to have another parameter class for non-critical constants and another task for propagating the values with a parameter? If we do that, the `assert` statement can be moved to the `EnrollmentManager`.

```
assert(validator_cycle >= PreimageRevealPeriod,
        format("The validator cycle must be equal to or greater than " ~
                "`PreimageRevalePeriod`(%d)", PreimageRevealPeriod));
```

Geod24 (changes requested):
As mentioned, this is not consensus-critical, and should be user-configurable through the config file. Only things that would cause *consensus failure* (two connected well-behaving nodes failing to reach agreement) should go in `ConsensusParams`.

linked0:
> As mentioned, this is not consensus-critical, and should be user-configurable through the config file. Only things that would cause _consensus failure_ (two connected well-behaving nodes failing to reach agreement) should go in `ConsensusParams`.

Right, I agree with @Geod24 and @AndrejMitrovic on the non-consensus-critical constant. I will make an `assert` statement in the code related to unit tests like Base.d, which sets the restriction that the validator cycle above 12 which is `PreimageRevealPeriod * 2`. You know, the value smaller than 12 will make a revelation of pre-images fail.

AndrejMitrovic:
So why not move this parameter to `NodeConfig` and then move the assert to the node's constructor?

linked0:
> So why not move this parameter to `NodeConfig` and then move the assert to the node's constructor?

Yes, I can. Before making that value configurable(#520 ), I will do a quick fix in a proper way.

linked0:
I found out the right approach and will apply it to another PR. So I'm closing it.

PR #974, opened 2020-06-18
Implement stake-based quorum balancer

### LINKED ISSUE #963: Take the node stakes into account when forming quorums
Now that #785 has been implemented we need to start taking the node's stakes into account.

Nodes which have a higher amount staked `should` have a greater chance of being included in a node's quorum.

AndrejMitrovic:
This is the second part of #240, and it implements #963.

There may be further plans to refine the algorithm:

- Take the preimages into account. This should be fairly simple, we will just add it as a second parameter to `getGenerator()`. However there are no mechanisms in place right now to retrieve preimages of all validators at a specific height.
- Take the history of the validator's performance into account. This could be a form of "score", a combination of the node's stake + historical voting data in the blockchain.

Fixes #963

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -522,6 +522,26 @@ public class EnrollmentManager
         return PublicKey(this.key_pair.V[]);
     }
 
+    /// Returns: true if this validator is currently enrolled
+    public bool isEnrolled (UTXOFinder finder) nothrow @safe
+    {
+        Hash[] utxo_keys;
+        assert(this.validator_set.getEnrolledUTXOs(utxo_keys));
+
+        const PublicKey key = this.getEnrollmentPublicKey();
+        foreach (utxo_key; utxo_keys)
+        {
+            UTXOSetValue value;
+            if (!finder(utxo_key, size_t.max, value))
+                assert(0, "UTXO for validator not found!");  // should never happen
+
+            if (value.output.address == key)
+                return true;
+        }
+
+        return false;
+    }
+
     /***************************************************************************
 
         Get the key for the enrollment data for this node
--- a/source/agora/consensus/Quorum.d
+++ b/source/agora/consensus/Quorum.d
@@ -13,161 +13,418 @@
 
 module agora.consensus.Quorum;
 
+import agora.common.Amount;
+import agora.common.BitField;
 import agora.common.Config;
 import agora.common.crypto.Key;
+import agora.common.Hash;
 import agora.common.Set;
 import agora.common.Types;
 import agora.consensus.data.Enrollment;
+import agora.consensus.data.Transaction;
 import agora.consensus.data.UTXOSetValue;
+import agora.consensus.EnrollmentManager;
+
+import scpd.Cpp;
+import scpd.types.Stellar_SCP;
+import scpd.types.Stellar_types : uint256, NodeID;
+import scpd.types.Utils;
+import scpd.types.XDRBase;
+import scpd.quorum.QuorumTracker;
+
+import ocean.core.Test;
 
 import std.algorithm;
 import std.array;
+import std.conv;
 import std.exception;
 import std.format;
 import std.math;
+import std.random;
 import std.range;
 import std.string;
+import std.typecons;
+
+version (unittest)
+{
+    import agora.utils.Test;
+    import std.stdio;
+}
+
+/// Maximum number of nodes to include in a quorum.
+private enum MAX_NODES_IN_QUORUM = 7;
+
+/// Threshold to use for the quorum configuration (e.g. 0.80 = 80%)
+private enum double THRESHOLD = 0.80;
 
 /*******************************************************************************
 
-    Build the quorum configuration for the given public key and the staked
-    enrollments.
+    Build the quorum configuration for the given public key and the registered
+    enrollment keys.
 
     Params:
-        node_key = the key of the node for which to generate the quorum
-        keys = the array of registered enrollment UTXO keys
+        key = the key of the node for which to generate the quorum
+        utxo_keys = the list of UTXO keys of all the active enrollments
         finder = UTXO finder delegate
-
-    Returns:
-        the generated quorum config
 
 *******************************************************************************/
 
-public QuorumConfig buildQuorumConfig ( const ref PublicKey node_key,
-    in Hash[] keys, in UTXOFinder finder )
+public QuorumConfig buildQuorumConfig ( const ref PublicKey key,
+    in Hash[] utxo_keys, UTXOFinder finder )
 {
-    Set!PublicKey all_nodes = getPublicKeys(keys, finder);
+    // special-case: only 1 validator is active
+    if (utxo_keys.length == 1)
+        return QuorumConfig(1, [key]);
+
+    // not including our own
+    NodeStake[] stakes = buildStakesDescending(key, utxo_keys, finder);
 
     QuorumConfig quorum;
-    all_nodes.each!(node => quorum.nodes ~= node);
-    quorum.nodes.sort();
-    quorum.threshold = quorum.nodes.length;
+    quorum.nodes ~= key;  // add ourself first
 
+    // for filtering duplicates from dice()
+    auto added = BitField!uint(stakes.length);
+    auto RNG_gen = getGenerator(key);
+    auto stake_amounts = stakes.map!(stake => stake.amount.integral);
+
+    // +1 as we already added ourself
+    const MaxNodes = min(stakes.length + 1, MAX_NODES_IN_QUORUM);
+    while (quorum.nodes.length < MaxNodes)
+    {
+        auto idx = dice(RNG_gen, stake_amounts);
+        if (added[idx])  // skip duplicate
+            continue;
+
+        quorum.nodes ~= stakes[idx].key;
+        added[idx] = true;  // mark used
+    }
+
+    quorum.nodes.sort;
+    quorum.threshold = max(1, cast(size_t)ceil(THRESHOLD * quorum.nodes.length));
     return quorum;
+}
+
+/// 1 node
+unittest
+{
+    auto keys = getKeys(1);
+    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.only, keys);
+    verifyQuorumsSanity(quorums);
+    verifyQuorumsIntersect(quorums);
+    test!"=="(countNodeInclusions(quorums, keys), [1]);
+}
+
+// 2 nodes
+unittest
+{
+    auto keys = getKeys(2);
+    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.repeat(2), keys);
+    verifyQuorumsSanity(quorums);
+    verifyQuorumsIntersect(quorums);
+    test!"=="(countNodeInclusions(quorums, keys), [2, 2]);
+}
+
+// 3 nodes with equal stakes
+unittest
+{
+    auto keys = getKeys(3);
+    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.repeat(3), keys);
+    verifyQuorumsSanity(quorums);
+    verifyQuorumsIntersect(quorums);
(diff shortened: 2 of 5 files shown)


AndrejMitrovic:
@TrustHenry help me I have a Windows linker error (your favorite)!

```
agora-unittests.obj : error LNK2019: unresolved external symbol "public: static class std::shared_ptr<class stellar::QuorumIntersectionChecker *> __cdecl stellar::QuorumIntersectionChecker::create(struct std::unordered_map<struct stellar::PublicKey,class std::shared_ptr<struct stellar::SCPQuorumSet>,struct std::hash<struct stellar::PublicKey>,struct std::equal_to<struct stellar::PublicKey>,class std::allocator<struct std::pair<struct stellar::PublicKey const ,class std::shared_ptr<struct stellar::SCPQuorumSet> > > > const &)" (?create@QuorumIntersectionChecker@stellar@@SA?AV?$shared_ptr@PEAVQuorumIntersectionChecker@stellar@@@std@@AEBU?$unordered_map@UPublicKey@stellar@@V?$shared_ptr@USCPQuorumSet@stellar@@@std@@U?$hash@UPublicKey@stellar@@@4@U?$equal_to@UPublicKey@stellar@@@4@V?$allocator@U?$pair@$$CBUPublicKey@stellar@@V?$shared_ptr@USCPQuorumSet@stellar@@@std@@@std@@@4@@4@@Z) referenced in function _D5agora9consensus6Quorum22verifyQuorumsIntersectFHSQBy6common6crypto3Key9PublicKeySQDeQBg5Types12QuorumConfigZv
```

TrustHenry:
> um.. Windows linker error

pragma(mangle, `?create@ ....

Wouldn't this be a temporary solution?

AndrejMitrovic:
> Could not deserialize value

Ok it seems that the ballot timer erroring out ("moved to an invalid state") is some kind of memory corruption bug. I'm looking into it.

AndrejMitrovic:
I found the root cause of the memory corruption bug (and it would happen in master too), this is the reason why we had "moved to a bad state" errors frequently.

I added a fixup commit, but it's not very efficient - I just wanted to unblock my work. I could split it into another PR too.

AndrejMitrovic:
I will split off some commits from this PR.

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+    while (quorum.nodes.length < MAX_NODES_IN_QUORUM)
+    {
+        auto idx = dice(rnd_gen, stake_amounts);
+        if (added[idx])  // skip duplicate
+            continue;
 
+        quorum.nodes ~= stakes[idx].key;
+        added[idx] = true;  // mark used
+
+        if (quorum.nodes.length >= stakes.length + 1)  // our stake not included
```
AndrejMitrovic: I think this condition could be moved inside the `while` condition.

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+            break;  // ran out of nodes
+    }
+
+    quorum.nodes.sort;
+    quorum.threshold = max(1, cast(size_t)ceil(THRESHOLD * quorum.nodes.length));
     return quorum;
 }
 
+version (unittest)
```
AndrejMitrovic: This should be moved somewhere else so it doesn't interfere with ddoc comments.

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+private const(PublicKey)[] getKeys (size_t count)
+{
+    return WK.Keys.byRange.take(count).map!(kp => kp.address).array;
+}
+
+/// 1 node
```
AndrejMitrovic, line 107: Only one of these unittests should be ddoc-documented, the rest doesn't have to be.

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+{
+    import libsodium.crypto_shorthash;
+    import std.bitmanip;
+
+    // generated once with 'crypto_shorthash_keygen'
+    static immutable ubyte[crypto_shorthash_KEYBYTES] PublicKey =
+        [111, 165, 189, 80, 37, 5, 16, 194, 39, 214, 156, 169, 235, 221, 21, 126];
+
+    ubyte[ulong.sizeof] short_hash;
+    crypto_shorthash(short_hash.ptr, hash[].ptr, 64, PublicKey.ptr);
```
AndrejMitrovic:
```suggestion
    crypto_shorthash(short_hash.ptr, hash[].ptr, hash[].length, PublicKey.ptr);
```

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+}
+
+/// Create a shorthash from a 64-byte blob to initialize the rnd generator
+private ulong toShortHash (const ref Hash hash) @trusted
+{
+    import libsodium.crypto_shorthash;
+    import std.bitmanip;
+
+    // generated once with 'crypto_shorthash_keygen'
+    static immutable ubyte[crypto_shorthash_KEYBYTES] PublicKey =
```
AndrejMitrovic: `PublicKey` is a bad name for this. We should just name it `iv` as a kind of initialization vector.

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
 
 /*******************************************************************************
 
-    For each enrollment find the associated public key,
-    and return the set of keys.
+    Create a random number generator which uses the node's public key as an
+    initializer for the RNG engine.
+
+    Using the Mersene Twister 19937 64-bit random number generator.
+    The public key is reduced to a short hash of 8 bytes,
```
AndrejMitrovic:
```suggestion
    The public key is reduced to a short hash of 8 bytes
```

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+
+    Returns:
+        a Mersenne Twister 64bit random generator
+
+*******************************************************************************/
+
+private auto getGenerator (PublicKey key)
+{
+    Mt19937_64 gen;
+    auto hash = hashFull(key[]);  // key is only 32 bytes, hash is 64
```
AndrejMitrovic: So why not just change `toShortHash` to take a 32 byte value?

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+
+/*******************************************************************************
+
+    Build a list of NodeStake's in descending stake order
+
+    Params
+        filter = the node's own key should be filtered here
         utxo_keys = the list of enrollments' UTXO keys
-        finder = UTXO finder delegate
+        finder = utxo to stake & public key lookup delegate
```
AndrejMitrovic:
```suggestion
        finder = lookup delegate to find the public key & stake
```

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
 
     Returns:
-        a set of all enrolled public keys
+        the list of stakes in descending stake order
 
 *******************************************************************************/
 
-private Set!PublicKey getPublicKeys (in Hash[] utxo_keys, in UTXOFinder finder)
+private NodeStake[] buildStakesDescending (const ref PublicKey filter,
+    in Hash[] utxo_keys, UTXOFinder finder)
```
AndrejMitrovic:
```suggestion
    in Hash[] utxo_keys, UTXOFinder finder) nothrow
```

Maybe?

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
 
-    Params:
-        num_enrollments = the number of enrollments to create
+    Build the quorum configs for the given enrollments and range of seeds,
+    and return a map of the number of times each node was included in
+    another node's quorum set.
 
     Returns:
-        a tuple of (Enrollment[], UTXOFinder)
+        the map of node => quorum set inclusion counts
```
AndrejMitrovic: This documentation is wrong.

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+    and return a map of the number of times each node was included in
+    another node's quorum set.
+
+    Returns:
+        the map of node => quorum set inclusion counts
+
+*******************************************************************************/
+
+version (unittest)
+private size_t[] countInclusions (QuorumConfig[PublicKey] quorums,
```
AndrejMitrovic:
```suggestion
private size_t[] countNodeInclusions (QuorumConfig[PublicKey] quorums,
```

source/agora/consensus/Quorum.d
```diff
@@ -13,181 +13,476 @@
...
+}
+
+/*******************************************************************************
+
+    Build the quorum configs for the given enrollments and range of seeds,
+    and return a map of the number of times each node was included in
+    another node's quorum set.
+
+    Returns:
+        the map of node => quorum set inclusion counts
```
AndrejMitrovic: It's not a map anymore, but an array. Document why.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -112,20 +112,34 @@ extern(D):
 
         Set or update the quorum configuration
 
+        The node additionally takes the list of all other validator's quorum
+        configurations and saves them for lookup in getQSet().
```
```suggestion
        configurations and saves them for later lookup by getQSet().
```

AndrejMitrovic:
The commit message should talk a bit more how the algorithm works.

AndrejMitrovic:
Some design comments:

- I think we can further iterate and alter the design of the algorithm as we move along and learn new things. It doesn't have to be perfect from the start.
- I thought about the network communication overhead costs. But I came to the conclusion that the quorum design alone won't have that much of an impact on latency. For example if node X is included in 10 other node's quorums, it doesn't necessarily mean those nodes must directly communicate with node X. They are only interested in there being a signed message from X, and that message may be propagated by other nodes. It should the the role of the network manager to find the most efficient way to get a message to a particular node.
- I also thought about whether "past performance" of a Validator should be taken into account in quorum balancing in addition to the node's stake. If for example there is a validator with a lot of stake but which fails / refuses to sign many blocks then the quorum balancer should punish that node by lowering its score. But thinking about this again, this becomes a little bit hard to do because we don't really have "identity", it is tied to the public key. A node operator can easily generate a new key, use a different UTXO, etc..
- I am not sure about the edge-cases yet. For example, probability could make it possible a validator is not included in any quorum (if its stake is very tiny compared to the rest of the network). I should try to write a test-case for this.

Anyway my theory is that we will refine this many more times before Coinnet.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -522,6 +522,26 @@ public class EnrollmentManager
         return PublicKey(this.key_pair.V[]);
     }
 
+    /// Returns: true if this validator is currently enrolled
+    public bool isEnrolled (UTXOFinder finder) nothrow @safe
```
Geod24, line 526:
```suggestion
    public bool isEnrolled (scope UTXOFinder finder) nothrow @safe
```

source/agora/consensus/Quorum.d
```diff
@@ -13,180 +13,478 @@
...
+        3, 2, 9, 8, 6, 7, 4, 4, 6, 4, 4, 7, 3, 3, 5, 5, 7, 6, 6, 6, 6, 7, 8, 5,
+        4, 4, 4, 5, 2, 2, 6, 3, 6, 5, 5, 7, 3, 3, 7, 4, 63]);
+}
+
+// 128 nodes where two nodes own 66% of the stake
+unittest
+{
+    auto keys = getKeys(128);
+    auto amounts = Amount(400000000000uL).repeat(128).array;
+    amounts[0] = amounts[$ - 1] = Amount(400000000000uL * 126);
```
Geod24, line 233: Please use `_` for big numbers

Geod24 (approved):
LGTM overall. Some nits but nothing blocking. As you said, we'll iterate over this quite a few times.

PR #975, opened 2020-06-19
Fix segfault in `sendPreimage` function

linked0:
The original code causes a segfault error because the `preimage` parameter has a `ref` modifier and the D compiler does not allocate memory for the parameter. So the `preimage` is corrupted after the fiber having called the delegate terminates. So the `ref` modifier must be removed at the moment.


AndrejMitrovic (approved):

Geod24 (approved):
For reference: https://issues.dlang.org/show_bug.cgi?id=20956

Issue #935, opened 2020-06-08
Should we add the ability to enroll as soon as we freeze ?


AndrejMitrovic:
Question about this: Should we add the ability to enroll as soon as we freeze ?
Pro:
- Genesis does it;
- It allows to potentially solve the censorship issue if paired with a rule that more validators => Higher priority;
Cons:
- Attack vector ?

_Originally posted by @Geod24 in https://github.com/bpfkorea/agora/pull/845_

TrustHenry:
If so, then it's the freeze Tx that the enrollment refers to at the same height of the block.

linked0:
It might be good for the freezing functionality to have a flag to determine that it's for being a validator or not. If a node is intended to be a validator, enrolling as soon as freezing could be good. This is not the exact answer to your question but I'm favor of adding the functionality(enroll as soon as freeze). Anyway, let's think about the attack vector.

Geod24:
Why would you freeze if you don't want to be a validator ?

MichaelKim20:
That's a good idea.
I think we should consider not only the validators of block creation but also the function of Congress voting.

linked0:
> Why would you freeze if you don't want to be a validator ?

I know we have only one use case for frozen UTXO. I just thought there might be use cases for which the UTXO could be frozen. Hmm, I didn't found out the actual cases yet.

Geod24:
Fixed by #989

PR #989, opened 2020-06-29
Allow enrollments to be present in the same block as the freezing transaction

### LINKED ISSUE #907: Add Enrollments to the hardcoded genesis blocks

Geod24:
This was never really specified, but it makes sense to allow it considering our recent work.
Since it's needed to fix (along with another PR) #907 and there's little time left for the demo, the implementation isn't great but works well enough.

### DIFF
--- a/source/agora/consensus/data/UTXOSetValue.d
+++ b/source/agora/consensus/data/UTXOSetValue.d
@@ -107,7 +107,7 @@ version (unittest) public class TestUTXOSet
     }
 
     /// Short hand to add a transaction
-    public void put (const Transaction tx)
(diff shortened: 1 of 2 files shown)


AndrejMitrovic (approved):

AndrejMitrovic:
"Make all PRs by Wednesday"

Geod24:
Fixed #935

PR #991, opened 2020-06-29
Trivial: Simplify some EnrollmentManager tests

Geod24:
See the first commit and the diff of the second.

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -797,7 +797,9 @@ public struct TxBuilder
     private Transaction data;
 }
 
-/// Returns: A range of Transactions which are spendable in the Genesis block
+/// Returns:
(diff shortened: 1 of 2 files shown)


source/agora/utils/Test.d
```diff
@@ -800,7 +800,10 @@ public struct TxBuilder
 /// Returns: A range of Transactions which are spendable in the Genesis block
```
AndrejMitrovic: Transactions => TxBuilders

AndrejMitrovic (changes requested):
Leaving this comment because it was pending. Now I'm really out.

source/agora/utils/Test.d
Geod24: Thanks! Fixed.

Issue #900, opened 2020-05-31
Avoid nominating and externalizing Enrollments that will fail Block validation


Geod24:
There were several issues with nomination discovered in #899:
- [x] There were leftover `Enrollment` in the `EnrollmentPool` which were picked up to be nominated, although they were also in the `ValidatorSet`;
- [x] There is no check that the result of `Ledger.prepareNominatingSet` can in fact be nominated;
- [x] Other node achieved consensus on duplicated Enrollments (!)

All of this needs to be fixed. #899 was just a stopgap measure but nominating enrollment (and nomination in general) needs to be much more thought out.

linked0:
The main thing of this issue is to remove the duplication of enrollments. So we need to add the following algorithm for this issue.
- We add the `requested_height` column in the `enrollment_pool` table, which means the height when the enrollment is inserted.
- If the `requested_height` of an enrollment is smaller than the `enrolled_height`, the enrollment is removed from the `EnrollmentPool` when the `getEnrollments` function of the `EnrollmentManager` is called.

All the issues to be resolved is all about the `getEnrollments` function of the `EnrollmentManager`.
It there are any other point of view or any suggestions, please let me know. I will review them positively.

Geod24:
Last point: Do not return `Enrollment` for which the `requested_height` is > than the height we are nominating for.

Geod24:
> We add the `requested_height` column in the `enrollment_pool` table, which means the height when the enrollment is inserted.

It should be "the height when the enrollment will be available".

PR #995, opened 2020-07-01
Nominate enrollments using `avail_height` on the `enrollment_pool` table

linked0:
I added the `avail_height` column into the `EnrollmentPool` in order to check if an enrollment can be a validator. If even the enrollment had been promoted as a validator, it can be used as the request to be a validator in the next cycle. So the `getEnrollments` function can select the enrollments which can be a validator for the next block with the `avail_height` and the height passed in as a parameter.

When the `getEnrollments` is called, it uses the `avail_height` column to get the right enrollments at the current block height. The rule is as follows.
1) If the `enrolled_height` of an enrollment is greater than the `avail_height`, we do not return the enrollment.
2) If the current height(the height for which we are nominating minus 1) is smaller than the `avail_height`, we do not return the enrollment.

Relates to #900

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -209,7 +209,7 @@ public class EnrollmentManager
     {
         auto enrolled_height = this.getEnrolledHeight(enroll.utxo_key);
 
-        // The last height at which a node plays a role as a validator.
+        // The first height at which the enrollment can be enrolled.
         ulong avail_height;
 
         if (enrolled_height == ulong.max)
@@ -241,12 +241,16 @@ public class EnrollmentManager
         assumeSafeAppend(enrolls);
 
         static Enrollment[] pool_enrolls;
-        this.enroll_pool.getEnrollments(pool_enrolls);
+        this.enroll_pool.getEnrollments(pool_enrolls, Height(height + 1));
         foreach (enroll; pool_enrolls)
         {
             const enroll_height = this.getEnrolledHeight(enroll.utxo_key);
+            const avail_height =
+                this.enroll_pool.getAvailableHeight(enroll.utxo_key);
+            assert(avail_height != Height(0));
             if (enroll_height == ulong.max ||
-                height >= enroll_height + this.params.ValidatorCycle - 1)
+                (avail_height >= enroll_height &&
+                    height >= enroll_height + this.params.ValidatorCycle - 1))
             {
                 enrolls ~= enroll;
             }
@@ -1054,4 +1058,60 @@ unittest
     auto e1 = EnrollmentManager.makeEnrollment(WK.Keys.A, utxo, 10, 0);
     auto e2 = EnrollmentManager.makeEnrollment(WK.Keys.B, utxo, 10, 0);
     assert(e1.random_seed != e2.random_seed);
+}
+
+/// Test for the height when the enrollment will be available
+unittest
+{
+    import agora.consensus.data.Transaction;
+
+    scope utxo_set = new TestUTXOSet;
+    KeyPair key_pair = KeyPair.random();
+    genesisSpendable().map!(txb => txb.refund(key_pair.address).sign(TxType.Freeze))
+        .each!(tx => utxo_set.put(tx));
+
+    // create an EnrollmentManager
+    const validator_cycle = 20;
+    scope man = new EnrollmentManager(":memory:", key_pair,
+        new immutable(ConsensusParams)(validator_cycle));
+
+    // create and add the first enrollment
+    Enrollment[] enrolls;
+    auto enroll = man.createEnrollment(utxo_set.keys[0]);
+    assert(man.addEnrollment(enroll, Height(10), &utxo_set.findUTXO));
+
+    // if the current height is smaller than the available height,
+    // we can get no enrollment
+    man.getEnrollments(enrolls, Height(9));
+    assert(enrolls.length == 0);
+
+    // if the current height is greater than or equal to the available height,
+    // we can get enrollments
+    man.getEnrollments(enrolls, Height(10));
+    assert(enrolls.length == 1);
+
+    // make the enrollment a validator
+    man.addValidator(enroll, Height(11), &utxo_set.findUTXO);
+    man.getEnrollments(enrolls, Height(11));
+    assert(enrolls.length == 0);
+
+    // add the enrollment that is already a validator, and check if
+    // the enrollment can be nominated at the height before the cycle end
+    assert(man.addEnrollment(enroll, Height(11), &utxo_set.findUTXO));
+    man.getEnrollments(enrolls, Height(validator_cycle + 9));
+    assert(enrolls.length == 0);
+    man.getEnrollments(enrolls, Height(validator_cycle + 10));
+    assert(enrolls.length == 1);
+
+    // make the enrollment a validator again
+    man.clearExpiredValidators(Height(validator_cycle + 11));
+    man.addValidator(enroll, Height(validator_cycle + 11), &utxo_set.findUTXO);
+
+    // add the enrollment that has the available height smaller than
+    // the enrolled height of the validator, we can get no enrollment
+    assert(man.addEnrollment(enroll, Height(validator_cycle + 10),
+        &utxo_set.findUTXO));
+    assert(man.pool.count() == 1);
+    man.getEnrollments(enrolls, Height(validator_cycle + 11));
+    assert(enrolls.length == 0);
 }
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -192,6 +192,36 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Add a enrollment data to the enrollment pool
+
+        Params:
+            enroll = the enrollment data to add
+            block_height = current block height
+            finder = the delegate to find UTXOs with
+
+        Returns:
+            true if the enrollment data has been added to the enrollment pool
+
+    ***************************************************************************/
+
+    public bool addEnrollment (const ref Enrollment enroll, Height block_height,
+        scope UTXOFinder finder) @safe nothrow
+    {
+        auto enrolled_height = this.getEnrolledHeight(enroll.utxo_key);
+
+        // The last height at which a node plays a role as a validator.
+        ulong avail_height;
+
+        if (enrolled_height == ulong.max)
+            avail_height = block_height + 1;
+        else
+            avail_height = enrolled_height + this.params.ValidatorCycle;
+
+        return this.pool.add(enroll, Height(avail_height), finder);
+    }
+
+    /***************************************************************************
+
         Get the unregistered enrollments that can be validator in the next
         block based on the current block height.
 
@@ -735,20 +765,20 @@ unittest
         EnrollmentManager.makeEnrollment(gen_key_pair, utxo_hash, 1008);
 
     auto enroll = man.createEnrollment(utxo_hash);
-    assert(!man.pool.add(fail_enroll, &storage.findUTXO));
-    assert(man.pool.add(enroll, &storage.findUTXO));
+    assert(!man.addEnrollment(fail_enroll, Height(1), &storage.findUTXO));
+    assert(man.addEnrollment(enroll, Height(1), &storage.findUTXO));
     assert(man.pool.count() == 1);
-    assert(!man.pool.add(enroll, &storage.findUTXO));
+    assert(!man.addEnrollment(enroll, Height(1), &storage.findUTXO));
 
     // create and add the second Enrollment object
     auto utxo_hash2 = utxo_hashes[1];
     auto enroll2 = man.createEnrollment(utxo_hash2);
-    assert(man.pool.add(enroll2, &storage.findUTXO));
+    assert(man.addEnrollment(enroll2, Height(1), &storage.findUTXO));
     assert(man.pool.count() == 2);
 
     auto utxo_hash3 = utxo_hashes[2];
     auto enroll3 = man.createEnrollment(utxo_hash3);
-    assert(man.pool.add(enroll3, &storage.findUTXO));
+    assert(man.addEnrollment(enroll3, Height(1), &storage.findUTXO));
     assert(man.pool.count() == 3);
 
     Enrollment[] enrolls;
@@ -792,7 +822,8 @@ unittest
     // Reverse ordering
     ordered_enrollments.sort!("a.utxo_key > b.utxo_key");
     foreach (ordered_enroll; ordered_enrollments)
-        assert(man.pool.add(ordered_enroll, &storage.findUTXO));
+        assert(man.addEnrollment(ordered_enroll, Height(10),
+            &storage.findUTXO));
     man.getEnrollments(enrolls, Height(man.params.ValidatorCycle + 8));
     assert(enrolls.length == 3);
     assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
@@ -863,7 +894,7 @@ unittest
 
     auto utxo_hash = utxo_hashes[0];
     auto enroll = man.createEnrollment(utxo_hash);
-    assert(man.pool.add(enroll, &storage.findUTXO));
+    assert(man.addEnrollment(enroll, Height(1), &storage.findUTXO));
 
     assert(man.params.ValidatorCycle - 101 == 907); // Sanity check
     assert(man.getValidatorPreimage(utxo_hash) == PreImageInfo.init);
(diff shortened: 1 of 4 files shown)


linked0:
Ready for review

Geod24 (changes requested):
I see three issues with this PR:
1) The test code and the implementation are separate. As repeated multiple times, they should come together.
2) The test does not test what it should. If I phrase what the test is doing, in English: "Get the value of a field in the SQL database and check that it is exactly what we put in". That is testing the database, *not* the actual behavior we want to test. What we want to test is that the precondition of #900 are here.
3) What we want is the height at which the enrollment is available, not when the enrollment was inserted. I added [a comment](https://github.com/bpfkorea/agora/issues/900#issuecomment-652717204) in the issue to clarify this.

linked0:
Ready for review.
If this PR is approved today, I can do the following task, or another PR next week to complete the issue #900 . If you review it, I will modify it as soon as possible.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -190,6 +190,34 @@ public class EnrollmentManager
...
+        Returns:
+            true if the enrollment data has been added to the enrollment pool
+
+    ***************************************************************************/
+
+    public bool addEnrollment (const ref Enrollment enroll, Height block_height,
+        scope UTXOFinder finder) @safe nothrow
+    {
+        auto enrolled_height = this.getEnrolledHeight(enroll.utxo_key);
+        ulong avail_height;
```
TrustHenry, line 207: Annotation for avail_height is required.

linked0: > Annotation for avail_height is required.

It would be better to describe the purpose of the `avail_height`. Thanks.

linked0: > Annotation for avail_height is required.

It would be better to describe the purpose of the avail_height. Thanks.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -190,6 +190,34 @@ public class EnrollmentManager
...
+
+        if (enrolled_height == ulong.max)
+            avail_height = block_height + 1 + this.params.ValidatorCycle;
```
Geod24: I don't understand this part. Why do you set `avail_height` to 1009 blocks in the future when there is no Enrollment ?

source/agora/consensus/EnrollmentPool.d
```diff
@@ -84,35 +87,42 @@ public class EnrollmentPool
...
+                        enroll);
+                    return false;
+                }
+                () @trusted {
+                    this.db.execute("INSERT INTO enrollment_pool " ~
+                        "(key, val, avail_height) VALUES (?, ?, ?)",
+                        enroll.utxo_key.toString(), buffer,
+                        avail_height.value);
+                }();
+            }
```
Geod24: https://www.sqlitetutorial.net/sqlite-replace-statement/

source/agora/consensus/EnrollmentManager.d
linked0: The `block_height` + 1 means the enrolled height if the enrollment has been enrolled at the next block of the `block_height`. So the `avail_height` becomes the last height at which a node plays a role as a validator. This assignment for the `avail_height` is only relevant to nodes requesting enrollment after the Genesis block.

source/agora/consensus/EnrollmentPool.d
linked0: > https://www.sqlitetutorial.net/sqlite-replace-statement/

I will apply it to the code. Thanks.

linked0:
Ready for review. Please!

linked0:
Ready for review.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -169,19 +175,25 @@ public class EnrollmentPool
...
     {
         try
         {
-            auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
-                "enrollment_pool WHERE key = ?)", enroll_hash.toString());
-            return results.front().peek!bool(0);
+            auto results = this.db.execute("SELECT avail_height FROM " ~
+                "enrollment_pool WHERE key = ?", enroll_hash.toString());
+            if (!results.empty &&
+                Height(results.oneValue!(size_t)) == avail_height)
```
Geod24: You don't need the `Height()` here

source/agora/consensus/EnrollmentPool.d
```diff
@@ -169,19 +175,25 @@ public class EnrollmentPool
...
+                return true;
+            else
+                return false;
```
Geod24: Why would you write:
```D
if (condition)
    return true;
else
    return false;
```

Instead of simply `return condition;` ? I feel that this pattern is repeated way too frequently, please remember the [KISS principle](https://en.wikipedia.org/wiki/KISS_principle).

Geod24 (changes requested):
There are too many things wrong here. I noted a few of them but the list is not exhaustive.
Let's discuss that IRL on Tuesday.

source/agora/consensus/EnrollmentPool.d
linked0: This needs `Height()` because of the below code. Let's talk about that IRL.
```
if (this.enroll_man.addEnrollment(enroll, this.ledger.getBlockHeight(),
            this.utxo_set.getUTXOFinder()))
{
    this.network.sendEnrollment(enroll);
}
```

source/agora/consensus/EnrollmentPool.d
```diff
@@ -169,19 +175,25 @@ public class EnrollmentPool
...
+    public bool hasEnrollment (const ref Hash enroll_hash, Height avail_height)
+        @trusted nothrow
     {
         try
         {
-            auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
-                "enrollment_pool WHERE key = ?)", enroll_hash.toString());
-            return results.front().peek!bool(0);
+            auto results = this.db.execute("SELECT avail_height FROM " ~
+                "enrollment_pool WHERE key = ?", enroll_hash.toString());
```
Geod24:
```suggestion
            auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
                "enrollment_pool WHERE key = ? AND avail_height = ?", enroll_hash.toString(), avail_heigth);
```

linked0:
Ready for review.
I added an additional commit for the `getEnrollments` function.
I will make additional PR needed today.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -190,6 +190,36 @@ public class EnrollmentManager
...
+            enroll = the enrollment data to add
+            block_height = current block height
+            finder = the delegate to find UTXOs with
+
+        Returns:
+            true if the enrollment data has been added to the enrollment pool
+
+    ***************************************************************************/
+
+    public bool addEnrollment (const ref Enrollment enroll, Height block_height,
```
AndrejMitrovic, line 201: If this function should always be used instead of `pool.add`, then we need to remove the `pool()` getter in this class.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -241,12 +241,16 @@ public class EnrollmentManager
         assumeSafeAppend(enrolls);
 
         static Enrollment[] pool_enrolls;
-        this.enroll_pool.getEnrollments(pool_enrolls);
+        this.enroll_pool.getEnrollments(pool_enrolls, height);
```
AndrejMitrovic: Hmm shouldn't this be `height + 1`?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1054,3 +1058,59 @@ unittest
...
+    import agora.consensus.data.Transaction;
+
+    scope utxo_set = new TestUTXOSet;
+    KeyPair key_pair = KeyPair.random();
+    genesisSpendable().map!(txb => txb.refund(key_pair.address).sign(TxType.Freeze))
+        .each!(tx => utxo_set.put(tx));
+
+    // create an EnrollmentManager
+    const validator_cycle = 20;
+    auto man = new EnrollmentManager(":memory:", key_pair,
```
```suggestion
    scope man = new EnrollmentManager(":memory:", key_pair,
```

source/agora/consensus/EnrollmentPool.d
```diff
@@ -237,20 +237,54 @@ public class EnrollmentPool
...
+    {
+        try
+        {
+            auto results = this.db.execute("SELECT avail_height " ~
+                "FROM enrollment_pool WHERE key = ?", enroll_hash.toString());
+            if (results.empty)
+                return Height(0);
+
+            return Height(results.oneValue!(size_t));
+
```
AndrejMitrovic:
```suggestion
```

source/agora/consensus/EnrollmentPool.d
```diff
@@ -237,20 +237,54 @@ public class EnrollmentPool
...
 
+    /***************************************************************************
+
+        Get the height when the enrollment will be available
+
+        Params:
+            enroll_hash = key for the available block height
+
+        Returns:
+            the available block height
```
AndrejMitrovic:
```suggestion
            the available block height, or 0 if the enrollment isn't found
```

source/agora/consensus/EnrollmentPool.d
```diff
@@ -237,20 +237,54 @@ public class EnrollmentPool
...
+        {
+            log.error("ManagedDatabase operation error: {}", ex.msg);
+            return Height(0);
+        }
+    }
+
+
     /***************************************************************************
 
         Get the unregistered enrollments in the block
```
AndrejMitrovic: This doc should be updated:

Get the unregistered enrollments from the pool

source/agora/consensus/EnrollmentManager.d
AndrejMitrovic: Because the `height` passed here is the already externalized block height, we want to know if the enrollment is available for the next block right?

AndrejMitrovic (dismissed):

source/agora/consensus/EnrollmentManager.d
linked0: > If this function should always be used instead of `pool.add`, then we need to remove the `pool()` getter in this class.

The `ValdiatorSet` and the `EnrollmentPool` are needed at the same time for the operation, so it might be better for the `EnrollmentManager` to have the function. I'm not sure which is better. Do you think putting the code in the Ledger would be better? Anyway, if removing the pool() getter is needed, I'll do it.

AndrejMitrovic: I'm saying what if the code calls `pool.add` instead of `EnrollmentManager.addEnrollment` by mistake? There is additional logic now in the enrollment manager which is not in the pool.

linked0: > I'm saying what if the code calls `pool.add` instead of `EnrollmentManager.addEnrollment` by mistake? There is additional logic now in the enrollment manager which is not in the pool.

Right, the `pool()` getter will be removed.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -190,6 +190,36 @@ public class EnrollmentManager
...
+            true if the enrollment data has been added to the enrollment pool
+
+    ***************************************************************************/
+
+    public bool addEnrollment (const ref Enrollment enroll, Height block_height,
+        scope UTXOFinder finder) @safe nothrow
+    {
+        auto enrolled_height = this.getEnrolledHeight(enroll.utxo_key);
+
+        // The last height at which a node plays a role as a validator.
```
linked0: The comment should be corrected.

source/agora/consensus/EnrollmentManager.d
linked0: > Because the `height` passed here is the already externalized block height, we want to know if the enrollment is available for the next block right?

The meaning of the parameter is the current block height and the function finds the enrollments that are available for the next height. But it can be modified as you suggested in order to reduce confusion.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -211,12 +241,16 @@ public class EnrollmentManager
         assumeSafeAppend(enrolls);
 
         static Enrollment[] pool_enrolls;
-        this.enroll_pool.getEnrollments(pool_enrolls);
+        this.enroll_pool.getEnrollments(pool_enrolls, height);
         foreach (enroll; pool_enrolls)
         {
             const enroll_height = this.getEnrolledHeight(enroll.utxo_key);
+            const avail_height =
+                this.enroll_pool.getAvailableHeight(enroll.utxo_key);
```
Geod24, line 243: Why would we need `getAvailableHeight` if we know `getEnrollments` already filters them out ???

linked0: > Why would we need getAvailableHeight if we know getEnrollments already filters them out ???

It's because we need to filter out the enrollment which left unintentionally in the pool although it had become a validator.

linked0:
Ready for review.
I applied the suggestions, and removed the `pool` function from the `EnrollmentManager`.

source/agora/node/Ledger.d
```diff
@@ -1025,12 +1025,13 @@ unittest
...
     foreach (const ref e; enrollments)
-        assert(ledger.enroll_man.pool.add(e, Height(3), findUTXO));
+        assert(ledger.enroll_man.addEnrollment(e, Height(3), findUTXO));
 
     Enrollment stored_enroll;
     foreach (idx, hash; utxo_hashes)
     {
-        assert(ledger.enroll_man.pool.getEnrollment(hash, stored_enroll));
+        assert((stored_enroll = ledger.enroll_man.getEnrollment(hash)) !=
+            Enrollment.init);
```
Geod24: Please break in two lines

source/agora/consensus/EnrollmentPool.d
```diff
@@ -201,40 +201,39 @@ public class EnrollmentPool
...
-                return true;
+                return deserializeFull!Enrollment(row.peek!(ubyte[])(1));
             }
         }
         catch (Exception ex)
         {
             log.error("Exception occured on getEnrollment: {}, " ~
                 "Key for enrollment: {}", ex.msg, enroll_hash);
-            return false;
+            return enroll;
```
Geod24: Returning `Enrollment.init` will be more explicit.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -201,40 +201,39 @@ public class EnrollmentPool
...
         }
 
-        return false;
+        return enroll;
```
Geod24: Same comment about `init` here

linked0:
I applied the suggestions. Thanks.

source/agora/consensus/EnrollmentPool.d
```diff
@@ -239,18 +239,51 @@ public class EnrollmentPool
...
+        @trusted nothrow
+    {
+        try
+        {
+            auto results = this.db.execute("SELECT avail_height " ~
+                "FROM enrollment_pool WHERE key = ?", enroll_hash.toString());
+            if (results.empty)
+                return Height(0);
+
+            return Height(results.oneValue!(size_t));
```
MichaelKim20, line 260: When it has multiple records, it will return a record with the smallest 'avail_height'?
However, 'avail_height' is not a primary key, so it should be checked.

linked0: > When it has multiple records, it will return a record with the smallest 'avail_height'?
> However, 'avail_height' is not a primary key, so it should be checked.

The `enroll_hash` which is a UTXO hash is a primary key for the table, so the `avail_height` for the key is only one. Thanks.

MichaelKim20: Oh, I see. We can always register only one data in advance.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -241,12 +241,16 @@ public class EnrollmentManager
...
+            assert(avail_height != Height(0));
             if (enroll_height == ulong.max ||
-                height >= enroll_height + this.params.ValidatorCycle - 1)
+                (avail_height >= enroll_height &&
+                    height >= enroll_height + this.params.ValidatorCycle - 1))
```
MichaelKim20, line 247: The person who enrolls the information will feel bored because the information he or she has enrolled is not contained in the block and there is no change.

linked0: The code is for the node which is willing to re-enroll as a validator although it is a validator. The code means as follows in fact.
`(enrolled_height != ulong.max && avail_height >= enroll_height && height >= enroll_height + this.params.ValidatorCycle - 1))`

If the `enrolled_height` equals to `ulong.max`, it will be contained in a block probably because it is a candidate for a validator.
Thanks.

Geod24 (approved):

PR #996, opened 2020-07-02
Change ConsensusData to take an array of Transaction

Geod24:
Using a set was a bit surprising, and didn't compose well with qualifiers,
e.g. we had to cast away constness and immutability.
Using an array is a bit more natural and ensures that `ConsensusData`
can be qualified.

### DIFF
--- a/source/agora/utils/SCPPrettyPrinter.d
+++ b/source/agora/utils/SCPPrettyPrinter.d
@@ -389,7 +389,7 @@ unittest
     Signature sig = Signature("0x000000000000000000016f605ea9638d7bff58d2c0c" ~
                               "c2467c18e38b36367be78000000000000000000016f60" ~
                               "5ea9638d7bff58d2c0cc2467c18e38b36367be78");
-    Enrollment record =
+    const Enrollment record =
     {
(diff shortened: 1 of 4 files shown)


source/agora/utils/SCPPrettyPrinter.d
```diff
@@ -389,24 +389,18 @@ unittest
...
-    enrollments ~= record;
-
-    auto tx_set = Set!Transaction.from([cast(Transaction)GenesisTransaction]);
-
-    ConsensusData cd =
+    const(ConsensusData) cd =
     {
-        tx_set: tx_set,
-        enrolls: enrollments,
+        tx_set:  GenesisBlock.txs[1 .. $],
```
linked0, line 402: Why not use just `GenesisBlock.txs`?

Geod24: It would require changing all the prints again.

linked0: OK, it's just test code, not an issue.

linked0 (approved):
LGTM

Issue #821, opened 2020-05-12
Replace setGenesisBlock / genesis config with hardcoded blocks


AndrejMitrovic:
As discussed with @Geod24 today. We should come up with an alternative to user-configurable genesis blocks. Initially we went with this solution to allow having a different genesis block for unittests / integration tests / testnet / coinnet. 

The problem is that a corrupt config file or just a user typo could cause the node to use a different genesis block.

One alternative was to use `version()` blocks and then build Agora with `-version=CoinNet`, `-version=TestNet`, but that is problematic because we then need to ship multiple binaries.

A better alternative suggested by @Geod24 is to have a runtime choice between several pre-configured genesis blocks. 

So instead of having `config.genesis_block`, we would have `./agora --mode=coinnet` | `./agora --mode=unittest` or something to that effect. It could also be a config option, but the point is that it's not a hex dump of a block but a choice with a limited pre-configured number of options.

Geod24:
So:
```sh
$ agora --testnet
# OR
$ agora --well-know-net
# The default is
$ agora --mainnet
```

Or something similar. I think we should keep `testnet` for, heh, testnet, because that's what people expect, and use a very unambiguous and non-similar name for the integration test (it's most likely to be only written in scripts, and will stand out).

AndrejMitrovic:
Do we want it to be a default? We probably don't want to accidentally run on mainnet by mistake.

I remember the days [at our prev employer] where we would `scp` binaries and then run them, but without it being obvious we're running on a test system or a live one. There were a few times when we ran test apps on the live system by mistake. :P

Geod24:
We should brainstorm it. I think the main difference between agora and the case you mentioned is the intended audience: `agora` is likely to be run by non-developer, hence the defaults need to "just work" as much as possible. But if we make the error message obvious enough I guess it could work, too.

Geod24:
Raising to prio-high because it's blocking us AFAIU.

AndrejMitrovic:
Some notes about this issue:
- Any time we change the built-in integration test genesis block, it may end up breaking all existing tests.
- We need to include Enrollments in the genesis block, it's hard to pre-determine how many so it works well with all tests (including the network tests).

PR #997, opened 2020-07-02
Simplify some tests using `genesisSpendable()` and cleanup some imports

Geod24:
This is needed for #821

### DIFF
--- a/source/agora/consensus/EnrollmentPool.d
+++ b/source/agora/consensus/EnrollmentPool.d
@@ -298,8 +298,6 @@ unittest
     import agora.consensus.data.ConsensusParams;
     import agora.consensus.data.Transaction;
     import std.algorithm;
-    import std.format;
-    import std.conv;
 
     scope storage = new TestUTXOSet;
     KeyPair key_pair = KeyPair.random();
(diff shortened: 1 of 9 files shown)


source/agora/consensus/EnrollmentPool.d
```diff
@@ -295,34 +295,16 @@ private Enrollment createEnrollment(const ref Hash utxo_key,
...
 
     scope storage = new TestUTXOSet;
-
-    auto gen_key_pair = WK.Keys.Genesis;
     KeyPair key_pair = KeyPair.random();
 
-    foreach (idx; 0 .. 8)
-    {
-        auto input = Input(hashFull(GenesisTransaction), idx.to!uint);
+    genesisSpendable().map!(txb => txb.refund(key_pair.address).sign(TxType.Freeze))
```
linked0, line 305: Do we need to use `WK.keys` instead of `key_pair`-like variables in the test like this in the future? If we need to, it would be good to use the `WK` in here.

linked0 (approved):
LGTM.
I left a comment though.

source/agora/consensus/EnrollmentPool.d
Geod24: We don't *have* to, but it's recommended.
In that case, the reason it wasn't done was because the test uses the same key for all enrollments. It's something I wanted to fix, but decided to do it in a separate commit as that is not blocking me on #821 right now.

Geod24:
Anything else ?
@linked0 : Following your comment I will submit another PR to use `WK.Keys`, but it also needs to refactor a few things (like remove the copy-pasta), hence why I think it's better on its own.

linked0:
No, everything's fine. I will merge it.

PR #999, opened 2020-07-02
Remove test-only makeGenesisBlock utility

Geod24:
As noted in the comment, now that issue 907 has been fixed,
this can be removed.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -532,36 +532,6 @@ version (unittest)
     }
 }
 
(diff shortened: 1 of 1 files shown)


linked0 (commented):
Nice, but the `remakeEnrollment` isn't needed anymore also. It can be deleted.

Geod24:
Right, but it seems it was unused before this PR, since I could not find an instance of it in the diff ?
Anyways, added a commit to remove it (and rebased on `v0.x.x`).

linked0 (approved):
LGTM. Thanks.

PR #1000, opened 2020-07-02
EnrollmentManager: Remove buggy call to setEnrollmentKey

Geod24:
```
This call is obviously wrong: it uses 'enroll' as a parameter,
which at this point is default-initialized.
So if an enrollment exists, the update would overwrite valid data
with all default-initialized data (most likely all 0).
After removing this wrong call, the 'setEnrollmentKey' function
is left unused, and thus was removed, without any effect on the tests.
```

Packed a small commit with it as will to avoid the double CI time.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -288,10 +288,6 @@ public class EnrollmentManager
         // R, signature noise
         this.signature_noise = this.createSignatureNoise(height);
 
-        // save enroll_key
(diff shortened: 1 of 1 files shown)


linked0 (commented):
It was an absolutely wrong call.
It would be good to remove `getEnrollmentKey` and `enroll_key` also.

linked0 (approved):
The `getEnrollmentKey` and `enroll_key` can be removed in another PR. So I approve of it.

PR #1001, opened 2020-07-02
Test: Use loops instead of copy-pasting code

Geod24:
```
This is a minor refactoring that uses loop and indexes instead
of copy pasting the same code and using `varname2` and `varname3`.
```

Found while refactoring code while addressing [@linked0 's comment](https://github.com/bpfkorea/agora/pull/997#discussion_r448691307).


linked0 (approved):
LGTM. Thanks.

PR #1002, opened 2020-07-03
Replace stateful makeEnrollment with stateless buildEnrollment

Geod24:
```
This allows to build `Enrollment` without changing the state of the Enrollment
stored in the EnrollmentManager, which is required in order to be able to
re-enroll before the end of the validation period.
```

Incidentally, this fixes a bug with `createSignatureNoise`: It shouldn't take a `height`, as it's hard, if not impossible, to know what height was used to create the noise. We moved the pre-image away from the height approach but this part was forgotten.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -255,45 +255,97 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Make an enrollment data for enrollment process
+        Build an `Enrollment` using `buildEnrollment`, stores and returns it
 
         Params:
-            frozen_utxo_hash = the hash of a frozen UTXO used to identify a validator
-                        and to generate a siging key
-            height = the starting index for generating pre-images
-            enroll = will contain the Enrollment if created
+            utxo = The hash of the frozen UTXO used as a stake.
+                   It must be owned by the private key this validator controls
+                   (`key_pair` argument to constructor).
 
         Returns:
-            true if the enrollment manager succeeded in creating the Enrollment
+            The `Enrollment` created by `buildEnrollment`
 
     ***************************************************************************/
 
-    public bool createEnrollment (Hash frozen_utxo_hash, Height height,
-        out Enrollment enroll) @trusted nothrow
+    public Enrollment createEnrollment (Hash utxo) @safe nothrow
     {
         // K, frozen UTXO hash
-        this.data.utxo_key = frozen_utxo_hash;
-        this.enroll_key = frozen_utxo_hash;
-
-        // N, cycle length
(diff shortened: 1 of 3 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -255,45 +255,96 @@ public class EnrollmentManager
...
+            offset = The number of times this private key has enrolled before.
+                     If `seed` is provided, this parameter is non-optional.
+            seed = Random seed to use
+
+        Returns:
+            An `Enrollment` refencing `utxo` signed with `key`
+
+    ***************************************************************************/
+
+    public static Enrollment makeEnrollment (
```
linked0, line 316: This function is used only in unit test, so it would be better with `version (unittest)` modifier, which make developers less confusing.

linked0 (changes requested):
Good work.
I have one suggestion though.

source/agora/consensus/EnrollmentManager.d
Geod24: Good catch! Fixed.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -255,45 +255,97 @@ public class EnrollmentManager
...
+        that accepts a `Hash` and thus does not incur the cost of re-generating
+        the pre-images.
+
+        Params:
+            key  = `KeyPair` that controls the `utxo`
+            utxo = `Hash` of a frozen UTXO value that can be used for enrollment
+            cycle_length = The cycle length to use (see `ConsensusParams`)
+            offset = The number of times this private key has enrolled before.
+                     If `seed` is provided, this parameter is non-optional.
+            seed = Random seed to use
```
TrustHenry: A good reading is done if the sort order of the parameter descriptions is the same as the parameter order of the function.

Geod24: What do you mean ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -255,45 +255,97 @@ public class EnrollmentManager
...
+
+        Returns:
+            An `Enrollment` refencing `utxo` signed with `key`
+
+    ***************************************************************************/
+
+    public static Enrollment makeEnrollment (
+        Pair key, const ref Hash utxo, uint cycle_length, Hash seed, ulong offset)
```
TrustHenry, line 317: ^ Parameter order

Geod24: Alright fixed

Geod24:
Anything else ?

TrustHenry (approved):
LGTM

linked0 (approved):

PR #1005, opened 2020-07-03
Split the Genesis block in 3 modules

### LINKED ISSUE #821: Replace setGenesisBlock / genesis config with hardcoded blocks

Geod24:
This is to ease with #821 .

### DIFF
--- a/source/agora/consensus/data/genesis/Coinnet.d
+++ b/source/agora/consensus/data/genesis/Coinnet.d
@@ -1,0 +1,91 @@
+/*******************************************************************************
+
+    Defines the genesis block used by the live network (CoinNet)
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.genesis.Coinnet;
+
+import agora.common.Amount;
+import agora.common.Hash;
+import agora.common.Types;
+import agora.common.crypto.Key;
+import agora.consensus.data.Block;
+import agora.consensus.data.Transaction;
+
+/// The genesis block as defined by CoinNet
+public immutable Block GenesisBlock =
+{
+    header:
+    {
+        prev_block:  Hash.init,
+        height:      Height(0),
+        merkle_root: GenesisMerkleRoot,
+    },
+    txs: [ GenesisTransaction ],
+    merkle_tree: [ GenesisMerkleRoot ],
+};
+
+///
+unittest
+{
+    assert(GenesisBlock.header.prev_block == Hash.init);
+    assert(GenesisBlock.header.height == 0);
+    assert(GenesisBlock.header.merkle_root == GenesisBlock.merkle_tree[0]);
+    assert(GenesisBlock.merkle_tree.length == 1);
+    assert(GenesisBlock.header.merkle_root == hashFull(GenesisTransaction));
+}
+
+///
+private immutable Hash GenesisMerkleRoot =
+    Hash(`0x5d7f6a7a30f7ff591c8649f61eb8a35d034824ed5cd252c2c6f10cdbd223671`
+         ~ `3dc369ef2a44b62ba113814a9d819a276ff61582874c9aee9c98efa2aa1f10d73`);
+
+/// The single transaction that are part of the genesis block
+private immutable Transaction GenesisTransaction =
+{
+    TxType.Payment,
+    outputs: [
+        Output(Amount(62_500_000L * 10_000_000L), GenesisOutputAddress),
+        Output(Amount(62_500_000L * 10_000_000L), GenesisOutputAddress),
+        Output(Amount(62_500_000L * 10_000_000L), GenesisOutputAddress),
+        Output(Amount(62_500_000L * 10_000_000L), GenesisOutputAddress),
+        Output(Amount(62_500_000L * 10_000_000L), GenesisOutputAddress),
+        Output(Amount(62_500_000L * 10_000_000L), GenesisOutputAddress),
+        Output(Amount(62_500_000L * 10_000_000L), GenesisOutputAddress),
+        Output(Amount(62_500_000L * 10_000_000L), GenesisOutputAddress),
+    ],
+};
+
+// TODO: Replace with the foundation's pubkey
+/// GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
+private immutable PublicKey GenesisOutputAddress = GenesisAddressUbyte;
+
+///
+private immutable ubyte[] GenesisAddressUbyte =
+    [
+        0x9D, 0x02, 0x38, 0xE0, 0xA1, 0x71, 0x40, 0x0B,
+        0xC6, 0xD6, 0x8A, 0x9D, 0x9B, 0x31, 0x6A, 0xCD,
+        0x51, 0x09, 0x64, 0x91, 0x13, 0xA0, 0x5C, 0x28,
+        0x4F, 0x42, 0x96, 0xD2, 0xB3, 0x01, 0x22, 0xF5,
+    ];
+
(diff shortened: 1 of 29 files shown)


source/agora/consensus/data/genesis/Coinnet.d
```diff
@@ -0,0 +1,92 @@
+/*******************************************************************************
+
+    Defines the genesis block used by the live network (CoinNet)
+
+    Copyright:
+        Copyright (c) 2019-202 BOS Platform Foundation Korea
```
AndrejMitrovic:
```suggestion
        Copyright (c) 2019-2020 BOS Platform Foundation Korea
```

source/agora/consensus/data/genesis/Test.d
```diff
@@ -1,17 +1,26 @@
 /*******************************************************************************
 
-    Contains primitives related to the genesis block
+    Defines a genesis block suitable for testing purpose
+
+    This genesis block is used in multiple places:
+    - Most unittests;
+    - Most network unittests;
```
AndrejMitrovic: Just a small note here but it will be used by networking tests in the future, not currently used yet because of: https://github.com/bpfkorea/agora/blob/4fbb4b3738a5e2be6ad03d7cf3bfa3467a5ff03c/source/agora/test/Base.d#L1019

source/agora/consensus/data/genesis/Test.d
```diff
@@ -1,17 +1,26 @@
...
+    - The system unit tests;
+    - The system integration tests;
+
+    The keys in this modules are well-known, and hence not suitable for anything
```
AndrejMitrovic:
```suggestion
    The keys in this module are well-known, and hence not suitable for anything
```

source/agora/consensus/data/genesis/package.d
```diff
@@ -0,0 +1,120 @@
...
+}
+
+
+/*******************************************************************************
+
+    Build a Genesis block from the provided data
+
+    This function is intended to be used when the hard-coded genesis block
+    needs to be replaced, e.g. when the parameters are changed or a new network
+    is created. It shouldn't be used in production code.
```
AndrejMitrovic: The term "production code" is a bit ambiguous here. Did you mean it shouldn't be used for CoinNet? Or that it should only be used for unittest / system integration tests?

source/agora/consensus/data/genesis/package.d
```diff
@@ -0,0 +1,120 @@
...
+*******************************************************************************/
+
+public immutable(Block) makeGenesis (
+    Transaction[] txs, Enrollment[] enrolls, Signature delegate(Hash) sigcb)
+{
+    import agora.consensus.validation.Block : isGenesisBlockInvalidReason;
+
+    Block genesis;
+    if (enrolls.length > txs.length)
+        throw new Exception("Cannot have more Enrollments than Transactions");
```
AndrejMitrovic: It seems like this check belongs in `isGenesisBlockInvalidReason`?

source/agora/consensus/data/genesis/package.d
```diff
@@ -0,0 +1,120 @@
...
+
+    Build a Genesis block from the provided data
+
+    This function is intended to be used when the hard-coded genesis block
+    needs to be replaced, e.g. when the parameters are changed or a new network
+    is created. It shouldn't be used in production code.
+
+*******************************************************************************/
+
+public immutable(Block) makeGenesis (
```
AndrejMitrovic, line 81: The function doesn't seem to be used anywhere? I assume the signature parameter is a delegate to allow signing with different private keys? We don't want to just pass a private key as a parameter here?

AndrejMitrovic: Also maybe `makeGenesisBlock` in `Base.d` should use this function now?

AndrejMitrovic (commented):
leftover comments from yesterday

source/agora/consensus/data/genesis/package.d
Geod24: For any "live" system. If one wishes to make their own, forked version of BOA, they shouldn't use those keys either.

source/agora/consensus/data/genesis/package.d
Geod24: I suppose I could remove this since the UTXOs wouldn't be found, or move it if the cost of `isInvalidGenesisBlock` matters (I guess not)

source/agora/consensus/data/genesis/Coinnet.d
```diff
@@ -0,0 +1,92 @@
...
+///
+private immutable Hash GenesisMerkleRoot =
+    Hash(`0xdb6e67f59fe0b30676037e4970705df8287f0de38298dcc09e50a8e85413959ca`
+         ~ `4c52a9fa1edbe6a47cbb6b5e9b2a19b4d0877cc1f5955a7166fe6884eecd2c3`);
+
+/// The single transaction that are part of the genesis block
+private immutable Transaction GenesisTransaction =
+{
+    TxType.Payment,
+    inputs: [ Input.init ],
```
TrustHenry:
```suggestion
    inputs: [],
```
Inputs must be empty.
Or remove the `inputs` code line.

source/agora/consensus/data/genesis/Coinnet.d
```diff
@@ -0,0 +1,92 @@
...
+{
+    header:
+    {
+        prev_block:  Hash.init,
+        height:      Height(0),
+        merkle_root: GenesisMerkleRoot,
+    },
+    txs: [ GenesisTransaction ],
+    merkle_tree: [ GenesisMerkleRoot ],
+};
```
MichaelKim20, line 34: Since there is no Enrollment data on CoinNet, we need another way to succeed in the **System Integration Test**.

source/agora/consensus/data/genesis/package.d
Geod24: As mentioned in the comment, this is here so that changing the genesis block is not such as hassle as it is now.
I might move it to a `utils` module at some point.

source/agora/consensus/data/genesis/package.d
Geod24: Changed to `non-test`

source/agora/consensus/data/genesis/Test.d
Geod24: Added a comment

source/agora/consensus/data/genesis/package.d
AndrejMitrovic: Ahh ok.

source/agora/consensus/data/genesis/package.d
Geod24: Removed

source/agora/consensus/data/genesis/Coinnet.d
Geod24: It still uses the same data after the change. So yeah the Coinnet block needs to be modified, but it belong to a separate PR, this is just a refactoring.

AndrejMitrovic:
Same here, if rebased it should go green now.

Geod24:
Rebased

Geod24:
Green!

AndrejMitrovic:
LGTM 🌲🌲

AndrejMitrovic (approved):

PR #1007, opened 2020-07-05
Fix #581: Remove or document usage of internal state in tests

### LINKED ISSUE #581: Remove usage of internal state in network tests
Our network tests are not supposed to use internal state of the nodes, they are intended to be black box testing.
However, we currently have 7 functions exposed in the API: https://github.com/bpfkorea/agora/blob/0cdc928b09eddec41f78a2c1a36459dac532a9c6/source/agora/test/Base.d#L427-L450

Out of those 7, two of them are just exposing functions which are normally private.
While not great, it is fine and justifiable.
The extra `printLog` function is also self-evident: If a test fails, we can call it to get some extra information.

The following 4 are much less obvious: `metaAddPeer`, `createEnrollmentData`, `getPreimage` and `updateEnrolledHeight`.
The later was added very recently, and the comment about it is in the test that uses it:
https://github.com/bpfkorea/agora/blob/0cdc928b09eddec41f78a2c1a36459dac532a9c6/source/agora/test/EnrollmentManager.d#L140-L142

Definition of done:
- [ ] The documentation about `TestAPI` to include the dos and donts
- [ ] Each function exposing internals documents why it needs to be there
- [ ] Hopefully, some of the functions are removed from the public API. I can't think of a reason to have `getPreimage` and `createEnrollmentData`. I remember `metaAddPeer` had a justification in the PR, and `updateEnrolledHeight` is a temporary solution.

Geod24:
```
The worst offenders listed in the issue have already been removed,
so this simply document the ones that are here to stay (start, printLog),
remove metaAddPeer and related, as the test succeeds even without calling
'TestAPIManager.addMetadata', and document createEnrollmentData and the API itself.
```

Fixes #581 (or more exactly, the last remnants of it)


linked0 (approved):
LGTM

PR #1008, opened 2020-07-05
Use a random seed in the quorum generator

### LINKED ISSUE #990: Quorum generator should use a random seed derived from the preimages
The current quorum algorithm design:

- A new (validated) block is about to be applied to the Ledger
- The validator set at the new block height changes (`onValidatorsChanged()` was called)
- The quorum configuration is regenerated

Currently the quorum generator derives its source of randomness from the [node's public key](https://github.com/bpfkorea/agora/blob/de69e629d20aa5d9647f64dfe40cab5cace81e07/source/agora/consensus/Quorum.d#L293). 

We want to add entropy to the this randomness (make it "more" random), while keeping the algorithm verifiable (given a specific set of parameters it should always generate the exact same quorum). Therefore we want to add a "random seed" parameter. This random seed value should be derived from the validator's preimages. 

What needs to be done:
- [ ] There needs to be an easy way to get the preimage of all active validators for a **specific block height**. Look into `ValidatorSet.getPreimage` as an example (hint: for preimages which are too new we need to hash them until we find the preimage at the desired height).
- [ ] The random seed should be derived from the concatenation of all the active validator's preimages for the new block height. As a trivial example: `R = hashMulti(S1, S2, S3)` where S1, S2, S3 are each validator's preimage at the given block height. Note that you will probably **not** use hashMulti for this because you don't know the number of parameters in advance. **It was just an example**. It's also possible to just use XOR for this.
- [ ] `buildQuorumConfigs` should take an additional random seed parameter, a hash parameter.
- [ ] `getGenerator` should take the random seed and hash it together with the node's public key, and feed it to `toShortHash`.
- [ ] `toShortHash` will need to be changed to take a `Hash` instead of a `PublicKey`

This will solve one leftover item in https://github.com/bpfkorea/agora/issues/240: `Validators are distributed among quorums in an unpredictable manner`.

AndrejMitrovic:
I will see if it's possible to add an additional networking test that tests that the exact expected quorum is generated. We don't expose this outside the node yet.

I've also found a small blocker which I've resolved with a workaround. Currently after generating the genesis block the nodes do not send each other any preimages for the enrollments in that genesis block.

Resolves https://github.com/bpfkorea/agora/issues/990

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -553,6 +553,41 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Generate the random seed reduced from the preimages for the provided
+        block height.
+
+        Params:
+            keys = the keys to look up (must be sorted)
+            height = the desired block height to look up the images for
+
+        Returns:
+            the random seed
+
+    ***************************************************************************/
+
+    public Hash getRandomSeed (in Hash[] keys, in Height height) @safe nothrow
+    in
+    {
+        assert(keys.length != 0);
+        assert(keys.isStrictlyMonotonic!((a, b) => a < b));
+    }
+    do
+    {
+        Hash rand_seed;
+        foreach (const ref key; keys)
+        {
+            const preimage = this.validator_set.getPreimageAt(key, height);
+            // this should not happen. validators which didn't reveal the
+            // preimage should not be in the active validator set 'keys'.
+            assert(preimage != PreImageInfo.init);
+            rand_seed = hashMulti(rand_seed, preimage);
+        }
+
+        return rand_seed;
+    }
+
+    /***************************************************************************
+
         Get validator's pre-image from the validator set.
 
         Params:
@@ -1255,4 +1290,73 @@ unittest
     assert(man.enroll_pool.count() == 1);
     man.getEnrollments(enrolls, Height(validator_cycle + 11));
     assert(enrolls.length == 0);
+}
+
+// test getRandomSeed()
+unittest
+{
+    import agora.common.crypto.Schnorr;
+    import agora.consensus.data.Transaction;
+    import std.conv;
+    import std.range;
+    import std.stdio;
+
+    scope storage = new TestUTXOSet;
+    Hash[] utxos;
+
+    // genesisSpendable returns 8 outputs
+    auto pairs = iota(8).map!(idx => WK.Keys[idx]).array;
+
+    genesisSpendable()
+        .enumerate
+        .map!(tup => tup.value
+            .refund(pairs[tup.index].address)
+            .sign(TxType.Freeze))
+        .each!((tx) {
+            storage.put(tx);
+            utxos ~= UTXOSetValue.getHash(tx.hashFull(), 0);
+        });
+
+    auto params = new immutable(ConsensusParams);
+    scope man = new EnrollmentManager(":memory:", KeyPair.random(), params);
+
+    foreach (idx, kp; pairs)
+    {
+        Pair pair;
+        pair.v = secretKeyToCurveScalar(kp.secret);
+        pair.V = pair.v.toPoint();
+
+        auto cycle = PreImageCycle(
+            0, 0, PreImageCache(PreImageCycle.NumberOfCycles, params.ValidatorCycle),
+            PreImageCache(params.ValidatorCycle, 1));
+
+        const seed = cycle.populate(pair.v, true);
+        const enroll = EnrollmentManager.makeEnrollment(
+            pair, utxos[idx], params.ValidatorCycle,
+            seed, cycle.index);
+        assert(man.addValidator(enroll, Height(1), &storage.findUTXO,
+            storage.storage) is null);
+
+        auto cache = PreImageCache(PreImageCycle.NumberOfCycles, params.ValidatorCycle);
+        cache.reset(hashMulti(pair.v, "consensus.preimages", 0));
+
+        PreImageInfo preimage = { enroll_key : utxos[idx],
+            distance : cast(ushort)params.ValidatorCycle,
+            hash : cache[$ - params.ValidatorCycle - 1] };
+
+        assert(man.addPreimage(preimage));
+    }
+
+    utxos.sort();  // must be sorted by enrollment key
+    assert(man.getRandomSeed(utxos, Height(1)) ==
+        Hash(`0xbc2a03ae4e9f00074ff201425bf5ad330c311dfd5c9ae54d38bfcfe4f2f02dbd99d6a25c975be9228fe4c9833e423bec9cb1039b05f4d0a23ca2c9310b936849`),
+        man.getRandomSeed(utxos, Height(1)).to!string);
+
+    assert(man.getRandomSeed(utxos, Height(504)) ==
+        Hash(`0xfd2e526f102abb279b21dfaa88ced3eae8bc373fbda3a5b377776bf7b5830c0776370fca30ab978f058a0690e05ae7e795ed65cc3cd069236e78ad486d216b61`),
+        man.getRandomSeed(utxos, Height(504)).to!string);
+
+    assert(man.getRandomSeed(utxos, Height(1008)) ==
+        Hash(`0xa9eca761735203ad896929790aa83c03a2154d8390137b2b59e92ca90150220c1992a1e3ebe318ad8049cf801b9a8b85119410131fc3c4d784ce430dd780a861`),
+        man.getRandomSeed(utxos, Height(1008)).to!string);
 }
--- a/source/agora/consensus/Quorum.d
+++ b/source/agora/consensus/Quorum.d
@@ -60,17 +60,20 @@ private enum double THRESHOLD = 0.80;
 /*******************************************************************************
 
     Build the quorum configuration for the given public key and the registered
-    enrollment keys.
+    enrollment keys. The random seed is used to introduce partial randomness
+    to the quorum assignment of each node.
 
     Params:
         key = the key of the node for which to generate the quorum
         utxo_keys = the list of UTXO keys of all the active enrollments
         finder = UTXO finder delegate
+        rand_seed = the random seed
 
 *******************************************************************************/
 
 public QuorumConfig buildQuorumConfig ( const ref PublicKey key,
-    in Hash[] utxo_keys, scope UTXOFinder finder ) @safe nothrow
+    in Hash[] utxo_keys, scope UTXOFinder finder, const ref Hash rand_seed ) 
+    @safe nothrow
 {
     // special-case: only 1 validator is active
     if (utxo_keys.length == 1)
@@ -84,7 +87,7 @@ public QuorumConfig buildQuorumConfig ( const ref PublicKey key,
 
     // for filtering duplicates from dice()
     auto added = BitField!uint(stakes.length);
-    auto RNG_gen = getGenerator(key);
+    auto RNG_gen = getGenerator(key, rand_seed);
     auto stake_amounts = stakes.map!(stake => stake.amount.integral);
 
     static assumeNothrow (T)(lazy T exp) nothrow
@@ -115,7 +118,8 @@ public QuorumConfig buildQuorumConfig ( const ref PublicKey key,
 unittest
 {
     auto keys = getKeys(1);
-    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.only, keys);
+    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.only, keys,
+        hashFull(1));
     verifyQuorumsSanity(quorums);
     verifyQuorumsIntersect(quorums);
     test!"=="(countNodeInclusions(quorums, keys), [1]);
@@ -125,7 +129,8 @@ unittest
 unittest
 {
     auto keys = getKeys(2);
-    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.repeat(2), keys);
+    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.repeat(2), keys,
+        hashFull(1));
     verifyQuorumsSanity(quorums);
     verifyQuorumsIntersect(quorums);
     test!"=="(countNodeInclusions(quorums, keys), [2, 2]);
@@ -135,7 +140,8 @@ unittest
 unittest
 {
     auto keys = getKeys(3);
-    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.repeat(3), keys);
+    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.repeat(3), keys,
+        hashFull(1));
     verifyQuorumsSanity(quorums);
     verifyQuorumsIntersect(quorums);
     test!"=="(countNodeInclusions(quorums, keys), [3, 3, 3]);
@@ -145,7 +151,8 @@ unittest
 unittest
 {
     auto keys = getKeys(4);
-    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.repeat(4), keys);
+    auto quorums = buildTestQuorums(Amount.MinFreezeAmount.repeat(4), keys,
+        hashFull(1));
     verifyQuorumsSanity(quorums);
     verifyQuorumsIntersect(quorums);
     test!"=="(countNodeInclusions(quorums, keys), [4, 4, 4, 4]);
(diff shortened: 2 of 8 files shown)


source/agora/consensus/ValidatorSet.d
```diff
@@ -397,6 +397,61 @@ public class ValidatorSet
...
+                     the preimage's current block height, the preimage will
+                     be hashed until the older preimage is retrieved.
+
+        Returns:
+            the PreImageInfo of the enrolled key if it exists,
+            otherwise PreImageInfo.init
+
+    ***************************************************************************/
+
+    public PreImageInfo getPreimageAt (const ref Hash enroll_key,
```
linked0, line 418: This function is kind of duplicate with the `getPreimage` function. So it would be good to merge this function with the original 'getPreimage`.

AndrejMitrovic:
I'll rewrite the networking test to remove the need to change the genesis block validation workaround. As discussed with Mathias yesterday, we want all the tests to start with the same genesis block.

AndrejMitrovic:
> I'll rewrite the networking test to remove the need to change the genesis block validation workaround. As discussed with Mathias yesterday, we want all the tests to start with the same genesis block.

Updated. Now there is a separate test which does the following:

- Starts with 6 enrolled nodes, validator cycle 10
- At the end of the cycle re-enrolls those validators and adds an additional two outsider validators
- Re-enrolls the 8 validators before they expire

We check the quorum configs each time enrollments change, and it shows that the preimages have an effect on the quorum shuffling.

------------

Now the only problem left is that sometimes the test-suite fails with segfaults. It seems to be reproducible on my machine (but the failure is not always in the same place).

source/agora/consensus/ValidatorSet.d
AndrejMitrovic: I'm not sure I can merge them. The use-cases are different.

Take a look at this call:

https://github.com/bpfkorea/agora/blob/56cd093013a57cd6f37c0fd9f7caaf663d256bd6/source/agora/consensus/ValidatorSet.d#L414

If I add a `height` parameter to `getPreimage`, what do we pass as a height to the `getPreimage` call?

In this case we just want the latest preimage with no condition on the enrollment height or the distance.

linked0: > If I add a `height` parameter to `getPreimage`, what do we pass as a height to the `getPreimage` call?
> 
> In this case we just want the latest preimage with no condition on the enrollment height or the distance.

There is only one pre-image for the `enroll_key` in the table, so it can be just selected and you can calculate the proper pre-image with several hashing.

AndrejMitrovic: Example 1:
- Validator submitted preimage for block height 20
- We just externalized block height 10
- We want to get the preimage at height 10. So we hash the existing preimage 10 times to get the preimage at height 10.

Example 2:

- The Validator wants to submit the preimage for block height 30
- We should not hash anything here, just validate the new preimage and replace the old one.

There are two separate use-cases here.

linked0: I thought the function can process with a default parameter for Example 2. That could make the function complex.
Thank you for your explanation.

AndrejMitrovic: It's not really about complexity. It's not possible to combine the two functions together.

In example 1 if the stored preimage is height 9 but we want a preimage for height 10 then the function should return PreImageInfo.init. So we cannot use `Height(ulong.max)` as the default value because it would return the preimage at height 9.

And if we used `Height(0)` as the default value then the function will unnecessarily hash the value when it's used in example 2.

AndrejMitrovic:
I added this assert into localrest and it triggers:

```diff
diff --git a/source/geod24/LocalRest.d b/source/geod24/LocalRest.d
index e69b6fb..e2d39fa 100644
--- a/source/geod24/LocalRest.d
+++ b/source/geod24/LocalRest.d
@@ -573,6 +573,7 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
         Variant[] await_msgs;

         scope exc = new ExitException();
+        bool has_shutdown;

         void runNode ()
         {
@@ -583,6 +584,9 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
             {
                 static if (is(T == Command))
                 {
+                    if (has_shutdown)
+                        assert(0);
+
                     scheduler.spawn(() => handleCommand(arg, node, control.filter));
                 }
                 else static if (is(T == Response))
@@ -600,6 +604,8 @@ public final class RemoteAPI (API, alias S = VibeJSONSerializer!()) : API
                     C.receiveTimeout(self, 10.msecs,
                         (ShutdownCommand!API e)
                         {
+                            has_shutdown = true;
+
                             if (e.callback !is null)
                                 e.callback(node);
                             exc.restart = e.restart;
```

So it seems that there are still events being handled for the node after `shutdown` was called.

AndrejMitrovic:
Ah I need to set it to false in the restart loop too. Let's see..

AndrejMitrovic:
I'm not sure whether merging https://github.com/bpfkorea/agora/pull/1019 and https://github.com/bpfkorea/agora/pull/1017 had anything to do with it, but the PR is suddenly green.

source/agora/test/Base.d
```diff
@@ -1159,6 +1167,18 @@ private immutable(Block) makeGenesisBlock (in KeyPair[] key_pairs,
         const utxo = UTXOSetValue.getHash(tx.hashFull(), 0);
         enrolls ~= EnrollmentManager.makeEnrollment(
             key_pair, utxo, validator_cycle);
+
+        ulong offset = 0;
+        const kp_v = secretKeyToCurveScalar(key_pair.secret);
+
```
```suggestion
```

source/agora/test/Base.d
```diff
@@ -1159,6 +1167,18 @@ private immutable(Block) makeGenesisBlock (in KeyPair[] key_pairs,
...
+        auto cache = PreImageCache(PreImageCycle.NumberOfCycles, validator_cycle);
+        cache.reset(hashMulti(kp_v, "consensus.preimages", offset));
+
+        PreImageInfo preimage;
+        preimage.enroll_key = utxo;
+        preimage.distance = cast(ushort)(validator_cycle);
```
```suggestion
        preimage.distance = cast(ushort)validator_cycle;
```

source/agora/test/Base.d
```diff
@@ -683,11 +684,15 @@ private mixin template TestNodeMixin ()
...
-    public this (Config config, Registry* reg, immutable(Block)[] blocks)
+    public this (Config config, Registry* reg, immutable(Block)[] blocks,
+        immutable(PreImageInfo)[] preimages)
     {
         this.registry = reg;
         this.blocks = blocks;
         super(config);
+
+        foreach (const ref preimage; preimages)
+            assert(this.enroll_man.addPreimage(preimage));
```
AndrejMitrovic: We already have the `getEnrollmentManager` injection routine. Maybe we can use that instead?

source/agora/node/Validator.d
```diff
@@ -60,6 +60,9 @@ public class Validator : FullNode, API
     /// The current required set of peer keeys to connect to
```
AndrejMitrovic, line 60: This commit needs a commit message.

source/agora/consensus/Quorum.d
```diff
@@ -251,7 +314,7 @@ private const(PublicKey)[] getKeys (size_t count)
 }
 
 /// Create a shorthash from a 64-byte blob for RNG initialization
```
AndrejMitrovic, line 324: Just a note: the comment was previously wrong (it was 32 bytes), but now it's correct.

source/agora/consensus/Quorum.d
```diff
@@ -284,16 +349,18 @@ unittest
 
     Params
         key = the public key of a node
+        rand_seed = the random seed
```
AndrejMitrovic, line 360: This parameter isn't mentioned in the docs above.

source/agora/node/Validator.d
```diff
@@ -134,8 +134,9 @@ public class Validator : FullNode, API
             assert(0);
         }
 
+        const rand_seed = hashFull(1);
```
AndrejMitrovic: Small note: The first commit does not use the preimages as the seed yet, because that change itself requires additional code. So I split it up into two commits.

source/agora/consensus/ValidatorSet.d
```diff
@@ -397,6 +397,61 @@ public class ValidatorSet
...
+
+        Get validator's pre-image for the given block height from the
+        validator set
+
+        Params:
+            enroll_key = The key for the enrollment in which the pre-image is
+                contained.
+            height = the desired preimage block height. If it's older than
+                     the preimage's current block height, the preimage will
+                     be hashed until the older preimage is retrieved.
```
AndrejMitrovic, line 409:
```suggestion
                     be hashed until the older preimage is retrieved.
                     Otherwise PreImageInfo.init is returned.
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -471,6 +471,33 @@ public class EnrollmentManager
...
     }
 
+    /***************************************************************************
+
+        Generate the random seed reduced from the preimages for the provided
+        block height.
+
+        Params:
+            keys = the keys to look up
+            height = the desired block height to look up the images in
```
```suggestion
            height = the desired block height to look up the images for
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -471,6 +471,33 @@ public class EnrollmentManager
...
+
+    ***************************************************************************/
+
+    public Hash getRandomSeed (in Hash[] keys, in Height height) @safe nothrow
+    {
+        Hash rand_seed;
+        foreach (const ref key; keys)
+        {
+            const preimage = this.validator_set.getPreimageAt(key, height);
+            assert(preimage != PreImageInfo.init);  // this should not happen
```
AndrejMitrovic: More specifically: If a validator has not revealed their preimage it means it should be removed from the validator set before this function is called. Therefore the `keys` parameter here will not contain that validator's key anymore.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -471,6 +471,33 @@ public class EnrollmentManager
...
+        Params:
+            keys = the keys to look up
+            height = the desired block height to look up the images in
+
+        Returns:
+            the random seed
+
+    ***************************************************************************/
+
+    public Hash getRandomSeed (in Hash[] keys, in Height height) @safe nothrow
```
AndrejMitrovic, line 568: add assert that keys is not empty

source/agora/consensus/EnrollmentManager.d
```diff
@@ -471,6 +471,33 @@ public class EnrollmentManager
...
+    {
+        Hash rand_seed;
+        foreach (const ref key; keys)
+        {
+            const preimage = this.validator_set.getPreimageAt(key, height);
+            assert(preimage != PreImageInfo.init);  // this should not happen
+            rand_seed = hashMulti(rand_seed, preimage);
```
AndrejMitrovic, line 583: Technically this does `hashMulti(Hash.init, preimage)` on the first loop iteration. However I don't think it really matters.

source/agora/consensus/EnrollmentManager.d
AndrejMitrovic, line 568: Forgot to write a unittest here. There is a networking unittest, but I think this function should be introduced in its own commit with a unittest.

source/agora/test/QuorumPreimage.d
```diff
@@ -0,0 +1,424 @@
+/*******************************************************************************
+
+    Tests preimage quorum configuration behavior.
```
AndrejMitrovic:
```suggestion
    Tests preimage quorum generation behavior.
```

source/agora/test/QuorumPreimage.d
```diff
@@ -0,0 +1,424 @@
...
+    version (none)
+    foreach (idx, node; nodes.enumerate)
+        writefln("Node %s: %s\n", idx, node.getQuorumConfig);
+
+    nodes.enumerate.each!((idx, node) =>
+        retryFor(node.getQuorumConfig() == quorums_1[idx], 5.seconds,
+            format("Node %s has quorum config %s. Expected: %s",
+                idx, node.getQuorumConfig(), quorums_1[idx])));
+
+    Transaction makePayTx (in Transaction prev, PublicKey[] addresses, uint index = 0)
```
AndrejMitrovic: I guess this should use the new TxBuilder routines. I haven't used them yet personally.

source/agora/test/Base.d
AndrejMitrovic: Hmm no it doesn't work. The issue is the enrollment manager must be instantiated first, then the ledger is instantiated (and it takes the manager in the ctor), then the ledger populates the validator set in the enrollment manager when loading the blocks.

So preimages need to be added after the enrollment manager and the ledger are constructed, but before `onValidatorsChanged` has been called.

source/agora/consensus/Quorum.d
```diff
@@ -284,16 +349,18 @@ unittest
...
 
     Returns:
         a Mersenne Twister 64bit random generator
 
 *******************************************************************************/
 
-private auto getGenerator (PublicKey key)
+private auto getGenerator (PublicKey key, Hash rand_seed)
```
AndrejMitrovic:
```suggestion
private auto getGenerator (PublicKey key, const ref Hash rand_seed)
```

AndrejMitrovic: Also for the key.

Geod24:
Not  a fan of the first commit. Rest looks nice, but I'll probably dig more into it a bit later.

AndrejMitrovic:
> Not a fan of the first commit.

Hey that makes two of us!

AndrejMitrovic:
Resolved most comments except the comment about using TxBuilder. I'll do this soon.

AndrejMitrovic:
I updated the last commit to use the TxBuilder.

It's still quite tricky to make this work because of the requirement to always need 8 transactions to build a single block.

source/agora/test/QuorumPreimage.d
Geod24: They are still there tho

AndrejMitrovic: I'll remove these.

source/agora/test/QuorumPreimage.d
```diff
@@ -0,0 +1,436 @@
...
+
+        auto signature = WK.Keys.Genesis.secret.sign(hashFull(tx)[]);
+        tx.inputs[0].signature = signature;
+        return tx;
+    }
+
+    auto spendable = network.blocks[$ - 1].txs
+        .filter!(tx => tx.type == TxType.Payment)
+        .map!(tx => iota(tx.outputs.length)
+        .map!(idx => TxBuilder(tx, cast(uint)idx)))
```
Geod24: That indentation is triggering me

AndrejMitrovic: > suggestion pls

Geod24: The `map` is on `iota`, not on the result of the previous `map`, so +4 indent
```suggestion
            .map!(idx => TxBuilder(tx, cast(uint)idx)))
```

AndrejMitrovic: Oh right. That was a mistake, it's easy to miss.

AndrejMitrovic:
> Not a fan of the first commit.

Right, so the thing that is problematic here is the disconnect between making the enrollment outside of the node (to *create* the genesis block) and the node's enrollment manager itself.

With https://github.com/bpfkorea/agora/pull/798/files the node should be able to load the enrollment from the genesis block and then restore its own preimage and send the latest one to the other nodes.

So I think https://github.com/bpfkorea/agora/pull/798 blocks this PR.

AndrejMitrovic:
I know what's causing the segfaults. It's a bug in the NetworkManager's usage of hashmaps:

https://github.com/bpfkorea/agora/blob/906709a22dfdd7a045163bc1931a64af6419de04/source/agora/network/NetworkManager.d#L727-L737

See if you can spot the problem!

source/agora/consensus/EnrollmentManager.d
```diff
@@ -471,6 +471,37 @@ public class EnrollmentManager
...
+            height = the desired block height to look up the images for
+
+        Returns:
+            the random seed
+
+    ***************************************************************************/
+
+    public Hash getRandomSeed (in Hash[] keys, in Height height) @safe nothrow
+    {
+        assert(keys.length != 0);
```
MichaelKim20, line 571: Shouldn't we check if the hash values in the `keys` are sorted in order?

AndrejMitrovic: Thanks, I'll add the assert.

AndrejMitrovic:
I'll rebase this.

Edit: this will need a bit more work as I ran into #1049

AndrejMitrovic:
Rebased and removed the workaround preimage commit.

source/agora/test/Base.d
```diff
@@ -686,6 +686,9 @@ public interface TestAPI : ValidatorAPI
     ***************************************************************************/
 
     public abstract Enrollment createEnrollmentData();
+
+    ///
```
Geod24, line 690: This needs an extra comment as per the other methods.

source/agora/consensus/Quorum.d
```diff
@@ -60,17 +60,20 @@ private enum double THRESHOLD = 0.80;
...
         key = the key of the node for which to generate the quorum
         utxo_keys = the list of UTXO keys of all the active enrollments
         finder = UTXO finder delegate
+        rand_seed = the random seed
 
 *******************************************************************************/
 
 public QuorumConfig buildQuorumConfig ( const ref PublicKey key,
-    in Hash[] utxo_keys, scope UTXOFinder finder ) @safe nothrow
+    in Hash[] utxo_keys, scope UTXOFinder finder, const ref Hash rand_seed )
```
Geod24, line 75: `foo (int bar)` not `foo ( int bar )` but that was already the case before this PR

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1256,3 +1291,72 @@ unittest
...
+    import agora.consensus.data.Transaction;
+    import std.conv;
+    import std.range;
+    import std.stdio;
+
+    scope storage = new TestUTXOSet;
+    Hash[] utxos;
+
+    // genesisSpendable returns 8 outputs
+    auto pairs = iota(8).map!(idx => WK.Keys[idx]).array;
```
Geod24, line 1308: We should really use `repeat` more often :)

Geod24: Also here you could also write `WK.Keys.byRange().take(8).array`.

Geod24 (approved):
Amazing, thanks. The comments don't really need to addressed now, apart from the one on getQuorumConfig

source/agora/consensus/Quorum.d
AndrejMitrovic: dfmt bro.

source/agora/consensus/EnrollmentManager.d
AndrejMitrovic: Oh neato. I forgot you added that.

PR #1009, opened 2020-07-06
Emit unittested modules in the CI

AndrejMitrovic:
This should at least let us know which module's unittests sometimes enter an infinite loop.


AndrejMitrovic:
I'll merge this to see what's stalling the tests in other PRs.

PR #1010, opened 2020-07-06
Take `NodeConfig.validator_cycle` into account

Geod24:
It started as a minor refactor until I realized things were not working quite well and the config option was flat out ignored.
See each commit message for an explanation.

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -112,13 +112,12 @@ public class FullNode : API
 
         Params:
             config = Config instance
-            params = the consensus-critical constants
             onValidatorsChanged = delegate to call when the active set of
                                   validators has changed (may be null)
 
     ***************************************************************************/
 
-    public this (const Config config, immutable(ConsensusParams) params,
+    public this (const Config config,
         void delegate () nothrow @trusted onValidatorsChanged = null)
     {
         // custom genesis block provided
@@ -137,7 +136,7 @@ public class FullNode : API
         this.metadata = this.getMetadata(config.node.data_dir);
 
         this.config = config;
-        this.params = params;
+        this.params = new immutable(ConsensusParams)(config.node.validator_cycle);
         this.taskman = this.getTaskManager();
         this.network = this.getNetworkManager(config.node, config.banman,
             config.network, config.dns_seeds, this.metadata, this.taskman);
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -368,9 +368,6 @@ public class TestAPIManager
     /// Contains the initial blockchain state of all nodes
     public immutable(Block)[] blocks;
 
-    /// Parameters for consensus-critical constants
-    public immutable(ConsensusParams) params;
-
     /// convenience: returns a random-access range which lets us access clients
     auto clients ()
     {
@@ -381,10 +378,9 @@ public class TestAPIManager
     protected Registry reg;
 
     ///
-    public this (immutable(Block)[] blocks, immutable(ConsensusParams) params)
+    public this (immutable(Block)[] blocks)
     {
         this.blocks = blocks;
-        this.params = params;
         this.reg.initialize();
     }
 
@@ -404,12 +400,12 @@ public class TestAPIManager
         if (conf.node.is_validator)
         {
             api = RemoteAPI!TestAPI.spawn!TestValidatorNode(conf, &this.reg,
-                this.blocks, this.params, conf.node.timeout.msecs);
+                this.blocks, conf.node.timeout.msecs);
         }
         else
         {
             api = RemoteAPI!TestAPI.spawn!TestFullNode(conf, &this.reg,
-                this.blocks, this.params, conf.node.timeout.msecs);
+                this.blocks, conf.node.timeout.msecs);
         }
 
         this.reg.register(conf.node.address, api.tid());
@@ -681,12 +677,11 @@ private mixin template TestNodeMixin ()
     private immutable(Block)[] blocks;
 
     ///
-    public this (Config config, Registry* reg, immutable(Block)[] blocks,
-        immutable(ConsensusParams) params = null)
+    public this (Config config, Registry* reg, immutable(Block)[] blocks)
     {
         this.registry = reg;
         this.blocks = blocks;
-        super(config, params);
+        super(config);
     }
 
     ///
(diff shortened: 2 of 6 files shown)


source/agora/test/Base.d
```diff
@@ -921,6 +919,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
             retry_delay : test_conf.retry_delay, // msecs
             max_retries : test_conf.max_retries,
             timeout : test_conf.timeout,
+            validator_cycle : cons_params.ValidatorCycle,
```
Geod24, line 922: This probably will go in the future, in favor of adding direct support in `TestConf`.
I'm also considering making 20 the default as we're likely to want to test cycles more often than not.

AndrejMitrovic:
I need the preimages from the enrollment manager in https://github.com/bpfkorea/agora/pull/1008/files, specifically this commit does this: https://github.com/bpfkorea/agora/pull/1008/commits/4c63a7404f11e805c9d7155b949c48f38da668f4

I couldn't think of a better solution yet.

Geod24:
Since we use well-known keypairs, you can just do something like this:
https://github.com/bpfkorea/agora/blob/dd84a839910444b22719a5ef5d3fd905a14ff871/source/agora/consensus/Genesis.d#L215-L231

AndrejMitrovic (approved):

AndrejMitrovic:
@linked0 what do you think?

linked0 (approved):
LGTM

AndrejMitrovic:
I'm not sure this was correct? It's generating the exact same `random_seed` for all the enrollments..

AndrejMitrovic:
```diff
diff --git a/source/agora/test/Base.d b/source/agora/test/Base.d
index 09c10691..dd0b051f 100644
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -1159,6 +1159,8 @@ private immutable(Block) makeGenesisBlock (in KeyPair[] key_pairs,
         const utxo = UTXOSetValue.getHash(tx.hashFull(), 0);
         enrolls ~= EnrollmentManager.makeEnrollment(
             key_pair, utxo, validator_cycle);
+        import std.stdio;
+        writefln("Enrollment: %s", enrolls[$ - 1]);
     }

     enrolls.sort!("a.utxo_key < b.utxo_key");
```

The second value is the seed:

```
Enrollment: Enrollment(0xb82cb96710af2e9804c59d1f1e1679f8b8b69f4c0f6cd79c8c12f365dd766c09aaa4febcc18b3665d33301cb248ac7afd343ac7b98b27beaf246ad12d3b3219a, 0xeea3fe41624a1071e073ba7b95ab277a3d358c57fd08daa4f5c67da5e47c9b1a2df84d201171d817db6169264c3ddc336e9401ccb53d18f3c3a5cf841c6c5b68, 10, 0x08948b2beb8d7c670f76308ea91bada39f21e04aaa98e6faad57bd9ac03fe72459681feeb9a3b68b18bca05ac811c75bf557a1029f110dd8b34ecb9720affe00)
Enrollment: Enrollment(0x4028965b7408566a66e4cf8c603a1cdebc7659a3e693d36d2fdcb39b196da967914f40ef4966d5b4b1f4b3aae00fbd68ffe8808b070464c2a101d44f4d7b0170, 0xeea3fe41624a1071e073ba7b95ab277a3d358c57fd08daa4f5c67da5e47c9b1a2df84d201171d817db6169264c3ddc336e9401ccb53d18f3c3a5cf841c6c5b68, 10, 0x0e6793d9c42301b03ba720f71d97ab6bade27c445e63e5ac51dda8d7f437ab2ee363196eed3f7b1efafe646bc5ef7816b3203cc4fdaa8c6d70260cb46391d06a)
Enrollment: Enrollment(0x190dd12450d9f1972e21e39d676d9e5bb690071d98ed5c0859077e409741912a4baf07dabc2c879725f0af67f783bbb0161c68d55ef6835639eaaedff2317561, 0xeea3fe41624a1071e073ba7b95ab277a3d358c57fd08daa4f5c67da5e47c9b1a2df84d201171d817db6169264c3ddc336e9401ccb53d18f3c3a5cf841c6c5b68, 10, 0x04f4cd55789338558fef622f8f331ee23f75960ad9be8d9ec9d135f0b644caece70b0b8f35a4d70a8bfd8c53183816ddaae73eefcbdc8ad715452b0ed62fedf0)
Enrollment: Enrollment(0x81a326afa790003c32517a2a2556613004e6147edac28d576cf7bcc2daadf4bb60be1f644c229b775e7894844ec66b2d70ddf407b8196b46bc1dfe42061c7497, 0xeea3fe41624a1071e073ba7b95ab277a3d358c57fd08daa4f5c67da5e47c9b1a2df84d201171d817db6169264c3ddc336e9401ccb53d18f3c3a5cf841c6c5b68, 10, 0x07244880b9eae2486e0a8a451a111872fda78f684f6b06e96c562c6d4d45852ba52bfc3af80cfff8f516625c469ecf18d7355239c66f2274bfec59dc952111f7)
```

Geod24:
Good catch. Fix here: https://github.com/bpfkorea/agora/pull/1013

source/agora/node/FullNode.d
```diff
@@ -137,7 +136,7 @@ public class FullNode : API
         this.metadata = this.getMetadata(config.node.data_dir);
 
         this.config = config;
-        this.params = params;
+        this.params = new immutable(ConsensusParams)(config.node.validator_cycle);
```
AndrejMitrovic, line 139: Wait this doesn't make any sense. Why is this code in `FullNode` instead of `Validator`?

AndrejMitrovic: Ah never mind.. I was confused about the name.

AndrejMitrovic: At first we had `validator_cycle` in the node config. It was supposed to be configurable by each validator node.

Then we decided we will have a hardcoded validator cycle for all validators.

Then we wanted to make it flexible for unittests, so we moved this constant into `ConsensusParams`.

And now we moved back again to step 1, where the validator cycle is configurable via the config.

So my questions is:

- Shouldn't we have just deleted `validator_cycle` from the config? It's supposed to only live in the `ConsensusParams` if it's supposed to be a constant

Geod24: The thing that I missed originally was that we might also want to apply those changes to a system integration test which is not compiled with `unittest`, hence why we have some consensus-critical values in the config (we have Genesis block too).

AndrejMitrovic: Oh right, system integration. Hmm..

So then consensus parameters will always live in the node config?

AndrejMitrovic: IOW would it make sense for these consensus params to just become a ConsensusConfig struct that lives inside of NodeConfig?

PR #1011, opened 2020-07-06
Tests: Move 'ValidatorCycle' argument to TestConf

Geod24:
This allows the configuration to be much more seamless,
and consistent accross the tests.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -847,6 +847,9 @@ public struct TestConf
     /// Number of nodes to instantiate
     size_t nodes = 4;
 
+    /// The value to give the `ValidatorCycle` of nodes
+    uint validator_cycle = 1008;
+
     /// whether to set up the peers in the config
     bool configure_network = true;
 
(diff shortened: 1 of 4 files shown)


source/agora/test/EnrollmentManager.d
```diff
@@ -34,10 +34,12 @@ import core.time;
 unittest
 {
     // generate 1007 blocks, 1 short of the enrollments expiring.
```
AndrejMitrovic: We should update this comment. generates 9 blocks.

source/agora/test/Base.d
```diff
@@ -847,6 +847,9 @@ public struct TestConf
     /// Number of nodes to instantiate
     size_t nodes = 4;
 
+    /// The value to give the `ValidatorCycle` of nodes
+    uint validator_cycle = 1008;
```
linked0, line 851: This constant is a protocol-level value. So I think it should be maintained in the source code.

Geod24: I'm not sure I understand the suggestion ?

source/agora/test/EnrollmentManager.d
Geod24: Added a commit

AndrejMitrovic (approved):

PR #1013, opened 2020-07-06
Fix makeEnrollment to generate different random seed

Geod24:
There was an issue discovered in this unittest-only function,
as the cache was not properly reset, it would always generate
the same random seed.
This function was luckily unittest-only.


linked0 (approved):
LGTM. It was good catch.

AndrejMitrovic:
> It was good catch.

About 2 hours of scratching my head, but thanks. 😄

PR #1020, opened 2020-07-08
Fix race condition in the Ledger test

AndrejMitrovic:
See commit description.


Geod24:
Well, to be fair, the node should support this...

AndrejMitrovic:
> Well, to be fair, the node should support this...

Adding tx's to the pool when they reference unknown utxos?

Geod24:
To some extent, I'd say yes. But since we can't really rely on it right now, let's go with this.

Geod24 (approved):
Merge at will

AndrejMitrovic:
I think we should file an issue for this. I can see us allowing tx's referencing UTXOs if they're at least partially valid. We could add a cleanup routine which periodically cleans up old transactions in the pool, to avoid DDoS.

PR #1022, opened 2020-07-09
Refactor some enrollment manager tests

AndrejMitrovic:


source/agora/consensus/EnrollmentManager.d
```diff
@@ -965,36 +965,39 @@ unittest
     auto utxo_hash1 = utxo_hashes[0];
     auto enrollment = man.createEnrollment(utxo_hash1);
     assert(man.pool.add(enrollment, &storage.findUTXO));
-    assert(man.getValidatorCount(block_height) + 1 == 1);
+    assert(man.getValidatorCount(block_height) == 0);  // not active yet
```
Geod24, line 966: Hum, if it's using  the test genesis  block, is that correct ?

AndrejMitrovic: There's no ledger here that loads the enrollment manager with the validator set.

Geod24 (approved):
Now I remember why the `+ 1` was in place: It was the number of entries in the pool. But I understand why it would be confusing, and this adds a few more asserts, so merge at will.

AndrejMitrovic:
Yeah but unfortunately that doesn't actually test the amount of items in the pool. It just assumes the count. :/

linked0 (approved):

PR #1023, opened 2020-07-09
Work around hashmap invalidation issue (+ small race condition fix)

AndrejMitrovic:
A better fix would be to use the `schedule` primitive as @Geod24 mentioned. But it's not implemented in vibe.d. We'll have to properly fix this later.

-----

Also I caught the second bug (fixed by second commit) by chance. The only bad side of the second bug is that the connection task would remain in the hashmap because `map.remove` could be called before `map[key] =`, so it is of no consequence really.

### DIFF
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -114,6 +114,16 @@ public class NetworkManager
             this.address = address;
             this.onHandshakeComplete = onHandshakeComplete;
             this.onFailedRequest = onFailedRequest;
+        }
+
+        /***********************************************************************
+
+            Start the connection task
+
+        ***********************************************************************/
+
+        public void start ()
+        {
             this.outer.taskman.runTask(&this.connect);
         }
 
@@ -432,6 +442,7 @@ public class NetworkManager
                 this.todo_addresses.remove(address);
                 this.connection_tasks[address] = new ConnectionTask(
                     address, &onHandshakeComplete, &onFailedRequest);
+                this.connection_tasks[address].start();
             }
         }
     }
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
To explain this a little bit further. I'll be using a bit of pseudo-code syntax here:

```D
int[int] map;

map[1] = 10;
map[2] = 20;
map[3] = 30;


void fiber1 ()
{
    foreach (key, val; map)
    {
          sendRequest(key, val);   // does a context switch to another fiber!
    }
}

void fiber2 ()
{
     while (1)
    {
        int index;
        map[++index] = index;
        Fiber.yield();   // or some other call which cause a fiber context switch
    }
}
```

Imagine that the two fibers have been scheduled to run. fiber1 uses an iterator to walk through the hashmap and for each request it does a context switch (this stops the fiber). In the meantime fiber 2 is adding new items to the map. **At some point**, the hashmap will rehash itself. This is done automatically by D's hashmap implementation, it's purpose is to keep hashmap insertions efficient.

The problem here is that rehashing the hashmap may invalidate all existing pointers to it. So the `foreach` loop in fiber 1 may be using an iterator that will point to invalid memory.

Geod24:
> This is done automatically by D's hashmap implementation, it's purpose is to keep hashmap insertions efficient.

Do you know when it happens ? I was looking for it but couldn't find any automatic call.

AndrejMitrovic:
Probably on insertion?

AndrejMitrovic:
I'll take a look in druntime.

MichaelKim20 (commented):
It looks good to me.
I have a question for you.
In this case of PR, don't we combine the two commit into one?

AndrejMitrovic:
> In this case of PR, don't we combine the two commit into one?

They're unrelated to each other. I could submit a separate PR for it, but it seemed unnecessary.

MichaelKim20 (approved):

linked0 (approved):
LGTM

PR #1026, opened 2020-07-10
Use proper Duration typing in the Config

AndrejMitrovic:
I added an overload to `opt` to make this possible.

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -36,6 +36,7 @@ import std.range;
 import std.traits;
 
 import core.stdc.time;
+import core.time;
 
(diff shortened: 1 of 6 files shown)


Geod24:
I think you can just use `get` here, it uses the Config's default if there's any. I'm not quite sure if `opt` is useful anymore.

AndrejMitrovic:
No that doesn't work because it cannot convert to Duration. The type in the actual file config is an integer (e.g. milliseconds), or floating point (e.g. 1.5), but `get` cannot know whether to call .seconds or .msecs, etc.

Geod24 (approved):

PR #1031, opened 2020-07-13
Make BlockStorage.load accept a genesis block as input

Geod24:
This remove the dependency on the global 'GenesisBlock' function.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -112,7 +112,7 @@ public class Ledger
         this.enroll_man = enroll_man;
         this.pool = pool;
         this.onValidatorsChanged = onValidatorsChanged;
-        if (!this.storage.load())
+        if (!this.storage.load(GenesisBlock))
             assert(0);
 
         // ensure latest checksum can be read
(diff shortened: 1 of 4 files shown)


source/agora/node/Ledger.d
```diff
@@ -112,7 +112,7 @@ public class Ledger
         this.enroll_man = enroll_man;
         this.pool = pool;
         this.onValidatorsChanged = onValidatorsChanged;
-        if (!this.storage.load())
+        if (!this.storage.load(GenesisBlock))
```
AndrejMitrovic, line 115: Shouldn't the Node pass the configured genesis block to the ledger, so it can be used here?

AndrejMitrovic: In other words, shouldn't it be a ctor parameter?

Geod24: I thought about it, but it meant having a useless field for the lifetime of the instance, since it's only ever used for `load`.
It felt like a bad design, but if you prefer I can amend the code.

AndrejMitrovic: Ah right this is not a "constant" here, it's changed in the call to setGenesisBlock.

AndrejMitrovic (approved):

PR #1034, opened 2020-07-16
Avoid nominating with duplicated Enrollments

linked0:
If a block is nominated with duplicated Enrollments, we must reject the block with checking that the Enrollments are valid candidates for the proposed block. I create a bad nominator and make it create duplicated enrollments for a network test in this pull request, 

Next time, I will submit another PR checking the result of `Ledger.prepareNominatingSet`.

Relates #900

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -422,6 +422,37 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Check if an enrollment is a valid candidate for the proposed height
+
+        Params:
+            enroll = The enrollment of the target to be checked
+            height = the height of proposed block
+            findUTXO = delegate to find the referenced unspent UTXOs with
+
+        Returns:
+            `null` if the enrollment can be a validator at the proposed height,
+            otherwise a string explaining the reason it is invalid.
+
+    ***************************************************************************/
+
+    public string isInvalidCandidateReason (const ref Enrollment enroll,
+        Height height, scope UTXOFinder findUTXO) @safe nothrow
+    {
+        if (auto fail_reason = enroll.isInvalidReason(findUTXO))
+            return fail_reason;
+
+        const enrolled = this.validator_set.getEnrolledHeight(enroll.utxo_key);
+        if (enrolled != ulong.max &&
+            height < (enrolled + this.params.ValidatorCycle))
+        {
+            return "Enrollment: Duplicated enrollments";
+        }
+
+        return null;
+    }
+
+    /***************************************************************************
+
         Clear up expired validators whose cycle for a validator ends
 
         The enrollment manager clears up expired validators from the set based
--- a/source/agora/test/EnrollmentManager.d
+++ b/source/agora/test/EnrollmentManager.d
@@ -297,3 +297,149 @@ unittest
     assert(network.clients().each!(
                client => client.getPreimage(enroll.utxo_key).distance == 0));
 }
+
+/// Situation: One misbehaving node sends an Enrollment for an
+///            already-enrolled validator.
+/// Expectation: The nomination is rejected.
+unittest
+{
+    import agora.common.Config;
+    import agora.common.Task;
+    import agora.common.crypto.Key;
+    import agora.consensus.protocol.Nominator;
+    import agora.consensus.data.ConsensusData;
+    import agora.node.Ledger;
+    import agora.node.Validator;
+    import agora.network.NetworkManager;
+
+    import geod24.Registry;
+    import core.atomic;
+
+    static class BadNominator : Nominator
+    {
+        private Ledger ledger;
+        private bool is_nominating;
+        private shared(size_t)* runCount;
+
+        /// Ctor
+        public this (NetworkManager network, KeyPair key_pair,
+                     Ledger ledger, TaskManager taskman, shared(size_t)* countPtr)
+        {
+            this.runCount = countPtr;
+            this.ledger = ledger;
+            super(network, key_pair, ledger, taskman);
+        }
+
+        ///
+        public override bool isNominating () @safe @nogc nothrow
+        {
+            return this.is_nominating;
+        }
+
+        ///
+        public override void tryNominate () @safe
+        {
+            // Most of the code, save for the line commented as such, are teken
+            // from `Validator.tryNominate`
+            if (this.is_nominating)
+                return;
+
+            this.is_nominating = true;
+            scope (exit) this.is_nominating = false;
+
+            ConsensusData data;
+            this.ledger.prepareNominatingSet(data);
+            if (data.tx_set.length == 0)
+                return;  // not ready yet
+
+            // This behavior is the only difference with a normal Validator:
+            // It adds an enrollment that is already for a node in the ValidatorSet
+            atomicOp!("+=")(*this.runCount, 1);
+            data.enrolls ~= GenesisBlock.header.enrollments[0];
+
+            // Nominate it
+            auto slot_idx = this.ledger.getBlockHeight() + 1;
+            this.nominate(slot_idx, data);
+        }
+    }
+
+    static class MisbehavingValidator : TestValidatorNode
+    {
+        private shared(size_t)* runCount;
+
+        /// Ctor
+        public this (Config config, Registry* reg, immutable(Block)[] blocks,
+                     shared(size_t)* countPtr)
+        {
+            this.runCount = countPtr;
+            super(config, reg, blocks);
+        }
+
+        ///
+        protected override Nominator getNominator (
+            NetworkManager network, KeyPair key_pair, Ledger ledger,
+            TaskManager taskman)
+        {
+            return new BadNominator(
+                network, key_pair, ledger, taskman, this.runCount);
+        }
+    }
+
+    static class BadAPIManager : TestAPIManager
+    {
+        // Make sure the code in BadNominator gets executed
+        shared size_t runCount;
+
+        ///
+        public this (immutable(Block)[] blocks)
+        {
+            super(blocks);
+        }
+
+        /// see base class
+        public override void createNewNode (Config conf)
+        {
+            if (this.nodes.length == 0)
+            {
+                auto api = RemoteAPI!TestAPI.spawn!MisbehavingValidator(
+                    conf, &this.reg, this.blocks, &this.runCount, conf.node.timeout);
+                this.reg.register(conf.node.address, api.tid());
+                this.nodes ~= NodePair(conf.node.address, api);
+            }
+            else
+                super.createNewNode(conf);
+        }
+    }
+
+    auto network = makeTestNetwork!BadAPIManager(TestConf.init);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto bad_validator = network.clients[0];
+    auto validator = network.clients[1];
+
+    // Sanity check: Check if genesis block has enrollments
+    const b0 = validator.getBlocksFrom(0, 2)[0];
+    assert(b0.header.enrollments.length >= 1);
+
+    // Make a block using Genesis
+    Transaction[] txs = genesisSpendable().take(Block.TxsInBlock).enumerate()
+        .map!(en => en.value.refund(WK.Keys.A.address).sign())
+        .array();
+    txs.each!(tx => validator.putTransaction(tx));
+
+    // Make sure that the code in validator gets executed
+    size_t loopCount;
+    while (atomicLoad(network.runCount) < 1)
+    {
+        // That's at least 5 seconds
+        assert(loopCount < 500);
+        loopCount++;
+        Thread.sleep(10.msecs);
+    }
+
+    // Ensure everyone is at the same level although there is a bad Validator
+    ensureConsistency(network.clients, 1);
+}
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -231,7 +231,7 @@ extern(D):
 
     ***************************************************************************/
 
-    private void nominate (ulong slot_idx, ConsensusData next) @trusted
+    protected void nominate (ulong slot_idx, ConsensusData next) @trusted
     {
         log.info("{}(): Proposing tx set for slot {}", __FUNCTION__, slot_idx);
 
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -398,10 +398,9 @@ public class Ledger
 
         foreach (enroll; data.enrolls)
         {
-            if (auto fail_reason = enroll.isInvalidReason(utxo_finder))
-            {
+            if (auto fail_reason = this.enroll_man.isInvalidCandidateReason(
+                enroll, expect_height, utxo_finder))
                 return fail_reason;
-            }
         }
 
         return null;


source/agora/consensus/EnrollmentManager.d
```diff
@@ -420,6 +420,37 @@ public class EnrollmentManager
...
+
+    ***************************************************************************/
+
+    public string isValidCandidate (const ref Enrollment enroll, Height height,
+        UTXOFinder findUTXO) @trusted nothrow
+    {
+        if (auto fail_reason = enroll.isInvalidReason(findUTXO))
+            return fail_reason;
+
+        const enrolled = this.validator_set.getEnrolledHeight(enroll.utxo_key);
```
TrustHenry, line 444:
```suggestion
        const Height enrolled = this.validator_set.getEnrolledHeight(enroll.utxo_key);
```

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,142 @@ unittest
...
+    auto nodes = network.clients;
+    auto validator = nodes[0];
+    auto bad_validator = nodes[1];
+
+    // Sanity check: Check if genesis block has enrollments
+    const b0 = validator.getBlocksFrom(0, 2)[0];
+    assert(b0.header.enrollments.length >= 1);
+
+    // Set an duplicated enrollment to `BadNominator`
+    BadNominator.dup_en = cast(shared)b0.header.enrollments[0];
```
TrustHenry: I'm not sure if the BadNominator was created.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -420,6 +420,37 @@ public class EnrollmentManager
...
+            height = the height of proposed block
+            findUTXO = delegate to find the referenced unspent UTXOs with
+
+        Returns:
+            `null` if the enrollment can be a validator at the proposed height,
+            otherwise a string explaining the reason it is invalid.
+
+    ***************************************************************************/
+
+    public string isValidCandidate (const ref Enrollment enroll, Height height,
```
AndrejMitrovic: This should be named `isInvalidCandidateReason`, not "isValid". Because you return a string when it's invalid, so `if (isValidCandidate(...))` will be `true` even though it failed validation.

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,142 @@ unittest
...
+            this.is_nominating = true;
+            scope (exit) this.is_nominating = false;
+
+            ConsensusData data;
+            this.ledger.prepareNominatingSet(data);
+            if (data.tx_set.length == 0)
+                return;  // not ready yet
+
+            // Add an enrollment that is already an validator
+            data.enrolls ~= cast(Enrollment)BadNominator.dup_en;
```
AndrejMitrovic: Couldn't you just do `data.enrolls ~= GenesisBlock.header.enrollments[0]`?

source/agora/consensus/EnrollmentManager.d
linked0: The format like "`const` {variable name}` is used in many places and we can catch the type easily in this code. So it's not bad I guess. Thanks.

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,142 @@ unittest
...
+    ensureConsistency(nodes, 1);
+
+    // Make a block in the `bad_validator`
+    txs = txs
+        .map!(ptx => TxBuilder(ptx).refund(WK.Keys.B.address).sign())
+        .array();
+    txs.each!(tx => bad_validator.putTransaction(tx));
+
+    // Ensure the block is not accepted
+    ensureConsistency(nodes, 1);
```
AndrejMitrovic: If you change this to:

```suggestion
    ensureConsistency(nodes, 2);
```

The tests will still pass. It's because the tests are not doing what you thought they were doing. :)

source/agora/test/EnrollmentManager.d
linked0: The `BadNominator` is created in the `createNewNode` of the `BadAPIManager` when the `makeTestNetwork`  is called. The code you point out is to set the duplicated enrollment to the static variable. Thanks.

source/agora/test/EnrollmentManager.d
TrustHenry: ^ That's right. I tested it the same way, but I still pass the test.

AndrejMitrovic: Remember that when you send a transaction to a node, it is gossiped to all other nodes. All of the Validator nodes will nominate their own value. 1 of these nodes here will nominate a bad value (duplicate enrollment), however it doesn't matter because the rest of the nodes will reach consensus on the other nominations which are valid.

AndrejMitrovic: There's a simpler test you can do:

- Spawn all BadValidator nodes (all 4 of them). Then send 8 transactions. Then do a thread.sleep call (to wait for propagation). Then check the block height of all nodes. Is it still height 0? That means nomination failed because the enrollment was rejected.

Actually it is very difficult to write correct networking tests right now because we don't get any information from the nodes - they don't tell us why the block height hasn't changed (was it a bad transaction? bad enrollment? no consensus reached?).

linked0: I'm adding another validator called `OneTimeValidator` that is nominating only one time. Thank you.

linked0:
Ready for review.
I made kind of normal validators that nominate only one time. After making the first block, the misbehaving validator can only nominate a block. As @AndrejMitrovic mentioned, the network doesn't tell us exactly why block height hasn't changed. But, I just tried to make the situation where the misbehaving validator makes the block having duplicated enrollments. Thank you guys for your comments and explanation.

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+
+    // Ensure everyone is at the same level
+    ensureConsistency(nodes, 1);
+
+    // Make a block in the `bad_validator`
+    txs = txs
+        .map!(ptx => TxBuilder(ptx).refund(WK.Keys.B.address).sign())
+        .array();
+    txs.each!(tx => bad_validator.putTransaction(tx));
+
```
AndrejMitrovic, line 442: You need to add a Thread.sleep call here. Otherwise the `ensureConsistency()` check can be too fast, before nomination is even attempted.

linked0: The function has already 2 seconds to wait for the result. Isn't that time enough?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -420,6 +420,37 @@ public class EnrollmentManager
...
+            findUTXO = delegate to find the referenced unspent UTXOs with
+
+        Returns:
+            `null` if the enrollment can be a validator at the proposed height,
+            otherwise a string explaining the reason it is invalid.
+
+    ***************************************************************************/
+
+    public string isInvalidCandidateReason (const ref Enrollment enroll,
+        Height height, UTXOFinder findUTXO) @trusted nothrow
```
Geod24: Why does this need to be `@trusted` ?

Geod24: Also the delegate needs to be `scope`

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+            scope (exit) this.is_nominating = false;
+
+            ConsensusData data;
+            this.ledger.prepareNominatingSet(data);
+            if (data.tx_set.length == 0)
+                return;  // not ready yet
+
+            // Add an enrollment that is already an validator
+            data.enrolls ~= GenesisBlock.header.enrollments[0];
+            // data.enrolls ~= Enrollment.init;
```
Geod24: Remove?

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+        ///
+        public override bool isNominating () @safe @nogc nothrow
+        {
+            return this.is_nominating;
+        }
+
+        ///
+        public override void tryNominate () @safe
+        {
+            // Checks done in `Nominator.tryNominate`
```
Geod24: I would change the comment to say: "// Most of the code, save for the line commented as such, are taken from `Validator.tryNominate`"

Or something similar

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+            this.ledger.prepareNominatingSet(data);
+            if (data.tx_set.length == 0)
+                return;  // not ready yet
+
+            // Add an enrollment that is already an validator
+            data.enrolls ~= GenesisBlock.header.enrollments[0];
+            // data.enrolls ~= Enrollment.init;
+
+            // note: we are not passing the previous tx set as we don't really
+            // need it at this point (might later be necessary for chain upgrades)
```
Geod24: Likewise, remove the comment, it doesn't matter here (it's copy-pasted code, that's what matters).

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+            this.is_nominating = true;
+            scope (exit) this.is_nominating = false;
+
+            ConsensusData data;
+            this.ledger.prepareNominatingSet(data);
+            if (data.tx_set.length == 0)
+                return;  // not ready yet
+
+            // Add an enrollment that is already an validator
+            data.enrolls ~= GenesisBlock.header.enrollments[0];
```
Geod24:
```suggestion
            // This behavior is the only difference with a normal Validator:
            // It adds an enrollment that is already for a node already in the ValidatorSet
            data.enrolls ~= GenesisBlock.header.enrollments[0];
```

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+                this.nodes ~= NodePair(conf.node.address, api);
+            }
+            else
+            {
+                auto api = RemoteAPI!TestAPI.spawn!OneTimeValidator(
+                    conf, &this.reg, this.blocks, conf.node.timeout);
+                this.reg.register(conf.node.address, api.tid());
+                this.nodes ~= NodePair(conf.node.address, api);
+            }
+        }
```
Geod24: Why do we need the `OneTimeValidator` ? I thought injecting a single node would be enough.

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+    }
+
+    auto network = makeTestNetwork!BadAPIManager(TestConf.init);
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.start();
```
Geod24:
```suggestion
    network.start();
    scope(exit) network.shutdown();
    scope(failure) network.printLogs();
```

Because if the network hasn't started, calling `shutdown` / `printLogs` is invaldi

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+    network.waitForDiscovery();
+    auto nodes = network.clients;
+    auto bad_validator = nodes[0];
+    auto validator = nodes[1];
+
+    // Sanity check: Check if genesis block has enrollments
+    const b0 = validator.getBlocksFrom(0, 2)[0];
+    assert(b0.header.enrollments.length >= 1);
+
+    // Make a bloack using Genesis
```
Geod24:
```suggestion
    // Make a block using Genesis
```

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+    Transaction[] txs;
+    txs = genesisSpendable().take(Block.TxsInBlock).enumerate()
```
Geod24: Merge those two statements

source/agora/test/EnrollmentManager.d
```diff
@@ -221,3 +221,188 @@ unittest
...
+    txs.each!(tx => validator.putTransaction(tx));
+
+    // Ensure everyone is at the same level
+    ensureConsistency(nodes, 1);
+
+    // Make a block in the `bad_validator`
+    txs = txs
+        .map!(ptx => TxBuilder(ptx).refund(WK.Keys.B.address).sign())
+        .array();
+    txs.each!(tx => bad_validator.putTransaction(tx));
```
Geod24: There's nothing special about the block being created by `bad_validator` though.
Also, the BadValidator should perhaps have a boolean which shows the code was executed. I will help you with this.

Geod24 (changes requested):
Overall looks good, a few changes / questions.
One thing that is problematic, though, is that nodes *should* reach consensus. Because only one node out of 4 provides bad data, the nomination protocol should weed that out. It currently does not.

source/agora/test/EnrollmentManager.d
linked0: > Why do we need the OneTimeValidator ? I thought injecting a single node would be enough.

I just wanted to make the case when the block won't be made if there is only a bad nominator. But, It may be more natural to have the situation where the nomination protocol weeds bad data out as you suggested.

source/agora/test/EnrollmentManager.d
linked0: > There's nothing special about the block being created by bad_validator though.

Ah, I know what you mean. The bad validator only nominates a block with bad data although it makes valid transactions.

> Also, the BadValidator should perhaps have a boolean which shows the code was executed. I will help you with this.

Yeah, thanks in advance.

Geod24:
Thanks for all the fixes ! Rebased it on `v0.x.x`, slightly edited the commit message (because we nominate `ConsensusData`, not a block, so we don't reject a `Block`).
Added some `shared` magic to ensure that the Nominator code gets executed.

source/agora/test/EnrollmentManager.d
Geod24: Yes, the issue was purely about rejecting that block at *nomination* time. Making it to a block would be a grave violation of consensus.

Geod24 (approved):

source/agora/test/EnrollmentManager.d
Geod24: Are you satisfied with the new code @AndrejMitrovic ?

linked0:
> Thanks for all the fixes ! Rebased it on `v0.x.x`, slightly edited the commit message (because we nominate `ConsensusData`, not a block, so we don't reject a `Block`).
> Added some `shared` magic to ensure that the Nominator code gets executed.

Thanks, I will double check the modified code.

source/agora/test/EnrollmentManager.d
AndrejMitrovic: Yes that's better.

AndrejMitrovic (approved):

source/agora/test/EnrollmentManager.d
AndrejMitrovic: Just to make things clearer for everyone, what I was worried about was:

- checkConsistency(1)
- putTransactions...
- checkConsistency(1)

The last check could be too early because calling putTransaction will start an asynchronous nomination round (remember that tx's are gossiped to all nodes, they will start nominating too).

But Mathias fixed this by having only one checkConsistency() call. Consensus should still be reached even if one node nominates a bad enrollment - because the other nodes will select a valid set of transactions and enrollments.

Issue #1033, opened 2020-07-13
Add event notification to Agora, starting with block reception


Geod24:
> As a developer, I want to programmatically be able to respond to certain events happening in the node.

For the moment, I would recommend to have it simply part of the configuration. In the future we might switch to [SSE](https://en.wikipedia.org/wiki/Server-sent_events) but for the moment let's keep it simple.
The idea is to come up with a list of events, and for each of them, have a section in the config file that specifies an URL to push the payload (HTTP POST) when said event happens.

Suggested events:
- A block is externalized, either via catch-up or direct participation;
- A valid pre-image is received (needed for stoa to continue development);

Definition of done:
- There is a YAML section for as follows:
```yaml
event_handlers:
  block_externalized:
    - http://abcdef.net/push
    - 127.0.0.1:3333/push
  preimage_received:
    - SomeOtherURL
```

When the corresponding event happens, the payload is sent to the URL via an HTTP POST request, with the payload as JSON body. Implementation tip: I would recommend implementing each event handler as a single-method `interface` and use the REST generator, e.g.:
```D
interface BlockExternalizedHandler
{
    @method(HTTPMethod.POST)
    @path("/push")
     // Using `/` for `@path` might be possible, in which case the path is passed to the `RestInterfaceClient`:
     // scope client = new RestInterfaceClient!BlockExternalizedHandler("http://127.0.0.1/push/");
    void push (Block block);
}
```

This way we can easily do dependency injection in the test-suite, and could even build some utilities to use this for testing.

AndrejMitrovic:
> This way we can easily do dependency injection in the test-suite

~I don't follow. How does event handling help with doing dependency injection?~

Edit: Resolved IRL.

MichaelKim20:
Aren't you considering the way Agora responds to external requests?

AndrejMitrovic:
I suggested yesterday that these requests will probably be local only. The Stoa server and the Agora node will run on the same machine.

But maybe @Geod24 thought of other use-cases?

MichaelKim20:
When handling events in D language, which library do you use?

AndrejMitrovic:
I'm not sure why the issue seems complicated. It's just about spawning a separate list of `RestInterfaceClient`'s as set in the configuration. Maybe we need to whiteboard at 4PM.

PR #1036, opened 2020-07-16
Add a node for the event handler to the configuration file

MichaelKim20:
We need to push data outside when a special event occurs.
For this, I implemented to add event handlers to the configuration file.

Related to #1033

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -85,3 +85,14 @@ dns:
 logging:
   # Values: Trace, Info, Warn, Error, Fatal, None (default)
   level: None
+
+################################################################################
+##                               Event Handlers                               ##
+################################################################################
+event_handlers:
+  # URLs to push a data when a block is externalized. (path is "/block_externalized")
+  block_externalized:
+    - http://127.0.0.1:3836/block_externalized
+  # URLs to push a data when a pre-image is updated. (path is "/preimage_received")
+  preimage_received:
+    - http://127.0.0.1:3836/preimage_received
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -87,6 +87,9 @@ public struct Config
 
     /// Logging config
     public LoggingConfig logging;
+
+    /// Event handler config
+    public EventHandlerConfig event_handlers;
 }
 
 /// Node config
@@ -158,6 +161,16 @@ public struct LoggingConfig
 
     /// The logging level
     LogLevel log_level = LogLevel.Error;
+}
+
+/// Configuration for URLs to push a data when an event occurs
+public struct EventHandlerConfig
+{
+    /// URLs to push a data when a block is externalized
+    public immutable string[] block_externalized_handler_addresses;
+
+    /// URLs to push a data when a pre-image is updated
+    public immutable string[] preimage_updated_handler_addresses;
 }
 
 /// Parse the command-line arguments and return a GetoptResult
@@ -265,6 +278,7 @@ private Config parseConfigImpl (ref const CommandLine cmdln, Node root)
         network : assumeUnique(parseSequence("network")),
         dns_seeds : assumeUnique(parseSequence("dns", true)),
         logging: parseLoggingSection("logging" in root, cmdln),
+        event_handlers: parserEventHandlers("event_handlers" in root, cmdln),
     };
 
     enforce(conf.network.length > 0, "Network section is empty");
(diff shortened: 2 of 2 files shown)


doc/config.example.yaml
```diff
@@ -85,3 +85,14 @@ dns:
...
+################################################################################
+##                               Event Handlers                               ##
+################################################################################
+event_handlers:
+  # URLs to push a data when a block is externalized. (path is "/push")
+  block_externalized:
+    - http://127.0.0.1:3836
+  # URLs to push a data when a pre-image is updated. (path is "/push_preimage")
+  preimage_received:
+    - http://127.0.0.1:3836
```
Geod24: Let's make them match the event name for simplicity.
So:
```suggestion
  # URLs to push a data when a block is externalized. (path is "/block_externalized")
  block_externalized:
    - http://127.0.0.1:3836
  # URLs to push a data when a pre-image is updated. (path is "/preimage_received")
  preimage_received:
    - http://127.0.0.1:3836
```

Geod24 (approved):
Nice!

source/agora/common/Config.d
```diff
@@ -160,6 +163,16 @@ public struct LoggingConfig
     LogLevel log_level = LogLevel.Error;
 }
 
+/// Configuration for URLs to push a data when an event occurs
+public struct EventHandlerConfig
+{
+    /// URLs to push a data when a block is externalized
+    public immutable string[] block_externalized_handler_address;
```
AndrejMitrovic: It's not a single address, so the word `address` isn't right. Maybe:

```D
block_externalized_handlers;
block_externalized_handler_addresses;
```

source/agora/common/Config.d
```diff
@@ -560,3 +574,79 @@ unittest
     auto scp_quorum = toSCPQuorumSet(quorum);
     assert(scp_quorum.toQuorumConfig() == quorum);
 }
+
+/*******************************************************************************
+
+    Parse the `event_handlers` config section
+
+    Params:
+        ptr = pointer to the Yaml node containing the loggers configuration
```
AndrejMitrovic: It's not `loggers`. Please be careful when copy-pasting code.

source/agora/common/Config.d
```diff
@@ -560,3 +574,79 @@ unittest
...
+            enforce(node[section].type == NodeType.sequence,
+                format("`%s` section must be a sequence", section));
+        else 
+            return null;
+
+        string[] result;
+        foreach (string item; node[section])
+            result ~= item;
+
+        return result;
```
AndrejMitrovic: The double lookup is unnecessary.

```suggestion
        auto elem = section in node;
        if (elem is null)
            return null;

        enforce(elem.type == NodeType.sequence,
            format("`%s` section must be a sequence", section));

        string[] result;
        foreach (string item; *elem)
            result ~= item;

        return result;
```

doc/config.example.yaml
MichaelKim20: @Geod24 
First of all, thank you for your review.
But I had a better idea, so I tested a few things.
Please check this. 
I think we can unify the handlers in this way.
https://gist.github.com/MichaelKim20/2ee0fd4590cc0a9d56adf5aa0c94b020

source/agora/common/Config.d
```diff
@@ -560,3 +574,79 @@ unittest
...
+    Params:
+        ptr = pointer to the Yaml node containing the loggers configuration
+        c = the parsed command line arguments, for override
+
+    Returns:
+        the parsed event handlers
+
+*******************************************************************************/
+
+private EventHandlerConfig parserEventHandlers (Node* node, const ref CommandLine c)
```
AndrejMitrovic, line 591: Why are you parsing the CommandLine and then not using it?

source/agora/common/Config.d
MichaelKim20: > It's not `loggers`. Please be careful when copy-pasting code.

I made a mistake.

source/agora/common/Config.d
MichaelKim20: `node[section]`,I can get rid of this. That's a good point.

source/agora/common/Config.d
MichaelKim20: That's a good point.

source/agora/common/Config.d
MichaelKim20: You want to add this, right?
```
        if (auto val = section in c.overrides)
            return *val;
```

AndrejMitrovic: No, I am just saying that it should not be a parameter.

AndrejMitrovic: However you have already included it in the latest change, so it's ok to keep it.

AndrejMitrovic (approved):

PR #1038, opened 2020-07-17
Add BlockExternalized and PreImageReceived handler events to Agora

### LINKED ISSUE #1033: Add event notification to Agora, starting with block reception

TrustHenry:
Add event handlers to be used when the block is externalized and when receiving the preimage.

Call `pushBlock` when the block is externalized.
When receiving a preImage, call `pushPreimage`.

Convert data into JSON format is sent to POST using RestInterfaceClient.

TestNetworkManager does not support it.

Related to #1033

### DIFF
--- a/source/agora/api/handler/BlockExternalizedHandler.d
+++ b/source/agora/api/handler/BlockExternalizedHandler.d
@@ -1,0 +1,38 @@
+/*******************************************************************************
+
+    Definitions of the BlockExternalizedHandler
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.api.handler.BlockExternalizedHandler;
+
+import agora.consensus.data.Block;
+
+import vibe.web.rest;
+import vibe.http.common;
+
+public interface BlockExternalizedHandler
+{
+// The REST generator requires @safe methods
+@safe:
+
+    /***************************************************************************
+
+        Externalize block data in JSON format with HTTP POST
+
+        API:
+            See config.event_handlers.block_externalized_handler_addresses
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.POST)
+    @path("/")
+    public void pushBlock (const Block block);
+}
--- a/source/agora/api/handler/PreImageReceivedHandler.d
+++ b/source/agora/api/handler/PreImageReceivedHandler.d
@@ -1,0 +1,38 @@
+/*******************************************************************************
+
+    Definitions of the PreImageReceivedHandler
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.api.handler.PreImageReceivedHandler;
+
+import agora.consensus.data.PreImageInfo;
+
+import vibe.web.rest;
+import vibe.http.common;
+
+public interface PreImageReceivedHandler
+{
+// The REST generator requires @safe methods
+@safe:
+
+    /***************************************************************************
+
+        Push a preImage in JSON format with HTTP POST
+
+        API:
+            See config.event_handlers.preimage_updated_handler_addresses
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.POST)
+    @path("/")
+    public void pushPreImage (const PreImageInfo preimage);
+}
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -23,6 +23,8 @@
 module agora.network.NetworkManager;
 
 import agora.api.Validator;
+import agora.api.handler.BlockExternalizedHandler;
+import agora.api.handler.PreImageReceivedHandler;
 import agora.common.BanManager;
 import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
@@ -823,6 +825,62 @@ public class NetworkManager
         // allocates, called infrequently though
         return format("http://%s:%s", this.node_config.address,
             this.node_config.port);
+    }
+
+    /***************************************************************************
+
+        Instantiates a client object implementing `BlockExternalizedHandler`
+
+        In the default implementation, this returns a `BlockExternalizedHandler`
+
+        Params:
+            address = The address (IPv4, IPv6, hostname) of target Server
+        Returns:
+            A BlockExternalizedHandler to communicate with the server
+            at `address`
+
+    ***************************************************************************/
+
+    public BlockExternalizedHandler getBlockExternalizedHandler
+        (Address address)
+    {
+        import vibe.http.client;
+
+        auto settings = new RestInterfaceSettings;
+        settings.baseURL = URL(address);
+        settings.httpClientSettings = new HTTPClientSettings;
+        settings.httpClientSettings.connectTimeout = this.node_config.timeout;
+        settings.httpClientSettings.readTimeout = this.node_config.timeout;
+
+        return new RestInterfaceClient!BlockExternalizedHandler(settings);
+    }
+
+    /***************************************************************************
+
+        Instantiates a client object implementing `PreImageReceivedHandler`
+
+        In the default implementation, this returns a `PreImageReceivedHandler`
+
+        Params:
+            address = The address (IPv4, IPv6, hostname) of target Server
+        Returns:
+            A PreImageReceivedHandler to communicate with the server
+            at `address`
+
+    ***************************************************************************/
+
+    public PreImageReceivedHandler getPreimageReceivedHandler
+        (Address address)
+    {
+        import vibe.http.client;
+
+        auto settings = new RestInterfaceSettings;
+        settings.baseURL = URL(address);
+        settings.httpClientSettings = new HTTPClientSettings;
+        settings.httpClientSettings.connectTimeout = this.node_config.timeout;
+        settings.httpClientSettings.readTimeout = this.node_config.timeout;
+
+        return new RestInterfaceClient!PreImageReceivedHandler(settings);
     }
 }
 
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -467,7 +467,10 @@ public class FullNode : API
         log.trace("Received Preimage: {}", prettify(preimage));
 
         if (this.enroll_man.addPreimage(preimage))
+        {
             this.network.sendPreimage(preimage);
+            this.pushPreImage(preimage);
+        }
     }
 
     /// GET: /preimage
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -154,7 +154,8 @@ public class FullNode : API
         this.enroll_man = this.getEnrollmentManager(config.node.data_dir,
             config.node, params);
         this.ledger = new Ledger(config.node, params, this.utxo_set,
-            this.storage, this.enroll_man, this.pool, onValidatorsChanged);
+            this.storage, this.enroll_man, this.pool, &this.pushBlock,
+            onValidatorsChanged);
         this.exception = new RestException(
             400, Json("The query was incorrect"), string.init, int.init);
 
@@ -169,6 +170,11 @@ public class FullNode : API
             config.event_handlers.preimage_updated_handler_addresses)
             this.preimage_handlers[address] = this.network
                 .getPreimageReceivedHandler(address);
+
+        // Special case
+        // Block externalized handler is set and push for Genesis block.
+        if (this.block_handlers.length > 0 && this.getBlockHeight() == 0)
+            this.pushBlock(GenesisBlock);
     }
 
     /***************************************************************************
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -14,6 +14,8 @@
 module agora.node.FullNode;
 
 import agora.api.FullNode;
+import agora.api.handler.BlockExternalizedHandler;
+import agora.api.handler.PreImageReceivedHandler;
 import agora.consensus.data.Block;
 import agora.common.Amount;
 import agora.common.BanManager;
@@ -106,6 +108,12 @@ public class FullNode : API
     /// If a custom genesis block is set it will be stored here
     private immutable Block genesis_block;
 
+    /// Block Externalized Handler list
+    protected BlockExternalizedHandler[Address] block_handlers;
+
+    /// PreImage Received Handler list
+    protected PreImageReceivedHandler[Address] preimage_handlers;
+
     /***************************************************************************
 
         Constructor
@@ -149,6 +157,18 @@ public class FullNode : API
             this.storage, this.enroll_man, this.pool, onValidatorsChanged);
         this.exception = new RestException(
             400, Json("The query was incorrect"), string.init, int.init);
+
+        // Make `BlockExternalizedHandler` from config
+        foreach (address;
+            config.event_handlers.block_externalized_handler_addresses)
+            this.block_handlers[address] = this.network
+                .getBlockExternalizedHandler(address);
+
+        // Make `PreImageReceivedHandler` from config
+        foreach (address;
+            config.event_handlers.preimage_updated_handler_addresses)
+            this.preimage_handlers[address] = this.network
+                .getPreimageReceivedHandler(address);
     }
 
     /***************************************************************************
@@ -465,5 +485,66 @@ public class FullNode : API
 
     protected void onAcceptedTransaction () @safe
     {
+    }
+
+    /***************************************************************************
+
+        Push the block data to the `block_handlers` target server list
+        set in config.
+
+        Convert block data to JSON serialization and send it POST using Rest.
+
+        Params:
+            block = externalized block data
+
+    ***************************************************************************/
+
+    private void pushBlock (const Block block) @safe
+    {
+        foreach (address, handler; this.block_handlers)
+        {
+            runTask({
+                try
+                {
+                    handler.pushBlock(block);
+                }
+                catch (Exception e)
+                {
+                    log.error("Error sending block height #{} to {} :{}",
+                        block.header.height, address, e);
+                }
+            });
+        }
+    }
+
+    /***************************************************************************
+
+        Push the preimage to the `preimage_handlers` target server list
+        set in config.
+
+        Convert `PreImageInfo` to JSON serialization and send it POST
+        using Rest.
+
+        Params:
+            preImage = Received `PreImageInfo`
+
+    ***************************************************************************/
+
+    private void pushPreImage (const PreImageInfo pre_image) @safe
+    {
+        foreach (address, handler; this.preimage_handlers)
+        {
+            runTask({
+                try
+                {
+                    handler.pushPreImage(pre_image);
+                }
+                catch (Exception e)
+                {
+                    log.error("Error sending preImage (enroll_key: {}) to {} :{}",
+                        pre_image.enroll_key, address, e);
+                }
+            });
+        }
     }
 }
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -76,9 +76,12 @@ public class Ledger
     /// Node config
     private NodeConfig node_config;
 
-    /// If not null call this delegate if the validator set changed after
-    /// a block was externalized
+    /// If not null call this delegate when a new block was added to the ledger
     private void delegate () nothrow @safe onValidatorsChanged;
+
+    /// If not null call this delegate
+    /// A block was externalized
+    private void delegate (const Block) @safe onAcceptedBlock;
 
     /// Parameters for consensus-critical constants
     private immutable(ConsensusParams) params;
@@ -94,6 +97,8 @@ public class Ledger
             storage = the block storage
             enroll_man = the enrollmentManager
             pool = the transaction pool
+            onAcceptedBlock = optional delegate to call
+                              when a block was added to the ledger
             onValidatorsChanged = optional delegate to call after the validator
                                   set changes when a block was externalized
 
@@ -103,6 +108,7 @@ public class Ledger
         NodeConfig node_config, immutable(ConsensusParams) params,
         UTXOSet utxo_set, IBlockStorage storage,
         EnrollmentManager enroll_man, TransactionPool pool,
+        void delegate (const Block) @safe onAcceptedBlock = null,
         void delegate () nothrow @safe onValidatorsChanged = null)
     {
         this.node_config = node_config;
@@ -111,6 +117,7 @@ public class Ledger
         this.storage = storage;
         this.enroll_man = enroll_man;
         this.pool = pool;
+        this.onAcceptedBlock = onAcceptedBlock;
         this.onValidatorsChanged = onValidatorsChanged;
         if (!this.storage.load(GenesisBlock))
             assert(0);
@@ -246,6 +253,7 @@ public class Ledger
         and add all of its outputs to the UTXO set.
         If there are any enrollments in the block,
         add enrollments to the validator set.
+        If not null call the `onAcceptedBlock` delegate.
 
         Params:
             block = the block to add
@@ -275,6 +283,9 @@ public class Ledger
 
         if (this.onValidatorsChanged !is null && validators_changed)
             this.onValidatorsChanged();
+
+        if (this.onAcceptedBlock !is null)
+            this.onAcceptedBlock(block);
     }
 
     /***************************************************************************
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -563,6 +563,9 @@ public class TestAPIManager
 
 public class TestNetworkManager : NetworkManager
 {
+    import agora.api.handler.BlockExternalizedHandler;
+    import agora.api.handler.PreImageReceivedHandler;
+
     ///
     public Registry* registry;
 
@@ -606,6 +609,20 @@ public class TestNetworkManager : NetworkManager
         cstring _data_dir)
     {
         return new FakeClockBanManager(conf);
+    }
+
+    ///
+    protected final override BlockExternalizedHandler getBlockExternalizedHandler
+        (Address address)
+    {
+        assert(0, "Not supported");
+    }
+
+    ///
+    protected final override PreImageReceivedHandler getPreimageReceivedHandler
+        (Address address)
+    {
+        assert(0, "Not supported");
     }
 }
 


source/agora/api/handler/PreImageReceivedHandler.d
```diff
@@ -0,0 +1,36 @@
...
+        Push a preImage in JSON format with HTTP POST
+
+        API:
+            POST /preimage_received
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.POST)
+    @path("/preimage_received")
+    public void pushPreImage (string[string] preimage);
```
AndrejMitrovic: Just make the type `Preimage` and let the json serializer do its work. There is no need to do it manually.

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,71 @@ public class FullNode : API
...
+            block = externalized block data
+
+    ***************************************************************************/
+
+    protected void pushBlock (const ref Block block) @safe
+    {
+        foreach (BlockExternalizedHandler handler; block_handlers)
+        {
+            runTask({
+                auto jsonStr = block.serializeToJsonString();
```
AndrejMitrovic: This is doing manual work which the vibe.d serializer already does for you under the hood.

And secondly, using a map type was unnecessary.

AndrejMitrovic:
~Hold on, I will find exactly how to make it use Json again.~

Edit: Yeah, so Json serialization is already in place. Just use the types like `Preimage`, `Block` in the `BlockExternalizedHandler` and `PreImageReceivedHandler`.

And don't use hashmaps, it's wasteful.

TrustHenry:
History Agora send 
https://github.com/Geod24/agora/blob/625476b2ba10965ec50fa0c90715b9e930a53b88/source/agora/cli/history/main.d#L94-L96

The code is the code we tested before.

Stoa /push block
https://github.com/bpfkorea/stoa/blob/847c60f82524dc1d91ab302f8037934f0ba1be6c/src/Stoa.ts#L130-L134

If we insert JSON data directly into the `Request.body` , there is no 'block' key, so we need to change our Stoa source.
I wanted to put the block key in the JSON data conversion and I used the hash map as a method.
I removed it because of the double processing part in the @AndrejMitrovic comment.
I need to check if the request.body.data contains the data.
And I applied the suggestion now.

@AndrejMitrovic Thanks for the review comment

source/agora/network/NetworkManager.d
```diff
@@ -824,6 +825,62 @@ public class NetworkManager
...
+        Params:
+            address = The address (IPv4, IPv6, hostname) of target Server
+        Returns:
+            A BlockExternalizedHandler to communicate with the server
+            at `address`
+
+    ***************************************************************************/
+
+    public BlockExternalizedHandler getBlockExternalizedHandler
+        (Address address)
```
AndrejMitrovic, line 845:
```suggestion
    public BlockExternalizedHandler getBlockExternalizedHandler (Address address)
```

source/agora/network/NetworkManager.d
```diff
@@ -824,6 +825,62 @@ public class NetworkManager
...
+        Params:
+            address = The address (IPv4, IPv6, hostname) of target Server
+        Returns:
+            A PreImageReceivedHandler to communicate with the server
+            at `address`
+
+    ***************************************************************************/
+
+    public PreImageReceivedHandler getPreimageReceivedHandler
+        (Address address)
```
AndrejMitrovic, line 873:
```suggestion
    public PreImageReceivedHandler getPreimageReceivedHandler (Address address)
```

source/agora/node/FullNode.d
```diff
@@ -149,6 +157,18 @@ public class FullNode : API
             this.storage, this.enroll_man, this.pool, onValidatorsChanged);
         this.exception = new RestException(
             400, Json("The query was incorrect"), string.init, int.init);
+
+        // Make `BlockExternalizedHandler` from config
+        foreach (block_externalize_address;
+            config.event_handlers.block_externalized_handler_addresses)
+            this.block_handlers ~= this.network.getBlockExternalizedHandler
+                (block_externalize_address);
```
AndrejMitrovic:
```suggestion
            this.block_handlers ~= this.network
                .getBlockExternalizedHandler(block_externalize_address);
```

There is a reason I suggest this. It's to make it easier to find function calls across the codebase.

source/agora/node/FullNode.d
```diff
@@ -149,6 +157,18 @@ public class FullNode : API
...
+
+        // Make `PreImageReceivedHandler` from config
+        foreach (preimage_updated_address;
+            config.event_handlers.preimage_updated_handler_addresses)
+            this.preimage_handlers ~= this.network.getPreimageReceivedHandler
+                (preimage_updated_address);
```
AndrejMitrovic:
```suggestion
            this.preimage_handlers ~= this.network
                .getPreimageReceivedHandler(preimage_updated_address);
```

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,69 @@ public class FullNode : API
...
+        Convert block data to JSON serialization and send it POST using Rest.
+
+        Params:
+            block = externalized block data
+
+    ***************************************************************************/
+
+    public void pushBlock (const ref Block block) @safe
+    {
+        foreach (BlockExternalizedHandler handler; block_handlers)
```
AndrejMitrovic:
```suggestion
        foreach (BlockExternalizedHandler handler; this.block_handlers)
```

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,69 @@ public class FullNode : API
...
+            runTask({
+                try
+                {
+                    // API POST: /block_externalized
+                    handler.pushBlock(block);
+                }
+                catch (Exception e)
+                {
+                    log.error(
+                        "Error sending block height #", block.header.height
```
AndrejMitrovic: Will this actually work? I didn't realize it was supported. I thought we had to use `log.error("block height #{}", height)` - the formatting string.

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,69 @@ public class FullNode : API
...
+                        ," to \"", this.config.node.address, "\": ", e);
```
AndrejMitrovic: This is wrong. It's not sending the block to itself.

AndrejMitrovic: Also in the future you can use this:

```
` to"`
```

This is simpler than using `\"`.

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,69 @@ public class FullNode : API
...
+        Convert `PreImageInfo` to JSON serialization and send it POST using Rest.
+
+        Params:
+            preImage = Received `PreImageInfo`
+
+    ***************************************************************************/
+
+    public void pushPreImage (const ref PreImageInfo preImage) @safe
+    {
+        foreach (PreImageReceivedHandler handler; preimage_handlers)
```
AndrejMitrovic:
```suggestion
        foreach (PreImageReceivedHandler handler; this.preimage_handlers)
```

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,69 @@ public class FullNode : API
...
+                {
+                    // API POST: /preimage_received
+                    handler.pushPreImage(preImage);
+                }
+                catch (Exception e)
+                {
+                    log.error(
+                        "Error sending preImage #",
+                        " (enroll_key: ", preImage.enroll_key, ") to \"",
+                        this.config.node.address, "\": ", e);
```
AndrejMitrovic: Again wrong destination address.

AndrejMitrovic (changes requested):

source/agora/network/NetworkManager.d
TrustHenry: Unfortunately, it's over 80 lines. Are you okay?

AndrejMitrovic:
> If we insert JSON data directly into the Request.body , there is no 'block' key, so we need to change our Stoa source.

So just add another parameter to `pushBlock` then.

```D
public void pushBlock (Hash hash, const Block block);
```

source/agora/network/NetworkManager.d
AndrejMitrovic: Ah sorry, I didn't realize. No it's fine then.

But we should probably think about reducing the length of our symbol names.

source/agora/network/NetworkManager.d
AndrejMitrovic: Ignore this suggestion.

source/agora/network/NetworkManager.d
TrustHenry: I wanted to reduce it, but I can't think of anything. Could you recommend it?
https://github.com/bpfkorea/agora/issues/1033
Actually, I decided to follow the example of the issue.

AndrejMitrovic: It's fine for now. Let's get this through and think about refactoring later.

TrustHenry: That's a good idea.

source/agora/node/FullNode.d
TrustHenry: Yes, it looks much better.

source/agora/node/FullNode.d
TrustHenry: Good catch!

source/agora/node/FullNode.d
TrustHenry: Yes, that's right.
This is the remnant used before adding config.

source/agora/node/FullNode.d
TrustHenry: It's fixed now.

TrustHenry:
Associative Arrays was used to print the target address of the event handler in the event of an error.

TrustHenry:
> > If we insert JSON data directly into the Request.body , there is no 'block' key, so we need to change our Stoa source.
> 
> So just add another parameter to `pushBlock` then.
> 
> ```d
> public void pushBlock (Hash hash, const Block block);
> ```

Post in 'pushBlock' We will just push JSON block data from body and fix respond from stoa.
https://github.com/bpfkorea/stoa/pull/40

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,66 @@ public class FullNode : API
...
+        Convert block data to JSON serialization and send it POST using Rest.
+
+        Params:
+            block = externalized block data
+
+    ***************************************************************************/
+
+    public void pushBlock (const ref Block block) @safe
+    {
+        foreach (handler_address; this.block_handlers.byKey)
```
MichaelKim20: I think this would be better.

```suggestion
        foreach (address, handler; this.block_handlers)
```
Link to reference)
https://dlang.org/spec/hash-map.html#aa_example_iteration

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,66 @@ public class FullNode : API
...
+        using Rest.
+
+        Params:
+            preImage = Received `PreImageInfo`
+
+    ***************************************************************************/
+
+    public void pushPreImage (const ref PreImageInfo preImage) @safe
+    {
+        foreach (handler_address; this.preimage_handlers.byKey)
```
MichaelKim20: I think this would be better.

```suggestion
        foreach (address, handler; this.preimage_handlers)
```
Link to reference)
https://dlang.org/spec/hash-map.html#aa_example_iteration

source/agora/api/handler/BlockExternalizedHandler.d
```diff
@@ -0,0 +1,38 @@
...
+
+import agora.consensus.data.Block;
+
+import vibe.web.rest;
+import vibe.http.common;
+
+public interface BlockExternalizedHandler
+{
+// The REST generator requires @safe methods
+@safe:
```
MichaelKim20, line 24: Is this really necessary?

source/agora/api/handler/PreImageReceivedHandler.d
```diff
@@ -0,0 +1,38 @@
...
+
+import agora.consensus.data.PreImageInfo;
+
+import vibe.web.rest;
+import vibe.http.common;
+
+public interface PreImageReceivedHandler
+{
+// The REST generator requires @safe methods
+@safe:
```
MichaelKim20, line 24: Is this really necessary?

MichaelKim20 (commented):
I think I need to check a few things.
The rest is good.

source/agora/node/FullNode.d
TrustHenry: Good! This looks better.

MichaelKim20 (approved):
LGTM

source/agora/network/NetworkManager.d
```diff
@@ -824,6 +826,62 @@ public class NetworkManager
...
+    {
+        import vibe.http.client;
+
+        auto settings = new RestInterfaceSettings;
+        settings.baseURL = URL(address);
+        settings.httpClientSettings = new HTTPClientSettings;
+        settings.httpClientSettings.connectTimeout = this.node_config.timeout;
+        settings.httpClientSettings.readTimeout = this.node_config.timeout;
+
+        return new RestInterfaceClient!PreImageReceivedHandler(address);
```
linked0: Don't you need to set `settings` instead of the `address`?

source/agora/api/handler/PreImageReceivedHandler.d
AndrejMitrovic: > Is this really necessary?

Don't you see the comment? It says `REST generator requires @safe`

source/agora/node/FullNode.d
```diff
@@ -106,6 +108,12 @@ public class FullNode : API
     /// If a custom genesis block is set it will be stored here
     private immutable Block genesis_block;
 
+    /// Block Externalized Handler list
+    protected BlockExternalizedHandler[string] block_handlers;
```
AndrejMitrovic:
```suggestion
    protected BlockExternalizedHandler[Address] block_handlers;
```

source/agora/node/FullNode.d
```diff
@@ -106,6 +108,12 @@ public class FullNode : API
...
+
+    /// PreImage Received Handler list
+    protected PreImageReceivedHandler[string] preimage_handlers;
```
AndrejMitrovic:
```suggestion
    protected PreImageReceivedHandler[Address] preimage_handlers;
```

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,65 @@ public class FullNode : API
...
+        set in config.
+
+        Convert block data to JSON serialization and send it POST using Rest.
+
+        Params:
+            block = externalized block data
+
+    ***************************************************************************/
+
+    public void pushBlock (const ref Block block) @safe
```
AndrejMitrovic: Why is this public?

And why is it never called?

source/agora/node/FullNode.d
```diff
@@ -466,4 +486,65 @@ public class FullNode : API
...
+
+        Convert `PreImageInfo` to JSON serialization and send it POST
+        using Rest.
+
+        Params:
+            preImage = Received `PreImageInfo`
+
+    ***************************************************************************/
+
+    public void pushPreImage (const ref PreImageInfo preImage) @safe
```
AndrejMitrovic:
```suggestion
    public void pushPreImage (const ref PreImageInfo pre_image) @safe
```

Styling rules.

AndrejMitrovic: Or just `preimage`

source/agora/node/FullNode.d
```diff
@@ -149,6 +157,18 @@ public class FullNode : API
...
+        foreach (block_externalize_address;
+            config.event_handlers.block_externalized_handler_addresses)
+            this.block_handlers[block_externalize_address] = this.network
+                .getBlockExternalizedHandler(block_externalize_address);
+
+        // Make `PreImageReceivedHandler` from config
+        foreach (preimage_updated_address;
+            config.event_handlers.preimage_updated_handler_addresses)
+            this.preimage_handlers[preimage_updated_address] = this.network
+                .getPreimageReceivedHandler(preimage_updated_address);
```
AndrejMitrovic: This can be simplified because the variables are not re-used outside the foreach statement.

```suggestion
        // Make `BlockExternalizedHandler` from config
        foreach (address;
            config.event_handlers.block_externalized_handler_addresses)
            this.block_handlers[address] = this.network.getBlockExternalizedHandler(address);

        // Make `PreImageReceivedHandler` from config
        foreach (address;
            config.event_handlers.preimage_updated_handler_addresses)
            this.preimage_handlers[address] = this.network.getPreimageReceivedHandler(address);
```

source/agora/network/NetworkManager.d
TrustHenry: Not a single value.
Multiple `PreImageRecoveredHandler settings create a new instance `RestInterfaceClient`.

AndrejMitrovic: He means you should write this:
```suggestion
        return new RestInterfaceClient!PreImageReceivedHandler(settings);
```

source/agora/node/FullNode.d
TrustHenry: Good! Use the defined type.

source/agora/node/FullNode.d
TrustHenry: I think it will be called outside the FullNode class.

TrustHenry: I'm preparing for the next PR.

source/agora/node/FullNode.d
TrustHenry: Good! The code is more concise.

source/agora/node/FullNode.d
AndrejMitrovic: > I think it will be called outside the FullNode class.

???

TrustHenry: I'm thinking about when to call.
BlockStorage.saveblock thinks it's the starting point for `pushBlock` delegate events.

AndrejMitrovic: You want BlockStorage to be the trigger point that calls FullNode.pushBlock? You want to add FullNode as an instance to the BlockStorage class?

source/agora/network/NetworkManager.d
TrustHenry: Oh, I didn't see it.
That's right. It's wrong.

AndrejMitrovic:
@TrustHenry you can do something like this:

```diff
diff --git a/source/agora/node/FullNode.d b/source/agora/node/FullNode.d
index 035a2205..a84249c9 100644
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -154,7 +154,8 @@ public class FullNode : API
         this.enroll_man = this.getEnrollmentManager(config.node.data_dir,
             config.node, params);
         this.ledger = new Ledger(config.node, params, this.utxo_set,
-            this.storage, this.enroll_man, this.pool, onValidatorsChanged);
+            this.storage, this.enroll_man, this.pool, &this.pushBlock,
+            onValidatorsChanged);
         this.exception = new RestException(
             400, Json("The query was incorrect"), string.init, int.init);
 
@@ -499,7 +500,7 @@ public class FullNode : API
 
     ***************************************************************************/
 
-    public void pushBlock (const ref Block block) @safe
+    private void pushBlock (const ref Block block) @safe
     {
         foreach (address, handler; this.block_handlers)
         {
diff --git a/source/agora/node/Ledger.d b/source/agora/node/Ledger.d
index edc1e786..de2e7c27 100644
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -79,6 +79,7 @@ public class Ledger
     /// If not null call this delegate if the validator set changed after
     /// a block was externalized
     private void delegate () nothrow @safe onValidatorsChanged;
+    private void delegate (const ref Block) nothrow @safe onAcceptedBlock;
 
     /// Parameters for consensus-critical constants
     private immutable(ConsensusParams) params;
@@ -103,6 +104,7 @@ public class Ledger
         NodeConfig node_config, immutable(ConsensusParams) params,
         UTXOSet utxo_set, IBlockStorage storage,
         EnrollmentManager enroll_man, TransactionPool pool,
+        void delegate (const ref Block) nothrow @safe onAcceptedBlock,
         void delegate () nothrow @safe onValidatorsChanged = null)
     {
         this.node_config = node_config;
@@ -111,6 +113,7 @@ public class Ledger
         this.storage = storage;
         this.enroll_man = enroll_man;
         this.pool = pool;
+        this.onAcceptedBlock = onAcceptedBlock;
         this.onValidatorsChanged = onValidatorsChanged;
         if (!this.storage.load(GenesisBlock))
             assert(0);
@@ -275,6 +278,8 @@ public class Ledger
 
         if (this.onValidatorsChanged !is null && validators_changed)
             this.onValidatorsChanged();
+
+        this.onAcceptedBlock(block);
     }
 
     /***************************************************************************
```

And for preimages you can call `pushPreImage` in the `receivePreimage` method.

TrustHenry:
@AndrejMitrovic Thank you. I applied the suggestion now.
 and added two new commits.

Call pushBlock when the block is externalized.
When receiving a preImage, call pushPreimage.

TrustHenry:
Ready to review!

source/agora/node/Ledger.d
```diff
@@ -80,6 +80,10 @@ public class Ledger
     /// a block was externalized
     private void delegate () nothrow @safe onValidatorsChanged;
 
+    /// If not null call this delegate
+    /// A block was externalized
```
AndrejMitrovic, line 83:
```suggestion
    /// If not null call this delegate when a new block was added to the ledger
```

It does not necessarily mean "externalized", because this delegate will be called when a node is in catch-up mode. For example when calling `getBlocksFrom`.

This is why I don't like to call the handler "BlockExternalizedHandler".

However we don't have to worry about naming now, we can change it later.

source/agora/node/Ledger.d
```diff
@@ -94,6 +98,8 @@ public class Ledger
             storage = the block storage
             enroll_man = the enrollmentManager
             pool = the transaction pool
+            onAcceptedBlock = optional delegate to call
+                              when a block was externalized
```
```suggestion
                              when a block was added to the ledger
```

source/agora/node/Ledger.d
```diff
@@ -103,6 +109,7 @@ public class Ledger
         NodeConfig node_config, immutable(ConsensusParams) params,
         UTXOSet utxo_set, IBlockStorage storage,
         EnrollmentManager enroll_man, TransactionPool pool,
+        void delegate (const ref Block) @safe onAcceptedBlock = null,
```
AndrejMitrovic: This should not be an optional parameter (remove `null`).

AndrejMitrovic: The FullNode always provides a delegate, and a Validator inherits the full node so it has the same behavior. So it is not an optional parameter.

source/agora/node/Ledger.d
```diff
@@ -275,6 +284,9 @@ public class Ledger
 
         if (this.onValidatorsChanged !is null && validators_changed)
             this.onValidatorsChanged();
+
+        if (this.onAcceptedBlock !is null)
```
AndrejMitrovic, line 287: It should never be null.

AndrejMitrovic (changes requested):
Requested a few small changes.

This is a good first step, however it would be best to integrate this code into the NetworkManager in the future. Because it has request retrying support, and the current code does not. However, care needs to be taken (we should never ban event handlers, the ban manager should not be used).

TrustHenry:
Thank you. I applied now all of @AndrejMitrovic's suggestions.

source/agora/node/Ledger.d
TrustHenry: That's right. I was concerned about that.

source/agora/node/Ledger.d
TrustHenry: I also considered not focusing on Block Externalized. I think a few word have been added.

source/agora/node/Ledger.d
```diff
@@ -94,6 +97,8 @@ public class Ledger
             storage = the block storage
             enroll_man = the enrollmentManager
             pool = the transaction pool
+            onAcceptedBlock = optional delegate to call
```
AndrejMitrovic, line 100:
```suggestion
            onAcceptedBlock = delegate to call
```

AndrejMitrovic: Ah @TrustHenry I didn't realize this was also used in unittests.

Maybe making it optional was actually a good idea. I'm sorry to give you more work, it might be worth reverting the change and making it optional (with the `!is null` check).

source/agora/node/Ledger.d
```diff
@@ -538,7 +548,7 @@ version (unittest)
             super(config, params, new UTXOSet(":memory:"),
                 new MemBlockStorage(blocks),
                 new EnrollmentManager(":memory:", config.key_pair, params),
-                new TransactionPool(":memory:"));
+                new TransactionPool(":memory:"), (const ref Block) @safe {});
```
AndrejMitrovic: This is the part I didn't see before. Thanks for making it clearer to me @TrustHenry.

AndrejMitrovic (approved):
Approved, there were just a few issues.

source/agora/node/Ledger.d
TrustHenry: The code has been reverted. OnAcceptedBlock is set to Optional. with the !is null check
Thanks for your review.

TrustHenry:
It's kind of weird.
I didn't see all the CI tests in this PR turn green.
One of the CI tests continues to lag and fails to output logs.

TrustHenry:
@Geod24 ping

source/agora/node/FullNode.d
```diff
@@ -466,4 +495,65 @@ public class FullNode : API
...
+    ***************************************************************************/
+
+    private void pushBlock (const ref Block block) @safe
+    {
+        foreach (address, handler; this.block_handlers)
+        {
+            runTask({
+                try
+                {
+                    handler.pushBlock(block);
```
MichaelKim20, line 518: I have a question.
Will this value(block) be valid here?

https://github.com/bpfkorea/agora/blob/8056af52285f23eb6937102c3bd7662c013eec2c/source/agora/consensus/protocol/Nominator.d#L395

TrustHenry: I'm not sure, 
I think he's asking if the block is consensused block. Is this what you meant?
And do you want to know if here the `pushBlock` call point is okay?

AndrejMitrovic: `pushBlock` is called when a **validated** block is added to the ledger. It is unrelated to nomination.

MichaelKim20: @AndrejMitrovic 
When a local variable created by another function is approached within the runTask,
The local variable's contents are missing.
How should I solve this?
I upload the sample code I made together.
Please.
https://gist.github.com/MichaelKim20/0c626a54dfd1f68593fb886da059c20b

AndrejMitrovic: Ah that's right, we should not use `ref` here. The reference may be out of scope when the task starts.

@TrustHenry you will need to remove `ref` from the function parameter.

TrustHenry: Yes, it certainly does.
Thank you.
I will apply `pushPreImage` as well.

TrustHenry:
CI test was performed after the rebase(https://github.com/bpfkorea/agora/pull/1046) but failed to pass ubuntu-18.04 ldc-master.
Previously, errors only occurred in macOS-10.15,ldc-1.22.0 

https://github.com/bpfkorea/agora/pull/1038/checks?check_run_id=896701409

````
+ dchatty=1
+ timeout -s SEGV 8m ./build/agora-unittests
./build/agora-unittests(+0x772b47)[0x5608d25c2b47]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x128a0)[0x7f6f86a6e8a0]
/lib/x86_64-linux-gnu/libpthread.so.0(pthread_cond_wait+0x243)[0x7f6f86a699f3]
./build/agora-unittests(+0x772f20)[0x5608d25c2f20]
./build/agora-unittests(+0x7602ba)[0x5608d25b02ba]
./build/agora-unittests(+0x760028)[0x5608d25b0028]
./build/agora-unittests(+0x1b2ee3)[0x5608d2002ee3]
./build/agora-unittests(+0x1b1960)[0x5608d2001960]
./build/agora-unittests(+0x7729d1)[0x5608d25c29d1]
./build/agora-unittests(+0x77e4cb)[0x5608d25ce4cb]
./build/agora-unittests(+0x77e3f3)[0x5608d25ce3f3]
./build/agora-unittests(+0x77e24e)[0x5608d25ce24e]
./build/agora-unittests(+0x1e6865)[0x5608d2036865]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xe7)[0x7f6f860d6b97]
./build/agora-unittests(+0x7c65a)[0x5608d1ecc65a]
##[error]The operation was canceled.
```

TrustHenry:
The run time of the build&test run task stopped at 29m 21s.

MichaelKim20 (approved):

TrustHenry:
@Geod24 Ready to review.

source/agora/node/FullNode.d
```diff
@@ -466,4 +495,65 @@ public class FullNode : API
...
+    ***************************************************************************/
+
+    private void pushPreImage (const PreImageInfo pre_image) @safe
+    {
+        foreach (address, handler; this.preimage_handlers)
+        {
+            runTask({
+                try
+                {
+                    handler.pushPreImage(pre_image);
```
TrustHenry, line 549: There is a problem that I think.
When matching `PreImageInfo` information with `enrolled enrolment` in `Stoa`
If there are multiple `enrolled enrolments` with the same UTXO_key,
We have to think about which `PreImageInfo` matches which of the two other `Enrollments`.
Because there is no information about `PreImageCycle` in `Stoa`.
Therefore, we need to make sure that `Hashing` matches `Random seed` of `Enrollment` by doing `Distance` of `PreImageInfo`.
However, it can be inefficient if Stoa does Hashing every time it receives `Preimage`.

AndrejMitrovic: > If there are multiple enrolled enrolments with the same UTXO_key,

Multiple enrollments with the same UTXO? That shouldn't be possible.

linked0: > Multiple enrollments with the same UTXO? That shouldn't be possible.

He means continuous enrollments with the same UTXO. He's worried about the situation where some applications could proceed during a period of more than one enrollment. So he wants to know which enrollment is the one for a specific pre-image among continuous enrollments.

TrustHenry: > Multiple enrollments with the same UTXO? That shouldn't be possible.

@AndrejMitrovic  Hold on plz ,,, Most likely to use the same UTXO.

@linked0  That's right.

TrustHenry: Here's what @linked0  describes as a `continuous enrollments`
For example, multiple enrollments(continuous enrollments) mean to the PreImageCycle 0 enrollment and the PreImageCycle 1 enrollment.

TrustHenry: The word `multiple enrollments` seems to cause confusion, so I'll call it `continuous  enrollments`

AndrejMitrovic: But they cannot be active at the same block height, right?

TrustHenry: > But they cannot be active at the same block height, right?

Yes, I'd like to believe that way.

AndrejMitrovic:
Can you please rebase? I wonder if the LocalRest fix is related to the timeout problem..

AndrejMitrovic:
This is strange, now there is a different error.

AndrejMitrovic:
SIGBUS instead of SIGSEGV, which means it was not the `timeout` which killed the binary.

TrustHenry:
It's strange, but I'll try again.

TrustHenry:
@Geod24 do you want to look into it more, or it looks good to you?

source/agora/node/FullNode.d
AndrejMitrovic: So is there a problem here or not? Is the PR ready to go?

Geod24 (approved):

source/agora/node/FullNode.d
TrustHenry: > So is there a problem here or not? Is the PR ready to go?

I think Stoa needs to figure it out.

PR #1041, opened 2020-07-21
unittests: link Phobos with debugging symbols

AndrejMitrovic:
I couldn't find a way to append to `dflags`, so it's copy-pasting the C++ flag here too.


Geod24 (approved):

Geod24:
> The plural of information is information

TIL

PR #1042, opened 2020-07-21
System integration tests: Set Genesis block based on the nodes

Geod24:
As a client, we don't know what is the "correct" genesis block at the moment.
The best option so far is to call `setGenesisBlock` using node-provided data.


AndrejMitrovic (approved):

tests/system/source/main.d
```diff
@@ -80,6 +85,7 @@ int main (string[] args)
 
         checkBlockHeight(addresses, 1);
     }
+    // Warning: GenesisBlock() is invalid here
```
linked0: Could you explain the reason why the GenesisBlock is invalid in this line, 88?

linked0 (approved):
LGTM

AndrejMitrovic:
I think if you rebase this it will go green.

tests/system/source/main.d
Geod24: Sure! I expanded the description.

Geod24:
Good thing is that this is not block #1005 anymore. But it's still not passing for unknown reasons.

Geod24:
Green :O

PR #1044, opened 2020-07-21
Refactor a Ledger test to be more straightforward

Geod24:
```
It was actually quite hard to tell what this test intended to do due to the copy-pasting
and how much duplication there was.
By using TxBuilder and std.algorithm, we can drastically reduce the number of LoCs
while improving readability. The test description was also updated to reflect its intent.
```

Note: I think this test might be redundant, but my intent was solely to clean the legacy stuff in the ledger, we can consolidate things later.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -836,7 +836,9 @@ private KeyPair[] getRandomKeyPairs ()
     return res;
 }
 
-// Use a transaction with the type 'TxType.Freeze' to create a block and test UTXOSet.
+/// Situation : Create two blocks, one containing only `Payment` transactions,
+///             the other containing only `Freeze` ones
+/// Expectation: Block creation succeeds
 unittest
 {
     NodeConfig config = {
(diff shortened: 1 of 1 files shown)


source/agora/node/Ledger.d
```diff
@@ -836,7 +836,9 @@ private KeyPair[] getRandomKeyPairs ()
     return res;
 }
 
-// Use a transaction with the type 'TxType.Freeze' to create a block and test UTXOSet.
+/// Situation : Create two blocks, one containing only `Payment` transactions,
+///             the other containing only `Freeze` ones
+/// Expectation: Block creationg succeeds
```
```suggestion
/// Expectation: Block creation succeeds
```

AndrejMitrovic (approved):

source/agora/node/Ledger.d
Geod24: Fixed

linked0 (approved):
Good, It's a necessary refactoring.

PR #1045, opened 2020-07-21
Ledger tests: Simplify splitGenesisTransaction

Geod24:
```
Since splitGenesis and splitGenesisTransaction were only called from a single place,
they were inlined into the calller.
Additionally, splitGenesisTransaction was further simplified by using TxBuilder and
ranges, which makes its actual goal much more readable.
```


TrustHenry (approved):
LGTM

PR #1046, opened 2020-07-21
CI: Implement a script-side timeout and retry mechanism

Geod24:
Currently the CI is extremely flakey and that affects development speed.
This is a band-aid in an attempt to reduce the failures.

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -3,7 +3,15 @@
 set -xeu
 set -o pipefail
 
-dchatty=1 dub test -b unittest-cov --skip-registry=all --compiler=${DC}
+# Only build the unittest binary, but don't run it
+# We want to run it ourselves to catch any bug / set timeout, etc...
+dub build -b unittest-cov -c unittest --skip-registry=all --compiler=${DC}
+
+dchatty=1
+# A run currently (2020-07-21) takes < 6 minutes on Linux
+# Try a total of three times
+timeout -s SEGV 8m ./build/agora-unittests || timeout -s SEGV 8m ./build/agora-unittests || timeout -s SEGV 8m ./build/agora-unittests
+
 rdmd --compiler=${DC} ./tests/runner.d --compiler=${DC} -cov
 dub build --skip-registry=all --compiler=${DC}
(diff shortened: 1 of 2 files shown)


ci/run.sh
```diff
@@ -3,7 +3,13 @@
 set -xeu
 set -o pipefail
 
-dchatty=1 dub test -b unittest-cov --skip-registry=all --compiler=${DC}
+dchatty=1 timeout -v -s SEGV 12m dub test -b unittest-cov --skip-registry=all --compiler=${DC}
```
AndrejMitrovic: It might actually be useful to build the unittest binary and run this binary rather than dub. That way when we  segfault it we could generate a core dump. (also needs `ulimit -u unlimited` set).

So:

```
dub build -c unittest -b unittest-cov --skip-registry=all --compiler=${DC}
dchatty=1 timeout -v -s SEGV 12m ./build/agora-unittests
```

What do you think?

AndrejMitrovic: Or would it make no difference?

Geod24:
It worked. See ldc-1.22 on Mac OS where it had to retry to get it to pass.

ci/run.sh
```diff
@@ -3,7 +3,15 @@
...
 
-dchatty=1 dub test -b unittest-cov --skip-registry=all --compiler=${DC}
+# Only build the unittest binary, but don't run it
+# We want to run it ourselves to catch any bug / set timeout, etc...
+dub build -b unittest-cov -c unittest --skip-registry=all --compiler=${DC}
+
+dchatty=1
+# A run currently (2020-07-21) takes < 6 minutes on Linux
+# Try a total of three times
+timeout -s SEGV 8m ./build/agora-unittests || timeout -s SEGV 8m ./build/agora-unittests || timeout -s SEGV 8m ./build/agora-unittests
```
AndrejMitrovic, line 13: gotta love bash :D

AndrejMitrovic (approved):

ci/run.sh
Geod24: Well the `$?` solution didn't work because we `set -e`. If we used the default behavior I could've looped.

PR #1047, opened 2020-07-21
Check the result of `Ledger.prepareNominatingSet`

### LINKED ISSUE #900: Avoid nominating and externalizing Enrollments that will fail Block validation

linked0:
This checks the validity of a `ConsensusData` before nominating it. It should never happen, so we make this a `log.fatal` call.

Fixes #900

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -179,6 +179,14 @@ extern(D):
         if (data.tx_set.length == 0)
             return;  // not ready yet
 
+        // check whether the consensus data is valid before nominating it.
+        if (auto msg = this.ledger.validateConsensusData(data))
+        {
+            log.fatal("tryNominate(): Invalid consensus data: {}. Data: {}",
+                    msg, data);
+            return;
+        }
+
         // note: we are not passing the previous tx set as we don't really
         // need it at this point (might later be necessary for chain upgrades)
         auto slot_idx = this.ledger.getBlockHeight() + 1;


source/agora/consensus/protocol/Nominator.d
```diff
@@ -179,6 +179,13 @@ extern(D):
         if (data.tx_set.length == 0)
             return;  // not ready yet
 
+        // check whether the consensus data is valid before nominating it.
+        if (auto msg = this.ledger.validateConsensusData(data))
+        {
+            log.fatal("tryNominate(): Invalid consensus data: {}", msg);
```
AndrejMitrovic: It's not fatal if the node continues running. Fatal means the node will shut down.

AndrejMitrovic: Ok it looks like @Geod24 wants this to be more than `log.error`, but we don't have `log.critical` for this purpose. So `log.fatal` is the next best thing, despite its name.

Keep this changeset.

AndrejMitrovic: Do we want to log out the actual consensus data too and not just the message?

linked0: > Do we want to log out the actual consensus data too and not just the message?

Logging out the data would be good.

Geod24:
Can you rebase ?

Geod24 (approved):
Merge at will

source/agora/consensus/protocol/Nominator.d
```diff
@@ -179,6 +179,14 @@ extern(D):
         if (data.tx_set.length == 0)
             return;  // not ready yet
 
+        // check whether the consensus data is valid before nominating it.
+        if (auto msg = this.ledger.validateConsensusData(data))
```
AndrejMitrovic, line 183: ~Although I question why this check is in here rather than inside of `prepareNominatingSet`. It seems like leaky abstractions?~

Question answered on slack.

PR #1053, opened 2020-07-23
sqlite: Don't use multiple handles to the same file

AndrejMitrovic:
Part of: https://github.com/bpfkorea/agora/issues/836

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -146,10 +146,10 @@ public class EnrollmentManager
             /* preimages: */ PreImageCache(this.params.ValidatorCycle, 1)
         );
 
-        this.validator_set = new ValidatorSet(db_path, params);
-        this.enroll_pool = new EnrollmentPool(db_path);
-
         this.db = new ManagedDatabase(db_path);
+        this.validator_set = new ValidatorSet(this.db, params);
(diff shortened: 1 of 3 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -146,10 +146,10 @@ public class EnrollmentManager
             /* preimages: */ PreImageCache(this.params.ValidatorCycle, 1)
         );
 
-        this.validator_set = new ValidatorSet(db_path, params);
-        this.enroll_pool = new EnrollmentPool(db_path);
-
         this.db = new ManagedDatabase(db_path);
+        this.validator_set = new ValidatorSet(this.db, params);
```
AndrejMitrovic, line 150: Alternatively, maybe these should be separate files. But I'm not sure what's more ideal here.

TrustHenry: How about applying it in WAL(Write-Ahead Logging) option?
https://www.sqlite.org/wal.html

AndrejMitrovic: It seems somewhat unrelated. It enables a reader and writer at the same time, but in our case we have multiple writers to the same file.

Anyway it's worth looking into WAL later, sure.

Geod24:
Tested, works!

Geod24 (approved):

PR #1054, opened 2020-07-23
Update LocalRest to get range error fix

AndrejMitrovic:
@Geod24 I am not fan of this manually maintained submodules readme. What do we need it for? 

And the commit date for the previous localrest update was wrong (July 22 instead of July 21).

### DIFF
--- a/submodules/README.md
+++ b/submodules/README.md
@@ -3,7 +3,7 @@
 This module contains the dependencies of Agora.
 Following is a recap of their capabilities, generated from [this site](https://www.tablesgenerator.com/markdown_tables).
 
-This data can also be gathered by running `git submodules`.
+This data can also be gathered by running `git submodule`.
 
 
 | Dependency name  | Version      | Commit date | Write access |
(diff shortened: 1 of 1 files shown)


submodules/README.md
```diff
@@ -3,7 +3,7 @@
 This module contains the dependencies of Agora.
 Following is a recap of their capabilities, generated from [this site](https://www.tablesgenerator.com/markdown_tables).
 
-This data can also be gathered by running `git submodules`.
+This data can also be gathered by running `git submodule`.
```
TrustHenry, line 6: Nice Catch!

linked0 (approved):

TrustHenry (approved):

Geod24:
> @Geod24 I am not fan of this manually maintained submodules readme. What do we need it for?

It was just to have an overview, especially the repository owned. But now that we have many forks, the point is almost moot.

> And the commit date for the previous localrest update was wrong (July 22 instead of July 21).

Looks like I used the date at which we updated Agora instead of the commit date.

PR #1056, opened 2020-07-23
Do not store the enrollment in the EnrollmentManager

linked0:
We use the `addValidator` function on restoring validators from the chain. The function also restores the enrollment information of the node. So, we do not need to use the table for the `enroll_key`. And the `data` variable is not used anymore, so I remove it from the `EnrollmentManager`.


Geod24 (approved):
LGTM, just a nit on the title

PR #1057, opened 2020-07-23
Build with address sanitizer

AndrejMitrovic:
This shows some interesting stack overflow bugs on both Linux and Mac OS.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -64,7 +64,7 @@
         "source/scpp/build/uint128_t.obj"
     ],
 
-    "dflags": [ "-extern-std=c++14" ],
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
I think one or more of these errors are a false-positive. The address where it's triggered is in druntime here:

https://github.com/ldc-developers/druntime/blob/cc97ccd00d4082221eee1d5afdbd775201d75877/src/rt/dwarfeh.d#L95-L99

But that function is clearly labeled for what it does.

Edit: What needs to be done here is to set a whitelist for the address sanitizer so it doesn't trigger there, and then find if there are any other locations where there are legitimate use-after-free and similar bugs.

PR #1060, opened 2020-07-24
Fix memory leak in restoreSCPState() and fix O(n) loading

AndrejMitrovic:


AndrejMitrovic:
See also: https://github.com/bpfkorea/agora/blob/c06e8b534119039f0a104899262c15c976bda71c/source/scpd/scp/SCP.d#L95

We need to start using this at some point.

Geod24 (approved):

PR #1063, opened 2020-07-27
Remove default configuration interface from integration tests

Geod24:
We're already using the default value, so there's no need to specify it.


TrustHenry (approved):
LGTM

linked0 (approved):

PR #1065, opened 2020-07-27
Optimize (de)serialization of static arrays of small types

Geod24:
Don't deserialize elements one by one if the type is some chars,
`ubyte`, or `bool`.
Also do validation for chars, and remove the `BitBlob` special case,
since it implicitly converts to a static array.

### DIFF
--- a/source/agora/common/Serializer.d
+++ b/source/agora/common/Serializer.d
@@ -407,22 +407,19 @@ public void serializePart (T) (scope const auto ref T record, scope SerializeDg 
                                CompactMode compact = CompactMode.Yes)
     @safe
 {
-    import geod24.bitblob;
-
     // Custom serialization handling trumps everything else
     static if (hasSerializeMethod!T)
         record.serialize(dg);
 
-    // BitBlob are fixed size and thus, value types
-    // If we use an `ubyte[]` overload, the length gets serialized
-    else static if (is(T : BitBlob!N, size_t N))
-        dg(record[]);
-
     // Static array needs to be handled before arrays (because they convert)
     else static if (is(T : E[N], E, size_t N))
     {
-        foreach (ref elem; record)
-            serializePart(elem, dg, compact);
+        // Small type optimization
+        static if (!hasSerializeMethod!E && (E.sizeof == 1 || isSomeChar!E))
+            dg(cast(const(ubyte[]))record);
+        else
+            foreach (ref elem; record)
+                serializePart(elem, dg, compact);
     }
 
     // Strings can be encoded as binary data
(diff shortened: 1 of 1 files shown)


source/agora/common/Serializer.d
```diff
@@ -407,22 +407,19 @@ public void serializePart (T) (scope const auto ref T record, scope SerializeDg
                                CompactMode compact = CompactMode.Yes)
```
AndrejMitrovic, line 407: Commit message:

> if the type is some chars, `ubyte`, or `bool`.

Did you mean "is one of: char, ubyte, or bool?

"type is some chars" sounds really strange.

Geod24: [I didn't pick the name](https://dlang.org/phobos/std_traits.html#isSomeChar)

source/agora/common/Serializer.d
```diff
@@ -407,22 +407,19 @@ public void serializePart (T) (scope const auto ref T record, scope SerializeDg
...
-    else static if (is(T : BitBlob!N, size_t N))
-        dg(record[]);
-
     // Static array needs to be handled before arrays (because they convert)
     else static if (is(T : E[N], E, size_t N))
     {
-        foreach (ref elem; record)
-            serializePart(elem, dg, compact);
+        // Small type optimization
+        static if (!hasSerializeMethod!E && (E.sizeof == 1 || isSomeChar!E))
```
AndrejMitrovic, line 418: `isSomeChar` will also return true for `dchar` which are 4 bytes, is this desired? And if it is, then why not write `E.sizeof == 4`?

source/agora/common/Serializer.d
AndrejMitrovic: I mean that's literally adding whitespace to a function name and calling it English. nono I will send you to English class with Michael & Henry.

source/agora/common/Serializer.d
Geod24: Returns true for `char`, `wchar`, `dchar`, and their qualified version.
Chars are a bit special, because they require validation. Casting straight from `ubyte[]` to `Char[]` is an error, even when `is(Char == dchar)` because the value could be outside the range of valid `dchar`.
But still, we don't want to serialize the `char` element piece-meal.

Looking at this from another angle, why can't we just do this optimization for any static array that doesn't have an element which defines `serialize` ? Well because this only checks one level down. Which makes me think, this approach might be a bit too simple...

Geod24: Yeah I don't think it's worth it (ATM) to check for nested static array, so let's keep it simple.

source/agora/common/Serializer.d
Geod24: Fixed 😊

source/agora/common/Serializer.d
```diff
@@ -588,27 +595,38 @@ public T deserializeFull (T) (scope const(ubyte)[] data) @safe
...
+            return process()[0 .. N];
         }
+        // `bytes` and `bool` and qualified
+        else static if (!hasFromBinaryFunction!E && E.sizeof == 1)
+            return (() @trusted { return (cast(E[]) dg(N)); })()[0 .. N];
         // Note: This does not allocate because `staticMap` yields a tuple
-        return [ Repeat!(N, deserializeEntry) ];
+        else
+        {
+            E deserializeEntry () ()
```
AndrejMitrovic, line 624: Doesn't this allocate a closure? Or is the compiler smart enough to optimize it?

Geod24: Compiler smart as far as I recall. See the comment. I would add a `@nogc` test if I could.

AndrejMitrovic (approved):

PR #1070, opened 2020-07-28
Remove and replace most network topologies

AndrejMitrovic:
This will enable writing a few tests to finally close issue #716.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -972,7 +972,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         ban_duration: 300
     };
 
-    Config makeConfig (size_t idx, NodeConfig self, NodeConfig[] node_confs)
+    Config makeMainConfig (size_t idx, NodeConfig self, NodeConfig[] node_confs)
     {
         auto other_nodes =
             node_confs
@@ -1005,7 +1005,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
     auto keys = refRange(&key_range);
 
     NodeConfig[] node_configs;
-    Config[] configs;
+    Config[] main_configs;
 
     keys.take(test_conf.validators)
         .each!(key => node_configs ~= makeNodeConfig(true, key));
@@ -1014,7 +1014,8 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
 
     // network these nodes together using the configured NetworkTopology
     node_configs.enumerate.each!(
-        pair => configs ~= makeConfig(pair.index, pair.value, node_configs));
+        pair => main_configs ~= makeMainConfig(
+            pair.index, pair.value, node_configs));
 
     NodeConfig[] extra_node_configs;
     keys.take(test_conf.outsider_validators)
@@ -1022,12 +1023,13 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
     keys.take(test_conf.outsider_full_nodes)
         .each!(key => extra_node_configs ~= makeNodeConfig(false, key));
 
-    // generate the outsider configs. pair.index is correct here despite
+    // generate the outsider main_configs. pair.index is correct here despite
     // passing 'node_configs'. in 'MinimallyConnected' mode the connection will be
     // n2 <- n1 <- n2 and n1 <- o1 and n2 <- o2, (o = outsider).
     // otherwise it will be [n1 <-> n2] <- o1 and [n1 <-> n2] <- o2
     extra_node_configs.enumerate.each!(
-        pair => configs ~= makeConfig(pair.index, pair.value, node_configs));
+        pair => main_configs ~= makeMainConfig(
+            pair.index, pair.value, node_configs));
 
     auto gen_block = makeGenesisBlock(
         node_configs
@@ -1039,14 +1041,14 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         .serializeFull()
         .toHexString();
 
-    foreach (ref conf; configs)
+    foreach (ref conf; main_configs)
         conf.node.genesis_block = gen_block_hex;
 
     immutable(Block)[] blocks = generateBlocks(gen_block,
         test_conf.extra_blocks);
 
     auto net = new APIManager(blocks);
-    foreach (ref conf; configs)
+    foreach (ref conf; main_configs)
         net.createNewNode(conf);
 
     return net;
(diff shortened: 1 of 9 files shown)


source/agora/test/Base.d
```diff
@@ -974,108 +972,61 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
...
-    case NetworkTopology.TwoOutsiderValidators:
-        node_configs ~= iota(test_conf.nodes).map!(idx => makeNodeConfig(true, WK.Keys[idx])).array;
-        configs = iota(test_conf.nodes)
-            .map!(idx => makeConfig(node_configs[idx], node_configs)).array;
-        break;
-    }
+    test_conf.validators.iota
+        .each!(_ => node_configs ~= makeNodeConfig(true, nextKey()));
+    test_conf.full_nodes.iota
+        .each!(_ => node_configs ~= makeNodeConfig(false, nextKey()));
```
Geod24: That looks like a prime use case for `zip`.

Geod24: Or `refRange`

source/agora/test/NetworkManager.d
```diff
@@ -180,6 +180,7 @@ unittest
 
     // node test will accept its blocks from node_validator,
     // as the blocks in node_bad do not pass validation
-    retryFor(node_test.getBlockHeight() == 10, 4.seconds);
+    retryFor(node_test.getBlockHeight() == 10, 4.seconds,
+        node_test.getBlockHeight().to!string);
```
Geod24: Yo

Geod24 (approved):
Sweeeeeeet, love it!

image

source/agora/test/Base.d
AndrejMitrovic: But refrange doesn't change the semantics..

source/agora/test/NetworkManager.d
AndrejMitrovic: Sorry. But it was useful during testing. :P

source/agora/test/Base.d
AndrejMitrovic: Or I just don't know how to use it properly?

```D
import std.range;
import std.stdio;

void main ()
{
    auto keys = iota(5);

    auto r = refRange(&keys);

    foreach (x; r)
        writeln(x);
    foreach (x; r)
        writeln(x);
}
```

Prints the same range twice.

Geod24:
```D
auto keys = WK.Keys.byRange().refRange();
keys.take(test_conf.validators).each!(key => node_configs ~= makeNodeConfig(true, key));
keys.take(test_conf.full_nodes).each!(key => node_configs ~= makeNodeConfig(false, key));
```

Would this work ?

AndrejMitrovic: First of all refRange takes a pointer.

AndrejMitrovic: (don't shoot the messenger)

source/agora/test/Base.d
```diff
@@ -843,42 +843,36 @@ public class TestValidatorNode : Validator, TestAPI
...
-    /// One FullNode is not part of the network for any other nodes
-    OneFullNodeOutsider,
-
-    /// Two Validators are not part of the network for any other nodes
-    TwoOutsiderValidators,
+    /// The nodes know about each other's IPs,
+    /// and additionally outsider nodes will connect to them.
+    FullyConnected,
+
+    /// The nodes are connected in a chain: v1 < v2 < v3 < v1,
```
linked0: The `<` looks like an inequality sign, so how about using `<-`.

source/agora/test/Base.d
```diff
@@ -843,42 +843,36 @@ public class TestValidatorNode : Validator, TestAPI
...
+    /// and additionally outsider nodes are minimally connected to them
+    /// via v1 < o1, v2 < o2 (o = outsider node)
```
linked0: Same as the above comment.

source/agora/test/Base.d
Geod24: You got bamboozled by `foreach`. I have a bug report open for this.
```D
import std.range;
import std.stdio;
import std.algorithm;

void main ()
{
    auto keys = iota(5);

    auto r = refRange(&keys);
    r.each!writeln;
    r.each!writeln;
}
```

FTFY

AndrejMitrovic: > /Users/andrejmitrovic/Downloads/ldc2-1.22.0-osx-x86_64/bin/../import/std/range/package.d(11502,13): Error: cannot modify struct instance *this._range of type Range because it contains const or immutable members

haha ranges go compiler errrrrror

source/agora/test/Base.d
```diff
@@ -974,108 +972,61 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
...
+            network : assumeUnique(network),
         };
 
         return conf;
     }
 
+    auto keys = WK.Keys.byRange();
+    auto nextKey () { scope (exit) keys.popFront(); return keys.front; }
     NodeConfig[] node_configs;
     Config[] configs;
```
linked0: I know the name `configs` has been used from start but it looks confusing. The `node_configs` and `configs` were confusing to me as a newcomer. How about using `main_configs` instead of `configs`.

source/agora/test/Base.d
AndrejMitrovic: I'm reducing it.

linked0 (commented):
Great!
And, I have some comments.

source/agora/test/Base.d
AndrejMitrovic: I actually had this change in mind too! I'll fix it.

source/agora/test/Base.d
AndrejMitrovic: Reduced test-case:

```D
import std.range;

public static auto byRange ()
{
    static struct Range
    {
        size_t lbound = 0;

        // trigger point (remove const and it compiles)
        const size_t hbound = 100;

        size_t length () const { return 0; }
        bool empty () const { return false; }
        int front () const { return 0; }
        void popFront () { }
    }

    return Range(0);
}

void main ()
{
    auto keys = byRange();
    auto r = refRange(&keys);
}
```

Time to submit it..

AndrejMitrovic: we can also work around it by making `hbound` an enum

AndrejMitrovic:
Fixed `const` => `enum` for the workaround, using `refRange` now, fixed the `<` to `<-`, and renamed `config` => `main_config`.

Geod24 (approved):

linked0 (approved):

Issue #1071, opened 2020-07-28
Preimage revealing unittest is not actually testing preimage revealing


AndrejMitrovic:
https://github.com/bpfkorea/agora/blob/9d0f3b9ea188e3d98e7c0b0cf337756812da052e/source/agora/test/EnrollmentManager.d#L92-L99

This is not doing what it's claiming to do. It is only testing the **commitment** - meaning the enrollment, not any preimages.

For example:

```diff
diff --git a/source/agora/test/EnrollmentManager.d b/source/agora/test/EnrollmentManager.d
index de298c19..1b4647ce 100644
--- a/source/agora/test/EnrollmentManager.d
+++ b/source/agora/test/EnrollmentManager.d
@@ -97,6 +97,10 @@ unittest
                  node.getPreimage(enroll_2.utxo_key) != PreImageInfo.init &&
                  node.getPreimage(enroll_3.utxo_key) != PreImageInfo.init,
             5.seconds));
+
+    import std.stdio;
+    writefln("Preimage for %s: %s", enroll_0.utxo_key,
+        nodes[0].getPreimage(enroll_0.utxo_key));
 }

 // Test for re-enroll before the validator cycle ends
```

This will print out:

```
Preimage for 0xb82cb96710af2e9804c59d1f1e1679f8b8b69f4c0f6cd79c8c12f365dd766c09aaa4febcc18b3665d33301cb248ac7afd343ac7b98b27beaf246ad12d3b3219a: PreImageInfo(0xb82cb96710af2e9804c59d1f1e1679f8b8b69f4c0f6cd79c8c12f365dd766c09aaa4febcc18b3665d33301cb248ac7afd343ac7b98b27beaf246ad12d3b3219a, 0xabb640e7c36df355074971ae3dc98759b5a80ee0270f98973b9045c9000f6c583ebf425ffb08e5c84466a1a4521c004141268c34aa34f6c224759d66b9b69251, 0)
```

This has a preimage height of zero - it's the commitment.

-----

The timer here should be configurable in the config:

https://github.com/bpfkorea/agora/blob/9d0f3b9ea188e3d98e7c0b0cf337756812da052e/source/agora/node/Validator.d#L189

For testing purposes I set it to a small value (1 second) and I shuffled the tests a bit:

```diff
diff --git a/source/agora/node/Validator.d b/source/agora/node/Validator.d
index 91754fdf..94bb350d 100644
--- a/source/agora/node/Validator.d
[...]

Issue #772, opened 2020-04-27
system integration test node is connecting to itself


AndrejMitrovic:
While writing a test with 8 **total** nodes, I've noticed this in the log:

> node-0_1  | 2020-04-27T01:55:50.623040704Z 2020-04-27 01:55:50,621 Info [agora.network.NetworkManager] - Discovery reached. 8 peers connected.

And also: 

> node-0_1  | 2020-04-27T01:55:49.167923309Z 2020-04-27 01:55:49,137 Info [agora.network.NetworkManager] - Established new connection with peer: GBCVDA6EJDTC76WP552UX4NM7ENRX6AK4JW27GKVHVOCLRWOGQES3UPP

This is the node's key, so the node is connecting with itself. It probably has something to do with IP filtering (banning).

I've also noticed the nodes are using several different address types to connect to:

```
ode-0_1  | Establishing connection with http://0.0.0.0:7826...
node-0_1  | Establishing connection with http://node-0:2826...
node-0_1  | Establishing connection with http://node-6:8826...
node-0_1  | Establishing connection with http://node-1:3826...
node-0_1  | Establishing connection with http://0.0.0.0:8826...
node-0_1  | Establishing connection with http://0.0.0.0:3826...
node-0_1  | Establishing connection with http://node-5:7826...
node-0_1  | Failed to connect to 172.21.0.8:6826: refused
node-0_1  | Failed to connect to 172.21.0.6:4826: refused
node-0_1  | Failed to connect to 172.21.0.7:5826: refused
```

So it tries connecting with `node-*:*`, `0.0.0.0:*`, and `172.21.0.*:*`, the latter of which are probably IPs assigned by docker.

Geod24:
The nodes now ban themselves, so this issue has been solved.

PR #1072, opened 2020-07-28
Implement periodic quorum shuffling

### LINKED ISSUE #1069: Quorums should be regularly re-shuffled
Currently the quorum sets are only changed when there is a change in the validator set - either through addition of new enrollments or by expiring enrollments.

Shuffling should be changed to be a frequent process, for example: shuffle quorums every N blocks where N is a consensus parameter.

Be aware that this will require some support code first. In our unittests we have tests such as:

- Set validation cycle to 20
- Generate genesis
- Create 19 blocks
- Start the nodes

The nodes won't have any preimages on startup except the commitment in the enrollment in the genesis block. If we set the "periodic shuffle" N parameter to 10, then on boot-up the nodes would generate the quorums for block height 0 and then on block height 10. But it's not possible to test this right now because the preimages are missing for height 10, so this would lead to an assertion failure.

AndrejMitrovic:
Notes:

- This does not implement saving of the preimages as mentioned in #1069. That's a separate issue. In the tests I simply poll and wait for the preimages to be revealed.
- When the shuffle cycle is set to a lower value (20), the EnrollmentManager tests frequently fail (because they generate around 21 blocks..). I'm pretty certain it's because the tests are written in a wrong way, one example shown was https://github.com/bpfkorea/agora/issues/1071 but there are more tests that need to be fixed. I've set the default shuffle cycle to 30 for now until this is resolved.
- This PR should be split up. It includes 4 bugfix/styling commits, an enhancement by making the quorum generator parameters configurable, and then the actual quorum shuffling implementation. All of the commits are required though.

Fixes: #1069


AndrejMitrovic:
Rebased.

MichaelKim20:
I have some suggestions and write them.
I think introducing a shuffle cycle is a good choice.
I think there should be a clearer standard for the value of the shuffle cycle.
I calculated it and I think up to 40 is possible.
I think it would be good to connect the shuffle cycle with the block height.
Cycle 0: Block Height 0-39
Cycle 1: Block Height 40-79
Cycle 2: Block Height 80-119

AndrejMitrovic:
> I think there should be a clearer standard for the value of the shuffle cycle.

It's a configurable option. We can change the default later.

> I think it would be good to connect the shuffle cycle with the block height.

That could cause unnecessary extra shuffles if a new validator has just enrolled.

Consider this case:

- Shuffle cycle set to 40
- At block height 38 a new Validator was added
- At block height 39 a new quorum is generated because the new Validator was added
- At block height 40 a shuffle occurs

It is unnecessary to shuffle the quorum at height 40, a new quorum was already generated at height 39. It should be shuffled +40 of the last quorum change height (which was 39, therefore 39 + 40 = 79, the next quorum shuffle).

source/agora/node/Validator.d
```diff
@@ -66,6 +67,11 @@ public class Validator : FullNode, API
     /// Quorum generator parameters
     protected QuorumParams quorum_params;
 
+    /// The last height at which a quorum shuffle took place.
+    /// If a new block is externalized and `last_shuffle_height`
+    /// is >= `QuorumShuffleCycle` then the quorum will be reshuffled.
+    private Height last_shuffle_height = Height(0);
```
Geod24: I don't think that comment is right, because what you are describing is a target, not a "last". Maybe missing an offset ?

source/agora/test/QuorumShuffle.d
```diff
@@ -0,0 +1,469 @@
...
+        network.blocks[$ - 1].txs, 1);
+    txs.each!(tx => nodes[0].putTransaction(tx));
+
+    // at block height 3 a shuffle should occur
+    nodes.enumerate.each!((idx, node) =>
+        retryFor(node.getBlockHeight() == 3, 5.seconds,
+            format("Node %s has block height %s. Expected: %s",
+                idx, node.getBlockHeight(), 3)));
+
+    enum quorums_2 = [
```
Geod24: That seems like a common occurence... Maybe we should make this a bit easier / less invasive ?

source/agora/test/QuorumShuffle.d
```diff
@@ -0,0 +1,469 @@
...
+import agora.consensus.data.genesis.Test;
+import agora.node.FullNode;
+import agora.test.Base;
+
+import std.algorithm;
+import std.format;
+import std.range;
+
+import core.thread;
+import core.time;
```
Geod24: Some of those imports should be useless

source/agora/test/QuorumShuffle.d
```diff
@@ -0,0 +1,469 @@
...
+        max_listeners : 7,
+        extra_blocks : 2,  // 1 short of shuffle cycle
+        validator_cycle : 10,
+        max_quorum_nodes : 4,  // makes it easier to test shuffle cycling
+        quorum_shuffle_cycle : 3
+    };
+    auto network = makeTestNetwork(conf);
+    network.start();
+    scope(exit) network.shutdown();
+    //scope(failure) network.printLogs();
```
Geod24: ?

source/agora/test/QuorumShuffle.d
```diff
@@ -0,0 +1,469 @@
...
+        QuorumConfig(4, [
+            WK.Keys.D.address,
+            WK.Keys.F.address,
+            WK.Keys.A.address,
+            WK.Keys.C.address]),
+    ];
+
+    version (none)
+    foreach (idx, node; nodes.enumerate)
+        writefln("Node %s: %s\n", idx, node.getQuorumConfig);
```
Geod24: Leftover ?

source/agora/test/QuorumShuffle.d
```diff
@@ -0,0 +1,469 @@
...
+    nodes.enumerate.each!((idx, node) =>
+        retryFor(node.getQuorumConfig() == quorums_2[idx], 5.seconds,
+            format("Node %s has quorum config %s. Expected: %s",
+                idx, node.getQuorumConfig(), quorums_2[idx])));
+
+    checkDistance(6);  // make sure we can create the next 3 blocks
+
+    void createBlock (ulong height)
+    {
+        txs = makeChainedTransactions(WK.Keys.Genesis, txs, 1);
```
Geod24: `TxBuilder` ?

source/agora/test/QuorumShuffle.d
```diff
@@ -0,0 +1,469 @@
...
+        QuorumConfig(4, [
+            WK.Keys.D.address,
+            WK.Keys.F.address,
+            WK.Keys.A.address,
+            WK.Keys.C.address]),
+    ];
+
+    version (none)
+    foreach (idx, node; nodes.enumerate)
+        writefln("Node %s: %s\n", idx, node.getQuorumConfig);
```
Geod24: Leftover

Geod24 (changes requested):

source/agora/test/QuorumShuffle.d
AndrejMitrovic: It's inside of `version (none)`, I left it in purposefully for when we change the quorum generator settings. Then you can uncomment `version (none)` and easily modify the test when the results are expected to change.

AndrejMitrovic: Should I name it something like `version (PrintQuorums)`? I guess that could prevent needing to recompile.

AndrejMitrovic: Prior art: https://github.com/Geod24/localrest/blob/37460b237c3bf14544dcd46d8d673bbe36d2d4db/source/geod24/LocalRest.d#L1437-L1438

source/agora/test/QuorumShuffle.d
AndrejMitrovic: Sure. I'll fix it.

source/agora/test/QuorumShuffle.d
AndrejMitrovic: Oh yes. Only 2 of them are actually used. Thanks.

source/agora/node/Validator.d
AndrejMitrovic: No I really am describing when the last shuffle occured.

source/agora/test/QuorumShuffle.d
Geod24: `version (none)` works for me, was just thinking you forgot them. I'm happy to keep them.

source/agora/node/Validator.d
AndrejMitrovic: And it is implied that one happened in the genesis block, because that's the first time the quorum "changes".

source/agora/test/QuorumShuffle.d
AndrejMitrovic: Hmm. I can try to calculate the preimage sum here and then call the quorum generator directly from the unittests and compare the quorums to see if they shuffled. It's definitely better than having to do these steps manually - thanks.

AndrejMitrovic:
Thanks for the feedback, there were some nice ideas in here.

source/agora/node/Validator.d
Geod24: Hum... I see: `last_shuffle_height >= QuorumShuffleCycle`. I thought `QuorumShuffleCycle` was the config option, but it's the target ?

AndrejMitrovic: I don't understand your terminologies here. Option vs Target.

We want there to be a quorum shuffle at least every X blocks, but they may happen more often if the set of enrollments has changed in the meantime. That's what I commented here https://github.com/bpfkorea/agora/pull/1072#issuecomment-669645266

Geod24: Mondays are hard.

`last_shuffle_height` is height at which the quorum was most recently shuffled;
`target_shuffle_height`, if it existed, would be the height at which the quorum is scheduled to be next shuffled;
`quorum_shuffle_cycle` is the difference between those two values;

AndrejMitrovic: target would be a misnomer though.

If we have `quorum_shuffle_cycle` set to 10 it might seem a good idea to set `target_shuffle_height` to 10 at the genesis block 0. But if there is a new enrollment at block height 5 then at height 5 we will already shuffle the quorums, so then we would have to shift the target to `15`. 

It seemed unnecessary to me to introduce another state variable when we can just check when was the last time the quorum was shuffled and if the diff is `quorum_shuffle_cycle` then shuffle it again.

AndrejMitrovic: I did just notice there is a bug in the logic. I was using `>` instead of `>=`.

I'll change the names of these variables.

AndrejMitrovic: Ah the bug was that `quorum_shuffle_cycle` was not passed to `ConsensusParams`..

This default parameter stuff needs to be removed: https://github.com/bpfkorea/agora/blob/b602254d522f5bc2e5b5e96525c4d61e7c9180ff/source/agora/consensus/data/ConsensusParams.d#L51

It's just going to become a source of bugs.

AndrejMitrovic:
I renamed the parameter to make it a little bit more obvious what it does. And I fixed the tests. The quorums are now all unique on every shuffle as expected.

source/agora/node/Validator.d
```diff
@@ -192,7 +192,7 @@ public class Validator : FullNode, API
     public override void start ()
     {
         this.startPeriodicDiscovery();
-        this.taskman.setTimer(10.seconds, &this.checkRevealPreimage, Periodic.Yes);
+        this.taskman.setTimer(1.seconds, &this.checkRevealPreimage, Periodic.Yes);
```
AndrejMitrovic: Let's turn this into a node config option.

AndrejMitrovic:
Ah I see what the last problem remains with the CircleCI failure:

```
node-4_1_1bd1427fc5a0 | 2020-08-13 05:02:03,512 Info [agora.network.NetworkManager] - Found new Validator: http://node-7:7826 (key: GDNODE7J5EUK7T6HLEO2FDUBWZEXVXHJO7C4AF5VZAKZENGQ4WR3IX2U)
node-4_1_1bd1427fc5a0 | 2020-08-13 05:02:03,890 Info [agora.network.NetworkManager] - Found new Validator: http://node-3:3826 (key: GDNODE3EWQKF33TPK35DAQ3KXAYSOT4E4ACDOVJMDZQDVKP66IMJEACM)
node-4_1_1bd1427fc5a0 | 2020-08-13 05:02:04,335 Info [agora.network.NetworkManager] - Found new Validator: http://node-4:4826 (key: GDNODE4KTE7VQUHVBLXIGD7VEFY57X4XV547P72D37SDG7UEO7MWOSNY)
node-4_1_1bd1427fc5a0 | 2020-08-13 05:02:05,454 Info [agora.network.NetworkManager] - Found new Validator: http://node-2:2826 (key: GDNODE2IMTDH7SZHXWDS24EZCMYCEJMRZWB3S4HLRIUP6UNGKVVFLVHQ)
node-4_1_1bd1427fc5a0 | 2020-08-13 05:02:05,547 Info [agora.network.NetworkManager] - Found new Validator: http://node-6:6826 (key: GDNODE6ZXW2NNOOQIGN24MBEZRO5226LSMHGQA3MUAMYQSTJVR7XT6GH)
node-4_1_1bd1427fc5a0 | 2020-08-13 05:02:05,731 Info [agora.network.NetworkManager] - Found new FullNode: http://node-0:1826
Linking...
```

Node 4 is connecting to itself, but not to node 5. This is a symptom of https://github.com/bpfkorea/agora/issues/772.

Now the question is why didn't it connect to node 5. Did it stall somehow? I'll add some logging.

AndrejMitrovic:
```
node-5_1  | 2020-08-13 05:41:22,288 Info [agora.network.NetworkManager] - Todo addresses { _set: [ "http://172.20.0.4:39057": true, "http://172.20.0.7:48129": true, "http://172.20.0.7:57391": true, "http://172.20.0.8:58561": true, "http://172.20.0.5:58077": true, "http://172.20.0.8:54461": true, "http://172.20.0.7:43823": true, "http://172.20.0.2:47551": true, "http://172.20.0.6:42397": true, "http://172.20.0.3:37267": true, "http://172.20.0.2:55307": true ] }
```

~It looks like there is memory corruption coming from somewhere.~

Edit: it's not memory corruption.

AndrejMitrovic:
What we're actually seeing is that the node runs out of connection tasks (there is a preconfigured limit of 10). But the reason it runs out is because it's connecting to a very large set of IP & Port ranges. For example:

```
node-0_1  | 2020-08-13 05:56:03,518 Info [...] - registerListener: http://172.20.0.8:54471
node-0_1  | 2020-08-13 05:56:04,349 Info [...] - registerListener: http://172.20.0.7:34819
node-0_1  | 2020-08-13 05:56:04,682 Info [...] - registerListener: http://172.20.0.5:53267
node-0_1  | 2020-08-13 05:56:04,721 Info [...] - registerListener: http://172.20.0.6:36671
node-0_1  | 2020-08-13 05:56:04,730 Info [...] - registerListener: http://172.20.0.2:47281
node-0_1  | 2020-08-13 05:56:04,742 Info [...] - registerListener: http://172.20.0.3:44341
```

```
node-3_1  | 2020-08-13 05:56:04,722 Info [...] - registerListener: http://172.20.0.5:46665
node-3_1  | 2020-08-13 05:56:04,725 Info [...] - registerListener: http://172.20.0.6:36691
node-3_1  | 2020-08-13 05:56:04,730 Info [...] - registerListener: http://172.20.0.2:37641
node-3_1  | 2020-08-13 05:56:04,742 Info [...] - registerListener: http://172.20.0.3:34367
node-3_1  | 2020-08-13 05:56:04,981 Info [...] - registerListener: http://172.20.0.4:51453
node-3_1  | 2020-08-13 05:56:05,150 Info [...] - registerListener: http://172.20.0.8:38775
```

Notice the ports are different for the two nodes. So when the nodes share their set of known IPs to each other the total amount of IP & port combinations explodes. For example:

```diff
diff --git a/source/agora/network/NetworkManager.d b/source/agora/network/NetworkManager.d
index 61d3d1d27..6517e8067 100644
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -401,6 +401,7 @@ public class NetworkManager
         this.discovery_task.add(node.client);
         this.metadata.peers.put(node.address);
         this.known_addresses.put(node.address);
+        log.info("Total known IPs: {}", this.known_addresses.length);
         this.connection_tasks.remove(node.address);

         this.registerAsListener(node.client);
@@ -474,6 +475,7 @@ public class NetworkManager

     public void registerListener (Address address)
     {
+        log.info("registerListener: {}", address);
         if (this.shouldEstablishConnection(address))
             this.addAddress(address);
     }
@@ -519,7 +521,10 @@ public class NetworkManager
         // reasoning: while *we* cannot establish a connection with
         // a node it's possible other nodes in the network might be able to.
         if (!this.isOurOwnAddress(address))
+        {
             this.known_addresses.put(address);
+            log.info("Total known IPs: {}", this.known_addresses.length);
+        }
     }

     /***************************************************************************
```

And the result:

```
node-5_1  | 2020-08-13 06:03:56,789 Info [agora.network.NetworkManager] - Total known IPs: 51
```

I think we need to fix this code: https://github.com/bpfkorea/agora/blob/a490dade705e5d9367e74c40793f2f430ae52b76/source/agora/node/Runner.d#L79

But I'm not sure how to proceed.

AndrejMitrovic:
One way to fix this is to not add the client's IP & Port to the known addresses.

Geod24:
(Rebased)

source/agora/test/Base.d
```diff
@@ -836,6 +837,9 @@ public interface TestAPI : ValidatorAPI
     ***************************************************************************/
 
     public void registerListenerAddress (Address address);
+
+    /// Get the list of expected quorum configs
+    public QuorumConfig[] getExpectedQuorums (in PublicKey[], Height);
```
Geod24, line 842: Should document parameters

source/agora/test/QuorumShuffle.d
```diff
@@ -0,0 +1,203 @@
...
+    Transaction[] txs = network.blocks[$ - 1].txs.serializeFull
+        .deserializeFull!(Transaction[]);
+
+    void makeBlock (ulong height)
+    {
+        txs = txs.filter!(tx => tx.type == TxType.Payment)
+            .map!(tx => iota(tx.outputs.length)
+                .map!(idx => TxBuilder(tx, cast(uint)idx)))
+            .joiner().takeExactly(8)
+            .map!(txb => txb.refund(WK.Keys.Genesis.address).sign()).array;
```
Geod24: - Use `spendable()`;
- Don't use `Genesis`

source/agora/test/QuorumShuffle.d
Geod24: 👀

Geod24 (approved):
Just a few comments, otherwise LGTM. Feel free to merge after they are addressed.

source/agora/test/QuorumShuffle.d
AndrejMitrovic: Oh I just realized this is a duplicate function of `makeBlock` above. Heh.

source/agora/test/QuorumShuffle.d
AndrejMitrovic: What do you mean by "don't use Genesis". The special Genesis address?

Geod24: > The special Genesis address?

Yes

AndrejMitrovic: `spendable` requires us to do a lot of back and forth btw - I can see this is going to become another code segment that is copy-pasted everywhere:


```D
const(Block)[] blocks = [network.blocks[$ - 1]];

void makeBlock (ulong height)
{
    auto txs = blocks[$ - 1].spendable.takeExactly(8)
        .map!(txb => txb.refund(WK.Keys.Genesis.address).sign()).array;
    txs.each!(tx => nodes[0].putTransaction(tx));
        retryFor(node.getBlockHeight() == height, 5.seconds,
                idx, node.getBlockHeight(), height)));
    retryFor((blocks = nodes[0].getBlocksFrom(height, 1)).length == 1, 5.seconds,
            format("Node 0 getBlocksFrom(%s, 1) failed", height));
}
```

Geod24: 1) Do we need `takeExactly` since we're transitioning away from tx-count-based block creation ?
2) Why do you set the refund address ? `sign()` is enough.
3) Since the transactions are  not  used as an array anywhere, no need to `array` then `each`, just `each` directly.

In short:
```D
blocks[$ - 1].spendable.each!(txb => nodes[0].putTransaction(txb.sign()));
```

AndrejMitrovic: Oh sorry I didn't realize `sign` does this already.

> Do we need takeExactly since we're transitioning away from tx-count-based block creation ?

Well we're still transitioning. :) 

Once we transition we can get rid of that, sure.

Geod24: > Oh sorry I didn't realize `sign` does this already.

Yes, the only requirement for `TxBuilder` is a refund address. If you provide it with a single output, it'll use the beneficiary as refund address, and if you provide it with a transaction, it'll use the first output's beneficiary by default.

PR #1077, opened 2020-07-29
Change to more detailed document and fix typos

MichaelKim20:


TrustHenry (approved):

source/agora/cli/client/Result.d
```diff
@@ -1,6 +1,6 @@
 /*******************************************************************************
 
-    The Agora client Result Constans
+    The Agora client Result Constant
```
Geod24:
```suggestion
    Constants returned by the agora CLI client
```

Geod24: Although honestly I'd rather get rid of this module, I don't think a module with 3 `int` constant makes sense.

source/agora/consensus/UTXOSet.d
```diff
@@ -114,7 +114,7 @@ public class UTXOSet
         get an UTXOSetValue in the UTXO set.
 
         Params:
-            hash = the hash of transation
+            hash = the hash of transaction
```
Geod24:
```suggestion
            hash = the hash of the transaction introducing the `Output`
```

source/agora/consensus/UTXOSet.d
```diff
@@ -171,7 +171,7 @@ public class UTXOSet
         Find an UTXOSetValue in the UTXO set.
 
         Params:
-            hash = the hash of transation
+            hash = the hash of transaction
```
Geod24:
```suggestion
            hash = the hash of the transaction introducing the `Output`
```

source/agora/consensus/data/Transaction.d
```diff
@@ -26,7 +26,7 @@ import agora.common.Serializer;
 
 import std.algorithm;
 
-/// The Agora Transaction Type constans
+/// The Agora Transaction Type constant
```
Geod24:
```suggestion
/// Type of the transaction: defines the content that follows and the semantic of it
```

Geod24 (commented):
Since we're fixing it, I rephrased it a bit more. Nice catch!

PR #1078, opened 2020-07-29
Fix setting the is_nominating flag

AndrejMitrovic:
This is a bug which I've noticed while inspecting leader selection.

When originally writing the Nominator class I failed to realize that `scp.nominate` may actually return early. For example if the current node is not the leader then SCP will start a timer 1 second in the future to [run scp.nominate() again](https://github.com/bpfkorea/agora/blob/1af3731d063a8b52f1e114a4b31e5ac65ad70869/source/scpp/src/scp/NominationProtocol.cpp#L528) and increase the round number to 2. This goes on an on as the number of rounds continue to increase (if nomination failed at every round).

I will have more PRs incoming which add additional tests - and I'll also write up a few more issues about what's still missing in the consensus networking tests.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -171,13 +171,12 @@ extern(D):
         if (this.is_nominating)
             return;
 
-        this.is_nominating = true;
-        scope (exit) this.is_nominating = false;
-
         ConsensusData data;
         this.ledger.prepareNominatingSet(data);
         if (data.tx_set.length == 0)
             return;  // not ready yet
+
+        this.is_nominating = true;
 
         // check whether the consensus data is valid before nominating it.
         if (auto msg = this.ledger.validateConsensusData(data))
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
```
+ dchatty=1
+ timeout -s SEGV 8m ./build/agora-unittests
+ rdmd --compiler=ldc2 ./tests/runner.d --compiler=ldc2 -cov
[fatal]	Failed to execute program (No such file or directory)
[fatal]	Target 'agora' doesn't exist. Is the target type set to "none" in the package recipe?
```

That's gotta be a new one?

AndrejMitrovic:
I'm converting this back into draft. I thought of a possible scenario that would lead to the node stopping making progress:

- Node starts nomination
- Experiences network failures so it keeps retrying to nominate
- Two blocks are externalized
- Network goes back online
- Node receives the second block before the first one (messages are not ordered), and rejects it because it's the wrong block height compared to its ledger.

Ideally here the node would just use the periodic catchup routine, however if `is_nominating` is still true then the periodic catchup will never be applied to the ledger.

I'll have to write a test-case for this.

AndrejMitrovic:
I'll fix this properly as part of time-based nomination.

PR #1086, opened 2020-07-31
Add a 16-node and 32-node quorum test

AndrejMitrovic:
Unfortunately the 32-node test had to be commented out because the nodes time out when creating the block.

The question is how much overhead we have, and whether a single system is able to run 32 nodes with ease.

Also the test /could/ be made generic with a little bit more code magic. For now I kept it like this. In the future we could rewrite it as:

```D
unittest
{
    testNodes(16);
    testNodes(32);
}
```

The only slight complexity is that blocks are limited to 8 blocks, so we have to do a bit of calculation when doing freeze TXs.


source/agora/test/Base.d
```diff
@@ -179,6 +180,9 @@ private UnitTestResult customModuleUnitTester ()
     foreach (mod; parallel(mod_tests))
         runTest(mod);
 
+    foreach (mod; parallel(heavy_tests))
+        runTest(mod);
```
Geod24, line 184: Should they be run in parallel ?

AndrejMitrovic: Well there's only one such test, but I'm thinking there might be more in the future.

Geod24 (approved):

PR #1091, opened 2020-08-03
Move the Genesis block to ConsensusParams

### LINKED ISSUE #955: Move the genesis block to the `ConsensusParams`
In the middle of making the validator cycle of an enrollment part of the Consensus parameter that is `ConsensusParams` class. we realized the genesis block can be a part of the class because it never changes on the network and is related to consensus.

This issue is separated from the issue #894 .

Geod24:
This will allow to simplify the test and avoid some consistency errors we had in the past.
Fixes #955


AndrejMitrovic (approved):

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -28,19 +30,34 @@ public immutable class ConsensusParams
...
         this.MaxQuorumNodes = max_quorum_nodes;
     }
+
+    /// Transitional method to avoid having to rewrite all unittests
+    /// TODO: Remove / replace with something more principled
+    version (unittest) public this (
+        uint validator_cycle = 1008, uint max_quorum_nodes = 7)
+    {
+        import agora.consensus.data.genesis.Test;
+        this(GenesisBlock, validator_cycle);
```
linked0: I guess it's kind of complicated to check in even this class to confirm which genesis block is used for the Genesis block. But, I'm not sure what's best.

linked0 (approved):
LGTM and I have one comment.

source/agora/consensus/data/ConsensusParams.d
Geod24: I don't follow your question ? The `GenesisBlock` symbol refers to the one in `agora.consensus.data.genesis.Test`.
I could turn it into a selective import to make it more obvious, but I was hoping the scoped symbol would do that.

Geod24: Fixed

linked0: I know what you mean, but I just thought it would be good to set the genesis block at higher level classes which are using the `ConsensusParams` class. Anyway, thanks for making the code more understandable.

Geod24: I originally did that, but the diff was huge, and very repetitive. Most of the time, you want to use the test genesis block in `unittest`.

Geod24:
Saw approval, rebasing (it might get us rid of some of the spurious failure).

source/agora/consensus/data/ConsensusParams.d
linked0: I think it was a good decision if there were so many changes and repetitive code.

Issue #1004, opened 2020-07-03
Transition away from creating a block based on the number of transactions in the pool


Geod24:
Currently: We attempt to nominate a set of transactions as soon as we have 8 available transactions.

What we want is to create blocks at a regular interval, regardless of the number of transaction.
We want this to be able to use the block height as a measure of time.
Our current thinking targets an interval of  10 minutes between blocks, but this should be configurable through `ConsensusParams` so we can write tests.

Questions to answer:
- How do we take into account the time it takes to nominate a block ?
  - How long before the "deadline" should nomination start ?
  - What timeout strategy do we use for nomination ?
  - How do we ensure the network is always live ?

Useful links:
Search everything related to `Block.TxsInBlock`.

AndrejMitrovic:
I will be working on this.

AndrejMitrovic:
I got a working draft version of this (test-suite passes).

But I don't have answers to the questions yet. Need to research that.

With regards to being able to measure progress of time, I wonder if it would make sense to include a timestamp in the block header.

AndrejMitrovic:
What do we do when there's not enough transactions to form a block? Do we create an empty block?

AndrejMitrovic:
s/not enough/not even 1 transaction.

AndrejMitrovic:
This was moved to done but was not closed.

PR #1096, opened 2020-08-04
Make nominating more controllable

AndrejMitrovic:
This PR contains code split off from the rest of the work that's in progress for https://github.com/bpfkorea/agora/issues/1004

These changes look nice on their own and they can be merged before the rest of the work is submitted.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -340,21 +340,19 @@ public class Ledger
 
     /***************************************************************************
 
-        Try to collect a set of transactions to nominate.
+        Collect up to a maximum number of transactions to nominate
 
         Params:
             txs = will contain the transaction set to nominate,
                   or empty if not enough txs were found
+            max_txs = the maximum number of transactions to prepare.
 
     ***************************************************************************/
 
-    public void prepareNominatingSet (ref ConsensusData data) @safe
+    public void prepareNominatingSet (ref ConsensusData data, ulong max_txs)
+        @safe
     {
-        assert(data.tx_set.length == 0);
-
-        if (this.pool.length < Block.TxsInBlock)
-            return;
-
+        data = ConsensusData.init;
         const next_height = Height(this.getBlockHeight() + 1);
         auto utxo_finder = this.utxo_set.getUTXOFinder();
 
@@ -367,17 +365,12 @@ public class Ledger
             else
                 data.tx_set ~= tx;
 
-            if (data.tx_set.length >= Block.TxsInBlock)
+            if (data.tx_set.length >= max_txs)
             {
                 data.tx_set.sort();
                 return;
             }
         }
-
-        // not enough txs were found
-        () @trusted {
-            data = ConsensusData.init;
-        }();
     }
 
     /***************************************************************************
@@ -533,8 +526,8 @@ version (unittest)
     private void forceCreateBlock (Ledger ledger)
     {
         ConsensusData data;
-        ledger.prepareNominatingSet(data);
-        assert(data.tx_set.length > 0);
+        ledger.prepareNominatingSet(data, Block.TxsInBlock);
+        assert(data.tx_set.length == Block.TxsInBlock);
         assert(ledger.onExternalized(data));
     }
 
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
Fixed commit message.

source/agora/node/Ledger.d
```diff
@@ -340,44 +340,38 @@ public class Ledger
...
-        if (this.pool.length < Block.TxsInBlock)
-            return;
-
+        data = ConsensusData.init;
         const next_height = Height(this.getBlockHeight() + 1);
         auto utxo_finder = this.utxo_set.getUTXOFinder();
 
         this.enroll_man.getEnrollments(data.enrolls,
             Height(this.getBlockHeight()));
+
```
AndrejMitrovic: Whitespace change.

source/agora/node/Ledger.d
```diff
@@ -527,14 +521,17 @@ public class Ledger
     }
 }
 
+/// Note: these unittests historically assume a block always contains
```
AndrejMitrovic: This comment belongs in a future commit, not here.

source/agora/test/Base.d
```diff
@@ -351,6 +352,18 @@ public class FakeClockBanManager : BanManager
     public override void dump () { }
 }
 
+/// Nominator with custom rules for when blocks should be nominated
+public extern (C++) class TestNominator : Nominator
+{
+extern(D):
+
```
AndrejMitrovic: ///

source/agora/test/Base.d
```diff
@@ -853,6 +866,13 @@ public class TestValidatorNode : Validator, TestAPI
     {
         return this.qc;
     }
+
+    /// Returns an instance of a TestNominator with custom nominating behavior
```
AndrejMitrovic: custom -> customizable

source/agora/test/Base.d
```diff
@@ -356,12 +357,37 @@ public class FakeClockBanManager : BanManager
 public extern (C++) class TestNominator : Nominator
 {
 extern(D):
+    /// number of txs required for nomination
+    private ulong txs_to_nominate;
```
AndrejMitrovic: This doesn't belong in this PR.

AndrejMitrovic:
Ok resolved comments and removed code which didn't belong here.

source/agora/test/Base.d
```diff
@@ -853,6 +866,13 @@ public class TestValidatorNode : Validator, TestAPI
     {
         return this.qc;
     }
+
+    /// Returns an instance of a TestNominator with customizable behavior
+    protected override TestNominator getNominator ( NetworkManager network,
```
linked0, line 871: This function is overridden so the signature should be the same as that of the superclass. But I know what you intended.

linked0 (commented):
Very good refactoring, but I have some comments.

source/agora/test/EnrollmentManager.d
```diff
@@ -376,7 +356,7 @@ unittest
         }
 
         ///
-        protected override Nominator getNominator (
+        protected override TestNominator getNominator (
```
linked0, line 359: Same as the above comment.

source/agora/test/Base.d
AndrejMitrovic: https://en.wikipedia.org/wiki/Covariant_return_type

AndrejMitrovic: There is also a very good reason to do this. There is an override of this method here: https://github.com/bpfkorea/agora/blob/97bfb80edad10731d0a5f41647428913a51dbafb/source/agora/test/EnrollmentManager.d#L383

If I didn't use `TestNominator` in `TestValidatorNode` then the code in `EnrollmentManager` could be written as:

```D
static class BadNominator : Nominator  // NOT TestNominator
{
    ...
}

protected override Nominator getNominator ( ... ) { return new BadNominator(...); }
```

This means it would circumvent the TestNominator class. But we want all derived classes to be forced to instantiate either `TestNominator` or a derived class from `TestNominator`. Changing the return type to be narrower (TestNominator in this PR line) will force the compiler to check this. It will be a compile-time error if the derived class returns an instance of a Nominator and not a TestNominator.

AndrejMitrovic:
@linked0 can you approve / merge this? It will unblock the rest of my work.

linked0 (approved):

AndrejMitrovic:
Thank you!

PR #1097, opened 2020-08-05
Introduce expectBlock() for testing heights

AndrejMitrovic:
I factored this out of the rest of the work in #1004 - it will make it much easier to test time-based block creation without having to touch most of the test code.


AndrejMitrovic:
Note that we still need the rest of the existing functions such as `expectConsistency()` - in particular when we don't want to shift the time. It's not obvious in this PR but `expectBlock()` will forward the clock times of each node in a subsequent PR.

Geod24 (approved):
Love it

Issue #407, opened 2019-10-28
Design and implement transaction selection filter


Geod24:
As a developer, I need to have a transaction filter algorithm, based on #406, that derives a filter used to include transactions in a block in order to prevent transaction flood attacks by validators.

Description
The filter should be loose enough that it allows transactions to be finalized timely in a healthy network, but tight enough that it prevents long lasting flood attack of transactions. The randomness of hash functions ensures the filter will be well distributed.

Definition of done
- [ ] Incoming transactions are included in a block based on a filter derived from the source of randomness;
- [ ] Blocks that have filtered-out transactions have their validators slashed;
- [ ] The filter *could* adapt itself based on transaction load of previous blocks;
- [ ] Transactions fees are taken into account (on the node side)
- [ ] The filter does not necessarily work on the transaction itself, but rather on the set of transactions, making it increasingly more expensive to add transactions;

Example:
A challenge on the merkle root + children, or on the merkle root's children + grandchildren.

Geod24:
We won't have this for CoinNet, as it's a major protocol change. Moved to the `MainNet` release.

PR #1098, opened 2020-08-05
Remove restriction for 8-transaction blocks

AndrejMitrovic:
Now there is a new TestConf option called `txs_to_nominate` which is passed to the `TestNominator`.

This does not change the nominating behavior of the production node yet.

Note also that this is **not** a consensus parameter because we don't want a transaction count consensus rule - we rather want a block size rule. However this is outside the scope of #1004.

This is the first part of the support code for #1004. In the next PRs I'll add a timer-based nominator.

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -128,14 +128,17 @@ unittest
 
     The block which contains the block header and its body (the transactions).
 
-    In the current preliminary design a block contains a single transaction.
-
 *******************************************************************************/
 
 public struct Block
 {
-    /// number of transactions that constitutes a block
-    public enum TxsInBlock = 8;
+    // some unittests still assume a block contains 8 txs. Once they're fixed
+    // this constant should be removed.
+    version (unittest)
+    {
+        /// number of transactions that constitutes a block
+        public enum TxsInTestBlock = 8;
+    }
 
     ///
     public BlockHeader header;
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -170,8 +170,8 @@ extern(D):
 
     protected bool prepareNominatingSet (out ConsensusData data) @safe
     {
-        this.ledger.prepareNominatingSet(data, Block.TxsInBlock);
-        if (data.tx_set.length != Block.TxsInBlock)
+        this.ledger.prepareNominatingSet(data, 8);
+        if (data.tx_set.length != 8)
             return false;  // not ready to nominate yet
 
         // check whether the consensus data is valid before nominating it.
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -35,6 +35,7 @@ import agora.common.TransactionPool;
 import agora.common.crypto.Key;
 import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
+import agora.consensus.data.ConsensusData;
 import agora.consensus.data.ConsensusParams;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.Transaction;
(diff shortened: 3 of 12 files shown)


AndrejMitrovic:
CircleCI don't like me no more.

source/agora/consensus/data/Block.d
```diff
@@ -128,15 +128,10 @@ unittest
...
     The block which contains the block header and its body (the transactions).
 
-    In the current preliminary design a block contains a single transaction.
-
 *******************************************************************************/
 
 public struct Block
 {
-    /// number of transactions that constitutes a block
-    public enum TxsInBlock = 8;
```
Geod24, line 138: Perhaps just make it a `version (unittest)` ? At the moment there's a bunch of changes in the test suite which are of little relevance.

AndrejMitrovic: Well I thought future tests should stop making these assumptions so I removed the constant.

source/agora/test/Base.d
```diff
@@ -356,12 +357,38 @@ public class FakeClockBanManager : BanManager
...
+            this.txs_to_nominate ? this.txs_to_nominate : ulong.max);
+        if (data.tx_set.length < this.txs_to_nominate)
+            return false;  // not enough txs
+
+        // defensive coding, same as base class
+        // (but may be overruled by derived classes)
+        if (auto msg = this.ledger.validateConsensusData(data))
+        {
+            // todo: need to log this
+            // log.fatal("tryNominate(): Invalid consensus data: {}. Data: {}", msg, data);
```
Geod24: Why is it a TODO ?

source/agora/test/Base.d
```diff
@@ -415,17 +446,18 @@ public class TestAPIManager
 
         Params:
             conf = the configuration passed on to the Node constructor
+            test_conf = test configuration
 
     ***************************************************************************/
 
-    public void createNewNode (Config conf)
+    public void createNewNode (Config conf, )
```
Geod24: :|

source/agora/test/Base.d
```diff
@@ -415,17 +446,18 @@ public class TestAPIManager
 
         Params:
             conf = the configuration passed on to the Node constructor
+            test_conf = test configuration
```
Geod24: :|

source/agora/consensus/data/Block.d
Geod24: You're right, but changing an easily-grepped-for constant for `8` isn't removing the assumption

source/agora/test/Base.d
AndrejMitrovic: The `log` is private and in the module scope of `Nominator`. So I wasn't sure what to do about this.

source/agora/consensus/data/Block.d
AndrejMitrovic: Fair enough.

source/agora/test/Base.d
AndrejMitrovic: Woops.

source/agora/test/Base.d
AndrejMitrovic: Happens.

source/agora/test/Base.d
Geod24: Makes sense.

AndrejMitrovic:
I've renamed the constant and versioned it.

AndrejMitrovic:
> ##[error]The process '/usr/local/bin/git' failed with exit code 1

Nothing is reliable these days.

AndrejMitrovic:
Green.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -170,8 +170,8 @@ extern(D):
 
     protected bool prepareNominatingSet (out ConsensusData data) @safe
     {
-        this.ledger.prepareNominatingSet(data, Block.TxsInBlock);
-        if (data.tx_set.length != Block.TxsInBlock)
+        this.ledger.prepareNominatingSet(data, 8);
```
linked0, line 173: There are two `8`s in the `Nominator` class, which feels like the `8` is something special. Should it be constant like `defaultTxsInBlock`? If your big picture is to remove the two 8s in the end, it's OK.

AndrejMitrovic: This will be replaced with time-based transaction selection. And the maximum limit of the block size is something we need to define later. For example https://github.com/bpfkorea/agora/issues/407

Geod24 (approved):

PR #1100, opened 2020-08-05
Implement a time-based block creation algorithm

AndrejMitrovic:
This implements one part of #1004.

There is no synchronization of clocks between nodes - but this is something I'm working on and might file as a separate issue.

Note that I'm using `shared(time_t)` rather than a `TestNode` API endpoint to get & set a custom clock time because I want to avoid request overheads when setting the time (especially because these requests may be delayed).


AndrejMitrovic:
Of all the targets to fail, it had to be Windows. 🤦‍♂️😄

AndrejMitrovic:
> source\agora\test\Base.d(553,16): Error: cannot implicitly convert expression `cast(ulong)this.genesis_start_time + height.value * cast(ulong)this.test_conf.block_interval_sec` of type `ulong` to `int`

Huh, why is `time_t` 32-bit on Windows? That is not at all what the docs say it should be: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/time-time32-time64?view=vs-2019

> time is a wrapper for _time64 and time_t is, by default, equivalent to __time64_t. If you need to force the compiler to interpret time_t as the old 32-bit time_t, you can define _USE_32BIT_TIME_T

But apparently in druntime it's different:

```
C:\dev> rdmd --eval="import core.stdc.time; writeln(time_t.sizeof);"
4
```

AndrejMitrovic:
Edit: also with `-m64`. I think DMD defaults to -m32 on Windows if I'm not mistaken. I'll ask in the D forums.

AndrejMitrovic:
Update: Added a clock-drift test-case and it works exactly as expected.

source/agora/common/Config.d
```diff
@@ -138,6 +138,9 @@ public struct NodeConfig
 
     /// Maximum number of nodes to include in an autogenerated quorum set
     public uint max_quorum_nodes = 7;
+
+    /// Threshold to use in the autogenerated quorum
+    public double quorum_threshold = 0.80;
```
Geod24: That feels like going back to QuorumConfig, just with a global value.
Do you envision this as a temporary solution, or do you think it's going to be permanent ?

source/agora/network/Clock.d
```diff
@@ -0,0 +1,39 @@
+/*******************************************************************************
+
+    Contains a Clock implementation.
```
Geod24, line 3: I love this commit <3

source/agora/common/Config.d
```diff
@@ -102,6 +102,13 @@ public struct NodeConfig
     /// in place of the built-in genesis block as defined by CoinNet
     public string genesis_block;
 
+    /// The Unix timestamp of the Genesis block. Blocks are created at a
+    /// regular interval starting from this time.
+    public time_t genesis_start_time = 1596179709;
```
Geod24: Ooooohhhh didn't think of this. Right. I would say, if we parse it from the config file, let's make it parse an ISO8601 string. Also, mention the time is UTC.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -181,7 +229,8 @@ extern(D):
 
     ***************************************************************************/
 
-    protected bool prepareNominatingSet (out ConsensusData data) @safe
+    protected bool prepareNominatingSet (out ConsensusData data,
+        in time_t cur_time) @safe
```
Geod24: I see what you did here

source/agora/consensus/protocol/Nominator.d
```diff
@@ -200,31 +249,61 @@ extern(D):
...
-        scope (exit) this.is_nominating = false;
+        const cur_time = this.clock.time();
+        if (cur_time < this.params.GenesisStartTime)
+        {
+            log.fatal("Clock is out of sync. " ~
+                "Current time: {}. Genesis time: {}", cur_time,
+                this.params.GenesisStartTime);
+            return;
+        }
+
```
Geod24, line 313: Maybe we should just check it at startup and `assert` here. I can see a lot of headache otherwise.

Geod24: Check it at startup: Needs to be user-friendly for the user in regular build, and dev friendly in UT

source/agora/node/FullNode.d
```diff
@@ -277,11 +277,8 @@ public class FullNode : API
             return;
 
         if (this.ledger.acceptTransaction(tx))
-        {
-            // gossip first
             this.network.gossipTransaction(tx);
-            this.onAcceptedTransaction();
-        }
+
```
Geod24: That's a sin

source/agora/node/Validator.d
```diff
@@ -67,6 +67,10 @@ public class Validator : FullNode, API
     /// Quorum generator parameters
     protected QuorumParams quorum_params;
 
+    /// Workaround: onRegenerateQuorums() is called from within the ctor,
+    /// but LocalScheduler is not instantiated yet.
+    private bool started;
```
Geod24, line 78: How can the timer be caller if the scheduler isn't running ? Can I have an ELIMM (Explain Like It's Monday Morning) ?

source/agora/test/Base.d
```diff
@@ -405,6 +408,23 @@ public struct NodePair
...
+    @property time_t time () @trusted @nogc nothrow
+    {
+        return atomicLoad(*this.cur_time);
+    }
+
+    /// Set the new time
+    @property void time (time_t new_time) @trusted @nogc nothrow
+    {
+        atomicStore(*this.cur_time, new_time);
+    }
```
Geod24, line 471: Why not in the `TestAPI` ?

Geod24 (commented):
This needs a rebase.

source/agora/test/Base.d
AndrejMitrovic: It's explained in the PR description.

source/agora/node/Validator.d
AndrejMitrovic: We cannot use `setTimer` until `LocalScheduler` is instantiated, and it is instantiated **after** the constructor finishes. In the constructor of the node `onRegenerateQuorums` is called. But `onRegenerateQuorums` is also called when a block is externalized.

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: I'll add another check at start-up, however I will leave this here as well. There is another PR coming which implements synchronization of the clock across the network, together with a lot of documentation about clock time drift and so on..

source/agora/common/Config.d
AndrejMitrovic: Yep. Also I'm not sure whether we need or don't need a timestamp in the block header yet. It might be possible to omit it for now since the estimated block timestamp is derived from its height.

source/agora/network/Clock.d
AndrejMitrovic: It gets bigger. https://github.com/AndrejMitrovic/agora/blob/b057fe56b79f54df5b9fc39276c648c18b695cab/source/agora/network/Clock.d#L1-L92

source/agora/test/Base.d
Geod24: Ah, missed that. The overhead is negligible TBH (we have a resolution of 1 second and the overhead is usually 10ms, 400ms is the highest I've seen it). But I'm fine if we leave it this way for the moment.

source/agora/common/Config.d
AndrejMitrovic: It's only for the unittests. In this case it makes it significantly easier to write a simple test-case - otherwise I would have to instantiate many more nodes and I prefer to have tests which are simple to understand.

I don't actually want this to be in the node configs. But there was a commit which removed the `ConsensusParams` from the `FullNode` ctor parameter list and `ConsensusParams` was instead instantiated directly in the constructor body using the node config parameters.

AndrejMitrovic:
> This needs a rebase.

Ok.

Geod24:
This has looooooads of conflicts.  Feel free to submit things piece-meal if that's easier.

AndrejMitrovic:
This can be reviewed.

Note that synchronization of clocks across nodes will be done in a separate PR.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -139,14 +162,67 @@ extern(D):
...
+        Stop the current nominating round
+
+        This does not stop the periodic nomination check timer,
+        but it stops the node from continuing nominating any transaction sets
+        which were proposed based on a previous ledger state.
+
+        This routine should be called after every externalize event and
+        after any valid block is received from the network (getBlocksFrom()).
+
```
AndrejMitrovic, line 212: Missing `Params:`

source/agora/consensus/protocol/Nominator.d
```diff
@@ -168,7 +244,8 @@ extern(D):
 
     ***************************************************************************/
 
-    protected bool prepareNominatingSet (out ConsensusData data) @safe
+    protected bool prepareNominatingSet (out ConsensusData data,
```
AndrejMitrovic: Needs `Params` docs. `cur_time` is used by derived classes, and will also be used here but by subsequent PRs.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -187,31 +264,61 @@ extern(D):
 
     /***************************************************************************
 
-        Try to begin a nomination round.
+        Gets the expected block nomination time for for the provided height.
+
```
AndrejMitrovic, line 271: Missing `Params`

source/agora/consensus/protocol/Nominator.d
```diff
@@ -187,31 +264,61 @@ extern(D):
...
+        Returns:
+            the expected block nomination time for the provided height
+
+    ***************************************************************************/
+
+    private ulong getExpectedBlockTime (Height height) @safe @nogc nothrow pure
```
AndrejMitrovic, line 280: We should change all `time_t` to `ulong` since `time_t` doesn't seem to be portable on Windows.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -187,31 +264,61 @@ extern(D):
...
+
+        The function will return early if either one of these are true:
+        - We're already in the asynchronous stage of nominating or balloting
+        - The current time is < getExpectedBlockTime(slot_idx)
+        - There are no transactions in the pool to nominate yet
 
     ***************************************************************************/
 
-    public void tryNominate () @safe
+    public void checkNominate () @safe
```
AndrejMitrovic: > This function is called periodically by the nominating timer.

Should this be `private` then?

source/agora/test/Base.d
AndrejMitrovic: I forgot to reply here. There is also the issue of filtering - I want to circumvent active filters on the node but still have the ability to change its time.

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -50,19 +59,23 @@ public immutable class ConsensusParams
...
-            quorum_shuffle_interval = quorum shuffle block cycle length
+            genesis_start_time = unix timestamp of the Genesis block
 
     ***************************************************************************/
 
     public this (immutable(Block) genesis, uint validator_cycle = 1008,
                  uint max_quorum_nodes = 7, uint quorum_threshold = 80,
-                 uint quorum_shuffle_interval = 30)
+                 uint quorum_shuffle_interval = 30,
+                 time_t genesis_start_time = 1596179709,
```
TrustHenry, line 69:
```suggestion
                 time_t genesis_start_time = 1596153600,
```

`assert(SysTime(DateTime.fromISOString("20200731T000000"), UTC()).toUnixTime == 1596153600);`

According to the time declared in config value, the unix timestamp value is as above.

AndrejMitrovic: Ah yes. In any case these are just placeholder values. We need to change them to the actual time of when mainnet will be deployed. But we don't know this yet.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -82,16 +86,19 @@ extern(D):
...
             network = the network manager for gossiping SCP messages
             key_pair = the key pair of this node
             ledger = needed for SCP state restoration & block validation
             taskman = used to run timers
 
     ***************************************************************************/
 
-    public this (NetworkManager network, KeyPair key_pair, Ledger ledger,
+    public this (Clock clock,
+        NetworkManager network, KeyPair key_pair, Ledger ledger,
```
Geod24: You have weird life choices

source/agora/network/Clock.d
```diff
@@ -0,0 +1,39 @@
...
+
+module agora.network.Clock;
+
+import agora.utils.Log;
+
+import core.stdc.time;
+
+mixin AddLogger!();
+
+/// The current implementation uses NTP
```
Geod24, line 25: ?

Geod24 (approved):

source/agora/network/Clock.d
AndrejMitrovic: Oh well outdated comment. But under the hood the OS uses NTP. Interestingly enough Ethereum (or was it Bitcoin?) tried to roll out their own NTP code and realized it was worse than what the OS already implements, so they rolled back.

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: Oh you mean the order of parameters. I kind of tried to order them by importance, but this is purely subjective. I'll add params at the end from now on.

PR #1101, opened 2020-08-06
Fix occasional circle-ci failures

AndrejMitrovic:


AndrejMitrovic:
Ah the numbers in the commit message are wrong. Lemme fix.

AndrejMitrovic:
Fixed.

AndrejMitrovic:
Ah node 7 was also missing from the FullNode config. Fixed.

tests/system/node/0/config.yaml
```diff
@@ -36,6 +36,7 @@ network:
   - http://node-4:4826
   - http://node-5:5826
   - http://node-6:6826
+  - http://node-7:7826
```
Geod24: Node 0 is not a validator, so it doesn't need to know the full set of nodes. That's why it had a lower number of listeners and was missing node 7. Maybe I should have added a comment.

tests/system/source/main.d
```diff
@@ -62,7 +62,7 @@ int main (string[] args)
         foreach (const ref addr; addresses)
             clients ~= new RestInterfaceClient!API(addr);
 
-        waitForDiscovery(clients, 5.seconds);
+        waitForDiscovery(clients, 10.seconds);
```
Geod24: 🤦‍♂️

tests/system/node/0/config.yaml
AndrejMitrovic: But the Validator's configs have the minimum set to 6. So they need to connect to the FullNode too.

tests/system/source/main.d
AndrejMitrovic: Yeah this didn't fix problems either. sigh

tests/system/node/0/config.yaml
AndrejMitrovic: I think maybe the right fix is to lower the minimum set in the Validator configs.

And actually it's a little dangerous to have this value be applied to the validators because if it's too high the validator will never start validating.

AndrejMitrovic:
But I think there is an underlying issue somewhere. There really shouldn't be any request failures.

AndrejMitrovic:
Two successful CI runs. I'll run CircleCI a few more times to see if it's consistent.

AndrejMitrovic:
Third run failed. We will need a much better solution than tinkering with the config options.

AndrejMitrovic:
Failed experiment.

PR #1102, opened 2020-08-06
Add quorum threshold to consensus params & improve ensureConsistency()

AndrejMitrovic:
Split off from #1100


Geod24 (approved):
LGTM, but can we avoid using floating point until necessary ? And it's only necessary if we don't have a fixed precision.

AndrejMitrovic:
I could make it a fixed integer range from 1 to 100 I suppose.

PR #1106, opened 2020-08-07
ConnectionTask: Fix request failure retrying

### LINKED ISSUE #1105: ConnectionTask is not handling request timeouts gracefully
https://github.com/bpfkorea/agora/blob/ef0b8e70fb4a2afef110dc960f998b6d1ec1a9ec/source/agora/network/NetworkManager.d#L162

This code can throw an `Exception` (in NetworkClient) which will not be caught by the try/catch.

I have a fix for it, but I need to write a test-case.

AndrejMitrovic:
This might resolve the frequent CircleCI failures we've seen recently.

Fixes #1105


source/agora/network/NetworkManager.d
```diff
@@ -162,11 +162,14 @@ public class NetworkManager
...
                 {
-                    // 404 => API not implemented, it's a FullNode
-                    if (ex.status == 404)
-                        break;
+                    if (auto http = cast(HTTPStatusException)ex)
+                    {
+                        // 404 => API not implemented, it's a FullNode
+                        if (http.status == 404)
+                            break;
+                    }
```
Geod24, line 172: How does it work with LocalRest ?

AndrejMitrovic: It's the NetworkClient which throws the exception if there was a timeout.

Without this change and with the test-case I've added ordinarily this would happen:

```
#### FATAL ERROR: Request failure to Validator #2 (GDC2...YD36) after 2 attempts
This node was started at source/agora/test/Base.d:459
```

AndrejMitrovic:
The CI still failed so I'm trying another thing in a second commit.

AndrejMitrovic:
> The CI failed

But the "uncaught exception" errors are now gone so this is good.

Geod24 (approved):

Issue #1104, opened 2020-08-07
Remove the confusing log about the `receivePreimage` endpoint


linked0:
When a pre-image is received, which has the same distance as the existing one in the `ValidatorSet`, the log as follows comes out. This log appears so many times and is kind of confusing by showing the information something wrong happened.

```
node-3_1_90d86e12c2e8 | 2020-08-07 07:39:22,571 Info [agora.consensus.ValidatorSet] - Invalid pre-image data: The height of new pre-image is not greater than that of the previous one. Pre-image: { /*omitted*/ }
```

But the situation is very normal because the pre-image had been just received before a new request is received. So, we should check the pre-image received is already in the `ValidatorSet` and we do not call the `addPreimage` function if it is. And then we can remove the log above.

AndrejMitrovic:
Looks like it's resolved now. Thanks @kchulj.

PR #1108, opened 2020-08-10
Ignore same height pre-image because validators will gossip them

kchulj:
```console
node-7_1_7e2b1f0640fc | 2020-08-06T07:54:49.463088747Z 2020-08-06 07:54:49,461 Info [agora.consensus.ValidatorSet] - Invalid pre-image data: The height of new pre-image is not greater than that of the previous one. Pre-image: { enroll_key: 0xdb3931bd87d2cea097533d82be0a5e36c54fec8e5570790c3369bd8300c65a03d76d12a74aa38ec3e6866fd64ae56091ed3cbc3ca278ae0c8265ab699ffe2d85, hash: 0x033d75ea3cf5f2d3e6b4be76ca6cc66bac6c101b79614dabe930111b34396fa2a5445ca43d0225d96c49874264caaea87cbaee692f60a8aac1b586b231202fb3, distance: 11 }
```
This message was showing up in the system integration test. The node gossips the pre-image even though the receiving nodes already have the information. Gossiping is not a malicious behavior, so we can ignore it because it prints the same message several times.

~~This addition is a workaround for~~ 
Fixes issue [#1104](https://github.com/bpfkorea/agora/issues/1104)

### DIFF
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -475,6 +475,10 @@ public class ValidatorSet
             return false;
         }
 
+        // Ignore same height pre-image because validators will gossip them
+        if (prev_preimage.distance == preimage.distance)
+            return false;
+
         if (auto reason = isInvalidReason(preimage, prev_preimage,
             this.params.ValidatorCycle))
         {


Geod24 (approved):

source/agora/consensus/ValidatorSet.d
```diff
@@ -475,6 +475,10 @@ public class ValidatorSet
             return false;
         }
 
+        // Ignore same height pre-image because validators will gossip them
+        if (prev_preimage.distance == preimage.distance)
```
AndrejMitrovic, line 479: Isn't it better to compare `if (prev_preimage == preimage)`? Because if the distance is the same but the hash is different we won't get any errors - but we should get the errors in the logs..

Geod24: Should we ? Does it really matter ?

TrustHenry (approved):
LGTM

linked0 (approved):
LGTM

source/agora/consensus/ValidatorSet.d
kchulj: This addition is supposed to act as an additional filter for [`cb941ec`](https://github.com/bpfkorea/agora/commit/cb941ecab41a2c0256427289e054f286d1596c3e). The error for the wrong hash should be triggered by another validity check in `isInvalidReason()`. I will examine the code again.

AndrejMitrovic (approved):

linked0:
I think this is not just a workaround for the issue #1104 , but resolve the issue. So you can replace the comment of "This addition is a workaround for issue #1104" with "Fixes #1104".

PR #1109, opened 2020-08-10
Fix HTTPStatusException propagation

AndrejMitrovic:


AndrejMitrovic:
Well I really thought this would fix it, but apparently there are more hidden bugs. Sigh.

AndrejMitrovic:
Ah maybe it also requires Jay's PR.

AndrejMitrovic:
> Ah maybe it also requires Jay's PR.

The reason I say this is because there are still no logs like this: https://github.com/bpfkorea/agora/blob/b602254d522f5bc2e5b5e96525c4d61e7c9180ff/source/agora/network/NetworkManager.d#L186

AndrejMitrovic:
> Ah maybe it also requires Jay's PR.

Nope it didn't work with the combination of that.

So my PR fixes one bug but it doesn't solve the underlying problem.

AndrejMitrovic:
With the second commit the log files finally show connecting to the FullNode:

> node-3_1_5841cdea84e9 | 2020-08-10 08:51:32,299 Info [agora.network.NetworkManager] - Found new FullNode: http://node-0:1826

But this is just treating the symptom and not fixing the underlying problem. However maybe we can at least be unblocked..

Issue #855, opened 2020-05-19
Ensure that an already-enrolled Validator starts to validate immediately 


Geod24:
As a validator, it is possible that I would boot up when already enrolled (e.g. if the node is a back-up node for another node).

We should have a test that:
- Starts a network with a few nodes;
- Has one node enrolled that is stopped *and wiped clean*;
- Wait at least 2 blocks;
- Restart the node, verify that it catches up, and starts to validate again;

AndrejMitrovic:
> Wait at least 2 blocks;

If a Validator is offline while 2 blocks are externalized, doesn't that imply that the Validator will be slashed? Therefore this Validator shouldn't be able to validate again when it catches up.

linked0:
> > Wait at least 2 blocks;
> 
> If a Validator is offline while 2 blocks are externalized, doesn't that imply that the Validator will be slashed? Therefore this Validator shouldn't be able to validate again when it catches up.

Right, I see what you mean. I'll modify the code of the PR #1110 accordingly.

PR #1110, opened 2020-08-12
Test case for validating immediately after a current validator restarts

linked0:
This is the test case for making sure that an already enrolled `Validator` starts validating immediately after it restarts with all the data wiped clean and two blocks behind.

Relates #855

### DIFF
--- a/source/agora/test/ValidatorCleanRestart.d
+++ b/source/agora/test/ValidatorCleanRestart.d
@@ -141,3 +141,64 @@ unittest
     new_txs.each!(tx => set_b[0].putTransaction(tx));
     network.expectBlock(set_b, Height(11), 5.seconds);
 }
+
+/// Situation: A validator is stopped and wiped clean after the block height
+///     is 1, and then, two blocks are inserted into the ledger. After the
+///     validator restarts, another new block is in the middle of a consensus.
+/// Expectation: The new block is inserted into the ledger because the validator
+///     has started to validate immediately.
+unittest
+{
+    TestConf conf = { validators : 3, full_nodes : 1 , quorum_threshold : 66 };
+    auto network = makeTestNetwork(conf);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    // The node_1, node_2, node_3 are the validators
+    auto nodes = network.clients;
+    auto node_1 = nodes[0];
+    auto node_2 = nodes[1];
+    auto on_nodes = nodes[1 .. $-1];
+
+    // Create a block from the Genesis block
+    auto txs = genesisSpendable().map!(txb => txb.sign()).array();
+    txs.each!(tx => node_1.putTransaction(tx));
+    network.expectBlock(Height(1), 5.seconds);
+
+    // The node_1 restarts and is disabled to respond
+    network.restart(node_1);
+    node_1.ctrl.sleep(5.seconds);
+
+    // Make 2 blocks
+    txs = txs.map!(tx => TxBuilder(tx).sign()).array();
+    txs.each!(tx => node_2.putTransaction(tx));
+    network.expectBlock(on_nodes, Height(2), 5.seconds);
+
+    txs = txs.map!(tx => TxBuilder(tx).sign()).array();
+    txs.each!(tx => node_2.putTransaction(tx));
+    network.expectBlock(on_nodes, Height(3), 5.seconds);
+
+    // Wait for node_1 to wake up
+    node_1.ctrl.withTimeout(10.seconds,
+        (scope TestAPI api) {
+            api.getPublicKey();
+        }
+    );
+
+    network.expectBlock(Height(3), 5.seconds);
+
+    // The node_2 restart and is disabled to respond, which means that
+    // the node_2 will be slashed soon.
+    network.restart(node_2);
+    node_2.ctrl.sleep(5.seconds);
+
+    // A new block is in the middle of a consensus.
+    txs = txs.map!(tx => TxBuilder(tx).sign()).array();
+    txs.each!(tx => node_1.putTransaction(tx));
+
+    // The new block has been inserted to the ledger with the approval
+    // of the node_1, although node_2 was shutdown.
+    network.expectBlock(Height(4), 10.seconds);
+}


source/agora/test/ValidatorGeneral.d
```diff
@@ -0,0 +1,70 @@
...
+import agora.test.Base;
+
+/// Situation: A validator is stopped and wiped clean after the network starts.
+///     And then, two blocks are inserted into the ledger. After the validator
+///     restarts, another new block is in the middle of a consensus.
+/// Expectation: The new block is inserted into the ledger because the validator
+///     has started to validate immediately.
+unittest
+{
+    TestConf conf = { validators : 3, full_nodes : 1 };
```
AndrejMitrovic:
```suggestion
    TestConf conf = { validators : 3, full_nodes : 1, quorum_threshold : 1.0 };
```

source/agora/test/ValidatorGeneral.d
```diff
@@ -0,0 +1,70 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.ValidatorGeneral;
```
AndrejMitrovic: I recommend renaming it to `ValidatorRestart` or `ValidatorCleanRestart`.

source/agora/test/ValidatorGeneral.d
```diff
@@ -0,0 +1,70 @@
...
+
+    txs = txs.enumerate()
+        .map!(en => TxBuilder(en.value)
+            .refund(WK.Keys[8 + en.index].address)
+            .sign())
+        .array();
+    txs.each!(tx => node_1.putTransaction(tx));
+    network.expectBlock(Height(2), 8.seconds);
+
+    // The node_1 restarts, which means that all the date is wiped clean
```
AndrejMitrovic:
```suggestion
    // The node_1 restarts, which means that all the data is wiped clean
```

source/agora/test/ValidatorGeneral.d
```diff
@@ -0,0 +1,70 @@
...
+    network.restart(node_1);
+    assert(node_1.getBlockHeight() == 0);
```
AndrejMitrovic: There is a race condition here. The block height is zero but only for a brief moment while the node is loading.

if you add:

```
retryFor(node_1.getBlockHeight() == 2, 5.seconds, node_1.getBlockHeight().to!string);
```

Then you'll see this passes the test.

Therefore I suggest removing the line `assert(node_1.getBlockHeight() == 0);`.

I know that the intention was to test that the node started with an empty blockchain - but we can't have this assert here because it might fail.

AndrejMitrovic (changes requested):
A few comments.

source/agora/test/ValidatorGeneral.d
```diff
@@ -0,0 +1,70 @@
...
+    // Create two blocks from the Genesis block
+    auto txs = genesisSpendable().enumerate()
+        .map!(en => en.value.refund(WK.Keys[en.index].address).sign())
+        .array();
+    txs.each!(tx => node_1.putTransaction(tx));
+    network.expectBlock(Height(1), 8.seconds);
+
+    txs = txs.enumerate()
+        .map!(en => TxBuilder(en.value)
+            .refund(WK.Keys[8 + en.index].address)
```
AndrejMitrovic: By the way the `+8` here and `+16` below is strange and distracts from the test. Why not just refund to the same keys? Or even to just one public key?

It makes no difference but it makes the tests simpler. I was just wondering why it was done.

linked0:
@AndrejMitrovic, Thank you for your review. I checked all the comments and will apply them to the code.

AndrejMitrovic (approved):

AndrejMitrovic:
It looks good to me, but I think the description in #855 is different to the implementation here.

linked0:
> It looks good to me, but I think the description in #855 is different to the implementation here.

I'll check it out. Thanks.

AndrejMitrovic:
I took a look at other uses of static variables with `-transition=tls`. All uses seem to be OK. Either buffers or the logs.

linked0:
Ready for review again.
I modified the test case according to [the @AndrejMitrovic's comment](https://github.com/bpfkorea/agora/pull/1110#issuecomment-675853217) and the description of the issue.

Geod24:
This needs a rebase (?)

linked0:
I rebased it and made the code simpler about making transactions.

source/agora/test/ValidatorCleanRestart.d
```diff
@@ -141,3 +143,60 @@ unittest
...
+    txs = txs.map!(tx => TxBuilder(tx).sign()).array();
+    txs.each!(tx => node_2.putTransaction(tx));
+    network.expectBlock(on_nodes, Height(2), 5.seconds);
+
+    txs = txs.map!(tx => TxBuilder(tx).sign()).array();
+    txs.each!(tx => node_2.putTransaction(tx));
+    network.expectBlock(on_nodes, Height(3), 5.seconds);
+
+    // Wait for node_1 to wake up
+    Thread.sleep(5.seconds);
```
Geod24: Instead of having a rough sleep, do a dummy call to it with a 10 seconds timeout.
E.g. call `getPublicKey` within `node_1.ctrl.withTimeout`

source/agora/test/ValidatorCleanRestart.d
```diff
@@ -141,3 +143,60 @@ unittest
...
+    node_2.ctrl.sleep(5.seconds);
+
+    // A new block is in the middle of a consensus.
+    txs = txs.map!(tx => TxBuilder(tx).sign()).array();
+    txs.each!(tx => node_1.putTransaction(tx));
+
+    // The new block has been inserted to the ledger with the approval
+    // of the node_1, although node_2 was slashed.
+    network.expectBlock(Height(4), 10.seconds);
+}
```
Geod24: Those comments aren't correct yet, since slashing is not implemented yet.

Geod24 (approved):
LGTM, just fix the `withTimeout` thing and we can merge.

Geod24 (approved):
Merge when green

Issue #1024, opened 2020-07-10
Properly fix hashmap invalidation issue 


AndrejMitrovic:
This PR implemented a workaround: https://github.com/bpfkorea/agora/pull/1023

However it is inefficient, we should not use an array for all the list of clients since the array may grow too much (although we could put limitations in place, and remove unused clients).

Here is a test-case showing the issue:

```D
import std.stdio;
import agora.utils.Test;
import agora.common.crypto.Key;

void test () @safe
{
    int[PublicKey] map;
    map[WK.Keys[0].address] = 10;
    map[WK.Keys[1].address] = 20;
    map[WK.Keys[2].address] = 30;
    map[WK.Keys[3].address] = 40;
    map[WK.Keys[4].address] = 50;
    size_t count;
    foreach (ref key, ref value; map)
    {
        writeln("In map (before blowItUp) Key: ", key, ", value: ", value);
        ++count;
        if (count == 3)
            blowItUp(map);
        writeln("In map (after blowItUp) Key: ", key, ", value: ", value);
    }
}

void blowItUp (ref int[PublicKey] map) @safe
{
    for (int v = 6; v < 512; ++v)
    {
        map[WK.Keys[v].address] = v * 10;
        writeln("Adding: Key: ", WK.Keys[v].address, ", value: ", v * 10);
    }
}

unittest
{
    test();
}
```

Notice the corruption here in the printouts:

```
n map (after blowItUp) Key: GDC22CFFKB4ZNRZUP6EMRIGVZSQEPSNH2CBMWLU5GLGKE36M3KX5YD36, value: 30
In map (before blowItUp) Key: GAEAAAAAAAAAAAGPHEVAWAIAAAAAAAAAAAAAAAAAIAA5SCQBAAAABOWV, value: 10
In map (after blowItUp) Key: GAEAAAAAAAAAAAGPHEVAWAIAAAAAAAAAAAAAAAAAIAA5SCQBAAAABOWV, value: 10
In map (before blowItUp) Key: GABQAAAAAAAAAAFKHEVAWAIAAAAAAAAAAAAAAAAAUD5NQCQBAAAAB2H7, value: 10
In map (after blowItUp) Key: GABQAAAAAAAAAAFKHEVAWAIAAAAAAAAAAAAAAAAAUD5NQCQBAAAAB2H7, value: 10
[...]

PR #130, opened 2019-08-01
[trivial] Reorder imports according to our style guide

AndrejMitrovic:
https://github.com/bpfkorea/agora/blob/v0.x.x/doc/Style.md#modules

PR #1114, opened 2020-08-13
Fix #825 - building agora based on agora-builder image as opposed to alpine image

### LINKED ISSUE #825: Provide a "base docker build" image
Currently, building the docker image from scratch can be quite slow, as it installs 45 packages, including an upgraded GCC.
Since we're depending on edge, this happens quite often, slowing developers and CI alike.

ferencdg:
building agora based on agora-builder image as opposed to alpine image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 # Build Agora from source
-FROM alpine:3.12.0 AS Builder
+FROM bpfk/agora-builder:latest AS Builder
 ARG DUB_OPTIONS
-RUN apk --no-cache add build-base dtools-rdmd dub git ldc libsodium-dev openssl openssl-dev sqlite-dev zlib-dev
 ADD . /root/agora/
 WORKDIR /root/agora/
 RUN dub build --skip-registry=all --compiler=ldc2 ${DUB_OPTIONS}


Geod24 (approved):

TrustHenry (approved):
LGTM

AndrejMitrovic:
If you rebase it should go green.

Geod24:
Looks like you added an extra commit here.

Dockerfile
```diff
@@ -1,7 +1,6 @@
 # Build Agora from source
```
ferencdg, line 1: this issue of #825 is very trivial, so I included my #1024 changes here too, please let me know if you want me to separate the two

source/agora/network/NetworkManager.d
```diff
@@ -326,10 +328,10 @@ public class NetworkManager
     protected ConnectionTask[Address] connection_tasks;
 
     /// List of validator clients
-    protected NetworkClient[] validators;
+    protected DList!NetworkClient validators;
```
ferencdg: cannot use single list here, as addition to the end of SList is O(n)

source/agora/network/NetworkManager.d
```diff
@@ -660,21 +662,21 @@ public class NetworkManager
     }
 
     ///
-    private bool minPeersConnected ()  pure nothrow @safe @nogc
+    private bool minPeersConnected ()  pure nothrow @safe
     {
         return this.required_peer_keys.length == 0 &&
-            this.peers.length >= this.node_config.min_listeners;
+            this.peers[].walkLength >= this.node_config.min_listeners;
```
ferencdg: This is an O(n) operation on DList, but in case the peer list is not long, it should be fine. Please let me know if I should find an other DList implementation that can produce the list size in O(1)

source/agora/network/NetworkManager.d
```diff
@@ -660,21 +662,21 @@ public class NetworkManager
     }
 
     ///
-    private bool minPeersConnected ()  pure nothrow @safe @nogc
+    private bool minPeersConnected ()  pure nothrow @safe
```
ferencdg: walkLength needs GC

source/agora/network/NetworkManager.d
```diff
@@ -660,21 +662,21 @@ public class NetworkManager
...
     {
         return this.required_peer_keys.length == 0 &&
-            this.peers.filter!(node =>
+            this.peers[].filter!(node =>
                 !this.banman.isBanned(node.address)).count >= this.node_config.max_listeners;
     }
 
     /// Returns: the list of node IPs this node is connected to
-    public NodeInfo getNetworkInfo () pure nothrow @safe @nogc
+    public NodeInfo getNetworkInfo () pure nothrow @safe
```
linked0, line 679: This kind of modification should be another PR or commit because it's not related to this commit.

source/agora/node/FullNode.d
```diff
@@ -238,7 +238,7 @@ public class FullNode : API
     }
 
     /// GET: /node_info
-    public override NodeInfo getNodeInfo () pure nothrow @safe @nogc
+    public override NodeInfo getNodeInfo () pure nothrow @safe
```
linked0, line 241: The same as the above comment.

linked0 (commented):
The second and third commits are not related to this PR because the title of this PR doesn't explain the commits it contains. So you should separate the commits into another PR.
If any questions, please let us know on the Slack.
Thanks.

ferencdg (commented):
> The second and third commits are not related to this PR because the title of this PR doesn't explain the commits it contains. So you should separate the commits into another PR.
> If any questions, please let us know on the Slack.
> Thanks.

yes, already removed and will create a separate PR

AndrejMitrovic:
It should be fairly easy to wrap `DList` in another struct which keeps track of the length. Something like:

```D
struct List (T)
{
    private DList!T dlist;
    private size_t _length;

    public size_t length () @nogc { return _length; }
    // implement forwarding of opSlice / insertBack, update _length when insertBack is called
}
```

ferencdg:
> It should be fairly easy to wrap `DList` in another struct which keeps track of the length. Something like:
> 
> ```d
> struct List (T)
> {
>     private DList!T dlist;
>     private size_t _length;
> 
>     public size_t length () @nogc { return _length; }
>     // implement forwarding of opSlice / insertBack, update _length when insertBack is called
> }
> ```

I was also thinking about this idea, but there aren't many elements in the container. Discussed it with Mathias, and for now we would just use the original DList.

Dockerfile
linked0: I know what you intended as you explained. But, please refer to [the example of a simple PR](https://github.com/bpfkorea/agora/pull/130)

In addition, it would be good to add the issue number in the pull request's description like this.
`Related to #825` or `Fixes #825`
Please refer to other PRs.

Thanks!

linked0:
I'm removing improper comments about function attributes like @nogc.
Sorry for the confusion.

Geod24:
Can you clean up the commits ?

PR #1116, opened 2020-08-13
Use uint instead of floating-point for thresholds

### LINKED ISSUE #1112: Replace floating-point with integrals in the Config
For example the `quorum_threshold` can be set between 0 to 100.0, but we don't actually need more than 100 steps.

AndrejMitrovic:
Fixes #1112


source/agora/common/Config.d
```diff
@@ -319,8 +319,8 @@ private NodeConfig parseNodeConfig (Node* node, const ref CommandLine cmdln)
     auto port = get!(ushort, "node", "port")(cmdln, node);
     auto validator_cycle = get!(uint, "node", "validator_cycle")(cmdln, node);
     auto max_quorum_nodes = get!(uint, "node", "max_quorum_nodes")(cmdln, node);
-    auto quorum_threshold = get!(double, "node", "quorum_threshold")(cmdln, node);
-    assert(quorum_threshold > 0.0 && quorum_threshold <= 1.0);
+    auto quorum_threshold = get!(uint, "node", "quorum_threshold")(cmdln, node);
+    assert(quorum_threshold > 1 && quorum_threshold <= 100);
```
linked0: I think it should be `quorum_threshold >= 1`, not `quorum_threshold > 1`.

linked0 (approved):

Issue #1117, opened 2020-08-13
List of issues causing CircleCI failures


AndrejMitrovic:
#### `HTTPStatusException` was not propagated properly (Fixed in https://github.com/bpfkorea/agora/pull/1109)

-----

#### Properly handle `HTTPStatusException` (Fixed in https://github.com/bpfkorea/agora/pull/1106)

-----

#### Request retry delays and timeouts might be too high (Fixed in #1118)

Some further ideas to improve the CircleCI succeeding:

- Reduce the `retry_delay` option. This is essentially the "sleeping time" between each time a request is retried. Currently it's set to 1.5 seconds. I think this is too long of a waiting time between retries.
- Increase the `max_retries` option so a node doesn't get prematurely banned.
- Change the ban manager logging to use `.info` instead of `.trace`. Right now our loggers only emit `info` level messages, but the ban manager still uses `trace` level - so if a node does get banned we won't see it in the logs in our CI.

-----

#### IP & Port when calling `registerListener` might be incorrect

As mentioned here: https://github.com/bpfkorea/agora/pull/1115#issue-467223207

However the following issue might be related:

-----

#### `listenHTTP` is called too late: (Fixed in #1118)

```
node-2_1_d220995f0bb5 | 2020-08-13 12:13:33,996 Info [agora.node.Validator] - start()
node-2_1_d220995f0bb5 | 2020-08-13 12:13:33,996 Info [agora.node.Validator] - startPeriodicDiscovery
node-2_1_d220995f0bb5 | 2020-08-13 12:13:33,996 Info [agora.node.Validator] - startPeriodicDiscovery inside task
[...]

PR #1118, opened 2020-08-13
Make the node open its ports before doing discovery()

AndrejMitrovic:
Another small part towards fixing #1117.


Geod24:
We really need that `schedule` primitive...

linked0:
All the calls in the `node.start()` were also executed asynchronously before through the `Vibe.d`'s timer and asynchronous tasks. Is there a big difference between before and after? The commit message says that this makes the node not to wait for the first discovery round to complete though.

AndrejMitrovic:
I'll rename the commit and change the description.

AndrejMitrovic:
The issue wasn't really about asynchronicity, it was about scheduling.

The `runTask` call would spawn a new fiber and suspend the calling fiber, and the new fiber would spawn many other requests. This starved the first fiber (the one in `main.d` calling `runNode`) - so it only gets to call `listenHTTP` later.

But we want `listenHTTP` to be called right away to make the node available for incoming connections.

Fixed the description.

linked0:
With your explanation, I've just realized that this is about scheduling and just seen the log shown in #1117. Thanks for the detailed explanation.

linked0 (approved):

AndrejMitrovic:
Thanks. I'll merge it with the approval.

Issue #1107, opened 2020-08-09
Remove `makeChainedTransactions`


Geod24:
We have a code comment to do it, might as well create an issue for it:
https://github.com/bpfkorea/agora/blob/b602254d522f5bc2e5b5e96525c4d61e7c9180ff/source/agora/utils/Test.d#L147-L148

There is also the "makeTransactionsForFreezing" in Ledger.d which is essentially a copy of it:
https://github.com/bpfkorea/agora/blob/b602254d522f5bc2e5b5e96525c4d61e7c9180ff/source/agora/node/Ledger.d#L795-L845

Geod24:
@kchulj : Seems there are just a few usages left:
```
source/agora/consensus/validation/Block.d:589:        makeChainedTransactions(gen_key, prev_txs, 1, 20_000_000));
source/agora/consensus/validation/Block.d:618:    auto txs_1 = makeChainedTransactions(gen_key, null, 1,
source/agora/consensus/validation/Block.d:740:    auto txs_1 = makeChainedTransactions(gen_key, null, 1,
source/agora/node/Ledger.d:764:    //       Currently due to the design of `makeChainedTransactions`,
source/agora/test/Base.d:1408:        auto txs = makeChainedTransactions(WK.Keys.Genesis,
source/agora/test/NetworkManager.d:81:                auto txs = makeChainedTransactions(prev_key, last_tx, 1);
source/agora/utils/Test.d:168:public Transaction[] makeChainedTransactions (KeyPair key_pair,
source/agora/utils/Test.d:244:    auto txes = makeChainedTransactions(gen_key, null, 1);
source/agora/utils/Test.d:254:    txes = makeChainedTransactions(gen_key, txes, 1);
source/agora/utils/Test.d:264:    txes = makeChainedTransactions(gen_key, prev_txs, 1, TotalSpend);
source/agora/utils/Test.d:281:    const(Transaction)[] txes = makeChainedTransactions(gen_key, null, 1);
source/agora/utils/Test.d:282:    txes = makeChainedTransactions(gen_key, txes, 1);
```

kchulj:
Yes, some of the usages have `Amount` as a parameter and I am working on adding it to `TxBuilder`.
The usages in `Test.d` are planning to be removed because they are specifically for `makeChainedTransactions`.

PR #1119, opened 2020-08-14
Make genesisSpendable() available to non-genesis blocks

kchulj:
The function `GenesisSpendable ()` filters the `Payment` output of a genesis block and returns a range as below :
```D
GenesisBlock.txs
        .filter!(tx => tx.type == TxType.Payment)
        .map!(tx => iota(tx.outputs.length).map!(idx => TxBuilder(tx, cast(uint)idx)))
        .joiner();
```

Since we are transitioning away from `makeChainedTransactions` to `TxBuilder`, we need to use this function for the next blocks as well. This brief change shifts the original function into a generic `spendable ()` and still makes `genesisSpendable ()` available for the unit tests.

Developed with @Geod24 
Relates issue #1107

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -35,6 +35,7 @@ import agora.common.crypto.Schnorr;
 import agora.common.Hash;
 import agora.common.Serializer;
 import agora.common.Types;
+import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.genesis.Test;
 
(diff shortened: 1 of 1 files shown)


source/agora/utils/Test.d
```diff
@@ -740,15 +741,21 @@ public struct TxBuilder
 
 /// Returns:
 ///   A range of `TxBuilder`s which reference each Payment output of
-///   the Genesis block
-public auto genesisSpendable () @safe
+///   the input block
+public auto spendable (in Block b) @safe
```
TrustHenry: Document in input parameter`in Block b` is missing.

TrustHenry:
```suggestion
public auto spendable (const ref Block block) @safe pure nothrow
```
What do you think?

source/agora/utils/Test.d
```diff
@@ -740,15 +741,21 @@ public struct TxBuilder
...
 {
-    return GenesisBlock.txs
+    return b.txs
```
TrustHenry:
```suggestion
    return block.txs
```

source/agora/utils/Test.d
linked0: For the name of the parameter `b`, `block` would be more readable. The function is very short though.

kchulj: LGTM!

source/agora/utils/Test.d
```diff
@@ -738,17 +739,34 @@ public struct TxBuilder
...
 {
-    return GenesisBlock.txs
+    return block.txs
         .filter!(tx => tx.type == TxType.Payment)
         .map!(tx => iota(tx.outputs.length).map!(idx => TxBuilder(tx, cast(uint)idx)))
         .joiner();
 }
 
+/// Convenience function for Genesis Block unittest
+public auto genesisSpendable () @safe
```
AndrejMitrovic:
```suggestion
public auto genesisSpendable () @safe pure nothrow
```

Geod24 (approved):

PR #1123, opened 2020-08-18
Fix #1024 - fix segfault caused by incorrect hashmap iteration 

### LINKED ISSUE #1024: Properly fix hashmap invalidation issue

ferencdg:
issue #1024: fix segfault caused by incorrect hashmap iteration with addition

### DIFF
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -47,9 +47,11 @@ import vibe.web.rest;
 
 import std.algorithm;
 import std.array;
+import std.container : DList;
 import std.exception;
 import std.format;
 import std.random;
+import std.range : walkLength;
 
 import core.stdc.time;
 import core.time;
@@ -326,10 +328,10 @@ public class NetworkManager
     protected ConnectionTask[Address] connection_tasks;
 
     /// List of validator clients
-    protected NetworkClient[] validators;
+    protected DList!NetworkClient validators;
 
     /// All connected nodes (Validators & FullNodes)
-    protected NetworkClient[] peers;
+    protected DList!NetworkClient peers;
 
     /// Easy lookup of currently connected peers
     protected Set!Address connected_peers;
@@ -390,11 +392,11 @@ public class NetworkManager
     private void onHandshakeComplete (scope ref NodeConnInfo node)
     {
         this.connected_peers.put(node.address);
-        this.peers ~= node.client;
+        this.peers.insertBack(node.client);
 
         if (node.is_validator)
         {
-            this.validators ~= node.client;
+            this.validators.insertBack(node.client);
             this.required_peer_keys.remove(node.key);
         }
 
@@ -539,7 +541,7 @@ public class NetworkManager
         {
             void catchup ()
             {
-                if (this.peers.length == 0)  // no clients yet (discovery)
+                if (this.peers.empty())  // no clients yet (discovery)
                     return;
 
                 this.getBlocksFrom(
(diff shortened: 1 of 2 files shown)


source/agora/network/NetworkManager.d
```diff
@@ -792,7 +794,7 @@ public class NetworkManager
 
     public void sendPreimage (PreImageInfo preimage) @safe
     {
-        foreach (ref node; this.peers)
+        foreach (ref node; this.peers[])
```
Geod24, line 797: `DList` does not define `opApply` ?

Geod24: Ah, `opSlice` returns a `Range`, how weird.

Geod24 (approved):
LGTM

linked0 (approved):
LGTM

PR #1121, opened 2020-08-14
Test for start validating after new validators restart

linked0:
This is the test case for checking new validators start validating after they restart with all the data missing. After restarting, they need to do the catch-up process and then start validating for new blocks being created.

Relates #854

PR #1126, opened 2020-08-19
Don't use a static variable for a status flag

AndrejMitrovic:
Found in https://github.com/bpfkorea/agora/pull/1121


Geod24:
LocalRest catches bugs <3

Geod24 (approved):

linked0:
The `is_running` should be reset like in the `shutdown` function, shouldn't it be?

AndrejMitrovic:
> The is_running should be reset like in the shutdown function, shouldn't it be?

When LocalRest restarts the node it destroys the Validator/FullNode class and instantiates a new one.

linked0 (approved):

PR #1127, opened 2020-08-19
Replace most `makeChainedTransactions` with `TxBuilder`

kchulj:
`makeChainedTransactions` is an outdated function which assumes a block always contains exactly 8 transactions.
Since we have transitioned away from creating a block based on the number of transactions, we should use `TxBuilder` instead.

Relates #1107

### DIFF
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -435,7 +435,7 @@ unittest
     auto gen_hash = GenesisBlock.header.hashFull();
 
     GenesisBlock.txs.each!(tx => utxos.put(tx));
-    auto block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, null, 1));
+    auto block = GenesisBlock.makeNewBlock(genesisSpendable().map!(txb => txb.sign()));
 
     // height check
     assert(block.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
@@ -498,7 +498,7 @@ unittest
     prev_txs.each!(tx => utxos.put(tx));  // these will be spent
 
     auto prev_block = block;
-    block = block.makeNewBlock(makeChainedTransactions(gen_key, prev_txs, 1));
+    block = block.makeNewBlock(prev_txs.map!(tx => TxBuilder(tx).sign()));
     assert(block.isValid(prev_block.header.height, prev_block.header.hashFull(),
         findUTXO, Enrollment.MinValidatorCount));
 
@@ -547,7 +547,7 @@ unittest
     };
 
     // consumed all utxo => fail
-    block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, null, 1));
+    block = GenesisBlock.makeNewBlock(genesisSpendable().map!(txb => txb.sign()));
     assert(block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
             findNonSpent, Enrollment.MinValidatorCount));
 
@@ -568,7 +568,7 @@ unittest
     // we stopped validation due to a double-spend
     assert(used_set.length == double_spend.length - 1);
 
-    block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, prev_txs, 1));
+    block = GenesisBlock.makeNewBlock(prev_txs.map!(tx => TxBuilder(tx).sign()));
     assert(block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
         findUTXO, Enrollment.MinValidatorCount));
 
--- a/source/agora/test/BanManager.d
+++ b/source/agora/test/BanManager.d
@@ -57,11 +57,25 @@ unittest
     // generate enough transactions to form 'count' blocks
     Transaction[] genBlockTransactions (size_t count)
     {
-        auto txes = makeChainedTransactions(gen_key, last_txs, count);
-        // keep track of last tx's to chain them to
-        last_txs = txes[$ - 8 .. $];
-        all_txs ~= txes;
-        return txes;
+        assert(count > 0);
+        Transaction[] txes;
+
+         if (!last_txs.length)
+         {
+             txes = genesisSpendable().map!(txb => txb.sign()).array();
+             last_txs = txes[$ - 8 .. $];
+             all_txs ~= last_txs;
+             count--;
+         }
+
+         foreach (idx; 0 .. count)
+         {
+             txes = last_txs.map!(tx => TxBuilder(tx).sign()).array();
+             // keep track of last tx's to chain them to
+             last_txs = txes[$ - 8 .. $];
+             all_txs ~= txes;
+         }
+         return txes;
     }
 
     genBlockTransactions(1).each!(tx => node_1.putTransaction(tx));
--- a/source/agora/test/Ledger.d
+++ b/source/agora/test/Ledger.d
@@ -45,21 +45,24 @@ unittest
 
     Transaction[][] block_txes; /// per-block array of transactions (genesis not included)
     Transaction[] last_txs;
+
+    // Get the genesis block, make sure it's the only block externalized
+    auto blocks = node_1.getBlocksFrom(0, 2);
+    assert(blocks.length == 1);
+
     foreach (block_idx; 0 .. 10)  // create 10 blocks
     {
         // create enough tx's for a single block
-        auto txs = makeChainedTransactions(WK.Keys.Genesis, last_txs, 1);
+        auto txs = blocks[block_idx].spendable().map!(txb => txb.sign()).array();
 
         // send it to one node
         txs.each!(tx => node_1.putTransaction(tx));
         network.expectBlock(Height(block_idx + 1), 4.seconds);
 
+        blocks ~= node_1.getBlocksFrom(block_idx + 1, 1);
         block_txes ~= txs.sort.array;
         last_txs = txs;
     }
-
-    // get all the blocks (including genesis block)
-    auto blocks = node_1.getBlocksFrom(0, 101);
 
     assert(blocks[0] == network.blocks[0]);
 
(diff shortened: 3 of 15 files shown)


source/agora/utils/Test.d
```diff
@@ -736,13 +736,17 @@ public struct TxBuilder
 }
```
AndrejMitrovic: This commit should be squashed with the previous one.

source/agora/common/crypto/Key.d
```diff
@@ -120,8 +120,9 @@ public struct PublicKey
     }
 
     /// Uses Stellar's representation instead of hex
-    public string toString () const @trusted
+    public string toString () const @trusted nothrow
     {
+        scope (failure) assert(0);
```
AndrejMitrovic: Unrelated change. I think you may want to rebase your PR first because this was merged a little while ago: https://github.com/bpfkorea/agora/pull/1125

source/agora/consensus/data/Block.d
```diff
@@ -439,7 +439,7 @@ unittest
 *******************************************************************************/
 
 public Block makeNewBlock (Transactions)(const ref Block prev_block,
-    Transactions txs, Enrollment[] enrollments) @safe nothrow
+    Transactions txs, Enrollment[] enrollments) @safe
```
AndrejMitrovic: Ditto comment about https://github.com/bpfkorea/agora/pull/1125 being merged, so I think this should remain being `nothrow`.

source/agora/consensus/validation/Block.d
```diff
@@ -435,7 +435,7 @@ unittest
     auto gen_hash = GenesisBlock.header.hashFull();
 
     GenesisBlock.txs.each!(tx => utxos.put(tx));
-    auto block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, null, 1));
+    auto block = GenesisBlock.makeNewBlock(genesisSpendable().map!(txb => txb.sign()).array());
```
```suggestion
    auto block = GenesisBlock.makeNewBlock(genesisSpendable().map!(txb => txb.sign()));
```

`makeNewBlock` accepts an input range, so there's no need to call `.array` - it's an unnecessary allocation.

source/agora/consensus/validation/Block.d
```diff
@@ -498,7 +498,7 @@ unittest
     prev_txs.each!(tx => utxos.put(tx));  // these will be spent
 
     auto prev_block = block;
-    block = block.makeNewBlock(makeChainedTransactions(gen_key, prev_txs, 1));
+    block = block.makeNewBlock(prev_txs.map!(tx => TxBuilder(tx).sign()).array());
```
```suggestion
    block = block.makeNewBlock(prev_txs.map!(tx => TxBuilder(tx).sign()));
```

source/agora/consensus/validation/Block.d
```diff
@@ -547,7 +547,7 @@ unittest
     };
 
     // consumed all utxo => fail
-    block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, null, 1));
+    block = GenesisBlock.makeNewBlock(genesisSpendable().map!(txb => txb.sign()).array());
```
```suggestion
    block = GenesisBlock.makeNewBlock(genesisSpendable().map!(txb => txb.sign()));
```

source/agora/consensus/validation/Block.d
```diff
@@ -568,7 +568,7 @@ unittest
     // we stopped validation due to a double-spend
     assert(used_set.length == double_spend.length - 1);
 
-    block = GenesisBlock.makeNewBlock(makeChainedTransactions(gen_key, prev_txs, 1));
+    block = GenesisBlock.makeNewBlock(prev_txs.map!(tx => TxBuilder(tx).sign()).array());
```
```suggestion
    block = GenesisBlock.makeNewBlock(prev_txs.map!(tx => TxBuilder(tx).sign()));
```

source/agora/test/NetworkManager.d
```diff
@@ -76,16 +77,22 @@ unittest
 
             Block last_block;
             // make 20 blocks which have an invalid previous hash
-            foreach (idx; 0 .. 20)
+
+            // genesis case
+            if (!last_tx.length)
```
AndrejMitrovic: Styling nit: when there's only a single statement then there is no need to add braces. So:

```D
if (!last_tx.length)
    last_tx = ...
```

source/agora/test/NetworkManager.d
```diff
@@ -76,16 +77,22 @@ unittest
 
             Block last_block;
             // make 20 blocks which have an invalid previous hash
-            foreach (idx; 0 .. 20)
+
```
AndrejMitrovic: I am not sure if this is the correct change. This code segment created invalid transactions **on purpose**, to test the behavior of the system. I will have to double-check the code again tomorrow to see if it does the right thing here.

source/agora/test/NetworkManager.d
```diff
@@ -151,11 +158,20 @@ unittest
 
     Transaction[][] block_txes; /// per-block array of transactions (genesis not included)
     Transaction[] last_txs;
-    foreach (block_idx; 0 .. 10)  // create 10 blocks
+
+    // create genesis block
+    if (!last_txs.length)
```
AndrejMitrovic: This check shouldn't be necessary, `last_txs` is empty here for sure.

source/agora/test/NetworkManager.d
```diff
@@ -151,11 +158,20 @@ unittest
...
     {
-        // create enough tx's for a single block
-        auto txs = makeChainedTransactions(WK.Keys.Genesis, last_txs, 1);
+        last_txs = genesisSpendable().map!(txb => txb.sign()).array();
+        last_txs.each!(tx => node_validator.putTransaction(tx));
+        network.expectBlock([node_validator], Height(1), 4.seconds);
+        block_txes ~= last_txs.sort.array;
+    }
 
+    foreach (block_idx; 1 .. 10)  // create a total of 10 blocks
```
AndrejMitrovic:
```suggestion
    foreach (block_idx; 1 .. 10)  // create 9 additional blocks
```

Because this loop creates 9 blocks (even if the `if` block before created 1).

source/agora/utils/Test.d
```diff
@@ -235,12 +236,10 @@ public Transaction[] makeChainedTransactions (KeyPair key_pair,
 ///
 unittest
 {
-    import agora.consensus.data.Block;
-
     auto gen_key = WK.Keys.Genesis;
 
     /// should spend genesis block's outputs
-    auto txes = makeChainedTransactions(gen_key, null, 1);
+    auto txes = genesisSpendable().map!(txb => txb.sign()).array();
```
AndrejMitrovic: I don't think the changes in these tests make sense. These tests are for the `makeChainedTransactions` function, not for `TxBuilder`.

When a future commit removes `makeChainedTransactions` then these unittests will also be removed. But if `makeChainedTransaction` is not removed yet in this PR then its tests should not be touched.

source/agora/utils/Test.d
```diff
@@ -274,11 +273,9 @@ unittest
 /// example of chaining
 unittest
 {
-    import agora.consensus.data.Block;
-
     auto gen_key = WK.Keys.Genesis;
-    const(Transaction)[] txes = makeChainedTransactions(gen_key, null, 1);
-    txes = makeChainedTransactions(gen_key, txes, 1);
+    auto txs = genesisSpendable().map!(txb => txb.sign()).array();
```
AndrejMitrovic: Ditto about not touching this test.

AndrejMitrovic (changes requested):
A few comments, otherwise looks good.

source/agora/test/NetworkManager.d
kchulj: It was not the correct change because it was not creating invalid transactions. I left it untouched in the new commit.
`TxBuilder` does not allow one to build invalid transactions, therefore, it would have to be modified after it has been built.
I will try this in a separate issue.

source/agora/common/crypto/Key.d
kchulj: Thanks for the suggestions @AndrejMitrovic 
The problem with the rebase has been fixed and the changes have been applied.
Ready for review again.

AndrejMitrovic (approved):

source/agora/test/BanManager.d
```diff
@@ -53,15 +53,29 @@ unittest
 
     Transaction[] all_txs;
     Transaction[] last_txs;
+    Transaction[] txes;
```
Geod24: Why is this outside the function's scope ?

source/agora/test/Ledger.d
```diff
@@ -45,22 +45,25 @@ unittest
 
     Transaction[][] block_txes; /// per-block array of transactions (genesis not included)
     Transaction[] last_txs;
+
+    // get all the blocks (including genesis block)
+    auto blocks = node_1.getBlocksFrom(0, 101);
+    assert(blocks.length == 1);
```
Geod24: If the expectation is that the length is exactly 1, then:
```suggestion
    // Get the genesis block, make sure it's the only block externalized
    auto blocks = node_1.getBlocksFrom(0, 2);
    assert(blocks.length == 1);
```

source/agora/test/NetworkManager.d
```diff
@@ -23,6 +23,7 @@ import agora.common.Metadata;
 import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.ConsensusParams;
+import agora.consensus.data.genesis.Test;
```
Geod24: Why the import ?

source/agora/test/ValidatorCount.d
```diff
@@ -22,6 +22,7 @@ module agora.test.ValidatorCount;
 
 version (unittest):
 
+import agora.consensus.data.Block;
```
Geod24: Why the import ?

source/agora/test/ValidatorCount.d
```diff
@@ -45,24 +46,31 @@ unittest
     auto node_1 = nodes[0];
 
     auto gen_key_pair = WK.Keys.Genesis;
+    const(Block)[] blocks = node_1.getBlocksFrom(0, 101);
```
Geod24: I'll add a comment here / make it consistent with the other call.

Geod24 (changes requested):
Just a few nitpicks, otherwise looks good.

linked0 (approved):
LGTM

Geod24 (approved):

PR #1132, opened 2020-08-21
fix #1128 - logger segfault

### LINKED ISSUE #1128: fixing Log.d that can cause segfaults under special circumstances
The current implementation of agora logger in Log.d doesn't handle object lifetime correctly and in some corner cases it can even cause segfaults or log messages might be lost. I will post some code to reprocude those segfaults later.

ferencdg:
logger segfault, see issue #1128

### DIFF
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -48,19 +48,12 @@ version (unittest)
 
 *******************************************************************************/
 
+mixin AddLogger!();
+
 static this ()
 {
-    log = Logger(__MODULE__);
     TransactionPool.initialize();
 }
-
-static ~this ()
-{
-    destroy(log);
-}
-
-/// Logger instance
-private Logger log;
 
 /// A transaction pool that is serializable to disk, backed by SQLite
 public class TransactionPool
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -49,21 +49,24 @@ public struct Logger
         import core.memory;
         this.logger = Ocean.Log.lookup(moduleName);
         this.logger.buffer(new char[](16384));
-        GC.addRoot(cast(void*)this.logger);
-    }
-
-    /// dtor
-    public ~this ()
-    {
-        import core.memory;
-        GC.removeRoot(cast(void*)this.logger);
     }
 
     public void opDispatch (string call, Args...) (Args args)
     {
         try
         {
-            mixin("this.logger." ~ call ~ "(args);");
+            import core.memory : GC;
+            if (GC.inFinalizer())
+            {
+                writeln("allowing logging from the destructor on a GC thread would risk running into segfaults, please see issue #1128");
+                return;
+            }
+            else if (logger is null)
+            {
+                assert(0,"\nplease make sure you are declaring the \nmixin AddLogger!(); statement on top, followed by:\nstatic this{}; followed by:\nstatic ~this{};");
+                return;
+            }
+            mixin("this.logger." ~ call ~ "(args);");            
         }
         catch (Exception ex)
         {
@@ -91,15 +94,10 @@ public struct Logger
 /// Insert a logger in the current scope, named log
 public template AddLogger (string moduleName = __MODULE__)
 {
-    private Logger log;
+    private Logger* log;
     static this ()
     {
-        log = Logger(moduleName);
-    }
-
-    static ~this()
-    {
-        destroy(log);
+        log = new Logger(moduleName);
     }
 }
 


source/agora/utils/Log.d
```diff
@@ -49,21 +49,24 @@ public struct Logger
...
     }
 
     public void opDispatch (string call, Args...) (Args args)
     {
         try
         {
-            mixin("this.logger." ~ call ~ "(args);");
+            import core.memory : GC;
+            if (GC.inFinalizer())
+            {
```
ferencdg, line 60: Please let me know if you know why stderr.writeln doesn't work inside the if statement, I got no useful message from the compiler. It seems ldc2 is having problems with outputting the real error from iside opDispatch.

image

outputs

image

AndrejMitrovic: I think it's a compiler bug. I ran into something like this before and at some point removed a workaround:

https://github.com/bpfkorea/agora/commit/4b60dc4ee703138a79ea342e82086b0fb04d9a15#diff-bea1a77bdfb5d54a9e7d698db79ed2dc

However the underlying problem might still be there.

AndrejMitrovic: Btw, ideally we wouldn't need this `opDispatch` trampoline. It's only there to make the logger `nothrow`. There should probably be a better way to do this - e.g. just infer nothrow from the call site if possible.

ferencdg: > I think it's a compiler bug. I ran into something like this before and at some point removed a workaround:
> 
> [4b60dc4#diff-bea1a77bdfb5d54a9e7d698db79ed2dc](https://github.com/bpfkorea/agora/commit/4b60dc4ee703138a79ea342e82086b0fb04d9a15#diff-bea1a77bdfb5d54a9e7d698db79ed2dc)
> 
> However the underlying problem might still be there.

it is very puzzling because 
std.stdio.writeln("") works, but std.stdio.stderr.writeln("") doesn't

and in my simplified example above, I can use stderr.writeln("") without any problems, it is just causing problems inside agora(althouh even in my simplified example I would get no useful compiler error message in case there is some problem within opDispatch)

ferencdg: anyway reported the issue: https://github.com/ldc-developers/ldc/issues/3545

AndrejMitrovic: It's actually a DMD front-end issue. LDC uses DMD's front-end. Most compile-time bugs are actually DMD bugs.

AndrejMitrovic: s/It's/should be

AndrejMitrovic: Yeah it fails with DMD too:

```diff
diff --git a/dub.json b/dub.json
index 5cb74b88b..7bf021ac5 100644
--- a/dub.json
+++ b/dub.json
@@ -70,7 +70,6 @@
     "buildRequirements": [ "allowWarnings" ],
 
 	"toolchainRequirements": {
-		"dmd": "no",
 		"gdc": "no",
 		"ldc": "~>1.19"
 	},
diff --git a/source/agora/utils/Log.d b/source/agora/utils/Log.d
index e3cd83a6f..22ccecc8a 100644
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -58,7 +58,7 @@ public struct Logger
             import core.memory : GC;
             if (GC.inFinalizer())
             {
-                writeln("allowing logging from the destructor on a GC thread would risk running into segfaults, please see issue #1128");
+                stderr.writeln("allowing logging from the destructor on a GC thread would risk running into segfaults, please see issue #1128");
                 return;
             }
             else if (logger is null)
@@ -66,7 +66,7 @@ public struct Logger
                 assert(0,"\nplease make sure you are declaring the \nmixin AddLogger!(); statement on top, followed by:\nstatic this{}; followed by:\nstatic ~this{};");
                 return;
             }
-            mixin("this.logger." ~ call ~ "(args);");            
+            mixin("this.logger." ~ call ~ "(args);");
         }
         catch (Exception ex)
         {
```

and:

```
dub build --compiler=dmd
```

Same error.

ferencdg: thanks, let me complain there too

ferencdg: known issue and marked as critcal... 
https://issues.dlang.org/show_bug.cgi?id=14145

AndrejMitrovic: Feel free to comment on it. Sometimes commenting helps bring an issue back to focus - there might be compiler devs interested in fixing it.

Geod24:
I'm not sure I understand this fix correctly: This is to prevent SEGV when using the Logger from a GC-run destructor ? But does that ever happen ? The lifetime of those log object is bound to the thread lifetime, so as long as they aren't accessed when the Thread is destroyed, we should be find, or am I missing something ?

ferencdg:
> I'm not sure I understand this fix correctly: This is to prevent SEGV when using the Logger from a GC-run destructor ? But does that ever happen ? The lifetime of those log object is bound to the thread lifetime, so as long as they aren't accessed when the Thread is destroyed, we should be find, or am I missing something ?

I think in my original issue, I mentioned 2 cases when segfault can occur:
1. static destructor/static constructor/log template is arranged in a special order
2. the case pertaining to your question

to answer your question this was the original setup:
client wishing to use logger(usually GC allocated) -> Agora logger object(thread lifetime) -> Oceans logger(GC allocated)

we saw it from the ManagedDatabase case that the client allocated on the GC can outlive the thread destruction, so it would use an object(s) that can potentially be destroyed(Agora logger object and Oceans logger)

ferencdg:
> > I'm not sure I understand this fix correctly: This is to prevent SEGV when using the Logger from a GC-run destructor ? But does that ever happen ? The lifetime of those log object is bound to the thread lifetime, so as long as they aren't accessed when the Thread is destroyed, we should be find, or am I missing something ?
> 
> I think in my original issue, I mentioned 2 cases when segfault can occur:
> 
> 1. static destructor/static constructor/log template is arranged in a special order
> 2. the case pertaining to your question
> 
> to answer your question this was the original setup:
> client wishing to use logger(usually GC allocated) -> Agora logger object(thread lifetime) -> Oceans logger(GC allocated)
> 
> we saw it from the ManagedDatabase case that the client allocated on the GC can outlive the thread destruction, so it would use an object(s) that can potentially be destroyed(Agora logger object and Oceans logger)

I do have to make a small adjustment: I need to create the Agora logger object on the GC heap too... If we have the Agora logger object tied to the thread, then its memory might already be freed and when we would try to use it from the cient it could cause SEGFAULT. However if we allocate Agora logger on the GC heap then although the GC might finalize the Agora logger(and so replace it with a default constructed Agora logger), but the memory will NOT be feed(so no SEGFAULT) until the client is finalized.

source/agora/common/TransactionPool.d
```diff
@@ -47,21 +47,13 @@ version (unittest)
     be initialized if `loggerCallback` was to be called immediately.
 
 *******************************************************************************/
+mixin AddLogger!();
```
Geod24:
```suggestion
*******************************************************************************/

mixin AddLogger!();
```

source/agora/utils/Log.d
```diff
@@ -49,21 +49,24 @@ public struct Logger
...
+                writeln("allowing logging from the destructor on a GC thread would risk running into segfaults, please see issue #1128");
+                return;
+            }
+            else if (logger is null)
+            {
+                assert(0,"\nplease make sure you are declaring the \nmixin AddLogger!(); statement on top, followed by:\nstatic this{}; followed by:\nstatic ~this{};");
+                return;
```
Geod24, line 67: Return is not needed if you use `assert(0)` or `throw new Exception`, the compiler does basic flow analysis

source/agora/utils/Log.d
```diff
@@ -49,21 +49,24 @@ public struct Logger
...
-            mixin("this.logger." ~ call ~ "(args);");
+            import core.memory : GC;
+            if (GC.inFinalizer())
+            {
+                writeln("allowing logging from the destructor on a GC thread would risk running into segfaults, please see issue #1128");
+                return;
+            }
+            else if (logger is null)
+            {
+                assert(0,"\nplease make sure you are declaring the \nmixin AddLogger!(); statement on top, followed by:\nstatic this{}; followed by:\nstatic ~this{};");
```
Geod24, line 66: Do we really want to crash here ? Might as well segv and get a core dump if we do. What's the rationale for using `assert(0)` ?

Geod24 (changes requested):
Few questions. Really not a fan of this workaround but let's go with it for the time being.

source/agora/utils/Log.d
ferencdg: https://github.com/bpfkorea/agora/issues/1128#issuecomment-676932589

this code corresponds to the point 1(where the static constructor/destructor is arranged in an unexpected order and causes SEGFAULTS at every startup consistently during executing the static section) -> this is the less serious issue compared to the other one. As you said I can just let it segfault, but I thought I would give a human readable error message. I can remove this code and just let it SEGFAULT.

The more serious issue which this PR resolves correspond to the case when we get SEGFAULTS by logging from the destructors. I plan to give more detail about that tomorrow.

ferencdg:
> Few questions. Really not a fan of this workaround but let's go with it for the time being.

I tried to come up with something that patches the potential segfaults without rewriting too much. If we had only 1 global logger potentially running on its separate receiving immutable log messages then both the lifetime issue of the logger and both the serialization of those immutable messages would be easy to solve.

Currently we can have as many loggers as modules and their lifetime is tied to the currently executing thread(well some part of the logging is tied, some not and that is the reason for the potential segfaults).

Geod24:
Just to be clear, my "not a fan" means that I think there *should* be a better way but I can't think of it and the fix is needed.
So once the small comments are addressed I still think we should move forward with this.

Geod24 (approved):

PR #1133, opened 2020-08-21
Update Vibe.d & co to fix DNS timeout bug

Geod24:
One issue we are seeing in CircleCI is the lack of timeout for DNS queries.
Since the containers don't all start at the same time, some DNS queries get fired
before the containers are online, but an answer is never returned and the nodes
just hang. With this fix to Vibe.d, this should work correctly.


AndrejMitrovic (approved):

AndrejMitrovic:
Thanks a lot for pushing this forward!

Geod24:
The first run still failed... But I think it was the client being slow. Didn't seem like the nodes were getting stuck this time, so good!

AndrejMitrovic:
Yeah. There's still a few more issues to resolve, but this was one of the obstacles.

PR #1140, opened 2020-08-25
Simplify the implementation of quorum regeneration

AndrejMitrovic:
This should be much better now. I ran into a bug and realized this can be simplified a lot.


Geod24 (approved):

linked0 (approved):
LGTM

PR #1143, opened 2020-08-25
Replace more makeChainedTransactions with TxBuilder

kchulj:
`makeChainedTransactions` has `key_pair` and `spend_amount` as parameters, but `TxBuilder` does not. It also does not allow invalid transactions to be created. As a result, some of the unit tests were not doing what they were supposed to do.

Below are some of the improvements:
1. Replace one transaction with a different amount, in order to get a different merkle root
2. Modify valid transactions by replacing and signing with a `foreach` loop
3. Delete unnecessary declaration of `Transaction` arrays

Improvement in `Base.d` developed with @AndrejMitrovic 

Relates #1107

### DIFF
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -584,9 +584,9 @@ unittest
         findUTXO, Enrollment.MinValidatorCount));
     const last_root = block.header.merkle_root;
 
-    // txs with a different amount
-    block = GenesisBlock.makeNewBlock(
-        makeChainedTransactions(gen_key, prev_txs, 1, 20_000_000));
+    block = GenesisBlock.makeNewBlock(prev_txs.enumerate.map!(en =>
+        TxBuilder(en.value).split(WK.Keys.byRange().take(en.index + 1).map!(k => k.address)).sign()));
+
     assert(block.isValid(GenesisBlock.header.height, GenesisBlock.header.hashFull(),
         findUTXO, Enrollment.MinValidatorCount));
 
@@ -615,8 +615,7 @@ unittest
     foreach (ref tx; GenesisBlock.txs)
         utxo_set.put(tx);
 
-    auto txs_1 = makeChainedTransactions(gen_key, null, 1,
-        400_000_000_000 * 8).sort.array;
+    auto txs_1 = genesisSpendable().map!(txb => txb.sign()).array();
 
     auto block1 = makeNewBlock(GenesisBlock, txs_1);
     assert(block1.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
@@ -737,8 +736,7 @@ unittest
     foreach (ref tx; GenesisBlock.txs)
         utxo_set.put(tx);
 
-    auto txs_1 = makeChainedTransactions(gen_key, null, 1,
-        400_000_000_000 * 10 * 8).sort.array;
+    auto txs_1 = genesisSpendable().map!(txb => txb.sign()).array();
 
     auto block1 = makeNewBlock(GenesisBlock, txs_1);
     assert(block1.isValid(GenesisBlock.header.height, gen_hash, findUTXO,
--- a/source/agora/test/NetworkManager.d
+++ b/source/agora/test/NetworkManager.d
@@ -19,6 +19,7 @@ import agora.api.Validator;
 import agora.common.BanManager;
 import agora.common.Config;
 import agora.common.crypto.Key;
+import agora.common.Hash;
 import agora.common.Metadata;
 import agora.common.Types;
 import agora.consensus.data.Block;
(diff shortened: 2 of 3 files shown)


source/agora/consensus/validation/Block.d
```diff
@@ -615,10 +622,26 @@ unittest
     foreach (ref tx; GenesisBlock.txs)
         utxo_set.put(tx);
 
-    auto txs_1 = makeChainedTransactions(gen_key, null, 1,
-        400_000_000_000 * 8).sort.array;
```
kchulj, line 619: Does the `spend_amount` here matter for this unittest?
If not, we can use the default amount and the loop can be deleted.

source/agora/test/NetworkManager.d
```diff
@@ -169,16 +197,28 @@ unittest
         last_txs = txs;
     }
 
-    // the validator node has 10 blocks, but bad node pretends to have 20
-    assert(node_validator.getBlockHeight() == 10, node_validator.getBlockHeight().to!string);
-    assert(node_bad.getBlockHeight() == 20);
+    // create genesis block
+    last_txs = genesisSpendable().map!(txb => txb.sign()).array();
+    last_txs.each!(tx => node_validator.putTransaction(tx));
+    network.expectBlock([node_validator], Height(1), 4.seconds);
```
AndrejMitrovic: The foreach loop before this code creates 9 blocks, so how come this test here passes?

source/agora/test/NetworkManager.d
```diff
@@ -169,16 +197,28 @@ unittest
         last_txs = txs;
     }
 
-    // the validator node has 10 blocks, but bad node pretends to have 20
-    assert(node_validator.getBlockHeight() == 10, node_validator.getBlockHeight().to!string);
-    assert(node_bad.getBlockHeight() == 20);
+    // create genesis block
```
AndrejMitrovic: it creates block height 1, not genesis

source/agora/consensus/validation/Block.d
AndrejMitrovic: It doesn't matter - this test is specifically testing freeze transactions as far as I can tell.

source/agora/test/NetworkManager.d
kchulj: It seems like some of my changes were not applied during the rebase.
Will check again.

source/agora/consensus/validation/Block.d
```diff
@@ -584,9 +584,16 @@ unittest
...
+    block = GenesisBlock.makeNewBlock(prev_txs.map!(tx => TxBuilder(tx).sign()));
+
+    // one tx with a different amount
+    block.txs[0].outputs[0].value = Amount(20_000_000 / 8);
+    block.txs[0].inputs[0].signature = gen_key.secret.sign(
+        hashFull(block.txs[0])[]);
+
+    block.txs.sort;
+    block.header.merkle_root = block.buildMerkleTree();
+
```
Geod24: I would say something as simple as:
```suggestion
    block = GenesisBlock.makeNewBlock(prev_txs.enumerate.map!(en => TxBuilder(en.value).split(WK.Keys.byRange().take(en.index)).sign()));
```

Would have done it, no ?

source/agora/consensus/validation/Block.d
```diff
@@ -615,10 +622,26 @@ unittest
...
+        foreach (idx2, txs2; block1.txs[idx1].outputs)
+            block1.txs[idx1].outputs[idx2].value = Amount(400_000_000_000);
+    }
+
+    foreach (idx1, txs1; block1.txs)
+    {
+        foreach (idx2, txs2; block1.txs[idx1].inputs)
+            block1.txs[idx1].inputs[idx2].signature = gen_key.secret.sign(
+                hashFull(block1.txs[idx1])[]);
+    }
```
linked0: There are some variables unused and some code is kind of redundant. The code could be simpler like as follows.

```suggestion
    foreach (tx; block1.txs)
    {
        iota(tx.outputs.length).map!(
            idx => tx.outputs[idx].value = Amount(400_000_000_000));
        iota(tx.inputs.length).map!(
            idx => tx.inputs[idx].signature = gen_key.secret.sign(hashFull(tx)[]));
    }
```

kchulj: Thank you for your suggestion! I will keep it in mind for future `foreach` loops.
However, this part will be deleted because the `Amount` does not matter for this test.

source/agora/consensus/validation/Block.d
```diff
@@ -615,10 +615,10 @@ unittest
     foreach (ref tx; GenesisBlock.txs)
         utxo_set.put(tx);
 
-    auto txs_1 = makeChainedTransactions(gen_key, null, 1,
-        400_000_000_000 * 8).sort.array;
+    auto txs_1 = genesisSpendable().map!(txb => txb.sign()).array();
 
     auto block1 = makeNewBlock(GenesisBlock, txs_1);
+
```
AndrejMitrovic: Unrelated change.

Otherwise LGTM, I'll merge after you fix this. :)

source/agora/test/NetworkManager.d
```diff
@@ -75,22 +76,44 @@ unittest
...
-                auto block = makeNewBlock(last_block, txs);
+                block = makeNewBlock(last_block, txs);
 
                 blocks ~= block;
                 last_block = block;
             }
 
+            auto signTx (Transaction tx) @trusted { return prev_key.secret.sign(hashFull(tx)[]); }
+
+            foreach (idx1, txs1; blocks)
```
linked0: Some more things, plz. You might also have to modify the related code like `foreach (tx; blocks[idx1].txs)`
```suggestion
            foreach (ref block; blocks)
```

source/agora/test/NetworkManager.d
```diff
@@ -75,22 +76,44 @@ unittest
...
+                foreach (tx; blocks[idx1].txs)
+                {
+                    iota(tx.outputs.length).map!(
+                        idx => tx.outputs[idx].address = prev_key.address);
+                    iota(tx.inputs.length).map!(
+                        idx => tx.inputs[idx].signature = signTx(blocks[idx1].txs[idx]));
+                }
+            }
+
+            iota(blocks.length).map!(idx => blocks[idx].header.merkle_root = blocks[idx].buildMerkleTree());
```
linked0: Some more things, plz.
```suggestion
            blocks.map!(block => block.header.merkle_root = block.buildMerkleTree());
```

source/agora/test/NetworkManager.d
```diff
@@ -75,22 +76,43 @@ unittest
...
+            auto signTx (Transaction tx) @trusted { return prev_key.secret.sign(hashFull(tx)[]); }
+
+            foreach (ref block1; blocks)
+            {
+                foreach (ref tx; block1.txs)
+                {
+                    iota(tx.outputs.length).map!(
+                        idx => tx.outputs[idx].address = prev_key.address);
+                    iota(tx.inputs.length).map!(
+                        idx => tx.inputs[idx].signature = signTx(block1.txs[idx]));
```
AndrejMitrovic: Range are lazy. This doesn't actually do anything.

source/agora/test/NetworkManager.d
```diff
@@ -75,22 +76,43 @@ unittest
...
+                }
+            }
+            blocks.map!(block => block.header.merkle_root = block.buildMerkleTree());
```
AndrejMitrovic: Same comment, this is a no-op.

source/agora/test/NetworkManager.d
Geod24: @linked0 : `map` should not be used to modify the input. As @AndrejMitrovic pointed out, this doesn't do anything because `map` does not iterate by itself.

AndrejMitrovic (approved):

kchulj:
Ready for review again. Thanks everyone for the suggestions.
The embedded `foreach` is currently there to imitate `makeChainedTransactions`.
However, as @AndrejMitrovic said, this loop is trivial and can be replaced with returning `block.init`, because we just need invalid blocks.
Let me know what you guys think.

linked0 (approved):
LGTM and sorry for the confusion about my suggestions.
Anyway, thanks for all the efforts.

Geod24 (approved):

PR #1144, opened 2020-08-25
Disable ManyValidators test as it frequently fails

AndrejMitrovic:
`dchatty=1` revealed this is the test that's consistently causing Github CI failures.

The test is likely running into a performance problem. With 16 nodes the amount of messages being sent throughout the network seems to be quite large - it's something we need to fix. It would be good to take a look at how Stellar implements the `emitEnvelope` routine.

### DIFF
--- a/source/agora/test/ManyValidators.d
+++ b/source/agora/test/ManyValidators.d
@@ -13,7 +13,9 @@
 
 module agora.test.ManyValidators;
 
-version (unittest):
+// temporarily disabled until failures are resolved
+// see #1145
+version (none):
 
 import agora.api.Validator;
 import agora.common.Amount;


AndrejMitrovic:
This is something I suspected for a while: https://github.com/stellar/stellar-core/blob/9500dc1d5e1182ca88694f1a226de7d8eec87941/src/overlay/Floodgate.cpp#L84

ferencdg:
Do we know with what error message it fails? I remember on Debian, I had to increase the GC's max memory for the unittests with "--DRT-gcopt=maxPoolSize:YYY" otherwise they didn't finish

AndrejMitrovic:
It's a timeout in the `expectBlock` call. That means the nodes are very busy handling requests. And we don't have any sort of request priority system in place in `-unittest` mode (LocalRest).

Geod24:
That's like, our most complete test

source/agora/test/ManyValidators.d
```diff
@@ -13,7 +13,8 @@
 
 module agora.test.ManyValidators;
 
-version (unittest):
+// temporarily disabled until failures are resolved
+version (none):
```
Geod24: Create an issue plz

AndrejMitrovic:
Both this PR and https://github.com/bpfkorea/agora/pull/1100 went green with this commit.

I've opened https://github.com/bpfkorea/agora/issues/1145.

Geod24:
> I've opened #1145.

The idea was to link it in the code too

linked0 (approved):

PR #1147, opened 2020-08-26
Begin working on a new schnorr scheme

AndrejMitrovic:


AndrejMitrovic:
This is an idea to replace R in the challenge, which was previously the sum of R for all the nodes. If we replace it with the signer's own R then we can arbitrarily add new signatures without having to predetermine the N/M ratio of signatures.

And because the signers have already pre-commited to R (thanks to our preimage scheme), the signers cannot manipulate their own R and therefore signatures cannot be forged.

It would be good to look hard and try to break this scheme. But if it works, it could open up a lot of possibilities and would simplify signature collection drastically.

AndrejMitrovic:
I pushed another commit that removes `R` from the hash (the commitment / message).

I'll ask about the security of this scheme on stackexchange soon.

AndrejMitrovic:
https://github.com/bpfkorea/agora/pull/1162 is a better example, it includes a commit with the "multi-sig" example.

PR #1150, opened 2020-08-27
[RFC] Implement a per-node request queue for gossiping

### LINKED ISSUE #1149: Implement a per-node request queue for gossiping
https://github.com/bpfkorea/agora/issues/1145 revealed that there are numerous fibers spawned for every single gossiping requests.

That means in the example of 10 nodes gossiping 8 transactions would lead to the creation of 80 fibers.

The `ManyValidators` test would spawn over 1600 fibers over the course of the run of the test.

AndrejMitrovic:
Also includes a bugfix.

This does not solve the issue of having too many requests to begin with. Our gossiping in the network is quadratic. Given N nodes there will be N to the power of 2 total messages propagated for every gossiping message. We need to make this smarter.

Fixes #1149

### DIFF
--- a/source/agora/common/Task.d
+++ b/source/agora/common/Task.d
@@ -33,6 +33,9 @@ public enum Periodic : bool
 /// Exposes primitives to run tasks through Vibe.d
 public class TaskManager
 {
+    /// stats: total number of task started
+    protected ulong tasks_started;
+
     /***************************************************************************
 
         Run an asynchronous task in vibe.d's event loop
@@ -45,6 +48,7 @@ public class TaskManager
     public void runTask (void delegate() dg) nothrow
     {
         static import vibe.core.core;
+        this.tasks_started++;
         vibe.core.core.runTask(dg);
     }
 
@@ -92,9 +96,21 @@ public class TaskManager
     public ITimer setTimer (Duration timeout, void delegate() dg,
         Periodic periodic = Periodic.No) nothrow
     {
+        this.tasks_started++;
         assert(dg !is null, "Cannot call this delegate if null");
         static import vibe.core.core;
         return new VibedTimer(vibe.core.core.setTimer(timeout, dg, periodic));
+    }
+
+    /***************************************************************************
+
+        Log out the request stats
+
+    ***************************************************************************/
+
+    public void logStats () @safe nothrow
+    {
+        log.info("Tasks started: {}", this.tasks_started);
     }
 }
 
--- a/source/agora/network/NetworkClient.d
+++ b/source/agora/network/NetworkClient.d
@@ -204,7 +204,7 @@ class NetworkClient
         this.taskman.runTask(
         {
             // if the node already has this tx, don't send it
-            if (this.attemptRequest!(API.hasTransactionHash, Throw.Yes,
+            if (this.attemptRequest!(API.hasTransactionHash, Throw.No,
                 LogLevel.Trace)(this.api, tx_hash))
                 return;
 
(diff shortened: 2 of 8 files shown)


source/agora/common/Task.d
```diff
@@ -92,10 +96,22 @@ public class TaskManager
...
         return new VibedTimer(vibe.core.core.setTimer(timeout, dg, periodic));
     }
+
+    /***************************************************************************
+
+        Log out the request stats
+
+    ***************************************************************************/
+
+    public void logStats ()
```
AndrejMitrovic: Missing attributes.

source/agora/network/NetworkManager.d
```diff
@@ -347,6 +347,9 @@ public class NetworkManager
     /// See 'minPeersConnected'
     private Set!PublicKey required_peer_keys;
 
+    /// Ditto
```
AndrejMitrovic: This ditto is wrong, this should be documented properly.

source/agora/network/NetworkClient.d
```diff
@@ -201,15 +201,12 @@ class NetworkClient
         import agora.common.Hash;
         const tx_hash = tx.hashFull();
 
-        this.taskman.runTask(
-        {
-            // if the node already has this tx, don't send it
-            if (this.attemptRequest!(API.hasTransactionHash, Throw.No,
-                LogLevel.Trace)(this.api, tx_hash))
-                return;
+        // if the node already has this tx, don't send it
```
AndrejMitrovic, line 204: Should consider squashing this with the code that introduces the RequestQueue, otherwise this commit might break some assumptions.

source/agora/network/NetworkManager.d
```diff
@@ -58,6 +58,127 @@ import core.time;
 
 mixin AddLogger!();
 
+///
```
AndrejMitrovic: Needs documentation.

source/agora/network/NetworkManager.d
```diff
@@ -58,6 +58,127 @@ import core.time;
...
+
+    ***************************************************************************/
+
+    private void run () @safe nothrow
+    {
+        this.is_running = true;
+
+        while (this.is_running)
+        {
+            // if banned sleep this fiber until the scheduled unban time
```
AndrejMitrovic:
```suggestion
            // if banned then sleep this fiber until the scheduled unban time
```

AndrejMitrovic:
Well the `ManyNodes` test still fails - so I removed that commit. I think the reason why it fails is because there are still too many requests being sent. Our gossiping has quadratic cost, for N nodes there will be N to the power of 2 total messages propagated for every unique gossiping message.

But at least with the new code we don't spawn a thousand fibers anymore.

AndrejMitrovic:
Hmm.. this only solves the problem of too many requests sent to other nodes. It's a good problem to solve, but there's one more thing to fix.

It doesn't solve the problem of incoming requests - in particular only for LocalRest. Whenever another node contacts this node LocalRest will spawn a new fiber. I think this can be optimized.

But then I think I just ran into the issue of the lack of channels in LocalRest..

AndrejMitrovic:
My idea worked!

The Github CI is now very green.

I'm going to submit the LocalRest improvements soon.

AndrejMitrovic:
> I'm going to submit the LocalRest improvements soon.

Haha, I ran into the classical problem in concurrency - locking. N1 -> N2 -> N3 -> N1.

I'll have to rethink the idea.

AndrejMitrovic:
This needs to be solved on the lower-level by implementing fiber pools in LocalRest. I'll split off the non-controversial commits into a separate PR.

PR #1152, opened 2020-08-28
Preventing multiple Node instance from using using the same cache directory.

### LINKED ISSUE #829: Implement a lock on the caching directory
It is way too easy to start two instances of the node in the same working directory and accidentally corrupt data. And someone could hypothetically spend a couple hours investigating why.

ferencdg:
This change was necessary so separate agora instances will not corrupt
each other's cache.

fixes #829

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -64,6 +64,14 @@ public struct CommandLine
     }
 }
 
+/// Thrown when there is a configuration error
+class ConfigException : Exception
+{
+    this(string msg, string file = __FILE__, size_t line = __LINE__) {
+        super(msg, file, line);
+    }
+}
+
 /// Main config
 public struct Config
 {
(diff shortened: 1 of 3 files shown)


source/agora/common/Config.d
```diff
@@ -225,16 +233,54 @@ public GetoptResult parseCommandLine (ref CommandLine cmdline, string[] args)
 
 public Config parseConfigFile (ref const CommandLine cmdln)
 {
```
ferencdg, line 235: let me fix this wierd formatting issues

source/agora/common/Config.d
```diff
@@ -64,6 +64,14 @@ public struct CommandLine
...
 }
 
+/// Thrown when there is a configuration error
+class ConfigException : Exception
+{
+    this(string msg, string file = __FILE__, size_t line = __LINE__) {
+        super(msg, file, line);
+    }
+}
+
```
Geod24, line 74: None of the changes in this file seems to be relevant. It seems you are introducing `ConfigException` only to transform it into `Exception` soon after. I can't see any special handling of `ConfigException` vs `Exception`.

source/agora/common/FileBasedLock.d
```diff
@@ -0,0 +1,160 @@
...
+
+import core.thread.osthread: getpid;
+
+/*******************************************************************************
+
+    Implements file based locking that can help synchronize processes
+
+*******************************************************************************/
+
+struct FileBasedLock
```
Geod24:
```suggestion
public struct FileBasedLock
```

source/agora/common/FileBasedLock.d
```diff
@@ -0,0 +1,160 @@
...
+{
+
+    ///
+    private string lock_file_name;
```
Geod24:
```suggestion
    ///
    private string file_name;
```

Since the aggregate contains `Lock` already, no need to repeat the name.

Geod24 (changes requested):
The issue is only about the `.cache` directory. I don't think we need to restrict the node to 1 process, which this seems to be doing ?

source/agora/common/Config.d
ferencdg: I transform Exception to ConfigException and catch ConfigException from the main

Geod24: But... Why not just catch `Exception` as before ?

ferencdg: Because that would catch the Exception thrown from the file locker, and also the current catch block in the main say the exception is related to 'config' issues

ferencdg: if you would prefer I could do this:
1. leave the ConfigException as it is in the main
2. throw a specific LockException and catch that in the main too, so I can handle it separately and also print out some user friendly error message

every other exception should propage through main(although there is none current as the runTask doesn't leak Exceptions)

Geod24: Well, there's a few ways around it:
- We expose `lock` and `lockThrow`. Why ? We could just have a boolean and avoid the `throw`ing part;
- We don't need to modify the config, we can just add a specific exception to `lock` and catch it first;
- Change the message in a way that does not make assumption, *or* break the `try` block up;

Actually the later sounds like the best idea. Let me give it a quick try.

ferencdg: "We expose lock and lockThrow. Why ?
we have 3 public functions
lockTry, lockBlock, lockThrow we don't expose the lock function, I think it is the lockTry that return a boolean. 

"We don't need to modify the config, we can just add a specific exception to lock and catch it first;"
I think it is a bit messy, it is not guaranteed that every Exception is a ConfigException(in the future)

How about my idea of having specific LockException and LockConfig: we could just create a simple mixin and use to define those exception classes like:
1. mixin CustomException!(LockException);
2. mixin CustomException!(ConfigException);

ferencdg: should I do 
'Change the message in a way that does not make assumption, or break the try block up;'?

Geod24: > Change the message in a way that does not make assumption, or break the try block up;

Did the latter [here](https://github.com/bpfkorea/agora/pull/1153).

> I think it is a bit messy, it is not guaranteed that every Exception is a ConfigException(in the future)

Agreed, that's why IMO reducing the scope to just `parseConfigFile` would be better.

ferencdg:
I will create a new branch with a new pull request, this branch is a mess by now

PR #1156, opened 2020-08-31
Add randomness to leader selection

### LINKED ISSUE #1079: Leader selection needs to be improved
Currently the leader selection in SCP is based entirely on the weight of the node:

https://github.com/bpfkorea/agora/blob/1af3731d063a8b52f1e114a4b31e5ac65ad70869/source/scpp/src/scp/NominationProtocol.cpp#L210
https://github.com/bpfkorea/agora/blob/1af3731d063a8b52f1e114a4b31e5ac65ad70869/source/scpp/src/scp/NominationProtocol.cpp#L265
https://github.com/bpfkorea/agora/blob/1af3731d063a8b52f1e114a4b31e5ac65ad70869/source/scpp/src/scp/LocalNode.cpp#L97

The current algorithm makes the leader selection too predictable. This is especially true in our case because all of our validators are registered and known.

We should take our source of randomness into account when doing leader selection for each round. 

And as a slight optimization we could replace the `getNodeWeight` so it doesn't have to walk through any quorum sets. We know how many times each node appears in a quorum.

-----

Side-note: the reason this is not an issue in Stellar is two-fold: 
1) the validator set is not known and can change at any moment (no registration necessary)
2) Stellar is trust-based, which partially justifies having a predictable leader selection algorithm

linked0:
By adding the `random seed` for all the validators in calculating the hash value of nodes, we can add additional randomness when doing leader selection for each round. This PR has the implementation of that algorithm.

Fixes #1079

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -17,8 +17,9 @@ import agora.common.crypto.ECC;
 import agora.common.crypto.Key;
 import agora.common.crypto.Schnorr;
 import agora.common.Config;
-import agora.common.Hash : Hash, HashDg, hashPart, hashFull;
+import agora.common.Hash : Hash, HashDg, hashPart, hashFull, hashMulti;
 import agora.common.ManagedDatabase;
+import agora.common.SCPHash;
 import agora.common.Serializer;
 import agora.common.Set;
 import agora.common.Task;
@@ -730,6 +731,52 @@ extern(D):
             return;
         this.active_timers[type] = this.taskman.setTimer(
             timeout.msecs, { callCPPDelegate(callback); });
+    }
+
+    /***************************************************************************
+
+        Used by the nomination protocol to randomize the order of messages
+        between nodes.
+
+        Params:
+            slot_idx = the slot index we're currently reaching consensus for.
+            prev = the previous data set for the provided slot index.
+            is_priority = the flag to check that this call is for priority.
+            round_num = the nomination round
+            node_id = the id of the node for which this computation is being made
+
+        Returns:
+            the 8-byte hash
+
+    ***************************************************************************/
+
+    public override uint64_t computeHashNode (uint64_t slot_idx,
+        ref const(Value) prev, bool is_priority, int32_t round_num,
+        ref const(NodeID) node_id) nothrow
+    {
+        const uint hash_N = 1;
+        const uint hash_P = 2;
+
+        const seed = this.ledger.getValidatorRandomSeed(Height(slot_idx - 1));
+        uint512 hash;
+        try
+        {
+            hash = uint512(hashMulti(slot_idx, prev[],
+                is_priority ? hash_P : hash_N, round_num, node_id, seed));
+        }
+        catch (Exception ex)
+        {
+            log.fatal("Computing hash of the node({}) failed: {}, Data was: " ~
+                "slot_idx: {}, prev: {}, is_priority: {}, seed: {}",
+                node_id, ex.msg, slot_idx, prev, is_priority, seed);
+            assert(0);
+        }
+
+        uint64_t res = 0;
+        for (size_t i = 0; i < res.sizeof; i++)
+            res = (res << 8) | hash[][i];
+
+        return res;
     }
 }
 
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -518,6 +518,31 @@ public class Ledger
             return null;
         return block.getMerklePath(index);
     }
+
+    /***************************************************************************
+
+        Generate the random seed reduced from the preimages for the provided
+        block height.
+
+        Params:
+            height = the desired block height to look up the images for
+
+        Returns:
+            the random seed
+
+    ***************************************************************************/
+
+    public Hash getValidatorRandomSeed (Height height) nothrow
+    {
+        Hash[] keys;
+        if (!this.enroll_man.getEnrolledUTXOs(keys) || keys.length == 0)
+        {
+            log.fatal("Could not retrieve enrollments / no enrollments found");
+            assert(0);
+        }
+
+        return this.enroll_man.getRandomSeed(keys, height);
+    }
 }
 
 /// Note: these unittests historically assume a block always contains
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -566,6 +566,42 @@ public class TestAPIManager
 
     /***************************************************************************
 
+        Checks the needed pre-images are revealed, sets the clock time to the
+        expected clock time to produce a block at the given height, and verifies
+        that the nodes have generated a block at the given block height.
+
+        The overload allows passing a subset of nodes to verify the block
+        heights for only these nodes. Note that the clock time is adjusted
+        for all nodes (this is what most tests expect).
+
+        Params:
+            height = the expected block height
+            enroll_header = the header which contains enrollment information
+            timeout = the request timeout to each node
+
+    ***************************************************************************/
+
+    public void expectBlock (Height height, const(BlockHeader) enroll_header,
+        Duration timeout, string file = __FILE__, int line = __LINE__)
+    {
+        this.expectBlock(this.clients, height, enroll_header, timeout, file,
+            line);
+    }
+
+    /// Ditto
+    public void expectBlock (Clients)(Clients clients, Height height,
+        const(BlockHeader) enroll_header, Duration timeout,
+        string file = __FILE__, int line = __LINE__) if (isInputRange!Clients)
+    {
+        assert(height > enroll_header.height);
+        auto distance = cast(ushort)(height - enroll_header.height - 1);
+        waitForPreimages(clients, enroll_header.enrollments, distance,
+            timeout);
+        this.expectBlock(clients, height, timeout, file, line);
+    }
+
+    /***************************************************************************
+
         Checks if all the nodes contain the given distance of pre-images for
         the given enrollments.
 
--- a/source/agora/test/EnrollmentManager.d
+++ b/source/agora/test/EnrollmentManager.d
@@ -165,7 +165,7 @@ unittest
                 .map!(ptx => TxBuilder(ptx).refund(WK.Keys[count].address).sign())
                 .array();
         txs.each!(tx => nodes[1].putTransaction(tx));
-        network.expectBlock(Height(count + 1), 3.seconds);
+        network.expectBlock(Height(count + 1), b0.header, 3.seconds);
     }
 
     // Now create an Enrollment for nodes[0], create block #20, and restart nodes[0]
@@ -175,7 +175,7 @@ unittest
         .map!(ptx => TxBuilder(ptx).refund(WK.Keys[20].address).sign())
         .array();
     txs.each!(tx => nodes[1].putTransaction(tx));
-    network.expectBlock(nodes.take(1), Height(validator_cycle), 3.seconds);
+    network.expectBlock(nodes.take(1), Height(validator_cycle), b0.header, 3.seconds);
     retryFor(nodes[0].getBlocksFrom(20, 1)[0].header.enrollments.length == 1,
         2.seconds);
 
@@ -188,7 +188,8 @@ unittest
         .map!(ptx => TxBuilder(ptx).refund(WK.Keys[21].address).sign())
         .array();
     txs.each!(tx => nodes[1].putTransaction(tx));
-    network.expectBlock(nodes.take(1), Height(21), 3.seconds);
+    const b20 = nodes[0].getBlocksFrom(20, 2)[0];
+    network.expectBlock(nodes.take(1), Height(21), b20.header, 3.seconds);
 
     PreImageInfo org_preimage = PreImageInfo(enroll.utxo_key, enroll.random_seed, 0);
 
@@ -400,6 +401,7 @@ unittest
         .map!(en => en.value.refund(WK.Keys.A.address).sign())
         .array();
     txs.each!(tx => validator.putTransaction(tx));
+    network.waitForPreimages(b0.header.enrollments, 1, 2.seconds);
     network.setTimeFor(Height(1));  // trigger consensus
 
     // Make sure that the code in validator gets executed
--- a/source/agora/test/Ledger.d
+++ b/source/agora/test/Ledger.d
@@ -57,7 +57,7 @@ unittest
 
         // send it to one node
         txs.each!(tx => node_1.putTransaction(tx));
-        network.expectBlock(Height(block_idx + 1), 4.seconds);
+        network.expectBlock(Height(block_idx + 1), blocks[0].header, 4.seconds);
 
         blocks ~= node_1.getBlocksFrom(block_idx + 1, 1);
         block_txes ~= txs.sort.array;
@@ -185,6 +185,10 @@ unittest
 
     auto txs = network.blocks[0].spendable.map!(txb => txb.sign()).array();
     txs.each!(tx => node_1.putTransaction(tx));
+
+    // wait for preimages to be revealed before making blocks
+    network.waitForPreimages(network.blocks[0].header.enrollments, 6, 5.seconds);
+
     network.expectBlock(Height(1), 3.seconds);
 
     txs = txs.map!(tx => TxBuilder(tx).sign()).array();
--- a/source/agora/test/NetworkManager.d
+++ b/source/agora/test/NetworkManager.d
@@ -54,7 +54,8 @@ unittest
 
     nodes[0].clearFilter();
     nodes[1].clearFilter();
-    network.expectBlock(Height(1), 2.seconds);
+    const b0 = nodes[0].getBlocksFrom(0, 2)[0];
+    network.expectBlock(Height(1), b0.header, 2.seconds);
 }
 
 /// test behavior when a node sends bad block data
@@ -170,6 +171,10 @@ unittest
     auto node_validators = nodes[0 .. 4];  // validators, create blocks
     auto node_test = nodes[4];  // full node, does not create blocks
     auto node_bad = nodes[5];  // full node, returns bad blocks in getBlocksFrom()
+
+    // wait for preimages to be revealed before making blocks
+    network.waitForPreimages(network.blocks[0].header.enrollments, 6,
+        5.seconds);
 
     // enable filtering first
     node_validators.each!(node => node.filter!(API.getBlocksFrom));
--- a/source/agora/test/Quorum.d
+++ b/source/agora/test/Quorum.d
@@ -52,7 +52,8 @@ unittest
     network.waitForDiscovery();
 
     auto nodes = network.clients;
-    network.expectBlock(Height(8), 5.seconds);
+    const b0 = nodes[0].getBlocksFrom(0, 2)[0];
+    network.expectBlock(Height(8), b0.header,  5.seconds);
 
     // create block with 6 payment and 2 freeze tx's
     auto txs = network.blocks[$ - 1].spendable().map!(txb => txb.sign()).array();
@@ -74,7 +75,7 @@ unittest
     txs.each!(tx => nodes[0].putTransaction(tx));
 
     // at block height 9 the freeze tx's are available
-    network.expectBlock(Height(9), 5.seconds);
+    network.expectBlock(Height(9), b0.header, 5.seconds);
 
     // now we can create enrollments
     Enrollment enroll_0 = nodes[$ - 2].createEnrollmentData();
@@ -99,7 +100,7 @@ unittest
     new_txs.each!(tx => nodes[0].putTransaction(tx));
 
     // at block height 10 the validator set has changed
-    network.expectBlock(Height(10), 3.seconds);
+    network.expectBlock(Height(10), b0.header, 3.seconds);
 
     //// these are un-enrolled now
     nodes[0 .. $ - 2].each!(node => node.sleep(10.minutes, true));
@@ -108,11 +109,12 @@ unittest
     txs = new_txs.map!(tx => TxBuilder(tx).sign()).array();
     txs.each!(tx => nodes[$ - 2].putTransaction(tx));
 
-    network.expectBlock(nodes[$ - 2 .. $], Height(11), 3.seconds);
+    const b10 = nodes[$ - 2].getBlocksFrom(10, 2)[0];
+    network.expectBlock(nodes[$ - 2 .. $], Height(11), b10.header, 3.seconds);
 
     // force wake up
     nodes[0 .. $ - 2].each!(node => node.sleep(0.seconds, false));
 
     // all nodes should have same block height now
-    network.expectBlock(Height(11), 10.seconds);
+    network.expectBlock(Height(11), b10.header, 10.seconds);
 }
--- a/source/agora/test/QuorumPreimage.d
+++ b/source/agora/test/QuorumPreimage.d
@@ -52,7 +52,8 @@ unittest
     network.waitForDiscovery();
 
     auto nodes = network.clients;
-    network.expectBlock(Height(8), 10.seconds);
+    const b0 = nodes[0].getBlocksFrom(0, 2)[0];
+    network.expectBlock(Height(8), b0.header, 10.seconds);
 
     enum quorums_1 = [
         // 0
@@ -146,7 +147,7 @@ unittest
     txs.each!(tx => nodes[0].putTransaction(tx));
 
     // at block height 9 the freeze txs are available
-    network.expectBlock(Height(9), 10.seconds);
+    network.expectBlock(Height(9), b0.header, 10.seconds);
 
     // now we re-enroll existing validators (extension),
     // and enroll 2 new validators.
@@ -175,7 +176,7 @@ unittest
     makeBlock();
 
     // at block height 10 the validator set has changed
-    network.expectBlock(Height(10), 3.seconds);
+    network.expectBlock(Height(10), b0.header, 3.seconds);
 
     // check if the needed pre-images are revealed timely
     enrolls.each!(enroll =>
@@ -276,12 +277,13 @@ unittest
                 idx, node.getQuorumConfig(), quorums_2[idx])));
 
     // create 9 blocks (1 short of all enrollments expiring)
+    const b10 = nodes[0].getBlocksFrom(10, 2)[0];
     foreach (idx; 0 .. 9)
     {
         makeBlock();
 
         // at block height 10 the validator set has changed
-        network.expectBlock(Height(10 + idx + 1), 3.seconds);
+        network.expectBlock(Height(10 + idx + 1), b10.header, 3.seconds);
     }
 
     // re-enroll all validators before they expire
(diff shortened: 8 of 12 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -577,9 +577,8 @@ public class EnrollmentManager
         foreach (const ref key; keys)
         {
             const preimage = this.validator_set.getPreimageAt(key, height);
-            // this should not happen. validators which didn't reveal the
-            // preimage should not be in the active validator set 'keys'.
-            assert(preimage != PreImageInfo.init);
```
linked0: > assert(preimage != PreImageInfo.init);

This should be restored, which will be dealt with in another PR. Currently, this makes many assertion failures due to timing issues which means that the function is called before the right pre-images are not revealed in many test cases.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -556,6 +563,61 @@ extern(D):
...
+            node_id = the id of the node for which this computation is being made
+
+    ***************************************************************************/
+
+    public override uint64_t computeHashNode(uint64_t slot_idx,
+        ref const(Value) prev, bool is_priority, int32_t round_number,
+        ref const(NodeID) node_id) nothrow
+    {
+        uint512 hash;
+        const random_seed = this.ledger.getValidatorRandomSeed(Height(slot_idx));
```
AndrejMitrovic: You're supposed to use the previous height, not the current one..

AndrejMitrovic: Lets consider the case where block height 1 was just externalized. We will have at least the preimages for height 1, but there is no guarantee we have preimages for heights 2+.

So when we start nominating block height 2, we can only take the most recent guaranteed preimages - the ones for height 1 

So you just have to do `slot_idx - 1` in this call.

linked0: > So when we start nominating block height 2, we can only take the most recent guaranteed preimages - the ones for height 1

I got it. Thanks.

source/agora/common/SCPHash.d
```diff
@@ -60,3 +61,11 @@ public uint512 getHashOf (uint64_t slot_idx, ref const(Value) prev, uint32_t has
 {
     return uint512(hashMulti(slot_idx, prev[], hash, round_num, value[]));
 }
+
+/// Ditto
+public uint512 getHashOf (uint64_t slot_idx, ref const(Value) prev, uint32_t hash,
+    int32_t round_num, ref const(NodeID) node_id, ref const(AgoraHash) random_seed)
+{
+    return uint512(hashMulti(slot_idx, prev[], hash, round_num, node_id,
```
AndrejMitrovic: There is no need for this function. You can move this entire call inline into `computeHashNode`

source/agora/consensus/EnrollmentManager.d
```diff
@@ -577,9 +577,8 @@ public class EnrollmentManager
         foreach (const ref key; keys)
         {
             const preimage = this.validator_set.getPreimageAt(key, height);
-            // this should not happen. validators which didn't reveal the
```
AndrejMitrovic: Please revert these changes, this assert is a preventative measure so the function is not called with the wrong height.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -556,6 +563,61 @@ extern(D):
...
+        Params:
+            slot_idx = the slot index we're currently reaching consensus for.
+            prev = the previous data set for the provided slot index.
+            is_priority = the flag to check that this call is for priority.
+            round_number = the nomination round
+            node_id = the id of the node for which this computation is being made
+
+    ***************************************************************************/
+
+    public override uint64_t computeHashNode(uint64_t slot_idx,
```
TrustHenry: No document definition for the return value. `uint64_t`

source/agora/consensus/protocol/Nominator.d
```diff
@@ -556,6 +563,61 @@ extern(D):
...
+    /***************************************************************************
+
+        Convert the `uint512` type of hash value into the `uint64_t` type
+
+        Params:
+            hash = the hash value of `uint512` type.
+
+    ***************************************************************************/
+
+    static private uint64_t hashHelper(ref const(uint512) hash) nothrow
```
TrustHenry: No document definition for the return value. `uint64_t`

source/agora/node/Ledger.d
```diff
@@ -507,6 +507,31 @@ public class Ledger
...
+
+        Params:
+            height = the desired block height to look up the images for
+
+        Returns:
+            the random seed
+
+    ***************************************************************************/
+
+    public Hash getValidatorRandomSeed (Height height) nothrow
```
AndrejMitrovic, line 535: This is awkward code, it does not belong here.

There is nothing this function uses in the Ledger itself other than the enrollment manager. 

Why not move `this.enroll_man.getEnrolledUTXOs(keys)` inside of `EnrollmentManager.getRandomSeed` and remove the `in Hash[] keys` parameter. And then also provide the `EnrollmentManager` instance to the `Nominator` and make it call `getRandomSeed(slot_idx - 1)`?

source/agora/consensus/protocol/Nominator.d
```suggestion
    private static uint64_t hashHelper (ref const(uint512) hash) nothrow
```

`private` goes first. And the `(` styling fix.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -556,6 +563,61 @@ extern(D):
...
+
+        return hashHelper(hash);
+    }
+
+    /***************************************************************************
+
+        Convert the `uint512` type of hash value into the `uint64_t` type
+
+        Params:
+            hash = the hash value of `uint512` type.
```
AndrejMitrovic: For very simple functions we can just use a `Params` and `Returns` section if the function returns something.

```suggestion
        Params:
            hash = the hash value of `uint512` type.

        Returns:
            An 8-byte `uint64_t` value reduced from the 64-byte `uint512` hash value
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -556,6 +563,61 @@ extern(D):
...
+        try
+        {
+            hash = getHashOf(slot_idx, prev, is_priority ? hash_P : hash_N,
+                round_number, node_id, random_seed);
+        }
+        catch (Exception ex)
+        {
+            log.fatal("Computing hash of the node({}) failed: {}, Data was: ",
+                node_id, ex.msg);
+            log.fatal("\tslot_idx: {}, prev: {}, is_priority: {}, randon_seed: {}",
```
```suggestion
            log.fatal("\tslot_idx: {}, prev: {}, is_priority: {}, random_seed: {}",
```

Why isn't this all in one log call? You can add `\n` if you really want to add a newline.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -75,6 +76,12 @@ public extern (C++) class Nominator : SCPDriver
     /// Whether we're in the process of nominating
     private bool is_nominating;
 
+    // values used to switch hash function between priority and neighborhood checks
+    // which are from SCPDriver.cpp
+    const uint hash_N = 1;
+    const uint hash_P = 2;
+    const uint hash_K = 3;
```
AndrejMitrovic: Please move these inside of `computeHashNode`. Constants should only be in an outer scope if they're used by other code too, but they are not used anywhere else.

source/agora/consensus/protocol/Nominator.d
```suggestion
    public override uint64_t computeHashNode (uint64_t slot_idx,
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -556,6 +563,61 @@ extern(D):
...
+        catch (Exception ex)
+        {
+            log.fatal("Computing hash of the node({}) failed: {}, Data was: ",
+                node_id, ex.msg);
+            log.fatal("\tslot_idx: {}, prev: {}, is_priority: {}, randon_seed: {}",
+                slot_idx, prev, is_priority, random_seed);
+            assert(0);
+        }
+
+        return hashHelper(hash);
```
AndrejMitrovic: Copy-paste the implementation of `hashHelper` inline here, because it's not used anywhere else.

AndrejMitrovic (changes requested):
Few changes requested.

source/agora/node/Ledger.d
linked0: > There is nothing this function uses in the Ledger itself other than the enrollment manager.

All right. I have already felt something weird about that and you have suggested the right approach. I'll apply it to the code. Thanks.

source/agora/common/SCPHash.d
linked0: Yes I can, but It's overloaded from the `getHashOf` and it's not bad to show that it's overloaded for the new feature. Don't you agree with this?

AndrejMitrovic: There is a difference. Those other overloads are used by C++ code. This function is not.

source/agora/consensus/protocol/Nominator.d
linked0: No document definition for the return value. uint64_t

> I will add a description for that. Thanks.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -556,6 +557,53 @@ extern(D):
...
+    {
+        const uint hash_N = 1;
+        const uint hash_P = 2;
+
+        uint512 hash;
+        const random_seed =
+            this.ledger.getValidatorRandomSeed(Height(slot_idx - 1));
+        try
+        {
+            hash = uint512(hashMulti(slot_idx, prev[], hash, round_num, node_id,
```
```suggestion
            hash = uint512(hashMulti(slot_idx, prev[], isPriority ? hash_P : hash_N, hash, round_num, node_id,
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -556,6 +557,53 @@ extern(D):
...
+    public override uint64_t computeHashNode (uint64_t slot_idx,
+        ref const(Value) prev, bool is_priority, int32_t round_num,
+        ref const(NodeID) node_id) nothrow
+    {
+        const uint hash_N = 1;
+        const uint hash_P = 2;
+
+        uint512 hash;
+        const random_seed =
+            this.ledger.getValidatorRandomSeed(Height(slot_idx - 1));
```
```suggestion
        const random_seed =
            this.ledger.getValidatorRandomSeed(Height(slot_idx - 1));
        uint512 hash;
```

The declaration should be moved the closest to where it's initialized.

source/agora/consensus/EnrollmentManager.d
linked0: I'm trying to restore the removed assertion, `assert(preimage != PreImageInfo.init);` by adding the code into the test codes like `QuorumPreimage.d` as follows.

```
foreach (enroll; enrolls)
{
    nodes.each!(node => 
        retryFor(node.getPreimage(enroll.utxo_key).distance >= 1 , 10.seconds));
}
```

But, this code fails after outsider validators are enrolled. I guess it's because the outsider validators don't reveal their pre-images timely although they are enrolled as a validator. I need more debugging for that. Is there any variable to distinguish original validators from outsider validators in the code? Any clues would be very helpful. Thanks in advance.

linked0:
```
foreach (enroll; blocks[0].header.enrollments)
{
    nodes.each!(node =>
        retryFor(node.getPreimage(enroll.utxo_key).distance >= 1 , 10.seconds));
}
```

With the codes like above, we can avoid assertions failure in many tests in `agora.test`. But, In QuorumPreimage.d, some failures occur continuously, which is related to outsiders I think. I’m debugging for that.

source/agora/node/Ledger.d
```diff
@@ -518,6 +518,48 @@ public class Ledger
...
+
+        Params:
+            tx = the transaction hash
+
+        Returns:
+            true if the transaction pool has the transaction hash.
+
+    ***************************************************************************/
+
+    public bool hasTransactionHash (const ref Hash tx) @safe
```
AndrejMitrovic: Why did you reintroduce this? it was removed in a previous PR.

source/agora/node/Ledger.d
```diff
@@ -518,6 +518,48 @@ public class Ledger
...
+        Params:
+            height = the desired block height to look up the images for
+
+        Returns:
+            the random seed
+
+    ***************************************************************************/
+
+    public Hash getValidatorRandomSeed (Height height) nothrow
+    {
```
AndrejMitrovic, line 536: Why did you reintroduce it again?

source/agora/test/Base.d
```diff
@@ -553,6 +553,23 @@ public class TestAPIManager
                     idx, node.getBlockHeight(), height), file, line));
     }
 
+    public void expectBlock (Height height, const(BlockHeader) enrolled_header,
```
AndrejMitrovic: This needs documentation. Actually if it only checks the enrollments then it might be better to just pass `b0.header.enrollments` instead as the parameter.

source/agora/test/Base.d
```diff
@@ -553,6 +553,23 @@ public class TestAPIManager
...
+        Duration timeout, string file = __FILE__, int line = __LINE__)
+    {
+        this.expectBlock(this.clients, height, enrolled_header, timeout, file,
+            line);
+    }
+
+    public void expectBlock (Clients)(Clients clients, Height height,
```
AndrejMitrovic, line 592: Needs a `/// Ditto`

source/agora/consensus/protocol/Nominator.d
```diff
@@ -707,6 +708,53 @@ extern(D):
...
+
+        Params:
+            slot_idx = the slot index we're currently reaching consensus for.
+            prev = the previous data set for the provided slot index.
+            is_priority = the flag to check that this call is for priority.
+            round_num = the nomination round
+            node_id = the id of the node for which this computation is being made
+
+        Returns:
+            the 64-byte hash which is unsigned long long type
```
```suggestion
            the 8-byte hash
```

It's 64-bit (8-byte), not 64-byte. And `unsigned long long` is C++ terminology, we should not use it here.

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: Ping.

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: Ping.

AndrejMitrovic (changes requested):

linked0:
After rebasing, `SIGABRT` suddenly occurs in the unit test of agora.test.EnrollmentManager as follows. It occurs during calling the `rollback` of the `ManagedDatabase` in `addValidatedBlock` function.


```
(lldb) r
Process 50685 launched: '/Users/jaylee/work/agora/build/agora-unittests' (x86_64)
Process 50685 stopped
* thread #4, stop reason = signal SIGABRT
    frame #0: 0x00007fff6a52d33a libsystem_kernel.dylib`__pthread_kill + 10
libsystem_kernel.dylib`__pthread_kill:
->  0x7fff6a52d33a <+10>: jae    0x7fff6a52d344            ; <+20>
    0x7fff6a52d33c <+12>: movq   %rax, %rdi
    0x7fff6a52d33f <+15>: jmp    0x7fff6a527629            ; cerror_nocancel
    0x7fff6a52d344 <+20>: retq
Target 0: (agora-unittests) stopped.
(lldb) bt
error: need to add support for DW_TAG_base_type 'immutable(char)' encoded with DW_ATE = 0x10, bit_size = 8
* thread #4, stop reason = signal SIGABRT
  * frame #0: 0x00007fff6a52d33a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff6a5e9e60 libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff6a4b4808 libsystem_c.dylib`abort + 120
    frame #3: 0x00000001001bdbbb agora-unittests`_D5agora6common15ManagedDatabaseQr8rollbackFNbNeZv at ManagedDatabase.d:165:17
    frame #4: 0x00000001001bf590 agora-unittests`_D5agora4node6LedgerQh17addValidatedBlockMFNfKxSQBu9consensus4data5BlockQgZv(this=0x00000001b5d0d5c0, block=0x00000002b6617590) at Ledger.d:290:25
    frame #5: 0x00000001001bf2fb agora-unittests`_D5agora4node6LedgerQh11acceptBlockMFNfKxSQBo9consensus4data5BlockQgZb(this=0x00000001b5d0d5c0, block=0x00000002b6617590) at Ledger.d:233:9
    frame #6: 0x00000001001befe1 agora-unittests`_D5agora4node6LedgerQh14onExternalizedMFNeSQBp9consensus4data13ConsensusDataQpZb(this=0x00000001b5d0d5c0, data=ConsensusData @ 0x00000002b66176b0) at Ledger.d:208:9
    frame #7: 0x0000000100154193 agora-unittests`Nominator::valueExternalized(this=0x00000001b5cbb160, slot_idx=1, value=<unavailable>) at Nominator.d:579:13
    frame #8: 0x0000000100630186 agora-unittests`stellar::BallotProtocol::setConfirmCommit(this=0x00000001044c9548, c=0x00000002b6617a40, h=0x00000002b6617a20) at BallotProtocol.cpp:1517:26
    frame #9: 0x000000010062fc29 agora-unittests`stellar::BallotProtocol::attemptConfirmCommit(this=0x00000001044c9548, hint=0x00000002b66190c8) at BallotProtocol.cpp:1493:16
    frame #10: 0x00000001006277e4 agora-unittests`stellar::BallotProtocol::advanceSlot(this=0x00000001044c9548, hint=0x00000002b66190c8) at BallotProtocol.cpp:1877:15
    frame #11: 0x00000001006260e7 agora-unittests`stellar::BallotProtocol::processEnvelope(this=0x00000001044c9548, envelope=0x00000002b66190c8, self=true) at BallotProtocol.cpp:199:13
    frame #12: 0x00000001006e1bb1 agora-unittests`stellar::Slot::processEnvelope(this=0x00000001044c9528, envelope=0x00000002b66190c8, self=true) at Slot.cpp:143:35
    frame #13: 0x0000000100628ccf agora-unittests`stellar::BallotProtocol::emitCurrentStateStatement(this=0x00000001044c9548) at BallotProtocol.cpp:626:19
    frame #14: 0x000000010062c8d6 agora-unittests`stellar::BallotProtocol::setPreparedAccept(this=0x00000001044c9548, ballot=0x00000002b6619590) at BallotProtocol.cpp:917:9
    frame #15: 0x000000010062be1f agora-unittests`stellar::BallotProtocol::attemptPreparedAccept(this=0x00000001044c9548, hint=0x00000002b661ad30) at BallotProtocol.cpp:881:20
    frame #16: 0x0000000100627736 agora-unittests`stellar::BallotProtocol::advanceSlot(this=0x00000001044c9548, hint=0x00000002b661ad30) at BallotProtocol.cpp:1871:15
    frame #17: 0x00000001006260e7 agora-unittests`stellar::BallotProtocol::processEnvelope(this=0x00000001044c9548, envelope=0x00000002b661ad30, self=false) at BallotProtocol.cpp:199:13
    frame #18: 0x00000001006e1bb1 agora-unittests`stellar::Slot::processEnvelope(this=0x00000001044c9528, envelope=0x00000002b661ad30, self=false) at Slot.cpp:143:35
    frame #19: 0x00000001006d0f80 agora-unittests`stellar::SCP::receiveEnvelope(this=0x00000001064a8e80, envelope=0x00000002b661ad30) at SCP.cpp:33:38
    frame #20: 0x0000000100154c33 agora-unittests`_D5agora9consensus8protocol9NominatorQk15receiveEnvelopeMFNeMKxS4scpd5types11Stellar_SCP11SCPEnvelopeZv(this=0x00000001b5cbb160, envelope=0x00000002b661ad30) at Nominator.d:466:9
    frame #21: 0x0000000100155a1e agora-unittests`_D5agora4node9ValidatorQk15receiveEnvelopeMFNfS4scpd5types11Stellar_SCP11SCPEnvelopeZv(this=0x0000700011813700, envelope=SCPEnvelope @ 0x00000002b661ad30) at Validator.d:260:9
    frame #22: 0x000000010017ec6f agora-unittests`_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPITSQBaQxQu10SerializerZQCg13handleCommandFSQDuQDq7CommandQDdSQEmQEi9FilterAPIZv(cmd=Command @ 0x00000002b661bec0, node=0x0000700011813960, filter=FilterAPI @ 0x00000002b661bea0) at LocalRest.d-mixin-477:494:33
    frame #23: 0x00000001001a9c03 agora-unittests`_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPITSQBaQxQu10SerializerZQCg__T7spawnedTCQCkQCh17EnrollmentManager18__unittest_L287_C1FZ20MisbehavingValidatorZQDcFNbSQGp11concurrency3TidAyaiSQGi6common6ConfigQhPOSQIk8RegistryQjAySQHv9consensus4data5BlockQgmPOlPOmZ7runNodeMFZ__T6handleTSQLgQLc7CommandZQyMFQuZ9__lambda2MFZv at LocalRest.d:602:43
    frame #24: 0x0000000100787b19 agora-unittests`_D6geod2411concurrency14FiberScheduler6createMFNbDFZvZ4wrapMFZv at concurrency.d:916:13
    frame #25: 0x0000000100c5a36f agora-unittests`_D4core6thread7context8Callable6opCallMFZv(this=0x0000000000000002) at context.d:46:17
    frame #26: 0x0000000100c5a4fe agora-unittests`_D4core6thread5fiber5Fiber3runMFZv(this=0x0000000279240880) at fiber.d:1127:9
    frame #27: 0x0000000100c5a433 agora-unittests`fiber_entryPoint at fiber.d:184:13
```


In fact, [this line](https://github.com/bpfkorea/agora/blob/320a7981ca93eaaee4bf067cd16b78db603a1041/source/agora/test/EnrollmentManager.d#L416) makes the `SIGABRT`.

Will analyze more.

source/agora/test/Base.d
linked0: In fact, it needs `height` because the `waitForPreimages` in the function is called with a distance which is the requested height minus the enrolled height. I documented that.

linked0:
Ready for review.
I solved the above `SIGABRT` issue by calling the `waitForPreimage` function.

source/agora/test/Base.d
```diff
@@ -564,6 +567,25 @@ public class TestAPIManager
...
+        this.expectBlock(this.clients, height, enrolled_header, timeout, file,
+            line);
+    }
+
+    /// Ditto
+    public void expectBlock (Clients)(Clients clients, Height height,
+        const(BlockHeader) enrolled_header, Duration timeout,
+        string file = __FILE__, int line = __LINE__) if (isInputRange!Clients)
+    {
+        auto distance = cast(ushort)(height - enrolled_header.height - 1);
```
AndrejMitrovic: Could you add `assert(height > enrolled_header.height)`? Otherwise this will overflow and turn into `ulong.max`.

source/agora/node/Ledger.d
```diff
@@ -518,6 +518,48 @@ public class Ledger
...
+            tx = the transaction hash
+
+        Returns:
+            true if the transaction pool has the transaction hash.
+
+    ***************************************************************************/
+
+    public bool hasTransactionHash (const ref Hash tx) @safe
+    {
+        return this.pool.hasTransactionHash(tx);
```
AndrejMitrovic: @linked0 Ping. This forwarding function does not need to exist.

AndrejMitrovic:
LGTM. Just please remove that `hasTransactionHash` wrapper. I think it was left in accidentally.

source/agora/test/Base.d
linked0: > Could you add assert(height > enrolled_header.height)? Otherwise this will overflow and turn into ulong.max.

Yes. I will. I think that's needed.

source/agora/node/Ledger.d
linked0: > This forwarding function does not need to exist.

I'll check it. Ty.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -17,8 +17,9 @@ import agora.common.crypto.ECC;
 import agora.common.crypto.Key;
 import agora.common.crypto.Schnorr;
 import agora.common.Config;
-import agora.common.Hash : Hash, HashDg, hashPart, hashFull;
+import agora.common.Hash : Hash, HashDg, hashPart, hashFull, hashMulti;
```
Geod24, line 20: At this point, just import the whole module

source/agora/consensus/EnrollmentManager.d
Geod24: ~Yes, this needs to be reverted. I can take a look.~

source/agora/consensus/EnrollmentManager.d
```diff
@@ -543,8 +543,9 @@ public class EnrollmentManager
         foreach (const ref key; keys)
         {
             const preimage = this.validator_set.getPreimageAt(key, height);
-            // TODO: The preimage should not have the .init value. So, checking
-            // that is needed.
+            // this should not happen. validators which didn't reveal the
+            // preimage should not be in the active validator set 'keys'.
+            assert(preimage != PreImageInfo.init);
```
Geod24: Looks like  the second commit should come first, so the assertion doesn't need to be deleted in the first place ?

source/agora/test/Base.d
```diff
@@ -540,6 +540,9 @@ public class TestAPIManager
         heights for only these nodes. Note that the clock time is adjusted
         for all nodes (this is what most tests expect).
 
+        Other overloads allow passing a `BlockHeader` of an enrolled height to
+        get enrollments and a height for checking pre-images are revealed.
```
Geod24: This seems to me like an abuse of overloading. The other overloads do much more (checking pre-images) than this. Perhaps the other overloads should just be the default ? Although the client doesn't care much about pre-images.

Geod24 (commented):
The second commit looks like a workaround for a deeper problem, but if it allows to get this moved forward, I think it's okay.
Just invert the commits, so the removal of the assertion doesn't need to happen.

source/agora/consensus/EnrollmentManager.d
linked0: Yeah, it would be good to come first.

source/agora/test/Base.d
linked0: I'm thinking about what the best document for that is. I'll add the right description.

source/agora/consensus/protocol/Nominator.d
linked0: > At this point, just import the whole module

If I import the whole module, the compiler outputs error as follows.
```
source/agora/consensus/protocol/Nominator.d(84,35): Error: alias agora.common.Hash.Hash at source/agora/common/Hash.d(86,8) conflicts with alias scpd.types.Stellar_types.Hash at source/scpd/types/Stellar_types.d(32,1) 
```

linked0:
Ready for review. Thanks for your reviews.

AndrejMitrovic (approved):

Issue #1083, opened 2020-07-30
Ledger routines return boolean status codes but also throw


AndrejMitrovic:
https://github.com/bpfkorea/agora/blob/1af3731d063a8b52f1e114a4b31e5ac65ad70869/source/agora/node/Ledger.d#L207

And also:

https://github.com/bpfkorea/agora/blob/1af3731d063a8b52f1e114a4b31e5ac65ad70869/source/agora/node/Ledger.d#L268-L269

Here the assert might not get triggered because `saveBlock` is not `nothrow`.

All of this has to be fixed.

Geod24:
This has been mostly fixed, there might be one or two inconsistencies left in other classes but not in the mentioned methods.

PR #1161, opened 2020-09-01
Make ledger routines consistent

kchulj:
The ledger routines are currently inconsistent, with a routine returning either boolean status codes, exceptions, or `assert`.

They should not return boolean status codes **and** also throw. There was a situation in the code below
https://github.com/bpfkorea/agora/blob/1af3731d063a8b52f1e114a4b31e5ac65ad70869/source/agora/node/Ledger.d#L268-L269
where the `assert` might not get triggered because `saveBlock` is not nothrow. `saveBlock` is now `nothrow`, but the other routines need to be fixed.
I would like some suggestions on how to structure the routines.

Relates #1083

### DIFF
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -128,26 +128,34 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public bool add (Transaction tx) @safe
+    public bool add (Transaction tx) @safe nothrow
     {
         static ubyte[] buffer;
 
         // check double-spend
         if (!isValidTransaction(tx))
             return false;
+        try
+        {
+            // insert each input information of the transaction
+            foreach (const ref input; tx.inputs)
+                this.input_set.put(input.hashFull());
 
-        // insert each input information of the transaction
-        foreach (const ref input; tx.inputs)
-            this.input_set.put(input.hashFull());
+            serializeToBuffer(tx, buffer);
 
-        serializeToBuffer(tx, buffer);
+            () @trusted {
+                db.execute("INSERT INTO tx_pool (key, val) VALUES (?, ?)",
+                    hashFull(tx)[], buffer);
+            }();
 
-        () @trusted {
-            db.execute("INSERT INTO tx_pool (key, val) VALUES (?, ?)",
-                hashFull(tx)[], buffer);
-        }();
-
-        return true;
+            return true;
+        }
+        catch (Exception ex)
+        {
+            log.error("Error adding transaction: {}" ~
+                "Transaction: {}", ex.msg, tx);
+            return false;
+        }
     }
 
     /***************************************************************************
@@ -244,10 +252,19 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public bool hasTransactionHash (const ref Hash tx) @trusted
+    public bool hasTransactionHash (const ref Hash tx) @trusted nothrow
     {
-        return this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
-            "tx_pool WHERE key = ?)", tx[]).front.peek!bool(0);
+        try
+        {
+            return this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
+                "tx_pool WHERE key = ?)", tx[]).front.peek!bool(0);
+        }
+        catch (Exception ex)
+        {
+            log.error("Error finding transaction hash: {}, " ~
+                "Hash: {}", ex.msg, tx);
+            return false;
+        }
     }
 
     /***************************************************************************
@@ -262,7 +279,7 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    private bool isValidTransaction (const ref Transaction tx) @trusted
+    private bool isValidTransaction (const ref Transaction tx) @trusted nothrow
     {
         auto txHash = tx.hashFull();
 
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -411,9 +411,16 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public size_t validatorCount () @safe
+    public size_t validatorCount () @safe nothrow
     {
-        return this.validator_set.count();
+        try
+        {
+            return this.validator_set.count();
+        }
+        catch (Exception ex)
+        {
+            log.error("Error while getting validator count: {}", ex.msg);
+        }
     }
 
     /***************************************************************************
@@ -482,7 +489,7 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public bool getNextPreimage (out PreImageInfo preimage) @safe
+    public bool getNextPreimage (out PreImageInfo preimage) @safe nothrow
     {
         const enrolled = this.getEnrolledHeight(this.enroll_key);
         if (enrolled == ulong.max)
@@ -631,7 +638,7 @@ public class EnrollmentManager
     }
 
     /// Returns: true if this validator is currently enrolled
-    public bool isEnrolled (UTXOFinder finder) nothrow @safe
+    public bool isEnrolled (UTXOFinder finder) @safe nothrow
     {
         Hash[] utxo_keys;
         assert(this.validator_set.getEnrolledUTXOs(utxo_keys));
@@ -830,7 +837,7 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public Enrollment getEnrollment (const ref Hash enroll_hash) @trusted
+    public Enrollment getEnrollment (const ref Hash enroll_hash) @trusted nothrow
     {
         return this.enroll_pool.getEnrollment(enroll_hash);
     }
@@ -844,7 +851,7 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public void removeEnrollment (const ref Hash enroll_hash) @trusted
+    public void removeEnrollment (const ref Hash enroll_hash) @trusted nothrow
     {
         this.enroll_pool.remove(enroll_hash);
     }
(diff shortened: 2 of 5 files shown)


source/agora/common/TransactionPool.d
```diff
@@ -135,26 +135,35 @@ public class TransactionPool
...
+                    hashFull(tx)[], buffer);
+            }();
 
-        return true;
+            return true;
+        }
+        catch (Exception ex)
+        {
+            log.error("Error adding transaction {}" ~
+            "Transaction: {}", ex.msg, tx);
```
linked0: An indentation needed.

source/agora/common/TransactionPool.d
```diff
@@ -204,20 +213,26 @@ public class TransactionPool
...
+                auto tx = deserializeFull!Transaction(row.peek!(ubyte[])(1));
 
+                // break
+                if (auto ret = dg(hash, tx))
+                    return ret;
+            }
+        }
+        catch (Exception ex)
+        {
+            // nothing we can do
```
linked0: Shouldn't we just add a message with log.trace?

linked0 (commented):
Looks good but one more comment.

source/agora/common/TransactionPool.d
AndrejMitrovic: We should never catch exceptions and then ignore them. This is also known as the "pokemon exception handling".

Instead:

- Let's make the delegate nothrow
- try/catch on the `execute` call, but then `assert(0)`.

In the future https://github.com/bpfkorea/agora/issues/836 will have to be looked at again - we will need to assess each `execute` call and see if we can handle any sqlite failures - we may want to retry the statement or abort.

source/agora/common/TransactionPool.d
```diff
@@ -233,10 +248,20 @@ public class TransactionPool
...
+    public bool hasTransactionHash (const ref Hash tx) @trusted nothrow
     {
-        return this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
-            "tx_pool WHERE key = ?)", tx[]).front.peek!bool(0);
+        try
+        {
+            return this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
+                "tx_pool WHERE key = ?)", tx[]).front.peek!bool(0);
+        }
+
```
AndrejMitrovic:
```suggestion
```

source/agora/common/TransactionPool.d
```diff
@@ -204,20 +213,66 @@ public class TransactionPool
...
+                    return ret;
+            }
+            return 0;
+        }();
 
+        if (dg is null)
+        {
+            Transaction tx;
+            dg(tx);
+        }
```
kchulj: During the pair programming, @Geod24 and I modified the `opApply` overload with a template.
Now, there is a separate overload for parameters `(Transaction)` and `(Hash, Transaction)`.
Since `dg` is never `null`, this specific block of code alters `opApply`'s attributes, under the hood.

source/agora/common/TransactionPool.d
```diff
@@ -204,20 +213,66 @@ public class TransactionPool
...
+        // can be removed
+        if (dg is null)
+        {
+            Transaction tx;
+            dg(tx);
         }
+        return 0;
+    }
+
+    public int opApply (DG2T) (scope DG2T dg)
```
AndrejMitrovic: If you annotate this overload with `@disable` you'll see it's never actually used anymore. So we may want to just get rid of it?

source/agora/common/TransactionPool.d
```diff
@@ -280,13 +344,13 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    version (unittest) private Transaction[] take (size_t count) @safe
+    version (unittest) private const(Transaction)[] take (size_t count)
```
AndrejMitrovic: Why was `@safe` removed?

kchulj: In this function, there is a `foreach` iteration over `Transaction[]`.
Since the `opApply` overload is not `@safe`, the `@safe` here has been removed to match the attribute.

AndrejMitrovic: You made it templated, so safety is inferred. Add the `@safe` back and you'll see it compiles.

source/agora/common/TransactionPool.d
kchulj: Yes, it compiles with `@disable`.
Does it mean that the overload for `(Transaction)` works for `(Hash, Transaction)` too?

AndrejMitrovic: No it means it's not used anymore! Because you changed all of the calls to only use `tx`.

Previously, for whatever reason we used `foreach (hash, tx)` in many places, but then discarded the hash. But you've corrected that, and now they're just doing `foreach (tx) ...`. So you don't need the second overload anymore, just the one taking transactions.

source/agora/common/TransactionPool.d
kchulj: Yes, it compiles with the `@safe`.
Thanks for the feedback.

kchulj:
Ready for review.

AndrejMitrovic (approved):

source/agora/node/FullNode.d
```diff
@@ -521,9 +521,18 @@ public class FullNode : API
...
     {
-        this.pushBlock(block);
+        try
+        {
+            this.pushBlock(block);
+        }
+        catch (Exception ex)
+        {
+            log.error("Error pushing block: {}" ~
+                "Block: {}", ex.msg, block);
```
TrustHenry, line 534:
```suggestion
                "Block: {}", ex.msg, prettify(block));
```
I don't know if the information needed for debugging should look pretty.

TrustHenry (approved):
Nice! LGTM

source/agora/node/FullNode.d
kchulj: Yes, that is a good idea. It is also used in other `log`s.

source/agora/node/Ledger.d
```diff
@@ -347,11 +347,12 @@ public class Ledger
 
         this.enroll_man.getEnrollments(data.enrolls,
             Height(this.getBlockHeight()));
-        foreach (hash, tx; this.pool)
+        foreach (ref Transaction tx; this.pool)
         {
             if (auto reason = tx.isInvalidReason(utxo_finder, next_height))
                 log.trace("Rejected invalid ('{}') tx: {}", reason, tx);
             else
+                // TODO: Make `ConsensusData.tx_set` const correct
```
Geod24: This can be removed since the cast is gone

source/agora/node/FullNode.d
```diff
@@ -521,9 +521,18 @@ public class FullNode : API
...
+            assert(0);
```
Geod24: So `pushBlock` can never `throw` ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -779,6 +787,7 @@ public class EnrollmentManager
         catch (Exception ex)
         {
             log.error("ManagedDatabase operation error {}", ex);
+            assert(0);
```
Geod24: Those are a bit scary.

Geod24 (changes requested):
This is a good example of a PR that should be split in at least 2 commits. The change to the `opApply` which makes it templated should be its own commit.

Additionally, I am not sure why `assert(0);` is the right solution. It looks like it could sometimes trigger for `ManagedDatabase` ?

source/agora/node/FullNode.d
kchulj: Since `pushBlock` is pushing to a handler, it should be recoverable.
Therefore, `assert(0)` is not the right solution.

linked0:
Rebase it, please.

TrustHenry:
Please update the git submodules.
`git submodule update`

source/agora/common/TransactionPool.d
```diff
@@ -128,26 +128,35 @@ public class TransactionPool
...
 
-        return true;
+            return true;
+        }
+        catch (Exception ex)
+        {
+            log.error("Error adding transaction: {}" ~
+                "Transaction: {}", ex.msg, tx);
+            return false;
+        }
```
Geod24: Can the whole function throws ? It looks like at least what's before the `foreach` cannot.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -417,9 +417,17 @@ public class EnrollmentManager
...
     {
-        return this.validator_set.count();
+        try
+        {
+            return this.validator_set.count();
+        }
+        catch (Exception ex)
+        {
+            log.error("Error while getting validator count: {}", ex.msg);
+            assert(0);
```
Geod24: Just providing a summary of the discussion we had:
Those `assert(0)` are problematic, because they mean "this should **never** happen". Then why is the called function not `nothrow` ?
Usually, this can happen for one main reason: the calling function **can** throw with some parameters, or in some state, but in the state/parameters we're providing, it will never. In this case, there is no parameter provided to `count`, so it has to be tied to the state.
If we look [at the code](https://github.com/bpfkorea/agora/blob/dc869bd16b84188606f3c514b4aefb197e7c26e3/source/agora/consensus/ValidatorSet.d#L120-L139), we see it's just doing a `SELECT`.
So why would it `throw` ? One can make guesses: the lack of type safety means `oneValue` means that it *has* to `throw` (because it's using `std.conv` internally). In this case, perhaps adding a counting function in the sqlite3 bindings would solve the issue. It could also `throw` on non existing tables. Then perhaps there is another construct that would make sure we get a reference to a table that can never be deleted ? Lastly, what if it's simply an IO error ? In this case, all bets are off. If it reads from disk somewhere, and that leads to a `throw`, it **could** happen and thus our `assert(0)` has an invalid assumption.

Geod24 (changes requested):
Commented on my concerns, we need to discuss / whiteboard.

Geod24:
Thanks, that was a good start but it prompted bigger discussion. I think we can close this as stale, and address it piece-meal.

PR #1162, opened 2020-09-02
Demonstrate one bad and one potentially good signature scheme

AndrejMitrovic:
This demonstrates what was written on https://tlu.tarilabs.com/cryptography/digital_signatures/introduction_schnorr_signatures.html#why-do-we-need-the-nonce, that not including 'r' in the signature can make it easy to extract the private key.

Note that it might still be possible to have a scheme like this:

```
c = hash(msg)
s = r + (k * c)
```

Where 'r' is included in signing but is not included in the hash. In fact I don't think the hashing part is important at all. The hashing part is what we commit to, but we have already committed to all P (public keys) and R (preimages) beforehand so it's unnecessary to re-commit to it by including it in the hash. However I'll try to write a test-case for this too.

### DIFF
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -166,6 +166,24 @@ public struct Scalar
     {
         dg(this.data[]);
     }
+
+    /***************************************************************************
+
+        Returns:
+            the inverted scalar.
+
+        See_Also: https://libsodium.gitbook.io/doc/advanced/point-arithmetic
+        See_Also: https://tlu.tarilabs.com/cryptography/digital_signatures/introduction_schnorr_signatures.html#why-do-we-need-the-nonce
+
+    ***************************************************************************/
+
+    public Scalar invert () @trusted
+    {
+        Scalar scalar = this;  // copy
+        assert(crypto_core_ed25519_scalar_invert(scalar.data[].ptr,
+            this.data[].ptr) == 0);
+        return scalar;
+    }
 }
 
 ///
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -172,8 +172,9 @@ public struct Scalar
         Returns:
             the inverted scalar.
 
-        See_Also: https://libsodium.gitbook.io/doc/advanced/point-arithmetic
-        See_Also: https://tlu.tarilabs.com/cryptography/digital_signatures/introduction_schnorr_signatures.html#why-do-we-need-the-nonce
+        See_Also:
+            https://libsodium.gitbook.io/doc/advanced/point-arithmetic
+            https://tlu.tarilabs.com/cryptography/digital_signatures/introduction_schnorr_signatures.html#why-do-we-need-the-nonce
 
     ***************************************************************************/
 
--- a/source/agora/common/crypto/Schnorr.d
+++ b/source/agora/common/crypto/Schnorr.d
@@ -291,3 +291,62 @@ nothrow @nogc @safe unittest
     Scalar stolen_key = s * c.invert();
     assert(stolen_key == kp.v);
 }
+
+// possibly secure signature scheme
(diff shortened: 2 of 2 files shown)


source/agora/common/crypto/Schnorr.d
```diff
@@ -269,3 +269,25 @@ nothrow @nogc @safe unittest
...
+    Scalar s = (kp.v * c);  // signature
+
+    // known public data of the node
+    Point K = kp.V;
+
+    // other nodes verify
+    assert(s.toPoint() == K * c);
+
+    // but the other node can also extract the private key!
+    Scalar stolen_key = s * c.invert();
```
TrustHenry, line 291: There's a problem. The verification test is good.

AndrejMitrovic: I don't understand your meaning?

TrustHenry: Which means you found the wrong signature case.

AndrejMitrovic: That's what the PR is trying to demonstrate. It says "bad signature scheme" in the title and the documentation and the commit message explain it.

AndrejMitrovic:
@hewison-chris I added another commit demonstrating the suggested scheme.

hewison-chris:
> @hewison-chris I added another commit demonstrating the suggested scheme.

This looks really good. I hope we can get it verified as secure and then we can start using it for block signing for validators with only n of m signatures.

source/agora/common/crypto/ECC.d
```diff
@@ -166,6 +166,24 @@ public struct Scalar
...
         dg(this.data[]);
     }
+
+    /***************************************************************************
+
+        Returns:
+            the inverted scalar.
+
+        See_Also: https://libsodium.gitbook.io/doc/advanced/point-arithmetic
+        See_Also: https://tlu.tarilabs.com/cryptography/digital_signatures/introduction_schnorr_signatures.html#why-do-we-need-the-nonce
```
```suggestion
        See_Also:
            https://libsodium.gitbook.io/doc/advanced/point-arithmetic
            https://tlu.tarilabs.com/cryptography/digital_signatures/introduction_schnorr_signatures.html#why-do-we-need-the-nonce
```

Could you also move it after `opUnary`, I think that makes more sense.

source/agora/common/crypto/Schnorr.d
```diff
@@ -269,3 +269,80 @@ nothrow @nogc @safe unittest
...
+
+    // verification of individual signatures
+    assert(s_1.toPoint() == R_1 + (K_1 * c));
+    assert(s_2.toPoint() == R_2 + (K_2 * c));
+
+    // "multi-sig" - collection of one or more signatures
+    auto sum_s = s_1 + s_2;
+    assert(sum_s.toPoint() ==
+        (R_1 + (K_1 * c)) +
+        (R_2 + (K_2 * c)));
```
Geod24, line 347:
```suggestion
    assert(sum_s.toPoint() ==
        (R_1 + (K_1 * c)) +
        (R_2 + (K_2 * c)));
    // Or the equivalent:
    assert(sum_s.toPoint() ==
        (R_1 + R2) + (K_1 * c) + (K_2 * c));
```

Geod24: The reason for this suggestion is that it will be closer to what's done during signature aggregation.

Geod24 (approved):
LGTM

source/agora/common/crypto/Schnorr.d
AndrejMitrovic: I'm not sure I follow but I've applied it regardless.

AndrejMitrovic:
Actually I think what I had in mind w.r.t. the signature scheme might even be very different to what other people had in mind. That's just my gut feeling.

I'll open up another PR soon demonstrating how it works.

PR #1168, opened 2020-09-03
Change the way pre-images are revealed

### LINKED ISSUE #1166: Pre-image received before new validator's information is saved in a block
This problem appears in PR #1163 and in the middle of modifying the code of PR #1156 and only occurs in the network tests.

Suppose that one outsider(A) and an already enrolled validator(B) are being enrolled in a block for a new validator cycle.
When the outsider A is enrolled as a validator, it receives an pre-image of validator B before the new block having the information of the new enrollments is added into the chain. Technically, `receivedPreimage` from the validator B is called on A side before `addValidatedBlock` is called while the new block containing the new validators is being made on A side.

linked0:
There was a timing issue where the pre-image of a new validator cycle arrives before the new cycle does not start yet which means a receiving node never gets the pre-image again. So, we need to change the way a validator reveals pre-images. We have decided that validators keep revealing the pre-image set to be revealed in the revealing period. After the period, validators start and keep revealing new pre-image for a new period.

Fixes #1166

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -496,24 +496,6 @@ public class EnrollmentManager
 
         const height = Height(min(this.next_reveal_height + PreimageRevealPeriod * 2,
                                 enrolled + this.params.ValidatorCycle));
-        return getPreimage(height, preimage);
-    }
-
-    /***************************************************************************
-
-        Get a pre-image at a certain height
-
-        Params:
-            height = the number of the height at which the pre-image exists
-            preimage = will contain the PreImageInfo if exists
-
-        Returns:
-            true if the pre-image exists
-
-    ***************************************************************************/
-
-    public bool getPreimage (Height height, out PreImageInfo preimage) @safe
-    {
         const enrolled_height =
             this.validator_set.getEnrolledHeight(this.enroll_key);
         assert(enrolled_height != ulong.max);
@@ -997,18 +979,10 @@ unittest
     // clear up all validators
     man.clearExpiredValidators(Height(1018));
 
-    // get a pre-image at a certain height
-    // A validation can start at the height of the enrolled height plus 1.
-    // So, a pre-image can only be got from the start height.
+    // A validation is enrolled at the height of 10.
     PreImageInfo preimage;
     enroll = man.createEnrollment(utxo_hash);
     assert(man.addValidator(enroll, Height(10), &utxo_set.findUTXO, utxos) is null);
-    assert(!man.getPreimage(Height(10), preimage));
-    assert(man.getPreimage(Height(11), preimage));
-    assert(preimage.hash == man.cycle.preimages[$ - 1]);
-    assert(man.getPreimage(Height(10 + man.params.ValidatorCycle), preimage));
-    assert(preimage.hash == man.cycle.preimages[0]);
-    assert(!man.getPreimage(Height(11 + man.params.ValidatorCycle), preimage));
 
     /// test for the functions about periodic revelation of a pre-image
     assert(man.needRevealPreimage(Height(10)));
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -96,8 +96,8 @@ public class EnrollmentManager
     /// Random key for enrollment
     private Pair signature_noise;
 
-    /// Next height for pre-image revelation
-    private Height next_reveal_height;
+    /// Distance of the preimage being revealed next time
+    private uint next_reveal_distance;
 
     /// The period for revealing a preimage
     /// It is an hour interval if a block is made in every 10 minutes
@@ -157,7 +157,7 @@ public class EnrollmentManager
         this.key_pair.V = this.key_pair.v.toPoint();
 
         // load next height for preimage revelation
-        this.next_reveal_height = this.getNextRevealHeight();
+        this.next_reveal_distance = this.getNextRevealDistance();
 
         // load the count of 'populate' of `PreImageCycle`
         const uint populate_count = this.getCycleIndex();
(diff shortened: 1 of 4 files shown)


source/agora/node/Validator.d
```diff
@@ -352,14 +352,13 @@ public class Validator : FullNode, API
...
     {
-        if (!this.enroll_man.needRevealPreimage(this.ledger.getBlockHeight()))
-            return;
-
         PreImageInfo preimage;
         if (this.enroll_man.getNextPreimage(preimage))
         {
-            this.receivePreimage(preimage);
-            this.enroll_man.increaseNextRevealHeight();
+            this.enroll_man.addPreimage(preimage);
```
AndrejMitrovic, line 358: Yes, this is a good change! We shouldn't call an API end-point but instead use the enrollment manager directly. 👍

source/agora/node/Validator.d
```diff
@@ -352,14 +352,13 @@ public class Validator : FullNode, API
...
+            this.network.sendPreimage(preimage);
+            this.pushPreImage(preimage);
+            this.enroll_man.updateRevealDistance(this.ledger.getBlockHeight());
```
AndrejMitrovic, line 361: By the way, a future improvement could be to **stop** the preimage revealing timer once a node's enrollment has expired. It's unrelated to this PR.

source/agora/test/QuorumPreimage.d
```diff
@@ -175,6 +177,10 @@ unittest
     // at block height 10 the validator set has changed
     network.expectBlock(Height(10), 3.seconds);
 
+    // check if the needed pre-images are revealed timely
+    nodes.each!(node =>
+        retryFor(node.getPreimage(enrolls[0].utxo_key).distance >= 6, 5.seconds));
```
AndrejMitrovic: We should check all the enrollments though:

```suggestion
    enrolls.each!(enroll =>
        nodes.each!(node =>
            retryFor(node.getPreimage(enroll.utxo_key).distance >= 6, 5.seconds)));
```

(Not tested)

source/agora/consensus/EnrollmentManager.d
```diff
@@ -96,8 +96,8 @@ public class EnrollmentManager
     /// Random key for enrollment
     private Pair signature_noise;
 
-    /// Next height for pre-image revelation
-    private Height next_reveal_height;
+    /// Distance of the preimage being revealed next time
+    private uint next_reveal_dist;
```
```suggestion
    private uint next_reveal_distance;
```

I prefer it longer like this because `dist` can mean other things. Like "distribution".

source/agora/consensus/EnrollmentManager.d
```diff
@@ -716,63 +658,48 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Get the next block height to reveal a pre-image
-
-        Returns:
-            the next block height to reveal a pre-image. if any problem in
-            getting the value, it returns the MAX ulong value.
+        Get the distance of the pre-image that will be revealed next time
```
AndrejMitrovic, line 661:
```suggestion
        Returns:
            Get the distance of the pre-image that will be revealed next time,
            or 0 if there's no preimage to reveal
```

The comment I made about `0` I'm not sure about, you probably know better than me.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -716,63 +658,48 @@ public class EnrollmentManager
...
+        return 0;
     }
 
     /***************************************************************************
 
-        Set the next block height to reveal a pre-image
-
-        Params:
-            height = the next block height to reveal a pre-image
+        Set the distance of the pre-image that will be revealed next time
```
AndrejMitrovic, line 690: Needs `Params`. And let's rename it to `distance`.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -716,63 +658,48 @@ public class EnrollmentManager
...
         try
         {
             return () @trusted {
                 auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
-                    "WHERE key = ?", "next_reveal_height");
+                    "WHERE key = ?", "next_reveal_distance");
                 if (!results.empty)
-                    return Height(results.oneValue!(ulong));
-                return Height(ulong.max);
+                    return results.oneValue!(uint);
```
AndrejMitrovic, line 677: Good change.

AndrejMitrovic (changes requested):
Very good work!

linked0:
Ready for review.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -831,6 +766,28 @@ public class EnrollmentManager
...
+    {
+        const enrolled = this.validator_set.getEnrolledHeight(this.enroll_key);
+        if (enrolled == ulong.max)
+            return;
+
+        assert(height >= enrolled);
+        auto dist = cast(uint)(height - enrolled);
+        dist = min(PreimageRevealPeriod * ((dist / PreimageRevealPeriod) + 1),
+                    this.params.ValidatorCycle - 1);
+        this.setNextRevealDistance(dist);
```
AndrejMitrovic: This is using the same variable for two different values. We should never re-use existing variables for different purposes. Instead I suggest this:

```suggestion
        auto cur_distance = cast(uint)(height - enrolled);
        auto next_distance = min(PreimageRevealPeriod * ((cur_distance / PreimageRevealPeriod) + 1),
                    this.params.ValidatorCycle - 1);
        this.setNextRevealDistance(next_distance);
```

AndrejMitrovic (approved):
LGTM except one change requested.

source/agora/consensus/EnrollmentManager.d
linked0: > This is using the same variable for two different values. We should never re-use existing variables for different purposes. Instead I suggest this:

How about using `cur` and `next` instead of `cur_distance` and `next_distance` because the modified line would be too long?

AndrejMitrovic: Sure. Although `curr` is usually the shortened form of `current`.

Geod24 (approved):
Oh I like it!

PR #1172, opened 2020-09-07
Remove makePayTx and makeFreezeTx

kchulj:
The unittest in `test/Quorum.d` was using outdated functions `makePayTx` and `makeFreezeTransaction`.
These are copies of `makeChainedTransactions` and have been replaced by `TxBuilder`.

Improves #1107

### DIFF
--- a/source/agora/test/Quorum.d
+++ b/source/agora/test/Quorum.d
@@ -54,52 +54,23 @@ unittest
     auto nodes = network.clients;
     network.expectBlock(Height(8), 5.seconds);
 
-    Transaction makePayTx (in Transaction prev, PublicKey[] addresses, uint index = 0)
-    {
-        auto input = Input(hashFull(prev), index);
-
-        Transaction tx =
-        {
-            TxType.Payment,
-            [input],
-            addresses.map!(addr => Output(Amount.MinFreezeAmount, addr)).array
-        };
-
-        auto signature = WK.Keys.Genesis.secret.sign(hashFull(tx)[]);
-        tx.inputs[0].signature = signature;
-        return tx;
-    }
-
-    Transaction makeFreezeTransaction (in Transaction prev, PublicKey address)
-    {
-        auto input = Input(hashFull(prev), 0);
-
-        Transaction tx =
-        {
-            TxType.Freeze,
-            [input],
-            [Output(Amount.MinFreezeAmount, address)]
-        };
-
-        auto signature = WK.Keys.Genesis.secret.sign(hashFull(tx)[]);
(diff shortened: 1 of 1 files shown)


AndrejMitrovic (approved):

AndrejMitrovic:
Thanks a lot.

linked0 (approved):
LGTM

source/agora/test/Quorum.d
```diff
@@ -54,52 +54,26 @@ unittest
...
-            TxType.Freeze,
-            [input],
-            [Output(Amount.MinFreezeAmount, address)]
-        };
-
-        auto signature = WK.Keys.Genesis.secret.sign(hashFull(tx)[]);
-        tx.inputs[0].signature = signature;
-        return tx;
-    }
+    KeyPair[] genesisKeyPair = [WK.Keys.Genesis];
```
Geod24: Why this ?

kchulj: `split()` takes a `range` of `KeyPair`, but I only need the genesis `KeyPair`. Instead of mapping the `WK.Keys.byRange`, I am using it like this: `.split(genesisKeyPair[0].address.repeat(3))`
Do you have a suggestion on how to structure this?

source/agora/test/Quorum.d
```diff
@@ -54,52 +54,26 @@ unittest
...
     // create block with 6 payment and 2 freeze tx's
     auto txs = network.blocks[$ - 1].spendable().map!(txb => txb.sign()).array();
 
     // rewrite 3rd to last tx to multiple outputs so we can create 8 spend tx's
     // in next block
-    txs[$ - 3] = makePayTx(network.blocks[$ - 1].txs[$ - 3],
-        [WK.Keys.Genesis.address, WK.Keys.Genesis.address,
-        WK.Keys.Genesis.address]);
+    txs[$ - 3] = TxBuilder(network.blocks[$ - 1].txs[$ - 3])
+        .split(genesisKeyPair[0].address.repeat(3))
```
Geod24:
```suggestion
        .split(WK.Keys.Genesis.address.repeat(3))
```

And then you don't need the array, no ?

source/agora/test/Quorum.d
Geod24: A range of `PublicKey`, no ? And you're using `repeat` already. `only` would also work for a single element.

source/agora/test/Quorum.d
kchulj: Yes, that's better.

PR #1175, opened 2020-09-08
Make `TransactionPool.opApply` templated

kchulj:
This allows us to have proper attributes being inferred on the users, as currently the delegate has no attributes.

Split off from #1161
Relates issue #1083 
Developed with @Geod24

### DIFF
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -137,7 +137,7 @@ public class TransactionPool
             return false;
 
         // insert each input information of the transaction
-        foreach (input; tx.inputs)
+        foreach (const ref input; tx.inputs)
             this.input_set.put(input.hashFull());
 
         serializeToBuffer(tx, buffer);
@@ -162,7 +162,7 @@ public class TransactionPool
     public void remove (const ref Transaction tx) @trusted
     {
         // delete inputs of transaction from the set of Input hashes
-        foreach (ref input; tx.inputs)
+        foreach (const ref input; tx.inputs)
             this.input_set.remove(input.hashFull());
 
         auto hash = tx.hashFull();
@@ -269,7 +269,7 @@ public class TransactionPool
         if (this.hasTransactionHash(txHash))
             return false;  // double-spend
 
-        foreach (input; tx.inputs)
+        foreach (const ref input; tx.inputs)
         {
             auto hash = input.hashFull();
             if (hash in this.input_set)
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -109,7 +109,7 @@ public class TransactionPool
             "(key BLOB PRIMARY KEY, val BLOB NOT NULL)");
 
         // populate the input set from the tx'es in the DB
-        foreach (_, tx; this)
+        foreach (const ref Transaction tx; this)
         {
             foreach (const ref input; tx.inputs)
                 this.input_set.put(input.hashFull());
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
Was this split off from another PR? I thought I already saw this code.

Geod24:
You can use `Co-Authored-By: $AUTHOR` in the commit message.
But really, I was not the author, just the committer ;)
Also when splitting off PR, please add a reference, in this case #1161

source/agora/common/TransactionPool.d
```diff
@@ -197,20 +197,38 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public int opApply (scope int delegate(Hash hash, Transaction tx) dg) @trusted
+    public int opApply (DG1T) (scope DG1T dg)
```
AndrejMitrovic: Why the name `DG1T`?

Geod24: Originally because it was the version with a single parameter (as opposed to `DG2T`).

source/agora/node/Ledger.d
```diff
@@ -347,7 +347,7 @@ public class Ledger
 
         this.enroll_man.getEnrollments(data.enrolls,
             Height(this.getBlockHeight()));
-        foreach (hash, tx; this.pool)
+        foreach (ref Transaction tx; this.pool)
```
linked0, line 350:
```suggestion
        foreach (const ref Transaction tx; this.pool)
```

source/agora/node/Ledger.d
```diff
@@ -379,7 +379,7 @@ public class Ledger
         const expect_height = Height(this.getBlockHeight() + 1);
         auto utxo_finder = this.utxo_set.getUTXOFinder();
 
-        foreach (tx; data.tx_set)
+        foreach (ref tx; data.tx_set)
```
linked0, line 382:
```suggestion
        foreach (const ref tx; data.tx_set)
```

source/agora/node/Ledger.d
```diff
@@ -390,7 +390,7 @@ public class Ledger
         if (data.enrolls.length + active_enrollments < Enrollment.MinValidatorCount)
             return "Enrollment: Insufficient number of active validators";
 
-        foreach (enroll; data.enrolls)
+        foreach (ref enroll; data.enrolls)
```
linked0:
```suggestion
        foreach (const ref enroll; data.enrolls)
```

source/agora/common/TransactionPool.d
kchulj: I think it can be just `DGT` now.

kchulj:
@AndrejMitrovic yes, this was split off from #1161 for organization.
I will make sure to mention this in the comment.

source/agora/common/TransactionPool.d
Geod24: Yes

source/agora/common/TransactionPool.d
```diff
@@ -162,7 +162,7 @@ public class TransactionPool
     public void remove (const ref Transaction tx) @trusted
     {
         // delete inputs of transaction from the set of Input hashes
-        foreach (ref input; tx.inputs)
+        foreach (const ref input; tx.inputs)
```
Geod24, line 165: So technically this should be in the second commit right ?

kchulj: Correct, I'll rebase it :)

Geod24:
Windows is already a PITA.

AndrejMitrovic:
> Windows is already a PITA.

I think the issue is the timeout. The Windows build doesn't use a `parallel` loop, this slows down the tests by at least 3 minutes (on my Windows machine).

So we will need to increase the timeout.

Geod24 (approved):

source/agora/node/Ledger.d
```diff
@@ -379,7 +379,7 @@ public class Ledger
         const expect_height = Height(this.getBlockHeight() + 1);
         auto utxo_finder = this.utxo_set.getUTXOFinder();
 
-        foreach (tx; data.tx_set)
+        foreach (const ref tx; data.tx_set)
```
TrustHenry, line 382: It doesn't seem relevant to the first commit.
Move to the second commit right ?

source/agora/node/Ledger.d
```diff
@@ -390,7 +390,7 @@ public class Ledger
         if (data.enrolls.length + active_enrollments < Enrollment.MinValidatorCount)
             return "Enrollment: Insufficient number of active validators";
 
-        foreach (enroll; data.enrolls)
+        foreach (const ref enroll; data.enrolls)
```
TrustHenry, line 393: Move to the second commit right ?

kchulj: The 2nd commit was for `inputs` only and I was thinking of keeping everything else in the 1st commit.
Since this is `enroll` and not related to `tx`, it might be a good idea to move it.
What do you think?

TrustHenry: How about applying the title and contents of the second commit comprehensively in the revision?

Issue #1142, opened 2020-08-25
Blocks loaded from disk/memory are not validated when the UTXO set is rebuilt


AndrejMitrovic:
Thanks to @kchulj for finding this one. While refactoring `makeBlocks` it was found that a node will generate its (empty!) utxo set from the loaded blocks but will not do any validation on the loaded blocks:

https://github.com/bpfkorea/agora/blob/4fc127860126d78610676c92d49fdad55576aa66/source/agora/node/Ledger.d#L134-L135

So `updateUTXOSet` is called (on a block assumed to be valid), and there may be an internal assertion failure in `UTXOSet.d` if the block is not actually valid.

For defensive measures we should validate every block which is read during the regeneration of the UTXO set / Validator set.

PR #1184, opened 2020-09-09
Blocks loaded from the ledger are validated when the UTXO set is rebuilt

TrustHenry:
For defensive measures, we should validate every block which is read
during the regeneration of the UTXO set / Validator set.

Related to #1142

### DIFF
--- a/source/agora/consensus/ValidatorSet.d
+++ b/source/agora/consensus/ValidatorSet.d
@@ -162,6 +162,24 @@ public class ValidatorSet
 
     /***************************************************************************
 
+        Remove all validators from the validator set
+
+    ***************************************************************************/
+
+    public void removeAll () @trusted nothrow
+    {
+        try
+        {
+            this.db.execute("DELETE FROM validator_set");
+        }
+        catch (Exception ex)
+        {
+            log.error("Error while calling ValidatorSet.removeAll(): {}", ex);
+        }
+    }
+
+    /***************************************************************************
+
         In validatorSet DB, return the enrolled block height.
 
         Params:
@@ -577,14 +595,13 @@ unittest
     assert(keys.length == 3);
     assert(keys.isStrictlyMonotonic!("a < b"));
 
-    // remove an enrollment
+    // remove ValidatorSet
     set.remove(utxos[1]);
     assert(set.count() == 2);
     assert(set.hasEnrollment(utxos[0]));
     set.remove(utxos[0]);
     assert(!set.hasEnrollment(utxos[0]));
-    set.remove(utxos[1]);
-    set.remove(utxos[2]);
+    set.removeAll();
     assert(set.count() == 0);
 
     Enrollment[] ordered_enrollments;
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -119,37 +119,71 @@ public class Ledger
             throw new Exception("Genesis block loaded from disk is " ~
                 "different from the one in the config file");
 
-        // need to regenerate the UTXO set, starting from the genesis block
+        // need to rebuild the UTXO set and Validator set,
+        // starting from the genesis block
         if (this.utxo_set.length == 0)
         {
-            Block block;
+            ManagedDatabase.beginBatch();
+            scope (failure) ManagedDatabase.rollback();
+
+            // clear validator set
+            this.enroll_man.removeAllValidators();
+
+            Block last_read_block = gen_block;
             foreach (height; 0 .. this.last_block.header.height + 1)
             {
+                Block block;
                 this.storage.readBlock(block, Height(height));
+                if (height == 0)
+                {
+                    if (auto reason = block.isGenesisBlockInvalidReason())
+                        throw new Exception(
+                            "Genesis block loaded from disk is invalid: " ~
+                            reason);
+                }
+                else
+                {
+                    const active_enrollments = enroll_man.getValidatorCount(
+                        block.header.height);
+
+                    if (auto fail_reason = block.isInvalidReason(
+                        last_read_block.header.height,
+                        last_read_block.header.hashFull,
+                        this.utxo_set.getUTXOFinder(),
+                        active_enrollments))
+                        throw new Exception(
+                            "A block loaded from disk is invalid: " ~
+                            fail_reason);
+                }
                 this.updateUTXOSet(block);
+                this.updateValidatorSet(block);
+                last_read_block = block;
             }
+            ManagedDatabase.commitBatch();
         }
-
-        // +1 because the genesis block counts as one
-        const ulong block_count = this.last_block.header.height + 1;
-
-        // we are only interested in the last 1008 blocks,
-        // because that is the maximum length of an enrollment.
-        const Height min_height =
-            block_count >= this.params.ValidatorCycle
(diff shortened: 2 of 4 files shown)


source/agora/node/Ledger.d
```diff
@@ -129,10 +129,25 @@ public class Ledger
         if (this.utxo_set.length == 0)
         {
             Block block;
+            Block last_block = gen_block;
             foreach (height; 0 .. this.last_block.header.height + 1)
             {
                 this.storage.readBlock(block, Height(height));
+
+                if (height != 0)
```
AndrejMitrovic: You implemented `isGenesisBlockInvalidReason`. Why not use it here?

source/agora/node/Ledger.d
```diff
@@ -129,10 +129,25 @@ public class Ledger
...
+                        block.header.height);
+
+                    if (auto fail_reason = block.isInvalidReason(last_block.header.height,
+                        last_block.header.hashFull,
+                        this.utxo_set.getUTXOFinder(),
+                        active_enrollments))
+                        throw new Exception("Rejected read block: " ~ fail_reason);
+                }
                 this.updateUTXOSet(block);
+                this.updateValidatorSet(block);
```
AndrejMitrovic, line 159: On line 168 we are again rebuilding the validator set. I wonder if we can avoid doing it twice.

Additionally, @linked0 we seem to have both `restoreValidators` and `updateValidatorSet`. It seems that either one can be used, according to this code and line 172 below. Do you think we should simplify this and only have one function? Or is the code here wrong and it should use `restoreValidators`?

AndrejMitrovic (changes requested):
A few comments.

source/agora/node/Ledger.d
linked0: Right. We already are rebuilding the validator set using the latest 1008 of blocks in `restoreValidators` on line 158.  So we don't need to call the  `updateValidatorSet` function here.  And, those two functions @AndrejMitrovic said really have different operations.

AndrejMitrovic: No I think it's definitely needed here. What I'm saying is that *if* its done here, we can skip the block on line 158.

TrustHenry: `updateValidatorSetThis` may be necessary.

Because it violates block verification logic.
https://github.com/bpfkorea/agora/blob/dc869bd16b84188606f3c514b4aefb197e7c26e3/source/agora/consensus/validation/Block.d#L111-L112
I think we need to integrate with `restoreValidatorslogic` below.

linked0: @AndrejMitrovic @TrustHenry You're right. We can remove the lines from 153 to 159. And we can remove `restoreValidators` also from `EnrollmentManager`. Sorry for the confusion!

AndrejMitrovic: There are several cases we need to cover:
- utxo set is empty -> clear and rebuild both utxo set and validator set. Clearing is important because you might have an empty utxo set but a non-empty validator set
- only validator set is empty -> only rebuild validator set

So I don't think it's correct to remove that code.

source/agora/node/Ledger.d
TrustHenry: It can be used now. :)

AndrejMitrovic: Excellent.

source/agora/consensus/ValidatorSet.d
```diff
@@ -160,6 +160,25 @@ public class ValidatorSet
...
+
+        Remove all validators from the validator set
+
+    ***************************************************************************/
+
+    public void removeAll () @trusted nothrow
+    {
+        try
+        {
+            // SQLite has added a TRUNCATE optimizer to the DELETE statement.
```
AndrejMitrovic: What is this comment for? Is it saying we should use TRUNCATE? Or not? It's only confusing, maybe we can remove it..

source/agora/node/Ledger.d
```diff
@@ -125,37 +125,72 @@ public class Ledger
             throw new Exception("Genesis block loaded from disk is " ~
                 "different from the one in the config file");
 
-        // need to regenerate the UTXO set, starting from the genesis block
+        // need to rebuild the UTXO set and Validator set,
+        // starting from the genesis block
         if (this.utxo_set.length == 0)
         {
             Block block;
```
AndrejMitrovic: Please move this to line 143. It's not used outside the `foreach` loop

source/agora/node/Ledger.d
```diff
@@ -125,37 +125,72 @@ public class Ledger
...
+                if (height == 0)
+                {
+                    if (auto reason = block.isGenesisBlockInvalidReason())
+                        throw new Exception(
+                            "Genesis block loaded from disk is invalid: " ~
+                            reason);
+                }
+                else
+                {
+                    size_t active_enrollments = enroll_man.getValidatorCount(
```
```suggestion
                    const active_enrollments = enroll_man.getValidatorCount(
```

Prefer using `const` or `const size_t`, either works. If we don't modify variables it's always best to make them const.

source/agora/consensus/ValidatorSet.d
TrustHenry: Most RDBMS uses `TRUNCATE` in this case.
SQLite is information that is optimized for delete without supporting TRUNCATE.

AndrejMitrovic: Then this comment is unnecessary, it's an optimization detail that has nothing to do with Agora. Please remove it. :)

source/agora/node/Ledger.d
```diff
@@ -125,37 +125,72 @@ public class Ledger
...
-        // need to regenerate the UTXO set, starting from the genesis block
+        // need to rebuild the UTXO set and Validator set,
+        // starting from the genesis block
         if (this.utxo_set.length == 0)
         {
             Block block;
+            Block last_read_block = gen_block;
+
+            ManagedDatabase.beginBatch();
+            scope (failure) ManagedDatabase.rollback();
```
AndrejMitrovic, line 127: Interesting. I think this was a good decision. 👍

source/agora/node/Ledger.d
TrustHenry: I forgot to move it.

source/agora/node/Ledger.d
TrustHenry: Good features should be used. Thanks @AndrejMitrovic

source/agora/node/Ledger.d
```diff
@@ -119,37 +119,71 @@ public class Ledger
...
+
+            // clear validator set
+            this.enroll_man.removeAllValidators();
+
             Block block;
+            Block last_read_block = gen_block;
             foreach (height; 0 .. this.last_block.header.height + 1)
             {
                 this.storage.readBlock(block, Height(height));
```
AndrejMitrovic, line 136: @TrustHenry can you move `Block block;` inside here please?

AndrejMitrovic (approved):
Looks good to me.

source/agora/node/Ledger.d
TrustHenry: Yes, ~

linked0 (approved):

AndrejMitrovic:
Thanks for the speedy work!

Issue #1185, opened 2020-09-09
Implement `waitForPreimages (Height)` in Base.d : TestAPIManager


AndrejMitrovic:
The need for this becomes more apparent once we start requiring preimages for reaching consensus & validating block signatures.

linked0:
I will take this issue for this sprint after Brian's confirmation. The requirement of this issue is also needed for issue #1079 and I was already planning to implement the function like that as part of #1079. I think the `expectBlock` is a good reference for the implementation.

PR #1186, opened 2020-09-09
Fix one EnrollmentManager test

AndrejMitrovic:
This fixes only **one** test. However I've quickly realized we would end up having to duplicate this code segment in every test.

So instead what I suggest is to work on implementing `waitForPreimages(...)` and adding calls in all the tests. I've opened https://github.com/bpfkorea/agora/issues/1185 for resolving this. If anyone is brave enough please take a shot at working on this!

### DIFF
--- a/source/agora/test/EnrollmentManager.d
+++ b/source/agora/test/EnrollmentManager.d
@@ -48,6 +48,12 @@ unittest
     auto nodes = network.clients;
     network.expectBlock(Height(validator_cycle - 1), 2.seconds);
 
+    // wait for preimages to be revealed before making block 10
+    nodes.each!(node =>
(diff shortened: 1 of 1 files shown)


linked0 (approved):
LGTM.
The `waitForPreimages` would be needed every part of the tests related to the consensus. In fact, I also need the function because issue #1079 requires that preimages are needed for adding randomness to the weight of validators. So I am planning to implement the function as part of the #1079.

AndrejMitrovic:
@linked0 #1185 was already assigned, so please co-ordinate with Chris. Let's make sure we don't duplicate our work.

linked0:
> @linked0 #1185 was already assigned, so please co-ordinate with Chris. Let's make sure we don't duplicate our work.

OK, thanks.

PR #1187, opened 2020-09-10
Some fixups to ECC primitives

Geod24:
First two commits are fixups for #1162 and last one is an improvement I noticed while reviewing it.

### DIFF
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -135,6 +135,25 @@ public struct Scalar
         return result;
     }
 
+    /***************************************************************************
+
+        Returns:
+            the inverted scalar.
+
+        See_Also:
+            https://libsodium.gitbook.io/doc/advanced/point-arithmetic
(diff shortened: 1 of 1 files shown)


source/agora/common/crypto/ECC.d
```diff
@@ -146,11 +146,11 @@ public struct Scalar
 
     ***************************************************************************/
 
-    public Scalar invert () @trusted
+    public Scalar invert () const @nogc @trusted
     {
         Scalar scalar = this;  // copy
-        assert(crypto_core_ed25519_scalar_invert(scalar.data[].ptr,
```
AndrejMitrovic: Whoa. No invert in release mode huh?

This makes me think that we should additionally run the unittests in `-release` mode too (can we?).

AndrejMitrovic (approved):

source/agora/common/crypto/ECC.d
Geod24: We could, but realistically, I don't think we are going to use `-release` any time soon.

PR #1188, opened 2020-09-10
Implement a proof-of-concept schnorr multisignature of a block

AndrejMitrovic:
This will not pass all tests until #1185 is fixed. However it will pass some tests: `dtest=agora.test.Simple dub test`

This is not complete. There is still the issue of keeping the old key index set around for collecting old signatures. But this is a minor issue to fix.

The signature scheme is described in the last commit. It differs from what was suggested in #365, my scheme is a little bit simpler.

Anyway we should whiteboard this.

-----

The scheme is:

- Node enrolls with rand seed (`X`) and uses the nonce (`r`) to sign,
  and includes `r * G` (`R`) in the `(R, s)` signature.

From here on out `r` implies the `r` used in the signing of the enrollment. And `R` implies the public nonce in the `(R, s)` signature of the enrollment. Notice that we always reuse this `(R, r)` pair, only we add a preimage offset to it for each new block height we want to sign / validate.

At block height #1:
- `r1 = r + X` => for signing
- `R1 = R + (X * G)` => for validation

At height #2:
- `r2 = r + preimage(X)` => for signing
- `R2 = R + preimage(X) * G` => for validation

At height #3:
- `r3 = r + preimage(preimage(X))` => for signing
- `R3 = R + preimage(preimage(X)) * G` => for validation

### DIFF
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -170,7 +170,7 @@ public struct Scalar
     }
 
     /// Scalar should be greater than zero and less than L:2^252 + 27742317777372353535851937790883648493
-    public bool isValid() () nothrow @nogc @safe
+    public bool isValid () nothrow @nogc @safe const
     {
         const auto ED25519_L =  BitBlob!256("0x1000000000000000000000000000000014def9dea2f79cd65812631a5cf5d3ed");
         const auto ZERO =       BitBlob!256("0x0000000000000000000000000000000000000000000000000000000000000000");
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -70,7 +70,7 @@ public struct Scalar
     /// Internal state
     package BitBlob!(crypto_core_ed25519_SCALARBYTES * 8) data;
 
-    private this (typeof(this.data) data) @safe
+    private this (typeof(this.data) data) @safe inout
     {
         this.data = data;
     }
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -201,6 +201,13 @@ public struct Scalar
     {
         dg(this.data[]);
     }
+
+    /// Convenience overload to allow this to be converted to an `opaque_array`,
+    /// as used in the `confirm_t` statement signature in SCP
+    public const(ubyte)[] opSlice () const @safe pure nothrow @nogc
+    {
+        return this.data[];
+    }
 }
 
 // Test Scalar fromString / toString functions
--- a/source/agora/common/crypto/Schnorr.d
+++ b/source/agora/common/crypto/Schnorr.d
@@ -122,7 +122,7 @@ public Point extractNonce (Signature sig) @safe @nogc nothrow pure
 
 *******************************************************************************/
 
-package struct Sig
+public struct Sig
 {
     /// Commitment
     public Point R;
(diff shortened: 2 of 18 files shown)


AndrejMitrovic:
There was a typo in the last commit message. Fixed.

AndrejMitrovic:
I've added another commit before the last one which adds schnorr signature validation in `isInvalidReason`. It works well. I haven't added any additional tests though.

AndrejMitrovic:
Edit: Pushed the wrong branch. It should be properly updated now.

AndrejMitrovic:
I added a description of the scheme in the PR description.

AndrejMitrovic:
@hewison-chris do you feel comfortable taking this over, to help you resolve #365?

hewison-chris:
> @hewison-chris do you feel comfortable taking this over, to help you resolve #365?

Sure, I'll give it a go and when I get stuck I will ask for your help.

AndrejMitrovic:
Alright, perfect.

Some of these commits might not be really necessary. I've included them to make it easier for me to prototype.

For example, I'm not sure whether it's a good idea to store the public key associated with an enrollment. Perhaps it's enough to just use UTXOs - but during development it became difficult to use UTXOs because I needed some way of getting to the public key, so I would have to manually maintain a UTXO => PublicKey map.

Then there is the issue of having two different sets of Keys in Agora:
- https://github.com/bpfkorea/agora/blob/v0.x.x/source/agora/common/crypto/Key.d
- https://github.com/bpfkorea/agora/blob/v0.x.x/source/agora/common/crypto/ECC.d

I really hope we can just get rid of the former (is it possible?). The frequent need to convert between `PublicKey` <=> `Point` is making the code too verbose.

Then, I've also avoided using `Scalar` as the data type in the `confirm_t` SCP statement because it would mean the `scpd.*` package would need to import into `agora.common.crypto.ECC`. However we could work around this problem by moving all the Schnorr code into a separate repository. @Geod24 mentioned that it might be a good idea to move it to another repo, and I agree. If we do, then we could import it and use it directly in `confirm_t` for the signature and avoid the conversion routines from `opaque_array` <=> `Scalar`. So we wouldn't need the `Scalar: Add opSlice support` commit.

Additionally I recommend splitting of the first two commits into its own PR. These commits not controversial, and we can already move forward with signing envelopes - and leave out the signing of `confirm_t` statements to a separate PR. There's a commented out unittest in those commits that would need to be fixed.

Let me know if you need any help, let's work as a team.

AndrejMitrovic:
I've implemented the idea I mentioned in https://github.com/bpfkorea/agora/pull/1206#issuecomment-693763789 in this PR as a proof-of-concept. Check out the last two commits.

AndrejMitrovic:
One thing we need to be careful about is nonce bias attacks. There's a few interesting papers & presentations about it:

- https://ecc2017.cs.ru.nl/slides/ecc2017-tibouchi.pdf
- https://eprint.iacr.org/2019/023.pdf - Attacks on ECDSA for R where R is < 32 bytes, and also for R =32 byte but is biased in some of its bits. Please note that for Schnorr this attack is allegedly even easier.
- https://www.monerooutreach.org/monero-konferenco/andrew-poelstra.html

This quote is relevant:

> partial info on k the nonce (e.g. N known bits) should translate to partial info on x the private key (N bits of info)

Meaning: if the nonce is biased and some of its bits often remain the same, you can extract the same bits from the private key.

We should at least try to analyze the generated R's and plot them. And of course, we should get an expert to review our Schnorr implementation & signature scheme.

AndrejMitrovic:
An even better write-up is here: https://medium.com/blockstream/musig-dn-schnorr-multisignatures-with-verifiably-deterministic-nonces-27424b5df9d6

AndrejMitrovic:
And just to be a little bit more specific here: We do (under our block signature scheme) want to force validators to never sign two different blocks at the same height. Otherwise they would reveal their private key. This is something we want. But we don't want the private key to leak due to an algorithm which produces biased nonces over time.

AndrejMitrovic:
Closing in favor of #1349.

PR #1190, opened 2020-09-11
Implement `waitPreimages(distance)`

linked0:
This PR is for #1185 of which requirements are adding the utility function checking pre-images for reaching consensus & validating block signatures. In fact, the signature of the function described in the issue has a parameter for `Height`, but using the `distance` is a simple way to implement. If the function has a `Height` as a parameter, it also has to have another `Height` for an enrolled height. And I think we might have another approach I could come up with if needed.

@AndrejMitrovic Please check if the code meets the requirement. It's a draft PR for sharing the concept. If any other opinions, please let me know.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -519,6 +519,38 @@ public class TestAPIManager
 
     /***************************************************************************
 
+        Checks if all the nodes contain the given distance of pre-images for
+        the given enrollments.
+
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
I guess we are forced to keep track of all enrollments in order to use this API. But I don't see a way to work around this, so it's good like it is now. 👍

AndrejMitrovic (approved):

PR #1192, opened 2020-09-11
Remove quorum hashes from SCP protocol messages

### LINKED ISSUE #982: Remove quorum set hashes from SCP protocol messages
In our quorum algorithm design the quorum is always known for all the active validator nodes in the system.

Therefore sending the quorum hash in each SCP message becomes superfluous. Removing it should save quite a bit on network bandwidth too.

This should be fairly simple to do because SCP already abstracts this away in the `getQuorumSetFromStatement` function.

AndrejMitrovic:
The quorum hashes can be implicitly derived from the `node_id` and `slot_idx` fields which are part of every `SCPEnvelope` message.

This reduces the total network overhead.

Fixes #982

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -125,13 +125,13 @@ extern(D):
         // store the list of other node's quorum hashes
         foreach (qc; other_quorums)
         {
-            auto quorum_set = buildSCPConfig(qc);
+            auto quorum_set = buildSCPQuorumSet(qc);
             auto shared_set = makeSharedSCPQuorumSet(quorum_set);
             this.known_quorums[hashFull(quorum_set)] = shared_set;
         }
 
         // set up our own quorum
-        auto quorum_set = buildSCPConfig(quorum);
+        auto quorum_set = buildSCPQuorumSet(quorum);
         () @trusted { this.scp.updateLocalQuorumSet(quorum_set); }();
         auto shared_set = makeSharedSCPQuorumSet(quorum_set);
         this.known_quorums[hashFull(quorum_set)] = shared_set;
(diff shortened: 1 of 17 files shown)


TrustHenry:
I think the idea is very good.
Access to the quorum set has become more familiar.

TrustHenry:
However, quorumSetHash (D) removed,  so we need to see more of the effects.

AndrejMitrovic:
> we need to see more of the effects.

I'm not sure what you mean by this.

The quorum set can now be looked up through the Node ID. `getQSet (Hash)` has been replaced with `getNodeQSet (PublicKey)`, which SCP calls when it needs to find the quorum set of a given message.

Geod24:
I like it, but that's a rather large modification of SCP. As we move in that direction, it's going to become harder to integrate upstream change. I have an idea to deal with it, just need to sit and implement it (hint: `patch`).

AndrejMitrovic:
Yes that's one of the problems..

This doesn't need to be merged right now. I tagged it as CoinNet because it would be good to have around that milestone.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -105,36 +105,29 @@ extern(D):
...
         assert(!this.is_nominating);
-        () @trusted { this.known_quorums.clear(); }();
+        () @trusted { this.quorum_set.clear(); }();
 
-        // store the list of other node's quorum hashes
-        foreach (qc; other_quorums)
+        foreach (qpair; quorums.byKeyValue)  // opApply2 is not nothrow..
         {
-            auto quorum_set = buildSCPConfig(qc);
+            auto quorum_set = buildSCPConfig(qpair.value);
```
linked0: Two suggestions about the code.
1. This name of `quroum_set` is confusing with `this.quorum_set` , so it should have another name like `temp_qset`.
2. The name of `buildSCPQuorumSet` would more representative for the operation than `buildSCPConfig`. How about changing the name?

linked0 (approved):
Nice! It makes the code more compact by removing the hash.
I leave a request for some modifications.

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: Good point about the naming. I was thinking the same thing, but forgot to rename it. Good suggestion about the rename too.

Geod24:
We should still do this, but this PR is now quite outdated, so closing.

PR #1199, opened 2020-09-14
Implement registry to be able to lookup network addresses

### LINKED ISSUE #1173: Implement network registry server
We need a network registry server to allow nodes to be findable based on their public key.

Requirements:
- Allow a node to update its IP address / name using a message signed by its private key;
- Allow any client to look up a node based on an IP address;
- Integrate with the DNS seed feature;

Additionally:
- Having a way to register a full node would be nice to have (but can be moved to another issue);
- It needs to integrate with the existing testsuite (localrest, integration);
- It should probably be a separate project that can be integrated into Agora;
- D is preferred but not strictly mandatory;
- Pointer: https://github.com/sipa/bitcoin-seeder

[Existing whiteboarding session](https://drive.google.com/file/d/1aqhuT1M_N_nMnkcS6dokhuNWscQN-Z2Q/view?usp=sharing)

ferencdg:
fixes #1173

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -52,6 +52,13 @@ validator:
   # Private seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
   # Public address:  GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
   seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  # Address of the name registry
+  registry_address: http://127.0.0.1:3003
+  # Network addresses that will be registered with the public key (Validator only)
+  # If left empty, all public network addresses of the node will be registered
+  addresses_to_register:
+    - 88.88.88.88
+    - best.validator.io
 
 ################################################################################
 ##                         Ban manager configuration                          ##
--- a/dub.json
+++ b/dub.json
@@ -67,6 +67,7 @@
     "dflags": [ "-extern-std=c++14" ],
     "lflags-posix": [ "-lstdc++", "-lsqlite3" ],
     "lflags-windows": [  "sqlite3.lib", "/nodefaultlib:msvcetd.lib" ],
+    "libs-windows": [ "iphlpapi" ],
     "buildRequirements": [ "allowWarnings" ],
 
 	"toolchainRequirements": {
@@ -80,7 +81,13 @@
             "name": "server",
             "targetName": "agora",
             "mainSourceFile": "source/agora/node/main.d",
-            "excludedSourceFiles": ["source/agora/cli/*/*"]
+            "excludedSourceFiles": ["source/agora/cli/*/*", "source/agora/registry/main.d"]
+        },
+        {
+            "name": "name-registry",
+            "targetName": "name-registry",
+            "mainSourceFile": "source/agora/registry/main.d",
+            "excludedSourceFiles": ["source/agora/cli/*/*", "source/agora/node/main.d"]
         },
         {
             "name": "client",
@@ -102,13 +109,14 @@
                 "source/agora/consensus/SCPEnvelopeStore.d",
                 "source/agora/cli/checkvtable/*",
                 "source/agora/cli/multi/*",
-                "source/agora/cli/vanity/*"
+                "source/agora/cli/vanity/*",
+                "source/agora/registry/main.d"
             ]
         },
         {
             "name": "unittest",
             "targetName": "agora-unittests",
-            "excludedSourceFiles": ["source/agora/cli/*/main.d"],
+            "excludedSourceFiles": ["source/agora/cli/*/main.d", "source/agora/registry/main.d"],
             "sourceFiles-posix": [
                 "source/scpp/build/DSizeChecks.o",
                 "source/scpp/build/DLayoutChecks.o"
@@ -127,7 +135,8 @@
                 "source/agora/cli/checkvtable/*",
                 "source/agora/cli/client/*",
                 "source/agora/cli/vanity/*",
-                "source/agora/node/main.d"
+                "source/agora/node/main.d",
+                "source/agora/registry/main.d"
             ]
         },
         {
@@ -141,7 +150,8 @@
                 "source/agora/cli/client/*",
                 "source/agora/cli/multi/*",
                 "source/agora/cli/vanity/*",
-                "source/agora/node/main.d"
+                "source/agora/node/main.d",
+                "source/agora/registry/main.d"
             ]
         },
         {
@@ -156,7 +166,8 @@
                 "source/agora/cli/client/*",
                 "source/agora/cli/multi/*",
                 "source/agora/cli/vanity/*",
-                "source/agora/node/main.d"
+                "source/agora/node/main.d",
+                "source/agora/registry/main.d"
             ]
         }
     ],
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -280,38 +280,39 @@ network:
     }
 }
 
+///
+private const(string)[] parseSequence (string section, ref const CommandLine cmdln, Node root, bool optional = false)
+{
+    if (auto val = section in cmdln.overrides)
+        return *val;
+
+    if (auto node = section in root)
+        enforce(root[section].type == NodeType.sequence,
+            format("`%s` section must be a sequence", section));
+    else if (optional)
+        return null;
+    else
+        throw new Exception(
+            format("The '%s' section is mandatory and must " ~
+                "specify at least one item", section));
+
+    string[] result;
+    foreach (string item; root[section])
+        result ~= item;
+
+    return result;
+}
+
 /// ditto
 private Config parseConfigImpl (ref const CommandLine cmdln, Node root)
 {
-    const(string)[] parseSequence (string section, bool optional = false)
-    {
-        if (auto val = section in cmdln.overrides)
-            return *val;
-
-        if (auto node = section in root)
-            enforce(root[section].type == NodeType.sequence,
-                format("`%s` section must be a sequence", section));
-        else if (optional)
-            return null;
-        else
-            throw new Exception(
-                format("The '%s' section is mandatory and must " ~
-                    "specify at least one item", section));
-
-        string[] result;
-        foreach (string item; root[section])
-            result ~= item;
-
-        return result;
-    }
-
     Config conf =
     {
         banman : parseBanManagerConfig("banman" in root, cmdln),
         node : parseNodeConfig("node" in root, cmdln),
         validator : parseValidatorConfig("validator" in root, cmdln),
-        network : assumeUnique(parseSequence("network")),
-        dns_seeds : assumeUnique(parseSequence("dns", true)),
+        network : assumeUnique(parseSequence("network", cmdln, root)),
+        dns_seeds : assumeUnique(parseSequence("dns", cmdln, root, true)),
         logging: parseLoggingSection("logging" in root, cmdln),
         event_handlers: parserEventHandlers("event_handlers" in root, cmdln),
     };
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -169,6 +169,12 @@ public struct ValidatorConfig
 
     /// The seed to use for the keypair of this node
     public immutable KeyPair key_pair;
+
+    // Network addresses that will be registered with the public key (Validator only)
+    public immutable string[] addresses_to_register;
+
+    // Registry address
+    public string registry_address;
 }
 
 /// Admin API config
@@ -281,7 +287,8 @@ network:
 }
 
 ///
-private const(string)[] parseSequence (string section, ref const CommandLine cmdln, Node root, bool optional = false)
+private const(string)[] parseSequence (string section, ref const CommandLine cmdln,
+        Node root, bool optional = false)
 {
     if (auto val = section in cmdln.overrides)
         return *val;
@@ -421,10 +428,13 @@ private ValidatorConfig parseValidatorConfig (Node* node, const ref CommandLine 
     if (!enabled)
         return ValidatorConfig(false);
 
+    auto registry_address = get!(string, "validator", "registry_address")(cmdln, node);
     ValidatorConfig result = {
         enabled: true,
         key_pair:
             KeyPair.fromSeed(Seed.fromString(get!(string, "validator", "seed")(cmdln, node))),
+        registry_address: registry_address,
+        addresses_to_register : assumeUnique(parseSequence("addresses_to_register", cmdln, *node, true)),
     };
     return result;
 }
@@ -440,6 +450,7 @@ unittest
 validator:
   enabled: true
   seed: SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  registry_address: http://127.0.0.1:3003
 `;
         auto node = Loader.fromString(conf_example).load();
         auto config = parseValidatorConfig("validator" in node, cmdln);
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -39,7 +39,10 @@ import agora.consensus.data.Transaction;
 import agora.network.Clock;
 import agora.network.NetworkClient;
 import agora.node.Ledger;
+import agora.registry.NameRegistryAPI;
+import agora.utils.InetUtils;
 import agora.utils.Log;
+import agora.utils.Utility;
 
 import scpd.types.Stellar_SCP;
 
@@ -318,6 +321,9 @@ public class NetworkManager
     /// Config instance
     protected const NodeConfig node_config = NodeConfig.init;
 
+    /// Validator instance
+    protected const ValidatorConfig validator_config = ValidatorConfig.init;
+
     /// Task manager
     private TaskManager taskman;
 
@@ -362,18 +368,20 @@ public class NetworkManager
     /// Clock instance
     protected Clock clock;
 
+    /// Registry client
+    private NameRegistryAPI registry_client;
+
     /// Maximum connection tasks to run in parallel
     private enum MaxConnectionTasks = 10;
 
     /// Ctor
-    public this (in NodeConfig node_config, in BanManager.Config banman_conf,
-        in string[] seed_peers, in string[] dns_seeds, Metadata metadata,
-        TaskManager taskman, Clock clock)
+    public this (in Config config, Metadata metadata, TaskManager taskman, Clock clock)
     {
         this.taskman = taskman;
-        this.node_config = node_config;
+        this.node_config = config.node;
+        this.validator_config = config.validator;
         this.metadata = metadata;
-        this.banman = this.getBanManager(banman_conf, clock,
+        this.banman = this.getBanManager(config.banman, clock,
             node_config.data_dir);
         this.discovery_task = new AddressDiscoveryTask(&this.addAddresses);
         this.clock = clock;
@@ -390,11 +398,11 @@ public class NetworkManager
         else
         {
             // add the IP seeds
-            this.addAddresses(Set!Address.from(seed_peers));
+            this.addAddresses(Set!Address.from(config.network));
 
             // add the DNS seeds
-            if (dns_seeds.length > 0)
-                this.addAddresses(resolveDNSSeeds(dns_seeds));
+            if (config.dns_seeds.length > 0)
+                this.addAddresses(resolveDNSSeeds(config.dns_seeds));
         }
     }
 
@@ -553,6 +561,20 @@ public class NetworkManager
         return false;
     }
 
+    /// Periodically registers network addresses
+    public void startPeriodicNameRegistration ()
+    {
+        this.registry_client = this.getNameRegistryClient(
+            validator_config.registry_address, 2.seconds);
+        if (this.registry_client is null)
+            return;
+        this.onRegisterName();  // avoid delay
+        // we re-register in every 2 minutes, in order to cope with the situation below
+        // 1. network registry server is restarted
+        // 2. client running agora node acquired some new IPs
+        this.taskman.setTimer(2.minutes, &this.onRegisterName, Periodic.Yes);
+    }
+
     /// Discover the network, connect to all required peers
     /// Some nodes may want to connect to specific peers before
     /// discovery() is considered complete
@@ -564,6 +586,25 @@ public class NetworkManager
         this.required_peer_keys = Set!PublicKey.from(
             required_peer_keys.byKey()
             .filter!(key => key !in this.connected_validator_keys));
+
+        if (registry_client !is null)
+            foreach (key; this.required_peer_keys)
+                taskman.runTask
+                ({
+                    retry!
+                    ({
+                        auto registryPayload = registry_client.getValidator(key);
+                        if (!registryPayload.verifySignature(key))
+                        {
+                            log.warn("RegistryPayload signature is incorrect for {}", key);
+                            return false;
+                        }
+                        foreach (addr; registryPayload.data.addresses)
+                            this.addAddress(addr);
+                        return true;
+                    },
+                    )(taskman, 3, 2.seconds, "Exception happened while trying to get validator addresses");
+                });
 
         // actually just runs it once, but we need the scheduler to run first
         // and it doesn't run in the constructor yet (LocalRest)
@@ -723,6 +764,36 @@ public class NetworkManager
         return new BanManager(banman_conf, clock, data_dir);
     }
 
+    /// register network addresses into the name registry
+    private void onRegisterName ()
+    {
+        const(Address)[] addresses = this.validator_config.addresses_to_register;
+        if (!addresses.length)
+            addresses = InetUtils.getPublicIPs();
+
+        RegistryPayload payload =
+        {
+            data:
+            {
+                public_key : this.validator_config.key_pair.address,
+                addresses : addresses,
+                seq : time(null)
+            }
+        };
+
+        payload.signPayload(this.validator_config.key_pair.secret);
+
+        try
+        {
+            this.registry_client.putValidator(payload);
+        }
+        catch (Exception ex)
+        {
+            log.info("Couldn't register our address: {}. Trying again later..",
+                ex);
+        }
+    }
+
     /***************************************************************************
 
         Retrieve blocks starting from block_height up to the highest block
@@ -856,6 +927,38 @@ public class NetworkManager
         settings.httpClientSettings.readTimeout = timeout;
 
         return new RestInterfaceClient!API(settings);
+    }
+
+    /***************************************************************************
+
+        Instantiates a client object implementing `NameRegistryAPI`
+
+        This function simply returns a name registry object implementing
+        `NameRegistryAPI`. In the default implementation, this returns a
+        `RestInterfaceClient`. However, it can be overriden in test code to
+        return an in-memory client.
+
+        Params:
+          address = The address of the name registry server
+          timeout = the timeout duration to use for requests
+
+        Returns:
+          An object to communicate with the name registry server
+
+    ***************************************************************************/
+
+    public NameRegistryAPI getNameRegistryClient (Address address, Duration timeout)
+    {
+        import vibe.http.client;
+        if (address=="disabled")
+            return null;
+        auto settings = new RestInterfaceSettings();
+        settings.baseURL = URL(address);
+        settings.httpClientSettings = new HTTPClientSettings();
+        settings.httpClientSettings.connectTimeout = timeout;
+        settings.httpClientSettings.readTimeout = timeout;
+
+        return new RestInterfaceClient!NameRegistryAPI(settings);
     }
 
     /***************************************************************************
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -35,6 +35,7 @@ import agora.network.NetworkManager;
 import agora.node.BlockStorage;
 import agora.node.FullNode;
 import agora.node.Ledger;
+import agora.registry.NameRegistryAPI;
 import agora.utils.Log;
 import agora.utils.PrettyPrinter;
 
@@ -211,6 +212,7 @@ public class Validator : FullNode, API
     public override void start ()
     {
         this.started = true;
+        this.network.startPeriodicNameRegistration();
         this.startPeriodicDiscovery();
         this.clock.startSyncing();
         this.taskman.setTimer(this.config.node.preimage_reveal_interval,
--- a/source/agora/registry/NameRegistryAPI.d
+++ b/source/agora/registry/NameRegistryAPI.d
@@ -1,0 +1,99 @@
+/*******************************************************************************
+
+    Definitions of the name registry API
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.registry.NameRegistryAPI;
+
+import agora.common.Hash;
+import agora.common.Types;
+import agora.common.crypto.Key;
+
+import vibe.http.common;
+import vibe.web.rest;
+
+///
+public struct RegistryPayloadData
+{
+    /// the public key that we want to register
+    public PublicKey public_key;
+
+    /// network addresses associated with the public key
+    public const(Address)[] addresses;
+
+    /// monotonically increasing sequence number
+    public ulong seq;
+}
+
+///
+public struct RegistryPayload
+{
+    ///
+    public RegistryPayloadData data;
+
+    /// signature over the `data` member
+    public Signature signature;
+
+    ///
+    public void signPayload(const ref SecretKey secret_key) nothrow
+    {
+        signature = secret_key.sign(hashFull(data)[]);
+    }
+
+    ///
+    public bool verifySignature(const ref PublicKey public_key) const nothrow @nogc @safe
+    {
+        return public_key.verify(this.signature, hashFull(this.data)[]);
+    }
+}
+
+/// API allowing to store network addresses under a public key
+public interface NameRegistryAPI
+{
+    @safe:
+
+    /***************************************************************************
+
+        Get network addresses corresponding to a public key
+
+        Params:
+            public_key = the public key that was used to register
+                         the network addresses
+
+        Returns:
+            Network addresses associated with the `public_key`
+
+        API:
+            GET /validator
+
+    ***************************************************************************/
+
+    public const(RegistryPayload) getValidator (PublicKey public_key);
+
+    /***************************************************************************
+
+        Register network addresses corresponding to a public key
+
+        Params:
+            registry_payload =
+                the data we want to register with the name registry server
+
+        Returns:
+            empty string, if the registration was successful, otherwise returns
+            the error message
+
+        API:
+            PUT /validator
+
+    ***************************************************************************/
+
+    public void putValidator (RegistryPayload registry_payload);
+}
--- a/source/agora/registry/NameRegistryImpl.d
+++ b/source/agora/registry/NameRegistryImpl.d
@@ -1,0 +1,90 @@
+/*******************************************************************************
+
+    Definitions of the name registry API Implementation
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.registry.NameRegistryImpl;
+
+import agora.common.Hash;
+import agora.common.Types;
+import agora.common.crypto.Key;
+import agora.registry.NameRegistryAPI;
+import agora.utils.Log;
+
+import vibe.core.core;
+import vibe.http.common;
+import vibe.web.rest;
+
+mixin AddLogger!();
+
+/// Implementation of `NameRegistryAPI` using associative arrays
+public final class NameRegistryImpl: NameRegistryAPI
+{
+    ///
+    private RegistryPayload[PublicKey] registry_map;
+
+    /***************************************************************************
+
+        Get network addresses corresponding to a public key
+
+        Params:
+            public_key = the public key that was used to register
+                         the network addresses
+
+        Returns:
+            Network addresses associated with the `public_key`
+
+        API:
+            GET /validator
+
+    ***************************************************************************/
+
+    public override const(RegistryPayload) getValidator (PublicKey public_key)
+    {
+        if (auto payload = public_key in registry_map)
+            return *payload;
+        return RegistryPayload.init;
+    }
+
+    /***************************************************************************
+
+        Register network addresses corresponding to a public key
+
+        Params:
+            registry_payload =
+                the data we want to register with the name registry server
+
+        Returns:
+            empty string, if the registration was successful, otherwise returns
+            the error message
+
+        API:
+            PUT /validator
+
+    ***************************************************************************/
+
+    public override void putValidator (RegistryPayload registry_payload)
+    {
+        // verify signature
+        if (!registry_payload.verifySignature(registry_payload.data.public_key))
+            throw new Exception("incorrect signature");
+
+        // check if we received stale data
+        if (auto previous = registry_payload.data.public_key in registry_map)
+            if (previous.data.seq > registry_payload.data.seq)
+                throw new Exception("registry already has a more up-to-date version of the data");
+
+        // register data
+        log.info("Registering network addresses: {} for public key: {}", registry_payload.data.addresses,
+            registry_payload.data.public_key.toString());
+        registry_map[registry_payload.data.public_key] = registry_payload;
+    }
+}
--- a/source/agora/registry/main.d
+++ b/source/agora/registry/main.d
@@ -1,0 +1,76 @@
+/*******************************************************************************
+
+    Standalone server to implement a name registry
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.registry.main;
+
+import agora.registry.NameRegistryAPI;
+import agora.registry.NameRegistryImpl;
+
+import vibe.core.core;
+import vibe.http.router;
+import vibe.http.server;
+import vibe.web.rest;
+
+import std.conv;
+import std.getopt;
+import std.stdio;
+
+///
+struct CommandlineArgs
+{
+    string bind_address = "0.0.0.0";
+    ushort bind_port = 3003;
+}
+
+/// Parse the command-line arguments and return a GetoptResult
+public GetoptResult parseCommandLine (ref CommandlineArgs cmdline_args, string[] args)
+{
+    return getopt(
+        args,
+        "bind-host|h",
+            "Address where the name register server will bind to, defaults to: " ~ CommandlineArgs.init.bind_address,
+            &cmdline_args.bind_address,
+        "bind-port|p",
+            "Port where the name register server will bind to, defaults to: " ~ to!string(CommandlineArgs.init.bind_port),
+            &cmdline_args.bind_port,
+        );
+}
+
+///
+private int main (string[] args)
+{
+    CommandlineArgs cmdline_args;
+    try
+    {
+        auto help = parseCommandLine(cmdline_args, args);
+        if (help.helpWanted)
+        {
+            defaultGetoptPrinter("Name Registry Server", help.options);
+            return 0;
+        }
+    }
+    catch (Exception ex)
+    {
+        writefln("Error parsing command-line arguments '%(%s %)': %s", args, ex.message);
+        return -1;
+    }
+
+    auto router = new URLRouter();
+    router.registerRestInterface(new NameRegistryImpl());
+
+    auto settings = new HTTPServerSettings;
+    settings.port = cmdline_args.bind_port;
+    settings.bindAddresses = [cmdline_args.bind_address];
+    listenHTTP(settings, router);
+    return runEventLoop();
+}
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -53,6 +53,8 @@ import agora.node.BlockStorage;
 import agora.node.FullNode;
 import agora.node.Ledger;
 import agora.node.Validator;
+import agora.registry.NameRegistryAPI;
+import agora.registry.NameRegistryImpl;
 import agora.utils.Log;
 import agora.utils.PrettyPrinter;
 public import agora.utils.Utility : retryFor;
@@ -513,6 +515,9 @@ public class TestAPIManager
     /// Registry holding the nodes
     protected Registry reg;
 
+    /// Name registry
+    private RemoteAPI!NameRegistryAPI name_registry;
+
     ///
     public this (immutable(Block)[] blocks, TestConf test_conf,
         time_t genesis_start_time)
@@ -522,6 +527,7 @@ public class TestAPIManager
         this.genesis_start_time = genesis_start_time;
         this.initial_time = this.getBlockTime(blocks[$ - 1].header.height);
         this.reg.initialize();
+        this.createNameRegistry();
     }
 
     /***************************************************************************
@@ -681,6 +687,18 @@ public class TestAPIManager
 
     /***************************************************************************
 
+        Create a new name registry
+
+    ***************************************************************************/
+
+    public void createNameRegistry ()
+    {
+        this.name_registry = RemoteAPI!NameRegistryAPI.spawn!NameRegistryImpl();
+        this.reg.register("name.registry", this.name_registry.tid());
+    }
+
+    /***************************************************************************
+
         Start each of the nodes
 
         Params:
@@ -734,6 +752,9 @@ public class TestAPIManager
         }
 
         this.nodes = null;
+
+        enforce(this.reg.unregister("name.registry"));
+        name_registry.ctrl.shutdown();
     }
 
     /***************************************************************************
@@ -787,11 +808,10 @@ public class TestAPIManager
 
     ***************************************************************************/
 
-    public void waitForDiscovery (string file = __FILE__, size_t line = __LINE__)
+    public void waitForDiscovery (Duration timeout = 5.seconds, string file = __FILE__, size_t line = __LINE__)
     {
         try
         {
-            const timeout = 5.seconds;
             this.nodes.each!(node =>
                 retryFor(node.client.getNodeInfo().ifThrown(NodeInfo.init)
                     .state == NetworkState.Complete,
@@ -864,12 +884,11 @@ public class TestNetworkManager : NetworkManager
     public Registry* registry;
 
     /// Constructor
-    public this (NodeConfig config, BanManager.Config ban_conf,
-        in string[] peers, in string[] dns_seeds, Metadata metadata,
+    public this (Config config, Metadata metadata,
         TaskManager taskman, Clock clock, Registry* reg)
     {
         this.registry = reg;
-        super(config, ban_conf, peers, dns_seeds, metadata, taskman, clock);
+        super(config, metadata, taskman, clock);
     }
 
     /// No "http://" in unittests, we just use the string as-is
@@ -886,6 +905,16 @@ public class TestNetworkManager : NetworkManager
         if (tid != typeof(tid).init)
             return new RemoteAPI!TestAPI(tid, timeout);
         assert(0, "Trying to access node at address '" ~ address ~
+               "' without first creating it");
+    }
+
+    ///
+    public override RemoteAPI!NameRegistryAPI getNameRegistryClient (Address address, Duration timeout)
+    {
+        auto tid = this.registry.locate("name.registry");
+        if (tid != typeof(tid).init)
+            return new RemoteAPI!NameRegistryAPI(tid, timeout);
+        assert(0, "Trying to access name registry at address '" ~ address ~
                "' without first creating it");
     }
 
@@ -1105,13 +1134,10 @@ private mixin template TestNodeMixin ()
 
     /// Return an instance of the custom TestNetworkManager
     protected override NetworkManager getNetworkManager (
-        in NodeConfig node_config, in BanManager.Config banman_conf,
-        in string[] peers, in string[] dns_seeds, Metadata metadata,
-        TaskManager taskman, Clock clock)
+        in Config config, Metadata metadata, TaskManager taskman, Clock clock)
     {
         assert(taskman !is null);
-        return new TestNetworkManager(node_config, banman_conf, peers,
-            dns_seeds, metadata, taskman, clock, this.registry);
+        return new TestNetworkManager(config, metadata, taskman, clock, this.registry);
     }
 
     /// Return an enrollment manager backed by an in-memory SQLite db
@@ -1388,6 +1414,9 @@ public struct TestConf
     /// max listener nodes. If set to 0, set to this.nodes - 1
     size_t max_listeners;
 
+    /// registry address
+    string registry_address = "name.registry";
+
     /// Number of transactions nominated for each nomination slot.
     /// This is only used for the TestNominator - it's not part of Consensus rules.
     /// Many existing tests have been originally written with the assumption that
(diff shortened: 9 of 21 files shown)


source/agora/node/Validator.d
```diff
@@ -86,6 +94,58 @@ public class Validator : FullNode, API
         // currently we are not saving preimage info,
         // we only have the commitment in the genesis block
         this.regenerateQuorums(Height(0));
+
+        if(config.node.registry_address.length && config.addresses_to_register.length)
```
AndrejMitrovic: Any usage of vibe.d should be wrapped so we can use LocalRest during unittest runs and only use vibe.d in non-unittest builds. That's one of the items in #1173:

> It needs to integrate with the existing testsuite (localrest, integration);

There should likely be something like `getClient`, perhaps called `getNameRegistryClient` in the `NetworkManager` which by defaults instantiates a `RestInterfaceClient!NameRegistryAPI`, but is overriden in unittests in `Base.d` to provide a non-network name registry node - similar to how it's done [here](https://github.com/bpfkorea/agora/blob/0a45c1359a9c819acae588aae25036f00c33e4af/source/agora/test/Base.d#L858). That code provides a LocalRest-based interface to the node, while the node itself is spawned for example [here](https://github.com/bpfkorea/agora/blob/0a45c1359a9c819acae588aae25036f00c33e4af/source/agora/test/Base.d#L638). For the name server node we could spawn it just once, for example somewhere inside of [here](https://github.com/bpfkorea/agora/blob/0a45c1359a9c819acae588aae25036f00c33e4af/source/agora/test/Base.d#L1338) and change the node's `registry_address` to match the same string as set with a call to `reg.register` like [here](https://github.com/bpfkorea/agora/blob/0a45c1359a9c819acae588aae25036f00c33e4af/source/agora/test/Base.d#L654).

A bit of pseudo-code:

```D
Base.d:

class TestAPIManager
{
    void makeNameRegistry ()
    {
        // TestNameRegistry would be the actual class which implements NameRegistryAPI,
        // just using a hashmap internally
        auto tid = RemoteAPI!NameRegistryAPI.spawn!TestNameRegistry(ctor_args...);
        this.reg.register("name.registry", api.tid());
    }
}

class TestNetworkManager
{
    /// get a localrest-backed client
    TestAPI getNameRegistryClient (Address address,
        Duration timeout)
    {
        auto tid = this.registry.locate(address);
        if (tid != typeof(tid).init)
            return new RemoteAPI!getNameRegistryAPI(tid, timeout);
        assert(0, "Trying to access node at address '" ~ address ~
               "' without first creating it");
    }
}

function makeTestNetwork ()
{
    // for each node's config set its `registry_address` to "name.registry" as set above
    foreach (node; nodes..)
        ...
    
    network.makeNameRegistry();
}
```

AndrejMitrovic (changes requested):

source/agora/node/Validator.d
ferencdg: > ```d
> makeTestNetwork
> ```

thanks, I will modify the code according to this

source/agora/common/Config.d
```diff
@@ -404,13 +413,16 @@ node:
 private ValidatorConfig parseValidatorConfig (Node* node, const ref CommandLine cmdln)
 {
     const enabled = get!(bool, "validator", "enabled")(cmdln, node);
-    if (!enabled)
+    if (!enabled || node is null)
```
AndrejMitrovic: This looks like a bugfix? Should it be in its own commit?

source/agora/common/crypto/Key.d
```diff
@@ -104,7 +104,7 @@ public struct PublicKey
     /// Alias to the BitBlob type
     private alias DataType = BitBlob!(crypto_sign_ed25519_PUBLICKEYBYTES * 8);
 
-    /*private*/ DataType data;
+    DataType data;
```
AndrejMitrovic: Unrelated change. Actually the comment was there because we want to make it private, but due to current compiler limitations we can't.

In any case please try not to mix unrelated styling changes and semantic changes into the same commit.

source/agora/network/NetworkManager.d
```diff
@@ -718,6 +719,39 @@ public class NetworkManager
...
+        import vibe.http.client;
+
+        auto rest_interface_settings = new RestInterfaceSettings();
+        rest_interface_settings.baseURL = URL(address);
+        rest_interface_settings.httpClientSettings = new HTTPClientSettings();
+        with(rest_interface_settings.httpClientSettings)
+        {
+            connectTimeout = timeout;
+            readTimeout = timeout;
+        }
```
AndrejMitrovic: You won't need the `with` if you keep the code simple just like in `getClient`.

```suggestion
        auto settings = new RestInterfaceSettings();
        settings.baseURL = URL(address);
        settings.httpClientSettings = new HTTPClientSettings();
        settings.httpClientSettings.connectTimeout = timeout;
        settings.httpClientSettings.readTimeout = timeout;
```

source/agora/node/Validator.d
```diff
@@ -91,6 +99,50 @@ public class Validator : FullNode, API
         // currently we are not saving preimage info,
         // we only have the commitment in the genesis block
         this.regenerateQuorums(Height(0));
+
+        if(config.validator.registry_address.length && config.validator.addresses_to_register.length)
```
AndrejMitrovic: This whole bit should be moved into its own function and called from the constructor.

source/agora/common/Config.d
```diff
@@ -324,12 +308,37 @@ private Config parseConfigImpl (ref const CommandLine cmdln, Node root)
     return conf;
 }
 
+///
+private const(string)[] parseSequence (string section, ref const CommandLine cmdln,
```
AndrejMitrovic: I'm not sure why the function was moved, but this changeset should likely need to be moved into its own commit.

doc/config.example.yaml
```diff
@@ -52,6 +52,13 @@ validator:
   # Private seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
   # Public address:  GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
   seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  # Address of the name registry
+  registry_address: http://127.0.0.1:3003
+# Network addresses that will be registered with the public key (Validator only)
+# If left empty, all public network addresses of the node will be registered
```
Geod24: Indentation

doc/config.example.yaml
```diff
@@ -52,6 +52,13 @@ validator:
...
+  addresses_to_register:
+    - 88.88.88.88
+    - 77.77.77.77
```
Geod24: Perhaps replace one of the IP with a DNS name to show it's supported ?

dub.json
```diff
@@ -66,7 +66,8 @@
 
     "dflags": [ "-extern-std=c++14" ],
     "lflags-posix": [ "-lstdc++", "-lsqlite3" ],
-    "lflags-windows": [  "sqlite3.lib", "/nodefaultlib:msvcetd.lib" ],
+    "lflags-windows": [  "sqlite3.lib", "/nodefaultlib:msvcetd.lib"],
```
Geod24: No change but a space remove, revert this line

source/agora/registry/NameRegistryServer.d
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    auto router = new URLRouter();
+	router.registerRestInterface(new NameRegistryImpl());
+
+	auto settings = new HTTPServerSettings;
+	settings.port = cmdline_args.bind_port;
+    settings.bindAddresses = [cmdline_args.bind_address];
+	listenHTTP(settings, router);
+    runEventLoop();
```
Geod24: Indentation. Also `return runEventLoop();`

source/agora/registry/NameRegistryServer.d
```diff
@@ -0,0 +1,78 @@
...
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.registry.NameRegistryServer;
```
Geod24: Please rename to `main` to follow the existing naming convention. One reason to do this is that `dub` does not run unittests in `mainSourceFile`.

source/agora/registry/NameRegistryImpl.d
```diff
@@ -0,0 +1,93 @@
...
+        // check if we received stale data
+        if ((registry_payload.data.public_key in registry_map) &&
+             registry_map[registry_payload.data.public_key].seq > registry_payload.data.seq)
+            return "registry already has a more up-to-date version of the data";
+
+        // register data
+        log.info("Registering network addresses: {} for public key: {}", registry_payload.data.addresses,
+            registry_payload.data.public_key.toString());
+        registry_map[registry_payload.data.public_key] = registry_payload.data;
+        return "";
```
Geod24: Always return `null` over `""`. There's a small but important reason for this:
```D
if (string s = foo())
    assert(s.length);
```
This can fail if you return `""`. But most people expect the length to be tested, not the pointer. Likewise, `"abcd"[0 .. 0]` gives you an empty slice with a non-`null` pointer.

source/agora/registry/NameRegistryImpl.d
```diff
@@ -0,0 +1,93 @@
...
+import agora.utils.Log;
+
+import vibe.core.core;
+import vibe.http.common;
+import vibe.web.rest;
+
+mixin AddLogger!();
+
+/// Implementation of `NameRegistryAPI` using associative arrays
+static class NameRegistryImpl: NameRegistryAPI
```
Geod24:
```suggestion
public final class NameRegistryImpl: NameRegistryAPI
```

`static` in this context is redundant.

source/agora/registry/NameRegistryClient.d
```diff
@@ -0,0 +1,82 @@
...
+module agora.registry.NameRegistryClient;
+
+import agora.common.Hash;
+import agora.common.Types;
+import agora.common.crypto.Key;
+import agora.registry.NameRegistryAPI;
+
+import core.stdc.time;
+
+/// Thin wrapper around `NameRegistryApi`
```
Geod24: What's the reason for this wrapper ? Currently it holds no state, and (almost) the same API can be achieved through `registerNetworkAddresses`.

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,95 @@
...
+
+    ///
+    public void sign_payload(const ref SecretKey secret_key)
+    {
+        signature = secret_key.sign(hashFull(data)[]);
+    }
+};
+
+/// API allowing to store network addresses under a public key
+static interface NameRegistryAPI
```
Geod24:
```suggestion
public interface NameRegistryAPI
```

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,95 @@
...
+
+        Params:
+            public_key = the public key that was used to register
+                         the network addresses
+
+        Returns:
+            Network addresses associated with the `public_key`
+
+        API:
+            GET /network_addresses
```
Geod24: Is that really a name we want ? I would suggest something like:
```suggestion
            GET /validators
```
OR:
```suggestion
            GET /nodes
```

`/addresses` would also be okay I guess, but the above feels more extensible (because it's tied to the content of the collection more than its type).

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,95 @@
...
+        Params:
+            registry_payload =
+                the data we want to register with the name registry server
+
+        Returns:
+            empty string, if the registration was successful, otherwise returns
+            the error message
+
+        API:
+            PUT /register_network_addresses
```
Geod24: If we're going with `PUT`, I'd recommend to have the same endpoint. so perhaps ` putValidators` (if the previous endpoint is renamed) ?

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,95 @@
...
+struct RegistryPayloadData
+{
+    /// the public key that we want to register
+    public PublicKey public_key;
+
+    /// network addresses associated with the public key
+    public Address[] addresses;
+
+    /// monotonically increasing sequence number
+    public long seq;
```
Geod24:
```suggestion
    public ulong seq;
```

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,95 @@
...
+struct RegistryPayload
+{
+    ///
+    public RegistryPayloadData data;
+
+    /// signature over the `data` member
+    public Signature signature;
+
+    ///
+    public void sign_payload(const ref SecretKey secret_key)
```
Geod24: Since it's only signed from one place, this seems superfluous.

source/agora/node/Validator.d
Geod24: This needs to be validated when the config file is parsed, not when the validator is instantiated

source/agora/common/crypto/Key.d
Geod24: Irrelevant

source/agora/common/Config.d
Geod24: If `node is null`, `enabled` will be `false`.

source/agora/test/Base.d
```diff
@@ -1509,8 +1553,17 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager,
...
-        net.createNewNode(conf, file, line);
+    {
+        auto node_api = net.createNewNode(conf, file, line);
+        if(conf.validator.enabled)
+        {
+            string tid_sink;
+            node_api.tid().toString((const(char)[] tid){tid_sink = to!string(tid);});
+            name_registry_client.registerNetworkAddresses(conf.validator.key_pair, [tid_sink]);
+        }
+    }
```
Geod24: Why do you need to link the tid ? And the `registerNetworkAddresses` should probably be done by the validator as part of the startup code, it shouldn't be test specific.

source/agora/test/Base.d
```diff
@@ -1296,6 +1336,9 @@ public struct TestConf
     /// max retries before a request is considered failed
     size_t max_retries = 20;
 
+    /// address of the name registry
+    string registry_address = "name.registry";
+
```
Geod24: I don't think we need to configure this.

Geod24 (changes requested):
First round of review.

source/agora/test/Base.d
ferencdg: "Why do you need to link the tid ?"

During the network tests the nodes communicate through the Localrest, and to get a handle for an other Node(to communicate with it), we need a Tid. There wouldn't  be to much use if we register the (same) IP address for all the test Nodes as we couldn't use the NameRegistry to get a handle for a Node.

"And the registerNetworkAddresses should probably be done by the validator as part of the startup code, it shouldn't be test specific."
There is a specific code in the validator for the reason I mentioned above.

source/agora/node/Validator.d
ferencdg: the reason why I validated it here is that I assumed having a correct name registry is not mandatory requirement for a Validator to run

source/agora/test/Base.d
AndrejMitrovic: The code you wrote in `Validator.d` is never executed because `addresses_to_register` was never set in the config in the tests. If it was executed you would see there were exceptions being thrown. The code is not doing what you've intended it to do.

I'll try to give some hints at how this can be improved - but I will need a bit of time. So please stay tuned.

AndrejMitrovic (dismissed):

AndrejMitrovic:
@ferencdg: take a look at the changes I've made here: https://github.com/bpfkorea/agora/commit/8b2dd7bfce288be6e9b7bc85f14e408aaa751ba9

You can run the test with: `dtest=agora.test.NameRegistry dub test`.

Some notes in particular follow:

- Avoid introducing new utility functions unless they have a good use-case and are used from many other call sites. Rule of three applies: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)
- Avoid writing extremely simple wrapper classes / structs.
- If you do write wrapper classes / structs , make sure they're self-contained. Your `RegistryPayload` struct implements `sign_payload`, but it does not implement validation. Instead validation was moved inside of `NameRegistryImpl.registerNetworkAddresses`. This is a bit of leaky abstraction. But it's also **incorrect**, because:
- The `NameRegistry` server implemented validation, but the data returned by the server is an array of addresses. This means a node cannot verify the authenticity of the data returned by the name registry. The `NameRegistry` should return the payload so the node can validate the data before accepting it as legitimate. - Note that I have not fixed this in the commit I linked above.
- Unittests are **absolutely required**. In fact your code wasn't even executing because the config option `addresses_to_register` was never set. If it was set, it would have crashed - this is due to the fact we cannot use the RemoteAPI until it started its event loop. In this case, it's safe to start using it after `Validator.start()` begins.
- `getNetworkAddresses()` was never called from any of your code.
- There should be no vibe.d imports in any modules except network-related modules such as `NetworkManager` or `NetworkClient`. In particular vibe.d should not be imported from within Validator.d. And all networking-related code should live in one of the network-related modules, it should not be moved to the main classes like `Validator` or `FullNode`.
- @safe / nothrow and potentially @nogc are all missing
- Styling is still not being followed.
- There is a mix of new functionality, utility functions, and unrelated styling fixes all mixed-in into a single commit. Commits should be atomic and self-contained.

Your work is appreciated, and thank you for the progress you've made so far! Let's just try to aim towards improving this work in the future. :)

source/agora/common/Config.d
Geod24: It's a wrong bugfix, `node is null` is handled by `get`.

source/agora/network/NetworkManager.d
Geod24: Agreed, no need to be overly verbose here.

source/agora/registry/NameRegistryClient.d
ferencdg: I think my idea was that we don't want to repeat code like this:
 RegistryPayload registry_payload =
        {
            data:
            {
                public_key : key_pair.address,
                addresses : addresses,
                seq : time(null)
            }
        };
        registry_payload.sign_payload(key_pair.secret);

but I agree we can live without this class

ferencdg:
> @ferencdg: take a look at the changes I've made here: [8b2dd7b](https://github.com/bpfkorea/agora/commit/8b2dd7bfce288be6e9b7bc85f14e408aaa751ba9)
> 
> You can run the test with: `dtest=agora.test.NameRegistry dub test`.
> 
> Some notes in particular follow:
> 
> * Avoid introducing new utility functions unless they have a good use-case and are used from many other call sites. Rule of three applies: https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)
> * Avoid writing extremely simple wrapper classes / structs.
> * If you do write wrapper classes / structs , make sure they're self-contained. Your `RegistryPayload` struct implements `sign_payload`, but it does not implement validation. Instead validation was moved inside of `NameRegistryImpl.registerNetworkAddresses`. This is a bit of leaky abstraction. But it's also **incorrect**, because:
> * The `NameRegistry` server implemented validation, but the data returned by the server is an array of addresses. This means a node cannot verify the authenticity of the data returned by the name registry. The `NameRegistry` should return the payload so the node can validate the data before accepting it as legitimate. - Note that I have not fixed this in the commit I linked above.
> * Unittests are **absolutely required**. In fact your code wasn't even executing because the config option `addresses_to_register` was never set. If it was set, it would have crashed - this is due to the fact we cannot use the RemoteAPI until it started its event loop. In this case, it's safe to start using it after `Validator.start()` begins.
> * `getNetworkAddresses()` was never called from any of your code.
> * There should be no vibe.d imports in any modules except network-related modules such as `NetworkManager` or `NetworkClient`. In particular vibe.d should not be imported from within Validator.d. And all networking-related code should live in one of the network-related modules, it should not be moved to the main classes like `Validator` or `FullNode`.
> * @safe / nothrow and potentially @nogc are all missing
> * Styling is still not being followed.
> * There is a mix of new functionality, utility functions, and unrelated styling fixes all mixed-in into a single commit. Commits should be atomic and self-contained.
> 
> Your work is appreciated, and thank you for the progress you've made so far! Let's just try to aim towards improving this work in the future. :)

thanks a lot for the review and the code,  I found it very useful!
I do aggree with most of it but, but have some comments:

1. "Avoid introducing new utility functions unless they have a good use-case"
retry is a 'functional style' retry function and it is more expressive then coding retry logic every time: you can just look at the name of the function and you know it does retry. Containers like map also have simple utility functions like: filter, map...

2. "The NameRegistry server implemented validation, but the data returned by the server is an array of addresses..."
I think Mathias and me had a similar conversation(please let me know if this is not the case) and we concluded even if the attacker would change the contents of the IP Adresses, it couldn't do much with it, as it doesn't have the private key to participate in the voting protocol. Of course he could just return some random addresses and it could cause DoS.
However, if an attacker can change the content of the array, then he could also just decide not to return anything and causing DoS, so returning the signed version wouldn't help much.

3. "getNetworkAddresses() was never called from any of your code"
I think the original issue was just to implement the name registry server, I suppose we would need a separate issue ticket for the actual usage. Let me know if I should raise a ticket and how exactly this getNetworkAddress should be used.

4. "Styling is still not being followed."
Please let me know when I made mistake, I try to follow it as much as possible, but hard to get use to it.

AndrejMitrovic:
> retry is a 'functional style' retry function and it is more expressive then coding retry logic every time: you can just look at the name of the function and you know it does retry. Containers like map also have simple utility functions like: filter, map...

Alright, but if it's useful then it should be used from multiple places. Hence why it would make sense to split this into a separate commit or possibly a PR that also uses `retry` in other places (such as the NetworkClient).

> I think Mathias and me had a similar conversation(please let me know if this is not the case) and we concluded even if the attacker would change the contents of the IP Adresses, it couldn't do much with it, as it doesn't have the private key to participate in the voting protocol.

Hmm.. I'm not sure I agree with @Geod24. Consider the following case:

- Attacker owns the name registry
- Victim sends request for addresses to the registry
- It receives IPs based on the keys it wants to look up, it establishes connection to these IPs but the requests frequently time out. Maybe the BanManager will ban these IPs, but there will still be considerable time wasted trying to use these IPs - lack of liveness.

At this point the victim doesn't know whether the name registry is trustworthy or not. Maybe it's experiencing actual connection issues, or maybe the name registry is giving it false information.

If the addresses returned from the registry are signed then the victim node can at least verify the signature. If the signature is bad the victim can proceed to ban the name registry and try another registry instead. Without verifying the signature the victim can't tell what's wrong.

Also it would be good to have the timestamp in the return value. If the timestamp is too old then we probably do not want to establish a connection with such a node.

> I think the original issue was just to implement the name registry server

Alright. But the work should be split up into smaller chunks:
- Build the business logic in Agora first (using LocalRest)
- Build an actual name server

Instead the PR as it stands implemented most of the second part, but only implemented partially the first part.

> Please let me know when I made mistake, I try to follow it as much as possible, but hard to get use to it.

We'll try to be more clear about the styling. In general it should follow the style of the rest of the codebase, and parts of this are documented here: https://github.com/bpfkorea/agora/blob/v0.x.x/doc/Style.md

We'll try to set up an automatic style check in the CI sometime later. For now it's a very manual process.

Geod24:
> "The NameRegistry server implemented validation, but the data returned by the server is an array of addresses..."

My memory is fuzzy on the details, so we might have had a misunderstanding. I fully agree that the registry should return the full registration (signature included) so the node can verify it. Maybe the discussion was about a registry returning old data ?

ferencdg:
> Hmm.. I'm not sure I agree with @Geod24. Consider the following case:
> 
> * Attacker owns the name registry
> * Victim sends request for addresses to the registry
> * It receives IPs based on the keys it wants to look up, it establishes connection to these IPs but the requests frequently time out. Maybe the BanManager will ban these IPs, but there will still be considerable time wasted trying to use these IPs - lack of liveness.
> 
> At this point the victim doesn't know whether the name registry is trustworthy or not. Maybe it's experiencing actual connection issues, or maybe the name registry is giving it false information.
> 
> If the addresses returned from the registry are signed then the victim node can at least verify the signature. If the signature is bad the victim can proceed to ban the name registry and try another registry instead. Without verifying the signature the victim can't tell what's wrong.
> 
> Also it would be good to have the timestamp in the return value. If the timestamp is too old then we probably do not want to establish a connection with such a node.
> 

Now I remember the original conversation with @Geod24, it was about the timestamp, not about this specific scenario when the attacker owns the registry. I think in my scenario, the attacker obtained the IP of a well-behaving validator(who has already registered his IP), and because there was no timestamp returned the other validators thought the IP still belongs to a well-behaving node. I think at this point Mathias told that it is okay, because the attacker doesn't have the private key.

ferencdg:
> > "The NameRegistry server implemented validation, but the data returned by the server is an array of addresses..."
> 
> My memory is fuzzy on the details, so we might have had a misunderstanding. I fully agree that the registry should return the full registration (signature included) so the node can verify it. Maybe the discussion was about a registry returning old data ?

yes, right our discussion was about the timestamp, sorry

ferencdg:
"If the addresses returned from the registry are signed then the victim node can at least verify the signature. If the signature is bad the victim can proceed to ban the name registry and try another registry instead. Without verifying the signature the victim can't tell what's wrong."

just some comment on this one: on top of verifying the signature the client could connect to multiple name registries at the same time and query the IP corresponding to PUB1 and use some majority decision... This would have the advantage that some validators might register their addresses into different name registries... 

This could also be an alternative of requiring the registry to send back signed IP addresses(but we can implement it on top of the signing too).

Or we could get the name registries talk to each other, but that could be harder to implement.

source/agora/common/Config.d
ferencdg: it is moved out because it is called from parseValidatorConfig too

source/agora/registry/NameRegistryAPI.d
ferencdg: time() returns clock_t on POXIS and 
alias c_long     clock_t;

and

alias long  c_long;

will this cause problems?

time actually could be negative if we want to refer to time before the unix epoch

Geod24: Then it's not a sequence number ?

source/agora/common/Config.d
```diff
@@ -165,6 +165,12 @@ public struct ValidatorConfig
 
     /// The seed to use for the keypair of this node
     public immutable KeyPair key_pair;
+
+    // Network addresses that will be registered with the public key (Validator only)
+    public immutable string[] addresses_to_register;
+
+    // Registry address
+    public string registry_address = "http://127.0.0.1:3003";
```
Geod24: This shouldn't have a default value.

source/agora/common/Config.d
```diff
@@ -324,12 +308,37 @@ private Config parseConfigImpl (ref const CommandLine cmdln, Node root)
...
+    return result;
+}
+
 /// Parse the node config section
 private NodeConfig parseNodeConfig (Node* node, const ref CommandLine cmdln)
 {
     auto min_listeners = get!(size_t, "node", "min_listeners")(cmdln, node);
     auto max_listeners = get!(size_t, "node", "max_listeners")(cmdln, node);
     auto address = get!(string, "node", "address")(cmdln, node);
+
```
Geod24: Unrelated

source/agora/common/Config.d
Geod24: The moves LGTM, but as mentioned, in its own commit.

source/agora/utils/Utility.d
```diff
@@ -0,0 +1,56 @@
...
+            duration = the time between retrying to execute the delegate
+
+        Returns:
+            returns Nullable() in case the delegate cannot be executed after X
+            retries, otherwise it returns the original return value of
+            the delegate wrapped into a Nullable object
+
+***************************************************************************/
+
+Nullable!(ReturnType!dg) retry (alias dg, int max_retry, Duration duration, T) (T waiter, string error_msg = "")
```
Geod24: I known it's tempting to move as much as possible to CT argument, but in my experience, it's a bad idea and leads to bloat. In this case, I would move `max_retry` and `duration` to runtime.

Additionally, never use `""` for empty string, as mentioned. Use `null` (or `string.init` if you need a strong type).
But in this case, we can use a sensible default error message, such as `"Operation timed out"`.

Last but not least, this is very similar to the `retryFor` utility we have. I think one should be expressed in terms of the other, **and** there should be unittests.

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,186 @@
...
+                (ip_parts[0]==127);
+
+        }
+    }
+
+    private static string[] filterIPs(bool function(string ip) filter_func)
+    {
+        return filter!(ip => filter_func(ip))(getAllIPs()).array();
+    }
+
```
Geod24:
```suggestion
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,186 @@
...
+            // 127.0.0.0   - 127.255.255.255
+
+            ubyte[] ip_parts = ip.split(".").map!(ip_part => to!ubyte(ip_part)).array();
+            return
+                (ip_parts[0]==10) ||
+                ((ip_parts[0]==172) && (ip_parts[1]>=16 && ip_parts[1]<=31)) ||
+                (ip_parts[0]==192 && ip_parts[1]==168) ||
+                (ip_parts[0]==169 && ip_parts[1]==254) ||
+                (ip_parts[0]==127);
+
```
Geod24:
```suggestion
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,186 @@
...
+                }
+                string ip = buffer[0 .. strlen(buffer.ptr)].idup();
+                ips ~= ip;
+            }
+        }
+
+        return ips;
+    }
+}
+version(Windows)
```
Geod24:
```suggestion
version (Windows)
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,186 @@
...
+{
+
+import std.socket;
```
Geod24:
```suggestion
{
import std.socket;
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,186 @@
...
+
+import core.sys.windows.iphlpapi;
+import core.sys.windows.iptypes;
+import core.sys.windows.windef;
+import core.sys.windows.winsock2;
+import core.stdc.stdlib: malloc, free;
+import core.stdc.string: strlen;
+
+    public static string[] getAllIPs()
+    {
```
Geod24: What's up with the indentation ? You either fo with indent or not, but it should be consistent. I'd recommend to just indent it.

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,186 @@
...
+        string[] ips = [];
```
Geod24:
```suggestion
        string[] ips;
```

doc/config.example.yaml
```diff
@@ -52,6 +52,13 @@ validator:
   # Private seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
   # Public address:  GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
   seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  # Address of the name registry
+  registry_address: http://127.0.0.1:3003
```
Geod24, line 56: Shouldn't this be an array as well ?

source/agora/network/NetworkManager.d
```diff
@@ -426,6 +442,27 @@ public class NetworkManager
...
             this.required_peer_keys.remove(key);
 
+
+        foreach (key; this.required_peer_keys)
+            taskman.runTask
+            ({
+                retry!
+                ({
+                    auto registryPayload = registry_client.getValidator(key);
+                    if(!registryPayload.verifySignature(key))
```
Geod24:
```suggestion
                    if (!registryPayload.verifySignature(key))
```

source/agora/network/NetworkManager.d
```diff
@@ -426,6 +442,27 @@ public class NetworkManager
...
+                    {
+                        log.warn("RegistryPayload signature is incorrect for {}", key.toString());
```
Geod24:
```suggestion
                        log.warn("RegistryPayload signature is incorrect for {}", key);
```

The logger will call the right overload of `toString` by itself

source/agora/network/NetworkManager.d
```diff
@@ -426,6 +442,27 @@ public class NetworkManager
         foreach (key; this.connected_validator_keys)
             this.required_peer_keys.remove(key);
 
+
+        foreach (key; this.required_peer_keys)
+            taskman.runTask
+            ({
+                retry!
```
Geod24: There's probably a better API to provide for this case, e.g. `GET /validators?limit=10`.
Iterating on `required_peer_keys` sounds dangerous, especially given the default value is `init`.

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,101 @@
...
+
+import agora.common.Hash;
+import agora.common.Types;
+import agora.common.crypto.Key;
+
+import vibe.http.common;
+import vibe.web.rest;
+
+///
+struct RegistryPayloadData
```
Geod24:
```suggestion
public struct RegistryPayloadData
```

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,101 @@
...
+
+    /// network addresses associated with the public key
+    public const(Address)[] addresses;
+
+    /// monotonically increasing sequence number
+    public ulong seq;
+}
+
+///
+struct RegistryPayload
```
Geod24:
```suggestion
public struct RegistryPayload
```

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,101 @@
...
+            the error message
+
+        API:
+            PUT /validator
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.PUT)
+    public string putValidator (RegistryPayload registry_payload);
+
```
Geod24:
```suggestion
```

source/agora/registry/NameRegistryImpl.d
```diff
@@ -0,0 +1,94 @@
...
+import vibe.core.core;
+import vibe.http.common;
+import vibe.web.rest;
+
+mixin AddLogger!();
+
+/// Implementation of `NameRegistryAPI` using associative arrays
+public final class NameRegistryImpl: NameRegistryAPI
+{
+
```
Geod24:
```suggestion
```

source/agora/network/NetworkManager.d
```diff
@@ -583,6 +620,36 @@ public class NetworkManager
...
+
+        try
+        {
+            this.registry_client.putValidator(payload);
+        }
+        catch (Exception ex)
+        {
+            log.info("Couldn't register our address: {}. Trying again later..",
+                ex);
+        }
```
Geod24, line 794: When would we get an Exception ? The registry will return a string with an error in case of e.g. invalid signature, but we never check the return value.

source/agora/network/NetworkManager.d
```diff
@@ -415,6 +423,14 @@ public class NetworkManager
         client.registerListener();
     }
 
+    public void startPeriodicNameRegistration ()
+    {
+        this.registry_client = this.getNameRegistryClient(
+            validator_config.registry_address, 2.seconds);
+        this.onRegisterName();  // avoid delay
+        this.taskman.setTimer(5.minutes, &this.onRegisterName, Periodic.Yes);
```
Geod24: So we send a registration every 5 minutes, no matter what ? That works, but it should be documented as a conscious decision.

source/agora/registry/NameRegistryImpl.d
```diff
@@ -0,0 +1,94 @@
...
+        // check if we received stale data
+        if ((registry_payload.data.public_key in registry_map) &&
+             registry_map[registry_payload.data.public_key].data.seq > registry_payload.data.seq)
+            return "registry already has a more up-to-date version of the data";
+
+        // register data
+        log.info("Registering network addresses: {} for public key: {}", registry_payload.data.addresses,
+            registry_payload.data.public_key.toString());
+        registry_map[registry_payload.data.public_key] = registry_payload;
```
Geod24: As I mentioned in the calling code, this might just need to `throw` if we want to report error (see `RestException`).
You can instantiate a single `Exception` in the constructor and just throw that.

source/agora/registry/main.d
```diff
@@ -0,0 +1,78 @@
...
+    auto router = new URLRouter();
+    router.registerRestInterface(new NameRegistryImpl());
+
+    auto settings = new HTTPServerSettings;
+    settings.port = cmdline_args.bind_port;
+    settings.bindAddresses = [cmdline_args.bind_address];
+    listenHTTP(settings, router);
+    runEventLoop();
+
+    return 0;
```
Geod24:
```suggestion
    return runEventLoop();
```

source/agora/test/EnrollmentManager.d
```diff
@@ -237,7 +237,7 @@ unittest
         }
 
         /// see base class
-        public override void createNewNode (Config conf, string file, int line)
+        public override RemoteAPI!TestAPI createNewNode (Config conf, string file, int line)
```
Geod24: Why this change ? It doesn't seem to be used.

source/agora/test/Base.d
```diff
@@ -653,6 +662,19 @@ public class TestAPIManager
 
         this.reg.register(conf.node.address, api.tid());
         this.nodes ~= NodePair(conf.node.address, api, time);
+        return api;
+    }
+
+    /***************************************************************************
+
+    Create a new name registry
```
```suggestion
        Create a new name registry
```

source/agora/test/Base.d
```diff
@@ -653,6 +662,19 @@ public class TestAPIManager
...
+
+    ***************************************************************************/
+
+    public void createNameRegistry ()
+    {
+        this.name_registry = RemoteAPI!NameRegistryAPI.spawn!NameRegistryImpl();
+        this.reg.register("name.registry", this.name_registry.tid());
```
Geod24, line 697: 👍

source/agora/test/Base.d
```diff
@@ -865,6 +889,16 @@ public class TestNetworkManager : NetworkManager
                "' without first creating it");
     }
 
+    ///
+    public final override RemoteAPI!NameRegistryAPI getNameRegistryClient (Address address, Duration timeout)
```
```suggestion
    public override RemoteAPI!NameRegistryAPI getNameRegistryClient (Address address, Duration timeout)
```

Geod24 (approved):
Mostly stylistic comments. One major comment about error handling, otherwise it LGTM overall. Let me know when comments are addressed.

source/agora/utils/InetUtils.d
ferencdg: you mean indent the imports too?

doc/config.example.yaml
ferencdg: we need to discuss what semantic do we want when we get the values: 
1. do we want to query all registries at once and use some majority decision if there are differences in the return value for a public key?
2. do we want to query the registries one by one

It didn't seem a straigtforward problem, so I defered it.

source/agora/network/NetworkManager.d
ferencdg: 1. what is the problem with init default value? in this case we don't do anything in the loop
2. I don't understand this limit part: we are passing a public key to getValidator, are you suggesting we should pass an array of public keys? I was also thinking about it for efficiency, but again, this requires me some time to implement

source/agora/network/NetworkManager.d
ferencdg: 1. ughh, I forgot to check the return value
2. we would get an exception if it times out, right?

source/agora/network/NetworkManager.d
ferencdg: document it just alongside with the code?

source/agora/test/EnrollmentManager.d
ferencdg: I used to use the return value at some point, but during refactoring the calling code was removed. I will revert back.

Geod24:
Still a few comments left to address.

source/agora/network/NetworkManager.d
ferencdg: > So we send a registration every 5 minutes, no matter what ? That works, but it should be documented as a conscious decision.

there are multiple reasons why we want to register in every 5 mintues:
1. what if the registry server is restarted?
2. what if on the client machine a new ip was added -> we want to register that one too

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+    }
+    else
+    {
+        import core.sys.darwin.ifaddrs;
+    }
+}
+
+struct InetUtils
+{
+version(OSX)
```
TrustHenry:
```suggestion
version (OSX)
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+{
+    import core.sys.posix.netdb;
+    import core.sys.posix.netinet.in_;
+    import core.sys.posix.sys.socket;
+
+    mixin AddGetAllIPs;
+}
+version(linux)
```
TrustHenry:
```suggestion
version (linux)
```

source/agora/utils/Utility.d
```diff
@@ -0,0 +1,125 @@
...
+            the delegate wrapped into a Nullable object
+
+***************************************************************************/
+
+Nullable!(ReturnType!dg) retry (alias dg, T) (T waiter, long max_retry, Duration duration, string error_msg = "Operation timed out: ")
+{
+    alias RetType = Nullable!(ReturnType!dg);
+    foreach (i; 0 .. max_retry)
+        try
+            if(auto res = dg())
```
Geod24:
```suggestion
            if (auto res = dg())
```

source/agora/utils/Utility.d
```diff
@@ -0,0 +1,125 @@
...
+    Duration timeout, lazy string msg = "",
+    string file = __FILE__, size_t line = __LINE__)
+{
+    // wait 100 msecs between attempts
+    const SleepTime = 100;
+    auto attempts = timeout.total!"msecs" / SleepTime;
+    const TotalAttempts = attempts;
+
+    ThreadWaiter thread_waiter;
+    if(!retry!({return check;})(thread_waiter, attempts, SleepTime.msecs).isNull)
```
Geod24:
```suggestion
    if (!retry!(() => check)(thread_waiter, attempts, SleepTime.msecs).isNull)
```

source/agora/utils/Utility.d
```diff
@@ -0,0 +1,125 @@
...
+
+///
+public struct ThreadWaiter
+{
+    public void wait (Duration duration) nothrow
+    {
+        import core.thread;
+        Thread.sleep(duration);
+    }
+}
```
Geod24, line 113: Why the struct ?

source/agora/test/BanManager.d
```diff
@@ -22,6 +22,7 @@ import agora.common.Hash;
 import agora.consensus.data.Block;
 import agora.consensus.data.Transaction;
 import agora.test.Base;
+import agora.utils.Utility;
```
Geod24: Just add a public import to `Utility` in `agora.utils.Test`  or to `agora.test.Base`, since `retryFor` is used everywhere.

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
+module agora.utils.InetUtils;
```
Geod24, line 14: Missing module doc

source/agora/utils/InetUtils.d
Geod24: This

source/agora/utils/InetUtils.d
Geod24: This

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+import core.stdc.string;
+
+mixin AddLogger!();
+
+///
+mixin template AddGetAllIPs()
+{
+    public static string[] getAllIPs()
+    {
+        string[] ips = [];
```
Geod24:
```suggestion
        string[] ips;
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+
+        ifaddrs *if_address_head_poi;
+        ifaddrs *if_address_poi;
```
Geod24:
```suggestion
        ifaddrs* if_address_head_poi;
        ifaddrs* if_address_poi;
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+
+        getifaddrs (&if_address_head_poi);
```
Geod24:
```suggestion
        getifaddrs(&if_address_head_poi);
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+        scope(exit) freeifaddrs(if_address_head_poi);
+
+        for (if_address_poi = if_address_head_poi; if_address_poi; if_address_poi = if_address_poi.ifa_next)
+        {
+            if (if_address_poi.ifa_addr &&
+            (if_address_poi.ifa_addr.sa_family==AF_INET || if_address_poi.ifa_addr.sa_family==AF_INET6))
+            {
+                const ipv6 = if_address_poi.ifa_addr.sa_family==AF_INET6;
```
Geod24: Space around operators, indentation is off

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+                                if_address_poi.ifa_addr,
+                                sockaddr_len,
+                                buffer.ptr,
+                                buffer.length,
+                                null,
+                                0,
+                                NI_NUMERICHOST);
+                if (name_info_res)
+                {
+                    log.error("error happened during a call to getnameinfo, name_info_res code:", name_info_res);
```
Geod24:
```suggestion
                    log.error("error happened during a call to getnameinfo, name_info_res code: {}", name_info_res);
```

At the moment this isn't logging anything

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+        string[] ips;
+        PIP_ADAPTER_INFO adapter_info_head;
+        PIP_ADAPTER_INFO adapter_info;
+        DWORD ret_adapters_info;
+        ULONG buff_length = IP_ADAPTER_INFO.sizeof;
+        adapter_info_head = cast(IP_ADAPTER_INFO *) malloc(IP_ADAPTER_INFO.sizeof);
+        if (adapter_info_head == NULL)
+        {
+            log.error("Error allocating memory needed to call GetAdaptersinfo - 1");
+            return [];
```
Geod24:
```suggestion
            return null;
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+    {
+        string[] ips;
+        PIP_ADAPTER_INFO adapter_info_head;
+        PIP_ADAPTER_INFO adapter_info;
+        DWORD ret_adapters_info;
+        ULONG buff_length = IP_ADAPTER_INFO.sizeof;
+        adapter_info_head = cast(IP_ADAPTER_INFO *) malloc(IP_ADAPTER_INFO.sizeof);
+        if (adapter_info_head == NULL)
+        {
+            log.error("Error allocating memory needed to call GetAdaptersinfo - 1");
```
Geod24: 1 ?

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+    import core.stdc.string: strlen;
+
+    public static string[] getAllIPs()
+    {
+        string[] ips;
+        PIP_ADAPTER_INFO adapter_info_head;
+        PIP_ADAPTER_INFO adapter_info;
+        DWORD ret_adapters_info;
+        ULONG buff_length = IP_ADAPTER_INFO.sizeof;
+        adapter_info_head = cast(IP_ADAPTER_INFO *) malloc(IP_ADAPTER_INFO.sizeof);
```
Geod24: Merge with the declaration

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+        {
+            log.error("Error allocating memory needed to call GetAdaptersinfo - 1");
+            return [];
+        }
+        scope(exit) free(adapter_info_head);
+        // find out the real size we need to allocate
+        if (GetAdaptersInfo(adapter_info_head, &buff_length) == ERROR_BUFFER_OVERFLOW)
+        {
+            free(adapter_info_head);
+            adapter_info_head = cast(IP_ADAPTER_INFO *) malloc(buff_length);
```
Geod24, line 148: realloc ?

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+
+    public static string[] getPrivateIPs()
+    {
+        return filterIPs(&isPrivateIP);
+    }
+
+    private static bool isPrivateIP(string ip)
+    {
+        bool is_ipv6 = ip.canFind(':');
+        if(is_ipv6)
```
Geod24:
```suggestion
        if (ip.canFind(':')) // IPv6
```

source/agora/common/Config.d
Geod24: On the last commit, you still have a diff on this function because it's moved around. You should revert it.

source/agora/registry/NameRegistryAPI.d
```diff
@@ -0,0 +1,100 @@
...
+        Returns:
+            empty string, if the registration was successful, otherwise returns
+            the error message
+
+        API:
+            PUT /validator
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.PUT)
```
Geod24: You can remove  this, and the import, now

source/agora/registry/NameRegistryImpl.d
```diff
@@ -0,0 +1,91 @@
...
+            GET /validator
+
+    ***************************************************************************/
+
+    public override const(RegistryPayload) getValidator (PublicKey public_key)
+    {
+        if(auto payload = public_key in registry_map)
+            return *payload;
+        else
+            return RegistryPayload.init;
```
Geod24:
```suggestion
        if (auto payload = public_key in registry_map)
            return *payload;
        return RegistryPayload.init;
```

We could also use `Nullable` for  this in the future.

source/agora/registry/NameRegistryImpl.d
```diff
@@ -0,0 +1,91 @@
...
+
+    public override void putValidator (RegistryPayload registry_payload)
+    {
+        // verify signature
+        if (!registry_payload.verifySignature(registry_payload.data.public_key))
+            throw new Exception("incorrect signature");
+
+        // check if we received stale data
+        if ((registry_payload.data.public_key in registry_map) &&
+             registry_map[registry_payload.data.public_key].data.seq > registry_payload.data.seq)
```
Geod24:
```suggestion
        if (auto previous = registry_payload.data.public_key in registry_map)
            if (previous.data.seq > registry_payload.data.seq)
```

The point of having the `in` operator in D is exactly to avoid double lookups

source/agora/test/Base.d
```diff
@@ -1575,6 +1604,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
     return net;
 }
 
+
```
Geod24, line 1615: Revert

Geod24 (changes requested):
A few more things

source/agora/utils/Utility.d
ferencdg: I didn't think this class will ever be part of any inheritance, so declared it as struct. Would class be better?

Geod24: Ah, ignore this comment, the question was more about why an aggregate for it, but I understood later it was because of the template.

source/agora/utils/InetUtils.d
ferencdg: There are two calls to GetAdaptersinfo - once we call it to allocate just one IP_ADAPTER_INFO, and the second time we call it to allocate several ones. I used the postfix -1 and -2 to differentiate between those. I can just be more verbose though to differentiate between those and not use the postfix.

source/agora/utils/InetUtils.d
ferencdg: Realloc can be more more costly than malloc, in the case where there isn't enough space after the buffer previously allocated by malloc. In that case realloc would not only call malloc, but would unnecessarily copy all the contents from the previous memory location.

Geod24: Good point, lets leave it as is.

source/agora/utils/InetUtils.d
Geod24: Change the other message to "Error **re**allocation memory"... ?

AndrejMitrovic:
It would be great to later upstream the `getAllIPs()` functionality into Phobos. There are many more eyes in the Phobos team and they can suggest improvements and catch any edge-cases which we haven't thought about.

AndrejMitrovic:
> It would be great to later upstream the getAllIPs() functionality into Phobos. There are many more eyes in the Phobos team and they can suggest improvements and catch any edge-cases which we haven't thought about.

Alternatively it could be upstreamed into vibe.d in fact.

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+            {
+                auto ip_tmp = cast(char *) adapter_info.IpAddressList.IpAddress.String;
+                string ip = ip_tmp[0 .. strlen(ip_tmp)].idup;
+                if(!ip.length || ip == "0.0.0.0")
+                {
+                    adapter_info = adapter_info.Next;
+                    continue;
+                }
+                ips ~= ip;
+                adapter_info = adapter_info.Next;
```
AndrejMitrovic:
```suggestion
                if (ip.length > 0 && ip != "0.0.0.0")
                    ips ~= ip;
                
                adapter_info = adapter_info.Next;
```

source/agora/utils/InetUtils.d
```diff
@@ -0,0 +1,224 @@
...
+            }
+        }
+        else
+        {
+            log.error("GetAdaptersInfo failed with error: {}", ret_adapters_info);
```
AndrejMitrovic, line 170:
```suggestion
            log.error("GetAdaptersInfo failed with error: {}", ret_adapters_info);
            return null;
```

And the `return ips;` below should be moved up into the last line of the `if ((ret_adapters_info = GetAdaptersInfo(adapter_info_head, &buff_length)) == NO_ERROR)` body.

ferencdg:
> It would be great to later upstream the `getAllIPs()` functionality into Phobos. There are many more eyes in the Phobos team and they can suggest improvements and catch any edge-cases which we haven't thought about.

it is okay from my side, but let me ask @Geod24 as this takes some time

AndrejMitrovic:
Sure. I didn't mean to block this PR though, we can decide to upstream this later.

source/agora/utils/InetUtils.d
ferencdg: I think it is good if we only have one exit point from a function which is usually at the end of the function. Sometimes like when there is an exceptional case we can return early.

Is there any rationale behind returning in the if statement?

AndrejMitrovic: If you have only one exit point and it returns a variable that is declared at the top of the function then you don't know what you're returning without inspecting the code flow.

ferencdg: > If you have only one exit point and it returns a variable that is declared at the top of the function then you don't know what you're returning without inspecting the code flow.

that is true, but lets say someone later refactors the code and add some filtering or manipulation code to the 'ips' variable after the premature return. In this case he will need to change premature return in every line which is error phrone.(I am talking in general when let's say we can return ips from multiple branches)

Anyways now I see there are pros and cons, I will follow your suggestion and return early

AndrejMitrovic: I see your point. In this case I would argue that:

- We should handle all error conditions first.
- If the function becomes so big that it's easy to miss there was some filtering done afterwards then the function should be split into multiple smaller functions. We do not want 1000 line functions in our code.

Geod24 (approved):

PR #1197, opened 2020-09-14
Split KeyPair from Node to Validator Config

kchulj:
Currently, both full nodes and validators share a common `NodeConfig`, and are distinguished by the `is_validator` boolean.
However, a full node should not have a `KeyPair` (and other validator-related attributes) as its attribute.

As discussed, there are two ways of handling this issue:
1. Make a separate `ValidatorConfig` struct which contains attributes exclusive to a `Validator` **(current approach)**
2. Embed `NodeConfig` into a new `ValidatorConfig`

The current approach is ok, but gets messy in `Base.d` because `makeMainConfig` takes an extra `ValidatorConfig` as its parameter. The `ValidatorConfig` is also created for full nodes, just with initial values. The function could be overloaded to remove this parameter for full nodes.

In the 2nd approach, the main `Config` has to be templated to either take `NodeConfig` or `ValidatorConfig`. The disadvantage would be that more code changes will be required in agora as a whole.

Relates #816

Issue #816, opened 2020-05-12
Split off keypair from Config and into a separate ValidatorConfig


AndrejMitrovic:
Currently the keypair is still in the main Config, but it doesn't make sense for FullNodes. Keeping it here will cause bugs, because code will assume the keypair is always valid.

PR #1200, opened 2020-09-15
Move validator attributes to ValidatorConfig

kchulj:
~~Split off from #1197~~
~~The commits will be re-organized once #1197 is complete.~~

~~Relates #816~~
This PR has been closed as the parameters are consensus-critical and will also affect full nodes.

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -138,6 +138,16 @@ public struct NodeConfig
 
     /// Path to the data directory to store metadata and blockchain data
     public string data_dir = "/var/lib/agora/";
+}
+
+/// Validator config
+public struct ValidatorConfig
+{
+    /// Whether or not this node should try to act as a validator
+    public bool enabled;
+
+    /// The seed to use for the keypair of this node
+    public immutable KeyPair key_pair;
 
     /// The cycle length for a validator
     public uint validator_cycle = 1008;
@@ -155,16 +165,6 @@ public struct NodeConfig
 
     /// How often should the periodic preimage reveal timer trigger (in seconds)
     public Duration preimage_reveal_interval = 10.seconds;
-}
-
-/// Validator config
-public struct ValidatorConfig
-{
-    /// Whether or not this node should try to act as a validator
-    public bool enabled;
-
-    /// The seed to use for the keypair of this node
-    public immutable KeyPair key_pair;
 }
 
 /// Admin API config
@@ -344,14 +344,6 @@ private NodeConfig parseNodeConfig (Node* node, const ref CommandLine cmdln)
 
     string data_dir = get!(string, "node", "data_dir")(cmdln, node);
     auto port = get!(ushort, "node", "port")(cmdln, node);
-    auto validator_cycle = get!(uint, "node", "validator_cycle")(cmdln, node);
-    auto max_quorum_nodes = get!(uint, "node", "max_quorum_nodes")(cmdln, node);
-    auto quorum_threshold = get!(uint, "node", "quorum_threshold")(cmdln, node);
-    assert(quorum_threshold >= 1 && quorum_threshold <= 100);
-    auto quorum_shuffle_interval = get!(uint, "node", "quorum_shuffle_interval")(
-        cmdln, node);
-    auto preimage_reveal_interval = get!(Duration, "node", "preimage_reveal_interval",
-        str => str.to!ulong.seconds)(cmdln, node);
 
     NodeConfig result = {
             min_listeners : min_listeners,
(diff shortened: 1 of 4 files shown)


source/agora/test/Base.d
```diff
@@ -1389,13 +1387,26 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager,
         return conf;
     }
 
+    ValidatorConfig makeValidatorConfig (KeyPair node_key)
+    {
+        ValidatorConfig conf =
+        {
+            enabled : true,
+            key_pair : node_key,
+            // preimage_reveal_interval : 1.seconds,  // check revealing frequently
```
Geod24: Just remove the comment here and add it in the next commit

source/agora/test/Base.d
```diff
@@ -1421,45 +1432,72 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager,
...
+            pair.index % node_configs.length, pair.value[0], node_configs,
+                makeValidatorConfig(pair.value[1])));
+
+    extra_node_configs.drop(test_conf.validators).enumerate(test_conf.validators).each!(
         pair => main_configs ~= makeMainConfig(
             pair.index % node_configs.length, pair.value, node_configs));
 
+    /*extra_node_configs.enumerate.each!(
+        pair => main_configs ~= makeMainConfig(
+            pair.index % node_configs.length, pair.value, node_configs));*/
```
Geod24: ?

source/agora/test/Base.d
```diff
@@ -1421,45 +1432,72 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager,
...
+
     auto gen_block = makeGenesisBlock(
-        node_configs
-            .filter!(conf => conf.is_validator)
+        validator_configs
+            // .filter!(conf => conf.validator.enabled)
```
Geod24: Remove ?

Geod24 (changes requested):
Unfortunately you can't quite split commit this way. Remember, commits should be atomic: It means that on their own, they should compile and pass test, which is clearly not the case here. I can see that the first, second, and last commits are related.

kchulj:
@Geod24 Since the first and second commit rely on each other, do you suggest that I combine these?
Once this one is merged, then I would make the last commits.

Geod24:
Yes. But the last commit is also needed, because without it, the integration tests won't pass. A good litmus test is: "can this commit be alone in it's own PR ?".

source/agora/test/Base.d
```diff
@@ -1421,45 +1432,72 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager,
...
     // n2 <- n1 <- n2 and n1 <- o1 and n2 <- o2, (o = outsider).
     // otherwise it will be [n1 <-> n2] <- o1 and [n1 <-> n2] <- o2
-    extra_node_configs.enumerate.each!(
+
+    extra_node_configs.zip(outsider_validator_keys).enumerate
+        .each!(pair => main_configs ~= makeMainConfig(
+            pair.index % node_configs.length, pair.value[0], node_configs,
+                makeValidatorConfig(pair.value[1])));
+
+    extra_node_configs.drop(test_conf.validators).enumerate(test_conf.validators).each!(
```
linked0: I guess it's a very clever code but makes the overall code complex. So how about making another variable like `fullnode_configs` for nodes that are not validators. This idea could make the code redundant but more readable.

kchulj:
Okay, I will complete #1197 before moving on here.

source/agora/test/Base.d
kchulj: Thanks for your suggestion.
The change has been made in #1197.

source/agora/common/Config.d
```diff
@@ -327,49 +327,32 @@ private Config parseConfigImpl (ref const CommandLine cmdln, Node root)
...
-            max_retries : max_retries,
-            timeout : timeout,
-            data_dir : data_dir,
-            validator_cycle : validator_cycle,
-            max_quorum_nodes : max_quorum_nodes,
-            quorum_threshold : quorum_threshold,
-            quorum_shuffle_interval : quorum_shuffle_interval,
-            preimage_reveal_interval : preimage_reveal_interval,
+        genesis_block:
+            opt!(string, "node", "genesis_block")(cmdln, node),
```
kchulj, line 332: Can this be one line?

Geod24:
As discussed, this is a different issue. Those parameters are consensus-critical and will also affect full nodes.

AndrejMitrovic:
> Those parameters are consensus-critical and will also affect full nodes.

`genesis_start_time` and `validator_cycle` definitely belong to both node types.

However I think `max_quorum_nodes`, `quorum_threshold`, `quorum_shuffle_interval`, and `preimage_reveal_interval` can be moved to ValidatorConfig because full nodes don't have to care about the quorums, they only care about the enrolled validator set and the preimages.

Issue #1170, opened 2020-09-04
The latest SCP envelopes need to be saved & restored in case of a crash


AndrejMitrovic:
If a validator node crashes during a consensus round then we must restore the node's latest nomination & ballot protocol messages the next time the node boots up.

SCP explicitly supports this exact scenario: https://github.com/bpfkorea/agora/blob/0bb013fbd6e7ed22999fb1bb9a86a92c2cd58fdc/source/scpp/src/scp/SCP.h#L97-L99

We need to store `getCurrentState()` and **not** `getLatestMessagesSend()`. The former contains all the latest messages for each node, while the latter only contains the most recently broadcasted message by this node.

From the SCP whitepaper what we need to save is the `N` from Fig. 14. from the nomination protocol (because `X`, `Y`, and `Z` can be regenerated using it), and the `M` from Fig. 16 from the ballot protocol (latest ballot message from each node).

AndrejMitrovic:
In addition to this we also need to look into what happens to the timer routines. Presumably a crashed node will have to call `scp.nominate` again, however it will start the protocol from nomination round 0 - which may be wrong depending on when the node crashed.

In fact the Stellar devs have recently run into this same problem: https://github.com/stellar/stellar-core/issues/2081

Therefore we should monitor that issue closely and also work towards updating our bindings to the latest SCP version.

AndrejMitrovic:
This issue will solve https://github.com/bpfkorea/agora/issues/640 as well. You can take a look at how Stellar does it:

https://github.com/stellar/stellar-core/blob/e88d1006292cf523b7c6b21c097d1e3cd24db8a6/src/herder/HerderImpl.cpp#L1557

It stores the latest envelopes on disk, and then reloads them. We would do something similar. 

In fact this entire code section is wrong:

https://github.com/bpfkorea/agora/blob/6150d764a227e636b192e5f4196b23939cc5ec6d/source/agora/consensus/protocol/Nominator.d#L250-L268

We should not be restoring SCP state from the blockchain, but from the envelopes themselves. And only the latest envelopes (the latest slot index) are needed. Essentially we would store `getCurrentState()` data to disk, and later load it and call `scp.setStateFromEnvelope` with it.

TrustHenry:
@AndrejMitrovic I confusion has arisen.
Among the envelopes obtained by `getCurrentState()`

```
[...]

PR #1204, opened 2020-09-16
Add LatestEnvelopes and store & restore envelopes

TrustHenry:
It restores the node's latest nomination & ballot protocol messages
the next time the node boots up.
Store the latest envelopes in DB and restore to 'scp.setStateFromEnvelope'.

Related to #1170

### DIFF
--- a/source/agora/consensus/SCPEnvelopeStore.d
+++ b/source/agora/consensus/SCPEnvelopeStore.d
@@ -1,0 +1,199 @@
+/*******************************************************************************
+
+    Contains supporting code for storing the latest SCP envelopes,
+    using SQLite as a backing store.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.SCPEnvelopeStore;
+
+import agora.common.ManagedDatabase;
+import agora.common.Serializer;
+import agora.common.Types;
+import agora.utils.Log;
+
+import d2sqlite3.library;
+import d2sqlite3.results;
+import d2sqlite3.sqlite3;
+
+import scpd.types.Stellar_SCP;
+
+import std.file : exists;
+import std.range;
+
+mixin AddLogger!();
+
+
+/// Ditto
+public class SCPEnvelopeStore
+{
+    /// SQLite db instance
+    private ManagedDatabase db;
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            db_path = path to the database file, or in-memory storage if
+                      :memory: was passed
+
+    ***************************************************************************/
+
+    public this (in string db_path)
+    {
+        const db_exists = db_path.exists;
+        if (db_exists)
+            log.info("Loading database from: {}", db_path);
+
+        this.db = new ManagedDatabase(db_path);
+
+        this.db.execute("CREATE TABLE IF NOT EXISTS scp_envelopes " ~
+            "(seq INTEGER PRIMARY KEY AUTOINCREMENT, envelope BLOB NOT NULL)");
+    }
+
+    /***************************************************************************
+
+        Store the envelope to the database.
+
+        First, clean with 'removeAll' before Adding it in new envelopes.
+
+        Params:
+            envelope = the envelop to add
+
+        Returns:
+            true if the envelope has been added to the database
+
+    ***************************************************************************/
+
+    public bool add (const ref SCPEnvelope envelope) @safe nothrow
+    {
+        static ubyte[] envelope_bytes;
+
+        try
+        {
+            envelope_bytes = serializeFull(envelope);
+        }
+        catch (Exception ex)
+        {
+            log.error("Serialization error: {}, Data was: {}", ex.msg,
+                envelope_bytes);
+            return false;
+        }
+
+        try
+        {
+            () @trusted {
+                db.execute("INSERT INTO scp_envelopes (envelope) VALUES(?)",
+                    envelope_bytes);
+            }();
+        }
+        catch (Exception ex)
+        {
+            log.error("Unexpected error while adding envelope: {}", ex.msg);
+            return false;
+        }
+        return true;
+    }
+
+    /***************************************************************************
+
+        Remove all envelopes from the database
+
+    ***************************************************************************/
+
+    public void removeAll () @trusted nothrow
+    {
+        try
+        {
+            this.db.execute("DELETE FROM scp_envelopes");
+        }
+        catch (Exception ex)
+        {
+            log.error("Error while calling SCPEnvelopeStore.removeAll(): {}", ex);
+        }
+    }
+
+    /***************************************************************************
+
+        Walk over the envelopes in the database
+        and call the provided delegate
+
+        Params:
+            dg = the delegate to call
+
+        Returns:
+            the status code of the delegate, or zero
+
+    ***************************************************************************/
+
+    public int opApply (scope int delegate(const ref SCPEnvelope) dg)
+    {
+        () @trusted
+        {
+            auto results = this.db.execute(
+                "SELECT envelope FROM scp_envelopes");
+
+            foreach (ref row; results)
+            {
+                auto env = deserializeFull!(const SCPEnvelope)(row.peek!(ubyte[])(0));
+
+                if (auto ret = dg(env))
+                    return ret;
+            }
+            return 0;
+        }();
+
+        return 0;
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the number of envelope in the database
+
+    ***************************************************************************/
+
+    public size_t length () @safe
+    {
+        return () @trusted {
+            return db.execute("SELECT count(*) FROM scp_envelopes")
+                .oneValue!size_t;
+        }();
+    }
+}
+
+/// add & opApply & remove tests
+unittest
+{
+    auto envelope_store = new SCPEnvelopeStore(":memory:");
+
+    SCPEnvelope[] envelopes;
+
+    foreach (_; 0 .. 2)
+    {
+        envelopes ~= SCPEnvelope.init;
+    }
+
+    foreach (env; envelopes)
+    {
+        envelope_store.add(env);
+    }
+
+    assert(envelope_store.length == 2);
+
+    foreach (const ref SCPEnvelope env; envelope_store)
+    {
+        assert(env == SCPEnvelope.init);
+    }
+
+    envelope_store.removeAll();
+    assert(envelope_store.length == 0);
+}
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -18,6 +18,7 @@ import agora.common.crypto.Key;
 import agora.common.crypto.Schnorr;
 import agora.common.Config;
 import agora.common.Hash : Hash, HashDg, hashPart, hashFull;
+import agora.common.ManagedDatabase;
 import agora.common.Serializer;
 import agora.common.Set;
 import agora.common.Task;
@@ -27,6 +28,7 @@ import agora.consensus.data.ConsensusData;
 import agora.consensus.data.ConsensusParams;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.Transaction;
+import agora.consensus.SCPEnvelopeStore;
 import agora.network.Clock;
 import agora.network.NetworkManager;
 import agora.node.Ledger;
@@ -48,6 +50,7 @@ import core.stdc.stdlib : abort;
 import core.stdc.time;
 
 import std.conv;
+import std.path : buildPath;
 import core.time : msecs, seconds;
 
 mixin AddLogger!();
@@ -95,6 +98,9 @@ public extern (C++) class Nominator : SCPDriver
     /// Note that Clock network synchronization is not yet implemented.
     private ITimer nomination_timer;
 
+    /// SCPEnvelopeStore instance
+    protected SCPEnvelopeStore scp_envelope_store;
+
 extern(D):
 
     /***************************************************************************
@@ -108,12 +114,13 @@ extern(D):
             key_pair = the key pair of this node
             ledger = needed for SCP state restoration & block validation
             taskman = used to run timers
+            data_dir = path to the data directory
 
     ***************************************************************************/
 
     public this (immutable(ConsensusParams) params, Clock clock,
         NetworkManager network, KeyPair key_pair, Ledger ledger,
-        TaskManager taskman)
+        TaskManager taskman, string data_dir)
     {
         this.params = params;
         this.clock = clock;
@@ -125,7 +132,8 @@ extern(D):
         this.scp = createSCP(this, node_id, IsValidator, no_quorum);
         this.taskman = taskman;
         this.ledger = ledger;
-        this.restoreSCPState(ledger);
+        this.scp_envelope_store = this.getSCPEnvelopeStore(data_dir);
+        this.restoreSCPState();
     }
 
     /***************************************************************************
@@ -379,50 +387,19 @@ extern(D):
 
     /***************************************************************************
 
-        Restore SCP's internal state based on the provided ledger state
-
-        Params:
-            ledger = the ledger instance
+        Restore SCP's internal state based on the stored latest envelopes
 
     ***************************************************************************/
 
-    private void restoreSCPState (Ledger ledger)
+    protected void restoreSCPState ()
     {
-        import agora.common.Serializer;
-        import scpd.types.Stellar_SCP;
-        import scpd.types.Utils;
-        import scpd.types.Stellar_types : StellarHash = Hash, NodeID;
-        import std.range;
-
-        auto pub_key = NodeID(uint256(PublicKey(this.schnorr_pair.V[])));
-        auto block = ledger.getBlocksFrom(ledger.getBlockHeight()).front;
-        auto serialized = block.serializeFull();
-        auto vec = serialized.toVec();
-        auto value = duplicate_value(&vec);
-        SCPStatement statement =
+        foreach (const ref SCPEnvelope envelope; this.scp_envelope_store)
         {
-            nodeID: pub_key,
-            slotIndex: block.header.height,
-            pledges: {
-                type_: SCPStatementType.SCP_ST_EXTERNALIZE,
-                externalize_: {
-                    commit: {
-                        counter: 0,
-                        value: value,
-                    },
-                    nH: 0,
-                },
-            },
-        };
-
-        SCPEnvelope env = SCPEnvelope(statement);
-        this.scp.setStateFromEnvelope(block.header.height, env);
-        if (!this.scp.isSlotFullyValidated(block.header.height))
-            assert(0);
-
-        // there should at least be a genesis block
-        if (this.scp.empty())
-            assert(0);
+            this.scp.setStateFromEnvelope(envelope.statement.slotIndex,
+                envelope);
+            if (!this.scp.isSlotFullyValidated(envelope.statement.slotIndex))
+                assert(0);
+        }
     }
 
     /***************************************************************************
@@ -464,6 +441,53 @@ extern(D):
 
         if (this.scp.receiveEnvelope(envelope) != SCP.EnvelopeState.VALID)
             log.info("Rejected invalid envelope: {}", scpPrettify(&envelope));
+    }
+
+    /***************************************************************************
+
+        Store the latest SCP sate for restore
+
+    ***************************************************************************/
+
+    public void storeLatestState () @safe
+    {
+        vector!SCPEnvelope envelopes;
+
+        () @trusted
+        {
+            if (this.scp.empty())
+                return;
+            envelopes = this.scp.getLatestMessagesSend(this.scp.getHighSlotIndex());
+        }();
+
+        ManagedDatabase.beginBatch();
+        scope (failure) ManagedDatabase.rollback();
+
+        // Clean the previous envelopes from the DB
+        this.scp_envelope_store.removeAll();
+
+        // Store the latest envelopes
+        foreach (const ref env; envelopes)
+            this.scp_envelope_store.add(env);
+
+        ManagedDatabase.commitBatch();
+    }
+
+    /***************************************************************************
+
+        Returns an instance of an SCPEnvelopeStore
+
+        Params:
+            data_dir = path to the data directory
+
+        Returns:
+            the SCPEnvelopeStore instance
+
+    ***************************************************************************/
+
+    protected SCPEnvelopeStore getSCPEnvelopeStore (string data_dir)
+    {
+        return new SCPEnvelopeStore(buildPath(data_dir, "scp_envelopes.dat"));
     }
 
     extern (C++):
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -87,7 +87,8 @@ public class Validator : FullNode, API
             this.params.QuorumThreshold);
 
         this.nominator = this.getNominator(this.params, this.clock,
-            this.network, this.config.validator.key_pair, this.ledger, this.taskman);
+            this.network, this.config.validator.key_pair, this.ledger, this.taskman,
+            this.config.node.data_dir);
 
         // currently we are not saving preimage info,
         // we only have the commitment in the genesis block
@@ -274,6 +275,7 @@ public class Validator : FullNode, API
             key_pair = the key pair of the node
             ledger = Ledger instance
             taskman = the task manager
+            data_dir = path to the data directory
 
         Returns:
             An instance of a `Nominator`
@@ -282,9 +284,10 @@ public class Validator : FullNode, API
 
     protected Nominator getNominator (immutable(ConsensusParams) params,
         Clock clock, NetworkManager network, KeyPair key_pair, Ledger ledger,
-        TaskManager taskman)
+        TaskManager taskman, string data_dir)
     {
-        return new Nominator(params, clock, network, key_pair, ledger, taskman);
+        return new Nominator(params, clock, network, key_pair, ledger, taskman,
+            data_dir);
     }
 
     /***************************************************************************
@@ -389,5 +392,17 @@ public class Validator : FullNode, API
             this.pushPreImage(preimage);
             this.enroll_man.updateRevealDistance(this.ledger.getBlockHeight());
         }
+    }
+
+    /***************************************************************************
+
+        Calls the base class `shutdown` and store the latest SCP state
+
+    ***************************************************************************/
+
+    override void shutdown ()
+    {
+        super.shutdown();
+        this.nominator.storeLatestState();
     }
 }
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -43,6 +43,7 @@ import agora.consensus.data.UTXOSetValue;
 import agora.consensus.UTXOSet;
 import agora.consensus.EnrollmentManager;
 import agora.consensus.data.genesis.Test;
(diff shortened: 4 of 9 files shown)


TrustHenry:
I have set sqlite as memory but I will replace it with a file.
I have an unknown error during the unit test.
Could you find a problem with my draftPR?

source/agora/consensus/LatestEnvelopes.d
```diff
@@ -0,0 +1,165 @@
...
+import agora.common.ManagedDatabase;
+import agora.common.Serializer;
+import agora.common.Types;
+import agora.utils.Log;
+
+import d2sqlite3.library;
+import d2sqlite3.results;
+import d2sqlite3.sqlite3;
+
+import core.stdc.inttypes;
```
TrustHenry: It can be removed now.

AndrejMitrovic:
> I have an unknown error during the unit test.

Run your unittests with LLDB and when it segfaults type `bt` and hit enter.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -123,7 +128,8 @@ extern(D):
         this.scp = createSCP(this, node_id, IsValidator, no_quorum);
         this.taskman = taskman;
         this.ledger = ledger;
-        this.restoreSCPState(ledger);
+        this.restoreSCPState();
+        this.latest_envelopes = this.getLatestEnvelopes(":memory:");
```
AndrejMitrovic: I'll give you a hint: You are restoring SCP state, and then retrieving the latest envelopes after.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -451,8 +429,38 @@ extern(D):
             return;  // slot was already externalized, ignore outdated message
         }
 
-        if (this.scp.receiveEnvelope(envelope) != SCP.EnvelopeState.VALID)
+        if (this.scp.receiveEnvelope(envelope) == SCP.EnvelopeState.VALID)
+        {
+            vector!SCPEnvelope envelopes =
+                this.scp.getCurrentState(envelope.statement.slotIndex);
```
AndrejMitrovic: I thought about this more. I think it's inefficient if we do this every time we receive an envelope.

Instead we should do it:
- On node shutdown
- On node crash

For the first part I suggest adding an override of `shutdown` in the `Validator` class:

```
override void shutdown { }
{
    super.shutdown();
    this.nominator.storeLatestState();
}
```

And implementing `storeLatestState` which will store the envelopes to the database.

For node crashes we actually need to implement a crash / segfault event handler, however this is a separate problem and should not be solved as part of #1170.

TrustHenry: Yes, that's right.
I thought it was inefficient when I saw the frequency of calls.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -377,56 +383,28 @@ extern(D):
...
+            this.scp.setStateFromEnvelope(env.statement.slotIndex, env);
+            if (!this.scp.isSlotFullyValidated(env.statement.slotIndex))
+                assert(0);
+        }
 
         // there should at least be a genesis block
-        if (this.scp.empty())
-            assert(0);
+        // if (this.scp.empty())
+        //     assert(0);
```
TrustHenry: This verification is essential?

AndrejMitrovic: This can be removed. For the genesis block we won't have any state in SCP yet, and therefore this check is unnecessary.

source/agora/test/SCPState.d
```diff
@@ -0,0 +1,156 @@
...
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto re_validator = network.clients[0];
+    auto txes = genesisSpendable().map!(txb => txb.sign()).array();
+    txes.each!(tx => re_validator.putTransaction(tx));
+    network.expectBlock(Height(1), 2.seconds);
+
+    // Now shut down & restart one node
+    network.restart(re_validator);
```
TrustHenry: I think I need to check more when node shutdown.

source/agora/consensus/LatestEnvelopes.d
```diff
@@ -0,0 +1,204 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.LatestEnvelopes;
```
AndrejMitrovic: `LatestEnvelope` is not a data type. So I think this is not a good name.

How about `agora.consensus.SCPEnvelopeDB` or `agora.consensus.SCPEnvelopeStore`?

source/agora/consensus/LatestEnvelopes.d
```diff
@@ -0,0 +1,204 @@
...
+    public this (in string db_path)
+    {
+        const db_exists = db_path.exists;
+        if (db_exists)
+            log.info("Loading database from: {}", db_path);
+
+        this.db = new ManagedDatabase(db_path);
+
+        this.db.execute("CREATE TABLE IF NOT EXISTS store_envelopes " ~
+            "(seq INTEGER PRIMARY KEY AUTOINCREMENT, envelope BLOB NOT NULL)");
```
AndrejMitrovic: This is wrong. This must be the slot index. It doesn't always start from zero. Don't use autoincrement.

source/agora/consensus/LatestEnvelopes.d
```diff
@@ -0,0 +1,204 @@
...
+            SCPEnvelope envelope;
+            dg(envelope);
+        }
+        return 0;
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the number of envelope in the latestEnvelopes
```
AndrejMitrovic:
```suggestion
            the number of envelopes in the database
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -455,6 +429,46 @@ extern(D):
...
+
+        Params:
+            data_dir = path to the data directory
+
+        Returns:
+            the latestEnvelopes instance
+
+    ***************************************************************************/
+
+    protected LatestEnvelopes getLatestEnvelopes (string data_dir)
```
AndrejMitrovic: This will also have to be renamed when you rename the module & class. Something like `getSCPEnvelopeDB`

source/agora/node/Validator.d
```diff
@@ -360,4 +362,12 @@ public class Validator : FullNode, API
             this.enroll_man.updateRevealDistance(this.ledger.getBlockHeight());
         }
     }
+
+    /// Ditto
+    override void shutdown ()
+    {
+        super.shutdown();
+        const slot_idx = this.ledger.getBlockHeight() + 1;
```
AndrejMitrovic: This is wrong. We should just get the latest state. Please make another commit where the slot index is stored in the Nominator as a field:

https://github.com/bpfkorea/agora/blob/e704747aec3c681976d50127ac507e15e524360c/source/agora/consensus/protocol/Nominator.d#L314

Then, you can just call `this.nominator.storeLatestState()` and all the logic will be contained within it.

source/agora/consensus/LatestEnvelopes.d
```diff
@@ -0,0 +1,204 @@
+/*******************************************************************************
+
+    Contains supporting code for restoring the latest envelopes,
```
AndrejMitrovic:
```suggestion
    Contains supporting code for restoring the latest SCP envelopes,
```

source/agora/consensus/LatestEnvelopes.d
```diff
@@ -0,0 +1,204 @@
...
+
+        this.db = new ManagedDatabase(db_path);
+
+        this.db.execute("CREATE TABLE IF NOT EXISTS store_envelopes " ~
+            "(seq INTEGER PRIMARY KEY AUTOINCREMENT, envelope BLOB NOT NULL)");
+    }
+
+    /***************************************************************************
+
+        Add a envelope to the latestEnvelopes
```
AndrejMitrovic:
```suggestion
        Store the envelope to the database.
````

File: source/agora/consensus/LatestEnvelopes.d [hunk 13]
```diff
@@ -0,0 +1,204 @@
...
+
+        First, clean with 'removeAll' before Adding it in new envelopes.
+
+        Params:
+            tx = the latestEnvelopes to add
```
AndrejMitrovic: Wrong argument name `tx`

source/agora/consensus/protocol/Nominator.d
```diff
@@ -455,6 +429,48 @@ extern(D):
...
+        this.latest_envelopes.removeAll();
+
+        // Store the lastest envelopes
+        foreach(const ref env; envelopes)
+            this.latest_envelopes.add(env);
+    }
+
+    /***************************************************************************
+
+        Returns an instance of a LatestEnvelopes
```
TrustHenry: I added the missing document.

source/agora/consensus/LatestEnvelopes.d
```diff
@@ -0,0 +1,204 @@
...
+
+        Remove all envelopes from the latestEnvelopes
+
+    ***************************************************************************/
+
+    public void removeAll () @trusted nothrow
+    {
+        try
+        {
+            this.db.execute("DELETE FROM store_envelopes");
```
AndrejMitrovic: Shouldn't this be:

```suggestion
            this.db.execute("DELETE * FROM store_envelopes");
```

I guess both work?

source/agora/consensus/LatestEnvelopes.d
```diff
@@ -0,0 +1,204 @@
...
+
+    public this (in string db_path)
+    {
+        const db_exists = db_path.exists;
+        if (db_exists)
+            log.info("Loading database from: {}", db_path);
+
+        this.db = new ManagedDatabase(db_path);
+
+        this.db.execute("CREATE TABLE IF NOT EXISTS store_envelopes " ~
```
AndrejMitrovic:
```suggestion
        this.db.execute("CREATE TABLE IF NOT EXISTS scp_envelopes " ~
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -106,12 +111,13 @@ extern(D):
             key_pair = the key pair of this node
             ledger = needed for SCP state restoration & block validation
             taskman = used to run timers
+            data_dir = path to the data dirctory
```
```suggestion
            data_dir = path to the data directory
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -455,6 +429,48 @@ extern(D):
...
+            data_dir = path to the data directory
+
+        Returns:
+            the latestEnvelopes instance
+
+    ***************************************************************************/
+
+    protected LatestEnvelopes getLatestEnvelopes (string data_dir)
+    {
+        return new LatestEnvelopes(buildPath(data_dir, "store_envelopes.dat"));
```
```suggestion
        return new LatestEnvelopes(buildPath(data_dir, "scp_envelopes.dat"));
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -455,6 +429,48 @@ extern(D):
...
+
+    public void storeLatestState (ulong slot_index) @safe
+    {
+        vector!SCPEnvelope envelopes;
+        () @trusted
+        {
+            envelopes = this.scp.getCurrentState(slot_index);
+        }();
+
+        // Clean the lastest envelopes DB
```
```suggestion
        // Clean the previous envelopes from the DB
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -455,6 +429,48 @@ extern(D):
...
+        this.latest_envelopes.removeAll();
+
+        // Store the lastest envelopes
```
```suggestion
        // Store the latest envelopes
```

source/agora/node/Validator.d
```diff
@@ -280,9 +281,10 @@ public class Validator : FullNode, API
 
     protected Nominator getNominator (immutable(ConsensusParams) params,
         Clock clock, NetworkManager network, KeyPair key_pair, Ledger ledger,
-        TaskManager taskman)
+        TaskManager taskman, string data_dir)
```
AndrejMitrovic, line 287: Needs parameter documentation.

source/agora/node/Validator.d
```diff
@@ -360,4 +362,12 @@ public class Validator : FullNode, API
             this.enroll_man.updateRevealDistance(this.ledger.getBlockHeight());
         }
     }
+
+    /// Ditto
```
AndrejMitrovic: What do you mean `/// Ditto`? That means `shutdown` becomes part of the `checkRevealPreimage` documentation because it's the previous symbol. Please document the function properly.

source/agora/test/Base.d
```diff
@@ -440,6 +441,12 @@ extern(D):
 
         return true;
     }
+
+    /// Ditto
```
AndrejMitrovic: Wrong Ditto, please document this properly.

source/agora/test/SCPState.d
```diff
@@ -0,0 +1,157 @@
+/*******************************************************************************
+
+    SCPState Related Test
```
AndrejMitrovic:
```suggestion
    Tests restoring SCP Envelope state on restart
```

source/agora/test/SCPState.d
```diff
@@ -0,0 +1,157 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.SCPState;
```
AndrejMitrovic:
```suggestion
module agora.test.RestoreSCPState;
```

source/agora/test/SCPState.d
```diff
@@ -0,0 +1,157 @@
...
+            TaskManager taskman, string data_dir, ulong txs_to_nominate)
+        {
+            super(params, clock, network, key_pair, ledger, taskman, data_dir,
+                txs_to_nominate);
+        }
+
+        ///
+        override void restoreSCPState ()
+        {
+            uint i;
```
AndrejMitrovic: You should add `assert(this.latest_envelopes.length > 0);` or `assert(i > 0)` after the foreach.

AndrejMitrovic (changes requested):
First round of review done. I will review more after the fixes.

source/agora/consensus/LatestEnvelopes.d
TrustHenry: I've been thinking about a lot of names.
It's good name `agora.consensus.SCPEnvelopeStore`

source/agora/consensus/LatestEnvelopes.d
TrustHenry: `seq` is a meaningless unique key.
When checking the data in the envelope in the test code, the envelopes in the nodes in the same slot_index were identified.
If so, slot_idx cannot be used as a unique key.
slot_index is included in SCEnvelope.SCPStatement structure.
Is there a reason to use slot_index as a key?

AndrejMitrovic: Ah it's in the statement. Yes you're right. Then it should be fine.

source/agora/consensus/LatestEnvelopes.d
TrustHenry: In standard SQL, most of the time,
Isn't it `DELETE FROM table name`?
Can't have * in a DELETE statement.

source/agora/node/Validator.d
TrustHenry: I was thinking about this.
Is there any way to get the latest slot_index in SCP.d?

Are these functions related?
https://github.com/bpfkorea/agora/blob/f7e514747f94a223b0d3ab99711220bd247e6657/source/scpd/scp/SCP.d#L117-L120

source/agora/consensus/protocol/Nominator.d
TrustHenry: All names are being changed.

source/agora/node/Validator.d
TrustHenry: https://github.com/bpfkorea/agora/blob/e704747aec3c681976d50127ac507e15e524360c/source/agora/consensus/protocol/Nominator.d#L314
If I put this slot_idx in a variable, does it stay up to date?

AndrejMitrovic: Yes you can use `getHighSlotIndex`.

TrustHenry:
@AndrejMitrovic Thank you for your review.
Apply the suggestion now.
names and documents have been updated.
I have changed the reference place for slot_index.

source/agora/test/SCPState.d
TrustHenry: Yes, there will be an error with the corresponding assert.
I'am looking for the cause.

source/agora/test/RestoreSCPState.d
```diff
@@ -0,0 +1,159 @@
...
+            TaskManager taskman, string data_dir, ulong txs_to_nominate)
+        {
+            super(params, clock, network, key_pair, ledger, taskman, data_dir,
+                txs_to_nominate);
+        }
+
+        ///
+        override void restoreSCPState ()
+        {
+            //assert(this.scp_envelope_store.length > 0);
```
AndrejMitrovic: The reason it fails is because the node instantiates a new in-memory database when it restarts. It's empty.

TrustHenry: Yes, that's right.
I need to keep an instance.

TrustHenry: And `libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: Cannot set state after starting ballot protocol`
Oh!...

TrustHenry: I don't know if the ballot protocol started.
I think need to make an exception.

AndrejMitrovic: I don't see this error anywhere in the CI.

AndrejMitrovic: And, you can only restore SCP state before nominations, before the event loop starts.

TrustHenry: I checked the message after testing it locally with FileDB.
I will keep the DB instance above and try to reproduce it.

Geod24:
My understanding is that this PR still needs work and will carry over to the next sprint, correct ?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -466,6 +441,48 @@ extern(D):
...
+
+    public void storeLatestState () @safe nothrow
+    {
+        vector!SCPEnvelope envelopes;
+
+        () @trusted
+        {
+            if (this.scp.empty())
+                return;
+            envelopes = this.scp.getLatestMessagesSend(this.scp.getHighSlotIndex());
```
TrustHenry, line 460: Changed from getCurrentState() to getLatestMessagesSend()
getLatestMessagesSend is returns the latest messages sent for the given slot.

TrustHenry:
> My understanding is that this PR still needs work and will carry over to the next sprint, correct ?

Yes, It is necessary to verify that SCP has implemented as intended.

source/agora/consensus/protocol/Nominator.d
TrustHenry: The problem is not visible in this version.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -466,6 +441,48 @@ extern(D):
...
+        }();
+
+        // Clean the previous envelopes from the DB
+        this.scp_envelope_store.removeAll();
+
+        // Store the latest envelopes
+        foreach(const ref env; envelopes)
```
```suggestion
        foreach (const ref env; envelopes)
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -466,6 +441,48 @@ extern(D):
...
+
+        () @trusted
+        {
+            if (this.scp.empty())
+                return;
+            envelopes = this.scp.getLatestMessagesSend(this.scp.getHighSlotIndex());
+        }();
+
+        // Clean the previous envelopes from the DB
+        this.scp_envelope_store.removeAll();
```
AndrejMitrovic, line 467: I think this and the adding needs to be atomic with `db.begin` / `db.commit`. If removing succeeds but adding new envelopes fails, then we will have lost the old envelope messages.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -466,6 +441,48 @@ extern(D):
...
+
+        // Store the latest envelopes
+        foreach(const ref env; envelopes)
+            this.scp_envelope_store.add(env);
+    }
+
+    /***************************************************************************
+
+        Returns an instance of a SCPEnvelopeStore
```
```suggestion
        Returns an instance of an SCPEnvelopeStore
```

(English is difficult, I know. :/
"A" and "An" is used based on pronunciation).

source/agora/consensus/protocol/Nominator.d
```diff
@@ -466,6 +441,48 @@ extern(D):
...
+
+        Params:
+            data_dir = path to the data directory
+
+        Returns:
+            the latestEnvelopes instance
```
```suggestion
            the SCPEnvelopeStore instance
```

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: Just as a note to readers: We discussed this IRL. Stellar does the same thing here, using `getLatestMessagesSend()` instead of `getCurrentState`. It does mean that messages from other nodes are not recovered. However, the node will accept **new** messages from other nodes later when they are sent. It's only its own internal state that it needs to recover - IOW the node's own messages that it gossips.

This may not be very clear to people unless they understand SCP well.

TrustHenry:
I did rebase done.

source/agora/consensus/SCPEnvelopeStore.d
```diff
@@ -0,0 +1,199 @@
+/*******************************************************************************
+
+    Contains supporting code for restoring the latest SCP envelopes,
+    using SQLite as a backing store.
```
AndrejMitrovic:
```suggestion
    Contains supporting code for storing the latest SCP envelopes,
    using SQLite as a backing store.
```

This code does not deal with restoring anything, just storing the data (being able to load it is implicit). The Nominator is in charge of doing the work of restoring.

AndrejMitrovic (changes requested):

source/agora/consensus/protocol/Nominator.d
TrustHenry: That's a good point.

source/agora/consensus/SCPEnvelopeStore.d
TrustHenry: Yes

source/agora/consensus/protocol/Nominator.d
TrustHenry: I apply the suggestion.

source/agora/test/RestoreSCPState.d
```diff
@@ -0,0 +1,171 @@
...
+    {
+        return store.length;
+    }
+}
+
+/// A test to store SCP state and recover SCP state
+unittest
+{
+    __gshared bool CheckEnvelopeStore = false;
+
```
AndrejMitrovic, line 83:
```suggestion
__gshared bool Checked = false;
```

source/agora/test/RestoreSCPState.d
```diff
@@ -0,0 +1,171 @@
...
+        {
+            super(params, clock, network, key_pair, ledger, taskman, data_dir,
+                txs_to_nominate);
+        }
+
+        ///
+        override void restoreSCPState ()
+        {
+            if (CheckEnvelopeStore)
+                assert(TestSCPEnvelopeStore.store.length == 2);
```
AndrejMitrovic:
```suggestion
            // on first boot the envelope store will be empty,
            // but on restart it should contain two envelopes
            // (nominate & ballot protocol messages)
            if (CheckEnvelopeStore)
            {
                assert(TestSCPEnvelopeStore.store.length == 2);
                Checked = true;
            }
```

source/agora/test/RestoreSCPState.d
```diff
@@ -0,0 +1,171 @@
...
+    auto re_validator = network.clients[0];
+    auto txes = genesisSpendable().map!(txb => txb.sign()).array();
+    txes.each!(tx => re_validator.putTransaction(tx));
+    network.expectBlock(Height(1), 2.seconds);
+
+    CheckEnvelopeStore = true;
+    // Now shut down & restart one node
+    network.restart(re_validator);
+    network.waitForDiscovery();
+    network.expectBlock(Height(1), 5.seconds);
```
AndrejMitrovic, line 176:
```suggestion
    network.expectBlock(Height(1), 5.seconds);
    assert(Checked);
```

AndrejMitrovic (changes requested):
One final set of changes requested.

source/agora/test/RestoreSCPState.d
```diff
@@ -0,0 +1,171 @@
...
+import geod24.Registry;
+
+import scpd.Cpp;
+import scpd.types.Stellar_SCP;
+
+import core.stdc.time;
+import core.thread;
+
+/// Class containing gshared store for SCPEnvelopeStoreTest
+public class TestSCPEnvelopeStore :  SCPEnvelopeStore
```
linked0:
```suggestion
public class TestSCPEnvelopeStore : SCPEnvelopeStore
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -379,50 +387,18 @@ extern(D):
...
-                    commit: {
-                        counter: 0,
-                        value: value,
-                    },
-                    nH: 0,
-                },
-            },
-        };
-
-        SCPEnvelope env = SCPEnvelope(statement);
```
linked0, line 418: How about using the name of `envelope`. The `env` make me feel like 'environment'. But it depends on you.

linked0 (approved):
LGTM

source/agora/test/RestoreSCPState.d
```diff
@@ -0,0 +1,178 @@
...
+        ///
+        override void restoreSCPState ()
+        {
+            // on first boot the envelope store will be empty,
+            // but on restart it should contain two envelopes
+            // (nominate & ballot protocol messages)
+            if (Checked)
+            {
+                assert(TestSCPEnvelopeStore.store.length == 2);
+                Checked = true;
```
AndrejMitrovic: There was some misunderstanding. I wanted you to use two booleans. However we can use a single boolean too. But, we need to change the code once more.

```suggestion
                Checked = false;
```

source/agora/test/RestoreSCPState.d
```diff
@@ -0,0 +1,178 @@
...
+    auto txes = genesisSpendable().map!(txb => txb.sign()).array();
+    txes.each!(tx => re_validator.putTransaction(tx));
+    network.expectBlock(Height(1), 2.seconds);
+
+    Checked = true;
+    // Now shut down & restart one node
+    network.restart(re_validator);
+    network.waitForDiscovery();
+    network.expectBlock(Height(1), 5.seconds);
+    assert(Checked);
```
AndrejMitrovic:
```suggestion
    assert(!Checked);
```

AndrejMitrovic (changes requested):
There was a misunderstanding.

source/agora/test/RestoreSCPState.d
TrustHenry: Yes, that's right.

TrustHenry:
@AndrejMitrovic @linked0  Thanks for the review.
Apply the suggestions.

AndrejMitrovic (approved):

AndrejMitrovic:
> Related to #1170

@TrustHenry please use `Fixes #1170` in the PR description and the commit message so the associated issue is automatically closed in the future.

PR #1207, opened 2020-09-16
Emit hidden version of a secret key in log.trace

### LINKED ISSUE #719: Logging should not expose any secret keys
```
node-2_1_37a10f21eb7b | 2020-04-02T01:15:19.329464369Z 2020-04-02 01:15:19,329 Trace [agora.node.Node] - Config is: { banman: { max_failed_requests: 10, ban_duration: 86400 }, node: { is_validator: true, min_listeners: 2, max_listeners: 10, address: "0.0.0.0", port: 4826, key_pair: { address: GA3DMXTREDC4AIUTHRFIXCKWKF7BDIXRWM2KLV74OPK2OKDM2VJ235GN, secret: 0xad53d56c28a7d573fcd7a534b3f1a2117e5156898b4a3c9322c0c520715e36364bdcf8f0442d0de2ff5f954e092c4d1cbc43ae832b336eb7e80da9a7ba45c911, seed: SAI4SRN2U6UQ32FXNYZSXA5OIO6BYTJMBFHJKX774IGS2RHQ7DOEW5SJ }, retry_delay: 1500, max_retries: 5, timeout: 1000, data_dir: ".cache" }, admin: { enabled: false, address: "127.0.0.1", port: 2827 }, network: ["http://node-0:2826", "http://node-1:3826"], dns_seeds: [], quorum: { threshold: 2, nodes: [GBUVRIIBMHKC4PE6BK7MO2O26U2NJLW4WGGWKLAVLAA2DLFZTBHHKOEK, GBJABNUCDJCIL5YJQMB5OZ7VCFPKYLMTUXM2ZKQJACT7PXL7EVOMEKNZ, GA3DMXTREDC4AIUTHRFIXCKWKF7BDIXRWM2KLV74OPK2OKDM2VJ235GN], quorums: [] }, logging: { log_level: 0 } }
```

> secret: 0xad53d56c28a7d573fcd7a534b3f1a2117e5156898b4a3c9322c0c520715e36364bdcf8f0442d0de2ff5f954e092c4d1cbc43ae832b336eb7e80da9a7ba45c911, seed: SAI4SRN2U6UQ32FXNYZSXA5OIO6BYTJMBFHJKX774IGS2RHQ7DOEW5SJ

This is a security issue. If someone finds a bug in Agora and wants to file a bug then they might provide us with a log dump. The log should not emit any private keys.

Rather than just removing its output, I can think of alternatives:
- Emit a hashed version of the key. It may help with logging in case that key appears in multiple error messages.
- Emit a shortened pretty-printed version of the key.

@Geod24 any thoughts?

kchulj:
Logging should not expose any secret keys. Therefore, `log.trace` now emits a hidden version of `SecretKey`, `Seed`, and `Scalar` by default (`PrintMode.Obfuscated`). Switching the argument to `PrintMode.Clear` argument to `toString` prints the original value.

`PrintMode.Obfuscated`:
```
validator: { enabled: true, key_pair: { address: GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ, secret: **SECRET**, seed: **SEED** }
```
`PrintMode.Clear`:
```
validator: { enabled: true, key_pair: { address: GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ, secret: 0xf52201b3d296424f285ca01391640951cd6a319b9d8ad6c60b4071a1e038029d94b84c9c7e3c72d9c836046c5cd87b5beb6c8e6498ea1b757207e7c22d29c5a7, seed: SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4 } 
```
Fixes #719

### DIFF
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -24,6 +24,8 @@ import agora.common.Types;
 import geod24.bitblob;
 import libsodium;
 
+import std.format;
+
 ///
 nothrow @nogc unittest
 {
@@ -90,17 +92,58 @@ public struct Scalar
         crypto_core_ed25519_scalar_reduce(this.data[].ptr, param[].ptr);
     }
 
-    /// Expose `toString`
-    public void toString (scope void delegate(const(char)[]) @safe sink)
-        const @safe
+    /***************************************************************************
+
+        Print the scalar
+
+        By default, this function prints the hidden version of the scalar.
+        Changing the mode to `Clear` makes it print the original value.
+        This mode is intended for debugging.
+
+        Params:
+          sink = The sink to write the piecemeal string data to
+          mode = The `PrintMode` to use for printing the content.
+                 By default, the hidden value is printed.
+
+    ***************************************************************************/
+
+    public void toString (scope void delegate(const(char)[]) @safe sink,
+                          PrintMode mode = PrintMode.Obfuscated) const @safe
     {
-        this.data.toString(sink);
+        final switch (mode)
+        {
+        case PrintMode.Obfuscated:
+            formattedWrite(sink, "**SCALAR**");
+            break;
+
+        case PrintMode.Clear:
+            formattedWrite(sink, "%s", this.data);
+            break;
+        }
     }
 
     /// Ditto
-    public string toString () const @safe
+    public string toString (PrintMode mode = PrintMode.Obfuscated) const @safe
     {
-        return this.data.toString();
+        string result;
+        this.toString((data) { result ~= data; }, mode);
+        return result;
+    }
+
+    ///
+    unittest
+    {
+        auto s = Scalar("0x0e00a8df701806cb4deac9bb09cc85b097ee713e055b9d2bf1daf668b3f63778");
+
+        assert(s.toString(PrintMode.Obfuscated) == "**SCALAR**");
+        assert(s.toString(PrintMode.Clear) ==
+               "0x0e00a8df701806cb4deac9bb09cc85b097ee713e055b9d2bf1daf668b3f63778");
+
+        // Test default formatting behavior with writeln, log, etc...
+        import std.format : phobos_format = format;
+        import ocean.text.convert.Formatter : ocean_format = format;
+        assert(phobos_format("%s", s) == "**SCALAR**");
+        assert(ocean_format("{}", s) == "**SCALAR**");
     }
 
     /// Vibe.d deserialization
@@ -207,7 +250,7 @@ public struct Scalar
 @safe unittest
 {
     static immutable string s = "0x1000000000000000000000000000000014def9dea2f79cd65812631a5cf5d3ec";
-    assert(Scalar.fromString(s).toString == s);
+    assert(Scalar.fromString(s).toString(PrintMode.Clear) == s);
 }
 
 // Test valid Scalars
--- a/source/agora/common/crypto/ECC.d
+++ b/source/agora/common/crypto/ECC.d
@@ -91,10 +91,10 @@ public struct Scalar
     }
 
     /// Expose `toString`
-    public void toString (scope void delegate(const(char)[]) @safe dg)
+    public void toString (scope void delegate(const(char)[]) @safe sink)
         const @safe
     {
-        this.data.toString(dg);
+        this.data.toString(sink);
     }
 
     /// Ditto
@@ -178,7 +178,7 @@ public struct Scalar
     }
 
     /// Return the point corresponding to this scalar multiplied by the generator
-    public Point toPoint () const nothrow @nogc @trusted 
+    public Point toPoint () const nothrow @nogc @trusted
     {
         Point ret = void;
         if (crypto_scalarmult_ed25519_base_noclamp(ret.data[].ptr, this.data[].ptr) != 0)
--- a/source/agora/common/crypto/Key.d
+++ b/source/agora/common/crypto/Key.d
@@ -23,6 +23,7 @@ import base32;
 import libsodium;
 
 import std.exception;
+import std.format;
 
 
 /// Simple signature example
@@ -260,6 +261,63 @@ public struct SecretKey
 
     /***************************************************************************
 
+        Print the secret key
+
+        By default, this function prints the hidden version of the secret key.
+        Changing the mode to `Clear` makes it print the original value.
+        This mode is intended for debugging.
+
+        Params:
+          sink = The sink to write the piecemeal string data to
+          mode = The `PrintMode` to use for printing the content.
+                 By default, the hidden value is printed.
+
+    ***************************************************************************/
+
+    public void toString (scope void delegate(const(char)[]) @safe sink,
+                          PrintMode mode = PrintMode.Obfuscated) const @safe
+    {
+        final switch (mode)
+        {
+        case PrintMode.Obfuscated:
+            formattedWrite(sink, "**SECRET**");
+            break;
+
+        case PrintMode.Clear:
+            formattedWrite(sink, "%s", this.data);
+            break;
+        }
+    }
+
+    /// Ditto
+    public string toString (PrintMode mode = PrintMode.Obfuscated) const @safe
+    {
+        string result;
+        this.toString((data) { result ~= data; }, mode);
+        return result;
+    }
+
+    ///
+    unittest
+    {
+        auto sk = KeyPair.fromSeed(Seed.fromString(
+                  "SDQJFVGII75JPFVRDNFJ55L7DE4H7V3RHXLRKW55NDPBRBYUZLTW4TJS"))
+                  .secret;
+
+        assert(sk.toString(PrintMode.Obfuscated) == "**SECRET**");
+        assert(sk.toString(PrintMode.Clear) ==
+               "0xbca6b465149f8d3f4bc1794c246b2df64dcab0e65beb1862c5a0843d3e06e2f5" ~
+               "6ee7ca148718de68bd5b15d73d71d77f38197ff59e4a1bb19697fa47c8d492e0");
+
+        // Test default formatting behavior with writeln, log, etc...
+        import std.format : phobos_format = format;
+        import ocean.text.convert.Formatter : ocean_format = format;
+        assert(phobos_format("%s", sk) == "**SECRET**");
+        assert(ocean_format("{}", sk) == "**SECRET**");
+    }
+
+    /***************************************************************************
+
         Signs a message with this private key
 
         Params:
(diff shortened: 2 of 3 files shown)


source/agora/common/crypto/ECC.d
```diff
@@ -90,11 +91,29 @@ public struct Scalar
         crypto_core_ed25519_scalar_reduce(this.data[].ptr, param[].ptr);
     }
 
+    public enum PrintMode
```
AndrejMitrovic: Needs `///` style comment. And also a description of what each mode does in the comment.

source/agora/common/crypto/Key.d
```diff
@@ -259,6 +260,30 @@ public struct SecretKey
         this.data = args;
     }
 
+    public enum PrintMode
```
AndrejMitrovic: Also needs a comment.

source/agora/common/crypto/Key.d
```diff
@@ -259,6 +260,30 @@ public struct SecretKey
...
+    {
+        Obfuscated,
+        Clear,
+    }
+
+    public void toString (scope void delegate(const(char)[]) @safe sink,
```
AndrejMitrovic, line 277: Also needs documentation.

source/agora/common/crypto/Key.d
```diff
@@ -311,6 +336,12 @@ public struct Seed
         this.data = args;
     }
 
+    public enum PrintMode
```
AndrejMitrovic: Also documentation needed.

source/agora/consensus/PreImage.d
```diff
@@ -75,7 +75,7 @@ public struct PreImageCache
     ***************************************************************************/
 
     public void toString (scope void delegate(scope const(char)[]) @safe sink,
-                          PrintMode mode = PrintMode.Bounds) const
+                          PrintMode mode = PrintMode.Bounds) const @safe
```
AndrejMitrovic: This belongs in its own commit.

source/agora/common/crypto/Key.d
```diff
@@ -271,7 +296,7 @@ public struct SecretKey
 
     ***************************************************************************/
 
-    public Signature sign (scope const(ubyte)[] msg) const
+    public Signature sign (scope const(ubyte)[] msg) const nothrow @nogc
```
AndrejMitrovic, line 331: This should ideally be moved into its own commit. So one commit replacing the `nothrow @nogc:` with this, and a separate commit which only implements what the issue is about.

AndrejMitrovic (dismissed):
Where are the unittests?

source/agora/common/crypto/ECC.d
```diff
@@ -20,6 +20,7 @@ module agora.common.crypto.ECC;
 import agora.common.Hash;
```
AndrejMitrovic: The commit message needs a description and a note of which issue it fixes. For example:

```
Emit hashed version of a secret key by default

To avoid leaking private keys into logs, we emit a shortened hashed version of the key. 
The behavior can toggled off with the `PrintMode` argument to `toString`.

Fixes #719
```

source/agora/common/crypto/ECC.d
```diff
@@ -90,11 +91,46 @@ public struct Scalar
         crypto_core_ed25519_scalar_reduce(this.data[].ptr, param[].ptr);
     }
 
-    /// Expose `toString`
-    public void toString (scope void delegate(const(char)[]) @safe sink)
-        const @safe
+    /// Describe print modes for `toString`
+    public enum PrintMode
+    {
+        /// Default value, print hashed version
```
AndrejMitrovic:
```suggestion
        /// Print shortened hashed version
```

The default value is in the parameter itself. And users know that the first member of an enum is the enum's `.init` value, it doesn't need to be explained here.

source/agora/common/crypto/ECC.d
```diff
@@ -90,11 +91,46 @@ public struct Scalar
...
+            this.data[].hashFull().prettify().toString(sink);
+            break;
+
+        case PrintMode.Clear:
+            this.data.toString(sink);
+            break;
+
+        default:
+            assert(0);
+        }
```
AndrejMitrovic:
```suggestion
        final switch (mode)
        {
        case PrintMode.Obfuscated:
            this.data[].hashFull().prettify().toString(sink);
            break;

        case PrintMode.Clear:
            this.data.toString(sink);
            break;
        }
```

Final switch is a feature in D. If you have a missing `case` it will error at compile-time, rather than having the exception being thrown at runtime.

source/agora/common/crypto/Key.d
```diff
@@ -258,6 +260,49 @@ public struct SecretKey
...
+          sink = The sink to write the piecemeal string data to
+          mode = The `PrintMode` to use for printing the content.
+                 By default, the hashed value is printed.
+
+    ***************************************************************************/
+
+    public void toString (scope void delegate(const(char)[]) @safe sink,
+                          PrintMode mode = PrintMode.Obfuscated) const @safe
+    {
+        switch (mode)
```
AndrejMitrovic: Ditto please use final switch here.

source/agora/common/crypto/Key.d
```diff
@@ -320,14 +374,41 @@ public struct Seed
...
+
+    public void toString (scope void delegate(const(char)[]) @safe sink,
+                          PrintMode mode = PrintMode.Obfuscated) const
     {
-        ubyte[1 + Seed.Width + 2] bin;
-        bin[0] = VersionByte.Seed;
-        bin[1 .. $ - 2] = this.data[];
-        bin[$ - 2 .. $] = checksum(bin[0 .. $ - 2]);
-        Base32.encode(bin, sink);
+        switch (mode)
```
AndrejMitrovic: Ditto final switch.

source/agora/common/crypto/Key.d
```diff
@@ -258,6 +260,49 @@ public struct SecretKey
...
+    }
+
+    /***************************************************************************
+
+        Print the secret key
+
+        By default, this function prints the hashed version of the secret key.
+        Changing the mode to `Clear` makes it print the non-hashed value.
+        This mode is intended for debugging. Additionally, no parameterless
+        `toString` is provided for this reason.
```
AndrejMitrovic: This is a very confusing sentence. How about instead:

```suggestion
        This mode is intended for debugging. The parameterless `toString` overload is not implemented, for safety reasons.
```

kchulj:
Any idea why it's failing completely on Mac and Windows?

AndrejMitrovic:
Linker issues, related to dub. There's a way to fix it.

source/agora/common/crypto/ECC.d
```diff
@@ -20,6 +20,7 @@ module agora.common.crypto.ECC;
 import agora.common.Hash;
 import agora.common.Serializer;
 import agora.common.Types;
+import agora.utils.PrettyPrinter;
```
Geod24: This is your problem. And a big no.

source/agora/common/crypto/Key.d
```diff
@@ -15,8 +15,10 @@ module agora.common.crypto.Key;
 
 import agora.common.crypto.Crc16;
 import agora.common.crypto.ECC;
+import agora.common.Hash;
 import agora.common.Types;
 import agora.common.Serializer;
+import agora.utils.PrettyPrinter;
```
Geod24: So is this

source/agora/common/crypto/ECC.d
AndrejMitrovic: What did @Geod24 tell you yesterday? Always blame drey.

source/agora/common/crypto/Key.d
```diff
@@ -322,7 +321,7 @@ public struct Seed
     }
 
     /// Make sure the sink overload of BitBlob is not picked
-    public void toString (scope void delegate(const(char)[]) sink) const
+    public void toString (scope void delegate(const(char)[]) @safe sink) const
```
Geod24: If you force `@safe` on the delegate, you should force `@safe` on `toString`.

kchulj: Is it ok if `toString` for `Seed` is not `@safe`?
It's using `Base32.encode`, which is not `@safe`.

Geod24: Then why did you change the delegate to be `@safe` ?

kchulj: I was assuming it's possible for only one of them to be `@safe`.. 😅

AndrejMitrovic:
It would be good to also check (for another PR) whether `r` is exposed anywhere. If anyone discovers the `r`, they can derive the private key - and then we're back to having the same problem.

AndrejMitrovic:
`r` being the nonce for the schnorr signatures.

source/agora/common/crypto/ECC.d
```diff
@@ -103,6 +135,20 @@ public struct Scalar
...
+    {
+        Scalar s = Scalar.random();
+        char[] buff1;
+        char[] buff2;
+        scope dg1 = (const(char)[] data) { buff1 ~= data; };
+        scope dg2 = (const(char)[] data) { buff2 ~= data; };
+
+        s.toString(dg1, PrintMode.Obfuscated);
+        s.toString(dg2, PrintMode.Clear);
+        assert(buff1 != buff2);
```
AndrejMitrovic: This is not really a good test, because it's not actually testing the output, only that it's different. It should test explicitly whether a private key is hidden. So what I suggest is:

- generate a random scalar
- print it out manually
- then hardcode it back into the first line in this test
- assert for both print modes exactly with the expected string output

kchulj: I updated the `unittest`s. What do you think?

source/agora/common/crypto/ECC.d
```diff
@@ -90,11 +90,43 @@ public struct Scalar
...
+
+    ***************************************************************************/
+
+    public void toString (scope void delegate(const(char)[]) @safe sink,
+                          PrintMode mode = PrintMode.Obfuscated) const @safe
+    {
+        final switch (mode)
+        {
+        case PrintMode.Obfuscated:
+            sink("********");
```
```suggestion
            sink("**SCALAR**");
```

So if we have multiple obfuscated value, we can tell the types apart.

source/agora/common/crypto/ECC.d
```diff
@@ -103,6 +135,21 @@ public struct Scalar
         return this.data.toString();
```
Geod24: This is problematic now

source/agora/common/crypto/ECC.d
kchulj: Good idea.

source/agora/common/crypto/Key.d
```diff
@@ -320,14 +423,19 @@ public struct Seed
         return Base32.encode(bin).assumeUnique;
```
Geod24: This needs fixing

source/agora/common/crypto/ECC.d
Geod24: Move this enum to `agora.common.Types` to avoid the duplication

source/agora/common/crypto/ECC.d
kchulj: Yes, and these `string toString ()` methods are used in other `struct`s and `unittest`s.
Would it be a good idea to remove the duplicate and use `scope void delegate` everywhere?

source/agora/common/crypto/Key.d
kchulj: Should I remove it, since it's a duplicate of `PrintMode.Clear`?

Geod24 (approved):

PR #1208, opened 2020-09-16
Fix revelation of wrong pre-images after re-enrollment

linked0:
There was a critical bug when `EnrollmentManager` reveals pre-images in the case when a validator re-enrolls before the validator cycle ends. In the previous code, we `reset` pre-images while creating new `Enrollment`, which means that the pre-images of the current cycle have been replaced with a new set of pre-images. So I make the `populate` function only generate the initial random seed for enrollment. After being a validator for the new cycle, the `populate` generates the whole pre-images for the new cycle.

Relates #1079

### DIFF
--- a/source/agora/consensus/PreImage.d
+++ b/source/agora/consensus/PreImage.d
@@ -362,24 +362,32 @@ public struct PreImageCycle
             this.seeds.reset(cycle_seed);
         }
 
-        // Populate the current enrollment round cache
-        // The index into `seeds` is the absolute index of the cycle,
-        // not the number of round, hence why we use `byStrides`
-        // The alternative would be:
-        // [$ - (this.index + 1) * this.preimages.length]
-        this.preimages.reset(this.seeds.byStride[$ - 1 - this.index]);
-
-        // Increment index if there are rounds left in this cycle
         if (consume)
         {
+            // Populate the current enrollment round cache
+            // The index into `seeds` is the absolute index of the cycle,
+            // not the number of round, hence why we use `byStrides`
+            // The alternative would be:
+            // [$ - (this.index + 1) * this.preimages.length]
+            this.preimages.reset(this.seeds.byStride[$ - 1 - this.index]);
+
+            // Increment index if there are rounds left in this cycle
             this.index += 1;
             if (this.index >= NumberOfCycles)
             {
                 this.index = 0;
                 this.nonce += 1;
             }
+            return this.preimages[$ - 1];
         }
-
-        return this.preimages[$ - 1];
+        else
+        {
+            // If this is used for getting initial pre-image for enrollment
+            // it just creates initial pre-image of the cycle seed.
+            auto next_images = PreImageCache(this.preimages.data.length,
+            this.preimages.interval);
+            next_images.reset(this.seeds.byStride[$ - 1 - this.index]);
+            return next_images[$ - 1];
+        }
     }
 }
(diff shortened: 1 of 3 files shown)


source/agora/consensus/PreImage.d
```diff
@@ -365,24 +365,35 @@ public struct PreImageCycle
...
+            this.preimages.reset(this.seeds.byStride[$ - 1 - this.index]);
+
+            // Increment index if there are rounds left in this cycle
             this.index += 1;
             if (this.index >= NumberOfCycles)
             {
                 this.index = 0;
                 this.nonce += 1;
             }
+            preimage = this.preimages[$ - 1];
```
AndrejMitrovic:
```suggestion
            return this.preimages[$ - 1];
```

source/agora/consensus/PreImage.d
```diff
@@ -365,24 +365,35 @@ public struct PreImageCycle
...
+        }
+        else
+        {
+            // If this is used for getting initial pre-image for enrollment
+            // it just creates initial pre-image of the cycle seed.
+            auto next_images = PreImageCache(this.preimages.data.length,
+            this.preimages.interval);
+            next_images.reset(this.seeds.byStride[$ - 1 - this.index]);
+            preimage = next_images[$ - 1];
```
AndrejMitrovic:
```suggestion
            return next_images[$ - 1];
```

source/agora/consensus/PreImage.d
```diff
@@ -365,24 +365,35 @@ public struct PreImageCycle
...
         }
 
-        return this.preimages[$ - 1];
+        return preimage;
```
AndrejMitrovic:
```suggestion
```

source/agora/consensus/PreImage.d
```diff
@@ -365,24 +365,35 @@ public struct PreImageCycle
...
 
-        // Populate the current enrollment round cache
-        // The index into `seeds` is the absolute index of the cycle,
-        // not the number of round, hence why we use `byStrides`
-        // The alternative would be:
-        // [$ - (this.index + 1) * this.preimages.length]
-        this.preimages.reset(this.seeds.byStride[$ - 1 - this.index]);
-
-        // Increment index if there are rounds left in this cycle
+        Hash preimage;
```
AndrejMitrovic: Avoid unnecessary state variables (see also other comments)

```suggestion
```

AndrejMitrovic (dismissed):

AndrejMitrovic:
I'm not entirely sure I understand what the reported issue was about. Maybe @Geod24 can help reviewing this.

Geod24:
@linked0 explained the problem to me yesterday. The issue is that, now that we accept `Enrollment` when an `Enrollment` is still active, we end up resetting the pre-image too early. See this code:
https://github.com/bpfkorea/agora/blob/952cd01b14778d5ca5e34c049c8217c59037614f/source/agora/consensus/EnrollmentManager.d#L268-L293

As soon as `addValidator` is performed, we consume the pre-image. If we try to access it later on, we'll get its pre-image.
However, this doesn't seem like the right solution. The way the cache works is fine, we should just not consume that early.
Also, this is lacking a test.

linked0:
> However, this doesn't seem like the right solution. The way the cache works is fine, we should just not consume that early.
Also, this is lacking a test.

@Geod24 I will research the right solution, but If you have one in your mind already, could you share it with me? And I will add a unit test for that.

linked0:
@Geod24 and I discussed what it should be. 
```
auto next_images = PreImageCache(this.preimages.data.length,
this.preimages.interval);
next_images.reset(this.seeds.byStride[$ - 1 - this.index]);
return next_images[$ - 1];
```

Unfortunately, we should calculate the first pre-image from the next cycle's seed because the pre-image is a pre-image of the seed of the current cycle.
I added a unit test for not consuming pre-images before being a validator.

Geod24 (approved):

PR #1194, opened 2020-09-11
Implement network clock time synchronization

AndrejMitrovic:
This work is based on https://github.com/bpfkorea/agora/pull/1100.

Fixes #1174

PR #1215, opened 2020-09-18
Small fixes in the Nominator

AndrejMitrovic:
I've split this off from https://github.com/bpfkorea/agora/pull/1194

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -489,26 +489,11 @@ extern(D):
     public override ValidationLevel validateValue (uint64_t slot_idx,
         ref const(Value) value, bool nomination) nothrow
     {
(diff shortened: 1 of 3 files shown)


Geod24 (approved):
LGTM. Last commit is a bit over-reaching (we eventually trust the Vibe.d code that send network requests), but let's go with it anyway.

TrustHenry (approved):

AndrejMitrovic:
Yeah we need to get rid of the @trusted stuff, but there is so much unsafe code everywhere else that there is no way of using it without wrapping it in @trusted..

AndrejMitrovic:
Ah it's LocalRest.wait that is unsafe.

AndrejMitrovic:
Well, I think we should wait for Omer's work rather than changing anything in the FiberScheduler to make it safe.

PR #1216, opened 2020-09-18
Change dub.json to workaround dub bug 2012

Geod24:
Also try to remove the workaround, let's see how this goes.
https://github.com/dlang/dub/issues/2012


Geod24:
The issue has been fixed upstream

Issue #1213, opened 2020-09-17
Add commons budget address to `ConsensusParams` & well-known keys


Geod24:
It's just a matter of adding another variable. Probably need to mine it to match a pattern in the future, but right now we're fine with anything. This will be used later to store data in the chain.

PR #1232, opened 2020-09-23
Add commons budget address to `ConsensusParams` & well-known keys

MichaelKim20:
I hide the commons budget seed key in the test.
Why don't we add a seed key when we need a seed key later?

I set the key pair like this in the test.
Seed:    SCNRULE3Q7NBNX7UIJVVI6HI5DQKP3MNDEHJA66FJZXDEGGH4SRRDSDR
Address: GCOMMONBGUXXP4RFCYGEF74JDJVPUW2GUENGTKKJECDNO6AGO32CUWGU

Related to #1213

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -102,6 +102,10 @@ public struct NodeConfig
     static assert(!hasUnsharedAliasing!(typeof(this)),
         "Type must be shareable accross threads");
 
+    /// If set, a commons budget address to use
+    /// in place of the built-in commons budget address as defined by CoinNet
+    public PublicKey commons_budget_address;
+
     /// If set, a hexdump serialized representation of the genesis block to use
     /// in place of the built-in genesis block as defined by CoinNet
     public string genesis_block;
@@ -330,6 +334,12 @@ private NodeConfig parseNodeConfig (Node* node, const ref CommandLine cmdln)
     auto min_listeners = get!(size_t, "node", "min_listeners")(cmdln, node);
     auto max_listeners = get!(size_t, "node", "max_listeners")(cmdln, node);
     auto address = get!(string, "node", "address")(cmdln, node);
+    auto commons_budget = opt!(string, "node", "commons_budget_address")(cmdln, node);
+
+    auto commons_budget_address = (commons_budget.length > 0)
+        ? PublicKey.fromString(commons_budget)
+        : PublicKey.init;
+
     auto genesis_block = opt!(string, "node", "genesis_block")(cmdln, node);
     auto genesis_start_time = get!(time_t, "node", "genesis_start_time",
         str => SysTime(DateTime.fromISOString(str)).toUnixTime)(cmdln, node);
@@ -356,6 +366,7 @@ private NodeConfig parseNodeConfig (Node* node, const ref CommandLine cmdln)
     NodeConfig result = {
             min_listeners : min_listeners,
             max_listeners : max_listeners,
+            commons_budget_address : commons_budget_address,
             genesis_block : genesis_block,
             genesis_start_time : genesis_start_time,
             block_interval_sec : block_interval_sec,
@@ -389,6 +400,7 @@ node:
   data_dir: .cache
   quorum_shuffle_interval: 10
   preimage_reveal_interval: 5
+  commons_budget_address: GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
 `;
         auto node = Loader.fromString(conf_example).load();
         auto config = parseNodeConfig("node" in node, cmdln);
@@ -397,6 +409,7 @@ node:
         assert(config.data_dir == ".cache");
         assert(config.quorum_shuffle_interval == 10);
         assert(config.preimage_reveal_interval == 5.seconds);
+        assert(config.commons_budget_address.toString() == "GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ");
     }
 }
 
--- a/source/agora/consensus/data/ConsensusParams.d
+++ b/source/agora/consensus/data/ConsensusParams.d
@@ -19,6 +19,7 @@
 
 module agora.consensus.data.ConsensusParams;
 
+import agora.common.crypto.Key;
 import agora.consensus.data.Block;
 
 import core.stdc.time;
@@ -50,6 +51,9 @@ public immutable class ConsensusParams
     /// How often blocks should be created
     public uint BlockIntervalSeconds;
 
+    /// The address of commons budget
+    public PublicKey CommonsBudgetAddress;
+
     /***************************************************************************
 
         Constructor
@@ -63,13 +67,16 @@ public immutable class ConsensusParams
 
     ***************************************************************************/
 
-    public this (immutable(Block) genesis, uint validator_cycle = 1008,
+    public this (immutable(Block) genesis,
+                 in PublicKey commons_budget_address,
+                 uint validator_cycle = 1008,
                  uint max_quorum_nodes = 7, uint quorum_threshold = 80,
                  uint quorum_shuffle_interval = 30,
                  time_t genesis_start_time = 1596179709,
                  uint block_interval_sec = 1)
     {
         this.Genesis = genesis;
+        this.CommonsBudgetAddress = commons_budget_address,
         this.ValidatorCycle = validator_cycle;
         this.MaxQuorumNodes = max_quorum_nodes;
         this.QuorumThreshold = quorum_threshold;
@@ -84,6 +91,7 @@ public immutable class ConsensusParams
         uint quorum_threshold = 80)
     {
         import agora.consensus.data.genesis.Test : GenesisBlock;
-        this(GenesisBlock, validator_cycle, max_quorum_nodes, quorum_threshold);
+        import agora.utils.WellKnownKeys;
+        this(GenesisBlock, CommonsBudget.address, validator_cycle, max_quorum_nodes, quorum_threshold);
     }
 }
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -544,7 +544,7 @@ version (unittest)
                 ? params_
                 : (blocks.length > 0
                    // Use the provided Genesis block
-                   ? new immutable(ConsensusParams)(cast(immutable)blocks[0])
+                   ? new immutable(ConsensusParams)(cast(immutable)blocks[0], WK.Keys.CommonsBudget.address)
                    // Use the unittest genesis block
                    : new immutable(ConsensusParams)());
 
@@ -1232,7 +1232,7 @@ unittest
         assert(ex.msg == "Genesis block loaded from disk is different from the one in the config file");
     }
 
-    immutable good_params = new immutable(ConsensusParams)(new_gen_block);
+    immutable good_params = new immutable(ConsensusParams)(new_gen_block, WK.Keys.CommonsBudget.address);
     // will not fail
     scope ledger = new TestLedger(WK.Keys.A, [new_gen_block], good_params);
     // Neither will the default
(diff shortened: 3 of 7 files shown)


source/agora/utils/WellKnownKeys.d
```diff
@@ -1434,6 +1434,28 @@ static immutable Genesis = KeyPair(
...
+    range returned by `byRange`, nor can it be indexed by `size_t`,
+    to avoid it being mistakenly used.
+
+    Address: GCOMMONBGUXXP4RFCYGEF74JDJVPUW2GUENGTKKJECDNO6AGO32CUWGU
+
+*******************************************************************************/
+
+static immutable CommonsBudgetKeyPair = KeyPair(
+    PublicKey([156, 198, 57, 161, 53, 47, 119, 242, 37, 22, 12, 66, 255, 137, 26, 106, 250, 91, 70, 161, 26, 105, 169, 73, 32, 134, 215, 120, 6, 118, 244, 42]),
+    SecretKey(),
```
MichaelKim20: Can we add when I need a seed key and a secret key later?

Geod24 (changes requested):
The point of WellKnownKeys is that the seed is known, so it should have `SecretKey` and `Seed` filled out. Also it's a  test-only module so shouldn't be imported by `ConsensusParams` outside of a `version (unittest)`.

AndrejMitrovic:
It would be nice to mine this key so it can be easily seen in the logs.

Geod24:
>     Address: G*COMMONB*GUXXP4RFCYGEF74JDJVPUW2GUENGTKKJECDNO6AGO32CUWGU

AndrejMitrovic:
Oh thanks I didn't see it.

source/agora/utils/WellKnownKeys.d
```diff
@@ -1434,6 +1434,29 @@ static immutable Genesis = KeyPair(
...
+
+    In unittests, we need the commons budget key pair to be known for us to be
+    able to write tests.
+    In the real network, there are different values.
+
+    Note that while this is a well-known keys, it is not part of the
+    range returned by `byRange`, nor can it be indexed by `size_t`,
+    to avoid it being mistakenly used.
+
+    Seed:    SCNRULE3Q7NBNX7UIJVVI6HI5DQKP3MNDEHJA66FJZXDEGGH4SRRDSDR`
```
AndrejMitrovic:
```suggestion
    Seed:    SCNRULE3Q7NBNX7UIJVVI6HI5DQKP3MNDEHJA66FJZXDEGGH4SRRDSDR
```

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -67,7 +72,8 @@ public immutable class ConsensusParams
                  uint max_quorum_nodes = 7, uint quorum_threshold = 80,
                  uint quorum_shuffle_interval = 30,
                  time_t genesis_start_time = 1596179709,
-                 uint block_interval_sec = 1)
+                 uint block_interval_sec = 1,
+                 const ref PublicKey commons_budget_address = CommonsBudgetKeyPair.address)
```
Geod24: Second parameter, `in`, no  default value

Geod24: The default value should be passed  at L94

kchulj:
There seems to be a `segfault`. I'm trying to find out what's causing it.

source/agora/common/Config.d
```diff
@@ -330,6 +334,13 @@ private NodeConfig parseNodeConfig (Node* node, const ref CommandLine cmdln)
     auto min_listeners = get!(size_t, "node", "min_listeners")(cmdln, node);
     auto max_listeners = get!(size_t, "node", "max_listeners")(cmdln, node);
     auto address = get!(string, "node", "address")(cmdln, node);
+    auto commons_budget = opt!(string, "node", "commons_budget_address")(cmdln, node);
+
+    import agora.utils.WellKnownKeys;
+    auto commons_budget_address = (commons_budget.length > 0)
+        ? PublicKey.fromString(commons_budget)
+        : CommonsBudgetKeyPair.address;
```
Geod24: https://github.com/bpfkorea/agora/pull/1232#pullrequestreview-494067435

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -80,10 +89,11 @@ public immutable class ConsensusParams
 
     /// Default for unittest, uses the test genesis block
     version (unittest) public this (
+        const ref PublicKey commons_budget_address = CommonsBudgetKeyPair.address,
```
Geod24: This was not my recommendation

source/agora/utils/Test.d
```diff
@@ -214,6 +217,12 @@ public struct WK
         {
             return wellKnownKeyByIndex(idx);
         }
+
+        /// Returns the commons budget key pairs
+        public static KeyPair CommonsBudget ()
+        {
+            return CommonsBudgetKeyPair;
+        }
```
Geod24: Why ?

source/agora/utils/WellKnownKeys.d
```diff
@@ -1434,6 +1434,29 @@ static immutable Genesis = KeyPair(
...
+    Note that while this is a well-known keys, it is not part of the
+    range returned by `byRange`, nor can it be indexed by `size_t`,
+    to avoid it being mistakenly used.
+
+    Seed:    SCNRULE3Q7NBNX7UIJVVI6HI5DQKP3MNDEHJA66FJZXDEGGH4SRRDSDR
+    Address: GCOMMONBGUXXP4RFCYGEF74JDJVPUW2GUENGTKKJECDNO6AGO32CUWGU
+
+*******************************************************************************/
+
+static immutable CommonsBudgetKeyPair = KeyPair(
```
Geod24: Why this name ? **ALL** other KeyPair are named in a human-readable way. `NODE{2..7}`, `A`..`Z`, `AA`..`ZZ`... None of them has a `KeyPair` suffix, so why does this one have one ?

Geod24 (changes requested):

source/agora/consensus/data/ConsensusParams.d
MichaelKim20: I fixed the code.

MichaelKim20:
> There seems to be a `segfault`. I'm trying to find out what's causing it.

Thank you, Check it out again.

source/agora/utils/Test.d
AndrejMitrovic: Yeah this is unnecessary. `Keys` is used for indexing, like `WK.Keys[idx]`.

source/agora/common/Config.d
AndrejMitrovic: This type of logic belongs somewhere here: https://github.com/bpfkorea/agora/blob/59d9463500e81f279e1ef57861d64246bf6e4e3b/source/agora/node/FullNode.d#L131

Then the budget address would be:

```D
auto commons_budget = 
    config.node.commons_budget_address != PublicKey.init ? 
    config.node.commons_budget_address : CNG.CommonsBudget;
```

And you would need to define a (fake) PublicKey inside of `agora.consensus.data.genesis.Coinnet`.

@Geod24 am I in the right here?

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -80,10 +89,11 @@ public immutable class ConsensusParams
 
     /// Default for unittest, uses the test genesis block
     version (unittest) public this (
+        in PublicKey commons_budget_address = CommonsBudgetKeyPair.address,
```
AndrejMitrovic: Move this parameter last and you won't have to modify any other tests.

source/agora/consensus/data/ConsensusParams.d
```diff
@@ -63,13 +68,17 @@ public immutable class ConsensusParams
...
+    public this (immutable(Block) genesis,
+                 in PublicKey commons_budget_address,
+                 uint validator_cycle = 1008,
                  uint max_quorum_nodes = 7, uint quorum_threshold = 80,
                  uint quorum_shuffle_interval = 30,
                  time_t genesis_start_time = 1596179709,
                  uint block_interval_sec = 1)
     {
         this.Genesis = genesis;
+
```
AndrejMitrovic:
```suggestion
```

source/agora/utils/WellKnownKeys.d
AndrejMitrovic: Missing `public`.

Also `static` doesn't do anything in the scope of a module. Remove it please.

source/agora/consensus/data/ConsensusParams.d
Geod24: No, don't make it a parameter. The Genesis block is not a parameter either.

AndrejMitrovic: Right good point, this is for tests only.

source/agora/utils/WellKnownKeys.d
Geod24: The whole file is `static immutable` though (the code was generated)

AndrejMitrovic: Ahh right, codegen.

MichaelKim20: > Why this name ? **ALL** other KeyPair are named in a human-readable way. `NODE{2..7}`, `A`..`Z`, `AA`..`ZZ`... None of them has a `KeyPair` suffix, so why does this one have one ?

I see. I'll change it to CommonsBudget.

source/agora/consensus/data/ConsensusParams.d
MichaelKim20: Yeah, I'll remove it from the parameters, and I'll do the same thing as the Genesis block.

source/agora/utils/Test.d
MichaelKim20: I added it because there was another problem.
I'll remove it here because I've already solved it.

source/agora/common/Config.d
MichaelKim20: I'll change it according to @AndrejMitrovic 's suggestion.

MichaelKim20:
Ready for review

AndrejMitrovic (approved):
LGTM

TrustHenry (approved):
LGTM

source/agora/common/Config.d
```diff
@@ -102,6 +102,10 @@ public struct NodeConfig
     static assert(!hasUnsharedAliasing!(typeof(this)),
         "Type must be shareable accross threads");
 
+    /// If set, a commons budget address to use
+    /// in place of the built-in commons budget address as defined by CoinNet
+    public PublicKey commons_budget_address;
```
Geod24, line 107: Does it actually  work with string, or does it expect an array of ubyte ?

source/agora/node/Ledger.d
```diff
@@ -536,6 +536,8 @@ version (unittest)
     /// A `Ledger` with sensible defaults for `unittest` blocks
     private final class TestLedger : Ledger
     {
+        import WKC = agora.utils.WellKnownKeys;
```
Geod24: The key is already imported via `agora.utils.Test`.

source/agora/common/Config.d
MichaelKim20: At first, it was declared a string.
But I judged that it would be safer to do verification by converting from this into a structure than to deliver it internally without its verification. And it has data in the ubyte array inside.

source/agora/node/Ledger.d
MichaelKim20: Thank you. I will fix

source/agora/common/Config.d
Geod24: Nevermind, I misread the diff L337. This works fine 👌

Geod24 (approved):
Just fix the imports in Ledger.d and  this is good to go.

MichaelKim20:
> Just fix the imports in Ledger.d and this is good to go.

@Geod24: I fixed the imports in Ledger.d
I could use it as follows.
WK.Keys.CommonsBudget.address

PR #1233, opened 2020-09-23
Change JSON serialization of `Height`

### LINKED ISSUE #1211: Change JSON serialization of `Height`
Currently, `height` is output as `{ value: XXX }` instead of just `XXX`, because we're using a struct to ensure strict typing.
This should be fixed, but it will affect Stoa.

TrustHenry:
The Height using JsonString is output as {value: XXX } because the structure.
When applied Height is output as XXX instead of {value: XXX }.

Fixes #1211

### DIFF
--- a/source/agora/common/Types.d
+++ b/source/agora/common/Types.d
@@ -76,4 +76,32 @@ public struct Height
 
     /// Provides implicit conversion to `ulong`
     public alias value this;
+
+    /// Support for Vibe.d serialization to JSON
+    public string toString () const @safe
+    {
+        import std.conv : to;
+        return this.value.to!string;
+    }
+
+    /// Support for Vibe.d deserialization
+    public static Height fromString (scope const(char)[] str) pure @safe
+    {
+        import std.conv : to;
+        immutable ul = str.to!ulong;
+        return Height(ul);
+    }
+}
+
+///
+unittest
+{
+    import vibe.data.json;
+
+    const h = Height(1000);
+    const(char)[] str_h = "1000";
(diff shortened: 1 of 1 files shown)


MichaelKim20 (approved):
LGTM

source/agora/common/Types.d
```diff
@@ -76,4 +76,38 @@ public struct Height
 
     /// Provides implicit conversion to `ulong`
     public alias value this;
+
+    /// Helper type for `toString`
+    private alias SinkT = void delegate(scope const(char)[] v) @safe;
```
Geod24: I don't think we need this, it's only used from within a function

source/agora/common/Types.d
```diff
@@ -76,4 +76,38 @@ public struct Height
...
+
+    /// Support for Vibe.d serialization to JSON
+    public string toString () const @safe
```
Geod24, line 81: Can you also provide a sink overload ?

source/agora/common/Types.d
TrustHenry: I Changed the string conversion method.
Using std.conv

source/agora/common/Types.d
TrustHenry: Well, I think we need to make another issue.
A problem with `Height` sink overload has been found.
But it seems to be separate from the related issues.

Geod24: This function leads to new memory allocations when logging, hence why I mentioned it.

PR #1234, opened 2020-09-24
Implement automated transaction generator

### LINKED ISSUE #1230: Implement the ability to continuously generate transactions and send them to Agora nodes
This is one feature to simulate activity on an Agora network.
The tool should have the ability to continuously generate new valid transactions and send these to one (or more) Agora nodes. This process should be automated.

Part of #1227

kchulj:
A tool which periodically generates a set of valid transactions and sends them to one node. The node should gossip the transactions to the other nodes. The interval at which the transactions are sent and the number of transactions can be specified.

Usage:
```
agora-client gentx [--dump] [--interval <interval>] [--count <count>] --ip <host> --port <port>
```
```
agora-client gentx -i eu-002.bosagora.io -p 2826
```
Custom interval and number of transactions sent at once:
```
agora-client gentx -i eu-002.bosagora.io -p 2826 -t 15 -c 4
```

Definition of done:
- [x] Send multiple `Transaction`s
- [x] Gossip and create block 

Fixes #1230

### DIFF
--- a/source/agora/cli/client/GenTxProcess.d
+++ b/source/agora/cli/client/GenTxProcess.d
@@ -1,0 +1,184 @@
+/*******************************************************************************
+
+    A tool which periodically generates a set of valid transactions
+    and sends them to one node. The interval at which the transactions
+    are sent and the number of transactions can be specified.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.client.GenTxProcess;
+
+import agora.api.FullNode;
+import agora.client.Result;
+import agora.common.Amount;
+import agora.common.crypto.Key;
+import agora.common.Hash;
+import agora.common.Task;
+import agora.common.Types;
+import agora.consensus.data.Block;
+import agora.consensus.data.Transaction;
+import agora.consensus.data.genesis.Test;
+import agora.utils.Log;
+import agora.utils.PrettyPrinter;
+import agora.utils.Test;
+
+import std.algorithm;
+import std.conv;
+import std.format;
+import std.getopt;
+import std.range;
+import std.stdio;
+
+import core.time;
+
+import vibe.core.core;
+
+mixin AddLogger!();
+
+/// Option required to generate and send transactions
+private struct GenTxOption
+{
+    /// IP address of node
+    public string host;
+
+    /// Port of node
+    public ushort port;
+
+    /// Interval of sending transactions
+    public Duration interval = 5.seconds;
+
+    /// Number of transactions sent at once (1 to 8)
+    public uint count = 8;
+
+    /// dump output option
+    public bool dump;
+}
+
+/// Parse the command-line arguments of gentx
+public GetoptResult parseGenTxOption (ref GenTxOption op, string[] args)
+{
+    return getopt(
+        args,
+        "ip|i",
+            "IP address of node",
+            &op.host,
+
+        "port|p",
+            "Port of node",
+            &op.port,
+
+        "interval|t",
+            "Interval of sending transactions (default: 5 seconds)",
+            (string option, string value) { op.interval = value.to!int.seconds; },
+
+        "count|c",
+            "Number of transactions sent at once (default: 8)",
+            &op.count,
+
+        "dump|o",
+            "Dump output option",
+            &op.dump
+            );
+}
+
+/// Print help
+public void printGenTxHelp (ref string[] outputs)
+{
+    outputs ~= "usage: agora-client gentx [--dump] [--interval <interval>]";
+    outputs ~= "                          [--count <count>] --ip <host>";
+    outputs ~= "                          --port <port>";
+    outputs ~= "";
+    outputs ~= "   gentx      Generate and send a transaction to node";
+    outputs ~= "";
+    outputs ~= "        -o --dump       Dump output option";
+    outputs ~= "        -i --ip         IP address of node";
+    outputs ~= "        -p --port       Port of node";
+    outputs ~= "        -t --interval   Interval of sending transactions";
+    outputs ~= "        -c --count      Number of transactions sent at once";
+    outputs ~= "";
+}
+
+/*******************************************************************************
+
+    Input arguments, generate the transaction and send it to the node
+
+    Params:
+        args = client command line arguments
+
+*******************************************************************************/
+
+public int genTxProcess (string[] args, ref string[] outputs,
+                         API delegate (string address) api_maker)
+{
+    GenTxOption op;
+    GetoptResult res;
+
+    try
+    {
+        res = parseGenTxOption(op, args);
+        if (res.helpWanted)
+        {
+            printGenTxHelp(outputs);
+            return CLIENT_SUCCESS;
+        }
+    }
+    catch (Exception ex)
+    {
+        log.info("Exception while generating transactions");
+        printGenTxHelp(outputs);
+        return CLIENT_EXCEPTION;
+    }
+
+    if (op.count > 8 || op.count <= 0)
+    {
+        printGenTxHelp(outputs);
+        outputs ~= format("Cannot send more than 8 transactions. %d requested."
+                          , op.count);
+        return CLIENT_INVALID_ARGUMENTS;
+    }
+
+    // connect to the node
+    string ip_address = format("http://%s:%s", op.host, op.port);
+    auto node = api_maker(ip_address);
+    auto taskman = new TaskManager;
+    auto last_block = node.getBlocksFrom(node.getBlockHeight(), 1)[0];
+    auto txs = last_block.spendable().take(op.count)
+                         .map!(txb => txb.split(
+                             WK.Keys.A.address.only()).sign())
+                         .array();
+
+    // function to generate and send transactions
+    void genTxs ()
+    {
+        txs.each!(tx => node.putTransaction(tx));
+        writefln("%s transactions sent to %s.\nTransactions:",
+            op.count, ip_address);
+        txs.each!(tx => writeln(prettify(tx)));
+        txs = txs.map!(txb => TxBuilder(txb).sign()).array();
+    }
+
+    if (op.dump)
+    {
+        outputs ~= format("IP = %s", op.host);
+        outputs ~= format("Port = %s", op.port);
+        outputs ~= format("Interval = %s", op.interval);
+        outputs ~= format("Count = %s", op.count);
+        outputs ~= format("Transactions =");
+        txs.each!(tx => outputs ~= format("%s", prettify(tx)));
+        outputs ~= format("Hash =");
+        txs.each!(tx => outputs ~= format("%s", hashFull(tx)));
+        return CLIENT_SUCCESS;
+    }
+
+    genTxs();
+    taskman.setTimer(op.interval, &genTxs, Periodic.Yes);
+
+    return runEventLoop();
+}
--- a/dub.json
+++ b/dub.json
@@ -101,7 +101,6 @@
                 "source/agora/network/NetworkClient.d",
                 "source/agora/network/NetworkManager.d",
                 "source/agora/node/*",
-                "source/agora/utils/PrettyPrinter.d",
                 "source/agora/utils/SCPPrettyPrinter.d",
                 "source/scpp/*",
                 "source/scpd/*",
--- a/source/agora/cli/client/DefaultProcess.d
+++ b/source/agora/cli/client/DefaultProcess.d
@@ -50,6 +50,8 @@ public void printDefaultHelp (ref string[] outputs)
     outputs ~= "";
     outputs ~= "These are commands:";
     outputs ~= "   sendtx      Send a transaction to node";
+    outputs ~= "   gentx       Generate transactions and";
+    outputs ~= "               send to node";
     outputs ~= "";
 }
 
--- a/source/agora/cli/client/main.d
+++ b/source/agora/cli/client/main.d
@@ -16,6 +16,7 @@ module agora.client.main;
 import agora.api.FullNode;
 import agora.client.Result;
 import agora.client.DefaultProcess;
+import agora.client.GenTxProcess;
 import agora.client.SendTxProcess;
 
 import vibe.core.core;
@@ -53,6 +54,10 @@ int runProcess (string[] args, ref string[] outputs)
     {
         case "sendtx":
             return sendTxProcess(args, outputs, (address) {
+                return new RestInterfaceClient!API(address);
+            });
+        case "gentx":
+            return genTxProcess(args, outputs, (address) {
                 return new RestInterfaceClient!API(address);
             });
         default :


source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
...
+    public string host = "0.0.0.0";
+
+    /// Port of node
+    public ushort port = 2826;
+
+    /// The address key to send the output
+    public string address;
+
+    /// Interval of sending transaction
+    public uint interval = 5;
```
AndrejMitrovic:
```suggestion
    public Duration interval = 5.seconds;
```

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
...
+            "Port of node (default: 2826)",
+            &op.port,
+
+        "dest|d",
+            "The address key to send the output",
+            &op.address,
+
+        "interval|t",
+            "Interval of sending transaction",
+            &op.interval,
```
AndrejMitrovic:
```suggestion
            (string value ) { op.interval = value.to!int.seconds; },
```

This assumes that the interval given at command-line always specifies seconds, and not milliseconds for example.

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
...
+    }
+
+    if (!isValid)
+        return CLIENT_INVALID_ARGUMENTS;
+
+    // connect to the node
+    string ip_address = format("http://%s:%s", op.host, op.port);
+    auto node = api_maker(ip_address);
+    auto taskman = new TaskManager;
+    auto txs = genesisSpendable().map!(txb => txb.sign()).array()[0];
```
AndrejMitrovic: Maybe you plan to support multiple txs in the future, but this variable only contains one. So it should be named appropriately.

```suggestion
    auto tx = genesisSpendable().map!(txb => txb.sign()).array()[0];
```

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
...
+
+    Input arguments, generate the transaction and send it to the node
+
+    Params:
+        args = client command line arguments
+
+*******************************************************************************/
+
+public int genTxProcess (string[] args, ref string[] outputs,
+                           APIMaker api_maker)
```
AndrejMitrovic:
```suggestion
                         APIMaker api_maker)
```

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
...
+
+    if (op.address.length == 0)
+    {
+        if (isValid) printGenTxHelp(outputs);
+        outputs ~= "Address is not entered. [--dest]";
+        isValid = false;
+    }
+
+    if (!isValid)
+        return CLIENT_INVALID_ARGUMENTS;
```
AndrejMitrovic: In general we should always prefer setting a validity flag to false by default. It's good programming style because if you forget to set it to false then the rest of the code assumes the flag is correct. It's better to assume it's invalid by default.

But in any case you don't really need a flag here.

```suggestion
    if (op.address.length == 0)
    {
        printGenTxHelp(outputs);
        outputs ~= "Address is not entered. [--dest]";
        return CLIENT_INVALID_ARGUMENTS;
    }
```

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
+/*******************************************************************************
+
+    A simple tool which generates valid transactions and sends
+    one transaction periodically to one given node.
```
AndrejMitrovic:
```suggestion
    A simple tool which generates valid transactions and sends
    each transaction periodically to one given node.
```

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
...
+    The interval at which the transaction is sent can be specified.
```
AndrejMitrovic:
```suggestion
    The interval at which the transactions are sent can be specified.
```

"the transaction" implies that it keeps sending the same one.

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
...
+import agora.utils.PrettyPrinter;
+import agora.utils.Test;
+
+import core.time;
+import std.algorithm;
+import std.format;
+import std.getopt;
+import std.range;
+import std.stdio;
+
```
AndrejMitrovic: I think this is our style: Phobos imports first, then Druntime.

```suggestion
import std.algorithm;
import std.format;
import std.getopt;
import std.range;
import std.stdio;

import core.time;
```

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,176 @@
...
+public void printGenTxHelp (ref string[] outputs)
+{
+    outputs ~= "usage: agora-client gentx [--dump] [--ip] [--port] [--dest] [--interval]";
+    outputs ~= "";
+    outputs ~= "   gentx      Generate and send a transaction to node";
+    outputs ~= "";
+    outputs ~= "        -i --ip         IP address of node";
+    outputs ~= "        -p --port       Port of node";
+    outputs ~= "        -d --dest       The address key to send the output";
+    outputs ~= "        -t --interval   Interval of sending transaction";
```
AndrejMitrovic:
```suggestion
    outputs ~= "        -t --interval   Interval of sending transactions";
```

AndrejMitrovic (changes requested):
So far so good!

AndrejMitrovic:
@kchulj have you tried using it with our test nodes?

kchulj:
> @kchulj have you tried using it with our test nodes?

Not yet. I will try that!

kchulj:
Update: the tool now supports multiple `Transaction`s.

kchulj:
@AndrejMitrovic The `Transaction`s can be sent to our test nodes. How can I check if the node is receiving them on the server-side?

AndrejMitrovic:
We'd have to inspect the logs. Which server did you send it to?

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,199 @@
...
+private struct GenTxOption
+{
+    /// IP address of node
+    public string host = "0.0.0.0";
+
+    /// Port of node
+    public ushort port = 2826;
+
+    /// The address key to send the output
+    public string address;
```
linked0: Please rename it because it's confusing with the description of 'IP address of node'. I know it means an address in the Blockchain world. :)

AndrejMitrovic:
I suggest changing this approach into something more error-proof. The tool could retrieve the latest block from a node, find the UTXOs, and then start creating transactions spending those UTXOs. That way the tool can be restarted and it would continue to work.

kchulj:
~~Update: the tool now supports sending transactions to multiple `node`s.~~
The tool sends to one node. The node should gossip the transactions to the other nodes.

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+    public string host;
+
+    /// Port of node
+    public ushort port;
+
+    /// Interval of sending transactions
+    public Duration interval = 5.seconds;
+
+    /// Number of transactions sent at once (1 to 8)
+    public uint count = 1;
```
Geod24: Why is it limited to 8 ?

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+
+    /// dump output option
+    public bool dump;
```
Geod24, line 61: Looking at the implementation, it's more of a dry-run mode, no ?

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+            &op.count,
+
+        "dump|o",
+            "Dump output option",
+            &op.dump
+            );
+}
+
+/// Print help
+public void printGenTxHelp (ref string[] outputs)
```
Geod24, line 92: `getopt` gives you this for free, check the documentation.
Although it's not a big deal (aka you don't have to change it now), because the current version is 👍 and works.

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+{
+    outputs ~= "usage: agora-client gentx [--dump] [--ip] [--port]";
```
Geod24: Side note: There is a certain syntax for parameter documentation in UNIX. `[--dump]` means the argument is optional and doesn't take a value (`[]` denotes optionality). So obviously `--ip` and `--port` are not optional.
Also, they take a value (as do `--interval` and `--count`.
To give you an example, this is `git rebase` SYNOPSYS:
```
SYNOPSIS
       git rebase [-i | --interactive] [<options>] [--exec <cmd>]
               [--onto <newbase> | --keep-base] [<upstream> [<branch>]]
       git rebase [-i | --interactive] [<options>] [--exec <cmd>] [--onto <newbase>]
               --root [<branch>]
       git rebase (--continue | --skip | --abort | --quit | --edit-todo | --show-current-patch)
```

You can clearly see that if `--exec` is used, it takes a parameter that is a command, for example.

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+        printGenTxHelp(outputs);
+        return CLIENT_EXCEPTION;
+    }
+
+    if (op.count > 8)
+    {
+        printGenTxHelp(outputs);
+        outputs ~= format("Cannot send %s transactions", op.count);
+        return CLIENT_INVALID_ARGUMENTS;
+    }
```
Geod24: Two things:
1) Why ?
2) When a client makes an error, it's not enough to tell them something is wrong. Try to give them a hint at why, and, in case it's not obvious (it is here), how to fix things. Here: "Cannot send more than 8 transactions (%d requested)".

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+        outputs ~= format("Hash =");
+        txs.each!(tx => outputs ~= format("%s", hashFull(tx)));
+        return CLIENT_SUCCESS;
+    }
+
+    genTxs();
+    taskman.setTimer(op.interval, &genTxs, Periodic.Yes);
+    runEventLoop();
+
+    return CLIENT_SUCCESS;
```
Geod24:
```suggestion
    return runEventLoop();
```

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+    if (op.count > 8)
+    {
+        printGenTxHelp(outputs);
+        outputs ~= format("Cannot send %s transactions", op.count);
+        return CLIENT_INVALID_ARGUMENTS;
+    }
+
+    // connect to the node
+    string ip_address = format("http://%s:%s", op.host, op.port);
+    auto node = api_maker(ip_address);
```
Geod24, line 149: Why do you need the `api_maker` ?

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+    auto taskman = new TaskManager;
+    auto last_block = node.getBlocksFrom(node.getBlockHeight(), 1)[0];
+    auto txs = last_block.spendable().take(op.count)
+                         .map!(txb => txb.split(
+                             WK.Keys.byRange.map!(k => k.address).take(1)).sign())
```
Geod24: This is always sending to `WK.Keys.A` right ?

Geod24 (commented):
It's a decent start. Did you try it on the live nodes ?

source/agora/cli/client/GenTxProcess.d
kchulj: It's temporarily 8 for simplicity, since I'm chaining from the 8 transactions in the genesis.

source/agora/cli/client/GenTxProcess.d
kchulj: Ah I didn't know this. Thanks for the explanation.

source/agora/cli/client/GenTxProcess.d
kchulj: I used it to create a node. Is there a different way to do it?

source/agora/cli/client/GenTxProcess.d
kchulj: Correct

kchulj:
> It's a decent start. Did you try it on the live nodes ?

It sends the transactions to the ~~live~~ local nodes. However, they are rejecting the transactions after the first one because the consensus is not reached. As discussed with @AndrejMitrovic , there needs to be either all nodes running simultaneously, or a custom genesis block which only has one validator. Any suggestions?

AndrejMitrovic:
> there needs to be either all nodes running simultaneously, or a custom genesis block which only has one validator. Any suggestions?

I was talking about testing locally. The nodes in our EU/Canada servers are already all connected together.

kchulj:
> > there needs to be either all nodes running simultaneously, or a custom genesis block which only has one validator. Any suggestions?
> 
> I was talking about testing locally. The nodes in our EU/Canada servers are already all connected together.

How can I modify the block creation interval in the servers?

AndrejMitrovic:
> How can I modify the block creation interval in the servers?

I don't think it's necessary because the `genesis_start_time` is already set in the past, therefore the nodes should be creating blocks as fast as possible already to catch up with the current time. There must be another reason why the nodes haven't started creating blocks.

AndrejMitrovic:
I can't seem to use this tool. This is what I get:

```
$ ./build/agora-client gentx -i "127.0.0.1" -p 2826 -t 3 -c 1

usage: agora-client gentx [--dump] [--ip] [--port]
                          [--interval] [--count]

   gentx      Generate and send a transaction to node

        -o --dump       Dump output option
        -i --ip         IP address of node
        -p --port       Port of node
        -t --interval   Interval of sending transactions
        -c --count      Number of transactions sent at once
```

But it's not telling me what's wrong. Which parameter is wrong / missing?

AndrejMitrovic:
It would be good to print the exception message and not just the help string if an exception was thrown.

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,181 @@
...
+            "IP address of node",
+            &op.host,
+
+        "port|p",
+            "Port of node",
+            &op.port,
+
+        "interval|t",
+            "Interval of sending transactions (default: 5 seconds)",
+            (string value) { op.interval = value.to!int.seconds; },
```
AndrejMitrovic: This is an unfortunate GetOpt design. The `value` here will actually be the string `"interval|t"`. Because it's not a value, it's an option. I think we need to use this:

```suggestion
            (string option, string value) { op.interval = value.to!int.seconds; },
```

source/agora/cli/client/GenTxProcess.d
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+    catch (Exception ex)
+    {
+        log.info("Exception while generating transactions");
+        printGenTxHelp(outputs);
+        return CLIENT_EXCEPTION;
+    }
+
+    if (op.count > 8)
```
AndrejMitrovic: Don't forget to check for zero.

source/agora/cli/client/GenTxProcess.d
AndrejMitrovic: So why not change it to this then?:

```suggestion
                             WK.Keys.A.address).sign())
```

AndrejMitrovic (changes requested):
Looks good, just two issues.

AndrejMitrovic:
```
commit c4101b1ab2879d6ac896cf0ac482b27baff744cc (HEAD, kuchul/tool-tx, pull1234)
Author: Ku Chul Jung <kuchul.jung@bpfkorea.org>
Date:   Wed Sep 23 13:31:45 2020 +0900

    Implement automated transaction generator

    A tool which periodially generates a set of valid transactions
    and sends them to one node. The interval at which the transactions
    are sent and the number of transactions can be specified.

commit 320a7981ca93eaaee4bf067cd16b78db603a1041
Author: Andrej Mitrovic <andrej.mitrovich@gmail.com>
Date:   Thu Sep 17 12:36:37 2020 +0900

    Tests: simplifly the node config creation

    The logic became too hard to follow.
```

I think you need to rebase your PR against `master`. The last commit in `master` is cdd5cc13be087be6da2a6973f090cb74d2838cc5

kchulj:
Update: After the tool sends the transactions, the nodes are able gossip and create a block.
However, the following error is shown when trying to create the 2nd block.

```
core.exception.AssertError@source/agora/consensus/EnrollmentManager.d(548): Assertion failure
```
This is triggered by:
```D
PreImageInfo preimage;
if (this.enroll_man.getNextPreimage(preimage))
{
    this.enroll_man.addPreimage(preimage);
```
As @AndrejMitrovic mentioned, it may be blocked by #1241

AndrejMitrovic:
~I have bisected it. It seems a27601afa74a762d5646d6a318aa9e8533979460 is the first commit where this bug starts to appear.~

Sorry I was wrong, the config option changes there so I have to revert to an earlier config layout. I'm continuing bisecting.

AndrejMitrovic:
I bisected it to b51025e4ef0a3bf707e51fe338eb82a17da75ae4, the first commit which shows the wrong preimage error message.

AndrejMitrovic:
> I bisected it to b51025e, the first commit which shows the wrong preimage error message.

But actually, maybe that commit is not the root cause. Before before this commit the node would not attempt to send the preimage, so that's why the error isn't shown. So I think the actual bug is even further behind this commit.

AndrejMitrovic:
@kchulj in the meantime, can you mark this as ready? I think we can go ahead with it.

AndrejMitrovic:
```
unittesting agora.test.ValidatorCleanRestart..
0   agora-unittests                     0x000000010ca07f78 _D4core7runtime9backtraceFNbPPviZi + 88
1   agora-unittests                     0x000000010ca07eec _D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPSQCk3sys5posix6signal9siginfo_tPvZv + 76
2   libsystem_platform.dylib            0x00007fff52da0f5a _sigtramp + 26
./ci/run.sh: line 13:  4125 Segmentation fault: 11  timeout -s SEGV 8m ./build/agora-unittests
```

It looks like we still have a bug somewhere in our main branch.

AndrejMitrovic:
Thank you for your work!

PR #1235, opened 2020-09-24
Give the BanManager a Clock instance

### LINKED ISSUE #1226: BanManager should use the new Clock class
The BanManager derives its time from this function:

https://github.com/bpfkorea/agora/blob/bbf523199b6efd1667ee0e3d79ad59d4c3ea93c2/source/agora/common/BanManager.d#L275

And then in unittests we use a derived `FakeBanManager` with its own custom clock:

https://github.com/bpfkorea/agora/blob/bbf523199b6efd1667ee0e3d79ad59d4c3ea93c2/source/agora/test/Base.d#L386-L389

Now that we have a [Clock instance](https://github.com/bpfkorea/agora/blob/bbf523199b6efd1667ee0e3d79ad59d4c3ea93c2/source/agora/network/Clock.d), we should unify this code:

- Give the BanManager an instance of the Clock
- Remove `FakeBanManager` entirely
- Change the BanManager [unittest](https://github.com/bpfkorea/agora/blob/v0.x.x/source/agora/common/BanManager.d) and use the `setTimeFor` routine from `Base.d` to test un-banning behavior.

linked0:
Now that we have a Clock instance, we need to unify this code:

* Give the BanManager an instance of the Clock
* Remove FakeBanManager and add a new class like TestBanManager with no operations on the `dump` and `load`.
* Change the BanManager unittest testing un-banning behavior.

Fixes #1226


source/agora/network/Clock.d
```diff
@@ -48,7 +48,6 @@
```
AndrejMitrovic, line 48: typo in commit message: cuased => caused

source/agora/common/BanManager.d
```diff
@@ -274,7 +279,7 @@ public class BanManager
 
     protected time_t getCurTime () @safe nothrow @nogc
     {
-        return time(null);
+        return clock.localTime();
```
AndrejMitrovic:
```suggestion
        return this.clock.localTime();
```

source/agora/test/Base.d
```diff
@@ -902,6 +902,7 @@ public class TestNetworkManager : NetworkManager
         Params:
             conf = ban manager config
             clock = clock instance
+            _data_dir = path to the data directory
```
AndrejMitrovic: Let's also rename the parameter to `data_dir` to be consistent with the `TestBanManager` constructor.

AndrejMitrovic (changes requested):
LGTM, just a few nits.

source/agora/common/BanManager.d
```diff
@@ -272,9 +277,9 @@ public class BanManager
 
     ***************************************************************************/
 
-    protected time_t getCurTime () @safe nothrow @nogc
+    protected time_t getCurTime () @safe nothrow
```
AndrejMitrovic: Can't we add `@nogc` to the Clock instead?

linked0: > Can't we add @nogc to the Clock instead?

That makes an error in CI builds as follows

source\agora\network\Clock.d(111,25): Error: `@nogc` function `agora.network.Clock.Clock.networkTime` cannot call non-@nogc function `std.conv.to!int.to!long.to`

AndrejMitrovic: But that's `networkTime()`, not `localTime()`.

linked0: Ah, Sorry. Another error also occurred.

source/agora/common/BanManager.d(245,56): Error: @nogc function agora.common.BanManager.BanManager.isBanned cannot call non-@nogc function agora.common.BanManager.BanManager.getCurTime

AndrejMitrovic: But I'm saying keep `getCurTime` and `isBanned` as `@nogc` and mark `Clock.localTime` as `@nogc`. Shouldn't it work then?

linked0: I see what you mean. I thought that the two functions, `localTime()` and `networkTime()` should have the same modifier. I will modify and check it. Thanks.

linked0:
On CI builds, the assertion failure occurs as follows. I'm debugging it now.
```
Running test on /home/runner/work/agora/agora/./tests/unit/BanManager.d
[stdout]	2020-09-24 08:12:01,233 Trace [agora.common.BanManager] - BanManager: Address 127.0.0.1 banned until 1600935181
[stderr]	core.exception.AssertError@/home/runner/work/agora/agora/submodules/eventcore/source/eventcore/core.d(24): setupEventDriver() was not called for this thread.
```

It's the error that the code in the tests/unit/BanManager.d makes, I guess.
```
auto banman = new BanManager(conf, new Clock(null, null), path);
banman.load();  // should not throw

const IP = "127.0.0.1";
10.iota.each!(_ => banman.onFailedRequest(IP));
assert(banman.isBanned(IP));
```

AndrejMitrovic:
The `getNetTimeOffset` parameter must be non-null. I should have added an `assert` in there but I guess I missed it.

linked0:
> The `getNetTimeOffset` parameter must be non-null. I should have added an `assert` in there but I guess I missed it.

Ah, thanks for the clue.

linked0:
Even if the main() in the tests.unit.BanManager module has no code, the error message is also occurring as follows.

`setupEventDriver() was not called for this thread.`

You can see the results in the CI/Run logs.

AndrejMitrovic:
I think the issue is that we have an import into vibe.d. Clock imports TaskManager, which imports vibe.d

vibe.d has a module destructor and this is where the exception is thrown:

```
[stderr]	core.exception.AssertError@/Users/andrejmitrovic/dev/agora/submodules/eventcore/source/eventcore/core.d(24): setupEventDriver() was not called for this thread.
[stderr]	----------------
[stderr]	??:? _d_assert_msg [0x10e1d4a96]
[stderr]	??:? nothrow @property @nogc @safe eventcore.driver.EventDriver eventcore.core.eventDriver() [0x10dfbff66]
[stderr]	??:? void vibe.core.core.shutdownDriver() [0x10dfd68a8]
[stderr]	??:? void vibe.core.core._sharedStaticDtor_L1543_C1() [0x10dfd66dc]
[stderr]	??:? int rt.minfo.rt_moduleDtor().__foreachbody1(ref rt.sections_elf_shared.DSO) [0x10e2024a8]
[stderr]	??:? int rt.sections_elf_shared.DSO.opApplyReverse(scope int delegate(ref rt.sections_elf_shared.DSO)) [0x10e203650]
[stderr]	??:? rt_term [0x10e1f9130]
[stderr]	??:? void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).runAll() [0x10e1f968a]
[stderr]	??:? _d_run_main2 [0x10e1f943e]
[stderr]	??:? _d_run_main [0x10e1f929c]
[stderr]	??:? main [0x10e0ad544]
```

I think it's a problem in vibe.d's design. It has assumptions about the code which uses it. It assumes that `setupEventDriver` was called.

We will have to work around this, let me think of a good solution.

source/agora/node/FullNode.d
```diff
@@ -373,7 +374,10 @@ public class FullNode : API
     protected Clock getClock (TaskManager taskman)
     {
         // non-synchronizing clock (for now)
-        return new Clock(taskman, (out long time_offset) { return true; });
+        return new Clock(
+            (out long time_offset) { return true; },
+            (Duration duration, void delegate() cb) nothrow @safe
+                {()@trusted {this.taskman.setTimer(duration, cb, Periodic.Yes);}();});
```
AndrejMitrovic: Since this is a one-liner function it's ok to just mark it entirely as `@trusted`:

```suggestion
            (Duration duration, void delegate() cb) nothrow @trusted
                { this.taskman.setTimer(duration, cb, Periodic.Yes); });
```

source/agora/node/Validator.d
```diff
@@ -306,16 +306,17 @@ public class Validator : FullNode, API
...
                 // not enrolled - no need to synchronize clocks
                 if (!this.enroll_man.isEnrolled(this.utxo_set.getUTXOFinder()))
                     return false;
 
                 return this.network.getNetTimeOffset(this.qc.threshold,
                     time_offset);
-            });
+            },
+            (Duration duration, void delegate() cb) nothrow @safe
+                {()@trusted {this.taskman.setTimer(duration, cb, Periodic.Yes);}();});
```
```suggestion
            (Duration duration, void delegate() cb) nothrow @trusted
                { this.taskman.setTimer(duration, cb, Periodic.Yes); });
```

source/agora/network/Clock.d
```diff
@@ -75,6 +74,9 @@ public class Clock
     /// how often the clock should be synchronized with the network
     public const Duration ClockSyncInterval = 1.minutes;
 
+    /// used to set timer for syncing
```
AndrejMitrovic, line 77:
```suggestion
    /// used to set timer for syncing
    /// (not using TaskManager due to bringing in vibe.d dependency)
```

source/agora/test/Base.d
```diff
@@ -1163,7 +1156,9 @@ public class TestClock : Clock
     public this (TaskManager taskman, GetNetTimeOffset getNetTimeOffset,
         shared(time_t)* cur_time)
     {
-        super(taskman, getNetTimeOffset);
+        super(getNetTimeOffset,
+            (Duration duration, void delegate() cb) nothrow @safe
+                {()@trusted {taskman.setTimer(duration, cb, Periodic.Yes);}();});
```
```suggestion
            (Duration duration, void delegate() cb) nothrow @trusted
                { this.taskman.setTimer(duration, cb, Periodic.Yes); });
```

AndrejMitrovic (changes requested):
Just a few more things to simplify.

In the future we can improve this, for now I'm ok with the code duplication.

AndrejMitrovic (approved):
Great work!

linked0:
> Great work!

Thanks, but I will get another link error like "undefined reference to XXX" in the CI builds soon which is what I saw just before. OMG.

AndrejMitrovic:
I know the problem, hold on.

AndrejMitrovic:
Try this:

```diff
diff --git a/dub.json b/dub.json
index bcfccca81..9b0645046 100644
--- a/dub.json
+++ b/dub.json
@@ -91,7 +91,8 @@
                 "source/agora/common/Config.d",
                 "source/agora/common/SCPHash.d",
                 "source/agora/consensus/Quorum.d",
-                "source/agora/network/*",
+                "source/agora/network/NetworkClient.d",
+                "source/agora/network/NetworkManager.d",
                 "source/agora/node/*",
                 "source/agora/utils/PrettyPrinter.d",
                 "source/agora/utils/SCPPrettyPrinter.d",
```

Alternatively we may need to move `Clock` out of the `network` package. But I placed it there precisely because by default it uses network communication. Oh well.

linked0:
Ready for review.

AndrejMitrovic:
I think enough work went into this.

PR #1243, opened 2020-10-05
Refine the formulas for next distance of pre-image revelation

linked0:
The previous formulas for calculating the distance for pre-image revelation is not exactly correct because we have already implemented the time-based revelation. So we just need to reveal pre-images in advance as much as the specified interval based on the current height.

Relates to #1079 which is already closed.


linked0:
Ready for review.

Geod24 (approved):

PR #1249, opened 2020-10-08
Added support for collecting various statistics

ferencdg:
Currently using Prometheus to collect the following statistics:
Transactions: Received, accepted to TxPool, rejected (perhaps even with reason), TxPool size, total amount in pool;
Endpoints: Number of request received per endpoints (e.g. GET /public_key);
Blocks: For each externalized block, new enrollments, transactions, tx total amount;
Validators: Number of validator, number of pre-images known ahead of time;

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -33,6 +33,10 @@ node:
   timeout: 2500
   # Path to the data directory (if the path doesn't exist it will be created)
   data_dir: .cache
+  # The local address where the stats server (currently Prometheus)
+  # is going to connect to, for example: http://0.0.0.0:8008
+  # It can also be set to -1 do disable listening
+  stats_listening_port: 9110
 
 ################################################################################
 ##                             Validator configuration                        ##
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -62,6 +62,8 @@ import agora.consensus.EnrollmentPool;
 import agora.consensus.PreImage;
 import agora.consensus.validation;
 import agora.consensus.ValidatorSet;
+import agora.stats.Utils;
+import agora.stats.Validator;
 import agora.utils.Log;
 version (unittest) import agora.utils.Test;
 
@@ -113,6 +115,12 @@ public class EnrollmentManager
     /// Parameters for consensus-critical constants
     private immutable(ConsensusParams) params;
 
+    /// Validator count stats
+    private ValidatorCountStats validator_count_stats;
+
+    /// Validator preimages stats
+    private ValidatorPreimagesStats validator_preimages_stats;
+
     /***************************************************************************
 
         Constructor
@@ -159,6 +167,8 @@ public class EnrollmentManager
         const uint populate_count = this.getCycleIndex();
         foreach (_; 0 .. populate_count)
             this.cycle.populate(this.key_pair.v, true);
+
+        Utils.getCollectorRegistry().addCollector(&collectValidatorStats);
     }
 
     /***************************************************************************
@@ -709,6 +719,25 @@ public class EnrollmentManager
         {
             log.error("ManagedDatabase operation error {}", ex);
         }
+    }
+
+    ///
+    private void collectValidatorStats (Collector collector)
+    {
+        auto validator_count = validator_set.count();
+        if (!validator_count)
+            validator_count_stats.setMetricTo!"agora_validators_gauge"(validator_count);
+        Hash[] keys;
+        if (getEnrolledUTXOs(keys))
+            foreach (const ref key; keys)
+                validator_preimages_stats.setMetricTo!"agora_preimages_gauge"(
+                    validator_set.getPreimage(key).distance, key.toString());
+
+        foreach (stat; validator_count_stats.getStats())
+            collector.collect(stat.value);
+
+        foreach (stat; validator_preimages_stats.getStats())
+            collector.collect(stat.value, stat.label);
     }
 
     /***************************************************************************
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -41,6 +41,10 @@ import agora.node.BlockStorage;
 import agora.node.Ledger;
 import agora.utils.Log;
 import agora.utils.PrettyPrinter;
+import agora.utils.Utility;
+import agora.stats.EndpointReq;
+import agora.stats.Server;
+import agora.stats.Utils;
 
 import scpd.types.Utils;
 
@@ -51,6 +55,7 @@ import vibe.http.server;
 import vibe.web.rest;
 
 import std.algorithm;
+import std.conv : to;
 import std.exception;
 import std.file;
 import std.path : buildPath;
@@ -90,6 +95,9 @@ public class FullNode : API
     /// Clock instance
     protected Clock clock;
 
+    /// Stats server
+    protected StatsServer stats_server;
+
     /// Network of connected nodes
     protected NetworkManager network;
 
@@ -116,6 +124,8 @@ public class FullNode : API
 
     /// PreImage Received Handler list
     protected PreImageReceivedHandler[Address] preimage_handlers;
+
+    protected EndpointRequestStats endpoint_request_stats;
 
     /***************************************************************************
 
@@ -197,7 +207,11 @@ public class FullNode : API
         // Block externalized handler is set and push for Genesis block.
         if (this.block_handlers.length > 0 && this.getBlockHeight() == 0)
             this.pushBlock(this.params.Genesis);
+
+        Utils.getCollectorRegistry().addCollector(&collectStats);
     }
+
+    mixin DefineCollectorForStats!("endpoint_request_stats", "collectStats");
 
     /***************************************************************************
 
@@ -209,6 +223,7 @@ public class FullNode : API
     {
         this.startPeriodicDiscovery();
         this.network.startPeriodicCatchup(this.ledger);
+        this.startStatsServer();
     }
 
     /***************************************************************************
@@ -245,6 +260,7 @@ public class FullNode : API
         log.info("Shutting down..");
         this.taskman.logStats();
         this.network.dumpMetadata();
+        this.stopStatsServer();
         this.pool = null;
         this.utxo_set = null;
         this.enroll_man = null;
@@ -269,6 +285,7 @@ public class FullNode : API
     /// GET: /node_info
     public override NodeInfo getNodeInfo () pure nothrow @safe
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "node_info", "http");
         return this.network.getNetworkInfo();
     }
 
@@ -286,6 +303,7 @@ public class FullNode : API
 
     public override void putTransaction (Transaction tx) @safe
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "transaction", "http");
         auto tx_hash = hashFull(tx);
         if (this.pool.hasTransactionHash(tx_hash))
             return;
@@ -300,12 +318,14 @@ public class FullNode : API
     /// GET: /has_transaction_hash
     public override bool hasTransactionHash (Hash tx) @safe
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "has_transaction_hash", "http");
         return this.pool.hasTransactionHash(tx);
     }
 
     /// GET: /block_height
     public override ulong getBlockHeight ()
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "block_height", "http");
         return this.ledger.getBlockHeight();
     }
 
@@ -313,6 +333,7 @@ public class FullNode : API
     public override const(Block)[] getBlocksFrom (ulong block_height,
         uint max_blocks)  @safe
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "blocks_from", "http");
         return this.ledger.getBlocksFrom(Height(block_height))
             .take(min(max_blocks, MaxBatchBlocksSent)).array;
     }
@@ -333,6 +354,25 @@ public class FullNode : API
             an instance of a NetworkManager
 
     ***************************************************************************/
+
+    /// Start the StatsServer
+    public void startStatsServer()
+    {
+        stats_server = getStatsServer();
+    }
+
+    /// Stop the StatsServer
+    public void stopStatsServer()
+    {
+        assert (stats_server !is null, "Stats server is stopped without being started");
+        stats_server.shutdown();
+    }
+
+    /// Returns a newly constructed StatsServer
+    protected StatsServer getStatsServer()
+    {
+        return new StatsServer(config.node.stats_listening_port);
+    }
 
     protected NetworkManager getNetworkManager (in Config config,
         Metadata metadata, TaskManager taskman, Clock clock)
@@ -494,12 +534,16 @@ public class FullNode : API
     /// GET: /merkle_path
     public override Hash[] getMerklePath (ulong block_height, Hash hash) @safe
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "merkle_path", "http");
         return this.ledger.getMerklePath(Height(block_height), hash);
     }
 
     /// PUT: /enroll_validator
     public override void enrollValidator (Enrollment enroll) @safe
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "enroll_validator", "http");
+        log.trace("Received Enrollment: {}", prettify(enroll));
+
         if (this.enroll_man.addEnrollment(enroll, this.ledger.getBlockHeight(),
             this.utxo_set.getUTXOFinder()))
         {
@@ -511,12 +555,16 @@ public class FullNode : API
     /// GET: /enrollment
     public override Enrollment getEnrollment (Hash enroll_hash) @safe
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "enrollment", "http");
         return this.enroll_man.getEnrollment(enroll_hash);
     }
 
     /// PUT: /receive_preimage
     public override void receivePreimage (PreImageInfo preimage) @safe
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "receive_preimage", "http");
+        log.trace("Received Preimage: {}", prettify(preimage));
+
         if (this.enroll_man.addPreimage(preimage))
         {
             log.info("Accepted preimage: {}", prettify(preimage));
@@ -528,12 +576,14 @@ public class FullNode : API
     /// GET: /preimage
     public override PreImageInfo getPreimage (Hash enroll_key)
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "preimage", "http");
         return this.enroll_man.getValidatorPreimage(enroll_key);
     }
 
     /// GET /local_time
     public override time_t getLocalTime () @safe nothrow
     {
+        endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(1, "local_time", "http");
         return this.clock.localTime();
     }
 
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -40,10 +40,14 @@ import agora.consensus.UTXOSet;
 import agora.consensus.EnrollmentManager;
 import agora.consensus.validation;
 import agora.node.BlockStorage;
+import agora.stats.Block;
+import agora.stats.Tx;
+import agora.stats.Utils;
 import agora.utils.Log;
 import agora.utils.PrettyPrinter;
 
 import std.algorithm;
+import std.conv : to;
 import std.exception;
 import std.range;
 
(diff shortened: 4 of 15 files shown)


source/agora/stats/EndpointReqestStats.d
```diff
@@ -0,0 +1,48 @@
+module agora.stats.EndpointReqestStats;
```
ferencdg, line 1: I am going to use some compile time machinery to reuse this code accross different statistics, so functions like increase_total_calls_by can be reused

source/agora/node/FullNode.d
```diff
@@ -41,9 +41,12 @@ import agora.node.BlockStorage;
 import agora.node.Ledger;
 import agora.utils.Log;
 import agora.utils.PrettyPrinter;
+import agora.stats.EndpointReqestStats;
+import agora.stats.StatsServer;
```
Geod24: One general point: Since we have package, there's no need to keep on repeating names.
E.g. `agora.stats.StatsServer` => `agora.stats.Server` is enough.

source/agora/network/NetworkManager.d
```diff
@@ -368,6 +369,9 @@ public class NetworkManager
     /// Clock instance
     protected Clock clock;
 
+    /// Stats server
+    protected StatsServer stats_server;
+
```
Geod24: This whole diff feels like it belongs more to the `FullNode`, doesn't it ?

source/agora/network/NetworkManager.d
```diff
@@ -561,6 +565,18 @@ public class NetworkManager
         return false;
     }
 
+    /// Start the StatsServer
+    public void startStatsServer()
+    {
+        stats_server = new StatsServer(node_config.stats_listening_port);
+    }
```
Geod24: Note that for dependency injection, the usual pattern is to `return` the value, not to rely on side effect.

source/agora/node/FullNode.d
```diff
@@ -269,6 +285,7 @@ public class FullNode : API
     /// GET: /node_info
     public override NodeInfo getNodeInfo () pure nothrow @safe
     {
+        endpoint_request_stats.increaseAgoraEndpointCallsTotalBy(1, "getNodeInfo", "http");
```
Geod24: I guess this is the only call so far just for testing ?

Geod24: ~Also where is `increaseAgoraEndpointCallsTotalBy` defined ?~ Ah found it, it's generated
It seems weird to have a "Agora endpoints call total", instead we just give the most precise info (GET to endpoint XXX) and let upstream generate aggregations.

source/agora/stats/EndpointReqestStats.d
```diff
@@ -0,0 +1,16 @@
...
+public struct EndpointRequestStatsLabel
+{
+    string endpoint;
+    string protocol;
+}
+
+public struct EndpointRequestStatsValue
+{
+    ulong agora_endpoint_calls_total;
+}
```
Geod24: If that's an aggregation, that is better left to upstream, as discussed in the other comment

source/agora/stats/EndpointReqestStats.d
```diff
@@ -0,0 +1,16 @@
+module agora.stats.EndpointReqestStats;
+
+import agora.stats.Stats;
+
+public struct EndpointRequestStatsLabel
+{
+    string endpoint;
+    string protocol;
+}
```
Geod24: Question: Can we log the originating IP ?
That might create too many dimensions though, so another possibility is log the peer only if the request is authenticated (comes from a validator).

source/agora/test/Base.d
```diff
@@ -954,6 +955,12 @@ public class TestNetworkManager : NetworkManager
                "' without first creating it");
     }
 
+    ///
+    public override void startStatsServer()
+    {
+        stats_server = new StatsServer(0);
```
Geod24: In Linux, port `0` means "pick whatever". I'm wondering, why does it need to be a different server / port though ? Why not simply expose it in the API (or make an API just for it) ?

source/agora/stats/StatsServer.d
```diff
@@ -0,0 +1,43 @@
+module agora.stats.StatsServer;
+
+import ocean.util.prometheus.collector.CollectorRegistry;
+import vibe.vibe;
+
+import std.stdio;
+
+class StatsServer
+{
+    private HTTPListener http_listener;
```
Geod24: Must of this complication can be avoided by using the REST generator and an API.

E.g.
```D
interface PrometheusAPI
{
    string getMetrics();
}

class FullNode : FullNodeAPI, PrometheusAPI {}
```

Then it just fits into LocalRest. We can also do `registerRestInterface!(PrometheusAPI)(node)` to expose *only* the Prometheus API to Vibe.d's REST generator, and handle the rest ourselves.

Geod24 (changes requested):

source/agora/network/NetworkManager.d
AndrejMitrovic: Also if `stopStatsServer` is called without a call to `startStatsServer`, there will be a segfault.

source/agora/network/NetworkManager.d
ferencdg: when I was implementing the Name Registry I was told not to put network related code into FullNode, that is why we put    
private NameRegistryAPI registry_client;
into NetworkManager

should I move stats_server into the FullNode?

source/agora/network/NetworkManager.d
ferencdg: > stopStatsServer

that would be a programming error though, no one is supposed to call the stopStatsServer without calling startStatsServer. I think it is better to fail with a segfault (or assert)

source/agora/node/FullNode.d
ferencdg: a lot of Prometheus function is dependent on monotonically increasing metrics like 
rate(prometheus_tsdb_head_chunks_created_total[1m])

I think when it comes to http_calls, the Promethues examples also espect aggregated statistics for example:
prometheus_http_requests_total{code="200",handler="/api/v1/label/:name/values"} 1

source/agora/stats/EndpointReqestStats.d
ferencdg: we can log it(as long as I can get the IP somehow), it only increases the dimension number by 1

source/agora/test/Base.d
ferencdg: would you prefer if the user specify -1 if he wants to disable startup up the metrics server?

source/agora/stats/StatsServer.d
ferencdg: we can do that but the real networking code in the StatsServer is very simple:
```D
        auto router = new URLRouter;
        router.get("/metrics", &handle_metrics);

        auto settings = new HTTPServerSettings;
        settings.port = port;
        http_listener = listenHTTP(settings, router);
```

and

```D
 private void handle_metrics (HTTPServerRequest req, HTTPServerResponse res)
 {
     res.writeBody(cast(const(ubyte[])) getCollectorRegistry().collect(),"text/plain");
 }
```
that would be the only thing Rest would help us to avoid

source/agora/node/FullNode.d
ferencdg: > I guess this is the only call so far just for testing ?

yes for now, I only added one call, I am planning to do the rest in a similar manner

ferencdg: > a lot of Prometheus function is dependent on monotonically increasing metrics like
> rate(prometheus_tsdb_head_chunks_created_total[1m])
> 
> I think when it comes to http_calls, the Promethues examples also espect aggregated statistics for example:
> prometheus_http_requests_total{code="200",handler="/api/v1/label/:name/values"} 1

also Promethues has some built-in timeseries 'fix' for the total count: if we restart agora node and the total count starts from 0, Promethues will detect it

source/agora/stats/StatsServer.d
Geod24: Well not the only thing, because then you can call this endpoint from LocalRest for free, no need to version it out / do dependency injection.

source/agora/node/FullNode.d
Geod24: Thanks for the explanation. It LGTM.

source/agora/stats/StatsServer.d
Geod24: Discussed via Slack, and unfortunately the REST generator is too limited for now.

source/agora/stats/Stats.d
```diff
@@ -0,0 +1,163 @@
...
+    public alias StatValueLabelTup = Tuple!(const ValueType, "value", const LabelType, "label");
+    private ValueType[string] stats_maps;
+
+    private static immutable SEPARATOR = "|SEPARATORSYM|";
+
+    /// generating functions with signature
+    /// increase<CamelCaseNameOfMetric>By
+    ///    (LabelTs...)(ulong number, LabelTs labels_packed) pure nothrow @safe
+    static foreach (member_name; __traits(allMembers, ValueType))
+    {
```
ferencdg, line 55: I am going to change all the mixins to use the q{...} tmr, but let me know any other problems

source/agora/node/Ledger.d
```diff
@@ -306,6 +326,28 @@ public class Ledger
...
+            collector.collect(stat.value);
+    }
+
+    ///
+    private ulong get_tx_amount (T)(ref T transactions)
+    {
+        ulong tx_amount;
+        foreach (const ref Transaction tx; transactions)
+            foreach (const ref output; tx.outputs)
+                tx_amount += to!ulong(output.value.toString());
```
ferencdg: any reason why we cannot get the raw amount from the Amount class?

source/agora/stats/Server.d
```diff
@@ -0,0 +1,62 @@
...
+
+            Params
+                port = port on which the server will listen on,
+                -1 to disable listening
+
+    ***************************************************************************/
+
+    public this (int port)
+    {
+        if (port == -1)
```
ferencdg, line 38: if this triggers anyone, I can create a MockStatsServer that does nothing and use it for the unit tests

doc/config.example.yaml
```diff
@@ -1,4 +1,4 @@
-################################################################################
+ ################################################################################
```
Geod24: Superfluous diff

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +718,24 @@ public class EnrollmentManager
...
+    {
+        auto validator_count = validator_set.count();
+        if(!validator_count)
+            validator_stats.setAgoraValidatorsGaugeTo(validator_count);
+        ulong preimage_count;
+        Hash[] keys;
+        if (getEnrolledUTXOs(keys))
+        {
+            foreach(const ref key; keys)
+                preimage_count += validator_set.getPreimage(key).distance;
```
AndrejMitrovic: Hmm.. This is a bit of a strange metric. If the validator re-enrolls then the distance becomes zero again, so the stats will end up varying a lot based on when they are emitted, and also based on when each validator has enrolled.

We should probably think about what would be a more useful statistic w.r.t. preimages. @Geod24 do you have any ideas for a good metric?

source/agora/node/FullNode.d
```diff
@@ -286,6 +299,7 @@ public class FullNode : API
 
     public override void putTransaction (Transaction tx) @safe
     {
+        endpoint_request_stats.increaseAgoraEndpointCallsTotalBy(1, "transaction", "http");
```
AndrejMitrovic: It would be very useful to differentiate between received txs and also accepted txs, two different stats.

Also these stats need to be consistently named. For example the API `hasTransactionHash` stat is named `has_transaction_hash`, but this one is named `transaction`. Let's give it the full name `put_transaction`.

Alternatively we could add a prefix `api_put_transaction`, `api_has_transaction_hash`, so it's very easy to find these in the stats.

source/agora/stats/Stats.d
AndrejMitrovic: Wouldn't it be possible to instead write it to something like:

```D
public void increaseStats (string name, LabelTs...)(ulong number, LabelTs labels)
```

That way we can actually look up functions. Right now with the use of mixins we cannot go to the function definition easily.

Also we should avoid using mixins here, we can use `__traits(getMember, object, "name")` for this use-case.

doc/config.example.yaml
```diff
@@ -33,6 +33,10 @@ node:
   timeout: 2500
   # Path to the data directory (if the path doesn't exist it will be created)
   data_dir: .cache
+  # The local address where the stats server (currently Prometheus)
+  # is going to connect to, for example: http://0.0.0.0:8008
+  # It can also be set to 0 do disable listening
+  stats_listening_port: 8008
```
Geod24: Set the default to `9110`. Prometheus itself listen on `9090` and the `node-exporter` on 9100.

source/agora/network/NetworkManager.d
Geod24: I think it would make more sense in `FullNode` yes, unless I'm missing something.

source/agora/node/FullNode.d
```diff
@@ -197,6 +202,15 @@ public class FullNode : API
         // Block externalized handler is set and push for Genesis block.
         if (this.block_handlers.length > 0 && this.getBlockHeight() == 0)
             this.pushBlock(this.params.Genesis);
+
+
+        StatsServer.getCollectorRegistry().addCollector(&collectStats);
+    }
+
+    void collectStats ( Collector collector )
```
Geod24:
```suggestion
    void collectStats (Collector collector)
```

source/agora/node/FullNode.d
```diff
@@ -197,6 +202,15 @@ public class FullNode : API
...
+    {
+        foreach(stat; endpoint_request_stats.getStats())
```
Geod24:
```suggestion
        foreach (stat; endpoint_request_stats.getStats())
```

source/agora/stats/EndpointReqestStats.d
```diff
@@ -0,0 +1,16 @@
+module agora.stats.EndpointReqestStats;
```
Geod24: Needs ddocc

source/agora/stats/Stats.d
```diff
@@ -0,0 +1,42 @@
+module agora.stats.Stats;
```
Geod24, line 14: Needs ddoc

source/agora/stats/Stats.d
```diff
@@ -0,0 +1,42 @@
...
+import std.array : join, split, byPair;
+import std.typecons : Tuple, tuple;
+
+
+public struct Stats (ValueType, LabelType)
+{
+    public alias StatValueLabelTup = Tuple!(ValueType, "value", LabelType, "label");
+    private ValueType[string] stats_maps;
+
+    static foreach(member_name; __traits(allMembers, ValueType))
```
Geod24:
```suggestion
    static foreach (member_name; __traits(allMembers, ValueType))
```

source/agora/stats/Stats.d
```diff
@@ -0,0 +1,42 @@
...
+            auto key = labels.join(\"|SEPARATOR|\");
+            if(auto metric = key in stats_maps)
+                (*metric)." ~ member_name ~ " += number;
+            else
+            {
+                stats_maps[key] = ValueType.init;
+                stats_maps[key]." ~ member_name ~" = number;
+            }
+        }");
+    }
```
Geod24: That is a weird interface. A single or two-levels associative array would do just fine without all this.
You could even declare the "right" interface using multi-parameters `opIndex`.

source/agora/stats/StatsServer.d
```diff
@@ -0,0 +1,43 @@
+module agora.stats.StatsServer;
```
Geod24: DDOC

source/agora/stats/StatsServer.d
```diff
@@ -0,0 +1,43 @@
...
+
+import ocean.util.prometheus.collector.CollectorRegistry;
+import vibe.vibe;
```
Geod24: Import too broad

source/agora/stats/StatsServer.d
```diff
@@ -0,0 +1,43 @@
...
+
+import std.stdio;
+
+class StatsServer
+{
+    private HTTPListener http_listener;
+
+    public this (ushort port)
+    {
+        if(!port)
```
Geod24:
```suggestion
        if (!port)
```

source/agora/stats/StatsServer.d
```diff
@@ -0,0 +1,43 @@
...
+    public void shutdown ()
+    {
+        if (http_listener !is HTTPListener.init)
+            http_listener.stopListening();
+    }
+
+    public static CollectorRegistry getCollectorRegistry ()
+    {
+        static CollectorRegistry collector_registry;
+        if(collector_registry is null)
```
Geod24:
```suggestion
        if (collector_registry is null)
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +718,24 @@ public class EnrollmentManager
         }
     }
 
+    ///
+    private void collectValidatorStats (Collector collector)
+    {
+        auto validator_count = validator_set.count();
+        if(!validator_count)
```
```suggestion
        if (!validator_count)
```

source/agora/consensus/EnrollmentManager.d
Geod24: I actually suggested this to get an idea of a validator's view over the network. Given that a cycle ends every 2 weeks, I don't think the noticeable drop pattern is such an issue.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +718,24 @@ public class EnrollmentManager
...
+            validator_stats.setAgoraValidatorsGaugeTo(validator_count);
+        ulong preimage_count;
+        Hash[] keys;
+        if (getEnrolledUTXOs(keys))
+        {
+            foreach(const ref key; keys)
```
```suggestion
            foreach (const ref key; keys)
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -711,6 +718,24 @@ public class EnrollmentManager
...
+                preimage_count += validator_set.getPreimage(key).distance;
+            validator_stats.setAgoraPreimagesGaugeTo(validator_count);
+        }
+        foreach(stat; validator_stats.getStats())
```
```suggestion
        foreach (stat; validator_stats.getStats())
```

source/agora/node/Ledger.d
Geod24: Exactly to avoid this usage. `add` must be used to avoid overflow.

Geod24: Also: https://github.com/bpfkorea/agora/blob/0b9f1248f00ba1e96be5d344e8d2d455fabac95e/source/agora/consensus/data/Transaction.d#L263

Geod24 (approved):
Nice. I reviewed commit-by-commit so some comments are updated. The interface is not ideal but not something we should concern ourserves over. Fix the few nits and let's merge.

source/agora/consensus/EnrollmentManager.d
AndrejMitrovic: Umm ok. I think we'll have to wait and see. My prediction is this output won't be very useful. :>

For me it would be more useful to have this stat per-node. But let's go ahead with this for now.

Geod24: Right, I missed this. @ferencdg : The `key` should be a dimension.

source/agora/stats/Stats.d
ferencdg: > That is a weird interface. A single or two-levels associative array would do just fine without all this.
> You could even declare the "right" interface using multi-parameters `opIndex`.

we can have unlimited number of dimensions, how would a single or two-level asociative array would support those?

Geod24: The first level would be for the `member_name`, the second would use `|SEPARATOR|` like your current implementation.

ferencdg: > The first level would be for the `member_name`, the second would use `|SEPARATOR|` like your current implementation.

I think my reasoning here was that we have to eventually create variables of type ValueTypeT(as Ocean's needs it) and we can do that here or in the getStat method. The second reason why I created this extra increase/set methods is to have a little bit of more abstraction over what we do in the increase/set method vs what an associative array can do

source/agora/stats/Stats.d
ferencdg: > Wouldn't it be possible to instead write it to something like:
> 
> ```d
> public void increaseStats (string name, LabelTs...)(ulong number, LabelTs labels)
> ```
> 
> That way we can actually look up functions. Right now with the use of mixins we cannot go to the function definition easily.
> 
> Also we should avoid using mixins here, we can use `__traits(getMember, object, "name")` for this use-case.

that is a good idea, we can get rid of the mixins that way

source/agora/node/FullNode.d
ferencdg: "It would be very useful to differentiate between received txs and also accepted txs, two different stats."
we have those accepted/rejected/received metrics in Transaction.d, this one is only for the endpoint calls

 "Also these stats need to be consistently named. For example the API hasTransactionHash stat is named has_transaction_hash, but this one is named transaction. Let's give it the full name put_transaction."

I named everything based on the name of the http endpoint:
hasTransactionHash function -> has_transaction_hash http endpoint
putTransaction functin -> transaction http endpoint
it would be great if vibe.d would expose the generated http endpoints, so I wouldn't even need to hardcode them

"Alternatively we could add a prefix api_put_transaction, api_has_transaction_hash, so it's very easy to find these in the stats."

we can search for metrics with metric name agora_endpoint_call_total to get all the endpoint(API) metrics

source/agora/stats/EndpointReqestStats.d
ferencdg: hmm I am very confused all these formatting issues were fixed and pushed yesterday, but your comments came in 4 hours ago and refer to the old state of code

ferencdg: oh, I think you mentioned you went commit by commit.... these were fixed in my latest commit, but not squashed yet

PR #1252, opened 2020-10-13
Name registry: don't unnecessarily check signatures & emit better logs

AndrejMitrovic:
We should also move the entire `if (this.registry_client !is null)` block logic into its own task (just like there is `AddressDiscoveryTask`), however I don't have time for this now.


TrustHenry (approved):
LGTM

linked0 (approved):

Geod24 (approved):
One nit: First commit title

PR #1257, opened 2020-10-15
Implement proof of concept of QR code generation

TrustHenry:
Implement a PoC of the 2 QR code generation using two APIs inside Agora.
The first API returns the data that is to be given for login, the second for encryption key.

Related to #1229

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -176,6 +176,7 @@
     ],
 
     "dependencies": {
+        "barcode":          { "version": "*" },
         "base32":           { "version": "*" },
         "bitblob":          { "version": "*" },
         "d2sqlite3":        { "version": "*" },
--- a/dub.selections.json
+++ b/dub.selections.json
@@ -1,6 +1,7 @@
 {
 	"fileVersion": 1,
 	"versions": {
+		"barcode": {"path":"submodules/barcode/"},
 		"base32": {"path":"submodules/base32/"},
 		"bitblob": {"path":"submodules/bitblob/"},
 		"botan": {"path":"submodules/botan/"},
--- a/source/agora/node/Runner.d
+++ b/source/agora/node/Runner.d
@@ -96,6 +96,10 @@ public NodeListenerTuple runNode (Config config)
 
     setTimer(0.seconds, &node.start, Periodic.No);  // asynchronous
 
+    if (config.validator.enabled && config.admin.enabled)
+        // schedule
+        setTimer(0.seconds, &((cast(Validator)node).admin_start), Periodic.No);
+
     log.info("About to listen to HTTP: {}", settings.port);
     return tuple!("node","http_listener")(node, listenHTTP(settings, router));
 }
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -32,6 +32,7 @@ import agora.consensus.protocol.Nominator;
 import agora.consensus.Quorum;
 import agora.network.Clock;
 import agora.network.NetworkManager;
+import agora.node.admin.AdminInterface;
 import agora.node.BlockStorage;
 import agora.node.FullNode;
 import agora.node.Ledger;
@@ -79,6 +80,9 @@ public class Validator : FullNode, API
     /// but LocalScheduler is not instantiated yet.
     private bool started;
 
+    /// admin interface
+    protected AdminInterface admin_interface;
+
     /// Ctor
     public this (const Config config)
     {
@@ -94,6 +98,9 @@ public class Validator : FullNode, API
         // currently we are not saving preimage info,
         // we only have the commitment in the genesis block
         this.regenerateQuorums(Height(0));
+
+        this.admin_interface = new AdminInterface(config,
+            this.config.validator.key_pair, this.clock);
     }
 
     /***************************************************************************
@@ -222,6 +229,17 @@ public class Validator : FullNode, API
 
         if (this.enroll_man.isEnrolled(this.utxo_set.getUTXOFinder()))
             this.nominator.startNominatingTimer();
+    }
+
+    /***************************************************************************
+
+        Begins asynchronous tasks for admin interface.
+
+    ***************************************************************************/
+
+    public void admin_start ()
+    {
+        this.admin_interface.start();
     }
 
     /***************************************************************************
@@ -410,5 +428,7 @@ public class Validator : FullNode, API
     {
         super.shutdown();
         this.nominator.storeLatestState();
+        if (this.config.admin.enabled)
+            this.admin_interface.stop();
     }
 }
--- a/source/agora/node/admin/AdminInterface.d
+++ b/source/agora/node/admin/AdminInterface.d
@@ -1,0 +1,134 @@
+/*******************************************************************************
+
+    Provide a web-based administrative interface to the node
+
+    To ease administration and configuration, users can interact with the node
+    using a web interface. This interface is served by Vibe.d,
+    usually on a port adjacent to the node port (2827 by default).
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.admin.AdminInterface;
+
+import agora.common.Config;
+import agora.common.crypto.Key;
+import agora.node.admin.QRCodeInterface;
+import agora.utils.Log;
+import agora.network.Clock;
+
+import vibe.core.core;
+import vibe.data.json;
+import vibe.http.router;
+import vibe.http.server;
+
+import std.format;
+import std.meta;
+import std.traits;
+
+mixin AddLogger!();
+
+/*******************************************************************************
+
+    Class for Admin interface
+
+    If admin is enabled this interface listens by AdminConfig's address
+    of config
+    Contains services that include authentication QR.
+    This interface must be in a path accessible only to the owner of the node.
+
+    This will added by node management and monitoring interfaces.
+
+    API:
+        GET /loginQR : Respond with QR code containing login information.
+        GET /encryptionKeyQR : Respond with QR code containing EncryptionKey.
+
+*******************************************************************************/
+
+public class AdminInterface
+{
+    /// HTTP listener
+    private HTTPListener listener;
+
+    /// Config instance
+    private const Config config;
+
+    /// The keypair of this node
+    private const KeyPair key_pair;
+
+    /// Clock instance
+    private Clock clock;
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            config = config instance
+            key_pair = the keypair of this node
+            clock = clock instance
+
+    ***************************************************************************/
+
+    public this (const Config config, const KeyPair key_pair, Clock clock)
+    {
+        this.config = config;
+        this.key_pair = key_pair;
+        this.clock = clock;
+    }
+
+    /***************************************************************************
+
+        Start listening for requests
+
+        Begins asynchronous tasks for admin interface
+
+    ***************************************************************************/
+
+    public void start ()
+    {
+        if (!this.config.admin.enabled)
+            assert(0, "Admin interface is not enabled in config settings.");
+
+        auto settings = new HTTPServerSettings(this.config.admin.address);
+        settings.port = this.config.admin.port;
+
+        auto router = new URLRouter;
+
+        // API endpoints for this validator
+        if (this.config.validator.enabled)
+        {
+            auto qrCode = new QRCodeInterface(this.key_pair, this.clock);
+
+            router.get("/loginQR", &qrCode.loginQRCreate);
+            router.get("/encryptionKeyQR", &qrCode.keyQRCreate);
+        }
+
+        log.trace("Started admin interface ...");
+        this.listener = listenHTTP(settings, router);
+
+        log.info("Admin interface listening to HTTP: http://{}:{}",
+            settings.bindAddresses[0], settings.port);
+    }
+
+    /***************************************************************************
+
+        Stop listening for requests
+
+    ***************************************************************************/
+
+    public void stop ()
+    {
+        if (!this.config.admin.enabled)
+            assert(0, "Admin interface is not enabled");
+
+        log.trace("Stopping admin interface ...");
+        this.listener.stopListening();
+    }
+}
--- a/source/agora/node/admin/QRCodeInterface.d
+++ b/source/agora/node/admin/QRCodeInterface.d
@@ -1,0 +1,289 @@
+/*******************************************************************************
+
+    Provides QR service interface used by the node's web-based management
+    interface
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.admin.QRCodeInterface;
+
+import agora.common.Config;
+import agora.common.crypto.Key;
+import agora.common.Hash;
+import agora.common.Types;
+import agora.consensus.data.PreImageInfo;
+import agora.utils.Log;
+import agora.network.Clock;
+
+import barcode;
+
+import vibe.data.json;
+import vibe.http.router;
+
+import core.stdc.time;
+import std.datetime : SysTime, unixTimeToStdTime;
+import std.datetime.timezone : UTC;
+import std.format;
+
+import core.time : days;
+
+mixin AddLogger!();
+
+/*******************************************************************************
+
+    Implementation of the Agora User's credentials QR services
+
+    A class to provide a login or encryption key to QRs to verify that
+    the owner of the validator from outside agora.
+    Respond SVG format's QR according to the get requests.
+    Contains QR code generator.
+
+*******************************************************************************/
+
+public class QRCodeInterface
+{
+    /// The keypair of this node
+    private const KeyPair key_pair;
+
+    /// Clock instance
+    private Clock clock;
+
+    /***************************************************************************
+
+        Constructor
+
+        Params:
+            key_pair = the keypair of this node
+
+    ***************************************************************************/
+
+    public this (const KeyPair key_pair, Clock clock)
+    {
+        this.key_pair = key_pair;
+        this.clock = clock;
+    }
+
+    /***************************************************************************
+
+        Respond with QR code containing login information.
+
+        Enter the name of the APP where the login information will be
+        used as the get query.
+
+        Convert LoginInfo to JSON format and respond with the QR code
+        of SVG format.
+        All public key cryptography is done on Curve25519.
+        The signature of the vectorCard provides verification function in the SDK.
+        This LoginInfo has an expiration time.
+        It provide a randomly generated temporary private key encoded
+        in strings on demand.
+        In the future, there is a need to enter a temporary key.
+
+    ***************************************************************************/
+
+    public void loginQRCreate (scope HTTPServerRequest req,
+        scope HTTPServerResponse res)
+    {
+        // The randomly generated temporary KeyPair
+        KeyPair temp_kp = KeyPair.random();
+
+        time_t current_time = this.clock.networkTime();
+        // A 'nonce' used to allow expiring
+        current_time += 60 * 60 * 24 * 90; // (default: +90 days)
+        SysTime expires_time = SysTime(unixTimeToStdTime(current_time));
+
+        LoginInfo login_info = LoginInfo(
+            temp_kp.secret.toString(PrintMode.Clear),
+            VoterCard(
+                this.key_pair.address,
+                temp_kp.address,
+                expires_time.toISOExtString()
+            )
+        );
+
+        login_info.voter_card.signature =
+            this.key_pair.secret.sign(hashFull(login_info.voter_card)[]);
+
+        res.headers["Content-Type"] = "image/svg+xml";
+        res.headers["Vary"] = "Accept-Encoding";
+        res.writeBody(createQRcode(login_info));
+    }
+
+    /***************************************************************************
+
+        Respond with QR code containing EncryptionKey.
+
+        Enter the name of the APP & height where the encryptionKey will be used
+        as the get query.
+
+    ***************************************************************************/
+
+    public void keyQRCreate (scope HTTPServerRequest req,
+        scope HTTPServerResponse res)
+    {
+        enforceHTTP("app" in req.query && "height" in req.query,
+            HTTPStatus.badRequest, "Missing app/height field.");
+
+        auto app = req.query["app"];
+        auto height = Height.fromString(req.query["height"]);
+
+        // TODO: PreImage to be released at that height
+        // Or it can be pre-released, so it can be `height + PreimageRevealPeriod`
+        PreImageInfo pre_image;
+
+        // blake2b hashing
+        const Hash key = hashMulti(pre_image, app);
+
+        EncryptionKey encryptionKey = EncryptionKey(
+            app,
+            height,
+            key
+        );
+
+        res.headers["Content-Type"] = "image/svg+xml";
+        res.headers["Vary"] = "Accept-Encoding";
+        res.writeBody(createQRcode(encryptionKey));
+    }
+
+    /***************************************************************************
+
+        SVG format QR code generator
+
+        Use barcode library to make QR code of JSON format.
+        Convert the data structure to JSON and return the QR code image
+        of the SVG format.
+
+        Params:
+            t = Type of struct to QR code
+
+        Returns:
+            QR code of XML string
+
+    ***************************************************************************/
+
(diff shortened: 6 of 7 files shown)


source/agora/node/Validator.d
```diff
@@ -38,6 +38,7 @@ import agora.node.Ledger;
 import agora.registry.NameRegistryAPI;
 import agora.utils.Log;
 import agora.utils.PrettyPrinter;
+import agora.test.QRCodeInterface;
```
Geod24: No module outside of `agora.test` should *ever* import from `agora.test`.

dub.json
```diff
@@ -179,6 +179,7 @@
         "libsodiumd":       { "path": "submodules/libsodiumd/", "version": "*" },
         "localrest":        { "path": "submodules/localrest/", "version": "*" },
         "ocean":            { "path": "submodules/ocean/", "version": "*" },
-        "vibe-d":           { "path": "submodules/vibe.d/", "version": "*" }
+        "vibe-d":           { "path": "submodules/vibe.d/", "version": "*" },
+        "barcode":           { "path": "submodules/barcode/", "version": "*" }
```
```suggestion
        "barcode":          { "path": "submodules/barcode/", "version": "*" }
```

source/agora/node/QRCodeInterface.d
```diff
@@ -0,0 +1,278 @@
...
+mixin AddLogger!();
+
+/*******************************************************************************
+
+    Class for Agora User's credentials QR services Interface
+
+    A class to provide a login or encryption key to QRs to verify that
+    the owner of the validator from outside agora.
+
+    This interface listens on 127.0.0.1:2829.
```
Geod24: This has to be configurable, otherwise you can't run 2 servers on the same machine (which we do)

source/agora/node/QRCodeInterface.d
```diff
@@ -0,0 +1,278 @@
...
+    public void start ()
+    {
+        auto settings = new HTTPServerSettings("127.0.0.1");
+        settings.port = 2829;
+        auto router = new URLRouter;
+        router
+            .get("/loginQR", &this.loginQRCreate)
+            .get("/encryptionKeyQR", &this.keyQRCreate);
+        this.listener = listenHTTP(settings, router);
+        scope (exit) this.listener.stopListening();
```
Geod24: That looks wrong. `listenHTTP` doesn't block, so  we `stopListening` immediately.

source/agora/node/QRCodeInterface.d
```diff
@@ -0,0 +1,278 @@
...
+        enforceHTTP("app" in req.query,
+            HTTPStatus.badRequest, "Missing app field.");
+
+        auto app = req.query["app"];
+
+        // The randomly generated temporary KeyPair
+        KeyPair temp_kp = KeyPair.random();
+
+        SysTime expires_time = Clock.currTime();
+        auto expires_term = 1.days;
```
Geod24: 1 day is way too short. We should talk to service planning, but for the time being, make it 3 months.

source/agora/node/QRCodeInterface.d
```diff
@@ -0,0 +1,278 @@
...
+
+unittest
+{
+    import std.functional : toDelegate;
+    import vibe.http.client;
+    import vibe.stream.operations;
+    import vibe.core.core;
+    import core.stdc.time;
+    import core.thread;
+    import std.stdio;
```
Geod24: We should **never** use Vibe.d in unittests

source/agora/node/data/EncryptionKey.d
```diff
@@ -0,0 +1,32 @@
...
+    public Height height;
+
+    public Hash value;
+}
+
+unittest
+{
+    import agora.common.Serializer;
+    testSymmetry!EncryptionKey();
+}
```
Geod24: Since this is only used inside of the QR code facility and not shared with the rest of the code, it should be in that module. Also we don't need  the symmetry test (although they don't hurt) because we never serialize it.

source/agora/node/data/LoginInfo.d
```diff
@@ -0,0 +1,66 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.data.LoginInfo;
```
Geod24: Same comment as per `EncryptionKey.d`

Geod24 (changes requested):

source/agora/node/QRCodeInterface.d
TrustHenry: I will to set this as a configuration option.
Should I run two servers(agora, this) on the same machine in different ports?

source/agora/node/Validator.d
TrustHenry: I missed the change. I changed the package path.

source/agora/node/QRCodeInterface.d
TrustHenry: I Set this expiration date to 90 days.

source/agora/node/QRCodeInterface.d
TrustHenry: Remove the Unit Test .

source/agora/node/data/EncryptionKey.d
TrustHenry: I will move these structures into `QRCodeInterface.d`

source/agora/node/QRCodeInterface.d
TrustHenry: @Geod24 I will to set this as a configuration option.
Should I run two servers(agora, this) on the same machine in different ports?

Geod24: It should be the same process listening on two ports, yes

TrustHenry: https://github.com/bpfkorea/agora/blob/1e6bafc901588eef708d8177537a9c3639d777c9/source/agora/common/Config.d#L186-L198
@Geod24 What do you think about using the same address and separate port as the admin that will be applied in the future?

TrustHenry: @Geod24 ping
Please review the newly changed code.

Geod24: Actually, same address and same port makes sense.

TrustHenry: If the port configuration is the same as admin, the QR interface is:
Should I consider it with the admin server?

Geod24: Yes

TrustHenry: I created the admin interface by referring to # 483.
This is the admin interface with different settings than node initialization logic.
In the future, the management or monitoring interface can be added to the admin interface.

TrustHenry:
@Geod24 Do I end these codes with proof of concept?
Or will we continue to work to merge?

source/agora/node/FullNode.d
```diff
@@ -128,6 +129,9 @@ public class FullNode : API
     /// Endpoint request stats
     protected EndpointRequestStats endpoint_request_stats;
 
+    /// admin interface
+    protected AdminInterface adminInterface;
```
AndrejMitrovic: Please follow the style guide. `admin_interface`

source/agora/node/admin/AdminInterface.d
```diff
@@ -0,0 +1,130 @@
...
+    /***************************************************************************
+
+        Stop listening for requests
+
+    ***************************************************************************/
+
+    public void stop ()
+    {
+        log.trace("Stopping admin interface ...");
+        this.listener.stopListening();
```
AndrejMitrovic, line 132: You have a check for `if (!this.config.admin.enabled)` in `start()`, but not in `stop()`. That means you will call `stop()` on a non-initialized `HTTPListener`.

It's better to just move this `this.config.admin.enabled` up into the class that uses it (FullNode), and instead use `assert` for these conditions in `start` and `stop`.

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,261 @@
+/*******************************************************************************
+
+    Provides QR service interface used by the node's web-based management
+    interface
+
+    Copyright:
+        Copyright (c) 2019-2020 BOS Platform Foundation Korea
```
AndrejMitrovic:
```suggestion
        Copyright (c) 2020 BOS Platform Foundation Korea
```

Copyright begins at the date the module is first introduced.

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,261 @@
...
+import vibe.http.router;
+
+import std.datetime.date : DateTime;
+import std.datetime.systime : SysTime, Clock;
+import std.datetime.timezone : UTC;
+import std.format;
+
+import core.time : days;
+
+import barcode;
```
AndrejMitrovic: This import should be moved before the `vibe` import.

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,261 @@
...
+    A class to provide a login or encryption key to QRs to verify that
+    the owner of the validator from outside agora.
+    Respond SVG format's QR according to the get requests.
+    Contains QR code generator.
+
+*******************************************************************************/
+
+public class QRCodeInterface
+{
+
```
AndrejMitrovic:
```suggestion
```

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,261 @@
...
+        Params:
+            key_pair = the keypair of this node
+
+    ***************************************************************************/
+
+    public this (const KeyPair key_pair)
+    in
+    {
+        assert(KeyPair.init != key_pair);
```
AndrejMitrovic: But you are explicitly instantiating this class in the `FullNode`. And the full node's key pair will be `KeyPair.init` if it really is a full node.

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,261 @@
...
+
+        auto app = req.query["app"];
+
+        // The randomly generated temporary KeyPair
+        KeyPair temp_kp = KeyPair.random();
+
+        SysTime expires_time = Clock.currTime();
+        auto expires_term = 90.days;
+        // A 'nonce' used to allow expiring
+        expires_time += expires_term;
```
AndrejMitrovic: There is no need for a variable if it's only used once.

```suggestion
        // A 'nonce' used to allow expiring
        expires_time += 90.days;
```

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,261 @@
...
+
+        LoginInfo loginInfo = LoginInfo(
```
AndrejMitrovic: Style guide.

```suggestion
        LoginInfo login_info = LoginInfo(
```

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,261 @@
...
+        The encryption key is derived from the pre-image as follows:
+        Hash(pre_image, "APPNAME");
+
+    ***************************************************************************/
+
+    public void keyQRCreate (scope HTTPServerRequest req,
+        scope HTTPServerResponse res)
+    {
+        enforceHTTP("app" in req.query && "height" in req.query,
+        HTTPStatus.badRequest, "Missing app/height field.");
```
AndrejMitrovic:
```suggestion
            HTTPStatus.badRequest, "Missing app/height field.");
```

source/agora/node/admin/AdminInterface.d
```diff
@@ -0,0 +1,130 @@
...
+        else
+        {
+        // API endpoints for this fullnode
+        }
+
+        this.listener = listenHTTP(settings, router);
+
+        string addr = format("http://%s:%d",
+            settings.bindAddresses[0], settings.port);
+        log.info("Admin interface listening to HTTP: {}", addr);
```
AndrejMitrovic:
```suggestion
        log.info("Admin interface listening to HTTP: http://{}:{}", 
            settings.bindAddresses[0], settings.port);
```

source/agora/node/admin/AdminInterface.d
```diff
@@ -0,0 +1,130 @@
...
+        this.key_pair = key_pair;
+    }
+
+    /***************************************************************************
+
+        Start listening for requests
+
+    ***************************************************************************/
+
+    public void start ()
```
AndrejMitrovic, line 94: Can this function throw? It should be documented.

AndrejMitrovic (dismissed):
First wave of reviews.

TrustHenry:
@Geod24 @AndrejMitrovic  Thank you for the reviews.
All suggestions applied

source/agora/node/Validator.d
```diff
@@ -222,6 +231,9 @@ public class Validator : FullNode, API
 
         if (this.enroll_man.isEnrolled(this.utxo_set.getUTXOFinder()))
             this.nominator.startNominatingTimer();
+
+        if (this.config.admin.enabled)
+            setTimer(0.seconds, &admin_interface.start, false);  // asynchronous
```
```suggestion
            setTimer(0.seconds, &this.admin_interface.start, Periodic.No);  // asynchronous
```

Is this used as a timer or as a `schedule`? If it's used as a schedule please add a comment.

TrustHenry: Use as timer.
admin is the same as the drive lifecycle of the current node.
If admin controls the operation of the node or if it is Initialization, it may be different.

AndrejMitrovic: > Use as timer.

That doesn't make sense. Why would you call `start()` multiple times?

AndrejMitrovic: You set the timer for 0 seconds, right? And there is no repeating. That means it's a `schedule` call. So it should be documented like that.

TrustHenry: I guess I was mistaken. Okay, I'll fix it. Thank you.

AndrejMitrovic: Yes please just add a comment `// schedule`

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,269 @@
...
+
+        // blake2b hashing
+        const Hash key = hashMulti(pre_image, app);
+
+        EncryptionKey encryptionKey = EncryptionKey(
+            app,
+            height,
+            key
+        );
+
```
MichaelKim20, line 149: The protocol ID must be entered as get parameter.
Hash key and proposal ID once again. The final value is the encryption key.
https://github.com/bpfkorea/infrastructure/tree/master/docs/votera#getting-an-encryption-key

TrustHenry: It has to be hashed at the Votera app.
`Hash prop_encryption_key = blake2bHash(encryption_key, propID);`

source/agora/node/Validator.d
```diff
@@ -41,6 +42,8 @@ import agora.utils.PrettyPrinter;
 
 import scpd.types.Stellar_SCP;
 
+import vibe.core.core;
+
```
Geod24: We can't import `vibe.core.core` here as it would break the business code separation we have (and thus LocalRest)

Geod24: You can move it to the `Runner`

source/agora/node/admin/AdminInterface.d
```diff
@@ -0,0 +1,127 @@
...
+        Begins asynchronous tasks for admin interface
+
+    ***************************************************************************/
+
+    public void start ()
+    {
+        if (!this.config.admin.enabled)
+            assert("Admin interface is not enabled in config settings.");
+
+        log.trace("Started admin interface ...");
```
Geod24: It's only started at L107

source/agora/node/admin/AdminInterface.d
```diff
@@ -0,0 +1,127 @@
...
+    /***************************************************************************
+
+        Stop listening for requests
+
+    ***************************************************************************/
+
+    public void stop ()
+    {
+        if (!this.config.validator.enabled)
+            assert("Admin interface is not enabled");
```
Geod24: This is an `assert(true);`

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,269 @@
...
+        The encryption key is derived from the pre-image as follows:
+        Hash(pre_image, "APPNAME");
+
+    ***************************************************************************/
+
+    public void keyQRCreate (scope HTTPServerRequest req,
+        scope HTTPServerResponse res)
+    {
+        enforceHTTP("app" in req.query && "height" in req.query,
+            HTTPStatus.badRequest, "Missing app/height field.");
```
Geod24, line 132: Nice

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,269 @@
...
+*******************************************************************************/
+
+public struct LoginInfo
+{
+    // The randomly generated temporary private key encoded as string
+    public string private_key;
+
+    public VoterCard voter_card;
+
+}
```
Geod24: Need a bit of documentation / style fix.

Geod24 (commented):
Nice! Looks mostly good. I'll give it another round tomorrow, but I think this is ready. Anything missing from your side ?

dub.selections.json
```diff
@@ -21,4 +22,4 @@
 		"vibe-core": {"path":"submodules/vibe-core/"},
 		"vibe-d": {"path":"submodules/vibe.d/"}
 	}
-}
+}
```
AndrejMitrovic: Why did this happen again? Didn't you fix your text editor?

source/agora/node/admin/AdminInterface.d
```diff
@@ -0,0 +1,128 @@
...
+        Start listening for requests
+
+        Begins asynchronous tasks for admin interface
+
+    ***************************************************************************/
+
+    public void start ()
+    {
+        if (!this.config.admin.enabled)
+            assert("Admin interface is not enabled in config settings.");
```
AndrejMitrovic: ?

You are asserting on a string.

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,269 @@
...
+
+module agora.node.admin.QRCodeInterface;
+
+import agora.common.Config;
+import agora.common.crypto.Key;
+import agora.common.Hash;
+import agora.common.Types;
+import agora.consensus.data.PreImageInfo;
+import agora.utils.Log;
+import agora.network.Clock;
```
AndrejMitrovic, line 23: Why are you importing this Clock but then using the one from Phobos?

AndrejMitrovic: If we want to write tests for this module, then we should use the Clock from the Network package. That way we can set the exact date and can even write tests. Obviously no tests were added to this PR, so it will have to be done later..

source/agora/node/Validator.d
TrustHenry: I change the start of the admin interface to start from Runner.

source/agora/node/admin/AdminInterface.d
TrustHenry: Fixed it.

dub.selections.json
TrustHenry: There was a big change in the upstream, It fell out while working rebase. ;)

source/agora/node/admin/AdminInterface.d
TrustHenry: Fixed it.

source/agora/node/admin/QRCodeInterface.d
TrustHenry: I was going to use it at first. The change was big, so I left it for a while.
But now I've applied it.

TrustHenry:
>  Nice! Looks mostly good. I'll give it another round tomorrow, but I think this is ready. Anything missing from your side ?

There is something that we think should proceed in the next step rather than in the missing part.
The definition of the pre-image is not clear in the part of the encryption key derived from the pre-image.
I think we need to create a new issue for this.

source/agora/node/admin/QRCodeInterface.d
TrustHenry: Added document and removed a line.

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,289 @@
...
+    // Public key of the `private_key` field
+    public PublicKey address;
+
+    // A 'nonce' used to allow expiring / replacing credentials
+    // In this case, we just use an expiration date in ISO8601
+    public string expires;
+
+    // Finally, the signature, made using `validator_address` private key
+    // (IOW: the `Validator` private key) on this whole packet.
+    public Hash signature;
```
AndrejMitrovic: Why is this a `Hash` instead of a `Signature`?

TrustHenry: I declared it hash for compatibility with boa-sdk.
Signature is BitBlob!512 is an alias, so I will modify it to Signature .

TrustHenry:
**[AD] Ready for the review!**

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,289 @@
...
+            temp_kp.secret.toString(PrintMode.Clear),
+            VoterCard(
+                this.key_pair.address,
+                temp_kp.address,
+                expires_time.toISOExtString()
+            )
+        );
+
+        login_info.voter_card.signature =
+            this.key_pair.secret.sign(hashFull(login_info)[]);
```
MichaelKim20: Isn't this a good way to do it?
```suggestion
            this.key_pair.secret.sign(hashFull(login_info. voter_card)[]);
```

TrustHenry: It makes sense.

source/agora/node/admin/QRCodeInterface.d
```diff
@@ -0,0 +1,289 @@
...
+    public void loginQRCreate (scope HTTPServerRequest req,
+        scope HTTPServerResponse res)
+    {
+        // The randomly generated temporary KeyPair
+        KeyPair temp_kp = KeyPair.random();
+
+        time_t current_time = this.clock.networkTime();
+        // A 'nonce' used to allow expiring
+        current_time += 60 * 60 * 24 * 90; // (default: +90 days)
+        SysTime expires_time = SysTime(unixTimeToStdTime(current_time));
```
MichaelKim20, line 100: I just ran it and I was able to get this.
expores_time: "2021-01-28T10:24:06"
It's working fine.

TrustHenry:
**[Advertising] Ready for the review! :speech_balloon:**

MichaelKim20 (approved):
LVGTM

TrustHenry:
`./build/agora -c doc/config.example.yaml`

You can see QR at this URLs.
```
http://127.0.0.1:2827/loginQR
http://127.0.0.1:2827/encryptionKeyQR?app=Votera&height=100
```

source/agora/node/Runner.d
```diff
@@ -96,6 +96,10 @@ public NodeListenerTuple runNode (Config config)
 
     setTimer(0.seconds, &node.start, Periodic.No);  // asynchronous
 
+    if (config.validator.enabled && config.admin.enabled)
+        // schedule
+        setTimer(0.seconds, &((cast(Validator)node).admin_start), Periodic.No);
+
```
Geod24, line 102: 99% of casts in D are a mistake, because the type system is powerful enough that you don't need them.
In this case, one could trivially move the `setTimer` call in the `if` clause and use `inst` which is of type `Validator`.

But an even better approach is to just merge `admin_start` with `start` and remove this all together.

source/agora/node/Validator.d
```diff
@@ -94,6 +98,9 @@ public class Validator : FullNode, API
         // currently we are not saving preimage info,
         // we only have the commitment in the genesis block
         this.regenerateQuorums(Height(0));
+
+        this.admin_interface = new AdminInterface(config,
+            this.config.validator.key_pair, this.clock);
```
Geod24, line 103: What's the point of passing the `config` AND the `key_pair` ?
I had a look at the code, and it is quite tangled:
- If the node needs the full `config` object (which is extremely rare), then it doesn't need to receive `key_pair` as well;
- If the node doesn't need the full `config` object, how can we reduce what we pass.

Looking at the code, it does a bunch of checks that `validator` and `admin` are enabled. This is the wrong way to think about it. If `start` gets called, we just start, without checking the config again.

The important point is to separate concern: You want one part of the code (object, function) to be responsible for only one thing. Duplicating those checks means that responsibility is not clear and things gets tangled (and unworkable) pretty quickly from there.

Geod24 (approved):
Nice.

Issue #1214, opened 2020-09-17
Add a way to store data in a block


Geod24:
For Votera (and other apps), we need to be able to store raw data in the chain.

## High level requirements:

- It must come at a cost, which might not be linear (to allow small packet of data but forbid images & co, at least in a contiguous chunk);
- This ability has been exploited in Bitcoin to introduce illegal content, or nuisible content (e.g. signature of a virus to trigger the AV);
- A vote data should be < 400 bytes, perhaps even 300;
- If storing data relies on UTXO, it should not pollute the UTXO (retain the data in the UTXO set).

## Suggested approach

Since storing data puts a price to the network, the suggested is to special-case a transaction.
Create the following structure for a transaction:
1) Perhaps, introduce a new transaction type (not mandatory but might help);
2) The transaction must have 1 or 2 outputs;
3) One of the outputs MUST be towards the commons budget (see #1213);
4) The second output is for refund;
5) If those rules are satisfied, the transaction can contain a certain amount of data, of which the size is dependent on the amount in the output towards the common budget;

Geod24:
Depends on #1213

Geod24:
We have https://github.com/bpfkorea/agora/pull/1258 but it's a WIP, so will not be closed this sprint.

PR #1258, opened 2020-10-15
[WIP] Implement a new transaction type to store raw data

MichaelKim20:
I defined a new transaction type to store raw data.
I added a member to the Transaction.
When deserializing the binary block data, an error occurred in the test code. I fixed this test code.

Relates to #1214


source/agora/consensus/data/Transaction.d
```diff
@@ -53,6 +54,9 @@ public struct Transaction
     /// The list of newly created outputs to put in the UTXO
     public Output[] outputs;
 
+    /// The data to store
+    public ubyte[] data;
+
```
Geod24: It needs to be a different type altogether. We cannot have empty fields like this. The deserializer is the one that will have to do the magic.

MichaelKim20:
I implemented the transaction type that can store raw data, and added some to the transaction validation.

source/agora/common/TransactionData.d
```diff
@@ -0,0 +1,284 @@
...
+
+    public static TransactionDataT fromBinary (TransactionDataT = TransactionData) (
+        scope DeserializeDg dg, const ref DeserializerOptions opts) @safe
+    {
+        // This is not worked
+        /*
+        auto tx_data_opts = DeserializerOptions.Default;
+        tx_data_opts.maxLength = MAX_SIZE;
+        ubyte[] data = deserializeFull!(ubyte[])(dg, tx_data_opts);
+        return TransactionData(data);
```
MichaelKim20: I don't understand why this code doesn't work.

source/agora/common/TransactionData.d
```diff
@@ -0,0 +1,284 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.TransactionData;
```
Geod24: Module should be in `data` package, like other structs

source/agora/common/TransactionData.d
```diff
@@ -0,0 +1,284 @@
...
+
+import agora.common.Hash;
+import agora.common.Serializer;
+
+static import std.ascii;
```
Geod24: Why the static import ?

source/agora/common/TransactionData.d
```diff
@@ -0,0 +1,284 @@
...
+    invariant()
+    {
+        assert(length <= MAX_SIZE);
+    }
+
+    /// The maximum size
+    public static immutable MAX_SIZE = 1024;
+
+    /// The byte array of transaction data
+    public ubyte[MAX_SIZE] data;
```
Geod24: This means the data will always be 1kb in memory, even for 300 bytes. That's not desireable.

source/agora/common/TransactionData.d
```diff
@@ -0,0 +1,284 @@
...
+
+    /// The number of bytes of transaction data
+    public size_t length;
+
+    /// If the input data is larger than MAX_SIZE, this value is true, if not false.
+    public bool is_oversize;
```
Geod24: What's the reasoning for such a flag ?

source/agora/common/TransactionData.d
```diff
@@ -0,0 +1,284 @@
...
+        foreach (chunk; range.chunks(2))
+        {
+            (() @trusted {
+                this.data[offset++] = cast(ubyte)((chunk[0] << 4) + chunk[1]);
+            })();
+            if (offset == MAX_SIZE)
+                break;
+        }
+        this.length = offset;
+    }
```
Geod24: Why do we need an explicit constructor for string => this struct ?

source/agora/common/TransactionData.d
```diff
@@ -0,0 +1,284 @@
...
+        size_t idx;
+        char[] buffer;
+        buffer.length = this.length * 2 + 2;
+        scope sink = (const(char)[] v) {
+            buffer[idx .. idx + v.length] = v;
+            idx += v.length;
+        };
+        this.toString(sink);
+        return buffer.idup;
+    }
```
Geod24: A large portion of this file seems to be a copy of `BitBlob`. I don't understand why you didn't just use `BitBlob`, and add the `oversize` flag, instead of copying the code.

source/agora/consensus/UTXOSet.d
```diff
@@ -81,6 +81,19 @@ public class UTXOSet
...
+        if ((tx.type == TxType.Data)
+            && tx.inputs.any!(input =>
+                (
+                    (input.previous != Hash.init) &&
+                    (this.getUTXOSetValue(input.previous, input.index).type == TxType.Freeze)
+                )
+            )
+        )
+        {
+            assert(0, "A frozen UTXO cannot be used for data storage.");
```
Geod24: Why is this an `assert` ?

source/agora/consensus/data/Transaction.d
```diff
@@ -53,6 +55,9 @@ public struct Transaction
     /// The list of newly created outputs to put in the UTXO
     public Output[] outputs;
 
+    /// The data to store
+    public TransactionData data;
+
```
Geod24, line 61: This adds > 1kb of data to *every single transaction*. Not something acceptable.

source/agora/consensus/validation/Transaction.d
```diff
@@ -89,14 +89,14 @@ public string isInvalidReason (
             if (auto fail_reason = isInvalidInput(input, utxo_value, sum_unspent))
                 return fail_reason;
 
-            if (utxo_value.type != TxType.Payment)
-                return "Transaction: Can only freeze a Payment transaction";
+            if ((utxo_value.type != TxType.Payment) && (utxo_value.type != TxType.Data))
+                return "Transaction: Can freeze a Payment or Data transaction";
```
Geod24: `only` was valid here

source/agora/consensus/validation/Transaction.d
```diff
@@ -119,9 +119,22 @@ public string isInvalidReason (
...
+            if ((count_freeze > 0) && (count_freeze != tx.inputs.length))
+                return "Transaction: Rejected combined inputs (freeze & payment)";
+        }
+        else
+        {
+            if (count_freeze > 0)
+                return "Transaction: Frozen UTXO is used for data storage";
+
+            if (tx.data.is_oversize)
+                return "Transaction: Data is too large (Maximum size is 1024 bytes)";
```
Geod24: This just reads the data sent by the client. If the client make its own binary data and set the flag to false, this will pass.

source/agora/common/TransactionData.d
```diff
@@ -0,0 +1,284 @@
...
+
+        Params:
+            dg = serialize function accumulator
+
+    ***************************************************************************/
+
+    public void serialize (scope SerializeDg dg) const @safe
+    {
+        serializePart(this.data[0 .. this.length], dg);
+    }
```
Geod24: I see that the 1kb only applies to data in memory, not the one on disk. It's better, but still, why not simply use a dynamic array ?

Geod24 (changes requested):
Non-data transactions shouldn't be affected like they are in this PR.

source/agora/common/TransactionData.d
MichaelKim20: I think it's weird, too.
I misinterpreted the meaning of `empty fields` in your last proposal.
I'll fix it.
https://github.com/bpfkorea/agora/pull/1258#discussion_r507196744

source/agora/common/TransactionData.d
MichaelKim20: If the data is larger than the maximum, this transaction will be ignored.
I did this to eliminate factors that could be a problem with the stability of the system due to the excessive waste of memory.

source/agora/common/TransactionData.d
MichaelKim20: This is to process hexadecimal strings when JSON serialized and deserialized.

source/agora/common/TransactionData.d
MichaelKim20: Now I know there's a way to use BitBlob right away, thank you for letting me know.

MichaelKim20: I tried to use BitBlob, but it's not working.
Because BitBlob only supports fixed sizes.

source/agora/common/TransactionData.d
AndrejMitrovic: But you don't have to do any of this. Just add a `ubyte[] payload` to the `Transaction`. Vibe.d already does all the serialization/deserialization.

source/agora/common/TransactionData.d
Geod24: Those are separate points: We do not want to have a fixed size, for the reasons mentioned, but if we did, this should have used `BitBlob` instead of copying all the code over. It's a basic case where [composition](https://en.wikipedia.org/wiki/Object_composition) could have been used.

source/agora/consensus/data/Transaction.d
```diff
@@ -268,3 +294,46 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
...
+        Fees payable to data storage
+
+*******************************************************************************/
+
+public Amount clculateDataFee(ulong data_size, ulong tx_fee, uint factor) @safe nothrow
+{
+    const double decimal = 100.0;
+    ulong fee = cast(ulong)((round((exp(cast(double)data_size / cast(double)factor) - 1.0) * decimal) / decimal) * 10_000_000L);
+    return Amount(fee + tx_fee);
+}
```
MichaelKim20: The formula is as follows:

I only dealt with fees for data storage except for normal transaction fees.

source/agora/consensus/data/Transaction.d
```diff
@@ -268,3 +294,46 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
...
+unittest
+{
+    import vibe.data.json;
+    Transaction old_tx = Transaction(
+        TxType.Data,
+        [Input(Hash.init, 0)],
+        [Output.init],
+        [1,2,3]
+    );
+    auto json_str = old_tx.serializeToJsonString();
```
MichaelKim20, line 288: I tried to express it in hexstring when expressed in JSON.
However, I finally decided that it was better to express it in an array in terms of performance.

source/agora/consensus/data/Transaction.d
```diff
@@ -53,6 +55,9 @@ public struct Transaction
     /// The list of newly created outputs to put in the UTXO
     public Output[] outputs;
 
+    /// The data to store
+    public ubyte[] data;
```
MichaelKim20: I was able to confirm that the deserializer works well even if I use ubyte[] without having to implement fromBinary by making a separate type.
Moreover, when printing as JSON characters, I finally implemented the code using ubyte[] because toString, fromString, is not necessary if JSON array is used without hexstring.

source/agora/consensus/data/Transaction.d
MichaelKim20: If we adjust the factor, we can change the fee appropriately.
I added factor to ConsensusParams.

source/agora/consensus/data/Params.d
```diff
@@ -54,6 +54,12 @@ public immutable class ConsensusParams
     /// The address of commons budget
     public PublicKey CommonsBudgetAddress;
 
+    /// The maximum size of transaction data
+    public uint MaxSizeTransactionData;
```
MichaelKim20, line 58: In Bitcoin, there is OP_RETURN.
The maximum value is 83 Bytes.
Bitcoin Cash has a maximum value of 200 Bytes

MichaelKim20: See) https://docs.cryptoapis.io/rest-apis/blockchain-as-a-service-apis/btc/index#btc-transactions-creating-transactions

source/agora/consensus/data/Transaction.d
```diff
@@ -247,3 +274,45 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
...
+public Amount clculateDataFee(ulong data_size, uint factor) @safe nothrow
+{
+    const double decimal = 100.0;
+    ulong fee = cast(ulong)((round((exp(cast(double)data_size / cast(double)factor) - 1.0) * decimal) / decimal) * 10_000_000L);
+    return Amount(fee);
+}
+
+unittest
+{
+    assert(clculateDataFee(0, 200) == Amount(0L));
```
TrustHenry, line 331: Are you going to allow data_size = 0?
Do you have any intentions?

MichaelKim20: When the data size is zero, the fee for data storage is zero.

source/agora/consensus/data/Transaction.d
```diff
@@ -247,3 +274,45 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
...
+    Params:
+        data_size = The size of the data
+        factor = The factor to calculate for the fee of transaction data
+
+    Return:
+        Fees payable to data storage
+
+*******************************************************************************/
+
+public Amount clculateDataFee(ulong data_size, uint factor) @safe nothrow
```
MichaelKim20, line 309: The formula is as follows:

I only dealt with fees for data storage except for normal transaction fees.







If we adjust the factor, we can change the fee appropriately.
I added factor to ConsensusParams.

source/agora/consensus/data/Transaction.d
MichaelKim20: A fee of 0.01 for a transaction will be applied once the future development direction is established.

MichaelKim20:
The `TransactionData` related PR was moved to #1320. I would appreciate it if you continue to review it there.

MichaelKim20:
I have moved the fee-related PR according to data storage to #1321 .
I would appreciate it if you continue to review it there.

PR #1262, opened 2020-10-16
Some prometheus metrics were missing

### LINKED ISSUE #1260: Some Prometheus metrics were missing
due to a bug

list:
  agora_validators_gauge

ferencdg:
list
   agora_validators_gauge

fixes: #1260

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -733,9 +733,7 @@ public class EnrollmentManager
     private void collectValidatorStats (Collector collector)
     {
         auto validator_count = validator_set.count();
-        if (!validator_count)
-            validator_count_stats.setMetricTo!"agora_validators_gauge"(
-                validator_count);
+        validator_count_stats.setMetricTo!"agora_validators_gauge"(validator_count);
         Hash[] keys;
         if (getEnrolledUTXOs(keys))
             foreach (const ref key; keys)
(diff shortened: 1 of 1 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -733,7 +733,7 @@ public class EnrollmentManager
     private void collectValidatorStats (Collector collector)
     {
         auto validator_count = validator_set.count();
-        if (!validator_count)
+        if (validator_count)
```
AndrejMitrovic: Actually I don't think you need the if check at all. There should never be zero active validators in the system.

ferencdg: zero can be database error, if you look into the count() function, I think it is better not to report in this case, it might be an unnecessary noise in the data

AndrejMitrovic: Yeah, we're gonna have to fix all of that later (w.r.t. ValidatorSet itself).

AndrejMitrovic (approved):

source/agora/consensus/EnrollmentManager.d
Geod24: Well, I think it's better to have a `0` than no data (no data can happen if the node is offline for example).
So I'd just remove the check as @AndrejMitrovic mentioned.

PR #1268, opened 2020-10-20
Add UTXO hash component in making pre-images

### LINKED ISSUE #849: Discussion: Should a node be able to enroll multiple times with the same address ?

linked0:
Our previous pre-image scheme relied only on the private key, a nonce, and a fixed string. But the scheme made it unable for a node to enroll with different frozen UTXOs because it made the same pre-images for the UTXOs.  So we have made it possible to enroll with different UTXOs by adding UXTO hash to our pre-image scheme.

Fixes #849

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -256,6 +256,9 @@ public class EnrollmentManager
         This also gets and stores the information about this node's enrollment
         information which could be lost in abnormal situations.
 
+        Note: When a new Enrollment is encountered that is to start before
+        the existing one runs out, this just warns the situation.
+
         Params:
             enroll = Enrollment structure to add to the validator set
             block_height = enrolled blockheight
@@ -278,6 +281,15 @@ public class EnrollmentManager
 
         if (enroll.utxo_key in self_utxos)
         {
+            if (this.enroll_key != Hash.init)
+            {
+                auto enrolled =
+                    this.validator_set.getEnrolledHeight(this.enroll_key);
+                log.warn("Since enrollment for a validator with the same key " ~
+                    "at block #{}! New enrollment: {}, current one: {} (since #{})",
+                    block_height, enroll.utxo_key, this.enroll_key, enrolled);
+            }
+
             this.setEnrollmentKey(enroll.utxo_key);
 
             // consume pre-images
@@ -308,9 +320,6 @@ public class EnrollmentManager
 
     public Enrollment createEnrollment (Hash utxo) @safe nothrow
     {
-        // K, frozen UTXO hash
-        this.enroll_key = utxo;
-
         // X, final seed data and preimages of hashes
         this.catchupPreImageCycle();
         const seed = this.cycle.populate(this.key_pair.v, utxo, false);
@@ -1027,30 +1036,11 @@ unittest
     assert(preimage.hash ==
         man.cycle.preimages[$ - 1 - man.PreimageRevealPeriod]);
 
+    // check the cycle count for enrollment
+    assert(man.getCycleIndex() == 2);
+
     // test for getting validators' UTXO keys
     Hash[] keys;
-
-    // validator A with the `utxo_hash` and the enrolled height of 10.
-    // validator B with the 'utxo_hash2' and the enrolled height of 11.
-    // validator C with the 'utxo_hash3' and no enrolled height.
-    assert(man.addValidator(enroll2, Height(11), &utxo_set.findUTXO, utxos) is null);
-    man.clearExpiredValidators(Height(11));
-    assert(man.validatorCount() == 2);
-    assert(man.getEnrolledUTXOs(keys));
-    assert(keys.length == 2);
-
-    // set an enrolled height for validator C
-    // set the block height to 1019, which means validator B is expired.
-    // there is only one validator in the middle of 1020th block being made.
-    assert(man.addValidator(enroll3, Height(1019), &utxo_set.findUTXO, utxos) is null);
-    man.clearExpiredValidators(Height(1019));
-    assert(man.validatorCount() == 1);
-    assert(man.getEnrolledUTXOs(keys));
-    assert(keys.length == 1);
-    assert(keys[0] == enroll3.utxo_key);
-
-    // check the cycle count for enrollment
-    assert(man.getCycleIndex() == 4);
 }
 
 /// Test for adding and getting pre-images
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -160,11 +160,6 @@ public class EnrollmentManager
         // load enrollment key
         this.enroll_key = this.getEnrollmentKey();
 
-        // load the count of 'populate' of `PreImageCycle`
-        const uint populate_count = this.getCycleIndex();
-        foreach (_; 0 .. populate_count)
-            this.cycle.populate(this.key_pair.v, true);
-
         Utils.getCollectorRegistry().addCollector(&this.collectValidatorStats);
     }
 
@@ -286,7 +281,8 @@ public class EnrollmentManager
             this.setEnrollmentKey(enroll.utxo_key);
 
             // consume pre-images
-            this.cycle.populate(this.key_pair.v, true);
+            this.catchupPreImageCycle();
+            this.cycle.populate(this.key_pair.v, enroll.utxo_key, true);
 
             // save the count of 'populate' of `PreImageCycle`
             this.setCycleIndex(
@@ -316,7 +312,8 @@ public class EnrollmentManager
         this.enroll_key = utxo;
 
         // X, final seed data and preimages of hashes
-        const seed = this.cycle.populate(this.key_pair.v, false);
+        this.catchupPreImageCycle();
+        const seed = this.cycle.populate(this.key_pair.v, utxo, false);
         const enroll = makeEnrollment(
             this.key_pair, utxo, this.params.ValidatorCycle,
             seed, this.cycle.index);
@@ -381,7 +378,7 @@ public class EnrollmentManager
         // Generate the random seed to use
         auto cache = PreImageCache(PreImageCycle.NumberOfCycles, cycle_length);
         assert(offset < cache.length);
-        cache.reset(hashMulti(kp.v, "consensus.preimages", offset));
+        cache.reset(hashMulti(kp.v, utxo, "consensus.preimages", offset));
 
         return makeEnrollment(kp, utxo, cycle_length, cache[$ - offset - 1], offset);
     }
@@ -804,6 +801,25 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Catch up on the cache of pre-images to the current index
+
+        A node needs to catch up on the cache of pre-images to the current
+        index once after it starts.
+
+    ***************************************************************************/
+
+    public void catchupPreImageCycle () @safe nothrow
+    {
+        auto current_cycle = this.getCycleIndex();
+        // If there is any other case diffrent from the following assertion,
+        // it must be that an abnormal situation happened
+        assert(this.cycle.index == 0 || this.cycle.index == current_cycle);
+        while (this.cycle.index < current_cycle)
+            this.cycle.populate(this.key_pair.v, this.enroll_key, true);
+    }
+
+    /***************************************************************************
+
         Update the distance of the pre-image that will be revealed next time
 
         Params:
@@ -1082,6 +1098,7 @@ unittest
 
     auto secret = Scalar.random();
     Scalar fake_secret; // Used whenever `secret` *shouldn't* be used
+    auto utxo = hashFull(Scalar.random());
     foreach (uint cycleGroupCount; 0 .. 10)
     {
         foreach (outerIndex; 1 .. PreImageCycle.NumberOfCycles + 1)
@@ -1091,7 +1108,7 @@ unittest
             assert(outerIndex - 1  == cycle.index);
             // Only provide `secret` on the first iteration
             const commitment =
-                cycle.populate(outerIndex == 1 ? secret : fake_secret, true);
+                cycle.populate(outerIndex == 1 ? secret : fake_secret, utxo, true);
             const lastInCycle = outerIndex == PreImageCycle.NumberOfCycles;
             // Sanity check #2
             assert(cycleGroupCount + lastInCycle == cycle.nonce);
@@ -1124,7 +1141,7 @@ unittest
             {
                 const Index = seedIndex % Enrollment.ValidatorCycle;
                 if (Index == (Enrollment.ValidatorCycle - 1))
-                    cycle.populate(seedIndex == cycle.seeds.length ? secret :fake_secret);
+                    cycle.populate(seedIndex == cycle.seeds.length ? secret :fake_secret, utxo);
                 assert(cycle.seeds[seedIndex] == cycle.preimages[Index]);
                 if (seedIndex == 0) break;
                 seedIndex--;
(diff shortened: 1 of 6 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -660,6 +666,58 @@ public class EnrollmentManager
...
+
+            if (!results.empty)
+                en_key = results.oneValue!(ubyte[]).deserializeFull!(Hash);
+        }
+        catch (Exception ex)
+        {
+            log.error("ManagedDatabase operation error: {}", ex);
+        }
+
+        return en_key;
```
AndrejMitrovic: A slight simplification to make it obvious when `Hash.init` is returned.

```suggestion
        try
        {
            auto results = this.db.execute("SELECT val FROM node_enroll_data " ~
                "WHERE key = ?", "enroll_key");

            if (!results.empty)
                return results.oneValue!(ubyte[]).deserializeFull!(Hash);
        }
        catch (Exception ex)
        {
            log.error("ManagedDatabase operation error: {}", ex);
        }

        return Hash.init;
```

AndrejMitrovic (changes requested):
LGTM, just one thing.

linked0:
I applied Drey's requests. Thanks.
But CI builds have some errors on all the `ldc-master` environments about `ocean` as follows. If anyone knows how to deal with it, please let me know.

```
/home/runner/work/agora/agora/submodules/ocean/src/ocean/core/array/Search.d:73: error: undefined reference to '_D5ocean4core5array17DefaultPredicatesQt__T7IsEqualTaZQl6opCallFNaNbNiNfIaIaZb'
/home/runner/work/agora/agora/submodules/ocean/src/ocean/core/array/Transformation.d:71: error: undefined reference to '_D5ocean4core5array17DefaultPredicatesQt__T6IsLessTaZQk6opCallFNaNbNiNfIaIaZb'
/home/runner/work/agora/agora/submodules/ocean/src/ocean/core/array/Transformation.d:73: error: undefined reference to '_D5ocean4core5array17DefaultPredicatesQt__T6IsLessTaZQk6opCallFNaNbNiNfIaIaZb'
/home/runner/work/agora/agora/submodules/ocean/src/ocean/core/array/Transformation.d:123: error: undefined reference to '_D5ocean4core5array17DefaultPredicatesQt__T6IsLessTaZQk6opCallFNaNbNiNfIaIaZb'
```

AndrejMitrovic:
It's not a bug with your PR. It seems latest ldc-master is broken: https://github.com/bpfkorea/agora/runs/1279452560

linked0:
You can review my PR, all the builds on ldc-master have failed though.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -160,6 +160,9 @@ public class EnrollmentManager
         // create Pair object from KeyPair object
         this.key_pair = Pair.fromScalar(secretKeyToCurveScalar(key_pair.secret));
 
+        // load enrollment key
+        this.enroll_key = this.getEnrollmentKey();
+
```
Geod24: This value is never read. Currently a validator wishing to enroll will call the enrollment manager with the UTXO key. Do we need to read it here ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -168,8 +168,11 @@ public class EnrollmentManager
 
         // load the count of 'populate' of `PreImageCycle`
         const uint populate_count = this.getCycleIndex();
+
+        // if there exists an erollment key, `populate_count` must be above 0.
+        assert(this.enroll_key == Hash.init || populate_count > 0);
         foreach (_; 0 .. populate_count)
-            this.cycle.populate(this.key_pair.v, true);
+            this.cycle.populate(this.key_pair.v, this.enroll_key, true);
```
Geod24: I see now why you are loading it in the  constructor. I think we should rather move this outside of the constructor, so pre-images only get generated when needed (and possible). Otherwise the first time a validator runs, tons of pre-images will be generated for nothing.

source/agora/consensus/PreImage.d
```diff
@@ -351,14 +352,14 @@ public struct PreImageCycle
 
     ***************************************************************************/
 
-    public Hash populate (scope const ref Scalar secret, bool consume)
-        @safe nothrow
+    public Hash populate (scope const ref Scalar secret,
+        scope const ref Hash enroll_key, bool consume) @safe nothrow
```
Geod24, line 356: Adding the `enroll_key` changes quite a bit. When we do see another validator with the same private key, we don't want to override our own pre-images with their own.

Geod24 (changes requested):
This needs a network test with two nodes.

source/agora/consensus/EnrollmentManager.d
linked0: It would be better to make pre-images when needed. I will find the right place to do it.

linked0:
I added a network test code in the module, agora.test.PreImage. But, it's not the final code. I just uploaded the code for sharing what I'm trying to do.
@Geod24 I'd like you to check if it's the right approach. In the code, the `SameKeyValidator` is using the `WK.Keys.A` and the second UTXO for the key in order to make an Enrollment.

linked0:
I'm writing a network test in this PR for a new validator with the same address as the existing validator. But, I'm not sure it's the right approach. Some whiteboarding session is needed I guess. With the last commit about network test, I have some assertion failure as follows. I'm debugging it now.

```
(lldb) bt
error: need to add support for DW_TAG_base_type 'immutable(char)' encoded with DW_ATE = 0x10, bit_size = 8
* thread #64, stop reason = signal SIGABRT
  * frame #0: 0x00007fff6a52d33a libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0x00007fff6a5e9e60 libsystem_pthread.dylib`pthread_kill + 430
    frame #2: 0x00007fff6a4b4808 libsystem_c.dylib`abort + 120
    frame #3: 0x00007fff67714458 libc++abi.dylib`abort_message + 231
    frame #4: 0x00007fff677058a7 libc++abi.dylib`demangling_terminate_handler() + 238
    frame #5: 0x00007fff692405b1 libobjc.A.dylib`_objc_terminate() + 104
    frame #6: 0x00007fff67713887 libc++abi.dylib`std::__terminate(void (*)()) + 8
    frame #7: 0x00007fff67716387 libc++abi.dylib`__cxa_rethrow + 99
    frame #8: 0x000000010073f0ac agora-unittests`stellar::Slot::processEnvelope(this=0x000000010c7182d8, envelope=0x0000000240412d30, self=false) at Slot.cpp:155:9
    frame #9: 0x000000010072e4a0 agora-unittests`stellar::SCP::receiveEnvelope(this=0x0000000104734aa0, envelope=0x0000000240412d30) at SCP.cpp:33:38
    frame #10: 0x0000000100142933 agora-unittests`_D5agora9consensus8protocol9NominatorQk15receiveEnvelopeMFNeMKxS4scpd5types11Stellar_SCP11SCPEnvelopeZv(this=0x00000001d901f2c0, envelope=0x0000000240412d30) at Nominator.d:444:9
    frame #11: 0x00000001001443a1 agora-unittests`_D5agora4node9ValidatorQk15receiveEnvelopeMFNfS4scpd5types11Stellar_SCP11SCPEnvelopeZv(this=0x000070000608a648, envelope=SCPEnvelope @ 0x0000000240412d30) at Validator.d:266:9
    frame #12: 0x000000010019670f agora-unittests`_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPITSQBaQxQu10SerializerZQCg13handleCommandFSQDuQDq7CommandQDdSQEmQEi9FilterAPIZv(cmd=Command @ 0x0000000240413ec0, node=0x000070000608a900, filter=FilterAPI @ 0x0000000240413ea0) at LocalRest.d-mixin-477:494:33
    frame #13: 0x00000001001aab53 agora-unittests`_D6geod249LocalRest__T9RemoteAPITC5agora4test4Base7TestAPITSQBaQxQu10SerializerZQCg__T7spawnedTCQCkQChQCf17TestValidatorNodeZQBnFNbSQFa11concurrency3TidAyaiSQEt6common6ConfigQhPOSQGv8RegistryQjAySQGg9consensus4data5BlockQgmPOlZ7runNodeMFZ__T6handleTSQJoQJk7CommandZQyMFQuZ9__lambda2MFZv at LocalRest.d:602:43
    frame #14: 0x00000001007e58f9 agora-unittests`_D6geod2411concurrency14FiberScheduler6createMFNbDFZvZ4wrapMFZv at concurrency.d:915:13
    frame #15: 0x0000000100d4551f agora-unittests`_D4core6thread7context8Callable6opCallMFZv(this=0x0000000000000002) at context.d:46:17
    frame #16: 0x0000000100d456ae agora-unittests`_D4core6thread5fiber5Fiber3runMFZv(this=0x00000001d951e680) at fiber.d:1129:9
    frame #17: 0x0000000100d455e3 agora-unittests`fiber_entryPoint at fiber.d:186:13
```

linked0:
Ready for review.
The network test in the last commit should have been more complete. But, [the problem](https://github.com/bpfkorea/agora/pull/1268#issuecomment-716384194) could not be easily solved. If a validator re-enroll with another UTXO before its current enrollment expires, the problem occurs. So it would be good to handle it as a separate issue because it would take some time for me to analyze and get better knowledge about that. I'll create another issue for that if possible.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1367,3 +1432,20 @@ unittest
...
+    auto utxo_set = new TestUTXOSet;
+    genesisSpendable()
+        .map!(txb => txb.refund(WK.Keys.A.address).sign(TxType.Freeze))
+        .each!(tx => utxo_set.put(tx));
+    auto man = new EnrollmentManager(":memory:", WK.Keys.A,
+        new immutable(ConsensusParams)(10));
+
+    man.setEnrollmentKey(utxo_set.keys[0]);
+    assert(man.getEnrollmentKey()[] == utxo_set.keys[0][]);
+    assert(man.getEnrollmentKey()[] != utxo_set.keys[1][]);
```
Geod24: I would add an `assert` on the length of the UTXO set (should be 8) before L1448, and use a different value (e.g. `keys[3]`), so you make sure it doesn't pick the first one in the `keys` for whatever reason.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -664,6 +667,68 @@ public class EnrollmentManager
         return false;
     }
 
+    /***************************************************************************
+
+        Get the key for the enrollment for this node
+
+        Returns:
+            the key for the enrollment
```
Geod24: Note that this function can "fail silently". No exception being thrown, and only a log message.
Originally I thought it was a bug, but it makes sense given the current design of the `EnrollmentManager`.
If that's intentional, it should be documented, and also the `log.error` should be downgraded to a non-error.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -664,6 +667,68 @@ public class EnrollmentManager
...
+
+        return Hash.init;
+    }
+
+    /***************************************************************************
+
+        Set the key for the enrollment for this node
+
+        Params:
+            enroll_key = the key for the enrollment
```
Geod24: Ditto about failure documentation and logging.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -871,19 +799,22 @@ public class EnrollmentManager
         Params:
             height = block height to check
 
+        Returns:
+            The distance
```
Geod24: Or `uint.max` if we're not enrolled

source/agora/consensus/EnrollmentManager.d
```diff
@@ -278,15 +281,24 @@ public class EnrollmentManager
...
-                (this.cycle.nonce * this.cycle.NumberOfCycles) + this.cycle.index);
+                // save the count of 'populate' of `PreImageCycle`
+                this.setCycleIndex(
+                    (this.cycle.nonce * this.cycle.NumberOfCycles)
+                        + this.cycle.index);
+            }
+            else
+            {
+                log.warn("New enrollment appears before the current one is " ~
+                    "expired, New: {}, Old: {}", enroll.utxo_key, this.enroll_key);
```
Geod24: Rephrasing: "Seen enrollment for a validator with the same private key! New enrollment: {}, current one: {}".
Adding block height would be even better, if possible:
"Seen enrollment for a validator with the same private key at block {}! New enrollment: {}, current one: {} (since #{})"

source/agora/consensus/EnrollmentManager.d
```diff
@@ -278,15 +281,24 @@ public class EnrollmentManager
...
 
-            // save the count of 'populate' of `PreImageCycle`
-            this.setCycleIndex(
-                (this.cycle.nonce * this.cycle.NumberOfCycles) + this.cycle.index);
+                // save the count of 'populate' of `PreImageCycle`
+                this.setCycleIndex(
+                    (this.cycle.nonce * this.cycle.NumberOfCycles)
+                        + this.cycle.index);
+            }
+            else
```
Geod24: If you move it as the first check and `return` you can simplify the code / diff.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -309,7 +321,7 @@ public class EnrollmentManager
     public Enrollment createEnrollment (Hash utxo) @safe nothrow
     {
         // K, frozen UTXO hash
-        this.enroll_key = utxo;
+        // this.enroll_key = utxo;
```
Geod24: Remove then ?

source/agora/test/SameAddressPreImage.d
```diff
@@ -0,0 +1,198 @@
+module agora.test.SameAddressPreImage;
```
Geod24: DDOC

Geod24 (commented):
Nice. Looks like this is almost ready. The network test usage of shared leaves me a bit skeptical, and might be the cause of your issues. Let's look at it together tomorrow.

linked0:
> Nice. Looks like this is almost ready. The network test usage of shared leaves me a bit skeptical and might be the cause of your issues. Let's look at it together tomorrow.

Thank you for the review. I'll fix what you commented on.
In fact, I didn't add the code in the last commit that makes the problem happen. But I will add the code as another network test with a  `version (none)` modifier.

linked0:
Ready for review after applying reviewers' comments.
I also added the network test code that makes [the problem](https://github.com/bpfkorea/agora/pull/1268#issuecomment-716384194) happen with the `version(none)` modifier.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -664,6 +667,77 @@ public class EnrollmentManager
         return false;
     }
 
+    /***************************************************************************
+
+        Get the key for the enrollment for this node
+
+        If getting an enrollment key fails, this displays a log messge and
```
Geod24: message

source/agora/consensus/EnrollmentManager.d
```diff
@@ -664,6 +667,77 @@ public class EnrollmentManager
...
+        returns `Hash.init`. If the database operation fails, it means that node
+        is not in the normal situation overall.
```
Geod24: "the node is not in the normal situation" is not quite a good explanation. If we detect and error situation, we usually try to remedy it. In this code, I was assuming that the failure was "normal" in case the UTXO wasn't propagated yet. But that doesn't seem to be the case ?

Geod24: Looking at the code again, when the node is not enrolled, I assume the `SELECT` will return an empty result, so no exception will be thrown ? If so that should be documented.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -664,6 +667,77 @@ public class EnrollmentManager
...
+        }
+
+        return Hash.init;
+    }
+
+    /***************************************************************************
+
+        Set the key for the enrollment for this node
+
+        If saving an enrollment key fails, this displays a log messges and just
```
Geod24: Ditto typo

Geod24:
Doesn't the third commit invalidate the test Genesis block ? It would seem so.
I extracted the first 2 commits here: https://github.com/bpfkorea/agora/pull/1303
Will look into the network test failure and the third commit.

Geod24:
Partially rebased on v0.x.x, added the changes to the test genesis block, likely the reason your network test didn't pass.
I'll continue working on this.

Geod24:
Okay, this actually uncovers TONS of problems. We generally identify validators by their public key, but since public keys are not unique, that creates issues. For example, the network registry relies on it.

I think we need to whiteboard (as a team) on this.

AndrejMitrovic:
> but since public keys are not unique

You left us on a cliffhanger here. :D

Issue #1227, opened 2020-09-21
Build a tool (or extend agora-cli) to simulate activity on an Agora network


AndrejMitrovic:
We want to start testing Agora in an environment which simulates a real-world network. You may think of this as TestNet, however this system of nodes does **not need** to be public - therefore it does not need to be TestNet. It could be any network with a custom-defined genesis block. For example, it could be a network of Agora nodes running only on the developer's computer for just a short period of time. Or it could run on a dedicated cloud service that only we have access to. 

Exactly where the network is running **is not important** for this issue.

Currently, we have support for:
- Starting a new Agora node (docker)
- Creating transactions and sending them to an Agora node (agora-cli #206)

In order to fully simulate a network we also need some additional features:

- An ability to continuously generate new valid transactions and send these to one (or more) Agora nodes
- An ability to purposefully DDoS an Agora node with well-formed API requests. These requests contain parameters which are valid (e.g. valid transactions)
- An ability to purposefully DDoS an Agora node with malformed API requests. This will require the use of randomness. For example, the tool could try to flip random bits or send garbage data to an API end-point. The tool could also try to call into end-points which don't exist. An Agora node should handle all these cases properly.

It may be possible to extend `agora-cli` for this purpose, or it may be necessary to build a new tool.

The objective of this tool is to:
- Find any hidden bugs that could cause an Agora node to become faulty. It's important to catch as many as these bugs before CoinNet is deployed.
- Find performance bottlenecks, and thereafter optimize Agora.

As a developer your goal is to:
- Find a good set of features that this tool should support, and the use-cases for such a feature. Some of them are listed above.
- Implement those features. If the scope of implementing the features is too large, feel free to break this down into several smaller issues.

@kchulj would you consider this to be an interesting feature to work on?

@Geod24 please share any ideas about the features we may need, or if the scope of the requirements should be re-adjusted.

Geod24:
There was https://github.com/bpfkorea/agora/pull/1234 merged and worked on this sprint.
Currently the feature set is quite limited, and it is not deployed. Also, the current design (a CLI tool) makes it hard to work with, as such a tool needs knowledge of the UTXO. This is an issue @kchulj is aware of (we discussed it) and I believe he's working on it, however I don't think we can consider this closed.

Geod24:
The basic work is done. The rest should be split into smaller issues in bpfkorea/faucet.

PR #1285, opened 2020-10-26
[WiP] Implement server-based tool

kchulj:
Starting point for the server application tool.
The tool is currently able to send transactions periodically and keep track of its own `UTXOSet`.

Part of #1227

### DIFF
--- a/source/agora/consensus/UTXOSet.d
+++ b/source/agora/consensus/UTXOSet.d
@@ -31,7 +31,7 @@ mixin AddLogger!();
 public class UTXOSet
 {
     /// Utxo cache backed by a database
-    private UTXODB utxo_db;
+    protected UTXODB utxo_db;
 
     /// Keeps track of spent outputs during the validation of a Tx / Block
     private Set!Hash used_utxos;
@@ -216,7 +216,7 @@ public class UTXOSet
 private class UTXODB
 {
     /// SQLite db instance
-    private ManagedDatabase db;
+    public ManagedDatabase db;
 
 
     /***************************************************************************
--- a/source/agora/tool/main.d
+++ b/source/agora/tool/main.d
@@ -1,0 +1,149 @@
+/+
+ dub.sdl:
+ name "sometest"
+ dependency "vibe-d" version="~>0.8"
++/
+
+import agora.common.Types;
+import agora.api.FullNode;
+import agora.common.crypto.Key;
+import agora.common.Serializer;
+import agora.consensus.data.genesis.Test;
+import agora.consensus.data.Transaction;
+import agora.consensus.data.UTXOSetValue;
+import agora.consensus.UTXOSet;
+import agora.utils.Test;
+
+import std.algorithm;
+import std.random;
+import std.range;
+import std.stdio;
+
+import core.time;
+
+import vibe.core.core;
+import vibe.core.log;
+import vibe.web.rest;
+
+final class MyUTXOSet : UTXOSet
+{
+    public this ()
+    {
+        super(":memory:");
+    }
+
+    public int opApply (scope int delegate (const Hash, const UTXOSetValue) dg)
+    {
+        auto results = this.utxo_db.db.execute("SELECT key, val FROM utxo_map");
+        foreach (row; results)
+        {
+            const key = deserializeFull!Hash(row.peek!(ubyte[])(0));
+            const value = deserializeFull!UTXOSetValue(row.peek!(ubyte[])(1));
+            if (auto ret = dg(key, value))
+                return ret;
+        }
+        return 0;
+    }
+}
+
+public int sendTransaction (string address, uint take)
+{
+    auto node = new RestInterfaceClient!API(address);
+    logInfo("The address of node is %s", address);
+    const height = node.getBlockHeight();
+    logInfo("The initial block height is: %s", height);
+    const blocks = node.getBlocksFrom(0, cast(uint) (height + 1));
+    scope utxos = new MyUTXOSet();
+
+    foreach (ref b; blocks)
+        foreach (ref tx; b.txs)
+            if (tx.type == TxType.Payment)
+                utxos.updateUTXOCache(tx, Height(0));
+
+    const utxo_len = utxos.length();
+    logInfo("Populated %s UTXOs", utxo_len);
+
+    // If there are less than 50 UTXOs
+    // then print the current UTXO set
+    if (utxo_len < 50)
+        foreach (key, utxo; utxos)
+            logInfo("UTXO: [%s] %s", key, utxo);
+
+    immutable size_t WKKeysCount = 1378;
+
+    auto txs = blocks[$ - 1].spendable().map!(txb => txb.split(
+                    WK.Keys.byRange().drop(uniform(0, WKKeysCount - take, rndGen))
+                    .take(take).map!(k => k.address)).sign()).array();
+
+    void send ()
+    {
+        foreach (tx; txs)
+        {
+            node.putTransaction(tx);
+            logInfo("Transaction sent: %s", tx);
+        }
+
+        txs = txs.map!(txb => TxBuilder(txb).split(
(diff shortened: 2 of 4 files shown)


source/agora/consensus/UTXOSet.d
```diff
@@ -216,7 +216,7 @@ public class UTXOSet
 private class UTXODB
 {
     /// SQLite db instance
-    private ManagedDatabase db;
+    public ManagedDatabase db;
```
Geod24, line 219: This was just a temporary measure, we need to do this in a principled way ;)

source/agora/consensus/UTXOSet.d
```diff
@@ -31,7 +31,7 @@ mixin AddLogger!();
 public class UTXOSet
 {
     /// Utxo cache backed by a database
-    private UTXODB utxo_db;
+    protected UTXODB utxo_db;
```
Geod24, line 34: This is okay, but in its own commit ;)

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
+/+
+ dub.sdl:
+ name "sometest"
+ dependency "vibe-d" version="~>0.8"
++/
+
```
Geod24, line 6: @AndrejMitrovic : I was thinking of moving this to its own repository, using submodules for the Agora dependency, and I think everything would *just work*. There would be a need to update the Agora submodule from time to time but it wouldn't put any pressure on Agora development. Thoughts ?

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+import agora.consensus.UTXOSet;
+import agora.utils.Test;
+
+import std.algorithm;
+import std.random;
+import std.range;
+import std.stdio;
+
+import vibe.core.core;
+import vibe.d;
```
Geod24: Too wide import here

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+        auto results = this.utxo_db.db.execute("SELECT key, val FROM utxo_map");
+        foreach (row; results)
+        {
+            const key = deserializeFull!Hash(row.peek!(ubyte[])(0));
+            const value = deserializeFull!UTXOSetValue(row.peek!(ubyte[])(1));
+            if (auto ret = dg(key, value))
+                return ret;
+        }
+        return 0;
+    }
```
Geod24, line 46: How do we handle this ? We could just move it to the superclass (removing the need to inherit). The current design with the DB being separated doesn't quite fulfill  the open/close principle.

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+}
+
+public int sendTransaction (API delegate (string address) api_maker)
```
Geod24: The only reason you would need this is if you were to write a test suite. But even if you did, you would most likely pass an `API` instance, not a delegate, so I'd just simplify this for the time being.

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+{
+    string address = "http://na-002.bosagora.io:7826";
```
Geod24: Make this a parameter

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+    auto node = api_maker(address);
+    logInfo("The address is %s", address);
+    const height = node.getBlockHeight();
+    logInfo("Node's block height: %s", height);
```
Geod24: Nice, we need to provide users with info on the current state of the chain 👍
I would however change the phrasing to something like "Initial block height for node is:" so that it's easier to grep if you later add more logging.

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+        foreach (ref tx; b.txs)
+            if (tx.type == TxType.Payment)
+                utxos.updateUTXOCache(tx, Height(0));
+
+    const utxo_len = utxos.length();
+    logInfo("Populated %s UTXOs", utxo_len);
+
+    if (utxo_len < 42)
+        foreach (key, utxo; utxos)
+            logInfo("UTXO: [%s] %s", key, utxo);
```
Geod24: Maybe pick a slightly different value... And add a comment on what we're doing here, it's not quite obvious for people  that didn't write the code :)

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+
+    auto txs = blocks[$ - 1].spendable().map!(txb => txb.split(
+                   WK.Keys.byRange().drop(uniform(0, 1378 - 501, rndGen))
```
Geod24: Likewise, use named constant here. Something as simple as:
```D
/// Number of well known keys (returned by `WK.Keys.byRange()`)
immutable size_t WKKeysCount = 1378;
```
at the top is enough.

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+                   .take(5).map!(k => k.address)).sign()).array();
```
Geod24: If you only take `5`, then uniform should be `WKKeysCount -  5`. Another named constant might be warranted.

source/agora/tool/main.d
```diff
@@ -0,0 +1,98 @@
...
+
+    void send ()
+    {
+        foreach(tx; txs)
```
Geod24:
```suggestion
        foreach (const ref tx; txs)
```

source/agora/utils/Test.d
```diff
@@ -377,6 +385,15 @@ public struct TxBuilder
...
 
+    /// Ditto
+    public ref typeof(this) attach (const Output tx, Hash utxo_hash)
+        @safe pure nothrow return
+    {
+        this.inputs ~= OutputRef(Transaction.init, 1, utxo_hash, tx);
+        this.leftover.value.mustAdd(tx.value);
+        return this;
+    }
+
```
Geod24, line 396: Those were hacks that needs to be fixed :)

Geod24 (changes requested):
Good start

AndrejMitrovic:
The PR title should be more informative. What does the tool do? It should be in the title.

Also "tool" is not a good name for this tool. It's too generic.

source/agora/tool/main.d
AndrejMitrovic: Yes I like this too.

source/agora/consensus/UTXOSet.d
Geod24: Use `agora.consensus.data.UTXOSetValue : TestUTXOSet` instead

source/agora/tool/main.d
```diff
@@ -0,0 +1,149 @@
...
+        auto results = this.utxo_db.db.execute("SELECT key, val FROM utxo_map");
+        foreach (row; results)
+        {
+            const key = deserializeFull!Hash(row.peek!(ubyte[])(0));
+            const value = deserializeFull!UTXOSetValue(row.peek!(ubyte[])(1));
+            if (auto ret = dg(key, value))
+                return ret;
+        }
+        return 0;
+    }
```
Geod24, line 46: Ditto remove

source/agora/tool/main.d
```diff
@@ -0,0 +1,149 @@
...
+    logInfo("The address of node is %s", address);
+    const height = node.getBlockHeight();
+    logInfo("The initial block height is: %s", height);
+    const blocks = node.getBlocksFrom(0, cast(uint) (height + 1));
+    scope utxos = new MyUTXOSet();
+
+    foreach (ref b; blocks)
+        foreach (ref tx; b.txs)
+            if (tx.type == TxType.Payment)
+                utxos.updateUTXOCache(tx, Height(0));
```
Geod24, line 61: Factor into a function

Geod24: And use correct `Height`

Geod24:
We're moving to bpfkorea/faucet .
Also this depends on #1286

kchulj:
Will close this and make another PR at `faucet`.

PR #1291, opened 2020-10-28
Implement a script execution engine

### LINKED ISSUE #237: Basic script execution engine
We need to come up with a set of primitives for our script execution that will satisfy the needs of Flash.
Obviously, that means we need to come up with a list of primitives required for Flash to work.

Related (sub) issues:
- #243 
- #242
- #245

In the future, we might want to split this issue into smaller, more granular tasks once the initial research is completed.

AndrejMitrovic:
I've loosely based it on the Bitcoin execution engine, adding code attribution for any borrowed code.

Unlike Bitcoin, we do not have to support historic opcodes such as `P2SH` (see [isPayToScriptHash](https://github.com/bitcoin/bitcoin/blob/55b1ffcd259cc64139dad646a52fab5c597c95c2/src/script/script.cpp#L201-L208)), but may instead replace it with something smaller by introducing a locking script tag.

Missing features:
- Time locks (todo later)
- SIGHASH support (needed for Eltoo as well as advanced scripting)
- Integer arithmetic. We may want to add OP_ADD, etc, however inadvertently this brings in platform-specific behavior and needs lots of care. E.g. signed vs unsigned.
- Integration into Agora

There will be more work put into this, but I'll likely do it in another branch.

Part of: #237

### DIFF
--- a/source/agora/script/Engine.d
+++ b/source/agora/script/Engine.d
@@ -1,0 +1,1404 @@
+/*******************************************************************************
+
+    Contains the script execution engine.
+
+    Note that Bitcoin-style P2SH scripts are not detected,
+    instead one should use LockType.Redeem in the Lock script tag.
+
+    Things not currently implemented:
+        - opcode weight calculation
+        - opcode total cost limit
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.script.Engine;
+
+import agora.common.crypto.ECC;
+import Schnorr = agora.common.crypto.Schnorr;
+import agora.common.Hash;
+import agora.consensus.data.Transaction;
+import agora.script.Lock;
+import agora.script.Opcodes;
+import agora.script.ScopeCondition;
+import agora.script.Script;
+import agora.script.Stack;
+
+import std.bitmanip;
+import std.conv;
+import std.range;
+import std.traits;
+
+version (unittest)
+{
+    import agora.common.crypto.ECC;
+    import agora.common.crypto.Schnorr;
+    import agora.common.Hash;
+    import agora.utils.Test;
+    import ocean.core.Test;
+    import std.stdio : writefln, writeln;  // avoid importing LockType
+}
+
+/// Ditto
+public class Engine
+{
+    /// Opcodes cannot be pushed on the stack. We use a byte array as a marker.
+    /// Conditional opcodes require the top item on the stack to be one of these
+    private static immutable ubyte[1] TrueValue = [OP.TRUE];
+    /// Ditto
+    private static immutable ubyte[1] FalseValue = [OP.FALSE];
+
+    /// Maximum total stack size
+    private immutable ulong StackMaxTotalSize;
+
+    /// Maximum size of an item on the stack
+    private immutable ulong StackMaxItemSize;
+
+    /***************************************************************************
+
+        Initializes the script execution engine with the configured consensus
+        limits.
+
+        Params:
+            StackMaxTotalSize = the maximum allowed stack size before a
+                stack overflow, which would cause the script execution to fail.
+                the script execution fails.
+            StackMaxItemSize = maximum allowed size for a single item on
+                the stack. If exceeded, script execution will fail during the
+                syntactical validation of the script.
+
+    ***************************************************************************/
+
+    public this (ulong StackMaxTotalSize, ulong StackMaxItemSize)
+    {
+        assert(StackMaxItemSize > 0 && StackMaxTotalSize >= StackMaxItemSize);
+        this.StackMaxTotalSize = StackMaxTotalSize;
+        this.StackMaxItemSize = StackMaxItemSize;
+    }
+
+    /***************************************************************************
+
+        Main dispatch execution routine.
+
+        The lock type will be examined, and based on its type execution will
+        proceed to either simple script-less payments, or script-based payments.
+
+        Params:
+            lock = the lock
+            unlock = may contain a `signature`, `signature, key`,
+                     or `script` which only contains stack push opcodes
+            tx = the spending transaction
+            input = the input which contained the unlock
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    public string execute (in Lock lock, in Unlock unlock, in Transaction tx,
+        in Input input) nothrow @safe
+    {
+        if (lock.bytes.length == 0)
+            return "Lock cannot be empty";
+
+        final switch (lock.type)
+        {
+        case LockType.Key:
+        case LockType.KeyHash:
+            if (auto error = this.handleBasicPayment(lock, unlock, tx))
+                return error;
+            break;
+
+        case LockType.Script:
+            if (auto error = this.executeBasicScripts(lock, unlock, tx, input))
+                return error;
+            break;
+
+        case LockType.Redeem:
+            if (auto error = this.executeRedeemScripts(lock, unlock, tx, input))
+                return error;
+            break;
+        }
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Handle stack-less and script-less basic payments.
+
+        If the lock is a `Lock.Key` type, the unlock must only
+        contain a `signature`.
+        If the lock is a `Lock.KeyHash` type, the unlock must contain a
+        `signature, key` tuple.
+
+        Params:
+            lock = must contain a `pubkey` or a `hash`
+            unlock = must contain a `signature` or `signature, key` tuple
+            tx = the spending transaction
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    private string handleBasicPayment (in Lock lock, in Unlock unlock,
+        in Transaction tx) nothrow @safe
+    {
+        // assumed sizes
+        static assert(Point.sizeof == 32);
+        static assert(Hash.sizeof == 64);
+
+        switch (lock.type)
+        {
+        case LockType.Key:
+            if (lock.bytes.length != Point.sizeof)
+                return "LockType.Key requires 32-byte key argument in the lock script";
+            const Point key = Point(lock.bytes);
+            if (!key.isValid())
+                return "LockType.Key 32-byte public key in lock script is invalid";
+
+            if (unlock.bytes.length != Signature.sizeof)
+                return "LockType.Key requires a 64-byte signature in the unlock script";
+            const sig = Signature(unlock.bytes);
+            if (!Schnorr.verify(key, sig, tx))
+                return "LockType.Key signature in unlock script failed validation";
+
+            break;
+
+        case LockType.KeyHash:
+            if (lock.bytes.length != Hash.sizeof)
+                return "LockType.KeyHash requires a 64-byte key hash argument in the lock script";
+            const Hash key_hash = Hash(lock.bytes);
+
+            const(ubyte)[] bytes = unlock.bytes;
+            if (bytes.length != Signature.sizeof + Point.sizeof)
+                return "LockType.KeyHash requires a 64-byte signature "
+                     ~ "and a 32-byte key in the unlock script";
+            const sig = Signature(bytes[0 .. Signature.sizeof]);
+            bytes.popFrontN(Signature.sizeof);
+
+            const Point key = Point(bytes);
+            if (!key.isValid())
+                return "LockType.KeyHash public key in unlock script is invalid";
+
+            if (hashFull(key) != key_hash)
+                return "LockType.KeyHash hash of key does not match key hash set in lock script";
+
+            if (!Schnorr.verify(key, sig, tx))
+                return "LockType.KeyHash signature in unlock script failed validation";
+
+            break;
+
+        default:
+            assert(0);
+        }
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Execute a `LockType.Script` type of lock script with the associated
+        unlock script.
+
+        The unlock script may only contain stack pushes.
+        The unlock script is ran, producing a stack.
+        Thereafter, the lock script will run with the stack of the
+        unlock script.
+
+        For security reasons, the two scripts are not concatenated together
+        before execution. You may read more about it here:
+        https://bitcoin.stackexchange.com/q/80258/93682
+
+        Params:
+            lock = the lock script
+            unlock = the unlock script
+            tx = the spending transaction
+            input = the input which contained the unlock
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    private string executeBasicScripts (in Lock lock,
+        in Unlock unlock, in Transaction tx, in Input input) nothrow @safe
+    {
+        assert(lock.type == LockType.Script);
+
+        Script unlock_script;
+        if (auto error = validateScript(ScriptType.Unlock, unlock.bytes,
+            this.StackMaxItemSize, unlock_script))
+            return error;
+
+        Script lock_script;
+        if (auto error = validateScript(ScriptType.Lock, lock.bytes,
+            this.StackMaxItemSize, lock_script))
+            return error;
+
+        Stack stack = Stack(this.StackMaxTotalSize, this.StackMaxItemSize);
+        if (auto error = this.executeScript(unlock_script, stack, tx, input))
+            return error;
+
+        if (auto error = this.executeScript(lock_script, stack, tx, input))
+            return error;
+
+        if (this.hasScriptFailed(stack))
+            return "Script failed";
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Execute a `LockType.Redeem` type of lock script with the associated
+        lock script.
+
+        The 64-byte hash of the redeem script is read from `lock_bytes`,
+        `unlock_bytes` is evaluated as a set of pushes to the stack where
+        the last push is the redeem script. The redeem script is popped from the
+        stack, hashed, and compared to the previously extracted hash from the
+        lock script. If the hashes match, the redeem script is evaluated with
+        any leftover stack items of the unlock script.
+
+        Params:
+            lock = must contain a 64-byte hash of the redeem script
+            unlock = must contain only stack push opcodes, where the last
+                     push is the redeem script itself
+            tx = the associated spending transaction
+            input = the input which contained the unlock
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    private string executeRedeemScripts (in Lock lock, in Unlock unlock,
+        in Transaction tx, in Input input) nothrow @safe
+    {
+        assert(lock.type == LockType.Redeem);
+
+        if (lock.bytes.length != Hash.sizeof)
+            return "LockType.Redeem requires 64-byte script hash in the lock script";
+        const Hash script_hash = Hash(lock.bytes);
+
+        Script unlock_script;
+        if (auto error = validateScript(ScriptType.Unlock, unlock.bytes,
+            this.StackMaxItemSize, unlock_script))
+            return error;
+
+        Stack stack = Stack(this.StackMaxTotalSize, this.StackMaxItemSize);
+        if (auto error = this.executeScript(unlock_script, stack, tx, input))
+            return error;
+
+        if (stack.empty())
+            return "LockType.Redeem requires unlock script to push a redeem script to the stack";
+
+        const redeem_bytes = stack.pop();
+        if (hashFull(redeem_bytes) != script_hash)
+            return "LockType.Redeem unlock script pushed a redeem script "
+                 ~ "which does not match the redeem hash in the lock script";
+
+        Script redeem;
+        if (auto error = validateScript(ScriptType.Redeem, redeem_bytes,
+            this.StackMaxItemSize, redeem))
+            return error;
+
+        if (auto error = this.executeScript(redeem, stack, tx, input))
+            return error;
+
+        if (this.hasScriptFailed(stack))
+            return "Script failed";
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Execute the script with the given stack and the associated spending
+        transaction. This routine may be called for all types of scripts,
+        lock, unlock, and redeem scripts.
+
+        An empty script will not fail execution. It's up to the calling code
+        to differentiate when this is an allowed condition.
+
+        Params:
+            script = the script to execute
+            stack = the stack to use for the script. May be non-empty.
+            tx = the associated spending transaction
+            input = the input which contained the unlock
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    private string executeScript (in Script script, ref Stack stack,
+        in Transaction tx, in Input input) nothrow @safe
+    {
+        // tracks executable condition of scopes for use with IF / ELSE / etc
+        ScopeCondition sc;
+        const(ubyte)[] bytes = script[];
+        while (!bytes.empty())
+        {
+            OP opcode;
+            if (!bytes.front.toOPCode(opcode))
+                assert(0, "Script should have been syntactically validated");
+            bytes.popFront();
+
+            if (opcode.isConditional())
+            {
+                if (auto error = this.handleConditional(opcode, stack, sc))
+                    return error;
+                continue;
+            }
+
+            // must consume payload even if the scope is currently false
+            const(ubyte)[] payload;
+            switch (opcode)
+            {
+            case OP.PUSH_DATA_1:
+                if (auto reason = this.readPayload!(OP.PUSH_DATA_1)(
+                    bytes, payload))
+                    return reason;
+                break;
+
+            case OP.PUSH_DATA_2:
+                if (auto reason = this.readPayload!(OP.PUSH_DATA_2)(
+                    bytes, payload))
+                    return reason;
+                break;
+
+            case 1: .. case OP.PUSH_BYTES_75:
+                const payload_size = opcode;  // encoded in the opcode
+                if (bytes.length < payload_size)
+                    assert(0);  // should have been validated
+
+                payload = bytes[0 .. payload_size];
+                bytes.popFrontN(payload.length);
+                break;
+
+            default:
+                assert(!opcode.isPayload());  // missing cases
+                break;
+            }
+
+            // whether the current scope is executable
+            // (all preceeding outer conditionals were true)
+            if (!sc.isTrue())
+                continue;
+
+            switch (opcode)
+            {
+            case OP.TRUE:
+                if (!stack.canPush(TrueValue))
+                    return "Stack overflow while pushing TRUE to the stack";
+                stack.push(TrueValue);
+                break;
+
+            case OP.FALSE:
+                if (!stack.canPush(FalseValue))
+                    return "Stack overflow while pushing FALSE to the stack";
+                stack.push(FalseValue);
+                break;
+
+            case OP.PUSH_DATA_1:
+                if (!stack.canPush(payload))
+                    return "Stack overflow while executing PUSH_DATA_1";
+                stack.push(payload);
+                break;
+
+            case OP.PUSH_DATA_2:
+                if (!stack.canPush(payload))
+                    return "Stack overflow while executing PUSH_DATA_2";
+                stack.push(payload);
+                break;
+
+            case 1: .. case OP.PUSH_BYTES_75:
+                if (!stack.canPush(payload))
+                    return "Stack overflow while executing PUSH_BYTES_*";
+
+                stack.push(payload);
+                break;
+
+            case OP.DUP:
+                if (stack.empty)
+                    return "DUP opcode requires an item on the stack";
+
+                const top = stack.peek();
+                if (!stack.canPush(top))
+                    return "Stack overflow while executing DUP";
+                stack.push(top);
+                break;
+
+            case OP.HASH:
+                if (stack.empty)
+                    return "HASH opcode requires an item on the stack";
+
+                const ubyte[] top = stack.pop();
+                const Hash hash = hashFull(top);
+                if (!stack.canPush(hash[]))  // e.g. hash(1 byte) => 64 bytes
+                    return "Stack overflow while executing HASH";
+                stack.push(hash[]);
+                break;
+
+            case OP.CHECK_EQUAL:
+                if (stack.count() < 2)
+                    return "CHECK_EQUAL opcode requires two items on the stack";
+
+                const a = stack.pop();
+                const b = stack.pop();
+                stack.push(a == b ? TrueValue : FalseValue);  // canPush() check unnecessary
+                break;
+
+            case OP.VERIFY_EQUAL:
+                if (stack.count() < 2)
+                    return "VERIFY_EQUAL opcode requires two items on the stack";
+
+                const a = stack.pop();
+                const b = stack.pop();
+                if (a != b)
+                    return "VERIFY_EQUAL operation failed";
+                break;
+
+            case OP.CHECK_SIG:
+                bool is_valid;
+                if (auto error = this.verifySignature!(OP.CHECK_SIG)(
+                    stack, tx, is_valid))
+                    return error;
+
+                // canPush() check unnecessary
+                stack.push(is_valid ? TrueValue : FalseValue);
+                break;
+
+            case OP.VERIFY_SIG:
+                bool is_valid;
+                if (auto error = this.verifySignature!(OP.VERIFY_SIG)(
+                    stack, tx, is_valid))
+                    return error;
+
+                if (!is_valid)
+                    return "VERIFY_SIG signature failed validation";
+                break;
+
+            case OP.INVALID:
+                return "Script panic while executing OP.INVALID opcode";
+
+            default:
+                assert(0);  // should have been handled
+            }
+        }
+
+        if (!sc.empty())
+            return "IF / NOT_IF requires a closing END_IF";
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Handle a conditional opcode like `OP.IF` / `OP.ELSE` / etc.
+
+        The initial scope is implied to be true. When a new scope is entered
+        via `OP.IF` / `OP.NOT_IF`, the condition is checked. If the condition
+        is false, then all the code inside the `OP.IF` / `OP.NOT_IF`` block
+        will be skipped until we exit into the first scope where the condition
+        is true.
+
+        Execution will fail if there is an `OP.ELSE` or `OP.END_IF` opcode
+        without an associated `OP.IF` / `OP.NOT_IF` opcode.
+
+        Currently trailing `OP.ELSE` opcodes are not rejected.
+        This is also a quirk in the Bitcoin language, and should
+        be fixed here later.
+        (e.g. `IF { } ELSE {} ELSE {} ELSE {}` is allowed).
+
+        Params:
+            opcode = the current conditional
+            stack = the stack to evaluate for the conditional
+            sc = the scope condition which may be toggled by a condition change
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    private string handleConditional (in OP opcode,
+        ref Stack stack, ref ScopeCondition sc) nothrow @safe
+    {
+        switch (opcode)
+        {
+        case OP.IF:
+        case OP.NOT_IF:
+            if (!sc.isTrue())
+            {
+                sc.push(false);  // enter new scope, remain false
+                break;
+            }
+
+            if (stack.count() < 1)
+                return "IF/NOT_IF opcode requires an item on the stack";
+
+            const top = stack.pop();
+            if (top != TrueValue && top != FalseValue)
+                return "IF/NOT_IF may only be used with OP.TRUE / OP.FALSE values";
+
+            sc.push((opcode == OP.IF) ^ (top == FalseValue));
+            break;
+
+        case OP.ELSE:
+            if (sc.empty())
+                return "Cannot have an ELSE without an associated IF / NOT_IF";
+            sc.tryToggle();
+            break;
+
+        case OP.END_IF:
+            if (sc.empty())
+                return "Cannot have an END_IF without an associated IF / NOT_IF";
+            sc.pop();
+            break;
+
+        default:
+            assert(0);
+        }
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Checks if the script has failed execution by examining its stack.
+        The script is considered sucessfully executed only if its stack
+        contains exactly one item, and that item being `TrueValue`.
+
+        Params:
+            stack = the stack to check
+
+        Returns:
+            true if the script is considered to have failed execution
+
+    ***************************************************************************/
+
+    private bool hasScriptFailed (/*in*/ ref Stack stack) // peek() is not const
+        pure nothrow @safe
+    {
+        return stack.empty() || stack.peek() != TrueValue;
+    }
+
+    /***************************************************************************
+
+        Reads the length and payload of the associated `PUSH_DATA_*` opcode,
+        and advances the `opcodes` array to the next opcode.
+
+        The length is read in little endian format.
+
+        Params:
+            OP = the associated `PUSH_DATA_*` opcode
+            opcodes = the opcode / data byte array
+            payload = will contain the payload if successfull
+
+        Returns:
+            null if reading the payload was successfull,
+            otherwise the string explaining why it failed
+
+    ***************************************************************************/
+
+    private string readPayload (OP op)(ref const(ubyte)[] opcodes,
+        out const(ubyte)[] payload) nothrow @safe /*@nogc*/
+    {
+        static assert(op == OP.PUSH_DATA_1 || op == OP.PUSH_DATA_2);
+        alias T = Select!(op == OP.PUSH_DATA_1, ubyte, ushort);
+        if (opcodes.length < T.sizeof)
+            assert(0);  // script should have been validated
+
+        const T size = littleEndianToNative!T(opcodes[0 .. T.sizeof]);
+        if (size == 0 || size > this.StackMaxItemSize)
+            assert(0);  // ditto
+
+        opcodes.popFrontN(T.sizeof);
+        if (opcodes.length < size)
+            assert(0);  // ditto
+
+        payload = opcodes[0 .. size];
+        opcodes.popFrontN(size);  // advance to next opcode
+        return null;
+    }
+
+    /***************************************************************************
+
+        Reads the Signature and Public key from the stack,
+        and validates the signature against the provided
+        spending transaction.
+
+        If the Signature and Public key are missing or in an invalid format,
+        an error string is returned.
+
+        Otherwise the signature is validated and the `sig_valid` parameter
+        is set to the validation result.
+
+        Params:
+            OP = the opcode
+            stack = should contain the Signature and Public Key
+            tx = the transaction that should have been signed
+
+        Returns:
+            an error string if the Signature and Public key are missing or
+            invalid, otherwise returns null.
+
+    ***************************************************************************/
+
+    private string verifySignature (OP op)(ref Stack stack, in Transaction tx,
+        out bool sig_valid) nothrow @safe //@nogc  // stack.pop() is not @nogc
+    {
+        static assert(op == OP.CHECK_SIG || op == OP.VERIFY_SIG);
+
+        // if changed, check assumptions
+        static assert(Point.sizeof == 32);
+        static assert(Signature.sizeof == 64);
+
+        if (stack.count() < 2)
+        {
+            static immutable err1 = op.to!string
+                ~ " opcode requires two items on the stack";
+            return err1;
+        }
+
+        const key_bytes = stack.pop();
+        if (key_bytes.length != Point.sizeof)
+        {
+            static immutable err2 = op.to!string
+                ~ " opcode requires 32-byte public key on the stack";
+            return err2;
+        }
+
+        const point = Point(key_bytes);
+        if (!point.isValid())
+        {
+            static immutable err3 = op.to!string
+                ~ " 32-byte public key on the stack is invalid";
+            return err3;
+        }
+
+        const sig_bytes = stack.pop();
+        if (sig_bytes.length != Signature.sizeof)
+        {
+            static immutable err4 = op.to!string
+                ~ " opcode requires 64-byte signature on the stack";
+            return err4;
+        }
+
+        const sig = Signature(sig_bytes);
+        sig_valid = Schnorr.verify(point, sig, tx);
+        return null;
+    }
+}
+
+version (unittest)
+{
+    // sensible defaults
+    private const TestStackMaxTotalSize = 16_384;
+    private const TestStackMaxItemSize = 512;
+}
+
+// OP.INVALID
+unittest
+{
+    scope engine = new Engine(TestStackMaxTotalSize, TestStackMaxItemSize);
+    test!("==")(engine.execute(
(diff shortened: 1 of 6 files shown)


source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+/// The supported opcodes
+/// Opcodes named `CHECK_*` push their result to the stack,
+/// whereas `VERIFY_*` opcodes invalidate the transaction if the result is false.
+/// Can encode up to 255 opcodes (one of which is INVALID).
+/// Note that the range between `PUSH_BYTES_1` and `PUSH_BYTES_64` is
+/// purposefully reserved to encode a push to the stack between 1 .. 64 bytes,
+/// without having to encode it in a separate length byte. 64 was chosen as
+/// the upper bound because our signatures are 64 bytes.
+/// For pushes of data longer than 64 bytes use the `PUSH_DATA_*` opcodes.
+enum OP : ubyte
```
TrustHenry: Great! a completely unique new OPcode definition.

AndrejMitrovic:
I'll take care of the code coverage soon. It seems I missed some test-cases, though 92.60% is pretty good. :P

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+                    return error;
+                break;
+
+            case LockType.Script:
+                if (auto error = executeBasicScripts(lock, unlock, tx))
+                    return error;
+                break;
+
+            case LockType.Redeem:
+                if (auto error = executeScriptHash(lock, unlock, tx))
```
AndrejMitrovic: This should maybe be renamed to `executeRedeemScript` to make it consistent.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+        }
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Handle stack-less and script-less basic payments.
+
+        If the lock is a P2K, the unlock must only contain a <signature>.
```
AndrejMitrovic: This documentation assumes the reader knows about Bitcoin, it should describe better what the lock script is expected to be.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+            return "Script failed";
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Execute a P2SH (Pay 2 Script Hash) type of lock script.
+
+        The 64-byte hash of the redeem script `H` is read from `lock_bytes`,
```
AndrejMitrovic: This should say "referred to as "H" in this text" or something, it's not immediately obvious what `H` is.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+        `unlock_bytes` is evaluated as a set of pushes to the stack where
+        the last push is the redeem script. The redeem script is popped from the
+        stack, hashed, and compared to `H` from the lock script. Then it's
```
AndrejMitrovic: Also missing: Only if the hashes match, is the **redeem** script evaluated.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+        evaluated with any leftover stack items.
+
+        Params:
+            lock = must contain a 64-byte hash of the redeem script
+            unlock = must contain only stack push opcodes, where the last
+                           push is the redeem script itself
```
AndrejMitrovic: bad indentation

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+            tx = the associated spending transaction
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    private string executeScriptHash (Lock lock, Unlock unlock,
```
AndrejMitrovic: Rename the function as previously suggested.

AndrejMitrovic: Why aren't all the parameters marked with `in`?

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+            unlock = may contain a <signature>, <signature, key>,
+                           or <script> which only contains stack push opcodes
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    public string execute (Lock lock, Unlock unlock, in Transaction tx)
```
AndrejMitrovic: Why aren't all the parameters marked with `in`?

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+
+        Main dispatch execution routine.
+
+        The lock type will be examined, and based on its type execution will
+        proceed to either simple script-less payments, or script-based payments.
+
+        Params:
+            lock = the lock
+            unlock = may contain a <signature>, <signature, key>,
+                           or <script> which only contains stack push opcodes
```
AndrejMitrovic: Fix indentation.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+    import agora.utils.Test;
+    import ocean.core.Test;
+    import std.stdio : writefln, writeln;  // avoid importing LockType
+}
+
+/// Ditto
+public class Engine
+{
+    /// Conditional opcodes require the top item on the stack to be one of these
+    private static immutable ubyte[1] TRUE = [OP.TRUE];
```
AndrejMitrovic: It isn't immediately obvious what the difference is between `OP.TRUE` and `TRUE`.

One is a value on the stack, and the other is the opcode which pushes that value to the stack. So how about renaming it to something like `TrueValue`?

It should be very clear to the reader that you can't push an opcode to the stack (though serialized redeem scripts are an exception!).

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+
+    public string execute (Lock lock, Unlock unlock, in Transaction tx)
+        nothrow @safe
+    {
+        if (lock.bytes.length == 0)
+            return "Lock cannot be empty";
+
+        final switch (lock.type)
+        {
+            case LockType.Key:
```
AndrejMitrovic: Inconsistent indentation. This should be un-indented.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+            return error;
+
+        Stack stack;
+        if (auto error = executeScript(unlock_script, stack, tx))
+            return error;
+
+        if (auto error = executeScript(lock_script, stack, tx))
+            return error;
+
+        if (hasStackFailed(stack))
```
AndrejMitrovic: There is missing `this.` syntax for all the method calls in this entire class.

There's a reason why. Originally all of these methods were static. In fact there is no state that the engine keeps around, so we could still mark it static. But I removed static because I thought in the future we may end up wanting to keep some state around.

In any case, let's add `this.` back to all the calls again.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+            return "Script failed";
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Execute the script with the given stack and the associated transaction
```
AndrejMitrovic: Make a note that this function executes the script regardless if it's a lock or an unlock script.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
+
+    ***************************************************************************/
+
+    private string executeScript (in Script script, ref Stack stack,
+        in Transaction tx) nothrow @safe
+    {
+        ScopeCondition sc;
```
AndrejMitrovic, line 352: It wouldn't hurt to add a comment here, even though the `ScopeCondition` documentation itself is sufficient.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+        const(ubyte)[] bytes = script[];
```
AndrejMitrovic, line 353: This could use another comment - for example whether an empty script is ok or not. The `Returns:` docs should reflect this too.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+
+            case OP.PUSH_DATA_2:
+                if (auto reason = pushToStack!(OP.PUSH_DATA_2)(stack, bytes))
+                    return reason;
+                break;
+
+            case OP.PUSH_BYTES_1: .. case OP.PUSH_BYTES_64:
+                const payload_size = opcode;  // encoded in the opcode
+                if (bytes.length < payload_size)
+                    assert(0);  // should have been validated
```
AndrejMitrovic: Extend this comment by saying how & when the script should have been validated.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+
+        if (!sc.empty())
+            return "IF requires a closing END_IF";
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Handle a conditional opcode
```
AndrejMitrovic: This could use more documentation.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+
+        Params:
+            opcode = the current conditional
+            stack = the stack to evaluate for the conditional
+            sc = the scope condition which may be toggled
+
+        Returns:
+            null if there were no errors,
+            or a string explaining the reason execution failed
```
AndrejMitrovic, line 535: This should explain briefly in which cases execution of conditionals can fail.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+                return "IF/NOT_IF opcode requires an item on the stack";
+
+            const top = stack.pop();
+            if (top != TRUE && top != FALSE)
+                return "IF/NOT_IF may only be used with OP.TRUE / OP.FALSE values";
+
+            sc.push((opcode == OP.IF) ^ (top == FALSE));
+            break;
+
+        case OP.ELSE:
```
AndrejMitrovic, line 562: Trailing `OP.ELSE` opcodes are not checked here yet - e.g. `IF { } ELSE {} ELSE {} ELSE {}`. This is a known quirk of the Bitcoin scripting language, and we should try not to replicate its mistakes.

I haven't thought about a good solution for this yet - mainly because I'm busy with more important stuff right now like Lightning Network support. This is not technically a vulnerability, btw.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+        default:
+            assert(0);
+        }
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Checks if the stack has a failing condition.
```
AndrejMitrovic: "condition" could be misread as being one of the conditional opcodes like `IF`. Instead the docs should be explicit about what constitutes a "failed stack". Maybe "hasStackFailed" is a bad name for this function too..

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+        pure nothrow @safe
+    {
+        return stack.empty() || stack.peek() != TRUE;
+    }
+
+    /***************************************************************************
+
+        Reads the length and payload of the associated `PUSH_DATA_*` opcode,
+        tries to push the payload onto the stack, and if successfull it advances
+        the `opcodes` array to the next opcode.
```
AndrejMitrovic: Add a note that script syntax validation should have been done first, because the asserts will fail if the script size exceeds the configured limits.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+    private string pushToStack (OP op)(ref Stack stack,
+        ref const(ubyte)[] opcodes) nothrow @safe /*@nogc*/
+    {
+        static assert(op == OP.PUSH_DATA_1 || op == OP.PUSH_DATA_2);
+        alias T = Select!(op == OP.PUSH_DATA_1, ubyte, ushort);
+        if (opcodes.length < T.sizeof)
+            assert(0);  // script should have been validated
+
+        const T size = littleEndianToNative!T(opcodes[0 .. T.sizeof]);
+        if (size == 0 || size > MAX_STACK_ITEM_SIZE)
```
AndrejMitrovic: This should be part of `ConsensusParams`.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+    private bool hasStackFailed (/*in*/ ref Stack stack) // peek() is not const
+        pure nothrow @safe
+    {
+        return stack.empty() || stack.peek() != TRUE;
+    }
+
+    /***************************************************************************
+
+        Reads the length and payload of the associated `PUSH_DATA_*` opcode,
+        tries to push the payload onto the stack, and if successfull it advances
```
AndrejMitrovic: Mention that the data is pushed in LE order.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1002 @@
...
+        stack.push(payload);  // push to stack
+        opcodes.popFrontN(size);  // advance to next opcode
+        return null;
+    }
+}
+
+// workaround to allow pure
+extern (C) int crypto_core_ed25519_is_valid_point (const(ubyte)* p) pure nothrow;
+
+/// See #1279
```
AndrejMitrovic: This could still use some documentation beyond just linking to an issue.

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+
+*******************************************************************************/
+
+module agora.script.Opcodes;
+
+import std.traits : EnumMembers;
+
+/// The supported opcodes
+/// Opcodes named `CHECK_*` push their result to the stack,
+/// whereas `VERIFY_*` opcodes invalidate the transaction if the result is false.
```
AndrejMitrovic: They don't invalidate the transaction, they just make `engine.execute` fail.

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+/// Can encode up to 255 opcodes (one of which is INVALID).
```
AndrejMitrovic: We may replace `INVALID` with something like `ABORT` or `PANIC` - if the only use-case is to halt executing the script and marking the transaction as invalid. Additionally there are no tests for this opcode.

Bitcoin implements this opcode, but we may not need it (yet).

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+/// Note that the range between `PUSH_BYTES_1` and `PUSH_BYTES_64` is
+/// purposefully reserved to encode a push to the stack between 1 .. 64 bytes,
+/// without having to encode it in a separate length byte. 64 was chosen as
+/// the upper bound because our signatures are 64 bytes.
```
AndrejMitrovic: ".., and that is one of our largest types".

But this may change in the future.

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+
+    /// Checks the signature with the given public key,
+    /// pushes the result to the stack
+    CHECK_SIG = 0x4E,
+
+    /// Checks the lock time against the current block height
+    VERIFY_LOCK_TIME = 0x4F,
+
+    /// Encodes a set of web assembly instructions
+    WEB_ASM = 0x50,
```
AndrejMitrovic: Not used yet, so we may just remove it in this PR.

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+
+    Params:
+        opcode = opcode to check
+
+    Returns:
+        true if the opcode is one of the conditional opcodes
+
+*******************************************************************************/
+
+public bool isConditional (OP opcode) pure nothrow @safe @nogc
```
AndrejMitrovic: Needs test-cases.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+
+public struct Stack
+{
+    /// The actual stack
+    private SList!(const(ubyte)[]) stack;
+
+    /// The number of items on the stack
+    private ulong num_items;
+
+    /// Total used bytes for this stack
```
AndrejMitrovic: Mention what it's used for.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the top item on the stack, without popping it
+
+    ***************************************************************************/
+
+    public const(ubyte)[] peek () /*const*/ pure nothrow @safe @nogc
```
AndrejMitrovic: Should mention why `const` can't be used here (hint: blame Phobos).

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+    {
+        assert(!this.stack.empty());
+        return this.stack.front();
+    }
+
+    /***************************************************************************
+
```
AndrejMitrovic, line 148: Mention that `.empty()` should be checked first.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+    ***************************************************************************/
+
+    public ulong usedBytes () const pure nothrow @safe @nogc
+    {
+        return this.used_bytes;
+    }
+
+    /***************************************************************************
+
+        Get the number of items on the stack. Explicitly typed as ulong to
```
AndrejMitrovic: Improve the docs here: the "platform-dependent" behavior is related to the behavior of the opcodes. For example if there was theoretically an opcode which hashed the size of the stack, it would have different behavior on x86 and x86-64 if the return value was a `size_t` here.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+            true if the stack is empty
+
+    ***************************************************************************/
+
+    public bool empty () const pure nothrow @safe @nogc
+    {
+        return this.stack.empty();
+    }
+
+    /// SList uses reference semantics by default. For the user to either use
```
AndrejMitrovic: Docs seem wrong. "For the user to" should be removed.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+    /// `ref` or to explicitly copy the stack via `copy()`.
+    public @disable this(this);
+
+    /***************************************************************************
+
+        Returns:
+            a copy of the stack. The two stacks may then be modified
+            independently of each other. Items may not be modified as
```
AndrejMitrovic: may not => cannot

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+            they're immutable, making the stack safe.
+
+    ***************************************************************************/
+
+    public Stack copy () /*const @nogc*/ pure nothrow @safe
```
AndrejMitrovic: Document why these attributes can't be used.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /***************************************************************************
+
+        Returns:
+            a range over the stack items, from top to bottom
+
+    ***************************************************************************/
+
+    public auto opSlice () /*const @nogc*/ pure nothrow @safe
```
AndrejMitrovic: Also mention why these attribute can't be used.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+    assert(stack.pop() == [1, 2, 3]);
+    assert(stack.count() == 0);
+    test!"=="(stack.usedBytes(), 0);
+    assert(stack.empty());
+    assert(copy.count() == 2);     // did not consume copy
+    assert(copy.usedBytes() == 4); // ditto
+    assert(!copy.empty());         // ditto
+    assert(stack.canPush(ubyte(42).repeat(100).array));
+    assert(!stack.canPush(ubyte(42).repeat(MAX_STACK_ITEM_SIZE + 1).array));
+    // todo: MAX_STACK_TOTAL_SIZE check is non-trivial (expensive)
```
AndrejMitrovic: If we use `ConsensusParams` then this check is no longer expensive because we can tailor it to our needs.

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
+/*******************************************************************************
```
AndrejMitrovic, line 1: The commit message definitely needs more documentation.

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+    /***************************************************************************
+
+        Validates the script syntactically, but not semantically.
+        Each opcode is checked if it's recognized, and any push opcodes have
+        their length & payload checked for size constraints.
+
+        The semantics of the script are not checked here. They may only be
+        checked by the script execution engine.
+
+        Ditto, but returns the string reason when the script is
```
AndrejMitrovic: This `Ditto` is out of place.

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+
+    public const(ubyte)[] opSlice () const pure nothrow @safe @nogc
+    {
+        return this.data[];
+    }
+
+    /***************************************************************************
+
+        Validates the script syntactically, but not semantically.
+        Each opcode is checked if it's recognized, and any push opcodes have
```
AndrejMitrovic: opcode / recognized => byte / is equal to one of the known opcodes.

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+        their length & payload checked for size constraints.
+
+        The semantics of the script are not checked here. They may only be
```
AndrejMitrovic: may / can

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+    ***************************************************************************/
+
+    public string isInvalidSyntaxReason (ScriptType type)
+        const pure nothrow @safe @nogc
+    {
+        const(ubyte)[] bytes = this.data[];
+        if (bytes.empty)
+            return null;  // empty scripts are syntactically valid
+
+        // todo: add script size checks (based on consensus params)
```
AndrejMitrovic: This todo should actually be moved somewhere else. Because the script size itself is determined during deserialization.

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+            otherwise the string explaining the reason why it's invalid
+
+    ***************************************************************************/
+
+    public string isInvalidSyntaxReason (ScriptType type)
+        const pure nothrow @safe @nogc
+    {
+        const(ubyte)[] bytes = this.data[];
+        if (bytes.empty)
+            return null;  // empty scripts are syntactically valid
```
AndrejMitrovic: In the Engine we unconditionally return an error if the Lock script is empty. So maybe we should move that check here, and get rid of the special check in the Engine?

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+
+        Checks the validity of a `PUSH_DATA_*` opcode and advances
+        the `bytes` array if the payload does not exceed the array.
+
+        Params:
+            OP = the associated `PUSH_DATA_*` opcode
+            bytes = the opcode byte array
+
+        Returns:
+            null if the opcode is syntactically valid,
```
AndrejMitrovic: The docs are wrong. It returns `null` if the data does not exceed the size constraints. But again, these should be moved to `ConsensusParams`.

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+    Params:
+        key_hash = the key hash to encode in the P2PKH lock script
+
+    Returns:
+        a P2PKH lock script which can be unlocked with the matching
+        public key & signature
+
+*******************************************************************************/
+
+public Script createLockP2PKH (Hash key_hash) pure nothrow @safe
```
AndrejMitrovic, line 325: This should be in a `version (unittest)`. And actually, it belongs in the `Engine.d` module.

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+    Params:
+        sig = the signature
+        pub_key = the public key
+
+    Returns:
+        a P2PKH unlock script which can be used with the associated lock script
+
+*******************************************************************************/
+
+public Script createUnlockP2PKH (Signature sig, Point pub_key)
```
AndrejMitrovic, line 346: Ditto move this to `Engine.d` and make it unittest-only.

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+
+    Script unlock_script = createUnlockP2PKH(sig, kp.V);
+    assert(unlock_script.isInvalidSyntaxReason(ScriptType.Unlock) is null);
+}
+
+/*******************************************************************************
+
+    Creates `PUSH_DATA_*` opcodes based on the length of the data.
+    It does not verify data limits for the stack, as this check
+    belongs in the Stack / Engine. Additionally it's useful to be able to
```
AndrejMitrovic: Stack => Script syntax validation?

source/agora/script/Script.d
```diff
@@ -0,0 +1,350 @@
...
+    create exceeding data buffers with this function for testing.
```
AndrejMitrovic: Then this should be moved elsewhere or put inside of `version(unittest)`

source/agora/script/Lock.d
```diff
@@ -0,0 +1,102 @@
+/*******************************************************************************
+
+    Contains the Lock / Unlock definitions
```
AndrejMitrovic, line 3: This is not adequate documentation.

source/agora/script/Lock.d
```diff
@@ -0,0 +1,102 @@
...
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.script.Lock;
+
+import std.traits : EnumMembers;
+
+/// Contains the Lock, which is a tag and either a Hash or set of opcodes
```
AndrejMitrovic: is => contains

source/agora/script/Lock.d
```diff
@@ -0,0 +1,102 @@
...
+public struct Lock
+{
+    /// Specifies the type of lock script
+    public LockType type;
+
+    /// May either be a Hash, or a sequence of opcodes
+    public const(ubyte)[] bytes;
+}
+
+/// Contains the Unlock, which can be a data tuple or a set of push opcodes
```
AndrejMitrovic: Bad docs. We can't say that the Unlock struct contains "the Unlock", it's too self-referential and doesn't make sense.

source/agora/script/Lock.d
```diff
@@ -0,0 +1,102 @@
...
+public enum LockType : ubyte
+{
+    /// lock is a 64-byte public key, unlock is the signature
+    Key = 0x0,
+
+    /// lock is a 64-byte public key hash, unlock is a (sig, key) pair
+    KeyHash = 0x01,
+
+    /// lock is a script, unlock may be anything required by the lock script
+    Script = 0x2,
```
AndrejMitrovic: `0x02`

source/agora/script/Lock.d
```diff
@@ -0,0 +1,102 @@
...
+
+    /// lock is a 64-byte hash of a script, unlock is the script containing
+    /// only stack pushes which will push the redeem script last
+    Redeem = 0x3,
```
AndrejMitrovic: `0x03`

source/agora/script/Lock.d
```diff
@@ -0,0 +1,102 @@
...
+    Params:
+        value = the byte containing the opcode
+        opcode = will contain the opcode if it was recognized
+
+    Returns:
+        true if the value is a recognized opcode
+
+*******************************************************************************/
+
+public bool toLockType (ubyte value, out LockType opcode)
```
AndrejMitrovic: This needs tests.

source/agora/script/ScopeCondition.d
```diff
@@ -0,0 +1,229 @@
+/*******************************************************************************
+
+    Keeps track of scopes and their conditions (TRUE or FALSE).
+    This struct can be used to implement conditional (IF/ELSE/ENDIF) logic.
+
+    It does this pushing a new scope for each visited `IF` opcode,
```
AndrejMitrovic: this => this by

source/agora/script/ScopeCondition.d
```diff
@@ -0,0 +1,229 @@
...
+    popping a scope for every visited `ENDIF` opcode, and toggling the scope's
+    condition for every visited `ELSE` opcode.
+
+    Unlike C-like programming languages, we do not support GOTO and therefore
+    may only increment the program counter one instruction at a time.
```
AndrejMitrovic, line 11: ... "Making the implementation of conditional scopes simple." etc

source/agora/script/ScopeCondition.d
```diff
@@ -0,0 +1,229 @@
...
+    This implementation is largely based on Bitcoin's `ConditionStack`,
+    as it's the most optimal O(1) solution we can think of.
+
+    For a description on how code flow control works (for a previous version),
+    see: https://building-on-bitcoin.com/docs/slides/Thomas_Kerin_BoB_2018.pdf
+
+    Copyright:
+        Copyright (c) 2009-2010 Satoshi Nakamoto
+        Copyright (c) 2009-2020 The Bitcoin Core developers
+        Copyright (c) 2020 BOS Platform Foundation Korea
```
AndrejMitrovic, line 23: I'm not actually sure this passes the legalese. The code is based on Bitcoin's implementation, but I reordered some code and renamed a bunch of the fields. I did this to make the code clearer to understand.

source/agora/script/ScopeCondition.d
```diff
@@ -0,0 +1,229 @@
+/*******************************************************************************
```
AndrejMitrovic, line 1: In the commit message:

> It's used for tracking which conditional scope are true or not,

It's used for tracking which conditional scopes are true - code may be executed,
or not true - code execution will be skipped until the scope's condition is true again,

AndrejMitrovic (commented):
Some comments.

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+    /// may be used with `case PUSH_BYTES_1: .. case PUSH_BYTES_64:` syntax.
+    PUSH_BYTES_1 = 0x01,
+    PUSH_BYTES_64 = 0x40, // 64 decimal
+
+    /// The next 1 byte contains the number of bytes to push onto the stack
+    PUSH_DATA_1 = 0x41,
+
+    /// The next 2 bytes (ushort in LE format) contains the number of bytes to
+    /// push onto the stack
+    PUSH_DATA_2 = 0x42,
```
AndrejMitrovic: There should be another test here: do not use `PUSH_DATA_2` if the data length can fit within `PUSH_DATA_1`, and also do not use `PUSH_DATA_1` if the data can fit within `PUSH_BYTES_*`.

Even though there is technically nothing wrong with using the bigger opcodes to encode smaller pushes, if we enforce this at the protocol level we can make sure the scripts are as optimized as possible (with regards to space usage).

source/agora/script/Opcodes.d
AndrejMitrovic: Missing `public`

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+
+    /// The next 4 bytes (ushort in LE format) contains the number of bytes to
+    /// push onto the stack
+    RESERVED_PUSH_DATA_3 = 0x43,
+
+    /// Pushes True onto the stack
+    TRUE = 0x44,
+
+    /// Pushes False onto the stack
+    FALSE = 0x00,
```
AndrejMitrovic: Make a note about the semantic meaning of these values. For example: they can be used with conditionals, and also as the final result of the script based on the top item on the stack.

source/agora/script/Script.d
AndrejMitrovic: The documentation should also describe the language itself. It's a bitcoin-style language (forth-style), where values are pushed to the stack before conditionals are executed.

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+    {
+        opcode = cast(OP)value;  // dirty, but avoids need to define all pushes
+        return true;
+    }
+
+    return false;
+}
+
+/// Ditto, but assumes the opcode is valid (safe to use after validation)
+public OP toOPCode (ubyte value) pure nothrow @safe @nogc
```
AndrejMitrovic: This is still problematic because client code could accidentally use `if (toOPCode(value)) { }`. Let's just get rid of it.

source/agora/script/Engine.d
AndrejMitrovic, line 353: It seems a bit strange that we have a `Script` that is built using `ubyte[]`, then manually validated, and then in here it's sliced again.

It would be better to move the syntactical validation into a static routine which **returns** a valid script. That way we can know for sure that when a function receives a `Script` parameter, it is by definition valid.

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,167 @@
...
+    /// The next 1 byte contains the number of bytes to push onto the stack
+    PUSH_DATA_1 = 0x41,
+
+    /// The next 2 bytes (ushort in LE format) contains the number of bytes to
+    /// push onto the stack
+    PUSH_DATA_2 = 0x42,
+
+    /// The next 4 bytes (ushort in LE format) contains the number of bytes to
+    /// push onto the stack
+    RESERVED_PUSH_DATA_3 = 0x43,
```
AndrejMitrovic: One question remains. How do we maintain backward compatibility when introducing new opcodes?

Bitcoin does it by causing the reserved opcodes to become a simple push to the stack. And then when the opcode is later defined to do something in a soft-fork, the redeem script actually has to use a `OP_DROP` opcode ~first~ after the opcode.

There's more info here: https://bitcoin.stackexchange.com/a/85472/93682

I will likely work on this another time, I don't want to stall the PR.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+    }
+
+    /***************************************************************************
+
+        Returns:
+            the number of bytes used by this stack
+
+    ***************************************************************************/
+
+    public ulong usedBytes () const pure nothrow @safe @nogc
```
AndrejMitrovic: This is actually not used by any other code. We have `canPush` instead. Let's remove it.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,233 @@
...
+import std.range;
+
+version (unittest)
+{
+    import ocean.core.Test;
+    import std.stdio;
+}
+
+/// Maximum total stack size
+public enum MAX_STACK_TOTAL_SIZE = 16_384;
```
AndrejMitrovic: `MAX_STACK_*` => `STACK_MAX_*` for better visual categorization. Also, we can just move it to consensus-style constructor parameters.

source/agora/script/Lock.d
```diff
@@ -0,0 +1,102 @@
...
+
+    Converts the byte to an opcode,
+    or returns false if it's an unrecognized opcode.
+
+    Params:
+        value = the byte containing the opcode
+        opcode = will contain the opcode if it was recognized
+
+    Returns:
+        true if the value is a recognized opcode
```
AndrejMitrovic, line 79: Bad docs.

AndrejMitrovic:
I've fixed all the comments.

I now have to increase the code coverage.

AndrejMitrovic:
I've increased code coverage to 100%

(minus the `assert(0);` which is unreachable in some places)

AndrejMitrovic:
I've originally thought that I don't have to add support for a multisig opcode because we have Schnorr, but I've realized that our  signature scheme is insecure for regular users. I think for now I'll skip adding the opcode, but we'll have to think of a good solution to the following problem.

It's secure for the Validators between them to use them because they already committed and signed with their private key (through the enrollment), proving they own the key. But it's not secure for users because of the possibility of rogue key attacks. MuSig is one key aggregation scheme which tries to solve it. The following is a good read (I've highlighted the relevant part):

https://blockstream.com/2018/01/23/en-musig-key-aggregation-schnorr-signatures/#:~:text=fuss

We currently don't have anyone working on Schnorr. But this will become important soon.

For users to initiate channels and therefore use the flash layer we need to have a secure multi-party signature scheme for the funding transactions.

It doesn't block us from any work right now. We can build an implementation now, and harden the security later. I'll open up a Schnorr-related issue in the coming days.

AndrejMitrovic:
For anyone who is interested in MuSig (and also other schemes), I highly recommend this guide: https://tlu.tarilabs.com/cryptography/musig-schnorr-sig-scheme/The_MuSig_Schnorr_Signature_Scheme.html

There are other chapters in this book which are great to study.

source/agora/script/Lock.d
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/// Contains a data tuple or a set of push opcodes
+public struct Unlock
+{
+    /// May be: <signature>, <signature, key>, <push opcodes>
+    public const(ubyte)[] bytes;
+}
+
+/// The output lock types. If the lock is a 64-byte array it's derived
```
AndrejMitrovic: Input, not output!

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+            case OP.FALSE:
+                if (!stack.canPush(FalseValue))
+                    return "Stack overflow while pushing FALSE to the stack";
+                stack.push(FalseValue);
+                break;
+
+            case OP.PUSH_DATA_1:
+                if (auto reason = this.handlePushData!(OP.PUSH_DATA_1)(
+                    stack, bytes))
+                    return reason;
```
AndrejMitrovic: No coverage for this line!

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,187 @@
...
+    but is small enough to leave room for introducing other 1-byte opcodes.
+
+    For pushes of data longer than 64 bytes use one of the `PUSH_DATA_*`
+    opcodes.
+
+*******************************************************************************/
+
+public enum OP : ubyte
+{
+    /// Using this is an error and will invalidate the transaction
```
AndrejMitrovic: Executing, not using!

source/agora/script/ScopeCondition.d
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    /***************************************************************************
+
+        Push a new scope with the given condition.
+        If this is the first scope with a FALSE condition then
+        it sets the earliest FALSE scope index to the current scope.
+
+        Params:
+            cond = the evaluated condition of a visited IF opcode
```
AndrejMitrovic: This can also `NOT_IF`.

source/agora/script/Stack.d
```diff
@@ -0,0 +1,276 @@
...
+    public void push (const(ubyte)[] data) @safe nothrow
+    {
+        assert(this.canPush(data));
+        this.stack.insertFront(data);
+        this.used_bytes += data.length;
+        this.num_items++;
+    }
+
+    /***************************************************************************
+
```
AndrejMitrovic, line 131: Needs mention of calling `empty()`

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+    /// Maximum size of an item on the stack
+    private immutable ulong StackMaxItemSize;
+
+    /***************************************************************************
+
+        Initializes the script execution engine with the configured consensus
+        limits.
+
+        Params:
+            StackMaxTotalSize = the maximum allowed stack size before a
```
AndrejMitrovic, line 69: Wrong docs, this was copied from the Stack docs.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+        Execute a `LockType.Script` type of lock script with the associated
+        unlock script.
+
+        The unlock script may only contain stack pushes.
+        The unlock script is ran, producing a stack.
+
+        Thereafter, the lock script will run with the stack of the
+        unlock script.
+
+        Params:
```
AndrejMitrovic, line 222: Add a note why the two scripts should not be concatenated together before execution. Someone might think they're smart and try to optimize this, but it would cause a security vulnerability.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+        if (auto error = this.executeScript(unlock_script, stack, tx))
+            return error;
+
+        if (stack.empty())
+            return "LockType.Redeem requires unlock script to push a redeem script to the stack";
+
+        const redeem_bytes = stack.pop();
+        if (hashFull(redeem_bytes) != script_hash)
+            return "LockType.Redeem unlock script pushed a redeem script "
+                 ~ "which does not match redeem hash in the lock script";
```
AndrejMitrovic: the redeem hash

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Handle a conditional opcode like `OP.IF` / `OP.ELSE` / etc.
+
+        The initial scope is implied to be true. When a new scope is entered
+        via `OP.IF`, the condition is checked. If the condition is false,
```
AndrejMitrovic: or `OP.NOT_IF`

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+        then all the code inside the `OP.IF` block will be skipped until
+        we exit into the first scope where the condition is true.
+
+        Execution will fail if there is an `OP.ELSE` or `OP.END_IF` opcode
+        without an associated `OP.IF` opcode.
```
AndrejMitrovic: again not_if too

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+        "Script panic while executing OP.INVALID opcode");
+}
+
+// OP.DUP
+unittest
+{
+    scope engine = new Engine(TestStackMaxTotalSize, TestStackMaxItemSize);
+    test!("==")(engine.execute(
+        Lock(LockType.Script, [OP.DUP]), Unlock.init, Transaction.init),
+        "DUP opcode requires an item on the stack");
```
AndrejMitrovic, line 733: This lacks a legitimate test.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+}
+
+// OP.HASH
+unittest
+{
+    scope engine = new Engine(TestStackMaxTotalSize, TestStackMaxItemSize);
+    test!("==")(engine.execute(
+        Lock(LockType.Script, [OP.HASH]), Unlock.init, Transaction.init),
+        "HASH opcode requires an item on the stack");
```
AndrejMitrovic, line 753: ditto

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+        Unlock.init, tx),
+        null);
+    test!("==")(engine.execute(
+        Lock(LockType.Script,
+            [OP.PUSH_BYTES_1, 2, OP.PUSH_BYTES_1, 1, OP.VERIFY_EQUAL, OP.TRUE]),
+        Unlock.init, tx),
+        "VERIFY_EQUAL operation failed");
+}
+
+// OP.CHECK_SIG
```
AndrejMitrovic, line 814: lacks positive test

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+
+    Signature invalid_sig;
+    test!("==")(engine.execute(
+        Lock(LockType.Script, [ubyte(OP.PUSH_BYTES_64)] ~ invalid_sig[]
+            ~ [ubyte(32)] ~ valid_key[]
+            ~ [ubyte(OP.CHECK_SIG)]), Unlock.init, tx),
+        "Script failed");
+}
+
+// Native P2PK (Pay to Public Key), consumes 33 bytes
```
AndrejMitrovic: LockType.Key

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1213 @@
...
+        Lock(LockType.Key, kp.V[]),
+        Unlock(ubyte(42).repeat(32).array), tx),
+        "LockType.Key requires a 64-byte signature in the unlock script");
+    test!("==")(engine.execute(
+        Lock(LockType.Key, kp.V[]),
+        Unlock(ubyte(42).repeat(65).array), tx),
+        "LockType.Key requires a 64-byte signature in the unlock script");
+}
+
+// Native P2PKH (Pay to Public Key Hash), consumes 65 bytes
```
AndrejMitrovic: LockType.KeyHash

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1257 @@
...
+        in Transaction tx) nothrow @safe
+    {
+        // tracks executable condition of scopes for use with IF / ELSE / etc
+        ScopeCondition sc;
+        const(ubyte)[] bytes = script[];
+        while (!bytes.empty())
+        {
+            OP opcode;
+            if (!bytes.front.toOPCode(opcode))
+                assert(0, "Script should have been syntactically validated");
```
AndrejMitrovic, line 358: This seems to fail sometimes, even though the syntactical validation should have caught it.

It could mean there is a mismatch between the `handlePushData` implementation and `isInvalidPushReason`. It might be good to merge these two routines so they stay in sync.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1257 @@
...
+            if (opcode.isConditional())
+            {
+                if (auto error = this.handleConditional(opcode, stack, sc))
+                    return error;
+                continue;
+            }
+
+            // whether the current scope is executable
+            // (all preceeding outer conditionals were true)
+            if (!sc.isTrue())
```
AndrejMitrovic, line 400: This is problematic. Because it's not skipping the stack push opcodes. So if an `ELSE` condition turns the scope to false, we won't pop the right amount of opcodes and the data payload will be interpreted as an opcode, and fail with an assert.

AndrejMitrovic:
Resolved proper payload extraction when the scope is false.

source/agora/script/Engine.d
```diff
@@ -0,0 +1,1287 @@
...
+            const sig = Signature(unlock.bytes);
+            if (!Schnorr.verify(key, sig, tx))
+                return "LockType.Key signature in unlock script failed validation";
+
+            break;
+
+        case LockType.KeyHash:
+            if (lock.bytes.length != Hash.sizeof)
+                return "LockType.KeyHash requires a 64-byte key hash argument in the lock script";
+            const Hash key_hash = Hash(lock.bytes);
```
AndrejMitrovic, line 180: There is a missing check that the key hashes match here.

source/agora/script/Opcodes.d
```diff
@@ -0,0 +1,217 @@
...
+    /// Executing this is an error and will cause script execution to fail.
+    /// Purposefully located first to default OPs to errors.
+    INVALID = 0x45,
+
+    /// Used to encode a small data push to the stack (up to 64 bytes),
+    /// may be used with `case PUSH_BYTES_1: .. case PUSH_BYTES_64:` syntax.
+    PUSH_BYTES_1 = 0x01,
+
+    /// Ditto
+    PUSH_BYTES_64 = 0x40, // 64 decimal
```
AndrejMitrovic: This will need to be extended. If we want to support `SIGHASH` then we need to encode it after the signature. So that will make needing 65-byte pushes common.

Let's be on the safe side and allow up to 75-byte pushes.

AndrejMitrovic:
Pushed a change to add the `Input` parameter to `execute()`. It's needed to implement `SigHash.NoInput` and other sighash types for the signatures. It's folded into the previous commit.

AndrejMitrovic:
Removed `isValidPointBytes()` as it's now part of `Point.isValid()` in a recent master change. And also fixed the accidental unittest breakage.

AndrejMitrovic:
Update: Added `OP.VERIFY_SIG` because it doesn't make sense to split it into a separate PR. This change also introduced a `verifySignature` method which is used by both `CHECK_SIG` and `VERIFY_SIG`. It will become even more useful once `SigHash` is added.

AndrejMitrovic:
I will merge this later today if there are no further comments.

This PR does not integrate the engine with Agora yet, it's part of another PR.

AndrejMitrovic:
This went through a lengthy self-review and received no other comments in the ~20 days it's been open. I'll merge and continue working on the other integration & feature PRs.

PR #1293, opened 2020-10-28
Validator recurring enrollments

### LINKED ISSUE #1254: Validator does not re-enroll after cycle expires
Currently, the `validator_cycle` is set to 20. The validator does not re-enroll after 20 blocks so more blocks cannot be created.

```
Fatal [agora.consensus.protocol.Nominator] - tryNominate(): Invalid consensus data: Enrollment: Insufficient number of active validators.
```

Triggered by `Ledger.d` :
```D
if (block.header.enrollments.length + active_enrollments < Enrollment.MinValidatorCount)
```

There should also be an option in which the validator does not want to re-enroll.

omerfirmak:
Validators used to not re-enroll at the end of their validator
cycle. Now they re-enroll with the existing UTXO one block before
their enrollment expire.

Some tests rely on the old behaviour. So a field for configuring the
re-enrollment is also added. Recurring enrollments are disabled by default.
We may enable it for tests that don't rely on the old behaviour.

hopefully fixes #1254

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -183,6 +183,9 @@ public struct ValidatorConfig
 
     // Registry address
     public string registry_address;
+
+    // If the enrollments will be renewed or not at the end of the cycle
+    public bool recurring_enrollment = true;
 }
 
 /// Admin API config
@@ -443,12 +446,14 @@ private ValidatorConfig parseValidatorConfig (Node* node, const ref CommandLine 
         return ValidatorConfig(false);
 
     auto registry_address = get!(string, "validator", "registry_address")(cmdln, node);
+    const recurring_enrollment = get!(bool, "validator", "recurring_enrollment")(cmdln, node);
     ValidatorConfig result = {
         enabled: true,
         key_pair:
             KeyPair.fromSeed(Seed.fromString(get!(string, "validator", "seed")(cmdln, node))),
         registry_address: registry_address,
         addresses_to_register : assumeUnique(parseSequence("addresses_to_register", cmdln, *node, true)),
+        recurring_enrollment : recurring_enrollment
     };
     return result;
 }
@@ -465,12 +470,14 @@ validator:
   enabled: true
   seed: SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
   registry_address: http://127.0.0.1:3003
+  recurring_enrollment : false
 `;
         auto node = Loader.fromString(conf_example).load();
         auto config = parseValidatorConfig("validator" in node, cmdln);
         assert(config.enabled);
         assert(config.key_pair == KeyPair.fromSeed(
             Seed.fromString("SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4")));
+        assert(!config.recurring_enrollment);
     }
     {
     immutable conf_example = `
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -637,7 +637,13 @@ public class EnrollmentManager
     }
 
     /// Returns: true if this validator is currently enrolled
-    public bool isEnrolled (UTXOFinder finder) nothrow @safe
+    public bool isEnrolled (scope UTXOFinder finder) nothrow @safe
+    {
+        return this.getEnrolledUTXO(finder) != Hash.init;
+    }
+
+    /// Returns: The UTXO hash that this Validator is enrolled with or Hash.init if not enrolled
+    public Hash getEnrolledUTXO (scope UTXOFinder finder) nothrow @safe
     {
         Hash[] utxo_keys;
         assert(this.validator_set.getEnrolledUTXOs(utxo_keys));
@@ -650,10 +656,10 @@ public class EnrollmentManager
                 assert(0, "UTXO for validator not found!");  // should never happen
 
             if (value.output.address == key)
-                return true;
+                return utxo_key;
         }
 
-        return false;
+        return Hash.init;
     }
 
     /***************************************************************************
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -14,6 +14,7 @@
 module agora.node.Validator;
 
 import agora.api.Validator;
+import agora.common.Amount;
 import agora.common.Config;
 import agora.common.Hash;
 import agora.common.crypto.Key;
@@ -363,6 +364,10 @@ public class Validator : FullNode, API
         // regenerate the quorums
         if (validators_changed || need_shuffle)
             this.regenerateQuorums(block.header.height);
+
+        // Re-enroll if our enrollment is about to expire
+        if (this.config.validator.recurring_enrollment)
+            this.checkAndEnroll(block.header.height);
     }
 
     /***************************************************************************
@@ -422,5 +427,51 @@ public class Validator : FullNode, API
         this.nominator.storeLatestState();
         if (this.config.admin.enabled)
             this.admin_interface.stop();
+    }
+
+    /***************************************************************************
+
+        Check if the current enrollment is about to expire or validator is not
+        enrolled. Enroll when necessary.
+
+        Params:
+            block_height = Current block height
+
+    ***************************************************************************/
+
+    private void checkAndEnroll (Height block_height) @safe
+    {
+        Hash enroll_key = this.enroll_man.getEnrolledUTXO(this.utxo_set.getUTXOFinder());
+
+        if (enroll_key == Hash.init && (enroll_key = this.getFrozenUTXO()) == Hash.init)
+            return; // Not enrolled and no frozen UTXO
+
+        const enrolled = this.enroll_man.getEnrolledHeight(enroll_key);
+
+        // This validators enrollment will expire next cycle or not enrolled at all
+        if (enrolled == ulong.max || block_height + 1 >= enrolled + this.params.ValidatorCycle)
+            this.network.sendEnrollment(this.enroll_man.createEnrollment(enroll_key));
+    }
+
+    /***************************************************************************
+
+        Get a frozen UTXO owned by the Validator
+
+        Return:
+            Returns hash of a UTXO eligible for staking
+
+    ***************************************************************************/
+
+    private Hash getFrozenUTXO () @safe
+    {
+        const pub_key = this.getPublicKey();
+        foreach (key, utxo; this.utxo_set.getUTXOs(pub_key))
+        {
+            if (utxo.type == TxType.Freeze &&
+                utxo.output.value.integral() >= Amount.MinFreezeAmount.integral())
+                return key;
+        }
+
+        return Hash.init;
     }
 }
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -233,6 +233,8 @@ public class Validator : FullNode, API
 
         if (this.enroll_man.isEnrolled(this.utxo_set.getUTXOFinder()))
             this.nominator.startNominatingTimer();
+        else if (this.config.validator.recurring_enrollment)
+            this.checkAndEnroll(this.ledger.getBlockHeight());
     }
 
     /***************************************************************************
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -452,7 +452,11 @@ public class Validator : FullNode, API
 
         // This validators enrollment will expire next cycle or not enrolled at all
         if (enrolled == ulong.max || block_height + 1 >= enrolled + this.params.ValidatorCycle)
+        {
+            log.trace("Sending Enrollment at height {} for {} cycles with {}", block_height,
+                                            this.params.ValidatorCycle , enroll_key);
             this.network.sendEnrollment(this.enroll_man.createEnrollment(enroll_key));
+        }
     }
 
     /***************************************************************************
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -1687,6 +1687,9 @@ public struct TestConf
 
     /// How often blocks should be created - in seconds
     uint block_interval_sec = 1;
+
+    /// If the enrollments will be renewed or not at the end of the cycle
+    bool recurring_enrollment = true;
 }
 
 /*******************************************************************************
@@ -1775,7 +1778,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         {
             banman : ban_conf,
             node : makeNodeConfig(self_address),
-            validator : ValidatorConfig(true, key_pair, [self_address]),
+            validator : ValidatorConfig(true, key_pair, [self_address], "", test_conf.recurring_enrollment),
             network : makeNetworkConfig(idx, addresses),
         };
 
--- a/source/agora/test/ValidatorRecurringEnrollment.d
+++ b/source/agora/test/ValidatorRecurringEnrollment.d
@@ -1,0 +1,65 @@
+/*******************************************************************************
+
+    Ensures validators re-enroll at the end of their validator cycle
+    when configured to do so
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.ValidatorRecurringEnrollment;
+
+import agora.test.Base;
+import agora.consensus.data.Transaction;
+
+unittest
+{
+    TestConf conf = {
+        quorum_threshold : 100
+    };
+    auto network = makeTestNetwork(conf);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto nodes = network.clients;
+    auto node_1 = nodes[0];
+
+    // Get the genesis block, make sure it's the only block externalized
+    auto blocks = node_1.getBlocksFrom(0, 2);
+    assert(blocks.length == 1);
+
+    Transaction[] txs;
+
+    void createAndExpectNewBlock (Height new_block_height)
+    {
+        // create enough tx's for a single block
+        txs = blocks[new_block_height - 1].spendable().map!(txb => txb.sign()).array();
+
+        // send it to one node
+        txs.each!(tx => node_1.putTransaction(tx));
+
+        network.expectBlock(new_block_height, blocks[0].header);
+
+        // add next block
+        blocks ~= node_1.getBlocksFrom(new_block_height, 1);
+    }
+
+    // create GenesisValidatorCycle - 1 blocks
+    foreach (block_idx; 1 .. GenesisValidatorCycle)
+    {
+        createAndExpectNewBlock(Height(block_idx));
+    }
+
+    // Create one last block
+    // if Validators don't re-enroll, this would fail
+    createAndExpectNewBlock(Height(GenesisValidatorCycle));
+    // Check if all validators in genesis are enrolled again
+    assert(blocks[blocks.length - 1].header.enrollments.length == blocks[0].header.enrollments.length);
+}
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -63,6 +63,9 @@ validator:
   addresses_to_register:
     - 88.88.88.88
     - best.validator.io
+  # Whether or not the Validator will enroll automatically at the startup or
+  # at the end of Validator cycle
+  recurring_enrollment: true
 
 ################################################################################
 ##                         Ban manager configuration                          ##
--- a/source/agora/test/EnrollmentManager.d
+++ b/source/agora/test/EnrollmentManager.d
@@ -35,7 +35,10 @@ import core.time;
 /// test for enrollment process & revealing a pre-image periodically
 unittest
 {
-    auto network = makeTestNetwork(TestConf.init);
+    TestConf conf = {
+        recurring_enrollment : false,
+    };
+    auto network = makeTestNetwork(conf);
     network.start();
     scope(exit) network.shutdown();
     scope(failure) network.printLogs();
@@ -65,8 +68,10 @@ unittest
 unittest
 {
     import agora.consensus.data.genesis.Test;
-
-    auto network = makeTestNetwork(TestConf.init);
+    TestConf conf = {
+        recurring_enrollment : false,
+    };
+    auto network = makeTestNetwork(conf);
     scope(exit) network.shutdown();
     scope(failure) network.printLogs();
     network.start();
@@ -97,7 +102,8 @@ unittest
 ///     reveals its pre-images periodically.
 unittest
 {
-    auto network = makeTestNetwork(TestConf.init);
+    TestConf conf = { recurring_enrollment : false };
+    auto network = makeTestNetwork(conf);
     scope(exit) network.shutdown();
     scope(failure) network.printLogs();
     network.start();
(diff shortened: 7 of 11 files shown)


source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -0,0 +1,30 @@
+module agora.test.ValidatorRecurringEnrollment;
```
AndrejMitrovic: Missing module header and trailing newline at the end of the file.

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -0,0 +1,30 @@
...
+
+    auto txes = genesisSpendable().map!(txb => txb.sign()).array();
+
+    Height expected_height;
+    foreach (tx; txes)
+    {
+        node_1.putTransaction(tx);
+        Thread.sleep(1.seconds);
+        network.expectBlock(++expected_height);
+    }
```
AndrejMitrovic, line 52: It would be good to add an assert here for the `expected_height` after the `foreach`.

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -0,0 +1,30 @@
...
+
+import agora.test.Base;
+
+import core.thread.osthread;
+
+import std.stdio;
+
+unittest
+{
+    TestConf conf = { validator_cycle : 3, txs_to_nominate : 1, recurring_enrollment : true };
```
AndrejMitrovic: Please set the `quorum_threshold` to `100`.

AndrejMitrovic: We might even consider changing the default to `100` in `TestConf` because it's a bit dangerous that it's implicitly lower, it circumvents expectations in the tests.

source/agora/test/Base.d
```diff
@@ -1644,7 +1647,8 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
             en.value,
             validator_addresses[en.index],
             enrolled_addresses.filter!(  // don't connect the validator to itself
-                addr => addr != validator_addresses[en.index]).array));
+                addr => addr != validator_addresses[en.index]).array,
+            test_conf.recurring_enrollment));
```
AndrejMitrovic: You don't need to pass this argument to `makeValidatorConfig`, that function can access `test_conf` directly.

source/agora/node/Validator.d
```diff
@@ -411,4 +415,29 @@ public class Validator : FullNode, API
...
+
+        Check if the current enrollment is about to expire and re-enroll
+        if necessary.
+
+        Params:
+            block_height = Current block height
+
+    ***************************************************************************/
+
+    private void checkEnrollmentExpiration (Height block_height) @safe
```
AndrejMitrovic: The function should be renamed. It does more than just check, it also re-enrolls.

I'm not sure what a better name would be without making it very long.

AndrejMitrovic: Maybe `checkAndReEnroll`? Looks a bit ugly though.

source/agora/node/Validator.d
```diff
@@ -411,4 +415,29 @@ public class Validator : FullNode, API
...
+    {
+        Hash enroll_key = this.enroll_man.getEnrolledUTXO(this.utxo_set.getUTXOFinder());
+        if (enroll_key != Hash.init)
```
AndrejMitrovic: Personal preference by me & Mathias is to prefer early returns rather than nesting indentation.

So:

```D
if (enroll_key == Hash.init)
    return;   // not enrolled
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -658,10 +664,12 @@ public class EnrollmentManager
                 assert(0, "UTXO for validator not found!");  // should never happen
 
             if (value.output.address == key)
-                return true;
+            {
+                return utxo_key;
+            }
```
AndrejMitrovic: Styling (simple expressions don't need braces):

```suggestion
                return utxo_key;
```

AndrejMitrovic (dismissed):
Just a few comments. Looks good!

source/agora/test/ValidatorRecurringEnrollment.d
kchulj: `validator_cycle` should be a part of `NodeConfig`.

source/agora/node/Validator.d
omerfirmak: `renewExpiringEnrollment` ? It is equally ugly tho :)

AndrejMitrovic: `checkAndEnroll` could work. I mean the fact that it's a "re"-enrollment doesn't really have that much of a semantical difference, my two cents..

kchulj (commented):
Nice!

source/agora/test/ValidatorRecurringEnrollment.d
omerfirmak: I believe there is no way to set `validator_cycle`, with `validator_cycle` being removed from `TestConf` right?

omerfirmak:
Updated according to comments

source/agora/test/Base.d
```diff
@@ -1597,7 +1600,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         {
             banman : ban_conf,
             node : makeNodeConfig(self_address),
-            validator : ValidatorConfig(true, key_pair, [self_address]),
+            validator : ValidatorConfig(true, key_pair, [self_address], "", test_conf.recurring_enrollment),
```
AndrejMitrovic, line 1781: I've just noticed that `registry_address` from `TestConf` is not used. That's a bug.

source/agora/node/Validator.d
```diff
@@ -411,4 +415,28 @@ public class Validator : FullNode, API
...
+    ***************************************************************************/
+
+    private void checkAndEnroll (Height block_height) @safe
+    {
+        Hash enroll_key = this.enroll_man.getEnrolledUTXO(this.utxo_set.getUTXOFinder());
+
+        if (enroll_key == Hash.init)
+            return; // not enrolled
+
+        auto enrolled = this.enroll_man.getEnrolledHeight(enroll_key);
```
Geod24:
```suggestion
        const enrolled = this.enroll_man.getEnrolledHeight(enroll_key);
```

source/agora/common/Config.d
```diff
@@ -183,6 +183,9 @@ public struct ValidatorConfig
 
     // Registry address
     public string registry_address;
+
+    // If the enrollments will be renewed or not at the end of the cycle
+    public bool recurring_enrollment;
```
Geod24: I'm not sure about making it a config option that is off by default. I feel like the sensible default is to have it to `true`, but if you want to re-enroll (or stop enrolling) it would be a bit weird to edit your config and reload / restart.
But since it's implemented I don't mind, just make it default to `true`.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -646,6 +646,12 @@ public class EnrollmentManager
 
     /// Returns: true if this validator is currently enrolled
     public bool isEnrolled (UTXOFinder finder) nothrow @safe
+    {
+        return this.getEnrolledUTXO(finder) != Hash.init;
+    }
+
+    /// Returns: The UTXO hash that this Validator is enrolled with or Hash.init if not enrolled
+    public Hash getEnrolledUTXO (UTXOFinder finder) nothrow @safe
```
Geod24: Can you also make both `UTXOFinder` parameter `scope` ? When using `scope` parameter, passing a delegate literal won't allocate with the GC.

source/agora/node/Validator.d
```diff
@@ -411,4 +415,28 @@ public class Validator : FullNode, API
...
+            block_height = Current block height
+
+    ***************************************************************************/
+
+    private void checkAndEnroll (Height block_height) @safe
+    {
+        Hash enroll_key = this.enroll_man.getEnrolledUTXO(this.utxo_set.getUTXOFinder());
+
+        if (enroll_key == Hash.init)
+            return; // not enrolled
```
Geod24: How does it behave if the node starts, there is a frozen UTXO, but it was never enrolled ?

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -0,0 +1,64 @@
...
+
+    // create GenesisValidatorCycle - 1 blocks
+    foreach (block_idx; 1 .. GenesisValidatorCycle)
+    {
+        createAndExpectNewBlock(Height(block_idx));
+    }
+
+    // Create one last block
+    // if Validators don't re-enroll, this would fail
+    createAndExpectNewBlock(Height(GenesisValidatorCycle));
```
Geod24, line 62: No check for `enrollments.length` ?

Geod24 (approved):
Very nice. Just two points:
- What about frozen UTXOs that have never been enrolled before ?
- The test could be a tiny bit more explicit in its check;

source/agora/common/Config.d
omerfirmak: and turn it off if a test relies on the existing behaviour?

source/agora/node/Validator.d
omerfirmak: Currently, it does nothing. Is `onAcceptedBlock` still a good place to enroll for the first time? Could there be a case where we should send an enrollment request but no blocks can be created?

source/agora/test/ValidatorRecurringEnrollment.d
omerfirmak: yeah, will add that.

source/agora/node/Validator.d
```diff
@@ -411,4 +416,50 @@ public class Validator : FullNode, API
...
+    private void checkAndEnroll (Height block_height) @safe
+    {
+        Hash enroll_key = this.enroll_man.getEnrolledUTXO(this.utxo_set.getUTXOFinder());
+
+        if (enroll_key == Hash.init && (enroll_key = this.getFrozenUTXO()) == Hash.init)
+            return; // Not enrolled and no frozen UTXO
+
+        const enrolled = this.enroll_man.getEnrolledHeight(enroll_key);
+
+        // This validators enrollment will expire next cycle or not enrolled at all
```
AndrejMitrovic, line 453: How can `enrolled` be `ulong.max` here? Shouldn't the previous `if` block take care of that?

omerfirmak: It would be `ulong.max` when the Validator is not enrolled but has a Frozen UTXO.

AndrejMitrovic: But the comment above says `Not enrolled and no frozen UTXO`

omerfirmak: Yes, so the early return handles the case where validator is not currently enrolled and there is no frozen utxo that we can use to enroll. If we dont return early we definitely have a frozen utxo. We will send an enrollment request in 2 cases, if the validator is not enrolled or the current enrollment is about to expire. `enrolled == ulong.max` covers the first case, `block_height + 1 >= enrolled + this.params.ValidatorCycle` covers the second.

AndrejMitrovic: Ah yes it makes sense now. Am a bit slow today. :)

source/agora/node/Validator.d
```diff
@@ -411,4 +416,50 @@ public class Validator : FullNode, API
...
+        if (enrolled == ulong.max || block_height + 1 >= enrolled + this.params.ValidatorCycle)
+            this.network.sendEnrollment(this.enroll_man.createEnrollment(enroll_key));
```
AndrejMitrovic, line 458: I think it would be useful to log this before sending the enrollment.

source/agora/node/Validator.d
omerfirmak: I am wondering if `checkAndEnroll()` should be called in `start()` or somewhere. I am trying to see if that would be a problem if validator havent caught up with the network yet.

source/agora/node/Validator.d
omerfirmak: Would that be a `log.trace` or a higher level?

source/agora/node/Validator.d
AndrejMitrovic: Thanks.  You made me notice a problem with our enrollments: https://github.com/bpfkorea/agora/issues/1297 :)

source/agora/node/Validator.d
AndrejMitrovic: trace is fine!

source/agora/node/Validator.d
omerfirmak: Accidental contribution!

omerfirmak:
> I am wondering if `checkAndEnroll()` should be called in `start()` or somewhere. I am trying to see if that would be a problem if validator havent caught up with the network yet.

I think we do. Think of the case where exactly minimum number of Validators are enrolled and all of their cycles are about to end. All but one decide to re-enroll. Unless we would still have minimum number of validators at the next block, nothing would be nominated. So a new Validator should be able to enroll without any new blocks being created.

AndrejMitrovic:
> All but one decide to re-enroll. 

Actually the protocol disallows this. It's in the block validation routines, there must always be a minimum number of active validators.

AndrejMitrovic:
> So a new Validator should be able to enroll without any new blocks being created.

Well a node can decide to enroll at any point, however their enrollment is not effective until it's part of an externalized block.

AndrejMitrovic:
In Stellar, for example, nodes do not enroll. They can enter the system at any point in time, regardless of the state of the blockchain. However, the network depends on reputation. Meaning that you as a validator will not be considered to be part of the validator network unless another node adds your node to their quorum. This is what Mathias refers to whenever he mentions that we want to have an open SCP network. In our system anyone can be a validator and they will become part of the quorum as long as they stake.

This does have its own challenges. For example, if our quorums only shuffle when a block is externalized then there can be a scenario where all the nodes are stuck (can't reach consensus), and can't re-shuffle (no block is externalized). This is something we'll likely have to improve in the future.

Anyway I thought I'd drop a bit of knowledge dump in here. :)

omerfirmak:
> > All but one decide to re-enroll.
> 
> Actually the protocol disallows this. It's in the block validation routines, there must always be a minimum number of active validators.

An already enrolled Validator may decide to un-enroll at the end of their cycle tho. That is what I was trying to point out :) So a new Validator should be able send a enrollment request async to the block creation. 

Essentially, what I mean is that `checkAndEnroll` can not be called only from `onAcceptedBlock`. Validator should be able to send an enrollment request whenever it pleases.

Thank you for the knowledge dump as well :)

AndrejMitrovic:
This is a very good point you're making. If nodes are at height 9 and cannot reach consensus on proposed height 10 because there's not enough enrollments (protocol rule), then other nodes should not wait for `onAcceptedBlock` to suggest re-enrolling.

IOW, the re-enroll should be continuously emitted until accepted. Perhaps a recurring timer would do the trick?

omerfirmak:
Enrolling on a timer would most probably work. But I would refrain from having periodic tasks as long as I can :)

AndrejMitrovic:
Do you have any other ideas?

omerfirmak:
> Do you have any other ideas?

https://github.com/bpfkorea/agora/blob/v0.x.x/source/agora/node/Ledger.d#L475

I think this is already called from the nomination timer. We could use this check to broadcast a "validators needed" message?

AndrejMitrovic:
> I think this is already called from the nomination timer. We could use this check to broadcast a "validators needed" message?

That's a good hunch. However it's not the right location. Nodes can nominate any set of values they wish. So we don't want to broadcast a "validators needed" message unless it's really necessary. "The boy who cried wolf" comes to mind.

And actually new validators are always welcome to join - for the benefit of the health of the overall network. So I'm not sure another network message would solve anything here.

omerfirmak:
> > I think this is already called from the nomination timer. We could use this check to broadcast a "validators needed" message?
> 
> That's a good hunch. However it's not the right location. Nodes can nominate any set of values they wish. So we don't want to broadcast a "validators needed" message unless it's really necessary. "The boy who cried wolf" comes to mind.
> 
> And actually new validators are always welcome to join - for the benefit of the health of the overall network. So I'm not sure another network message would solve anything here.

When booting up they can always try to enroll in `start()`, while catching up or at the latest when a new block is created. They will have plenty of opportunity imo.

I feel like if a Validator is trying to nominate something and cant continue only because of the lack of validators it should be able to let the network know :smiley: The existing code may not work as it is though, I havent looked in depth. If I am not mistaken, at that point there are enough transactions and block creation period is up.

AndrejMitrovic:
> I feel like if a Validator is trying to nominate something and cant continue only because of the lack of validators it should be able to let the network know

The SCP protocol already handles that. Leader selection is not just based on block height (slot index), but also on timers within the same block height. In SCP parlance there are slots (proposed block height), and within those slots there are multiple rounds of nominations. For each round, a new validator or set of validators become the leaders.

So there is no need to inform the network that there is a lack of validators. If they know the latest state of the blockchain (and they must know, otherwise they cannot participate in reaching consensus), then they will by definition know the state of the active enrollments and will know that new enrollments are required.

So this check can be entirely local within the node. That being said, we can't force a validator to enroll without the user's permission. So the automatic re-enrolling would need to be enabled.

AndrejMitrovic:
What we can do, however, is have some kind of monitoring service. This I agree would be beneficial.

But we can use the stats for that. Set up a stats alert if there are not enough enrollments, and then inform the community (or start our own nodes).

Note that I don't expect us to ever be in a situation where we don't have enough active validators in the system. The Foundation will host their own nodes. At least the minimum amount that's set by the protocol.

omerfirmak:
> then they will by definition know the state of the active enrollments and will know that new enrollments are required.

Yes, I realize that. But implementation wise, I don't see an existing event handler that we can use to check for lack of enrollments reliably and act. That's why we either would go with a periodic check or create a new message/event.

Regardless, I think the code in this PR will account for most of the cases. Probably even all cases if we can rely on a human operator :smiley: We can merge it if you are all happy with it as well.

AndrejMitrovic:
> create a new message/event.

But who would you be messaging?

omerfirmak:
> > create a new message/event.
> 
> But who would you be messaging?

It would be a broadcasted message I guess. Not enrolled validators that somehow could not realize the need for an enrollment before would simply respond by enrolling.

AndrejMitrovic:
>    Not enrolled validators that somehow could not realize the need for an enrollment before would simply respond by enrolling.

If they have recurring enrollments configured to true, then they should have automatically re-enrolled. If it's set to false, then they should not re-enroll without the user's permission.

Secondly, we are building a trustless blockchain. How do you verify that a node which broadcasts "We need enrollments" message is legitimate? The node could be lying.

omerfirmak:
>    If they have recurring enrollments configured to true, then they should have automatically re-enrolled. If it's set to false, then they should not re-enroll without the user's permission.

This is only useful when the node is not enrolled in the first place. Re-enrolling would happen automatically, if configured, anyways.

>    Secondly, we are building a trustless blockchain. How do you verify that a node which broadcasts "We need enrollments" message is legitimate? The node could be lying.

Yes, but I cant see any side effects it would have though. The receiver of the message can always check for the need of enrollments right? The message would simply be a trigger for the check and creation of the enrollment, instead of a periodic timer. Nothing more.

AndrejMitrovic:
> The receiver of the message can always check for the need of enrollments right?

Yes, but they can do so on their own - without the need for another party to inform them. I don't see a need to introduce more messaging to the system.

source/agora/node/Validator.d
```diff
@@ -419,8 +419,8 @@ public class Validator : FullNode, API
```
AndrejMitrovic, line 433: Commit message: never enrolled => not currently enrolled?

source/agora/common/Config.d
```diff
@@ -183,6 +183,9 @@ public struct ValidatorConfig
```
AndrejMitrovic, line 183: commit message: `the their enrollment expire` => `their enrollment expires`

source/agora/common/Config.d
AndrejMitrovic: This needs to be added in the documented config.

source/agora/common/Config.d
```diff
@@ -185,7 +185,7 @@ public struct ValidatorConfig
     public string registry_address;
 
     // If the enrollments will be renewed or not at the end of the cycle
-    public bool recurring_enrollment;
+    public bool recurring_enrollment = true;
```
AndrejMitrovic, line 188: This needs to be folded into the previous commit. The previous commit actually says it's off by default, but then the next commit immediately flips it. The default of a feature should be set in the commit which introduces the feature.

AndrejMitrovic:
I think the first 4 commits need to be squashed. They do not seem to be atomic commits (e.g. first commit uses a function which is introduced in a later commit).

omerfirmak:
> > The receiver of the message can always check for the need of enrollments right?
> 
> Yes, but they can do so on their own - without the need for another party to inform them. I don't see a need to introduce more messaging to the system.

The question is, where in the code they should do it. The reason I am not into periodic tasks is, once you have enough of them, they are a nightmare to maintain and debug. They introduce so much uncertainty into the system and you start to see bugs that happen only on Wednesdays which happen to be an odd day of the year and the year is divisible by 17  :smile:

AndrejMitrovic:
> The question is, where in the code they should do it. The reason I am not into periodic tasks is, once you have enough of them, they are a nightmare to maintain and debug. 

Yes. We should limit our timers. Preferably it could be just one fiber routine which does a set of invariant checks.

> They introduce so much uncertainty into the system and you start to see bugs that happen only on Wednesdays which happen to be an odd day of the year and the year is divisible by 17 😄

I feel like there's a really good story hiding in here. Maybe you could tell it to us one day over a set of 🍻?

source/agora/common/Config.d
omerfirmak: Where is that exactly? :) I think it should be added to the parsing from file routines as well.

AndrejMitrovic: This one: https://github.com/bpfkorea/agora/blob/v0.x.x/doc/config.example.yaml

omerfirmak:
All of the comments should be addressed now.

omerfirmak:
> > The question is, where in the code they should do it. The reason I am not into periodic tasks is, once you have enough of them, they are a nightmare to maintain and debug.
> 
> Yes. We should limit our timers. Preferably it could be just one fiber routine which does a set of invariant checks.
> 
> > They introduce so much uncertainty into the system and you start to see bugs that happen only on Wednesdays which happen to be an odd day of the year and the year is divisible by 17 smile
> 
> I feel like there's a really good story hiding in here. Maybe you could tell it to us one day over a set of beers?

I am all about embarrassing stories :smile:

omerfirmak:
Is there anything blocking this PR that I should take care of? Do we want to implement periodic enrollments?

Geod24 (approved):

Geod24:
Just need a rebase and we can merge

source/agora/node/Validator.d
```diff
@@ -423,4 +430,54 @@ public class Validator : FullNode, API
...
+        if (enroll_key == Hash.init && (enroll_key = this.getFrozenUTXO()) == Hash.init)
+            return; // Not enrolled and no frozen UTXO
+
+        const enrolled = this.enroll_man.getEnrolledHeight(enroll_key);
+
+        // This validators enrollment will expire next cycle or not enrolled at all
+        if (enrolled == ulong.max || block_height + 1 >= enrolled + this.params.ValidatorCycle)
+        {
+            log.trace("Enrolling at height {} for {} cycles with {}", block_height,
+                                            this.params.ValidatorCycle , enroll_key);
```
Geod24:
```suggestion
            log.trace("Sending Enrollment at height {} for {} cycles with {}", block_height,
                                            this.params.ValidatorCycle , enroll_key);
```

It's important to have the distinction of "sending stuff" vs things actually being applied.

omerfirmak:
Rebased

AndrejMitrovic:
Well done!

PR #1299, opened 2020-10-30
Call GC in-between tests

omerfirmak:
I have checked to see that only maximum of 100 fibers are ever in the system at the same time.
But GC is not always triggered in-between tests which results in GC memory keep accumulating.

This fixes the OOM crash I get in my system with 8G of physical memory.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -265,6 +265,8 @@ private UnitTestResult customModuleUnitTester ()
 
         while (parallel_tests.length)
         {
+            import core.memory;
+
             auto test = parallel_tests.front;
             parallel_tests.popFront();
 
@@ -272,6 +274,7 @@ private UnitTestResult customModuleUnitTester ()
             available_cores.wait();
 
             (new WorkThread(test)).start();
+            core.memory.GC.collect();
         }
     }
 


AndrejMitrovic:
I'm a little worried about the stop-the-world GC behavior. If the threads are paused then some of the tests might time out if the collection cycle is too long.

omerfirmak:
> I'm a little worried about the stop-the-world GC behavior. If the threads are paused then some of the tests might time out if the collection cycle is too long.

Calling GC once for each test could help with that as well. More we delay it, higher the risk of a long collection cycle, right?

AndrejMitrovic:
> Calling GC once for each test could help with that as well. More we delay it, higher the risk of a long collection cycle, right?

You're right. Anyway we can always increase the timeouts. And if the collection takes too long, it means we are allocating too much anyway.

ferencdg:
I am confused by this, because I think the runtime should call GC.collect before throwing OOM. I made some comment on slack about this, let me repeat it here:

"
I remember under Debian I had to increase the maximum GC memory for dlang through a command line argument, and then the tests ran normally. Under Ubuntu I never had this problem. 

It is more like a workaround until someone profiles the

btw I had 32GB memory and still got OOM without the command line parameter
"

ferencdg:
I java we have something called weak reference, so an object can become 'strongly unreachable' but that weak reference can be converted back to 'strong reference' and then it becomes 'strongly reachable' and cannot be GC'd anymore, I don't think D has weak references

ferencdg:
maybe you manually calling the GC caused somehow to reduce the number of parallel tests, and it is just by coincidence that you no longer get OOM

Geod24:
D's GC is derived from the [Boehm GC](https://en.wikipedia.org/wiki/Boehm_garbage_collector), which was developed for languages without support for GC (C/C++).
It has evolved quite a bit since its inception, but it's still far from fully utilizing the full type information it could.
Note that we could disable [this hack](https://github.com/bpfkorea/agora/blob/9097aa132ec62022635ef06355fb7174d27e9f41/source/agora/node/main.d#L35-L40) and try to run it concurrently.

omerfirmak:
> I am confused by this, because I think the runtime should call GC.collect before throwing OOM. I made some comment on slack about this, let me repeat it here:
> 
> "
> I remember under Debian I had to increase the maximum GC memory for dlang through a command line argument, and then the tests ran normally. Under Ubuntu I never had this problem.
> 
> It is more like a workaround until someone profiles the
> 
> btw I had 32GB memory and still got OOM without the command line parameter
> "

From the top of my head, in POSIX fibers allocate stack space with mmap. It may not be triggering GC.

ferencdg:
> > I am confused by this, because I think the runtime should call GC.collect before throwing OOM. I made some comment on slack about this, let me repeat it here:
> > "
> > I remember under Debian I had to increase the maximum GC memory for dlang through a command line argument, and then the tests ran normally. Under Ubuntu I never had this problem.
> > It is more like a workaround until someone profiles the
> > btw I had 32GB memory and still got OOM without the command line parameter
> > "
> 
> From the top of my head, in POSIX fibers allocate stack space with mmap. It may be not triggering GC.

what was the actual error message you got? core.exception.OutOfMemoryError thrown by new? Or you get a OOM when you called mmap?

omerfirmak:
Apparently, [this](https://github.com/ldc-developers/druntime/blob/4e4e61c8b993cedc612c2b184442aa908679a11e/src/core/thread/fiber.d#L1283) is where the druntime aborts. The root cause is too many Fibers being terminated but not deconstructed (ie garbage collected). Each fiber uses mmap to allocate a stack and creates a mapping, which is limited per process. These mappings are only destroyed when Fiber is deconstructed.

omerfirmak:
> Apparently, [this](https://github.com/ldc-developers/druntime/blob/4e4e61c8b993cedc612c2b184442aa908679a11e/src/core/thread/fiber.d#L1283) is where the druntime aborts. The root cause is too many Fibers being terminated but not deconstructed (ie garbage collected). Each fiber uses mmap to allocate a stack and creates a mapping, which is limited per process. These mappings are only destroyed when Fiber is deconstructed.

As expected, setting `vm.max_map_count` parameter in Linux to a higher value also fixes the issue.  `vm.max_map_count`  defaults to 65530 on my machine.

omerfirmak:
This wont be needed when https://github.com/Geod24/localrest/pull/103 is merged.

PR #1301, opened 2020-10-30
Make UTXOFinder accept the UTXO hash as parameter

Geod24:
The UTXOFinder should find UTXOs by their key,
which is the hash of the UTXO, instead of relying
on the decomposed key (tx hash + index).
This is a break of abstraction that is also present in `Input`,
and this commit is a first step towards fixing bpfkora/agora#1286 .

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -654,7 +654,7 @@ public class EnrollmentManager
         foreach (utxo_key; utxo_keys)
         {
             UTXOSetValue value;
-            if (!finder(utxo_key, size_t.max, value))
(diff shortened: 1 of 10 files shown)


AndrejMitrovic (approved):
Yessssssssssssssssssss!

linked0 (approved):
Great!!!
How about replacing all the `utxo_hash` with `utxo` in `UTXOSet.d`. There are not many 'utxo_hash' in the file.

Geod24:
> There are not many 'utxo_hash' in the file.

10 of them. That's originally what I did, but some places use `utxo` to refer to the actual output, so I decided to defer any cosmetic improvement for later.

PR #1302, opened 2020-10-30
Minor fixup for the admin interface

Geod24:
Following review, those were obvious low hanging fruits.

### DIFF
--- a/source/agora/node/admin/AdminInterface.d
+++ b/source/agora/node/admin/AdminInterface.d
@@ -125,9 +125,6 @@ public class AdminInterface
 
     public void stop ()
     {
-        if (!this.config.admin.enabled)
-            assert(0, "Admin interface is not enabled");
-
         log.trace("Stopping admin interface ...");
         this.listener.stopListening();
     }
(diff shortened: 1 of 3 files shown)


source/agora/node/admin/AdminInterface.d
```diff
@@ -125,9 +125,6 @@ public class AdminInterface
 
     public void stop ()
     {
-        if (!this.config.admin.enabled)
```
AndrejMitrovic, line 128: Also this check is already here: https://github.com/bpfkorea/agora/blob/9097aa132ec62022635ef06355fb7174d27e9f41/source/agora/node/Validator.d#L431-L432

AndrejMitrovic (approved):

TrustHenry (approved):
LGTM

source/agora/node/admin/AdminInterface.d
Geod24: Yep, there are a few other checks which could be simplified, but this one stood out.

PR #1306, opened 2020-11-01
Rename a few modules, shorten names, and add documentation

Geod24:
Some parts of the consensus package didn't make sense to me.
I hope the first 2 commits are no-brainers, and the third one will make sense to everyone.
The last one is a renamed of a name I find annoying, but if people prefer to keep the long name, or find another shorter name, suggestions welcome!

### DIFF
--- a/source/agora/consensus/README.md
+++ b/source/agora/consensus/README.md
@@ -1,0 +1,18 @@
+# Consensus package
+
+The consensus package is aimed to be usable as a library by client code.
+It contains the definitions of the data types that are exchanged and stored,
+as well as the state a full node must keep.
+
+This classification derives from framing the problem as [state machine replication](https://en.wikipedia.org/wiki/State_machine_replication).
+
+Following the state machine's definition, we associate the following components:
+- **States**: `ValidatorSet` & `UTXOSet` represent the node's view of the network;
+- **Inputs**: The content of the blocks (`Block`, `Enrollment`, `Transaction`);
+- **Outputs**: In our system, output is equal to state, so there's no distinction;
+- **Transition function**: Most of the mutation methods in `state` (e.g. applying transactions to `UTXOSet`) and the content of the `validation` package;
+- **Output function**: Like outputs, we don't have a separate representation for this;
+- **Start**: This is simply the genesis block.
+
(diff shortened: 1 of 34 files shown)


kchulj:
`UTXO` LGTM. Since it's a data structure of a UTXO..

source/agora/consensus/README.md
```diff
@@ -0,0 +1,18 @@
...
+It contains the definitions of the data types that are exchanged and stored,
+as well as the state a full node must keep.
+
+This classification derives from framing the problem as [state machine replication](https://en.wikipedia.org/wiki/State_machine_replication).
+
+Following the state machine's definition, we associate the following components:
+- **States**: `ValidatorSet` & `UTXOSet` represent the node's view of the network;
+- **Inputs**: The content of the blocks (`Block`, `Enrollment`, `Transaction`);
+- **Outputs**: In our system, output is equal to state, so there's no distinction;
+- **Transition function**: Most of the mutation method in `state` (e.g. applying transactions to `UTXOSet`) and the ccontent of the `validation` package;
```
AndrejMitrovic: method => methods
ccontent => content

Sorry doing a suggest is hard on mobile.

source/agora/consensus/README.md
```diff
@@ -0,0 +1,18 @@
...
+- **Output function**: Like outputs, we don't have a separate representation for this;
+- **Start**: This is simply the genesis block.
+
+Our **Inputs** also includes some specific items, such as `agora.consensus.data.Params`,
+which should be constant for the lifetime of the machine, but is used to tweak behavior in tests.
```
AndrejMitrovic, line 18: I think we might also use it for protocol upgrades in the future?

AndrejMitrovic (approved):
Just a nit.

source/agora/consensus/README.md
Geod24: If we do we can update the docs :P

source/agora/consensus/README.md
Geod24: Thanks, fixed

Geod24:
Linking issue should be fixed (but also shouldn't happen in the first place, LDC is being funky).

Geod24:
Green

PR #1307, opened 2020-11-01
Change Input hash to match UTXO's hash

Geod24:
```
There was a minor difference in how hashing was done: UTXOs were hashed using (Hash, ulong),
while the hash of an Input was done using (Hash, uint).
As a pre-requisite to fixing this, we change it to be consistent (using ulong).
```

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -1332,15 +1332,15 @@ unittest
 
     utxos.sort();  // must be sorted by enrollment key
     assert(man.getRandomSeed(utxos, Height(1)) ==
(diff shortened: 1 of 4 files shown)


hewison-chris (approved):
LGTM

AndrejMitrovic:
This is probably the second time we used the wrong type.

This is one of the reasons we should take care to avoid leaky abstractions. And also having strong typing (like `Height`) helps.

Thanks for catching this.

Geod24:
Well that was actually an error because `hash(a, b)` is obviously not `hash(hash(a, b))`...
On the other hand, I found a simple way to do the change I actually wanted to do, so 🤷

PR #1308, opened 2020-11-02
Remove unused 'coinbase' functions

Geod24:
They were only present in tests, but never used in production,
and simply an heritage from the very early days.


Geod24:
Cleaned up

Geod24:
Green

AndrejMitrovic (approved):

PR #1312, opened 2020-11-02
Demonstrate Schnorr multi-sig rogue-key attack

AndrejMitrovic:
This attack is possible when the parties in a multi-signature process do not prove ownership of their private keys.

Our block multi-sig scheme is secure because each node proves ownership of their key through the signing of the Enrollment. We do not have to change that scheme.

However, for regular users using the existing schnorr multi-sig scheme is insecure.

Resources:
- https://tlu.tarilabs.com/cryptography/digital_signatures/introduction_schnorr_signatures.html#key-cancellation-attack
- https://blockstream.com/2018/01/23/en-musig-key-aggregation-schnorr-signatures/#:~:text=not%20secure.

I strongly suggest that we look into some of the proposed secure multi-sig schemes, such as MuSig (there are other alternatives too).

Resources:
- https://tlu.tarilabs.com/cryptography/musig-schnorr-sig-scheme/The_MuSig_Schnorr_Signature_Scheme.html

### DIFF
--- a/source/agora/common/crypto/Schnorr.d
+++ b/source/agora/common/crypto/Schnorr.d
@@ -346,6 +346,42 @@ nothrow @nogc @safe unittest
     assert(stolen_key != kp.v);
 }
 
+// rogue-key attack
+// see: https://tlu.tarilabs.com/cryptography/digital_signatures/introduction_schnorr_signatures.html#key-cancellation-attack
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
Hopefully this will serve as good documentation in the future. I'll merge it since it doesn't introduce any new code.

PR #1313, opened 2020-11-02
Fix #1286: Store UTXO hash in Input, not txhash + index

### LINKED ISSUE #1286: Change `Input` to reference UTXO by their Hash, not Transaction Hash + index
Currently our `Input` sends `tx_hash` + index:
https://github.com/bpfkorea/agora/blob/5a9198093fe98451a973ad9b561df7887bd1a213/source/agora/consensus/data/Transaction.d#L132-L136

This is a breach of abstraction, as we rely explicitly on the way UTXO are indexed. Instead, we should just reference the UTXO's hash.

@AndrejMitrovic that might also be useful for you, would it not ? It would allow to provide a script as input instead of an UTXO, for example.

Geod24:
CC @kchulj

### DIFF
--- a/source/agora/consensus/data/Transaction.d
+++ b/source/agora/consensus/data/Transaction.d
@@ -30,7 +30,7 @@ import std.algorithm;
 public enum TxType : ubyte
 {
     Payment,
-    Freeze
+    Freeze,
 }
 
 /*******************************************************************************
--- a/source/agora/consensus/data/Transaction.d
+++ b/source/agora/consensus/data/Transaction.d
@@ -149,6 +149,12 @@ public struct Input
         this.utxo = hashMulti(txhash, index);
     }
 
+    /// Ctor which does hashing based on the `Transaction` and index
+    public this (in Transaction tx, ulong index) nothrow @safe
+    {
+        this.utxo = hashMulti(tx.hashFull(), index);
+    }
+
     /***************************************************************************
 
         Implements hashing support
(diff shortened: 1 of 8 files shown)


source/agora/consensus/data/Transaction.d
```diff
@@ -129,15 +129,26 @@ public struct Output
...
-    /// Index of the `Output` in the `previous` `Transaction`
-    public uint index;
+    /// The hash of the UTXO to be spent
+    public Hash utxo;
 
     /// A signature that should be verified using the `previous[index].address` public key
     public Signature signature;
 
+    /// Simple ctor
+    public this (in Hash utxo_, in Signature sig = Signature.init)
```
AndrejMitrovic, line 139: Wouldn't it make sense to just remove the `Signature` parameter from the ctor?

I'm not sure we can sign anything before the utxo is set?

source/agora/utils/Test.d
```diff
@@ -360,20 +348,41 @@ public struct TxBuilder
...
+            .array;
         tx.outputs.each!(outp => this.leftover.value.mustAdd(outp.value));
         return this;
     }
 
     /// Ditto
     public ref typeof(this) attach (const Transaction tx, uint index)
+        @safe nothrow return
+    {
+
```
AndrejMitrovic: 🚔

AndrejMitrovic (approved):
Just two comments.

source/agora/utils/Test.d
Geod24: Oh god

source/agora/consensus/data/Transaction.d
Geod24: Well since the signature is on the `Hash`, we can.
And the reason this ctor is here is because the deserializer needs it.

AndrejMitrovic: It's on the hash of the transaction, is it not?

But in any case this stuff might change in the future (SIGHASH, etc). So no worries for now.

Geod24:
Fixed the abomination.

source/agora/consensus/data/Transaction.d
Geod24: Ah right! Still, if you remove it, it doesn't compile 🤷

PR #1315, opened 2020-11-02
BitField.d - no need to serialize _storage field name

### LINKED ISSUE #1288: `BitField` JSON serialization should not include `_storage`
Similar to https://github.com/bpfkorea/agora/issues/1211 we apparently include `_storage` in the serialization ATM (found in Stoa).

ferencdg:
_storage is the only element of the struct, no need to
serialize the field name
fixes: #1288

### DIFF
--- a/source/agora/common/BitField.d
+++ b/source/agora/common/BitField.d
@@ -34,6 +34,8 @@ import agora.common.Serializer;
 import agora.common.Types;
 
 import std.algorithm;
+import std.array : array, join, split;
+import std.conv : to;
 import std.math;
 import std.traits;
 
@@ -76,8 +78,8 @@ public struct BitField (T = uint)
     ///
     private enum BitsPerT = (T.sizeof * 8);
 
-    /// Backing store: Should be private but Vibe.d cannot serialize it...
-    public T[] _storage;
+    /// Backing store
+    private T[] _storage;
 
     /***************************************************************************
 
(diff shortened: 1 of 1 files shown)


source/agora/common/BitField.d
```diff
@@ -176,6 +178,18 @@ public struct BitField (T = uint)
         }
     }
 
+    /// Overrides default vibe.d serialization
+    public string toString () const pure nothrow @safe
+    {
+        return "[" ~ _storage.map!(num_part => num_part.to!string).join(",") ~ "]";
```
AndrejMitrovic: `return this._storage.to!string` has the same effect here.

source/agora/common/BitField.d
```diff
@@ -176,6 +178,18 @@ public struct BitField (T = uint)
...
+    }
+
+    /// Ditto
+    public static BitField fromString (scope const(char)[] str) @safe
+    {
+        return BitField!T(str[1 .. $-1].split(",").map!(str_part => str_part.to!T).array);
```
AndrejMitrovic: This has the same effect:

```
return BitField!T(str.to!(T[]));
```

AndrejMitrovic (changes requested):

source/agora/common/BitField.d
ferencdg: I like this, but it only works if I remove the 'nothrow' declaration

AndrejMitrovic: Ah, strange. I think I'll file a bug for this.

AndrejMitrovic: In any case, it would be good to provide a sink version like suggested before: 

https://github.com/bpfkorea/agora/pull/1233#discussion_r493465721

AndrejMitrovic (approved):

AndrejMitrovic:
This will likely break Stoa again right? We should probably file an issue for it, just to keep track of these changes.

PR #1316, opened 2020-11-02
Implement additional methods for TestUTXOSet

kchulj:
The storage of the `TestUTXOSet` is able to be updated.
The `UTXO`s could only be added to the storage, and not be deleted.
Now, they are cleared if they are spent.

Improves [faucet](https://github.com/bpfkorea/faucet)

### DIFF
--- a/source/agora/consensus/data/UTXO.d
+++ b/source/agora/consensus/data/UTXO.d
@@ -141,6 +141,70 @@ public class TestUTXOSet
         }
         return false;
     }
+
+    private UTXO getUTXO (Hash utxo) nothrow @safe
+    {
+        UTXO value;
+        if (!this.findUTXO_(utxo, value))
+            assert(0);
+        return value;
+    }
+
+    public void updateUTXOStorage (const ref Transaction tx, Height height) @safe
+    {
+        import std.algorithm : any;
+
+        // defaults to next block
+        Height unlock_height = Height(height + 1);
+
+        // for payments of frozen transactions, it will melt after 2016 blocks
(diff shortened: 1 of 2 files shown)


source/agora/consensus/data/UTXO.d
```diff
@@ -141,6 +141,68 @@ public class TestUTXOSet
...
         return false;
     }
+
+    private UTXO getUTXO (Hash utxo) nothrow @safe
+    {
+        UTXO value;
+        if (!this.findUTXO_(utxo, value))
+            assert(0);
+        return value;
+    }
```
Geod24, line 151: This is such a small wrapper it shouldn't be its own function

Geod24:
I wonder if we should have a base class for those two ?

kchulj:
> I wonder if we should have a base class for those two ?

Makes sense, since the methods are nearly identical.
Would have to overload it with `ManagedDatabase` and AA somehow.

source/agora/consensus/data/UTXO.d
```diff
@@ -141,6 +141,68 @@ public class TestUTXOSet
...
+                )
+            )
+        )
+        {
+            unlock_height = Height(height + 2016);
+        }
+
+        foreach (const ref input; tx.inputs)
+        {
+            this.storage.remove(input.utxo);
```
AndrejMitrovic, line 175: I would add an assert here making sure the utxo exists. It will help catch bugs.

kchulj:
@AndrejMitrovic what do you think about an abstract base class?

kchulj:
Will close this and instead inherit from the base class in #1318 .

PR #1318, opened 2020-11-03
Implement base class for UTXOSet

kchulj:
The methods for `UTXOSet` and `TestUTXOSet` are nearly identical.
The base class has been created so that the code becomes simpler.

Relates [faucet#10](https://github.com/bpfkorea/faucet/issues/10)


kchulj:
The CI is now green.
However, moving `findUTXO` to the base class is causing an assertion error for some reason.

kchulj:
What's a better name for the base class : `Set` or `Cache` ?

source/agora/consensus/state/UTXOSet.d
```diff
@@ -165,8 +165,8 @@ public class UTXOSet
         Find an UTXO in the UTXO set.
 
         Params:
-            hash = the hash of the UTXO (`hashFull(tx_hash, index)`)
-            output = will contain the UTXO if found
+            hash = the hash of the UTXO (`hashMulti(tx_hash, index)`)
+            value = the UTXO to find
```
Geod24: That change is wrong, it's an `out` parameter.

source/agora/consensus/state/UTXOSet.d
```diff
@@ -27,42 +27,11 @@ import std.file;
 
 mixin AddLogger!();
 
-/// ditto
-public class UTXOSet
+abstract class UTXOCache
```
Geod24, line 39: Missing documentation and visibility

source/agora/consensus/data/UTXO.d
```diff
@@ -53,6 +53,133 @@ public struct UTXO
...
 }
 
+/*******************************************************************************
+
+    This is the the base class of a UTXO storage.
+    It is currently inherited by UTXOSet and TestUTXOSet.
+
+*******************************************************************************/
+
+abstract class UTXOCache
```
Geod24: First thing, don't add code in one commit just to move it all in the next commit.
Second, instead of moving `UTXOCache` in the `data` package, we should move `TestUTXOSet` next to `UTXOCache`.

source/agora/consensus/data/UTXO.d
```diff
@@ -141,6 +265,24 @@ public class TestUTXOSet
...
+    protected override void add (Hash utxo, UTXO value) @safe
+    {
+        this.storage[utxo] = value;
+    }
+
+    /// Ditto
+    protected override bool find (Hash utxo, UTXO value) nothrow @safe
+    {
+        return this.peekUTXO(utxo, value);
+    }
```
Geod24: Remove `find` and just expose `peekUTXO` in both cases.

Geod24 (changes requested):

source/agora/consensus/data/UTXO.d
```diff
@@ -53,6 +53,133 @@ public struct UTXO
...
+    protected Set!Hash used_utxos;
+
+    /***************************************************************************
+
+        Add all of a transaction's outputs to the Utxo set,
+        and remove the spent outputs in the transaction from the set.
+
+        Params:
+            tx = the transaction
+            height = Height of the block where UTXO will be unlocked
```
AndrejMitrovic: Isn't this the wrong documentation? Because in the body of the function `+1` is used. Therefore `height` is just the current block height is it not?

source/agora/consensus/data/UTXO.d
kchulj: I used `peekUTXO` for `TestUTXOSet` since it doesn't have its own `find` and `peekUTXO` is doing the same thing. However, `UTXOSet` doesn't have the `peekUTXO` method.

As you suggested, it may the best to remove the `getUTXO` in the base class ? (it's only used there)

Geod24: I think the real UTXO should also have `peekUTXO`

source/agora/consensus/data/UTXO.d
kchulj: Good catch, it's the current height.

source/agora/consensus/data/UTXO.d
kchulj: What do you think about renaming the current `find` overload to `peekUTXO`? It is calling this in `UTXODB`.

https://github.com/bpfkorea/agora/blob/7a7a5b186443de3b2bc676ff08247f67bdb98551/source/agora/consensus/state/UTXOSet.d#L263-L276

Geod24: Yes that was the suggestion :)

source/agora/consensus/state/UTXOSet.d
```diff
@@ -137,18 +117,83 @@ public class UTXOSet
 
     ***************************************************************************/
 
-    public UTXOFinder getUTXOFinder () @trusted nothrow
+    public abstract UTXOFinder getUTXOFinder () nothrow @trusted;
```
Geod24: Why abstract ? The implementation is the same, isn't it ?

source/agora/consensus/state/UTXOSet.d
```diff
@@ -70,7 +50,7 @@ public class UTXOSet
 
         Params:
             tx = the transaction
-            height = Height of the block where UTXO will be unlocked
+            height = current height of the block
```
Geod24: No, the documentation was correct. I tried to avoid the "current Height" as much as possible in docs, because it promotes the idea that we know the state of the network. You **never** know the current `Height`. You have a vision of the network, and perhaps you know what `Height` other nodes should be at, but you don't actually know (even if you ask them, they could lie).
Also, imagine you are doing catch-up from genesis, and there are 1000 blocks to catch-up. The first time you call this function, with a `Height == 1`, is it correct to say "current Height" ?

source/agora/consensus/state/UTXOSet.d
```diff
@@ -165,7 +210,7 @@ public class UTXOSet
         Find an UTXO in the UTXO set.
 
         Params:
-            hash = the hash of the UTXO (`hashFull(tx_hash, index)`)
+            hash = the hash of the UTXO (`hashMulti(tx_hash, index)`)
```
Geod24, line 182: Good catch

source/agora/consensus/state/UTXOSet.d
```diff
@@ -190,6 +235,130 @@ public class UTXOSet
         log.trace("findUTXO: utxo_hash {} not found", utxo);
         return false;
     }
+
+    /// Ditto
```
Geod24: Those are not `/// Ditto`, since `/// Ditto` means "same as last symbol". Just `///` to make sure they are documented is enough.

source/agora/consensus/state/UTXOSet.d
```diff
@@ -190,6 +235,130 @@ public class UTXOSet
...
+    {
+        this.utxo_db[utxo] = value;
+    }
+}
+
+/*******************************************************************************
+
+    This is a simple UTXOSet, used when the AA behavior is desired
+
+    Most unittestsdo not need a full-fledged UTXOSet with all the DB and
```
Geod24:
```suggestion
    Most unittests do not need a full-fledged `UTXOSet` with all the DB and
```

Also, `fully-fledged` ? CC @hewison-chris .

Geod24 (commented):
Nice. I think you should just squash the last 2 commits together.

source/agora/consensus/state/UTXOSet.d
kchulj: Yes. I tried to implement it in the base class and making `findUTXO` abstract, but it caused too many errors. Probably due to the delegate and alias.
Currently, `UTXOSet` uses `findUTXO` and `TestUTXOSet` uses `findUTXO_`. They seemed to be overloadable but  required code changes in other parts.

source/agora/consensus/state/UTXOSet.d
AndrejMitrovic: Alright you can name it "local height". But the docs still lie. The utxo will be unlocked either at the next height or after unfreeze. Docs need fixing.

source/agora/consensus/state/UTXOSet.d
hewison-chris: British is fully fledged :-)

source/agora/consensus/state/UTXOSet.d
```diff
@@ -156,41 +234,98 @@ public class UTXOSet
...
+            };
+            this.storage[h] = v;
         }
+    }
+
+    /// Workaround 20559...
+    public void clear ()
+    {
+        this.storage.clear();
+    }
```
kchulj: Why is this a workaround ?

Geod24: https://issues.dlang.org/show_bug.cgi?id=20559

But I fixed it and it's not needed since v2.090.1.

Geod24 (approved):
LGTM!

source/agora/consensus/state/UTXOSet.d
kchulj: Oh ok. I'll remove this comment then.

Geod24:
```
Linking...
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite38internal4util6errmsgFNbPSQBk7sqlite3QiZAya: error: undefined reference to 'sqlite3_errmsg'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite38internal4util6errmsgFNbPSQBk7sqlite312sqlite3_stmtZAya: error: undefined reference to 'sqlite3_db_handle'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite37results11ResultRange6__ctorMFNcSQBr9statement9StatementZSQCqQCjQCe: error: undefined reference to 'sqlite3_step'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite37results11ResultRange6__ctorMFNcSQBr9statement9StatementZSQCqQCjQCe: error: undefined reference to 'sqlite3_column_count'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite37results11ResultRange8popFrontMFZv: error: undefined reference to 'sqlite3_step'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite37results3Row__T4peekTmZQiMFmZm: error: undefined reference to 'sqlite3_column_int64'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite37results3Row__T4peekTAhVEQBjQBc8PeekModei0ZQBcMFmZQBd: error: undefined reference to 'sqlite3_column_blob'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite37results3Row__T4peekTAhVEQBjQBc8PeekModei0ZQBcMFmZQBd: error: undefined reference to 'sqlite3_column_bytes'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite39statement9Statement7Payload6__dtorMFNbZv: error: undefined reference to 'sqlite3_finalize'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite39statement9Statement6__ctorMFNcSQBq8database8DatabaseAyaZSQCqQCjQCc: error: undefined reference to 'sqlite3_prepare_v2'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite39statement9Statement6__ctorMFNcSQBq8database8DatabaseAyaZSQCqQCjQCc: error: undefined reference to 'sqlite3_bind_parameter_count'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite39statement9Statement__T4bindTAhZQjMFiQiZv: error: undefined reference to 'sqlite3_bind_blob64'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite39statement9Statement__T4bindTAxhZQkMFiQjZv: error: undefined reference to 'sqlite3_bind_blob64'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite38database8Database7Payload6__dtorMFNbZv: error: undefined reference to 'sqlite3_progress_handler'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite38database8Database7Payload6__dtorMFNbZv: error: undefined reference to 'sqlite3_close'
tests/system/.dub/build/application-debug-linux.posix-x86_64-ldc_2094-2615C5A75083FE1BD2080603CD118633/systemtest-simple.o:basic_auth_client.d:function _D9d2sqlite38database8Database5closeMFZv: error: undefined reference to 'sqlite3_close'
```

The failure is not spurious :)

source/agora/consensus/state/UTXOSet.d
```diff
@@ -137,59 +117,214 @@ public class UTXOSet
...
+    public void put (const Transaction tx) nothrow @safe
+    {
+        Hash txhash = hashFull(tx);
+        foreach (size_t idx, ref output_; tx.outputs)
         {
-            log.trace("findUTXO: utxo_hash {} found in used_utxos: {}", utxo, used_utxos);
-            return false;  // double-spend
+            Hash h = UTXO.getHash(txhash, idx);
+            UTXO v = {
+                type: tx.type,
```
AndrejMitrovic, line 222: Something I have noticed with the `TestUTXOSet` is that it doesn't store the `unlock_height`. Could we add it as a parameter which defaults to 0?

Geod24: Good point. Although we now have `updateUTXOCache` through the base class which would do that for you, so I think we could just remove this method in the long run ?

AndrejMitrovic: Ah right. Yes then it makes sense to just remove it, otherwise it's duplicating the functionality (poorly).

kchulj: Correct, `updateUTXOCache` can now be called with `TestUTXOSet`. I was thinking of some use-cases where we would want to "stack" the UTXOs (not remove any).
If there isn't such a case, then we don't need it.

kchulj:
:white_check_mark:

Geod24 (approved):
Nice

source/agora/consensus/state/UTXOSet.d
kchulj: I tried to replace occurrences of `put` with `updateUTXOCache`, but it doesn't seem to be a simple replace.
Could it be the fact that `put` doesn't have `unlock_height`
```D
UTXO v = {
    type: tx.type,
    output: output_
};
```
and `updateUTXOCache` takes into account the `unlock_height`?

AndrejMitrovic: Setting it to `0` should work then, no?

PR #1320, opened 2020-11-04
Define a new type for storing data for use in `Transaction`

MichaelKim20:
I defined a new type to store raw data in Transaction.

Added computeHash for hash.
Added fromString, toString for JSON serialization, deserialization
Added fromBinary for deserialization
Added new encoding and decoding of the hexadecimal string
Added a way to handle errors that can occur when decoding the hexadecimal string

The maximum size has not been determined, but it will be checked by validation.

Relates to #1214

### DIFF
--- a/source/agora/consensus/data/DataPayload.d
+++ b/source/agora/consensus/data/DataPayload.d
@@ -1,0 +1,214 @@
+/*******************************************************************************
+
+    Contains a type for storing data in a transaction
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.DataPayload;
+
+import agora.common.Hash;
+import agora.common.Serializer;
+
+import std.algorithm.iteration;
+import std.conv: parse;
+import std.range;
+
+/// The structure of for storing data in a transaction
+public struct DataPayload
+{
+    /// The byte array of transaction data
+    public const(ubyte)[] data;
+
+    /***************************************************************************
+
+        Create a DataPayload from binary data
+
+        Params:
+            bin  = Binary data to store in this `DataPayload`
+
+    ***************************************************************************/
+
+    public this (const(ubyte)[] bin) pure nothrow @safe
+    {
+        this.data = bin;
+    }
+
+    /// Print `DataPayload`
+    public void toString (scope void delegate(const(char)[]) @safe sink) const @safe
+    {
+        if (this.data.length == 0)
+            return;
+
+        auto toHexDigit = (ubyte value) @safe nothrow @nogc
+        {
+            return cast(char)(value + ((value < 10) ? 0x30 : 0x57));
+        };
+
+        sink("0x");
+        char[2] hex;
+        this.data.each!(
+            (num)
+            {
+                hex[0] = toHexDigit(num >> 4);
+                hex[1] = toHexDigit(num & 0xF);
+                sink(hex);
+            }
+        );
+    }
+
+    /// Support for Vibe.d serialization
+    public string toString () const @safe
+    {
+        if (this.data.length == 0)
+            return "";
+
+        size_t idx;
+        char[] buffer;
+        buffer.length = this.data.length * 2 + 2;
+        scope sink = (const(char)[] v) {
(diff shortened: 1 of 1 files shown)


source/agora/consensus/data/TransactionData.d
```diff
@@ -0,0 +1,297 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.data.TransactionData;
```
AndrejMitrovic: How about calling it `DataPayload`? We now have `data.Transaction` and `data.TransactionData` and it's very confusing.

source/agora/consensus/data/TransactionData.d
```diff
@@ -0,0 +1,297 @@
...
+
+/// The structure of for storing data in a transaction
+public struct TransactionData
+{
+    /// The byte array of transaction data
+    public ubyte[] data;
+
+    /// The error for fromString(),
+    /// This will be checked in the validation of the transaction.
+    public string error;
```
AndrejMitrovic: Errors should not be embedded in the structures that will be stored in the UTXO set or in the blockchain.

source/agora/consensus/data/TransactionData.d
```diff
@@ -0,0 +1,297 @@
...
+    }
+
+    /// Support for comparison (rvalue overload)
+    public int opCmp (const typeof(this) s) const nothrow @safe @nogc
+    {
+        return this.opCmp(s);
+    }
+
+    /// Returns the number of elements in the data
+    @property size_t length() const pure nothrow @safe @nogc
```
AndrejMitrovic: `this.data` is already public, so why do you need this getter function?

source/agora/consensus/data/TransactionData.d
```diff
@@ -0,0 +1,297 @@
...
+    ***************************************************************************/
+
+    public static TransactionDataT fromBinary (TransactionDataT = TransactionData) (
+        scope DeserializeDg dg, const ref DeserializerOptions opts) @safe
+    {
+        return TransactionData(deserializeFull!(ubyte[])(dg, opts));
+    }
+
+    /// Support for comparison
+    public int opCmp (ref const typeof(this) s) const nothrow @safe @nogc
```
AndrejMitrovic: Why do we need comparison support?

AndrejMitrovic (changes requested):

source/agora/consensus/data/TransactionData.d
MichaelKim20: Yes, I will fix it

source/agora/consensus/data/TransactionData.d
MichaelKim20: There is no particular reason, so I will remove it

source/agora/consensus/data/TransactionData.d
MichaelKim20: There is no particular reason, so I will remove it

source/agora/consensus/data/TransactionData.d
MichaelKim20: I changed the function fromString to solve this problem.

source/agora/consensus/data/DataPayload.d
```diff
@@ -0,0 +1,215 @@
...
+        Params:
+            bin  = Binary data to store in this `DataPayload`
+
+    ***************************************************************************/
+
+    public this (scope const ubyte[] bin) pure nothrow @safe
+    {
+        this.data.length = bin.length;
+        this.data[] = bin[];
+    }
```
AndrejMitrovic: How about changing the data to `const`? After all, the data should never be modified after its read. And this way you avoid having to duplicate the array:

```suggestion
    /// The byte array of transaction data
    public const(ubyte)[] data;

    /***************************************************************************

        Create a DataPayload from binary data

        Params:
            bin  = Binary data to store in this `DataPayload`

    ***************************************************************************/

    public this (const(ubyte)[] bin) pure nothrow @safe
    {
        this.data.bin;
    }
```

MichaelKim20: That's a good amendment. I changed it.

AndrejMitrovic (approved):

PR #1321, opened 2020-11-04
Implement a formula for calculating fee for data storage

MichaelKim20:
I implemented a formula for calculating the fee for data storage

I separated from PR #1258 to deal with the matter of fee in more detail.

The size of the voting data is about 637 bytes when expressed in JSON, and about 320 bytes when expressed in binary.

I designed the formula to increase the fee exponentially as the data grows.
And I set up a factor so that I could adjust the curvature of the curve. (The smaller the factor in the formula, the greater the curvature.)
I suggest adding this factor to ConsnesusParams.

A graph of the formula according to the factor is shown below.

Relates to #1214

### DIFF
--- a/source/agora/node/Fee.d
+++ b/source/agora/node/Fee.d
@@ -1,0 +1,157 @@
+/*******************************************************************************
+
+    Contains the function to calculate the fees used to store the data
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.Fee;
+
+import agora.common.Amount;
+
+import std.math;
+
+/*******************************************************************************
+
+    Calculates the fee of transaction data to store
+
+    Params:
+        data_size = The size of the data
+        factor = The factor to calculate for the fee of transaction data
+
+    Returns:
+        A fee payable to data storage
+
+*******************************************************************************/
+
+public Amount calculateDataFee(ulong data_size, uint factor) pure nothrow @safe @nogc
+{
+    const ulong decimal = 100L;
+    ulong fee = cast(ulong)(
+        round(
+            (
+                exp((cast(double)data_size / cast(double)factor)) -
+                1.0
+            ) *
+            decimal
+        ) * 10_000_000L / decimal
+
+    );
+    return Amount(fee);
+}
+
+// Test for calculating a fee to store data
+unittest
+{
+    // When the factor is 100
+    assert(calculateDataFee(0, 100) == Amount(0L));
+    assert(calculateDataFee(10, 100) == Amount(1_100_000L));
+    assert(calculateDataFee(20, 100) == Amount(2_200_000L));
+    assert(calculateDataFee(30, 100) == Amount(3_500_000L));
+    assert(calculateDataFee(40, 100) == Amount(4_900_000L));
+    assert(calculateDataFee(50, 100) == Amount(6_500_000L));
+    assert(calculateDataFee(60, 100) == Amount(8_200_000L));
+    assert(calculateDataFee(70, 100) == Amount(10_100_000L));
+    assert(calculateDataFee(80, 100) == Amount(12_300_000L));
+    assert(calculateDataFee(90, 100) == Amount(14_600_000L));
+    assert(calculateDataFee(100, 100) == Amount(17_200_000L));
+    assert(calculateDataFee(110, 100) == Amount(20_000_000L));
+    assert(calculateDataFee(120, 100) == Amount(23_200_000L));
+    assert(calculateDataFee(130, 100) == Amount(26_700_000L));
+    assert(calculateDataFee(140, 100) == Amount(30_600_000L));
+    assert(calculateDataFee(150, 100) == Amount(34_800_000L));
+    assert(calculateDataFee(160, 100) == Amount(39_500_000L));
+    assert(calculateDataFee(170, 100) == Amount(44_700_000L));
+    assert(calculateDataFee(180, 100) == Amount(50_500_000L));
+    assert(calculateDataFee(190, 100) == Amount(56_900_000L));
+    assert(calculateDataFee(200, 100) == Amount(63_900_000L));
+    assert(calculateDataFee(210, 100) == Amount(71_700_000L));
+    assert(calculateDataFee(220, 100) == Amount(80_300_000L));
+    assert(calculateDataFee(230, 100) == Amount(89_700_000L));
+    assert(calculateDataFee(240, 100) == Amount(100_200_000L));
+    assert(calculateDataFee(250, 100) == Amount(111_800_000L));
+    assert(calculateDataFee(260, 100) == Amount(124_600_000L));
+    assert(calculateDataFee(270, 100) == Amount(138_800_000L));
+    assert(calculateDataFee(280, 100) == Amount(154_400_000L));
+    assert(calculateDataFee(290, 100) == Amount(171_700_000L));
+    assert(calculateDataFee(300, 100) == Amount(190_900_000L));
+    assert(calculateDataFee(310, 100) == Amount(212_000_000L));
+    assert(calculateDataFee(320, 100) == Amount(235_300_000L));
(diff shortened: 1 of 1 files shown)


source/agora/consensus/data/Transaction.d
```diff
@@ -247,3 +248,44 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
...
+    Params:
+        data_size = The size of the data
+        factor = The factor to calculate for the fee of transaction data
+
+    Return:
+        A fee payable to data storage
+
+*******************************************************************************/
+
+public Amount clculateDataFee(ulong data_size, uint factor) @safe nothrow
```
AndrejMitrovic:
```suggestion
public Amount calculateDataFee (ulong data_size, uint factor) @safe nothrow
```

source/agora/consensus/data/Transaction.d
```diff
@@ -247,3 +248,44 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
             return false;
     return true;
 }
+
+/*******************************************************************************
```
AndrejMitrovic: This function shouldn't live in this module. The `consensus/data/*` modules should only contain data definitions.

Technically the fees are only for the Validators. So maybe it should live somewhere close to the `Validator.d` module. Perhaps a `Fee.d` module would work.

source/agora/consensus/data/Transaction.d
```diff
@@ -247,3 +248,44 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
...
+            decimal
+        ) *
+        10_000_000L
+    );
+    return Amount(fee);
+}
+
+unittest
+{
+    // When the size of a data is 0 bytes the fee is 0 BOA
```
AndrejMitrovic: These are not sufficient tests. Try to test on a range of values. For example using power of two could work.

For example:

```D
clculateDataFee(0, 200);
clculateDataFee(2, 200);
clculateDataFee(4, 200);
clculateDataFee(8, 200);
clculateDataFee(16, 200);
...
```

and so on.. And also with a different factor too.

source/agora/consensus/data/Transaction.d
```diff
@@ -247,3 +248,44 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
             return false;
     return true;
 }
+
+/*******************************************************************************
+
+    Caculates the fee of transaction data to store
```
AndrejMitrovic: I think this function should not calculate the fees, but only a **multiplier**. 

Think about the following: if BOA is worth 1$ today, but 100$ next year, do you really want the fee to be hardcoded into the protocol?

Fees may be dynamic and also entirely up to the validator nodes. But it's good to have a helper function like this to help most validators pick the fees.

source/agora/consensus/data/Transaction.d
AndrejMitrovic: It can also be `@nogc` and maybe even `pure`.

source/agora/consensus/data/Transaction.d
```diff
@@ -247,3 +248,44 @@ public bool getSumOutput (const Transaction tx, ref Amount acc)
...
+
+    Params:
+        data_size = The size of the data
+        factor = The factor to calculate for the fee of transaction data
+
+    Return:
```
AndrejMitrovic:
```suggestion
    Returns:
```

AndrejMitrovic (changes requested):

source/agora/consensus/data/Transaction.d
MichaelKim20: There was something like this on the issue.
```
### High level requirements:

- It must come at a cost, which might not be linear (to allow small packet of data but forbid images & co, at least in a contiguous chunk);

### Suggested approach

5. If those rules are satisfied, the transaction can contain a certain amount of data, of which the size is dependent on the amount in the output towards the common budget;
```

If this value is determined by the validator's node, how can transaction validation be done?
For your information, I have already implemented code in PR #1258 that validates according to the amount of transaction data. Please check.

MichaelKim20: I'm trying to save the factor in consensusParams to compensate for the variation in the value of BOA. How is it?

AndrejMitrovic: Alright, I misunderstood what was written.

I'm still not sure how this can work by being hardcoded. But you don't have to change your code.

source/agora/node/Fee.d
```diff
@@ -0,0 +1,157 @@
...
+
+import std.math;
+
+/*******************************************************************************
+
+    Calculates the fee of transaction data to store
+
+    Params:
+        data_size = The size of the data
+        factor = The factor to calculate for the fee of transaction data
```
AndrejMitrovic, line 26: So when the factor is bigger, the fee is lower? That's what the tests are suggesting.

What is the "factor" here? What values can be accepted (which range)?

MichaelKim20: https://github.com/bpfkorea/agora/pull/1321#issue-515139655
We can see from the graph.

AndrejMitrovic (approved):

AndrejMitrovic:
I think a good next issue to work on would be to determine if the fee in a transaction is large enough.

The fee in the transaction is all the output sum minus all the input sum. This is the unspent amount.

`Transaction.isInvalidReason` already calculates this:

https://github.com/bpfkorea/agora/blob/a943507c2b333668ebe2b55703048c99f9ff660f/source/agora/consensus/validation/Transaction.d#L132

`sum_unspent` will contain the fee. I think it would be good to add a delegate parameter to `isInvalidReason` and call it with the fee. Something like:

```D
public string isInvalidReason (
    in Transaction tx, scope UTXOFinder findUTXO, in Height height,
    bool delegate (Amount) checkFee)
```

And then check:

```D
if (!checkFee(tx, sum_unspent)) return "Transaction: Not enough fees";
```

I think this approach could work?

source/agora/node/Fee.d
```diff
@@ -0,0 +1,157 @@
...
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.Fee;
```
Geod24, line 14: This is a consensus-critical module, why is it here ??

source/agora/node/Fee.d
```diff
@@ -0,0 +1,157 @@
...
+    Params:
+        data_size = The size of the data
+        factor = The factor to calculate for the fee of transaction data
+
+    Returns:
+        A fee payable to data storage
+
+*******************************************************************************/
+
+public Amount calculateDataFee(ulong data_size, uint factor) pure nothrow @safe @nogc
```
Geod24, line 33: Style

source/agora/node/Fee.d
```diff
@@ -0,0 +1,157 @@
...
+    ulong fee = cast(ulong)(
+        round(
+            (
+                exp((cast(double)data_size / cast(double)factor)) -
+                1.0
+            ) *
+            decimal
+        ) * 10_000_000L / decimal
+
+    );
```
Geod24, line 45: This is a landmine. We shouldn't use floating point for this. If we need a higher precision, use the same technique as in `Amount`, a fixed precision based on integer arithmetic.

source/agora/node/Fee.d
```diff
@@ -0,0 +1,157 @@
...
+    assert(calculateDataFee(430, 100) == Amount(727_000_000L));
+    assert(calculateDataFee(440, 100) == Amount(804_500_000L));
+    assert(calculateDataFee(450, 100) == Amount(890_200_000L));
+    assert(calculateDataFee(460, 100) == Amount(984_800_000L));
+    assert(calculateDataFee(470, 100) == Amount(108_9500_000L));
+    assert(calculateDataFee(480, 100) == Amount(1_205_100_000L));
+    assert(calculateDataFee(490, 100) == Amount(1_332_900_000L));
+    assert(calculateDataFee(500, 100) == Amount(1_474_100_000L));
+
+    // When the factor is 200
```
Geod24, line 105: Why do we test every 10 values ? That sounds very superfluous, and as soon as this will be updated, it'll be even less fun than the quorum tests.

Issue #1076, opened 2020-07-29
Design & implement slashing rules for missing pre-images


Geod24:
Currently, we are building rules based on the availability of pre-images: signature scheme and quorum balancing.
We also have a timer in place to ensure our pre-images are available long before we need them.
Now we need to add a way to detect and log whenever pre-images are missing, decide on a policy of what to do with it, and implement it.

### Detection

The first step IMO is to detect whenever we are missing a pre-image. This will allow us to make sure we have everything properly implemented at the moment. Once this is the case, we can move on to writing a test case.
That test case should have at least one node in the network doing the following:
- (A) Never send any pre-image after its initial enrollment;
- (B) Only send one pre-image after enrolling then no more;
- (C) Send pre-image for half of its cycle (potentially in one burst) then stop sending pre-image;
- (D) Send pre-image for half its cycle, stop it for a quarter of the cycle, resume until the end;
Those 4 situations should be tested *independently*, so not during the same cycle, and not using the same utxo key / public key.
This will allow us to see how the network currently behaves when faced with those situations, and correct any remaining issue.

### Policy decision

Once we have that basis in place, we will need to come up with the slashing policy for missing pre-images.
As discussed in the previous meetings, missing pre-images is one of the gravest mistakes a validator can commit, short of outright hostile actions (such as signing two blocks for the same height).

This will need to be further discussed with the team, but currently my mind is on two things:
- gradual punishment for validators missing to relay their pre-image, so that they still have an incentive to come back and propagate them if they miss a deadline. The more pre-images, the safer the randomness function.
- applying a loss-of-income penalty to the quorum: this gives a strong incentive to the quorum to *keep and propagate* pre-images for nodes in their quorum, ensuring that they will not simply ignore them to maximize the penalty of a validator;

### Implementation

Once we have the basic tests cases, and decided on a policy, we  can move to an implementation. I expect the policy decision process to yield more test cases which we can implement.

AndrejMitrovic:
> The more pre-images, the safer the randomness function.

One thing I still don't follow w.r.t. randomness: If all the preimages are revealed in advance then anyone can calculate the random value for any future block height (unless a new enrollment was added in the meantime).

How do we account for that? Is there another step that's missing?

Geod24:
> If all the preimages are revealed in advance

[...]

PR #1328, opened 2020-11-05
[WIP] Network test detecting whenever pre-images are missing

linked0:
This is the first step of #1076, which is a network test for detecting the situation where we are missing a pre-image. This contains 4 situations as follows. These are simple test cases, but we could make test cases for more complex situations like the validators having a different cycle. 

(A) Never send any pre-image after its initial enrollment;
(B) Only send one pre-image after enrolling then no more;
(C) Send pre-image for half of its cycle (potentially in one burst) then stop sending pre-image;
(D) Send pre-image for half its cycle, stop it for a quarter of the cycle, resume until the end;

Part of #1076

### DIFF
--- a/source/agora/test/MissingPreImageDetection.d
+++ b/source/agora/test/MissingPreImageDetection.d
@@ -1,0 +1,592 @@
+/*******************************************************************************
+
+    Contains tests for the validators not revealing their pre-images. There
+    are four cases that the validators do not reveal as follows.
+
+    (A) Never send any pre-image after its initial enrollment
+    (B) Only send one pre-image after enrolling then no more
+    (C) Send pre-image for half of its cycle (potentially in one burst)
+        then stop sending pre-image
+    (D) Send pre-image for half its cycle, stop it for a quarter of the cycle,
+        resume until the end
+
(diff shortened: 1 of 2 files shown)


source/agora/test/MissingPreImageDetection.d
```diff
@@ -0,0 +1,597 @@
...
+        The new transactions generated
+
+*******************************************************************************/
+
+private Transaction[] makeTransactions (Transaction[] old_txs, PublicKey address)
+{
+    return old_txs
+        .length.iota.map!(idx => TxBuilder(old_txs[idx], cast(uint)0))
+        .map!(txb => txb.refund(address).sign())
+        .array;
```
Geod24: That's a very convoluted way of writing:
```suggestion
    return old_txs.map!(tx => TxBuilder(tx, 0).refund(address).sign()).array();
```

Although I'm not sure why you only want to spend a single output ?

linked0: Will update. No special reason for a single output. But, I will update it also after referring to other codes, If needed. Ty.

PR #1330, opened 2020-11-05
Make UTXO-sensible tests use either `peekUTXO` or `getUTXOFinder`

kchulj:
The `alias` has been removed and the function `findUTXO_` has been renamed to `findUTXO`.
The alias was calling only `peekUTXO`. `findUTXO` has double-spend protection, whereas `peekUTXO` does not. In the unittests, these two functions should be called according to their purpose.

Improves #1318

### DIFF
--- a/source/agora/consensus/data/UTXO.d
+++ b/source/agora/consensus/data/UTXO.d
@@ -84,12 +84,8 @@ public class TestUTXOSet
     public UTXOFinder getUTXOFinder () @trusted nothrow
     {
         this.used_utxos.clear();
-        return &this.findUTXO_;
+        return &this.findUTXO;
     }
-
-    /// FIXME: Remove and make UTXO-sensible tests use either `peekUTXO`
-    /// or `getUTXOFinder`
-    public alias findUTXO = peekUTXO;
 
     /// Get an UTXO, no double-spend protection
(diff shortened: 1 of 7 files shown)


source/agora/consensus/data/UTXO.d
```diff
@@ -126,7 +122,7 @@ public class TestUTXOSet
     }
 
     /// Get an UTXO, does not return double spend
-    private bool findUTXO_ (Hash utxo, out UTXO value)
+    public bool findUTXO (Hash utxo, out UTXO value)
```
linked0, line 125: Is there any reason why the access modifier changed to `public`? The code means that we can call it directly.

Geod24 (approved):

linked0 (approved):
LGTM, but one question.

source/agora/consensus/data/UTXO.d
Geod24: Yes, that was the point of this change. Before `findUTXO` would alias to `peekUTXO`, meaning no double spend protection. This was done to keep tests from passing, because the `TestUTXOSet` didn't originally have double spend protection, so adding it would have meant fixing tons of tests otherwise. Which @kchulj eventually did :)

kchulj: Yes, that allows us to call it directly.

Issue #1267, opened 2020-10-20
Implement proof of concept of Eltoo


AndrejMitrovic:
As mentioned in https://github.com/bpfkorea/agora/issues/1266, there are currently two main competitive LN update mechanisms. LN-Penalty, and Eltoo. LN-Penalty has three compatible implementations, whereas Eltoo is still blocked from being developed & deployed on Bitcoin due to a requirement for a new sighash type ([BIP 118](https://github.com/bitcoin/bips/pull/943)).

LN developers seem to be in agreement that Eltoo is a great evolution over LN-Penalty, and may end up replacing the penalty update protocol. Note that both protocols can be used at the same time on Bitcoin.

In our case we have a big advantage over Bitcoin. We can develop the protocol without being blocked by the requirement of doing a soft-fork. In addition to that our PoS system could potentially make it easy to create multi-party channels - we could even make this multi-party channel creation be part of the protocol rules (think quorum shuffling, for example).

As a first step we can try to implement a proof of concept of Eltoo. There are many moving parts, and we should aim to have the simplest proof of concept.

This will probably require the following:
- Implement witness (script) fields
- Implement very simple stack-based execution engine with only the required opcodes (e.g. time locks, sequence matching)
- Implement sequence number for script matching (this might even possibly be optimized out later with schnorr tricks, I need to experiment)
- Add support for sighash_noinput
- Add HTLC support for multi-hop payments
- Allocate anything over >40K funds for channel creation

Thereafter it should be fairly simple to implement multi-party channels as well.

remyers:
I implemented a PoC eltoo test script using the original BIP-118 segwit version of eltoo. This might be of interest if you are implementing eltoo, but needs to updated for Taproot version of anyprevout.

https://github.com/remyers/bitcoin/commits/anyprevout

AndrejMitrovic:
Hey @remyers! Thanks, I was aware of that implementation. I was in contact with Christian Decker about existing implementations and ideas and he mentioned your work.

AndrejMitrovic:
Resolved in #1610

PR #1339, opened 2020-11-09
Implement transaction-level absolute and input-level relative time locks

### LINKED ISSUE #243: Define and implement time locks on transactions

AndrejMitrovic:
This will be needed for Eltoo (https://github.com/bpfkorea/agora/issues/1267) and also for multi-hop payment channels.

Note that the execution engine itself **is not** in charge of checking the block height. The engine only cares about executing opcodes in the scope of the Transaction itself and the Output it is spending - it does not care about block heights. This allows transactions to be cached - otherwise a transaction's script would have to be re-evaluated at each new block height to determine if it's valid. There are many BIPs explaining how this mechanism works, if you're curious.

Fixes https://github.com/bpfkorea/agora/issues/243

### DIFF
--- a/source/agora/consensus/data/Transaction.d
+++ b/source/agora/consensus/data/Transaction.d
@@ -130,24 +130,32 @@ public struct Input
     /// A signature that should be verified using the `previous[index].address` public key
     public Signature signature;
 
+    /// The UTXO this `Input` references must be at least `unlock_age` older
+    /// than the block height at which the spending transaction wants to be
+    /// included in the block. Use for implementing relative time locks.
+    public uint unlock_age = 0;
+
     /// Simple ctor
-    public this (in Hash utxo_, in Signature sig = Signature.init)
+    public this (in Hash utxo_, in Signature sig = Signature.init, uint unlock_age = 0)
         inout pure nothrow @nogc @safe
     {
         this.utxo = utxo_;
         this.signature = sig;
+        this.unlock_age = unlock_age;
     }
 
     /// Ctor which does hashing based on index
-    public this (Hash txhash, ulong index) nothrow @safe
+    public this (Hash txhash, ulong index, uint unlock_age = 0) nothrow @safe
     {
         this.utxo = hashMulti(txhash, index);
+        this.unlock_age = unlock_age;
     }
 
     /// Ctor which does hashing based on the `Transaction` and index
-    public this (in Transaction tx, ulong index) nothrow @safe
+    public this (in Transaction tx, ulong index, uint unlock_age = 0) nothrow @safe
     {
         this.utxo = hashMulti(tx.hashFull(), index);
+        this.unlock_age = unlock_age;
     }
 
     /***************************************************************************
@@ -162,6 +170,7 @@ public struct Input
     public void computeHash (scope HashDg dg) const nothrow @safe @nogc
     {
         dg(this.utxo[]);
+        hashPart(this.unlock_age, dg);
     }
 }
 
@@ -202,7 +211,7 @@ unittest
     );
 
     const tx_payment_hash = Hash(
-        `0x79a7bb3cae7e4d46a45674fefd3708557574890ce7a554c09c7d486346a31feb82005aa593620db332c1ae32c5ab44f0977c8834969883a1928e8db0d1b3ccac`);
+        `0x428d691addd27708b719a5e47cbac932618f0f843681dc0a46c97971ff8c419c817c65d90f3b74394db46801843b79537a583903bec5da57de70155276d2aa46`);
     const expected1 = payment_tx.hashFull();
     assert(expected1 == tx_payment_hash, expected1.toString());
 
@@ -213,7 +222,7 @@ unittest
     );
 
     const tx_freeze_hash = Hash(
-        `0x7b8f848dcc4deab1c2161aa4e91401b6d9e57f2ac8126493f0343efbc23a89217871e7b1c2ef8f1a40736cd4b331e26de43b48f593097fe97941a92673635895`);
+        `0x6ce6bdeac41ffa444e6c2250ec09e04652597c3ec92f54f69029cc16ae4fc84faa5372b56e24c4c8667d00a8d1c0a7bc550999e4cdcd039548361a15e72fa081`);
     const expected2 = freeze_tx.hashFull();
     assert(expected2 == tx_freeze_hash, expected2.toString());
 }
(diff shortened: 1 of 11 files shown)


AndrejMitrovic:
This is ready for review.

Note that the lock is based on the block height, and not an absolute time. We don't yet have block timestamping.

AndrejMitrovic:
@hewison-chris / @linked0 could I get a review of this?

source/agora/consensus/data/Transaction.d
```diff
@@ -124,24 +131,32 @@ public struct Input
     /// A signature that should be verified using the `previous[index].address` public key
     public Signature signature;
 
+    /// The UTXO this `Input` references must be at least `unlock_age` older
+    /// than the block height at which the spending transaction wants to be
+    /// included in the block. Use for implementing relative time locks.
+    public uint unlock_age = 0;
```
TrustHenry, line 136: Oh. Stoa needs to respond to this change.

AndrejMitrovic: Yes, to both changes.

source/agora/consensus/data/Transaction.d
```diff
@@ -53,6 +53,11 @@ public struct Transaction
     /// The list of newly created outputs to put in the UTXO
     public Output[] outputs;
 
+    /// This transaction may only be included in a block with `height >= unlock_height`.
+    /// Note that another tx with a lower lock time could double-spend this tx.
+    /// Using `uint` instead of `Height` as it's enough to cover 80k years.
+    public uint unlock_height = 0;
```
TrustHenry: Yeah, here it is. um.. uint

AndrejMitrovic: > Yeah, here it is. um.. uint

There is a comment above, please read it.

AndrejMitrovic:
@TrustHenry the changes should be fairly simple. Just the addition of two new fields. I think it should be an easy change on the Stoa side?

linked0 (approved):
LGTM.
Additionally, there's some point where someone could confuse about why the two kinds of heights have different types although `uint` is enough. And, It would be good to show a situation or a use case where the `unlock_age` is used.

AndrejMitrovic:
> Additionally, there's some point where someone could confuse about why the two kinds of heights have different types although uint is enough. 

I could change it back to `Height`. I guess we could change this later if we want to optimize it..

> And, It would be good to show a situation or a use case where the unlock_age is used.

The primary use-case is with hashed time-locked contracts (https://en.bitcoin.it/wiki/Hash_Time_Locked_Contracts), but this requires execution engine support. This is in another PR that I will rebase soon: https://github.com/bpfkorea/agora/pull/1341

TrustHenry:
> @TrustHenry the changes should be fairly simple. Just the addition of two new fields. I think it should be an easy change on the Stoa side?

Structural change itself is simple, but Stoa test codes are not.
Stoa needs to flexibly replace Agora's changes.
But now Stoa is exactly aligned with the Agora block structure.
And before merging PR, it would be good to have related issues in Stoa together.

TrustHenry:
The word `unlock_height` has similar functions as `Transaction.unlock_height` and `UTXO.unlock_height`, but the target is different.
When we talk about `unlock_height` later, we need to distinguish between `transaction` or `UTXO`.

AndrejMitrovic:
Alright, then to keep it simple I will rename it to `time_lock` which is similar to Bitcoin.

AndrejMitrovic:
Although `Height time_lock;` will look weird..

AndrejMitrovic:
`Height height_lock;` it is then. At least it will be easy to search for the different symbols.

AndrejMitrovic:
@bpfkorea/core I have an idea.

Since there's a lot of breaking changes that the flash layer introduces (transaction time locks, Input / Output changes), what do you think about adding a new `flash-layer` branch into `agora` so all of my flash layer pull requests would target this branch?

This way we can avoid breaking Stoa / Faucet until the flash layer is stabilized.

I would periodically merge the changes from `v0.x.x` into `flash-layer`. So you don't have to worry about this branch, your own PRs can still target only `v0.x.x`.

Then when Mathias returns we could get a final review, and merge `flash-layer` into `v0.x.x`.

Thoughts?

source/agora/consensus/validation/Transaction.d
```diff
@@ -76,6 +79,15 @@ public string isInvalidReason (
         if (!sum_unspent.add(utxo_value.output.value))
             return "Transaction: Input overflow";
 
+        // note: this is strictly not necessary to be here, the Input's script
+        // should be evaluated for validity and then the Input could be kept
+        // until the unlock height becomes valid. Alternatively we could reject
+        // it here right away and force the party to send it again at the right
+        // time. However, we run into a risk if we ever implement caching of
+        // rejecting them if they're submitted too early.
+        if (height < utxo_value.unlock_height + input.unlock_age)
```
TrustHenry, line 88: I think there's not enough testing on this, but it all looks good.

TrustHenry (approved):
LGTM!

source/agora/consensus/validation/Transaction.d
AndrejMitrovic: Can you explain? The unittests are not enough?

AndrejMitrovic:
I changed the target branch to `flash-layer`. This means merging this PR will not break Stoa / Faucet.

kchulj:
> Thoughts?

Good idea.

source/agora/consensus/validation/Transaction.d
TrustHenry: This is a question,
are you thinking of replacing the `utxo_value.unlock_height` with the `height_lock` later?
In my Think Verification methods are confusing when it is a Freezing transaction or a melting transaction.

AndrejMitrovic: This code you're commenting on does not use `height_lock`. This is the transaction-level lock. 

The code instead uses `unlock_age`, which is an input-level lock. Unlock age is the number of externalized blocks between when a UTXO is first ready to be spent and the current block height.

> In my Think Verification methods are confusing when it is a Freezing transaction or a melting transaction.

Yes it's a little bit complicated right now because we have this special freeze transaction.

But we cannot build the flash layer with just freeze transactions. We need transaction-level and input-level time locks too. This is issue #243.

AndrejMitrovic: #243 will be part of my upcoming demo, so this will all be explained there.

AndrejMitrovic:
Ok, there's two approvals so I'll merge this and continue with the work.

Just to make it clear again, this will not require changes to Stoa / Faucet until next year.

PR #1340, opened 2020-11-09
Implement policy finding candidates to be slashed

linked0:
This creates the policy manager class for slashing and defined the base criteria where a validator will be slashed as the first step of issue 1076. As a result of the whiteboarding session, we start with 10k on the missing pre-image, effective immediately. 

This PR does not contain how to impose penalty on the misbehaving validator yet. I will implement the feature imposing 10K BOA on the validator in other PR.

Part of #1076

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -521,6 +521,20 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Get pre-images of all the validators
+
+        Returns:
+            All the pre-images of validators
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getAllPreimages () @safe nothrow
+    {
+        return this.validator_set.getAllPreimages();
+    }
+
+    /***************************************************************************
+
         Generate the random seed reduced from the preimages for the provided
         block height.
 
--- a/source/agora/consensus/SlashPolicy.d
+++ b/source/agora/consensus/SlashPolicy.d
@@ -1,0 +1,152 @@
+/*******************************************************************************
+
+    Manages the slashing policy for the misbehaving validators that do not
+    publish pre-images timely.
+
+    This class currently has two responsibilities:
+    - It determines when the misbehaving validators will be slashed -- in other
+        words, how many times of missing pre-images it will allow.
+    - It determines what the penalty is for misbehaving validators -- in other
+        words, how many BOA it should pay for a penalty.
+
+    All the validators should publish their pre-images timely in order for
+    the network to maintain randomness. So we need a penalty as an incentive
+    to make validators publish them regularly. So this class manages all the
+    policies for penalty and slashing.
+    See https://github.com/bpfkorea/agora/issues/1076.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.SlashPolicy;
+
+import agora.common.Amount;
+import agora.common.crypto.Key;
+import agora.common.Types;
+import agora.consensus.EnrollmentManager;
+import agora.consensus.data.Enrollment;
+import agora.consensus.data.PreImageInfo;
+import agora.consensus.state.UTXODB;
+import agora.utils.Log;
+version (unittest) import agora.utils.Test;
+
+import std.exception;
+
+mixin AddLogger!();
+
+/*******************************************************************************
+
+    Manage the policy for slashing
+
+*******************************************************************************/
+
+public class SlashPolicy
+{
+    // The amount of the penalty set to 10K BOA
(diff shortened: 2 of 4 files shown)


source/agora/consensus/SlashPolicy.d
```diff
@@ -0,0 +1,159 @@
...
+                slashed_keys ~= utxo_value.output.address;
+            }
+        }
+
+        return slashed_keys;
+    }
+
+    public Amount getPenaltyAmount()
+    {
+        return this.penalty_amount;
```
AndrejMitrovic: I'm not sure this belongs here. Will this class be in charge of slashing? It seems not, because it has a public method `getSlashCandidates`, so it will be another class which does the actual slashing right? This one is just about selecting the candidates.

But then this class doesn't need to store the penalty amount, the other class will be in charge of that.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -519,6 +519,20 @@ public class EnrollmentManager
...
+    /***************************************************************************
+
+        Get pre-images of all the validators
+
+        Returns:
+            All the pre-images of validators
+
+    ***************************************************************************/
+
+    PreImageInfo[] getAllPreimages () @trusted nothrow
```
AndrejMitrovic: `public`. And please use `@safe` and make the function in `ValidatorSet` also `@safe`.

source/agora/consensus/state/ValidatorSet.d
```diff
@@ -414,6 +414,39 @@ public class ValidatorSet
...
+    /***************************************************************************
+
+        Get pre-images of all the validators
+
+        Returns:
+            All the pre-images of validators
+
+    ***************************************************************************/
+
+    PreImageInfo[] getAllPreimages () @trusted nothrow
```
AndrejMitrovic: `public` and `@safe`

source/agora/consensus/SlashPolicy.d
```diff
@@ -0,0 +1,159 @@
...
+    /***************************************************************************
+
+        Get validators to be slashed due to missing pre-images
+
+        This finds the misbehaving validators which fail to publish their
+        pre-images regularly. A validator will be the candidate for slashing
+        if the last revealed pre-image is behind, based on the current height.
+
+        Params:
+            block_height = current block height
```
AndrejMitrovic:
```suggestion
            height = current block height
```

source/agora/consensus/SlashPolicy.d
```diff
@@ -0,0 +1,159 @@
...
+
+        foreach (preimage; preimages)
+        {
+            auto enrolled = this.enroll_man.getEnrolledHeight(preimage.enroll_key);
+            assert(height >= enrolled);
+            ushort curr_distance = cast(ushort)(height - enrolled);
+            if (preimage.distance < curr_distance)
+            {
+                UTXO utxo_value;
+                finder(preimage.enroll_key, utxo_value);
```
AndrejMitrovic: UTXOFinder returns a boolean, please use it here instead of comparing with `UTXO.init`.

source/agora/consensus/SlashPolicy.d
```diff
@@ -0,0 +1,159 @@
...
+
+        Params:
+            enroll_man = the EnrollmentManager
+
+    ***************************************************************************/
+
+    public this (EnrollmentManager enroll_man)
+    {
+        this.enroll_man = enroll_man;
+        this.penalty_amount = Amount(100_000_000_000);
```
AndrejMitrovic, line 69: This ultimately belongs in the `ConsensusParams`. However, since this is WIP you may keep it hardcoded for now.

AndrejMitrovic (changes requested):
Just a few issues.

source/agora/consensus/SlashPolicy.d
linked0: Currently, the `SlashPolicy` class manages the amount but it can be moved into `ConsensusParams` as you commented. It will be done on other PR.

source/agora/consensus/EnrollmentManager.d
linked0: My mistake. I will update it.

source/agora/consensus/state/ValidatorSet.d
linked0: Ditto.

source/agora/consensus/SlashPolicy.d
linked0: Will update it.

source/agora/consensus/SlashPolicy.d
linked0: Ditto

source/agora/consensus/SlashPolicy.d
linked0: The description of the class is confusing. I will update it. Actually this class is in charge of managing slashing policy which determine the condition where the slashing must be done.

linked0:
This PR is not needed anymore.

PR #1341, opened 2020-11-10
Implement time-lock opcodes in the execution engine

AndrejMitrovic:
This previously required:

- https://github.com/bpfkorea/agora/pull/1291
- https://github.com/bpfkorea/agora/pull/1339

This does not yet integrate the engine with Agora, that will be part of another PR.

### DIFF
--- a/source/agora/script/Engine.d
+++ b/source/agora/script/Engine.d
@@ -509,6 +509,21 @@ public class Engine
 
                 break;
 
+            case OP.VERIFY_UNLOCK_AGE:
+                if (stack.count() < 1)
+                    return "VERIFY_UNLOCK_AGE opcode requires an unlock age on the stack";
+
+                const age_bytes = stack.pop();
+                if (age_bytes.length != uint.sizeof)
+                    return "VERIFY_UNLOCK_AGE unlock age must be a 4-byte number";
+
+                const uint unlock_age = littleEndianToNative!uint(
+                    age_bytes[0 .. uint.sizeof]);
+                if (unlock_age > input.unlock_age)
+                    return "VERIFY_UNLOCK_AGE unlock age of input is too low";
+
+                break;
+
             case OP.INVALID:
                 return "Script panic while executing OP.INVALID opcode";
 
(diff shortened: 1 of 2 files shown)


AndrejMitrovic:
@bpfkorea/core ready for review.

source/agora/script/Engine.d
```diff
@@ -493,6 +494,21 @@ public class Engine
...
+                if (stack.count() < 1)
+                    return "VERIFY_HEIGHT_LOCK opcode requires a lock height on the stack";
+
+                const height_bytes = stack.pop();
+                if (height_bytes.length != ulong.sizeof)
+                    return "VERIFY_HEIGHT_LOCK height lock must be an 8-byte number";
+
+                const Height height_lock = Height(littleEndianToNative!ulong(
+                    height_bytes[0 .. ulong.sizeof]));
+                if (height_lock > tx.height_lock)
```
linked0, line 507: Is the `tx.height_lock` always the height of a proposed block in the situation of OP.VERIFY_HEIGHT_LOCK being used?

AndrejMitrovic:
Alright, I'll proceed with the work.

linked0 (approved):
LGTM.
Thank you for the great work and one question.

source/agora/script/Engine.d
AndrejMitrovic: It's the height of some future block height, for example +10 blocks in the future.

The engine itself validates the height lock in the transaction, while the consensus rules validate that such a transaction won't be added to the blockchain until that height is reached. So a transaction can be valid, just not yet unlocked. 

It allows creating time contracts between multiple parties. The transaction will be valid, but it will be guaranteed that it will not be included in a block until the height in `height_lock` is reached.

AndrejMitrovic: The chapter on timelock is very useful: https://cypherpunks-core.github.io/bitcoinbook/ch07.html

linked0: I got it. You mean that the transaction having `tx.height_lock` higher than the current height is just stored in the transaction pool yet before the height is reached.
But isn't there a kind of double-spending problem? I guess some bad node can make two transactions, one transaction having `tx.height_lock` of current height and the other transaction having `tx.height_lock` of future block height.

AndrejMitrovic: That's right. And the book mentions that also. That's why input-level locks are better. However, transaction-level locks are still useful for HTLCs. For example https://bitcoin.stackexchange.com/questions/87968/why-do-we-use-cltv-instead-of-csv-in-the-htlc-output-of-the-commitment-transacti and https://en.bitcoin.it/wiki/Hash_Time_Locked_Contracts.

AndrejMitrovic: The flash layer will mainly use an input-level lock. However, for multi-hop payments we will have to use HTLCs.

linked0: Ah, I get to know, to some extent, why you introduce the input-level locks. Ty.

source/agora/script/Engine.d
```diff
@@ -493,6 +494,36 @@ public class Engine
                     return "VERIFY_SIG signature failed validation";
                 break;
 
+            case OP.VERIFY_HEIGHT_LOCK:
+                if (stack.count() < 1)
```
TrustHenry, line 498: `if (stack.empty)`
Does this check the same as this?

AndrejMitrovic: Oh, good eye! I will fix it up.

source/agora/script/Engine.d
```diff
@@ -493,6 +494,36 @@ public class Engine
...
+
+                const Height height_lock = Height(littleEndianToNative!ulong(
+                    height_bytes[0 .. ulong.sizeof]));
+                if (height_lock > tx.height_lock)
+                    return "VERIFY_HEIGHT_LOCK height lock of transaction is too low";
+
+                break;
+
+            case OP.VERIFY_UNLOCK_AGE:
+                if (stack.count() < 1)
```
TrustHenry, line 513: Here it is :)

PR #1343, opened 2020-11-10
Implement SigHash support in the execution engine

AndrejMitrovic:
This implements `SigHash.All` (default), and `SigHash.NoInput` which is described in the Lightning Network and Eltoo whitepapers.

The SigHash is embedded as a single byte after the signature. Different signatures can be pushed onto the stack depending on a script's execution path, this is why Bitcoin pairs the signature with the SigHash as a tuple rather than making the SigHash separate (e.g. as an Input field).

Requires:
- #1291 
- #1339 
- #1341

The first three commits will be removed once #1291, #1339, and #1341 are merged.


AndrejMitrovic:
This won't be necessary for now. There might still be a use-case for SigHash to add arbitrary fees to transactions, but  SigHash.NoInput on the signature itself turned out to not be unnecessary.

PR #1345, opened 2020-11-10
Allow TxBuilder to attach a range of outputs

kchulj:
`TxBuilder` can attach a range of tuples that have an `Output` and a `Hash` as its elements.
This can be used as a helper function to merge `UTXO`s.

Relates [faucet#14](https://github.com/bpfkorea/faucet/issues/14)

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -387,6 +387,41 @@ public struct TxBuilder
 
     /***************************************************************************
 
+        Attaches to a range of tuples.
+
+        The tuple should have an `Output` and a `Hash` as its first and
+        second element, respectively.
+
+        Params:
+            rng = the range of tuple
+
+        Returns:
+            A reference to `this` to allow for chaining
+
+    ***************************************************************************/
+
+    public ref typeof(this) attach (RNG) (scope RNG rng)
+        @safe pure nothrow return
+    {
+        alias ET = ElementType!RNG;
+
+        static assert(isInputRange!RNG);
+        static assert(ET.length == 2, "Parameter must be a range of tuple with
+            2 elements");
+        static assert(is(typeof(ET[0]) : const Output),
+            "Tuple's first element should be convertible to `Output`, not: `" ~
+            typeof(ET[0]).stringof ~ "`");
+        static assert(is(typeof(ET[1]) : const Hash),
+            "Tuple's second element should be convertible to `Hash`, not: `" ~
+            typeof(ET[1]).stringof ~ "`");
+
+        rng.each!(tup => this.attach(tup[0], tup[1]));
+
+        return this;
+    }
+
+    /***************************************************************************
+
         Finalize the transaction, signing the input, and reset the builder
 
         Params:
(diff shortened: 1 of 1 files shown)


source/agora/utils/Test.d
```diff
@@ -385,6 +385,42 @@ public struct TxBuilder
         return this;
     }
 
+    /***************************************************************************
+
+        Attaches to a range of tuples.
+        The tuple should have an `Output` and a `Hash` as its elements.
```
```suggestion
        Attaches to a range of tuples.

        The tuple should have an `Output` and a `Hash` as its first and
        second elements, respectively.
```

The newline matters, as it's the separator between [summary and description](https://dlang.org/spec/ddoc.html#sections). And specifying the order (and how it's accessed) matters to the reader.

source/agora/utils/Test.d
```diff
@@ -385,6 +385,42 @@ public struct TxBuilder
...
+
+    ***************************************************************************/
+
+    public ref typeof(this) attach (RNG) (scope RNG rng)
+        @safe pure nothrow return
+    {
+        alias ET = ElementType!RNG;
+
+        static assert(isInputRange!RNG);
+        static assert(ET.length == 2, "Parameter must be a tuple of 2 elements");
```
```suggestion
        static assert(ET.length == 2, "Parameter must be a range of tuple with 2 elements");
```

source/agora/utils/Test.d
```diff
@@ -385,6 +385,42 @@ public struct TxBuilder
...
+            typeof(ET[0]).stringof ~ "`");
+        static assert(is(typeof(ET[1]) : const Hash),
+            "Tuple's second element should be convertible to `Hash`, not: `" ~
+            typeof(ET[1]).stringof ~ "`");
+
+        rng.each!((tup)
+        {
+            this.inputs ~= OutputRef(tup[0], tup[1]);
+            this.leftover.value.mustAdd(tup[0].value);
+        });
```
Geod24: If you can reuse code, do it. In this case:
```suggestion
        rng.each!((tup) { this.attach(tup); });
```

Also this has an interesting side-effect: The user could actually provide a `Transaction / index`, and if it wasn't for the static assert, it would work. But let's not loosen the static assert for now.

kchulj: Does `attach (in Output utxo, in Hash hash)` automatically recognize the `tup` elements and assign them respectively?

If not, then I think it is better to assign manually:
```D
rng.each!((tup) { this.attach(tup[0], tup[1]); });
```

Geod24 (approved):

source/agora/utils/Test.d
```diff
@@ -385,6 +385,41 @@ public struct TxBuilder
...
+        Params:
+            rng = the range of tuple
+
+        Returns:
+            A reference to `this` to allow for chaining
+
+    ***************************************************************************/
+
+    public ref typeof(this) attach (RNG) (scope RNG rng)
+        @safe pure nothrow return
```
TrustHenry, line 404: Oh! A reference to `this` Nice!

kchulj: I noticed that all overloads of `attach` are using `ref typeof(this)`.

TrustHenry (approved):
LGTM!
How about adding a failure test? First Hash, Second Output

source/agora/utils/Test.d
Geod24: > If not, then I think it is better to assign manually:

It does. Tuples get expanded and then parameter matching is done.

MichaelKim20:
I think we can easily aggregate outputs scattered across multiple transactions. It's a very good feature.

PR #1346, opened 2020-11-10
Add sequence ID to the Transaction

AndrejMitrovic:
In the Eltoo whitepaper floating transactions are rebindable to any previous UTXO with the matching keypair. There needs to be a way to determine which update transaction is the latest one, and to prevent older stale update transactions from being able to rebind newer update transactions.

In the proposed Bitcoin implementation of Eltoo the authors proposed repurposing the transaction-level locktime field as a form of sequence ID. The reason they can get away with this is because there are many block heights which are already externalized. So for example a set of update transactions with locktimes of [1, 2, 1000, 20_000] can be accepted and externalized because their locktimes are in the past.

However in our case we don't have the luxury of using the locktime. We don't have a history of old block heights that we could repurpose as a sequence ID. The whitepaper mentions this, and recommends for other blockchain projects to instead introduce a purpose-built sequence ID field. That's what this PR does.

Requires:
- #1291 
- #1339 
- #1341 
- #1343


AndrejMitrovic:
I thought about this, and I think it could be possible to store the sequence ID in the unlock script rather than polluting the transaction size. Since we already use a special SigHash type, we could put the sequence ID in the unlock field and hash it to verify the signature.

AndrejMitrovic:
I came up with a better scheme. The sequence ID can be pushed to the stack by the unlock script. We can introduce a new opcode (`OP.CHECK_SEQUENCE_SIGNATURE`) which verifies the pushed sequence ID is >= the expected sequence ID by the lock script (similar to `OP.VERIFY_SEQ_ID` in this PR) and also verifies the signature. To prevent malleability, the pushed sequence ID is hashed when verifying the signature.

Therefore we won't have to pollute the transaction with a sequence ID in its header.

This actually rules out a big security risk in Eltoo. In Eltoo, it's possible to attach an update TX to any other utxo with a matching public key (hence why unique key generation is so important there). In their example a new feature `SigHash.NoInput` is used which binds to the signature itself, allowing it to attach to any UTXO with a matching key. In my new proposal, it's not the signature itself which decides on attachability, but the lock script.

Then, because the signature must include the sequence ID in its hash (`OP.CHECK_SEQUENCE_SIGNATURE`), it will fail to attach to regular pay transactions (as they will not include the sequence ID in the lock when verifying the signature with `OP.VERIFY_SIG`).

I'll write up a PR in the coming days.

PR #1347, opened 2020-11-10
Demonstrate examples of on-chain and off-chain Eltoo protocol scripts

AndrejMitrovic:
This PR builds on the previous work and also requires adding lock/unlock conditions in the `Output` and `Input`, respectively. That part has not been finished yet - this is the Engine integration part I will be working on later.

Requires:
- #1291
- #1339 
- #1341 
- #1343 
- #1346 
- Engine integration PR (WIP).


AndrejMitrovic:
This doesn't need to still be a PR. I'll integrate the set of helper routines in another PR. And the on-chain version was just for testing, it doesn't need to exist in the repo.

PR #1351, opened 2020-11-10
tests: get rid of unused parameter

AndrejMitrovic:


omerfirmak (approved):

kchulj:
Would it be possible to get rid of these functions?
I remember they were duplicates of `makeChainedTransactions`.

AndrejMitrovic:
> Would it be possible to get rid of these functions?

That would be ideal. Someone should take care of these in the future.

kchulj (approved):

kchulj:
> > Would it be possible to get rid of these functions?
> 
> That would be ideal. Someone should take care of these in the future.

Relates #1107

AndrejMitrovic:
Also it's currently blocking my work on the execution engine. I just have this one failing test, but the test was written in such a terrible copy-paste way that it's hard to see what the test is actually trying to accomplish.

Issue #1364, opened 2020-11-13
Agora node stops receiving transactions at some point


kchulj:
The `faucet` continuously sends transactions, but the `agora` node stops receiving them at some point.
This bug is reproducible, and the height at which it stops receiving them, ranges from `6` to `51`.

kchulj:
Solving #1362 may fix this, as this happens after the socket error.

kchulj:
I have observed that this issue originates from one or more nodes getting the `InvalidMemoryOperation` error, crashing, and being unable to restart. The issue is the same as #1367 .
I will close this issue.

PR #1365, opened 2020-11-13
Stop nominating immediately after externalize

AndrejMitrovic:
There is a potential for a context switch here.

I'm not sure if this is related to #1364, but it's definitely a bug. We'll test it with #1364 next week.

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -352,13 +352,13 @@ public class Validator : FullNode, API
     protected final override void onAcceptedBlock (const ref Block block,
         bool validators_changed) @safe
     {
-        super.onAcceptedBlock(block, validators_changed);
+        assert(block.header.height >= this.last_shuffle_height);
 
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM

source/agora/node/Validator.d
```diff
@@ -352,12 +352,13 @@ public class Validator : FullNode, API
...
-        super.onAcceptedBlock(block, validators_changed);
-
         // block received either via externalize or getBlocksFrom(),
-        // we need to cancel any existing nominating rounds
+        // we need to cancel any existing nominating rounds.
+        // note: must be called before any context switch
         this.nominator.stopNominationRound(block.header.height);
 
+        super.onAcceptedBlock(block, validators_changed);
+
```
Geod24: Wouldn't this be better after the `assert` ?

AndrejMitrovic:
I moved it below the assert.

Issue #1327, opened 2020-11-04
[Upstream] DUB: Support setting `-preview=in` for libraries


Geod24:
Currently we cannot make use of some flags (in particular, the newly released `-preview=in` which I implemented for us), because this does not propagate to the libraries we use.
E.g.:
```console
% dub init dubtest --type=vibe.d
Package recipe format (sdl/json) [json]:
Name [dubtest]:
Description [A simple vibe.d server application.]:
Author name [Mathias Lang]:
License [proprietary]:
Copyright string [Copyright © 2020, Mathias Lang]:
Add dependency (leave empty to skip) []: vibe-d
Adding dependency vibe-d ~>0.9.2
Add dependency (leave empty to skip) []:
Successfully created an empty project in '/Users/geod24/projects/bpfk/agora/dubtest'.
Package successfully created in dubtest
% cd dubtest
```

Change `dub.json` to:
```json
{
    "name": "dubtest",
    "dflags": [ "-preview=in" ],
    "dependencies": {
        "vibe-d": "~>0.9.2"
    }
}
```

Build:
```console
% dub build -v
[TONS OF OUTPUT]
vibe-d 0.9.2: building configuration "vibe-core"...
Ignoring all import libraries for static library build.
[...]

PR #1366, opened 2020-11-13
Enable -preview=in

Geod24:
This currently does not work because of issues with dub,
but is a good showcase of the diff made possible.

Depends on:
- Dropping support for < v2.094.0 (so let's wait for v2.095.0)
- #1327


Geod24:
This should be unblocked in a few weeks at most: https://forum.dlang.org/thread/hwrdxaxaqahnipnapawx@forum.dlang.org

Geod24:
Superseded by #1700

PR #1369, opened 2020-11-17
Stabilize quorum generator tests

AndrejMitrovic:
The quorum generator tests will no longer fail whenever there is a change in the hashing algorithm of the transactions.


linked0 (approved):
Nice!

AndrejMitrovic:
Maybe I should remove the file line from the hashing because it will break the tests if any line is moved..

PR #1372, opened 2020-11-18
Fix GitHub CI for windows

TrustHenry:


TrustHenry:
@hewison-chris Let's see if this works.

hewison-chris:
Maybe this PR is better as it keeps it similar to how we do the linux and Mac jobs. Also the path is known in this step.

hewison-chris (approved):
Good find

TrustHenry:
@hewison-chris  has a better idea. https://github.com/bpfkorea/agora/pull/1373

PR #1375, opened 2020-11-19
Engine: Implement OP.CHECK_SEQ_SIG to support floating transactions

AndrejMitrovic:
This feature enables implementing so-called "floating" transactions. Such a transaction may be attached to any previous matching transaction as long as its sequence number is equal to or greater than the minimum sequence as defined in the lock script.

This feature was originally introduced in the Eltoo whitepaper: https://blockstream.com/eltoo.pdf

2019-04-25-eltoo-figure-1

In the original Eltoo proposal, a new `SigHash` type is added which makes it possible for a signature to be attached to any other input as long as the key and the amount matches (part of the hash). However, this has a security risk: if the key is ever reused for another regular transaction then the funds could be stolen by simply attaching the signature to another input.

I came up with an alternative: Sequence signatures may only be attached to lock scripts which declare that the unlock script must include a sequence signature. This signature cannot be attached to an input which calls `OP.CHECK_SIG`, because the hash will include the non-blanked Input and therefore the signature will fail validation.

In addition to this approach, we also do not introduce a sequence number into the transaction as suggested. Instead we push the sequence to the stack. To make this work we make `OP.CHECK_SEQ_SIG` hash the sequence that's pushed on the stack when validating the signature - so the sequence must be signed. That way we prevent malleability.

-----

The code will make much more sense once I have the proof-of-concept PR ready next week.

### DIFF
--- a/source/agora/script/Engine.d
+++ b/source/agora/script/Engine.d
@@ -494,6 +494,26 @@ public class Engine
                     return "VERIFY_SIG signature failed validation";
                 break;
 
+            case OP.CHECK_SEQ_SIG:
+                bool is_valid;
+                if (auto error = this.verifySequenceSignature!(OP.CHECK_SEQ_SIG)(
+                    stack, tx, input, is_valid))
+                    return error;
+
+                // canPush() check unnecessary
+                stack.push(is_valid ? TrueValue : FalseValue);
+                break;
+
+            case OP.VERIFY_SEQ_SIG:
+                bool is_valid;
+                if (auto error = this.verifySequenceSignature!(OP.VERIFY_SEQ_SIG)(
+                    stack, tx, input, is_valid))
+                    return error;
+
+                if (!is_valid)
+                    return "VERIFY_SEQ_SIG signature failed validation";
+                break;
+
             case OP.VERIFY_HEIGHT_LOCK:
                 if (stack.empty())
                     return "VERIFY_HEIGHT_LOCK opcode requires a lock height on the stack";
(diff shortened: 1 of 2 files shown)


source/agora/script/Engine.d
```diff
@@ -735,6 +755,144 @@ public class Engine
...
+            return err4;
+        }
+
+        const seq_bytes = stack.pop();
+        if (seq_bytes.length != ulong.sizeof)
+        {
+            static immutable err5 = op.to!string
+                ~ " opcode requires 8-byte sequence on the stack";
+            return err5;
+        }
```
TrustHenry, line 839: I think the test code for seq_bytes is missing.
However, the length of `min_seq_bytes` has been tested.

AndrejMitrovic: Good eye! I'll add the test.

AndrejMitrovic:
Thank you. I added two missing test-cases.

TrustHenry (approved):
LGTM!

source/agora/script/Engine.d
```diff
@@ -735,6 +755,144 @@ public class Engine
...
+        the challenge as a hash
+
+*******************************************************************************/
+
+public Hash getSequenceChallenge (in Transaction tx, in ulong sequence,
+    in ulong input_idx) nothrow @safe
+{
+    assert(input_idx < tx.inputs.length, "Input index is out of range");
+
+    Transaction dup;
```
linked0, line 890: The variable name `dup` is confusing with the `dup` function.

source/agora/script/Engine.d
```diff
@@ -735,6 +755,144 @@ public class Engine
...
+
+        Returns:
+            an error string if the needed arguments on the stack are missing,
+            otherwise returns null
+
+    ***************************************************************************/
+
+    private string verifySequenceSignature (OP op)(ref Stack stack,
+        in Transaction tx, in Input input, out bool sig_valid)
+        nothrow @safe //@nogc  // stack.pop() is not @nogc
```
linked0, line 788: Do you intend to leave the comment?

source/agora/script/Engine.d
```diff
@@ -945,6 +1103,175 @@ unittest
...
+
+    const seq_0 = ulong(0);
+    const seq_1 = ulong(1);
+    const seq_0_bytes = nativeToLittleEndian(seq_0);
+    const seq_1_bytes = nativeToLittleEndian(seq_1);
+
+    test!("==")(engine.execute(
+        Lock(LockType.Script,
+              [ubyte(1), ubyte(1)]  // wrong pubkey size
+            ~ toPushOpcode(seq_0_bytes)
```
linked0, line 1146: We usually put the `~` operator at the end of a line, I guess.

linked0 (approved):
A good solution for the security risk. But I need more knowledge for full understanding :) 
Left some comments.

source/agora/script/Engine.d
AndrejMitrovic: It depends. In some code it's done that way.

But recently I've realized that putting it on its own line makes it easier to see the concatenations. Because they all get aligned. For example:

```D
int[] arr = [1, 2, 3] ~
    [4, 5, 6, 7, 8 9, 10, 11, 12] ~
    [7, 8];
```

Compared to:

```D
int[] arr = [1, 2, 3]
    ~ [4, 5, 6, 7, 8 9, 10, 11, 12]
    ~ [7, 8];
```

In he second case the concatenations are all on the same column and easy to see. I think this is also the style Mathias prefers.

source/agora/script/Engine.d
AndrejMitrovic: Yes because I wanted to make `pop()` `@nogc` in the future.

But maybe these comments are not so useful?

source/agora/script/Engine.d
AndrejMitrovic: Good point. I'll rename it in another PR.

source/agora/script/Engine.d
linked0: Ah, I understand this is our preference.

source/agora/script/Engine.d
linked0: Yes, useful.

Issue #1325, opened 2020-11-04
[Upstream] DUB: Make `excludedSourceFiles` recursive


Geod24:
`excludedSourceFiles` uses globbing, but that doesn't work quite well, as subpackages are not included by default.
Example:
https://github.com/bpfkorea/agora/blob/7a7a5b186443de3b2bc676ff08247f67bdb98551/dub.json#L109-L111

This is just stupid, as it's very unlikely (and frankly, bad practice) that one would want to exclude a package but not its sub-packages. So I suggest that we allow `excludedSourcePaths` to be recursive iff the path ends with a trailing slash, e.g. in the example above we would use `source/agora/cli/`.

omerfirmak:
Why dont we simply `source/agora/cli/*` wouldnt that work?

Geod24:
Well that's the point :D
`source/agora/cli/*` will *NOT* excludes files in subdirectories, only source files in `source/agora/cli/`. Give it a shot.

omerfirmak:
`BuildSettings.removePaths(["path1", "root/path1", "root/path2/test.d", "root2/path1"], ["root/*"]);` outputs `["path1", "root2/path1"]`. I will try it in dub.json as well.

omerfirmak:
File tree:

- source
   - app.d
   - folder1
       - test.d
       - failfolder
           - fail.d

fail.d containts a compile error.
with `"excludedSourceFiles": ["source/folder1/*"]`, project compiles successfully. Am I missing something here?

omerfirmak:
`std.path.globMatch` might have fixed its own end.

Geod24:
If it works nowadays, could you raise a PR against `dub.json` then ?

Geod24:
And perhaps take a look at older dub. Using the install script, you can do `~/dlang/install.sh dub-1.20.0` and then use `~/dlang/dub-1.20.0/bin/dub` to compile, and test if we need to require a minimum version of `dub`.

omerfirmak:
> https://github.com/bpfkorea/agora/blob/7a7a5b186443de3b2bc676ff08247f67bdb98551/dub.json#L109-L111
> 
> This is just stupid, as it's very unlikely (and frankly, bad practice) that one would want to exclude a package but not its sub-packages. So I suggest that we allow `excludedSourcePaths` to be recursive iff the path ends with a trailing slash, e.g. in the example above we would use `source/agora/cli/`.

[...]

PR #1377, opened 2020-11-20
Minor simplification to the dub.json

Geod24:
Since #1325 is a non-issue.


omerfirmak (approved):
At some point #1325 might have been an actual issue though. `source/agora/cli/*/*` suggest that to me.

Geod24:
Perhaps it was an issue in the past. I tested with v1.20.0 but I cannot test further on OSX.

PR #1379, opened 2020-11-23
Refuse Enrollments if the public key is already present

### LINKED ISSUE #1310: Refuse Enrollments if the public key is already present
We should refuse any Enrollment which would lead to the same public key being enrolled twice (despite the two UTXOs).
The goal is to prevent mishaps, as this was a scenario which was originally envisioned.
In the future, we can relax this limitation if there's a good use case for it (e.g. increasing liquidity).

linked0:
We should refuse any Enrollment which would lead to the same public key being enrolled twice (despite the two UTXOs). In other words, enrollments from the same public key can not be accepted in an overlapped cycles. But, a validator can enroll with any UTXO after a validator cycle ends.

This PR is for that requirement. This contains network tests also.

Fixes #1310

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -597,33 +597,6 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Restore validators' information from block
-
-        Params:
-            last_height = the latest block height
-            block = the block to update the validator set with
-            finder = the delegate to find UTXOs with
-            self_utxos = the UTXOs belonging to this node
-
-    ***************************************************************************/
-
-    public void restoreValidators (Height last_height, const ref Block block,
-        scope UTXOFinder finder, in UTXO[Hash] self_utxos)
-        @safe nothrow
-    {
-        assert(last_height >= block.header.height);
-        if (last_height - block.header.height < this.params.ValidatorCycle)
-        {
-            foreach (const ref enroll; block.header.enrollments)
-            {
-                this.addValidator(enroll, block.header.height, finder,
-                    self_utxos);
-            }
-        }
-    }
-
-    /***************************************************************************
-
         Get the public key of node that is used for a enrollment
 
         Returns:
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -192,6 +192,7 @@ public class EnrollmentManager
 
         Params:
             enroll = the enrollment data to add
+            pubkey = the public key of the enrollment
             block_height = current block height
             finder = the delegate to find UTXOs with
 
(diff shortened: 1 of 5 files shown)


source/agora/consensus/state/ValidatorSet.d
```diff
@@ -94,15 +95,20 @@ public class ValidatorSet
         if (this.hasEnrollment(enroll.utxo_key))
             return "This validator is already enrolled";
 
+        // check if an enrollment of the same public key is already present
+        if (this.hasPublicKey(pubkey))
```
linked0, line 99: This checking code should be also called in the `isInvalidCandidateReason` function of `EnrollmentManager`. I will fix it.

linked0:
After updating some code, I will request a review for it again.

source/agora/consensus/state/ValidatorSet.d
```diff
@@ -237,6 +243,34 @@ public class ValidatorSet
...
+    {
+        try
+        {
+            auto results = this.db.execute("SELECT EXISTS(SELECT 1 FROM " ~
+                "validator_set WHERE public_key = ?)", pubkey.toString());
+            return results.front().peek!bool(0);
+        }
+        catch (Exception ex)
+        {
+            log.error("Exception occured on hasEnrollment: {}, " ~
```
AndrejMitrovic: We need to stop writing code like this.

If there is an error in the database, then the enrollment will be accepted. Do you really want that to happen?

linked0: > If there is an error in the database, then the enrollment will be accepted. Do you really want that to happen?

I added `assert(0)` statements into the two critical functions, `hasEnrollment` and `hasPublicKey`. But we would need to check more about other functions as a separate issue.

linked0:
Ready for review

source/agora/consensus/EnrollmentManager.d
```diff
@@ -200,8 +201,9 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public bool addEnrollment (const ref Enrollment enroll, Height block_height,
-        scope UTXOFinder finder) @safe nothrow
+    public bool addEnrollment (const ref Enrollment enroll,
+        const ref PublicKey pubkey, Height block_height, scope UTXOFinder finder)
+        @safe nothrow
```
Geod24, line 206: Why do we need to pass the public key if we look up the UTXO ?

Geod24: Okay, looking at the rest of the code, it might make sense to just pass the `UTXO` and remove the `finder` and `pubkey` argument.

source/agora/consensus/state/ValidatorSet.d
AndrejMitrovic: Thanks. There is still an open issue about handling SQLite errors properly. Hopefully someone takes care of that.

Geod24:
First commit is great. Second and third should be merged, and note the comment about UTXO. Last commit, I'm not so sure about it. I don't think we need it right away anyway, do we ?

source/agora/consensus/EnrollmentManager.d
linked0: OK, I understand you meant we can remove also the `finder` argument on `isInvalidReason`. Right?
In fact, the `finder` is used for `isInvalidReason`  in  `EnrollmentPool`'s `add` function.

Geod24: Ah, I see, so it will be a bit complicated. Let's leave it for another day then.

Geod24:
Gave it a second read and realized I missed Drey's comment. Let's go ahead with this, thanks!

PR #1393, opened 2020-11-30
tests: wait for preimages to be revealed

AndrejMitrovic:
This is similar to a workaround we implemented in the localrest tests, but we never did it for the system integration tests.


linked0 (approved):
LGTM

Issue #1395, opened 2020-11-30
vibe.d bug with http requests (previously: memory corruption bug)


AndrejMitrovic:
> node-2_1  | 2020-11-30 09:24:40,909 Info [agora.network.NetworkManager] - Net time offsets: [{ key: GDNODE6ZXW2NNOOQIGN24MBEZRO5226LSMHGQA3MUAMYQSTJVR7XT6GH, node_time: 1, req_delay: 0, offset: -1606728279 }, { key: GDNODE3EWQKF33TPK35DAQ3KXAYSOT4E4ACDOVJMDZQDVKP66IMJEACM, node_time: 1, req_delay: 0, offset: -1606728279 }, { key: GDNODE2IMTDH7SZHXWDS24EZCMYCEJMRZWB3S4HLRIUP6UNGKVVFLVHQ, node_time: 1606728279, req_delay: 0, offset: 0 }, { key: GDNODE4KTE7VQUHVBLXIGD7VEFY57X4XV547P72D37SDG7UEO7MWOSNY, node_time: 1606728280, req_delay: 0, offset: 0 }, { key: GDNODE6ZXW2NNOOQIGN24MBEZRO5226LSMHGQA3MUAMYQSTJVR7XT6GH, node_time: 1606728280, req_delay: 0, offset: 0 }, { key: GDNODE7J5EUK7T6HLEO2FDUBWZEXVXHJO7C4AF5VZAKZENGQ4WR3IX2U, node_time: 1606728280, req_delay: 0, offset: 0 }, { key: GDNODE7J5EUK7T6HLEO2FDUBWZEXVXHJO7C4AF5VZAKZENGQ4WR3IX2U, node_time: 1606728280, req_delay: 0, offset: 0 }]
> node-2_1  | 2020-11-30 09:24:40,960 Info [agora.network.NetworkManager] - Retrieving blocks [2..1606728280] from http://node-6:6826..

Look closely. The logging which added the `node_time` logging emits the same number `1606728280` as in the `Retrieving blocks` log line. There seems to be some memory corruption which propagates everywhere.

I also added some additional logging and ran into this:

```
node-4_1  | RECEIVED INVALID PUBLIC KEY: '' from submodules/vibe.d/data/vibe/data/serialization.d(739)
```

This is in `Key.fromString (scope const(char)[] str)`. There is probably a separate bug here that we're using assertions in place of exceptions inside this function. But it's also highly unusual that legitimate local nodes are sending empty json strings.

[...]

PR #1400, opened 2020-12-02
[osx] Reproducible test-case for #1395 (vibe.d bug)

AndrejMitrovic:
This includes some commits which help reproduce the issue.

The last commit curiously gets rid of the bug. In it I replaced the return value of `getLocalTime()` from `time_t` to `struct { time_t }`. I didn't randomly come up with this workaround. I initially added several fields in this struct and populated them with some strings, just to see if they would end up getting corrupted. But instead the bug disappeared.

To reproduce the bug with commit 8b31aa9 (_Change logging_):

```
$ git checkout 8b31aa9
$ DFLAGS="-g --link-defaultlib-debug" dub build agora
```

Then, spawn 6 terminals and CD to each system integration test's directory:

```
$ cd agora/tests/system/node/7
$ cd agora/tests/system/node/6
$ cd agora/tests/system/node/5
$ cd agora/tests/system/node/4
$ cd agora/tests/system/node/3
$ cd agora/tests/system/node/2
```

Before running each test I highly recommend you always clear the cache:

```
$ rm -r -- tests/system/node/[0-9]/.cache
```

You don't need to start node 0 because that's the FullNode, it's not relevant to the test.

Then within each terminal you just have to run the node (do this once for each of the 6 nodes):

```
../../../../build/agora -c config.yaml
```

Then, build and run [faucet](https://github.com/bpfkorea/faucet):

```
$ ./bin/faucet http://127.0.0.1:2826
```

You just need to look into the first node output (node 7 in my case) to see what goes wrong. After running faucet eventually the node will crash with:

```
2020-12-02 19:46:22,266 Info [agora.network.NetworkManager] - Node time: 1606905982
sock error 54!
2020-12-02 19:46:22,872 Info [agora.network.NetworkManager] - Node time: 1
core.exception.AssertError@source/agora/network/NetworkManager.d(561): Assertion failure
```

I added this assert there. The node time should never be `1`. It can be `0` in case of request failure, or the node's local unix time. If you remove the assert you'll see that the values tend to be small but random.

Now, if you check out the last commit 73a6527 and you reproduce the steps above then the bug is gone and the above assertion does not trigger anymore.

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -46,6 +46,11 @@ public struct NodeInfo
     public Set!string addresses;
 }
 
+struct LocalTime
+{
+    time_t time;
+}
+
 /*******************************************************************************
 
     Define the API a full node exposes to the world
@@ -241,5 +246,5 @@ public interface API
 
     ***************************************************************************/
 
-    public time_t getLocalTime ();
+    public LocalTime getLocalTime ();
 }
--- a/source/agora/common/crypto/Key.d
+++ b/source/agora/common/crypto/Key.d
@@ -157,9 +157,9 @@ public struct PublicKey
     public static PublicKey fromString (scope const(char)[] str) @trusted
     {
         const bin = Base32.decode(str);
-        assert(bin.length == 1 + PublicKey.Width + 2);
-        assert(bin[0] == VersionByte.AccountID);
-        assert(validate(bin[0 .. $ - 2], bin[$ - 2 .. $]));
+        enforce(bin.length == 1 + PublicKey.Width + 2);
+        enforce(bin[0] == VersionByte.AccountID);
+        enforce(validate(bin[0 .. $ - 2], bin[$ - 2 .. $]));
         return PublicKey(typeof(this.data)(bin[1 .. $ - 2]));
     }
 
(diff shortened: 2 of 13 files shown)


AndrejMitrovic:
@omerfirmak does this look interesting for you to look at?

omerfirmak:
If I can reproduce it on my end, I would love to take a look :)

AndrejMitrovic:
Great! I think it's also reproducible on Alphine (the docker file), but I had an easier time debugging when running it locally.

AndrejMitrovic:
I just reproduced this in WSL 1.0 Ubuntu as well. Both commits as described before.

omerfirmak:
It didnt work for me on the first try, but after a couple of tries multiple nodes crashed when the first block was externalized. I will see what I can find.

AndrejMitrovic:
Ah I forgot to mention one more step. Before running faucet, you should wait for about ~5 seconds so all the nodes reveal their preimages in time. Otherwise there will be a crash related to missing preimages. That itself is another issue that will be resolved once slashing is implemented.

AndrejMitrovic:
There is also an option called `preimage_reveal_period` or something like that. It can be reduced to a small value so this waiting time can be avoided.

AndrejMitrovic:
This was resolved upstream right? I'll close the PR.

PR #1404, opened 2020-12-04
Add a endpoint to provide a consensus params

MichaelKim20:
When creating transactions, SDK needs to calculate the fees in the data payload and address the commons-budget.
So I add an endpoint that provides Agora's ConsensusParams.

Related to https://github.com/bpfkorea/boa-sdk-ts/issues/63


AndrejMitrovic:
The values for CoinNet and TestNet will be public and known. There is no need to create an API endpoint for this.

MichaelKim20:
I have confirmed that there are many problems with using this.
I will handle it with the SDK as constants.

PR #1409, opened 2020-12-08
Don't allow fiber switch before quorums are shuffled

AndrejMitrovic:
This code was modified several times.. Ideally we would have a `schedule()` primitive in both vibe.d and localrest so we don't have to write these fixes at the call site (it's leaky abstractions).


linked0 (approved):

kchulj (approved):
TY

PR #1418, opened 2020-12-14
ValidatorSet should use the respective cycle_length of each individual enrollment

omerfirmak:
Currently, ValidatorSet uses the local parameters to validate and expire Enrollments. But enrollments can have different cycle lengths.

This revealed some tests use TestGenesis, which has enrollments with 20-block cycle length , but assume 1008-block cycle lengths. The default parameter for tests should be 20 to reflect the TestGenesis enrollments better.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -1093,11 +1093,12 @@ unittest
     assert(man.addEnrollment(enroll, key_pair.address, Height(1),
             utxo_set.getUTXOFinder()));
 
-    assert(man.params.ValidatorCycle - 101 == 907); // Sanity check
     assert(man.getValidatorPreimage(utxo_hash) == PreImageInfo.init);
     assert(man.addValidator(enroll, key_pair.address, Height(2), &utxo_set.peekUTXO,
             utxos) is null);
-    auto preimage = PreImageInfo(utxo_hash, man.cycle.preimages[100], 907);
+    auto preimage = PreImageInfo(utxo_hash,
+                        man.cycle.preimages[(man.params.ValidatorCycle / 2) - 1],
+                        cast (ushort) man.params.ValidatorCycle / 2);
     assert(man.addPreimage(preimage));
     assert(man.getValidatorPreimage(utxo_hash) == preimage);
 }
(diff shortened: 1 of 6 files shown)


hewison-chris:
This looks good but I am not sure if it should be posible to have a different cycle_length than that defined in the consensus params.
In tests we use 20 as it is more convenient than 1008.
@Geod24 Should it be possible to enroll with a different cycle_length than the one defined in the consensus params?

linked0:
> This looks good but I am not sure if it should be posible to have a different cycle_length than that defined in the consensus params.
> In tests we use 20 as it is more convenient than 1008.
> @Geod24 Should it be possible to enroll with a different cycle_length than the one defined in the consensus params?

I agree with @hewison-chris's opinion although this looks good. I think we don't have to have a different cycle for Enrollment because that value is shared with the `ConsensusParam` object through all the objects. That's why the `ConsensusParam` object is `immutable`.

omerfirmak:
ConsensusParams is immutable and shared across an instance of agora but there is no guarantee that it will be same for different instances. This is more about consistency of the different nodes' view of the ValidatorSet, then supporting different cycle_lengths for Enrollments.

linked0:
> ConsensusParams is immutable and shared across an instance of agora but there is no guarantee that it will be same for different instances. This is more about consistency of the different nodes' view of the ValidatorSet, then supporting different cycle_lengths for Enrollments.

You're right. There's no guarantee for that at the moment, I think.
But, we have no plan for different `cycle_length` in the CoinNet environment because it is a protocol-level value. If there's some validator trying to enroll with a different `cycle_length`, we should reject the enrollment. I think we should add checking code for the validator doing that in the `agora.consensus.validation.Enrollment` module later.
Thank you.

AndrejMitrovic:
> ConsensusParams is immutable and shared across an instance of agora but there is no guarantee that it will be same for different instances. This is more about consistency of the different nodes' view of the ValidatorSet, then supporting different cycle_lengths for Enrollments.

`ConsensusParams` will be the same for all nodes which use the same genesis hash - in other words all nodes which are part of that network. The reason it exists is because CoinNet / TestNet / UnittestNet (which we use in the tests & our private testing network) will all have slightly different settings. For example, in unittests we may want to have a lower enrollment cycle so we don't have to generate thousands of blocks just in order to be able to test behavior of the nodes as they enroll / un-enroll. And the other fields allow us to test things like faster / slower block creation rate, etc.. 

If two nodes are running on the same `*Net`, they will have the same `ConsensusParams` values. That's why the word `Consensus`  is in the name, consensus can only be reached if these values are the same.

A node running on `CoinNet` cannot reach consensus with a node running on `TestNet` / `UnittestNet`. They'll have different genesis blocks, the UTXOs will not match between them, and they'll never be able to reach consensus if they use different values for the other fields as well.

Now, it's not the best idea that these parameters are in the `NodeConfig` because someone could try to tweak these and cause problems for themselves. Ideally these options would be well hidden from the user. But for now, they're exposed in the `NodeConfig`.

omerfirmak:
> ConsensusParams will be the same for all nodes which use the same genesis hash - in other words all nodes which are part of that network

If this is the end goal, then like @linked0 mentioned, `cycle_length` in enrollment should probably be validated against the value in `ConsensusParams`.

> That's why the word Consensus is in the name, consensus can only be reached if these values are the same.

If I am correct, this is not enforced anywhere tho. I can get `Simple.d` test to pass with different `ConsensusParams.ValidatorCycle` for each node.

AndrejMitrovic:
> If this is the end goal, then like @linked0 mentioned, cycle_length in enrollment should probably be validated against the value in ConsensusParams.

Yeah it should.

> If I am correct, this is not enforced anywhere tho. I can get Simple.d test to pass with different ConsensusParams.ValidatorCycle for each node.

So I guess we never added these checks..

Initially we planned to allow the validator to set any cycle they want. Later this was changed to 1008 (for CoinNet), but the `cycle_length` remained in the `Enrollment`. I suppose it's still there in case we want to relax this rule in the future. I don't recall the reasoning why the hardcoded number is the way we went with.

omerfirmak:
Regardless, I still feel like the information on the chain should be used. If we want constant cycle lengths across the network, the incoming `Enrollment`s should be validated for that.  This would make it easier to change the behaviour in the future, if we ever decide to do that. Just my 2 cents.

Thank you all for the insights!

AndrejMitrovic:
> Regardless, I still feel like the information on the chain should be used. If we want constant cycle lengths across the network, the incoming Enrollments should be validated for that. This would make it easier to change the behaviour in the future, if we ever decide to do that. Just my 2 cents.

Yes I agree. It's not a controversial change. I think we should add the validation as well.

omerfirmak:
Can finally close this since #2489  happened

Issue #1403, opened 2020-12-03
Implement pushTransaction for the listener


AndrejMitrovic:
To allow Stoa to build its transaction pool it needs to know when transactions are added to the pool. Agora can push those transactions to listener via a new `pushTransaction` API, similar to `pushPreimage` / `pushBlock`.

Then, the wallet can ask Stoa if a specific tx is currently in the pool (pending).

This is a better alternative to https://github.com/bpfkorea/agora/issues/1402

PR #1420, opened 2020-12-16
Implement pushTransaction for the transaction received handler

TrustHenry:
To allow Stoa to build its transaction pool it needs to know when transactions are added to the pool.
This allows it to set the incoming address by adding 'transaction_received_handler_addresses' to the setting.
Received a transaction push only valid a transaction.

Related to #1403


AndrejMitrovic (approved):

AndrejMitrovic:
CI failure is unrelated.

PR #1421, opened 2020-12-16
Fix #1297: Make Enrollments valid for nomination on only a single block height

### LINKED ISSUE #1297: Enrollments are susceptible to replay attacks
Currently the Enrollment consists of:

```
Hash utxo_key;
Hash random_seed;
uint cycle_length;
Signature enroll_sig;
```

A Validator may re-enroll with the same UTXO key as a previous Enrollment. But that also means they can re-use their signature. In other words, they could just re-send the exact same Enrollment as before, and it would be accepted under current protocol rules.

This presents a problem: An attacker could take a previous Enrollment belonging to a Validator (perhaps a Validator that has un-enrolled long ago) and gossip it. There is a big risk to this: if an attacker collects enough Enrollments for a set of un-enrolled Validators which still have those UTXOs frozen, the attacker could cause a quorum shuffle where most of the newly created quorum network is offline. Therefore we lose liveness. And if we have any slashing rules for inactive validators then we risk punishing innocent users.

As discussed with @Geod24 there are a few ways of fixing this:

- Check that the `random_seed` is different from the previous enrollment. But this requires keeping this state in memory for as long as the frozen UTXO exists (which could be arbitrarily long).
- Add a `max_block_height` to the Enrollment. When a node submits an Enrollment we can check if `block_height + enroll.cycle_length < enroll.max_block_height`, and then only accept the enrollment into a block if this condition is satisfied. This would also mean that if the Enrollment was rejected the node would have to create a new Enrollment with a new `max_block_height`.

There might be many more ways of fixing this. It would be good to research and come up with an optimal solution.

omerfirmak:
With this, `PreImage`s are consumed not by new enrollments but by new blocks. Commitments in new enrollments should be the `PreImage` that would be revealed on the height that Enrollment would be available.

More background info at https://github.com/bpfkorea/agora/issues/1297

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -238,8 +238,8 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public Enrollment[] getEnrollments (ref Enrollment[] enrolls, Height height)
-        @trusted nothrow
+    public Enrollment[] getEnrollments (ref Enrollment[] enrolls, Height height,
+                                    scope UTXOFinder peekUTXO) @trusted nothrow
     {
         enrolls.length = 0;
         assumeSafeAppend(enrolls);
@@ -248,16 +248,11 @@ public class EnrollmentManager
         this.enroll_pool.getEnrollments(pool_enrolls, Height(height + 1));
         foreach (enroll; pool_enrolls)
         {
-            const enroll_height = this.getEnrolledHeight(enroll.utxo_key);
-            const avail_height =
-                this.enroll_pool.getAvailableHeight(enroll.utxo_key);
-            assert(avail_height != Height(0));
-            if (enroll_height == ulong.max ||
-                (avail_height >= enroll_height &&
-                    height >= enroll_height + this.params.ValidatorCycle - 1))
-            {
+            UTXO utxo;
+            if (peekUTXO(enroll.utxo_key, utxo) &&
+                this.isInvalidCandidateReason(enroll, utxo.output.address,
+                                    Height(height + 1), peekUTXO) is null)
                 enrolls ~= enroll;
-            }
         }
         return enrolls;
     }
@@ -942,7 +937,7 @@ unittest
     }
 
     Enrollment[] enrolls;
-    man.getEnrollments(enrolls, Height(1));
+    man.getEnrollments(enrolls, Height(1), &utxo_set.peekUTXO);
     assert(enrolls.length == 3);
     assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
 
@@ -967,7 +962,7 @@ unittest
     assert(man.addValidator(ordered_enrollments[0], pairs[0].address, Height(9),
             utxo_set.getUTXOFinder(), utxos) !is null);
     assert(man.getEnrolledHeight(ordered_enrollments[1].utxo_key) == ulong.max);
-    man.getEnrollments(enrolls, Height(9));
+    man.getEnrollments(enrolls, Height(9), &utxo_set.peekUTXO);
     assert(enrolls.length == 1);
     // One Enrollment was moved to validator set
     assert(man.validator_set.count() == 1);
@@ -976,7 +971,8 @@ unittest
     man.removeEnrollment(utxo_hashes[0]);
     man.removeEnrollment(utxo_hashes[1]);
     man.removeEnrollment(utxo_hashes[2]);
-    assert(man.getEnrollments(enrolls, Height(9)).length == 0);
+    assert(man.getEnrollments(enrolls, Height(9), &utxo_set.peekUTXO)
+                                                                .length == 0);
 
     // clear up all validators
     man.validator_set.removeAll();
@@ -986,7 +982,8 @@ unittest
     foreach (idx, ordered_enroll; ordered_enrollments)
         assert(man.addEnrollment(ordered_enroll, pairs[idx].address, Height(10),
             &utxo_set.peekUTXO));
-    man.getEnrollments(enrolls, Height(man.params.ValidatorCycle + 8));
+    man.getEnrollments(enrolls, Height(man.params.ValidatorCycle + 8),
+                                                    &utxo_set.peekUTXO);
     assert(enrolls.length == 3);
     assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
 
@@ -1257,26 +1254,26 @@ unittest
 
     // if the current height is smaller than the available height,
     // we can get no enrollment
-    man.getEnrollments(enrolls, Height(8));
+    man.getEnrollments(enrolls, Height(8), &utxo_set.peekUTXO);
     assert(enrolls.length == 0);
 
     // if the current height is greater than or equal to the available height,
     // we can get enrollments
-    man.getEnrollments(enrolls, Height(9));
+    man.getEnrollments(enrolls, Height(9), &utxo_set.peekUTXO);
     assert(enrolls.length == 1);
 
     // make the enrollment a validator
     man.addValidator(enroll, key_pair.address, Height(10), &utxo_set.peekUTXO, utxos);
-    man.getEnrollments(enrolls, Height(11));
+    man.getEnrollments(enrolls, Height(11), &utxo_set.peekUTXO);
     assert(enrolls.length == 0);
 
     // add the enrollment that is already a validator, and check if
     // the enrollment can be nominated at the height before the cycle end
     enroll = man.createEnrollment(utxo_set.keys[0], Height(10 + validator_cycle));
     assert(man.addEnrollment(enroll, key_pair.address, Height(11), &utxo_set.peekUTXO));
-    man.getEnrollments(enrolls, Height(validator_cycle + 8));
+    man.getEnrollments(enrolls, Height(validator_cycle + 8), &utxo_set.peekUTXO);
     assert(enrolls.length == 0);
-    man.getEnrollments(enrolls, Height(validator_cycle + 9));
+    man.getEnrollments(enrolls, Height(validator_cycle + 9), &utxo_set.peekUTXO);
     assert(enrolls.length == 1);
 
     // make the enrollment a validator again
@@ -1289,7 +1286,7 @@ unittest
     // constraint
     assert(!man.addEnrollment(enroll, key_pair.address, Height(validator_cycle + 10),
         &utxo_set.peekUTXO));
-    man.getEnrollments(enrolls, Height(validator_cycle + 11));
+    man.getEnrollments(enrolls, Height(validator_cycle + 11), &utxo_set.peekUTXO);
     assert(enrolls.length == 0);
 }
 
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -220,7 +220,8 @@ public class EnrollmentManager
             return false;
         }
 
-        return this.enroll_pool.add(enroll, Height(avail_height), finder);
+        return this.enroll_pool.add(enroll, Height(avail_height), finder,
+                                    &this.validator_set.findRecentEnrollment);
     }
 
     /***************************************************************************
@@ -447,7 +448,8 @@ public class EnrollmentManager
         if (this.isPublicKeyEnrolled(enroll.utxo_key, pubkey))
             return "Enrollment: The same public key is already present";
 
-        if (auto fail_reason = enroll.isInvalidReason(findUTXO))
+        if (auto fail_reason = enroll.isInvalidReason(findUTXO, height,
+                                &this.validator_set.findRecentEnrollment))
             return fail_reason;
 
         const enrolled = this.validator_set.getEnrolledHeight(enroll.utxo_key);
@@ -994,7 +996,7 @@ unittest
             &utxo_set.peekUTXO, utxos) is null);
     assert(man.getNextPreimage(preimage, Height(10)));
     assert(preimage.hash ==
-        man.cycle.preimages[$ - 1 - man.PreimageRevealPeriod]);
+        man.cycle.preimages[$ - 1 - man.PreimageRevealPeriod - Height(10).value]);
 
     // test for getting validators' UTXO keys
     Hash[] keys;
@@ -1038,7 +1040,7 @@ unittest
     Hash[] utxo_hashes = utxo_set.keys;
 
     auto utxo_hash = utxo_hashes[0];
-    auto enroll = man.createEnrollment(utxo_hash, Height(1));
+    auto enroll = man.createEnrollment(utxo_hash, Height(2));
     assert(man.addEnrollment(enroll, key_pair.address, Height(1),
             utxo_set.getUTXOFinder()));
 
@@ -1046,7 +1048,7 @@ unittest
     assert(man.getValidatorPreimage(utxo_hash) == PreImageInfo.init);
     assert(man.addValidator(enroll, key_pair.address, Height(2), &utxo_set.peekUTXO,
             utxos) is null);
-    auto preimage = PreImageInfo(utxo_hash, man.cycle.preimages[100], 907);
+    auto preimage = PreImageInfo(utxo_hash, man.cycle.preimages[98], 907);
     assert(man.addPreimage(preimage));
     assert(man.getValidatorPreimage(utxo_hash) == preimage);
 }
@@ -1264,28 +1266,29 @@ unittest
     assert(enrolls.length == 1);
 
     // make the enrollment a validator
-    man.addValidator(enroll, key_pair.address, Height(11), &utxo_set.peekUTXO, utxos);
+    man.addValidator(enroll, key_pair.address, Height(10), &utxo_set.peekUTXO, utxos);
     man.getEnrollments(enrolls, Height(11));
     assert(enrolls.length == 0);
 
     // add the enrollment that is already a validator, and check if
     // the enrollment can be nominated at the height before the cycle end
+    enroll = man.createEnrollment(utxo_set.keys[0], Height(10 + validator_cycle));
     assert(man.addEnrollment(enroll, key_pair.address, Height(11), &utxo_set.peekUTXO));
-    man.getEnrollments(enrolls, Height(validator_cycle + 9));
+    man.getEnrollments(enrolls, Height(validator_cycle + 8));
     assert(enrolls.length == 0);
-    man.getEnrollments(enrolls, Height(validator_cycle + 10));
+    man.getEnrollments(enrolls, Height(validator_cycle + 9));
     assert(enrolls.length == 1);
 
     // make the enrollment a validator again
     man.clearExpiredValidators(Height(validator_cycle + 11));
-    man.addValidator(enroll, key_pair.address, Height(validator_cycle + 11),
-        &utxo_set.peekUTXO, utxos);
+    assert(man.addValidator(enroll, key_pair.address, Height(validator_cycle + 11),
+        &utxo_set.peekUTXO, utxos));
 
     // add the enrollment that has the available height smaller than
-    // the enrolled height of the validator, we can get no enrollment
-    assert(man.addEnrollment(enroll, key_pair.address, Height(validator_cycle + 10),
+    // the enrolled height of the validator, it will fail because of the commitment
+    // constraint
+    assert(!man.addEnrollment(enroll, key_pair.address, Height(validator_cycle + 10),
         &utxo_set.peekUTXO));
-    assert(man.enroll_pool.count() == 1);
     man.getEnrollments(enrolls, Height(validator_cycle + 11));
     assert(enrolls.length == 0);
 }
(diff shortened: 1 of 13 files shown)


source/agora/consensus/state/ValidatorSet.d
```diff
@@ -571,6 +593,50 @@ public class ValidatorSet
...
+        Params:
+            enroll_key = The key for the enrollment
+            state = struct to fill once an enrollment is found
+
+        Returns:
+            true if an enrollment with the key was ever accepted, false otherwise
+
+    ***************************************************************************/
+
+    bool findRecentEnrollment (in Hash enroll_key, out EnrollmentState state) @trusted nothrow
```
AndrejMitrovic, line 612: Could use a test-case in the module.

source/agora/consensus/PreImage.d
```diff
@@ -390,4 +391,98 @@ public struct PreImageCycle
...
+        This will calculate the `index` and `nonce` according to the given
+        height and populate the `seed` and `preimages` if neccesary.
+
+        This will be particularly usefull since the PreImages will now be
+        consumed not by new enrollments but creation of new blocks.
+
+        Params:
+          secret = The secret key of the node, used as part of the hash
+                   to generate the cycle seeds
+          height = Current height
```
AndrejMitrovic: current => desired/requested, it's not necessarily the latest height.

source/agora/consensus/PreImage.d
```diff
@@ -390,4 +391,98 @@ public struct PreImageCycle
...
+
+    ***************************************************************************/
+
+    public void seek (scope const ref Scalar secret, Height height) @safe @nogc nothrow
```
AndrejMitrovic: Should this be public?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -323,13 +311,13 @@ public class EnrollmentManager
 
     ***************************************************************************/
 
-    public Enrollment createEnrollment (in Hash utxo) @safe nothrow
+    public Enrollment createEnrollment (in Hash utxo, Height height) @safe nothrow
```
AndrejMitrovic, line 311: Needs `Params` for `height`

source/agora/consensus/validation/Block.d
```diff
@@ -468,12 +479,36 @@ public bool isGenesisBlockValid (const ref Block genesis_block)
...
-    size_t active_enrollments, scope PayloadChecker checkPayload) nothrow @safe
+    size_t active_enrollments, scope PayloadChecker checkPayload,
+    scope EnrollmentFinder findEnrollment) nothrow @safe
 {
     return isInvalidReason(block, prev_height, prev_hash, findUTXO,
-        active_enrollments, checkPayload) is null;
+        active_enrollments, checkPayload, findEnrollment) is null;
+}
+
+public EnrollmentFinder getGenesisEnrollmentFinder () nothrow @trusted
```
AndrejMitrovic, line 494: This is public but undocumented.

AndrejMitrovic (changes requested):
A few comments.

AndrejMitrovic:
@linked0 could you also take a look at this and tell us what you think?

source/agora/consensus/state/ValidatorSet.d
omerfirmak: Added the test for it in the `EnrollmentManager.d`

omerfirmak:
Thanks for the review @AndrejMitrovic , all of your comments should be addressed now.

AndrejMitrovic:
It would be good to add a network test in `source/agora/test/*`. There are some examples there where a custom node is instantiated which behaves badly, a similar test could be added where a node tries enroll (or re-enroll) with an enrollment with the wrong `random_seed`.

linked0:
> @linked0 could you also take a look at this and tell us what you think?

I'm reviewing it now. Sorry for the late review. Anyway, it looks really awesome. I will leave some comments or suggestions soon.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -248,16 +248,11 @@ public class EnrollmentManager
         this.enroll_pool.getEnrollments(pool_enrolls, Height(height + 1));
         foreach (enroll; pool_enrolls)
         {
-            const enroll_height = this.getEnrolledHeight(enroll.utxo_key);
-            const avail_height =
```
linked0, line 256: This part of the code should not be removed because it's not related to the current PR. The `avail_height` is to check whether some enrollments can be nominated for the next height or not. Some validators can try to re-enroll much earlier before the enrollment expires. For example, a node can be enrolled at the height of 10, and it can try to re-enroll at the height of 11 after being enrolled, and then the `Enrollment` is added to the pool. The added `Enrollment` can be nominated for the height of 1017. That is the original concept for the `avail_height`. If we decide to only use `checkAndEnroll` as the way of re-enrolling, we should remove all the code `avail_height`. Or the code about `avail_height` should remain.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1264,28 +1266,29 @@ unittest
     assert(enrolls.length == 1);
 
     // make the enrollment a validator
-    man.addValidator(enroll, key_pair.address, Height(11), &utxo_set.peekUTXO, utxos);
+    man.addValidator(enroll, key_pair.address, Height(10), &utxo_set.peekUTXO, utxos);
```
linked0, line 1266: I don't see why the `Height` changed into 10. The part of the code means that the enrollments are nominated at the height of 10 and enrolled at the height of 11, and there are no enrollments in the pool after enrollment. So this should be 11. This change does not match with the basic flow of the enrollment process.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1264,28 +1266,29 @@ unittest
...
     man.getEnrollments(enrolls, Height(11));
     assert(enrolls.length == 0);
 
     // add the enrollment that is already a validator, and check if
     // the enrollment can be nominated at the height before the cycle end
+    enroll = man.createEnrollment(utxo_set.keys[0], Height(10 + validator_cycle));
```
linked0, line 1272: The original purpose is to re-enroll with the previously used `Enrollment`. So I don't think it's needed.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1264,28 +1266,29 @@ unittest
...
     assert(man.addEnrollment(enroll, key_pair.address, Height(11), &utxo_set.peekUTXO));
-    man.getEnrollments(enrolls, Height(validator_cycle + 9));
+    man.getEnrollments(enrolls, Height(validator_cycle + 8));
```
linked0: I think it should be restored as it was because the change does not contain the original purpose.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1264,28 +1266,29 @@ unittest
...
     assert(enrolls.length == 0);
-    man.getEnrollments(enrolls, Height(validator_cycle + 10));
+    man.getEnrollments(enrolls, Height(validator_cycle + 9));
```
linked0: Ditto

linked0 (commented):
Good work and very important code change.
But the code about the `avail_height` was removed unintentionally. Some test code that changed does not go with the original purpose. Please check the comments.

omerfirmak:
> It would be good to add a network test in `source/agora/test/*`. There are some examples there where a custom node is instantiated which behaves badly, a similar test could be added where a node tries enroll (or re-enroll) with an enrollment with the wrong `random_seed`.

Yeah, I wanted to get some feedback before writing a network test.

source/agora/consensus/EnrollmentManager.d
omerfirmak: With these changes, `avail_height`  is actually completely redundant. This information is implicitly encoded in the `random_seed` since `random_seed` will only be valid for a single height. The validation checks in this function will make sure an enrollment in the pool will only be nominated at the height that it is valid.

We can get rid of `avail_height` in a separate PR I guess.

omerfirmak: and sending an early Enrollment request may not be desired any more. If you enroll at `Height(0)` with a validator cycle of 1008. You would have to reveal the preimage of `Height(1008)` prematurely just to send an enrollment request for the next cycle.

linked0: > and sending an early Enrollment request may not be desired any more.

This is the feature we already had, which is done in issue #824 . A validator can re-enroll much earlier before a cycle ends in the situation where `recurring_enrollment` is set to `false`. So the use of `avail_height` could be removed after some discussion and decision.

omerfirmak: By early I meant sending a request at `Height(1)` for an enrollment that would be available at `Height(1008)`. The behaviour described in #824 is still possible.

omerfirmak: > By early I meant sending a request at Height(1) for an enrollment that would be available at Height(1008).

this is possible too, just requires the node to reveal a preimage way early.

linked0: > this is possible too, just requires the node to reveal a preimage way early.

OK, I see what you intended in the code now. Thanks.

source/agora/consensus/EnrollmentManager.d
omerfirmak: This test actually enrolls the Validator one block late. I dont know if that was intentional. But with the new rules, this is not possible. All the height changes below are required because of this change, you cant enroll with the same enrollment more then once and you can use the enrollment only on a single height.

If a node does `createEnrollment(height = n)` and sends the request, that enrollment can only be included in the block #n. So the nodes create the enrollment for `Height(n)` at `Height(n-1)`.

See `checkAndEnroll`, it creates the enrollment for `block_height + 1` and that enrollment gets included in the next block.
https://github.com/bpfkorea/agora/blob/681fd853c1cb40c1df11959d47e0d81a9baca37d/source/agora/node/Validator.d#L466

source/agora/consensus/EnrollmentManager.d
omerfirmak: Ditto

source/agora/consensus/EnrollmentManager.d
omerfirmak: Ditto

source/agora/consensus/EnrollmentManager.d
omerfirmak: Counter ditto :)

source/agora/consensus/EnrollmentManager.d
linked0: > If a node does `createEnrollment(height = n)` and sends the request, that enrollment can only be included in the block #n. So the nodes create the enrollment for `Height(n)` at `Height(n-1)`.
> 
> See `checkAndEnroll`, it creates the enrollment for `block_height + 1` and that enrollment gets included in the next block.

I see what you mean. I know the situation is kind of confusing because the `avail_height` remains. :) 
With your reasoning, the code below should be succeeded because you create an enrollment for Height(n) and try to enroll at the Height(n-1), so the calling `getEnrollments` at the Height(n-1) should return 1.

```
 man.getEnrollments(enrolls, Height(9)); 
 assert(enrolls.length == 1);
```

omerfirmak: https://github.com/omerfirmak/agora/blob/681fd853c1cb40c1df11959d47e0d81a9baca37d/source/agora/consensus/EnrollmentManager.d#L248

`EnrollmentManager.getEnrollments` does the +1 trick as well.

https://github.com/omerfirmak/agora/blob/681fd853c1cb40c1df11959d47e0d81a9baca37d/source/agora/node/Ledger.d#L446

`Ledger` calls `getEnrollments` with the current height, `EnrollmenManager` request the enrollments valid for next block from the `EnrollmentPool`

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1300,44 +1248,45 @@ unittest
...
-    auto enroll = man.createEnrollment(utxo_set.keys[0]);
+    auto enroll = man.createEnrollment(utxo_set.keys[0], Height(10));
     assert(man.addEnrollment(enroll, key_pair.address, Height(10),
             utxo_set.getUTXOFinder()));
 
     // if the current height is smaller than the available height,
     // we can get no enrollment
-    man.getEnrollments(enrolls, Height(9));
+    man.getEnrollments(enrolls, Height(9), &utxo_set.peekUTXO);
     assert(enrolls.length == 0);
```
linked0, line 1258: > Ledger calls getEnrollments with the current height, EnrollmenManager request the enrollments valid for next block from the EnrollmentPool

Ledger calls getEnrollments with the current height(n-1 = 9), EnrollmenManager request the enrollments valid for next block(n = 10) from the EnrollmentPool.
This is what I meant.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1300,44 +1248,45 @@ unittest
 
     // create and add the first enrollment
     Enrollment[] enrolls;
-    auto enroll = man.createEnrollment(utxo_set.keys[0]);
+    auto enroll = man.createEnrollment(utxo_set.keys[0], Height(10));
     assert(man.addEnrollment(enroll, key_pair.address, Height(10),
```
linked0:
```suggestion
    assert(man.addEnrollment(enroll, key_pair.address, Height(9),
```

omerfirmak:
Added the network test.

source/agora/consensus/EnrollmentManager.d
omerfirmak: Ok I see what you mean now :+1:  This should work I guess? Should be identical to what `Ledger` does.

``` diff
diff --git a/source/agora/consensus/EnrollmentManager.d b/source/agora/consensus/EnrollmentManager.d
index 6ec43d67..e88b56e3 100644
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -1249,17 +1249,17 @@ unittest
     // create and add the first enrollment
     Enrollment[] enrolls;
     auto enroll = man.createEnrollment(utxo_set.keys[0], Height(10));
-    assert(man.addEnrollment(enroll, key_pair.address, Height(10),
+    assert(man.addEnrollment(enroll, key_pair.address, Height(9),
             utxo_set.getUTXOFinder()));
 
     // if the current height is smaller than the available height,
     // we can get no enrollment
-    man.getEnrollments(enrolls, Height(9), &utxo_set.peekUTXO);
+    man.getEnrollments(enrolls, Height(8), &utxo_set.peekUTXO);
     assert(enrolls.length == 0);
 
     // if the current height is greater than or equal to the available height,
     // we can get enrollments
-    man.getEnrollments(enrolls, Height(10), &utxo_set.peekUTXO);
+    man.getEnrollments(enrolls, Height(9), &utxo_set.peekUTXO);
     assert(enrolls.length == 1);
 ```

Comment (omerfirmak) [reply source/agora/consensus/EnrollmentManager.d]: Ok, I appllied the diff above.

Comment (linked0) [reply source/agora/consensus/EnrollmentManager.d]: > Ok I see what you mean now 👍 This should work I guess? Should be identical to what `Ledger` does.

Yes, That's what I meant. Thanks.

**Comment by omerfirmak** [comment]
Thanks for the review, @linked0 . I have rebased to `v0.x.x`.

File: source/agora/test/ValidatorRecurringEnrollment.d [hunk 13]
```diff
@@ -63,3 +63,114 @@ unittest
...
+    static class BadEnrollmentManager : EnrollmentManager
+    {
+        public this (string db_path, KeyPair key_pair,
+            immutable(ConsensusParams) params)
+        {
+            super(db_path, key_pair, params);
+        }
+
+        public override Enrollment createEnrollment (in Hash utxo,
+                                                Height height) @safe nothrow
```
linked0: Use only four spaces for indentation, please.

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -63,3 +63,114 @@ unittest
     // Check if all validators in genesis are enrolled again
     assert(blocks[blocks.length - 1].header.enrollments.length == blocks[0].header.enrollments.length);
 }
+
+// Recurring enrollment with wrong `random_seed`
```
linked0, line 67: A more detailed description needed.
Refer to other test code, please.
https://github.com/bpfkorea/agora/blob/a92aead454b8e93a79ef7c9fccec7f1d4aaacabe/source/agora/test/ValidatorCleanRestart.d#L116-L120

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -63,3 +63,114 @@ unittest
...
+        {
+            super(config, reg, blocks, test_conf, cur_time);
+        }
+
+        protected override EnrollmentManager getEnrollmentManager (
+            string data_dir, in ValidatorConfig validator_config,
+            immutable(ConsensusParams) params)
+        {
+            return new BadEnrollmentManager(":memory:",
+                                            validator_config.key_pair, params);
```
linked0: Ditto

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -63,3 +63,114 @@ unittest
...
+        }
+    }
+
+    static class BadAPIManager : TestAPIManager
+    {
+
+        public this (immutable(Block)[] blocks, TestConf test_conf,
+                                                            time_t initial_time)
```
linked0: Use only four spaces for indentation, please.

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -63,3 +63,114 @@ unittest
...
+        {
+            super(blocks, test_conf, initial_time);
+        }
+
+        public override void createNewNode (Config conf, string file = __FILE__,
+                                                            int line = __LINE__)
```
linked0: Ditto

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -63,3 +63,114 @@ unittest
...
+    auto blocks = node_1.getBlocksFrom(0, 2);
+    assert(blocks.length == 1);
+
+    Transaction[] txs;
+
+    void createAndExpectNewBlock (Height new_block_height)
+    {
+        // create enough tx's for a single block
+        txs = blocks[new_block_height - 1].spendable().map!(txb => txb.sign())
+                                                                    .array();
```
linked0: Ditto

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -63,3 +63,114 @@ unittest
...
+
+    // create GenesisValidatorCycle - 1 blocks
+    foreach (block_idx; 1 .. GenesisValidatorCycle)
+    {
+        createAndExpectNewBlock(Height(block_idx));
+    }
+
+    // Try creating one last block, should fail
+    assertThrown!AssertError(createAndExpectNewBlock(
+                                                Height(GenesisValidatorCycle)));
```
linked0: Ditto

source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -63,3 +63,114 @@ unittest
...
+            immutable(ConsensusParams) params)
+        {
+            return new BadEnrollmentManager(":memory:",
+                                            validator_config.key_pair, params);
+        }
+    }
+
+    static class BadAPIManager : TestAPIManager
+    {
+
```
linked0: Unnecessary line

linked0 (approved):
Great work.
Some suggestions.

source/agora/test/ValidatorRecurringEnrollment.d
omerfirmak: I feel like keeping stuff that overflow to the second line close to their original horizontal position gives a more readable result. But I'll change it to keep the codebase consistent :+1: Thanks!

AndrejMitrovic: It also spends a lot more whitespace if it's near the end of the line. E.g.:

```D
public override void foo (in int par1,
                          in int par2,
                          in int par3,
                          in int par4)
```

vs:

```D
public override void foo (in int par1,
    in int par2, int par3, int par4)
```

Actually in your case since you're already using a new line, why not just do:

```D
public override Enrollment createEnrollment (
    in Hash utxo, Height height )
```

Then they're grouped together nicely?

omerfirmak: That open parenthesis, looks so sad and alone  :smile:

AndrejMitrovic: It's like a more chonky egyptian brace.

omerfirmak:
``` d
public override Enrollment createEnrollment (
    in Hash utxo, Height height 
) @safe {

}
```
How about this? With no func attribute, it looks weird tho :smile:

AndrejMitrovic: It's very unusual compared to most D projects. Let's just be consistent with what we have, and later if we want to change anything we can use a code styling app to change the layout project-wide.

omerfirmak:
I've applied the changes you suggested @linked0. I'll take a look at the CircleCI failure. Integration tests on my local machine has no problems, its weird.

omerfirmak:
All green

AndrejMitrovic (approved):

AndrejMitrovic:
Thank you @omerfirmak, nice job!

Thanks @linked0 for the review!

PR #1424, opened 2020-12-16
Remove unnecessary test code about slashing protocol

linked0:
We do not have the `preimage_root` in the `ConsensusData` structure anymore. But, there remains some code about testing the `preimage_root` information. So we remove the unnecessary code.

@ferencdg have suggested removing the code. So I remove the code that might make unexpected side effects.


AndrejMitrovic (approved):

AndrejMitrovic:
It seems CircleCI got stuck? Could you try a `git amend` and push? This should trigger the CI to run again.

Issue #1428, opened 2020-12-17
Implement support for non-interactive threshold signatures (linear complexity, not MuSig)


AndrejMitrovic:
As of today there are only a few (maybe just one?) non-interactive Schnorr threshold signature schemes which exist. One of them is MuSig. There are some resources about it here:

- https://suredbits.com/schnorr-applications-threshold-signatures/
- https://blockstream.com/2018/01/23/en-musig-key-aggregation-schnorr-signatures/
- https://eprint.iacr.org/2018/068.pdf
- https://tlu.tarilabs.com/cryptography/musig-schnorr-sig-scheme/The_MuSig_Schnorr_Signature_Scheme.html

As of now I am not sure whether we would need any special support from Agora's side in order to validate MuSig signatures, or whether it's only the signing process that is complex and the signature validation process remains the same. If that's the case, then only the wallet would need to add support for MuSig. But since we've outsourced wallet development I don't have high hopes of it being properly implemented in due time. 

Since implementing support for MuSig is non-trivial it's not a bad idea to implement a stop-gap measure by adding support for multi-sigs in the same way as Bitcoin's `CHECK_MULTI_SIG`.

If we do end up delivering MuSig support before CoinNet we can then just remove the `CHECK_MULTI_SIG` opcode.

AndrejMitrovic:
Resolved.

PR #1429, opened 2020-12-17
Engine: Implement support for threshold signatures

AndrejMitrovic:
See https://github.com/bpfkorea/agora/issues/1428 for more info. If we don't get MuSig support in the Wallet soon then this is the scheme we should be using by default.

Fixes https://github.com/bpfkorea/agora/issues/1428


AndrejMitrovic:
Merging as no comments received and the wallet needs it.

PR #1430, opened 2020-12-17
Update d2sqlite3 and make it easier to access the DB

Geod24:
One issue I've seen in the live system is that it's hard to debug anything because the DB is unreadable.
Having binary data in the sqlite kind of defeat a great benefit of having sqlite in the first place, hence this change.
Open for debate.


AndrejMitrovic:
It makes sense to me

AndrejMitrovic:
There seems to be some flaky test triggering on mac + ldc. But it's not related to your PR.

AndrejMitrovic (approved):

PR #1436, opened 2020-12-18
Link system tests with sqlite3

omerfirmak:
[This](https://app.circleci.com/pipelines/github/bpfkorea/agora/5317/workflows/acd127b1-d13a-4ac5-86d8-3c24780a7360/jobs/5624) shows that system test now depend on sqlite somehow.

### DIFF
--- a/tests/system/dub.json
+++ b/tests/system/dub.json
@@ -3,8 +3,8 @@
     "targetType": "executable",
     "targetPath": "build",
     "dflags": [ "-i" ],
-    "libs-posix": [ "sodium" ],
-    "libs-windows": [ "libsodium" ],
+    "libs-posix": [ "sodium", "sqlite3" ],
+    "libs-windows": [ "libsodium", "libsqlite3" ],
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
Ah so it worked fine in the CI but not locally without it?

omerfirmak:
> Ah so it worked fine in the CI but not locally without it?

Nah, it didnt work locally either :) I am checking if I introduced an unnecessary import somewhere.

omerfirmak:
Even with narrow imports like this cause sql to be linked to system tests. Although `EnrollmentFinder` , `EnrollmentState` , `EnrollmentStatus` alone should not require db. We might have to link sqlite3.

``` d
import agora.consensus.state.ValidatorSet : EnrollmentFinder, EnrollmentState,
                                            EnrollmentStatus;
```

omerfirmak:
I am closing this and adding this commit to #1421 since it is causing this.

PR #1437, opened 2020-12-18
Add timestamp to the block

### LINKED ISSUE #1406: Add a timestamp to the block
Currently our goal is for the Agora network to produce a block every 10 minutes. However, this cannot be a hard requirement as the nodes may fail to produce the blocks in case of network outages, or in case there is inability to reach consensus in due time. 

Due to this unpredictable nature of the block creation frequency it's not possible to accurately timestamp transactions based solely on the block height of the block in which a transaction was included.

A UTC unix timestamp should be added to the block. This field will be part of the hash and therefore signed.

Since we do not have miner nodes we need to use SCP to reach consensus on the timestamp which will be set in the block. This does create some problems we need to solve:

- When should the timestamp be proposed? Should it be together with the `ConsensusData`? The problem here is that the total time it takes to **reach consensus** is unbounded, which means that by the time consensus is reached the proposed timestamp could already be quite old. Should the timestamp instead be proposed outside of `ConsensusData` in a separate round of voting after the transaction consensus has been reached? There is nothing preventing us from using SCP to vote on a separate proposal, so we might go with this approach. Voting on timestamps should be less contentious than voting on the transaction set.

To get the PoC out of the way, I suggest adding the timestamp to the `ConsensusData`. In a later implementation, we could decide to go with the alternate approach where we do a second round of voting for the timestamp. 

- How do the nodes agree on a common timestamp? In which cases would another node accept or reject a proposed timestamp? We could use [Clock.networkTime()](https://github.com/bpfkorea/agora/blob/3402716cf39f69b31995ba205730bd189aa761ab/source/agora/network/Clock.d#L109) for the baseline. But then we would need to define how much the proposed timestamp is allowed to deviate from the `networkTime()`.

I suggest adding some consensus parameter values such as `max_timestamp_deviation` which would be the max number of seconds that the proposed timestamp in `ConsensusData` may deviate from the `Clock.networkTime()`. We should also make it a rule that the proposed timestamp may only be `>=` previous block timestamp.

ferencdg:
The nodes reuse the consensus algorithm to make a joint decision on what timestamp should go into the block.

fixes: #1406


ferencdg:
Still in draft and I am planning to add more unit tests and fix up smaller things like method signatures/style. Please comment on the general idea/design or if something like MockClock triggers you.

source/agora/consensus/validation/Block.d
```diff
@@ -136,7 +140,31 @@ public string isInvalidReason (const ref Block block, Height prev_height,
...
+            res = "the timestamp in consensus data is not greater than the timestamp in the last block\n" ~
+                    "new block timestamp: " ~ to!string(SysTime(unixTimeToStdTime(new_block_ts))) ~ "\n" ~
+                    "previous block's timestamp: " ~ to!string(SysTime(unixTimeToStdTime(prev_block_ts)));
+        else if (new_block_ts > curr_timestamp + blockTimestampToleranceSecs)
+            res =  "(the timestamp in consensus data) is greater than (the timestamp in the last block + tolerance)\n" ~
+                    "new block timestamp: " ~ to!string(SysTime(unixTimeToStdTime(new_block_ts))) ~ "\n" ~
+                    "current timestamp: " ~ to!string(SysTime(unixTimeToStdTime(curr_timestamp))) ~ "\n" ~
+                    "tolerance in seconds: " ~to!string(blockTimestampToleranceSecs);
+    }
+    catch (Exception e) {}
```
omerfirmak: If the `try` block throws without setting `res`, this would return valid here. Should `res` be set to something in the `catch` block?

ferencdg: yeah, forgot to remove that, just did it for testing, thanks

source/agora/utils/Log.d
```diff
@@ -120,7 +120,7 @@ static this ()
 }
 
 /// Circular appender which appends to an internal buffer
-public class CircularAppender (size_t BufferSize = 2^^20) : Appender
+public class CircularAppender (size_t BufferSize = 2^^22) : Appender
```
ferencdg: ugh, let me remove this

AndrejMitrovic:
Is it possible to split off the bugfixes into a separate PR?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -324,10 +324,6 @@ extern(D):
```
AndrejMitrovic: I don't think this is correct. The node itself starts a periodic timer which recursively calls itself until nomination is over. See here: https://github.com/bpfkorea/agora/blob/b6c7b12b9c8dc3fbc88f6e3f8c8dbefd29af5a08/source/scpp/src/scp/NominationProtocol.cpp#L522-L528

Also if this was a bugfix it should have come with a test-case where the test fails without the bugfix but succeeds with the bugfix.

source/agora/common/Config.d
```diff
@@ -114,11 +114,6 @@ public struct NodeConfig
     /// and TODO: addresses should be different prefix (e.g. TN... for TestNet)
     public bool testing;
 
-    /// The Unix timestamp of the Genesis block. Blocks are created at a
-    /// regular interval starting from this time.
-    public time_t genesis_start_time =
```
AndrejMitrovic, line 119: Why did you remove this?

AndrejMitrovic: We should still have a target average for the block interval, which should be based on the genesis start time.

AndrejMitrovic: Ah you moved it to the genesis block header. That's alright then, but it would be good to make a comment in the commit message.

source/agora/consensus/data/Params.d
```diff
@@ -69,7 +69,9 @@ public immutable class ConsensusParams
             validator_cycle = cycle length for a validator
             max_quorum_nodes = max nodes to include in autogenerated quorum set
             quorum_threshold = the threshold to use for the generated quorums
-            genesis_start_time = unix timestamp of the Genesis block
+            block_timestamp_tolerance_secs = new block timestamp has to be greater
+                than previous block timestamp, but less than
```
AndrejMitrovic: Well we can't really have a limit to the upper bound. If the rule is it has to be less than `current time + block_timestamp_tolerance_secs` then that would mean if the nodes are late to nominate they would be stuck forever?

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: Btw, in the next commit this change seems to be reverted.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -324,17 +323,22 @@ extern(D):
...
+        if (cur_time < genesis_start_time)
         {
             log.fatal("Clock is out of sync. " ~
                 "Current time: {}. Genesis time: {}", cur_time,
-                this.params.GenesisStartTime);
+                genesis_start_time);
             return;
         }
 
         const slot_idx = this.ledger.getBlockHeight() + 1;
```
AndrejMitrovic: Could you move this line above `this.nominate(slot_idx, data);`? Because the `slot_idx` is not used in the next line anymore.

source/agora/consensus/validation/Block.d
```diff
@@ -80,7 +83,8 @@ version (unittest)
 
 public string isInvalidReason (const ref Block block, Height prev_height,
     in Hash prev_hash, scope UTXOFinder findUTXO,
-    size_t active_enrollments, scope PayloadChecker checkPayload) nothrow @safe
+    size_t active_enrollments, scope PayloadChecker checkPayload,
+	ulong prev_timestamp, ulong curr_timestamp, uint blockTimestampToleranceSecs) nothrow @safe
```
AndrejMitrovic: No tabs please.

source/agora/consensus/validation/Block.d
```diff
@@ -136,7 +140,34 @@ public string isInvalidReason (const ref Block block, Height prev_height,
             return fail_reason;
     }
 
-    return null;
+    return validateBlockTimestamp(prev_timestamp, block.header.timestamp, curr_timestamp, blockTimestampToleranceSecs);
+}
+
+static public string validateBlockTimestamp(ulong prev_block_ts, ulong new_block_ts, ulong curr_timestamp,
```
AndrejMitrovic: `static` has no meaning in module scope.

Also this function is public and doesn't have documentation.

source/agora/consensus/validation/Block.d
```diff
@@ -136,7 +140,34 @@ public string isInvalidReason (const ref Block block, Height prev_height,
...
+    try
+    {
+        import std.conv : to;
+        import std.datetime.systime;
+        if (new_block_ts <= prev_block_ts)
+            res = "the timestamp in consensus data is not greater than the timestamp in the last block\n" ~
+                    "new block timestamp: " ~ to!string(SysTime(unixTimeToStdTime(new_block_ts))) ~ "\n" ~
+                    "previous block's timestamp: " ~ to!string(SysTime(unixTimeToStdTime(prev_block_ts)));
+        else if (new_block_ts > curr_timestamp + blockTimestampToleranceSecs)
+            res =  "(the timestamp in consensus data) is greater than (the timestamp in the last block + tolerance)\n" ~
```
AndrejMitrovic, line 268: Again this doesn't really make sense. If the nodes are late to nominate then they will not be able to recover.

source/agora/network/Clock.d
```diff
@@ -156,3 +158,21 @@ public class Clock
             this.net_time_offset = time_offset;
     }
 }
+
+public class MockClock : Clock
```
AndrejMitrovic, line 161: Mocking stuff should be moved to one of the test modules like [Base.d](https://github.com/bpfkorea/agora/blob/v0.x.x/source/agora/test/Base.d).

There is already a `TestClock` there for example: https://github.com/bpfkorea/agora/blob/30956d13635a1e1ec77b7ce7c5f157f8d851c67d/source/agora/test/Base.d#L1441

source/agora/node/Ledger.d
```diff
@@ -211,6 +220,11 @@ public class Ledger
         Utils.getCollectorRegistry().addCollector(&this.collectBlockStats);
     }
 
+    public const(Block) getLastBlock() @safe @nogc nothrow pure const
```
AndrejMitrovic: Documentation please.

source/agora/node/Ledger.d
```diff
@@ -667,15 +689,17 @@ version (unittest)
                 ? params_
                 : (blocks.length > 0
                    // Use the provided Genesis block
-                   ? new immutable(ConsensusParams)(cast(immutable)blocks[0], WK.Keys.CommonsBudget.address)
+                   ? new immutable(ConsensusParams)(cast(immutable)blocks[0], WK.Keys.CommonsBudget.address,
+                            1008, 7, 80, 30, 1, 1024, 200, 600)
```
AndrejMitrovic: Why not keep these in the default parameters list in the ctor? That's why they were made, so they don't have to be copy-pasted to the call site.

source/agora/consensus/data/Params.d
ferencdg: 1. if we don't have an upper bound then nodes can agree on block timestamp that is 100 years into the future
2. what do you mean by 'if the nodes are late to nominate'? If a node clock is too far forward compared to other blocks, then yes it cannot nominate his own candidates, but it will re-nominate other node candidates

source/agora/consensus/validation/Block.d
ferencdg: still a draft :)

source/agora/consensus/data/Params.d
ferencdg: see https://github.com/bpfkorea/agora/pull/1437#discussion_r546591387

if you mean the node's current clock is too back in the past, so it will not accept anything: I think we already have some assumption about how close the clocks of the nodes have to be together. Why was the original clock synchronization algorithm added? Where else do we need the clocks to be synchronized and what happen there if they are not?

source/agora/consensus/validation/Block.d
ferencdg: see https://github.com/bpfkorea/agora/pull/1437#discussion_r546591387

if you mean the node's current clock is too back in the past, so it will not accept anything: I think we already have some assumption about how close the clocks of the nodes have to be together. Why was the original clock synchronization algorithm added? Where else do we need the clocks to be synchronized and what happen there if they are not?

source/agora/node/Ledger.d
ferencdg: the last parameter is NOT the default, that is why it has to be repeated

source/agora/network/Clock.d
ferencdg: I saw TestClock, but I don't feel it is easy enough to use in unit tests: it requires creating delegates and passing task manager... Even if I created a new constructor that only sets the time there are inherited methods like 'synchronize' that would use the delegate. Also I think it would have been better to define the Clock class as an interface.

source/agora/consensus/data/Params.d
AndrejMitrovic: > if we don't have an upper bound then nodes can agree on block timestamp that is 100 years into the future

We still use the median value of the network time. So this would only fail if the majority of nodes were conspiring to do this.

In case of network outages (think an hour of downtime) we'll have to create a block once the network issues are resolved. But this value here might prevent us from doing that unless `block_timestamp_tolerance_secs` is set to a really high value.

source/agora/consensus/validation/Block.d
AndrejMitrovic: > Why was the original clock synchronization algorithm added?

In this case the node itself may reject voting on a nomination, but if 51% of the network agrees on the vote then it will go through anyway.

In your case you're adding this check to the block validation itself. That's different. Because it means no block will ever be accepted if it exceeds the tolerance. But what do we set the tolerance to? 1 hour? 1 day?

What do other blockchains do? Is there a protocol-level rule like this?

source/agora/consensus/data/Params.d
ferencdg: > > if we don't have an upper bound then nodes can agree on block timestamp that is 100 years into the future
> 
> We still use the median value of the network time. So this would only fail if the majority of nodes were conspiring to do this.
> 
> In case of network outages (think an hour of downtime) we'll have to create a block once the network issues are resolved. But this value here might prevent us from doing that unless `block_timestamp_tolerance_secs` is set to a really high value.

1. if an attacker modifies the agora code and nominates block with 100 years in the future AND there is no upper limit, then his nomination might win and we will have blocks with incorrect timestamp

2. I don't see any problems with network outages, lets say the last block timestamp was: 
1111 and that was the correct current time
the network outage last until 2222 and suppose we should have created 2 blocks
any block timestamp will be fine between 1111 and 2222 + tolerance for those 2 blocks

ferencdg: the criterea for good block timestamp is:
greater then previous block timestamp AND 
less then the current network time + tolerance

source/agora/consensus/validation/Block.d
ferencdg: 1. let's say the node X clock's is really behind and 51% of the nodes accepted a nomination and the block got externalized
2. when the block reaches node X it will find it invalid because of the tolerance and it will exit as an externalized invalid block was received(not ideal, but we can use different tolerance value for voting and accepting blocks)
3. the node restarts next day
4. now the node's current time + tolerance will accept the block that was previously rejected

so the criteria of accepting the block is less stringent as the physical time goes on

AndrejMitrovic: Alright. Let's just get some test-cases written for this. It's better to confirm our expectations.

source/agora/network/Clock.d
AndrejMitrovic: Sure! I meant it would be good to place it either there or in the module where the `MockClock` is used.

source/agora/node/Ledger.d
AndrejMitrovic: > the last parameter is NOT the default, that is why it has to be repeated

Are you sure? That's not what I see. Unless you pushed an update that is.

ferencdg: I am 99% sure :) the test uses 600 secs tolerance while the real one uses 60 secs. I did one push, but that was to remove an unnecessary commit, not related to this.

source/agora/network/Clock.d
ferencdg: unfortunately putting MockClock into Base cases:
object.Error@rt/minfo.d(371): Cyclic dependency between module constructors/destructors of agora.node.Ledger and agora.node.Validator
agora.node.Ledger* ->
agora.test.Base ->
agora.node.Validator* ->
agora.node.Ledger*

when I am trying to unit test agora.node.Ledger

AndrejMitrovic: Damn. D strikes again. Just leave it there then.

AndrejMitrovic:
@ferencdg it looks good so far. However adding the timestamp to the block will have to be merged after the schnorr/slashing work is merged in, as it takes priority and has a hard deadline of needing to be delivered before the end of month. So this PR might get stuck in limbo for a while longer, I'm sorry.

In the meantime, it would be good to split up the bugfixes into a separate PR so we can focus on getting that in.

ferencdg:
> @ferencdg it looks good so far. However adding the timestamp to the block will have to be merged after the schnorr/slashing work is merged in, as it takes priority and has a hard deadline of needing to be delivered before the end of month. So this PR might get stuck in limbo for a while longer, I'm sorry.
> 
> In the meantime, it would be good to split up the bugfixes into a separate PR so we can focus on getting that in.

Sure, I just rebased to the latest master branch. I think the changes you talked about are already in, right? I hope my change can go in next Monday before other PRs. I had to rebase this one 4 times during this sprint and each time tons of conflict. I might have spent around 2 days just with rebasing and fixing failing unit tests.

source/agora/consensus/data/Block.d
```diff
@@ -484,11 +463,12 @@ unittest
         random_seed = Hash of random seed of the preimages
         missing_validators = list of indices to the validator UTXO set
             which have not revealed the preimage
+        timestamp = the block timestamp
```
AndrejMitrovic: This is in the wrong position.

source/agora/consensus/data/Block.d
```diff
@@ -538,10 +519,12 @@ version (unittest)
...
         Hash random_seed = Hash.init, uint[] missing_validators = null,
         KeyPair[] keys = genesis_validator_keys,
-        ulong delegate (PublicKey) cycleForValidator = (PublicKey k) => defaultCycleZero(k)) @safe nothrow
+        ulong delegate (PublicKey) cycleForValidator = (PublicKey k) => defaultCycleZero(k),
+        ulong timestamp = ulong.max) @safe nothrow
     {
-        auto block = makeNewBlock(prev_block, txs, enrollments, random_seed,
-            missing_validators);
+        auto block = makeNewBlock(prev_block, txs,
+                (timestamp == ulong.max) ? prev_block.header.timestamp + 1 : timestamp,
```
AndrejMitrovic, line 530: Shouldn't this be `+ ConsensusParams.BlockInterval`?

source/agora/consensus/data/Block.d
```diff
@@ -65,6 +65,9 @@ public struct BlockHeader
     /// List of indices to the validator UTXO set which have not revealed the preimage
     public uint[] missing_validators;
 
+    /// Block timestamp
```
AndrejMitrovic:
```suggestion
    /// Block unix timestamp
```

source/agora/network/Clock.d
```diff
@@ -62,9 +62,42 @@ public alias GetNetTimeOffset = bool delegate (out long) @safe nothrow;
...
+
+    /// do nothing
+    public override void synchronize () @safe nothrow {}
+}
+
+
 /// Ditto
 public class Clock
 {
+    protected this(){}
```
AndrejMitrovic: If it's only used by `MockClock` I'd prefer that you just pass the null arguments directly to the super constructor.

source/agora/node/Ledger.d
AndrejMitrovic: > the test uses 600 secs tolerance while the real one uses 60 secs

I mean, why not set it to 600 here then https://github.com/bpfkorea/agora/blob/da2c738e94c8918ec2356154f11737ae22b9fe88/source/agora/consensus/data/Params.d#L108 ?

AndrejMitrovic (commented):
Just a few comments, otherwise LGTM.

AndrejMitrovic:
> Sure, I just rebased to the latest master branch. I think the changes you talked about are already in, right? I hope my change can go in next Monday before other PRs. I had to rebase this one 4 times during this sprint and each time tons of conflict. I might have spent around 2 days just with rebasing and fixing failing unit tests.

Yeah all the PRs should be merged already, so the conflict days are over for now.

source/agora/node/Ledger.d
ferencdg: there are 2 versions of the constructor for ConsensusParameter
1. accept a block as the genesis block in the first parameter (this is the version that is used by prod code and by the test you referenced)
2. doesn't accept any blocks(used by some tests)

I don't think setting the consensus block_timestamp_tolerance_secs for the 1st type of constructor to 600 is a good idea, as that doesn't seem to be a reasonable tolerance for the prod code.

source/agora/consensus/data/Block.d
ferencdg: 1. a lot of unit tests will not adjust the clock, and lot of them are not related to 'time consensus', by adding 1 here will ensure that those unrelated unit tests will pass

2. the reason why I didn't add  ConsensusParams.BlockInterval is because at this level most unit tests have no idea about what the ConsensusParams are, and they are just not using it.

3. if some unit tests would like to change the clock or the block creation time, it can pass any timestamp it wants

ferencdg:
Finally the tests went to green, please let me know if anything else is needed.

source/agora/common/Config.d
```diff
@@ -193,6 +188,9 @@ public struct ValidatorConfig
 
     // If the enrollments will be renewed or not at the end of the cycle
     public bool recurring_enrollment = true;
+
+    /// Used only by the testing framework to override the timestamp for genesis block
```
ferencdg: it is very far from ideal to put this data here: what I would need is a way to be able to get the start time of the test in the nominator

I can fix this in a follow up PR where I change getExpectedBlockTime from private to protected and override that method in the TestNominator

source/agora/node/Ledger.d
```diff
@@ -835,15 +875,18 @@ version (unittest)
...
                    // Use the unittest genesis block
-                   : new immutable(ConsensusParams)());
+                   : new immutable(ConsensusParams)(1008, 7, 80, 30, 1, 1024, 200, 600));
```
ferencdg: I could most probably got away with specifying 600 for this second constructor here, but I think it is debatable whether the 2 constructors of the ConsensusParams should have different defaults(even the second constructor which is exclusively used by tests might make more sense with 60 default)

source/agora/common/Config.d
ferencdg: I made the improvement as part of this PR instead of raising a separate one, hopefully this will be enough to merge this PR

AndrejMitrovic:
LGTM.

It would be good to test this with Faucet and to also get https://github.com/bpfkorea/stoa/issues/241 ready before merging.

source/agora/consensus/data/Block.d
```diff
@@ -87,31 +90,7 @@ public struct BlockHeader
...
-        dg(this.merkle_root[]);
-        serializePart(this.validators, dg);
-        serializePart(this.signature, dg);
-        serializePart(this.enrollments.length, dg);
-        foreach (enrollment; this.enrollments)
-            serializePart(enrollment, dg);
-        dg(this.random_seed[]);
-        serializePart(this.missing_validators.length, dg);
-        foreach (validator; this.missing_validators)
-            serializePart(validator, dg);
```
Geod24, line 114: This could've been its own commit

source/agora/consensus/data/Block.d
```diff
@@ -123,17 +102,16 @@ unittest
     PublicKey pubkey = PublicKey.fromString(address);
 
     // above parts not @safe/@nogc yet
-    () @safe @nogc nothrow
+    () @safe
```
Geod24: The test was specifically designed to check for `@nogc/nothrow`-ness, so please revert this and the change to `assert`. Hopefully we'll get `-checkaction=context` soon-ish and will no more needs hacks like this.

source/agora/consensus/data/Block.d
Geod24: Could you add a comment summarizing this to the code then ?
E.g. "// This should really be `+ ConsensusParams.BlockInterval` instead of `+ 1`, but ..."

source/agora/consensus/data/Params.d
Geod24: @ferencdg : There is a difference between "consensus parameters" and "node parameters".
A consensus parameter is something that the node cannot break without having a visible effect on the network, while a node parameter is something that can be broken unseen. Think about the way nodes deal with transactions they receive: A node could choose to ignore every transaction which has an odd hash, and it would behave according to the consensus protocol for the rest of the world.
On the other hand, if we have a "hash filter" on the block, and node can only elect odd hash for odd block height and even hash for even height, then it becomes a consensus parameter.

What you are describing here is a node parameter. A node should absolutely have a way to configure how far into a future it accepts to nominate a block. But it cannot be a hard, consensus limit, as @AndrejMitrovic noted.

source/agora/consensus/data/genesis/Test.d
```diff
@@ -51,6 +51,7 @@ public immutable Block GenesisBlock = {
         prev_block:  Hash.init,
         height:      Height(0),
         merkle_root: GenesisMerkleTree[$ - 1],
+        timestamp:   1596153600,
```
Geod24: Won't that cause problems in tests, since it's set to 2020-07-31 (also please add a comment about it) ?

source/agora/consensus/validation/Block.d
```diff
@@ -219,7 +224,59 @@ public string isInvalidReason (const ref Block block, Height prev_height,
...
+        block_timestamp_tolerance_secs = the proposed block timestamp should be less
+                than curr_timestamp + block_timestamp_tolerance_secs
+
+    Returns:
+        `null` if the new_block_ts is valid, otherwise a string explaining
+        the reason it is invalid.
+
+*******************************************************************************/
+
+public string validateBlockTimestamp(ulong prev_block_ts, ulong new_block_ts, ulong curr_timestamp,
```
Geod24: Space before parenthesis

source/agora/consensus/validation/Block.d
```diff
@@ -96,6 +100,7 @@ public string isInvalidReason (const ref Block block, Height prev_height,
     scope EnrollmentFinder findEnrollment, size_t active_enrollments, size_t enrolled_validators,
     Point delegate (Height, ulong) nothrow @safe getValidatorAtIndex,
     Point delegate (const ref Point) nothrow @safe getCommitmentNonce,
+    ulong prev_timestamp, ulong curr_timestamp, uint block_timestamp_tolerance_secs,
```
Geod24: Instead of using a `uint` for `block_timestamp_tolerance_secs `, could we use a `core.time : Duration` ?

source/agora/consensus/validation/Block.d
```diff
@@ -219,7 +224,59 @@ public string isInvalidReason (const ref Block block, Height prev_height,
...
+{
+    string res;
+    try
+    {
+        import std.conv : to;
+        import std.datetime.systime : SysTime, unixTimeToStdTime;
+        log.trace("validateBlockTimestamp():\nnew block timestamp: {}\ncurrent timestamp: {}\ntolerance in seconds: {}",
+                    to!string(SysTime(unixTimeToStdTime(new_block_ts))),
+                    to!string(SysTime(unixTimeToStdTime(curr_timestamp))),
+                    block_timestamp_tolerance_secs);
```
Geod24: I would remove this `log.trace`, as the validation routines are made to be lightweight and  this is going to cause some noise

source/agora/consensus/validation/Block.d
```diff
@@ -219,7 +224,59 @@ public string isInvalidReason (const ref Block block, Height prev_height,
...
+
+        if (new_block_ts <= prev_block_ts)
+            res = "the timestamp in consensus data is not greater than the timestamp in the last block\n" ~
+                    "new block timestamp: " ~ to!string(SysTime(unixTimeToStdTime(new_block_ts))) ~ "\n" ~
+                    "previous block's timestamp: " ~ to!string(SysTime(unixTimeToStdTime(prev_block_ts)));
```
Geod24, line 266: We usually try to not allocate in the validation routines, that's why we're returning literals everywhere else.
We can give this a try, but it's at odds with the rest of the code.

source/agora/network/Clock.d
```diff
@@ -62,6 +62,38 @@ public alias GetNetTimeOffset = bool delegate (out long) @safe nothrow;
 public alias SetPeriodicTimer = void delegate (Duration, void delegate())
         @trusted nothrow;
 
+///
+public class MockClock : Clock
```
Geod24: Clock related changes could have gone in their own commit

source/agora/node/Ledger.d
```diff
@@ -827,6 +865,8 @@ version (unittest)
     /// A `Ledger` with sensible defaults for `unittest` blocks
     private final class TestLedger : Ledger
     {
+        public MockClock mock_clock;
+
```
Geod24: Why redefine it here ? Just use `Ledger.clock` ?

source/agora/node/Ledger.d
```diff
@@ -1137,6 +1180,62 @@ private KeyPair[] getRandomKeyPairs ()
...
+version (unittest)
+auto get_ledger()
+{
+    auto ledger = new TestLedger(WK.Keys.NODE2);
+    auto txs = genesisSpendable().enumerate()
+        .map!(en => en.value.refund(WK.Keys[en.index].address).sign())
+        .array();
+    txs.each!(tx => assert(ledger.acceptTransaction(tx)));
+    return ledger;
+}
```
Geod24: Nest this inside the unittest instead of making it global. Also, `getLedger ()`

source/agora/test/Base.d
```diff
@@ -447,15 +447,25 @@ extern(D):
     /// number of txs required for nomination
     protected ulong txs_to_nominate;
 
+    /// test start time
+    protected ulong test_start_time;
```
Geod24, line 451: Nitpick, but why not just `start_time` ? We know this refers to the tests for sure.

Geod24 (commented):
Broadly LGTM. Quite a few things, and this could have used a better separation.

source/agora/node/Ledger.d
ferencdg: Ledger.clock is private unfortunately and I need to access it from outside Ledger.d for testing. 

I was thinking to make it public only in debug modes. Should I do that instead of keeping a reference to the clock in the TestLedger?

source/agora/consensus/validation/Block.d
ferencdg: I will remove it, but it seems isInvalidReason is all over littered with trace logs right now.

source/agora/node/Ledger.d
ferencdg: ok, I will just make it a constructor parameter instead, it is more clear

source/agora/test/Base.d
ferencdg: we also have 'initial_time' which is alone conusing enough I think:

```D
/// The initial clock time of every spawned node. Note that if there were
/// any extra blocks loaded (`blocks` in the ctor) then the initial time
/// will be test_start_time + (last_height * block_interval)
protected time_t initial_time;
```

Geod24: Fair point, ignore then.

source/agora/consensus/data/genesis/Test.d
ferencdg: that is a good point and it took me days to find out why tests fails randomly when running in multi-threaded mode(nomination was kicking in too early)

last week I solved that by adding an overriden version of the 
```D
protected override ulong getExpectedBlockTime () @safe @nogc nothrow pure
```
method to the TestNominator

ferencdg:
thank you @Geod24 @AndrejMitrovic @omerfirmak for the comments they should all be addressed now

source/agora/common/Config.d
```diff
@@ -174,6 +169,10 @@ public struct NodeConfig
 
     /// The factor to calculate for the fee of data payload
     public uint tx_payload_fee_factor = 200;
+
+    /// The new block timestamp has to be greater than the previous block timestamp,
+    /// but less than current time + block_timestamp_tolerance_dur
+    public Duration block_timestamp_tolerance_dur = 60.seconds;
```
Geod24: No need for the `_dur` suffix

source/agora/common/Config.d
```diff
@@ -419,7 +416,8 @@ private NodeConfig parseNodeConfig (Node* node, in CommandLine cmdln)
             preimage_reveal_interval : preimage_reveal_interval,
             stats_listening_port : stats_listening_port,
             tx_payload_max_size : tx_payload_max_size,
-            tx_payload_fee_factor : tx_payload_fee_factor
+            tx_payload_fee_factor : tx_payload_fee_factor,
+            block_timestamp_tolerance_dur : block_timestamp_tolerance_secs.seconds
```
Geod24: Add a trailing comma for future additions

source/agora/network/Clock.d
```diff
@@ -62,6 +62,38 @@ public alias GetNetTimeOffset = bool delegate (out long) @safe nothrow;
...
+{
+    private ulong time;
+
+    ///
+    public this(ulong time)
```
Geod24: Space

source/agora/network/Clock.d
```diff
@@ -62,6 +62,38 @@ public alias GetNetTimeOffset = bool delegate (out long) @safe nothrow;
...
+    {
+        super(null, null);
+        this.time = time;
+    }
+
+    ///
+    public void setTime(ulong time)
```
Geod24: Space

source/agora/network/Clock.d
Geod24: Also move this *after* the `Clock` type please

source/agora/node/Ledger.d
```diff
@@ -100,6 +107,10 @@ public class Ledger
     /// The checker of transaction data payload
     private DataPayloadChecker payload_checker;
 
+    /// The new block timestamp has to be greater than the previous block timestamp,
+    /// but less than current time + block_timestamp_tolerance_dur
+    public Duration block_timestamp_tolerance_dur;
```
Geod24: Likewise, no need for a sufix

source/agora/node/Ledger.d
```diff
@@ -224,6 +244,20 @@ public class Ledger
...
+    /***************************************************************************
+
+        Returns the last block in the `Ledger`
+
+        Returns:
+            last block in the `Ledger`
+
+    ***************************************************************************/
+
+    public const(Block) getLastBlock() @safe @nogc nothrow pure const
```
Geod24: Space before parenthesis please

source/agora/node/Ledger.d
```diff
@@ -1137,6 +1189,61 @@ private KeyPair[] getRandomKeyPairs ()
...
 }
 
+unittest
+{
+    import agora.consensus.data.genesis.Test;
+    ConsensusData data;
+    auto genesis_ts = GenesisBlock.header.timestamp;
+    MockClock mock_clock = new MockClock(time(null));
+
+    auto getLedger(Clock clock)
```
Geod24: Space before parenthesis

Geod24 (approved):
LGTM, just a couple of style fixes, feel free to merge once addressed.

Geod24:
Actually, since the changes are so minor, I did them directly and will merge once green.

PR #1438, opened 2020-12-21
Implement slashing protocol

linked0:
With this PR, we can add information about validators being slashed into the block. After the validators being slashed, the UTXOs of the validators are split into the validators' address and the `CommonsBudget` address.

This does *NOT* contain the validation check for the information yet. But it will be implemented soon. The feature of PR #1421 is the need for the next step, which is to maintain un-enrolled validators in the `ValidatorSet`.

Part of #1076

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -702,12 +702,13 @@ public class EnrollmentManager
         foreach (const ref key; keys)
         {
             const preimage = this.validator_set.getPreimageAt(key, height);
-            // this should not happen. validators which didn't reveal the
-            // preimage should not be in the active validator set 'keys'.
+            // We could have a misbehaving validator that does not reveal
+            // its preimages in real world. In order to deal with the validator,
+            // We implement the slashing protocol.
             if (preimage == PreImageInfo.init)
             {
-                log.fatal("No preimage at height {} for validator key {}", height.value, key);
-                assert(0, "No preimage at expected height");
+                log.info("No preimage at height {} for validator key {}", height.value, key);
+                continue;
             }
             rand_seed = hashMulti(rand_seed, preimage);
         }
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -1095,6 +1095,20 @@ public class EnrollmentManager
     {
         return &this.validator_set.findRecentEnrollment;
     }
+
+    /***************************************************************************
+
+        Unenroll the validator from the validator set
+
+        Params:
+            enroll_hash = key for the validator to unenroll
+
+    ***************************************************************************/
+
+    public void unenrollValidator (in Hash enroll_hash) @trusted
+    {
+        this.validator_set.unenroll(enroll_hash);
+    }
 }
 
 /// tests for member functions of EnrollmentManager
@@ -1708,4 +1722,41 @@ unittest
             Height(params.ValidatorCycle), &utxo_set.peekUTXO,
                                                 utxo_set.storage) is null);
     assert(man.getValidatorCount(Height(params.ValidatorCycle)) == 1);
+}
+
+// Test for adding and removing validators
+unittest
+{
+    import agora.consensus.data.Transaction;
+    import std.array;
+    import std.range;
+
+    Hash[] utxos;
+    auto utxo_set = new TestUTXOSet;
+    genesisSpendable()
+        .enumerate.map!(tup => tup.value
+            .refund(WK.Keys[tup.index].address).sign(TxType.Freeze))
+        .each!((tx) {
+            utxo_set.put(tx);
+            utxos ~= UTXO.getHash(tx.hashFull(), 0);
+        });
+
+    auto man = new EnrollmentManager(":memory:", WK.Keys.A,
+        new immutable(ConsensusParams));
+    auto e1 = EnrollmentManager.makeEnrollment(
+        WK.Keys.A, utxos[0], 10, 0);
+    auto e2 = EnrollmentManager.makeEnrollment(
+        WK.Keys.B, utxos[1], 10, 0);
+
+    assert(man.addEnrollment(e1, WK.Keys.A.address, Height(1), &utxo_set.peekUTXO));
+    assert(man.addEnrollment(e2, WK.Keys.B.address, Height(1), &utxo_set.peekUTXO));
+
+    assert(man.addValidator(e1, WK.Keys.A.address, Height(2), &utxo_set.peekUTXO,
+        utxo_set.storage) is null);
+    assert(man.addValidator(e2, WK.Keys.B.address, Height(2), &utxo_set.peekUTXO,
+        utxo_set.storage) is null);
+
+    assert(man.getValidatorCount(Height(2)) == 2);
+    man.unenrollValidator(utxos[0]);
+    assert(man.getValidatorCount(Height(2)) == 1);
 }
--- a/source/agora/consensus/SlashPolicy.d
+++ b/source/agora/consensus/SlashPolicy.d
@@ -115,6 +115,36 @@ public class SlashPolicy
 
     /***************************************************************************
 
+        Get the UTXOs of the validators that do not reveal their pre-images
+        by indices
+
+        Params:
+            validators_utxos = will contain the UTXOs ot the validators
+            missing_validators = indices of validators being slashed
+
+    ***************************************************************************/
+
+    public void getMissingValidatorsUTXOs (ref Hash[] validators_utxos,
+        const uint[] missing_validators) @safe nothrow
+    {
+        validators_utxos.length = 0;
+        () @trusted { assumeSafeAppend(validators_utxos); }();
+
+        Hash[] keys;
+        if (!this.enroll_man.getEnrolledUTXOs(keys))
+        {
+            log.fatal("Could not retrieve enrollments");
+            assert(0);
+        }
+
+        foreach (idx; missing_validators)
+        {
+            validators_utxos ~= keys[idx];
+        }
+    }
+
+    /***************************************************************************
+
         Get the random seed reduced from the preimages for the provided
         block height.
 
@@ -322,18 +352,22 @@ unittest
     // get all the validators and find index of the first and second validastors
     uint first_validator;
     uint second_validator;
+    Hash first_validator_utxo;
+    Hash second_validator_utxo;
     Hash[] utxos;
     assert(enroll_man.getEnrolledUTXOs(utxos));
     foreach (idx, utxo; utxos)
         if (utxo == enrollments[0].utxo_key)
         {
             first_validator = cast(uint)idx;
+            first_validator_utxo = utxo;
             break;
         }
     foreach (idx, utxo; utxos)
         if (utxo == enrollments[1].utxo_key)
         {
             second_validator = cast(uint)idx;
+            second_validator_utxo = utxo;
             break;
         }
 
@@ -367,6 +401,12 @@ unittest
                 (missing_validators.length));
     assert(missing_validators.find(first_validator).empty());
     assert(missing_validators.find(second_validator).empty());
+
+    // get the UTXOs of the validators that do not reveals preimages
+    Hash[] validators_utxos;
+    slash_man.getMissingValidatorsUTXOs(validators_utxos, missing_validators);
+    assert(validators_utxos.find(first_validator_utxo).empty());
+    assert(validators_utxos.find(second_validator_utxo).empty());
 
     // get and check random seed
     Hash preimage_root = slash_man.getRandomSeed(Height(2));
--- a/source/agora/consensus/SlashPolicy.d
+++ b/source/agora/consensus/SlashPolicy.d
@@ -153,6 +153,47 @@ public class SlashPolicy
 
     /***************************************************************************
 
+        Get the random seed reduced from the preimages of validators
+        except the provided 'missing_validators'.
+
+        Params:
+            height = the desired block height to look up the hash for
+            missing_validators = the validators that did not reveal their
+                preimages for the height
+
+        Returns:
+            the random seed if there are one or more valid preimages,
+            otherwise Hash.init.
+
+    ***************************************************************************/
+
+    public Hash getExternalizedRandomSeed (in Height height,
+        const ref uint[] missing_validators) @safe nothrow
+    {
+        Hash[] keys;
+        if (!this.enroll_man.getEnrolledUTXOs(keys) || keys.length == 0)
+        {
+            log.fatal("Could not retrieve enrollments / no enrollments found");
+            assert(0);
+        }
+
+        Hash[] valid_keys;
+        foreach (idx, key; keys)
+        {
+            if (missing_validators.find(idx).empty())
+                valid_keys ~= key;
+        }
+
+        // TODO: This should not happen in fact, which situation will be handled
+        // in issue #1444. (https://github.com/bpfkorea/agora/issues/1444)
+        if (valid_keys.length == 0)
+            return Hash.init;
+
+        return this.enroll_man.getRandomSeed(valid_keys, height);
+    }
+
+    /***************************************************************************
+
         Check if a validator has a pre-image for the height
 
         Params:
@@ -330,5 +371,11 @@ unittest
     // get and check random seed
     Hash preimage_root = slash_man.getRandomSeed(Height(2));
     assert(preimage_root ==
+        hashMulti(hashMulti(Hash.init, preimage_1), preimage_2));
+
+    // get and check random seed when a block being externalized
+    Hash externalized_seed = slash_man.getExternalizedRandomSeed(Height(2),
+        missing_validators);
+    assert(externalized_seed ==
         hashMulti(hashMulti(Hash.init, preimage_1), preimage_2));
 }
--- a/source/agora/consensus/SlashPolicy.d
+++ b/source/agora/consensus/SlashPolicy.d
@@ -143,7 +143,7 @@ public class SlashPolicy
                 valid_keys ~= key;
         }
 
-        // NOTE: The preimage root of `Hash.init` value is currently
+        // NOTE: The random seed of `Hash.init` value is currently
         // checked in the `validateSlashingData` function of `Ledger`.
         if (valid_keys.length == 0)
             return Hash.init;
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -184,7 +184,9 @@ public struct Block
                 this.header.merkle_root,
                 validators,
                 signature,
-                this.header.enrollments.dup),
+                this.header.enrollments.dup,
+                this.header.random_seed,
+                this.header.missing_validators.dup),
             // TODO: Optimize this by using dup for txs also
             this.txs.map!(tx =>
                 tx.serializeFull.deserializeFull!Transaction).array,
@@ -479,11 +481,16 @@ unittest
         prev_block = the previous block
         txs = the transactions that will be contained in the new block
         enrollments = the enrollments that will be contained in the new block
+        random_seed = Hash of random seed of the preimages
+        missing_validators = list of indices to the validator UTXO set
+            which have not revealed the preimage
 
 *******************************************************************************/
 
 public Block makeNewBlock (Transactions)(const ref Block prev_block,
-    Transactions txs, Enrollment[] enrollments = null) @safe nothrow
+    Transactions txs, Enrollment[] enrollments = null,
+    Hash random_seed = Hash.init, uint[] missing_validators = null)
+    @safe nothrow
 {
     static assert (isInputRange!Transactions);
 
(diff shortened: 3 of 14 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -565,12 +565,13 @@ public class EnrollmentManager
         foreach (const ref key; keys)
         {
             const preimage = this.validator_set.getPreimageAt(key, height);
-            // this should not happen. validators which didn't reveal the
-            // preimage should not be in the active validator set 'keys'.
+            // We could have a misbehaving validator that does not reveal
```
AndrejMitrovic, line 705: The way I envisioned it is that we would simply return the preimage root that's stored in the block header.

This function is always called with a block height in the past.

AndrejMitrovic: But I guess the next commits do that. I'm still reviewing.

source/agora/consensus/data/Block.d
```diff
@@ -59,6 +59,12 @@ public struct BlockHeader
     /// Enrolled validators
     public Enrollment[] enrollments;
 
+    /// Hash of random seed of the preimages for this this height
```
AndrejMitrovic, line 62: in the commit message: whil => while

source/agora/node/Ledger.d
```diff
@@ -518,8 +522,8 @@ public class Ledger
 
     public string validateSlashingData (in ConsensusData data) @safe
     {
-        return this.slash_man.isInvalidPreimageRootReason(this.getBlockHeight(),
-                data.missing_validators);
+        return this.slash_man.isInvalidPreimageRootReason(
+                this.last_block.header.height, data.missing_validators);
```
AndrejMitrovic: Is this change just an optimization? Or is the height passed here different now?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -924,6 +924,20 @@ public class EnrollmentManager
...
+
+        Params:
+            enroll_hash = key for the validator to remove
+
+    ***************************************************************************/
+
+    public void removeValidator (in Hash enroll_hash) @trusted
+    {
+        this.validator_set.remove(enroll_hash);
+    }
```
AndrejMitrovic, line 1111: I think we're going with the other approach which is just tagging the disabled validators, and not removing them from the set. This is to support what was implemented in https://github.com/bpfkorea/agora/pull/1421/commits

source/agora/consensus/SlashPolicy.d
```diff
@@ -113,6 +113,36 @@ public class SlashPolicy
...
+
+        Hash[] keys;
+        if (!this.enroll_man.getEnrolledUTXOs(keys))
+        {
+            log.fatal("Could not retrieve enrollments");
+            assert(0);
+        }
+
+        foreach (idx; missing_validators)
+        {
```
AndrejMitrovic, line 141: This isn't actually doing anything? The array is empty.

source/agora/node/Ledger.d
linked0: I changed it unintentionally. There's no difference between before and after. I'll restore it.

source/agora/consensus/state/UTXOSet.d
```diff
@@ -49,18 +49,20 @@ abstract class UTXOCache
         Params:
             tx = the transaction
             height = local height of the block
+            slashed = whether the transaction is for slashing
```
AndrejMitrovic, line 52: typo in commit message: chian

source/agora/consensus/EnrollmentManager.d
linked0: > I think we're going with the other approach which is just tagging the disabled validators, and not removing them from the set. This is to support what was implemented in https://github.com/bpfkorea/agora/pull/1421/commits

Yes, right. The `remove` function will just change the flag of 'active' after the commit being merged.

source/agora/consensus/data/Block.d
linked0: > in the commit message: whil => while

corrected.

source/agora/consensus/state/UTXOSet.d
linked0: > typo in commit message: chian

corrected.

source/agora/consensus/EnrollmentManager.d
linked0: Added `unenrollValidator` function instead of `removeValidator` and new function called `unenroll` into the `ValidatorSet`.

source/agora/consensus/SlashPolicy.d
linked0: > This isn't actually doing anything? The array is empty.

The `missing_validators` which is in fact `missing_validators` in `BlockHeader` is passed from the Ledger when creating a new Block. Can you explain why do you think it's empty?

source/agora/consensus/EnrollmentManager.d
linked0: > The way I envisioned it is that we would simply return the preimage root that's stored in the block header.
> 
> This function is always called with a block height in the past.

I will think more about that.

source/agora/consensus/SlashPolicy.d
AndrejMitrovic: ahh I misread the code. I saw the other array set to zero.

source/agora/consensus/SlashPolicy.d
```diff
@@ -113,6 +113,36 @@ public class SlashPolicy
...
+        by indices
+
+        Params:
+            validators_utxos = will contain the UTXOs ot the validators
+            missing_validators = indices of validators being slashed
+
+    ***************************************************************************/
+
+    public void getMissingValidatorsUTXOs (ref Hash[] validators_utxos,
+        const ref uint[] missing_validators) @safe nothrow
```
AndrejMitrovic: This doesn't need to be `ref`.

source/agora/consensus/EnrollmentManager.d
linked0: > The way I envisioned it is that we would simply return the preimage root that's stored in the block header.
> 
> This function is always called with a block height in the past.

I thought about it. 
The `random_seed` of a block header is for the previous height and the `rand_seed` from this `getRandomSeed` function which is used for the next block is for the current height. So we should generate a random seed again in this function, I think.

AndrejMitrovic: > which is used for the next block 

I thought this function was used for the quorum generator. When it generates quorums for height N it uses preimages of N-1. So it could just read the root from the N-1 block header.

AndrejMitrovic: I think the problem I am seeing here is that `getRandomSeed` is used by both the quorum generator and the slasher.

The quorum generator needs a fixed (externalized) random seed, which we are now storing in the block header.

So maybe there should be two functions? One is the `getCandidateRandomSeed` and the other `getExternalizedRandomSeed`. Do you see what I mean?

linked0: > > which is used for the next block
> 
> I thought this function was used for the quorum generator. When it generates quorums for height N it uses preimages of N-1. So it could just read the root from the N-1 block header.

I thought it the same way, but I realized that the random seed from the N-1 block header is generated with preimages of N-2 in rethinking. Do you mean that the random seed is generated with pre-images of current validators and newly enrolled validators?

linked0: > So maybe there should be two functions? One is the `getCandidateRandomSeed` and the other `getExternalizedRandomSeed`. Do you see what I mean?

I got it. I thought it right before. We can generate the random seed for the `getExternalizedRandomSeed` with pre-images of current validators and newly enrolled validators when validating a block, which is stored in a block header after.

AndrejMitrovic: So I took another look now.

When a block is externalized, we call `rebuildQuorumConfig(.., block.height)`. That function calls `getRandomSeed()`.

What I am worried about is the case where a validator reveals their preimage in the small time window just before a block is externalized. Their preimage might not be included in the block's header (because it's too late), but it might get included in some node's local validator set because a new block was not externalized yet. Then, when `rebuildQuorumConfig()` is called it's possible some nodes will use a different random seed based on their current internal preimage state.

That's why I suggest to use a separate function for the quorum generator. It should take the random seed in the block header (always). I was wrong about it taking the `N-1`'th preimages, it's the `N`'th preimages.

Does that make sense?

linked0: > That's why I suggest to use a separate function for the quorum generator. It should take the random seed in the block header (always). I was wrong about it taking the N-1'th preimages, it's the N'th preimages.
>
> Does that make sense?

Yes, that makes sense. And there is one more thing to be considered. Which pre-images are used for the `random seed` for the block header? There is a situation some pre-images are NOT revealed at the current height (I mean, the height of the currently externalized block) So, we need to generate the random seed with N-1'th preimages of the existing validator and N'th preimages of newly enrolled validators. What do you think of it?

AndrejMitrovic: > There is a situation some pre-images are NOT revealed at the current height (I mean, the height of the currently externalized block)

I mean that's fine? After the block is externalized we use that random seed for the quorum generator. We don't care what's in it, just that it was voted and externalized.

linked0: > I mean that's fine? After the block is externalized we use that random seed for the quorum generator. We don't care what's in it, just that it was voted and externalized.

The `random_seed` is not part of `ConsensusData`. So we should determine what pre-images should used for the `random_seed`, I think. (EDITED)

AndrejMitrovic: I am just talking about the quorum generator though. Not the nomination process.

AndrejMitrovic: If this isn't making much sense to you, it's fine to keep it the way you wrote it.

I'll submit a separate issue for this later.

linked0: I will think about what you mean again if you submit a new issue about that. Anyway, I just wanted to know how to generate the `random_seed` for a block header.
Thank you.

linked0:
Ready for review.
Thanks.

source/agora/consensus/EnrollmentManager.d
AndrejMitrovic: > Anyway, I just wanted to know how to generate the random_seed for a block header.

Yes, that's fair. I was looking at the other use-case and not specifically your own. So it may have been off-topic for this PR. I apologize.

AndrejMitrovic:
Unfortunately there is a conflict now. I hope it won't be hard to rebase?

linked0:
> Unfortunately there is a conflict now. I hope it won't be hard to rebase?

I'm rebasing it now.

source/agora/consensus/EnrollmentManager.d
linked0: > > Anyway, I just wanted to know how to generate the random_seed for a block header.
> 
> Yes, that's fair. I was looking at the other use-case and not specifically your own. So it may have been off-topic for this PR. I apologize.

It's okay. There are some mistakes in expressing my opinion in English. Thanks.

AndrejMitrovic:
Thank you for rebasing. I'm doing a review now.

source/agora/utils/PrettyPrinter.d
```diff
@@ -355,7 +358,9 @@ private struct BlockHeaderFmt
 { utxo: 0xb20d...08eb, seed: 0xa050...2cb4, cycles: 20, sig: 0x052e...6b31 }
 { utxo: 0xdb39...2d85, seed: 0xdd1b...7bfa, cycles: 20, sig: 0x0e00...4fe2 }]
 Signature: 0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000,
-Validators: [0]`;
+Validators: [0],
+RandomSeed: [0x0000...0000],
+MPV: []`;
```
AndrejMitrovic: In the future I think we should name this just `MV` that stands for "missing validators".

source/agora/consensus/data/Block.d
```diff
@@ -541,9 +552,16 @@ version (unittest)
             "7caa4fcffdc5c068a07532637cf5042ae39b7af418847385480e620e1395987")
     };
 
-    auto block = makeNewBlock(GenesisBlock, [Transaction.init], [enr_1, enr_2]);
+    auto primage_root = Hash("0x47c993d409aa7d77651ecaa5a5d29e47a7aee609c7" ~
```
AndrejMitrovic: `primage_root` => `preimage_root`

source/agora/node/Ledger.d
```diff
@@ -229,7 +229,10 @@ public class Ledger
     public bool onExternalized (ConsensusData data)
         @trusted
     {
-        auto block = makeNewBlock(this.last_block, data.tx_set, data.enrolls);
+        Hash random_seed = slash_man.getRandomSeed(this.getBlockHeight());
```
AndrejMitrovic: One change can be made here. Take into account the `missing_validators` in the ConsensusData, and make sure their preimage is not included when calculating the random seed inside of `getRandomSeed`.

The reason why this is needed is because there can be a timing issue. A block could be externalized around the same time as another slow validator sends a preimage. So that validator could be inside of `missing_validators`, but the preimage could have been added just a second before externalization.

This is the issue I talked about before.

source/agora/node/Ledger.d
```diff
@@ -247,7 +250,8 @@ public class Ledger
 
     ***************************************************************************/
 
-    public bool acceptBlock (const ref Block block) @safe
+    public bool acceptBlock (const ref Block block)
+        @safe
```
AndrejMitrovic: Unrelated change.

source/agora/node/Ledger.d
```diff
@@ -535,7 +539,8 @@ public class Ledger
 
     ***************************************************************************/
 
-    public string validateBlock (const ref Block block) nothrow @safe
+    public string validateBlock (const ref Block block)
+        nothrow @safe
```
AndrejMitrovic: Unrelated change.

source/agora/consensus/state/UTXOSet.d
```diff
@@ -49,18 +49,20 @@ abstract class UTXOCache
...
+        bool slashed = false) @safe
     {
         import std.algorithm : any;
 
         // defaults to next block
         Height unlock_height = Height(height + 1);
 
         // for payments of frozen transactions, it will melt after 2016 blocks
-        if ((tx.type == TxType.Payment)
+        if ((tx.type == TxType.Payment && !slashed)
```
AndrejMitrovic, line 68: You need to document these kinds of changes, it's very important.

For example:

```
If this is an unfreeze and slashing transaction, don't melt but instead set the unlock height to the next block as it will be sent to the commons budget.
```

source/agora/node/Ledger.d
```diff
@@ -395,6 +395,45 @@ public class Ledger
...
+            auto remain_amount = Amount(utxo_value.output.value);
+            remain_amount.sub(this.slash_man.penalty_amount);
+            Transaction slashing_tx =
+            {
+                TxType.Payment,
+                inputs: [Input(utxo)],
+                outputs: [
+                    Output(this.slash_man.penalty_amount,
+                        this.slash_man.penalty_address),
+                    Output(remain_amount, utxo_value.output.address),
```
AndrejMitrovic, line 500: This means if the node gets punished then parts of their funds will be sent to the commons budget, but now the leftover part will be immediately unfrozen. I'm not sure this is the best idea. I would have hoped that the leftover part either remains frozen, or becomes a melting transaction.

But I understand it might be complicated to implement.

How about we leave it as is for now and we file a new issue for improving this later?

source/agora/node/Ledger.d
```diff
@@ -427,6 +466,31 @@ public class Ledger
...
+        Update the validators that are to be slashed
+
+        Params:
+            block = the block to update the Validator set with
+
+    ***************************************************************************/
+
+    protected void updateSlashedValidatorSet (const ref Block block) @safe
+    {
+        if(block.header.height == 0)
```
AndrejMitrovic: `if(` => `if (`

AndrejMitrovic (changes requested):
A few comments.

source/agora/node/Ledger.d
linked0: > How about we leave it as is for now and we file a new issue for improving this later?

Yes, it could be implemented with a more reasonable policy about the frozen UTXO.

source/agora/node/Ledger.d
linked0: > One change can be made here. Take into account the missing_validators in the ConsensusData, and make sure their preimage is not included when calculating the random seed inside of getRandomSeed.

Right, we should avoid the situation where the preimage arriving late could be used to generate the random seed. I'll fix it.

source/agora/utils/PrettyPrinter.d
linked0: > In the future I think we should name this just MV that stands for "missing validators".

I agree with it.

source/agora/consensus/state/UTXOSet.d
linked0: > You need to document these kinds of changes, it's very important.
> 
> For example:
> 
> ```
> If this is an unfreeze and slashing transaction, don't melt but instead set the unlock height to the next block as it will be sent to the commons budget.
> ```

Ah, I missed it although it's very important.

source/agora/node/Ledger.d
linked0: Updated.

source/agora/node/Ledger.d
linked0: Updated.

source/agora/node/Ledger.d
linked0: Updated.

source/agora/consensus/data/Block.d
linked0: Updated.

source/agora/utils/PrettyPrinter.d
```diff
@@ -331,12 +331,15 @@ private struct BlockHeaderFmt
     {
         try
         {
-            formattedWrite(sink, "Height: %d, Prev: %s, Root: %s, Enrollments: [%s]\nSignature: %s,\nValidators: %s",
+            formattedWrite(sink, "Height: %d, Prev: %s, Root: %s, Enrollments: [%s]\nSignature: %s,\nValidators: %s,\nRandomSeed: [%s],\nMPV: [%s]",
```
hewison-chris: Are we going to stick with MPV? I think `missing` or `slashed` might be better.

source/agora/test/SlashingMisbehavingValidator.d
```diff
@@ -0,0 +1,148 @@
...
+
+    // block 1
+    txs.each!(tx => nodes[0].putTransaction(tx));
+    network.expectBlock(Height(1));
+
+    // block 2
+    txs = txs.map!(tx => TxBuilder(tx).sign()).array();
+    txs.each!(tx => nodes[0].putTransaction(tx));
+    network.expectBlock(Height(2));
+    auto block2 = nodes[0].getBlocksFrom(2, 2)[0];
```
hewison-chris: Why up to 2 blocks in getBlockFrom?

hewison-chris (approved):
Only minor comments.

source/agora/test/SlashingMisbehavingValidator.d
linked0: There's no special meaning in fact. I'll fix it.

source/agora/utils/PrettyPrinter.d
linked0: @AndrejMitrovic gave the same opinion about that. I will change it to `Missing Validators` or `MV`.

linked0: Do you think `Missing` is better?

linked0:
All green in the end.
Ready for review.

source/agora/utils/PrettyPrinter.d
hewison-chris: In fact `Missing validators` is better.

AndrejMitrovic: Slashed validators is better I think.

"Missing" could sound like they just haven't signed the block yet.

AndrejMitrovic:
@linked0 you also need to add block validation of the random seed in the [validation routine](https://github.com/bpfkorea/agora/blob/b939c3b12798c4b876b621c012aebf9e19a20f02/source/agora/consensus/validation/Block.d#L94).

Do you want to do it in a separate PR?

linked0:
> @linked0 you also need to add block validation of the random seed in the [validation routine](https://github.com/bpfkorea/agora/blob/b939c3b12798c4b876b621c012aebf9e19a20f02/source/agora/consensus/validation/Block.d#L94).
> 
> Do you want to do it in a separate PR?

I'd like to do it in issue #1444 because it needs to modify some routine generating blocks and other related code so it will take some time.

AndrejMitrovic (approved):
LGTM. I think just the logging could be changed as mentioned.

linked0:
> LGTM. I think just the logging could be changed as mentioned.

I will do it now, thanks.

source/agora/utils/PrettyPrinter.d
linked0: Updated to 'Slashed validators`.

source/agora/test/SlashingMisbehavingValidator.d
linked0: Updated.

AndrejMitrovic:
Thank you @linked0!

PR #1450, opened 2020-12-24
Use exceptions instead of assertions

### LINKED ISSUE #1399: `fromString()` routines should not use assertions
This assertion is sometimes triggered due to what appears to be some kind of memory corruption (the `str` is actually empty):

https://github.com/bpfkorea/agora/blob/a4b9eb785f4c83f2146505e32ee7e515c735426a/source/agora/common/crypto/Key.d#L160

Nevertheless, even if the memory bug wasn't there we still shouldn't use assertions for external data. The DoS vector here is simple, pretend to be a node and send garbage data when `getPublicKey` is called and crash the calling node.

Exceptions should be used instead.

AndrejMitrovic:
Fixes #1399

### DIFF
--- a/source/agora/common/crypto/Key.d
+++ b/source/agora/common/crypto/Key.d
@@ -153,13 +153,25 @@ public struct PublicKey
         assert(writer.data() == address);
     }
 
-    /// Create a Public key from Stellar's string representation
+    /***************************************************************************
+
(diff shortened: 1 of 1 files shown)


linked0 (approved):
LGTM.

PR #1452, opened 2020-12-28
Implement transaction fee collection

### LINKED ISSUE #1435: Implement transaction fee collection
In order to support transaction replacement via bigger fees (https://github.com/bpfkorea/agora/issues/245), we also need to implement fee collection first.

Each transaction's fee is defined as the `Sum(Inputs) – Sum(Outputs)`.

In Bitcoin, **collecting** the fees is simple. The single miner which mined the block generates a `Coinbase` transaction which includes all of the transactions' combined fees and outputs them all to the miner's desired public key.

In our case, it's different. We don't have miners, we have consensus nodes. Therefore the fee needs to be distributed among the validator nodes. There are some open questions:

- How are the fees distributed among the nodes? Are they distributed evenly towards all (non-slashed) validators? Or does distribution depend on other factors such as: stake amount, number of total externalized Enrollments for that node..?
- After fee distribution is decided how do we update the UTXO set? In Bitcoin they use the Coinbase transaction for this. Do we need something similar? Or could we make it implicit in the protocol and not add any additional transactions to the block? If it's implicit then after externalization the UTXO set would need to be automatically updated to add new UTXOs to each validator node (the fees). But we need to be careful here because we're creating UTXOs from thin air basically. Maybe a Coinbase transaction would still be a better option.

omerfirmak:
Fixes: #1435


source/agora/consensus/validation/Transaction.d
```diff
@@ -132,13 +132,25 @@ public string isInvalidReason (
         auto error = checkPayload(tx);
         if (error !is null) return error;
     }
+    else if (tx.type == TxType.Coinbase)
+    {
+        if (tx.inputs.length != 1)
+            return "Transaction: Coinbase transactions must" ~
+                "include a single Input";
+
+        if (tx.inputs[0] != Input(Hash.init, height))
```
AndrejMitrovic: I'm not sure this is the best idea. Couldn't this end up in a situation where there are multiple coinbase transactions (in different blocks) with the same hash? That could end up overwriting a previous utxo.

source/agora/consensus/state/ValidatorSet.d
```diff
@@ -736,6 +742,47 @@ public class ValidatorSet
...
+            height = requested height
+
+        Returns:
+            `PublicKey`s and enrollment heights of `Validator`s whose enrollment
+            cycle have just ended
+
+    ***************************************************************************/
+
+    public ExpiringValidator[] getExpiringValidators (Height height,
+        out ExpiringValidator[] ex_validators)
```
AndrejMitrovic: Needs to be `ref`. `out` sets it to `null`, so setting the `length` later doesn't have the intended effect. Should also call `assumeSafeAppend` after the length is changed to zero.

source/agora/consensus/state/ValidatorSet.d
```diff
@@ -736,6 +742,47 @@ public class ValidatorSet
...
+
+        try
+        {
+            auto results = this.db.execute("SELECT enrolled_height, public_key" ~
+                " FROM validator_set WHERE enrolled_height + cycle_length = ?" ~
+                " AND active =  ?", height.value, EnrollmentStatus.Active);
+
+            foreach (row; results)
+            {
+                ex_validators ~= ExpiringValidator(Height(row.peek!(size_t)(0)),
```
AndrejMitrovic: `size_t` => `ulong`

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1129,6 +1129,25 @@ public class EnrollmentManager
     {
```
AndrejMitrovic, line 1133: commit message: `Add get ValidatorStakes` => `Add getValidatorStakes`

source/agora/consensus/state/ValidatorSet.d
```diff
@@ -783,6 +783,41 @@ public class ValidatorSet
...
+        Params:
+            peekUTXO =
+            utxos =
+
+        Returns:
+            
+
+    ***************************************************************************/
+
+    public UTXO[] getValidatorStakes (UTXOFinder peekUTXO, out UTXO[] utxos)
```
AndrejMitrovic: ditto comment about `ref` and `assumeSafeAppend`. And I suggest changing it in the EnrollmentManager too.

source/agora/node/Fee.d
```diff
@@ -229,6 +236,84 @@ public class FeeManager
...
+        Return:
+            `Amount` of fees that should be paid to each Validator
+
+    ***************************************************************************/
+
+    public Amount[] getValidatorFees (Amount tot_fee, Amount tot_data_fee,
+        UTXO[] stakes) nothrow @safe
+    {
+        // no stakes, no fees
+        if (stakes.length == 0)
```
AndrejMitrovic, line 272: Technically I think this should be an assertion. There should always be active validators.

source/agora/consensus/validation/Transaction.d
omerfirmak: I did this to avoid exactly that actually :) Bitcoin had that problem in early days apparently, see BIP34.

> there are multiple coinbase transactions (in different blocks) with the same hash

Wouldnt that be a hash collision anyways? Their inputs will have different heights, shouldnt they have different hashes? As long as we create single coinbase tx per address in a block, we should be fine IMO.

AndrejMitrovic: I don't think the height is part of the tx hash.

AndrejMitrovic: You could do a simple test. Create two exactly same coinbase txs at different heights and see if they produce unique utxos.

omerfirmak: Height actually goes into UTXO hash of the input.

I have tried before, but I'll add an explicit test for it anyways.

AndrejMitrovic: > Height actually goes into UTXO hash of the input.

That doesn't sound right to me. Maybe we're talking about different topics?

AndrejMitrovic: > I did this to avoid exactly that actually :) Bitcoin had that problem in early days apparently, see BIP34.

Yes but the way they did this is to actually include the height in the (hashing) of the coinbase transaction. But I don't see that change in your code?

omerfirmak: The ctor I use there gets a TX hash and output index to build an input. It builds UTXO hash with hash(TX hash, output index). I just repurposed it, the Input it creates has the utxo hash of hash(Hash.init, height). It is just a unique value for each height, but an invalid UTXO hash.

AndrejMitrovic: Oh, it was a sneaky change. Could we add an explicit overload of the ctor that takes a `Height`?

AndrejMitrovic:
The only part that I find tricky here is the use of floating point. This might (probably will) have platform-specific behavior.

omerfirmak:
> The only part that I find tricky here is the use of floating point. This might (probably will) have platform-specific behavior.

If you mean the `Amount.percentage()`, I intend to change that to integers only. I was just not sure if it would overflow at that time.

omerfirmak:
I'll add more tests if needed according to the coverage report

AndrejMitrovic:
It seems like some of these commits should be squashed together. In particular it doesn't really make sense to introduce functionality in one commit and then add attributes to it in another commit in the very same PR. It seems like a lot of the commits are redesigns of each other. We don't need to keep that history in git.

AndrejMitrovic:
I'll review once the logical squashing is done (I don't expect 1 giant commit but just squash the relevant commits together please)

omerfirmak:
> I'll review once the logical squashing is done (I don't expect 1 giant commit but just squash the relevant commits together please)

Phew, this took me longer than I'd like to admit. I aged 5 years from the rebase stress :D
Hope it's now easier to review!

omerfirmak:
It is weird, CircleCI env is missing https://github.com/dlang/druntime/pull/3012

Geod24:
Unfortunately we are *still* using LDC v1.20.0...

source/agora/consensus/validation/Transaction.d
```diff
@@ -95,8 +95,8 @@ public string isInvalidReason (
             if (auto fail_reason = isInvalidInput(input, utxo_value, sum_unspent))
                 return fail_reason;
 
-            if (utxo_value.type != TxType.Payment)
-                return "Transaction: Can only freeze a Payment transaction";
+            if (utxo_value.type == TxType.Freeze)
+                return "Transaction: Can't freeze a Freeze transaction";
```
AndrejMitrovic: Slightly better wording: `Can't freeze an already frozen transaction`

source/agora/consensus/validation/Transaction.d
```diff
@@ -132,13 +132,25 @@ public string isInvalidReason (
...
         if (error !is null) return error;
     }
+    else if (tx.type == TxType.Coinbase)
+    {
+        if (tx.inputs.length != 1)
+            return "Transaction: Coinbase transactions must" ~
+                "include a single Input";
+
+        if (tx.inputs[0] != Input(height))
+            return "Transaction: Coinbase transactions invalid Input";
```
AndrejMitrovic: maybe "Coinbase transaction contains invalid input"

source/agora/node/Ledger.d
```diff
@@ -358,10 +358,10 @@ public class Ledger
         auto reason = tx.isInvalidReason(this.utxo_set.getUTXOFinder(),
             expected_height, &this.payload_checker.check);
 
-        if (reason !is null || !this.pool.add(tx))
+        if (tx.type == TxType.Coinbase || reason !is null || !this.pool.add(tx))
```
AndrejMitrovic: It would be better to skip it before tx validation. Otherwise this might become a DDos vector.

source/agora/utils/Test.d
```diff
@@ -585,6 +585,29 @@ public struct TxBuilder
...
+        Params:
+            amount = Amount to deduct
+
+        Returns:
+            Reference to `this` for easy chaining
+
+    ***************************************************************************/
+
+    public ref typeof(this) deduct (Amount amount)
+        return
```
AndrejMitrovic, line 607: This wasn't mentioned in the commit message I think.

source/agora/node/Ledger.d
```diff
@@ -193,6 +193,8 @@ public class Ledger
                             "A block loaded from disk is invalid: " ~
                             fail_reason);
                 }
+                if (block.header.height)
```
AndrejMitrovic: I would make this explicit.

source/agora/common/Config.d
```diff
@@ -174,6 +174,13 @@ public struct NodeConfig
 
     /// The factor to calculate for the fee of data payload
     public uint tx_payload_fee_factor = 200;
+
+    /// The share that Validators would get out of the transction fees
+    /// Out of 100
```
AndrejMitrovic: The rest goes to commons budget? I would add that comment.

source/agora/common/Config.d
```diff
@@ -174,6 +174,13 @@ public struct NodeConfig
...
+    public ubyte validator_tx_fee_cut = 70;
+
+    /// How frequent the payments to Validators will be
```
AndrejMitrovic: should mention this is the number of blocks (?). Is 144 from the whitepaper?

source/agora/consensus/validation/Block.d
```diff
@@ -130,6 +132,26 @@ public string isInvalidReason (const ref Block block, Height prev_height,
...
+        new_tx.type = tx.type;
+        new_tx.inputs = tx.inputs.dup;
+        new_tx.outputs = tx.outputs.dup;
+        new_tx.payload = tx.payload;
+        expected_txs ~= new_tx;
+    }
+
+    appendCoinbaseTX(expected_txs, block.header.missing_validators);
+
+    if (!isPermutation(expected_txs, block.txs))
```
AndrejMitrovic: What does this do?

AndrejMitrovic: It seems like validation becomes memory intensive now.

AndrejMitrovic (changes requested):

source/agora/common/Config.d
omerfirmak: 144 was @Geod24's suggestion.

source/agora/consensus/validation/Block.d
omerfirmak: Well, I am definitely not proud of this part. I'll try to rework it a little.

omerfirmak: Please check the latest commit, I think it is much better. I will update the docs and squash it, if it gets the OK.

source/agora/consensus/validation/Block.d
```diff
@@ -139,25 +141,12 @@ public string isInvalidReason (const ref Block block, Height prev_height,
...
-
-    appendCoinbaseTX(expected_txs, block.header.missing_validators);
-
-    if (!isPermutation(expected_txs, block.txs))
-        return "Block: Invalid Coinbase transaction";
+    auto expected_cb_txs = getCoinbaseTX(block.txs,
+        block.header.missing_validators);
+    auto incoming_cb_txs = block.txs.filter!(
+            tx => tx.type == TxType.Coinbase);
+    if (!isPermutation(expected_cb_txs, incoming_cb_txs))
```
AndrejMitrovic, line 148: I'm still not sure about this design. There can only be one coinbase per block, no? Why are there multiple ones being compared here?

omerfirmak: Incoming block could have more than one, we cant rely on that assumption. It would be validated if it only has the single expected coinbase tx.

omerfirmak: If you ask, why I didnt just check for `length == 1` is to make the check somewhat generic.

AndrejMitrovic: I see. Well I guess it's less lines of code, although the actual intention is hard to figure out from the call.

The way I would design it is to always force the coinbase tx to be the first one. That does create issues with sorting, but I would basically only sort on `txs[1..$]`.

Anyway your approach works too.

AndrejMitrovic: My solution would probably create issues with the merkle tree. So yours is better.

omerfirmak: I generally avoid stuff that may create edge cases that I can't immediately recognize. The most generalized solution I can think of is what I go with most of the time.

omerfirmak:
Squashed the last commit and rebased.

source/agora/common/Amount.d
```diff
@@ -112,6 +112,24 @@ public struct Amount
...
 
+    /***************************************************************************
+
+        Returns:
+            Greatest common divisor of `Amount`s in `ams`
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public static Amount gcd (Amount[] ams)
```
AndrejMitrovic: `in Amount[]`?

source/agora/consensus/Fee.d
```diff
@@ -187,6 +189,12 @@ public class FeeManager
     /// Out of 100
     public immutable ubyte ValidatorTXFeeCut = 70;
 
+    /// Payout period
+    public immutable ubyte PayoutPeriod = 144;
+
+    /// Total Amount of fees accumulated per address
+    Amount[PublicKey] accumulated_fees;
```
AndrejMitrovic: Missing access attribute.

source/agora/consensus/Fee.d
```diff
@@ -314,6 +322,108 @@ public class FeeManager
...
+            stakes);
+
+        foreach (idx, stake; stakes)
+        {
+            this.accumulated_fees.update(stake.output.address,
+            { return validator_fees[idx]; },
+            (ref Amount so_far) {
+                so_far.add(validator_fees[idx]);
+                return so_far;
+            });
```
AndrejMitrovic: Needs indent here.

source/agora/consensus/Fee.d
```diff
@@ -314,6 +322,108 @@ public class FeeManager
...
+            this.accumulated_fees.clear();
+        } ();
+    }
+
+    /***************************************************************************
+
+        Calculate total fees of a Transaction set
+
+        Params:
+            tx_set = Transaction set
```
AndrejMitrovic, line 420: `peekUTXO` is not documented.

source/agora/node/Ledger.d
```diff
@@ -882,6 +904,26 @@ public class Ledger
...
+
+    ***************************************************************************/
+
+    public void accumulateFees (const ref Block block) nothrow @safe
+    {
+        UTXO[] stakes;
+        this.enroll_man.getValidatorStakes(&this.utxo_set.peekUTXO, stakes);
+        assert(isStrictlyMonotonic(block.header.missing_validators));
+        foreach (idx; block.header.missing_validators.retro)
+            stakes = stakes.remove(idx);
```
AndrejMitrovic: Wouldn't it be more efficient to provide these missing validators in the `getValidatorStakes` call?

I guess it doesn't matter much if slashing will be rare though.

AndrejMitrovic (commented):
Just a few comments. Otherwise LGTM.

source/agora/node/Ledger.d
omerfirmak: Yep, I'll do that. Performance benefits may be negligible but it tidies up the caller site. Thanks!

AndrejMitrovic:
Sorry about the conflict!

omerfirmak:
Rebased

omerfirmak:
I have added DB storage for FeeManager. Please take a look at the last commit.

source/agora/consensus/Fee.d
```diff
@@ -177,16 +182,33 @@ unittest
...
+            "(public_key TEXT PRIMARY KEY, fee TEXT)");
+
+        auto results = this.db.execute("SELECT public_key, fee " ~
+            " FROM accumulated_fees");
+
+        foreach (ref row; results)
+        {
+            const key = PublicKey.fromString(row.peek!(char[])(0));
+            const fee = Amount.fromString(row.peek!(char[])(1));
+            accumulated_fees[key] = fee;
```
AndrejMitrovic: this.accumulated_fees

source/agora/consensus/Fee.d
```diff
@@ -343,6 +365,19 @@ public class FeeManager
                     return so_far;
                 }
             );
+
+            try
+            {
+                auto new_fee = this.accumulated_fees[stake.output.address]
```
AndrejMitrovic, line 371: This will throw an Error if `accumulated_fees` doesn't have this public key. Is there any case where that could happen?

omerfirmak: Nope, L361 prevents that.

AndrejMitrovic: Got it. It didn't show in the diff.

omerfirmak:
Rebased

AndrejMitrovic (approved):

omerfirmak:
Rebased again.

omerfirmak:
If there is no new comments on this today, I'll rebase and merge.

omerfirmak:
@ferencdg , is the CI failure expected?

PR #1458, opened 2021-01-04
Use alpine:edge for the runner image

Geod24:
Let's see if this works as expected.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,12 +4,13 @@ ARG DUB_OPTIONS
 ARG AGORA_VERSION="HEAD"
 ADD . /root/agora/
 WORKDIR /root/agora/
-RUN apk --no-cache add llvm-libunwind-dev
 RUN AGORA_VERSION=${AGORA_VERSION} dub build --skip-registry=all --compiler=ldc2 ${DUB_OPTIONS}
 
 # Runner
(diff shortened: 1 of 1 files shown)


kchulj (approved):

Geod24:
This is blocked by:
- An update to dtools which currently doesn't work;
- Which in turns is blocked by LDC segfaulting when an exception is thrown;
- Which is fixed by backporting my libunwind patches to LDC v1.25.0
- Which is blocked by LDC v1.25.0 existing (I could backport to v1.24.0 but it'll be much more work);
- LDC v1.25.0 **should** be released in around a month at most, according to the usual schedule;

Geod24:
Should be unblocked. I regenerated `agora-builder` already. We don't yet have the `dub` upgrade, because of some regression, but we don't need it for LDC v1.25.0. Worst case scenario we can just compile our own.

Geod24:
💚

AndrejMitrovic:
I still see 4 runs in progress?

AndrejMitrovic (approved):

Geod24:
> I still see 4 runs in progress?

This PR only affects CircleCI (the only user of the docker image)

AndrejMitrovic:
Doh that's a good point. I'll merge it.

PR #1461, opened 2021-01-05
Fix 1382: Use Base64 when serializing ubyte[]

### LINKED ISSUE #1382: Use a smarter format for JSON binary serialization
Please see [issue 2437](https://github.com/vibe-d/vibe.d/issues/2437)

Geod24:
Simply updating Vibe.d and using @ferencdg 's excellent work.
Don't merge yet, I want to verify that everything works as expected (esp. w.r.t. inheritance).

### DIFF
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -326,25 +326,28 @@ extern(C++, (StdNamespace)) extern(C++, class) struct vector (T, Alloc = allocat
 
         alias opDollar = length;
 
+        static if (is(T : ubyte))
+        {
+            import vibe.data.serialization : Base64ArrayPolicy;
+            package alias SerPolicy = Base64ArrayPolicy;
+        }
+        else
+        {
+            package alias SerPolicy = DefaultPolicy;
+        }
+
         string toString() const @trusted
         {
-            bool first = true;
-            string ret = "[ ";
-            foreach (ref entry; this.constIterator())
-            {
-                if (!first)
-                    ret ~= ", ";
-                ret ~= entry.serializeToJsonString();
-                first = false;
-            }
-            ret ~= " ]";
-            return ret;
+            import std.array : appender, Appender;
+
+            auto app = appender!string();
+            serializeWithPolicy!(JsonStringSerializer!(Appender!string), SerPolicy)(this[], app);
+            return app.data;
         }
 
         static typeof(this) fromString(string src) @safe
         {
-            import scpd.types.Utils;
-            auto array = src.deserializeJson!(T[]);
+            auto array = src.deserializeWithPolicy!(JsonStringSerializer!string, SerPolicy, T[]);
             typeof(this) vec;
             foreach (ref item; array)
                 vec.push_back(item);
@@ -395,6 +398,20 @@ extern(C++, (StdNamespace)) extern(C++, class) struct vector (T, Alloc = allocat
             }
         }
     }
+}
+
+unittest
+{
+    import std.algorithm : each;
+    import std.conv : to;
+    import std.range : iota;
+
+    vector!ubyte vec_ubyte;
+    iota(5).each!((num) {ubyte b = cast(ubyte)num; vec_ubyte.push_back(b);});
+    auto serialized = vec_ubyte.toString();
+    assert(serialized == `"AAECAwQ="`, "actual serialized: " ~ serialized);
+    auto deserialized = vec_ubyte.fromString(serialized)[];
+    assert(deserialized == [0, 1, 2, 3, 4], "actual deserialized: " ~ to!string(deserialized));
 }
 
 unittest
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -242,7 +242,7 @@ public extern(C++, (StdNamespace))  struct equal_to (T = void) {}
     (e.g. push_back).
 
     Extra items, like `ConstIterator` and `toString` / `fromString` are for
-     ease of use (e.g. `to/fromString` actually allows vibe.d to deserialize it)
+    ease of use (e.g. `to/fromString` actually allows vibe.d to deserialize it)
 
 *******************************************************************************/
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,8 +1,11 @@
 # Part of the source tree, but not needed to build Agora
 .git/
+ci/
 doc/
 hooks/
 tests/
+LICENSE
+README.md
 
 # Various build folders
 .dub/
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -23,8 +23,9 @@ import agora.common.Set;
 import agora.common.Serializer;
 import agora.consensus.data.Transaction;
 
-import vibe.web.rest;
+import vibe.data.serialization;
 import vibe.http.common;
+import vibe.web.rest;
 
 // TODO: time_t is not a fixed-size, need another type for the API
 import core.stdc.time;
@@ -68,6 +69,7 @@ public struct NodeInfo
 *******************************************************************************/
 
 @path("/")
+@serializationPolicy!(Base64ArrayPolicy)
 public interface API
 {
 // The REST generator requires @safe methods
--- a/source/agora/api/Validator.d
+++ b/source/agora/api/Validator.d
@@ -50,6 +50,7 @@ static import agora.api.FullNode;
 
 import scpd.types.Stellar_SCP;
 
+import vibe.data.serialization;
 import vibe.web.rest;
 
 ///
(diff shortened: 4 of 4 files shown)


Geod24:
Testing locally:
```console
$ docker build . -t agora
$ cd tests/system && docker-compose up && cd - 
```
This will block on `docker-compose` (gives you log)
In another terminal:
```console
$ docker run -it -v$(pwd):/agora/ --net container:system_node-2_1 nicolaka/netshoot
Starting docker container...
$ tcpdump -s0 -w /agora/capture.pcap -i eth0 '(((ip[2:2] - ((ip[0]&0xf)<<2)) - ((tcp[12]&0xf0)>>2)) != 0)'
tcpdump: listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
tcpdump: pcap_loop: The interface went down <= Happens when you shut down node2
7575 packets captured
7575 packets received by filter
0 packets dropped by kernel
```
This will capture HTTP traffic and save it in `capture.pcap`

Finally, in a third window:
```console
$ dub --root tests/system/ -- http://127.0.0.1:4000 http://127.0.0.1:4002
```
Here I was just reducing the typing overhead, but you could pass all validators address. This will send the txs and trigger nomination. After the test is done, one can open `capture.pcap` with Wireshark to get a trace of what was going on in node2.

In this case, it allowed me to realize that policies are not herited.

Geod24:
@ferencdg : This doesn't work. Am I missing something ?

Screen Shot 2021-01-05 at 12 40 45

Geod24:
Obviously: https://github.com/bpfkorea/agora/blob/ad9b297f67a67a10e9b7bea9ba9c826bd8fd97d8/source/scpd/Cpp.d#L329-L352

I wonder if there's something more clever we could do here. Perhaps specialize for `ubyte` ? In any case, it's not a blocker for this PR.

ferencdg:
> Obviously:
> 
> https://github.com/bpfkorea/agora/blob/ad9b297f67a67a10e9b7bea9ba9c826bd8fd97d8/source/scpd/Cpp.d#L329-L352
> 
> I wonder if there's something more clever we could do here. Perhaps specialize for `ubyte` ? In any case, it's not a blocker for this PR.

Yes that is the problem, and creating an ubyte specialization seems a good idea. If we copied all the elements into a temporary array then we could just call the serializeWithPolicy vibe.d method with the Base64Policy

ferencdg:
> Obviously:
> 
> https://github.com/bpfkorea/agora/blob/ad9b297f67a67a10e9b7bea9ba9c826bd8fd97d8/source/scpd/Cpp.d#L329-L352
> 
> I wonder if there's something more clever we could do here. Perhaps specialize for `ubyte` ? In any case, it's not a blocker for this PR.

Also I see some cases in our codebase where we override the `toString` method and inside the method we call vibed serialization methods. The problem is that in the current vibed design, `toString` doesn't take a Policy template, so it cannot pass it down to subsequent serialize calls. This is more like a design problem with vibed serialization that has been there before my PR(policies were existing before my PR).

ferencdg:
> > Obviously:
> > https://github.com/bpfkorea/agora/blob/ad9b297f67a67a10e9b7bea9ba9c826bd8fd97d8/source/scpd/Cpp.d#L329-L352
> > 
> > I wonder if there's something more clever we could do here. Perhaps specialize for `ubyte` ? In any case, it's not a blocker for this PR.
> 
> Yes that is the problem, and creating an ubyte specialization seems a good idea. If we copied all the elements into a temporary array then we could just call the serializeWithPolicy vibe.d method with the Base64Policy



> > Obviously:
> > https://github.com/bpfkorea/agora/blob/ad9b297f67a67a10e9b7bea9ba9c826bd8fd97d8/source/scpd/Cpp.d#L329-L352
> > 
> > I wonder if there's something more clever we could do here. Perhaps specialize for `ubyte` ? In any case, it's not a blocker for this PR.
> 
> Also I see some cases in our codebase where we override the `toString` method and inside the method we call vibed serialization methods. The problem is that in the current vibed design, `toString` doesn't take a Policy template, so it cannot pass it down to subsequent serialize calls. This is more like a design problem with vibed serialization that has been there before my PR(policies were existing before my PR).

If we didn't have the above problem(policies are not passed down to the toString methods), then we could just create our own agora serialization policy that doesn't just work with ubyte8, but also with our custom vector!ubyte class. This is less intrusive then adding the specialized `toString` method to the vector class

```D
public class Base64AgoraArrayPolicy (R) if ((isArray!R && is(ElementType!R : ubyte)) || is(R : vector!ubyte))
{
	public static string toRepresentation (in R data) @safe pure
	{
		...
	}

	public static ubyte[] fromRepresentation (in string data) @safe pure
	{
		...
	}
}
```

Geod24:
How long do you think it'd take to implement?

ferencdg:
> How long do you think it'd take to implement?

If I don't want to break vibe.d backward compatibility, then it is kind of a messy change, because we already have now so many toString variants(type supporting sink or not/serializer supporting sink or not/toString can take policy parameter or not):

https://github.com/vibe-d/vibe.d/blob/master/data/vibe/data/serialization.d#L483-L493
(even after I make those changes on toString, vibed has a lot of other similar methods like toISOExtString, toRepresentation that doesn't have access to Policies)

I would need to duplicate and slightly modify a lot of already existing templates and write a lot of new unit tests. On the vibe.d side it wouldn't take more than 3 days and on agora side the new serialization policy with test is maximum 2 days.

Geod24:
Well, two things to consider:
- The `sink` version hasn't been released yet, so we can still "break" them;
- It's my understanding that `toRepresentation` and `toISOString` are here because the serialization options are too limited;

In particular, `toRepresentation` gives you a binary representation. That's not of much help for REST interface (unless you decide to use that as a fallback). Likewise, the inability to select a middleware for serialization is quite limiting. Express got it quite rare with [their middleware approach](https://expressjs.com/en/guide/using-middleware.html), although you can get away with much more in a dynamic language.

Ideally you should even be able to specify, in Vibe.d, different serialization depending on the content-type. Of course that's not useful for us, but it's one of the idea behind REST.

Geod24:
> If we didn't have the above problem(policies are not passed down to the toString methods)

Do we need the policy to be passed down ? We can just eliminate the `toString` method if we have a policy, can't we ?

ferencdg:
1. Background

    1. The usecase I mentioned is the SCPStatement which has the toString already overriden(I guess vibed 
    doesn't like unions), and inside we have calls into serializeJson
    https://github.com/bpfkorea/agora/blob/v0.x.x/source/scpd/types/Stellar_SCP.d#L114
    The end goal would be to somehow pass that policy down to vector.

    2. Policy is something that needs to be provided when someone call's vibed's 'serizalizeJson()' method, it is not a global settings

2. Manually passing policies <br/>
we could pass down our own serialization policy manually(SerializeUbyteVectorPolicy) in the code I referenced above, if we want(but ideally we would just pass in whatever we got in the 'SCPStatement.toString()' method. <br />
This SerializeUbyteVectorPolicy policy would take precende over the 'vector.toString()' method(which method we would still keep for other types) so for vector!ubyte we would use the base64 encoding and for everything else we would use the existing 'vector.toString()' implementation.<br />
We could also write a 'SerializeALLVectorPolicy' instead that takes care of serializing any kind of vectors and pass that down manually.<br />
One big disadvantage of this solution is that now if we have a new class called SuperEnvelope that also has a vector inside it, but SuperEnvelope DOESN't need to override the 'SuperEnvelope.toString()' method: now we do have to write 'SuperEnvelope.toString()', so we can pass down the policy manually(unless we do the vibe.d fix I mentioned in my previous post)

3. Specializing 'vector.toString'
However I think your original idea of just specializing 'vector.toString()' (and reusing Base64Encoding policy there) is easier and less messy.

source/scpd/Cpp.d
```diff
@@ -328,23 +328,44 @@ extern(C++, (StdNamespace)) extern(C++, class) struct vector (T, Alloc = allocat
...
+                import vibe.data.serialization : Base64ArrayPolicy;
+                import std.array : appender, Appender;
+
+                auto app = appender!string();
+                serializeWithPolicy!(JsonStringSerializer!(Appender!string), Base64ArrayPolicy)(this[], app);
+                return app.data;
+            }
+            else
+            {
+                bool first = true;
```
ferencdg: Any idea why the original code didn't just call the serialize method on this[]? I think that is equivalent to the below verbose code. I removed the old code, please let me know if you see any difference.

AndrejMitrovic:
LGTM.

But this will break Stoa correct? So maybe an issue should be filed to https://github.com/bpfkorea/stoa/issues/new to make sure it's ready.

Geod24:
> But this will break Stoa correct?

I don't think so, actually. Because Stoa doesn't deal with the SCP envelope, and `BitField` is `uint[]`. Do we have any `ubyte[]` ?

AndrejMitrovic (approved):

Issue #1407, opened 2020-12-07
libexecinfo (used on Alpine Linux) is buggy 


Geod24:
Currently we depend on `libexecinfo` for unwinding stack traces. Unfortunately it has some issues and can crash the program. E.g. see https://github.com/mikroskeem/libexecinfo/issues/2 and related thread.

One way to work around this would be to use `libunwind`, which is portable across platforms, and included in both GCC and LLVM. But that means druntime needs some adaptation.

Geod24:
First part of said adaptation: https://github.com/dlang/druntime/pull/3304

Geod24:
I have a fairly advanced WIP but I hit another issue in the compiler that is a blocker. It's a regression on lambda type inference introduce in v2.091.

Geod24:
Submitted a PR, and will get it through the finish line in the coming week. Expect it by v2.096.0 though.
https://github.com/dlang/druntime/pull/3321

Geod24:
Worked around in #1462

PR #1462, opened 2021-01-05
Work around buggy libexecinfo by defining our own traceHandler

Geod24:
On Alpine Linux, which uses Musl libc, there is no backtrace() and co.
As a result, to provide trace information when an exception is thrown,
druntime was made to rely on libexecinfo, which provides a backtrace-like interface.

However, it turned out that this library is buggy and routinely leads to SEGV
in various pieces of code. Hence, we are in the process of adding support
for libunwind to druntime, which is the industry standard when it comes to frame inspection.

The work upstream is almost complete, and the results we've seen are very
encouraging. However, the support will only land in v2.096.0, which means
LDC v1.26.0, which is two release away, estimated to be around 4 months from now.
It is likely that this support can be back-ported to LDC v1.25.0 with a little effort,
which would put the ETA at 1 month, much more reasonable.

Given this uncertain schedule, we want to work around the issue by copying over the support
from the druntime PR. The workaround is rather crude: we simply define a trace handler
that prints addresses and function names. This means we loose file/line information temporarily,
but Agora should be much more stable.

If the missing file/line information turns out to be problematic,
we could get it back by copying over the relevant parts of druntime,
as they have been made to be self-contained by our previous refactorings.

As a last note, this also means we force users to link in `libunwind` on other POSIX platforms.
It's not an issue on OSX, as it's distributed with the development environment,
but will create another constraint on Linux.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -62,9 +62,7 @@ jobs:
       if: runner.os == 'Linux'
       run: |
         sudo apt-get update
-        sudo apt-get install libsqlite3-dev
-        sudo apt-get install libsodium-dev
-        sudo apt-get install g++-9
+        sudo apt-get install g++-9 libsodium-dev libsqlite3-dev libunwind-dev
 
     - name: '[Windows] Install dependencies & setup environment'
       if: runner.os == 'Windows'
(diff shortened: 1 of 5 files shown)


Geod24:
Kinda sorta fixes https://github.com/bpfkorea/agora/issues/1407

Geod24:
Green

hewison-chris (approved):
LGTM.

AndrejMitrovic (approved):

PR #1464, opened 2021-01-05
Let Transactions using the same inputs be added to the pool

### LINKED ISSUE #1332: TransactionPool should not filter double-spend transactions
The current design of the TransactionPool prohibits the introduction of features such as:

- replace by fee
- transaction selection & filter mechanism #407

In the current design if two transactions which spend the same UTXO are sent to the node, only one of the transactions will be accepted in the pool (the first one). That makes the transaction selection a race condition. And it also means the transaction pools across the nodes are in a conflicted state - it may affect consensus for example.

Of course we also don't want to blow up the size of the pool. A possible middle-ground would be to add the ability to group double-spend txs together. The pool could for example allow up to N transactions with the same UTXO, and later on the transaction selection algorithm (#407) could pick the best candidate out of the double-spend set. When the block is externalized, all related tx's spending that same utxo would be removed from the pool, and not just the tx's that are in the block.

omerfirmak:
Fixes #1332

### DIFF
--- a/source/agora/common/TransactionPool.d
+++ b/source/agora/common/TransactionPool.d
@@ -81,8 +81,8 @@ public class TransactionPool
     /// SQLite db instance
     private ManagedDatabase db;
 
-    /// Map for hashes for Input objects of each transaction in TX pool
-    private Set!Hash input_set;
+    /// Keeps track of which TXs spend which inputs
+    private Set!Hash[Hash] spenders;
 
     /***************************************************************************
 
@@ -110,10 +110,7 @@ public class TransactionPool
 
         // populate the input set from the tx'es in the DB
         foreach (const ref Transaction tx; this)
-        {
-            foreach (const ref input; tx.inputs)
-                this.input_set.put(input.hashFull());
-        }
+            this.updateSpenderList(tx);
     }
 
     /***************************************************************************
@@ -137,8 +134,7 @@ public class TransactionPool
             return false;
 
         // insert each input information of the transaction
-        foreach (const ref input; tx.inputs)
-            this.input_set.put(input.hashFull());
+        this.updateSpenderList(tx);
 
         serializeToBuffer(tx, buffer);
 
@@ -156,17 +152,107 @@ public class TransactionPool
 
         Params:
             tx = the transaction to remove
+            rm_double_spent = remove the TXs that use the same inputs
 
     ***************************************************************************/
 
-    public void remove (const ref Transaction tx) @trusted
+    public void remove (const ref Transaction tx,
+        bool rm_double_spent = true) @trusted
     {
-        // delete inputs of transaction from the set of Input hashes
-        foreach (const ref input; tx.inputs)
-            this.input_set.remove(input.hashFull());
+        auto tx_hash = tx.hashFull();
+        if (!this.hasTransactionHash(tx_hash))
+            return;
 
-        auto hash = tx.hashFull();
-        this.db.execute("DELETE FROM tx_pool WHERE key = ?", hash);
+        this.db.execute("DELETE FROM tx_pool WHERE key = ?", tx_hash);
+
+        if (rm_double_spent)
+        {
+            // Incoming TX is accepted into the chain, so any other TXs in the pool
+            // that use the same inputs are now invalid, remove them too
+            Set!Hash inv_txs;
+
+            this.gatherDoubleSpentTXs(tx, inv_txs);
+            foreach (input; tx.inputs)
+                this.spenders.remove(input.hashFull());
+
+            inv_txs.each!(inv_tx_hash => this.remove(inv_tx_hash, false));
+        }
+        else
+            foreach (input; tx.inputs)
+                if (auto list = input.hashFull() in this.spenders)
+                    (*list).remove(tx_hash);
+    }
+
+    /// Ditto
+    public void remove (Hash tx_hash,
+        bool rm_double_spent = true) @trusted
+    {
+        if (!this.hasTransactionHash(tx_hash))
+            return;
+
+        auto results = this.db.execute("SELECT val FROM tx_pool " ~
+            "WHERE key = ?", tx_hash);
+        if (!results.empty)
+        {
+            auto tx = deserializeFull!Transaction(results
+                .front.peek!(ubyte[])(0));
+            this.remove(tx, rm_double_spent);
+        }
+    }
+
+    /***************************************************************************
+
+        Add the given TX to `spenders` list
+
+        Params:
+            tx = the transaction to add
+
+    ***************************************************************************/
+
+    private void updateSpenderList (const ref Transaction tx) @safe
+    {
+        auto tx_hash = tx.hashFull();
+
+        // insert each input information of the transaction
+        foreach (const ref input; tx.inputs)
+        {
+            const in_hash = input.hashFull();
+            // Update the spenders list
+            if (in_hash !in this.spenders)
+                this.spenders[in_hash] = Set!Hash();
+            this.spenders[in_hash].put(tx_hash);
+        }
+    }
+
+    /***************************************************************************
+
+        Gather TXs that share inputs with the given TX
+
+        Params:
+            tx = a transaction
+            double_spent_txs = container to write the double-spend TXs
+
+        Returns:
+            true if double-spend TXs where found, false otherwise
+
+    ***************************************************************************/
+
+    private bool gatherDoubleSpentTXs (const ref Transaction tx,
+        ref Set!Hash double_spent_txs)
+    {
+        double_spent_txs.clear();
+
+        const tx_hash = tx.hashFull();
+        foreach (const ref input; tx.inputs)
+        {
+            const in_hash = input.hashFull();
+            if (auto cur_spenders = in_hash in this.spenders)
+                foreach (spender; *cur_spenders)
+                    if (spender != tx_hash)
+                        double_spent_txs.put(spender);
+        }
+
+        return double_spent_txs.length > 0;
     }
 
     /***************************************************************************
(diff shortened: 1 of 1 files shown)


source/agora/common/TransactionPool.d
```diff
@@ -136,15 +133,15 @@ public class TransactionPool
         if (!isValidTransaction(tx))
             return false;
 
+        auto tx_hash = hashFull(tx);
```
linked0: It's used only once. So, I think it's not necessary.

source/agora/common/TransactionPool.d
```diff
@@ -156,17 +153,88 @@ public class TransactionPool
 
         Params:
             tx = the transaction to remove
+            rm_double_spent = remove the TXs that use the same inputs
 
     ***************************************************************************/
 
-    public void remove (const ref Transaction tx) @trusted
+    public void remove (const ref Transaction tx,
+        bool rm_double_spent = true) @trusted
```
linked0, line 170: Are there any use cases for the `rm_double_spent` of `false` value? It would be good to make the concept and code simple. So if there are not, how about removing the parameter for the moment?

source/agora/common/TransactionPool.d
```diff
@@ -156,17 +153,88 @@ public class TransactionPool
...
+            "WHERE key = ?", tx_hash);
+        if (!results.empty)
+        {
+            auto tx = deserializeFull!Transaction(results
+                .front.peek!(ubyte[])(0));
+            this.remove(tx, rm_double_spent);
+        }
+    }
+
+    ///
```
linked0: Please add document

source/agora/common/TransactionPool.d
```diff
@@ -156,17 +153,88 @@ public class TransactionPool
...
+            if (auto existing = in_hash in this.spenders)
+                *existing ~= tx_hash;
+            else
+                this.spenders[in_hash] = [tx_hash];
+        }
+    }
 
-        auto hash = tx.hashFull();
-        this.db.execute("DELETE FROM tx_pool WHERE key = ?", hash);
+    ///
```
linked0: Please add document.

source/agora/common/TransactionPool.d
```diff
@@ -274,11 +284,50 @@ public class TransactionPool
...
             {
-                auto tx = deserializeFull!Transaction(row.peek!(ubyte[])(0));
+                auto key = Hash(row.peek!(char[])(0));
+                if (key in skip_txs)
+                    continue;
+
+                auto tx = deserializeFull!Transaction(row.peek!(ubyte[])(1));
+
+                Set!Hash db_keys; // Gather double spent TX keys
+                if (this.selector && this.gatherDoubleSpentTXs(tx, db_keys))
```
linked0, line 320: The` selector` variable has been already set to the `defaultSelector` function. So it doesn't need to check if it's null.

source/agora/common/TransactionPool.d
```diff
@@ -466,18 +522,95 @@ unittest
...
+    assert(pool.add(tx1));
+    assert(pool.add(tx2));
+    assert(pool.length == 2);
+    pool.remove(tx2);
+    assert(pool.length == 0);
+}
 
-    // add duplicate tx into second pool => return true
-    assert(pool2.add(tx2));
+unittest
```
linked0, line 625: A simple description about the test is needed.

source/agora/common/TransactionPool.d
```diff
@@ -466,18 +522,95 @@ unittest
...
+    assert(pool.add(tx2));
+    assert(pool.add(tx3));
+    assert(pool.length == 3);
+    // tx3 should be removed as well.
+    pool.remove(tx2);
+    assert(pool.length == 1);
+    assert(pool.hasTransactionHash(tx1.hashFull()));
+}
+
+unittest
```
linked0, line 666: Ditto

source/agora/common/TransactionPool.d
```diff
@@ -614,3 +669,74 @@ unittest
     assert(pool.length == 1);
     assert(pool.hasTransactionHash(tx3.hashFull()));
 }
+
+unittest
```
linked0, line 706: A simple description about the test is needed.

linked0 (commented):
Good work!
I left some suggestions. If my comments are not understandable, please let me know.

source/agora/common/TransactionPool.d
```diff
@@ -81,8 +81,8 @@ public class TransactionPool
     /// SQLite db instance
     private ManagedDatabase db;
 
-    /// Map for hashes for Input objects of each transaction in TX pool
-    private Set!Hash input_set;
+    /// Keeps track of which TXs spend which inputs
+    private Hash[][Hash] spenders;
```
AndrejMitrovic: Shouldn't this still be a `Set!Hash[Hash]`? We don't want to add duplicate transactions (meaning ones with the exact same hash).

AndrejMitrovic: Well I guess `isValidTransaction` already handles duplicate txs. You can keep it as an array if it makes the code simpler I suppose.

source/agora/common/TransactionPool.d
```diff
@@ -55,6 +55,9 @@ static this ()
     TransactionPool.initialize();
 }
 
+/// A delegate type to select one of the double spent TXs
+public alias doubleSpentSelector = ulong delegate(Transaction[] txs);
```
AndrejMitrovic: Type definitions begin with uppercase, not lowercase. So `DoubleSpentSelector`.

source/agora/common/TransactionPool.d
```diff
@@ -274,11 +284,50 @@ public class TransactionPool
     {
         () @trusted
         {
-            auto results = this.db.execute("SELECT val FROM tx_pool");
+            // TXs that went through the "selection" process and should be
```
AndrejMitrovic, line 305: Well this works, but I wonder if in the future we can replace it with something simpler. What I had in mind was:

- TransactionPool provides a range interface, maybe through a `.range()` getter or `opSlice`
- The element type of the range is a `Transaction[]`. So if you would call `.array()` on this range you would actually get `Transaction[][]`.
- If the length of an element of this range is == 1, it's not a double-spend and it can be selected.
- If the length of an element is > 1, it means this set of transactions are double-spends. The calling code can then decide what to do with it

source/agora/common/TransactionPool.d
Geod24: I had the exact same question.

omerfirmak: See L179, that call gets redirected to this overload with `rm_double_spent = false`

source/agora/common/TransactionPool.d
omerfirmak: The duplicate TX addition has very little to do with the `spenders` structure, the main storage is the DB. But I dont think `Set!Hash[Hash]` would make the code any more complex and would better represent the intention. I will give it a try :+1:

source/agora/common/TransactionPool.d
omerfirmak: L118 calls `opApply` with a null `selector`. It is intentional, to not filter any double-spend transactions while rebuilding `spenders`.

source/agora/common/TransactionPool.d
omerfirmak: So, do we let double-spends through `opApply`?

AndrejMitrovic: Well I would get rid of `opApply` entirely. But there might be use-cases for wanting to walk through double-spends in the future..

AndrejMitrovic: Anyway it was just a suggestion, we can go ahead with your design for now.

omerfirmak: My intention was to keep the old behavior and API while extending the functionality.

source/agora/common/TransactionPool.d
omerfirmak: It even made it a touch simpler :) Thanks @AndrejMitrovic

omerfirmak:
Thanks for reviews! I think this is ready. All comments are addressed.

source/agora/common/TransactionPool.d
linked0: > See L179, that call gets redirected to this overload with rm_double_spent = false

Yeah, right.
But, calling each other `remove` functions is kind of complex to understand. It would be good to make only the second `remove` be able to call the first one. The first one can be called recursively. And then the second `remove` function does not need the `rm_double_spent` parameter, I think. It's just a suggestion.

source/agora/common/TransactionPool.d
omerfirmak: What you are describing is actually `spenders`. We can easily provide a way to access it down the line, if we ever need that.

source/agora/common/TransactionPool.d
omerfirmak: The second one is essentially a wrapper to the first one. So they either both have the `rm_double_spent` or not.

With the original `remove`, we still need to know if we are deleting the TX because it got accepted or because it is now invalid because of another TX that got accepted in to the chain. In the first case, we have to delete the TXs that was made invalid by acceptance of the Transaction `tx`.  In the second case, we just need to delete the `tx` itself. That is why we need the `rm_double_spent`

source/agora/common/TransactionPool.d
```diff
@@ -55,6 +55,9 @@ static this ()
     TransactionPool.initialize();
 }
 
+/// A delegate type to select one of the double spent TXs
+public alias DoubleSpentSelector = ulong delegate(Transaction[] txs);
```
Geod24: `size_t delegate (in Transaction[] txs)`

Geod24 (commented):
The first commit LGTM but why do we need the second commit ? Why can't the caller handle the double-spent case ?

omerfirmak:
> The first commit LGTM but why do we need the second commit ? Why can't the caller handle the double-spent case ?

With the current interface (`opApply`) it did not make sense to me to put that on the caller. Caller would have to detect the double-spent and group the related TXs. To do that, they would have to go through the entire pool even if they need only `N` TXs.

Also, if caller sets `selector` to `null`, it can bypass the filtering.

IMO, we either go with what @AndrejMitrovic [suggested](https://github.com/bpfkorea/agora/pull/1464#discussion_r552341179) or this.

Geod24 (approved):
Alright. Not too keen on the design, but I can't offer a better alternative ATM. Just fix the DG type and we're g2g.

omerfirmak:
> Alright. Not too keen on the design, but I can't offer a better alternative ATM. Just fix the DG type and we're g2g.

Done, rebased too.

PR #1465, opened 2021-01-06
Decouple TxBuilder from WellKnownKeys

### LINKED ISSUE #1348: Decouple TxBuilder from well-known KeyPair
`TxBuilder` was originally created as a convenient way to create any kind of transactions in the test suite.
However, it has proved good enough to be used as a general `Transaction`-building facility.

As it's design relies on the input being attributed to the well-known keypairs, it cannot yet be  used outside of the testing mode. This should be fixed  so that it both keeps the convenience it has in the test-suite, but works with not-yet-known keypairs.

There's multiple ways to handle this:
```D
/// TxBuilder was renamed to TxMaker:
struct TxMaker (alias LookupFn) { /* ... */ }
alias TxBuilder = TxMaker!(WK.Keys.opIndex);
```

Another approach is to use traditional OOP and to  simply expose a global object.
One of the goal here is to avoid having to edit every single use of it

kchulj:
`TxBuilder` could only be used with the well-known keypairs. Now, a transaction can be signed with not-yet-known keypairs.
If one wants to use not-yet-known keypairs, then they can call `signTx`.

Fixes #1348

### DIFF
--- a/source/agora/utils/Test.d
+++ b/source/agora/utils/Test.d
@@ -45,6 +45,7 @@ import std.algorithm;
 import std.array;
 import std.file;
 import std.format;
+import std.functional;
 import std.path;
 import std.range;
 
@@ -435,13 +436,16 @@ public struct TxBuilder
         Params:
             type = type of `Transaction`
             data = data payload of `Transaction`
+            lookupDg = delegate to look up the `KeyPair`
 
         Returns:
             The finalized & signed `Transaction`.
 
     ***************************************************************************/
 
-    public Transaction sign (TxType type = TxType.Payment, const(ubyte)[] data = []) @safe nothrow
+    public Transaction sign (TxType type = TxType.Payment, const(ubyte)[] data = [],
+        scope KeyPair delegate(PublicKey pubkey) @safe nothrow lookupDg = toDelegate(&WK.Keys.opIndex))
+        @safe nothrow
     {
         assert(this.inputs.length, "Cannot sign input-less transaction");
         assert(this.data.outputs.length || this.leftover.value > Amount(0),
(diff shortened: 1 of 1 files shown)


kchulj:
This is an example of a lookup function which can be used with our own `myKeys`. It would be an argument in `signTx`.
```D
public KeyPair lookup (PublicKey pubkey)
{
    KeyPair[] myKeys;
    SecretKey[PublicKey] mapping;
    foreach (const ref key; myKeys)
        mapping[key.address] = key.secret;

    auto result = mapping.byKeyValue().find!(kv => kv.key == pubkey);
    return result.empty ? KeyPair.init : KeyPair(pubkey, result.front);
}
```

source/agora/utils/Test.d
```diff
@@ -595,6 +598,27 @@ public struct TxBuilder
...
+            txb = a `TxBuilder` instance
+            type = type of `Transaction`
+            data = data payload of `Transaction`
+
+        Returns:
+            The finalized & signed `Transaction`.
+
+    ***************************************************************************/
+
+public Transaction signTx (ref TxBuilder txb, TxType type = TxType.Payment,
```
AndrejMitrovic: It seems the indentation of this (and the `spendable`) function is wrong. Shouldn't it be indented like the ddoc comment?

AndrejMitrovic: Also I don't see where this function is used?

kchulj: > It seems the indentation of this (and the spendable) function is wrong. Shouldn't it be indented like the ddoc comment?

I simply followed the style of `spendable`, however that's not the case for other ddocs. Perhaps the ddoc for this and `spendable` should have no indentation ? (they are not part of the `TxBuilder` struct)

AndrejMitrovic: Right then the docs should be unindented.

Geod24:
`toDelegate` is a nice find. But why do you need the overload then ?

source/agora/utils/Test.d
kchulj: > Also I don't see where this function is used?

"One of the goal here is to avoid having to edit every single use of it"
I am thinking of leaving the existing `sign`s, and this one can be used whenever one wants to use a not-well-known KeyPair. Open for suggestions.

kchulj:
> `toDelegate` is a nice find. But why do you need the overload then ?

Yes, the delegate can now be an optional argument.
Thank you @omerfirmak for the help :)

Geod24 (approved):
LGTM

PR #1468, opened 2021-01-06
Check validation for slashing information in a block

### LINKED ISSUE #1444: Slashing information validation
This issue is related to #1076 which is about implementing the slashing protocol. We need to split some tasks related to block header validation from issue #1076 because the tasks have many things to do and could consume some time.
 
We've been adding information about slashed validators into a block. The newly added information is as follows.

```
/// Hash of random seed of the preimages for this this height
Hash random_seed;

/// List of indices to the validator UTXO set which have not revealed the preimage
uint[] missing_validators;
```

The `random_seed` must not null in any case, which is generated from the preimages of current validators, and it must be the same as the local `random_seed` from a node itself. It's the things for the validation routine to check. In the current code of Agora, there are 2 cases the information is not valid as follows besides the case where a misbehaving validator manipulates the information.

1. In the catch-up process, there are no pre-images for generating the local random seed needed for the comparison. So we need a process to get pre-images from other nodes and add the information into the `ValidatorSet`.
2. In many unit tests or network tests, there are some cases that do not consider maintaining preimages needed for validation check and generating the `random_seed`.

As a goal of this issue, we have to add the validation routine about slashing information of the block header and make the code pass the above situations and tests.

linked0:
We've added information about slashed validators into a block like `random_seed` and `missing_validators`. So, when a new block is being made, we should check the validity of the slashing information. There could be a situation where preimages needed for validation are missing like in the catch-up process. So we should be able to get the preimages from the network. 

Fixes #1444

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -215,6 +215,24 @@ public interface API
 
     /***************************************************************************
 
+        Get validators' pre-image information
+
+        API:
+            GET /preimages
+
+        Params:
+            start_height = the starting enrolled height to begin retrieval from
+            end_height = the end enrolled height to finish retrieval to
+
+        Returns:
+            preimages' information of the validators
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getPreimages (ulong start_height, ulong end_height);
+
+    /***************************************************************************
+
         Reveals a pre-image
 
         API:
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -785,6 +785,35 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Add pre-images for enrolled validators
+
+        Params:
+            preimages = the pre-images to add
+
+        Returns:
+            true if preimages was added successfully
+
+    ***************************************************************************/
+
+    public bool addPreimages (const PreImageInfo[] preimages) @safe nothrow
+    {
+        foreach (image; preimages)
+        {
+            auto stored_image =
+                this.validator_set.getPreimage(image.enroll_key);
+            if (stored_image == PreImageInfo.init ||
+                stored_image.distance >= image.distance)
+                continue;
+
+            if (!this.validator_set.addPreimage(image))
+                return false;
+        }
+
+        return true;
+    }
+
+    /***************************************************************************
+
         Get the public key of node that is used for a enrollment
 
         Returns:
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -748,6 +748,26 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Get validators' pre-image information
+
+        Params:
+            start_height = the starting enrolled height to begin retrieval from
+            end_height = the end enrolled height to finish retrieval to
+
+        Returns:
+            preimages' information of the validators
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getValidatorPreimages (Height start_height,
+        Height end_height) @safe nothrow
+    {
+        return this.validator_set.getPreimages(start_height,
+            end_height);
+    }
+
+    /***************************************************************************
+
         Add a pre-image information to a validator data
 
         Params:
--- a/source/agora/consensus/SlashPolicy.d
+++ b/source/agora/consensus/SlashPolicy.d
@@ -213,11 +213,6 @@ public class SlashPolicy
                 valid_keys ~= key;
         }
 
-        // TODO: This should not happen in fact, which situation will be handled
-        // in issue #1444. (https://github.com/bpfkorea/agora/issues/1444)
-        if (valid_keys.length == 0)
-            return Hash.init;
-
         return this.enroll_man.getRandomSeed(valid_keys, height);
     }
 
--- a/source/agora/consensus/state/ValidatorSet.d
+++ b/source/agora/consensus/state/ValidatorSet.d
@@ -582,6 +582,48 @@ public class ValidatorSet
 
     /***************************************************************************
 
+        Get validators' pre-image information
+
+        Params:
+            start_height = the starting enrolled height to begin retrieval from
+            end_height = the end enrolled height to finish retrieval to
+
+        Returns:
+            preimages' information of the validators
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getPreimages (Height start_height,
+        Height end_height) @trusted nothrow
+    {
+        PreImageInfo[] preimages;
+
+        try
+        {
+            auto results = this.db.execute("SELECT key, preimage, distance " ~
+                "FROM validator_set WHERE enrolled_height >= ? AND " ~
+                "enrolled_height <= ?",
+                start_height, end_height);
+
+            foreach (row; results)
+            {
+                Hash enroll_key = Hash(row.peek!(char[])(0));
+                Hash preimage = Hash(row.peek!(char[])(1));
+                ushort distance = row.peek!ushort(2);
+                preimages ~= PreImageInfo(enroll_key, preimage, distance);
+            }
+        }
+        catch (Exception ex)
+        {
+            log.error("Exception occured on getPreimages: {}, heights " ~
+                "[{}..{}]", ex.msg, start_height, end_height);
+        }
+
+        return preimages;
+    }
+
+    /***************************************************************************
+
         Get validator's pre-image for the given block height from the
         validator set
 
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -83,6 +83,7 @@ version (unittest)
         active_enrollments = the number of enrollments that do not expire
             at the next block height (prev_height + 1).
         enrolled_validators = the number of validators enrolled at this height
+        random_seed = Hash of random seed of the preimages
         getValidatorAtIndex = delegate to provide validator Point K for
             given height and index into map
         getCommitmentNonce = delegate to provide the commitment Nonce of the
@@ -102,7 +103,8 @@ version (unittest)
 
 public string isInvalidReason (const ref Block block, Height prev_height,
     in Hash prev_hash, scope UTXOFinder findUTXO, scope FeeChecker checkFee,
-    scope EnrollmentFinder findEnrollment, size_t active_enrollments, size_t enrolled_validators,
+    scope EnrollmentFinder findEnrollment, size_t active_enrollments,
+    size_t enrolled_validators, in Hash random_seed,
     Point delegate (Height, ulong) nothrow @safe getValidatorAtIndex,
     Point delegate (const ref Point, const Height) nothrow @safe getCommitmentNonce,
     ulong prev_timestamp, ulong curr_timestamp, Duration block_timestamp_tolerance_dur,
@@ -178,6 +180,12 @@ public string isInvalidReason (const ref Block block, Height prev_height,
         if (auto fail_reason = VEn.isInvalidReason(enrollment, enrollmentsUTXOFinder,
                                             block.header.height, findEnrollment))
             return fail_reason;
+    }
+
+    if (block.header.random_seed == Hash.init
+        || block.header.random_seed != random_seed)
+    {
+        return "Block: Header's random seed does not match that of known pre-images";
     }
 
     Point sum_K;
@@ -640,14 +648,17 @@ version (unittest)
     public string isValidcheck (const ref Block block, Height prev_height,
         Hash prev_hash, scope UTXOFinder findUTXO,
         size_t active_enrollments, size_t enrolled_validators, scope FeeChecker checkFee,
-        scope EnrollmentFinder findEnrollment, ulong enrollment_cycle = 0,
-        ulong prev_timestamp = 0, ulong curr_timestamp = ulong.max,
+        scope EnrollmentFinder findEnrollment, Hash random_seed = Hash.init,
+        ulong enrollment_cycle = 0, ulong prev_timestamp = 0, ulong curr_timestamp = ulong.max,
         Duration block_timestamp_tolerance_dur = 100.seconds,
         string file = __FILE__, size_t line = __LINE__) nothrow @safe
     {
+        if (random_seed == Hash.init)
+            random_seed = getTestRandomSeed();
+
         return isInvalidReason(block, prev_height, prev_hash, findUTXO,
             checkFee, findEnrollment, active_enrollments, enrolled_validators,
-            (Height h, ulong i) @safe nothrow
+            random_seed, (Height h, ulong i) @safe nothrow
             {
                 return Point(genesis_validator_keys[i].address[]);
             },
@@ -670,14 +681,15 @@ version (unittest)
     public bool isValid (const ref Block block, Height prev_height,
         Hash prev_hash, scope UTXOFinder findUTXO,
         size_t active_enrollments, size_t enrolled_validators, scope FeeChecker checkFee,
-        scope EnrollmentFinder findEnrollment, ulong enrollment_cycle = 0,
-        ulong prev_timestamp = 0, ulong curr_timestamp = ulong.max,
+        scope EnrollmentFinder findEnrollment, Hash random_seed = Hash.init,
+        ulong enrollment_cycle = 0, ulong prev_timestamp = 0, ulong curr_timestamp = ulong.max,
         Duration block_timestamp_tolerance_dur = 100.seconds,
         string file = __FILE__, size_t line = __LINE__) nothrow @safe
     {
         string reason = isValidcheck(block, prev_height, prev_hash, findUTXO,
-            active_enrollments, enrolled_validators, checkFee, findEnrollment, enrollment_cycle,
-            prev_timestamp, curr_timestamp, block_timestamp_tolerance_dur);
+            active_enrollments, enrolled_validators, checkFee, findEnrollment,
+            random_seed, enrollment_cycle, prev_timestamp, curr_timestamp,
+            block_timestamp_tolerance_dur);
         if (reason !is null)
             log.trace("[{}:{}] block height {} is invalid: {}",
                 file, line, block.header.height, reason);
@@ -688,13 +700,14 @@ version (unittest)
     public bool isNotValid (const ref Block block, Height prev_height,
         Hash prev_hash, scope UTXOFinder findUTXO,
         size_t active_enrollments, size_t enrolled_validators, scope FeeChecker checkFee,
-        scope EnrollmentFinder findEnrollment, ulong enrollment_cycle = 0,
-        ulong prev_timestamp = 0, ulong curr_timestamp = ulong.max,
+        scope EnrollmentFinder findEnrollment, Hash random_seed = Hash.init,
+        ulong enrollment_cycle = 0, ulong prev_timestamp = 0, ulong curr_timestamp = ulong.max,
         Duration block_timestamp_tolerance_dur = 100.seconds) nothrow @safe
     {
         return isValidcheck(block, prev_height, prev_hash, findUTXO,
-            active_enrollments, enrolled_validators, checkFee, findEnrollment, enrollment_cycle,
-            prev_timestamp, curr_timestamp, block_timestamp_tolerance_dur) !is null;
+            active_enrollments, enrolled_validators, checkFee, findEnrollment,
+            random_seed, enrollment_cycle, prev_timestamp, curr_timestamp,
+            block_timestamp_tolerance_dur) !is null;
     }
 }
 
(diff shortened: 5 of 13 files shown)


source/agora/consensus/state/ValidatorSet.d
```diff
@@ -579,6 +579,48 @@ public class ValidatorSet
...
+                preimages ~= PreImageInfo(enroll_key, preimage, distance);
+            }
+        }
+        catch (Exception ex)
+        {
+            log.error("Exception occured on getPreimages: {}, heights " ~
+                "[{}..{}]", ex.msg, start_height, end_height);
+        }
+
+        return preimages;
```
hewison-chris, line 622: You could consider using fold to build the array directly from the results:-
```
public PreImageInfo[] getPreimages (ulong start_height,
        ulong end_height) @trusted nothrow
    {
        PreImageInfo[] empty;
        try
        {
            auto results = this.db.execute("SELECT key, preimage, distance " ~
                "FROM validator_set WHERE enrolled_height >= ? AND " ~
                "enrolled_height <= ?",
                start_height, end_height);

            return results.fold!((preimages, row) =>
                preimages ~= PreImageInfo(
                    Hash(row.peek!(char[])(0)),
                    Hash(row.peek!(char[])(1)),
                    row.peek!ushort(2)))(empty);
        }
        catch (Exception ex)
        {
            log.error("Exception occured on getPreimages: {}, heights " ~
                "[{}..{}]", ex.msg, start_height, end_height);
        }

        return empty;
    }
```

hewison-chris (commented):
You have a typo in the descritpion of the first commit. (catch-up process)

source/agora/consensus/state/ValidatorSet.d
linked0: Good suggestion, thanks.

linked0:
> You have a typo in the descritpion of the first commit. (catch-up process)

Fixed.

source/agora/test/RestoreSlashingInfo.d
```diff
@@ -0,0 +1,93 @@
...
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto nodes = network.clients;
+    auto set_a = network.clients[0 .. GenesisValidators];
+    auto set_b = network.clients[GenesisValidators .. $];
+
+    // generate 18 blocks, 2 short of the enrollments expiring.
+    network.generateBlocks(Height(GenesisValidatorCycle - 2));
+    network.expectBlock(Height(GenesisValidatorCycle - 2));
```
hewison-chris: This `expectBlock` is redundant as `generateBlocks` is already waiting for the blocks in `addBlock`. (In some tests we call expectBlock with a different range of clients)

source/agora/test/RestoreSlashingInfo.d
```diff
@@ -0,0 +1,93 @@
...
+
+    // Block 19 we add the freeze utxos for set_b validators
+    // prepare frozen outputs for outsider validators to enroll
+    blocks[0].spendable().drop(1)
+        .map!(txb => txb
+            .split(keys).sign(TxType.Freeze))
+        .each!(tx => set_a[0].putTransaction(tx));
+
+    network.generateBlocks(Height(GenesisValidatorCycle - 1));
+    network.expectBlock(Height(GenesisValidatorCycle - 1));
```
hewison-chris: Same here. The clients are defaulting to the same in both calls.

source/agora/test/RestoreSlashingInfo.d
```diff
@@ -0,0 +1,93 @@
...
+    network.generateBlocks(Height(GenesisValidatorCycle));
+    network.expectBlock(Height(GenesisValidatorCycle));
+
+    // Sanity check
+    auto b20 = set_a[0].getBlocksFrom(GenesisValidatorCycle, 1)[0];
+    assert(b20.header.enrollments.length == conf.outsider_validators);
+
+    // Block 21
+    network.generateBlocks(Height(GenesisValidatorCycle + 1));
+    network.expectBlock(Height(GenesisValidatorCycle + 1));
```
hewison-chris: Also

source/agora/test/RestoreSlashingInfo.d
linked0: I missed it. Will remove the unnecessary thing.

source/agora/test/RestoreSlashingInfo.d
```diff
@@ -0,0 +1,93 @@
...
+        retryFor(node.getBlockHeight() == GenesisValidatorCycle - 1, 2.seconds,
+            format!"Expected block height %s but outsider %s has height %s."
+                (GenesisValidatorCycle - 1, idx, node.getBlockHeight())));
+
+    // Now we enroll the set B validators.
+    set_b.enumerate.each!((idx, _) => network.enroll(GenesisValidators + idx));
+
+    // Block 20, After this the Genesis block enrolled validators will be expired.
+    network.generateBlocks(Height(GenesisValidatorCycle));
+    network.expectBlock(Height(GenesisValidatorCycle));
```
linked0: It's interesting that removing this one makes the test fail.

hewison-chris: Probably due to delay caused by this check. So maybe the problem is in the test code after this.

hewison-chris: I think I see what it is. Now the default clients is no longer correct. You need to pass all the clients to the generateBlocks now there are more validators.

linked0: Thanks for the explanation.

source/agora/consensus/state/ValidatorSet.d
linked0: The code you suggested outputs some error like this.

> Error: no property fold for type d2sqlite3.results.ResultRange

I have to learn more about the`fold`. It's not bad to leave the code as it is but I will try to use the `fold` in a similar context like this one. Thanks for introducing the new concept.

source/agora/test/RestoreSlashingInfo.d
linked0: Updated.

source/agora/test/RestoreSlashingInfo.d
linked0: Updated.

source/agora/test/RestoreSlashingInfo.d
linked0: Updated.

source/agora/test/RestoreSlashingInfo.d
linked0: Updated. Thanks.

source/agora/consensus/state/ValidatorSet.d
hewison-chris: I think you probably missed the import std.algorithm

hewison-chris: Sorry I forgot to include in the sugestion.

linked0:
Ready for review.

source/agora/consensus/state/ValidatorSet.d
hewison-chris: This does work:-
```
public PreImageInfo[] getPreimages (ulong start_height,
        ulong end_height) @trusted nothrow
    {
        import std.algorithm;
        PreImageInfo[] empty;
        try
        {
            auto results = this.db.execute("SELECT key, preimage, distance " ~
                "FROM validator_set WHERE enrolled_height >= ? AND " ~
                "enrolled_height <= ?",
                start_height, end_height);

            return results.fold!((preimages, row) =>
                preimages ~= PreImageInfo(
                    Hash(row.peek!(char[])(0)),
                    Hash(row.peek!(char[])(1)),
                    row.peek!ushort(2)))(empty);
        }
        catch (Exception ex)
        {
            log.error("Exception occured on getPreimages: {}, heights " ~
                "[{}..{}]", ex.msg, start_height, end_height);
        }

        return empty;
    }
```

hewison-chris: No need to change now but just for reference.

linked0: Right. I have already succeeded with your information yesterday. :) But, I decided to use the original code because using `foreach` is easier to understand, I think. I will apply the `fold` in a more adequate situation or context. Thanks.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -773,6 +773,25 @@ public class EnrollmentManager
...
+        Params:
+            preimage = the pre-images information to add
+
+    ***************************************************************************/
+
+    public void addPreimages (const PreImageInfo[] preimages) @safe nothrow
+    {
+        foreach (image; preimages)
+        {
+            this.validator_set.addPreimage(image);
```
AndrejMitrovic: You're ignoring the return value of `addPreimage`. What if a node sends you the wrong preimages that don't validate? Then you should exit early and try getting them from another node.

Specifically another node could send you a giant array of `PreImageInfo`. It would be good to return early if any of the preimages are invalid.

source/agora/consensus/SlashPolicy.d
```diff
@@ -173,6 +174,12 @@ public class SlashPolicy
                 valid_keys ~= key;
         }
 
+        version (unittest)
+        {
+            if (valid_keys.length == 0)
+                return getTestRandomSeed();
```
AndrejMitrovic: Why is this necessary?

source/agora/consensus/EnrollmentManager.d
linked0: Right, it would be better.
Will fix it.

source/agora/consensus/SlashPolicy.d
linked0: There are so many cases in unit tests where the pre-images needed don't exist. It's because the tests don't in fact need to focus on the existence of pre-images. So I decided to just add codes that work around errors due to missing pre-images because it will take so much time to provide proper preimages in the unit tests.

AndrejMitrovic: Isn't that what [waitForPreimages](https://github.com/bpfkorea/agora/blob/c7b78c6092a03efe42598348a2cb9b10cdb9e66b/source/agora/test/Base.d#L677) and [expectBlock](https://github.com/bpfkorea/agora/blob/c7b78c6092a03efe42598348a2cb9b10cdb9e66b/source/agora/test/Base.d#L657) which calls it is for?

linked0: Yes, right. But I mean unit tests, not network tests. The functions are not used in unit tests.

AndrejMitrovic: I see. The issue I see is that this code will also be used by the networking tests.

Maybe a better way is to provide a custom SlashPolicy in the non-networking tests, which overrides `getRandomSeed`.

linked0: I think it's better, thanks. And I agree that the code could affect the networking test.

AndrejMitrovic (approved):
Overall LGTM. Are the test failures just random?

linked0:
> Overall LGTM. Are the test failures just random?

Yes, right. It randomly happens.

linked0:
All green!

AndrejMitrovic:
LGTM. Anyone else?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -736,6 +736,26 @@ public class EnrollmentManager
...
+            start_height = the starting enrolled height to begin retrieval from
+            end_height = the end enrolled height to finish retrieval to
+
+        Returns:
+            preimages' information of the validators
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getValidatorPreimages (ulong start_height,
+        ulong end_height) @trusted nothrow
```
Geod24: Why is it `@trusted` ?

source/agora/consensus/state/ValidatorSet.d
```diff
@@ -579,6 +579,48 @@ public class ValidatorSet
...
+            start_height = the starting enrolled height to begin retrieval from
+            end_height = the end enrolled height to finish retrieval to
+
+        Returns:
+            preimages' information of the validators
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getPreimages (ulong start_height,
+        ulong end_height) @trusted nothrow
```
Geod24: Using `Height` type here

source/agora/network/NetworkManager.d
```diff
@@ -835,8 +862,9 @@ public class NetworkManager
 
     ***************************************************************************/
 
-    private void getBlocksFrom (Height block_height,
-        scope bool delegate(const(Block)[]) @safe onReceivedBlocks) nothrow
+    private void getBlocksFrom (Ledger ledger, Height block_height,
+        scope bool delegate(Ledger ledger, const(Block)[], const(PreImageInfo)[]) @safe onReceivedBlocks)
+        nothrow
```
Geod24: I think this mixes things up quite a bit. How come we can receive a valid block but not know the pre-images ?

source/agora/utils/Test.d
```diff
@@ -630,6 +630,12 @@ public auto genesisSpendable () @safe pure nothrow
     return GenesisBlock.spendable;
 }
 
+/// Get random seed for making new block
+public Hash getTestRandomSeed () @safe nothrow
+{
+    return hashFull(Hash.init);
+}
+
```
Geod24, line 665: What is the "random seed" supposed to be ?

source/agora/network/NetworkManager.d
linked0: > I think this mixes things up quite a bit. How come we can receive a valid block but not know the pre-images?

Before making a new block, we should know the pre-images as you know. So the following code will be running when the case you are saying happens which means getting blocks and pre-images from the next node. 

```
// one or more blocks were rejected, stop retrieval from node
if (!onReceivedBlocks(ledger, blocks, preimages))
    continue LNextNode;
```

I thought that there's a very low possibility that pre-images are not correct although blocks are valid.

In fact, there's one more thing we should think about it. What if there's no node from which we can get pre-images? We can imagine that there are nodes that only have the ledger and no normal database. Don't we need to store the last pre-images of un-enrolled validators into the ledger, including slashed validators? I know you don't want the ledger to be bloated due to that. I will file a new issue about that if needed.

source/agora/consensus/EnrollmentManager.d
linked0: Ah, I think it should be `@safe`.

source/agora/consensus/state/ValidatorSet.d
linked0: Will update.

source/agora/utils/Test.d
linked0: It is supposed to be the return value of `getRandomSeed` of the `Ledger`. 

```
public Hash getRandomSeed() nothrow @safe
{
    return this.slash_man.getRandomSeed(this.last_block.header.height);
}
```

There is no pre-image that is revealed by gossiping in unit tests so the `getTestRandomSeed` is needed.

source/agora/network/NetworkManager.d
linked0: > I think this mixes things up quite a bit. How come we can receive a valid block but not know the pre-images?

Anyway, we need to mix blocks and pre-images up in this function because making a new block needs proper pre-images of the current height. The fact that a block received from some node is valid means the pre-images from the node also valid because the blocks were created with the right pre-images. If there is some invalid information, the catch-up process can restart getting valid information from other nodes at the point where the failure occurs. If you have any other suggestions about that, please let me know or we can make things right with another new issue.

source/agora/api/FullNode.d
```diff
@@ -213,6 +213,24 @@ public interface API
...
+            start_height = the starting enrolled height to begin retrieval from
+            end_height = the end enrolled height to finish retrieval to
+
+        Returns:
+            preimages' information of the validators
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getPreimages (ulong start_height, ulong end_height);
+
```
Geod24, line 233: Why do we need this ? `getPreimage` should be enough, shouldn't it ?

source/agora/network/NetworkManager.d
```diff
@@ -850,6 +855,19 @@ public class NetworkManager
 
             do
             {
+                auto start_height =
+                    (block_height <= 1) ?  0 : block_height;
+                auto end_height = start_height +
+                    (block_height <= 1) ?  this.consensus_config.validator_cycle :
+                        this.consensus_config.validator_cycle - 1;
```
Geod24: Why is `1` a special value here ?

source/agora/network/NetworkClient.d
```diff
@@ -370,6 +370,29 @@ class NetworkClient
...
+        Returns:
+            the array of preimages of validators enrolling from `enrolled_height`
+            to `end_height`
+
+            If the request failed, returns an empty array
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getPreimages (ulong start_height,
+        ulong end_height) @trusted nothrow
```
Geod24: Why `@trusted` ?

source/agora/node/FullNode.d
```diff
@@ -300,16 +300,40 @@ public class FullNode : API
...
+        Params:
+            blocks = the blocks to be added
+            preimages = the preimages needed to check the validity of the blocks
+
+        Returns:
+            true if the blocks and preimages was added
+
+    ***************************************************************************/
+
+    bool addBlocks(const(Block)[] blocks, const(PreImageInfo)[] preimages) @safe
```
Geod24: Missing space before parenthesis

source/agora/node/FullNode.d
```diff
@@ -300,16 +300,40 @@ public class FullNode : API
...
+    {
+        foreach (block; blocks)
+        {
+            if(!this.ledger.addPreimages(preimages))
```
Geod24: Instead of adding a function in the ledger whose full role is to forward to the enrollment manager, call the enrollment manager directly and remove the Ledger function

source/agora/consensus/SlashPolicy.d
```diff
@@ -29,6 +29,7 @@ module agora.consensus.SlashPolicy;
 
 import agora.common.Amount;
 import agora.common.crypto.Key;
+import agora.common.Hash;
```
Geod24: Why the extra import ?

source/agora/consensus/validation/Block.d
```diff
@@ -179,6 +181,12 @@ public string isInvalidReason (const ref Block block, Height prev_height,
             return fail_reason;
     }
 
+    if (block.header.random_seed == Hash.init
+        || block.header.random_seed != random_seed)
+    {
+        return "Block: Random seed is wrong";
```
Geod24: "Block: Header's random seed does not match that of known pre-images"

source/agora/consensus/validation/Block.d
```diff
@@ -639,14 +647,17 @@ version (unittest)
...
         size_t active_enrollments, size_t enrolled_validators, scope FeeChecker checkFee,
-        scope EnrollmentFinder findEnrollment, ulong enrollment_cycle = 0,
-        ulong prev_timestamp = 0, ulong curr_timestamp = ulong.max,
+        scope EnrollmentFinder findEnrollment, Hash random_seed = Hash.init,
+        ulong enrollment_cycle = 0, ulong prev_timestamp = 0, ulong curr_timestamp = ulong.max,
         Duration block_timestamp_tolerance_dur = 100.seconds,
         string file = __FILE__, size_t line = __LINE__) nothrow @safe
     {
+        if (random_seed == Hash.init)
+            random_seed = getTestRandomSeed();
```
Geod24, line 657: Why not just give it as default parameter ?

source/agora/utils/Test.d
Geod24: Isn't that a bug though ? That pre-images are not revealed by gossiping ?

source/agora/node/Ledger.d
```diff
@@ -1067,6 +1071,22 @@ public class Ledger
...
+        Returns:
+            the random seed if there are one or more valid preimages,
+            otherwise Hash.init.
+
+    ***************************************************************************/
+
+    public Hash getRandomSeed() nothrow @safe
+    {
+        return this.slash_man.getRandomSeed(this.last_block.header.height);
+    }
```
Geod24, line 1079: This should be in the slashing manager directly and not in the Ledger. The problem is, we shouldn't need to do dependency injection here, because the random seed is derived from pre-images and shouldn't need to be changed. Or am I missing something ?

source/agora/network/NetworkManager.d
linked0: > Why is 1 a special value here ?

I found out that `1` is not needed if I make the formula more correct and simpler.
I changed it. Ty.

source/agora/consensus/validation/Block.d
linked0: > Why not just give it as default parameter ?

It's because there are some cases the `random_seed` is already set before calling the `isValidcheck` function. In many test cases, there are two situations where the `random_seed` is set or not.

source/agora/utils/Test.d
linked0: > Isn't that a bug though ? That pre-images are not revealed by gossiping ?

No, it's not. It's only used in unit tests without any gossiping.

source/agora/node/Ledger.d
linked0: I understand what you mean. But Getting the random seed is needed in unit tests like in `TestLedger` so I've got to make this function overridden which is used as follows.

```
static class ThrowingLedger : Ledger
{
     ///
     public override Hash getRandomSeed() nothrow @safe
     {
         return getTestRandomSeed();
     }
}
```

linked0:
Applied all the suggestions and rebased. Waiting for all green. 
Thanks for the review.

source/agora/consensus/validation/Block.d
Geod24: But if it's set, it can be passed as a parameter ? Maybe you have a link to the code that creates the issue ?

linked0: > But if it's set, it can be passed as a parameter? Maybe you have a link to the code that creates the issue?

I'll make the value be passed as a parameter. I think it's the last thing to do in this PR.

Geod24 (approved):

PR #1474, opened 2021-01-08
Add the absolute block height to the stats

### LINKED ISSUE #1405: Need stats for the current block height
Currently we have stats such as `agora_block_externalized_total`:

https://github.com/bpfkorea/agora/blob/3402716cf39f69b31995ba205730bd189aa761ab/source/agora/node/Ledger.d#L257

However this is not very useful. It's only counting how many new blocks were externalized since the last time the node booted up. So if the node restarts this count will go to zero. It's much more useful to just show the actual current block height. That way if there's any discrepancies between the nodes we'll be able to easily see them (and graph them).

Geod24:
This is more useful than the number of block externalized since start.
Fixes #1405


AndrejMitrovic (approved):

kchulj (commented):
Nice!

PR #1476, opened 2021-01-08
Refactor: Isolate state rebuilding in its own function

Geod24:
```
It is a large enough portion of the constructor to warrant its own private function,
even if this function is only ever called once.
```

Just moving code around.


AndrejMitrovic:
Looks like auto-merge doesn't work unless there is also an approval.

AndrejMitrovic (approved):

Geod24:
I think it's one of the branch requirement ? You can't rebase something without approving it first ?

PR #1483, opened 2021-01-11
Notify Validator of invalid nominations

omerfirmak:
This can be used as a self-test.
The example usage here makes sure all Validators keep track of the active Validator count and enroll if needed, asynchronously to the block creation.

Any thoughts?

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -651,6 +651,13 @@ public class Ledger
         }
     }
 
+    /// Error message describing the reason of validation failure
+    public static enum InvalidConsensusDataReason : string
+    {
+        NoTransactions = "Transaction set doesn't contain any transactions",
+        NotEnoughValidators = "Enrollment: Insufficient number of active validators",
+    }
+
     /***************************************************************************
 
         Check whether the consensus data is valid.
@@ -669,7 +676,7 @@ public class Ledger
         auto utxo_finder = this.utxo_set.getUTXOFinder();
 
         if (!data.tx_set.length)
-            return "Transaction set doesn't contain any transactions";
+            return InvalidConsensusDataReason.NoTransactions;
 
         foreach (const ref tx; data.tx_set)
         {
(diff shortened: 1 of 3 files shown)


source/agora/node/Ledger.d
```diff
@@ -680,7 +680,7 @@ public class Ledger
         size_t active_enrollments = enroll_man.getValidatorCount(expect_height);
 
         if (data.enrolls.length + active_enrollments < Enrollment.MinValidatorCount)
-            return "Enrollment: Insufficient number of active validators";
+            return "Enrollment: Insufficient number of active validators"; // If you change this, change all occurences
```
Geod24: Shouldn't we enter this function with some valid state instead of patching it here ?

omerfirmak: If you mean that we should check for total enrollments for the next block earlier than this, we could. But this function is also used to validate the incoming `ConsensusData`, so this check should still be there probably.

source/agora/node/Validator.d
```diff
@@ -508,4 +510,22 @@ public class Validator : FullNode, API
...
+        Params:
+            data = Invalid ConsensusData
+            msg = Reason
+
+    ***************************************************************************/
+
+    private void invalidNominationHandler (const ref ConsensusData data,
+        const ref string msg) @safe
+    {
+        if (msg == "Enrollment: Insufficient number of active validators")
```
AndrejMitrovic: So if someone accidentally changes the string, then this will no longer work. It's an odd choice. Why not just use an enum?

omerfirmak: That is why there are comments about being careful about changing the literals :smile: (Ledger.d:683)
Well, changing the return type would probably mean changing the entire `isInvalidReason` family of functions and refactor points they were called. I am not entirely sure if this is worth it.

But I fully agree this puts the weight entirely on our future selfs :smile:

omerfirmak: `validateConsensusData` now returns an enum. :+1:

AndrejMitrovic:
> This can be used as a self-test.

I would at least change the last commit message. It's not just notifying the validator, the validator tries to re-enroll too. That should be noted.

Geod24 (approved):
LGTM once @AndrejMitrovic 's comment is addressed.

omerfirmak:
Thanks for the reviews.
I have updated the commit message and added a comment above the new `checkAndEnroll` call.

PR #1484, opened 2021-01-11
Storing the build information in the binary

### LINKED ISSUE #1472: Store the build information in the binary
For monitoring purpose, we would like to have the version number inside Agora.
We have support for it in the stats, but we [currently just write "HEAD"] (https://github.com/bpfkorea/agora/blob/8f6bc417766428097703dafbc942a7443380d1f7/source/agora/node/FullNode.d#L211).

Suggested course of action:
- Write a little D script to do it that is cross-platform;
- Use the script in `preGenerateCommands`;
- The script can either write:
    - A `.d` file [like this](https://github.com/dlang/dub/blob/0c7a158f1914e045e246ebb9adc4b742d05eea74/build.d#L100-L154);
    - A text file with just the info which is read using [`import` expression](https://dlang.org/spec/expression.html#import_expressions) (**not** import declarations);
- The script should avoid rebuilding the app if the version hasn't changed;
- The script should have an environment variable it can read the version from (e.g. `AGORA_VERSION`);
- The script should otherwise query `git describe` for the version;

I recommend the `import` expression approach, as it will allow one to read the file and avoid rewriting the version file if it hasn't changed.

Few notes:
- This script will be used by developers when building;
- This script might also be run by people downloading the sources from Github (which is why the environment variable is important);
- The output of this script needs to be `.gitignore` / `.dockerignore`d;
- If the output is a simple file, it needs to be in a `stringImportPaths` or whatever dub calls it (see `dmd -J`);

ferencdg:
fixes: #1472

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -155,7 +155,7 @@ jobs:
 
       - name: Build documentation
         run: |
-          dub build -b ddox
+          AGORA_VERSION=HEAD dub build -b ddox
           # Filter out libraries
           # `x:ddocFilterArgs` in `dub.json` does not seem to work,
           # so we do it manually
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,11 @@
 # Build Agora from source
 FROM bpfk/agora-builder:latest AS Builder
 ARG DUB_OPTIONS
+ARG AGORA_VERSION="HEAD"
 ADD . /root/agora/
 WORKDIR /root/agora/
 RUN apk --no-cache add llvm-libunwind-dev
-RUN dub build --skip-registry=all --compiler=ldc2 ${DUB_OPTIONS}
+RUN AGORA_VERSION=${AGORA_VERSION} dub build --skip-registry=all --compiler=ldc2 ${DUB_OPTIONS}
 
 # Runner
 FROM alpine:3.12.0
--- a/source/agora/cli/version/main.d
+++ b/source/agora/cli/version/main.d
@@ -1,0 +1,105 @@
+#!/usr/bin/env dub
+/+
+ dub.json:
+ {
+     "name": "version_build"
+ }
+ +/
+/*******************************************************************************
+
+    Determine the version number that will be used during building the Agora node.
+
+    Version number is determined in the following order:
+    1. The 'AGORA_VERSION' environment variable will be used, if it is set to
+       a non-empty string.
+    2. Otherwise, the output of the `git describe` command will be used.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.cli.ver.main;
+
+import std.file : exists, mkdirRecurse, readText, write;
+import std.format;
+import std.path;
+import std.process;
+import std.stdio : writeln;
+import std.string;
+
+import core.stdc.stdlib : exit;
+
+private immutable VersionFileName = "VERSION";
+private immutable RootPath = __FILE_FULL_PATH__.dirName.dirName.dirName.dirName.dirName;
+private immutable VersionFileDir = RootPath.buildPath("build");
+private immutable VersionFilePath = VersionFileDir.buildPath(VersionFileName);
+private immutable EnvVersionName = "AGORA_VERSION";
+private immutable GitDescribe = ["git", "describe"];
+
+/// Utility function to run a command and throw on error if the command fails
+private string runCmd (const string[] cmd)
+{
+    writeln("Executing command: ", cmd);
+    auto proc_res = execute(cmd);
+    if (proc_res.status != 0)
+        throw new Exception(format("Command failed: %s", cmd));
+    return proc_res.output;
+}
+
+/// Determine the version number
+private void writeVersion ()
+{
+    string version_res;
+
+    // get the version from environment variables
+    if (auto from_env = environment.get(EnvVersionName))
+    {
+        from_env = from_env.strip();
+        if (!from_env.length)
+            throw new Exception(format("Environment variable %s is empty, please remove it or set to non-empty value",
+                                EnvVersionName));
+        version_res = from_env;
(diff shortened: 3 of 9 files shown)


Dockerfile
```diff
@@ -1,10 +1,11 @@
 # Build Agora from source
 FROM bpfk/agora-builder:latest AS Builder
 ARG DUB_OPTIONS
+ARG AGORA_VERSION="HEAD"
```
ferencdg, line 4: this is deliberately not an ENV as I would not like to preserve this environment variable after the docker build finishes

.dockerignore
```diff
@@ -7,11 +7,12 @@ tests/
 LICENSE
 README.md
 
-# Various build folders
+# Various build folders and other generated files
 .dub/
 build/
 docs/
 source/scpp/build/
+agora.version
```
Geod24: Just output the file in `build`

source/agora/cli/version/main.d
```diff
@@ -0,0 +1,105 @@
...
+import std.file : write, readText, exists;
+import std.format;
+import std.path;
+import std.process;
+import std.stdio : writeln;
+import std.string;
+
+import core.stdc.stdlib : exit;
+
+private immutable VersionFileName = "agora.version";
```
Geod24: I'd prefer something like `VERSION` as output file.

source/agora/cli/version/main.d
```diff
@@ -0,0 +1,105 @@
...
+    // get the version from environment variables
+    auto version_from_env = environment.get(EnvVersionName);
+    if (version_from_env !is null)
+    {
+        version_from_env = version_from_env.strip();
+        if (version_from_env.length)
+            version_res = version_from_env;
+        else
+            throw new Exception(format("Environment variable %s is empty, please remove it or set to non-empty value",
+            EnvVersionName));
```
Geod24:
```suggestion
        if (!version_from_env.length)
            throw new Exception(format("Environment variable %s is empty, please remove it or set to non-empty value",
            EnvVersionName));
            version_res = version_from_env;
```

Get error handling out of the way early for better flow

ferencdg:
should be fixed by now

.github/workflows/main.yml
```diff
@@ -155,7 +155,8 @@ jobs:
 
       - name: Build documentation
         run: |
-          dub build -b ddox
+          mkdir build
```
Geod24: Ah I see the issue. You can just add a `build/.keep` file then, to make sure the directly is always present.

source/agora/cli/version/main.d
```diff
@@ -0,0 +1,104 @@
...
+
+    Determine the version number that will be used during building the Agora node.
+
+    Version number is determined in the following order:
+    1. The 'AGORA_VERSION' environment variable will be used, if it is set to
+       a non-empty string.
+    2. Otherwise, the output of the `git describe` command will be used.
+
+    Copyright:
+        Copyright (c) 2019 BOS Platform Foundation Korea
```
Geod24, line 18: Bit out of date don't you think ? ;)

source/agora/cli/version/main.d
```diff
@@ -0,0 +1,104 @@
...
+{
+    writeln("Executing command: ", cmd);
+    auto proc_res = execute(cmd);
+    if (proc_res.status != 0)
+        throw new Exception(format("Command failed: %s", cmd));
+    return proc_res.output;
+}
+
+/// Determine the version number
+private string getVersion()
```
Geod24:
```suggestion
private string getVersion ()
```

source/agora/cli/version/main.d
```diff
@@ -0,0 +1,104 @@
...
+        version_from_file = VersionFilePath.readText().strip();
+
+    // if the version from the file is the same as the version deduced, then exit
+    if (version_from_file == version_res)
+        exit(0);
+
+    write(VersionFilePath, version_res);
+
+    return version_res;
+
```
Geod24:
```suggestion
```

source/agora/cli/version/main.d
```diff
@@ -0,0 +1,104 @@
...
+}
+
+///
+public void main(string[] args)
+{
+    try
+    {
+        getVersion();
```
Geod24: Since the return value is never used, and this really just writes the version, perhaps rename it, change the return value, and use `return` instead of `exit` in the above function ?

source/agora/cli/version/main.d
```diff
@@ -0,0 +1,104 @@
...
+        exit(0);
+
+    write(VersionFilePath, version_res);
+
+    return version_res;
+
+}
+
+///
+public void main(string[] args)
```
Geod24, line 91:
```suggestion
public void main (string[] args)
```

source/agora/cli/version/main.d
```diff
@@ -0,0 +1,104 @@
...
+        throw new Exception("Unable to determine agora version number");
+
+    // get the version from the version file
+    string version_from_file;
+    if (VersionFilePath.exists)
+        version_from_file = VersionFilePath.readText().strip();
+
+    // if the version from the file is the same as the version deduced, then exit
+    if (version_from_file == version_res)
+        exit(0);
```
Geod24: Thanks for adding comments. It would be nice to have an explanation of why instead of what.
For example, the comment about "git [repository] with no label" at L71 is good at doing this.
In this one, you might want to comment on why we don't want rebuild, rather than what the code is doing.

Also, always prefer `return` over `exit` when possible.

source/agora/node/FullNode.d
```diff
@@ -206,9 +206,11 @@ public class FullNode : API
 
         Utils.getCollectorRegistry().addCollector(&this.collectAppStats);
         Utils.getCollectorRegistry().addCollector(&this.collectStats);
+        enum build_version = import(VersionFileName);
+        pragma(msg, "Compiling agora version: " ~ build_version);
```
Geod24: Remove

Geod24 (approved):
Nice, LGTM

Geod24 (approved):
Did the fixes

PR #1490, opened 2021-01-12
Config: move preimage_reveal_interval to Validator config

Geod24:
Because this number is only relevant for validators,
as they are the only ones revealing pre-images.


linked0 (approved):
LGTM

kchulj (approved):
LGTM

Issue #1228, opened 2020-09-21
Replace transaction set and enrollments in ConsensusData with hashes and/or a merkle root


AndrejMitrovic:
This is the counter-issue to https://github.com/bpfkorea/agora/issues/993. #993 should likely be closed and this issue should be worked on instead. 

Voting on entire transaction sets is not scalable. This will become more apparent when:

- There are many transactions in the pool to nominate
- The block size limit is increased (imagine voting on 4MB transaction sets, it could take minutes to transfer this data)
- There are dozens, hundreds, or even thousands of validator nodes

As a first step we can do the following:
- Add an API end-point for a node to be able to retrieve a set of transactions from another node's transaction pool. What should the API look like? If we design it as `getTransaction (Hash tx)` then this may cause many requests to the node. It could be `getTransactions (Set!Hash txs)`. There might be other alternatives.
- Replace ConsensusData internal fields to be hashes of tx'es and hashes of enrollments.

Then we must also fix the nominator:

- Carefully inspect the `Nominator` validation routine. SCP supports returning `kMaybeValidValue` in the [validateValue](https://github.com/bpfkorea/agora/blob/bbf523199b6efd1667ee0e3d79ad59d4c3ea93c2/source/agora/consensus/protocol/Nominator.d#L539) SCP routine. See [this comment](https://github.com/bpfkorea/agora/blob/406a22df4c33dfd2f2e3a61811fe6e73aea94852/source/scpp/src/scp/SCPDriver.h#L56-L58). One possible approach is: Find as many transactions from the `ConsensusData` transaction hash set inside of the transaction pool. If all transactions are found and all are valid => return `kValidValue`. If some transactions were not found, but they all still valid, return `kMaybeValidValue`. If **any** found transaction is invalid => return `kInvalidValue`.

-----

As a final step we should consider replacing the hash set of transactions with a merkle root. It's possible to skip the intermediate development step of changing `transaction/enrollment set => hash set => merkle root` and go directly to the idea of changing `transaction/enrollment set => merkle root`, but I'm afraid this might become a multi-sprint issue. Doing this piece-meal might be better.

omerfirmak:
> If some transactions were not found, but they all still valid, return kMaybeValidValue

As far as I can see, stellar-core never does this. It does not make any decision about an envelope until it has all the necessary information.  They also vote on tx set hashes, not individual tx hashes.

[...]

PR #1491, opened 2021-01-13
Nominator: Allow multiple nominations per round

omerfirmak:
SCP rules allows nodes to nominate multiple values.
Only constraint is that once a node confirms a nomination, it should
stop nominating new values.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -109,6 +109,9 @@ public extern (C++) class Nominator : SCPDriver
     /// Whether we're in the asynchronous stage of nominating
     private bool is_nominating;
 
+    /// Last height that we finished the nomination round
+    private Height last_confirmed_height;
+
     /// Periodic nomination timer. It runs every second and checks the clock
     /// time to see if it's time to start nominating. We do not use the
     /// `BlockInterval` interval directly because this makes the timer
@@ -351,8 +354,9 @@ extern(D):
 
     private void checkNominate () @safe
     {
-        // already in the asynchronous stage of nominating
-        if (this.is_nominating)
+        const slot_idx = this.ledger.getBlockHeight() + 1;
+        // are we done nominating this round
+        if (this.last_confirmed_height >= slot_idx)
             return;
 
         const cur_time = this.clock.networkTime();
@@ -373,11 +377,9 @@ extern(D):
         if (!this.prepareNominatingSet(data))
             return;
 
-        log.info("Nomination started at {}", cur_time);
-        log.trace("Consensus data is: {}", data);
+        log.trace("Nominating {} at {}", data, cur_time);
         this.is_nominating = true;
 
-        const slot_idx = this.ledger.getBlockHeight() + 1;
         // note: we are not passing the previous tx set as we don't really
         // need it at this point (might later be necessary for chain upgrades)
         this.nominate(slot_idx, data);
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
>  it should stop nominating new values

Technically the node can choose to do whatever it wants, but other (non-faulty/non-byzantine) nodes will ignore the new values if they know that node has already confirmed something.

Anyway just trying to clear it up for everyone.

ferencdg:
I wonder what is the background of this change. At the end of the 10 mins interval when the nomination begins, the set of transactions will be fairly stable. Once we called the 'nominate' method of the SCP, then there is an internal timer inside SCP that kicks in, and tries to renominate the same set of transactions(and other node's nominations too) that were passed into the 'nominate' method for that round.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -817,6 +814,13 @@ extern(D):
                 "Error: {}", ex.msg);
             return ValidationLevel.kInvalidValue;
         }
+
+        // Per SCP rules, once we CONFIRM a NOMINATE; we can't
+        // nominate new values. Keep track of the biggest slot_idx we confirmed
+        // a NOMINATE on
+        if (slot_idx > this.done_nominating)
```
AndrejMitrovic: I don't think this is the right place to do this. It assumes the node will confirm something as soon as its validated and this function returns.

Isn't there a confirm event handler virtual method somewhere in SCP? Or something similar to that.

AndrejMitrovic: Maybe something from here?

https://github.com/stellar/stellar-core/blob/54b03f755ae5d5aa12a799c8f1ee4d87fc9d1a1d/src/scp/SCPDriver.h#L205-L240

omerfirmak: Other than `updatedCandidateValue` they look unrelated. It's called a bit further into the process, seems like a better place to do this. 

I moved it there.

omerfirmak:
> I wonder what is the background of this change. At the end of the 10 mins interval when the nomination begins, the set of transactions will be fairly stable. Once we called the 'nominate' method of the SCP, then there is an internal timer inside SCP that kicks in, and tries to renominate the same set of transactions(and other node's nominations too) that were passed into the 'nominate' method for that round.

Flexibility mostly, we should not tie our own hands. IMO, we should take advantage of the SCP as much as we can.
If I am not mistaken, nomination phase can take arbitrarily long. So letting nodes nominate multiple values could help.

source/agora/consensus/protocol/Nominator.d
omerfirmak: `emitEnvelope` would probably be even better. We could check for an outgoing CONFIRM envelope. What do you think?

omerfirmak: We can squash the last commit once we are happy with it.

linked0:
Please make a note of the original issue from which this PR comes, like "Part of #1228" in the description of this PR. It can make us find the original issue.

omerfirmak:
> Please make a note of the original issue from which this PR comes, like "Part of #1228" in the description of this PR. It can make us find the original issue.

Well, this is not necessarily part of any existing issue tho. It is more like an overall improvement.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -183,7 +183,6 @@ extern(D):
     public void setQuorumConfig (const ref QuorumConfig quorum,
         const(QuorumConfig)[] other_quorums) nothrow @safe
     {
-        assert(!this.is_nominating);
```
AndrejMitrovic: This was a useful assert to catch bugs.

Even though SCP by design allows quorums to change mid-nomination, in our design the quorums should be fixed within every new block height nomination.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -106,8 +106,8 @@ public extern (C++) class Nominator : SCPDriver
     /// Currently active timers grouped by type
     private ITimer[TimerType.max + 1] active_timers;
 
-    /// Whether we're in the asynchronous stage of nominating
-    private bool is_nominating;
+    /// Last height that we finished the nomination round
+    private Height done_nominating;
```
AndrejMitrovic: `done` usually implies a boolean. How about `last_confirmed_height`?

source/agora/consensus/protocol/Nominator.d
omerfirmak: I have added `is_nominating` back for this.

AndrejMitrovic:
Looks good now, just please squash.

omerfirmak:
> Looks good now, just please squash.

Done

AndrejMitrovic (approved):

PR #1492, opened 2021-01-14
tests: add option to skip printing logs on failure

AndrejMitrovic:
A common workflow I use is this:

- Do my work
- Keep rebasing, splitting and moving code
- Ensure each commit is atomic and passes the test-suite

The last part is more difficult than it should be because of the excessive logs we have when tests fail. When I'm trying to make sure each commit is atomic I run the entire test-suite with the `dchatty=1` flag. This shows me the list of files that fail. 

But the excessive node logs of failing tests drown out that list of files that were tested so they become hard to find in the giant screen log in the terminal. The ideal approach would be:

- Disable node logs (implemented in this PR) and run with `dchatty=1`
- After the tests finish collect all the failing modules
- For each failing module run the tests individually with `dtest=agora.test.FailingModule` to get the full log and fix each test accordingly.

That's where `dnologs=1` in this PR helps with.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -137,6 +137,9 @@ void testAssertHandler (string file, ulong line, string msg) nothrow
     throw new AssertError(msg, file, line);
 }
 
+/// Skip printing out per-node logs ony agora/test/* failures
+shared bool no_logs = false;
+
 /// Custom unnitest runner as a workaround for multi-threading issue:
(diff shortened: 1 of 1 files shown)


hewison-chris:
I usualy find "-- ST" in the output to skip past the logs, but I think this option for no logs is a good idea still.

hewison-chris (approved):

omerfirmak (approved):
@AndrejMitrovic doing god's work!

PR #1494, opened 2021-01-14
Implement inbound connection banning support (AGORA side)

### LINKED ISSUE #202: Implement inbound connection banning support in vibe.d
As a Developer, I need to prevent one form of DDoS attacks from other nodes, we need support for banning inbound connections from specific addresses.

**Definition of Done:**
- [x] When a request reaches the server, it can be made to be rejected before any significant processing / memory allocation happens
- [x] The functionality supports dynamically banning
- [x] The functionality supports dynamically unbanning
- [x] The functionality does not add much overhead on the request processing

ferencdg:
fixes: #202

### DIFF
--- a/source/agora/common/BanManager.d
+++ b/source/agora/common/BanManager.d
@@ -17,10 +17,12 @@ module agora.common.BanManager;
 import agora.common.Serializer;
 import agora.common.Types;
 import agora.network.Clock;
+import agora.utils.InetUtils;
 import agora.utils.Log;
 
 import std.file;
 import std.path;
+import std.socket : getAddressInfo, AddressFamily;
 import std.stdio;
 
 import core.stdc.stdlib;
@@ -174,6 +176,23 @@ public class BanManager
 
         if (status.fail_count >= this.config.max_failed_requests)
         {
+            try
+            {
+                immutable host_and_port = InetUtils.extractHostAndPort(address);
+                if (host_and_port.type == HostType.Domain)
+                    foreach (ref addr_info; getAddressInfo(host_and_port.host))
+                    {
+                       auto resolved_address = addr_info.address.toAddrString();
+                       if (addr_info.address.addressFamily == AddressFamily.INET6)
+                           resolved_address = InetUtils.expandIPv6(resolved_address);
+
+                       this.ban(resolved_address);
+                    }
+            }
+            catch (Exception e)
+            {
+                log.trace("Error happened while trying to resolve DNS name {}", address);
+            }
             status.fail_count = 0;
             this.ban(address);
         }
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -408,6 +408,13 @@ public class NetworkManager
         }
     }
 
+    /// Returns an already instantiated version of the BanManager
+    /// (please also see `NetworkManager.getBanMananger()`)
+    public BanManager getAlreadyCreatedBanManager () @safe @nogc nothrow pure
+    {
+        return this.banman;
+    }
+
     /// Called after a node's handshake is complete
     private void onHandshakeComplete (scope ref NodeConnInfo node)
     {
--- a/source/agora/utils/InetUtils.d
+++ b/source/agora/utils/InetUtils.d
@@ -15,15 +15,31 @@ module agora.utils.InetUtils;
 
 import agora.utils.Log;
 
+import vibe.inet.url : URL;
+
 import std.algorithm;
 import std.algorithm.searching;
 import std.array;
+import std.ascii : isAlpha, isDigit;
 import std.conv;
+import std.range : zip, iota;
+import std.regex : matchFirst, regex;
+import std.string : indexOf, strip;
 import std.socket;
+import std.typecons : tuple, Tuple;
 
 import core.stdc.string;
 
 mixin AddLogger!();
+
+///
+public enum HostType : ubyte
+{
+    IPv4,
+    IPv6,
+    Domain,
+    Invalid,
+}
 
 ///
 mixin template AddGetAllIPs()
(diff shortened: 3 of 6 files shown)


source/agora/utils/InetUtils.d
```diff
@@ -19,12 +19,24 @@ import std.algorithm;
...
 
 mixin AddLogger!();
 
+///
+public enum HostType : ubyte
+{
+    IPv4,
+    IPv6,
+    Domain,
+    Unknown
```
Geod24:
```suggestion
    Unknown,
```

source/agora/utils/InetUtils.d
```diff
@@ -174,6 +186,135 @@ version (Windows)
     }
 }
 
+    /***************************************************************************
+        Converts a compressed IPv6 representation to uncompressed representation
```
Geod24:
```suggestion
    /***************************************************************************
    
        Expand an IPv6 address to its canonical representation
```

source/agora/utils/InetUtils.d
```diff
@@ -174,6 +186,135 @@ version (Windows)
...
+
+        Params:
+            ip = compressed IPv6 address
+
+        Returns:
+            uncompressed representation of IPv6 address
+
+    ***************************************************************************/
+
+    public static string convertCompressedIPv6ToUncompressed (string ip) @safe pure
```
Geod24: `expandIPv6` or something shorter please.

Geod24:
```suggestion
    public static string convertCompressedIPv6ToUncompressed (const(char)[] ip) @safe pure
```

AFAICS

source/agora/utils/InetUtils.d
```diff
@@ -174,6 +186,135 @@ version (Windows)
...
+        auto parts = ip.split(":");
+        const first_empty_ind = parts.countUntil("");
+
+        // filter out all the empty parts, except the first one
+        auto res = zip(iota(parts.length), parts).filter!((tup) => tup[0] == first_empty_ind || !tup[1].empty()).
+                   map!(tup => tup[1]).array();
+
+        // fill in the empty part with the needed zeros
+        return res.map!(part => (part.empty() ? replicate(["0"],8 - res.length + 1) : [part]))
+                    .joiner.array.join(":");
```
Geod24:
```suggestion
                    .joiner(':').array();
```

?

source/agora/utils/InetUtils.d
```diff
@@ -174,6 +186,135 @@ version (Windows)
...
+            http://[1:2::3]/data
+
+        Params:
+          url = the URL from which we would like to extract the host and port
+
+        Returns: `HostPortTup` containing the extracted host, port and host type
+
+    ****************************************************************************/
+
+    public static HostPortTup extractHostAndPort (string url) @safe
```
Geod24, line 258: Couldn't we just use Vibe.d for this ?

Geod24 (commented):
Needs a rebase and adaptation to the new names

source/agora/network/NetworkManager.d
```diff
@@ -408,6 +408,13 @@ public class NetworkManager
         }
     }
 
+    /// Returns an already instantiated version of the BanManager
+    /// (please also see `NetworkManager.getBanMananger()`)
+    public BanManager getAlreadyCreatedBanManager () @safe @nogc nothrow pure
```
AndrejMitrovic, line 413: We might want to rename all the `get*` functions to `make*/create*` functions so we would have:

```
makeBanManager() or createBanManager()
getBanManager()
```

We can leave that for a future PR.

source/agora/utils/InetUtils.d
ferencdg: what I would like here is to flatten embedded maps:
```D
    auto a = [["1"],["2"],["0","0"],["5"]];
    writeln(a.map!(e => e).joiner.array.join(":"));
    // should print out 1:2:0:0:5
```

this line
```D
writeln(a.map!(e => e).joiner(":").array; 
```
returns a compile error

source/agora/utils/InetUtils.d
ferencdg: I might be able to make it work, but it lacks the main thing we need: classifying whether the 'host' is ipv4/ipv6/domain

tests/runner.d
```diff
@@ -34,7 +34,7 @@ private int main (string[] args)
     const lflags = getLflags();
     if (!importPaths.length || !lflags.length)
         return 1;
-    const Args = ["rdmd", "-vcolumns"] ~ args[1 .. $] ~
+    const Args = ["rdmd", "-vcolumns", "-i=stdx"] ~ args[1 .. $] ~
```
ferencdg, line 37: vibe.d(through the URL class) seems to pulled in stdx:allocator, let me know if you have a better idea of handling this

Geod24: This should be returned by `getImportPaths`, but I'm fine if we leave this as is.

source/agora/utils/InetUtils.d
```diff
@@ -174,6 +190,145 @@ version (Windows)
     }
 }
 
+    /***************************************************************************
+        Expand an IPv6 address to its canonical representation
```
Geod24: Space

source/agora/common/BanManager.d
```diff
@@ -174,6 +176,23 @@ public class BanManager
...
+                    foreach (ref addr_info; getAddressInfo(host_and_port.host))
+                    {
+                       auto resolved_address = addr_info.address.toAddrString();
+                       if (addr_info.address.addressFamily == AddressFamily.INET6)
+                           resolved_address = InetUtils.expandIPv6(resolved_address);
+
+                       this.ban(resolved_address);
+                    }
+            }
+            catch(Exception e)
```
Geod24: Space

source/agora/utils/InetUtils.d
```diff
@@ -322,10 +322,6 @@ version (Windows)
 
         assert(extractHostAndPort("seed.bosagora.io/blabla")
             == HostPortTup("seed.bosagora.io", 80, HostType.Domain));
-
-        // Negative tests
-        assert(extractHostAndPort("blab/blabla")
-            == HostPortTup("", 0, HostType.Invalid));
```
Geod24: Why is this removed ?

Geod24 (approved):
LGTM

source/agora/utils/InetUtils.d
ferencdg: for the 'getImportPaths': I was wondering if the reason for missing stdx is that in dub.json the submodule/stdx-allocator is not referenced.

source/agora/utils/InetUtils.d
ferencdg: In the newest version of the implementation "blab/blabla" is valid: the implementation will try to detect if the url has any schema or not, and if not then "http" is assumed. 

I can add more negative tests though, if you like.

Geod24: Well then this part of the diff should be squashed in the previous commit. Because the previous commit introduce a test, and this commit removes that diff. Since this commit does not touch the function that is tested, either the previous commit does not pass tests, or the test shouldn't be removed. From what you're telling me it sounds like the former ?

ferencdg: sorry my bad, fixed all the issues by now

source/agora/utils/InetUtils.d
Geod24: Still missing the newline

PR #1495, opened 2021-01-14
Replace transaction set in ConsensusData with hashes

### LINKED ISSUE #993: Propagate newly received transactions via ConsensusData to the TransactionPool
Currently, when we receive transactions via `ConsensusData` we do not store the transaction in the pool.
It would be wise to do so, and could also allow us to find out some issues (e.g. if we're stuck trying to nominate conflicting values).

Definition of done:
- [ ] When a node with an empty pool receive `ConsensusData` that contains transactions it is not aware of, it stores it in its `TransactionPool`, even if it does not reach consensus;
- [ ] A node does not store transactions which are conflicting;
- [ ] If a transaction conflicts, or was newly added, it is reported in the log (as debug);

omerfirmak:
Part of #1228

Fixes https://github.com/bpfkorea/agora/issues/993

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -57,7 +57,7 @@ import core.stdc.stdint;
 import core.stdc.stdlib : abort;
 import core.stdc.time;
 
-import std.algorithm : max;
+import std.algorithm : max, map;
 import std.conv;
 import std.format;
 import std.path : buildPath;
@@ -533,8 +533,18 @@ extern(D):
             const Block prev_block = blocks.front;
             Hash random_seed = this.ledger.getExternalizedRandomSeed(
                 this.ledger.getBlockHeight(), con_data.missing_validators);
-            const Block proposed_block = makeNewBlock(prev_block, con_data.tx_set,
-                con_data.timestamp, con_data.enrolls, random_seed, con_data.missing_validators);
+
+            Transaction[] received_tx_set;
+            if (auto fail_reason = this.ledger.getValidTXSet(con_data, received_tx_set))
+            {
+                log.info("Missing TXs while checking envelope signature : {}",
+                    prettify(envelope));
+                return; // We dont have all the TXs for this block. Try to catchup
+            }
+
+            const Block proposed_block = makeNewBlock(prev_block,
+                received_tx_set, con_data.timestamp, con_data.enrolls,
+                random_seed, con_data.missing_validators);
             const block_sig = ValidatorBlockSig(Height(envelope.statement.slotIndex),
                 public_key, Scalar(envelope.statement.pledges.confirm_.value_sig));
             if (!this.collectBlockSignature(block_sig, proposed_block))
@@ -716,8 +726,18 @@ extern(D):
         const prev_block = blocks.front;
         Hash random_seed = this.ledger.getExternalizedRandomSeed(
                 this.ledger.getBlockHeight(), con_data.missing_validators);
-        const proposed_block = makeNewBlock(prev_block, con_data.tx_set,
-            con_data.timestamp, con_data.enrolls, random_seed, con_data.missing_validators);
+
+        Transaction[] signed_tx_set;
+        if (auto fail_reason = this.ledger.getValidTXSet(con_data, signed_tx_set))
+        {
+            log.info("Missing TXs while signing confirm ballot {}",
+                prettify(envelope));
+            return;
+        }
+
+        const proposed_block = makeNewBlock(prev_block,
+            signed_tx_set, con_data.timestamp, con_data.enrolls,
+            random_seed, con_data.missing_validators);
 
         Height height = proposed_block.header.height;
         const Sig sig = createBlockSignature(proposed_block);
@@ -862,8 +882,19 @@ extern(D):
 
             Hash random_seed = this.ledger.getExternalizedRandomSeed(
                 this.ledger.getBlockHeight(), data.missing_validators);
-            const block = makeNewBlock(prev_block, data.tx_set, data.timestamp, data.enrolls,
-                random_seed, data.missing_validators);
+
+            Transaction[] externalized_tx_set;
+            if (auto fail_reason = this.ledger.getValidTXSet(data,
+                externalized_tx_set))
+            {
+                log.info("Missing TXs while externalizing at Height {}: {}",
+                    height, prettify(data));
+                return;
+            }
+
+            const block = makeNewBlock(prev_block,
+                externalized_tx_set, data.timestamp, data.enrolls, random_seed,
+                data.missing_validators);
 
             // If we did not sign yet then add signature and gossip to other nodes
             if (this.schnorr_pair.V !in this.slot_sigs[height])
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -37,6 +37,7 @@ import agora.common.Types;
 import agora.common.Metadata;
 import agora.common.Set;
 import agora.common.Task;
+import agora.common.Hash;
 import agora.consensus.data.Transaction;
 import agora.network.Clock;
 import agora.network.NetworkClient;
@@ -860,6 +861,37 @@ public class NetworkManager
         {
             log.error("Couldn't retrieve blocks: {}. Will try again later..",
                 ex.msg);
+        }
+    }
+
+    /***************************************************************************
+
+        Retrieve blocks starting from block_height up to the highest block
+        that's available from the connected nodes.
+
+    ***************************************************************************/
+
+    public void getUnknownTXs (Ledger ledger) @safe nothrow
+    {
+        auto unknown_txs = ledger.getUnknownTXHashes();
+
+        foreach (peer; this.peers[])
+        {
+            if (unknown_txs.length == 0)
+                break;
+
+            foreach (tx; peer.getTransactions(unknown_txs))
+            {
+                try
+                {
+                    ledger.acceptTransaction(tx);
+                    unknown_txs.remove(tx.hashFull());
+                }
+                catch (Exception e)
+                {
+                    log.info("Unknown TX {} threw {}", tx, e.msg);
+                }
+            }
         }
     }
 
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -289,6 +289,7 @@ public class FullNode : API
                     Height(this.ledger.getBlockHeight() + 1),
                     // if any blocks fail validation => short-circuit
                     blocks => blocks.all!(block => this.acceptBlock(block)));
+                this.network.getUnknownTXs(this.ledger);
             }
             catchup(); // avoid delay
             this.taskman.setTimer(this.network.node_config.block_catchup_interval, &catchup, Periodic.Yes);
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -82,6 +82,10 @@ mixin AddLogger!();
 /// Maximum number of blocks that will be sent in a call to getBlocksFrom()
 private enum uint MaxBatchBlocksSent = 20;
 
+/// Maximum number of transactions that will be sent in a call to
+/// getTransactionByHash
+private enum uint MaxBatchTranscationsSent = 100;
+
 /*******************************************************************************
 
     Implementation of the FullNode API
@@ -762,5 +766,30 @@ public class FullNode : API
                 }
             });
         }
+    }
+
+    /***************************************************************************
+
+        Params:
+            tx_hashes = A Set of Transaction hashes
+
+        Returns:
+            Transactions, if found in the pool, corresponding to the
+            requested hashes
+
+    ***************************************************************************/
+
+    public Transaction[] getTransactions (Set!Hash tx_hashes) @safe
+    {
+        Transaction[] found_txs;
+        foreach (hash; tx_hashes)
+        {
+            if (found_txs.length >= MaxBatchTranscationsSent)
+                break;
+            auto tx = this.pool.getTransactionByHash(hash);
+            if (tx != Transaction.init)
+                found_txs ~= tx;
+        }
+        return found_txs;
     }
 }
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -77,6 +77,9 @@ public class Ledger
     /// Pool of transactions to pick from when generating blocks
     private TransactionPool pool;
 
+    /// TX Hashes Ledger encountered but dont have in the pool
+    private Set!Hash unknown_txs;
+
     /// The last block in the ledger
     private Block last_block;
 
@@ -388,6 +391,8 @@ public class Ledger
             this.tx_stats.increaseMetricBy!"agora_transactions_rejected_total"(1);
             return false;
         }
+        // If we were looking for this TX, stop
+        this.unknown_txs.remove(tx.hashFull());
 
         this.tx_stats.increaseMetricBy!"agora_transactions_accepted_total"(1);
         return true;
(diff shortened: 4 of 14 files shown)


source/agora/node/FullNode.d
```diff
@@ -733,4 +737,29 @@ public class FullNode : API
...
     }
+
+    /***************************************************************************
+
+        Params:
+            tx_hashes = A Set of Transaction hashes
+
+        Returns:
+            Transactions corresponding to the requested hashes or
+            Transaction.init for hashes that can't be found in the pool
```
AndrejMitrovic: That's not what the code does though, it doesn't return `Transaction.init` but just skips those transactions entirely.

source/agora/node/FullNode.d
```diff
@@ -733,4 +737,29 @@ public class FullNode : API
...
+    ***************************************************************************/
+
+    public Transaction[] getTransactions (Set!Hash tx_hashes) @safe
+    {
+        Transaction[] found_txs;
+        foreach (hash; tx_hashes)
+        {
+            if (found_txs.length >= MaxBatchTranscationsSent)
+                break;
+            auto tx = this.pool.getTransactionByHash(hash);
```
AndrejMitrovic, line 790: I feel like we can optimize this in the future. I am pretty sure databases are well optimized for looking up several keys at once, and not just one at a time.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -800,6 +800,8 @@ extern(D):
             auto data = deserializeFull!ConsensusData(value[]);
             if (auto fail_reason = this.ledger.validateConsensusData(data))
             {
+                if (fail_reason == Ledger.InvalidConsensusDataReason.MayBeValid)
+                    return ValidationLevel.kInvalidValue;
```
AndrejMitrovic: By SCP's definition if you return `kInvalidValue` then it means the node may never vote this value as valid afterwards. Are you sure we're not supposed to use `kMaybeValidValue`?

omerfirmak: Well, initially we returned `kMaybeValidValue` here. But I had some problems in the balloting phase. Node started to confirm statements without having access to all the previous blocks. I completely went from these [two](https://github.com/bpfkorea/agora/blob/bbf523199b6efd1667ee0e3d79ad59d4c3ea93c2/source/agora/consensus/protocol/Nominator.d#L539) [comments](https://github.com/bpfkorea/agora/blob/406a22df4c33dfd2f2e3a61811fe6e73aea94852/source/scpp/src/scp/SCPDriver.h#L56-L58). and decided `kInvalidValue` was more appropriate. In nominating phase it does not matter what we return they are the same. In balloting we should still fully validate the value, but our ledger may be a bit behind to return `kMaybeValidValue`. That is indeed what Stellar does actually. They only return `kMaybeValidValue` [here](https://github.com/stellar/stellar-core/blob/54b03f755ae5d5aa12a799c8f1ee4d87fc9d1a1d/src/herder/HerderSCPDriver.cpp#L237), which is the case where the value is fully validated but ledger is not fully caught up.

source/agora/node/FullNode.d
omerfirmak: Agreed, this was just to get stuff working.

source/agora/node/FullNode.d
omerfirmak: Oops, copy paste error.

source/agora/consensus/protocol/Nominator.d
omerfirmak: > then it means the node may never vote this value as valid afterwards

I am not entirely sure, but if the same value is renominated, it should probably count as a new value.

AndrejMitrovic: Well okay. The docs are not amazing..

AndrejMitrovic: I mean the SCP docs.

source/agora/node/Ledger.d
```diff
@@ -677,15 +685,19 @@ public class Ledger
     {
         const expect_height = Height(this.getBlockHeight() + 1);
         auto utxo_finder = this.utxo_set.getUTXOFinder();
-        auto unknown_txs = false;
+        auto missing_ctx = false;
```
AndrejMitrovic: Is "ctx" supposed to mean "context"?

source/agora/network/NetworkManager.d
```diff
@@ -893,6 +895,37 @@ public class NetworkManager
...
+        foreach (peer; this.peers[])
+        {
+            if (unknown_txs.length == 0)
+                break;
+
+            foreach (tx; peer.getTransactions(unknown_txs))
+            {
+                try
+                {
+                    ledger.acceptTransaction(tx);
```
AndrejMitrovic, line 887: You need to check the return value.

AndrejMitrovic (commented):
Overall LGTM. Just a few issues.

source/agora/network/NetworkManager.d
omerfirmak: Do we want to log it, if its invalid? I could not think of any other reason for checking if it is valid or not.

source/agora/node/Ledger.d
omerfirmak: Yep

source/agora/network/NetworkManager.d
AndrejMitrovic: You can't think of any? Check the next line.

If the tx is invalid then the missing tx set should not be modified..

AndrejMitrovic: Wait I just realized something.

AndrejMitrovic: My assumption was wrong. In fact if the tx is invalid there is no other tx with a matching hash, that would make it a hash collision.

AndrejMitrovic: I wonder if this can be abused in some way. If a bad actor nominates a big tx set of invalid hashes, will that cause the node to constantly try to download these txes?

AndrejMitrovic: by invalid I mean fake hashes of non-existing txes.

omerfirmak: It was hard to follow but, yes :)

omerfirmak: Oops I didnt see the last 2 messages when I replied to you.
 
 > I wonder if this can be abused in some way. If a bad actor nominates a big tx set of invalid hashes, will that cause the node to constantly try to download these txes?
 
Until the end of that round, yes. We clear the unknown txs at the end of each round. "big tx set" will probably be bound by some upper limit too eventually. Maybe the block size or tx count?

omerfirmak: We can maybe give up on looking for hashes after a certain amount of tries

AndrejMitrovic: Anyway it shouldn't block the PR.

omerfirmak:
I rebased the fee dıstrıbutıon commits and added some changes to better handle the externalization with missing TXs.
Tests pass on my local machine but I feel like block signatures could be easily broken by this. @hewison-chris, can you check if these make sense to you, especially the early returns in the last commit .

The last commit will be formatted/documented and mostly squashed to previous commits.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -860,9 +870,14 @@ extern(D):
 
             Hash random_seed = this.ledger.getExternalizedRandomSeed(
                 this.ledger.getBlockHeight(), data.missing_validators);
+
+            Transaction[] externalized_tx_set;
+            if (this.ledger.getValidTXSet(data, externalized_tx_set))
+                return;
```
hewison-chris, line 892: I feel like this should be an assert(0) here as we can not ignore the invalid tx set when we should be externalizing the block.

omerfirmak: Since we now externalize TX hashes, we might not have all the TXs in our pool at the point the value was externalized. So that could fail for missing TXs as well. In that case we should be able to catch up to the network and sign the unsigned blocks along the way.

hewison-chris: So will this externalize be triggered again after the missing TXs are received?

omerfirmak: No, node will have to download the full block from the network.

hewison-chris: Perhaps I should also not assert in that case, when there are not enough signatures. The node can also wait to recieve the block during a periodic catchup and test again for required number of signatures before externalizing.

hewison-chris: I created [PR #1517](https://github.com/bpfkorea/agora/pull/1517) to address this for block sigs. I agree with your logic here.

omerfirmak: Great, thanks!

source/agora/consensus/protocol/Nominator.d
```diff
@@ -529,8 +529,14 @@ extern(D):
             const Block prev_block = blocks.front;
             Hash random_seed = this.ledger.getExternalizedRandomSeed(
                 this.ledger.getBlockHeight(), con_data.missing_validators);
-            const Block proposed_block = makeNewBlock(prev_block, con_data.tx_set,
-                con_data.timestamp, con_data.enrolls, random_seed, con_data.missing_validators);
+
+            Transaction[] received_tx_set;
+            if (this.ledger.getValidTXSet(con_data, received_tx_set))
```
AndrejMitrovic: So it returns `true` if it fails?

AndrejMitrovic: Ah wait it returns a string.

I thought this behavior was deprecated.

AndrejMitrovic: This would make it clearer:

`if (auto error = this.ledger.getValidTXSet(con_data, received_tx_set))`

Although you could argue that the string is unused anyway.

omerfirmak: Yeah, lets just make it explicit.

hewison-chris: If the string is not used then it may as well be a bool. It is not an `isInvalid...` function so it is not obvious what is success status.

AndrejMitrovic (approved):

AndrejMitrovic:
I'll let @hewison-chris finish his review & then we can merge.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -529,8 +529,14 @@ extern(D):
             const Block prev_block = blocks.front;
             Hash random_seed = this.ledger.getExternalizedRandomSeed(
                 this.ledger.getBlockHeight(), con_data.missing_validators);
-            const Block proposed_block = makeNewBlock(prev_block, con_data.tx_set,
-                con_data.timestamp, con_data.enrolls, random_seed, con_data.missing_validators);
+
+            Transaction[] received_tx_set;
+            if (auto fail_reason = this.ledger.getValidTXSet(con_data, received_tx_set))
+                return; // We dont have all the TXs for this block. Try to catchup
```
hewison-chris, line 542: We probably should log here the fail_reason as there is nothing logged in any of the called functions.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -856,8 +868,15 @@ extern(D):
 
             Hash random_seed = this.ledger.getExternalizedRandomSeed(
                 this.ledger.getBlockHeight(), data.missing_validators);
-            const block = makeNewBlock(prev_block, data.tx_set, data.timestamp, data.enrolls,
-                random_seed, data.missing_validators);
+
+            Transaction[] externalized_tx_set;
+            if (auto fail_reason = this.ledger.getValidTXSet(data,
+                externalized_tx_set))
+                return;
```
hewison-chris, line 892: Also here

hewison-chris (commented):
Looks good, only a couple of logging requests.

linked0 (approved):
LGTM

omerfirmak:
Thanks for the reviews everyone. I have made the changes and rebased. I think this is good to merge it once it goes green.

AndrejMitrovic:
> Thanks for the reviews everyone. I have made the changes and rebased. I think this is good to merge it once it goes green.

It would be good to test this in our test environment. It's a significant change.

omerfirmak:
> > Thanks for the reviews everyone. I have made the changes and rebased. I think this is good to merge it once it goes green.
> 
> It would be good to test this in our test environment. It's a significant change.

Together with the Coinbase TXs they make a dangerous duo. Once we get `faucet` to include fees we should be able to test extensively.

omerfirmak:
@hewison-chris , is this the failure we were seeing in NetworkManager tests?
https://github.com/bpfkorea/agora/pull/1495/checks?check_run_id=1739937365

hewison-chris:
I don't see any new logging. Did you think it will be too much logging?

omerfirmak:
> I don't see any new logging. Did you think it will be too much logging?

Check L536 and L880. I've added only to the places you commented on.

hewison-chris:
> @hewison-chris , is this the failure we were seeing in NetworkManager tests?
> https://github.com/bpfkorea/agora/pull/1495/checks?check_run_id=1739937365

Sorry, looks like the new network unit test for PeriodicCatchup can sometimes fail. Will take a look now.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -711,8 +721,14 @@ extern(D):
         const prev_block = blocks.front;
         Hash random_seed = this.ledger.getExternalizedRandomSeed(
                 this.ledger.getBlockHeight(), con_data.missing_validators);
-        const proposed_block = makeNewBlock(prev_block, con_data.tx_set,
-            con_data.timestamp, con_data.enrolls, random_seed, con_data.missing_validators);
+
+        Transaction[] signed_tx_set;
+        if (auto fail_reason = this.ledger.getValidTXSet(con_data, signed_tx_set))
+            return;
```
hewison-chris, line 735: Also needed here.

source/agora/node/Ledger.d
```diff
@@ -237,8 +240,13 @@ public class Ledger
         Hash random_seed = this.slash_man.getExternalizedRandomSeed(
             this.getBlockHeight(), data.missing_validators);
 
-        auto block = makeNewBlock(this.last_block, data.tx_set, data.timestamp, data.enrolls,
-            random_seed, data.missing_validators);
+        Transaction[] externalized_tx_set;
+        if (auto fail_reason = this.getValidTXSet(data, externalized_tx_set))
+            return false;
```
hewison-chris, line 247: Also here

source/agora/node/Ledger.d
```diff
@@ -254,8 +262,14 @@ public class Ledger
             .map!(idx => PublicKey(this.enroll_man.getValidatorAtIndex(next_block, idx)[]))
             .map!(K => WK.Keys[K])
             .array();
-        const block = makeNewTestBlock(this.last_block, data.tx_set,
-            data.enrolls, Hash.init, data.missing_validators, public_keys,
+
+        Transaction[] externalized_tx_set;
+        if (auto fail_reason = this.getValidTXSet(data, externalized_tx_set))
+            return false;
```
hewison-chris, line 274: Also here

hewison-chris (approved):
LGTM. Just a couple more log lines.

omerfirmak:
Added more logs. 

`PeriodicCatchup` is still failing though. Should we wait for you to take a look @hewison-chris ?

hewison-chris:
> Added more logs.
> 
> `PeriodicCatchup` is still failing though. Should we wait for you to take a look @hewison-chris ?

Yes I know why it fails. Fixing it now. I will have to gossip the signatures also after receiving block during catchup.

PR #1496, opened 2021-01-14
Pushing down the AGORA_VERSION to DockerHub build

ferencdg:


Geod24:
Why don't you just run the script and then read the resulting file ?

ferencdg:
> Why don't you just run the script and then read the resulting file ?

I was thinking about that, but then I guess I have to install rdmd before every build...

Geod24:
Good point. Current method is better then.

hooks/build
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env sh
+
```
Geod24, line 2: set -xeu
set -o pipefail

hooks/build
```diff
@@ -0,0 +1,17 @@
...
+GIT_DESCRIBE_OUT=`git describe`
+
+if [ -n "$GIT_DESCRIBE_OUT"]
+then
+	AGORA_VERSION=$GIT_DESCRIBE_OUT
+fi
+
+echo "Agora version is: $AGORA_VERSION"
+
+docker build --build-arg AGORA_VERSION=$AGORA_VERSION -f $DOCKERFILE_PATH -t $IMAGE_NAME .
```
Geod24, line 17: I guess all those variables are provided by docker hub ?

ferencdg: yes, I think if we didn't have the build file, then the command that is executed is:
`docker build -f $DOCKERFILE_PATH -t $IMAGE_NAME .`

Geod24:
Looks good. Why is it still a draft ?

Geod24:
I think this broke the dockercloud build, because it does a shallow clone:
```
Executing build hook...
+ set -o pipefail
+ AGORA_VERSION=HEAD
++ git describe
fatal: No names found, cannot describe anything.
+ GIT_DESCRIBE_OUT=
build hook failed! (128)
```

ferencdg:
> I think this broke the dockercloud build, because it does a shallow clone:
> 
> ```
> Executing build hook...
> + set -o pipefail
> + AGORA_VERSION=HEAD
> ++ git describe
> fatal: No names found, cannot describe anything.
> + GIT_DESCRIBE_OUT=
> build hook failed! (128)
> ```

my dockerhub build initially failed, but after I pushed smoe tag back to the remote repository I no longer got that error message

Geod24:
You mean you pushed them to your fork ?

ferencdg:
and it strangely worked after that...

Geod24:
Yeah, docker hub isn't great and tends to pull from the fork...

PR #1502, opened 2021-01-18
Add slashing information to `ConsensusParams'

### LINKED ISSUE #1486: Make the slashing policy part of the consensus parameter
We have the slashing policy for misbehaving validators that we slash the validator that never sends any pre-image after its initial enrollment. But the policy can be modified when needed as follows.
- Only send one pre-image after enrolling then no more;
- Send pre-image for half of its cycle (potentially in one burst) then stop sending pre-image;

By making the threshold of the count of missing pre-images part of the consensus parameter, by which we slash any misbehaving validators, we can change the slashing policy more flexibly.

**Definition Done**

- [ ] Make the threshold part of the consensus parameter
- [ ] Make code flexible to apply the threshold to slashing misbehaving validators

linked0:
We add two slashing information to `ConsensusParams` such as `amount of penalty` and `slash threshold`. The `amount of penalty` is trivial. And the `slash threshold` is the criteria by which we punish the misbehaving validators. Currently, we punish the validators as soon as we detect them not to reveal preimages. But with this implementation, we can set more flexible policies about punishing the validators like slashing after two heights of missing pre-images or one-third of the cycle of missing pre-images.

Fixes #1486


source/agora/common/Config.d
```diff
@@ -174,6 +175,9 @@ public struct NodeConfig
     /// The duration between requests for retrieving the latest blocks
     /// from all other nodes
     public Duration block_catchup_interval = 20.seconds;
+
+    /// The amount of a penalty for slashed validators
+    public Amount slash_penalty_amount = Amount(100_000_000_000);
```
AndrejMitrovic: How about adding a new commit which implements a `coins` function which returns:

```D
Amount coins (in ulong num_coins)
{
    return Amount(UnitPerCoin.value * num_coins, true);
}
```

And then it can be used as:

```D
public Amount slash_penalty_amount = 10_000.coins;
```

MinFreezeAmount does something similar, but it uses the private `value` field. So I suggest adding a helper function. See here: https://github.com/bpfkorea/agora/blob/51466376cef32f26d04ab0d272d43951d4aec89e/source/agora/common/Amount.d#L41-L42

AndrejMitrovic: Also it would be easier if the slashing value in the config is in coins rather than units.

linked0: > How about adding a new commit which implements a coins function which returns:
> Amount coins (in ulong num_coins)

Okay, it would be good. I will add it.

linked0:
Rebased and updated.
Ready for review.

source/agora/common/Config.d
```diff
@@ -189,6 +189,9 @@ public struct NodeConfig
     /// How frequent the payments to Validators will be in blocks
     public uint payout_period = 144;
 
+    /// The threshold to slash the validators did not reveal pre-images
```
AndrejMitrovic: Could we document what threshold means? Something like "total number of blocks a node may refuse to reveal its preimage before it becomes slashed".

AndrejMitrovic (approved):
Just a few comments.

source/agora/common/Config.d
linked0: Updated.

source/agora/consensus/SlashPolicy.d
```diff
@@ -105,7 +110,7 @@ public class SlashPolicy
 
         foreach (idx, utxo_key; keys)
         {
-            if (!this.hasRevealedPreimage(height, utxo_key))
+            if (!this.hasRevealedPreimage(this.heightForSlashing(height), utxo_key))
```
AndrejMitrovic: Am I understanding this correctly? This slashes the node if it doesn't reveal at least X preimages in advance, where X is the threshold?

linked0: The threshold, X is to specify how many blocks a validator can miss pre-images before being slashed. If it's set to 0 which is the default value, the validator is slashed as soon as the case of missing pre-images is detected. If 1 is set to the value, the validator can miss the pre-images for one block. This concept is based on the cases of missing preimages, which is described [the original design of the slashing protocol](https://github.com/bpfkorea/agora/issues/1076).

linked0:
Would you please review this? It's only approved by @AndrejMitrovic.

hewison-chris (approved):
LGTM. The test failures are from PeriodicCatchup which I am currently fixing.

source/agora/common/Config.d
```diff
@@ -197,6 +198,13 @@ public struct ConsensusConfig
 
     /// How frequent the payments to Validators will be in blocks
     public uint payout_period = 144;
+
+    /// Total number of blocks a node may refuse to reveal its preimage
+    /// before it becomes slashed
+    public uint slash_threshold = 0;
```
Geod24: Is that ever going to be more than 0 ?

source/agora/common/Config.d
```diff
@@ -474,6 +482,8 @@ private ConsensusConfig parseConsensusConfig (Node* node, in CommandLine cmdln)
     const tx_payload_fee_factor = get!(uint, "consensus", "tx_payload_fee_factor")(cmdln, node);
     const validator_tx_fee_cut = get!(ubyte, "consensus", "validator_tx_fee_cut")(cmdln, node);
     const payout_period = get!(uint, "consensus", "payout_period")(cmdln, node);
+    auto slash_threshold = opt!(uint, "node", "slash_threshold")(cmdln, node);
+    auto slash_penalty_value = opt!(ulong, "node", "slash_penalty_amount")(cmdln, node);
```
Geod24: You can use `get` here.

tests/system/node/0/config.yaml
```diff
@@ -19,6 +19,8 @@ node:
 
 consensus:
   validator_cycle: 20
+  slash_threshold: 0
+  slash_penalty_amount: 10000
```
Geod24: No need to write it if you're using the default value

Geod24 (commented):
What's the rationale for accepting missing pre-images ? I don't think we should accept any missing ones, because of how easy it is to set a watchtower, and because they can be revealed way ahead of time.

tests/system/node/0/config.yaml
linked0: Right. But It's just a template so that a user can change it.

linked0:
After discussing the necessity of  'slash_thresh`, we conclude that we don't need the feature. So I will remove the related code in this PR and request a review again.

linked0:
Ready for review!

AndrejMitrovic (approved):

AndrejMitrovic:
LGTM.

Issue #734, opened 2020-04-07
Merge the catchup code and the startup code for enrollment/preimage/utxo_set


Geod24:
Currently the periodic catch does this:
https://github.com/bpfkorea/agora/blob/05bfe591aec457544bfe9fc8b799f592d4d6b303/source/agora/network/NetworkManager.d#L208-L239

While the constructor of `Ledger` does this:
https://github.com/bpfkorea/agora/blob/05bfe591aec457544bfe9fc8b799f592d4d6b303/source/agora/node/Ledger.d#L93-L117

There's an obvious duplication here, that we should address.

linked0:
There are two options to implement it.
@Geod24 Do you mean the first option as the goal of this issue?

**First option**
```
public void startPeriodicCatchup (Ledger ledger, bool delegate() @safe isNominating) 
{ 
   this.taskman.runTask( 
   () 
   { 
       // periodic task 
       while (1) 
       { 
           while(auto block = this.ledger.getNextBlock()) <======                                                                         
               ledger.acceptBlock(block);                 <======
                                                                          
           this.getBlocksFrom( 
               ledger.getBlockHeight() + 1, 
               blocks => blocks.all!(block => 
                   // do not alter the state of the ledger if 
                   // we're currently nominating 
                   !isNominating() && ledger.acceptBlock(block))); 

           this.taskman.wait(2.seconds); 
       } 
   }); 
} 
```

After writing the code above, we can delete the code about reading block, and updating utxos and validators, in the constructor of the `Ledger`.

**Second option**
```
private void rebuildUTXOAndValidatorSet (ref Block gen_block)
{
    foreach (height; 0 .. this.last_block.header.height + 1)
    {
        Block block;
        this.storage.readBlock(block, Height(height));
        this.acceptBlock(block);
    }
}
```

With the code, we can remove the duplicate function calls from the `rebuildUTXOAndValidatorSet`, which are already called in the `validateBlock` and `addValidatedBlock` which are called in the `acceptBlock` in turn.

linked0:
The PR #1505 is approved and merged as the final code of this issue by @Geod24. So I close this.

PR #1505, opened 2021-01-18
Refactor the process loading blocks

linked0:
The catch-up process and the process loading blocks from the block storage do almost the same operation. But the two codes existed redundantly in different places. So we refactor the code only to use the `acceptBlock` function when the processes are going. This PR might be the intermediate step or it would be the final code. This can be discussed more in the middle of reviewing.

Part of #734

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -166,7 +166,18 @@ public class Ledger
                 "different from the one in the config file");
 
         if (this.utxo_set.length == 0)
-            this.rebuildUTXOAndValidatorSet(gen_block);
+        {
+            // clear validator set
+            this.enroll_man.removeAllValidators();
+
+            foreach (height; 0 .. this.last_block.header.height + 1)
+            {
+                Block block;
+                this.storage.readBlock(block, Height(height));
+                this.acceptBlock(block, false);
+                this.last_block = block;
+            }
+        }
         else if (this.enroll_man.validatorCount() == 0)
         {
             // +1 because the genesis block counts as one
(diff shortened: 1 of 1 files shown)


source/agora/node/Ledger.d
```diff
@@ -165,7 +165,19 @@ public class Ledger
...
+        {
+            // clear validator set
+            this.enroll_man.removeAllValidators();
+
+            foreach (height; 0 .. this.last_block.header.height + 1)
+            {
+                Block block;
+                this.storage.readBlock(block, Height(height));
+                if (!this.acceptBlock(block, false))
+                    break;
```
omerfirmak: This looks like dead code to me. With `externalize = false`, `acceptBlock` will throw instead of returning false, right?

linked0: > This looks like dead code to me. With externalize = false, acceptBlock will throw instead of returning false, right?

Right, will fix it.

linked0:
Ready for review.

source/agora/node/Ledger.d
```diff
@@ -326,24 +272,31 @@ public class Ledger
...
-            log.trace("Rejected block: {}: {}", fail_reason, block.prettify());
-            return false;
+            if (externalized)
+            {
+                log.trace("Rejected block: {}: {}", fail_reason, block.prettify());
+                return false;
+            }
+            else
+                throw new Exception("A block loaded from disk is invalid: " ~
+                    fail_reason);
```
Geod24, line 296: Why not `return false` here and throw in the constructor ?
Having a function that `throw` or return `false` depending on a boolean parameter is usually not a good idea.

linked0: The loading process of the constructor was throwing `Exception`. But, Just use of `return false` would be more consistent because the original `acceptBlock` just returned a boolean value. I will update it.

Geod24:
Can you rebase?

linked0:
> an you rebase?

I will do it soon.

linked0:
> Can you rebase?

Rebased.

linked0:
@Geod24 Do you think this PR is the final code about issue #734? I think so because I refactored the code so that two redundant codes reading and validating, and add blocks into the ledger are replaced with codes that only call the `acceptBlock` function. There is in fact another option that the code accepting blocks in the constructor of `Ledger` is moved into the `startPeriodicCatchup` function like this.

```
public void startPeriodicCatchup (Ledger ledger, bool delegate() @safe isNominating) 
{ 
...
       // periodic task 
       while (1) 
       { 
           while(auto block = this.ledger.getNextBlock()) <======     see this                                                                    
               ledger.acceptBlock(block);                 <====== see this
                                                                          
           this.getBlocksFrom( 
               ledger.getBlockHeight() + 1, 
               blocks => blocks.all!(block => 
                   // do not alter the state of the ledger if 
                   // we're currently nominating 
                   !isNominating() && ledger.acceptBlock(block))); 

...
} 
```

But If the `Ledger` can read the blocks from files by itself, it's not bad for the code accepting blocks to exist in the constructor of the `Ledger`.

Geod24:
> @Geod24 Do you think this PR is the final code about issue #734? 

Yes

Geod24:
Rebased, hopefully it'll work now.

Geod24 (approved):

PR #1507, opened 2021-01-18
Adding the ability to build name-registry image

ferencdg:
Prerequisite for bpfkorea/infrastructure#99

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -92,6 +92,7 @@
             "name": "name-registry",
             "targetName": "name-registry",
             "mainSourceFile": "source/agora/registry/main.d",
+            "libs-linux": [ "unwind" ],
             "excludedSourceFiles": ["source/agora/cli/*", "source/agora/node/main.d"]
         },
         {
--- a/.github/workflows/registry-docker.yml
+++ b/.github/workflows/registry-docker.yml
@@ -1,0 +1,31 @@
+name: Build registry
+
+on: [push, pull_request]
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        if: github.event_name == 'push'
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build_and_push
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ github.event_name == 'push' }}
+          file: ./Dockerfile.NameRegistry
+          tags: bpfk/name-registry
+
+      - name: Image digest
+        run: echo ${{ steps.docker_build.outputs.digest }}
--- a/Dockerfile.NameRegistry
+++ b/Dockerfile.NameRegistry
@@ -1,0 +1,15 @@
+# Build Agora from source
+FROM bpfk/agora-builder:latest AS RegistryBuilder
+ARG AGORA_VERSION="HEAD"
+ADD . /root/agora/
+WORKDIR /root/agora/
+RUN apk --no-cache add llvm-libunwind-dev dub
+RUN dub --version
+RUN AGORA_VERSION=${AGORA_VERSION} dub build --skip-registry=all --compiler=ldc2 --config name-registry
+
+# Runner
+FROM alpine:3.12.0
+RUN apk --no-cache add ldc-runtime llvm-libunwind libgcc libsodium libstdc++ sqlite-libs
+COPY --from=RegistryBuilder /root/agora/build/name-registry /usr/local/bin/name-registry
+WORKDIR /name-registry/
+ENTRYPOINT [ "/usr/local/bin/name-registry" ]


.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
...
+          build-args: |
+            AGORA_VERSION=HEAD
+
+      -
+        name: Build and push
+        if: github.event_name == 'push'
+        id: docker_build_and_push
+        uses: docker/build-push-action@v2
+        with:
+          push: true
```
Geod24: You can combine those in one step with:
```suggestion
          push: github.event_name == 'push'
```

.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
+name: docker
+
+on: [push, pull_request]
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Set up QEMU
```
Geod24:
```suggestion
      - name: Set up QEMU
```

Ditto for the others

.github/workflows/docker.yml
```diff
@@ -0,0 +1,47 @@
...
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      -
+        name: Only build
+        if: github.event_name == 'pull_request'
+        id: docker_build
+        uses: docker/build-push-action@v2
+        with:
+          push: false
+          file: ./DockerfileNameRegistry
+          tags: bpfk/agora:name-registry-test
```
Geod24: Created `bpfk/name-registry`

DockerfileNameRegistry
```diff
@@ -0,0 +1,15 @@
...
+ARG DUB_OPTIONS
+ARG AGORA_VERSION="HEAD"
+ADD . /root/agora/
+WORKDIR /root/agora/
+RUN apk --no-cache add llvm-libunwind-dev
+RUN AGORA_VERSION=${AGORA_VERSION} dub build --skip-registry=all --compiler=ldc2 --config name-registry ${DUB_OPTIONS}
+
+# Runner
+FROM alpine:3.12.0
+RUN apk --no-cache add ldc-runtime libexecinfo llvm-libunwind libgcc libsodium libstdc++ sqlite-libs
```
Geod24: You don't need `libexecinfo`

Geod24 (commented):
I'd also love to see this build and push Agora so we can ditch docker hub, but it can be left to another issue if you prefer.

.github/workflows/docker.yml
ferencdg: > You can combine those in one step with:

I wish that worked... and I tried it before... but it seems it always trigger even when I do a pull_request or they don't trigger even for push (I cannot remember which one)

.github/workflows/docker.yml
ferencdg: > Ditto for the others

I will clean up all those and more, this was really just a DRAFT to make sure the idea works

.github/workflows/docker.yml
ferencdg: > > You can combine those in one step with:
> 
> I wish that worked... and I tried it before... but it seems it always trigger even when I do a pull_request or they don't trigger even for push (I cannot remember which one)

just confirmed, it doesn't push the image:
https://github.com/bpfkorea/agora/actions/runs/495569212
I have to expicitely set it to 'true'

dub.json
```diff
@@ -92,6 +92,7 @@
             "name": "name-registry",
             "targetName": "name-registry",
             "mainSourceFile": "source/agora/registry/main.d",
+            "libs-linux": [ "unwind" ],
```
Geod24, line 95: It should only be Musl

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build_and_push
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ github.event_name != 'pull_request' }}
+          file: ./DockerfileNameRegistry
```
Geod24:
```suggestion
          file: ./Dockerfile.NameRegistry
```

In the hope this will give us syntax highlighting

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
...
+          tags: bpfk/name-registry
+          build-args: |
+            AGORA_VERSION=HEAD
```
Geod24: Since it's set in the Dockerfile, do we need it here ?

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Build and push
+        id: docker_build_and_push
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ github.event_name != 'pull_request' }}
```
Geod24: I think `== 'push'` would be more reliable, or is there a reason to have it like this ?

DockerfileNameRegistry
```diff
@@ -0,0 +1,15 @@
+# Build Agora from source
+FROM bpfk/agora-builder:latest AS RegistryBuilder
+ARG DUB_OPTIONS
```
Geod24: I don't think you need this (or `${DUB_OPTIONS}`). It was put to allow us to build Agora with coverage on in the integration tests.

.github/workflows/docker.yml
```diff
@@ -0,0 +1,33 @@
+name: docker
```
Geod24:
```suggestion
name: Build registry
```

So it doesn't get confused with Agora's docker job. Same for the file name, something like `registry-docker.yml` or whatever you see fit.

Geod24 (approved):
LGTM, only nits

.github/workflows/docker.yml
ferencdg: I planned to put both agora and name-registry jobs into the same workflow file, so if one fails, then we don't attempt to build the other. Anyways, let me separate them.

dub.json
ferencdg: I wonder if this is a dub bug or we are using an outdated version of dub on our agora build container, but 
"libs-linux-musl": [ "unwind" ] has no effect. The dub build on the docker is: 
`DUB version 1.21.0, built on May 15 2020`

Geod24 (approved):

PR #1521, opened 2021-01-22
Make BlockStorage.{load,saveBlock} not nothrow

Geod24:
```
They are only used in one place each, as they should,
and we trigger an HLT if they are to return false anyway.
```


Geod24:
Green

AndrejMitrovic (approved):

PR #1522, opened 2021-01-22
Document and simplify dependency injection

Geod24:
```
Since many places already relied on accessing the config object directly,
make it a guarantee that it'll be filled, as well as the consensus parameters.
It also allows to remove the oddity of requiring a ValidatorConfig to be
passed to a FullNode.
```


AndrejMitrovic (approved):
Very nice!

Geod24:
Just realized we had `getNominator` too in `Validator`, so did the changes there too.

Geod24:
Green

PR #1531, opened 2021-01-22
Re-enable parallel GC

Geod24:
It should work, as the oldest supported version is LDC v1.20, which has the fix.


omerfirmak (approved):

Geod24:
For ref, the issue is explained here: https://github.com/dlang/dub/issues/1812#issuecomment-568981831
TL;DR, druntime didn't account for the TLS when calling `pthread_create`, which caused the GC to have a ridiculously low stack.

Geod24:
Bypassing flakey CI.

PR #1533, opened 2021-01-22
Test: Respect environment variable values

omerfirmak:
We used to only check for existence of the variables. "dchatty=0 dub test" would still be chatty.


Geod24:
Wouldn't `to!bool` be better ? *might* recognize `true`. Probably not `yes` tho.

omerfirmak:
I wanted to avoid getting people confused when they use `dchatty=1`

Geod24 (approved):

PR #1534, opened 2021-01-24
Fixed: unable to terminate name-registry

ferencdg:
Name registry didn't terminate while pressing CTRL+C
due to the Vibe.d bug: https://github.com/vibe-d/vibe-core/issues/205

This is needed before for infra issue #99.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -92,6 +92,7 @@
             "name": "name-registry",
             "targetName": "name-registry",
             "mainSourceFile": "source/agora/registry/main.d",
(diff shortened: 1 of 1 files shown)


Geod24:
I don't follow... This bug has been fixed, hasn't it?

ferencdg:
> I don't follow... This bug has been fixed, hasn't it?

right, it seems Sonke fixed recently, haven't noticed that. However our eventcore is very old, and doesn't seem to have this fix.

ferencdg:
will be solved by updating eventcore

PR #1535, opened 2021-01-24
Update eventcore to v0.9.13

Geod24:
This also contain our workaround, which will become unnecessary once we switch to LDC v1.25


AndrejMitrovic (approved):

Geod24:
Yeah I'm not responsible for that OOM on docker hub.

PR #1538, opened 2021-01-25
Network section in config shouldn't be mandatory

### LINKED ISSUE #1536: Network section in config shouldn't be mandatory
currently if the network section is empty, agora exits with: 
`Failed to parse config file 'config.yaml'. Error: The 'network' section is mandatory and must specify at least one item`

However if we have a name registry set up, then the network section shouldn't be mandatory.

ferencdg:
If a validator node has a name registry in the config, then
the network section can be empty.

This is a pre-requisite for infra #99

fixes: #1536

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -305,11 +305,38 @@ public Config parseConfigString (string data, string path)
 unittest
 {
     assertThrown!Exception(parseConfigString("", "/dev/null"));
-    // Missing 'network' section
+
+    // Missing 'network' section for a non validator node
     {
         immutable conf_str = `
-node:
-  validator: false
+validator:
+  enabled: false
+`;
+        assertThrown!Exception(parseConfigString(conf_str, "/dev/null"));
+    }
+
+    // Missing 'network' section for a validator node with name registry
+    {
+        immutable conf_str = `
+validator:
+  enabled: true
+  registry_address: http://127.0.0.1:3003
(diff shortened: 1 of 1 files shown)


AndrejMitrovic (approved):

source/agora/common/Config.d
```diff
@@ -356,13 +383,15 @@ private Config parseConfigImpl (in CommandLine cmdln, Node root)
...
+        network : assumeUnique(parseSequence("network", cmdln, root, true)),
         dns_seeds : assumeUnique(parseSequence("dns", cmdln, root, true)),
         logging: parseLoggingSection("logging" in root, cmdln),
         event_handlers: parserEventHandlers("event_handlers" in root, cmdln),
     };
 
-    enforce(conf.network.length > 0, "Network section is empty");
+    enforce((conf.network.length > 0) ||
+            (conf.validator.enabled && conf.validator.registry_address != "disabled"),
+         "Network section is empty");
```
Geod24:
```suggestion
    if (conf.validator.enabled)
        enforce(conf.network.length || conf.validator.registry_address != "disabled",
            "Either the network section must not be empty, or 'validator.registry_address' must be set");
    else
        enforce(conf.network.length, "Network section must not be empty");    
```

This should provide a slightly better user experience.

Geod24 (approved):
Nice. Just a nitpick on  the error message.

Geod24 (approved):

kchulj (approved):

PR #1539, opened 2021-01-25
Building Agora images with github actions

### LINKED ISSUE #1513: Github should create bpfk:agora docker images instead of Dockerhub

ferencdg:
fixes: #1513

### DIFF
--- a/.github/workflows/agora-docker.yml
+++ b/.github/workflows/agora-docker.yml
@@ -1,0 +1,55 @@
+name: Build agora
+
+on: [push, pull_request]
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+
+      - name: Login to DockerHub
+        if: github.event_name == 'push'
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Check out repository
+        uses: actions/checkout@v2
+        with:
+          submodules: false
+          persist-credentials: false
+
+      - name: Determining Agora version number
+        shell: bash
+        env:
+          github_ref: ${{ github.ref }}
+          github_sha: ${{ github.sha }}
+        run: |
+          set -xeu
+          set -o pipefail
+          if [[ $github_ref == /refs/tags/* ]]
+          then
+            AGORA_VERSION="${github_ref#"/refs/tags/"}-${github_sha}"
+          else
+            AGORA_VERSION="$github_sha"
+          fi
+          echo "AGORA_VERSION=$AGORA_VERSION" >> $GITHUB_ENV
+
+        # This step will checkout the source with all the submodules
+      - name: Build and push
+        id: docker_build_and_push
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ github.event_name == 'push' }}
+          tags: bpfk/agora
+          build-args: |
+            AGORA_VERSION=${{ env.AGORA_VERSION }}
+
+      - name: Image digest
+        run: echo ${{ steps.docker_build.outputs.digest }}


ferencdg:
created the PR so I can test it, but not ready for review

.github/workflows/agora-docker.yml
```diff
@@ -0,0 +1,46 @@
...
+          GIT_DESCRIBE_OUT=`git describe --always`
+          echo "Version is $GIT_DESCRIBE_OUT"
+          echo "AGORA_VERSION=$GIT_DESCRIBE_OUT" >> $GITHUB_ENV
+
+      - name: Build and push
+        id: docker_build_and_push
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ github.event_name == 'push' }}
+          file: ./Dockerfile
```
Geod24: Should be the default, isn't it ?

.github/workflows/agora-docker.yml
```diff
@@ -0,0 +1,51 @@
...
+        if: github.event_name == 'push'
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+      - name: Check out repository
+        uses: actions/checkout@v2
+        with:
+          submodules: false
```
Geod24, line 25: That looks wrong

ferencdg: > That looks wrong

I need to check out the agora repo to execute the `git describe`, otherwise I get 'no .git repository found'. What is your suggestion?

.github/workflows/agora-docker.yml
```diff
@@ -0,0 +1,51 @@
...
+        run: |
+          set -xeu
+          set -o pipefail
+          AGORA_VERSION=HEAD
+          GIT_DESCRIBE_OUT=`git describe --always`
+          if [ -n "$GIT_DESCRIBE_OUT" ]
+          then
+            AGORA_VERSION=$GIT_DESCRIBE_OUT
+          fi
+          echo "AGORA_VERSION=$AGORA_VERSION" >> $GITHUB_ENV
```
Geod24: I would check if `${{ github.ref }}` starts with  `refs/tags/` and uses that (minus the prefix), or use `${{ github.sha }}`. Or just use sha as a first step.

Geod24: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#:~:text=Contexts%20are%20a%20way%20to,Contexts%20use%20the%20expression%20syntax.&text=Information%20about%20the%20workflow%20run,more%20information%2C%20see%20github%20context.

.github/workflows/agora-docker.yml
Geod24: I mean, this specific line, have `submodules: false`. Why ?

ferencdg: > I mean, this specific line, have `submodules: false`. Why ?

We don't need the submodules to execute git describe. Anyways if I try to put the Version number together from github env variables as you suggested, then I don't need to checkout anything.

Geod24: But you do need  the submodules to build Agora. You will always need to `checkout`, otherwise there's nothing to build.

.github/workflows/agora-docker.yml
ferencdg: I modified the code based on this, but the semantic of your suggested solution is different from mine. 

github.ref will only contain a tag if 'tag push' triggered the workflow, while in my solution git-describe will produce the latest tag regardless whether it was a 'tag push' or a 'branch push' or a PR.

.github/workflows/agora-docker.yml
ferencdg: it turned out the 'build docker' github action did checkout all the source, so this step can be removed

Geod24 (approved):

.github/workflows/agora-docker.yml
```diff
@@ -0,0 +1,55 @@
...
+on: [push, pull_request]
+
+jobs:
+  main:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+
+      - name: Set up Docker Buildx
```
AndrejMitrovic, line 12: Typo?
```suggestion
      - name: Set up Docker Build
```

Geod24: https://docs.docker.com/buildx/working-with-buildx/

.github/workflows/agora-docker.yml
Geod24: Perfect!

.github/workflows/agora-docker.yml
AndrejMitrovic: Talk about bad marketing..

PR #1554, opened 2021-01-26
Height: Add '+' and '+=' overloads

Geod24:
Clears up the clutter a bit.

### DIFF
--- a/source/agora/common/Types.d
+++ b/source/agora/common/Types.d
@@ -97,6 +97,19 @@ public struct Height
     {
         return Height(this.value++);
     }
+
+    /// Allow to offset an height by a fixed number
+    public Height opBinary (string op : "+") (ulong offset) const
+    {
+        return Height(this.value + offset);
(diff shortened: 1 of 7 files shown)


kchulj (approved):
Very nice

source/agora/common/Types.d
```diff
@@ -97,6 +97,19 @@ public struct Height
...
     }
+
+    /// Allow to offset an height by a fixed number
+    public Height opBinary (string op : "+") (ulong offset) const
+    {
+        return Height(this.value + offset);
+    }
+
+    /// Allow to offset an height by a fixed number
+    public ref Height opBinaryAssign (string op : "+=") (ulong offset) return
```
linked0, line 108: Is the `return` attribute right?

Geod24: Yes, because it returns a `ref` to itself (which allows chaining)

PR #1555, opened 2021-01-26
Test: Interrupted validator cycle

omerfirmak:
Validators that were unresponsive for a while should be able to catch up and
enroll again. Even if their enrollment expired completely.

### DIFF
--- a/source/agora/test/ValidatorRecurringEnrollment.d
+++ b/source/agora/test/ValidatorRecurringEnrollment.d
@@ -270,3 +270,63 @@ unittest
     assert(blocks[$ - 1].header.enrollments.length == 3);
     assert(blocks[$ - 2].header.enrollments.length == 3);
 }
+
+// Some nodes are interrupted during their validator cycles, they should
+// still manage to enroll when they are back online
+unittest
+{
+    TestConf conf = {
+        quorum_threshold : 66
+    };
+    auto network = makeTestNetwork(conf);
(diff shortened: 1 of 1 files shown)


source/agora/test/ValidatorRecurringEnrollment.d
```diff
@@ -270,3 +270,62 @@ unittest
...
+    nodes[$ - 1].ctrl.sleep(60.seconds, true);
+    nodes[$ - 2].ctrl.sleep(60.seconds, true);
+
+    network.generateBlocks(iota(GenesisValidators - 2),
+        Height(GenesisValidatorCycle - 1));
+
+    // Wake one up right before cycle ends
+    nodes[$ - 2].ctrl.sleep(0.seconds);
+    // Let it catch up
+    Thread.sleep(10.seconds);
```
AndrejMitrovic: There seem to be a lot of long unconditional sleep calls. Since a recent change, all the tests are running serially rather than in parallel. So this will further slow down the tests.

AndrejMitrovic: For now could we try to just reduce the time? 10 seconds seems a lot?

omerfirmak: Changed them with `expectBlock`

Geod24 (approved):

PR #1556, opened 2021-01-26
Use `std.traits.Parameters` in the test-suite to make `Nominator` / `getNominator` more DRY

Geod24:
Was working on some changes in the nominator and I kept on having errors everywhere.
So I was looking for ways to clear that up, when I noticed I must not be the only one, given the variety of errors in the parameters being passed :)

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -140,9 +140,9 @@ extern(D):
 
         Params:
             params = consensus params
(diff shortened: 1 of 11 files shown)


Geod24:
@omerfirmak : You might be interested in this :P

AndrejMitrovic (approved):
Nice work!

omerfirmak:
> @omerfirmak : You might be interested in this :P

I knew there was a better way :) Just did not know how :D

PR #1559, opened 2021-01-27
Validator: Only try to re-enroll for recent blocks

### LINKED ISSUE #1546: Eliminate unnecessary Enrollment requests during catch up.
If a Validator is N blocks behind the network, it will try to enroll N times during catch up and fail. Validator should not emit unnecessary requests when it knows that will fail for sure.

omerfirmak:
Fixes #1546

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -462,8 +462,10 @@ public class Validator : FullNode, API
         if (validators_changed || need_shuffle)
             this.regenerateQuorums(block.header.height);
 
-        // Re-enroll if our enrollment is about to expire
-        if (this.config.validator.recurring_enrollment)
+        // Re-enroll if our enrollment is about to expire and the block is recent enough
+        auto cur_offset = this.clock.networkTime() - this.params.GenesisTimestamp;
+        if (this.config.validator.recurring_enrollment &&
+            block.header.time_offset > cur_offset - 3 * this.params.BlockInterval.total!"seconds")
             this.checkAndEnroll(block.header.height);
 
         // note: may context switch, should be called last after quorums


source/agora/node/Validator.d
```diff
@@ -597,6 +620,6 @@ public class Validator : FullNode, API
         // Network needs Validators, see if we can enroll
         if (this.config.validator.recurring_enrollment &&
             msg == Ledger.InvalidConsensusDataReason.NotEnoughValidators)
-            this.checkAndEnroll(this.ledger.getBlockHeight());
+            this.checkAndEnroll(this.ledger.getBlockHeight(), true);
```
linked0: Can you explain why this calling has `true` for `dont_wait`?

omerfirmak: The case here is that we don't have enough validators for the next round. To get out of that situation faster we ignore the timeout.

source/agora/node/Validator.d
```diff
@@ -553,8 +557,27 @@ public class Validator : FullNode, API
...
+                block_height + 1);
+
+            // Disable existing timer
+            if (emitter)
+                emitter.stop();
+
+            // Network tests use 1 second block interval, timeout should be
+            // adjusted accordingly
+            version (unittest) enum EnrollmentDelay = 10.msecs;
+            else enum EnrollmentDelay = 10.seconds;
```
Geod24: That's a bit red flag

source/agora/node/Validator.d
```diff
@@ -533,11 +533,15 @@ public class Validator : FullNode, API
...
             block_height = Current block height
+            dont_wait = Immediately send the request.
 
     ***************************************************************************/
 
-    protected void checkAndEnroll (Height block_height) @safe
+    protected void checkAndEnroll (Height block_height, bool dont_wait = false) @safe
     {
+        // Enrollment emitter
+        static ITimer emitter;
```
Geod24: `static` variables within member functions are evil. Usually the unit of encapsulation is the object, but in this case it's the method, meaning that if we ever have 2 nodes running in the same thread, things will start to fail randomly. I'd recommend to steer clear from them.

source/agora/node/Validator.d
```diff
@@ -553,8 +557,27 @@ public class Validator : FullNode, API
...
+
+            // This timer will fire after EnrollmentDelay. Within that period,
+            // if we decide to request a different enrollment; next call to
+            // `checkAndEnroll` will disable this. This way we wont flood the
+            // network with invalid enrollment requests
+            () @trusted {
+                emitter = this.taskman.setTimer(dont_wait ? 0.msecs : EnrollmentDelay,
+                    {this.network.peers.each!(p => p.client.sendEnrollment(enrollment));}, 
+                    Periodic.No);
+            } ();
```
Geod24: So you essentially want to turn this into a "daemon" that gets periodically called and checks if it needs to enroll, right ? In which case taking the timer logic out of the function might be the best approach.

source/agora/node/Validator.d
ferencdg: I think we still try to retrieve 1024 blocks at once, so that probably needs to be changed too, otherwise 10secs might not be enough :)

source/agora/node/Validator.d
omerfirmak: Not really, it is a one shot timer. What I wanted is essentially a delayed and cancelable task.

source/agora/node/Validator.d
omerfirmak: I just checked, nodes ask for 1024 blocks but they only return `MaxBatchBlocksSent` which is 20.

omerfirmak:
Updated with the `time_offset` approach

AndrejMitrovic (approved):

Geod24 (approved):

PR #1560, opened 2021-01-27
Update vanity/system tests dub.json to new submodules

Geod24:
We're not using memutils not its botan-math dependency anymore,
so just remove them so they don't get pulled.
Additionally, the dub.json of vanity was changed
not to reference path, so it can be overridden.


kchulj:
Relates [faucet#30](https://github.com/bpfkorea/faucet/issues/30).
The submodules and `dub.json` there also need to be updated in order to build with Docker Hub.

Geod24:
I'm not sure I follow your comment. We don't build with docker hub anymore. And is there something else needed on Agora side ?

kchulj:
> I'm not sure I follow your comment. We don't build with docker hub anymore. And is there something else needed on Agora side ?

That was wrong, it is the GitHub Action. I meant, that it would be nice to have the same submodules on faucet as in agora.

Geod24:
@kchulj : So is this PR okay ?

kchulj (approved):
Yes LGTM.

PR #1562, opened 2021-01-27
Test: Ensure network can recover from no validator situation

### LINKED ISSUE #1545: Test for the cases where Enrollments don't get accepted immediately
With #1421, if an enrollment in the pool does not get immediately accepted in to the next block, it will no longer be valid. The Validator should keep trying until it successfully enrolls. The current implementation supports that, but we lack the tests for it.

omerfirmak:
Fixes #1545


linked0 (approved):
LGTM.
I think it needs retrying the ci builds.

Issue #1551, opened 2021-01-26
Ensure that nodes don't slash themselves


Geod24:
There should be a check that nodes don't externalize a block that would lead to their own slashing.
See #1537 for what can happen in such a case.

PR #1564, opened 2021-01-28
Not slash when there are not enough validators

linked0:
If there are not enough validators after some validators are slashed, we must not slash misbehaving validators. Instead, we can wait for the validators to reveal their pre-images again or other enrollment to be requested.

Part of #1551 

The original requirement of the issue is that validators do not slash themselves. The requirement could be handled in other PR.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -763,6 +763,8 @@ public class Ledger
             return fail_reason;
 
         size_t active_enrollments = enroll_man.getValidatorCount(expect_height);
+        assert(active_enrollments >= data.missing_validators.length);
+        active_enrollments -= data.missing_validators.length;
 
         if (data.enrolls.length + active_enrollments < Enrollment.MinValidatorCount)
             return InvalidConsensusDataReason.NotEnoughValidators;
--- a/source/agora/test/SlashingMisbehavingValidator.d
+++ b/source/agora/test/SlashingMisbehavingValidator.d
@@ -39,13 +39,27 @@ import geod24.Registry;
 // after enrollment.
 private class MissingPreImageEM : EnrollmentManager
 {
+    private shared bool* reveal_preimage;
+
     mixin ForwardCtor!();
+
+    ///
+    public this (Parameters!(EnrollmentManager.__ctor) args,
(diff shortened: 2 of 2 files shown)


linked0:
Ready for review.

source/agora/node/Ledger.d
```diff
@@ -760,7 +760,9 @@ public class Ledger
         if (auto fail_reason = this.getValidTXSet(data, tx_set))
             return fail_reason;
 
-        size_t active_enrollments = enroll_man.getValidatorCount(expect_height);
+        size_t active_enrollments = enroll_man.getValidatorCount(expect_height) -
+            data.missing_validators.length;
+        assert(active_enrollments >= 0);
```
AndrejMitrovic: Please use a signed type if you're comparing like this.

linked0: Updated. Ty.

linked0:
Rebased and ready for review.

Geod24:
This currently triggers in the live system, doesn't it ?

linked0:
> This currently triggers in the live system, doesn't it ?

No, it doesn't. It only happened in the local machine when a configuration has an old structure and can't be parsed correctly.

source/agora/node/Ledger.d
```diff
@@ -760,7 +760,9 @@ public class Ledger
         if (auto fail_reason = this.getValidTXSet(data, tx_set))
             return fail_reason;
 
-        size_t active_enrollments = enroll_man.getValidatorCount(expect_height);
+        long active_enrollments = enroll_man.getValidatorCount(expect_height) -
+            data.missing_validators.length;
+        assert(active_enrollments >= 0);
```
```suggestion
        size_t active_enrollments = enroll_man.getValidatorCount(expect_height);
        assert(active_enrollments >= data.missing_validators.length;
```

It's usually better to compare two values instead of subtract and compare two zero, since you don't risk overflowing when comparing.

linked0: Updated.

Geod24:
Needs a rebase, otherwise LGTM

linked0:
Rebased.

Geod24 (approved):

PR #1565, opened 2021-01-28
Fix #665 - Use Duration for ban time

### LINKED ISSUE #665: `BanManager.banFor` accepts negative values

Geod24:


kchulj (approved):

AndrejMitrovic (approved):

Geod24:
Rebased

PR #1576, opened 2021-01-29
Remove some single-use functions from NetworkManager

Geod24:
See each commit for details. This reduces dependencies between modules, ensuring that NetworkManager is used to send the same message to a set of peer, but not giving `NetworkManager` knowledge of every message type to send.
The "banned" messages were removed, as if a node is banned, it should not be in the list of peers to begin with.

### DIFF
--- a/source/agora/network/NetworkManager.d
+++ b/source/agora/network/NetworkManager.d
@@ -46,8 +46,6 @@ import agora.utils.InetUtils;
 import agora.utils.Log;
 import agora.utils.Utility;
 
-import scpd.types.Stellar_SCP;
-
 import vibe.http.common;
 import vibe.web.rest;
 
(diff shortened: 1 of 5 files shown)


Geod24:
Interestingly this consistently SEGV on Mac OSX.

source/agora/network/NetworkManager.d
```diff
@@ -1029,29 +1028,6 @@ public class NetworkManager
...
-
-    ***************************************************************************/
-
-    public void sendEnrollment (Enrollment enroll) @safe
-    {
-        foreach (ref node; this.peers[])
-        {
-            if (this.banman.isBanned(node.client.address))
-            {
-                log.trace("Not sending to {} as it's banned", node.client.address);
```
kchulj, line 1085: Is it ok if the `BanMan` log is removed?

Geod24: > The "banned" messages were removed, as if a node is banned, it should not be in the list of peers to begin with.

https://github.com/bpfkorea/agora/pull/1576#issue-563969595

AndrejMitrovic (approved):
LGTM.

AndrejMitrovic:
Perhaps the consistent failure is just a symptom of what happens when too many messages are sent, more than the Mac system it's running on can handle?

PR #1578, opened 2021-02-01
The leftover part of slashed UTXO becomes melting

### LINKED ISSUE #1487: The leftover part of slashed UTXO should be still frozen
In the recent implementation of the slashing protocol, if the node gets punished then parts of their funds will be sent to the commons budget, but the leftover part will be immediately unfrozen. 

https://github.com/bpfkorea/agora/blob/295b80b143b55b4d566620b4282c1710c0d96afc/source/agora/node/Ledger.d#L540-L549

This might not be the best idea. And we could discuss the policy about whether the leftover part either remains frozen or becomes a melting transaction. After establishing the policy about that, we will implement it as we determined.

linked0:
The leftover part of slashed UTXO either should remain frozen or become melting while the funds going to the CommonsBudget address is melted immediately. It would be better that the leftover part will be melting because the two outputs can be handled in a transaction.

Fixes #1487

### DIFF
--- a/source/agora/consensus/state/UTXOSet.d
+++ b/source/agora/consensus/state/UTXOSet.d
@@ -49,23 +49,19 @@ abstract class UTXOCache
         Params:
             tx = the transaction
             height = local height of the block
-            slashed = whether the transaction is for slashing
+            commons_budget = the CommonsBudget address
 
     ***************************************************************************/
 
     public void updateUTXOCache (const ref Transaction tx, Height height,
-        bool slashed = false) @safe
+        immutable PublicKey commons_budget) @safe
     {
         import std.algorithm : any;
 
-        // defaults to next block
-        Height unlock_height = height + 1;
+        bool melting = false;
 
-        // for payments of frozen transactions, it will melt after 2016 blocks
-        // If this is an unfreeze and slashing transaction, don't melt but
-        // instead set the unlock height to the next block as it will be sent to
-        // the commons budget.
-        if ((tx.type == TxType.Payment && !slashed)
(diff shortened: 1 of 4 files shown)


source/agora/consensus/state/UTXOSet.d
```diff
@@ -50,30 +50,27 @@ abstract class UTXOCache
             tx = the transaction
             height = local height of the block
             slashed = whether the transaction is for slashing
```
AndrejMitrovic: `slashed` was removed

linked0: Updated.

source/agora/consensus/state/UTXOSet.d
```diff
@@ -49,31 +49,27 @@ abstract class UTXOCache
...
 
     ***************************************************************************/
 
     public void updateUTXOCache (const ref Transaction tx, Height height,
-        bool slashed = false) @safe
+        immutable PublicKey commons_budget) @safe
     {
         import std.algorithm : any;
 
-        // defaults to next block
```
omerfirmak, line 61: Do you really need to introduce a new variable here? I think you can still use `unlock_height`.

Just set it to `height + 2016` for all frozen utxos and do something like;

``` d
auto utxo_value = UTXO(output.address != commons_budget ? 
           unlock_height : height + 1, tx.type, output); 
```

wouldnt that work?

linked0: Compact and nice code. I applied the code with some modifications.
Thanks.

Geod24 (approved):

PR #1579, opened 2021-02-01
Continue merging flash layer branch to v0.x.x (Part 2)

AndrejMitrovic:
Let's see if this works.

### DIFF
--- a/source/agora/script/Opcodes.d
+++ b/source/agora/script/Opcodes.d
@@ -36,10 +36,6 @@ import std.traits : EnumMembers;
 
 public enum OP : ubyte
 {
-    /// Executing this is an error and will cause script execution to fail.
-    /// Purposefully located first to default OPs to errors.
-    INVALID = 0x50,
-
     /// Used to encode a small data push to the stack (up to 75 bytes),
     /// may be used with `case PUSH_BYTES_1: .. case PUSH_BYTES_64:` syntax.
     PUSH_BYTES_1 = 0x01,
--- a/source/agora/script/Opcodes.d
+++ b/source/agora/script/Opcodes.d
@@ -122,6 +122,27 @@ public enum OP : ubyte
     /// script execution to fail if the signature is invalid
     VERIFY_SIG = 0x60,
 
+    /// Supports non-interactive threshold multi-signatures. A combination
+    /// of `N / M` is supported where `N <= M` and where the number of keys
+    /// and signatures does not exceed 5.
+    /// Pops an item from the stack which is the count of public keys that
+    /// should be read from the stack. It pops those public keys, and then
+    /// pops another item which is the number of required signatures.
+    /// Then it pops that many signatures from the stack. Then it walks through
+    /// all the public keys and validates the signatures in sequence. If a
+    /// key did not validate, it's removed from the internal list. If we've
+    /// ran out of keys and the number of valid signatures is less than
+    /// the required amount it means signature validation has failed.
+    /// It pushes the result of validation to the tack.
+    /// The keys and signatures must be placed in the same order on the stack.
+    /// Supports up to 5 public keys total as the signature validation is not
+    /// as efficient as alternative implementations such as MuSig.
+    CHECK_MULTI_SIG = 0x61,
+
+    /// Ditto, but instead of pushing the result to the stack it will cause the
+    /// script execution to fail if the signature is invalid
+    VERIFY_MULTI_SIG = 0x62,
+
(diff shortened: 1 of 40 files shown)


source/agora/script/Opcodes.d
```diff
@@ -61,67 +57,100 @@ public enum OP : ubyte
...
+    CHECK_SIG = 0x5F,
+
+    /// Ditto, but instead of pushing the result to the stack it will cause the
+    /// script execution to fail if the signature is invalid
+    VERIFY_SIG = 0x60,
+
+    /// Supports non-interactive threshold multi-signatures. A combination
+    /// of `N / M` is supported where `N <= M` and where the number of keys
+    /// and signatures does not exceed 5.
+    /// Pops an item from the stock which is the count of public keys that
```
omerfirmak:
```suggestion
    /// Pops an item from the stack which is the count of public keys that
```

AndrejMitrovic:
I attached gdb after the timeout and got the stack traces:

[thread 1](https://gist.github.com/AndrejMitrovic/3024ad9f6dca05d201626cf0e1ed49d9)
[thread 2](https://gist.github.com/AndrejMitrovic/7907b366bd6204708de8580a993a2498)
[thread 3](https://gist.github.com/AndrejMitrovic/f23a7fd246dd226a1bb8693d8dfc0c57)
[thread 4](https://gist.github.com/AndrejMitrovic/f75f95671aa0d361b4eca97cfac0e50c)
[thread 5](https://gist.github.com/AndrejMitrovic/b9620340e808e92970cb6097d11e7e8e)
[thread 6](https://gist.github.com/AndrejMitrovic/9847d2a16377c8bf1f8849ecdf035982)
[thread 7](https://gist.github.com/AndrejMitrovic/ac2418f97fc390a129e7de7075496bd0)
[thread 8](https://gist.github.com/AndrejMitrovic/b5d5e9983d0f4346d0aefc020e815680)
[thread 9](https://gist.github.com/AndrejMitrovic/b27bfb502bdbf430409225a1e7be3bf6)

I don't know what to make of it yet.

Geod24:
Which one segfaulted ?

AndrejMitrovic:
They didn't segfault, I attached GDB after a timeout. The unittests were running in the background.

AndrejMitrovic:
The issue is the tests never finish. And there's no logging either.

AndrejMitrovic:
Maybe I timed out too soon though. Could be that the tests just run a long time. I'll give this another try.

AndrejMitrovic:
> Maybe I timed out too soon though. Could be that the tests just run a long time. I'll give this another try.

So far it's been running for 40 minutes, way longer than the previous max at around ~20. I'll try to do another stack trace but after a longer timeout (maybe 30 mins).

AndrejMitrovic:
Running it for 30 minutes gets me a stack trace of only 3 threads:

[thread 1](https://gist.github.com/AndrejMitrovic/3a998eb06b6a59a97fbfd933433b1cd8)
[thread 2](https://gist.github.com/AndrejMitrovic/36312938190b20c50e98296a077d3e2c)
[thread 3](https://gist.github.com/AndrejMitrovic/020d3afe958632380ba3a305b6174f70)

AndrejMitrovic:
I'm trying to disable some tests individually now.

omerfirmak:
I don't know if you guys were able to reproduce this on your local machines. Limiting the unittests to only 2 cores, similar to the CI, reproduced the issue on my machine.

AndrejMitrovic:
I didn't know the CI runs on only two cores.. can't we increase this? It doesn't resolve the underlying problem but it will at least unblock us.

AndrejMitrovic:
Ah did you mean just using `dsinglethreaded`?

omerfirmak:
No, I meant actually forcing tests to run on only 2 cores. I used `taskset` on linux.

AndrejMitrovic:
It looks like the test-suite now passes! That means we can begin merging all the flash work into v0.x.x mainline.

However I need to fix up the commits so they're atomic. Some of the commits don't compile on their own.

AndrejMitrovic:
@omerfirmak I added a commit changing all stdout printouts to log calls. Also added these:

```
scope (failure) factory.printLogs();
```

The printouts will no longer appear in real-time or be interleaved, but they will at least be grouped per node and it will avoid stdout spam for everyone else.

AndrejMitrovic:
@omerfirmak I'll give you the honors to merge.

omerfirmak (approved):

PR #1580, opened 2021-02-01
time_t should not be used directly

### LINKED ISSUE #1451: time_t should not be used directly
1. time_t is uint on windows while ulong on linux
2. because of the above, we have to declare the timestamp in the block header as ulong directly, as the wire format of the blockheader should be the same accross all OS
3. all the Clock interface methods use time_t
4. there is a constant need for converting between time_t and ulong becuase of point 2 and 3, this happens for example when we want to initialize a timer in tests with time from the Genesis block's timestamp

I think, we shouldn't be using time_t in agora, we should have our own time type let's say
alias agora_time = ulong

when time_t is fixed to be the same on each platform, we should change 
alias agora_time = time_t

or just grep/replace agora_time with time_t

ferencdg:
fixes: #1451

### DIFF
--- a/source/agora/common/Types.d
+++ b/source/agora/common/Types.d
@@ -17,6 +17,9 @@ import agora.common.crypto.Key;
 
 import geod24.bitblob;
 
+/// Represents a specific point in time, it should be changed to time_t
+/// after time_t became platform independent
+public alias TimePoint = ulong;
 
 /// An array of const characters
 public alias cstring = const(char)[];
--- a/source/agora/network/Clock.d
+++ b/source/agora/network/Clock.d
@@ -48,6 +48,7 @@
 
 module agora.network.Clock;
 
+import agora.common.Types : TimePoint;
 import agora.utils.Log;
 
 import core.stdc.time;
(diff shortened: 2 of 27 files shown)


Geod24:
Your local copy seems out of date, I changed it to `Duration` a few days ago.

ferencdg:
> Your local copy seems out of date, I changed it to `Duration` a few days ago.

Right, I saw that when github complained about the merge issue. I have other commits in this PR that are still relevant.

source/agora/network/Clock.d
```diff
@@ -106,13 +107,10 @@ public class Clock
 
     ***************************************************************************/
 
-    public time_t networkTime () @safe nothrow
+    public TimePoint networkTime () @safe nothrow
     {
-        // https://issues.dlang.org/show_bug.cgi?id=21134
-        import std.conv : to;
-        // not a problem if #21134 is fixed before year 2038
         scope (failure) assert(0);
```
AndrejMitrovic: You can remove this line too. It was there for the `std.conv` call.

Geod24:
This needs a rebase

ferencdg:
> This needs a rebase

rebased

source/agora/common/Types.d
```diff
@@ -17,6 +17,9 @@ import agora.common.crypto.Key;
 
 import geod24.bitblob;
 
+/// Represents a specific point in time, it should be changed to TimePoint
+/// after TimePoint became platform independent
```
Geod24: I'm not sure I understand that comment ?

Geod24 (approved):
LGTM, just a question about the comment which seems unclear.

Geod24:
>  source/agora/test/SlashingMisbehavingValidator.d(193,16): Error: undefined identifier `time_t`

For ref I just rebased on `v0.x.x` again

ferencdg:
> LGTM, just a question about the comment which seems unclear.

rebased again, also fixed the comment(it was messed up by VSCode find and replace)

PR #1584, opened 2021-02-01
BlockStorage: Remove boolean-returning functions and embrace Exception

Geod24:
It's pretty obvious that the Exception-less API has been completely misused, leading to a lot of places just using `assert(0)`.
In hindsight, it's also relatively pointless to optimize against Exceptions when we perform disk IO and (de)serialization.
So clean up the API by returning `Block` and using `Exception`.

### DIFF
--- a/source/agora/node/BlockStorage.d
+++ b/source/agora/node/BlockStorage.d
@@ -366,8 +366,8 @@ public class BlockStorage : IBlockStorage
                 if (findex == this.file_index)
                     return true;
 
-                if (this.is_saving && !this.writeChecksum())
-                    assert(0);
+                if (this.is_saving)
+                    this.writeChecksum();
 
                 this.release();
             }
@@ -474,8 +474,7 @@ public class BlockStorage : IBlockStorage
 
         this.length += size_t.sizeof + block_size;
 
-        if (!this.writeChecksum())
-            throw new Exception("BlockStorage: Failed to write checksum");
+        this.writeChecksum();
 
         // add to index of height
         this.height_idx.insert(
@@ -554,9 +553,7 @@ public class BlockStorage : IBlockStorage
                 throw new Exception("failed to write block data size");
 
             this.length += size_t.sizeof + block_size;
-
-            if (!this.writeChecksum())
-                assert(0);
+            this.writeChecksum();
 
             return true;
         }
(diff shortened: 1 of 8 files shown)


Geod24:
Should be fixed now

Geod24:
Ready

source/agora/node/BlockStorage.d
```diff
@@ -70,16 +70,15 @@ public interface IBlockStorage
 
         Read the last block from the storage.
 
-        Params:
-            block = The last block
-
         Returns:
-            Returns true if success, otherwise returns false.
+            The blocks that has been read.
```
kchulj:
```suggestion
            The block that has been read.
```

kchulj (approved):

source/agora/node/BlockStorage.d
Geod24: Ty, rebased and fixed.

source/agora/node/BlockStorage.d
```diff
@@ -572,7 +572,14 @@ public class BlockStorage : IBlockStorage
     {
         try
         {
-            this.readBlockAtPosition(block, this.height_idx.findBlockPosition(height));
+            const block_pos = this.height_idx.findBlockPosition(height);
+            if (height != 0 && block_pos == 0)
```
omerfirmak: Can `findBlockPosition` fail for `height = 0`?

source/agora/node/BlockStorage.d
```diff
@@ -956,28 +953,15 @@ public class BlockStorage : IBlockStorage
         Read the file data, calculate checksum,
         and write checksum to file at start point
 
-        Returns:
```
omerfirmak, line 957: Does it need a `Throws`?

source/agora/node/BlockStorage.d
Geod24: No, `height = 0` means `pos = 0`

source/agora/node/BlockStorage.d
Geod24: Hum, it doesn't throw directly but is not `nothrow`... I plan to inline it in the future, and I don't think it's that important for private functions.

omerfirmak (approved):

PR #1586, opened 2021-02-02
Fix 1549: Merge ConsensusConfig and ConsensusParams

### LINKED ISSUE #1549: Merge ConsensusConfig and ConsensusParams
Doesn't quite make sense to have them separate. Also `ConsensusParams` should *probably* be a struct.

Geod24:
Remove some duplication. Last commit is a bonus that I spotted while doing this.
Note that currently `Genesis`, `BlockInterval[Seconds]` and `CommonsBudgetAddress` are not part of `ConsensusConfig`.
`Genesis` because it's configured indirectly (through `testing`), `BlockInterval[Seconds]` because it feels very much like a node params (and need further investigation), and lastly `CommonsBudgetAddress` **should** probably be in it, but that's another issue.


AndrejMitrovic (approved):

Geod24:
Reverted some changes to the test suite that broke it.

Issue #1221, opened 2020-09-20
Split Schnorr code into its own repository


Geod24:
The Schnorr code is pretty simple and self contained, and could be extracted in its own repository.
This could allow us to get reviews on it as well.

PR #1587, opened 2021-02-02
ECC: Remove redundant 'serialize' method

Geod24:
The serializer recognizes this struct as a value type and thus there
is no need to provide our own 'serialize' method.

Required for #1221


AndrejMitrovic (approved):

Geod24:
Hold on, we might actually go a different way

Geod24:
Nevermind, this would be good to go regardless.

PR #1589, opened 2021-02-02
FullNode: Introduce a simpler 'getBlock' function

Geod24:
This can be useful for clients, as it does not rely on a query parameter,
unlike `getBlocksFrom`, which makes it unusable in certain contexts.

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -146,6 +146,24 @@ public interface API
 
     /***************************************************************************
 
+        Expose blocks as a REST collection
+
+        API:
+            GET /blocks/:height
+
+        Params:
+            _height = The height of the block to return
+
+        Returns:
+            The block at height `_height`, or throw an `Exception` (404).
+
+    ***************************************************************************/
+
+    @path("/blocks/:height")
+    public const(Block) getBlock (ulong _height);
+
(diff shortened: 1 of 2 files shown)


AndrejMitrovic (approved):
LGTM. Wanted this!

Why `_height` though?

source/agora/api/FullNode.d
```diff
@@ -144,6 +144,24 @@ public interface API
...
+
+        Params:
+            _height = The height of the block to return
+
+        Returns:
+            The block at height `_height`, or throw an `Exception` (404).
+
+    ***************************************************************************/
+
+    @path("/blocks/:height")
```
linked0, line 162: Doesn't it need to be `/block/:height`?

AndrejMitrovic: I think he's trying to re-use the same end-point with an optional parameter.

Geod24:
> Why `_height` though?

https://vibed.org/api/vibe.web.rest/ see "Manual endpoint specification"

source/agora/api/FullNode.d
Geod24: Yep, also there's a convention in REST interfaces that collections contain an `s`.

linked0 (approved):

PR #1591, opened 2021-02-02
Set: Overload opSlice()

omerfirmak:
This is needed to use some `std.iteration` templates like `map()`

### DIFF
--- a/source/agora/common/Set.d
+++ b/source/agora/common/Set.d
@@ -116,6 +116,12 @@ public struct Set (T)
         return SetT.from!(SetT)(
             iota(length).map!(_ => deserializeFull!(T)(dg, opts)));
     }
+
+    /// Provide a range to iterate on
+    public typeof(this._set.byKey) opSlice ()
+    {
+        return this._set.byKey;
+    }
 }
 
 /// fill the buffer with the set's keys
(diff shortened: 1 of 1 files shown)


source/agora/common/Set.d
```diff
@@ -116,6 +116,12 @@ public struct Set (T)
         return SetT.from!(SetT)(
             iota(length).map!(_ => deserializeFull!(T)(dg, opts)));
     }
+
+    /// Provde a Range to iterate on
```
Geod24:
```suggestion
    /// Provide a range to iterate on
```

Geod24 (approved):

Geod24 (approved):

source/agora/common/Set.d
```diff
@@ -222,3 +228,10 @@ unittest
     arr.dropIndex(1);
     assert(arr == [1, 3]);
 }
+
+unittest
+{
+    auto set = Set!uint.from([2, 4, 6, 8]);
+    auto _2xset = set[].map!(elem => tuple(elem, 2*elem));
```
linked0, line 235: Do we need to use `_` for the name of a local variable?

omerfirmak: No, its because variable names cant start with numbers and I am lazy :)

linked0 (approved):
LGTM

source/agora/common/Set.d
```diff
@@ -116,6 +116,12 @@ public struct Set (T)
         return SetT.from!(SetT)(
             iota(length).map!(_ => deserializeFull!(T)(dg, opts)));
     }
+
+    /// Provide a range to iterate on
+    public typeof(this._set.byKey) opSlice ()
```
AndrejMitrovic, line 121: By the way, you can use `auto` in the future.

PR #1595, opened 2021-02-03
Fix a deprecation and a warning on Windows

Geod24:
Rather simply, just saw those in a build log while working on something else.


TrustHenry (approved):
LGTM

PR #1598, opened 2021-02-04
BlockHeader: Exchange signature and validators field

Geod24:
This will make the binary layout on disk more predictable,
as Signature and offset will now always be at a fixed location,
and allow us to update them in place.


hewison-chris (approved):
I agree this is better. Although the bitfield remains the same size for a certain block once it is written to disk, as it depends on the number of enrolled validators, the signature and bitfield offsets will be known for all heights with your change.

kchulj (approved):

PR #1599, opened 2021-02-04
BlockStorage: Remove nothrow on catch-all functions

Geod24:
Since the interface needs not be nothrow, just remove those intermediate casts.


hewison-chris (approved):
LGTM

kchulj (approved):
LG

PR #1600, opened 2021-02-04
Ensure that nodes don't slash themselves

### LINKED ISSUE #1551: Ensure that nodes don't slash themselves

linked0:
This code terminates a node that is trying to slash itself, which is decided by the discussion with @Geod24.
The `agora.test.EnrollDifferentUTXOs` network test in the second commit fails without the first commit which removes the statements saving enrollment information into the `EnrollmentManager` before consensus.

Fixes #1551

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -823,6 +823,34 @@ public class EnrollmentManager
         return PublicKey(this.key_pair.V[]);
     }
 
+    /***************************************************************************
+
+        Get the index of this node's enrollment in the sorted enrolled UTXOs
+
+        Returns:
+            the index of the enrollment, or ulong.max if this node is not
+            enrolled as a validator.
+
+    ***************************************************************************/
+
+    public ulong getIndexOfEnrollment () @safe nothrow
+    {
+        if (this.enroll_key == Hash.init)
+            return ulong.max;
+
+        Hash[] utxo_keys;
+        if (!this.validator_set.getEnrolledUTXOs(utxo_keys))
+            assert(0);
+
+        foreach (idx, key; utxo_keys)
+            if (key == this.enroll_key)
+                return idx;
+
+        log.fatal("Index of the node's enrollment not found. Enrollment: {}",
+            this.enroll_key);
+        assert(0);
+    }
+
     /// Returns: true if this validator is currently enrolled
     public bool isEnrolled (scope UTXOFinder finder) nothrow @safe
     {
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -415,9 +415,6 @@ public class EnrollmentManager
 
     public Enrollment createEnrollment (in Hash utxo, Height height) @safe nothrow
     {
-        // K, frozen UTXO hash
-        this.enroll_key = utxo;
-
         // X, final seed data and preimages of hashes
         const seed = this.cycle.getPreImage(this.key_pair.v, height);
         const enroll = makeEnrollment(
--- a/source/agora/consensus/SlashPolicy.d
+++ b/source/agora/consensus/SlashPolicy.d
@@ -233,6 +233,9 @@ public class SlashPolicy
     private bool hasRevealedPreimage (in Height height, in Hash utxo_key)
         @safe nothrow
     {
+        if (utxo_key == this.enroll_man.getEnrollmentKey())
+            return true;
+
         auto preimage = this.enroll_man.getValidatorPreimage(utxo_key);
         auto enrolled = this.enroll_man.getEnrolledHeight(preimage.enroll_key);
         assert(height >= enrolled);
@@ -264,6 +267,13 @@ public class SlashPolicy
         if (!this.enroll_man.getEnrolledUTXOs(keys) || keys.length == 0)
         {
             log.fatal("Could not retrieve enrollments / no enrollments found");
+            assert(0);
+        }
+
+        auto enroll_index = this.enroll_man.getIndexOfEnrollment();
+        if (enroll_index != ulong.max && !missing_validators.find(enroll_index).empty)
+        {
+            log.fatal("The node is slashing itself.");
             assert(0);
         }
 
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -808,6 +808,12 @@ extern(D):
                 return ValidationLevel.kInvalidValue;
             }
 
+            if (this.ledger.checkSelfSlashing(data))
+            {
+                log.error("validateValue(): Slasing itself");
+                return ValidationLevel.kInvalidValue;
+            }
+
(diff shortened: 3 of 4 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -823,6 +823,33 @@ public class EnrollmentManager
...
+
+    ***************************************************************************/
+
+    public ulong getIndexOfEnrollment () @safe nothrow
+    {
+        if (this.enroll_key == Hash.init)
+            return ulong.max;
+
+        Hash[] utxo_keys;
+        assert(this.validator_set.getEnrolledUTXOs(utxo_keys));
```
```
if (!this.validator_set.getEnrolledUTXOs(utxo_keys));
    assert(0);
```

Because currently the assert you wrote here will disappear in `-release` mode, so the function call will not exist.

AndrejMitrovic: assert(0) is a special-case: It's always there, even in `-release` mode.

linked0: I got it. Will update.
Thanks.

linked0:
Would you please review this? It's a too late review request but it's not complex and implemented as what @Geod24 and I have discussed already.

source/agora/test/SlashingMisbehavingValidator.d
```diff
@@ -154,6 +154,9 @@ unittest
     utxo_set.peekUTXO(frozen_hash, utxo);
     assert(utxo.type == TxType.Freeze);
 
+    // node[5] becomes unresponsive in order not to slash itself.
```
Geod24: I'm not sure I follow this comment. Will the unittest crash if the sleep isn't here ?

linked0: Yes, it will because I have added the assertion in case of slashing itself.
The statement makes the `node[5]` not to nominating a consensus data that contains the `missing_validators` having the index of the node itself.

Geod24: So the assertion *does* trigger. It should never trigger. When we discussed this, I suggested to add the assertion if the bug was fixed, but it seems it is not.

linked0: Sorry for the misunderstanding. I will add the code that does not make a consensus data if a node itself doesn't reveal its pre-images.

source/agora/consensus/SlashPolicy.d
```diff
@@ -105,7 +105,8 @@ public class SlashPolicy
 
         foreach (idx, utxo_key; keys)
         {
-            if (!this.hasRevealedPreimage(height, utxo_key))
+            if (utxo_key != this.enroll_man.getEnrollmentKey() &&
+                !this.hasRevealedPreimage(height, utxo_key))
```
Geod24: How can `hasRevealedPreimage` return true for itself ? That shouldn't happen.

source/agora/consensus/SlashPolicy.d
```diff
@@ -267,6 +268,13 @@ public class SlashPolicy
             assert(0);
         }
 
+        auto enroll_index = this.enroll_man.getIndexOfEnrollment();
+        if (enroll_index != ulong.max && !missing_validators.find(enroll_index).empty)
+        {
+            log.fatal("The node is slashing itself.");
+            assert(0);
+        }
```
Geod24, line 278: Isn't this `checkSelfSlashing` ?

source/agora/consensus/SlashPolicy.d
```diff
@@ -279,6 +287,32 @@ public class SlashPolicy
...
+    public bool checkSelfSlashing(in uint[] missing_validators) @safe nothrow
+    {
+        auto enroll_index = this.enroll_man.getIndexOfEnrollment();
+        if (enroll_index != ulong.max &&
+            !missing_validators.find(enroll_index).empty)
+        {
+            return true;
+        }
+        return false;
+    }
```
Geod24: This should probably be in `Ledger` since it's only accessed from there.

linked0: Moved

source/agora/consensus/SlashPolicy.d
linked0: Yes, it is.
Changed it to call `checkSelfSlashing` and moved the calling into the ledger.

source/agora/consensus/SlashPolicy.d
linked0: I don't follow you. This call is getting whether the requested pre-images exists in the `ValidatorSet`.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -823,6 +823,34 @@ public class EnrollmentManager
...
+
+        foreach (idx, key; utxo_keys)
+            if (key == this.enroll_key)
+                return idx;
+
+        log.fatal("Index of the node's enrollment not found. Enrollment: {}",
+            this.enroll_key);
+        assert(0);
+    }
+
```
hewison-chris, line 853: I think this index is already known by the validator without creating a new function:-
`public ulong getIndexOfValidator (const Height height, Point K) const nothrow @safe`

hewison-chris (changes requested):
Looks good but a small change request.

source/agora/consensus/SlashPolicy.d
Geod24: Yes, and it would make the code much simpler if `hasRevealedPreimage` would always returns `true` for `utxo_key == this.enroll_man.getEnrollmentKey()`. Because the node *knowns* the pre-images, so it should consider them "revealed".

linked0: Got it, Thanks.

source/agora/consensus/EnrollmentManager.d
linked0: The index of `getIndexOfValidator` is not the same as the index of `getIndexOfEnrollment` because the former is based on the order of public keys.
Thanks.

source/agora/test/SlashingMisbehavingValidator.d
linked0: Removed

linked0:
Ready for review.

source/agora/consensus/EnrollmentManager.d
hewison-chris: Perhaps we should be using the same sorting criteria for validators for both block signing and slashing. I think it is easier if we keep it consistent. Perhaps this can be done in a seperate PR though.

hewison-chris: Raised an issue to update the validator index to be consistent.

hewison-chris (approved):
LGTM

Geod24:
Can you rebase ? Otherwise LGTM.

linked0:
> Can you rebase ? Otherwise LGTM.

OK. I will.

PR #1602, opened 2021-02-04
Don't allow Coinbase only blocks

### LINKED ISSUE #1601: Coinbase only TX sets should not validate
Currently there are no rules against network creating blocks consisting of Coinbase TXs only. This is probably not ideal. We should both not create such nominations and invalidate any nominations/blocks that only include Coinbase TXs.

omerfirmak:
Fixes #1601

### DIFF
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -137,12 +137,17 @@ public string isInvalidReason (const ref Block block, Height prev_height,
     if (!block.txs.isSorted())
         return "Block: Transactions are not sorted";
 
+    bool only_coinbase = true;
     foreach (const ref tx; block.txs)
     {
+        only_coinbase &= tx.type == TxType.Coinbase;
         if (auto fail_reason = VTx.isInvalidReason(tx, findUTXO,
             block.header.height, checkFee))
             return fail_reason;
     }
+
+    if (only_coinbase)
+        return "Block: Must contain other transactions than Coinbase";
 
     auto expected_cb_txs = getCoinbaseTX(block.txs,
         block.header.missing_validators);
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -659,8 +659,10 @@ public class Ledger
         }
 
         const pre_cb_len = data.tx_set.length;
-        data.tx_set ~= this.getCoinbaseTX(tot_fee, tot_data_fee,
-            data.missing_validators).map!(tx => tx.hashFull()).array;
+        // Dont append a CB TX to an empty TX set
+        if (pre_cb_len > 0)
+            data.tx_set ~= this.getCoinbaseTX(tot_fee, tot_data_fee,
+                data.missing_validators).map!(tx => tx.hashFull()).array;
         // No more than 1 CB per block
         assert(data.tx_set.length - pre_cb_len <= 1);
     }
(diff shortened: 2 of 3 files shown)


source/agora/node/Ledger.d
```diff
@@ -1982,3 +1990,43 @@ unittest
             mpv_stake.output.address).array.length == 0);
     }
 }
+
+// Coinbase only ConsensusData and blocks
```
linked0: Please leave a little more detailed description of what the situation is and what the expectation is.

source/agora/test/Fee.d
```diff
@@ -219,3 +219,37 @@ unittest
     auto bad_node = nodes[0];
     assert(bad_node.getBlockHeight() == Height(0));
 }
+
+unittest
```
linked0, line 225: Please leave the description.

source/agora/node/Ledger.d
```diff
@@ -1982,3 +1990,43 @@ unittest
...
+    import agora.consensus.data.genesis.Test;
+    import std;
+    ConsensusConfig config = { validator_cycle: 20, payout_period: 1 };
+    auto params = new immutable(ConsensusParams)(GenesisBlock,
+        CommonsBudget.address, config);
+
+    const(Block)[] blocks = [ GenesisBlock ];
+    scope ledger = new TestLedger(WK.Keys.NODE2, blocks, params);
+
+    Amount per_tx_fee = Amount.UnitPerCoin;
```
linked0: The `per_tx_fee` doesn't need to be introduced.

linked0 (approved):
Good work.
You should add comments to the commit.

kchulj (approved):
The code LGTM.

source/agora/consensus/validation/Block.d
```diff
@@ -137,13 +137,18 @@ public string isInvalidReason (const ref Block block, Height prev_height,
...
     foreach (const ref tx; block.txs)
     {
+        only_coinbase &= tx.type == TxType.Coinbase;
         if (auto fail_reason = VTx.isInvalidReason(tx, findUTXO,
             block.header.height, checkFee))
             return fail_reason;
     }
 
+    if (only_coinbase)
+        return "Block: Coinbase only block";
```
Geod24:
```suggestion
        return "Block: Must contain other transactions than Coinbase";
```

We should always aim to tell what is wrong, not just the state. As a user, would you rather see "Error: Block has 1337 transactions" or "Error: Number of transaction per block must be a power of 2, not 1337" ?

source/agora/node/Ledger.d
```diff
@@ -718,6 +720,7 @@ public class Ledger
         NoTransactions = "Transaction set doesn't contain any transactions",
         NotEnoughValidators = "Enrollment: Insufficient number of active validators",
         MayBeValid = "May be valid",
+        OnlyCoinbaseTX = "Transaction set only includes Coinbase transactions",
```
```suggestion
        OnlyCoinbaseTX = "Transaction set only includes a Coinbase transaction",
```

We can only have one, right ?

source/agora/test/Fee.d
```diff
@@ -219,3 +219,39 @@ unittest
...
+
+    Transaction[] txs;
+    // create enough tx's for a single block
+    txs = blocks[$ - 1].spendable().map!(txb => txb
+        .deduct(Amount.UnitPerCoin).sign()).array();
+    // Send a single TX with fees to a node
+    node_1.putTransaction(txs[0]);
+    network.expectBlock(Height(1), blocks[0].header);
+    // The fees from the last block should not trigger a new block creation
+    assertThrown!AssertError(network.expectBlock(Height(2), blocks[0].header));
```
Geod24: We should *never* test for `AssertError`. The way to do it would be to instrument a validator.

omerfirmak:
Rebased and all comments addressed.

PR #1606, opened 2021-02-08
Refactor `getRandomSeed` function

linked0:
There are two `getRandomSeed` functions in the `EnrollmentManager` and `SlashPolicy` classes, which is unnecessary and confusing. We should only use the function in the `EnrollmentManager`. So this PR refactor the functions.


linked0:
This will be ready for review today after some more refinement.

Geod24:
This has been superseded by #2084

PR #1607, opened 2021-02-08
collectBlockSignature: Only pass the hash, not the full block

Geod24:
This method only cares about the block's hash, not its content,
so avoid a superfluous copy.


linked0 (approved):
LGTM

kchulj (approved):
LGTM

Geod24:
🤦 

I should be hashing the header

PR #1610, opened 2021-02-08
Implement Flash Layer with multi-hop payments

AndrejMitrovic:
Revival of https://github.com/bpfkorea/agora/pull/1415 with additional features.

As discussed with @omerfirmak there's still some wrong assumptions made in the code. These will have to be fixed, either in this PR or in a fixup later.

Also there is an occasional test failure, there's some timing assumptions that fail sometimes.

I plan to move the stuff out of `test/Flash.d` and into a separate class. My goal is for a FullNode / ValidatorNode to also be a FlashNode if they have any funding transactions on the blockchain for which they are a part of. In this way we get some free features like getting the latest gossiped block, transaction validation, etc. 

But I want to focus on the internals first before it's capable to run it in release mode.

### DIFF
--- a/source/agora/flash/API.d
+++ b/source/agora/flash/API.d
@@ -1,0 +1,292 @@
+/*******************************************************************************
+
+    Contains the Flash API.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.flash.API;
+
+import agora.common.Amount;
+import agora.common.crypto.ECC;
+import agora.common.Types;
+import agora.flash.Config;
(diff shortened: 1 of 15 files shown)


omerfirmak:
> My goal is for a FullNode / ValidatorNode to also be a FlashNode if they have any funding transactions on the blockchain for which they are a part of

Do we have any intentions to have thin Lightning Nodes as well, in the future?

AndrejMitrovic:
> Do we have any intentions to have thin Lightning Nodes as well, in the future?

I suppose that's not a bad idea. There could be an abstract flash node class, with implementations in the `LightFlashNode` and `FullNode` / `Validator` classes. Maybe like that..

omerfirmak (approved):
Lets just merge these.

AndrejMitrovic:
Fastest review in the west.

PR #1611, opened 2021-02-08
Flash: Lightning Network in-memory representation

### LINKED ISSUE #1583: Flash Layer: Network topology in-memory representation
This is a sub issue of #1528 

Problem of routing a payment through lightning network is essentially a graph problem. We should have an efficient in-memory representation of the network that a Lightning Node is aware of. So that we can traverse and build paths along the network efficiently.

omerfirmak:
This commit adds a graph representation of the lightning network and additional
book keeping needed to be done by the FlashNode.

It is missing removal of closed channels, does not take channel fees and balances
in to account.

Fixes #1583

### DIFF
--- a/source/agora/flash/Network.d
+++ b/source/agora/flash/Network.d
@@ -1,0 +1,334 @@
+/*******************************************************************************
+
+    Contains in-memory representation of Lightning Network topology
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.flash.Network;
+
+import agora.flash.Config;
+import agora.flash.Channel;
+import agora.flash.Route;
+
+import agora.common.crypto.ECC;
+import agora.common.Hash;
+import agora.common.Set;
+import agora.common.Amount;
+
+private struct NetworkNode
+{
+    /// Channels that this NetworkNode is a part of
+    private Set!Hash[Point] channels;
+}
+
+public class Network
+{
+    /// Nodes
+    private NetworkNode[Point] nodes;
+
+    /// Delegate to lookup Channels by their ID
+    private Channel delegate (Hash chan_id) @safe nothrow lookupChannel;
+
+    /// Ctor
+    this (Channel delegate (Hash chan_id) @safe nothrow lookupChannel)
+    {
+        this.lookupChannel = lookupChannel;
+    }
+
+    /***************************************************************************
+
+        Add a Channel to the network
+
+        Params:
+            chan_conf = ChannelConfig of the Channel to be added
+
+    ***************************************************************************/
+
+    public void addChannel (const ref ChannelConfig chan_conf) @safe nothrow
+    {
+        const funder_pk = chan_conf.funder_pk;
+        const peer_pk = chan_conf.peer_pk;
+
+        this.addChannel(funder_pk, peer_pk, chan_conf);
+        this.addChannel(peer_pk, funder_pk, chan_conf);
+    }
+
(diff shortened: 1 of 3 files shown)


source/agora/flash/Node.d
```diff
@@ -80,6 +81,9 @@ public abstract class FlashNode : FlashAPI
     /// any intermediaries.
     protected Hash[Hash] secrets;
 
+    /// Lightning Network topology
```
AndrejMitrovic: I don't think we should use the Lightning network naming in our code. We want to differentiate our solution to the existing LN. Just Flash network is fine.

source/agora/flash/Network.d
```diff
@@ -0,0 +1,331 @@
...
+
+import agora.common.crypto.ECC;
+import agora.common.Hash;
+import agora.common.Set;
+import agora.common.Amount;
+
+private struct NetworkNode
+{
+    /// Channels that this NetworkNode is a part of
+    private Set!Hash[Point] channels;
```
AndrejMitrovic, line 28: More specifically the key is a counter-party, and we can have multiple channels with another counter-party so that's why this is a set. Was that the intention?

source/agora/flash/Network.d
```diff
@@ -0,0 +1,331 @@
...
+    /***************************************************************************
+
+        Remove multiple Channels from the network
+
+        Params:
+            chns = List of ChannelConfig
+
+    ***************************************************************************/
+
+    public void removeChannel (const ref ChannelConfig[] chns) @safe nothrow
```
AndrejMitrovic: Inconsistent naming with `addChannels`. Should be `removeChannels` then.

source/agora/flash/Network.d
```diff
@@ -0,0 +1,331 @@
...
+            from_pk = Source node public key
+            to_pk = Destination node public key
+            ignore_chans = Channels to ignore
+
+        Returns:
+            If found, path from source to destination
+
+    ***************************************************************************/
+
+    public Hop[] getPath (Point from_pk, Point to_pk,
```
AndrejMitrovic: I would name this `getPaymentPath`. We will add the `Amount` later, so it will make more sense.

Or you could choose to rename it later when the Amount is added? I think that's fine too.

source/agora/flash/Network.d
```diff
@@ -0,0 +1,331 @@
...
+        {
+            fees[pk] = pk == from_pk ? Amount(0) : Amount.MaxUnitSupply;
+            unvisited.put(pk);
+        }
+
+        while (unvisited.length > 0)
+        {
+            import std.algorithm;
+            import std.array;
+
```
AndrejMitrovic, line 181:
```suggestion
            // Pick the node with the smallest fee
```

source/agora/flash/Network.d
```diff
@@ -0,0 +1,331 @@
...
+                    if (chans.length == 0)
+                        continue;
+
+                    // TODO: Pick the channel with the lowest fee from the list
+                    // TODO: Verify that the channel is open
+                    auto chan = chans[0];
+                    auto chan_fee = Amount(1);
+
+                    auto total_fee = fees[min_pk]; total_fee.mustAdd(chan_fee);
```
AndrejMitrovic: Can we keep statements on their own line? 

```suggestion
                    auto total_fee = fees[min_pk]; 
                    total_fee.mustAdd(chan_fee);
```

AndrejMitrovic (changes requested):

source/agora/flash/Network.d
omerfirmak: Yes, exactly.

AndrejMitrovic:
Please self-merge when ready as I'm heading out now.

PR #1630, opened 2021-02-10
Update flash-layer branch with latest localrest

omerfirmak:
Lets see if this will fix the Flash tests

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -606,7 +606,7 @@ public class TestAPIManager
 
     ***************************************************************************/
 
-    public Registry* getRegistry ()
+    public Registry!TestAPI* getRegistry ()
     {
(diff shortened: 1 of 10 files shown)


AndrejMitrovic:
There seems to be a race condition in the Flash code. I think the API will need to be amended with ACK / FAIL on each signing request. For example, currently if the node fails to sign it will not notify the counterparty. And then the failing node might start a new signing process with a new sequence ID which doesn't match the counterparty's sequence ID.

It will have to be fixed.

AndrejMitrovic:
Hmm I think I know what's going on. `proposeUpdate` is called by one node, but the other node might have already called a new `proposePayment` for the same channel. It's a timing issue.

omerfirmak:
Merged as part of #1631

PR #1631, opened 2021-02-15
Flash: Fix two-way payments

AndrejMitrovic:
This builds on top of the work in https://github.com/bpfkorea/agora/pull/1630

### DIFF
--- a/ci/run.sh
+++ b/ci/run.sh
@@ -15,6 +15,7 @@ dub build -b unittest-cov -c unittest --skip-registry=all --compiler=${DC}
 # Run this after unit tests have proven to compile ok
 rdmd --compiler=${DC} ./tests/runner.d --compiler=${DC} -cov
 
+export dtest=flash
 export dchatty=true
 export dsinglethreaded=true
 # Run a single test at a time to prevent resource issues and also see which test failed
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -239,14 +239,18 @@ public class Channel
     /***************************************************************************
 
         Returns:
-            the balance of the owner of the channel
+            the balance available for the provided payment direction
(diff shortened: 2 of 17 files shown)


omerfirmak:
I think we should put an upper bound to the `while(1)` retry loops. That may simplify the process

AndrejMitrovic:
I think another issue is that a payment from A->B can be started at the same time as an update from B->A. I think maybe the solution is to avoid spawning fibers everywhere, and to order the events by weight. Updates would be preferred over payments as this unlocks HTLCs rather than makes new ones.

AndrejMitrovic:
The last commit fixes the first unittest failing. I still think there might be issues with timing though.

AndrejMitrovic:
Ah there is a small bug in that commit too. I'll fix it.

AndrejMitrovic:
I've fixed that last commit again.

AndrejMitrovic:
What can we do about the CI not showing any logs until the tests are done? I can't get the raw logs until done either. So I have no idea where the CI is getting stuck.

omerfirmak:
> What can we do about the CI not showing any logs until the tests are done? I can't get the raw logs until done either. So I have no idea where the CI is getting stuck.

AFAIK, it should show them if you keep the page open from the start.

AndrejMitrovic:
> AFAIK, it should show them if you keep the page open from the start.

I know about this trick but it still fails to show the logs sometimes.

AndrejMitrovic:
All except one have finished green. I wonder if it's the GC thread getting stuck again..

@omerfirmak feel free to merge if you approve.

AndrejMitrovic:
I just realized I should have folded some of these commits. Well, let me know if I should do it.

omerfirmak (approved):

Issue #1623, opened 2021-02-09
Flash: handle failed payment routing and re-try with a different path


AndrejMitrovic:
This is related to https://github.com/bpfkorea/agora/issues/1528. But I think it can be implemented as two separate features.

Currently when a routing request fails, the sender just attempts to send it again. In fact there is no special ErrorCode for rejecting an HTLC. 

[This page](https://github.com/lnbook/lnbook/blob/develop/path-finding.asciidoc) shows the common LN implementation flow:

> If the payment settles, the sender updates its topological knowledge based on this new information for future payments. The algorithm terminates.
> 
> If the payment fails, the sender updates its topological knowledge based on this new information. It then selects a different path and starts the process again from the beginning.

So we need to have the ability for a hop to reject the HTLC. And the sending node should then forward this failure state back to *its* sender.

That means there can be a situation such as:
- Alice (htlc) => Bob (accepted)
- Bob => Charlie (rejected)

In this case Bob has already notified Alice that it accepted the HTLC via `proposeUpdate()`. So it cannot return an `ErrorCode` anymore when Charlie rejects its HTLC. Instead we should add a new API specifically for cancelling HTLCs, similar to how LN does it.

omerfirmak:
> This is related to #1528. But I think it can be implemented as two separate features.

I would say #1528 depends on this. As the sender has to know which channel failed to avoid it in the next payment attempt.

omerfirmak:
Similar to the LN, we need a mechanism to return error messages to the origin of the payment.

https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md#returning-errors

PR #1633, opened 2021-02-15
Flash: Propagate payment failure down the path

omerfirmak:
If a node on the path fails to route the payment
the already created HTLCs should unwind.

Part of #1623

### DIFF
--- a/source/agora/test/Flash.d
+++ b/source/agora/test/Flash.d
@@ -551,7 +551,7 @@ unittest
     // the utxo the funding tx will spend (only relevant to the funder)
     const bob_utxo = UTXO.getHash(hashFull(txs[1]), 0);
     const bob_charlie_chan_id = bob.openNewChannel(
-        bob_utxo, Amount(10_000), Settle_1_Blocks, charlie_pk);
+        bob_utxo, Amount(6_000), Settle_1_Blocks, charlie_pk);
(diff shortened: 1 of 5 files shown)


AndrejMitrovic:
On mobile it LGTM. I'll have another look soon.

AndrejMitrovic:
Had another look, LGTM!

omerfirmak:
Rebased & added some logs

source/agora/test/Flash.d
```diff
@@ -576,6 +576,13 @@ unittest
     writefln("Sleeping for 7 seconds..");
     Thread.sleep(7.seconds);
 
+    // Second payment attempt should fail because of insufficient funds
+    auto inv_2 = charlie.createNewInvoice(Amount(4_000), time_t.max, "payment 2");
+    alice.payInvoice(inv_2);
+
+    writefln("Sleeping for 7 seconds..");
+    Thread.sleep(7.seconds);
```
AndrejMitrovic: payInvoice is now blocking so I think we can get rid of the sleep calls between the payInvoice calls in this second test. I did get rid of them in the first test but I forgot about the second one.

omerfirmak: Done

AndrejMitrovic (approved):

PR #1636, opened 2021-02-16
Flash: Improve printouts

AndrejMitrovic:
A stop-gap measure, but it will help debugging significantly.


AndrejMitrovic:
This actually uncovered something. Payments A->B work well but not the other way around. Because the node always assumed it's A->B and not the reverse.

omerfirmak (approved):

AndrejMitrovic:
Folded into https://github.com/bpfkorea/agora/pull/1631

PR #1662, opened 2021-02-17
Large enough pre-images not to create new seeds

### LINKED ISSUE #1500: Nodes should be able to recover from exhausting their PreImageCycle
Once the preimages in a `PreImageCycle` are all used, nodes can't enroll again with the same UTXO.
They should be able to transfer their staked coins to a new UTXO, generate a new cycle of preimages and start enrolling with it.

This should probably be enabled by the `recurring_enrollment` config or a new one.

linked0:
Once the preimages in a `PreImageCycle` are all used, nodes can not enroll again with the same UTXO. So we decided that the number of Pre-Images generated when first enrolling should be large enough that there would unlikely be the need to ever re-enroll with a new pre-image seed.

The `nonce` is an unnecessary variable and the `populate` function of `PreImageCycle` is only used in unit tests, so I refactored the `PreImageCycle` class.

As @hewison-chris suggested, 5 million pre-images could be better and the `offset` should be removed. For removing `offset`, it will be processed in other PR. The number of pre-images generated can be modified in this PR.

Fixes #1500


source/agora/consensus/data/genesis/Test.d
```diff
@@ -59,44 +59,44 @@ public immutable Block GenesisBlock = {
             // NODE6
```
hewison-chris: I think these node comments will now need updating also. When https://github.com/bpfkorea/agora/pull/1626 is merged it will be easier as they get printed when the enrolments need updating.

hewison-chris (changes requested):
LGTM except the nodes indicated in the Test GenesisBlock probably need to be updated.

omerfirmak (approved):

source/agora/consensus/data/genesis/Test.d
linked0: Updated. Thanks.

Geod24:
Didn't we discuss doing away with the concept of pre-image being a multiple of cycles, and instead just going for the much simpler 1 pre-image = 1 block, in which case we should just specify the number of pre-images ?

linked0:
> Didn't we discuss doing away with the concept of pre-image being a multiple of cycles, and instead just going for the much simpler 1 pre-image = 1 block, in which case we should just specify the number of pre-images ?

Right. We just need to specify the number of pre-images. The `NumberOfCycles` remains unintentionally. I will remove it.

linked0:
Ready for review.
I promised to remove the concept of `cycle` and adopt the total number of pre-images to the code but I just changed the concept to that of the count of seeds. It's because accepting the count of `seed` is more efficient for unit tests or network tests. These codes will generate about 5 million pre-images for the `Coinnet` genesis block and about 100K pre-images for the `Test` genesis block. 

Setting the count of all the pre-images and creating the genesis block dynamically could be another issue, I think.

Thanks.

source/agora/consensus/PreImage.d
```diff
@@ -274,26 +274,17 @@ public struct PreImageCycle
...
-        This is the data used as a nonce in generating the cycle seed.
-        Named `nonce` to avoid any ambiguity. It is incremented once every
-        `EnrollPerCycle` period, currently 700 days.
-
-    ***************************************************************************/
-
-    public uint nonce;
+    /// The number of seeds for pre-images
+    /// If the cycle is 1008, about 5 million of pre-images will be generated.
+    public static immutable uint SeedCount = 5000;
```
Geod24: Am I missing something here ? Having a value which is dependent on the cycle will still create problems, won't it ?

kchulj: What is the `Seed` in this context?

linked0: > What is the `Seed` in this context?

We can't store all the pre-images like 5 million pre-images, so we just maintain 5000 seeds. When a new set of pre-images, we can a set of pre-images from a proper seed as follows.
`
[preimage 2016] [preimage 2015] ... [preimage 2] [seed 2] [preimage 1008] [preimage 1007] ... [[preimage 2] [seed1]
`
If you have a more proper name for the variable, please let me know. :)

hewison-chris: I think the cache is supposed to define which `preImages` are stored in the array and not the `seed`. I think if we are going to use a continuous sequence of `preImages` then a target count is easier to understand and follow.

kchulj: Thank you for the detailed explanation :)
That is similar to `PreImageCycle` vs `PreImageCache`.

linked0: > I think the cache is supposed to define which `preImages` are stored in the array and not the `seed`. I think if we are going to use a continuous sequence of `preImages` then a target count is easier to understand and follow.

I just use the `SeedCount` as a multiplier of a validator cycle because setting a total number of pre-images causes hashing 5 million times even in network tests or unit tests, which takes more than 1 second. That's why I used the seed count, not the total count.

hewison-chris: I thought if we needed to have it configurable for testing then we would make it configurable as `ConsensusParam` that we override in our unit tests. If you really want to stick with a multiplier of the `validatorCycle` then the original name `NumberOfCycles ` was more accurate.

linked0: > Am I missing something here ? Having a value which is dependent on the cycle will still create problems, won't it ?

Yes, it will. I think I should change the code. My intention was just not to hash so many times in unit tests or network tests but it will leave the problem also. I will use 5 million preimages for the Coinnet and 100K preimages for the Testnet. How about this? @Geod24 

@hewison-chris I understand what you mean now. I was misunderstanding some point.

source/agora/consensus/PreImage.d
```diff
@@ -274,14 +274,16 @@ public struct PreImageCycle
 {
     /// Make sure we get initialized by disabling the default ctor
     @disable public this();
+
     /// Ditto
-    public this (typeof(PreImageCycle.tupleof) args)
+    public this (uint seed_interval)
     {
-        this.tupleof = args;
+        this.seeds = PreImageCache(PreImageCycle.SeedCount, seed_interval);
```
Geod24: This is still using `SeedCount` as a multiplier.

Geod24 (approved):
LGTM. Want to do a  few more tests but that looks about right.
It doesn't to affect the test suite run time too much though ?

linked0:
> LGTM. Want to do a few more tests but that looks about right.
> It doesn't to affect the test suite run time too much though ?

Thanks. I think It adds more test time like a couple of minutes.

source/agora/consensus/PreImage.d
```diff
@@ -328,70 +318,6 @@ public struct PreImageCycle
...
-            auto next_images = PreImageCache(this.preimages.data.length,
-            this.preimages.interval);
-            next_images.reset(this.seeds.byStride[$ - 1 - this.index]);
-            return next_images[$ - 1];
-        }
-    }
-
     /***************************************************************************
 
         Seek to the `PreImage`s at height `height`
```
hewison-chris, line 323: The documentation still refers to `nonce`.

hewison-chris (approved):
LGTM - just documentation.

source/agora/consensus/PreImage.d
linked0: Updated

kchulj (approved):
LGTM.

AndrejMitrovic (approved):

Geod24:
I'll rebase this on Tuesday.

Geod24:
Should have said Thursday! I have a rebase locally, which passes test. Just need to add the WK preimages and will push.

linked0:
This PR is closing as the new PR #2252 is created instead.

Issue #1661, opened 2021-02-17
TxBuilder does not work as documented


Geod24:
https://github.com/bpfkorea/agora/blob/4c7f32d9a84a8ea666bdf362f742a5a931ec058c/source/agora/utils/Test.d#L283

It's documented like this, and then there's a test for the exact opposite behavior:
https://github.com/bpfkorea/agora/blob/4c7f32d9a84a8ea666bdf362f742a5a931ec058c/source/agora/utils/Test.d#L746-L747

Geod24:
Fixed by #1664

Issue #1440, opened 2020-12-22
In a freeze transaction, output corresponding to a refund shall not freeze.


MichaelKim20:
Currently, all outputs contained in a freeze transaction are treated as frozen UTXO.
However, the UTOX equivalent of the refund money should not be frozen and can be reused.
If a person wants to freeze only 40,000 BOA in UTXO with 50,000 BOA,
The refund equivalent of 10,000, another output of the same transaction, shall not be frozen.

AndrejMitrovic:
I see the problem. But how would you solve it?

MichaelKim20:
I've thought about a few things, but it's not enough to suggest such a good method.
1. To add a type to a transaction output
2. To set a rule that only the first item of transaction outputs is frozen

I've only thought about these solutions yet.

AndrejMitrovic:
I've had the same thoughts.

The reason why we didn't do `1.` initially, is because it would increase the size of the transactions.

The second rule might work.

PR #1664, opened 2021-02-17
Fix 1661, 1440: Refund output of freezing tx shouldn't be frozen

Geod24:
The fix for #1661 is necessary to make #1440 possible, so the two are in this PR.
CC @MichaelKim20

### DIFF
--- a/source/agora/consensus/state/UTXOSet.d
+++ b/source/agora/consensus/state/UTXOSet.d
@@ -13,6 +13,7 @@
 
 module agora.consensus.state.UTXOSet;
 
+import agora.common.Amount;
 import agora.common.crypto.Key;
 import agora.common.Serializer;
 import agora.common.Set;
@@ -82,11 +83,20 @@ abstract class UTXOCache
         {
             auto utxo_hash = UTXO.getHash(tx_hash, idx);
 
-            // Melting UTXOs will melt after 2016 blocks except ones that go to
-            // the CommonsBudget and others will melt at the next block.
-            auto utxo_value = UTXO((melting && output.address != commons_budget) ?
-                height + 2016 : height + 1, tx.type, output);
-            this.add(utxo_hash, utxo_value);
+            // In general, melting UTXOs will be available after 2016 blocks
+            // UTXOs which are towards the Commons Budget are excluded from this
+            // rule, and will be available from the next block.
+            if (melting && output.address != commons_budget)
+                this.add(utxo_hash, UTXO(height + 2016, tx.type, output));
+            // Additionally, if a transaction has two or more outputs,
+            // the first one is allowed to be under the minimum freezing amount.
+            // If it is, it will not be considered frozen.
+            else if (tx.type == TxType.Freeze && tx.outputs.length >= 2 &&
+                     (idx == 0 && output.value < Amount.MinFreezeAmount))
+                this.add(utxo_hash, UTXO(height + 1, TxType.Payment, output));
+            // Finally, the following is the most common case
+            else
+                this.add(utxo_hash, UTXO(height + 1, tx.type, output));
         }
     }
 
--- a/source/agora/test/Ledger.d
+++ b/source/agora/test/Ledger.d
@@ -232,3 +232,42 @@ unittest
     node_1.putTransaction(backup_tx);
     network.expectBlock(Height(3));  // new block finally created
 }
+
+// Ensure that when creating a frozen UTXO, the refund is not frozen too
+// See https://github.com/bpfkorea/agora/issues/1440
+unittest
+{
+    TestConf conf = { txs_to_nominate: 1 };
+    auto network = makeTestNetwork(conf);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    // Create a freezing tx with two outputs:
+    // 1) A refund of 1k
+    // 2) A very large frozen amount (60.999k)
+    Amount freezeAmount = network.blocks[0].txs[1].outputs[0].value;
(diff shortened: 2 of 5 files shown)


source/agora/consensus/state/UTXOSet.d
```diff
@@ -82,11 +83,23 @@ abstract class UTXOCache
         {
             auto utxo_hash = UTXO.getHash(tx_hash, idx);
 
-            // Melting UTXOs will melt after 2016 blocks except ones that go to
-            // the CommonsBudget and others will melt at the next block.
-            auto utxo_value = UTXO((melting && output.address != commons_budget) ?
-                height + 2016 : height + 1, tx.type, output);
-            this.add(utxo_hash, utxo_value);
+            // In general, melting UTXOs will be available after 2016 blocks
+            //
```
linked0: I don't think the blank line is needed.

source/agora/consensus/state/UTXOSet.d
```diff
@@ -82,11 +83,23 @@ abstract class UTXOCache
...
+            // UTXOs which are towards the Commons Budget are excluded from this
+            // rule, and will be available from the next block.
+            if (melting && output.address != commons_budget)
+                this.add(utxo_hash, UTXO(height + 2016, tx.type, output));
+
+            // Additionally, if a transaction has two or more outputs,
+            // the first one is allowed to be under the minimum freezing amount.
```
linked0, line 92: Is there any reason the amount only under `Amount.MinFreezeAmount` is unfrozen at the index 0?

linked0 (commented):
LGTM.

linked0 (approved):

source/agora/consensus/state/UTXOSet.d
Geod24: It would be (by convention) the refund transaction

source/agora/consensus/state/UTXOSet.d
Geod24: Cleaned up the whitespaces

AndrejMitrovic (approved):

source/agora/test/Ledger.d
```diff
@@ -232,3 +232,42 @@ unittest
...
+    assert(tx.outputs.length == 2);
+    network.clients[0].putTransaction(tx);
+
+    // Wait for the block to be created
+    network.expectBlock(Height(1));
+    const b1 = network.clients[0].getBlock(1);
+    assert(b1.txs.length == 1);
+
+    // Now spend the refund transaction
+    auto tx2 = TxBuilder(b1.txs[0], 0).sign();
```
MichaelKim20, line 267: Isn't the refund amount second output?

MichaelKim20: auto tx2 = TxBuilder(b1.txs[0], 1).sign();

source/agora/test/Ledger.d
```diff
@@ -232,3 +232,42 @@ unittest
...
+    assert(tx2.outputs.length == 1);
+
+    network.clients[0].putTransaction(tx2);
+    network.expectBlock(Height(2));
+    assert(network.clients[0].getBlock(2).txs.length == 1);
```
MichaelKim20, line 272: I think we need to make sure that the transaction of block height 2 can be used immediately.

source/agora/test/Ledger.d
Geod24: See the third commit in this PR.

source/agora/test/Ledger.d
Geod24: ??
It's a `Payment` transaction, of course it can be used directly. The real test is when `tx2` is created, as before, `tx2` would be rejected.

MichaelKim20: Thank you. I checked it

source/agora/consensus/state/UTXOSet.d
```diff
@@ -82,11 +83,20 @@ abstract class UTXOCache
...
+            // In general, melting UTXOs will be available after 2016 blocks
+            // UTXOs which are towards the Commons Budget are excluded from this
+            // rule, and will be available from the next block.
+            if (melting && output.address != commons_budget)
+                this.add(utxo_hash, UTXO(height + 2016, tx.type, output));
+            // Additionally, if a transaction has two or more outputs,
+            // the first one is allowed to be under the minimum freezing amount.
+            // If it is, it will not be considered frozen.
+            else if (tx.type == TxType.Freeze && tx.outputs.length >= 2 &&
+                     (idx == 0 && output.value < Amount.MinFreezeAmount))
```
MichaelKim20, line 95: @Geod24 I'm so happy that it's in a very simple code.
I am sending you a proposal because I have one question, so please review it.
There is a user with 100,000 BOA that is only one UTXO.
He wants to run a node after freezing 40,000 BOA. 
So he froze 40,000 BOA and sent a transaction with about 60,000 BOA to the refund.
The remaining about 60,000 BOA have also been frozen, so if he uses them for payment, others will not be able to use them for a certain period of time.
And if it is used for payment, other people who receive it will not be able to use it for a certain period of time.

Most users may not know that their 100,000 BOA is a single UTXO.
Therefore, I think it would be better to remove the limit on the amount of the refund.
I would like to know first why you added a limit on the amount of the refund for the freeze, and please kindly review my proposal.

Geod24: > I would like to know first why you added a limit on the amount of the refund for the freeze, and please kindly review my proposal.

Because otherwise it's impossible to create a frozen transaction that *does not* have a refund, but has more than one output. If a user wants to freeze only 40k out of > 80k, they will have to first split their UTXOs. The wallet could handle this automatically.
@AndrejMitrovic any thoughts on this ?

Issue #1550, opened 2021-01-26
Validation-related routines should be moved out of Ledger


Geod24:
The Ledger should be used to keep track of the state, and hence can be used by full nodes.
Having validator-related code in it leads to complications and makes it harder to use in a library.

Geod24:
Another issue is that `EnrollmentManager` is stored in the Ledger, however a full node doesn't need to manage (its own) enrollments.

PR #1665, opened 2021-02-17
Make `Ledger.validateBlock` handle genesis blocks

Geod24:
```
Instead of having to special case the call site,
just move the check inside of `validateBlock`.
```

Part of #1550


kchulj (approved):
LG

PR #1666, opened 2021-02-17
Ledger: Move IO/side effects to `acceptBlock`, make `addValidatedBlock` Ledger-only

Geod24:
```
There was a split of responsibilities between what `acceptBlock` and
`addValidatedBlock` were doing. `acceptBlock` was mostly forwarding
to `addValidatedBlock`, but also logging stats, except for one metric.

Because the node needed to potentially catch-up on startup,
a boolean (externalized) had to be added for this to work.
Since the boolean was only ever set to false a single time,
it's better to remove it, and instead have the validation duplicated
in the startup code.

In order for this to work, we also needed to carve the side-effect
performing code out of `addValidatedBlock`, as trying to save
an already saved block would have triggered an assertion.

The result is a public interface which does all the IO operations,
and forward to a private interface which does what the ledger
is supposed to do: update its state.
```


source/agora/node/Ledger.d
```diff
@@ -414,25 +419,15 @@ public class Ledger
...
         log.trace("Storing active validators for next block using height {}.", block.header.height);
         this.enroll_man.updateValidatorIndexMaps(Height(block.header.height + 1));
 
         // Clear the unknown TXs every round (clear() is not @safe)
         this.unknown_txs = Set!Hash.init;
 
-        if (externalized && this.onAcceptedBlock !is null)
-            this.onAcceptedBlock(block, validators_changed);
+        // Update the known "last block"
+        this.last_block = deserializeFull!Block(serializeFull(block));
```
Geod24, line 430: This is the only part that isn't quite clean, because we need to have a mutable `block` but we get a `const` one.
Planning on solving this later, but block externalization being a bit more expensive isn't a big deal.

source/agora/node/Ledger.d
```diff
@@ -393,18 +404,12 @@ public class Ledger
```
AndrejMitrovic, line 404: Need ddoc update.

AndrejMitrovic (approved):
Just 1 comment.

source/agora/node/Ledger.d
Geod24: Done

Geod24:
All 🟢

linked0 (approved):

kchulj (approved):

PR #1670, opened 2021-02-18
Flash: Dont propagate unknown payment hashes to be dropped

omerfirmak:
Second commit is unrelated but too small for a PR of its own.

### DIFF
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -741,8 +741,7 @@ public class Channel
         this.taskman.setTimer(0.seconds,
         {
             auto update_pair = this.update_signer.collectSignatures(
-                this.cur_seq_id,
-                new_outputs, priv_nonce, peer_nonce,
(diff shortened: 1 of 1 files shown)


source/agora/flash/Channel.d
```diff
@@ -727,10 +727,16 @@ public class Channel
         PrivateNonce priv_nonce = genPrivateNonce();
         PublicNonce pub_nonce = priv_nonce.getPublicNonce();
 
-        auto new_balance = this.foldHTLCs(this.cur_balance, secrets, rev_htlcs,
+        const old_balance = this.cur_balance;
```
linked0: I think there's no need to introduce the new variable `old_balance`. Is there any reason for that?

omerfirmak: Yep, PRs in the near future will be using that.

linked0 (approved):

AndrejMitrovic (approved):

PR #1671, opened 2021-02-18
Make BlockStorage.updateBlockMultiSig accepts the sig and bitfield

Geod24:
Instead of passing the whole block and serializing it,
just pass the fields to update along with the height
(to locate the block) and the hash (to check we have the right one).

### DIFF
--- a/tests/unit/BlockStorageMultiSig.d
+++ b/tests/unit/BlockStorageMultiSig.d
@@ -22,6 +22,7 @@ import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.genesis.Test;
 import agora.consensus.data.Transaction;
+import agora.crypto.Hash;
 import agora.node.BlockStorage;
 import agora.utils.Test;
 import agora.utils.PrettyPrinter;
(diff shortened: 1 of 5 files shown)


Geod24:
So, this fails because the `Height` is variably-serialized.

hewison-chris:
> So, this fails because the `Height` is variably-serialized.

Look like a compilation error in `agora/tests/unit/BlockStorageMultiSig.d`. I had a similar issue with my branch and refactored the file to make it easier to see why it fails to compile.

Geod24:
Should be fixed now

tests/unit/BlockStorageMultiSig.d
```diff
@@ -87,7 +87,7 @@ private void main ()
 
     /// Update each block adding an updated block multisig and set validator 0
     /// as also signed
-    iota(1, BlockCount).each!(h => {
+    iota(1, BlockCount).each!((h) {
```
hewison-chris, line 91: I have still seen cases where compile fails and it doesn't give the line number of the actual line that causes the problem. Only solution I think is to call inner function in lambda.

tests/unit/BlockStorageMultiSig.d
```diff
@@ -92,7 +93,7 @@ private void main ()
         assert(block2.header.validators[1] == true, format!"block at height %s:\n%s\n\n"(h, prettify(block2)));
         block2.header.signature = SIG2;
         block2.header.validators[0] = true;
-        storage.updateBlockMultiSig(block2);
+        storage.updateBlockSig(Height(h), block2.hashFull(), SIG2, block2.header.validators);
```
hewison-chris, line 96: This test fails now with `Hash of block to update differs from the expected one`

hewison-chris (changes requested):
Not quite ready yet.

Geod24:
Fixed

hewison-chris (approved):
LGTM

PR #1672, opened 2021-02-18
Flash: Route error information back to the origin 

omerfirmak:
This is a basic and naive implementation of the error routing.
It should be secured in multiple ways in the future.


AndrejMitrovic:
I guess we'll use onion-based error routing later, right?

AndrejMitrovic:
I rebased your branch after merging your other PR. I'll merge this one soon.

Issue #1614, opened 2021-02-09
Flash: Make onion packets fixed-size


AndrejMitrovic:
In https://github.com/bpfkorea/agora/issues/1529 onion encrypted packets were implemented. However their size varies based on the number of hops.

In LN the onion packet is always of a fixed size. This ensures no node can ever known how many hops the packet has traveled or how many hops it still needs to travel. The hops only know about the previous and next hops. This is an important privacy feature.

For this we should consult these pages:
- https://github.com/lnbook/lnbook/blob/develop/03_how_ln_works.asciidoc
- https://github.com/lightningnetwork/lightning-rfc/blob/master/04-onion-routing.md

AndrejMitrovic:
Resolved in #1673

PR #1673, opened 2021-02-19
Flash: Make onion packets fixed-size

AndrejMitrovic:
I went with a simpler array approach rather than recursive encryption for now. We could improve upon this later.

Fixes: https://github.com/bpfkorea/agora/issues/1614

### DIFF
--- a/source/agora/flash/OnionPacket.d
+++ b/source/agora/flash/OnionPacket.d
@@ -37,6 +37,7 @@ import agora.common.crypto.Schnorr;
 import agora.common.Hash;
 import agora.common.Serializer;
 import agora.common.Types;
+import agora.flash.Route;
 import agora.script.Lock;
 import agora.script.Script;
 import agora.utils.Log;
@@ -44,6 +45,8 @@ import agora.utils.Log;
 import libsodium.randombytes;
 import libsodium.crypto_secretbox;
 import libsodium.crypto_generichash;
+
+import std.range;
 
 import core.stdc.time;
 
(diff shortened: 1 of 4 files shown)


AndrejMitrovic:
Ah the first commit is unnecessary.

AndrejMitrovic:
Removed first commit.

AndrejMitrovic:
Need to fix the `payload` field, it's still not static size. But I need to override the serialization to avoid using VarInt, otherwise it can't be fixed-size.

AndrejMitrovic:
Added a commit making the payload / encrypted payload sizes static rather than a `byte[]`.

source/agora/flash/OnionPacket.d
```diff
@@ -101,11 +116,139 @@ public struct Payload
...
+    return packet;
+}
+
+/// Fill the payload with random encrypted data so it looks real but it ain't
+private void fillGarbage (ref EncryptedPayload payload)
+{
+    auto ephemeral_kp = Pair.random();
+
+    ubyte[32] key_bytes = ephemeral_kp.v[][0 .. 32];
+    randombytes_buf_deterministic(payload.nonce.ptr,
```
AndrejMitrovic: This should be fixed. The randomness of this may not generate a nonce that's on the curve. So it's better to just use `Scalar.random()` here.

AndrejMitrovic: Ah no I was wrong, this is not a point on the curve. But it is just a random value, so we can just use `randombytes_buf` for this.

AndrejMitrovic: Fixed.

AndrejMitrovic:
There was a typo while refactoring. Fixed it now.

AndrejMitrovic:
@omerfirmak can I get a merge here? I'll file a separate issue to make the structure recursive.

source/agora/flash/OnionPacket.d
```diff
@@ -67,11 +68,10 @@ public struct OnionPacket
...
 
-    // encrypted payload, decrypts to a `Payload`
-    public EncryptedPayload encrypted_payload;
-
-    // todo: replace with actual HMAC
-    public Hash hmac;
+    // encrypted payload. The first one decrypts & deserializes to a `Payload`.
+    // the next payload may or may not be legitimate, but it depends entirely
+    // on the next node's ability to decrypt the payload.
+    public EncryptedPayload[20] encrypted_payloads;
```
omerfirmak, line 74: Does this limit our hop count now?

AndrejMitrovic: Sure. Although we could technically set this to a higher limit if we want to.

omerfirmak (approved):

PR #1680, opened 2021-02-22
Improve code for SlashingMisbehavingValidator test

Geod24:
It's not perfect but get rid of some of the surprising code. Only left is getting rid of the shared `UTXOSet`.

### DIFF
--- a/source/agora/test/SlashingMisbehavingValidator.d
+++ b/source/agora/test/SlashingMisbehavingValidator.d
@@ -28,6 +28,7 @@ import agora.network.NetworkManager;
 import agora.utils.Test;
 import agora.test.Base;
 
+import core.atomic;
 import core.stdc.stdint;
 import core.thread;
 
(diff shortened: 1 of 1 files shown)


source/agora/test/SlashingMisbehavingValidator.d
```diff
@@ -52,10 +53,10 @@ private class MissingPreImageEM : EnrollmentManager
     public override bool getNextPreimage (out PreImageInfo preimage,
         Height height) @safe
     {
-        if (*this.reveal_preimage == false)
+        if (!atomicLoad(*this.reveal_preimage))
```
kchulj, line 56: What are advantages of using atomic operations here, and in L229?

Geod24: Avoiding bugs ;)

Jokes aside, it doesn't matter too much here, because it's  a simple boolean that is flipped, but you should never access shared data without proper synchronization. The basis of which is atomic operations. So a more serious answer would be "correctness".

linked0 (approved):
LGTM

kchulj (approved):
LGTM.

ferencdg:
I think other tests are also borken with the same problem like:
Byzantine.d
EnrollmentManager.d

Geod24:
> I think other tests are also borken with the same problem like:

I'll submit follow up PRs, thanks!

Issue #1644, opened 2021-02-17
Flash: Replace blocking pay/update APIs with accept/reject APIs and don't spawn fibers everywhere


AndrejMitrovic:
Currently we're spawning fibers everywhere, but we can replace this with a single fiber per channel. This makes sense since a channel can only be updated once at a time. That being said, we should be able to group multiple HTLCs in one go, but that's an optimization we can do later.

AndrejMitrovic:
I already began working on this, hope to have something finished soon.

AndrejMitrovic:
This was resolved in https://github.com/bpfkorea/agora/pull/1688

PR #1682, opened 2021-02-22
Re-architecture flash for fast payments

AndrejMitrovic:
No need for a review yet, it needs a cleanup. It's a CI test for now. It does work great on my local machine though.

Also includes https://github.com/bpfkorea/agora/pull/1673 in the first commit.

Fixes https://github.com/bpfkorea/agora/issues/1644

### DIFF
--- a/source/agora/flash/UpdateSigner.d
+++ b/source/agora/flash/UpdateSigner.d
@@ -116,6 +116,9 @@ public class UpdateSigner
     /// a new one should not be started.
     private bool is_collecting;
 
+    /// Set when the peer has confirmed collecting signatures for this update
+    private bool peer_confirmed_update;
+
     /***************************************************************************
 
         Constructor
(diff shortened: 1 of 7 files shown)


source/agora/flash/UpdateSigner.d
```diff
@@ -273,16 +297,23 @@ public class UpdateSigner
...
 
+        writefln("%s: Settlement signature %s from %s received",
+            this.kp.V.prettify, seq_id, this.peer_pk.prettify);
+
         if (auto error = this.isInvalidSettleMultiSig(this.pending_settle,
             settle_res.value, priv_nonce, peer_nonce))
         {
             // todo: inform? ban?
-            writefln("%s: Error during validation: %s. For settle signature "
+            writefln("%s: Error durin g validation: %s. For settle signature "
```
kchulj: I think this is a typo.

AndrejMitrovic: Yeah there's going to be lots of these here. But thanks.

AndrejMitrovic:
I can reproduce the thread bug where it's getting stuck on every run locally: https://github.com/AndrejMitrovic/agora/tree/thread-bug

Ran with: `dchatty=true dnologs=true dtest=agora.test.Flash dub test`.

Tried with latest ldc as well. It seems all the threads are waiting for something.

AndrejMitrovic:
Edit: removed old comment. I think the stack trace was deceiving.

AndrejMitrovic:
✅

AndrejMitrovic:
I think I might have reverted some of the error reporting routines. But I don't think they were designed right. The error reporting for a failed payment should be done through another API, not through a series of blocking calls.

omerfirmak:
> I think I might have reverted some of the error reporting routines. But I don't think they were designed right. The error reporting for a failed payment should be done through another API, not through a series of blocking calls.

IIRC, they werent blocking.

AndrejMitrovic:
> IIRC, they werent blocking.

Ah hold on, I think I see it now. I'll revert some of my changes.

AndrejMitrovic:
I re-added error reporting. I will write a separate test-case for this later though, as I don't want to further delay the PR.

AndrejMitrovic:
I'm gonna go ahead with this so we can continue rapid development.

PR #1686, opened 2021-02-23
Print out the first 8 chars of PubKey instead of 4

### LINKED ISSUE #1363: Logging the node short hash isn't as useful as it could be
e.g.:

```
Nov 13 08:06:52 eu-002 docker[14457]: 2020-11-13 08:06:52,28 
Info [agora.consensus.protocol.Nominator] - Rejected invalid envelope: 
{ statement: { node: GDNO...LVHQ, slotIndex:...
```

Node that `GDNO...` in this case is actually `GDNODE2...`, but we can't easily tell that because it's shortened in the output. We can still identify the nodes by their ending `...LVHQ`, but it's not debugging-friendly.

ferencdg:
SCPEnvelope's pretty printer now prints out the the first 8
characters of the PubKey instead of the previous 4.

fixes: #1363


AndrejMitrovic (approved):

linked0 (approved):
LGTM

kchulj (approved):

omerfirmak (approved):

PR #1687, opened 2021-02-23
Reduce usage of rdmd in tests

Geod24:
Since rdmd needs to call the compiler twice, it's quite slow.
Reduce its usage by directly using '-i -run' with the compiler instead.

### DIFF
--- a/ci/run.bat
+++ b/ci/run.bat
@@ -3,7 +3,9 @@ set AGORA_VERSION="HEAD"
 rem call dub test --skip-registry=all --compiler=%DC%
 call dub build --skip-registry=all --compiler=%DC% -c unittest -b unittest
 if %errorlevel% neq 0 exit /b %errorlevel%
-call rdmd --build-only --compiler=%DC% ./tests/runner.d --compiler=%DC% -cov
+rem call %DC% -i -run ./tests/runner.d %DC% -cov
+call %DC% -i ./tests/runner.d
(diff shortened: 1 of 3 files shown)


hewison-chris (approved):
LGTM

ci/run.bat
```diff
@@ -3,7 +3,7 @@ set AGORA_VERSION="HEAD"
 rem call dub test --skip-registry=all --compiler=%DC%
 call dub build --skip-registry=all --compiler=%DC% -c unittest -b unittest
 if %errorlevel% neq 0 exit /b %errorlevel%
-call rdmd --build-only --compiler=%DC% ./tests/runner.d --compiler=%DC% -cov
+call %DC% ./tests/runner.d %DC% -cov
```
linked0: Isn't there the option "-i -run" in Windows environment?

kchulj (approved):
Code LGTM.
Windows fails though..

ci/run.bat
Geod24: Correct

linked0 (approved):

ci/run.bat
Geod24: So actually, `-i` was missing, but `-run` shouldn't be used because we were not running those tests...

Geod24:
Brain just turned on, and I realized why it was failing, so setting auto-merge...

PR #1689, opened 2021-02-24
Upgrade minimum required version to LDC v1.25.0

Geod24:
And remove some workarounds for bugs which have been fixed.


Geod24:
Disabled one of the CI job due to dlang/dub#2110, otherwise it's green. I think one of the C++ mangling bug on Windows that was supposed to be fixed actually isn't, will track down later.

PR #1695, opened 2021-02-25
Reduce unnecessary fiber usage

AndrejMitrovic:
This should at least reduce memory usage and fiber spam.

### DIFF
--- a/source/agora/network/Client.d
+++ b/source/agora/network/Client.d
@@ -32,6 +32,7 @@ import vibe.http.client;
 
 import std.algorithm;
 import std.array;
+import std.container : DList;
 import std.format;
 import std.random;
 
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
Maybe this is the wrong level to optimize at, and we should rather optimize in LocalRest / Vibe.d. Although that means optimizing in two different codebases..

AndrejMitrovic:
This seems to mostly work minus the CircleCI troubles. I did get some failures previously in the Github CI, but I don't see where the problem is - seems random.

source/agora/network/Client.d
```diff
@@ -102,6 +164,75 @@ public class NetworkClient
...
+    }
+
+    /// Shut down the gossiping timer
+    public void shutdown ()
+    {
+        this.gossip_timer.stop();
+    }
+
+    /// For gossiping we don't want to block the calling fiber, so we use
+    /// use request queueing and a separate fiber to handle all the outgoing requests.
```
hewison-chris: double `use`

source/agora/network/Client.d
```diff
@@ -102,6 +164,75 @@ public class NetworkClient
...
+    private void gossipScheduler ()
+    {
+        while (1)
+        {
+            while (!this.gossip_queue.empty)
+            {
+                auto event = this.gossip_queue.front;
+                this.gossip_queue.removeFront();
+                this.handleGossip(event);
+                // yield w/ quick sleep return for next event
```
hewison-chris: In fact no sleep.

source/agora/network/Client.d
```diff
@@ -102,6 +164,75 @@ public class NetworkClient
...
+                this.taskman.wait(0.msecs);
+            }
+
+            this.taskman.wait(10.msecs);  // no events, yield w/ longer sleep
```
hewison-chris: is `w/` an abbreviation for with?

AndrejMitrovic: yes and I'll revert

source/agora/network/Client.d
```diff
@@ -102,6 +164,75 @@ public class NetworkClient
...
+    private void handleGossip (GossipEvent event) nothrow
+    {
+        switch (event.type) with (GossipType)
+        {
+        case Tx:
+            import agora.crypto.Hash;
+            const tx_hash = event.tx.hashFull();
+
+            // if the node already has this tx, don't send it
+            if (this.attemptRequest!(API.hasTransactionHash, Throw.No,
```
hewison-chris: I wonder if this is in fact slower as it is two API calls.

AndrejMitrovic: It depends on latency, bandwidth, and the tx size. We are checking the hash first before we send the full tx. But if latency is large then yes this will block the fiber. Maybe we should just send without asking?

hewison-chris (approved):
LGTM. Minor comment updates.

source/agora/network/Client.d
hewison-chris: Perhaps it can depend on the size of the `tx` which could be a configuration parameter. It doesn't have to be part of this PR though.

AndrejMitrovic: We could perhaps take a look at what Bitcoin is doing.

hewison-chris: They broadcast unconditionally according to [docs here](https://github.com/bitcoin/bitcoin/blob/ed25cb58f605ba583c735f330482df0bf9348f3a/src/rpc/rawtransaction.cpp#L819)

AndrejMitrovic: Yeah I was looking at the code and reached the same conclusion. I was the one who added this check here, so I'll just remove it and see what happens.

hewison-chris: 🤞

AndrejMitrovic:
It's green now.

hewison-chris (approved):
LGTM

kchulj (approved):

PR #1700, opened 2021-03-01
Enable `-preview=in` by default

Geod24:
Still a draft because I want to extend it to be used everywhere `Hash` is used.
In the process, I found some bug with Phobos, and realized that for this to be usable, we really need to have [9423](https://issues.dlang.org/show_bug.cgi?id=9423) fixed.

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -68,7 +68,7 @@
         "source/scpp/build/uint128_t.obj"
     ],
 
-    "dflags": [ "-extern-std=c++14" ],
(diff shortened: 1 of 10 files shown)


Geod24:
Upstream Vibe.d fix: https://github.com/vibe-d/vibe.d/pull/2538

Geod24:
Upstream fix merged, ready to go.

omerfirmak (approved):

PR #1702, opened 2021-03-02
Make the members of `KeyPair` the same as the members of `Pair`

### LINKED ISSUE #1645: Merge the two key pair types into one
We currently have two key types, `agora.common.crypto.Key : KeyPair` (with 3 types in it), and `agora.crypto.Schnorr : Pair`.
We should have `KeyPair` based on `Pair`, and just providing the routines to make it human readable (e.g. instead of printing `0xFFF...AAAA` for `Scalar`, print `SEEDGOESHERE`).

linked0:
We should merge the `KeyPair` and the `Pair` types into one. I made the components of the classes the same in this PR as the middle stage.
`SecretKey` of the `KeyPair` and `Scalar` of the `Pair` have `BitBlob!(crypto_core_ed25519_SCALARBYTES * 8)` data in this PR. And `PublicKey` of the `KeyPair` and `Point` of the `Pair` have `BitBlob!(crypto_core_ed25519_BYTES * 8)` data.

Feel free to post any opinions and suggestions. I will have a whiteboard if needed.

Fixes #1645


source/agora/common/crypto/Key.d
```diff
@@ -64,10 +66,15 @@ public struct KeyPair
...
+        BitBlob!(crypto_sign_ed25519_SECRETKEYBYTES * 8) sk_data;
         PublicKey pk;
-        if (crypto_sign_seed_keypair(pk[].ptr, sk[].ptr, seed[].ptr) != 0)
+        if (crypto_sign_seed_keypair(pk[].ptr, sk_data[].ptr, seed[].ptr) != 0)
             assert(0);
+
+        Scalar x25519_sk;
+        if (crypto_sign_ed25519_sk_to_curve25519(cast(ubyte*)(x25519_sk[].ptr), sk_data[].ptr) != 0)
+            assert(0);
+        SecretKey sk = SecretKey(x25519_sk[]);
```
Geod24: Are those different ? I expect they just hold the same value, don't they ?

source/agora/common/crypto/Key.d
```diff
@@ -248,9 +260,13 @@ public struct PublicKey
...
-        return 0 ==
-            crypto_sign_ed25519_verify_detached(
-                signature[].ptr, msg.ptr, msg.length, this.data[].ptr);
+        // return 0 ==
+        //     crypto_sign_ed25519_verify_detached(
+        //         signature[].ptr, msg.ptr, msg.length, this.data[].ptr);
+
+        import agora.crypto.Schnorr;
+        auto X = Point(this.data[]);
+        return agora.crypto.Schnorr.verify(X, signature, msg);
```
Geod24: Don't comment out code, remove it. Also if we're switching to `verify` then we don't need the `DataType` aliases anymore, do we ?

source/agora/common/crypto/Key.d
linked0: No, they are not. The data of the `SecretKey` has the `seed` and `Point` in its 64 bytes storage as follows.
https://github.com/jedisct1/libsodium/blob/57d950a54e6f7743084092ba5d31b8fa0641eab2/src/libsodium/crypto_sign/ed25519/ref10/keypair.c#L26-L27

source/agora/common/crypto/Key.d
linked0: Removed the comment.

But I don't understand why this change is related to the `DataType` alias. Anyway did you mean this kind of change?

```
public struct PublicKey
{
    BitBlob!(crypto_core_ed25519_BYTES * 8) data;
    alias data this;

    /// Construct an instance from binary data
    public this (const BitBlob!(crypto_core_ed25519_BYTES * 8) args)
        pure nothrow @safe @nogc
    {
        this.data = args;
    }
```

hewison-chris (approved):
LGTM, but I think in the commit description you have the Point and Scalar swapped.

source/agora/common/crypto/Key.d
```diff
@@ -123,7 +132,7 @@ public struct PublicKey
     /// Uses Stellar's representation instead of hex
     public string toString () const @trusted nothrow
     {
-        ubyte[1 + PublicKey.Width + 2] bin;
+        ubyte[1 + crypto_core_ed25519_BYTES + 2] bin;
```
Geod24: Revert, because they should be the same

source/agora/common/crypto/Key.d
```diff
@@ -133,7 +142,7 @@ public struct PublicKey
     /// Make sure the sink overload of BitBlob is not picked
     public void toString (scope void delegate(const(char)[]) sink) const @trusted
     {
-        ubyte[1 + PublicKey.Width + 2] bin;
+        ubyte[1 + crypto_core_ed25519_BYTES + 2] bin;
```
Geod24: Revert

source/agora/common/crypto/Key.d
```diff
@@ -169,7 +178,7 @@ public struct PublicKey
     public static PublicKey fromString (scope const(char)[] str) @trusted
     {
         const bin = Base32.decode(str);
-        enforce(bin.length == 1 + PublicKey.Width + 2);
+        enforce(bin.length == 1 + crypto_core_ed25519_BYTES + 2);
```
Geod24: Revert

source/agora/common/crypto/Key.d
```diff
@@ -245,12 +254,7 @@ public struct PublicKey
     public bool verify (Signature signature, scope const(ubyte)[] msg)
         const nothrow @nogc @trusted
     {
-        // The underlying function does not expose a safe interface,
-        // but we know (thanks to tests and careful inspection)
-        // that our data type match and they are unlikely to change
-        return 0 ==
-            crypto_sign_ed25519_verify_detached(
-                signature[].ptr, msg.ptr, msg.length, this.data[].ptr);
+        return agora.crypto.Schnorr.verify(Point(this.data[]), signature, msg);
```
Geod24: Just `this.data` ? It's already a `Point`

source/agora/common/crypto/Key.d
```diff
@@ -348,12 +348,13 @@ public struct SecretKey
 
     public Signature sign (scope const(ubyte)[] msg) const nothrow @nogc
     {
-        Signature result;
-        // The second argument, `siglen_p`, a pointer to the length of the
-        // signature, is always set to `64U` and supports `null`
-        if (crypto_sign_ed25519_detached(result[].ptr, null, msg.ptr, msg.length, this.data[].ptr) != 0)
-            assert(0);
-        return result;
+        Pair kp = Pair.fromScalar(Scalar(this.data[]));
```
Geod24: I think we should drop `SecretKey` completely and just move this to `KeyPair`.

source/agora/common/crypto/Key.d
```diff
@@ -348,12 +348,13 @@ public struct SecretKey
...
+        version (unittest)
+        {
+            const Pair noise = Pair.fromScalar(Scalar(hashMulti(kp.v, "sign")));
+            return agora.crypto.Schnorr.sign(kp, noise, msg);
+        }
```
Geod24: What is this ?

source/agora/common/crypto/Key.d
Geod24: Also, `this.data` is already a `Scalar`

source/agora/common/crypto/Key.d
```diff
@@ -540,11 +541,7 @@ unittest
 
 public static Scalar secretKeyToCurveScalar (SecretKey secret) nothrow @nogc
 {
-    Scalar x25519_sk;
-    // FIXME: We don't want to expose a mutable buffer but we need one here
-    if (crypto_sign_ed25519_sk_to_curve25519(cast(ubyte*)(x25519_sk[].ptr), secret[].ptr) != 0)
-        assert(0);
-    return x25519_sk;
+    return Scalar(secret[]);
```
Geod24: `return secret.data` ?

source/agora/consensus/Quorum.d
```diff
@@ -115,7 +115,10 @@ public QuorumConfig buildQuorumConfig (in PublicKey key,
         added[idx] = true;  // mark used
     }
 
-    quorum.nodes.sort;
+    () @trusted {
+        scope (failure) assert(0);
+        quorum.nodes.sort;
+    }();
```
Geod24: This looks unrelated ?

source/agora/common/crypto/Key.d
```diff
@@ -565,7 +562,7 @@ unittest
     assert(pair.V[] == kp.address[]);
     Signature enroll_sig = sign(pair, "BOSAGORA");
 
-    Point point_Address = Point(kp.address);
+    Point point_Address = Point(kp.address[]);
```
Geod24: I think we can just remove this unittest now ?

source/agora/consensus/validation/Enrollment.d
```diff
@@ -66,7 +66,7 @@ public string isInvalidReason (const ref Enrollment enrollment,
     if (utxo_set_value.type != typeof(utxo_set_value.type).Freeze)
         return "Enrollment: UTXO is not frozen";
 
-    Point address = Point(utxo_set_value.output.address);
+    Point address = Point(utxo_set_value.output.address[]);
```
Geod24: Why do we need to slice ? `address` has an `alias this` to a `Point`.

Geod24 (approved):
Broadly LGTM

source/agora/common/crypto/Key.d
Geod24: I see why you need it now. We can just use `sizeof` here.

source/agora/consensus/Quorum.d
Geod24: I see why. You get the error message because, while sorting, `opCmp` is called, and unfortunately it is not annotated. Will fix.

Geod24: For reference:
https://github.com/bpfkorea/crypto/blob/374d775327cb381a5037deb88edb7e352636de2f/source/agora/crypto/ECC.d#L382-L392
Calls https://github.com/Geod24/bitblob/blob/5418160fe7751eaa1ec99870ae7c62bd1f9c58e7/source/geod24/bitblob.d#L237-L251 ~but the later doesn't require annotations because it's templated~ Edit: Nvm the attributes are at the top.

Geod24: Depends on https://github.com/bpfkorea/crypto/pull/19

Geod24:
Pushed some fixes

source/agora/common/crypto/Key.d
Geod24: Let's keep removing `SecretKey` for later.

source/agora/consensus/Quorum.d
hewison-chris: BTW crypto submodule update has been merged.
Commit cbb4169a3a23092d18463b8723a68680f36c5010  includes the changes to attributes.

hewison-chris: Will be included now if you rebase on upstream.

linked0: OK, I will apply it. Thanks.

source/agora/common/crypto/Key.d
linked0: This is for the unit test of `PrettyPrinter` checking the formatted output of a created block. A random noise for signing changes every time as you know. Is there any better way for that?

source/agora/consensus/Quorum.d
linked0: Applied new `crypto` submodule.

linked0:
@hewison-chris @Geod24 Thank you for your review and comments. All the suggestions except [one](https://github.com/bpfkorea/agora/pull/1702#discussion_r589043952) are applied. 
I will remove `secretKeyToCurveScalar` and request reviews again.

linked0:
Removed the `secretKeyToCurveScalar` function which is not needed anymore because the `SecretKey` has the `Scalar` data now.

source/agora/common/crypto/Key.d
```diff
@@ -102,14 +114,11 @@ unittest
 /// Represent a public key / address
 public struct PublicKey
 {
-    /// Alias to the BitBlob type
-    private alias DataType = BitBlob!(crypto_sign_ed25519_PUBLICKEYBYTES * 8);
-
-    /*private*/ DataType data;
+    /*private*/ Point data;
```
hewison-chris, line 120: The commit message is wrong. 
Public key is Point and Secret key is Scalar.

source/agora/common/crypto/Key.d
```diff
@@ -123,7 +132,7 @@ public struct PublicKey
     /// Uses Stellar's representation instead of hex
     public string toString () const @trusted nothrow
     {
-        ubyte[1 + PublicKey.Width + 2] bin;
+        ubyte[1 + PublicKey.sizeof + 2] bin;
```
hewison-chris: Perhaps descriptive constants could be used for these extra byte lengths.

source/agora/common/crypto/Key.d
hewison-chris: I will take a look.

hewison-chris (changes requested):

source/agora/common/crypto/Key.d
linked0: Updated. Ty.

source/agora/common/crypto/Key.d
linked0: I see what you mean. I will add them in another commit.

source/agora/common/crypto/Key.d
linked0: Deleted with the support of @hewison-chris.

hewison-chris (approved):
LGTM

source/agora/common/crypto/Key.d
```diff
@@ -81,14 +88,19 @@ public struct KeyPair
...
-        if (crypto_sign_ed25519_sk_to_seed(seed[].ptr, sk[].ptr) != 0)
+        if (crypto_sign_ed25519_sk_to_seed(seed[].ptr, sk_data[].ptr) != 0)
             assert(0);
-        return KeyPair(pk, sk, seed);
+
+        Scalar x25519_sk;
+        if (crypto_sign_ed25519_sk_to_curve25519(cast(ubyte*)(x25519_sk[].ptr), sk_data[].ptr) != 0)
+            assert(0);
+        SecretKey sk = SecretKey(x25519_sk[]);
+        return KeyPair(PublicKey(pk[]), sk, seed);
```
Geod24, line 103: https://github.com/jedisct1/libsodium/blob/ae4add868124a32d4e54da10f9cd99240aecc0aa/src/libsodium/crypto_sign/ed25519/ref10/keypair.c#L69-L83

For reference

PR #1704, opened 2021-03-02
Use 'in' parameters in even more places + some minor fixes

Geod24:
Second commit is the main one. Best reviewed commit-by-commit.


source/agora/script/Script.d
```diff
@@ -434,7 +434,7 @@ unittest
 *******************************************************************************/
 
 version (unittest)
-public Script makeScript (in ubyte[] opcodes) pure nothrow @safe @nogc
+public Script makeScript (return const ubyte[] opcodes) pure nothrow @safe @nogc
```
linked0, line 437: Why is the `return` keyword used? I don't think it's related to the lifetime of the parameter `opcodes`. Could you explain?

Geod24: It is related to the lifetime. It means that `opcodes` will be part of what's returned by `makeScript`, i.e., part of the `Script` instance.

source/agora/consensus/validation/Block.d
```diff
@@ -673,14 +673,14 @@ version (unittest)
             },
             prev_timestamp, (curr_timestamp == ulong.max) ? block.header.timestamp : curr_timestamp,
             block_timestamp_tolerance_dur,
-            (const ref Transaction[] tx_set, const ref uint[] missing_validators)
+            (in Transaction[] tx_set, in uint[] missing_validators)
             {
-                return cast(Transaction[]) [];
+                return (Transaction[]).init;
```
kchulj, line 678: Are the parentheses needed, although the `cast` has been removed?

Geod24: Yes because it's an array. `Transaction.init` parses fine, but `Transaction[].init` will yield a compiler error.

linked0 (approved):

kchulj (approved):
LGTM

PR #1706, opened 2021-03-02
Update Shnorr signature datastructs during startup

### LINKED ISSUE #1703: signature related datastructures inside EnrollmentManager are not restored
If the validator is restarted and the UTXO array is provided while creating the Ledger object, then 'key_to_index' and 'index_to_key' datastructeres are not recovered(left empty) for past blocks. Those datastructures are recovered though if the Ledger is created without the UTXO array. 

@hewison-chris could you please confirm this? I haven't tested, just found this while looking at the code.

ferencdg:
'key_to_index' and 'index_to_key' were not filled in for the entire
cycle length during validator start up

fixes: #1703

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -526,6 +526,7 @@ public class Ledger
     protected void updateValidatorSet (in Block block) @safe
     {
         this.enroll_man.clearExpiredValidators(block.header.height);
+        this.enroll_man.updateValidatorIndexMaps(Height(block.header.height));
         PublicKey pubkey = this.enroll_man.getEnrollmentPublicKey();
         UTXO[Hash] utxos = this.utxo_set.getUTXOs(pubkey);
         foreach (idx, ref enrollment; block.header.enrollments)


Geod24 (approved):

source/agora/node/Ledger.d
```diff
@@ -526,6 +526,7 @@ public class Ledger
     protected void updateValidatorSet (in Block block) @safe
     {
         this.enroll_man.clearExpiredValidators(block.header.height);
+        this.enroll_man.updateValidatorIndexMaps(Height(block.header.height));
```
hewison-chris, line 529: This should only be called if the height is greater than 0. Otherwise we get errors logged for no validators at height 0.

ferencdg: apart from that I think we need to handle the case where 
no utxo_set is provided and 'this.enroll_man.validatorCount() != 0'

hewison-chris: I believe it is handled within this function [this.addValidatedBlock(block);](https://github.com/bpfkorea/agora/blob/5310eb4bec8dac40a00ac7fe8568eee45495173e/source/agora/node/Ledger.d#L191) which is called for each block in the case you mention.

PR #1707, opened 2021-03-02
Move `pushBlock` call out of ctor into `start` function

Geod24:
Removing IO in ctor is always a good thing. Here we move a network request from the node ctor and into the `start` function, which is also de-duplicated to make the code more straightforward.


linked0 (approved):
LGTM

kchulj (approved):

PR #1708, opened 2021-03-02
regenerateQuorums: Get block height instead of using Height(0)

### LINKED ISSUE #1663: getRandomSeed() needs to assert instead of returning Hash.init
It seems to me this could be the root cause of the problem in https://github.com/bpfkorea/agora/issues/1512

Specifically:

https://github.com/bpfkorea/agora/blob/1735931aa2b3e02c04b4453101cef9c9b61d5617/source/agora/node/Validator.d#L103

This calls the quorum generator with height zero, even though the restarted node's ledger may be newer than this. But if for whatever reason those nodes don't exist anymore, the `getRandomSeed` will return `Hash.init` and then the generated quorum for the node is in an invalid state.

In addition to that, this bug should be fixed:

```diff
diff --git a/source/agora/node/Validator.d b/source/agora/node/Validator.d
index 24fba457a..e8001aa07 100644
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -100,7 +100,7 @@ public class Validator : FullNode, API
         // currently we are not saving preimage info,
         // we only have the commitment in the genesis block
-        this.regenerateQuorums(Height(0));
+        this.regenerateQuorums(this.ledger.getBlockHeight());
         this.admin_interface = new AdminInterface(config,
             this.config.validator.key_pair, this.clock);
```

kchulj:
If `getRandomSeed` returns `Hash.init`, then the generated quorum for the node is in an invalid state.

The current quorum generator is called with `Height(0)`, even though the restarted node's ledger may be newer than this.

Fixes #1663

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -100,7 +100,7 @@ public class Validator : FullNode, API
 
         // currently we are not saving preimage info,
         // we only have the commitment in the genesis block
-        this.regenerateQuorums(Height(0));
+        this.regenerateQuorums(this.ledger.getBlockHeight());
 
         this.admin_interface = new AdminInterface(config,
             this.config.validator.key_pair, this.clock);


source/agora/consensus/EnrollmentManager.d
```diff
@@ -718,6 +718,7 @@ public class EnrollmentManager
                 continue;
             }
             rand_seed = hashMulti(rand_seed, preimage);
+            assert(rand_seed != Hash.init);
```
linked0: I think this line should be out of the foreach loop because if all the pre-images of the `keys` are PreImageInfo.init, the `rand_seed` will be Hash.init in the end.

kchulj: Good point. I was wondering whether to put this line here, or above the `return`.

kchulj: If it `continue`, then it will not reach the `hashMulti` right ?

linked0: Correct.

Geod24 (approved):
Hum, this loop is buggy on so many levels...

source/agora/consensus/EnrollmentManager.d
```diff
@@ -719,6 +719,7 @@ public class EnrollmentManager
             }
             rand_seed = hashMulti(rand_seed, preimage);
         }
+        assert(rand_seed != Hash.init);
```
Geod24: The assert gets hit during tests

PR #1712, opened 2021-03-02
Flash: Fix error routing

omerfirmak:


source/agora/test/Flash.d
```diff
@@ -515,7 +515,7 @@ unittest
     // the utxo the funding tx will spend (only relevant to the funder)
     const bob_utxo = UTXO.getHash(hashFull(txs[1]), 0);
     const bob_charlie_chan_id = bob.openNewChannel(
-        bob_utxo, Amount(10_000), Settle_1_Blocks, charlie_pk);
+        bob_utxo, Amount(3_000), Settle_1_Blocks, charlie_pk);
```
AndrejMitrovic, line 518: Could you please write a separate unittest for this? It's not great having one test doing multiple separate functionality tests.

AndrejMitrovic: Actually let me unblock this, and then let's just start writing new tests instead of modifying existing ones.

PR #1719, opened 2021-03-03
 Flash: Keep track of shared secrets

omerfirmak:
Channels should keep track of the shared secret that they use to
decrypt a message. This is needed in order to obfuscate the error
messages that may be caused by that particular payment.

### DIFF
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -86,8 +86,9 @@ public class Channel
     /// The list of any off-chain updates which happened on this channel
     private UpdatePair[] channel_updates;
 
-    /// Unresolved payment hashes which are part of existing HTLCs
-    private Set!Hash payment_hashes;
+    /// Unresolved payment hashes and their associated shared secret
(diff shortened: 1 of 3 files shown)


source/agora/flash/Channel.d
```diff
@@ -1113,7 +1130,7 @@ LOuter: while (1)
     public Result!PublicNonce onProposedPayment (in uint seq_id,
         in Hash payment_hash, in Amount amount, in Height lock_height,
         in OnionPacket packet, in Payload payload, in PublicNonce peer_nonce,
-        in Height height )
+        in Height height, in Point shared_secret)
```
AndrejMitrovic, line 1134: Could you add this to the param docs?

omerfirmak: Done

AndrejMitrovic (approved):

AndrejMitrovic:
I'll merge this first and then push the rebased `flash-layer` branch (it's outdated compared to v0.x.x).

PR #1718, opened 2021-03-03
Test height 2 in system integration test

AndrejMitrovic:
I thought it worked at first but I misread the logs. Let's see if the CI agrees.

Issue #1717, opened 2021-03-03
Integrate the Flash node and add system integration tests


AndrejMitrovic:
We already have enough code support to make this possible.

The first step is moving stuff outside of `test/Flash` and into a base class. Then, integrating the Flash node within FullNode / Validator. I have something WIP.

AndrejMitrovic:
Resolved in #1721

PR #1721, opened 2021-03-03
Integrate Flash into FullNode / Validator and provide FlashThinNode

AndrejMitrovic:
I've considered multiple approaches. But it's not easy due to D not supporting multiple-inheritance. So everything has to be done through interfaces or aggregation.

This is just one approach, maybe you can think of other ideas @omerfirmak?

I tried to add system integration tests, but that test-suite seems to be failing even in v0.x.x (https://github.com/bpfkorea/agora/pull/1718). So I'm blocked on that.

The next step would have been to deploy some flash nodes to our private testnet and have them create channels with each other and send some microtransactions. It's not obvious how this can be done since the nodes have a controller interface but it's not exposed remotely (should only be done locally like the Admin interface thingy).

Anyway this is all WIP.

Part of #1717

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -22,6 +22,7 @@ import agora.common.Types;
 import agora.common.Set;
 import agora.common.Serializer;
 import agora.consensus.data.Transaction;
+import agora.flash.API;
 
 import vibe.data.serialization;
 import vibe.http.common;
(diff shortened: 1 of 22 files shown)


omerfirmak:
I went over it and I think it is a solid start. We should probably merge this and see if we will come across any limitations as we go.

AndrejMitrovic:
I have improvements I'll push soon. Also getting rid of duplicate interfaces. And I just had the first successful vibe.d networking Flash test!

AndrejMitrovic:
I've split it up into multiple commits and I've added a `testing` mode. This is the best I could come up with given the limited time we have.

I've tested this locally with this diff: https://gist.github.com/AndrejMitrovic/12e907f433020bff5321e266b5b8b565

It works quite nicely. Alice pays Charlie's invoice through an intermediate channel. So Flash is confirmed working on TCP/IP.

The testing steps are hardcoded in the classes but we can fix this later.

omerfirmak (approved):
Great work!

AndrejMitrovic:
> Great work!

Aww thanks buddy!

PR #1722, opened 2021-03-03
Shut down timers while shutting down Agora

Geod24:
While this is unlikely to affect the live system,
it might have an effect in unittest where restarting nodes frequently fails.


Geod24:
Looks like it let the system integration tests shut down properly \o/

Geod24:
Found the damn bug, should be good to go now (minus the fact that Faucet is broken)

kchulj (approved):
LGTM

PR #1724, opened 2021-03-03
Flash: Obfuscate OnionError in each hop

omerfirmak:


source/agora/flash/Channel.d
```diff
@@ -1842,10 +1842,14 @@ LOuter: while (1)
         if (error.payment_hash in this.payment_hashes ||
             error.payment_hash in this.dropped_htlcs)
         {
+            const shared_secret = error.payment_hash in this.payment_hashes ?
+                                    this.payment_hashes[error.payment_hash] :
+                                    this.dropped_htlcs[error.payment_hash].shared_secret;
```
AndrejMitrovic, line 1847: Is it guaranteed the hash will be in `dropped_htlcs`?

source/agora/flash/OnionPacket.d
```diff
@@ -512,3 +512,49 @@ public struct OnionError
...
+        Obfuscated/deobfuscated OnionError
+
+*******************************************************************************/
+
+public OnionError obfuscate (OnionError error, Point secret) @trusted
+{
+    import libsodium.crypto_stream_chacha20;
+
+    const key = hashFull(secret);
+    assert(key[].length >= crypto_stream_chacha20_KEYBYTES);
```
AndrejMitrovic, line 534: Is this technique with chacha20 described in the LN RFC? From my brief glance at the RFC I thought the same technique would be used as the one for onion packets.

It's fine to use chacha though.

source/agora/flash/OnionPacket.d
```diff
@@ -512,3 +512,49 @@ public struct OnionError
...
+    assert(error.payment_hash[].length >= crypto_stream_chacha20_NONCEBYTES);
+
+    ubyte[error.sizeof] stream;
```
AndrejMitrovic: OnionError.sizeof specifically. Just to make it clear it's fixed-size.

AndrejMitrovic (commented):
Very clever, I like it. Just a few comments.

source/agora/flash/Channel.d
omerfirmak: Yep, the if above guarantees that.

source/agora/flash/OnionPacket.d
omerfirmak: RFC does not specify any algorithm but the implementation uses chacha. 

https://github.com/lightningnetwork/lightning-onion/blob/3c8c8d0b89ea82958d1fae0657e3eaf5f1267b5f/crypto.go#L138-L150

source/agora/flash/OnionPacket.d
omerfirmak: Done

AndrejMitrovic (approved):

PR #1725, opened 2021-03-04
Rename modules to not include their package names

Geod24:
Having the package name in the module name just adds redundancy,
so shorten their names. NetworkRegistryImpl was also renamed
to simply NetworkRegistry (to match FullNode / Validator,
which aren't named FullNodeImpl or ValidatorImpl).


linked0 (approved):
LGTM
I love all of your commits with good commit messages!

kchulj (approved):
Good idea.

PR #1727, opened 2021-03-04
Nominator: Use a constant for the empty value

Geod24:


source/agora/consensus/protocol/Nominator.d
```diff
@@ -418,11 +422,9 @@ extern(D):
     {
         log.info("{}(): Proposing tx set for slot {}", __FUNCTION__, slot_idx);
 
-        auto prev_value = ConsensusData.init.serializeFull().toVec();
         auto next_value = next.serializeFull().toVec();
-        auto prev_dup = duplicate_value(&prev_value);
```
kchulj, line 423: I guess we don't need to duplicate the `empty_value`?

Geod24: The point of duplicating was:
1) Ensure that what is allocated in C++ is deallocated in C++;
2) Ensure that the GC doesn't collect things in C++ land;

The first one is achieved by passing the vector by ref, the later is achieved by having the GC never collect it.

kchulj (approved):

PR #1728, opened 2021-03-04
eventcore: Switch back to upstream v0.9.13

Geod24:
Now that the minimum required version of LDC is v1.25.0,
we are guaranteed to have a workaround around dlang issue 21442,
so switch back to upstream to get proper cleanup.


kchulj (approved):
🟢

PR #1731, opened 2021-03-04
Flash: Retry failed payment attempts

omerfirmak:
See each commit

### DIFF
--- a/source/agora/test/Flash.d
+++ b/source/agora/test/Flash.d
@@ -419,3 +419,119 @@ unittest
     writefln("alice closing tx: %s", alice.getClosingTx(alice_bob_chan_id));
     assert(block12.txs[0] == alice.getClosingTx(alice_bob_chan_id));
 }
+
+/// Test path probing
+unittest
+{
+    TestConf conf = { txs_to_nominate : 1, payout_period : 100 };
+    auto network = makeTestNetwork(conf);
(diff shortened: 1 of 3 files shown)


AndrejMitrovic:
I will have a closer look after it's out of draft. You think we can merge https://github.com/bpfkorea/agora/pull/1721? I'll rebase against v0.x.x after.

omerfirmak:
Down the road we should probably add a max retry count and an increasing delay between each attempt.

source/agora/test/Flash.d
```diff
@@ -419,3 +419,120 @@ unittest
     writefln("alice closing tx: %s", alice.getClosingTx(alice_bob_chan_id));
     assert(block12.txs[0] == alice.getClosingTx(alice_bob_chan_id));
 }
+
+/// Test indirect channel payments
```
AndrejMitrovic: Should the comment be updates? Seems like copypasta. :)

I'll merge now and you can fix it.later.

AndrejMitrovic (approved):

source/agora/test/Flash.d
AndrejMitrovic: Or self-merge when ready!

omerfirmak: Thanks, updated.

PR #1732, opened 2021-03-05
NetworkManager: Filter our own key

AndrejMitrovic:
Something I've noticed during the Flash work. It's not an ideal solution as we should filter connecting to ourselves on the connection level. But at least it might help with the test-suite.


AndrejMitrovic:
Also this will still not prevent a FullNode from connecting to itself. As I mentioned, we need a better mechanism.

Geod24 (approved):
Nice

PR #1734, opened 2021-03-05
Move CircleCI to Github CI

Geod24:
We've been seeing quite a few issues with CircleCI lately, so this unifies everything under Github CI.


Geod24:
Yup, this works, will merge in a few minutes.

PR #1735, opened 2021-03-05
[CI test] Update flash-layer against latest v0.x.x

AndrejMitrovic:
Just a test before the force-push to `flash-layer` so I avoid breaking the branch again.


AndrejMitrovic:
I think this one is failing in the system integration tests because we're using a faucet commit which uses the `v0.x.x` branch instead of `flash-layer` - so all the signatures are invalid as they're not using Schnorr.

AndrejMitrovic:
I don't think there's any easy way fixing this for now except disabling the integration tests. So I'll add that as a commit.

hewison-chris:
> I don't think there's any easy way fixing this for now except disabling the integration tests. So I'll add that as a commit.

You could use a labelled version of faucet and put it in your dockerHub account. Then in the docker-compose use that.

AndrejMitrovic:
> You could use a labelled version of faucet and put it in your dockerHub account. Then in the docker-compose use that.

But the `flash-layer` branch isn't updated yet. So how do I make faucet target it? I think this is a chicken & egg problem.

AndrejMitrovic:
> But the flash-layer branch isn't updated yet. So how do I make faucet target it? I think this is a chicken & egg problem.

Never mind, Chris explained how to do it.

AndrejMitrovic:
CI is timing out. :(

AndrejMitrovic:
@omerfirmak does this branch work for you locally with `dub test`?

AndrejMitrovic:
ah it doesn't work for me locally anymore either. Some formatting issue. Fixing..

AndrejMitrovic:
Fixed formatting errors. Waiting for CI.

AndrejMitrovic:
> [source/agora/flash/OnionPacket.d:169] Assertion thrown during test:

Hmmm

omerfirmak:
> > [source/agora/flash/OnionPacket.d:169] Assertion thrown during test:
> 
> Hmmm

I think we are missing https://github.com/bpfkorea/agora/commit/4c5fe5911c9212c04acfac8261d0208a24948653

AndrejMitrovic:
> I think we are missing 4c5fe59

Right! I'll push again.

AndrejMitrovic:
It's green. I'm gonna push it to the branch.

PR #1737, opened 2021-03-07
NetworkManager: Remove buggy getAddress/isOurOwnAddress

Geod24:
```
isOurOwnAddress was not working as intended, since it relied on getAddress,
which itself relied on the config value of 'node.address', which is not
our own address, but rather a netmask of interfaces to listen to.
The most value for this field is actually '0.0.0.0',
meaning we're listening on all available interfaces,
so we would never attempt to connect to 0.0.0.0.
```

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -720,8 +720,7 @@ public class NetworkManager
 
     private bool shouldEstablishConnection (Address address)
     {
-        return !this.isOurOwnAddress(address) &&
-            !this.banman.isBanned(address) &&
+        return !this.banman.isBanned(address) &&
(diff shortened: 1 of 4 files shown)


kchulj (approved):
LGTM
but some tests fail.

Geod24:
> but some tests fail.

Because the test is broken: https://github.com/bpfkorea/agora/issues/1738
Will fix this sprint

omerfirmak:
https://github.com/Geod24/localrest/pull/121

Should fix the hangs in this PR.

Geod24:
> Should fix the hangs in this PR.

Oh yeeeees

PR #1739, opened 2021-03-07
Ledger: Separate functions relevant to Validators in their own class

### LINKED ISSUE #1550: Validation-related routines should be moved out of Ledger

Geod24:
Fixes #1550


hewison-chris (approved):
LGTM

PR #1742, opened 2021-03-08
Flash: Fix failing test

omerfirmak:
With retries this test is doomed to fail since there is no
alternative path. Retries are tested in another test anyways.

### DIFF
--- a/source/agora/test/Flash.d
+++ b/source/agora/test/Flash.d
@@ -395,15 +395,6 @@ unittest
     bob.waitForUpdateIndex(bob_charlie_chan_id, 2);
     charlie.waitForUpdateIndex(bob_charlie_chan_id, 2);
 
-    // Second payment attempt should fail because of insufficient funds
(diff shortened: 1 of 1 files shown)


source/agora/test/Flash.d
```diff
@@ -395,15 +395,6 @@ unittest
     bob.waitForUpdateIndex(bob_charlie_chan_id, 2);
     charlie.waitForUpdateIndex(bob_charlie_chan_id, 2);
 
-    // Second payment attempt should fail because of insufficient funds
-    auto inv_2 = charlie.createNewInvoice(Amount(4_000), time_t.max, "payment 2");
-    alice.payInvoice(inv_2);
-
-    // wait for payment + folding update indices
-    alice.waitForUpdateIndex(alice_bob_chan_id, 4);
```
AndrejMitrovic, line 403: In fact I think what this test did was add and revert HTLCs. But it didn't check what the update tx looks like after that.

Well anyway it's good to go. :)

AndrejMitrovic (approved):

PR #1744, opened 2021-03-09
Minor fixes around `SecretKey`

Geod24:
As mentioned in #1702 we should remove `SecretKey`. Those commits aim in that direction, by clearing up the field around it (removing one of the only direct usage), along with some little style fixes found along the way.

### DIFF
--- a/source/agora/common/crypto/Key.d
+++ b/source/agora/common/crypto/Key.d
@@ -85,6 +85,23 @@ public struct KeyPair
         KeyPair kp = KeyPair.fromSeed(Seed.fromString(seedStr));
     }
 
+    /***************************************************************************
+
+        Signs a message with this keypair's private key
+
+        Params:
+          msg = The message to sign
+
+        Returns:
+          The signature of `msg` using `this`
+
+    ***************************************************************************/
+
+    public Signature sign (scope const(ubyte)[] msg) const nothrow @nogc
+    {
+        return this.secret.sign(msg);
+    }
+
     /// Generate a new, random, keypair
     public static KeyPair random () @nogc
     {
--- a/source/agora/registry/API.d
+++ b/source/agora/registry/API.d
@@ -43,13 +43,13 @@ public struct RegistryPayload
     public Signature signature;
 
     ///
-    public void signPayload (const ref SecretKey secret_key) nothrow
+    public void signPayload (in SecretKey secret_key) nothrow
     {
         this.signature = secret_key.sign(hashFull(this.data)[]);
     }
 
     ///
-    public bool verifySignature (const ref PublicKey public_key) const nothrow @nogc @safe
+    public bool verifySignature (in PublicKey public_key) const nothrow @nogc @safe
     {
         return public_key.verify(this.signature, hashFull(this.data)[]);
     }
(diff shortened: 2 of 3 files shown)


source/agora/consensus/validation/Block.d
```diff
@@ -431,7 +431,7 @@ unittest
 
     // don't accept block height 0 from the network
     block.header.height = 0;
-    assert(block.isNotValid(Height(0), Hash.init, null,
+    assert(!block.isValid(Height(0), Hash.init, null,
         Enrollment.MinValidatorCount, Enrollment.MinValidatorCount, checker, findGenesisEnrollments));
```
hewison-chris: I think this will now log something even though we expect it to fail the check as it is not valid.

source/agora/consensus/validation/Block.d
```diff
@@ -693,20 +693,6 @@ version (unittest)
...
-
-    /// Ditto but returns `bool`, only usable in unittests
-    public bool isNotValid (in Block block, Height prev_height,
-        Hash prev_hash, scope UTXOFinder findUTXO,
-        size_t active_enrollments, size_t enrolled_validators, scope FeeChecker checkFee,
-        scope EnrollmentFinder findEnrollment, Hash random_seed = Hash.init,
-        ulong enrollment_cycle = 0, ulong prev_time_offset = 0, ulong curr_time_offset = ulong.max,
-        Duration block_time_tolerance = 100.seconds) nothrow @safe
-    {
-        return isValidcheck(block, prev_height, prev_hash, findUTXO,
```
hewison-chris: This one is dong the check but not logging. `isValid` will log if it is not. See comment in line 678.

hewison-chris (changes requested):
It was useful to get reason logged when running tests that are expected to be valid but are not. Maybe I should have added a comment to the `isNotValid` function also.

source/agora/registry/API.d
```diff
@@ -43,7 +43,7 @@ public struct RegistryPayload
     public Signature signature;
 
     ///
-    public void signPayload (in SecretKey secret_key) nothrow
+    public void signPayload (in KeyPair secret_key) nothrow
```
linked0: `key_pair` would be better for the parameter.

Geod24:
How I see now, thanks for the clarification. I think we should go with something like `assertValid` instead of `isValid` called from `assert` for the other function then.

source/agora/common/crypto/Key.d
```diff
@@ -85,6 +85,23 @@ public struct KeyPair
...
+          msg = The message to sign
+
+        Returns:
+          The signature of `msg` using `this`
+
+    ***************************************************************************/
+
+    public Signature sign (scope const(ubyte)[] msg) const nothrow @nogc
+    {
+        return agora.crypto.Schnorr.sign(Pair(this.secret.data, this.address.data), msg);
```
linked0: Suggested change.
```
return this.secret.sign(msg);
```

linked0 (approved):
LGTM with some requests.

source/agora/registry/API.d
Geod24: Good catch! Will go with `kp` though, as it's a common abbreviation.

Geod24:
Comments addressed: Applied most of @linked0 's diff and removed the last commit for now.

hewison-chris (commented):
LGTM

hewison-chris (approved):

PR #1746, opened 2021-03-09
Tests: Use improved addNewNode instead of repeating its code

Geod24:
This makes the code more DRY and generic.
We probably should factor the pattern of having only one node
injected in the network as well, but we'll leave that for another commit.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -758,13 +758,35 @@ public class TestAPIManager
             this.addNewNode!TestFullNode(conf, file, line);
     }
 
-    /// Convenience templated function to be called from overriding classes
-    public TestAPI addNewNode (NodeType : TestAPI) (
-        Config conf, string file = __FILE__, int line = __LINE__)
+    /***************************************************************************
+
+        Convenience templated function to be called from overriding classes
+
+        Params:
+          conf = The configuration for this node (usually forwarded from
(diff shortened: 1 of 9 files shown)


source/agora/test/Base.d
```diff
@@ -758,13 +758,35 @@ public class TestAPIManager
...
+        Note:
+          The "extra arguments" parameter, `eArgs`, makes a few assumptions
+          which might not hold in the future, most importantly:
+          - `TestFullNode` and `TestValidatorNode` have the same ctor args;
+          - Arguments for `NodeType` are in the same order as its parent;
+
+    ***************************************************************************/
+
+    public TestAPI addNewNode (NodeType : TestAPI) (Config conf,
+        Parameters!(NodeType.__ctor)[Parameters!(TestValidatorNode.__ctor).length .. $] eArgs,
```
Geod24, line 784: Tried factoring this monster out in its own `template`, but got the dreaded "is used as a type" error.
Could also go with:
```D
        Parameters!(NodeType.__ctor)[6 .. $] eArgs,
```

But I figured in this case the verbose version might be better ?

hewison-chris (approved):
LGTM. Is more DRY though.

Geod24:
Fixed commit message

kchulj (approved):

PR #1748, opened 2021-03-09
Minor: Improve error message for self-slashing ConsensusData

Geod24:
Since there was a typo in the message ('Slasing'),
might as well reword the log message to be less scary,
and decrease it to 'warn' as it *could* be triggered under normal circumstances,
but still should be checked by the node operator.


linked0 (approved):

kchulj (approved):
LG

PR #1757, opened 2021-03-10
Don't compare full blocks in the tests

AndrejMitrovic:
This makes the tests fail if the number
of signers doesn't exactly match between
the nodes - which is an expected occurence.

Instead just compare the block header hashes.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -1101,9 +1101,11 @@ public class TestAPIManager
                     (file, line, height, idx,
                         this.clients[idx].getBlockHeight())));
 
-        retryFor(client_idxs.map!(idx =>
-            this.clients[idx].getAllBlocks()).uniq().count() == 1,
-            5.seconds,
+        Block[] all_blocks = cast(Block[])(client_idxs.map!(idx =>
+            this.clients[idx].getAllBlocks()).joiner.array);
+        all_blocks.sort!((a, b) => a.header.hashFull > b.header.hashFull);
+
+        assert(all_blocks.map!(b => b.header.hashFull).uniq().count() == height + 1,
             format!"[%s:%s] Clients %s blocks are not all the same: %s"
                 (file, line, client_idxs, client_idxs.fold!((s, i) =>
                     s ~ format!"\n\n========== Client #%s ==========%s"


hewison-chris (changes requested):
There are tests where this check is required though. Perhaps it can be made optional so only those tests that really need to check that all signatures are updated are in fact updated.

AndrejMitrovic:
> There are tests where this check is required though. Perhaps it can be made optional so only those tests that really need to check that all signatures are updated are in fact updated.

Yeah that makes sense.

Also I think signature propagation should be fixed. This is the Simple test that fails.

AndrejMitrovic:
> There are tests where this check is required though. Perhaps it can be made optional so only those tests that really need to check that all signatures are updated are in fact updated.

Which specific test is it?

hewison-chris:
> > There are tests where this check is required though. Perhaps it can be made optional so only those tests that really need to check that all signatures are updated are in fact updated.
> 
> Which specific test is it?

At least test.PeriodicCatchup.d should include this check. I would also expect Simple to pass if given enough time.

source/agora/test/Base.d
```diff
@@ -1101,9 +1101,11 @@ public class TestAPIManager
                     (file, line, height, idx,
                         this.clients[idx].getBlockHeight())));
```
ferencdg, line 1103: there might be a little bit of race condition here:

1. at line 1097 all clients has a block height of 10 including client 1, however client 1 is too busy to respond to the testing framework
3. all clients almost ready for accepting block with height 11
4. client 1 is finally able to report block height 10, so line 1097 returns
5. all clients accepted block with height 11
6. the assert will trigger at 1108

this has probably low chance to manifest, but we have big timing delays in test and small block creation(1sec/block I think)

source/agora/test/Base.d
```diff
@@ -1101,9 +1101,11 @@ public class TestAPIManager
...
 
-        retryFor(client_idxs.map!(idx =>
-            this.clients[idx].getAllBlocks()).uniq().count() == 1,
-            5.seconds,
+        Block[] all_blocks = cast(Block[])(client_idxs.map!(idx =>
+            this.clients[idx].getAllBlocks()).joiner.array);
+        all_blocks.sort!((a, b) => a.header.hashFull > b.header.hashFull);
```
ferencdg, line 1106: we could get rid of the sorting/hashing and use some linear algorithm, if we want
(x is client_idxs.map!(idx => this.clients[idx].getAllBlocks())

```D
    int[] a = [1, 2, 3, 4];
    int[] b = [10, 20, 30, 40];
    int[] c = [100, 200, 300, 400];
    auto x = [a, b, c];
    iota(x.length).map!(i => x.transversal(i)).joiner.writeln;
```

Geod24:
@AndrejMitrovic : Where are we on this ? I see some outstanding comments by @ferencdg .

AndrejMitrovic:
> At least test.PeriodicCatchup.d should include this check. I would also expect Simple to pass if given enough time.

Shouldn't 5 seconds be enough time? Should I just increase the value to 10?

AndrejMitrovic:
It seems that https://github.com/bosagora/agora/pull/1860/commits/5705e8ac1a83052ce7344384f96f5b3538d3bd1d covers this, I'll close my PR.

PR #1758, opened 2021-03-10
Use a single timer for gossipping

Geod24:
```
Instead of calling wait a gazillion times,
which creates a timer every time with Vibe.d,
use a single recurring timer.
```

Also included a minor style fix that didn't warrant his own PR.

### DIFF
--- a/source/agora/network/Client.d
+++ b/source/agora/network/Client.d
@@ -164,7 +164,7 @@ public class NetworkClient
         this.exception = new Exception(
             format("Request failure to %s after %s attempts", address,
                 max_retries));
-        this.gossip_timer = this.taskman.setTimer(0.seconds, &this.gossipScheduler);
+        this.gossip_timer = this.taskman.setTimer(10.msecs, &this.gossipTask, Periodic.Yes);
     }
 
     /// Shut down the gossiping timer
@@ -175,20 +175,14 @@ public class NetworkClient
 
     /// For gossiping we don't want to block the calling fiber, so we use
     /// request queueing and a separate fiber to handle all the outgoing requests.
-    private void gossipScheduler ()
+    private void gossipTask ()
     {
-        while (1)
+        while (!this.gossip_queue.empty)
         {
-            while (!this.gossip_queue.empty)
-            {
-                auto event = this.gossip_queue.front;
-                this.gossip_queue.removeFront();
-                this.handleGossip(event);
-                // yield and reschedule for next event
-                this.taskman.wait(0.msecs);
-            }
-
-            this.taskman.wait(10.msecs);  // no events, yield with longer sleep
+            auto event = this.gossip_queue.front;
+            this.gossip_queue.removeFront();
+            this.handleGossip(event);
+            // yield and reschedule for next event
         }
     }
 
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -22,30 +22,31 @@ module agora.test.Base;
 
 version (unittest):
 
+import agora.api.FullNode : NodeInfo, NetworkState;
+import agora.api.Validator : ValidatorAPI = API;
 import agora.common.Amount;
 import agora.common.BanManager;
 import agora.common.BitField;
 import agora.common.Config;
-import agora.common.Types;
 import agora.common.Metadata;
 import agora.common.Set;
 import agora.common.Task;
-import agora.node.TransactionPool;
+import agora.common.Types;
 import agora.common.crypto.Key;
 import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
-import agora.consensus.protocol.Data;
+import agora.consensus.data.genesis.Test;
 import agora.consensus.data.Params;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.ValidatorBlockSig;
 import agora.consensus.data.Transaction;
-import agora.consensus.state.UTXODB;
 import agora.consensus.EnrollmentManager;
-import agora.consensus.data.genesis.Test;
-import agora.consensus.SCPEnvelopeStore;
+import agora.consensus.protocol.Data;
+import agora.consensus.Fee;
 import agora.consensus.protocol.Nominator;
 import agora.consensus.Quorum;
-import agora.consensus.Fee;
+import agora.consensus.SCPEnvelopeStore;
+import agora.consensus.state.UTXODB;
 import agora.crypto.Hash;
 import agora.network.Client;
 import agora.network.Clock;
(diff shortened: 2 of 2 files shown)


hewison-chris (approved):
LGTM

AndrejMitrovic (approved):
much better, ty

source/agora/network/Client.d
```diff
@@ -164,7 +164,7 @@ public class NetworkClient
         this.exception = new Exception(
             format("Request failure to %s after %s attempts", address,
                 max_retries));
-        this.gossip_timer = this.taskman.setTimer(0.seconds, &this.gossipScheduler);
+        this.gossip_timer = this.taskman.setTimer(10.msecs, &this.gossipTask, Periodic.Yes);
```
linked0, line 167: Is it safe to call `gossipTask` again even if the function called previously does not return?

Geod24: Good point! Let me test something.

kchulj (approved):

source/agora/network/Client.d
Geod24: I checked, it's safe

hewison-chris: Can't we stop the gossip_timer whilst we run gossipTask just to prevent them running on top of each other?

AndrejMitrovic: I don't understand the problem here. There is only a single fiber here at play, it's not going to trigger exactly every 10ms. It runs the function and doesn't return until it's done. It may yield, but there will **not** be another fiber spawned for this task.

Unless I'm misunderstanding how vibe.d does it..

Geod24: The function *might* (and probably be) will re-entered on yield (Vibe will spawn a new fiber if needed), but that isn't a problem even if the queue is not empty as popping an element will be done atomically.

Geod24:
```D
import vibe.d;

void slowTimer ()
{
    import core.thread;
    import std.stdio;

    static size_t count;
    auto lc = count;
    ++count;

    writefln("\t[%s] Timer before sleep: %s", lc, MonoTime.currTime());
    Thread.sleep(3.seconds);
    writefln("[%s] Timer after first sleep: %s", lc, MonoTime.currTime());

    requestHTTP("http://www.example.org/",
                (scope req) {
                    req.method = HTTPMethod.POST;
                    //req.writeJsonBody(["name": "My Name"]);
                },
                (scope res) {
                    writefln("[%s] Timer after response: %s", lc, MonoTime.currTime());
                }
        );
    writefln("[%s] Timer after query: %s", lc, MonoTime.currTime());
    Thread.sleep(3.seconds);
    writefln("[%s] Timer after second sleep: %s", lc, MonoTime.currTime());
}

int main (string[] args)
{
    setTimer(1.seconds, () => slowTimer(), true);
    return runApplication(&args);
}
```

```
Running ./test
	[0] Timer before sleep: MonoTime(66717500241453 ticks, 1000000000 ticks per second)
[0] Timer after first sleep: MonoTime(66720504006136 ticks, 1000000000 ticks per second)
	[1] Timer before sleep: MonoTime(66720504659656 ticks, 1000000000 ticks per second)
[1] Timer after first sleep: MonoTime(66723508781841 ticks, 1000000000 ticks per second)
	[2] Timer before sleep: MonoTime(66723509319560 ticks, 1000000000 ticks per second)
[2] Timer after first sleep: MonoTime(66726511478762 ticks, 1000000000 ticks per second)
	[3] Timer before sleep: MonoTime(66726512053924 ticks, 1000000000 ticks per second)
[3] Timer after first sleep: MonoTime(66729516220444 ticks, 1000000000 ticks per second)
	[4] Timer before sleep: MonoTime(66729516763934 ticks, 1000000000 ticks per second)
[4] Timer after first sleep: MonoTime(66732517022578 ticks, 1000000000 ticks per second)
	[5] Timer before sleep: MonoTime(66732518065253 ticks, 1000000000 ticks per second)
[5] Timer after first sleep: MonoTime(66735521812352 ticks, 1000000000 ticks per second)
	[6] Timer before sleep: MonoTime(66735522926300 ticks, 1000000000 ticks per second)
[6] Timer after first sleep: MonoTime(66738527103039 ticks, 1000000000 ticks per second)
[0] Timer after response: MonoTime(66738527609189 ticks, 1000000000 ticks per second)
[0] Timer after query: MonoTime(66738527781622 ticks, 1000000000 ticks per second)
[0] Timer after second sleep: MonoTime(66741530188874 ticks, 1000000000 ticks per second)
	[7] Timer before sleep: MonoTime(66741530589788 ticks, 1000000000 ticks per second)
[7] Timer after first sleep: MonoTime(66744532153784 ticks, 1000000000 ticks per second)
	[8] Timer before sleep: MonoTime(66744532474669 ticks, 1000000000 ticks per second)
[8] Timer after first sleep: MonoTime(66747534910757 ticks, 1000000000 ticks per second)
[1] Timer after response: MonoTime(66747535630245 ticks, 1000000000 ticks per second)
[1] Timer after query: MonoTime(66747535733046 ticks, 1000000000 ticks per second)
[1] Timer after second sleep: MonoTime(66750538149349 ticks, 1000000000 ticks per second)
[2] Timer after response: MonoTime(66750538377852 ticks, 1000000000 ticks per second)
[2] Timer after query: MonoTime(66750538445595 ticks, 1000000000 ticks per second)
[2] Timer after second sleep: MonoTime(66753541046982 ticks, 1000000000 ticks per second)
	[9] Timer before sleep: MonoTime(66753541239478 ticks, 1000000000 ticks per second)
[9] Timer after first sleep: MonoTime(66756543348677 ticks, 1000000000 ticks per second)
	[10] Timer before sleep: MonoTime(66756543759955 ticks, 1000000000 ticks per second)
[10] Timer after first sleep: MonoTime(66759548736211 ticks, 1000000000 ticks per second)
[3] Timer after response: MonoTime(66759549401285 ticks, 1000000000 ticks per second)
[3] Timer after query: MonoTime(66759549542981 ticks, 1000000000 ticks per second)
```

Geod24: The tabulated line is the function being entered

AndrejMitrovic: So vibe.d tries to guarantee this function is called every 10ms. LocalRest doesn't simulate this behavior though. So I wonder if we'll end up with bugs in production not caught during testing..

AndrejMitrovic: And the `wait` technique worked fine for LocalRest, but not for vibe.d. sigh

Geod24: One take away from this: We don't have to compensate for drift.

Geod24: @AndrejMitrovic : I tried to document the semantic a bit better: https://github.com/vibe-d/vibe-core/pull/255
We should probably align LocalRest with it.

ferencdg: this is the (dangerous) behavior of 'Periodic.Yes' that I tried to mention here too
https://github.com/bpfkorea/agora/issues/312#issuecomment-760896537

PR #1762, opened 2021-03-10
Tasks: Separate Vibe.d implementation from the interface

Geod24:
Ideally, we should be able to completely ignore Vibe.d when
building the test target, provided all usages of Vibe.d have
been properly abstracted.
This is currently not possible for a number of reasons,
including the fact that the API and the implementations depend on Vibe.d.
Separating the `TaskManager` gives us one less reason.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -334,7 +334,7 @@ public alias RemoteAPI (APIType) = geod24.LocalRest.RemoteAPI!(APIType, Serializ
 
 *******************************************************************************/
 
-public class LocalRestTaskManager : TaskManager
+public class LocalRestTaskManager : ITaskManager
 {
     static import geod24.LocalRest;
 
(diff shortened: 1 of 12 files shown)


source/agora/test/Base.d
```diff
@@ -349,7 +349,7 @@ public class LocalRestTaskManager : TaskManager
 
     public override void runTask (void delegate() dg) nothrow
     {
-        this.tasks_started++;
+        //this.tasks_started++;
```
AndrejMitrovic: Deliberate change?

source/agora/test/Base.d
```diff
@@ -389,7 +389,7 @@ public class LocalRestTaskManager : TaskManager
     public override ITimer setTimer (Duration timeout, void delegate() dg,
         Periodic periodic = Periodic.No) nothrow
     {
-        this.tasks_started++;
+        //this.tasks_started++;
```
AndrejMitrovic: ditto

source/agora/test/Base.d
Geod24: Nope, reverted

source/agora/test/Base.d
Geod24: Also

AndrejMitrovic (approved):

PR #1764, opened 2021-03-10
Remove duplicated 'sign' method from SecretKey

Geod24:
Since KeyPair is aimed at being the replacement for Pair,
we should only expose the 'sign' primitive here.
This way, we prevent re-generating the public key every time
we sign.


hewison-chris (approved):
LGTM

linked0 (approved):

PR #1765, opened 2021-03-11
SendTxProcess: Simplify signature code

Geod24:

PR #1766, opened 2021-03-11
Nominator: Use a KeyPair instead of crypto's Pair

Geod24:
```
The former is intended to be used in Agora, and is a wrapper around the former.
This also removes the duplication between the pair and node_public_key.
```

The middle commit was the main intent. In the process, I realized that `sign` should take more than just an array of bytes (which will unblock #1765) and that the keypair was used from `node.Validator` (??), so both issues were fixed.


linked0 (approved):
LGTM

kchulj (approved):
LGTM

PR #1767, opened 2021-03-11
Flash: Store a KeyPair instead of a Pair

Geod24:
We're moving away from `Pair`, and using `KeyPair` everywhere now,
since it contains the component of a `Pair` itself.
To keep the diff reviewable, this only affect places where `Pair`
was used. A later change will use `PublicKey` instead of `Point`.

### DIFF
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -30,6 +30,7 @@ import agora.consensus.data.UTXO;
 import agora.consensus.data.Transaction;
 import agora.crypto.ECC;
 import agora.crypto.Hash;
+import agora.crypto.Key;
 import agora.crypto.Schnorr;
 import agora.script.Engine;
 import agora.script.Lock;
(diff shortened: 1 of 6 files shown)


kchulj (approved):
LGTM

kchulj:
However, there is this log:
```
Could not print logs for node: Request to void geod24.LocalRest.RemoteAPI!(TestFlashAPI, Serializer).RemoteAPI.printLog() @safe couldn't be processed : base32.Base32Exception@submodules/base32/src/base32.d(1033): Invalid source length
308
```

linked0 (approved):

Geod24:
> Could not print logs for node: Request to void geod24.LocalRest.RemoteAPI!(TestFlashAPI, Serializer).RemoteAPI.printLog() @safe couldn't be processed : base32.Base32Exception@submodules/base32/src/base32.d(1033): Invalid source length

Added a commit to fix this.

kchulj:
Do we need to re-run the CI?

Geod24:
Nope, something is wrong, need to debug

PR #1768, opened 2021-03-11
Define and implement block rewards

### LINKED ISSUE #246: Define and implement block rewards
### Premise

The white-paper defines a few things w.r.t. rewards: categorization, amount, randomness...
However, freezing rewards will be removed, as they are a form of passive income which would make our coin a security.
While the total amount of rewards is likely to stay the same, the distribution among the various categories will change to align incentives with the goals stated in the white-paper.

### Incentive

One of the goal of the white-paper is to separate the political incentive from the economical incentive.
The direction we are heading towards is to have low rewards for simple validators, and increasingly higher rewards for L2 providers (validators with > 40k).
Since L2 liquidity cannot be slashed (that'd be double spend), we assume that every validator has exactly 40k.

Additionally we need to give incentive for validators to:
- Accept new validators in the validator set;
- Ensure that pre-images and other critical data are correctly propagated;

In the second case, the lack of propagation / cooperation (e.g. by withholding pre-images) could be used as an attempt to slash a well-behaving node.

prerequisite: #2245

ferencdg:
fixes: #246

### DIFF
--- a/source/agora/common/Amount.d
+++ b/source/agora/common/Amount.d
@@ -86,6 +86,14 @@ public struct Amount
         return ret;
     }
 
+    /// Non-throwing version of toString
+    /// Use it only for debugging when called from nothrow function
+    public string toStringNT () const @safe nothrow
+    {
+        try return this.toString();
+        catch (Exception e) {return "Exception happened when calling Amount.toStringNT()";}
+    }
+
     /// Support for Vibe.d deserialization
     public static Amount fromString (scope const(char)[] str) pure @safe
     {
--- a/source/agora/common/Amount.d
+++ b/source/agora/common/Amount.d
@@ -2,9 +2,9 @@
 
    Defines a monetary type used in the blockchain
 
-   This struct can hold an amount between 0 and 500_000_000 * 10^7,
-   as there is an absolute maximum supply of 500M and each coin can be divided
-   in up to 10^7 units.
+   This struct can hold an amount between 0 and 4_948_159_353 * 10^7,
+   as there is an absolute maximum supply of 4_948_159_353 coins and
+   each coin can be divided in up to 10^7 units.
 
    This struct does not expose operation overloading on purpose.
    Having operator overloading would mean that any error should be reported
@@ -33,7 +33,7 @@ public struct Amount
     /// Number of units ('cents') per coins
     public static immutable Amount UnitPerCoin   = Amount(10_000_000, true);
     /// Maximum number of BOA coins that can ever be in circulation
-    public static immutable ulong  MaxCoinSupply = 500_000_000;
+    public static immutable ulong  MaxCoinSupply = 4_948_159_353;
     /// Maximum amount of money that can ever be in circulation
     public static immutable Amount MaxUnitSupply =
         Amount(UnitPerCoin.value * MaxCoinSupply, true);
@@ -448,8 +448,8 @@ nothrow pure @nogc @safe unittest
 pure @safe unittest
 {
     import std.exception;
-    assert(Amount.fromString(`5000000000000000`) == Amount.MaxUnitSupply);
-    assertThrown!Exception(Amount.fromString(`5000000000000001`));
+    assert(Amount.fromString(`49481593530000000`) == Amount.MaxUnitSupply);
+    assertThrown!Exception(Amount.fromString(`49481593530000001`));
     Amount maxv = Amount.MaxUnitSupply;
     assert(maxv.isValid());
     assert(!maxv.add(Amount(1)));
@@ -461,7 +461,7 @@ pure @safe nothrow @nogc unittest
     assert(Amount.UnitPerCoin.integral() == 1);
     assert(Amount.UnitPerCoin.decimal() == 0);
 
-    assert(Amount.MaxUnitSupply.integral() == 500_000_000);
+    assert(Amount.MaxUnitSupply.integral() == 4_948_159_353);
     assert(Amount.MaxUnitSupply.decimal() == 0);
 
     assert(Amount(500).integral() == 0);
--- a/source/agora/common/Set.d
+++ b/source/agora/common/Set.d
@@ -109,6 +109,12 @@ public struct Set (T)
             this.put(key);
     }
 
+    /// Returns true if the set has the element e, otherwise returns false
+    public auto opBinary (string op)(in T e) const nothrow pure @nogc if (op == "in")
+    {
+        return e in _set;
+    }
+
     /***************************************************************************
 
         Serialization support
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -84,7 +84,7 @@ public class EnrollmentManager
 {
     /// The period for revealing a preimage
     /// It is an hour interval if a block is made in every 10 minutes
-    public static immutable uint PreimageRevealPeriod = 6;
+    public const uint PreimageRevealPeriod;
 
     /// Logger instance
     private Logger log;
@@ -166,11 +166,13 @@ public class EnrollmentManager
             cacheDB = The cache database, used by `EnrollmentPool`
             key_pair = the keypair of the owner node
             params = the consensus-critical constants
+            preimage_reveal_period = the period for revealing a preimage
 
     ***************************************************************************/
 
     public this (ManagedDatabase stateDB, ManagedDatabase cacheDB,
-                 KeyPair key_pair, immutable(ConsensusParams) params)
+                 KeyPair key_pair, immutable(ConsensusParams) params,
+                 uint preimage_reveal_period = 6)
     {
         this.log = Logger(__MODULE__);
         assert(params !is null);
@@ -182,6 +184,7 @@ public class EnrollmentManager
         this.validator_set = new ValidatorSet(stateDB, params);
         this.enroll_pool = new EnrollmentPool(cacheDB);
 
+        this.PreimageRevealPeriod = preimage_reveal_period;
 
         // create the table for enrollment data for a node itself
         this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
@@ -239,6 +242,27 @@ public class EnrollmentManager
             return ulong.max;
         }
         return this.keymap.key_to_index[height][K];
+    }
+
+    /***************************************************************************
+
+        Params:
+            height = the height at which to look up the mapping for
+            index = the index for which to find the associated Key
+
+        Returns:
+            the key belonging to this index,
+            or `PublicKey.init` if none was found
+
+    ***************************************************************************/
+
+    public PublicKey getValidatorAtIndex (in Height height, in ulong index)
+        const @safe nothrow
+    {
+        if (height !in this.keymap.index_to_key || index !in this.keymap.index_to_key[height])
+            return PublicKey.init;
+        else
+            return this.keymap.index_to_key[height][index];
     }
 
     /***************************************************************************
--- a/source/agora/consensus/data/Params.d
+++ b/source/agora/consensus/data/Params.d
@@ -88,7 +88,7 @@ public immutable class ConsensusParams
     public this (immutable(Block) genesis,
                  in PublicKey commons_budget_address,
                  ConsensusConfig config = ConsensusConfig.init,
-                 Duration block_interval = 1.seconds)
+                 Duration block_interval = 1.seconds) @safe pure nothrow
     {
         this.Genesis = genesis;
         this.CommonsBudgetAddress = commons_budget_address,
@@ -99,7 +99,7 @@ public immutable class ConsensusParams
     /// Default for unittest, uses the test genesis block
     version (unittest) public this (
         uint validator_cycle = 20, uint max_quorum_nodes = 7,
-        uint quorum_threshold = 80, Amount min_fee = 0)
+        uint quorum_threshold = 80, Amount min_fee = 0) @safe pure nothrow
     {
         import agora.consensus.data.genesis.Test : GenesisBlock;
         import agora.utils.WellKnownKeys;
--- a/source/agora/consensus/data/Params.d
+++ b/source/agora/consensus/data/Params.d
@@ -40,6 +40,7 @@ module agora.consensus.data.Params;
 
 import agora.common.Amount;
 import agora.consensus.data.Block;
+import agora.consensus.Reward;
 import agora.crypto.Key;
 
 import core.time;
@@ -72,6 +73,13 @@ public immutable class ConsensusParams
     mixin ROProperty!("PayoutPeriod", "payout_period");
     mixin ROProperty!("GenesisTimestamp", "genesis_timestamp");
     mixin ROProperty!("MinFee", "min_fee");
+    mixin ROProperty!("BlockRewardFactorA", "block_reward_factor_a");
+    mixin ROProperty!("BlockRewardFactorB", "block_reward_factor_b");
+    mixin ROProperty!("BlockRewardFactorC", "block_reward_factor_c");
+    mixin ROProperty!("BlockRewardGap", "block_reward_gap");
+    mixin ROProperty!("BlockRewardDelay", "block_reward_delay");
+    mixin ROProperty!("ValidatorBlockRewards", "validator_block_rewards");
+    mixin ROProperty!("FoundationBlockRewards", "foundation_block_rewards");
 
     /***************************************************************************
 
@@ -93,6 +101,9 @@ public immutable class ConsensusParams
         this.Genesis = genesis;
         this.CommonsBudgetAddress = commons_budget_address,
         this.BlockInterval = block_interval;
+
+        if (!config.validator_block_rewards.length)
+            config.validator_block_rewards = ConsensusConfig.DefaultValidatorBlockRewards;
         this.data = config;
     }
 
@@ -108,6 +119,7 @@ public immutable class ConsensusParams
             max_quorum_nodes: max_quorum_nodes,
             quorum_threshold: quorum_threshold,
             min_fee: min_fee,
+            validator_block_rewards: ConsensusConfig.DefaultValidatorBlockRewards,
         };
         this(GenesisBlock, CommonsBudget.address, config);
     }
@@ -148,6 +160,71 @@ public struct ConsensusConfig
     /// The minimum (transaction size adjusted) fee.
     /// Transaction size adjusted fee = tx fee / tx size in bytes.
     public Amount min_fee = Amount(700);
+
+    /// The number of blocks between 2 block reward payout
+    /// The value of 3 means block reward payout happens at the block 1, 4, 7...
+    public ushort block_reward_gap = 10;
+
+    /// factor 'a' in the block reward reduction function of f(x) = b*e^(a*x) + c
+    public double block_reward_factor_a = 0.046;
+
+    /// factor 'b' in the block reward reduction function of f(x) = b*e^(a*x) + c
+    public double block_reward_factor_b = 1;
+
+    /// factor 'c' in the block reward reduction function of f(x) = b*e^(a*x) + c
+    public double block_reward_factor_c = 1;
+
+    /// The delay (measured in blocks) after which block reward is payed
+    /// The value of 5 means the block reward payout for confirming blocks
+    /// [(X-confirmation_payout_gap + 1), (X)] happens at block (X + block_reward_delay)
+    public ushort block_reward_delay = 5;
+
+    /// Block rewards given to the validators for 128 years
+    public immutable(BlockRewardsTup)[] validator_block_rewards;
+
+    /// Block rewards given to the foundation for roughly 6 years
+    public immutable(BlockRewardsTup)[] foundation_block_rewards =
+    [
+         mbr(315_360_000_0000000), mbr(315_360_000_0000000), mbr(315_360_000_0000000),
+         mbr(315_360_000_0000000), mbr(315_360_000_0000000), mbr(223_200_000_0000000, 22_320_000)
+    ];
+
+    /// Default block rewards given to validators
+    /// Not assigned directly to `validator_block_rewards` here to avoid
+    /// 'cannot inline default argument' compiler error
+    private static immutable BlockRewardsTup[] DefaultValidatorBlockRewards =
+    [
+        mbr(170_294_400_0000000), mbr(159_548_823_0000000), mbr(149_481_293_0000000), mbr(140_049_023_0000000),
+        mbr(131_211_930_0000000), mbr(122_932_457_0000000), mbr(115_175_419_0000000), mbr(107_907_850_0000000),
+        mbr(101_098_865_0000000), mbr(94_719_526_0000000), mbr(88_742_724_0000000), mbr(83_143_058_0000000),
+        mbr(77_896_731_0000000), mbr(72_981_448_0000000), mbr(68_376_318_0000000), mbr(64_061_773_0000000),
+        mbr(60_019_475_0000000), mbr(56_232_246_0000000), mbr(52_683_991_0000000), mbr(49_359_631_0000000),
+        mbr(46_245_039_0000000), mbr(43_326_977_0000000), mbr(40_593_044_0000000), mbr(38_031_623_0000000),
+        mbr(35_631_828_0000000), mbr(33_383_460_0000000), mbr(31_276_963_0000000), mbr(29_303_387_0000000),
+        mbr(27_454_343_0000000), mbr(25_721_974_0000000), mbr(24_098_918_0000000), mbr(22_578_276_0000000),
+        mbr(21_153_587_0000000), mbr(19_818_795_0000000), mbr(18_568_229_0000000), mbr(17_396_574_0000000),
+        mbr(16_298_850_0000000), mbr(15_270_393_0000000), mbr(14_306_831_0000000), mbr(13_404_070_0000000),
+        mbr(12_558_273_0000000), mbr(11_765_846_0000000), mbr(11_023_421_0000000), mbr(10_327_843_0000000),
+        mbr(9_676_156_0000000), mbr(9_065_591_0000000), mbr(8_493_552_0000000), mbr(7_957_609_0000000),
+        mbr(7_455_484_0000000), mbr(6_985_043_0000000), mbr(6_544_287_0000000), mbr(6_131_342_0000000),
+        mbr(5_744_454_0000000), mbr(5_381_979_0000000), mbr(5_042_376_0000000), mbr(4_724_203_0000000),
+        mbr(4_426_105_0000000), mbr(4_146_818_0000000), mbr(3_885_154_0000000), mbr(3_640_001_0000000),
+        mbr(3_410_317_0000000), mbr(3_195_126_0000000), mbr(2_993_513_0000000), mbr(2_804_623_0000000),
+        mbr(2_627_651_0000000), mbr(2_461_846_0000000), mbr(2_306_504_0000000), mbr(2_160_963_0000000),
+        mbr(2_024_606_0000000), mbr(1_896_854_0000000), mbr(1_777_162_0000000), mbr(1_665_023_0000000),
+        mbr(1_559_960_0000000), mbr(1_461_527_0000000), mbr(1_369_305_0000000), mbr(1_282_901_0000000),
+        mbr(1_201_950_0000000), mbr(1_126_107_0000000), mbr(1_055_050_0000000), mbr(988_476_0000000),
+        mbr(926_103_0000000), mbr(867_666_0000000), mbr(812_917_0000000), mbr(761_622_0000000), mbr(713_563_0000000),
+        mbr(668_537_0000000), mbr(626_353_0000000), mbr(586_830_0000000), mbr(549_801_0000000), mbr(515_108_0000000),
+        mbr(482_605_0000000), mbr(452_153_0000000), mbr(423_622_0000000), mbr(396_891_0000000), mbr(371_847_0000000),
+        mbr(348_384_0000000), mbr(326_401_0000000), mbr(305_805_0000000), mbr(286_509_0000000), mbr(268_430_0000000),
+        mbr(251_492_0000000), mbr(235_623_0000000), mbr(220_755_0000000), mbr(206_825_0000000), mbr(193_775_0000000),
+        mbr(181_548_0000000), mbr(170_092_0000000), mbr(159_359_0000000), mbr(149_304_0000000), mbr(139_883_0000000),
+        mbr(131_056_0000000), mbr(122_786_0000000), mbr(115_038_0000000), mbr(107_780_0000000), mbr(100_979_0000000),
+        mbr(94_607_0000000), mbr(88_637_0000000), mbr(83_044_0000000), mbr(77_804_0000000), mbr(72_895_0000000),
+        mbr(68_295_0000000), mbr(63_986_0000000), mbr(59_948_0000000), mbr(56_165_0000000), mbr(52_621_0000000),
+        mbr(49_301_0000000), mbr(46_190_0000000), mbr(43_275_0000000)
+    ];
 }
 
 /// Inserts properties functions aliasing `ConsensusConfig`
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -26,6 +26,7 @@ import agora.common.Config;
 import agora.common.ManagedDatabase;
 import agora.common.Set;
 import agora.common.Types;
+import agora.consensus.Reward;
 import agora.consensus.data.Block;
 import agora.consensus.protocol.Data;
 import agora.consensus.data.Enrollment;
@@ -38,6 +39,7 @@ import agora.consensus.Fee;
 import agora.consensus.validation;
 import agora.consensus.validation.Block : validateBlockTimeOffset;
 import agora.consensus.Fee;
+import agora.crypto.ECC : Point;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.network.Clock;
@@ -115,6 +117,9 @@ public class Ledger
     /// but less than current time + block_time_offset_tolerance
     public Duration block_time_offset_tolerance;
 
+    /// Block Reward Calculator
+    private Reward reward;
+
     /***************************************************************************
 
         Constructor
@@ -154,6 +159,7 @@ public class Ledger
         this.clock = clock;
         this.block_time_offset_tolerance = block_time_offset_tolerance;
         this.storage.load(params.Genesis);
+        this.reward = new Reward(params);
 
         // ensure latest checksum can be read
         this.last_block = this.storage.readLastBlock();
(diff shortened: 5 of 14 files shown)


source/agora/test/BlockReward.d
```diff
@@ -0,0 +1,87 @@
+/*******************************************************************************
```
ferencdg, line 1: Will add 2 more tests: missing signatures and missing validators

source/agora/consensus/validation/Block.d
```diff
@@ -148,13 +148,6 @@ public string isInvalidReason (in Block block, Engine engine, Height prev_height
     if (only_coinbase)
         return "Block: Must contain other transactions than Coinbase";
 
-    auto expected_cb_txs = getCoinbaseTX(block.txs,
```
ferencdg: point 5
https://github.com/bpfkorea/agora/issues/246#issuecomment-791075528

source/agora/test/Fee.d
```diff
@@ -214,10 +214,6 @@ unittest
     {
         createAndExpectNewBlock(Height(block_idx));
     }
-
```
ferencdg, line 215: point 5
https://github.com/bpfkorea/agora/issues/246#issuecomment-791075528

source/agora/node/Ledger.d
```diff
@@ -609,19 +744,20 @@ public class Ledger
         Params:
             tx_set = Transaction set to generate the CoinBase TX for
             missing_validators = MPVs
+            height = the height for which the coinbase transaction should be returned
 
         Returns:
-            List of expected Coinbase TXs
+            One or zero Coinbase TX
```
ferencdg: explanation
https://github.com/bpfkorea/agora/pull/1768/commits/0cd1c01abddd12ac3c63358d31470bb3d66a0120#diff-d25eeef12241e794638a748cea9cc8e7b124f1d61ccdaa962f7fb64ed3c66cb6R725

actually, even before my change we had numerous asserts that this function only returns 1 or less elements in the array

source/agora/test/Fee.d
omerfirmak: Does this mean a node can be fed blocks with wrong CB txs? Do we just trust the peer when we are catching up?

ferencdg: > Does this mean a node can be fed blocks with wrong CB txs? Do we just trust the peer when we are catching up?

We are still checking if at least 50% of the validators signed the block, so not just trusting only 1 peer. 

------ background

However, Mathias described an attack before
1. in 2020, lets say we have 20 validators, 2 are malicious, 8 are lazy(they don't sign blocks), and 10 are well-behaving. Those 10 well-behaving is able to create a block in 2020.
2. in 2022, 8 validators(from 2020) sell their private keys(6 previously lazy ones and 2 previously well-behaving one are the sellers).
3. now those 2 malicious validators from 2020 along with the newly bought private keys can dupe a 'catching up' a validator.

Our block reward(based on validators signing blocks) will try to minimize the number of lazy validators, so let's in 2020 we didn't have 8 lazy validators, just 2, so the block was signed by 6+10=16 validators... With block reward in mind the person in 2022 have to buy more private keys, because we will also change our catchup logic at some point to the following one:

each validator needs to accept the block with the MOST signatures at a specific height

----- end of background

so if we implement all above, then a catching up validator can only be duped, if someone can create a block with let's say 60%-70% of all signatures

ferencdg:
ready for review - squashed some commits and cleaned up the code

source/agora/common/Amount.d
```diff
@@ -86,6 +86,13 @@ public struct Amount
         return ret;
     }
 
+    /// Non-throwing version of toString
+    public string toStringNT () const @safe nothrow
```
AndrejMitrovic, line 91: Why is this needed? Commit doesn't explain it.

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,650 @@
+/*******************************************************************************
+
+    Contains functions to calculate the block reward.
+
+    1. We will only distribute block rewards in every X blocks
+       as opposed to distributing it in every block. This was originally meant
+       to reduce blockchain size.
+    2. We will delay the distribution of the block rewards, so for
+       blocks A .. B we will distribute the rewards at block B + Y.
```
AndrejMitrovic, line 9: What is Y here?

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,650 @@
...
+       This is because we want to give enough time validators to share
+       block signatures which is the basis of the block reward.
+    3. Each validator will be rewarded based on how many blocks they signed
+       and the reward will NOT be propotional to how much coin they staked.
+    4  The block reward is divided between the validators and the Foundation.
```
AndrejMitrovic: s/Foundation/Commons Budget?

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,650 @@
...
+import std.array;
+import std.math;
+import std.range : repeat;
+import std.typecons : Tuple, tuple;
+
+import core.time;
+
+mixin AddLogger!();
+
+public class BlockRewardCalc
```
AndrejMitrovic: Needs `///` or ddoc is not generated.

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,650 @@
...
+
+    Returns: the validator reward for each validator based on how many blocks
+    each validator signed
+
+    ***************************************************************************/
+
+    public Amount[Point] getEachValidatorReward (in Height height, ushort[Point] validator_stakes,
+        out Amount remainder, double comply_perc, uint indent = 0) @safe const nothrow
+    {
+        auto indentStr = ' '.repeat(indent).array;
```
AndrejMitrovic: This will cause an allocation even if trace logging is disabled. Could perhaps turn it into a function?

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,650 @@
...
+        log.trace(indentStr ~ "Calculating each validator reward");
+        Amount[Point] each_validator_rewards;
+        Amount total_validator_reward = getReducedValidatorReward(height, comply_perc, indent + indentInc);
+        Amount gcd_validator_reward = total_validator_reward;
+
+        ulong total_stakes;
+        foreach (ref stake; validator_stakes.byValue())
```
AndrejMitrovic: `const ref`?

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,650 @@
...
+            total_stakes += stake;
+        assert(total_stakes != 0);
+
+        Amount min_validator_reward = total_validator_reward;
+        min_validator_reward.div(total_stakes);
+
+        Amount actual_total_validator_reward;
+        foreach (pkey_stake; validator_stakes.byKeyValue())
```
AndrejMitrovic: Maybe also `const ref` here?

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,650 @@
...
+            Amount validator_reward = min_validator_reward;
+            validator_reward.mul(pkey_stake.value);
+            each_validator_rewards[pkey_stake.key] = validator_reward;
+            actual_total_validator_reward.mustAdd(validator_reward);
+        }
+        remainder = total_validator_reward;
+        remainder.mustSub(actual_total_validator_reward);
+        foreach (amount_point_pair; each_validator_rewards.byKeyValue())
+            log.trace(indentStr ~ "{} -> {}", amount_point_pair.key, amount_point_pair.value);
+        log.trace(indentStr ~ "Remainder is: {}", remainder.toStringNT());
```
AndrejMitrovic: Can use comma instead of concatenation.

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,650 @@
...
+    Returns: the validator reward for each validator based on how many blocks
+    each validator signed
+
+    ***************************************************************************/
+
+    public Amount[Point] getEachValidatorReward (in Height height, ushort[Point] validator_stakes,
+        out Amount remainder, double comply_perc, uint indent = 0) @safe const nothrow
+    {
+        auto indentStr = ' '.repeat(indent).array;
+        log.trace(indentStr ~ "Calculating each validator reward");
```
AndrejMitrovic: Can use comma instead of concatenation.

source/agora/consensus/data/Params.d
```diff
@@ -1,4 +1,4 @@
-/*******************************************************************************
+ 		 /*******************************************************************************
```
AndrejMitrovic: Oops

source/agora/node/Ledger.d
```diff
@@ -356,6 +361,74 @@ public class Ledger
...
+    {
+        Point[] keys;
+        Block block;
+        try
+        {
+            block = this.storage.readBlock(height);
+        }
+        catch (Exception e) {
+            log.error("unable to read block at height: {}", height);
+            return keys;
```
AndrejMitrovic: Doesn't this just return `null` then?

AndrejMitrovic (commented):
Just a few comments.

I'm not sure whether using floating point is a good idea, there were reservations against using it in the past.

ferencdg:
> Just a few comments.
> 
> I'm not sure whether using floating point is a good idea, there were reservations against using it in the past.

Do you by any chance remember why floating point was not recommended? The only thing I could think is
1. double is kind of a waste, it takes 64 bytes
2. some cpu's use different floating point units and precision might differ -> it is not an issue in our case as I think 'double' type is standardized, so arithmetic on them should be the same on every fpu(vs 'real' type in dlang)
3. precision loss: 0.1 has a finite decimal representation, but infinete binary representation(0.00011001100110011011....) -> we don't care about this

the exponential function that I use for 'block reward reduction' returns a flotaing point value, so the only think I could do is represent the input(comply percentage) to that exponential function in decimal(I just multiply it by 10.000 let's say), so I can pass around that input parameter to different functions. However when I use it as the input in the exponential function, I still have to convert it back to floating point. (I don't think we want to create our own integer based exponential function like 2^x and then try to somehow fit the return value between 0 - 100 so it can be used as reduction percentage)

source/agora/common/Amount.d
ferencdg: > Why is this needed? Commit doesn't explain it.

I had some 'nothrow' functions from which I had to print out the 'Amount', I will modify the comment to make it more obvious that this 'non-throwing' version should mainly be used for debugging.

source/agora/consensus/BlockRewardCalc.d
ferencdg: > This will cause an allocation even if trace logging is disabled. Could perhaps turn it into a function?

hmm, Ocean's format function(called by log.trace) is defined as 
```D
public void format (Args...) (Level level, cstring fmt, Args args) @safe
```
so it cannot just lazily evaluate the argument.

the only thing I can think of is 
```D
class IndentClass(int indentation)
{
    void toString (scope FormatterSink sink)
    {
        sink(' '.repeat(indent).array);
    }
}
```

and then create an instance of this class on the stack(or just declare it as struct), and then pass that instance to log.trace :) 

do you think it is worth the effort?

AndrejMitrovic: Maybe it's overkill then. Let's fix it later if it's not a big deal.

ferencdg:
thanks for the review @AndrejMitrovic, if anyone else has a comment, please let me know, so this PR could be merged this sprint

source/agora/common/Amount.d
```diff
@@ -86,6 +86,14 @@ public struct Amount
         return ret;
     }
 
+    /// Non-throwing version of toString
+    /// Use it only for debugging when called from nothrow function
+    public string toStringNT () const @safe nothrow
+    {
+        try return this.toString();
+        catch (Exception e) {return "";}
```
omerfirmak: Maybe this return string should indicate the exception.

ferencdg: I was thinking returning null, but that is probably a bad idea: it would require the call site to do extra checks. So as you suggested I will just return some error message.

ferencdg:
rebased, let me know if any more changes needed

source/agora/common/Set.d
```diff
@@ -79,6 +79,12 @@ public struct Set (T)
             this.put(key);
     }
 
+    /// Returns true if the set has the element e, otherwise returns false
+    public auto opBinary (string op)(in T e) nothrow const pure @safe @nogc if (op == "in")
```
Geod24: Type ctor (`const`, `inout`, `shared`, etc...) always come first.
Also it's a template so you don't need to specify all attributes.

source/agora/common/Amount.d
Geod24: It's not used anywhere and stands out, I'd remove it.
If you need to debug you can use the `scope (failure) assert(0);` hack.

source/agora/test/Common.d
```diff
@@ -0,0 +1,101 @@
...
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.Common;
```
Geod24: That's what `Base` is for.

source/agora/test/Common.d
```diff
@@ -0,0 +1,101 @@
...
+import agora.consensus.data.PreImageInfo;
+import agora.consensus.state.UTXOSet;
+import agora.test.Base;
+
+import core.atomic;
+
+// EnrollmentManager that might or migh not reveal preimage based on the value
+// current value of an atomic shared variable
+version (unittest)
+package class MissingPreImageEM : EnrollmentManager
```
Geod24: It's a good idea to make them reusable, but those classes have a few issues. One of them being `public static shared UTXOSet utxo_set;`

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+    Copyright:
+        Copyright (c) 2021 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.consensus.BlockRewardCalc;
```
Geod24:
```suggestion
module agora.consensus.Rewards;
```

Let's keep it simple and spellable

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+       nonlinear function of the the total signatures,
+       so 2 times more signatures might increase the total validator reward by 4.
+       This is to encourage validators to share signatures.
+       The coins remaing is  'fixed block reward at height X' -
+       'total validator reward' will go to the Commons Budget.
+    6. The part of the total validator reward that cannot be didived equally
+       between the validators will also go to the Commons Budget
+
+    Copyright:
+        Copyright (c) 2021 BOS Platform Foundation Korea
```
Geod24:
```suggestion
        Copyright (c) 2019-2021 BOSAGORA Foundation
```

source/agora/consensus/data/Params.d
```diff
@@ -125,6 +137,71 @@ public struct ConsensusConfig
...
+    /// factor 'b' in the block reward reduction function of f(x) = b*e^(a*x) + c
+    public double block_reward_factor_b = 1;
+
+    /// factor 'c' in the block reward reduction function of f(x) = b*e^(a*x) + c
+    public double block_reward_factor_c = 1;
+
+    /// The delay (measured in blocks) after which block reward is payed
+    /// The value of 5 means the block reward payout for confirming blocks
+    /// [(X-confirmation_payout_gap + 1), (X)] happens at block (X + block_reward_delay)
+    public ushort block_reward_delay = 5;
```
Geod24, line 180: Do those need to be configurable ? If not they should be directly in `ConsensusParams`.

Added some information about why here: https://github.com/bosagora/agora/pull/1819

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
+/*******************************************************************************
+
+    Contains functions to calculate the block reward.
+
+    1. We will only distribute block rewards in every X blocks
+       as opposed to distributing it in every block. This was originally meant
+       to reduce blockchain size.
```
Geod24: Try to use a more neutral language when writing documentation. E.g. "We will only distribute" => "We will distribute". Also try to avoid usage of "we", e.g. "Block rewards are distributed every X blocks [...]".

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
+/*******************************************************************************
+
+    Contains functions to calculate the block reward.
+
+    1. We will only distribute block rewards in every X blocks
+       as opposed to distributing it in every block. This was originally meant
```
Geod24: originally ?

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+       to reduce blockchain size.
+    2. We will delay the distribution of the block rewards by Y blocks, so for
+       blocks A .. B we will distribute the rewards at block B + Y.
+       This is because we want to give enough time validators to share
+       block signatures which is the basis of the block reward.
+    3. Each validator will be rewarded based on how many blocks they signed
+       and the reward will NOT be propotional to how much coin they staked.
```
Geod24: Likewise, if you feel the lack of proportionality needs emphasis, something like "[...] how many blocks they signed, regardless of the amount staked". Or "Rewards do not depend on the amount staked by a validator.".

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+       however this doesn't mean that the total validator reward is fixed.
+       The total validator reward is based on the total number of signatures
+       on the blocks in question. The total validator reward is a
+       nonlinear function of the the total signatures,
+       so 2 times more signatures might increase the total validator reward by 4.
+       This is to encourage validators to share signatures.
+       The coins remaing is  'fixed block reward at height X' -
+       'total validator reward' will go to the Commons Budget.
+    6. The part of the total validator reward that cannot be didived equally
+       between the validators will also go to the Commons Budget
```
Geod24: A few typos (`didived`, `remaing`, ...). Also this doesn't feel much like documentation, as it lacks structure.
I would recommend to state the goals first (study supply, reward active participants, etc..), then expand on the actual system property and the details of the implementation after.

source/agora/common/Amount.d
```diff
@@ -33,7 +33,7 @@ public struct Amount
     /// Number of units ('cents') per coins
     public static immutable Amount UnitPerCoin   = Amount(10_000_000, true);
     /// Maximum number of BOA coins that can ever be in circulation
-    public static immutable ulong  MaxCoinSupply = 500_000_000;
+    public static immutable ulong  MaxCoinSupply = 4_948_159_353;
```
Geod24, line 36: When `Amount` was created, the goal was simply to ensure we never go over the limit. The idea was to later add a function which would give us the coin supply based on the block height. Do you think that's something we can achieve or it would create too much trouble (e.g. by forcing us to pass extra parameters everywhere) ?

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+    'Periods' represent time periods and are stored in the block_rewards array.
+
+    Params:
+        height = the height for which we want to get the period information
+        block_rewards = time periods
+
+    Returns: which 'period' the particular height falls into and whether that height
+    is the last height for that particular period
+
+    ***************************************************************************/
```
Geod24: Indentation is wrong

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+    {
+        ulong blocks_accumulated = 0;
+        foreach (period; 0 .. block_rewards.length)
+        {
+            blocks_accumulated += block_rewards[period].dist_period_secs / consensus_params.BlockInterval.total!"seconds";
+            if (blocks_accumulated >= height)
+                return CurrentPeriodTup(period, blocks_accumulated == height);
+        }
+        return CurrentPeriodTup(block_rewards.length, false);
+    }
```
Geod24: Doesn't look like it's using anything from the class, should it be either a free function, or at least a `static` one ?

source/agora/consensus/BlockRewardCalc.d
Geod24: There's a simple fix: https://github.com/sociomantic-tsunami/ocean/blob/ba566c473249341633d723904a0b82ba1201a55b/src/ocean/util/log/Logger.d#L385-L391

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+              the blocks for which we want to calculate the block reward
+        indent = number of spaces before log messages
+
+    Returns: the validator reward for each validator based on how many blocks
+    each validator signed
+
+    ***************************************************************************/
+
+    public Amount[Point] getEachValidatorReward (in Height height, ushort[Point] validator_stakes,
+        out Amount remainder, double comply_perc, uint indent = 0) @safe const nothrow
```
Geod24: Using `PublicKey` instead of `Point`

source/agora/node/Ledger.d
```diff
@@ -356,6 +361,74 @@ public class Ledger
...
+        catch (Exception e) {
+            log.error("unable to read block at height: {}", height);
+            return null;
+        }
+
+        immutable validator_cnt = enroll_man.getCountOfValidators(height);
+        foreach (i; 0 .. validator_cnt)
+        {
+            auto validator_pkey = enroll_man.getValidatorAtIndex(height, i);
+            if (block.header.validators[i] && !block.header.missing_validators.canFind(i))
```
Geod24, line 362: The second check will never be false, because a validator that signed had to sign using its pre-image, hence there's no way to be missing yet sign. If you are missing, you get slashed.

source/agora/node/Ledger.d
```diff
@@ -356,6 +361,74 @@ public class Ledger
...
+        Params:
+            height = block height
+
+        Returns: the public keys of the validators that signed the block at
+                 the specified height, and and also not
+                 classified as 'missing' validator
+
+    ***************************************************************************/
+
+    private Point[] getPKeysOfSignedNonMissingValidators (Height height) nothrow @safe
```
Geod24: It should not be possible to be both `missing` and signing, as explained under this comment.
Also, perhaps this function could use with a better name, such as `getRewardEligibleValidators` ? It's a more semantic / abstract name, which makes sense *unless* we have a need to get the list of signers for other purposes ?

source/agora/node/Ledger.d
```diff
@@ -356,6 +361,74 @@ public class Ledger
...
+
+    ***************************************************************************/
+
+    private ushort[Point] getPKeysOfSignedNonMissingValidators (Height start_height, Height end_height) nothrow @safe
+    {
+        ushort[Point] signed_cnt_keys_map;
+
+        foreach (height; start_height .. end_height + 1)
+            foreach (point; getPKeysOfSignedNonMissingValidators(height))
+                signed_cnt_keys_map[point]++;
```
Geod24, line 396: Does this work ? I would expect it to throw.

source/agora/node/Ledger.d
```diff
@@ -356,6 +361,74 @@ public class Ledger
...
+        Point[] keys;
+        Block block;
+        try
+        {
+            block = this.storage.readBlock(height);
+        }
+        catch (Exception e) {
+            log.error("unable to read block at height: {}", height);
+            return null;
+        }
```
Geod24, line 355: This is not handled at the call site. In general, we should probably just let it `throw`, much better to do so than to propose invalid consensus data.

source/agora/node/Ledger.d
```diff
@@ -579,17 +656,77 @@ public class Ledger
...
+            immutable period_end = Height(height - params.BlockRewardDelay);
+            immutable period_start = Height(period_end - params.BlockRewardGap + 1);
+            Amount remainder;
+
+            // calculate the penalty for missing singatures
+            auto pkey_sigcounts = getPKeysOfSignedNonMissingValidators(period_start, period_end);
+            double total_actual_sigcount = pkey_sigcounts.byValue().fold!((a, b) => a + b)(0);
+            log.trace("Total actual signatures count: {}", total_actual_sigcount);
+            double total_expected_sigcount = iota(period_start.value, period_end.value + 1)
+                .map!(height => this.enroll_man.getCountOfValidators(Height(height))).fold!((a, b) => a + b)(cast(size_t) 0);
```
Geod24, line 685: Uh, why the `cast` ?

source/agora/node/Ledger.d
```diff
@@ -579,17 +656,77 @@ public class Ledger
...
+        ///////
+        if (block_reward_calc.is_payout_time(height))
+        {
+            immutable period_end = Height(height - params.BlockRewardDelay);
+            immutable period_start = Height(period_end - params.BlockRewardGap + 1);
+            Amount remainder;
+
+            // calculate the penalty for missing singatures
+            auto pkey_sigcounts = getPKeysOfSignedNonMissingValidators(period_start, period_end);
+            double total_actual_sigcount = pkey_sigcounts.byValue().fold!((a, b) => a + b)(0);
```
Geod24: Shouldn't this be an integer ? Why double ? Also I think you can use `sum` here ;)

source/agora/node/Ledger.d
Geod24: Likewise, shouldn't this be an integer, not a double ?

source/agora/node/Ledger.d
```diff
@@ -579,17 +656,77 @@ public class Ledger
...
+                    return amount;
+                }
+            );
+
+            // merge the block reward with fee reward
+            foreach (ref pkey_reward; each_validator_reward.byKeyValue())
+                if (pkey_reward.value != Amount(0))
+                {
+                    if (auto pkey_ind_found = (pkey_reward.key in pkey_to_output_ind))
+                        coinbase_tx.outputs[*pkey_ind_found].value.mustAdd(pkey_reward.value);
```
Geod24: `mustAdd` shouldn't be used outside of tests

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+
+    Params:
+        height = the height of the block that will contain the CoinBase transaction
+
+    Returns:
+        true if block reward should be distributed at that particular height
+
+    ***************************************************************************/
+
+    public bool is_payout_time (Height height) @safe const pure nothrow
```
Geod24: `isPayoutTime` according to our style guide

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+{
+
+    ///
+    public this (immutable(ConsensusParams) consensus_params) @safe @nogc pure nothrow
+    {
+        this.consensus_params = consensus_params;
+    }
+
+    ///
+    private alias CurrentPeriodTup = Tuple!(ulong, "period", bool, "is_period_end");
```
Geod24: A `struct` would also work :)

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+        ulong total_stakes;
+        foreach (const ref stake; validator_stakes.byValue())
+            total_stakes += stake;
+        assert(total_stakes != 0);
+
+        Amount min_validator_reward = total_validator_reward;
+        min_validator_reward.div(total_stakes);
+
+        Amount actual_total_validator_reward;
+        foreach (const ref pkey_stake; validator_stakes.byKeyValue())
```
Geod24: Why the complicated pattern when you can just do:
```D
foreach (key; const ref value; validator_stakes)
```

?

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+    }
+
+    unittest
+    {
+        ushort[Point] validator_stakes;
+        Amount remainder;
+        auto block_reward = getBlockRewardCalc(2, 3, cast(immutable(BlockRewardsTup)[])[BlockRewardsTup(Amount(100), 10)], 1.seconds);
+
+        auto validator1 = Point.fromString("0xab4f6f6e85b8d0d38f5d5798a4bdc4dd444c8909c8a5389d3bb209a186105110");
+        auto validator2 = Point.fromString("0xab4f6f6e85b8d0d38f5d5798a4bdc4dd444c8909c8a5389d3bb209a186105111");
```
Geod24: Use well-known KPs

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+        comply_perc = (#actual signatures / #expected signatures) * 100 on all
+                      the blocks for which we want to calculate the block reward
+        indent = number of spaces before log messages
+
+    Returns:
+        validator whitepaper reward - penalty on validators due to missing signature
+
+    ***************************************************************************/
+
+    public Amount getReducedValidatorReward (in Height height, double comply_perc, uint indent = 0) @safe const nothrow
```
Geod24: Let me tell you a floating point joke:
A man has a computer problem. He decides to use floating point. Now he has 2.00000000001 problems.
Usually, we use an integer with a fixed precision for this kind of use case.

source/agora/consensus/BlockRewardCalc.d
```diff
@@ -0,0 +1,651 @@
...
+            consensus_params.BlockRewardFactorB *
+            exp(comply_perc * consensus_params.BlockRewardFactorA) +
+            consensus_params.BlockRewardFactorC);
+        log.trace("{}comply percentage: {}", indentStr, comply_perc);
+        log.trace("{}reduce percentage: {}", indentStr, reduce_perc);
+        reduce_perc = cast(ubyte) min(reduce_perc, 100);
+
+        if (reduce_perc == 100)
+            return total_validator_reward;
+        total_validator_reward.div(100);
```
Geod24: What about the remainder ?

source/agora/consensus/BlockRewardCalc.d
AndrejMitrovic: I don't quite clearly see why it shouldn't be proportional? Is this explained anywhere?

Geod24 (commented):
Did a first pass.

source/agora/consensus/BlockRewardCalc.d
ferencdg: If I remember correctly it was not my decision, but let me know if I am mistaken (@Geod24). A validator who has X * 40k BOA can and will run multiple Agora instances to maximimze its reward, so I agree with you on this.

source/agora/common/Set.d
ferencdg: not sure I am following this:
'Also it's a template so you don't need to specify all attributes.'

are those attributes stripped after the template instantiation? or are we expecting 
```D
return e in _set;
```
might throw for some instantiation?

Geod24: I mean that you can remove `@safe` because the compiler will check the body, see that it is `@safe`, and mark it `@safe` itself without you needing to do anything.

source/agora/test/Common.d
ferencdg: '
it's not used anywhere and stands out, I'd remove it.
If you need to debug you can use the scope (failure) assert(0); hack.
'

it is used in BlockRewardCalc.getEachValidatorReward

source/agora/test/Common.d
ferencdg: I have an other PR that tries to solve that problem:
https://github.com/bosagora/agora/pull/1817#issuecomment-804515733
any comment on that comment?

I do need to reuse those classes for block reward.

source/agora/consensus/BlockRewardCalc.d
ferencdg: it uses consensus_params

source/agora/node/Ledger.d
ferencdg: I can change, but we have a slashing policy, so what happens if in the future the slashing policy will not be based on pre-images or the signature will not be based pre-images?

source/agora/node/Ledger.d
ferencdg: if you mean line 427, it works

```D
    int[string] m;
    m["hello"]++;
    writeln(m["hello"]);
```

writes out 1

source/agora/node/Ledger.d
ferencdg: This function is very deep in the call hierarchy, and its parents are all nothrow. I can change it, but that will remove nothrow from core functions like valueExternalized, getValidTXSet.

source/agora/node/Ledger.d
ferencdg: doesn't compile without the cast

source/agora/consensus/BlockRewardCalc.d
ferencdg: ehhmm, I tried to explain it here:
https://github.com/bosagora/agora/pull/1768#issuecomment-799178954

any comment on that?

source/agora/consensus/BlockRewardCalc.d
ferencdg: remainder goes to the Commons Budget and calculated by: getExtraCommonsBudgetReward

source/agora/test/Common.d
ferencdg: > That's what `Base` is for.

my plan was to put the higher level classes like Validator/EnrollmentManager to a new file as the module description suggests:

```
Common classes and utilities shared between the network tests.
    Classes and utilities in this module are considered higher level than
    in Base.d.
```

Base.d is already a monster with 2000+ lines of code. I can move those classes to Base.d though

source/agora/consensus/data/Params.d
ferencdg: **block_reward_factor_a
block_reward_factor_b
block_reward_factor_c**
they are very similar to already existing things like
**tx_payload_fee_factor
validator_tx_fee_cut
slash_penalty_amount**
and they are all in ConsensusConfig

**block_reward_gap
block_reward_delay**
they are very similar to already existing things like
**payout_period**
and they are all in ConsensusConfig

Maybe we want to raise a separate ticket and do all that cleanup for all variables together, so we will not have any inconsistencies.

source/agora/common/Amount.d
ferencdg: > When `Amount` was created, the goal was simply to ensure we never go over the limit. The idea was to later add a function which would give us the coin supply based on the block height. Do you think that's something we can achieve or it would create too much trouble (e.g. by forcing us to pass extra parameters everywhere) ?

We would have problems when we want to add let's say 
```D
Amount a = Amount(10000, 2);
Amount b = Amount(20000, 4) 
```
where the 2 and 4 are the heights. Addition will no longer be associative, because 
```D
a.add(b) // might overflow(unless we change a's hight to 4)
b.add(a) // will not overflow(provided that at height 4 the amount of 30000 is okay)
```

Additionally there are around 800+ mentions of Amount(case sensitive/separate word) in our codebase, I think this could be a ticket on its own.

source/agora/consensus/BlockRewardCalc.d
ferencdg: Just wondering if struct would have any benefit over Tuples or if we should refine our coding guidelines. Tuples have some nice methods like expand, translate... Any suggestion on when to use one over the other?

source/agora/consensus/BlockRewardCalc.d
ferencdg: > Using `PublicKey` instead of `Point`

I am depending on other functions like getValidatorAtIndex() that was written long time ago and still uses Point. I think it would be better to clean up all usages at once instead of me doing from Point and to Point conversation whenever I was to call a legacy function.

source/agora/node/Ledger.d
ferencdg: > mustAdd

so if I use 'add', should I check if the resulting Amount is still valid? In case your answer is 'yes I should check' and it is an invalid amount after the addition, then I would like to assert(0). I am confused

source/agora/consensus/BlockRewardCalc.d
ferencdg: got a nasty DLang bug when tried:
source/agora/consensus/BlockRewardCalc.d(146,9): Error: _aaApply2 is not nothrow
source/agora/consensus/BlockRewardCalc.d(128,26): Error: nothrow function agora.consensus.BlockRewardCalc.BlockRewardCalc.getEachValidatorReward may throw

to reproduce
```D
void f() nothrow
{
    int[string] m;
    foreach(const ref k, v; m)
    {
    }
}
int testmain()
{
    f();
    return 0;
}
```

ferencdg:
thanks for the review @Geod24, I fixed what I could, but for the following ones I had suggestions/questions:

https://github.com/bosagora/agora/pull/1768#discussion_r600189742
https://github.com/bosagora/agora/pull/1768#discussion_r600213266
https://github.com/bosagora/agora/pull/1768#discussion_r601000883
https://github.com/bosagora/agora/pull/1768#discussion_r600199453
https://github.com/bosagora/agora/pull/1768#discussion_r601009640
https://github.com/bosagora/agora/pull/1768#discussion_r600201037

If I could get some comment on those that you disagree, then I might be able to finish this task this Sprint.

ferencdg:
rebase #3

ferencdg:
rebase #4

source/agora/consensus/EnrollmentManager.d
```diff
@@ -241,6 +244,27 @@ public class EnrollmentManager
...
+
+    public PublicKey getValidatorAtIndex (in Height height, in ulong index)
+        const @safe nothrow
+    {
+        if (height !in this.keymap.index_to_key || index !in this.keymap.index_to_key[height])
+            return PublicKey.init;
+        else
+            return this.keymap.index_to_key[height][index];
+    }
+
```
hewison-chris, line 267: Would prefer this in a separate commit (or with the changes it is created for) as it is unrelated to the other changes in this commit.

source/agora/test/Common.d
hewison-chris: Why not just make them `public` in one of the test modules and import them in the other test files. Otherwise `common` will become the next monster.

hewison-chris (commented):
I will continue the review after reading the `Rewards` section of the white paper.

source/agora/test/Common.d
ferencdg: I think

1. It would be hard to find that class if we just put it randomly into some of the test files from where we use it.
2. Someone already moved those classes into Base.d anyways in recent v0.x.x branch.

hewison-chris: In that case can you rebase please.

source/agora/node/Ledger.d
```diff
@@ -787,11 +929,6 @@ public class Ledger
             return "Internal error: Could not calculate random seed at current height";
         }
 
-        auto expected_cb_txs = this.getCoinbaseTX(block.txs, block.header.missing_validators);
```
ferencdg, line 790: https://github.com/bosagora/agora/pull/1768#discussion_r592191394

ferencdg:
rebase #5

hewison-chris:
Closing as will continue in new PR https://github.com/bosagora/agora/pull/2415

PR #1771, opened 2021-03-11
KeyPair: Remove 'seed' member variable

Geod24:
```
The Keypair struct was in a weird state:
- It had PublicKey, as it should;
- It has SecretKey, which used to be a public key and a private key,
  and now is just a private key;
- It had the seed, which is the private key,
  but was a leftover from the previous implementation
  and wasn't used anywhere anymore;

This commit removes the unused member. `fromSeed` is still present
and will be handled in a later PR.
```


Geod24:
Ready to go

linked0 (approved):
LG

PR #1783, opened 2021-03-12
Improve utility for vanity key mining

Geod24:
```
- Use Pair instead of KeyPair, ensuring we get correct Scalars;
- Use an extra array of boolean to avoid piggybacking on the first byte;
- Add support for special names (nodes, genesis, commons);
```

### DIFF
--- a/source/agora/cli/vanity/main.d
+++ b/source/agora/cli/vanity/main.d
@@ -19,7 +19,9 @@
 
 module agora.cli.vanity.main;
 
+import agora.crypto.ECC;
 import agora.crypto.Key;
+import agora.crypto.Types;
 
 import core.atomic;
 import std.parallelism;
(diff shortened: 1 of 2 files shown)


source/agora/cli/vanity/main.d
```diff
@@ -67,10 +81,10 @@ void main (string[] args)
     NextKey:
         while (atomicLoad(found) < result.length)
         {
-            auto tmp = KeyPair.random();
+            auto tmp = Pair.random();
 
             // TODO: Use binary to avoid `toString` call
-            const addr = tmp.address.toString();
+            const addr = PublicKey(tmp.V).toString();
```
Geod24, line 87: This can be reverted after #1785

linked0 (approved):

PR #1784, opened 2021-03-12
Implement Bech32 encoding/decoding

### LINKED ISSUE #1655: Switch address format from Stellar-style to bech32
See https://en.bitcoin.it/wiki/BIP_0173 for details. This probably should be addressed only after #1645 has been addressed.

linked0:
These are utilities for converting public keys of `ubyte` array type and bech32-encoded strings. We should switch the address format from Stellar-style to bech32. So this is the first step for the task.

This PR needs more comments and descriptions. And it also needs more tests. So it is now a WIP state.

Fixes #1655


source/agora/common/Bech32.d
```diff
@@ -0,0 +1,232 @@
...
+    combined ~= y;
+    return combined;
+}
+
+/** This function will compute what 6 5-bit values to XOR into the last 6 input values, in order to
+ *  make the checksum 0. These 6 values are packed together in a single 30-bit integer. The higher
+ *  bits correspond to earlier values. */
+uint polymod(const ubyte[] values)
+{
+    // The input is interpreted as a list of coefficients of a polynomial over F = GF(32), with an
```
AndrejMitrovic: This was clearly taken from some other code. Please keep existing copyrights in place.

linked0: I will. Thanks.

linked0:
This PR now only passes in unit tests of `agora.crypto.Key`. In order to make all the unit tests passed, we should change all the key pairs of the `WellKnownKeys` module. So, I'd like to have a whiteboard or pair-programming about it. And there are something more to share and discuss.
@Geod24 I will ask you for a whiteboard tomorrow. Thanks.

linked0:
Need to rebase and some updates

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+        http://www.opensource.org/licenses/mit-license.php.
+
+*******************************************************************************/
+
+module agora.crypto.Bech32;
+
+import agora.crypto.ECC;
+
+/// The Bech32 character set for encoding.
+const char* CHARSET = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
```
Geod24:
```suggestion
private immutable string CHARSET = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+
+/// The Bech32 character set for decoding.
+const byte[128] CHARSET_REV = [
```
Geod24:
```suggestion
private immutable byte[128] CHARSET_REV = [
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+
+    Params:
+        values = 5-bit values to XOR
+
+    Returns:
+        checksum calculated through the polynomial
+
+*******************************************************************************/
+
+uint polymod (const ubyte[] values)
```
Geod24:
```suggestion
private uint polymod (const ubyte[] values) @safe pure nothrow @nogc
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+
+    Params:
+        hrp = human redable part
+
+    Returns:
+        5-bits values for the hrp
+
+*******************************************************************************/
+
+ubyte[] expand_hrp (const string hrp)
```
Geod24:
```suggestion
private ubyte[] expand_hrp (in char[] hrp) @safe pure nothrow
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+    ret.reserve(hrp.length + 90);
+    ret.length = hrp.length * 2 + 1;
+    for (size_t i = 0; i < hrp.length; ++i)
+    {
+        ubyte c = hrp[i];
+        ret[i] = c >> 5;
+        ret[i + hrp.length + 1] = c & 0x1f;
+    }
+    ret[hrp.length] = 0;
+    return ret;
```
Geod24, line 326: So are we returning a string here ? The 0 termination would imply so.
Also we should probably return a slice, not just `ret`. This code seems buggy to me.

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
+/*******************************************************************************
```
Geod24, line 1: Could you move the public functions at the top of this module ?
I expect there's one to encode, one to decode, and they should be the first thing one see when opening this module.

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+
+    Utilities for converting between bech32-encoded string and `ubyte` array.
+
```
Geod24, line 4:
```suggestion

    See_Also:
        https://github.com/sipa/bech32
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+    Params:
+        hrp = human redable part
+        values = values to get checksum from
+
+    Returns:
+        an checksum
+
+*******************************************************************************/
+
+ubyte[] create_checksum (string hrp, const ubyte[] values)
```
Geod24:
```suggestion
ubyte[] create_checksum (in char[] hrp, in ubyte[] values)
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+{
+    ubyte[] enc = expand_hrp(hrp);
+    enc ~= cast(ubyte [])values;
```
Geod24: The cast shouldn't be needed ?

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+    enc.length = enc.length + 6;
+    uint mod = polymod(enc) ^ 1;
```
Geod24: Did you find a good resource about Bech32M ?

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,486 @@
...
+        // CoinNet Genesis Address
+        // GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
+        ValidAddressData(
+            "boa1xzwsyw8q59c5qz7x669fmxe3dtx4zztyjyf6qhpgfapfd54nqy302ft5v67",
+            Point(cast(ubyte[])[
+                0x9D, 0x02, 0x38, 0xE0, 0xA1, 0x71, 0x40, 0x0B,
+                0xC6, 0xD6, 0x8A, 0x9D, 0x9B, 0x31, 0x6A, 0xCD,
+                0x51, 0x09, 0x64, 0x91, 0x13, 0xA0, 0x5C, 0x28,
+                0x4F, 0x42, 0x96, 0xD2, 0xB3, 0x01, 0x22, 0xF5])
+        ),
```
Geod24: Those are self-referential. Would be better to use the reference implementation tests here: https://github.com/sipa/bech32/blob/master/ref/c%2B%2B/tests.cpp

source/agora/crypto/Bech32.d
Geod24: https://github.com/bitcoin/bips/blob/master/bip-0350.mediawiki

source/agora/crypto/Bech32.d
linked0: I see what you mean. Will update.

source/agora/crypto/Bech32.d
linked0: Right. It doesn't need it.

source/agora/crypto/Bech32.d
linked0: Thanks for the information.

source/agora/crypto/Bech32.d
linked0: Can you explain more about what you mean? I will look into it more anyway.

source/agora/crypto/Bech32.d
linked0: By analyzing it, it turns out that the 0 does not mean termination because it's not `char` array, but `ubyte` array.

source/agora/crypto/Bech32.d
Geod24: Those are not good tests, because they are based on the implementation, rather than written independently.
Writing an implementation then adding tests that match it doesn't confirm that we conform to the standard.
Instead, the D implementation should be checked against the same tests that the C++ implementation is using.

linked0: Added more test cases.

linked0:
Ready for review.

source/scpd/types/Stellar_SCP.d
```diff
@@ -321,41 +321,42 @@ struct SCPQuorumSet {
...
 
     const qc1 = toSCPQuorumSet(QuorumConfig(2,
-        [PublicKey.fromString("GBFDLGQQDDE2CAYVELVPXUXR572ZT5EOTMGJQBPTIHSLPEOEZYQQCEWN"),
-         PublicKey.fromString("GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5")]));
+        [PublicKey.fromString("boa1xrv266magda8dc87uche9zvj8842shz3sdavw0qecpgeykyv4ynss5caytq"),
+         PublicKey.fromString("boa1xrv266magd38dc87uche9zvj8842shz3sd3vw0qecpgeykyv4ynssw6rsgy")]));
 
     assert(qc1.hashFull() == Hash.fromString(
-        "0x7b56982f02dfbf3737ff560fe6511674e182d38475f0516cb8b0f338a3156543b0731e66c9a3ced01c238652b12e51c95207ec2bf6eae237f24b08a357a1bd2a"),
+        "0x2a37ad395b2c50d18f2a52e4b6759a8bc2e43f7835ef5f41805d1f717e528943f0de5a15ec9bf4916db6da76f4d1f04e7df0bfdd5d5973a734acce115f07ed67"),
```
Geod24: Why did the hash change ? The quorum set hash only depends on the binary representation of the public key, doesn't it ?

source/scpd/types/Stellar_SCP.d
```diff
@@ -321,41 +321,42 @@ struct SCPQuorumSet {
...
         qc1.hashFull().to!string);
+    import std.stdio;
```
Geod24: Leftover

source/agora/utils/SCPPrettyPrinter.d
```diff
@@ -448,8 +448,8 @@ unittest
     auto pair = WK.Keys.NODE2;
 
     auto qc = QuorumConfig(2,
-        [PublicKey.fromString("GBFDLGQQDDE2CAYVELVPXUXR572ZT5EOTMGJQBPTIHSLPEOEZYQQCEWN"),
-         PublicKey.fromString("GBYK4I37MZKLL4A2QS7VJCTDIIJK7UXWQWKXKTQ5WZGT2FPCGIVIQCY5")]);
+        [PublicKey.fromString("boa1xrv266cegdthdc87uche9zvj8842shz3sdyvw0qecpgeykyv4ynssf79ndd"),
+         PublicKey.fromString("boa1xzq266cegdthdc87uche9zvj8842shz3sdyvw0qecpgeykyv4ynss3dne57")]);
```
Geod24: Those two values seem awfully close to one another, how did you obtain them ?

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,757 @@
...
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1
+];
+
+enum Encoding {
+    Invalid,
+    Bech32,  // Bech32 encoding as defined in BIP173
+    Bech32m, // Bech32m encoding as defined in BIP350
+};
```
Geod24:
```suggestion
}
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,757 @@
...
+
+struct DecodeResult
+{
+    Encoding encoding;  // What encoding was detected in the result
+    char[] hrp;         // The human readable part
+    ubyte[] data;       // The payload (excluding checksum)
+};
```
Geod24:
```suggestion
}
```

source/agora/crypto/Key.d
```diff
@@ -134,27 +137,30 @@ public struct PublicKey
...
+        ubyte[VersionWidth + PublicKey.sizeof] bin;
         bin[0] = VersionByte.AccountID;
-        bin[VersionWidth .. $ - ChecksumWidth] = this.data[];
-        bin[$ - ChecksumWidth .. $] = checksum(bin[0 .. $ - ChecksumWidth]);
-        Base32.encode(bin, sink);
+        bin[VersionWidth .. $] = this.data[];
+        string encoded = encodeBech32(HumanReadablePart, bin,
+            Encoding.Bech32m, true).assumeUnique;
+        foreach (enc_char; encoded)
+            sink([enc_char]);
```
Geod24: The point of this overload is to avoid allocation, and this code completely fails this.
First and foremost, why the foreach ? It's iterating over *every element of the array* then allocating a *1-element array* to pass to the sink. Why not feed `encoded` directly to the sink ???

`encodeBech32` could be made to not allocate, but we will leave this for another PR. But I really can't understand what's the point of this `foreach`...

source/agora/crypto/Key.d
```diff
@@ -134,27 +137,30 @@ public struct PublicKey
     /// Uses Stellar's representation instead of hex
     public string toString () const @trusted nothrow
     {
-        ubyte[VersionWidth + PublicKey.sizeof + ChecksumWidth] bin;
+        scope(failure) assert(0);
```
Geod24: Why ?

source/agora/crypto/Key.d
```diff
@@ -112,6 +113,8 @@ unittest
 
 private immutable int VersionWidth = 1;
 private immutable int ChecksumWidth = 2;
+private immutable int HumanReadableWidth = 3;
```
Geod24: Remove and use `.length`

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,757 @@
...
+            Point(cast(ubyte[])[
+                216, 173, 107, 125, 67, 98, 118, 224,
+                254, 230, 47, 146, 137, 146, 57, 234,
+                168, 92, 81, 131, 98, 199, 60, 25,
+                192, 81, 146, 88, 140, 169, 39, 8])
+        ),
+
+    ];
+
+    foreach(input; addresses_bech32_m)
```
Geod24:
```suggestion
    foreach (input; addresses_bech32_m)
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,757 @@
...
+            Point(cast(ubyte[])[
+                216, 173, 107, 125, 67, 98, 118, 224,
+                254, 230, 47, 146, 137, 146, 57, 234,
+                168, 92, 81, 131, 98, 199, 60, 25,
+                192, 81, 146, 88, 140, 169, 39, 8])
+        ),
+
+    ];
+
+    foreach(input; addresses_bech32)
```
Geod24:
```suggestion
    foreach (input; addresses_bech32)
```

source/agora/crypto/Bech32.d
```diff
@@ -0,0 +1,757 @@
...
+}
+
+// Convert a input string to a lower case string
+version(unittest) char[] toLowerString(const(char)[] input) pure nothrow @safe
+{
+    char[] conv;
+    foreach (c; input)
+        conv ~= toLower(c);
+    return conv;
+}
```
Geod24: There's a function in the std library for this

source/agora/crypto/Key.d
linked0: The `encodeBech32` was not `nothrow` in the previous code, so this is leftover. I removed it.

source/agora/crypto/Bech32.d
linked0: > There's a function in the std library for this

You mean this function, [toLower](https://dlang.org/library/std/uni/to_lower.html#1)?

Geod24: Yes

linked0: Updated, Ty.

source/agora/utils/SCPPrettyPrinter.d
linked0: Only the first byte is different for each public address. The first address has the value, 216 for the first byte, and the second one has 128 for that.

Geod24: Please use the previous key's value so we don't have to update the hash, just the string representation.

linked0: Ah, OK. I will.

source/scpd/types/Stellar_SCP.d
linked0: Will update.

source/agora/crypto/Key.d
linked0: This is because I'm not familiar with the concept of the `sink`. I will look into the use of the `sink` and modify it right now.

linked0:
Ready for review

source/agora/utils/PrettyPrinter.d
```diff
@@ -172,22 +172,22 @@ private struct PubKeyFmt
     {
         try
         {
-            // Public keys are 56 characters, only take the first 8 and last 4
-            // Only format `0xABCDEFGH..JKLM`
-            enum StartUntil = 8;
-            enum EndFrom    = 56 - 4;
-            size_t count;
+            // Public keys are 63 characters, only take the first 12 and last 4
+            // Only format `0xboa1acdefghi..6789`
```
Geod24:
```suggestion
            // Only format `boa1acdefghi..6789`
```

Geod24: Actually I wonder if we should just not print the HRP ?

Geod24 (approved):
LGTM, just fix the one comment and we can merge.

source/agora/utils/PrettyPrinter.d
linked0: I thought about that also. Can't we use it as identifying the network like bitcoin's use case, like `bt`, `tb`?  Anyway, we can remove that now and restore it later if needed.

AndrejMitrovic:
Does this mean we have to mine the keys again?

PR #1785, opened 2021-03-12
Key.random(): Correct to always return valid Scalar

Geod24:
Before this change, `Keypair.random()` would return values which were outside
of the `]0; L[` range (with L=2^252+27742317777372353535851937790883648493).
This was the root cause that led to the situation described in issue 1776.


hewison-chris (approved):
LGTM

PR #1790, opened 2021-03-14
Flash: Use ChannelUpdates for fee calculations

omerfirmak:


source/agora/common/Amount.d
```diff
@@ -285,6 +285,22 @@ public struct Amount
         return this.value / other.value;
     }
 
+    /***************************************************************************
+
+        Params:
```
AndrejMitrovic, line 293: Could use some documentation about what the function actually does.

source/agora/flash/Config.d
```diff
@@ -129,3 +129,30 @@ unittest
...
 }
+
+/***************************************************************************
+
+    Calculate total fee from the given update and payment total
+
+    Params:
+        update = latest update for the channel
+        total = payment total
+
```
AndrejMitrovic, line 140: Missing `Returns`

source/agora/common/Amount.d
omerfirmak: Oops

omerfirmak:
Updated

AndrejMitrovic (approved):

Issue #1772, opened 2021-03-11
Assert triggered in OnionPacket.d:169


Geod24:
We have the following path:
https://github.com/bpfkorea/agora/blob/93b4db119f8809b7cb8681c9f7db098232dbf2d1/source/agora/flash/Node.d#L891-L897

And `getPaymentPath` may return `null`. However the above code doesn't handle it, and we later hit an `assert`:
https://github.com/bpfkorea/agora/blob/93b4db119f8809b7cb8681c9f7db098232dbf2d1/source/agora/flash/OnionPacket.d#L169

This is "possibly a timing issue" according to @AndrejMitrovic (seems to trigger more often on OSX machines which are indeed slower), and can be observed frequently in the CI.

omerfirmak:
Do we have a full stack trace? Multiple tests could trigger that.

Geod24:
https://github.com/bosagora/agora/blob/309989a54529ec708ecab235547662aabdc5740d/source/agora/flash/OnionPacket.d#L172

<img width="1340" alt="Screen Shot 2021-03-31 at 9 47 59" src="https://user-images.githubusercontent.com/2180215/113074565-46c97900-9206-11eb-939e-d60be062f37b.png">

Looks like this isn't fixed yet.

AndrejMitrovic:
Does it say which line in the test file it's coming from?

Geod24:
Unfortunately there's no more log than what you see above :(
It's currently still running: https://github.com/bosagora/agora/pull/1850/checks?check_run_id=2232746880

omerfirmak:
I think we should remove that assert anyways and propagate the failure to the caller.
Would probably give a more helpful error message.

AndrejMitrovic:
Making a PR soon for it.

AndrejMitrovic:
If this doesn't happen again in the next few days then we can close it.

Geod24:
> If this doesn't happen again in the next few days then we can close it.

It didn't.

PR #1791, opened 2021-03-15
Flash tests: wait for channel gossiping

AndrejMitrovic:
I'm pretty sure this is the root cause of https://github.com/bpfkorea/agora/issues/1772.

Obviously retries should be put in place as well. But for now this should fix the tests.


source/agora/test/Flash.d
```diff
@@ -588,7 +602,7 @@ unittest
     charlie.waitChannelOpen(bob_charlie_chan_id);
     /+++++++++++++++++++++++++++++++++++++++++++++/
 
-    /+ CHARLIE BOB => ALICE CHANNEL +/
+    /+ OPEN CHARLIE => ALICE CHANNEL +/
```
omerfirmak, line 605: :smile:

omerfirmak (approved):

PR #1792, opened 2021-03-15
Add `--initialize` and an endpoint to validate the config file

Geod24:
This is the skeleton of the setup interface.
When `--initialize` is passed to Agora, it enters a "setup"
step where it listens on 127.0.0.1:2827 for requests.

It will serve a single-page-application (SPA) codenamed Talos,
which at the end of the setup process will call the `/create` endpoint
with the generated config file. If the config file passes validation,
it will be written to disk, the setup interface / Talos be shut down,
and Agora will start.

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -49,6 +49,9 @@ public struct CommandLine
     /// Path to the config file
     public string config_path = "config.yaml";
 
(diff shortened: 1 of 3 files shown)


AndrejMitrovic (approved):
LGTM.

linked0 (approved):

PR #1793, opened 2021-03-15
Clean up the badges in README

Geod24:
Replace CircleCI with Github CI
Remove microbadger badges, as they've been broken since 2019,
and the site is going to shut down on July 1st, 2021.


kchulj (approved):
LGTM.

PR #1795, opened 2021-03-15
Flash: Care for Channel capacity when routing

omerfirmak:


source/agora/flash/Network.d
```diff
@@ -36,6 +36,19 @@ public class Network
     /// Nodes
     private NetworkNode[Point] nodes;
 
+    private static struct ChannelRoutingInfo
```
AndrejMitrovic, line 40: Needs a little `///` here. :)

source/agora/flash/Network.d
```diff
@@ -36,6 +36,19 @@ public class Network
...
+    {
+        /// Public key of the funder of the channel.
+        public Point funder_pk;
+
+        /// The total amount funded in this channel. This information is
+        /// derived from the Outputs of the funding transaction.
+        public Amount capacity;
```
AndrejMitrovic, line 47: But a capacity also depends on the direction of the payment. Or is this purely used as the first filter ? (e.g. filter by total capacity first, then filter by direction capacity)

Specifically what I mean is:

A <=> B. Capacity: 10 Boa.
...
A sends 6 Boa to B.
...
A <=> B. Capacity: 10 Boa. A may send up to 4 Boa, but B may send up to 6 Boa now.

Do you follow what I mean?

omerfirmak: Yep, but that information is only available to A and B. 

So for a node C, only way is to use previous failed attempts through a channel to estimate its balance. Even if we do that, I am not sure how reliable that information is since it can change at any moment. 

For nodes A and B, I don't think we need any additional logic here. Since the retry mechanism will be immediately triggered without any network activity and origin node will ignore that channel.

AndrejMitrovic: That's right, I completely forgot about that part. And emitting individual updates does not scale (it kind of defeats the purpose of microtransactions).

AndrejMitrovic (approved):

PR #1797, opened 2021-03-15
Upgrade vibe-core to v1.14.0

Geod24:
This fixes a regression w.r.t. timers, ensuring there is only one active call
at a time, and fixing a bug where timers are re-triggered after being stopped.


Geod24:
Good to go

kchulj (approved):
Looks green.

PR #1803, opened 2021-03-17
SlashPolicy unit tests independent of UTXO ordering

### LINKED ISSUE #1497: SlashPolicy unittest implicitly depends on UTXO ordering
These two can fail:

https://github.com/bpfkorea/agora/blob/da49c28bc1663ecbe991b27e27ae46d9f5a3422c/source/agora/consensus/SlashPolicy.d#L414
https://github.com/bpfkorea/agora/blob/da49c28bc1663ecbe991b27e27ae46d9f5a3422c/source/agora/consensus/SlashPolicy.d#L420

The code assumes that `enrollments[0].utxo_key < enrollments[1].utxo_key`, but actually if there is anything in the code which changes how utxo hashes are generated then this code could fail. In fact that's what happens in the `flash-layer` branch. As a temporary workaround I just fixed it by doing:

```diff
diff --git a/source/agora/consensus/SlashPolicy.d b/source/agora/consensus/SlashPolicy.d
index 5aba36e7..95d920a9 100644
--- a/source/agora/consensus/SlashPolicy.d
+++ b/source/agora/consensus/SlashPolicy.d
@@ -411,11 +411,11 @@ unittest
     // get and check random seed
     Hash preimage_root = slash_man.getRandomSeed(Height(2));
     assert(preimage_root ==
-        hashMulti(hashMulti(Hash.init, preimage_1), preimage_2));
+        hashMulti(hashMulti(Hash.init, preimage_2), preimage_1));

     // get and check random seed when a block being externalized
     Hash externalized_seed = slash_man.getExternalizedRandomSeed(Height(2),
         missing_validators);
     assert(externalized_seed ==
-        hashMulti(hashMulti(Hash.init, preimage_1), preimage_2));
+        hashMulti(hashMulti(Hash.init, preimage_2), preimage_1));
 }
```

But this is not reliable against future changes. And the test failing will be confusing to other devs in the future.

It's not trivial to properly fix. I tried by just sorting the `enrollments` array by their utxo key, but then the rest of the code depends implicitly on the mapping between the `enrollments` and `pairs` arrays.

linked0:
Some unit test assumed that `enrollments[0].utxo_key < enrollments[1].utxo_key`. We should tmake the code independent of UTXO ordering.

Fixes #1497

### DIFF
--- a/source/agora/consensus/SlashPolicy.d
+++ b/source/agora/consensus/SlashPolicy.d
@@ -424,14 +424,14 @@ unittest
     assert(validators_utxos.find(first_validator_utxo).empty());
     assert(validators_utxos.find(second_validator_utxo).empty());
 
-    // get and check random seed
+    // get and check random seed for two valid validators
     Hash preimage_root = slash_man.getRandomSeed(Height(2));
-    assert(preimage_root ==
-        hashMulti(hashMulti(Hash.init, preimage_2), preimage_1));
+    assert(preimage_root != Hash.init);
+    assert(preimage_root != hashMulti(Hash.init, preimage_1));
+    assert(preimage_root != hashMulti(Hash.init, preimage_2));
 
     // get and check random seed when a block being externalized
     Hash externalized_seed = slash_man.getExternalizedRandomSeed(Height(2),
         missing_validators);
-    assert(externalized_seed ==
-        hashMulti(hashMulti(Hash.init, preimage_2), preimage_1));
+    assert(externalized_seed == preimage_root);
 }


linked0:
Please review this. This is a simple PR. :)

kchulj (approved):
LGTM.

Geod24:
But the block validation guarantee that they are sorted, does it not ?

linked0:
> But the block validation guarantee that they are sorted, does it not ?

Right. The `random_seed` has originally the meaning of UTXOs sorted. But the ordering is not the focus in this test case, the focus is to check the two random seeds are the same, which are gotten when a block is being nominated and validated. This PR intends to avoid the situation where the ordering of UTXOs makes the test fail.

Geod24 (approved):
Fair point

PR #1805, opened 2021-03-18
Flash: Bound re-try loops in update and payment proposals

omerfirmak:


AndrejMitrovic (approved):

AndrejMitrovic:
The integration test seems to be failing consistently. But it shouldn't, it doesn't use the flash layer.

Edit: seems other PRs are failing too.

hewison-chris:
> The integration test seems to be failing consistently. But it shouldn't, it doesn't use the flash layer.
> 
> Edit: seems other PRs are failing too.

You will need to rebase from upstream.

hewison-chris:
> > The integration test seems to be failing consistently. But it shouldn't, it doesn't use the flash layer.
> > Edit: seems other PRs are failing too.
> 
> You will need to rebase from upstream.

Although maybe we have a new problem:-
`Check condition failed after timeout of 1 minute and 600 attempts: >>CI: node-0 has not completed discovery after 1 minute.`

omerfirmak:
Rebased

omerfirmak:
All green

PR #1810, opened 2021-03-21
Integration test: Checkout and build Faucet

Geod24:


kchulj (approved):
LGTM.

.github/workflows/integration.yml
```diff
@@ -26,14 +27,39 @@ jobs:
         sudo apt-get update
         sudo apt-get install g++-9 libsodium-dev libsqlite3-dev
 
+    - name: Checkout faucet
+      uses: actions/checkout@v2
+      with:
+        repository: bosagora/faucet
```
Geod24: So it turns out that actually checking out faucet instead of agora works better, who would've guessed.

kchulj: Hmm, so it was not checking out without the `repository`?

Geod24: It is checking out Agora (again)

Geod24:
Okay, finally it works.

PR #1812, opened 2021-03-21
Remove usages of populate from EnrollmentManager

Geod24:
populate is not used anymore in non-test code, and should be removed.
In test code that uses it, it can usually be replaced by a different
call to makeEnrollment, such as in every uses in EnrollmentManager.


linked0 (approved):
LGTM.
Please fix the misspelled word, `uusually` in the commit message.

Geod24:
Fixed `uusually` and `diffrent` ;)

kchulj (approved):
LGTM.

PR #1815, opened 2021-03-22
Flash: Add serialization support

### LINKED ISSUE #1741: Serialize & load Flash metadata to disk
We need to store the channel metadata to disk and make sure to load it when we start up the flash nodes.

AndrejMitrovic:
It's missing a test-case which I'll add soon.

Took @omerfirmak's advice and used occasional dumping rather than making it timer-based. I also used a template mixin trick so most of the existing doesn't need to be changed.

Fixes #1741

### DIFF
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -276,7 +276,7 @@ public class Channel
         // initial output allocates all the funds back to the channel creator
         const seq_id = 0;
         const Balance balance = { refund_amount : this.conf.capacity };
-        const Output[] outputs = this.buildBalanceOutputs(balance);
+        Output[] outputs = this.buildBalanceOutputs(balance);
 
         auto pair = this.update_signer.collectSignatures(0, outputs,
             priv_nonce, peer_nonce, this.conf.funding_tx);
@@ -897,7 +897,7 @@ LOuter: while (1)
         auto new_balance = this.buildUpdatedBalance(direction,
             this.cur_balance, payment.amount, payment.payment_hash,
             payment.lock_height, payment.height);
-        const new_outputs = this.buildBalanceOutputs(new_balance);
+        auto new_outputs = this.buildBalanceOutputs(new_balance);
 
         log.info("{}: Created outgoing payment balance request: {}",
             this.kp.address.flashPrettify, new_balance);
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -52,15 +52,6 @@ mixin AddLogger!();
 /// Ditto
 public class Channel
 {
-    /// The static information about this channel
-    public const ChannelConfig conf;
-
-    /// Key-pair used for signing and deriving update / settlement key-pairs
-    public const KeyPair kp;
-
-    /// Whether we are the funder of this channel (`funder_pk == this.kp.V`)
-    public const bool is_owner;
-
     /// Used to publish funding / trigger / update / settlement txs to blockchain
     public const void delegate (Transaction) txPublisher;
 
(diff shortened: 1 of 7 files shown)


AndrejMitrovic:
Need to add `UpdateSigner` serialization too. Working on it.

AndrejMitrovic:
I realize now that I could have used UDAs for this..

AndrejMitrovic:
Fixed serialization issues and added a test-case. There were lots of little details that needed to be taken into account.

AndrejMitrovic:
I'll file a separate issue for converting the mixin into just using UDAs for serialization. And also the need to add hashmap support to the serializer.

source/agora/flash/Node.d
```diff
@@ -812,9 +941,9 @@ public abstract class FlashNode : ControlFlashAPI
         assert(result.error == ErrorCode.None, result.to!string);
 
         auto channel = new Channel(chan_conf, this.kp, priv_nonce, result.value,
-            peer, this.engine, this.taskman, &this.putTransaction,
```
omerfirmak, line 843: I am not seeing why we need this extra `peer_pk` parameter. Is there something I am missing?

AndrejMitrovic: Maybe not! I'll try to remove it. I think I needed it briefly during some refactorings.

omerfirmak:
Could we have any reentrancy issues with `dump()`? I guess `ManagedDatabase` would handle any edge cases.

AndrejMitrovic:
> Could we have any reentrancy issues with dump()? 

With the current assumption of only having one work thread the function will never be entered twice. The routine doesn't yield the fiber either. Do you see another way it might be entered twice?

AndrejMitrovic:
Removed `peer_pk` from the ctor.

source/agora/flash/Channel.d
```diff
@@ -1884,3 +1766,134 @@ LOuter: while (1)
         return this.pending_close.tx;
     }
 }
+
+/// The closing transaction can spend from the funding transaction when
+/// the channel parties want to collaboratively close the channel.
+/// It requires both of the parties signatures. For safety reasons,
+/// the closing transaction should only be signed once the node marks
+/// the channel as 'PendingClose'.
+private static struct PendingClose
```
Geod24: `static` no more needed

source/agora/flash/Channel.d
```diff
@@ -1884,3 +1766,134 @@ LOuter: while (1)
...
+{
+    private Hash payment_hash;
+    private Amount amount;
+    private Height lock_height;
+    private OnionPacket packet;
+    private Height height;
+}
+
+/// All the channel metadata which we keep in the DB for storage
+private mixin template ChannelMetadata ()
```
Geod24, line 2046: This feels weird to me at first, that we have to use a mixin template here.

Geod24 (commented):
Maybe the Serializer isn't the best fit here ? I found myself shifting some data structures from the binary serializer to the SQLite db (e.g. storing `Hash` as strings) so that it's easier to debug. The serializer was only intended for blocks and network communication, originally.

source/agora/flash/Channel.d
AndrejMitrovic: I forgot about UDAs https://github.com/bosagora/agora/pull/1815#issuecomment-803796943

AndrejMitrovic:
> Maybe the Serializer isn't the best fit here ? I found myself shifting some data structures from the binary serializer to the SQLite db (e.g. storing Hash as strings) so that it's easier to debug. The serializer was only intended for blocks and network communication, originally.

Mmmm good point. I just stored everything as a blob and that does makes it hard to debug later. I think turning structs into their own tables is gonna take a little more effort though. How about I file an issue and we fix this up later?

AndrejMitrovic:
The failure seems to be unrelated to this PR. The timeouts happen in other tests but only Flash code was touched.

AndrejMitrovic:
Finally it's green.

AndrejMitrovic:
@omerfirmak would you consider merging this? It's blocking some other work. :>

omerfirmak (approved):

PR #1817, opened 2021-03-22
Using 'shared' variables without synchronization

### LINKED ISSUE #1676: using 'shared' variables without synchronization
Some of our intermittent test failures might be caused by not using synchronization(locks/atomic operations) while using 'shared' variables for example:

SlashingMisbehavingValidator uses
private shared bool* reveal_preimage;
from multiple threads without any syncronization

Some other affected tests:
Byzantine.d
EnrollmentManager.d

ferencdg:
fixes: #1676

### DIFF
--- a/doc/Style.md
+++ b/doc/Style.md
@@ -350,7 +350,7 @@ import agora.test.Base; // Base functions for the test environment
 unittest
 {
     const NodeCount = 4;
-    auto network = makeTestNetwork(NetworkTopology.Simple, NodeCount);
+    auto network = makeTestNetwork!TestAPIManager(NetworkTopology.Simple, NodeCount);
     network.start();
     network.waitForDiscovery();
 
--- a/source/agora/test/SlashingMisbehavingValidator.d
+++ b/source/agora/test/SlashingMisbehavingValidator.d
@@ -17,6 +17,7 @@ module agora.test.SlashingMisbehavingValidator;
 version (unittest):
 
 import agora.common.Config;
+import agora.common.ManagedDatabase;
 import agora.consensus.data.Block;
 import agora.consensus.data.Params;
 import agora.consensus.data.PreImageInfo;
@@ -26,6 +27,8 @@ import agora.consensus.state.UTXOSet;
 import agora.crypto.Schnorr;
 import agora.utils.Test;
 import agora.test.Base;
+
+import std.path : buildPath;
 
 import core.atomic;
 import core.stdc.stdint;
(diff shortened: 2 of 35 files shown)


source/agora/test/SlashingMisbehavingValidator.d
```diff
@@ -63,7 +63,7 @@ private class MissingPreImageEM : EnrollmentManager
 // `MissingPreImageEM` class
 private class NoPreImageVN : TestValidatorNode
 {
-    public static shared UTXOSet utxo_set;
+    public static UTXOSet utxo_set;
```
ferencdg: will put back the 'shared' here, even though there is no compiler enforcement

Geod24:
I was more hoping we could remove the usage of a shared UTXOSet instead of endorsing it.

ferencdg:
> I was more hoping we could remove the usage of a shared UTXOSet instead of endorsing it.

if we want to use the UTXOSet:

First, I was thinking about creating a new method on the test clients that takes a copy of the exisitng UTXOSet and returns it... However UTXOSet uses an in memory database, so not sure how easy would that be to 'copy'. 

I could
1. query the in-memory DB and return a copy all UTXOs and write methods that search on that returned object 
2. look into how to make a copy of the in-memory DB, so I can copy UTXOSet
3. make a new derived class of UTXOSet that I can easily copy(doesn't have in memory-db)
All of the above would have taken me more time, then the solution I chose.

At the end I just synced the UTXOSet, let me know which version you prefer

---------------------------
if we don't want to use the UTXOSet

We could maybe use the header to know who was slashed, but the original test tests for more than just who was slashed...

ferencdg:
also rebased, let me know if any changes are needed

Geod24:
Can you rebase ? Let's go with the solution we have now, and if we end up extending the test interface, this will benefit from it.

ferencdg:
rebased

Geod24 (approved):

Geod24:
The timeout seems consistent.

doc/Style.md
```diff
@@ -350,7 +350,7 @@ import agora.test.Base; // Base functions for the test environment
 unittest
 {
     const NodeCount = 4;
```
ferencdg, line 352: makeTestNetwork can no longer figure out the default type after signature change(seems Dlang compiler problem)

source/agora/test/SlashingMisbehavingValidator.d
```diff
@@ -63,14 +65,15 @@ private class MissingPreImageEM : EnrollmentManager
 // `MissingPreImageEM` class
 private class NoPreImageVN : TestValidatorNode
 {
-    public static shared UTXOSet utxo_set;
+    public __gshared UTXOSet utxo_set;
```
ferencdg, line 69: why '__gshared'? Using shared would have required us to define all methods in UTXOSet as 'shared'

ferencdg:
> The timeout seems consistent.

I regret ever trying to fix this test... Should work now, but the PR needs a second review.

ferencdg:
🟢 let me know if anything else is needed

doc/Style.md
Geod24: Having to modify all tests is a non-starter. I'll take a look at the failure soon. The rest looks good.

Geod24:
Just needs a rebase and let's merge this

ferencdg:
rebased

Geod24 (approved):

PR #1818, opened 2021-03-23
PublicKey.verify: Accept any kind of message

Geod24:
Since we are just forwarding, we should not needlessly restrict the interface.


hewison-chris (approved):
LGTM

PR #1822, opened 2021-03-24
Trivial: Simplify random_seed check

Geod24:
Also add some input parameter validation for good measure.

### DIFF
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -112,6 +112,10 @@ public string isInvalidReason (in Block block, Engine engine, Height prev_height
     if (block.header.height == 0)
         return "Block: Genesis block should be validated using isGenesisBlockInvalidReason";
 
+    assert(random_seed !is Hash.init);
+    // Validate this after the genesis check for better UX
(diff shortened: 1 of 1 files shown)


kchulj (approved):
LGTM.

source/agora/consensus/validation/Block.d
```diff
@@ -116,6 +116,9 @@ public string isInvalidReason (in Block block, Engine engine, Height prev_height
     import std.conv;
     import std.range;
 
+    assert(prev_hash !is Hash.init);
+    assert(random_seed !is Hash.init);
+
     if (block.header.height == 0)
         return "Block: Genesis block should be validated using isGenesisBlockInvalidReason";
```
ferencdg: considering that the GenesisBlock has prev_hash == Hash.init

```suggestion
    if (block.header.height == 0)
        return "Block: Genesis block should be validated using isGenesisBlockInvalidReason";

    assert(prev_hash !is Hash.init);
    assert(random_seed !is Hash.init);
```

we can also just delete the if

Geod24: Nice catch again, fixed

PR #1823, opened 2021-03-24
Implement FlashListenerAPI for Wallet integration

AndrejMitrovic:
This will allow integrating the Flash node with a Wallet or other front-end client.

### DIFF
--- a/source/agora/flash/Node.d
+++ b/source/agora/flash/Node.d
@@ -354,7 +354,7 @@ public abstract class FlashNode : ControlFlashAPI
         this.db = this.getManagedDatabase(db_path);
         this.load();
 
-        this.channels = Channel.loadChannels(this.db,
+        this.channels = Channel.loadChannels(this.conf, this.db,
             &this.getFlashClient,engine, taskman,
             &this.putTransaction, &this.paymentRouter,
             &this.onChannelOpen,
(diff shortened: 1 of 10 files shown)


source/agora/flash/Node.d
```diff
@@ -374,6 +405,12 @@ public abstract class FlashNode : ControlFlashAPI
 
         foreach (_, chan; this.channels)
             this.network.addChannel(chan.conf);
+
+        if (this.conf.listener_address.length != 0)
+            this.listener = this.getFlashListenerClient(
+                this.conf.listener_address, Duration.init);
+        else  // avoid null checks & segfaults
+            this.listener = new BlackHole!FlashListenerAPI();
```
AndrejMitrovic, line 413: Notes about this:

- Either we use the code in the PR
- Or we make `listener` optional and we always do `&& this.listener !is null` everywhere.
- Or we make `listener` a required option so it must not be null.

source/agora/flash/Node.d
```diff
@@ -834,12 +893,29 @@ public abstract class FlashNode : ControlFlashAPI
...
         }
 
         foreach (payment_hash; rev_htlcs)
+        {
             if (auto invoice = payment_hash in this.invoices)
-                this.payInvoice(*invoice); // Retry
+            {
+                ErrorCode error;
+                if (auto errors = payment_hash in this.payment_errors)
+                    error = (*errors)[0].err;  // pick first known reason
```
omerfirmak: This should  be the last known reason as we queue new failure reasons to the end.
But still, we may report wrong failure reason to the wallet. As error are propagated separately, there is a race between the error message and HTLC being revoked.

AndrejMitrovic: Yeah I had the same thoughts, not sure what is the best approach. I think I will change it to use the last error then.

AndrejMitrovic:
@omerfirmak let me know if you have any additional comments and I'll fix them up in another PR.

PR #1825, opened 2021-03-24
Clean up some SQL code and use string instead of BLOB

Geod24:
See each commit.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -61,7 +61,6 @@ import agora.crypto.ECC;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.crypto.Schnorr;
-import agora.serialization.Serializer;
 import agora.stats.Utils;
 import agora.stats.Validator;
 import agora.utils.Log;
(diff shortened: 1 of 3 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -161,7 +160,7 @@ public class EnrollmentManager
 
         // create the table for enrollment data for a node itself
         this.db.execute("CREATE TABLE IF NOT EXISTS node_enroll_data " ~
-            "(key CHAR(128) PRIMARY KEY, val BLOB NOT NULL)");
+            "(key CHAR(128) PRIMARY KEY, val CHAR(66) NOT NULL)");
```
TrustHenry:
```suggestion
            "(key CHAR(128) PRIMARY KEY, val TEXT(66) NOT NULL)");
```

Geod24: Actually, according to SQLite manual, it just ignore the length, so I'm going to make it `TEXT`

Geod24:
Updated, thanks for spotting this @TrustHenry !

TrustHenry (approved):
LGTM

Issue #1657, opened 2021-02-17
Review default values in configuration


Geod24:
Currently we have quite a few default values in our config files which are just "here", but are extremely suggestive, such as the ban rules (https://github.com/bpfkorea/agora/blob/4c7f32d9a84a8ea666bdf362f742a5a931ec058c/source/agora/common/BanManager.d#L37-L45), or the timeout (https://github.com/bpfkorea/agora/blob/4c7f32d9a84a8ea666bdf362f742a5a931ec058c/source/agora/common/Config.d#L152-L162). Those should be reviewed and corrected to match the expected CoinNet flow.

Geod24:
Four places to review:
- Infrastructure (TestNet);
- In code (`Config.d`);
- In the example file (`config.yaml`);
- In the system test;

linked0:
Should we review the `data_dir` value?

```
/// Path to the data directory to store metadata and blockchain data
public string data_dir = "/var/lib/agora/";
```

Geod24:
That's a good default for Linux, but not for Windows indeed. Neither for OSX.

Geod24:
Addressed

Geod24:
Where are we on this ?

PR #1828, opened 2021-03-26
Refactor BanManager.Config

linked0:
We moved `BanManager.Config` to `Config` as `BanConfig` because it's more convenient to manage configuration in one place. 

Part of #1657


ferencdg (approved):

linked0:
I found out that there are some reasons that the `BanManager` has its own `Config` type. So, I'm closing this PR.
@ferencdg Thanks for the review anyway.

PR #1830, opened 2021-03-26
Reduce the amount of module-level logger instances

### LINKED ISSUE #1443: Logger should not add module constructors to modules
there are constant module dependency problems like this:
```
Cyclic dependency between module constructors/destructors of agora.consensus.state.ValidatorSet and agora.consensus.validation.Block
agora.consensus.state.ValidatorSet* ->
agora.consensus.validation ->
agora.consensus.validation.Block* ->
agora.consensus.state.ValidatorSet*
```

Sometimes modules should not import each other(like in the above case), but there are genuine cases when modules would like to import each other AND the the modules also would like to use loggers.

Currently the logger implementation uses module constructors to instantiate the per module level logger. Andrei suggested changing this to lazy initialization and getting rid of module constructors.

Geod24:
This does some much-needed cleanup in the logger instances, and I hope that it will reduce the memory usage on the servers, as it seems a good chunk of the leaks come from the memory allocated by the Loggers. Normally having an allocation in module ctor wouldn't be an issue, but Vibe.d spawns and kills threads often, or so it seems.

### DIFF
--- a/source/agora/consensus/Fee.d
+++ b/source/agora/consensus/Fee.d
@@ -22,13 +22,10 @@ import agora.consensus.data.UTXO;
 import agora.consensus.data.Params;
 import agora.consensus.state.UTXOCache;
 import agora.crypto.Key;
-import agora.utils.Log;
 
 import std.math;
 import std.algorithm;
 import std.array;
-
-mixin AddLogger!();
 
 /// Delegate to check data payload
 public alias FeeChecker = string delegate (in Transaction tx,
@@ -358,7 +355,7 @@ public class FeeManager
     ***************************************************************************/
 
     public void accumulateFees (in Block block, UTXO[] stakes,
-        scope UTXOFinder peekUTXO) nothrow @trusted
+        scope UTXOFinder peekUTXO) @trusted
     {
         if (block.header.height % this.params.PayoutPeriod == 0)
             this.clearAccumulatedFees();
@@ -379,18 +376,9 @@ public class FeeManager
                 }
             );
 
-            try
-            {
-                auto new_fee = this.accumulated_fees[stake.output.address]
-                    .toString();
-                this.db.execute("REPLACE INTO accumulated_fees " ~
-                    "(fee, public_key) VALUES (?,?)", new_fee,
-                    stake.output.address.toString());
-            }
-            catch (Exception e)
-            {
-                log.error("ManagedDatabase operation error on accumulateFees");
-            }
+            this.db.execute(
+                "REPLACE INTO accumulated_fees (fee, public_key) VALUES (?,?)",
+                this.accumulated_fees[stake.output.address], stake.output.address);
         }
     }
 
@@ -413,14 +401,10 @@ public class FeeManager
     }
 
     /// Clears the accumulated fees
-    public void clearAccumulatedFees () nothrow @safe
+    public void clearAccumulatedFees () @safe
     {
         () @trusted {
-            try
-                this.db.execute("DELETE FROM accumulated_fees");
-            catch (Exception e)
-                log.error("ManagedDatabase operation error on clearAccumulatedFees");
-
+            this.db.execute("DELETE FROM accumulated_fees");
             this.accumulated_fees.clear();
         } ();
     }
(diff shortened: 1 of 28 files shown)


source/agora/consensus/SlashPolicy.d
```diff
@@ -160,10 +152,7 @@ public class SlashPolicy
     {
         Hash[] keys;
         if (!this.enroll_man.getEnrolledUTXOs(keys) || keys.length == 0)
-        {
-            log.fatal("Could not retrieve enrollments / no enrollments found");
-            assert(0);
-        }
+            assert(0, "Could not retrieve enrollments / no enrollments found");
```
hewison-chris, line 155: Perhaps a const string could be used for this as it is used several times.

hewison-chris (approved):
LGTM, one small change maybe

source/agora/consensus/SlashPolicy.d
Geod24: For assert, it's quite nice to be able to just grep the message and find the line, so if it triggers, I'd recommend deduplicating it instead.

hewison-chris: Sure, saves an extra grep.

source/agora/consensus/Fee.d
```diff
@@ -379,18 +376,10 @@ public class FeeManager
...
-                    .toString();
-                this.db.execute("REPLACE INTO accumulated_fees " ~
-                    "(fee, public_key) VALUES (?,?)", new_fee,
-                    stake.output.address.toString());
-            }
-            catch (Exception e)
-            {
-                log.error("ManagedDatabase operation error on accumulateFees");
-            }
+            auto new_fee =
```
linked0: It should be removed.

ferencdg:
this will also help us with circular dependency problems introduced by the logger, so I guess we can close this one too: https://github.com/bosagora/agora/issues/1443

linked0 (approved):
LGTM, one small change.

ferencdg:
> This does some much-needed cleanup in the logger instances, and I hope that it will reduce the memory usage on the servers, as it seems a good chunk of the leaks come from the memory allocated by the Loggers. Normally having an allocation in module ctor wouldn't be an issue, but Vibe.d spawns and kills threads often, or so it seems.

I wonder why those allocations from the module constructor are not freed, does vibe.d secretly holds onto the logger after the thread is terminated/reused? I remember creating a simple test app that allocated in the module constructor(without vibe.d and without ocean's Logger), and I saw no memory increase when new threads came and go away.

Geod24:
> I wonder why those allocations from the module constructor are not freed, does vibe.d secretly holds onto the logger after the thread is terminated/reused?

I am not even sure those are not freed. I actually expect them to be, but the very fact that those happen in the first place is bad. Since we have an OOP architecture, with classes that hold state, it's a no-brainer to have it moved to the class, as shown by the diff. You might notice I didn't move *everything*, like logger only appearing in unittests, and the one in `network/Client.d`.

One thing on TLS: D users tend to use TLS liberally, and I think that's a mistake. TLS comes at a cost, as each spawned thread  will require the space to be reserved, even if it never uses this portion of the code. Even worse, since TLS is handled natvely (e.g. by pthread), spawning a thread in C or C++ that never come in contact with D code will use more memory because it will reserve space for all the globals in Phobos / Vibe.d / whatever other D code.

source/agora/consensus/Fee.d
Geod24: Oopsie

Geod24:
Fixed

omerfirmak:
@Geod24 @ferencdg 

As far as I can see, D garbage collector does not leak. So that probably includes TLS and any module ctors.

ferencdg:
> @Geod24 @ferencdg
> 
> As far as I can see, D garbage collector does not leak. So that probably includes TLS and any module ctors.

I think the reason why I originally thought it is leaking is because when I profiled with tracy, then at that time our tracy integration in Agora was broken and it showed up as leaking. Maybe after the integration is fixed, it shows up as excessive, but reclaimable memory usage.

PR #1831, opened 2021-03-26
CI: Add retry hack to version_build

Geod24:
```
We get an odd 'no such file or directly' error on the CI sometimes,
which seems to come from a race condition in dub.
The retry hack used for the C++ build script was broken because it
used `sleep || ...`, which should have been `sleep && ...` (sleep doesn't fail).
So fix it, and apply the same hack to `version_dub`.
```

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -11,8 +11,8 @@
     "stringImportPaths" : [ "build" ],
     "excludedSourceFiles": [ "source/scpp/*.d", "source/scpd/quorum/QuorumIntersectionTests.d" ],
     "preGenerateCommands": [
-        "$DUB source/agora/cli/version/main.d",
-        "$DUB source/scpp/build.d || sleep 5s || $DUB source/scpp/build.d"
+        "$DUB source/agora/cli/version/main.d || (sleep 5s && $DUB source/agora/cli/version/main.d)",
+        "$DUB source/scpp/build.d || (sleep 5s && $DUB source/scpp/build.d)"
     ],
     "sourceFiles-posix": [
         "source/scpp/build/BallotProtocol.o",


linked0 (approved):
LGTM but please correct the commit message, "so uuse it for the version ..."

kchulj (approved):
The `macOS` CI complains :)

Geod24:
```
agora ~master: building configuration "server"...
source/agora/node/FullNode.d(237,30): Error: file `"VERSION"` cannot be found or not in a path specified with `-J`
source/agora/node/FullNode.d(237,30):        Path(s) searched (as provided by `-J`):
source/agora/node/FullNode.d(237,30):        [0]: `build`
Type 'FlashValidatorAPI' implements more than one interface: make sure the one describing the REST server is the first one
Type 'FlashFullNodeAPI' implements more than one interface: make sure the one describing the REST server is the first one
source/agora/node/main.d(60,34): Error: file `"VERSION"` cannot be found or not in a path specified with `-J`
source/agora/node/main.d(60,34):        Path(s) searched (as provided by `-J`):
source/agora/node/main.d(60,34):        [0]: `build`
ldc2 failed with exit code 1.
```

WOW that's weird, it looks like the command was successful but the file wasn't generated ??

dub.json
```diff
@@ -11,7 +11,7 @@
     "stringImportPaths" : [ "build" ],
     "excludedSourceFiles": [ "source/scpp/*.d", "source/scpd/quorum/QuorumIntersectionTests.d" ],
     "preGenerateCommands": [
-        "$DUB source/agora/cli/version/main.d",
+        "$DUB source/agora/cli/version/main.d || sleep 5s || $DUB source/agora/cli/version/main.d",
```
ferencdg: why do we have || instead of &&? 
I think whatever prints out 'Failed to execute' will return with an error code of non-zero(might be something trivial I don't get)

Shouldn't we see 2 'Failed to execute', if it is retried?

Geod24: Indeed, good catch!

Geod24:
Fixed

PR #1834, opened 2021-03-26
Account for sentinels in calls to Tracy

omerfirmak:
Some calls to tracy didnt account for the sentinel values, which can cause freed memory to be reported as leaked.
Aligned tracy calls with `log_free`/`log_malloc` calls.


Geod24 (approved):
Looks reasonable from a mile away

PR #1835, opened 2021-03-26
UnitTest runner: Stop early in single threaded mode

Geod24:
Currently, we have some flakey test in the CI, so we have a retry strategy in place.
Namely, we try to run the unittest binary twice.
However, our retry strategy is defeated by our module runner,
which will run all the tests regardless. This isn't very helpful,
especially considering the amount of output a single test failure produces.
So if a single threaded mode fails, consider it important enough to warrant
forcing the developer forcing it before continuing.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -172,6 +172,8 @@ private UnitTestResult customModuleUnitTester ()
         to!bool(environment["dnologs"]) : false;
     const all_single_threaded = ("dsinglethreaded" in environment) ?
         to!bool(environment["dsinglethreaded"]) : false;
+    const should_fail_early = ("dfailearly" in environment) ?
+        to!bool(environment["dfailearly"]) : true;
     auto filter = environment.get("dtest").toLower();
     size_t filtered;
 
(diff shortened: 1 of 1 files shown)


source/agora/test/Base.d
```diff
@@ -248,11 +249,19 @@ private UnitTestResult customModuleUnitTester ()
             CircularAppender!()().print(output);
             stdout.flush();
         }
+        return false;
     }
 
     // Run single-threaded tests
+    bool failed_single_threaded;
     foreach (mod; single_threaded)
```
ferencdg, line 259: I wonder if 'run in single threaded' and 'fail at first error' should be 2 separate aspects/env variables. The reason is that for local development I would still like to 'run single threaded', but not 'fail at first error'. If 4 tests fails, then I would just rerun those 4 one-by-one. (Sometimes tests fails even in single threaded on my machine, even if there are no bugs in the code, so I just generally rerun them one-by-one)

Geod24: How does it look now ?

source/agora/test/Base.d
```diff
@@ -172,6 +172,8 @@ private UnitTestResult customModuleUnitTester ()
         to!bool(environment["dnologs"]) : false;
     const all_single_threaded = ("dsinglethreaded" in environment) ?
         to!bool(environment["dsinglethreaded"]) : false;
+    const should_fail_early = ("dfailearly" in environment) ?
+        to!bool(environment["dfailearly"]) : true;
```
ferencdg, line 176: if this is the default, we need to change CI, right?

ferencdg (approved):

PR #1837, opened 2021-03-28
Ledger test: Change a test to use a cycle of 20

Geod24:
One of the main issue with the Ledger tests at the moment
is that it bypasses some checks, leading to tests using
invalid blocks. One of the many reasons blocks are invalid
is that there's a different cycle length being used
(ConsensusParams and Genesis are not consistent).
This will be fixed in a later PR, but this specific test
can be easily fixed now.


kchulj (approved):
LGTM

PR #1840, opened 2021-03-28
Improve logging when externalizing a block

Geod24:
- Remove a logging statement from the validation function, which is supposed to be idempotent;
- Add a summary of the block being externalized inside of addValidatedBlock;
- List the three metrics we're interested in: Active / Signing / Slashed;

Note that 'Active' includes 'Slashed'. Currently, a bug of the validation function
is that it allows a block that slashes all validators, which is obviously invalid,
as no one would (should) sign it, and there can be no progress after that.
This will be addressed in a future PR.


Geod24:
Screen Shot 2021-03-29 at 2 18 32
Screen Shot 2021-03-29 at 2 18 59

I'll improve the logging of BitFields in a later PR.

PR #1843, opened 2021-03-28
Logs: Remove noisy log about a valid signature

Geod24:
This log doesn't add any information to the developer anymore,
as the envelope is already logged in trace mode,
and the developer would expect that the signature is valid.
If that is not the case, the INVALID message will still trigger.


kchulj (approved):
✅

PR #1844, opened 2021-03-28
Reduce duplication and reduce special cases in Ledger test

Geod24:
The Ledger tests suffer from many issues, which originate from the organic growth of its tests.
As rules were added, not all tests were properly adapted, leading to some place using invalid blocks in tests.
This does some preliminary work by reducing the number of functions used to generate blocks, and instead of generating blocks ad-hoc, make it go through the ledger as it is the class with all the contextual information.


hewison-chris (approved):
Thanks these changes will help make it easier.

PR #1846, opened 2021-03-29
Minor improvements to Ledger unittests

Geod24:
All to make it easier to switch to a default of 20 for ValidatorCycle.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -1655,7 +1655,7 @@ unittest
     import agora.consensus.data.genesis.Test;
     import agora.consensus.PreImage;
 
-    auto params = new immutable(ConsensusParams)(10);
+    auto params = new immutable(ConsensusParams)(20);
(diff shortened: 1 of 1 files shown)


source/agora/node/Ledger.d
```diff
@@ -1572,24 +1572,23 @@ unittest
     import agora.consensus.data.genesis.Coinnet : CoinGenesis = GenesisBlock;
```
linked0, line 1572: Can't we remove `CoinGenesis`, using `GenesisBlock` instead?

linked0 (approved):
LGTM, and one question.

hewison-chris (approved):
LGTM

source/agora/node/Ledger.d
Geod24: No because it is still used (we need two Genesis block to have a mismatch).

kchulj (approved):
LGTM

PR #1856, opened 2021-03-31
Bech32: Minor style fixes + remove usage of logger

Geod24:
Pretty simple stuff, see each commit.

### DIFF
--- a/source/agora/crypto/Bech32.d
+++ b/source/agora/crypto/Bech32.d
@@ -36,12 +36,10 @@
 module agora.crypto.Bech32;
 
 import agora.crypto.ECC;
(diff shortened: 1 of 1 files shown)


linked0 (approved):

kchulj:
TIL the order of `@safe pure nothrow @nogc`.

Geod24:
We don't really have an order, besides having type qualifiers first followed by safety (and simply because it's so omniscient).
It's not something I would flag in code review either, as I don't think it's productive (we would need something like dfmt for it to be something worth even talking about).

kchulj (approved):

PR #1858, opened 2021-03-31
PreImageInfo: Rename 'enroll_key' to 'utxo' and other simplifications

Geod24:
The first commit renames 'enroll_key' to UTXO and is up for discussion.
Personally I find it more descriptive and less confusing.
The other commits are simplification of existing code (or removal) and should be fairly straightforward.

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -646,14 +646,20 @@ public class EnrollmentManager
     public bool getNextPreimage (out PreImageInfo preimage, in Height height)
         @safe
     {
-        uint next_dist = getRevealDistance(height);
-        if (next_dist >= this.params.ValidatorCycle)
+        const enrolled = this.validator_set.getEnrolledHeight(this.enroll_key);
+        if (enrolled == ulong.max)
+            return false;
+
+        assert(height >= enrolled);
+        const next_reveal = min(height + PreimageRevealPeriod,
+                                enrolled + (this.params.ValidatorCycle - 1));
+
+        if (next_reveal <= height)
             return false;
 
         preimage.utxo = this.enroll_key;
-        preimage.distance = cast(ushort)next_dist;
-        const enrolled = this.validator_set.getEnrolledHeight(this.enroll_key);
-        preimage.hash = this.cycle[enrolled + next_dist];
+        preimage.distance = cast(ushort)(next_reveal - enrolled);
+        preimage.hash = this.cycle[next_reveal];
         return true;
     }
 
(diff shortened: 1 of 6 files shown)


Geod24:
Ping

kchulj:
Is the name `utxo` in order to avoid conflict with `Enrollment`'s `utxo_key`?

hewison-chris (approved):
LGTM

source/agora/consensus/EnrollmentManager.d
```diff
@@ -966,26 +965,26 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Set the key for the enrollment for this node
+        Set the UTXO for the enrollment for this node
 
         If saving an enrollment key fails, this displays a log messages and just
```
linked0: The doc should be modified.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -722,19 +722,18 @@ public class EnrollmentManager
...
         Returns:
             the PreImageInfo of the enrolled key if it exists,
             otherwise PreImageInfo.init
 
     ***************************************************************************/
 
-    public PreImageInfo getValidatorPreimage (in Hash enroll_key)
-        @trusted nothrow
+    public PreImageInfo getValidatorPreimage (in Hash utxo)
+        @safe nothrow
```
linked0, line 720: Can you explain why it changed?

Geod24:
> Is the name `utxo` in order to avoid conflict with `Enrollment`'s `utxo_key`?

Nope. Usually I try to avoid repeating the type name in the variable name, unless it's trivial and generic (`Hash hash` or `UTXO utxo`).

`Hash utxo_key` doesn't conform to that criteria, because why wouldn't one just use `utxo` ? The `_key` part is a reference to the fact that this is not the actual UTXO struct, but the hash that references it. In practice, using hash to reference things is so common in blockchain that everyone in the field understands this concept.

In this case, if you see `enroll_key`, and it's of type `Hash`, and it's in the `Enrollment` struct, you just know it's an index of some sort, but you don't really know more than that. `utxo` at least tells you what kind of index it is. We could be a bit more verbose and use for example `collateral`, but I just went for the thing that was shorter to type.

linked0 (approved):
LGTM

source/agora/consensus/EnrollmentManager.d
Geod24: `@trusted` presents an interface as `@safe` without doing the checks.
If the function is a trivial wrapper for another function, it's likely that the `@trusted` should be on that other function.

For example, the following is fine:
```D
T* trustedMalloc (T) () @trusted
{
    void* ret = malloc(T.sizeof);
    initializeMemory(ret);
    return cast(T*) ret;
}
```
Because it wraps a `@system` function in a safe manner.
However, this is not:
```D
void* trustedMalloc (size_t size) @trusted { return malloc(size); }
```
Because if you can write this, it means `malloc` should be safe to call directly too (after all `trustedMalloc` only applies attributes, nothing else), which it clearly is not.

kchulj (approved):
Ty for explanation, LGTM.

source/agora/consensus/EnrollmentManager.d
linked0: Ty. That's what I really want because it's always kind of confusing. :)

source/agora/consensus/EnrollmentManager.d
Geod24: Done

source/agora/consensus/EnrollmentManager.d
```diff
@@ -966,26 +951,26 @@ public class EnrollmentManager
 
     /***************************************************************************
 
-        Set the key for the enrollment for this node
+        Set the UTXO for the enrollment for this node
 
-        If saving an enrollment key fails, this displays a log messages and just
+        If saving to the database fails, this displays a log messages and just
         returns. The `enroll_key` stores the value anyway and it can be restored
```
linked0, line 957: Ah, There's one more to be fixed.

Geod24: No, this one references the field in `EnrollmentManager` which has the same name.

PR #1861, opened 2021-03-31
Flash: Payment path error reporting & timeouts

AndrejMitrovic:


doc/config.example.yaml
```diff
@@ -75,6 +75,43 @@ validator:
...
+################################################################################
+flash:
+  # Whether or not the Flash protocol should be supported
+  enabled: true
+  # Timeout for each request in milliseconds
+  timeout: 10000
+  # This is a randomly generated keypair
+  # DO NOT USE THOSE VALUES ANYWHERE
+  # Private seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  # Public address:  GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
```
Geod24, line 91: Outdated

AndrejMitrovic: Then it means the rest of the documented config is also outdated.

source/agora/flash/ErrorCode.d
```diff
@@ -19,6 +19,9 @@ public enum ErrorCode : ushort
     /// No error.
     None = 0,
 
+    /// Couldn't find a path to another node, or it was too long
+    PathNotFound,
+
```
Geod24: Perhaps you want to keep `Unknown` next to `None` ?

doc/config.example.yaml
Geod24: True, let's just fix it all at once

Geod24 (approved):
LGTM, merge at will. One thing for the future is that we should probably be smarter about the config, since flash and validator are quite overlapping.

omerfirmak (approved):

PR #1865, opened 2021-04-01
vibe-core: v1.14.0 to v1.16.0

Geod24:
This includes an update to the logging system which will make our production
trace logs much less verbose.


kchulj (approved):
There is an error in `Change permission on lst files` of integration but I don't think it is related.

Geod24:
I think it is. I'll track down the underlying bug and will report it.

AndrejMitrovic:
> sudo: unable to execute /usr/bin/chmod: Argument list too long

Super weird error. Should we just retry?

Geod24:
I added a workaround for the issue. It does add 30s to the runtime of the CI, but it works. And opened #1868.

PR #1866, opened 2021-04-01
Simplify database handling; add documentation in FullNode; more Point -> PublicKey

Geod24:
See each commit, and let me know if this needs to be split.


AndrejMitrovic (approved):
LGTM!

PR #1869, opened 2021-04-01
Correct values in configuration on banning and retry

### LINKED ISSUE #1657: Review default values in configuration

linked0:
We should wait some time to ban any unanswering node after many retries and failed requests. We now set the waiting time to about 1 hour. Actually, the total time of multiplication of `retry_delay`, `max_retries`, `timeout`, and `max_failed_requests` in the commits could be one and a half hours.

We mostly refer to the values of `infrastructure` because it is used for the live test system. Anyway, one hour could be the best time to wait before banning a bad node.

Fixes #1657

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -26,11 +26,11 @@ node:
   # Port on which we bind
   port:    2826    # 0xB0A, default value
   # Number of milliseconds to wait between retrying requests
-  retry_delay: 5000
+  retry_delay: 3000
   # Maximum number of retries to issue before a request is considered failed
-  max_retries: 5
+  max_retries: 10
   # Timeout for each request in milliseconds
-  timeout: 2500
+  timeout: 5000
   # Path to the data directory (if the path doesn't exist it will be created)
   data_dir: .cache
   # The local address where the stats server (currently Prometheus)
@@ -117,7 +117,7 @@ flash:
 ################################################################################
 banman:
   # max failed requests until an address is banned
-  max_failed_requests: 1000
+  max_failed_requests: 100
   # the default duration of a ban
   ban_duration: 86400
 
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -160,10 +160,10 @@ public struct NodeConfig
     public Duration retry_delay = 3.seconds;
 
     /// Maximum number of retries to issue before a request is considered failed
-    public size_t max_retries = 5;
+    public size_t max_retries = 10;
 
     /// Timeout for each request
-    public Duration timeout = 500.msecs;
+    public Duration timeout = 5000.msecs;
 
     /// Path to the data directory to store metadata and blockchain data
     public string data_dir = "/var/lib/agora/";
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -166,7 +166,7 @@ public struct NodeConfig
     public Duration timeout = 5000.msecs;
 
     /// Path to the data directory to store metadata and blockchain data
-    public string data_dir = "/var/lib/agora/";
+    public string data_dir = ".cache";
 
     /// The duration between requests for retrieving the latest blocks
     /// from all other nodes
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -1752,13 +1752,13 @@ public struct TestConf
     bool configure_network = true;
 
     /// the delay between request retries
-    Duration retry_delay = 100.msecs;
+    Duration retry_delay = 500.msecs;
 
     /// minimum clients to connect to (defaults to nodes.length - 1)
     size_t min_listeners;
 
     /// max retries before a request is considered failed
-    size_t max_retries = 20;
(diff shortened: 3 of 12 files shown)


doc/config.example.yaml
```diff
@@ -26,11 +26,11 @@ node:
   # Port on which we bind
   port:    2826    # 0xB0A, default value
   # Number of milliseconds to wait between retrying requests
-  retry_delay: 5000
+  retry_delay: 500
```
Geod24: I think 5 seconds was actually better.

source/agora/common/Config.d
```diff
@@ -157,13 +157,13 @@ public struct NodeConfig
...
-    public Duration retry_delay = 3.seconds;
+    public Duration retry_delay = 500.msecs;
 
     /// Maximum number of retries to issue before a request is considered failed
-    public size_t max_retries = 5;
+    public size_t max_retries = 10;
 
     /// Timeout for each request
-    public Duration timeout = 500.msecs;
+    public Duration timeout = 5000.msecs;
```
Geod24: Those should indeed be consistent with the example doc, see my comments there

source/agora/test/Base.d
```diff
@@ -1752,13 +1752,13 @@ public struct TestConf
     bool configure_network = true;
 
     /// the delay between request retries
-    Duration retry_delay = 100.msecs;
+    Duration retry_delay = 500.msecs;
```
Geod24, line 1755: Those are slightly different though. We know that the nodes are in memory so it's going to be insanely faster.
However, increasing to 500 msecs (and reducing retries) is actually a good idea, so 👍

source/agora/test/NetworkDiscovery.d
```diff
@@ -20,7 +20,10 @@ import agora.test.Base;
 ///
 unittest
 {
-    TestConf conf = TestConf.init;
+    TestConf conf =
+    {
+        retry_delay : 100.msecs,
+    };
```
Geod24, line 26: Why ?

doc/config.example.yaml
```diff
@@ -117,7 +117,7 @@ flash:
 ################################################################################
 banman:
   # max failed requests until an address is banned
-  max_failed_requests: 1000
+  max_failed_requests: 100
```
Geod24, line 120: On the fence on this, I think both are valid, since it's client requests.

Geod24 (approved):
Except for the default `retry_delay`, all LGTM

source/agora/test/NetworkDiscovery.d
linked0: The first network test in this module would fail after the default `retry_delay` is set to `500.msecs` because the test checks the `NodeInfo` of all the nodes immediately.

doc/config.example.yaml
linked0: Will update.

source/agora/common/Config.d
linked0: I changed the value to `5000.msecs` in all the related files. Am I missing something?

kchulj: So I assume we want 5 seconds for both `retry_delay` and `timeout`.

linked0: OK, I got it. I'm applying it now. Ty.

doc/config.example.yaml
linked0: @Geod24 I'd like to confirm that the value(`5000` msecs) you suggested is right before applying it to the code. I think it's kind of long. Do you mean we should have 5 seconds for `retry_delay` and 5 seconds for `timeout`?

linked0: @Geod24 I'd like to confirm that the value(5000 msecs) you suggested is right before applying it to the code. I think it's kind of long. Do you mean we should have 5 seconds for `retry_delay` and 5 seconds for `timeout`?

Geod24: Yes

kchulj (approved):
✅

kchulj:
`Some index files failed to download. They have been ignored, or old ones used instead.` ?
There has been various `Integration` failures recently..

Geod24:
What's up with the changes to flash in the 3rd commit ?

linked0:
I will check the last commit. There might have been some problems in committing.

Geod24 (approved):

linked0:
The 5 seconds for the `retry_delay` makes the integration test fail constantly. So I set the value to 3 seconds, which makes the test succeed. I will check the reason why it affects the test.

PR #1870, opened 2021-04-01
Move signature validation to ledger & clean it up

Geod24:
See each commit for details. I have a feeling this won't pass the CI though.

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -772,60 +772,51 @@ public class Ledger
         const Scalar challenge = hashFull(block);
         const enrolled_validators = this.enroll_man.getCountOfValidators(block.header.height);
 
-        log.trace("Validators expected to sign this block is {}", enrolled_validators);
         // Check that more than half have signed
         auto signed = iota(0, enrolled_validators).filter!(i => block.header.validators[i]).count();
         if (signed <= enrolled_validators / 2)
         {
-            log.error("Only {} signed. Require more than {} out of {} validators to sign for externalizing slot height {}.",
-                      signed, enrolled_validators / 2, enrolled_validators, block.header.height);
-            return "Only " ~ to!string(signed) ~ " signed the block. Require more than " ~ to!string(enrolled_validators);
+            log.error("Block#{}: Signatures are not majority: {}/{}, signers: {}",
+                      block.header.height, signed, enrolled_validators,
+                      block.header.validators);
+            return "The majority of validators hasn't signed this block";
         }
+
+        log.trace("Checking signature, participants: {}/{}", signed, enrolled_validators);
         foreach (idx; 0 .. enrolled_validators)
         {
             const K = this.enroll_man.getValidatorAtIndex(block.header.height, idx);
-            log.trace("idx {} PublicKey: {}", idx, K);
-            if (K == PublicKey.init)
-            {
-                log.error("[{}:{}] idx #{}: Validator key not found for height {}",
-                          file, line, idx, block.header.height);
-                return "Block: Couldn't find a Validator for the given index "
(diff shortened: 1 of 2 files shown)


Geod24:
> I have a feeling this won't pass the CI though.

I should have more faith, this is G2G.

source/agora/node/Ledger.d
```diff
@@ -772,60 +772,52 @@ public class Ledger
...
 
-        log.trace("Validators expected to sign this block is {}", enrolled_validators);
         // Check that more than half have signed
         auto signed = iota(0, enrolled_validators).filter!(i => block.header.validators[i]).count();
         if (signed <= enrolled_validators / 2)
         {
-            log.error("Only {} signed. Require more than {} out of {} validators to sign for externalizing slot height {}.",
-                      signed, enrolled_validators / 2, enrolled_validators, block.header.height);
-            return "Only " ~ to!string(signed) ~ " signed the block. Require more than " ~ to!string(enrolled_validators);
+            // Can't use '51%' here because 50.000001% is enough
```
hewison-chris: I don't see the point of this comment.

source/agora/node/Ledger.d
```diff
@@ -772,60 +772,52 @@ public class Ledger
...
+            log.error("Block#{}: Signatures are not majority: {}/{} ({})",
```
hewison-chris: A little cryptic but I guess the intention is to keep it more compact?

source/agora/node/Ledger.d
```diff
@@ -772,60 +772,52 @@ public class Ledger
...
+                      block.header.height, signed, enrolled_validators,
+                      block.header.validators);
+            return "The majority of validators hasn't signed this block";
         }
+
+        log.trace("Checking signature, participants: {}/{}", enrolled_validators);
```
hewison-chris: Missing parameter.

hewison-chris (changes requested):
Some much needed tidy up of the signature logging. Although there will be some rebase conflicts as I also made a couple of changes to logging in the other PR which is also green.
Also a couple of small changes requested.

source/agora/node/Ledger.d
Geod24: Is it cryptic ? It'll show something like:
```
Block#4: Signatures are not majority: 2/6 ([12])
```
The `[12]` is very cryptic tho and will need to be fixed, but I thought "majority" and "2/6" will suffice. Suggestions welcome.

source/agora/node/Ledger.d
Geod24: True. Originally I wrote something like "Less than 51% signatures", but that could have been misleading.

Geod24: We could also just go with the `51%` anyway given how much of a corner case it is. "Block does not have absolute majority of signature" also comes to mind.

source/agora/node/Ledger.d
hewison-chris: Maybe if we add `Block#4: Signatures are not majority: 2/6, validators signed=[252]`.
We could also show the bits of the mask but then if we have hundreds of validators it may get messy.

source/agora/node/Ledger.d
hewison-chris: I think it is ok as is but just without the comment.

source/agora/node/Ledger.d
Geod24: Fixed

source/agora/node/Ledger.d
Geod24: How does it look now ?

source/agora/node/Ledger.d
Geod24: Removed this line

hewison-chris (approved):
LGTM

PR #1871, opened 2021-04-01
SCP: Do not print outdated envelopes

Geod24:
Moving the log statement after a few checks make the logs more readable.
We also remove the newline in the log message as flushing stdout conflict
with some tools that expect each line to be its own log statement.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -461,8 +461,6 @@ extern(D):
 
     public void receiveEnvelope (scope ref const(SCPEnvelope) envelope) @trusted
     {
-        log.trace("Receiving envelope: {}", scpPrettify(&envelope));
-
         // ignore messages if `startNominatingTimer` was never called or
         // if `stopNominatingTimer` was called
         if (this.nomination_timer is null)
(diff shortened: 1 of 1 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -475,22 +473,22 @@ extern(D):
...
             return;
         }
-        if (!verify(V, envelope.signature.toSignature(), challenge))
+        if (!verify(public_key, envelope.signature.toSignature(), challenge))
         {
-            log.trace("INVALID Envelope signature {} \nfor public key {} \n" ~
-                "envelope {}\nchallenge {}",
-                envelope.signature, public_key, scpPrettify(&envelope),
-                challenge.toString(PrintMode.Clear));
+            // If it fails signature verification, it might not originate from said key
```
hewison-chris, line 486: This signature is for a single validator so I do not understand the comment.

hewison-chris (changes requested):
LGTM just one thing.

source/agora/consensus/protocol/Nominator.d
Geod24: Anyone can produce an envelope claiming to be `X` with an invalid signature. Our message should not *attribute* the error to this node because it hasn't proved its identity.

hewison-chris: ok now I see what you meant. I was thinking it was related to multi sig.

hewison-chris (approved):

PR #1872, opened 2021-04-01
Runner: Flatten condition w.r.t. node

Geod24:
This should be more readable, at the expense of checking one boolean twice.


linked0 (approved):
Looks good!

PR #1873, opened 2021-04-01
Remove register[As]Listener

Geod24:
This was only used in unittest and not in the actual, deployed node.
Now that we have the registry, it should not be necessary.


AndrejMitrovic (approved):

Geod24:
Looks like pre-image gossipping is broken without `registerListener`

Geod24:
Hopefully #2244 will help.

Geod24:
#2511 did it

Issue #864, opened 2020-05-21
Test-suite sometimes fails with an odd error


AndrejMitrovic:
```
Run ./ci/run.sh
+ set -o pipefail
+ dub test -b unittest-cov --skip-registry=all --compiler=ldc2
Executable configuration "unittest" of package agora defines no main source file, this may cause certain build modes to fail. Add an explicit "mainSourceFile" to the package description to fix this.
Running custom 'unittest' configuration.
Running pre-generate commands for vibe-d:tls...
Running pre-generate commands for agora...
Failed to execute program (No such file or directory)
Command failed with exit code 2: /Users/runner/hostedtoolcache/dc/ldc2-4eaa2fd8/x64/ldc2-4eaa2fd8-osx-x86_64/bin/dub source/scpp/build.d
```

This is a strange one, it seems like there's a race condition somewhere where we try to run an executable which doesn't exist yet. I see it happen more and more often.

AndrejMitrovic:
https://github.com/bpfkorea/agora/pull/845/checks?check_run_id=695047246

TrustHenry:
If that's the case, If you print out the command of build.d, it will become clearer.
https://github.com/bpfkorea/agora/blob/c3eef128192bff125a868c7d6be7c16d97bdd5b1/source/scpp/build.d#L185

Geod24:
There's only three possibilities I can think of:
- `cpp` doesn't exists;
- `dub` doesn't exists;
- `source/scpp/build.d` doesn't exists;

Now one of the thing you would see, if the command failed, was [this](https://github.com/bpfkorea/agora/blob/c3eef128192bff125a868c7d6be7c16d97bdd5b1/source/scpp/build.d#L153-L154). Since we don't see it, uncommenting `writeln` won't do anything because the failure is before that.

Geod24:
I am starting to suspect cache pollution in the Github action.

AndrejMitrovic:
So the issue is multiple jobs running in parallel?

Geod24:
I will test this theory shortly

Geod24:
Just got this:
```
+ dub build --skip-registry=all --compiler=ldc2
Running pre-generate commands for vibe-d:tls...
Running pre-generate commands for agora...
[...]

PR #1874, opened 2021-04-01
[CI] Enable verbose dub output for preGenerateCommands

omerfirmak:
Will try to recreate #864

### DIFF
--- a/dub.json
+++ b/dub.json
@@ -11,8 +11,8 @@
     "stringImportPaths" : [ "build" ],
     "excludedSourceFiles": [ "source/scpp/*.d", "source/scpd/quorum/QuorumIntersectionTests.d" ],
     "preGenerateCommands": [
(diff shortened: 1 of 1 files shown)


Geod24:
You could probably modify the Github workflow to just run `dub build`. Even do it in loops until it fails.

omerfirmak:
I couldn't reproduce the issue. @Geod24, Should we merge this and wait for the next time we see it in the wild?

Geod24:
I think that it's a race condition, and so using verbose (and writing to stdout) is enough to delay things enough for it not to happen. Can you reproduce without verbose output ?

omerfirmak:
> Can you reproduce without verbose output ?

Sadly, no.

Geod24 (approved):
Can't hurt

PR #1881, opened 2021-04-05
Added double spent transaction replacement

### LINKED ISSUE #245: Define and implement rules for transaction replacement
Transaction replacement is what happens when a new transaction is received that compete with  another transaction in the MemPool.
In this case, some design could allow the transaction with higher fees to win.
We need to come up with our rule on this. Initially we will allow a replacement fee to occur if the fee is 20% higher than the previous one, rounded up to the next digit.

ferencdg:
implementation notes:
https://github.com/bosagora/agora/issues/245#issuecomment-809829999

fixes: #245

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -251,10 +251,10 @@ public class FullNode : API
         this.metadata = this.makeMetadata();
         this.network = this.makeNetworkManager(this.metadata, this.taskman, this.clock);
         this.storage = this.makeBlockStorage();
-        this.pool = this.makeTransactionPool();
-        this.utxo_set = this.makeUTXOSet();
-        this.enroll_man = this.makeEnrollmentManager();
         this.fee_man = this.makeFeeManager();
+        this.utxo_set = this.makeUTXOSet();
+        this.pool = this.makeTransactionPool();
+        this.enroll_man = this.makeEnrollmentManager();
         this.transaction_relayer = this.makeTransactionRelayer();
         const ulong StackMaxTotalSize = 16_384;
         const ulong StackMaxItemSize = 512;
@@ -716,7 +716,32 @@ public class FullNode : API
 
     protected TransactionPool makeTransactionPool ()
     {
-        return new TransactionPool(this.cacheDB);
+        return new TransactionPool(this.cacheDB, &getDoubleSpentSelector);
+    }
+
+    /***************************************************************************
+
+        Returns an instance of a DoubleSpentSelector
+
+        Subclasses can override this method and return
+        a DoubleSpentSelector
+
+        Returns:
+            an instance of DoubleSpentSelector
+
+    ***************************************************************************/
+
+    protected size_t getDoubleSpentSelector (Transaction[] txs) nothrow @safe
+    {
+        return maxIndex!((a, b)
+        {
+            Amount fee_a;
+            Amount fee_b;
+            fee_man.getAdjustedTXFee(a, &this.utxo_set.peekUTXO, fee_a);
+            fee_man.getAdjustedTXFee(b, &this.utxo_set.peekUTXO, fee_b);
+            return fee_a < fee_b;
+        }
+        )(txs);
     }
 
     /***************************************************************************
@@ -853,7 +878,8 @@ public class FullNode : API
         this.endpoint_request_stats
             .increaseMetricBy!"agora_endpoint_calls_total"(1, "transaction", "http");
         auto tx_hash = hashFull(tx);
-        if (this.pool.hasTransactionHash(tx_hash))
+        if (this.pool.hasTransactionHash(tx_hash) ||
+            !ledger.isAcceptableDoubleSpent(tx, config.node.double_spent_threshold_pct))
             return;
 
         this.tx_stats.increaseMetricBy!"agora_transactions_received_total"(1);
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -50,10 +50,12 @@ import agora.utils.Log;
 import agora.utils.PrettyPrinter;
 
 import std.algorithm;
+import std.algorithm.searching : maxElement;
 import std.conv : to;
 import std.exception;
 import std.format;
 import std.range;
+import std.typecons : Nullable, nullable;
 
 import core.time : Duration, seconds;
 
@@ -502,6 +504,48 @@ public class Ledger
 
     /***************************************************************************
 
+        Checks whether the `tx` is an acceptable double spend transaction.
+
+        If `tx` is not a double spend transaction, then returns true.
+        If `tx` is a double spend transaction, and its fee is considerable higher
+        than the existing double spend transactions, then returns true.
+        Otherwise this function returns false.
+
+        Params:
+            tx = transaction
+            threshold_pct = percentage by which the fee of the new transaction has
+              to be higher, than the previously highest double spend transaction
+
+        Returns:
+            whether the `tx` is an acceptable double spend transaction
+
+    ***************************************************************************/
+
+    public bool isAcceptableDoubleSpent (in Transaction tx, ubyte threshold_pct) @safe
+    {
+
+        Amount tx_fee;
+        if (fee_man.getAdjustedTXFee(tx, &utxo_set.peekUTXO, tx_fee) != null)
+            return false;
+
+        // only consider a double spend transaction, if its fee is
+        // considerably higher than the current highest fee
+        auto fee_threshold = getDoubleSpentHighestFee(tx);
+
+        // if the fee_threshold is null, it means there won't be any double
+        // spend transactions, after this transaction is added to the pool
+        if (!fee_threshold.isNull())
+            fee_threshold.get().percentage(threshold_pct + 100);
+
+        if (!fee_threshold.isNull() &&
+            (!tx_fee.isValid() || tx_fee < fee_threshold.get()))
+            return false;
+
+        return true;
+    }
+
+    /***************************************************************************
+
         Create the Coinbase TX for this nomination round and append it to the
         tx_set
 
@@ -879,6 +923,66 @@ public class Ledger
         this.enroll_man.getValidatorStakes(block.header.height, &this.utxo_set.peekUTXO, stakes,
             block.header.missing_validators);
         this.fee_man.accumulateFees(block, stakes, &this.utxo_set.peekUTXO);
+    }
+
+    /***************************************************************************
+
+        Returns the highest fee among all the transactions which would be
+        considered as a double spent, if `tx` transaction was in the transaction
+        pool.
+
+        If adding `tx` to the transaction pool would not result in double spent
+        transaction, then the return value is Nullable!Amount().
+
+        Params:
+            tx = transaction
+
+        Returns:
+            the highest fee among all the transactions which would be
+            considered as a double spend, if `tx` transaction was in the
+            transaction pool.
+
+    ***************************************************************************/
+
+    public Nullable!Amount getDoubleSpentHighestFee (in Transaction tx) @safe
+    {
+        Set!Hash tx_hashes;
+        pool.gatherDoubleSpentTXs(tx, tx_hashes);
+
+        const(Transaction)[] txs;
+        foreach (const tx_hash; tx_hashes)
+        {
+            const tx_ret = this.pool.getTransactionByHash(tx_hash);
+            if (tx_ret != Transaction.init)
+                txs ~= tx_ret;
+        }
+
+        if (!txs.length)
+            return Nullable!Amount();
+
+        return nullable(txs.map!((tx)
+            {
+                Amount tot_fee;
+                fee_man.getAdjustedTXFee(tx, &utxo_set.peekUTXO, tot_fee);
+                return tot_fee;
+            }).maxElement());
+    }
+
+    /***************************************************************************
+
+        Get a transaction from pool by hash
+
+        Params:
+            tx = the transaction hash
+
+        Returns:
+            Transaction or Transaction.init
+
+    ***************************************************************************/
+
+    public Transaction getTransactionByHash (in Hash hash) @trusted nothrow
+    {
+        return this.pool.getTransactionByHash(hash);
     }
 
     /***************************************************************************
--- a/source/agora/node/TransactionPool.d
+++ b/source/agora/node/TransactionPool.d
@@ -18,6 +18,7 @@ import agora.common.ManagedDatabase;
 import agora.common.Types;
 import agora.common.Set;
 import agora.consensus.data.Transaction;
+import agora.consensus.data.genesis.Test;
 import agora.crypto.Hash;
 import agora.serialization.Serializer;
 import agora.utils.Log;
@@ -29,6 +30,7 @@ import d2sqlite3.sqlite3;
 import std.algorithm;
 import std.conv : to;
 import std.exception : collectException, enforce;
+import std.functional : toDelegate;
 import std.range;
 import std.string;
 
@@ -98,7 +100,8 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public this (ManagedDatabase db)
+    public this (ManagedDatabase db,
+            DoubleSpentSelector double_spent_selector = toDelegate(&TransactionPool.defaultSelector))
     {
         this.db = db;
 
@@ -112,7 +115,7 @@ public class TransactionPool
 
         // Set selector after rebuilding the spender list, so nothing gets
         // filtered out
-        this.selector = &this.defaultSelector;
+        this.selector = double_spent_selector;
     }
 
     /// Used in unittests of this moduule only
@@ -424,7 +427,7 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    private size_t defaultSelector (Transaction[] txs)
+    private static size_t defaultSelector (Transaction[] txs)
     {
         return 0;
     }
@@ -720,7 +723,7 @@ unittest
     assert(pool.hasTransactionHash(tx3.hashFull()));
 }
 
-// test double-spend selection mechanism
+// test double-spend selection mechanism with max output selector
 unittest
 {
     // Pick the TX with max output value, assumes only one output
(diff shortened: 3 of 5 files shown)


ferencdg:
rebase #1

source/agora/utils/Test.d
```diff
@@ -677,6 +677,25 @@ public struct TxBuilder
...
+    ***************************************************************************/
+
+    public ref typeof(this) deductRemaining ()
+        return
+    {
+        this.leftover.value = Amount(0);
+
+        return this;
+    }
+
```
Geod24: Nope, this is one of those function that sounds good but will be a pain down the road.
The `TxBuilder` was made so that valid transactions are easy to make and invalid ones are hard.
This function does not, as it makes it trivial to create transactions that burns an arbitrary amount, something we seldom see IRL.

source/agora/consensus/Fee.d
```diff
@@ -401,6 +401,29 @@ public class FeeManager
...
+
+    ***************************************************************************/
+
+    public Amount getTXFee (in Transaction tx, scope UTXOFinder peekUTXO) nothrow @safe
+    {
+        Amount tot_fee;
+        Amount tot_data_fee;
+
+        getTXSetFees([tx], peekUTXO, tot_fee, tot_data_fee);
+        tot_fee.add(tot_data_fee);
```
Geod24: Bypasses the overflow check

source/agora/consensus/Fee.d
```diff
@@ -401,6 +401,29 @@ public class FeeManager
...
+
+        return tot_fee;
+    }
```
Geod24: Maybe you could just modify the other function to return the `sum`, since it currently returns `void` ?

source/agora/node/FullNode.d
```diff
@@ -491,6 +491,32 @@ public class FullNode : API
...
 
+        // all inputs have to be known in order to calculate the fee and
+        // compare it to the fee threshold
+        UTXO utxo;
+        foreach (const ref input; tx.inputs)
+            if (!utxo_set.peekUTXO(input.utxo, utxo))
+                return;
+
+        // only consider a double spend transaction, if its fee is
+        // considerably higher than the current highes fee
```
Geod24: highest

source/agora/node/Ledger.d
```diff
@@ -916,6 +918,44 @@ public class Ledger
...
+            tx = transaction
+
+        Returns:
+            the highest fee among all the transactions which would be
+            considered as a double spend, if `tx` transaction was in the
+            transaction pool.
+
+    ***************************************************************************/
+
+    public Nullable!Amount getDoubleSpentHighestFee(in Transaction tx) @safe
```
Geod24: Why not return `0` ?

source/agora/node/FullNode.d
```diff
@@ -491,6 +491,32 @@ public class FullNode : API
...
+        auto fee_threshold = ledger.getDoubleSpentHighestFee(tx);
+
+        // if the fee_threshold is null, it means there won't be any double
+        // spend transactions, after this transaction is added to the pool
+        if (!fee_threshold.isNull())
+        {
+            fee_threshold.div(100);
```
Geod24: This ignore the remainder, which means it might not respect the percentage

source/agora/node/FullNode.d
```diff
@@ -648,7 +674,29 @@ public class FullNode : API
...
+
+        Subclasses can override this method and return
+        a DoubleSpentSelector
+
+        Returns:
+            an instance of DoubleSpentSelector
+
+    ***************************************************************************/
+
+    protected DoubleSpentSelector getDoubleSpentSelector () nothrow @safe
```
Geod24: Why not make this function the double spend detector (and passing its address) instead of the extra dg ?

source/agora/node/TransactionPool.d
```diff
@@ -790,6 +791,69 @@ unittest
...
+        [Output(Amount(2), KeyPair.random.address)]
+    };
+
+    // double spent transaction, trying to spend parent
+    Transaction tx3 =
+    {
+        TxType.Payment,
+        [Input(tx1.hashFull(), 0)],
+        [Output(Amount(1), KeyPair.random.address)]
+    };
```
Geod24: Those transactions are not valid. Please use valid transactions in tests.

Geod24 (commented):
Few things

source/agora/node/FullNode.d
```diff
@@ -491,6 +491,32 @@ public class FullNode : API
...
+            fee_threshold.div(100);
+            fee_threshold.mul(100 + config.node.double_spent_threshold_pct);
+        }
+
+        // an attacker could easily send a transaction with a very high fee
+        // which would make the fee_threshold invalid after the multiplication
+        // because of this, there is an explicit check to fee_threshold.isValid()
+        if (!fee_threshold.isNull() &&
+           (!fee_threshold.isValid() || fee_man.getTXFee(tx, &utxo_set.peekUTXO) < fee_threshold))
+            return;
```
omerfirmak: Do we think this is the right place to do this? Shouldn't this all be abstracted away in Ledger/TransactionPool?

source/agora/node/FullNode.d
omerfirmak: `percentage()` + `add()` is probably the right way to go.

source/agora/utils/Test.d
ferencdg: I don't thing burning would make the transaction invalid. :) The suggested usecase for this function is txbuilder.draw(100, [pk]).deductRemaining() which essentially set the remaining amount as fee. I can make this function version(unittest) if you would like it.

We already have a deduct() function, so alternatively I could have defined a new function, that returned the remaining amount and pass it to deduct.

source/agora/consensus/Fee.d
ferencdg: I think it is better to have only 1 clearly defined responsibility for every function and combine thos functions, if necessary. Not everyone needs the sum of tot_fee and tot_data_fee, so we should not calculate it unnecessarily(even though this is just an addition).

(In fact, I think the original function should just return an array of tot_fee and tot_data fee and have a 3rd function aggregate it, if aggregation is necessary)

source/agora/node/Ledger.d
ferencdg: it would be a loss of information, having a double spend transaction with 0 fee is not the same as having no double spend transaction(the case when it is null)

source/agora/node/FullNode.d
ferencdg: I like that idea...

source/agora/node/TransactionPool.d
ferencdg: if you mean 
[Input(Hash.init, 0)],
is not a valid input, then I see this pattern in our code everywhere, I quick search reveals that every single Agora developer(except Chris) wrote tests like this at some point :)

should we clean those up too?

source/agora/node/FullNode.d
ferencdg: I agree, I was also thinking of moving this into a separate method in the Ledger class so it can be reused.

source/agora/node/TransactionPool.d
Geod24: > is not a valid input, then I see this pattern in our code everywhere

And we suffer from it every week :(

ferencdg: okay, let me fix this now, and also will also raise a ticket for cleaning up the entire codebase, and will do when I have more time

source/agora/node/FullNode.d
ferencdg: @omerfirmak not sure I understand your 'percentage() + add()' solution, let's say the amount is 199 and percentage is 20%

are you suggesting (199/100)*20 + 199 = 219? If we had real numbers then the result should be 238.8.

I think Mathias suggestion with the remainder is 199 + (99*20/100) + (199/100)*20 = 238

if we could just turn out the overflow check and would calculate 199*120/100 that would be 238 which is very close to 238.8 and would require less operations

ferencdg: in general if we do something like Y/X then the biggest mistake of integer calculation will be (X-1)/X and we amplify it by 20 when we multiply. It would be better to multiply first and then divide where our error is (X-1/X) maximum.

omerfirmak: @ferencdg , `Amount.percentage()`will first multiply then div by 100. So it should have minimal error.

Geod24: @ferencdg : To clarify: with an amount of 199 and 20%, we should only accept a fee `>= 239` if we can't represent `238.8`. Even if the result of the calculation was `238.00001` it should only allow `>= 239`.

ferencdg: > @ferencdg , `Amount.percentage()`will first multiply then div by 100. So it should have minimal error.

that can overflow though(the one below, only overflows when the final result overflows)... I think I just go ahead with 
199 + (99*20/100) + (199/100)*20

omerfirmak: With our current `Amount.MaxUnitSupply` it can't overflow.

ferencdg: > With our current `Amount.MaxUnitSupply` it can't overflow.

what if an attacker sends a fee of MaxUnitSupply - 1? At this point no checks are mode on whether the transaction is valid.

ferencdg: In theory we can have a transaction with a fee of Maxsupply/18 > fee > Maxsupply/20, and it will overflow ;) while the
199 + (99*20/100) + (199/100)*20 solution not

ferencdg: > @ferencdg : To clarify: with an amount of 199 and 20%, we should only accept a fee `>= 239` if we can't represent `238.8`. Even if the result of the calculation was `238.00001` it should only allow `>= 239`.

so do you know how to estimate the error, so I can correct the integer division(if we use the 199 + (99*20/100) + (199/100)*20) formula... Btw, do we really want to be this precise? I just randomly choose 20%....

omerfirmak: I was solely talking about the call to `percentage()`. I don't think it will overflow, ever. But after you add the `percentage()` result to the fee, it might overflow yes. At that point we can either accept or reject that TX.

source/agora/consensus/Fee.d
ferencdg: That was actually intentional. This method can be called with a partially validated transaction, and the fee might very well overflow. I can certainly not `assert` here, as this transaction can come from an attacker, and when that happens all debug build of agora intstance would crash. It should be the responsibility of the caller to check for validity and decide what to do. I also would not like to fully validate the input transaction before it is passed to this function, as I would like to fail fast, if the fee threshold is not respected.

I will modify the documentation and mention that the result should be checked. Alternatively, I could change the interface and return a bool.

ferencdg:
thanks for the review @Geod24, @omerfirmak, I made the code changes

source/agora/common/Amount.d
```diff
@@ -47,6 +47,12 @@ public struct Amount
     /// Internal data storage
     private ulong value;
 
+    /// The size of the Amount object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
```
Geod24: Just use `.sizeof` from the caller

source/agora/consensus/data/DataPayload.d
```diff
@@ -26,6 +26,11 @@ public struct DataPayload
     /// The byte array of transaction data
     public const(ubyte)[] data;
 
+    /// The size of the data DataPayload object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
```
Geod24: Style

source/agora/consensus/data/DataPayload.d
```diff
@@ -26,6 +26,11 @@ public struct DataPayload
...
+    {
+        return data.length * data[0].sizeof;
```
Geod24: `this`

Geod24 (commented):
This needs a rebase right ?

source/agora/common/Amount.d
ferencdg: > This needs a rebase right ?

rebase #2, also your latest comments were fixed in an already merged PR(those commits were chopped of from this PR)

Geod24:
Please rebase

ferencdg:
rebase #3

ferencdg:
:green_circle:

source/agora/node/Ledger.d
```diff
@@ -500,6 +502,48 @@ public class Ledger
...
+        // spend transactions, after this transaction is added to the pool
+        if (!fee_threshold.isNull())
+            fee_threshold.percentage(threshold_pct + 100);
+
+        if (!fee_threshold.isNull() &&
+           (!tx_fee.isValid() || tx_fee < fee_threshold))
+            return false;
+
+        return true;
+    }
```
Geod24: Doesn't this belong in the pool rather than the Ledger ?

source/agora/node/Ledger.d
```diff
@@ -881,6 +925,66 @@ public class Ledger
...
+            tx = transaction
+
+        Returns:
+            the highest fee among all the transactions which would be
+            considered as a double spend, if `tx` transaction was in the
+            transaction pool.
+
+    ***************************************************************************/
+
+    public Nullable!Amount getDoubleSpentHighestFee(in Transaction tx) @safe
```
Geod24: Ditto about being in the pool. Also, space before parens.

source/agora/node/TransactionPool.d
```diff
@@ -98,7 +100,8 @@ public class TransactionPool
 
     ***************************************************************************/
 
-    public this (ManagedDatabase db)
+    public this (ManagedDatabase db,
+            DoubleSpentSelector double_spent_selector = toDelegate(&TransactionPool.defaultSelector))
```
Geod24, line 104: Don't think we want the default value, since it's only instantiated a handful of times.
For unittests, there's a specific ctor.

source/agora/test/Base.d
```diff
@@ -1558,6 +1558,12 @@ private mixin template TestNodeMixin ()
         return new ManagedDatabase(":memory:");
     }
 
+    /// Return a transaction pool backed by an in-memory SQLite db
+    public override TransactionPool makeTransactionPool ()
+    {
+        return new TransactionPool(new ManagedDatabase(":memory:"), &getDoubleSpentSelector);
```
Geod24: Why ? The default version uses `cacheDB`. I actually planned to remove those, as nowadays cacheDB / stateDB is the only place where dependency injection is done.

source/agora/test/TransactionReplacement.d
```diff
@@ -0,0 +1,86 @@
...
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto nodes = network.clients;
+    auto node0 = nodes[0];
+    auto node1 = nodes[1];
+
+    auto input_tx = GenesisBlock.txs.filter!(tx => tx.type == TxType.Payment).front();
```
Geod24, line 44: So `genesisSpendable.front()` ?

source/agora/test/TransactionReplacement.d
```diff
@@ -0,0 +1,86 @@
...
+        .deductRemaining().sign()).array();
+
+    // send the transaction with 10000 fee to node0
+    nodes[0].putTransaction(txs[0]);
+
+    // wait until the transaction propages to all nodes
+    nodes.each!(node => node.hasTransactionHash(hashFull(txs[0])).retryFor(2.seconds));
+
+    // send transactions with fee 10100, 10200, 11000 to node0
+    iota(1,4).each!(tx_ind => node0.putTransaction(txs[tx_ind]));
```
Geod24: IMO the following is much easier to read:
```suggestion
    txs[1 .. 4].each!(tx => node0.putTransaction(tx));
```

source/agora/test/TransactionReplacement.d
```diff
@@ -0,0 +1,86 @@
...
+
+    // wait 2 seconds and make sure none of the nodes have those transactions
+    Thread.sleep(2.seconds);
+    iota(1,4).each!(tx_ind => nodes.each!(node => !node.hasTransactionHash(hashFull(txs[tx_ind]))));
```
Geod24: Ditto iota => Slice

source/agora/test/TransactionReplacement.d
```diff
@@ -0,0 +1,86 @@
...
+    // send a transaction with 13000s fee to node0
+    nodes[0].putTransaction(txs[4]);
+
+    // wait until the transaction propages to all nodes
+    nodes.each!(node => node.hasTransactionHash(hashFull(txs[4])).retryFor(2.seconds));
+
+    // wait until the first block is created and return it
+    network.expectHeight(Height(1));
+    const block = node1.getBlock(Height(1));
+    auto payment_txs = block.txs.filter!(tx => tx.type == TxType.Payment).array();
```
Geod24: Well technically there will only be one tx, so we might make the assumptions here simpler ?

Geod24 (approved):
LGTM

source/agora/node/Ledger.d
```diff
@@ -500,6 +502,48 @@ public class Ledger
...
+            return false;
+
+        // only consider a double spend transaction, if its fee is
+        // considerably higher than the current highest fee
+        auto fee_threshold = getDoubleSpentHighestFee(tx);
+
+        // if the fee_threshold is null, it means there won't be any double
+        // spend transactions, after this transaction is added to the pool
+        if (!fee_threshold.isNull())
+            fee_threshold.percentage(threshold_pct + 100);
```
Geod24: This line, along with 541, triggers a new deprecation message. It should call `.get()`

source/agora/test/TransactionReplacement.d
Geod24: Ah, `genesisSpendable()` gives you a range of `TxBuilder` but this needs txs. I see.

source/agora/test/TransactionReplacement.d
Geod24: Nevermind, I forgot about coinbase :)

Geod24:
I did the style fixes, the rest is not big enough to delay the merge any longer.

PR #1885, opened 2021-04-06
Allow Agora to listen to multiple interfaces

Geod24:
Contains a bunch of refactoring, but the end goal is to make Agora able to listen either on HTTP or RPC.
Part of #203 .

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -91,6 +91,9 @@ public struct Config
     /// The node config
     public NodeConfig node;
 
+    /// Configuration for interfaces the node expose (only http for now)
+    public immutable InterfaceConfig[] interfaces = InterfaceConfig.Default;
+
     /// Consensus parameters for the chain
     public ConsensusConfig consensus;
 
@@ -114,6 +117,38 @@ public struct Config
 
     /// Event handler config
     public EventHandlerConfig event_handlers;
+}
+
+/// Used to specify endpoint-specific configuration
+public struct InterfaceConfig
+{
+    static assert(!hasUnsharedAliasing!(typeof(this)),
+        "Type must be shareable accross threads");
+
+    /// Type of interface one is able to register
+    public enum Type
+    {
+        // FIXME: https://github.com/sociomantic-tsunami/ocean/issues/846
+        // /// Cannonical name is in upper case
+        // HTTP = 0,
+        /// Convenience alias for parsing
+        http = 0,
+    }
+
+    /// Ditto
+    public Type type;
+
+     /// Bind address
+    public string address;
+
+    /// Bind port
+    public ushort port;
+
+    /// Default values when none is given in the config file
+    private static immutable InterfaceConfig[/* Type.max */ 1] Default = [
+        // Publicly enabled by default
+        { type: Type.http, address: "0.0.0.0", port: 0xB0A, },
+    ];
 }
 
 /// Node config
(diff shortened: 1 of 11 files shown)


Geod24:
Don't merge, I will probably add a second interface in the system integration test to make sure this works properly.
So Faucet and the test would contact Agoras on HTTP, and the Agoras would talk to themselves via another interface.
I might even mix the interfaces just so that we get a very good coverage.

Geod24:
Damn, found a Formatter bug: https://github.com/sociomantic-tsunami/ocean/issues/846

Geod24:
Can I haz review ?

source/agora/common/Config.d
```diff
@@ -116,6 +119,38 @@ public struct Config
...
+public struct InterfaceConfig
+{
+    static assert(!hasUnsharedAliasing!(typeof(this)),
+        "Type must be shareable accross threads");
+
+    /// Type of interface one is able to register
+    public enum Type
+    {
+        // FIXME: https://github.com/sociomantic-tsunami/ocean/issues/846
+        // /// Cannonical name is in upper case
```
linked0, line 132: Kind of a strange format for the comment.

source/agora/common/Config.d
```diff
@@ -116,6 +119,38 @@ public struct Config
...
+     /// Bind address
+    public string address;
+
+    /// Bind port
+    public ushort port;
+
+    /// Default values when none is given in the config file
+    private static immutable InterfaceConfig[/* Type.max */ 1] Default = [
+        // Publicly enabled by default
+        { type: Type.http,  address:   "0.0.0.0", port: 0xB0A, },
```
linked0: There are unnecessary spaces and a period.

source/agora/common/Config.d
```diff
@@ -377,13 +406,42 @@ private const(string)[] parseSequence (string section, in CommandLine cmdln,
...
+
+            interfaces ~= InterfaceConfig(type, address, port);
+        }
+
+        if (!interfaces.length)
+            throw new Exception("The 'interfaces' section must be empty or have valid values");
+    }
+    else
+        interfaces = InterfaceConfig.Default;
+
```
linked0, line 436: Why don't you make a function for that, like `parseInterfaceConfig`?

source/agora/node/Runner.d
```diff
@@ -126,16 +124,23 @@ public Listeners runNode (Config config)
 
     setTimer(0.seconds, &result.node.start, Periodic.No);  // asynchronous
 
-    scope (exit)
+
+
+    // HTTP interfaces for the node
+    foreach (interface_; config.interfaces.filter!(i => i.type == InterfaceConfig.Type.http))
```
linked0, line 130: `interf` would be better than `interface_`.

linked0 (approved):
LGTM just some nits

source/agora/node/Runner.d
Geod24: https://dlang.org/dstyle.html#naming_keywords

source/agora/common/Config.d
Geod24: Ah, it'll align in a future commit, with the TCP interface

source/agora/common/Config.d
Geod24: It's because the two lines are commented out. If I used `///` the documentation would incorrectly be applied to the next member. It isn't great but I hope to get this bug fixed in the near-ish future.

source/agora/common/Config.d
Geod24: But I removed them in the meantime

source/agora/common/Config.d
Geod24: Feels a bit overkill, given the simplicity of this code. Also, `parseSequence` should really be able to handle this.

Geod24:
Rebased and fixed spaces, as it was green before, merging.

PR #1886, opened 2021-04-06
Exclude validators from being banned by the BanManager

### LINKED ISSUE #1674: Exclude validators from being banned by the BanManager
Now that we have network registries, we should make sure that the validators are *not* banned.

omerfirmak:
Fixes #1674

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -55,7 +55,7 @@ import agora.consensus.data.PreImageInfo;
 import agora.consensus.EnrollmentPool;
 import agora.consensus.PreImage;
 import agora.consensus.validation;
-import agora.consensus.state.ValidatorSet;
+public import agora.consensus.state.ValidatorSet;
 import agora.consensus.state.UTXOCache;
 import agora.crypto.ECC;
 import agora.crypto.Hash;
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -1241,6 +1241,36 @@ public class NetworkManager
 
         return new RestInterfaceClient!TransactionReceivedHandler(settings);
     }
+
+    /***************************************************************************
+
+        Whitelist a PublicKey to avoid banning it
+
+        Params:
+            key = the PublicKey to whitelist
+
+    ***************************************************************************/
+
+    public void whitelist (PublicKey key)
+    {
+        this.peers[].filter!(p => p.key == key)
+            .each!(p => this.banman.whitelist(p.client.address));
+    }
+
+    /***************************************************************************
+
+        Unwhitelist a PublicKey to allow banning it
+
+        Params:
+            key = the PublicKey to unwhitelist
+
+    ***************************************************************************/
+
+    public void unwhitelist (PublicKey key)
+    {
+        this.peers[].filter!(p => p.key == key)
+            .each!(p => this.banman.unwhitelist(p.client.address));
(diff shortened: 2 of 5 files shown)


source/agora/network/Manager.d
```diff
@@ -194,6 +194,7 @@ public class NetworkManager
                 }
 
                 log.info("Found new Validator: {} (key: {})", address, key);
+                this.outer.banman.whitelist(address);
```
Geod24: Well... This just checks if the node implements a certain API. It would be trivial to write a node that exploits this not to get banned, no ?

omerfirmak: Yep, we could whitelist them as long as they are enrolled maybe? A not enrolled validator is no different than a FullNode anyways.

Geod24: Yes we should. But obviously this means connecting the `BanManager` to the `Ledger`.

omerfirmak: How about `onAcceptedBlock`to get a list of changes in validator set instead of just a bool?

omerfirmak: @Geod24, whitelisting enrolled validators is not a perfect solution as well. Anyone could impersonate a Validator at network level right? Maybe discovery handshake should include an authentication step.

Geod24: > Maybe discovery handshake should include an authentication step.

You're taking the words out of my mouth :)

Geod24: > How about `onAcceptedBlock` to get a list of changes in validator set instead of just a bool?

Sounds about right.

omerfirmak:
Updated to whitelist enrolled validators only.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -55,7 +55,7 @@ import agora.consensus.data.PreImageInfo;
 import agora.consensus.EnrollmentPool;
 import agora.consensus.PreImage;
 import agora.consensus.validation;
-import agora.consensus.state.ValidatorSet;
+public import agora.consensus.state.ValidatorSet;
```
AndrejMitrovic, line 58: Deliberate change?

source/agora/node/FullNode.d
```diff
@@ -416,8 +416,16 @@ public class FullNode : API
 
     protected bool acceptBlock (const ref Block block) @trusted
     {
+        ExpiringValidator[] ex_validators;
+        this.enroll_man.getExpiringValidators(block.header.height, ex_validators);
         // Attempt to add block to the ledger (it may be there by other means)
-        this.ledger.acceptBlock(block);
+        if (this.ledger.acceptBlock(block))
+        {
+            ex_validators[].each!(ex => this.network.unwhitelist(ex.pubkey));
```
AndrejMitrovic: You don't actually need to slice an array, it's already a slice. :)

Just curious, do you do it to make code more readable?

omerfirmak: It's just me being a bad D programmer:)

source/agora/consensus/EnrollmentManager.d
omerfirmak: Yep, `ExpiringValidator` is defined in `ValidatorSet`. That should be visible to modules using `EnrollmentManager.getExpiringValidators`

AndrejMitrovic (approved):

hewison-chris (approved):
LGTM

PR #1889, opened 2021-04-06
Stop nominating_timer on shutdown

Geod24:


omerfirmak (approved):

Geod24:
Looks like this causes more issue. I think there might be some more timers lingering, will have a look this afternoon.

ferencdg (approved):

PR #1892, opened 2021-04-07
Change Flash retry logic to use backoff delays

### LINKED ISSUE #1890: Flash: Replace dumb retry logic with a truncated exponential backoff algorithm
Most importantly, we need to use randomness (jitter). And we need to add an additional max backoff config option. There's good resources on this here:

https://en.wikipedia.org/wiki/Exponential_backoff
https://cloud.google.com/iot/docs/how-tos/exponential-backoff
https://docs.microsoft.com/en-us/azure/architecture/best-practices/transient-faults
https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

A part of this was implemented by @omerfirmak (thanks!), but it's not used consistently across the code and it lacks the truncation & jitter.

Edit: This approach is needed not just for connection errors (which would ideally be handled in vibe.d), but also race conditions between two flash nodes. In particular the nodes may not have the exact same sequence ID / channel state / block height at any given time - therefore the nodes should attempt to retry their requests when it makes sense. If for example the sequence ID doesn't match (yet), the request should be retried. But if the node sends an invalid signature, then the request doesn't need to be retried because it shows there was a logical error or the node was byzantine.

AndrejMitrovic:
This also includes a bunch of fixes to the internals.

But let's see what the CI says.

Fixes #1890

### DIFF
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -36,6 +36,7 @@ import agora.crypto.Schnorr;
 import agora.script.Engine;
 import agora.script.Lock;
 import agora.serialization.Serializer;
+import agora.utils.Backoff;
 import agora.utils.Log;
 
 import std.array;
@@ -104,6 +105,9 @@ public class Channel
     /// node ctor is done (see Flash restart tests)
     private FlashAPI delegate (in Point peer_pk, Duration timeout) getFlashClient;
 
+    /// Retry delay algorithm
+    private Backoff backoff;
+
 
     /***************************************************************************
 
@@ -155,6 +159,8 @@ public class Channel
         this.onChannelOpen = onChannelOpen;
         this.onPaymentComplete = onPaymentComplete;
         this.onUpdateComplete = onUpdateComplete;
+        this.backoff = new Backoff(this.flash_conf.retry_multiplier,
+            this.flash_conf.max_retry_delay.total!"msecs".to!uint);
 
         this.dump();
 
@@ -200,6 +206,8 @@ public class Channel
         this.onChannelOpen = onChannelOpen;
         this.onPaymentComplete = onPaymentComplete;
         this.onUpdateComplete = onUpdateComplete;
+        this.backoff = new Backoff(this.flash_conf.retry_multiplier,
+            this.flash_conf.max_retry_delay.total!"msecs".to!uint);
 
         this.taskman.setTimer(0.seconds,
             (this.channel_updates.length == 0)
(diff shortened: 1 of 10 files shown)


AndrejMitrovic:
> submodules/agora/source/agora/flash/UpdateSigner.d(35,8): Error: module `Backoff` is in file 'agora/utils/Backoff.d' which cannot be read

So every time we add a new module we first have to update Faucet's `dub.json` file and add that file there, even without knowing whether the PR will be accepted. This is ridiculous.

source/agora/flash/Channel.d
```diff
@@ -1157,13 +1185,24 @@ LOuter: while (1)
     {
         const old_balance = this.cur_balance;
 
+        const old_seq_id = this.cur_seq_id;
```
omerfirmak: `collectSignatures` already takes `update.seq_id` as an argument. Could we wait until `collectSignatures` to return to update the `cur_seq_id`?

source/agora/flash/Node.d
```diff
@@ -853,10 +853,11 @@ public abstract class FlashNode : ControlFlashAPI
         // our own secret (we are the payee)
         if (auto secret = payment_hash in this.secrets)
         {
-            assert(!error); // Payee should never fail to receive
-            channel.learnSecrets([*secret], [], this.last_block_height);
+            if (!error)
```
omerfirmak: I can't see how, is there an example you can give? AFAICS, `onPaymentComplete` is never called with an error for incoming payments anyways.

source/agora/flash/Channel.d
AndrejMitrovic: Sure, good idea.

source/agora/flash/Node.d
AndrejMitrovic: The signing process can fail. I added a notification for failed incoming payments, but now that I think about it maybe it should be a separate callback.

I'll rework it a bit.

AndrejMitrovic:
Thanks for the tips @omerfirmak. I removed the incoming payment change and fixed up the sequences.

AndrejMitrovic:
It looks like https://github.com/bosagora/faucet/pull/56 isn't ready yet so we'll have to do this in a funky way by juggling submodules..

AndrejMitrovic:
Edit: s/submodules/docker images

AndrejMitrovic:
It might be easier to just merge it first, then fix faucet to add the new import, and then update agora again. Let's wait for green from the CI.

AndrejMitrovic:
Merging this, and I'll fix up the faucet version soon.

PR #1893, opened 2021-04-07
Logger: Replace opDispatch by overloads

Geod24:
While working on #1885 I found a bug in Ocean, but in order to diagnose it I first add to apply this diff.
Since it's a recurring event, might as well upstream it.


AndrejMitrovic (approved):

kchulj (approved):
LGTM.

PR #1894, opened 2021-04-07
Don't wait for parallel/heavy tests if failed early

AndrejMitrovic:
I think this explains those 60 minute run times..

For example this code blocks:

```D
import core.sync.semaphore;
import std.stdio;

void main ()
{
    auto finished_tasks_num = new Semaphore(0);
    finished_tasks_num.wait();  // blocked
    writeln("Done");
}
```


Geod24 (approved):
That's embarassing

kchulj (approved):

PR #1896, opened 2021-04-07
Properly work around network discovery bug

AndrejMitrovic:
Previous fix in https://github.com/bosagora/agora/pull/1895 wrongly assumed that the issue was too many connected nodes. But sometimes it's exactly the number we need (Genesis - 1 for the self node).

Hopefully this makes the test stop failing.

### DIFF
--- a/source/agora/test/NetworkDiscovery.d
+++ b/source/agora/test/NetworkDiscovery.d
@@ -33,7 +33,7 @@ unittest
     foreach (key, node; network.nodes)
     {
         auto addresses = node.client.getNodeInfo().addresses.keys;
-        assert(addresses.sort.uniq.count >= GenesisValidators,  // may connect to itself (#772)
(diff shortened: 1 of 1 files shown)


AndrejMitrovic:
```
Unittesting agora.test.EnrollmentManager..
================================ ASSERT HANDLER ===============================
[source/agora/consensus/validation/Enrollment.d:88] Assertion thrown during test:
```

This looks like a new bug.

AndrejMitrovic:
It looks like this check was added recently in this commit: https://github.com/bosagora/agora/commit/2419d636422e4ce986dba4805b2031202973bae4

AndrejMitrovic:
And the tests still took 60 minutes to finish despite the early failure. Hmm..

omerfirmak (approved):

AndrejMitrovic:
Segfault in the Flash tests. 🤦‍♂️

I'll re-run this again.

PR #1897, opened 2021-04-07
NetworkManager: Authenticate Validator node identity

omerfirmak:
Nodes could impersonate another Validator since there was
no authentication at NetworkManager.

### DIFF
--- a/source/agora/api/Validator.d
+++ b/source/agora/api/Validator.d
@@ -104,7 +104,7 @@ public interface API : agora.api.FullNode.API
 
     ***************************************************************************/
 
-    public Identity getPublicKey (PublicKey key);
+    public Identity getPublicKey (PublicKey key = PublicKey.init);
 
     /***************************************************************************
 
--- a/source/agora/api/Validator.d
+++ b/source/agora/api/Validator.d
@@ -46,6 +46,7 @@ import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.ValidatorBlockSig;
 import agora.crypto.Hash;
 import agora.crypto.Key;
+import agora.crypto.Schnorr;
 import agora.flash.api.FlashAPI;
 static import agora.api.FullNode;
 
@@ -62,6 +63,16 @@ public import agora.api.FullNode;
 @serializationPolicy!(Base64ArrayPolicy)
 public interface FlashValidatorAPI : API, ExtendedFlashAPI
 {
+}
+
+/// Identity of a Validator node
+public struct Identity
+{
+    /// Public Key of the node
+    PublicKey key;
+
+    /// MAC
+    ubyte[] mac;
 }
 
 /*******************************************************************************
@@ -82,15 +93,18 @@ public interface API : agora.api.FullNode.API
 
     /***************************************************************************
 
+        Params:
+            key = key of the peer
+
         Returns:
-            The public key of this node
+            The public key of this node and a secret to prove identity
 
         API:
             GET /public_key
 
     ***************************************************************************/
 
-    public PublicKey getPublicKey ();
+    public Identity getPublicKey (PublicKey key);
 
     /***************************************************************************
 
--- a/source/agora/network/Client.d
+++ b/source/agora/network/Client.d
@@ -223,17 +223,21 @@ public class NetworkClient
 
     /***************************************************************************
 
+        Params:
+            key = key of the peer
+
         Returns:
-            the node's public key.
+            The public key of this node and a secret to prove identity
 
         Throws:
             `Exception` if the request failed.
 
     ***************************************************************************/
 
-    public PublicKey getPublicKey () @trusted
+    public Identity getPublicKey (PublicKey key = PublicKey.init) @trusted
     {
-        return this.attemptRequest!(API.getPublicKey, Throw.Yes)(this.api);
+        return this.attemptRequest!(API.getPublicKey, Throw.Yes)(this.api,
+            key);
     }
 
     /***************************************************************************
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -159,9 +159,30 @@ public class NetworkManager
             {
                 try
                 {
+                    import agora.flash.OnionPacket : generateSharedSecret;
+                    import libsodium.crypto_auth;
+
                     // LocalRest will return PublicKey.init,
                     // vibe.d will throw HTTPStatusException with status == 404
-                    key = client.getPublicKey();
+                    const ephemeral_kp = KeyPair.random();
+                    auto id = client.getPublicKey(ephemeral_kp.address);
+
+                    if (id.key == PublicKey.init)
+                        break;
+
+                    Hash shared_sec = generateSharedSecret(true,
+                        ephemeral_kp.secret, id.key).hashFull();
+                    static assert(shared_sec.sizeof >= crypto_auth_KEYBYTES);
+
+                    if (id.mac.length != crypto_auth_KEYBYTES ||
+                        crypto_auth_verify(id.mac.ptr, id.key[].ptr,
+                            id.key[].length, shared_sec[].ptr) != 0)
+                    {
+                        this.outer.banman.ban(this.address);
+                        return;
+                    }
+
+                    key = id.key;
                     break;
                 }
                 catch (Exception ex)
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -256,11 +256,25 @@ public class Validator : FullNode, API
     }
 
     /// GET /public_key
-    public override PublicKey getPublicKey () pure nothrow @safe
+    public override Identity getPublicKey (PublicKey key) nothrow @safe
     {
+        import agora.flash.OnionPacket : generateSharedSecret;
+        import libsodium.crypto_auth;
         endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(
             1, "public_key", "http");
-        return this.config.validator.key_pair.address;
+
+        Identity id = Identity(this.config.validator.key_pair.address);
+        if (key == PublicKey.init)
+            return id;
+
+        Hash shared_sec = generateSharedSecret(false,
+            this.config.validator.key_pair.secret, key).hashFull();
+        static assert(shared_sec.sizeof >= crypto_auth_KEYBYTES);
+
+        id.mac.length = crypto_auth_BYTES;
+        () @trusted { crypto_auth (id.mac.ptr, id.key[].ptr,
+            id.key[].length, shared_sec[].ptr); } ();
+        return id;
     }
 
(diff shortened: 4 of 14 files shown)


source/agora/api/Validator.d
```diff
@@ -82,15 +93,18 @@ public interface API : agora.api.FullNode.API
 
     /***************************************************************************
 
+        Params:
+            challange = a challange for Validator to prove identity
+
         Returns:
-            The public key of this node
+            The public key of this node and signature for the challange
```
Geod24: I like that you are consistent in your typos :D
`challenge`

source/agora/network/Client.d
```diff
@@ -223,17 +223,21 @@ public class NetworkClient
...
 
     ***************************************************************************/
 
-    public PublicKey getPublicKey () @trusted
+    public Identity getPublicKey (ulong challange) @trusted
     {
-        return this.attemptRequest!(API.getPublicKey, Throw.Yes)(this.api);
+        return this.attemptRequest!(API.getPublicKey, Throw.Yes)(this.api,
+            challange);
     }
```
Geod24: We might want to think about the challenge for a bit, because we're essentially offering an endpoint to sign a random 8 bytes... That could be dangerous. Luckily there's a proven way to securely exchange keys... Diffie Hellman.

source/agora/test/Base.d
```diff
@@ -1398,6 +1398,16 @@ public interface TestAPI : ValidatorAPI
     ***************************************************************************/
 
     public TimePoint getNetworkTime ();
+
+    /***************************************************************************
+
+        Returns:
+            the adjusted clock time taking into account the clock drift compared
```
linked0: Should be the right description.

linked0 (approved):
LGTM but needs a correction.

source/agora/network/Client.d
omerfirmak: Very good point, thanks. What about this one?

source/agora/node/Validator.d
```diff
@@ -264,11 +264,14 @@ public class Validator : FullNode, API
...
 
     /// GET /public_key
-    public override PublicKey getPublicKey () pure nothrow @safe
+    public override Identity getPublicKey (ulong challenge) nothrow @safe
     {
+        import std.random;
         endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(
             1, "public_key", "http");
-        return this.config.validator.key_pair.address;
+        auto rand = uniform!ulong;
```
AndrejMitrovic: Also when it comes to randomness, we should always prefer to use libsodium.

omerfirmak: Updated, thanks!

AndrejMitrovic (approved):

source/agora/network/Client.d
Geod24: It's still susceptible to replay attacks, isn't it ?

omerfirmak: Yes:( Let's just append the random numbers instead of adding them. Pushing in a second.

omerfirmak:
Updated to implement;

``` dlang
unittest
{
    import agora.flash.OnionPacket;
    auto n1_ephemeral_kp = KeyPair.random();
    auto n2_kp = KeyPair.random();
    // n1 sends the ephemeral public key to n2
    // n2 generates a shared secret
    auto n2_shared_sec = generateSharedSecret(false, n2_kp.secret, n1_ephemeral_kp.address);
    // n2 sends it's PublicKey and n2_shared_sec back to n1.
    // n1 should be able to generate the same shared secret
    auto n1_shared_sec = generateSharedSecret(true, n1_ephemeral_kp.secret, n2_kp.address);
    assert(n2_shared_sec == n1_shared_sec); // authenticated
}
```

source/agora/network/Manager.d
```diff
@@ -159,9 +159,30 @@ public class NetworkManager
...
-                    key = client.getPublicKey();
+                    const ephemeral_kp = KeyPair.random();
+                    auto id = client.getPublicKey(ephemeral_kp.address);
+
+                    if (id.key == PublicKey.init)
+                        break;
+
+                    Hash shared_sec = generateSharedSecret(true,
+                        ephemeral_kp.secret, id.key).hashFull();
+                    assert(shared_sec[].length >= crypto_auth_KEYBYTES);
```
```suggestion
                    static assert(shared_sec.sizeof >= crypto_auth_KEYBYTES);
```

source/agora/node/Validator.d
```diff
@@ -264,11 +264,25 @@ public class Validator : FullNode, API
...
     {
+        import agora.flash.OnionPacket : generateSharedSecret;
+        import libsodium.crypto_auth;
         endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(
             1, "public_key", "http");
-        return this.config.validator.key_pair.address;
+
+        Identity id = Identity(this.config.validator.key_pair.address);
+        if (key == PublicKey.init)
+            return id;
```
Geod24, line 268: That should never happen, could it ?

source/agora/node/Validator.d
```diff
@@ -264,11 +264,25 @@ public class Validator : FullNode, API
...
+
+        Hash shared_sec = generateSharedSecret(false,
+            this.config.validator.key_pair.secret, key).hashFull();
+        assert(shared_sec[].length >= crypto_auth_KEYBYTES);
```
Geod24: Ditto `static assert`

source/agora/test/NetworkDiscovery.d
```diff
@@ -83,3 +87,65 @@ unittest
...
+    __gshared NodePair impersonator_node;
+
+    static class ImpersonatorValidator : TestValidatorNode
+    {
+        mixin ForwardCtor!();
+
+        /// GET /public_key
+        protected override Identity getPublicKey (PublicKey key) nothrow @safe
+        {
+            () @trusted { call_count++; } ();
```
Geod24: `atomicOp`

source/agora/test/NetworkDiscovery.d
```diff
@@ -83,3 +87,65 @@ unittest
...
+    }
+
+    TestConf conf;
+    auto network = makeTestNetwork!ImpersonatorManager(conf);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+
+    int count = 0;
+    while (call_count < GenesisValidators - 1)
```
Geod24: `atomicLoad`

source/agora/test/NetworkDiscovery.d
```diff
@@ -83,3 +87,65 @@ unittest
...
+    {
+        if (count++ > 30) // 3 secs
+            break;
+        Thread.sleep(100.msecs);
+    }
+
+    assert(call_count >= GenesisValidators - 1);
```
Geod24: `atomicLoad`

source/agora/test/NetworkDiscovery.d
```diff
@@ -83,3 +87,65 @@ unittest
...
+        {
+            if (impersonator_node != NodePair.init && conf.validator.enabled)
+                this.addNewNode!TestValidatorNode(conf, file, line);
+            else if (conf.validator.enabled)
+            {
+                this.addNewNode!ImpersonatorValidator(conf, file, line);
+                impersonator_node = this.nodes[$ - 1];
+            }
+            else
+                this.addNewNode!TestFullNode(conf, file, line);
```
Geod24:
```suggestion
            if (impersonator_node == NodePair.init && conf.validator.enabled)
            {
                this.addNewNode!ImpersonatorValidator(conf, file, line);
                impersonator_node = this.nodes[$ - 1];
            }
            else
                super.createNewNode(conf, file, line);
```

source/agora/test/NetworkDiscovery.d
```diff
@@ -83,3 +87,65 @@ unittest
...
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+
+    int count = 0;
+    while (call_count < GenesisValidators - 1)
+    {
+        if (count++ > 30) // 3 secs
+            break;
+        Thread.sleep(100.msecs);
+    }
```
Geod24: Aren't you just sleeping 3s here ?

source/agora/test/Quorum.d
```diff
@@ -53,7 +54,7 @@ unittest
     // generate 18 blocks, 1 short of the enrollments expiring.
     network.generateBlocks(Height(GenesisValidatorCycle - 2));
 
-    const keys = nodes.map!(node => node.getPublicKey())
+    const keys = nodes.map!(node => node.getPublicKey(PublicKey.init).key)
```
Geod24: Maybe a default value on the API would be reasonable here ?

source/agora/node/Validator.d
omerfirmak: Well, there is nothing keeping someone from calling this endpoint with `PublicKey.init`.

source/agora/test/NetworkDiscovery.d
omerfirmak: How? `call_count` should eventually reach `GenesisValidators - 1` and terminate the loop.

source/agora/test/Quorum.d
omerfirmak: LocalRest does not like default values. We could do it in `NetworkClient` but doesnt help the tests I guess.

Geod24: Oh we should probably have it implemented in LocalRest.

Geod24: But actually, it shouldn't need special handling if we keep a reference to the interface, does it ?

source/agora/test/NetworkDiscovery.d
Geod24: Brain was disconnected, ignore.

source/agora/test/Quorum.d
omerfirmak: Seems like it, see the last commit.

source/agora/test/Base.d
```diff
@@ -521,6 +521,11 @@ public struct NodePair
     /// used in the Nomination protocol.
     private shared(TimePoint)* cur_time;
 
+    ///
+    public TestAPI api;
+
+    alias api this;
+
```
Geod24, line 557: Client already implicitly converts to `TestAPI`.

Geod24 (changes requested):

source/agora/test/Base.d
omerfirmak: >  But actually, it shouldn't need special handling if we keep a reference to the interface, does it ?

Wasn't this what you meant here?

Geod24 (approved):

Geod24:
Can you rebase ?

omerfirmak:
> Can you rebase ?

Rebased

PR #1899, opened 2021-04-08
Revert BanManager minimum failed request from 100 to 1000

Geod24:
This seems to make the integration test fails more often.


linked0:
How about decreasing the value of `retry_delay` to 1000 (msecs) in every config.yaml file?

Geod24:
No, the ban manager is not the problem, as it doesn't ban validators anymore.

Issue #1552, opened 2021-01-26
When running on the docker, no blocks above height 19 are created.


MichaelKim20:
I tested in the following environment.
The height of the block is not greater than 19. 
I think this phenomenon may be related to the `validator_cycle`.
I attached the log of Agora nodes as a result.

### **Basic Infomation**
**Agora**
_git address_
https://github.com/MichaelKim20/agora
_branch_
localtest

**Stoa**
_git address_
https://github.com/MichaelKim20/stoa
_branch_
test-101

**BOA-SDK-TS**
_git address_
https://github.com/MichaelKim20/boa-sdk-ts
_branch_
localtest5

### **Clone & Install**
git clone https://github.com/MichaelKim20/agora
cd agora
git checkout localtest

git submodule update --init
docker build . -t agora

git clone https://github.com/MichaelKim20/stoa
cd stoa
git checkout test-101
npm install

git clone https://github.com/MichaelKim20/boa-sdk-ts
cd boa-dk-ts
git checkout localtest5
npm install

### **Test**
**Run Agora** 
cd localtest
./drm.sh
compose up

**Run Stoa**
npm start -- -c conf/config.yaml

**Run Stoa**
npm run sendtx

### Result
This is a log of Agora
https://gist.github.com/MichaelKim20/f52ad19dbc4cb16809c51a0e963db6cb

kchulj:
Based on the log, the nodes have already retrieved block 19. But why do they try again? It fails after 25 attempts.

kchulj:
The bug is now reproducible in live production.
What could be possible causes for this? This happens after the enrollments have been accepted.

AndrejMitrovic:
Can we get the logs of all the nodes from start up to the point they get stuck?

AndrejMitrovic:
I do think I find a bug in the block retrieval though, but I'm not sure if this is the cause of the stalls. I'll submit a PR soon.

kchulj:
> Can we get the logs of all the nodes from start up to the point they get stuck?

[Here](https://gist.github.com/kchulj/d96f56f697c0d0f35d67a6d4197cec9c) is a quick overview of the logs.
From the start to this point, there is nothing interesting. I will check again whether the logs look similar for all the nodes (I think they do).

AndrejMitrovic:
[...]

PR #1905, opened 2021-04-08
Block retrieval fixup

AndrejMitrovic:
This is just an optimization bugfix, but actually it uncovered a whole host of issues. The biggest one is that there are numerous methods in the Ledger which return status codes (`bool` return), but which also **throw**.

That means these exceptions can be accidentally caught and swallowed. I attempted marking some nothrow, but there's a very large chain of calls that needs to be marked nothrow to make this work. So we'll have to do this in another PR.

The commit messages explain in more detail.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -875,7 +875,7 @@ public class NetworkManager
 
         node_pairs.sort!((a, b) => a.height > b.height);
 
-        LNextNode: foreach (pair; node_pairs) try
+        LNextNode: foreach (pair; node_pairs)
         {
             if (height > pair.height)
                 continue;  // this node does not have newer blocks than us
@@ -904,18 +904,22 @@ public class NetworkManager
                 log.info("Received blocks [{}..{}]",
                     blocks[0].header.height, blocks[$ - 1].header.height);
 
-                // one or more blocks were rejected, stop retrieval from node
-                if (!onReceivedBlocks(blocks, preimages))
-                    continue LNextNode;
+                try
+                {
+                    // one or more blocks were rejected, stop retrieval from node
+                    if (!onReceivedBlocks(blocks, preimages))
+                        continue LNextNode;
+                }
+                catch (Exception ex)
+                {
+                    // @BUG: Ledger routines should be marked nothrow,
+                    // or else storage issues should be handled differently.
+                    log.error("Error in onReceivedBlocks(): {}", ex);
+                }
 
                 height += blocks.length;
             }
             while (height < pair.height);
-        }
-        catch (Exception ex)
-        {
-            log.error("Couldn't retrieve blocks: {}. Will try again later..",
-                ex.msg);
         }
     }
 
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -850,8 +850,8 @@ public class NetworkManager
     ***************************************************************************/
 
     public void getBlocksFrom (Height height,
-        scope bool delegate(const(Block)[],
-        const(PreImageInfo)[]) @safe onReceivedBlocks) nothrow
+        scope Height delegate(const(Block)[], const(PreImageInfo)[])
+            @safe onReceivedBlocks) nothrow
     {
         struct Pair { Height height; NetworkClient client; }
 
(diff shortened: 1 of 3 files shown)


omerfirmak (approved):

Geod24:
> The biggest one is that there are numerous methods in the Ledger which return status codes (`bool` return), but which also **throw**.

Yeah, and I'm starting to think the status code approach was a bit too eager. We had way too many places where return codes and exceptions alike were ignored, but at least with Exceptions, we will see it.

AndrejMitrovic:
>  but at least with Exceptions, we will see it.

Unless they're swallowed. :)

Some kind of `@nodiscard` would really help. Wasn't it proposed recently?

source/agora/node/FullNode.d
```diff
@@ -360,21 +360,23 @@ public class FullNode : API
...
     ***************************************************************************/
 
-    public bool addBlocks (const(Block)[] blocks, const(PreImageInfo)[] preimages)
+    public Height addBlocks (const(Block)[] blocks, const(PreImageInfo)[] preimages)
         @safe
     {
+        Height last_height;
         foreach (block; blocks)
         {
             if(!this.ledger.enrollment_manager.addPreimages(preimages))
```
Geod24: This line also looks super buggy, especially now with preimages being potentially disjoint between cycles.
If we get blocks [18, 19, 20, 21] and pre-images for 21, the EM would reject them ATM.

source/agora/network/Manager.d
```diff
@@ -905,18 +905,21 @@ public class NetworkManager
...
                 // one or more blocks were rejected, stop retrieval from node
-                if (!onReceivedBlocks(blocks, preimages))
-                    continue LNextNode;
-
-                block_height += blocks.length;
+                try
+                {
+                    block_height = onReceivedBlocks(blocks, preimages);
+                    if (block_height == 0)
+                        continue LNextNode;
```
Geod24: The special casing on `0` seems odd. Why not return the number of blocks added ?

AndrejMitrovic: Right I seem to have missed this. Will fix.

AndrejMitrovic: In fact I can get rid of the do/while loop now too, because I made the function nothrow.

source/agora/node/FullNode.d
AndrejMitrovic: Oh then I think we should add a `rollback()` call here.

AndrejMitrovic: Now that I look at this function more it looks really sketchy. `addPreimages` is called multiple times...

AndrejMitrovic: I think we need to rework the catching up phase w.r.t. preimages. I'll file a separate issue.

AndrejMitrovic: Scratch that, I'll just fix the function. I've also noticed that this function will return `false` if we receive a duplicate preimage (same height), which means that if we receive a newer preimage and *then* do a `getBlocksFrom + getPreimages` the `addBlocks` function will terminate early and the blocks would be rejected.

AndrejMitrovic:
I think this should fix it now. I think we could also optimize the last commit (maybe in another PR?). We don't need to keep trying to re-add preimages which were successfully added. We could filter them by their enrollment start height.

AndrejMitrovic:
@kchulj I am placing a bet of one beer (any beer of your choice) that this will fix https://github.com/bosagora/agora/issues/1552 🍺

kchulj:
> @kchulj I am placing a bet of one beer (any beer of your choice) that this will fix #1552 🍺

Related to #1083 ?

AndrejMitrovic:
> Related to #1083 ?

Somewhat. But there were multiple bugs found in the catchup code here. The net effect is that blocks would be silently rejected. That might explain why there was nothing interesting in the logs in #1552.

source/agora/node/FullNode.d
Geod24: IMO it's still broken ATM, because upon receiving PI for 21, the Ledger (at 17) will reject them and so adding block 18 will not work. When we move from distance to height, we will need to accept PI beyond the enrolled cycle (or adjust them to be at the end of the cycle, but there's no downside in accepting them beyond the cycle within a reasonable margin).

Geod24:
> @kchulj I am placing a bet of one beer (any beer of your choice) that this will fix #1552 🍺

I'll take the bet if Ku doesn't :P
I don't think #1552 is related to catch up, although I would love to be wrong.

PR #1916, opened 2021-04-11
CI: Test with LDC 1.26.0-beta1

Geod24:


Geod24:
`ldc-master` is working now so we don't need this.

PR #1924, opened 2021-04-12
Registry: Add stats

omerfirmak:


Geod24 (approved):

Geod24:
> #11 164.4 source/agora/registry/Server.d(20,8): Error: module `Registry` is in file 'agora/stats/Registry.d' which cannot be read

Need a small change to `dub.json` tho

source/agora/registry/main.d
```diff
@@ -65,6 +70,10 @@ private int main (string[] args)
         return -1;
     }
 
+    StatsServer stats_server;
+    if (cmdline_args.stats_port != 0)
```
ferencdg, line 74: once I got scolded for using port 0 for disabling stats server(there used to be a check in the StatsServer that if 0 is passed to the constructor, then it will not start up)
https://github.com/bosagora/agora/pull/1249#discussion_r502983154

ferencdg (approved):

source/agora/registry/main.d
omerfirmak: Well, I don't see why anyone would want to bind a server socket to a random port so I think we are fine:)

Geod24: > Well, I don't see why anyone would want to bind a server socket to a random port so I think we are fine:)

"I can't think of any way someone would use that" is a sure way to be surprise :)
I remember a system where that was done, as the number of nodes to spawn wasn't known (and varied during the system lifetime), and there was simply a way for consumers to discover ports. So it's not *so odd*.

PR #1926, opened 2021-04-12
 Flash: Authenticate error messages 

omerfirmak:
While the payee node deobfuscates the error message it can
detect which node the message originates from. Ignore the error if the
origin node of the message is not a peer of the failing channel.

### DIFF
--- a/source/agora/flash/Node.d
+++ b/source/agora/flash/Node.d
@@ -884,20 +884,31 @@ public abstract class FlashNode : ControlFlashAPI
 
             auto chans = (*path).map!(hop => hop.chan_id);
             OnionError deobfuscated = err;
-            foreach (secret; shared_secrets)
+            size_t failing_hop_idx = shared_secrets.length - 1;
+            foreach (idx, secret; shared_secrets)
             {
                 if (chans.canFind(deobfuscated.chan_id))
+                {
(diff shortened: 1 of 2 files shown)


source/agora/flash/Node.d
```diff
@@ -884,20 +884,29 @@ public abstract class FlashNode : ControlFlashAPI
...
-                    deobfuscated);
-                this.payment_errors[deobfuscated.payment_hash] ~= deobfuscated;
-                this.dump();
-            }
+            if (failing_hop_idx == size_t.max)
+                return;
+
+            const failing_node_pk = (*path)[failing_hop_idx].pub_key;
+            const failing_chan = this.known_channels[deobfuscated.chan_id];
+            if (failing_chan.funder_pk != failing_node_pk &&
```
AndrejMitrovic, line 904: Some `//` comments here about what the code is doing would be great.

AndrejMitrovic (changes requested):
Just a comment.

source/agora/flash/Node.d
omerfirmak: Done, made a small change to the loop above too.

AndrejMitrovic (approved):

AndrejMitrovic:
Also LGTM.

PR #1935, opened 2021-04-14
PreImage: Fix a unittest not to assume ValidatorCycle's value

Geod24:
We want to change it from 1008 to 20 in unittests,
but this unittest was making the assumption it was 1008.


hewison-chris (commented):
LGTM

PR #1936, opened 2021-04-14
Tests: Remove duplicated createEnrollment functions, use EM.makeEnrollment

Geod24:
Simply removing redundant and (almost) dead code.


hewison-chris (commented):
LGTM

PR #1937, opened 2021-04-14
Fix #1646 - Separate TxBuilder in its own module

### LINKED ISSUE #1646: Separate TxBuilder into its own module
`TxBuilder` started as a testing utility, and was put in `agora.utils.Test`, but it evolved into a really usable tool to make transaction, as seen by its usage in `Faucet`. It should be separated in its own module to reflect this, one that does not rely on the well-known keypairs too.

Geod24:
Fairly simple


Geod24:
Will self merge since this is non-controversial and will update Faucet

PR #1938, opened 2021-04-14
Use the human-readable part for the prefix of addresses

### LINKED ISSUE #1354: We should have a different prefix for addresses used for testing and live networks
Perhaps we should adopt a scheme like all addresses used when running in test mode have prefix T.
Maybe we could generate vanity addresses for our tests to use:-
TV1, TV2... for test validator nodes 
TF1, TF2... for test full nodes
TC for commons budget address
TG for Genesis (Foundation address)

Bitcoin addresses for example:- https://en.bitcoin.it/wiki/List_of_address_prefixes

### LINKED ISSUE #1878: Give meaningful names to well-known keypairs
Since the move to Bech32, our well-known addresses no longer match an easy to recognize pattern.
We need to come up with a new pattern to use (`node` will not work, `o` is forbidden in bech32), change the vanity program, and mine them.

linked0:
Before applying `Bech32` encoding, all the well-known addresses contain special prefixes used to identify them easily. But after applying that, we lost the prefixes because the `Bech32` makes different names with the same `Point` value. So we are using the `human-readable` part for the prefix of addresses.

Fixes #1878
Fixes #1354

If there are any suggestions about modifying the prefixes of well-known addresses, like removing the `gd` part of each prefix, please let me know.


linked0:
Need to check errors of CI builds.

Geod24:
> #11 169.5 ldc2 failed with exit code -9. This may indicate that the process has run out of memory.

Yep, I think it doesn't like the new WKKeyData.

source/agora/consensus/Quorum.d
```diff
@@ -191,16 +191,18 @@ unittest
...
         hashFull(2), QuorumParams.init, 7);
     verifyQuorumsSanity(quorums_2);
     verifyQuorumsIntersect(quorums_2);
     test!"=="(countNodeInclusions(quorums_2, keys),
-        [8, 7, 8, 11, 8, 5, 8, 6, 6, 9, 4, 9, 3, 7, 6, 7]);
+        [8, 12, 6, 6, 6, 6, 9, 6, 4, 6, 8, 5, 7, 7, 7, 9]);
 }
 
+import std.stdio;
+
```
Geod24: 👀

kchulj:
Hmm is it affecting faucet too?

source/agora/crypto/Key.d
```diff
@@ -121,35 +147,51 @@ public struct PublicKey
...
+        nothrow @safe @nogc
     {
         this.data = args;
+
+        auto wkdata = WKKeyData.Keys[this.data];
+        if (wkdata != KeyData.init)
+            this.hrp = wkdata.hrp;
+        else
+            this.hrp = hrp;
     }
```
Geod24: This is not a sound interface. There are two main cases for the second parameter:
- The user provides it, expecting it to be respected;
- The user doesn't provide it, expects a sane default;

The former use case is not covered: If the user provides it, the constructor might still overwrite it.
A much better interface here is to expose two overloads, one that doesn't take a second parameter and do the HRP lookup, and the other that does and just use that.

source/agora/crypto/Key.d
```diff
@@ -121,35 +147,51 @@ public struct PublicKey
...
 
     /// Uses Stellar's representation instead of hex
     public string toString () const @trusted nothrow
     {
-        ubyte[VersionWidth + PublicKey.sizeof] bin;
+        ubyte[VersionWidth + PublicKey.data.sizeof] bin;
```
Geod24: Hum that means that `PublicKey` is not a value type anymore, that's a big downside

source/agora/crypto/Key.d
```diff
@@ -210,12 +252,16 @@ public struct PublicKey
...
-            auto exp_hash = Hash("0xdb3b785e31c05c9383f498aa12a5c054fcb6f99b2" ~
-                "0bf7ff25fdb24d5fe6d1bb5768f1821de9d5d31faa7ebafc79837ba32dc4" ~
-                "774bffea9918411d0c4c8ac403c");
+            auto exp_hash = Hash("0x39007e97ccdc88c27bb8bded39016d95dfee4713a" ~
+                "c08e9be6025a33e64179effa8b827a5115f2a570f0ac16e9fa9f3c4e51f4" ~
+                "6642a1c6b40a2141c86418aea11");
+
+            // scope(failure) assert(0);
+            // import std.stdio;
+            // writeln("hash: ", hash);
```
Geod24: :eyes:

source/agora/crypto/Key.d
```diff
@@ -210,12 +252,16 @@ public struct PublicKey
...
-        () nothrow @safe @nogc
+        () nothrow @safe
         {
             auto hash = hashFull(pubkey);
-            auto exp_hash = Hash("0xdb3b785e31c05c9383f498aa12a5c054fcb6f99b2" ~
-                "0bf7ff25fdb24d5fe6d1bb5768f1821de9d5d31faa7ebafc79837ba32dc4" ~
-                "774bffea9918411d0c4c8ac403c");
+            auto exp_hash = Hash("0x39007e97ccdc88c27bb8bded39016d95dfee4713a" ~
+                "c08e9be6025a33e64179effa8b827a5115f2a570f0ac16e9fa9f3c4e51f4" ~
+                "6642a1c6b40a2141c86418aea11");
```
Geod24: Since the HRP can be freely modified, it shouldn't affect the hash, otherwise a node could just mount attacks by using different HRP for different nodes.

linked0:
> > #11 169.5 ldc2 failed with exit code -9. This may indicate that the process has run out of memory.
> 
> Yep, I think it doesn't like the new WKKeyData.

Hmm, I think so. Need to find another way or a workaround.

source/agora/utils/WellKnownKeyData.d
```diff
@@ -0,0 +1,7041 @@
...
+    Copyright:
+        Copyright (c) 2019-2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.utils.WellKnownKeyData;
```
Geod24: Wait why a new module ? We already have one for this ?

Geod24 (changes requested):
Shouldn't affect the change, and I don't understand the point of the module duplication.

linked0:
I'm retrying other ways to solve the issues reviewers raised on this PR. I will request a whiteboard after retrying other ways and reorganizing my thoughts.

source/agora/crypto/Key.d
linked0: I moved the logic into the `toString` function. Ty.

source/agora/crypto/Key.d
linked0: It's reverted.

source/agora/crypto/Key.d
linked0: Reverted

source/agora/utils/WellKnownKeyData.d
linked0: Actually we need this because `agora.crypto.Key` module can not refer to `agora.utils.WellKnownKeys`. I created the module `agora.utils.WellKnownHRPs' module and made the module more compact anyway.

linked0:
Ready for review.
The `Integration` test will definitely fail because the newly added file, `WellKnownHRPs.d` in Agora is not added to `faucet` automatically. If this PR is merged with the `WellKnownHRPs.d`, I will apply the change into `faucet` with the support of @kchulj.

linked0:
Ping

source/agora/crypto/Key.d
```diff
@@ -20,6 +20,7 @@ import agora.crypto.ECC;
 import agora.crypto.Hash;
 import agora.crypto.Schnorr;
 import agora.serialization.Serializer;
+import agora.utils.WellKnownHRPs : getWellKnownHRP;
```
Geod24: No, we should absolutely not have a dependence between this code and `WellKnownHRPs`.
The well-known keys (and HRP) are test utilities only. They can depend on the production code, but the production code can NEVER depend on them.

linked0: I see what you mean. I will find another way soon. Thanks.
However, I only made human-readable parts for [a .. zz] in `WellKnownHRPs` because if I make all the HRPs for [a .. azz], it makes the testing time increase 40%. I don't know why for sure but only know that the very long AA for HRPs causes it. If you have any ideas or suggestions for that, please tell me about that. Anyway, if I found a good way for implementing this feature, that problem would be solved by itself.

linked0:
Ready for review.
I have managed for the `Key` module to refer `WellKnownHRPs` only in the unit tests.

linked0:
I'd like to have a whiteboard about this PR after Sprint Planning.

Geod24:
As discussed with @linked0 , using HRP proved not to be a good idea, so this is superseded by #2008

PR #1941, opened 2021-04-14
Flash: Make sure Payee sets enough lock_time

omerfirmak:
While building the onion, payee will try to accompany
all the htlc_delta values on route. If they don't result in
enough lock time, should be increased according to the settle_time.

We ignore settle_time configurations on the path, nodes should be expected
to set appropriate htlc_delta.


AndrejMitrovic (approved):
LGTM. Any new tests to add?

omerfirmak:
> LGTM. Any new tests to add?

I am not sure if this creates new test cases.

AndrejMitrovic:
> I am not sure if this creates new test cases.

Alright, then I'll merge it.

PR #1951, opened 2021-04-15
Add `{EM,ValidatorSet}.countActive` to respect the given height

Geod24:
```
The behavior of `getValidatorCount` could be surprising to the reader:
it was getting the number of potential validators at the height after the provided one.
The documentation and the parameter name didn't advertise this fact properly,
and while most usages were correct, there are a few red flags.

Since it is widely used with both "current block" and "next block" semantics,
the first step is to allow one to use both separately,
by making `ValidatorSet` return the correct value,
and later on remove `EnrollmentManager.getValidatorCount`.
```

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -994,11 +994,22 @@ public class EnrollmentManager
             `height`.
             Returns 0 in case of error.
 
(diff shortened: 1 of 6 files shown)


hewison-chris (approved):
So there were three validator count functions. I removed one last week.

PR #1954, opened 2021-04-16
EPool.getEnrollments: Do not reuse memory, use new GC allocations

Geod24:
This pattern was premature optimization, which we all know to be the root of all evil.
The memory wasn't actually reused much, since `ConsensusData` was always cleared because of `out`.
It however did complicate the API, and made the code not re-entrant, which is now fixed.


linked0 (approved):
LGTM

Issue #1917, opened 2021-04-12
Limiting the rate with which transactions get relayed 


ferencdg:
this is to prevent network flooding with low fee transactions

details:

https://github.com/bosagora/agora/issues/236#issuecomment-813833479

part of #236

PR #1955, opened 2021-04-16
Various commits chopped off from PR 1881

ferencdg:
PR 1881 stuck in a permanent review state, so I chopped off some commits that I need for other PRs. Comments for those commits are on the PR 1881 page. Please merge, so I can continue creating other PRs, otherwise I am **blocked**.

### DIFF
--- a/source/agora/common/Amount.d
+++ b/source/agora/common/Amount.d
@@ -255,16 +255,37 @@ public struct Amount
         Take a percentage of the `value`
 
         Params:
-            percentage = Requested percentage
+            pct = Requested percentage
 
     ***************************************************************************/
 
     pragma(inline, true)
-    public void percentage (ubyte percentage)
+    public bool percentage (ushort pct)
     {
-        assert(percentage <= 100);
-        this.value *= percentage;
+        bool overflow;
+        this.value = mulu(this.value, pct, overflow);
+        // deliberately not checking using isValid(), because after the divsion by 100,
+        // the value can become valid
+        if (overflow)
+            return false;
         this.value /= 100;
+        return this.isValid();
+    }
+
+    /***************************************************************************
+
+        Take a percentage of the `value` and assert if the result is invalid
+
+        Params:
+            pct = Requested percentage
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public Amount mustPercentage (ushort pct)
+    {
+        this.percentage(pct) || assert(0);
+        return this;
     }
 
     /***************************************************************************
@@ -496,12 +517,21 @@ unittest
 unittest
 {
     Amount amt = Amount(100);
-    amt.percentage(33);
+    bool is_valid = amt.percentage(33);
     assert(amt == Amount(33));
-    amt.percentage(33);
+    assert(is_valid);
+
+    is_valid = amt.percentage(33);
     assert(amt == Amount(10));
-    amt.percentage(10);
+    assert(is_valid);
+
+    is_valid = amt.percentage(10);
     assert(amt == Amount(1));
+    assert(is_valid);
+
+    amt = Amount(ulong.max);
+    is_valid = amt.percentage(2);
+    assert(is_valid);
 }
 
 unittest
--- a/source/agora/consensus/Fee.d
+++ b/source/agora/consensus/Fee.d
@@ -403,6 +403,34 @@ public class FeeManager
 
     /***************************************************************************
 
+        Calculate the transaction fee and adjust the fee based on the
+        transaction's size measured in bytes.
+
+        The bigger the transaction size is, the smaller the the fee becomes.
+
+        Params:
+            tx = transaction for which we want to calculate the adjusted fee
+            peekUTXO = UTXO finder
+            tot_fee = adjusted fee
+
+        Returns: string describing the error, if an error happened, null otherwise
+
+    ***************************************************************************/
+
+    public string getAdjustedTXFee (in Transaction tx, scope UTXOFinder peekUTXO,
+        out Amount tot_fee) nothrow @safe
+    {
+        Amount tot_data_fee;
+        const(Transaction)[1] txs = [tx];
+
+        auto res = this.getTXSetFees(txs, peekUTXO, tot_fee, tot_data_fee);
+        tot_fee.div(tx.sizeInBytes());
+
+        return res;
+    }
+
+    /***************************************************************************
+
         Calculate total fees of a Transaction set
 
         Params:
@@ -411,9 +439,11 @@ public class FeeManager
             tot_fee = Total fee (incl. data fees)
             tot_data_fee = Total data fee
 
+        Returns: string describing the error, if an error happened, null otherwise
+
     ***************************************************************************/
 
-    public void getTXSetFees (in Transaction[] tx_set,
+    public string getTXSetFees (in Transaction[] tx_set,
         scope UTXOFinder peekUTXO, ref Amount tot_fee, ref Amount tot_data_fee)
         nothrow @safe
     {
@@ -427,17 +457,19 @@ public class FeeManager
             foreach (input; tx.inputs)
             {
                 UTXO utxo;
-                assert(peekUTXO(input.utxo, utxo));
+                if (!peekUTXO(input.utxo, utxo))
+                    return "Unable to find input for utxo";
                 tot_in.mustAdd(utxo.output.value);
             }
 
-            assert(tx.getSumOutput(tot_out), "Not validated block in" ~
-                "getTXSetFees");
+            if (!tx.getSumOutput(tot_out))
+                return "Transaction output value is invalid";
             // sum(inputs) - sum(outputs)
             tot_in.mustSub(tot_out);
             tot_fee.mustAdd(tot_in);
             tot_data_fee.mustAdd(this.getDataFee(tx.payload.data.length));
         }
+        return null;
     }
 
     /// For unittest
--- a/source/agora/consensus/data/DataPayload.d
+++ b/source/agora/consensus/data/DataPayload.d
@@ -26,6 +26,11 @@ public struct DataPayload
     /// The byte array of transaction data
     public const(ubyte)[] data;
 
+    /// The size of the data DataPayload object
+    public ulong sizeInBytes () const nothrow pure @safe @nogc
+    {
+        return this.data.length * this.data[0].sizeof;
+    }
     /***************************************************************************
 
         Create a DataPayload from binary data
--- a/source/agora/consensus/data/Transaction.d
+++ b/source/agora/consensus/data/Transaction.d
@@ -60,6 +60,17 @@ public struct Transaction
     /// The data to store
     public DataPayload payload;
 
+    /// The size of the Transaction object
+    public ulong sizeInBytes () const nothrow pure @safe @nogc
+    {
+        ulong size = this.type.sizeof + this.payload.sizeInBytes();
+        foreach (const ref input; this.inputs)
+            size += input.sizeInBytes();
+        foreach (const ref output; this.outputs)
+            size += output.sizeInBytes();
+        return size;
+    }
+
     /// This transaction may only be included in a block with `height >= lock_height`.
     /// Note that another tx with a lower lock time could double-spend this tx.
     public Height lock_height = Height(0);
@@ -122,6 +133,12 @@ public struct Output
     /// The lock condition for this Output
     public Lock lock;
 
+    /// The size of the Output object
+    public ulong sizeInBytes () const nothrow pure @safe @nogc
+    {
+        return this.value.sizeof + this.lock.sizeInBytes();
+    }
+
     /// Ctor
     public this (Amount value, inout(Lock) lock) inout pure nothrow @trusted
     {
@@ -182,6 +199,12 @@ public struct Input
     /// than the block height at which the spending transaction wants to be
     /// included in the block. Use for implementing relative time locks.
     public uint unlock_age = 0;
+
+    /// The size of the Input object
+    public ulong sizeInBytes () const nothrow pure @safe @nogc
+    {
+        return this.unlock.sizeInBytes() + this.unlock_age.sizeof + this.utxo.sizeof;
+    }
 
     /// Simple ctor
     public this (in Hash utxo_, Unlock unlock = Unlock.init, uint unlock_age = 0)
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -669,8 +669,8 @@ public class Ledger
         in uint[] missing_validators) nothrow @safe
     {
         Amount tot_fee, tot_data_fee;
-        this.fee_man.getTXSetFees(tx_set, &this.utxo_set.peekUTXO, tot_fee,
-            tot_data_fee);
+        if (auto fee_res = this.fee_man.getTXSetFees(tx_set, &this.utxo_set.peekUTXO, tot_fee, tot_data_fee))
+            assert(0, fee_res);
         return this.getCoinbaseTX(tot_fee, tot_data_fee, missing_validators);
     }
 
--- a/source/agora/script/Lock.d
+++ b/source/agora/script/Lock.d
@@ -66,6 +66,12 @@ public struct Lock
             break;
         }
     }
+
+    /// The size of the Lock object
+    public ulong sizeInBytes () const nothrow pure @safe @nogc
+    {
+        return this.type.sizeof + this.bytes.length * this.bytes[0].sizeof;
+    }
 }
 
 unittest
@@ -88,6 +94,12 @@ public struct Unlock
     {
         sink("0x");
         UbyteHexString(this.bytes).toString(sink);
+    }
+
+    /// The size of the Unlock object
+    public ulong sizeInBytes () const nothrow pure @safe @nogc
+    {
+        return this.bytes.length * this.bytes[0].sizeof;
     }
 }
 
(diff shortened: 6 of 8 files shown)


source/agora/common/Amount.d
```diff
@@ -260,11 +260,27 @@ public struct Amount
     ***************************************************************************/
 
     pragma(inline, true)
-    public void percentage (ubyte percentage)
+    public bool percentage (ushort pct)
```
Geod24, line 263: Need to update the UT:
```
source/agora/common/Amount.d:499:    amt.percentage(33);
source/agora/common/Amount.d:501:    amt.percentage(33);
source/agora/common/Amount.d:503:    amt.percentage(10);
```

Also commit message: "percentage", not "percantage".

source/agora/common/Amount.d
```diff
@@ -260,11 +260,27 @@ public struct Amount
...
     {
-        assert(percentage <= 100);
-        this.value *= percentage;
+        this.value *= pct;
```
Geod24: There should be a validity check to avoid overflow which would turn an invalid value into a valid one.

source/agora/common/Amount.d
Geod24, line 263: Why did you change the parameter name ?

source/agora/common/Amount.d
```diff
@@ -260,11 +260,27 @@ public struct Amount
...
         this.value /= 100;
+        return isValid();
+    }
+
+    /***************************************************************************
+
+        Take a percentage of the `value` and assert if the result is invalid
+
+        Params:
+            percentage = Requested percentage
```
Geod24: Mismatch with the parameter name

source/agora/utils/TxBuilder.d
```diff
@@ -424,6 +424,25 @@ public struct TxBuilder
...
+    ***************************************************************************/
+
+    public ref typeof(this) deductRemaining ()
+        return
+    {
+        this.leftover.value = Amount(0);
+
+        return this;
+    }
+
```
Geod24: Since fee seems to be the only use case, I think it'd be better to have something that reference it, such as `addFee`. Also this function is lacking attributes.

source/agora/common/Amount.d
```diff
@@ -47,6 +47,12 @@ public struct Amount
     /// Internal data storage
     private ulong value;
 
+    /// The size of the Amount object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
+    {
+        return value.sizeof;
+    }
+
```
Geod24: Just call sizeof from the caller ?

source/agora/consensus/Fee.d
```diff
@@ -427,17 +456,19 @@ public class FeeManager
...
-                assert(peekUTXO(input.utxo, utxo));
+                if (!peekUTXO(input.utxo, utxo))
+                    return "unable to find input for utxo";
                 tot_in.mustAdd(utxo.output.value);
             }
 
-            assert(tx.getSumOutput(tot_out), "Not validated block in" ~
-                "getTXSetFees");
+            if (!tx.getSumOutput(tot_out))
+                return "Not validated block in getTXSetFees";
```
Geod24: That message doesn't make much sense, think we can reword it ?

source/agora/node/Ledger.d
```diff
@@ -669,8 +669,9 @@ public class Ledger
         in uint[] missing_validators) nothrow @safe
     {
         Amount tot_fee, tot_data_fee;
-        this.fee_man.getTXSetFees(tx_set, &this.utxo_set.peekUTXO, tot_fee,
-            tot_data_fee);
+        auto fee_res = this.fee_man.getTXSetFees(tx_set, &this.utxo_set.peekUTXO, tot_fee, tot_data_fee);
+        if (fee_res != null)
+            assert(0, fee_res);
```
```D
if (auto fee_res = ....)
    assert(0, fee_res);
```

source/agora/consensus/Fee.d
```diff
@@ -427,17 +456,19 @@ public class FeeManager
             foreach (input; tx.inputs)
             {
                 UTXO utxo;
-                assert(peekUTXO(input.utxo, utxo));
+                if (!peekUTXO(input.utxo, utxo))
+                    return "unable to find input for utxo";
```
Geod24: Other messages are capitalized.

source/agora/consensus/Fee.d
```diff
@@ -401,6 +401,33 @@ public class FeeManager
...
+        Returns: string describing the error, if an error happened, null otherwise
+
+    ***************************************************************************/
+
+    public string getAdjustedTXFee (in Transaction tx, scope UTXOFinder peekUTXO,
+        out Amount tot_fee) nothrow @safe
+    {
+        Amount tot_data_fee;
+
+        auto res = getTXSetFees([tx], peekUTXO, tot_fee, tot_data_fee);
```
Geod24: `getTXSetFees` should really accept a range, or you could pass `(&tx)[0 .. 1]` to avoid the needless allocation.
But let's fix this later as it might create other issues with scoping and `@safe`.

source/agora/common/Amount.d
linked0, line 263: The return value is only used in `mustPercentage`. Why don't you call the `isValid` with `assert`?

Geod24 (changes requested):
The `getSizeInBytes` doesn't seem very documented and is not tested. Since it will dictate the fees in the protocol, it'd be nice to know the expectation. Additionally, I think we should limit it to the Transaction module to avoid contaminating all data structures.

source/agora/common/Amount.d
```diff
@@ -260,11 +260,27 @@ public struct Amount
...
+        Take a percentage of the `value` and assert if the result is invalid
+
+        Params:
+            percentage = Requested percentage
+
+    ***************************************************************************/
+
+    pragma(inline, true)
+    public Amount mustPercentage (ushort pct)
```
linked0, line 285: Another name like `getPercentage would be better.

source/agora/consensus/data/Transaction.d
```diff
@@ -60,6 +60,17 @@ public struct Transaction
     /// The data to store
     public DataPayload payload;
 
+    /// The size of the Transaction object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
+    {
+        ulong size = type.sizeof + payload.sizeInBytes();
+        foreach (const ref input; inputs)
```
linked0: Suggested change.
`this.inputs`

source/agora/consensus/data/Transaction.d
```diff
@@ -60,6 +60,17 @@ public struct Transaction
...
+            size += input.sizeInBytes();
+        foreach (const ref output; outputs)
```
linked0: Suggested change.
`this.outputs`

source/agora/consensus/Fee.d
ferencdg: I can rename it, it came from a previous commit though

source/agora/consensus/data/Transaction.d
```diff
@@ -122,6 +133,12 @@ public struct Output
     /// The lock condition for this Output
     public Lock lock;
 
+    /// The size of the Output object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
+    {
+        return value.sizeInBytes() + lock.sizeInBytes();
```
linked0: Shouldn't we add `this` to the member variables?

source/agora/consensus/data/Transaction.d
```diff
@@ -183,6 +200,12 @@ public struct Input
     /// included in the block. Use for implementing relative time locks.
     public uint unlock_age = 0;
 
+    /// The size of the Input object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
+    {
+        return unlock.sizeInBytes() + unlock_age.sizeof + utxo.sizeof;
```
linked0: Ditto

source/agora/script/Lock.d
```diff
@@ -45,13 +45,25 @@ public struct Lock
 
     /// May either be a Hash, or a sequence of opcodes
     public const(ubyte)[] bytes;
+
+    /// The size of the Lock object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
+    {
+        return type.sizeof + bytes.length * bytes[0].sizeof;
```
linked0: Ditto

source/agora/script/Lock.d
```diff
@@ -45,13 +45,25 @@ public struct Lock
...
 /// Contains a data tuple or a set of push opcodes
 public struct Unlock
 {
     /// May be: <signature>, <signature, key>, <push opcodes>
     public const(ubyte)[] bytes;
+
+    /// The size of the Unlock object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
+    {
+        return bytes.length * bytes[0].sizeof;
```
linked0: Ditto

source/agora/consensus/Fee.d
```diff
@@ -401,6 +401,33 @@ public class FeeManager
...
+            tx = transaction for which we want to calculate the adjusted fee
+            peekUTXO = UTXO finder
+            tot_fee = adjusted fee
+
+        Returns: string describing the error, if an error happened, null otherwise
+
+    ***************************************************************************/
+
+    public string getAdjustedTXFee (in Transaction tx, scope UTXOFinder peekUTXO,
+        out Amount tot_fee) nothrow @safe
```
linked0, line 421: Out of curiosity, what does the `tot` stand for?

linked0 (commented):
Good work and some comments.

source/agora/common/Amount.d
ferencdg: because now we have a method called 
```D
    pragma(inline, true)
    public Amount mustPercentage (ushort pct)
    {
        percentage(pct) || assert(0);
        return this;
    }
```

and this method calls the percentage method, so I had to change the percenatege parameter to pct to be consistent

source/agora/common/Amount.d
ferencdg: I wanted to be consisten with mustAdd/mustSub

source/agora/consensus/data/Transaction.d
ferencdg: kind of contraversial :) If we modify the the style guide that we have to prefix every class variable with this then I am happy to comply. VSCode uses different colors for class members vs local variables though.

source/agora/consensus/Fee.d
ferencdg: total -> same name used in getTXFee

source/agora/common/Amount.d
ferencdg: percentage and mustPercantage methods can be called by anyone and they behave differently:
percentage - returns false if invalid
mustPercantage - asserts if invalid

hence the `must` prefix

this is the same behaviour as with mustSub/sub and mustAdd/add

ferencdg:
> The `getSizeInBytes` doesn't seem very documented and is not tested. 

what else could I say?
```D
    /// The size of the Transaction object
    public ulong sizeInBytes() const nothrow pure @safe @nogc
```
> and is not tested. 

if someone adds a new field to the Transaction object, and forget to change the sizeInBytes method, then I will not be able to catch that with unit tests. If someone removes a field, and forget the change the sizeInBytes method, he will receive a compile error. I can add a unit test that calculates the size and then I canculate it on paper and compare... But I already have done that.


> Additionally, I think we should limit it to the Transaction module to avoid contaminating all data structures.

do you mean use `package` modifier on the sizeInBytes() method? I am going to use that method in Fee for example and that is not in the same package

source/agora/utils/TxBuilder.d
ferencdg: We already have an addFee method called: 'deduct', but tho use it you need to know how much value is left in the output. That makes using 'deduct' method inconvinient in some cases(like when using it with filters/maps...). I can call it deductRemaining, if that is okay, or I can remove both methods to addFee and addRemainingAsFee

source/agora/consensus/Fee.d
ferencdg: Just using a fixed size array with length 1 would avoid allocations(on the heap).

ferencdg:
thanks for the review @Geod24, @linked0 I fixed everything, and still have 1 pending question: https://github.com/bosagora/agora/pull/1955#issuecomment-820846559

if you are okay with my answer, then this could be merged

Geod24:
This triggers a (valid) assert:
```
#### FATAL ERROR: 
This node was started at source/agora/test/Flash.d:528
Full error: core.exception.AssertError@source/agora/common/Amount.d(341)
----------------
runtime.d:815 [0x563497768a4b]
runtime.d:774 [0x563497768273]
dmain2.d:292 [0x56349777bf00]
deh.d:46 [0x5634977b226f]
dwarfeh.d:332 [0x56349777cb02]
/home/runner/work/agora/agora/source/agora/test/Base.d:140 [0x563496f089b1]
exception.d:430 [0x56349776773f]
exception.d:595 [0x563497767c24]
/home/runner/work/agora/agora/source/agora/common/Amount.d:341 [0x563496e06413]
/home/runner/work/agora/agora/source/agora/consensus/Fee.d:329 [0x5634973bcfdc]
/opt/hostedtoolcache/dc/ldc2-1.25.0/x64/ldc2-1.25.0-linux-x86_64/bin/../import/std/algorithm/iteration.d:1026 [0x5634973bac9c]
/home/runner/work/agora/agora/source/agora/consensus/Fee.d:329 [0x563496fdc189]
/home/runner/work/agora/agora/source/agora/node/Ledger.d:620 [0x563496eb000e]
/home/runner/work/agora/agora/source/agora/node/Ledger.d:1173 [0x563496f0aef3]
/home/runner/work/agora/agora/source/agora/test/Base.d:478 [0x563496e63651]
```

ferencdg:
> This triggers a (valid) assert:
> 
> ```
> #### FATAL ERROR: 
> This node was started at source/agora/test/Flash.d:528
> This most likely means that the node crashed due to an uncaught exception
> If not, please file a bug at https://github.com/Geod24/localrest/
> Full error: core.exception.AssertError@source/agora/common/Amount.d(341)
> ----------------
> runtime.d:815 [0x563497768a4b]
> runtime.d:774 [0x563497768273]
> dmain2.d:292 [0x56349777bf00]
> deh.d:46 [0x5634977b226f]
> dwarfeh.d:332 [0x56349777cb02]
> /home/runner/work/agora/agora/source/agora/test/Base.d:140 [0x563496f089b1]
> exception.d:430 [0x56349776773f]
> exception.d:595 [0x563497767c24]
> /home/runner/work/agora/agora/source/agora/common/Amount.d:341 [0x563496e06413]
> /home/runner/work/agora/agora/source/agora/consensus/Fee.d:329 [0x5634973bcfdc]
> /opt/hostedtoolcache/dc/ldc2-1.25.0/x64/ldc2-1.25.0-linux-x86_64/bin/../import/std/algorithm/iteration.d:1026 [0x5634973bac9c]
> /home/runner/work/agora/agora/source/agora/consensus/Fee.d:329 [0x563496fdc189]
> /home/runner/work/agora/agora/source/agora/node/Ledger.d:620 [0x563496eb000e]
> /home/runner/work/agora/agora/source/agora/node/Ledger.d:1173 [0x563496f0aef3]
> /home/runner/work/agora/agora/source/agora/test/Base.d:478 [0x563496e63651]
> ```

fixed, anything else needed before this PR can be merged? It is blocking #1917

source/agora/common/Amount.d
Geod24: Well `this.percentage(percentage)` would have worked but I see your point now.

source/agora/consensus/data/Transaction.d
Geod24: It's not in the style guide ? Because we've always done it. Indeed should add it.

source/agora/consensus/data/DataPayload.d
```diff
@@ -26,6 +26,11 @@ public struct DataPayload
     /// The byte array of transaction data
     public const(ubyte)[] data;
 
+    /// The size of the data DataPayload object
+    public ulong sizeInBytes() const nothrow pure @safe @nogc
```
Geod24: Also missing the space before parens in all those functions

source/agora/consensus/Fee.d
```diff
@@ -427,17 +457,19 @@ public class FeeManager
...
-                assert(peekUTXO(input.utxo, utxo));
+                if (!peekUTXO(input.utxo, utxo))
+                    return "Unable to find input for utxo";
                 tot_in.mustAdd(utxo.output.value);
             }
 
-            assert(tx.getSumOutput(tot_out), "Not validated block in" ~
-                "getTXSetFees");
+            if (!tx.getSumOutput(tot_out))
+                return "Transaction output value is invalid";
```
Geod24, line 466: If those get triggered, isn't it logic error ?

ferencdg: this function can get called (and is called in PR #1881) with partially verified transactions (transactions that got relayed to us by an attacker through putTransaction). It can very well happen that those transactions didn't have their input in the UTXO or their total sum doesn't add up, so we shouldn't assert

source/agora/consensus/data/Transaction.d
ferencdg: > It's not in the style guide ? Because we've always done it. Indeed should add it.

Yeah, I just checked it is not there. I will go through this PR and add 'this' and also change the style guide

linked0:
The last commit should be fixuped to the fourth one, I think.

Geod24 (approved):
Needs a rebase but g2g

ferencdg:
> Needs a rebase but g2g

rebased and also prefixed everything with 'this'...

PR #1956, opened 2021-04-16
FullNode: Remove unused version (unittest)

Geod24:
This method is overridden in unittests (agora.test.Base) and this statement never executes.


kchulj (approved):
LGTM.
If the `MemoryStorage` will not be used, then the the ddoc needs update.

Geod24:
Good point, fixed doc

linked0 (approved):

PR #1958, opened 2021-04-16
ConsensusData validation: Check number of validators against MinValidatorCount

Geod24:
```
The code was checking that we didn't slash too many validators,
but it was not checking that there was enough validators left
after slashing, and those two conditions would only be true
iff Enrollment.MinValidatorCount == 1.
```

Hope this makes the logic clearer, it did take me a few tries.


Geod24:
I got three runs of `Integration` to pass out of three... Could it be ?

hewison-chris (approved):
LGTM although I think we need to make it robust in relation to the FIXME comment.

PR #1962, opened 2021-04-16
Minor improvements to SQL usability

Geod24:
Shortening names so join and manual queries are easier, and using TEXT so that the DB is (a bit) more readable.
Also having the `pubkey_hash` field as TEXT will allow to join `utxo` with `validator`.

### DIFF
--- a/doc/Database.md
+++ b/doc/Database.md
@@ -11,11 +11,11 @@ For more informations, see [this description](https://github.com/bosagora/agora/
 
 Our oldest table, only contain binary serialized data, should be changed to a readable format.
 
-| Field name  | SQL Type | D type    | Attributes  | Comment                                                |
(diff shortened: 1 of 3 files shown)


ferencdg:
I see a lot of mixing up of `pubkey_hash` and `pubkey` in our code like this
image

seems we store the pubkey, but keep refering to it as pubkey_hash... I wonder if we could clean up all those too

Geod24:
Correct, and renaming those parameter is planned, but the priority would be to have `val` be correctly serialized instead of a blob.
See https://github.com/bosagora/agora/blob/v0.x.x/doc/Database.md#state-db where this is mentioned.

Geod24:
Added a commit to rename `pubkey_hash`

hewison-chris (approved):
LGTM

PR #1966, opened 2021-04-16
Fix `.gitignore` and extend `ValidatorCleanRestart` timeout

Geod24:
See each commit.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,8 @@
 *.exe
 build/
 tracy_build
+source/agora/cli/version/version_build
+source/scpp/cpp_build
 
 # DUB
 .dub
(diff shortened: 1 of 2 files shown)


AndrejMitrovic (approved):

.gitignore
```diff
@@ -15,6 +15,8 @@
 *.exe
 build/
 tracy_build
+source/agora/cli/version/version_build
+source/scpp/cpp_build
```
kchulj, line 19: These files were created when I run `dub build`..

Geod24: Yes they are binaries. What's interesting is that they were not created before, or we would have seen it. So maybe verbose mode changed something ?

linked0 (approved):
LGTM

.gitignore
kchulj: Might as well try without the verbose mode.

PR #1968, opened 2021-04-16
Throw on missing validator in `Ledger.getActiveValidatorPublicKeys`

Geod24:
```
This is a situation that should never happen,
and we better throw an Exception here to notice it.
```

### DIFF
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -1641,6 +1641,10 @@ unittest
     const params = new immutable(ConsensusParams)();
 
     // normal test: UTXO set and Validator set updated
+    version (none)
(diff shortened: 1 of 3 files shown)


source/agora/node/Ledger.d
```diff
@@ -1641,6 +1641,10 @@ unittest
     const params = new immutable(ConsensusParams)();
 
     // normal test: UTXO set and Validator set updated
+    version (none)
+        // FIXME: This test is invalid as it is missing Enrollments
+        // However Enrollments cannot be trivially added, as pre-images are
+        // also necessary, hence this is temporarily disabled
```
Geod24, line 1647: @hewison-chris : I feel really stupid for not just doing this earlier...

PR #1976, opened 2021-04-18
Fix a couple bugs manifesting in the test-suite and improve usability

Geod24:
Currently HEAD is a bit broken, mostly due to `ValidatorCleanRestart` and sometimes `ValidatorRecurringEnrollment`.
This does not fully fix the issues, but removes a few spurious error message, makes things easier to debug, and removes a race condition. There's still an issue happening in the nomination process which I am currently debugging.


Geod24:
Merging because at least it fails timely now.

PR #1978, opened 2021-04-18
Make UTXODB store the UTXOs in a readable format

Geod24:
This will make debugging the live system much easier and also allows us
to join the UTXO set from the validator set, opening the door for smaller
footprint and reduction in duplication.


Geod24:
We got one (real) green run (ubuntu 18 / ldc-master), which is as good as it's going to get for the current state of HEAD, plus integration, so merging.

PR #1983, opened 2021-04-19
Allow loggers to be configured and to write to files

### LINKED ISSUE #1649: Separate the logger output into multiple files
Currently we rely on the log output being to `stdout` in the live system.
This is however showing its limit, as we can no longer follow the stream of messages.
Sociomantic had a pretty handy mean to configure loggers via config file, and we just need to adapt it to our config parser:
https://github.com/sociomantic-tsunami/ocean/blob/v6.x.x/src/ocean/util/log/Config.d

Geod24:
Replicate what Ocean is doing, but reading a YAML file.
Fixes https://github.com/bosagora/agora/issues/1649

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -153,12 +153,57 @@ dns:
 
 ################################################################################
 ##                               Logging options                              ##
+##                                                                            ##
+## Our logging system is hierarchical: Logger names are, e.g. `a.b.c.d`.      ##
+## Inside agora, every module that logs have a module-level logger matching   ##
+## it's module name, e.g. `agora.node.main`.                                  ##
+##                                                                            ##
+## In addition, some modules can have more nested loggers: one such use case  ##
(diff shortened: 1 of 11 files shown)


doc/config.example.yaml
```diff
@@ -153,12 +153,56 @@ dns:
...
+  level: Info
+  # Whether or not to log output to the console
+  console: true
+  # Output file to write the logging output to
+  # Note that output of a more specialized logger that uses another file won't be
+  # written to this file.
+  # The path is relative to `data_dir` unless an absolute path is supplied.
+  # Intermediate directories will be created as needed.
+  # This setting is optional, as no file would be written to if empty / not supplied.
+  file: log/root.log
```
Geod24: This is incorrect, currently we require `root:`

Geod24: Fixed

PR #1991, opened 2021-04-19
Fix #1989 - Make nomination interval configurable

### LINKED ISSUE #1989: Nomination timer should be configurable
https://github.com/bosagora/agora/blob/ee609ac6628833ea104086f68cc88a70ff69fd02/source/agora/consensus/protocol/Nominator.d#L224-L225

We're currently doing brute-force, and it seems to not work on the live system because of the latency.
It should at the very least be a config parameter so we can experiment with different values.

Geod24:
Pair programming with @kchulj


kchulj (approved):
✅

PR #1992, opened 2021-04-19
Fix #1990: Add time logging to API endpoints

### LINKED ISSUE #1990: Add timing on time spent in endpoints
Currently, we don't have timing on the time spent between a request arriving and being fully processed.
We might be inadvertently blocking some clients and causing timeouts if we increase some delays (e.g. #1989).
We should have a way to measure the endpoint timings.

Geod24:
CC @kchulj

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -68,6 +68,7 @@ import vibe.web.rest;
 
 import std.algorithm;
 import std.conv : to;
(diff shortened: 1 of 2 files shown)


Geod24:
Updated

Geod24:
I don't think this is worth merging. We found the issue, and the code can always be resurrected if a similar issue re-surface. An external service monitoring the node would be more accurate.

PR #1997, opened 2021-04-20
Limiting the rate of tx flooding

### LINKED ISSUE #1917: Limiting the rate with which transactions get relayed

ferencdg:
fixes: #1917

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -41,6 +41,19 @@ node:
   # The duration between requests for retrieving the latest blocks
   # from all other nodes
   block_catchup_interval_secs: 20
+  # The maximum number of transactions relayed in every batch.
+  # Value 0 means no limit.
+  relay_tx_max_num : 100
+  # Transaction relay batch is triggered in every `relay_tx_interval_secs`.
+  # Value 0 means, the transaction will be relayed immediately.
+  relay_tx_interval_secs : 30
+  # The minimum amount of fee a transaction has to have to be relayed.
+  # The fee is adjusted by the transaction size:
+  # adjusted fee = fee / transaction size in bytes.
+  relay_tx_min_fee : 0
+  # Transaction put into the relay queue will expire, and will be removed
+  # after `relay_tx_cache_exp_secs`.
+  relay_tx_cache_exp_secs : 1200
 
 # Each entry in this array is an interface Agora will listen to, allowing to
 # expose the same node on more than one network interface or with different
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -207,6 +207,27 @@ public struct NodeConfig
     /// The new block time offset has to be greater than the previous block time offset,
     /// but less than current time + block_time_offset_tolerance_secs
     public Duration block_time_offset_tolerance = 60.seconds;
+
+    // The percentage by which the double spend transaction's fee should be
+    // increased in order to be added to the transaction pool
+    public ubyte double_spent_threshold_pct = 20;
+
+    /// The maximum number of transactions relayed in every batch.
+    /// Value 0 means no limit.
+    public uint relay_tx_max_num;
+
+    /// Transaction relay batch is triggered in every `relay_tx_interval`.
+    /// Value 0 means, the transaction will be relayed immediately.
+    public Duration relay_tx_interval;
+
+    /// The minimum amount of fee a transaction has to have to be relayed.
+    /// The fee is adjusted by the transaction size:
+    /// adjusted fee = fee / transaction size in bytes.
+    public Amount relay_tx_min_fee;
+
+    /// Transaction put into the relay queue will expire, and will be removed
+    /// after `relay_tx_cache_exp`.
+    public Duration relay_tx_cache_exp;
 }
 
 /// Validator config
@@ -498,6 +519,11 @@ private NodeConfig parseNodeConfig (Node* node, in CommandLine cmdln)
     const block_time_offset_tolerance_secs = opt!(uint, "node", "block_time_offset_tolerance_secs")(cmdln, node);
     const network_discovery_interval = opt!(uint, "node", "network_discovery_interval_secs")(cmdln, node);
     const block_catchup_interval = opt!(uint, "node", "block_catchup_interval_secs")(cmdln, node);
+    const double_spent_threshold_pct = opt!(ubyte, "node", "double_spent_threshold_pct")(cmdln, node);
+    const relay_tx_max_num = opt!(ushort, "node", "relay_tx_max_num")(cmdln, node, 100);
+    Duration relay_tx_interval = opt!(ulong, "node", "relay_tx_interval_secs")(cmdln, node, 15).seconds;
+    const relay_tx_min_fee = Amount(opt!(ulong, "node", "relay_tx_min_fee")(cmdln, node, 0));
+    Duration relay_tx_cache_exp = opt!(ulong, "node", "relay_tx_cache_exp_secs")(cmdln, node, 1200).seconds;
 
     NodeConfig result = {
             min_listeners : min_listeners,
@@ -514,6 +540,11 @@ private NodeConfig parseNodeConfig (Node* node, in CommandLine cmdln)
             block_time_offset_tolerance : block_time_offset_tolerance_secs.seconds,
             network_discovery_interval : network_discovery_interval.seconds,
             block_catchup_interval : block_catchup_interval.seconds,
+            double_spent_threshold_pct : double_spent_threshold_pct,
+            relay_tx_max_num : relay_tx_max_num,
+            relay_tx_interval : relay_tx_interval,
+            relay_tx_min_fee : relay_tx_min_fee,
+            relay_tx_cache_exp : relay_tx_cache_exp,
     };
     return result;
 }
--- a/source/agora/node/TransactionRelayer.d
+++ b/source/agora/node/TransactionRelayer.d
@@ -1,0 +1,474 @@
+/*******************************************************************************
+
+    TransactionRelayer is responsible for storing transactions, and relaying
+    some of them to all known network clients periodically. The current
+    implementation relays transactions with the highest fees.
+
+    The class contains several optimizations
+    $(UL
+        $(LI Transaction hashes are stored in a RedBlackTree sorted by the
+             transaction fee. By doing so, getting the N transactions with
+             the highest fees can be done quickly.)
+        $(LI Transaction fee is calculated, and stored as part of the key
+             to avoid the expensive fee recalculation every time an addition/remove
+             operation is done on the RedBlackTree)
+        $(LI Separate hash set is used to be able to quickly decide whether a
+             transaction is in the RedBlackTree or not)
+        $(LI Only the transaction hash(and not the entire transaction) is
+             stored in the RedBlackTree to reduce memory usage)
+        $(LI There is a periodic cleanup of old entries in the Hash set and in
+             the RedBlackTree to reduce memory usage)
+    )
+
+    Copyright:
+        Copyright (c) 2019-2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.node.TransactionRelayer;
+
+import agora.common.Amount;
+import agora.common.Config;
+import agora.common.Task;
+import agora.common.Types;
+import agora.common.Set;
+import agora.consensus.data.Transaction;
+import agora.consensus.state.UTXOSet;
+import agora.crypto.Hash;
+import agora.utils.Log;
+import agora.network.Client;
+import agora.network.Clock;
+import agora.network.Manager;
+import agora.node.TransactionPool;
+
+import std.algorithm;
+import std.array : array;
+import std.container : DList, RedBlackTree;
+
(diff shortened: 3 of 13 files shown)


doc/config.example.yaml
```diff
@@ -41,6 +41,19 @@ node:
...
+  # Transaction relay batch is triggered in every `relay_tx_interval_secs`.
+  # Value 0 means, the transaction will be relayed immediately.
+  relay_tx_interval_secs : 15
+  # The minimum amount of fee a transaction has to have to be relayed.
+  # The fee is adjusted by the transactiction size:
+  # adjusted fee = fee / transaction size in bytes.
+  relay_tx_min_fee : 0
+  # Transaction put into the relay queue will expire, and will be removed
+  # after `relay_tx_cache_exp_secs`.
+  relay_tx_cache_exp_secs : 20
```
Geod24: 20 seconds ??

doc/config.example.yaml
```diff
@@ -41,6 +41,19 @@ node:
   # The duration between requests for retrieving the latest blocks
   # from all other nodes
   block_catchup_interval_secs: 20
+  # The maximum number of transactions relayed in every batch.
+  # Value 0 means no limit.
+  relay_tx_max_num : 100
+  # Transaction relay batch is triggered in every `relay_tx_interval_secs`.
+  # Value 0 means, the transaction will be relayed immediately.
+  relay_tx_interval_secs : 15
```
Geod24: I suggest 30 just to have a bit more leeway at first

doc/config.example.yaml
```diff
@@ -41,6 +41,19 @@ node:
...
+  # The minimum amount of fee a transaction has to have to be relayed.
+  # The fee is adjusted by the transactiction size:
```
```suggestion
  # The fee is adjusted by the transaction size:
```

doc/config.example.yaml
```diff
@@ -41,6 +41,19 @@ node:
...
+  # adjusted fee = fee / transaction size in bytes.
+  relay_tx_min_fee : 0
```
Geod24, line 53: How does this interact with the protocol-level fee lower limit ?

source/agora/common/Config.d
```diff
@@ -207,6 +207,27 @@ public struct NodeConfig
...
     /// but less than current time + block_time_offset_tolerance_secs
     public Duration block_time_offset_tolerance = 60.seconds;
+
+    // The percentage by which the double spend transaction's fee should be
+    // increased in order to be added to the transaction pool
+    public ubyte double_spent_threshold_pct = 20;
+
+    // The maximum number of transactions relayed in every batch.
+    // Value 0 means no limit.
+    ushort relay_tx_max_num;
```
Geod24:
```suggestion
    /// The maximum number of transactions relayed in every batch (0 = unlimited)
    public size_t relay_tx_max_num;
```

Is there are reason to put it `ushort` ?

source/agora/common/Config.d
```diff
@@ -207,6 +207,27 @@ public struct NodeConfig
...
+    Duration relay_tx_interval;
+
+    // The minimum amount of fee a transaction has to have to be relayed.
+    // The fee is adjusted by the transactiction size:
+    // adjusted fee = fee / transaction size in bytes.
+    Amount relay_tx_min_fee;
+
+    // Transaction put into the relay queue will expire, and will be removed
+    // after `relay_tx_cache_exp`.
+    Duration relay_tx_cache_exp;
```
Geod24: Style: `public` and DDOC

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+             the highest fees can be done quickly.)
+        $(LI Transaction fee is calculated, and stored as part of the key
+             to avoid the expensive fee recalculation every time an addition/remove
+             operation is done on the RedBlackTree)
+        $(LI Separate hash set is used to be able to quickly decide whether a
+             transaction is in the RedBlackTree or not)
+        $(LI Only the transaction hash(and not the entire transaction) is
+             stored in the RedBlackTree to reduce memory usage)
+        $(LI There is a periodic cleanup of old entries in the Hash set and in
+             the RedBlackTree to reduce memory usage)
```
Geod24, line 20: Nice usage of DDOC! But you are missing a closing parenthesis. Also DDOC now supports subset of Markdown by default, and that includes lists IIRC, so you could probably get away with that.

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+            pool = transaction pool
+            config = config
+            clients = network clients
+            taskman = task manager
+            clock = clock
+            getAdjustedTXFee = delegate calculating transaction fee
+
+    ***************************************************************************/
+
+    public this (ref TransactionPool pool, immutable Config config, DList!NodeConnInfo* clients,
```
Geod24: Why `ref` ?

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+        this.taskman = taskman;
+        this.clock = clock;
+        this.getAdjustedTXFee = getAdjustedTXFee;
+        this.txs = new RedBlackTree!(TxHolder, TxHolder.cmpFeeLess, true)();
+        this.log = Logger(__MODULE__);
+
+        // if relay_tx_interval == 0, then we immediately relay the transaction
+        if (this.config.node.relay_tx_interval != 0.seconds)
+            this.timers ~= this.taskman.setTimer(config.node.relay_tx_interval, &this.relayTransactions, Periodic.Yes);
+        this.timers ~= this.taskman.setTimer(1.minutes, &this.cleanRelayQueue, Periodic.Yes);
```
Geod24: Since we have a `stop` method, perhaps have a `start` one too ?

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+    }
+
+    /// Called during node shutdown.
+    public void shutdown () @safe nothrow
+    {
+        foreach (timer; this.timers)
+            timer.stop();
```
Geod24: Should probably clear that array too

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+            this.txs.removeFront();
+        }
+        return txs_gathered;
+    }
+
+    /// Sends an array of transactions ordered by fee to known network clients.
+    public void relayTransactions () @safe nothrow
+    {
+        this.getRelayTransactions().each!(tx => (*this.clients).each!(node_info => node_info.client.sendTransaction(tx)));
+    }
```
Geod24, line 223: This seems very wasteful. Perhaps we could have an endpoint that accepts a batch, directly ? But I don't mind for now, just thinking about later.

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+
+        Params:
+            tx = transaction to be added to the relay queue
+
+        Returns:
+            null, if there was no error, otherwise string detailing the error
+
+    ***************************************************************************/
+
+    public override string addTransaction(in Transaction tx)  @safe nothrow
```
Geod24:
```suggestion
    public override string addTransaction (in Transaction tx) @safe nothrow
```

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+        this.txs.insert(TxHolder(fee, tx_hash,
+            this.clock.networkTime() + this.config.node.relay_tx_cache_exp.total!"seconds"));
+        this.tx_hashes.put(tx_hash);
+
+        return null;
+    }
+
+    /***************************************************************************
+
+        Adds a transaction to the relay queue. Unlike addTransactionImp, this
```
Geod24:
```suggestion
        Adds a transaction to the relay queue.
        
        Unlike addTransactionImp, this [...]
```

I would phrase it as "can be overridden by child class to perform additional behavior", or the likes, since I suppose you did this to override it in UT ?

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+            this.relayTransactions();
+        return res;
+    }
+
+    ///
+    public static enum TransactionRelayError : string
+    {
+        TxAlreadyAdded = "Transaction has already been added to the relay queue",
+        TxFeeInvalid = "Transaction fee is invalid",
+        TxFeeTooLow = "Transaction fee is too low"
```
Geod24:
```suggestion
        TxFeeTooLow = "Transaction fee is too low",
```

Trailing commas in D are nice

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+            import std.format;
+
+            // Workaround for missing @safe on the delegate passed to RedBlackTree.toString().
+            auto safe_dg = cast(SafeSinkT) dg;
+
+            formattedWrite(safe_dg, "{ amount: ");
+            this.fee.toString(safe_dg);
+            formattedWrite(safe_dg, ", hash: ");
+            this.hash.toString(safe_dg);
+            formattedWrite(safe_dg, ", expiry: %d }", this.expiry);
```
Geod24: You could just merge it all into one statement, `formattedWrite` recognizes `sink`

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+        }
+
+        ///
+        public string toString () const @safe
+        {
+            string ret;
+            void delegate(scope const(char)[] v) mydg ;
+            scope SinkT dg = mydg;
+            this.toString(dg);
+            return ret;
```
Geod24: Why are you passing a null delegate ?

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+        this.pool = new TransactionPool(cacheDB);
+        this.utxo_set = new TestUTXOSet();
+        auto fee_man = new FeeManager(stateDB, params);
+        this.getAdjustedTXFee = (in Transaction tx, out Amount tot_fee)
+        {
+            return fee_man.getAdjustedTXFee(tx, &utxo_set.peekUTXO, tot_fee);
+        };
+
+        this.log = Logger(__MODULE__);
+        this.txs = new RedBlackTree!(TxHolder, TxHolder.cmpFeeLess, true)();
```
Geod24: Can this call the other ctor instead ? `this(...);`

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+    ***************************************************************************/
+
+    version (unittest)
+    static Transaction getTX (ushort gen_out_ind, Amount fee)
+    {
+        import agora.consensus.data.genesis.Test;
+        import agora.crypto.Key;
+        import std.array;
+
+        auto genesis_tx = GenesisBlock.txs.filter!(tx => tx.type == TxType.Payment).array()[0];
```
Geod24: Instead of `array()[0]` which does an allocation just to take the first element, use `front`.

Geod24 (commented):
Did a first pass, looks nice. My instinct would have been to integrate this in either the `TransactionPool` or the `NetworkManager` (probably the former) but the code is clean so this works. I skipped over the tests for now, will give it another eye after a coffee :P

doc/config.example.yaml
ferencdg: Do we have any protocol-level fee now, or do we ever want to have? 

Validators are rewarded from the block reward too, even if the block has no transactions with fee. Also if we limit the size of the block, and put the highest fee transaction into the block, then that is going to put a limit on the minimum fee too.

source/agora/common/Config.d
ferencdg: I think we don't want to delay relaying the transactions for too long, so let's say relay_tx_interval will be around 30 secs or 1 minute. I don't think we would like to relay more transactions than ushort.max in 30 secs. I looked into Ethereum and at the peak times, it has 10.000 transactions per block.

AndrejMitrovic: > it has 10.000 transactions per block.

That doesn't sound right.

source/agora/node/TransactionRelayer.d
```diff
@@ -0,0 +1,462 @@
...
+
+        Params:
+            tx = transaction to be added to the relay queue
+
+        Returns:
+            null, if there was no error, otherwise string detailing the error
+
+    ***************************************************************************/
+
+    private string addTransactionImp (in Transaction tx) @safe nothrow
```
linked0, line 252: There are two functions named `addTransactionImp` and `addTransaction`, which is kind of confusing. Why don't you merge the two functions into one or name them differently?

linked0 (commented):
Good work!
Please add more explanation for the commit message.

ferencdg:
fixed

AndrejMitrovic (approved):

ferencdg:
Any chance of merging this? My current issue ticket would need parts of this PR to be merged in.

Geod24 (approved):
LGTM, rebase it and we can merge.

ferencdg:
rebased

PR #2007, opened 2021-04-22
Restore Node logging capacity in unittests

Geod24:
PR1983, while making logger configurable, inadvertently removed code that
allowed logs to be saved and later accessed while running network tests.
This means that nowadays, when a test fails, we don't have any insight into what happened.


hewison-chris (approved):
LGTM

linked0 (approved):

PR #2008, opened 2021-04-22
Updated Vanity to work with Bech32; Update WK keypairs

Geod24:
🤷


linked0 (approved):
LGTM and thank you!
I think the two issues, or #1354 and #1878, are solved by this PR.

Geod24:
~Integration will not pass because Faucet needs to be updated with the latest keys. Will try locally to see if it's the only issue.~
But it should because we build Faucet...

PR #2009, opened 2021-04-22
Nomination interval: Raise default and parse as seconds

Geod24:
Having nomination interval expressed in milliseconds is overkill and only makes sense in tests,
where the overhead is very low, and we don't parse the config anyway.
Experience has shown that a low value create issues in the network,
so raise it to something more tolerant.

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -230,8 +230,8 @@ public struct ValidatorConfig
     /// How often should the periodic preimage reveal timer trigger (in seconds)
     public Duration preimage_reveal_interval = 10.seconds;
 
-    /// How often the nomination timer should trigger, in milliseconds
-    public Duration nomination_interval = 1.seconds;
+    /// How often the nomination timer should trigger, in seconds
+    public Duration nomination_interval = 5.seconds;
 }
 
 /// Admin API config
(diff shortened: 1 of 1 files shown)


source/agora/common/Config.d
```diff
@@ -620,7 +620,7 @@ private ValidatorConfig parseValidatorConfig (Node* node, in CommandLine cmdln)
     const preimage_reveal_interval = get!(Duration, "validator", "preimage_reveal_interval",
         str => str.to!ulong.seconds)(cmdln, node);
     const nomination_interval = get!(Duration, "validator", "nomination_interval",
-        str => str.to!ulong.msecs)(cmdln, node);
+        str => str.to!ulong.secs)(cmdln, node);
```
linked0: Should be `seconds`.

kchulj: Ahh I was wondering why the compiler doesn't like it..

source/agora/common/Config.d
```diff
@@ -231,7 +231,7 @@ public struct ValidatorConfig
     public Duration preimage_reveal_interval = 10.seconds;
 
     /// How often the nomination timer should trigger, in milliseconds
```
kchulj, line 233:
```suggestion
    /// How often the nomination timer should trigger
```

kchulj:
```suggestion
    /// How often the nomination timer should trigger
```

kchulj (dismissed):

Geod24:
All fixed

linked0 (approved):

PR #2010, opened 2021-04-22
Added minimum transaction fee

ferencdg:
Discussed with @Geod24 that this could help us with:
1. limiting the block size
2. limiting the size of memory pool

### DIFF
--- a/source/agora/consensus/Fee.d
+++ b/source/agora/consensus/Fee.d
@@ -224,6 +224,7 @@ public class FeeManager
 
         Params:
             tx = `Transaction`
+            tx_fee = transaction fee
 
         Return:
             `null` if the transaction is valid, a string explaining the reason it
@@ -231,7 +232,7 @@ public class FeeManager
 
     ***************************************************************************/
 
-    public string check (in Transaction tx, Amount sum_unspent) nothrow @safe
+    public string check (in Transaction tx, Amount tx_fee) nothrow @safe
     {
         if (tx.payload.data.length == 0)
             return null;
@@ -241,8 +242,8 @@ public class FeeManager
 
         const required_fee = calculateDataFee(tx.payload.data.length,
             this.params.TxPayloadFeeFactor);
-        if (sum_unspent < required_fee)
-            return "Transaction: There is not enough fee.";
+        if (tx_fee < required_fee)
+            return "Transaction: There is not enough data fee";
 
         return null;
     }
--- a/source/agora/consensus/Fee.d
+++ b/source/agora/consensus/Fee.d
@@ -245,6 +245,12 @@ public class FeeManager
         if (tx_fee < required_fee)
             return "Transaction: There is not enough data fee";
 
+        Amount minimumFee = params.MinFee;
+        if (!minimumFee.mul(tx.sizeInBytes()))
+            return "Fee: Transaction size overflows fee cap";
+        if (tx_fee < minimumFee)
+            return "Transaction: Fee is less than adjusted minimum fee";
+
         return null;
     }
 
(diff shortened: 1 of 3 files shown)


source/agora/consensus/Fee.d
```diff
@@ -245,6 +245,10 @@ public class FeeManager
         if (tx_fee < required_fee)
             return "Transaction: There is not enough data fee";
 
+        tx_fee.div(tx.sizeInBytes());
+        if (tx_fee < params.MinFee)
+            return "Transaction: Fee is less than adjusted minimum fee";
```
Geod24: Two things:
- `div` returns an `Amount`. If the return is non-zero, it means that the condition should be `<=`.
- Wouldn't a better approach be:
```D
Amount minimumFee = params.MinFee;
if (!minimumFee.mul(tx.sizeInBytes())
    return "Fee: Transaction size overflows fee cap";
if (tx_fee < minimumFee)
    return "...";
```

source/agora/consensus/data/Params.d
```diff
@@ -96,14 +97,15 @@ public immutable class ConsensusParams
...
-        uint quorum_threshold = 80)
+        uint quorum_threshold = 80, Amount min_fee = 0)
     {
         import agora.consensus.data.genesis.Test : GenesisBlock;
         import agora.utils.WellKnownKeys;
         ConsensusConfig config = {
             validator_cycle: validator_cycle,
             max_quorum_nodes: max_quorum_nodes,
             quorum_threshold: quorum_threshold,
+            min_fee: min_fee,
```
Geod24, line 108: Whenever we have a new consensus parameter, while it's good to add it to the struct, we really don't need to make it configurable unless there's a reason to. ATM I don't really see one for this ?

Geod24 (changes requested):
Simple and concise, but needs test. Also I would simply not make it configurable, what's the point ?

source/agora/consensus/data/Params.d
ferencdg: Our network tests doesn't have any fee, and the default minimum fee is 700, so I have to override it

source/agora/consensus/Fee.d
ferencdg: ugh, cannot count how many times I failed with this

source/agora/consensus/data/Params.d
```diff
@@ -143,6 +145,10 @@ public struct ConsensusConfig
 
     /// The amount of a penalty for slashed validators
     public Amount slash_penalty_amount = 10_000.coins;
+
+    /// The minimum - transaction size adjusted - fee.
```
linked0: All the three `-`s are minus signs?

ferencdg: `- transaction size adjusted -`
here it is a dash :) I can change to ()

linked0: Ah, `()` would be better. Ty. :)

source/agora/consensus/data/Params.d
ferencdg: In our unit tests, we almost never use fees and the default fee is 700, so they would fail.

source/agora/consensus/Fee.d
ferencdg: I think the original code is correct, but I might have misunderstood your comment.

minfee = 1 BOA/byte
txsize = 100
if tx fee is 0 - 99 then adjusted fee is 0, so we don't allow the transaction
if tx fee is 100 or above, then adjusted fee is 1+, so we allow the transaction 

minfee = 0 BOA/byte
txsize = 100
if tx fee is 0 - 99 then adjusted fee is 0, so we allow the transaction
if tx fee is 100 or above, then adjusted fee is 1+, so we allow the transaction

ferencdg:
fixed

Geod24 (approved):

PR #2016, opened 2021-04-22
Switch SCP build from g++ to clang++

### LINKED ISSUE #2015: Switch SCP build from g++ to clang++
https://github.com/bosagora/agora/blob/c1d67bf97c663e2b4e6c2db118f693aef034283e/source/scpp/build.d#L62

We should be using `clang++` here. We are doing so already on Mac, because `g++` just aliases to `clang++`, but not yet on Linux.

omerfirmak:
Fixes #2015

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -34,7 +34,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install libsqlite3-dev
           sudo apt-get install libsodium-dev
-          sudo apt-get install g++-9
+          sudo apt-get install clang
 
       - name: Build documentation
         run: |
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -53,7 +53,21 @@ extern(C++, (StdNamespace)) {
         else
             private alias TPtr = T*;
 
-        ~this() @safe {}
+        this (ref typeof(this) rhs) @trusted nothrow
+        {
+            copyCtorCPPObject!(typeof(this))(&this, &rhs);
+        }
+
+        ~this () @trusted nothrow
+        {
+            dtorCPPObject!(typeof(this))(&this);
+        }
+
+        void opAssign()(auto ref typeof(this) rhs) @trusted nothrow
+        {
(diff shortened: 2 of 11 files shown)


Geod24:
Well you're in luck, not only there!

source/scpd/Cpp.d
```diff
@@ -44,6 +44,7 @@ extern(C++, (StdNamespace)) {
             private alias TPtr = T*;
 
         ~this() @safe {}
+        this (const ref shared_ptr!T) @safe nothrow;
```
omerfirmak: Weirdly, this links fine on Linux with both GCC and Clang runtimes but fails on MacOS. Any ideas? @Geod24

Geod24: I'm pretty sure it gets inlined on Mac OSX. I'm surprised it does not on Linux, that sounds absurdly inefficient. 
This is the definition on my machine. Note the `inline`:
```C++
template<class _Tp>
template<class _Yp>
inline
shared_ptr<_Tp>::shared_ptr(const shared_ptr<_Yp>& __r,
                            typename enable_if<__compatible_with<_Yp, element_type>::value, __nat>::type)
         _NOEXCEPT
    : __ptr_(__r.__ptr_),
      __cntrl_(__r.__cntrl_)
{
    if (__cntrl_)
        __cntrl_->__add_shared();
}
```

omerfirmak:
Containers should be updated but tests seems to be passing now.

.github/workflows/documentation.yml
```diff
@@ -34,7 +34,7 @@ jobs:
           sudo apt-get update
           sudo apt-get install libsqlite3-dev
           sudo apt-get install libsodium-dev
-          sudo apt-get install g++-9
+          sudo apt-get install g++-9 clang
```
Geod24:
```suggestion
          sudo apt-get install clang
```

Don't think we need `g++-9` anymore, do we ? LDC might still use it for linking, but that's it.

Geod24 (commented):
You can update the containers there: https://github.com/bosagora/docker-agora-builder

PR #2017, opened 2021-04-22
Use lower ValidatorCycle in some EnrollmentManager tests

Geod24:
Those tests were using and EnrollmentCycle of 1008,
way above the 20 value we use for most of the tests,
and inconsistent with `agora.consensus.data.genesis.Test`.

While changing the default is still too invasive (it breaks Ledger tests),
changing those tests is pretty straightforward.


hewison-chris (approved):
LGTM

linked0 (approved):
LGTM

PR #2026, opened 2021-04-26
Validator: Don't needlessly update signature maps before acceptBlock

Geod24:
`acceptBlock` is the call that externalize new data (slashed validators,
new enrollments, etc...), hence trying to get that data from the ValidatorSet
before having externalized the block is bound to be error-prone.
The call was redundant since it is also performed in `Ledger.addValidatedBlock`.

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -296,17 +296,15 @@ public class Validator : FullNode, API
         import std.range;
         import std.format;
 
+        if (!super.acceptBlock(block))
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM.

kchulj (approved):

linked0 (approved):
LGTM.
How about adding the necessary space between the function name, `acceptBlock` and the left parenthesis on L291.

PR #2027, opened 2021-04-26
SCP: Use last block hash in computeHashNode

Geod24:
The random seed is not stable and changes as new pre-images get externalized,
so it could be used to influence a node, while the previous block hash is stable.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -1114,7 +1114,7 @@ extern(D):
         const uint hash_N = 1;
         const uint hash_P = 2;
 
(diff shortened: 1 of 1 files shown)


ferencdg (approved):

linked0 (approved):

kchulj:
What does stable mean in this context?

Geod24:
It has the potential to seemingly randomly change during a round as new pre-images get externalized.

ferencdg:
> What does stable mean in this context?

During the nominations 'computeHashNode' is called as part of something called 'leader selection', and the return of this function influences who will be the leader at a particular round.

This method was not stable, because new preimages could have been received during the nomination, and this method's return value was depending on the new preimages received.

kchulj:
Thank you for the explanation!

PR #2029, opened 2021-04-27
Flash: Publish settlement after timeout is reached

### LINKED ISSUE #1948: Flash: Make sure settlement is published after the time lock
https://github.com/bosagora/agora/blob/24c7efe3cf18333feb54416ec6b3e7cd5e963538/source/agora/flash/Channel.d#L709

We need to mark the channel as awaiting publishing a settlement, and then making sure we publish it after the timelock is complete.

AndrejMitrovic:
Fixes #1948

### DIFF
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -539,6 +539,9 @@ public class Channel
                     this.peer_pk));
         }
 
+        // check on start-up once in case this is a preloaded channel
+        this.checkPublishSettlement();
+
 LOuter: while (1)
         {
             scope (success)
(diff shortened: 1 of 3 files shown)


source/agora/flash/Channel.d
```diff
@@ -1580,6 +1606,8 @@ LOuter: while (1)
                 this.onSettleTxExternalized(tx);
             }
         }
+
+        this.checkPublishSettlement();
```
omerfirmak, line 1629: Does this mean that we would wait for the next block to broadcast the settlement once we detect the update TX in a block?

AndrejMitrovic: Ah you mean we shouldn't wait +1 blocks, but try to publish it right away? Let me check if it's doing that and fix it if not.

AndrejMitrovic: > Does this mean that we would wait for the next block to broadcast the settlement once we detect the update TX in a block?

I think we do need to wait for the first block it's unlocked at because otherwise the tx would be rejected when it tries to be added to the pool. It won't be able to attach to the Output of the update tx as the script execution will fail.

omerfirmak (approved):

source/agora/flash/Channel.d
Geod24: We should really fix the pool to accept future TXs. But in the meantime, yes.

PR #2030, opened 2021-04-27
Remove `createEnrollmentData` endpoint

### LINKED ISSUE #1659: Remove `createEnrollmentData`, use regular enrollment rules
https://github.com/bpfkorea/agora/blob/0b4304b800cfbe373442cf5de17596999e375610/source/agora/test/Base.d#L1300-L1310

This is a unittest special case that needs to be removed to better match the real system.

linked0:
The function, `createEnrollmentData` does not match with the way a node tries to enroll. So we remove the function and use `checkAndEnroll` instead.

Fixes #1659

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -1128,8 +1128,7 @@ public class TestAPIManager
     {
         static assert (isInputRange!Idxs);
 
-        Enrollment enroll = this.clients[client_idx].createEnrollmentData();
-        clients[client_idx].enrollValidator(enroll);
+        auto enroll = clients[client_idx].setRecurringEnrollment(true);
         client_idxs.each!(idx =>
             retryFor(this.clients[idx].getEnrollment(enroll.utxo_key) == enroll,
                 5.seconds,
@@ -1394,15 +1393,21 @@ public interface TestAPI : ValidatorAPI
 
     /***************************************************************************
 
-        TEMPORARY: Create a valid `Enrollment` for this node
+        Toggle enrollment
 
-        This method is a temporary workaround to create an Enrollment for a node
-        in tests. In the future it will be replaced by a simple function call,
-        once Enrollment catch-up is fixed and all node keypairs are well-known.
+        Check if the next enrollment is available or validator is not enrolled.
+        Make a node to enroll when necessary.
+
+        Params:
+            doIt = if the enrollments will be renewed continuouly or not
+
+        Returns:
+            The `Enrollment` used to enroll with,
+            or `Enrollment.init` if an enrollment is not possible
 
     ***************************************************************************/
 
-    public abstract Enrollment createEnrollmentData();
+    public abstract Enrollment setRecurringEnrollment (bool doIt);
 
     ///
     public QuorumConfig getQuorumConfig ();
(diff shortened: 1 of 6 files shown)


source/agora/test/Base.d
```diff
@@ -1778,33 +1780,12 @@ public class TestValidatorNode : Validator, TestAPI
         super(config);
     }
 
-    /// Create an enrollment data used as information for an validator
-    public override Enrollment createEnrollmentData ()
+    /// ditto
+    public Enrollment toggleEnroll ()
```
Geod24: Why is it returning the enrollment ? The point of this was to have something like:
```suggestion
    public void setRecurringEnrollment (bool doIt)
```

And then toggle it on/off from test. This PR makes things better, because at least we're not duplicating code anymore, but it doesn't yet go all the way.

linked0: Some tests need to check all the nodes to have the enrollment which is returned from the function like as follows.
```
Enrollment new_enroll = nodes[0].setRecurringEnrollment(true);
Thread.sleep(3.seconds);  // enrollValidator() can take a while..
nodes.each!(node =>
        retryFor(node.getEnrollment(new_enroll.utxo_key) == Enrollment.init, 1.seconds));
```
I think the name of the function does not match with the operation of that too.

linked0:
Ready for review.
I think I need to add a test case for the 'setRecurringEnrollment` function. Before adding that, can you review this?

source/agora/test/EnrollmentManager.d
```diff
@@ -78,8 +78,7 @@ unittest
     auto first_node = network.clients[0];
 
     // Request enrollment at the height of 15
-    auto enroll = first_node.createEnrollmentData();
-    first_node.enrollValidator(enroll);
+    auto enroll = first_node.setRecurringEnrollment(true);
```
Geod24, line 81: Unused variable

source/agora/test/Base.d
```diff
@@ -1776,35 +1784,29 @@ public class TestValidatorNode : Validator, TestAPI
...
+            this.recurring_enrollment &&
+            this.config.validator.recurring_enrollment &&
+                block.header.time_offset > cur_offset - 3 * this.params.BlockInterval.total!"seconds")
         {
-            if (utxo.type == TxType.Freeze &&
-                utxo.output.value.integral() >= Amount.MinFreezeAmount.integral())
-            {
-                utxo_hashes ~= key;
-            }
+            checkAndEnroll(this.ledger.getBlockHeight());
```
```suggestion
            this.checkAndEnroll(this.ledger.getBlockHeight());
```

Geod24 (approved):
LGTM just two nits

source/agora/test/EnrollmentManager.d
```diff
@@ -78,8 +78,7 @@ unittest
     auto first_node = network.clients[0];
 
     // Request enrollment at the height of 15
-    auto enroll = first_node.createEnrollmentData();
-    first_node.enrollValidator(enroll);
+    first_node.setRecurringEnrollment(true);
```
Geod24: Ah it was used L89, sorry... I think we can remove L89 too.

Geod24 (approved):
Test LGTM 👍

source/agora/test/EnrollmentManager.d
linked0: Ah, I should have checked that. Will update. Thanks.

PR #2031, opened 2021-04-28
Fix shutdown by using our own signal handler

Geod24:
```
Currently, we sometimes fail to shutdown because the Vibe.d
signal handler does not call our shutdown procedure.
This should fix that and remove any issues we saw with shutting down the node.
Note that this doesn't handle broken pipe (as Vibe.d does), as there currently
doesn't seem to be a need for it.
```


linked0 (approved):
LGTM

PR #2040, opened 2021-04-30
Integrate Talos (setup interface) into Agora

### LINKED ISSUE #1984: Integrate Talos into Agora
We got the source code, now only need to glue it together.

Geod24:
Fixes #1984.

The interface is not yet perfect, but it's usable.

The following needs to be fixed:
- Error messages in case of misconfiguration are not correctly displayed;
- After success, no message is given as to what the user needs to do (close the window);
- Many configuration options are missing, some are too prominent (e.g. BanManager);
- The network section is required even though DNS are provided (Agora bug?)
- The written configuration is essentially JSON, instead it should be pretty printed by the node;
- We're directly writing the body which isn't very safe;


Geod24:
Ah, and the code was committed as is, we need to update the dependencies. But since it broke horribly when I did so, I decided to focus on integrating what works now.

Geod24:
Want through dependency hell and back, should be good to go now!

Unfortunately this extends the build time quite a bit:
```
 => [builder 5/7] RUN npm ci && npm run build                                                                   129.5s
 => [builder 6/7] WORKDIR /root/agora/                                                                            0.0s
 => [builder 7/7] RUN AGORA_VERSION=HEAD dub build --skip-registry=all --compiler=ldc2 ${DUB_OPTIONS}           190.5s
```

Also dependencies need to be updated in the (near) future. I didn't do that originally because I wanted to separate concerns (and it broke horribly), but I had to update some already to make it work on Alpine Linux anyway, so I think the hardest part is done.

```
found 4706 vulnerabilities (4680 low, 26 high) in 2641 scanned packages
  run `npm audit fix` to fix 4705 of them.
  1 vulnerability requires manual review. See the full report for details.
```

Geod24:
🟢

PR #2041, opened 2021-04-30
Simplify Flash node hierarchy & use custom IP & port for the control interface

### LINKED ISSUE #1998: Flash: Use a different IP & port for the ControlAPI interface
The `ControlAPI`, just like the `AdminConfig`, should have a configurable binding address and default to `localhost`.

This API is the control interface and shouldn't be exposed to the open internet by default.

Kindly suggested by @MichaelKim20

AndrejMitrovic:
This gets rid of the custom FlashFullNode / FlashValidatorNode classes and now has a single AgoraFlashNode class (could be renamed later).

Requires: https://github.com/bosagora/agora/pull/2014

Fixes #1998


AndrejMitrovic:
Integration test fails due to faucet because I added a new import.

omerfirmak (approved):

Issue #2042, opened 2021-05-01
Agora HEAD crashes with Invalid Memory Operation


Geod24:
```
#0  0x00007fc52ba1b710 in _d_throw_exception () from /usr/lib/libdruntime-ldc-shared.so.96
#1  0x00007fc52b9e6dfe in onInvalidMemoryOperationError () from /usr/lib/libdruntime-ldc-shared.so.96
#2  0x00007fc52b9f5cf4 in core.internal.gc.impl.conservative.gc.ConservativeGC.runLocked!(core.internal.gc.impl.conservative.gc.ConservativeGC.mallocNoSync(ulong, uint, ref ulong, const(TypeInfo)), core.internal.gc.impl.conservative.gc.mallocTime, core.internal.gc.impl.conservative.gc.numMallocs, ulong, uint, ulong, const(TypeInfo)).runLocked(ref ulong, ref uint, ref ulong, ref const(TypeInfo)) () from /usr/lib/libdruntime-ldc-shared.so.96
#3  0x00007fc52b9f8ae5 in _DThn16_4core8internal2gc4impl12conservativeQw14ConservativeGC6mallocMFNbmkxC8TypeInfoZPv ()
   from /usr/lib/libdruntime-ldc-shared.so.96
#4  0x00007fc52b9e7135 in _d_assert () from /usr/lib/libdruntime-ldc-shared.so.96
#5  0x0000556700c577eb in vibe.internal.freelistref.FreeListRef!(vibe.stream.counting.EndCallbackInputStream, true).FreeListRef.checkInvariants() const (this=0x0) at /root/agora/submodules/vibe-core/source/vibe/internal/freelistref.d:175
#6  0x0000556700c576e5 in vibe.internal.freelistref.FreeListRef!(vibe.stream.counting.EndCallbackInputStream, true).FreeListRef.clear() (this=0x0) at /root/agora/submodules/vibe-core/source/vibe/internal/freelistref.d:149
#7  0x0000556700c576b1 in vibe.internal.freelistref.FreeListRef!(vibe.stream.counting.EndCallbackInputStream, true).FreeListRef.~this() (this=0x0) at /root/agora/submodules/vibe-core/source/vibe/internal/freelistref.d:121
#8  0x0000556700c34a10 in vibe.http.client.HTTPClientResponse.__fieldDtor() (this=0x7fc5263cd000) at client.d:1078
[...]

PR #2049, opened 2021-05-03
Provide an assertHandler to see assertion failures in dtors

Geod24:
Part of #2042


Geod24:
Self merging so we can deploy this

PR #2053, opened 2021-05-03
Revert "Human-readable part for test network"

Geod24:
This reverts commit 3ec9303e176305abf900526536f7f307e1dc3e14.
The commit broke the test network, as the name registry,
and any tool consuming addresses, will now trip on the enforce
in Key.d:186, which checks that the prefix matches.
In order to restore the live system to a workable state,
this needs to be temporarily reverted.


linked0 (approved):
LGTM.
I will check the cause.

PR #2055, opened 2021-05-04
Make pretty printer more generic

### LINKED ISSUE #1903: PrettyPrinter should be generic
Currently the process of adding pretty-printing to a type is tedious:

https://github.com/bosagora/agora/blob/ef355e6792dd49ef2d43816ffaf1b09c11a156e5/source/agora/utils/PrettyPrinter.d#L59-L80

It would be nice if it supported any aggregate that happens to embed some of these structures. Pseudocode (untested):

```diff
diff --git a/source/agora/utils/PrettyPrinter.d b/source/agora/utils/PrettyPrinter.d
index d9eb613e..8c79ef67 100644
--- a/source/agora/utils/PrettyPrinter.d
+++ b/source/agora/utils/PrettyPrinter.d
@@ -78,6 +78,19 @@ public auto prettify (T) (const auto ref T input) nothrow
         return ConsensusDataFmt(input);
     else static if (is(T : const QuorumConfig))
         return QuorumConfigFmt(input);
+    else static if (is(T : struct))
+    {
+        struct Formatted
+        {
+            public void toString (scope void delegate (scope const char[]) @safe sink) @safe nothrow
+            {
+                foreach (field; input.tupleof)
+                    formattedWrite(sink, "%s", prettify(field));
+            }
+        }
+
+        return Formatted();
+    }
     else
         return input;
 }
```

Something like that.

AndrejMitrovic:
Should probably leave your suggestions on what is the best formatting layout.

Also note that calling `prettify` on a struct which embeds unions will fail because `toString` has been marked as `@safe` here. I think this is actually a good thing because it prevents us from printing out all of the union values at once (only one of the unions would ever be valid).

Fixes #1903

### DIFF
--- a/source/agora/utils/PrettyPrinter.d
+++ b/source/agora/utils/PrettyPrinter.d
@@ -78,8 +78,64 @@ public auto prettify (T) (const auto ref T input) nothrow
         return ConsensusDataFmt(input);
     else static if (is(T : const QuorumConfig))
         return QuorumConfigFmt(input);
+    else static if (is(T == struct))  // recurse into fields and auto-prettify
+    {
+        struct Formatted
+        {
+            public void toString (scope void delegate (scope const char[]) @safe sink) @safe nothrow
+            {
+                import std.traits;
+                try
+                {
+                    sink("{ ");
+                    foreach (idx, field; FieldNameTuple!T)
+                    {
+                        formattedWrite(sink, "%s: %s", field,
+                            prettify(__traits(getMember, input, field)));
+                        if (idx + 1 < FieldNameTuple!T.length)
+                            sink(", ");
+                    }
+                    sink(" }");
+                }
+                catch (Exception ex)
+                {
+                    assert(0, ex.msg);
+                }
+            }
+        }
+
+        return Formatted();
+    }
     else
         return input;
+}
+
+unittest
+{
+    static struct S
+    {
+        static struct Nested
+        {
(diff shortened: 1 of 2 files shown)


source/agora/utils/PrettyPrinter.d
```diff
@@ -78,10 +78,65 @@ public auto prettify (T) (const auto ref T input) nothrow
...
+        {
+            public void toString (scope void delegate (scope const char[]) @safe sink) @safe nothrow
+            {
+                import std.traits;
+                try
+                {
+                    sink("{ ");
+                    foreach (idx, field; FieldNameTuple!T)
+                    {
+                        formattedWrite(sink, field ~ ": %s", prettify(__traits(getMember, input, field)));
```
AndrejMitrovic: Gonna fix this allocation, hold up.

Geod24 (approved):
Nice.

source/agora/utils/PrettyPrinter.d
Geod24: Is it allocating ? I think the compiler const fold. But better to make it explicitly use `%s: %s` yes.

AndrejMitrovic: You know that's even better, I'll do that.

kchulj (approved):

linked0 (approved):

PR #2056, opened 2021-05-04
Merge QRCodeInterface into AdminInterface 

Geod24:
CC @TrustHenry : Can you test this ?


source/agora/node/admin/AdminInterface.d
```diff
@@ -68,47 +74,225 @@ public class AdminInterface
...
-        return listenHTTP(settings, router);
+    public override string encryptionKeyQR (
+        string app, ulong height, out string contentType, out string vary)
+    {
+        // TODO: PreImage to be released at that height
+        // Or it can be pre-released, so it can be `height + PreimageRevealPeriod`
+        PreImageInfo pre_image;
+
+        // blake2b hashing
+        const Hash key = hashMulti(pre_image, app);
```
hewison-chris, line 151: Can it be pre_image.hash ?

Geod24: I just moved the code. It always uses the default-initialized pre-image so it's not valid anyway...

Geod24:
For reference, this was triggered because I am working on adding an endpoint to dynamically reconfigure the loggers via HTTP queries.

hewison-chris (approved):
LGTM

linked0 (approved):
LGTM. We need to reformat the docs of `NodeControlAPI` functions though.

source/agora/node/admin/AdminInterface.d
TrustHenry: This should be fixed to a block height-based pre-image later.

TrustHenry (approved):
LGTM

PR #2057, opened 2021-05-04
Make KeyPair / SecretKey methods safe

Geod24:
Should be a no brainer

### DIFF
--- a/source/agora/crypto/Key.d
+++ b/source/agora/crypto/Key.d
@@ -55,6 +55,8 @@ unittest
 /// To construct addresses (PublicKey), see `fromString`
 public struct KeyPair
 {
+    @safe:
+
     /// Public key
     public const PublicKey address;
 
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM

source/agora/crypto/Key.d
```diff
@@ -322,7 +326,7 @@ public struct SecretKey
         import std.format : phobos_format = format;
         import ocean.text.convert.Formatter : ocean_format = format;
         assert(phobos_format("%s", sd) == "**SECRET**");
-        assert(ocean_format("{}", sd) == "**SECRET**");
+        () @trusted { assert(ocean_format("{}", sd) == "**SECRET**"); }();
```
linked0, line 329: Why is the `ocean_format` nested in `@trusted`?

Geod24: Because `format` isn't `@safe`, and having the `@safe` at the top also applies to `unittest`.

linked0: Is `phobos_format` `@safe`? I can't find `@safe` in the prototype of the `format` function in `std.format`.

linked0 (approved):

PR #2058, opened 2021-05-06
Expose an API to allow reconfiguring loggers dynamically

Geod24:
This currently does not fully work (changing appenders has bugs),
but works for log level which is the most useful feature.

### DIFF
--- a/source/agora/api/Admin.d
+++ b/source/agora/api/Admin.d
@@ -16,6 +16,8 @@ module agora.api.Admin;
 import vibe.data.serialization;
 import vibe.web.rest;
 
+import ocean.util.log.ILogger;
+
 import std.typecons;
 
 
@@ -24,6 +26,20 @@ import std.typecons;
 public interface NodeControlAPI
 {
     @safe:
+
+    /***************************************************************************
+
+        Set the configuration of a Logger
+
+    ***************************************************************************/
+
+    public void postLogger (
+        @viaQuery("name") string name,
+        @viaQuery("propagate") bool propagate = true,
+        @viaQuery("level") Nullable!(ILogger.Level) level = Nullable!(ILogger.Level).init,
+        @viaQuery("additive") Nullable!bool additive = Nullable!(bool).init,
+        @viaQuery("console") Nullable!bool console = Nullable!(bool).init,
+        @viaQuery("file") Nullable!string file = Nullable!(string).init);
 
     /***************************************************************************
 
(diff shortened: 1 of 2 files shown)


source/agora/api/Admin.d
```diff
@@ -25,6 +27,19 @@ public interface NodeControlAPI
...
     @safe:
 
+    /***************************************************************************
+
+        Set the configuration of a Logger
+
+    ***************************************************************************/
+
+    public void postLogger (
+        @viaQuery("name") string name, @viaQuery("propagate") bool propagate = true,
```
linked0:
```suggestion
        @viaQuery("name") string name,
        @viaQuery("propagate") bool propagate = true,
```

linked0: Where can I get information about `@viaQuery` in the `dlang` site ?

source/agora/node/admin/AdminInterface.d
```diff
@@ -86,6 +89,39 @@ public class AdminInterface : NodeControlAPI
...
+
+        if (!additive.isNull())
+            logger.additive = additive.get();
+        if (!level.isNull())
+            logger.level(level.get(), propagate);
+
+        if (!console.isNull() || !file.isNull())
+        {
+            logger.clear();
+            if (!console.isNull())
```
linked0: It would be better to take this `if` part out  of `if (!console.isNull() || !file.isNull())` part.

source/agora/node/admin/AdminInterface.d
```diff
@@ -86,6 +89,39 @@ public class AdminInterface : NodeControlAPI
...
+            {
+                auto appender = new AppendConsole();
+                appender.layout(new AgoraLayout());
+                logger.add(appender);
+            }
+            if (!file.isNull())
```
linked0: Ditto

linked0 (approved):
LGTM just some nits.

source/agora/api/Admin.d
Geod24: It's a parameter UDA: https://vibed.org/api/vibe.web.common/viaQuery

source/agora/node/admin/AdminInterface.d
Geod24: What do you mean ? `logger.clear()` needs to be caller if either is true, but it needs to only be called once, ahead of setting the new appender.

Geod24:
Added a comment and applied styling suggestions

source/agora/node/admin/AdminInterface.d
linked0: I mean it like follows. It looks simpler for me. But it doesn't matter.

```
if (!console.isNull() || !file.isNull())
{
    logger.clear();
}

if (!console.isNull())
{
}

if (!file.isNull())
{
}
```

Geod24: Yeah I did that change

PR #2059, opened 2021-05-06
Devel: Add a config file to run a single validator locally

Geod24:
```
This can come in very handy to test changes that don't need network,
for example the node API.
```

Inspired from https://github.com/Geod24/rns-test-env
Note that in order to use faucet, one need to run it locally, as docker images don't have access to the host's network on OSX.

### DIFF
--- a/devel/config-single.yaml
+++ b/devel/config-single.yaml
@@ -1,0 +1,48 @@
+#
+# This configuration file is intended to provide a standalone local network
+# with a single validator.
+#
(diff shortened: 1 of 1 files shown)


linked0 (approved):

kchulj (approved):

Geod24:
My node doesn't seem to be re-enrolling locally. But it's not a configuration issue, probably a small bug in the way we handle Enrollments.

PR #2060, opened 2021-05-06
Add syntactical validation for lock scripts

### LINKED ISSUE #1380: Execution engine: Lock scripts should be syntactically checked (+ conditional opcodes checks)
While writing the networking tests I've realized the lock scripts' conditionals are only checked once they're executed together with the unlock scripts. But that presents a danger of allowing badly formed lock scripts into the blockchain. Someone could have their funds locked forever due to a bad script such as:

```
IF
    ...
ELSE
   ...
// missing END_IF
```

The above only becomes an error once the unlock script is presented. We need to syntactically validate new lock scripts in the `lock + unlock => [new lock]` scenario and not just `lock + unlock` combos.

AndrejMitrovic:
Also see comment in https://github.com/bosagora/agora/issues/1380#issuecomment-833241704

Fixes #1380

### DIFF
--- a/source/agora/consensus/validation/Transaction.d
+++ b/source/agora/consensus/validation/Transaction.d
@@ -86,6 +86,9 @@ public string isInvalidReason (
         if (tx.type == TxType.Freeze && idx > 0 &&
             output.value < Amount.MinFreezeAmount)
             return "Transaction: All non-refund outputs must be over the minimum freezing amount";
+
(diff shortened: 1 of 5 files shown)


AndrejMitrovic:
This seems to fail consistently in the system integration test, but the logs don't show anything interesting. 🤔

hewison-chris:
Continued in https://github.com/bosagora/agora/pull/2136

PR #2061, opened 2021-05-06
Directly add our own enrollments to the pool when re-enrolling

Geod24:
```
Previously, we would not save our own enrollment to the pool when re-enrolling,
instead relying on the gossipping done by other nodes.
To make this more straightforward, we also split the 'EnrollmentPool.add' method
between add (with validation) and 'addValidated', in a manner similar to the Ledger.
```

Testing locally with the single-node network.

### DIFF
--- a/source/agora/consensus/EnrollmentPool.d
+++ b/source/agora/consensus/EnrollmentPool.d
@@ -115,7 +115,7 @@ public class EnrollmentPool
             () @trusted {
                 this.db.execute("REPLACE INTO enrollment_pool " ~
                     "(key, val, avail_height) VALUES (?, ?, ?)",
-                    enroll.utxo_key.toString(), buffer, avail_height.value);
+                    enroll.utxo_key, buffer, avail_height.value);
             }();
         }
         catch (Exception ex)
--- a/source/agora/consensus/EnrollmentPool.d
+++ b/source/agora/consensus/EnrollmentPool.d
@@ -67,19 +67,20 @@ public class EnrollmentPool
 
     /***************************************************************************
 
-        Add a enrollment data to the enrollment pool
+        Add a enrollment data to the enrollment pool after validating it
 
         Params:
             enroll = the enrollment data to add
             avail_height = height at which the enrollment is available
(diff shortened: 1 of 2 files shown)


AndrejMitrovic (approved):

Geod24:
```
2021-05-06 16:10:13,551 Info [agora.consensus.protocol.Nominator] - Nominating { tx_set: [0x5c64...4138], enrolls: [{ utxo: 0x096b...7e64, seed: 0x39ff...164f, cycles: 20, sig: 0xa060...834b }], missing_validators: [], time_offset: 10825788 } at 1620285013
2021-05-06 16:10:13,551 Info [agora.consensus.protocol.Nominator] - agora.consensus.protocol.Nominator.Nominator.nominate(): Proposing tx set for slot 20
2021-05-06 16:10:13,620 Info [agora.consensus.protocol.Nominator] - Externalized consensus data set at 20: { tx_set: [0x5c64...4138], enrolls: [{ utxo: 0x096b...7e64, seed: 0x39ff...164f, cycles: 20, sig: 0xa060...834b }], missing_validators: [], time_offset: 10825788 }
2021-05-06 16:10:13,633 Info [agora.node.Ledger] - Beginning externalization of block #20
2021-05-06 16:10:13,633 Info [agora.node.Ledger] - Transactions: 1 - Enrollments: 1
2021-05-06 16:10:13,633 Info [agora.node.Ledger] - Validators: Active: 0 - Signing: [128] - Slashed: []
```

Yep, it works well. That might unblock us on some other issues too.

kchulj (approved):
LGTM but CI 🔴

Geod24:
Looks like this uncovered the race condition

Geod24:
Found the bug, we were not re-sending enrollments because of a check in `checkAndEnroll`.
The check would previously only trigger after we're heard from other nodes of our own enrollment, but seemed to interfere with `ValidatorRecurringEnrollment`. As mentioned, since it's only an optimisation, I removed it for the time being.

linked0 (approved):
LGTM

PR #2063, opened 2021-05-07
Improve encapsulation of validator index maps

Geod24:
This is by no means perfect but it's enough of a step up that it deserves its own PR IMO.

Here's the first commit message which explains the reasoning behind the change:
```
Those maps are slightly redundant (especially when it comes to the most recent height),
and problematic as they are currently stored in the EnrollmentManager.
Since they are validator-related, it would seem more appropriate to have them in
either the Ledger or the ValidatorSet. For such a transition to be easy,
it would be better if all methods related to it were properly encapsulated,
which this commit starts doing.
Further commits will move elements of the function inside the new structure.
This wasn't done in one go as this refactor also provides opportunity to
remove other usages, and to improve error handling (e.g. some parts depend
on the logger inside EnrollmentManager, which conflicts with our goal).
```

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -134,10 +134,28 @@ public class EnrollmentManager
 
         /// used for validating the signature
         private PublicKey[ulong][Height] index_to_key;
+
+        /***********************************************************************
+
+            Update the validator key index maps
+
+            Params:
+                height = the block height the validators will next sign
+
+        ***********************************************************************/
+
+        public void update (in Height height, in PublicKey[] keys) @safe nothrow
+        {
+            foreach (idx, key; keys)
+            {
+                this.key_to_index[height][key] = idx;
+                this.index_to_key[height][idx] = key;
+            }
+        }
     }
 
     /// Ditto
-    private ValidatorKeyMap keymap;
+    public ValidatorKeyMap keymap;
 
     /***************************************************************************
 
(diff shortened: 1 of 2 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -114,18 +114,30 @@ public class EnrollmentManager
...
 
-    /// In order to support collecting signatures *after* a block is
-    /// externalized we must know the key index for the block header bitmask
-    /// for the active validator set *for that block height*.
-    /// Once a block is externalized the validator set might change
-    /// - so we store for each block height (can be optimised later) the maps
-    ///   for key_to_index and index_to_key
+    /***************************************************************************
+
+        A two-way maps between public key and indexes at a given height
```
hewison-chris: maps -> map

source/agora/node/Ledger.d
```diff
@@ -833,22 +833,28 @@ public class Ledger
         Point sum_K;
         Point sum_R;
         const Scalar challenge = hashFull(block);
-        const enrolled_validators = this.enroll_man.getCountOfValidators(block.header.height);
+        PublicKey[] validators;
+        try
+            validators = this.enroll_man.getActiveValidatorPublicKeys(block.header.height);
+        catch (Exception exc)
+        {
+            this.log.error("Exception thrown bt getActiveValidatorPublicKey while externalizing valid block: {}", exc);
```
hewison-chris: bt -> by

source/agora/node/Ledger.d
```diff
@@ -557,7 +557,11 @@ public class Ledger
                 assert(0);
             }
         }
-        this.enroll_man.updateValidatorIndexMaps(block.header.height + 1);
+
+        const nheight = block.header.height + 1;
```
hewison-chris: nheight -> next

hewison-chris (approved):
LGTM, just a few typos

source/agora/node/Ledger.d
Geod24: That one was actually on purpose (nheight => next height), but I guess next is better.

hewison-chris: I thought maybe that was the case but think next is still better.

Geod24:
Fixed typos

linked0 (approved):
LGTM

PR #2065, opened 2021-05-07
Update SCP to v16.0.0

AndrejMitrovic:
In case I don't finish tracking down the bugs someone else could take a look at it too.

Related to: https://github.com/bosagora/agora/issues/505

NOTES:

- Targeting v16.0.0 rather than v17.0.0 as the newest one uses C++17 and I'm not sure about the compatibility with D yet.

CHANGES:

- Removed `atomic<bool>` from the quorum intersection checker because it's unnecessary for us. Our quorum generator should not produce giant quorums, unlike in Stellar where they're mostly user-defined (to an extent).


AndrejMitrovic:
The nomination has been fixed. But now `combineCandidates` segfaults. It's probably something to do with the data size & layout of the structs.

AndrejMitrovic:
Found the second bug. I was passing by value instead of by ref. The simple test now passes. I'll let the CI do its run.

AndrejMitrovic:
Passes the CI on ubuntu. I still have to re-enable the quorum checker code, but I think I need https://github.com/bosagora/agora/pull/2066 first to resolve the runtime bugs with using `unorered_map`.

AndrejMitrovic:
I've rebased it against v0.x.x but there's many linker errors. The issue is missing instantiations of `cppSizeOf` and a few other symbols. It's a royal PITA trying to instantiate the right macros.

AndrejMitrovic:
@omerfirmak any idea why I'm still getting this linker error with the last commit?:

>   "int getCPPSizeof<std::__1::unordered_map<int, int, std::__1::RandHasher<int, std::__1::hash<int> >, std::__1::equal_to<int>, std::__1::allocator<std::__1::pair<int const, int> > > >()", referenced from:

source/scpd/Cpp.d
```diff
@@ -89,6 +89,8 @@ extern(C++, (StdNamespace)) extern(C++, class) struct allocator (T) {}
 
 extern(C++, (StdNamespace)) extern(C++, class) struct less (T) {}
 
+extern(C++, (StdNamespace)) extern(C++, class) struct RandHasher (T, Hasher = hash!T) { }
```
omerfirmak:
```suggestion
extern(C++, `std`) extern(C++, class) struct RandHasher (T, Hasher = hash!T) { }
```

should fix it i guess

AndrejMitrovic: Hey that worked! Thanks!

Geod24:
> +3,420 −1,079

Walter, is that you ? :trivial:

Geod24:
For anyone ~mad~ kind enough to review: https://github.com/stellar/stellar-core/compare/v11.2.0...v16.0.0

source/scpd/scp/SCPDriver.d
```diff
@@ -108,6 +155,12 @@ nothrow:
         void* toShortString(ref const(PublicKey) pk) const;
     }
 
+    // `getHashOf` computes the hash for the given vector of byte vector
+    abstract Hash getHashOf(ref vector!Value vals) const;
+
+    // Agora: routing through xdr_to_opaque to get the same hashing behavior
+    Hash getHashOfQuorum(ref const(SCPQuorumSet) qSet) const;
```
Geod24: Move the `@trusted` here ?

source/scpp/src/util/Logging.cpp
```diff
@@ -19,8 +19,8 @@ DLogger::DLogger(int level, std::string const& loggerName)
 
 DLogger::~DLogger()
 {
-    if (mLevel == TRACE && !Logging::logTrace("SCP"))
-        return;
+    //if (mLevel == TRACE && !Logging::logTrace("SCP"))
+    //    return;
```
Geod24: What's the story here ?

source/scpp/src/scp/QuorumSetUtils.cpp
```diff
@@ -34,47 +35,37 @@ class QuorumSetSanityChecker
...
+            "Total number of nodes in a quorum must be within 1 and 1000";
     }
-    else if (mCount > 1000)
-    {
-        *reason = "Number of validator nodes exceeds the limit of 1000";
-        mIsSane = false;
-    }
-
-    // only one of the two may be true
-    assert(mIsSane ^ (*reason != nullptr));
```
Geod24, line 62: Such a loss

source/scpd/Cpp.d
```diff
@@ -113,7 +115,7 @@ extern(C++, (StdNamespace)) {
         else
             private alias TPtr = T*;
 
-        mixin CPPBindingMixin!(shared_ptr!T);
+        mixin CPPBindingMixin!(typeof(this));
```
Geod24: You could also use:
```suggestion
        mixin CPPBindingMixin!(shared_ptr);
```

As the template name refers to the current instantiation. I'm curious why the change was necessary though ?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -1231,6 +1241,38 @@ extern(D):
...
+        }
+        void trusted () @trusted
+        {
+            crypto_generichash_final(&state, hash[].ptr, Hash.sizeof);
+        }
+        trusted();
+
+        ubyte[64] bytes = hash[][0..64];
+        return StellarHash(bytes);
+    }
```
Geod24, line 1275: I think we can improve this but I need to look at it with fresh eyes

source/scpd/Cpp.d
```diff
@@ -151,14 +153,14 @@ private pure nothrow @nogc @safe extern(C++) size_t cpp_unordered_map_length (K,
 
 /// Rudimentary bindings for std::unordered_map
 extern(C++, (StdNamespace))
-public struct unordered_map (Key, T, Hash = hash!Key, KeyEqual = equal_to!Key, Allocator = allocator!(pair!(const Key, T)))
+public struct unordered_map (Key, T, Hash = RandHasher!(Key), KeyEqual = equal_to!Key, Allocator = allocator!(pair!(const Key, T)))
```
Geod24, line 156: Did we get this wrong or it's a new thing in C++ ?

source/scpd/Cpp.d
```diff
@@ -532,6 +534,13 @@ unittest
 /// Invoke an std::function pointer (note: must be void* due to mangling issues)
 extern(C++) void callCPPDelegate (void* cb);
 
+public mixin template NonMovableOrCopyable ()
+{
+    @disable this ();
+    @disable this (this);
+    @disable ref typeof (this) opAssign () (auto ref typeof(this) rhs);
+}
```
Geod24, line 542: 👍

source/scpd/quorum/QuorumIntersectionChecker.d
```diff
@@ -27,7 +27,8 @@ public abstract class QuorumIntersectionChecker
   public:
     /// Create & initialize a QuorumIntersectionChecker with the given map
     static shared_ptr!QuorumIntersectionChecker create (
-        ref const(QuorumTracker.QuorumMap) map);
+        ref const(QuorumTracker.QuorumMap) map,
+	bool quiet = false);
```
Geod24, line 31: Indent

source/scpd/quorum/QuorumTracker.d
```diff
@@ -69,6 +84,14 @@ public:
...
     /// returns the known quorum map of the entire network
     ref const(QuorumMap) getQuorum () const;
+
+    // Given a node in the transitive quorum, but _outside_ of the local quorum
+    // set, find all nodes in the local quorum set with the shortest distance to
+    // `nodeOutsideQset`. If a node from the local quorum set is passed in,
+    // return itself. If the local node is passed in, return an empty set.
+    // todo:
+    //std::set<NodeID> const&
+    //findClosestValidators(NodeID const& nodeOutsideQset);
```
Geod24: Not needed ?

source/scpd/scp/BallotProtocol.d
```diff
@@ -57,22 +57,39 @@ extern(C++, class) public struct BallotProtocol
...
+    extern (C++, class) public struct SCPBallotWrapper
+    {
+    extern(C++):
+        ValueWrapperPtr mWvalue;
+        const SCPBallot ballot;
+
+      public:
+        this(uint c, ValueWrapperPtr vw);
+        this(ref const(SCPBallotWrapper) o);
+        ~this();
```
Geod24: Don't see the dtor on the C++ side, I assume it's implicit due to fields ?

source/scpd/scp/BallotProtocol.d
```diff
@@ -297,6 +320,12 @@ extern(C++, class) public struct BallotProtocol
     void startBallotProtocolTimer();
     void stopBallotProtocolTimer();
     void checkHeardFromQuorum();
+
+    SCPBallotWrapperUPtr makeBallot(SCPBallot* b) const;
+    SCPBallotWrapperUPtr makeBallot(uint c, Value* v) const;
+
+    // todo
+    //std::string ballotToStr(SCPBallotWrapperUPtr const& ballot) const;
```
Geod24: What's the issue here ?

source/scpd/scp/SCPDriver.d
```diff
@@ -23,6 +23,47 @@ import core.stdc.inttypes;
...
+    ref const(SCPEnvelope) getEnvelope() const;
+    ref const(SCPStatement) getStatement() const;
+}
+
+alias SCPEnvelopeWrapperPtr = shared_ptr!SCPEnvelopeWrapper;
+
+extern (C++, class) public struct WrappedValuePtrComparator
+{
+extern(C++):
+    bool operator()(ref const(ValueWrapperPtr) l, ref const(ValueWrapperPtr) r) const;
```
Geod24: Should be `opCall`

source/scpd/tests/GlueTypes.d
```diff
@@ -59,6 +65,20 @@ alias TypesNoLayout = AliasSeq!
...
+    //ValueWrapperPtr,
+    //BallotProtocol.SCPBallotWrapper,
+    //SCPEnvelopeWrapper,
+    //ValueWrapper,
+
+    //LocalNode,
+    //BallotProtocol,
+    //NominationProtocol,
+    //SCP,
+    //Slot,
```
Geod24: Size is off ?

Geod24 (commented):
Did a first pass. I'm wondering, isn't the only reason we're adding SHA / Curve25519 because of the hashing behavior of Stellar ? IIRC the changes to get rid of those are minimal, as we don't need to match the hashes, just to provide unique hashes (and it's easier nowadays as we have the same hash size).

source/scpd/Cpp.d
AndrejMitrovic: Ah some of the instantiations only partially passed the parameters (not all of the template parameters). I'll use the template name trick though.

source/scpp/src/util/Logging.cpp
AndrejMitrovic: Debugging, will revert.

AndrejMitrovic:
> Did a first pass. I'm wondering, isn't the only reason we're adding SHA / Curve25519 because of the hashing behavior of Stellar ?

It seems there were some pointless imports to SHA.h in most of the modules. But if I remove them it still compiles. So I'll get rid of these modules.

source/agora/consensus/protocol/Nominator.d
AndrejMitrovic: It would be nice if the hashing routines had support for ranges. I want to be able to do something like this:

```D
const hash = hashFull(vals[].map!(val => val[]));
```

AndrejMitrovic: I wonder if I just have to flatten the range.. 🤔

AndrejMitrovic: There is a very strange error when passing ranges to `hashFull`:

```
submodules/crypto/source/agora/crypto/Hash.d(177,22): Error: expression `*record` is `void` and has no value
```

But this part of code is:

```D
// Pointers are handled as arrays, only their size must be 0 or 1
else static if (is(T : E*, E))
{
    if (record is null)
        hashVarInt(uint(0), hasher);
    else
    {
        hashVarInt(uint(1), hasher);
        hashPart(*record, hasher);
    }
}
```

How did a range got converted into a pointer? Could it be something to do with `in`?

AndrejMitrovic: Anyway I think we can add `isInputRange` as a special case in `hashPart` and it could work?

AndrejMitrovic: Hmm not really possible with input ranges because we only know the length after exhaustion of the input range, but the hash routine should hash the length first.

source/scpd/tests/GlueTypes.d
AndrejMitrovic: I've re-enabled most of them except two, which I haven't been able to match the size of on mac/linux reliably without guessing the size. Maybe @omerfirmak wants to take a look at these later.

AndrejMitrovic:
Most comments resolved!

source/scpd/Cpp.d
Geod24: Ping

source/scpd/quorum/QuorumIntersectionChecker.d
Geod24: Ping

source/scpd/Cpp.d
AndrejMitrovic: It's a new thing in SCP:

https://github.com/stellar/stellar-core/blob/master/src/util/UnorderedSet.h
https://github.com/stellar/stellar-core/blob/master/src/util/UnorderedMap.h

I haven't looked into why they switched things.

source/scpp/src/crypto/XDRHasher.h
```diff
@@ -0,0 +1,105 @@
...
+#include <xdrpp/endian.h>
+#include <xdrpp/marshal.h>
+
+namespace stellar
+{
+
+// CRTP base class to use in XDR-hashing xdrpp-compatible archiver objects. All
+// you need to provide is a `void Derived::hashBytes(unsigned char const*,
+// size_t)` member.
+template <typename Derived> struct XDRHasher
```
Geod24, line 16: Do we need this (and shorthash / siphash) ?

AndrejMitrovic: I think the problematic part is the QuorumIntersectionChecker. It's technically part of Herder and not SCP, so it brings in a lot of dependencies. For example:

ShortHash uses XdrHasher & SipHash
HashOfHash uses ShortHash
SecretKey uses HashOfHash
QuorumIntersectionChecker uses SecretKey

source/scpp/src/scp/LocalNode.h
```diff
@@ -9,6 +9,7 @@
 #include <vector>
 
 #include "scp/SCP.h"
+#include "util/HashOfHash.h"
```
Geod24, line 12: Okay so I was looking at this locally. This is what pulls in the hash & secretkey dependency AFAICS.
If you look at our fork of SCP, I actually removed it. I believe it could be removed upstream, but it is used in the implementation (the `.cpp`). 

However, I think those are just lingering import. If you remember, Stellar reduced dependencies by exposing a `sign` method, instead of relying on `SecretKey`. So I believe we should be able to remove all `SecretKey`-related code from our code, but we'll have to remove some (unused) imports.

Geod24 (approved):
Amazing work, thanks! Let's get it in, and remove the unused dependencies later...

Geod24:
Note that you didn't edit `dub.json` so all the new `.cpp` files are actually not linked in.

source/scpp/src/util/Math.cpp
```diff
@@ -17,20 +22,135 @@ rand_fraction()
...
+        return std::set<double>(points.begin(), points.end());
+    }
+
+    std::set<double> centroids;
+
+    auto backlog = points;
+
+    auto moveIndexToCentroid = [&](size_t index) {
+        // @BUG: undefined symbol printAssertFailureAndThrow()
+        // releaseAssertOrThrow(index < backlog.size());
```
Geod24, line 72: This can be uncommented if we add `GlobalChecks.o` to `dub.json`.
But I'd like to either get rid of it (and the `Backtace.cpp` dependency) or have it implemented in D.

PR #2066, opened 2021-05-07
Add default ctor, copy ctor and dtor for multiple CPP bindings

### LINKED ISSUE #753: Bindings for `std::unordered_map` are wrong
```c++
#include <unordered_map>
char (*__kaboom)[sizeof(std::unordered_map<int, int>)] = 1;
```
(Trick taken from https://stackoverflow.com/a/35261673/1021344)

This will print:
```console
clang++ uno.cpp
uno.cpp:3:8: error: cannot initialize a variable of type 'char (*)[40]' with an rvalue of type 'int'
char (*__kaboom)[sizeof(std::unordered_map<int, int>)] = 1;
       ^                                                 ~
1 error generated.
```

So the size is 40. However our bindings only have [*a single pointer*](https://github.com/bpfkorea/agora/blob/eb4ab3b0083c72fa2b7eb815ad2cd7a7290ffad4/source/scpd/Cpp.d#L85-L107).

That's on OSX with the latest clang++. I expect things to be wrong on other platforms too.

### LINKED ISSUE #2015: Switch SCP build from g++ to clang++

omerfirmak:
Fixes #753 
Fixes #2015

### DIFF
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -34,6 +34,76 @@ import std.meta;
 
 import vibe.data.json;
 
+public enum CppCtor { Use = 0 }
+
+extern(C++) {
+@trusted nothrow @nogc:
+    void defaultCtorCPPObject(T) (T* ptr);
+    void dtorCPPObject(T) (T* ptr);
+    void opAssignCPPObject(T) (T* lhs, T* rhs);
+    void copyCtorCPPObject(T) (T* ptr, inout(T)* rhs);
+    void copyCtorCPPObject(T) (immutable(T)* ptr, inout(T)* rhs);
+    int getCPPSizeof(T) ();
+}
+
+private mixin template CPPBindingMixin (T, bool Copyable = true, bool DefaultConstructable = false)
+{
+    extern(D)
+    {
+        static if (!DefaultConstructable) @disable this();
+        this (CppCtor use) @trusted nothrow @nogc
+        {
+            defaultCtorCPPObject!T(&this);
+        }
+
+        ~this () @trusted nothrow @nogc
+        {
+            dtorCPPObject!T(&this);
+        }
+
+        static if (Copyable)
+        this (ref return scope inout T rhs) @trusted nothrow @nogc
+        {
+            copyCtorCPPObject!T(&this, &rhs);
+        }
+
+        static if (Copyable)
+        this (ref return scope inout T rhs) immutable @trusted nothrow @nogc
+        {
+            copyCtorCPPObject!T(&this, &rhs);
+        }
+
+        static if (Copyable)
+        void opAssign()(auto ref T rhs) @trusted nothrow @nogc
+        {
+            opAssignCPPObject!T(&this, &rhs);
+        }
+    }
+}
+
+/// Can't import `core.stdcpp.allocator` because it transitively imports
+/// `core.stdcpp.exception`
+/// In this case we just need to get the name right for `vector`
+
+extern(C++, (StdNamespace)) extern(C++, class) struct allocator (T) {}
+
+extern(C++, (StdNamespace)) extern(C++, class) struct less (T) {}
+
+extern(C++, (StdNamespace)) extern(C++, class) struct default_delete (T) {}
+
+// simplistic std::pair bindings
+public extern(C++, (StdNamespace)) struct pair (T1, T2)
+{
+    T1 first;
+    T2 second;
+}
+
+// fake std::hash (for mangling)
+public extern(C++, (StdNamespace))  struct hash (T) {}
+
+// fake std::equal_to (for mangling)
+public extern(C++, (StdNamespace))  struct equal_to (T = void) {}
+
 extern(C++, (StdNamespace)) {
     /// Simple binding to `std::shared_ptr`
     extern(C++, class) struct shared_ptr (T)
@@ -43,19 +113,22 @@ extern(C++, (StdNamespace)) {
         else
             private alias TPtr = T*;
 
-        ~this() @safe {}
+        mixin CPPBindingMixin!(shared_ptr!T);
+
         TPtr ptr;
         void* _control_block;
         alias ptr this;
     }
 
     /// Simple binding to `std::unique_ptr`
-    extern(C++, class) struct unique_ptr (T)
+    extern(C++, class) struct unique_ptr (T, Deleter = default_delete!T)
     {
         static if (is(T == class) || is(T == interface))
             private alias TPtr = T;
         else
             private alias TPtr = T*;
+
+        mixin CPPBindingMixin!(unique_ptr!T, false);
 
         TPtr ptr;
         alias ptr this;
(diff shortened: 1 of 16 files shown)


source/scpd/Cpp.d
```diff
@@ -76,40 +79,39 @@ private nothrow @nogc extern(C++) void cpp_unordered_map_assign (K, V)(
...
-
-/// Create a new unordered map with the given Key/T types, and return a pointer to it
-private pure nothrow @nogc @safe extern(C++) void* cpp_unordered_map_create (K, V)();
-
 /// Rudimentary bindings for std::unordered_map
 extern(C++, (StdNamespace))
 public struct unordered_map (Key, T, Hash = hash!Key, KeyEqual = equal_to!Key, Allocator = allocator!(pair!(const Key, T)))
 {
-    private void* ptr;
+    private ulong[56 / ulong.sizeof] _data;
```
Geod24: `ulong.sizeof` is *always* 8. Did you mean `size_t` or `cpp_ulong` ?

AndrejMitrovic: How'd you get this number?

omerfirmak: It's a magic number, verified here

https://github.com/bosagora/agora/pull/2066/files#diff-2fff5f741967021f44722f88e37effa07123b287d13e843ff246b43d91a7d8d4R478

omerfirmak: Nah, I made that a `ulong[]` hoping that it would help with alignment and hard coding length to 7 just looked odd.

omerfirmak: It will probably be different for Mac OSX tho, CI will tell

AndrejMitrovic: Yes I'm having similar problems with https://github.com/bosagora/agora/pull/2065. There's a size mismatch on one platform (linux) but not another (macos). So much for C++ portability huh? :o)

omerfirmak: Our `shared_ptr`binding is faulty too, it causes C++ runtime to free the pointer we pass to it. Maybe this and https://github.com/bosagora/agora/pull/2016 could help you.

source/scpd/Cpp.d
```diff
@@ -76,40 +79,39 @@ private nothrow @nogc extern(C++) void cpp_unordered_map_assign (K, V)(
...
+
+    @disable this ();
+
+    extern(D):
+    this (char dummy)
+    {
+        defaultCtorCPPObject!(typeof(this))(&this);
+    }
```
Geod24: What's the reasoning for having a dummy ctor over, say, a `static` `make` function ?

source/scpp/extra/DUtils.h
```diff
@@ -74,3 +74,21 @@ VectorT duplicate(const VectorT& this_)
...
+void defaultCtorCPPObject (T* ptr)
+{
+    new(ptr) T();
+}
+
+template<typename T>
+void dtorCPPObject (T* ptr)
+{
+    ptr->~T();
+}
```
Geod24, line 94: Could https://github.com/dlang/druntime/blob/661c2d8975d5a5c6bfbab6b18850059aac857046/src/object.d#L3955-L3971 help ?

source/scpd/quorum/QuorumTracker.d
```diff
@@ -70,6 +70,3 @@ public:
     /// returns the known quorum map of the entire network
     ref const(QuorumMap) getQuorum () const;
 }
-
-// should be 48, see #753
-static assert(QuorumTracker.sizeof == 16);
```
Geod24, line 75: Why is this removed and not fixed ?

source/scpd/Cpp.d
omerfirmak: I think this lets you construct the object anywhere (stack/heap), without explicitly passing a pointer or copying the whole thing, right?

source/scpd/quorum/QuorumTracker.d
omerfirmak: While I think that does not really serve any purpose, I'll try to rewrite it better.

source/scpp/extra/DUtils.h
omerfirmak: I'll give it a try but in the cases where linker can't link `~object()` directly, that probably won't work either.

source/scpp/extra/DUtils.h
```diff
@@ -74,3 +74,21 @@ VectorT duplicate(const VectorT& this_)
...
+template<typename T>
+int getCPPSizeof ()
+{
+    return sizeof(T);
+}
+
+template<typename T>
+void defaultCtorCPPObject (T* ptr)
+{
+    new(ptr) T();
```
linked0, line 87: Is it the same as follows?
`new T(ptr);`

linked0: Ah, there was some misunderstanding. Never mind. :)

omerfirmak:
All green

omerfirmak:
Apart from `std::vector`, all containers now have proper copy ctor and `opAssign`

source/scpd/Cpp.d
```diff
@@ -34,7 +34,72 @@ import std.meta;
...
+    void copyCtorCPPObject(T) (T* ptr, inout(T)* rhs);
+    void copyCtorCPPObject(T) (immutable(T)* ptr, inout(T)* rhs);
+    int getCPPSizeof(T) ();
+}
+
+private mixin template CPPBindingMixin (T)
+{
+    extern(D)
+    {
+        this (char dummy) @trusted nothrow @nogc
```
AndrejMitrovic: Would it be a good idea to add `@disable this();` so people remember to call this ctor?

source/scpd/types/XDRBase.d
```diff
@@ -45,48 +45,32 @@ uint size32(size_t s)
 ////////////////////////////////////////////////////////////////
 
 //! Generic class of XDR unmarshaling errors.
-class xdr_runtime_error : runtime_error
-{
-}
```
Geod24: Why ? `alias` is almost like a `#define` for type, so it would e.g. give the wrong name if used as template parameter.

source/scpd/Cpp.d
Geod24: Yes to `@disable this();`

I would also suggest creating a special enum and only using this, e.g. `enum CppCtor { Use = 0; }` and then:
```D
public this (CppCtor use);
auto x = CppType(CppCtor.Use);
```

Geod24 (commented):
How does the second commit work given we can't pass `CppRuntime_Clang` to LDC ?

source/scpd/types/XDRBase.d
omerfirmak: Reverted those

source/scpd/Cpp.d
omerfirmak: Yep, default ctor was meant to be disabled.

omerfirmak:
> How does the second commit work given we can't pass `CppRuntime_Clang` to LDC ?

Well, it just changes the compiler but we still use `libstd++` on linux.

source/scpd/types/XDRBase.d
omerfirmak: and it immediately broke macos.

Geod24:
```
  "std::__1::logic_error::what() const", referenced from:
      __D4scpd5types7XDRBase15xdr_wrong_union6__vtblZ in agora.o
      __D4scpd3Cpp11logic_error6__vtblZ in agora.o
  "std::__1::runtime_error::what() const", referenced from:
      __D4scpd5types7XDRBase17xdr_runtime_error6__vtblZ in agora.o
      __D4scpd5types7XDRBase12xdr_overflow6__vtblZ in agora.o
      __D4scpd5types7XDRBase18xdr_stack_overflow6__vtblZ in agora.o
      __D4scpd5types7XDRBase20xdr_bad_message_size6__vtblZ in agora.o
      __D4scpd5types7XDRBase20xdr_bad_discriminant6__vtblZ in agora.o
      __D4scpd5types7XDRBase18xdr_should_be_zero6__vtblZ in agora.o
      __D4scpd5types7XDRBase20xdr_invariant_failed6__vtblZ in agora.o
```

Don't think Exceptions are in `__1` are they ?

Geod24:
```
[source/agora/test/Flash.d:1343] Assertion thrown during test: assert(info.state == ChannelState.SettingUp || info.state == ChannelState.WaitingForFunding) failed
/Users/runner/work/agora/agora/source/agora/test/Flash.d:1343 void agora.test.Flash.__unittest_L1211_C1() [0x11006133e]
/Users/runner/work/agora/agora/source/agora/test/Flash.d agora.test.Flash.__unittest [0x110079c5e]
/Users/runner/work/agora/agora/source/agora/test/Base.d:264 bool agora.test.Base.customModuleUnitTester().runTest(agora.test.Base.customModuleUnitTester().ModTest) [0x110099a60]
/Users/runner/work/agora/agora/source/agora/test/Base.d:283 core.runtime.UnitTestResult agora.test.Base.customModuleUnitTester() [0x11009835c]
runtime.d:607 runModuleUnitTests [0x111109a28]
dmain2.d:551 void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).runAll() [0x1111236e2]
dmain2.d:525 void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x111123662]
dmain2.d:584 _d_run_main2 [0x11112351d]
dmain2.d:384 _d_run_main [0x1111231d4]
/Users/runner/hostedtoolcache/dc/ldc2-1.26.0/x64/ldc2-1.26.0-osx-x86_64/bin/../import/core/internal/entrypoint.d:42 main [0x1100a4774]
```

Never seen this one before

omerfirmak:
green after multiple tries

omerfirmak:
Needs approval too @Geod24

Geod24 (approved):

PR #2072, opened 2021-05-11
Remove two Exception-swallowing `scope (failure)`

Geod24:
We're seeing the assert in `UTXODB.find` triggered in the live system.
However, due to the usage of `scope (failure)`, we don't actually know what the error is.

### DIFF
--- a/source/agora/consensus/state/UTXODB.d
+++ b/source/agora/consensus/state/UTXODB.d
@@ -83,22 +83,31 @@ package class UTXODB
 
     public bool find (in Hash key, out UTXO value) @trusted nothrow
     {
-        scope(failure) assert(0);
-        auto results = this.db.execute(
-            "SELECT unlock_height, type, amount, locktype, lock FROM utxo WHERE hash = ?", key);
-
-        foreach (row; results)
+        try
         {
-            auto unlock_height = Height(row.peek!ulong(0));
-            auto type = row.peek!TxType(1);
-            // DMD BUG: Cannot construct the object directly, `inout` bug
-            Output output;
-            output.value = Amount(row.peek!ulong(2));
-            output.lock  = Lock(row.peek!(LockType)(3), row.peek!(ubyte[])(4));
-            value = UTXO(unlock_height, type, output);
-            return true;
+            auto results = this.db.execute(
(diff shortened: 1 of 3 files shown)


source/agora/consensus/state/UTXODB.d
```diff
@@ -83,22 +83,31 @@ package class UTXODB
...
-            // DMD BUG: Cannot construct the object directly, `inout` bug
-            Output output;
-            output.value = Amount(row.peek!ulong(2));
-            output.lock  = Lock(row.peek!(LockType)(3), row.peek!(ubyte[])(4));
-            value = UTXO(unlock_height, type, output);
-            return true;
+            auto results = this.db.execute(
+                "SELECT unlock_height, type, amount, locktype, lock FROM utxo WHERE hash = ?", key);
+
+            foreach (row; results)
```
hewison-chris, line 91: Is there any reason why we use foreach here and not the following:-
`if (!results.empty && results.oneValue!(byte[]).length != 0)`

hewison-chris (approved):
LGTM

source/agora/consensus/state/UTXODB.d
Geod24: Nothing I can think of ATM.

source/agora/consensus/state/UTXODB.d
```diff
@@ -83,22 +83,31 @@ package class UTXODB
...
+                value = UTXO(unlock_height, type, output);
+                return true;
+            }
+            return false;
+        }
+        catch (Exception exc)
+        {
+            import std.stdio;
+            try writeln("Exception thrown in UTXODB.find: ", exc);
+            catch (Exception exc) { assert(0, "Last chance writeln thrown: Giving up"); }
```
kchulj, line 108: Is it common practice to have a `writeln` `try catch` block inside of `catch`?

Geod24: Well if `writeln` was `nothrow` we wouldn't need the `try` / `catch`

kchulj (approved):
LGTM

linked0 (approved):

Geod24:
Finally. This was very confusing because the `OnInvalidMemoryOperation` was actually triggered in Faucet (most likely because shutdown is broken), but that wasn't clear to me at the time. Should be G2G.

PR #2080, opened 2021-05-12
Registry with UTXO instead of PublicKey

### LINKED ISSUE #1309: Change registries (network, test) to use the UTXO instead of the public key
Spin off from #849

As discussed, public keys are not currently unique, while utxo will always be. When referring the nodes individually, we should then use the UTXO instead of the public key. This will allow to prevent mishap if someone ever does something funny with enrollments.

linked0:
When referring to the nodes individually, we should then use the UTXO instead of the public key because public keys are not currently unique.

Relates to #1309

### DIFF
--- a/source/agora/flash/Node.d
+++ b/source/agora/flash/Node.d
@@ -1562,11 +1562,23 @@ public class AgoraFlashNode : FlashNode
 
         foreach (pair; this.getManagedKeys())
         {
+            // TODO: This should use UTXO instead of PublicKey
+            // We should change `managed_keys` to be a set of UTXO hash
+            Hash utxo_key;
+            UTXO utxo;
+            this.network.findUTXO(pair.key, utxo_key, utxo);
+            if(utxo_key == Hash.init)
+            {
+                log.fatal("Couldn't find the UTXO for this node: {}.", pair.key);
+                assert(0);
+            }
+
             RegistryPayload payload =
             {
                 data:
                 {
-                    public_key : pair.key,
+                    utxo_key : utxo_key,
+                    utxo : utxo,
                     addresses : addresses,
                     seq : time(null)
                 }
@@ -1625,7 +1637,13 @@ public class AgoraFlashNode : FlashNode
 
         log.info("getFlashClient searching peer: {}", peer_pk);
 
-        auto payload = this.registry_client.getValidator(peer_pk);
+        // TODO: This should use UTXO instead of PublicKey
+        // We should change `managed_keys` to be a set of UTXO hash
+        Hash utxo_key;
+        UTXO utxo;
+        this.network.findUTXO(peer_pk, utxo_key, utxo);
+        auto payload = this.registry_client.getValidator(utxo_key);
+
         if (payload == RegistryPayload.init)
         {
             log.warn("Could not find mapping in registry for key {}", peer_pk);
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -255,6 +255,7 @@ public class FullNode : API
         this.utxo_set = this.makeUTXOSet();
         this.pool = this.makeTransactionPool();
         this.enroll_man = this.makeEnrollmentManager();
+        this.network.setEnrollmentUTXOGetter(&this.getEnrollmentUTXO);
         this.transaction_relayer = this.makeTransactionRelayer();
         const ulong StackMaxTotalSize = 16_384;
         const ulong StackMaxItemSize = 512;
@@ -594,6 +595,28 @@ public class FullNode : API
                 commons_budget,
                 config.consensus,
                 config.node.block_interval_sec.seconds);
+    }
+
+    /***************************************************************************
+
+        Get the enrolled UTXO of this validator
+
+        Params:
+            utxo_key = key for a UTXO used in enrollment
+            utxo = will contain the UTXO for the `utxo_key`
+
+    ***************************************************************************/
+
+    public void getEnrollmentUTXO (ref Hash utxo_key, out UTXO utxo)
+        @trusted nothrow
+    {
+        if (utxo_key == Hash.init)
+            utxo_key = this.enroll_man.getEnrollmentKey();
+
+        if (utxo_key == Hash.init)
+            return;
+
+        this.utxo_set.peekUTXO(utxo_key, utxo);
     }
 
     /// Returns a newly constructed StatsServer
--- a/source/agora/registry/API.d
+++ b/source/agora/registry/API.d
@@ -14,6 +14,7 @@
 module agora.registry.API;
 
 import agora.common.Types;
+import agora.consensus.data.UTXO;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.crypto.Schnorr: Signature;
@@ -24,8 +25,11 @@ import vibe.web.rest;
 ///
 public struct RegistryPayloadData
 {
-    /// the public key that we want to register
-    public PublicKey public_key;
+    /// the hash of the UTXO that we want to register
+    public Hash utxo_key;
+
+    /// the UTXO that we want to register
+    public UTXO utxo;
 
     /// network addresses associated with the public key
     public const(Address)[] addresses;
@@ -66,8 +70,7 @@ public interface NameRegistryAPI
         Get network addresses corresponding to a public key
 
         Params:
-            public_key = the public key that was used to register
-                         the network addresses
+            key = the key that was used to register the network addresses
 
         Returns:
             Network addresses associated with the `public_key`
@@ -77,7 +80,7 @@ public interface NameRegistryAPI
 
     ***************************************************************************/
 
-    public const(RegistryPayload) getValidator (PublicKey public_key);
+    public const(RegistryPayload) getValidator (Hash hash);
 
     /***************************************************************************
 
--- a/source/agora/registry/Server.d
+++ b/source/agora/registry/Server.d
@@ -31,7 +31,7 @@ mixin AddLogger!();
 public final class NameRegistry: NameRegistryAPI
 {
     ///
-    private RegistryPayload[PublicKey] registry_map;
+    private RegistryPayload[Hash] registry_map;
 
     /// Validator count stats
     private RegistryStats registry_stats;
@@ -47,8 +47,7 @@ public final class NameRegistry: NameRegistryAPI
         Get network addresses corresponding to a public key
 
         Params:
-            public_key = the public key that was used to register
-                         the network addresses
+            key = the key that was used to register the network addresses
 
         Returns:
             Network addresses associated with the `public_key`
@@ -58,9 +57,9 @@ public final class NameRegistry: NameRegistryAPI
 
     ***************************************************************************/
 
-    public override const(RegistryPayload) getValidator (PublicKey public_key)
+    public override const(RegistryPayload) getValidator (Hash key)
     {
-        if (auto payload = public_key in registry_map)
+        if (auto payload = key in registry_map)
             return *payload;
         return RegistryPayload.init;
     }
@@ -85,18 +84,18 @@ public final class NameRegistry: NameRegistryAPI
     public override void putValidator (RegistryPayload registry_payload)
     {
         // verify signature
-        if (!registry_payload.verifySignature(registry_payload.data.public_key))
+        if (!registry_payload.verifySignature(registry_payload.data.utxo.output.address))
             throw new Exception("incorrect signature");
 
         // check if we received stale data
-        if (auto previous = registry_payload.data.public_key in registry_map)
+        if (auto previous = registry_payload.data.utxo_key in registry_map)
             if (previous.data.seq > registry_payload.data.seq)
                 throw new Exception("registry already has a more up-to-date version of the data");
 
         // register data
         log.info("Registering network addresses: {} for public key: {}", registry_payload.data.addresses,
-            registry_payload.data.public_key.toString());
-        registry_map[registry_payload.data.public_key] = registry_payload;
+            registry_payload.data.utxo.output.address.toString());
+        registry_map[registry_payload.data.utxo_key] = registry_payload;
         this.registry_stats.setMetricTo!"registry_record_count"(registry_map.length);
     }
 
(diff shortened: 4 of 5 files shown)


source/agora/registry/API.d
```diff
@@ -43,6 +47,25 @@ public struct RegistryPayload
...
+        Params:
+            pubkey = the address used to make a key for a `RegistryPayload`
+            utxo = the UTXO used to make a key for a `RegistryPayload`
+
+        Returns:
+            a key of `Hash` type
+
+    ***************************************************************************/
+
+    static public Hash makeRegistryKey (in PublicKey pubkey, in UTXO utxo)
```
ferencdg: do we really need to has the pubkey and utxo together? utxo has an output, and if the utxo is a payment transaction (as opposed to be a script), then the output should contain the pubkey(or its hash)

source/agora/node/Validator.d
```diff
@@ -449,6 +449,21 @@ public class Validator : FullNode, API
...
 
+    /// Ditto
+    public override UTXO getValidatorUTXO (in PublicKey pubkey) @trusted nothrow
+    {
+        auto hash = this.enroll_man.validator_set.getEnrolledUTXO(
+                this.ledger.getBlockHeight(), pubkey);
+        if (hash != Hash.init)
+        {
+            UTXO utxo;
+            this.utxo_set.peekUTXO(hash, utxo);
```
ferencdg: peekUTXO has a bool return value to indicate if the peeking was successful

source/agora/registry/API.d
linked0: Good point! But there are some cases where `outsider_validators` don't yet have UTXOs in unit tests. I will consider another way to deal with it. Ty.

source/agora/node/Validator.d
```diff
@@ -449,6 +449,21 @@ public class Validator : FullNode, API
         return new AdminInterface(this.config.validator.key_pair, this.clock);
     }
 
+    /// Ditto
+    public override UTXO getValidatorUTXO (in PublicKey pubkey) @trusted nothrow
+    {
+        auto hash = this.enroll_man.validator_set.getEnrolledUTXO(
+                this.ledger.getBlockHeight(), pubkey);
```
ferencdg: here we are referring to the block height as 'this.ledger.getBlockHeight()' - but how would this work with a validator that is catching up, and still doesn't have the correct hight?

source/agora/node/FullNode.d
```diff
@@ -534,6 +535,24 @@ public class FullNode : API
...
+
+        Returns:
+            An UTXO if there is a validator for the public key,
+            otherwise UTXO.init
+
+    ***************************************************************************/
+
+    public UTXO getValidatorUTXO (in PublicKey pubkey) @trusted nothrow
+    {
+        assert(0);
```
ferencdg: just a small thing: can we have assert(0, "some string here") - when we will switch to release build we might not have access to the stack trace, so we might not know what the actual error was

source/agora/node/Validator.d
linked0: What do you mean? I used the function to get a UTXO for the hash.

ferencdg: > What do you mean? I used the function to get a UTXO for the hash.

what if 
this.utxo_set.peekUTXO(hash, utxo); 
returns false? it would mean 
utxo == UTXO.init and we would return that

linked0: You mean the line of `if (hash != Hash.init)` is not needed?

source/agora/registry/Server.d
```diff
@@ -84,19 +83,22 @@ public final class NameRegistry: NameRegistryAPI
 
     public override void putValidator (RegistryPayload registry_payload)
```
ferencdg, line 84: I was thinking a bit more about this:
let's say an attacker(node A) with a pubkey A signs the payload and the payload contains
1. the attacker pubkey A
2. UTXO X where UTXO X is actually a UTXO for node B

Currently we only check if the payload is properly signed, but we never check if UTXO X belongs to node B or node A

For this reason it might make sense to include the pubkey into that hash(or add the extra check as above which is what I think we should do)

source/agora/node/Validator.d
linked0: I will check the returned value with assert. Ty.

source/agora/node/FullNode.d
linked0: Updated. There was a mistake. The code in `Validator` was in fact moved into this function because the `Validator` doesn't need to have it. Ty.

source/agora/node/Validator.d
linked0: I don't think it could be a big problem because even if there are multiple `RegistryPayload`s for a `PublicKey` in the registry, they will indicate the same address. I hope this explanation makes sense to you. :)

ferencdg: > I don't think it could be a big problem because even if there are multiple `RegistryPayload`s for a `PublicKey` in the registry, they will indicate the same address. I hope this explanation makes sense to you. :)

they can actually indicate different addresses, if the validator's IP changed :) However we do authenticate validators during the handshake, so I agree this is not a big problem.

source/agora/registry/API.d
```diff
@@ -43,6 +47,25 @@ public struct RegistryPayload
...
+        @safe nothrow @nogc
+    {
+        return hashMulti(pubkey, utxo);
+    }
+
```
Geod24: I don't understand why we hash both together. The UTXO hash should be enough.

source/agora/registry/API.d
```diff
@@ -24,7 +25,10 @@ import vibe.web.rest;
 ///
 public struct RegistryPayloadData
 {
-    /// the public key that we want to register
+    /// the UTXO that we want to register
+    public UTXO utxo;
+
+    /// the public key that we want to register if there's no UTXO
     public PublicKey public_key;
```
Geod24: Since the `UTXO` contains the `PublicKey`, the second field is redundant.

source/agora/consensus/state/ValidatorSet.d
```diff
@@ -325,6 +325,43 @@ public class ValidatorSet
...
+            else
+                return Hash.init;
+        }
+        catch (Exception ex)
+        {
+            log.fatal("Exception occured on getEnrolledUTXO: {}, " ~
+                "PublicKey for enrollment: {}", ex.msg, pubkey);
+            assert(0);
+        }
+    }
```
Geod24: We shouldn't need such a function. Its presence indicates that the switch is not complete, and we're still using `PublicKey` instead of `UTXO` / `Hash` in some places.
Additionally, it turns a "recoverable" error (`Exception`) into an unrecoverable one (`assert(0)`). I think we should leave the caller of this function decide whether or not the error is recoverable.

source/agora/flash/Node.d
```diff
@@ -1595,10 +1595,17 @@ public class AgoraFlashNode : FlashNode
 
         foreach (pair; this.getManagedKeys())
         {
+            UTXO utxo = this.network.getValidatorUTXO(pair.key);
+            if(utxo == UTXO.init)
+            {
+                log.trace("Couldn't find the UTXO for this node: {}.", pair.key);
+            }
```
Geod24: That looks buggy. If we can't find the `utxo`, we send a payload with `UTXO.init`.

Geod24 (commented):
This does 80% of the change, I believe. I'll have a look at it locally.

source/agora/registry/API.d
linked0: It's because there are cases where `outsider` validators don't have any UTXO in unit tests.

source/agora/registry/API.d
ferencdg: I also mentioned this in a previous comment, but if we use only the UTXO as the key, we might need to do some extra checks: https://github.com/bosagora/agora/pull/2080#discussion_r631492163

source/agora/registry/API.d
ferencdg: outsider validators shouldn't register themselves I think

linked0: Currently, the outsider validators register themselves. But I know we have a plan to remove outsider validators in our test codes.

source/agora/consensus/state/ValidatorSet.d
linked0: I will implement another approach soon without introducing a new function like this.

linked0:
How about this approach? Please leave your opinions. I will fix some nits anyway.

@Geod24, I think we need to have a function to get UTXOs for a public key because the `required_peer_keys` parameter of the `discover` function of `NetworkManager` is a set of `PublicKey`s. Do you think we should change the `required_peer_keys` should be `Set!Hash(UTXO)` instead of `Set!PublicKey`?

source/agora/flash/Node.d
```diff
@@ -1595,10 +1595,17 @@ public class AgoraFlashNode : FlashNode
 
         foreach (pair; this.getManagedKeys())
         {
+            UTXO utxo = this.network.getEnrollmentUTXO();
+            if(utxo == UTXO.init)
+            {
+                log.trace("Couldn't find the UTXO for this node: {}.", pair.key);
+            }
+
```
Geod24: Urgh, the flash layer really shouldn't piggyback on the registry here... I think this will break the flash layer big time.

linked0: Ah, Sorry. This is the intermediate code, not the final code. I just wanted to upload my intermediate code into GitHub.

linked0:
I'm working on the issue by starting to add the following code in the third commit.
```
package(scpd)
struct UTXOHash {
    extern(D) this(uint512 value) @safe pure nothrow @nogc
    {
        this.utxo_hash_ = value;
    }

    extern(D) this(typeof(this.tupleof) args) @safe pure nothrow @nogc
    {
        this.tupleof = args;
    }

    int32_t type_;
    uint512 utxo_hash_;
    alias utxo_hash_ this;
}

... // Omitted
alias NodeID = UTXOHash;
```
By adding the above code and the related code, there are so many errors. And I'm working on the errors.

@Geod24  Could you leave your opinion about the third commit in this PR and give me some guidance about how to change code in the Stellar part?
I'm wondering I'm going in the right direction on this issue. It would be good to have a whiteboard soon.

source/agora/common/Config.d
```diff
@@ -939,15 +939,15 @@ private auto get (string section, string name, Converter) (
 public SCPQuorumSet toSCPQuorumSet (in QuorumConfig quorum_conf) @safe nothrow
 {
     import std.conv;
-    import scpd.types.Stellar_types : uint256, NodeID;
+    import scpd.types.Stellar_types : uint256, uint512, NodeID;
```
Geod24: You can remove the uint256 import I think.

source/agora/node/Validator.d
```diff
@@ -143,8 +143,8 @@ public class Validator : FullNode, API
         static QuorumConfig[] other_qcs;
         this.rebuildQuorumConfig(this.qc, other_qcs, height);
         this.nominator.setQuorumConfig(this.qc, other_qcs);
-        buildRequiredKeys(this.config.validator.key_pair.address, this.qc,
-            this.required_peer_keys);
+        buildRequiredKeys(this.config.validator.key_pair.address,
+            this.qc, this.required_peer_keys, this.required_peer_utxos);
```
Geod24: I think we can / should remove `required_peer_keys`

source/scpd/scp/SCPDriver.d
```diff
@@ -148,11 +148,11 @@ nothrow:
 
         // `toStrKey` returns StrKey encoded string representation
         //std::string toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
-        void* toStrKey(ref const(PublicKey) pk, bool fullKey = true) const;
+        void* toStrKey(ref const(UTXOHash) pk, bool fullKey = true) const;
```
Geod24: Don't think that's going to work, this is for keys, not hashes

source/scpd/types/Stellar_types.d
```diff
@@ -68,6 +72,23 @@ struct PublicKey {
...
+
+    extern(D) this(typeof(this.tupleof) args) @safe pure nothrow @nogc
+    {
+        this.tupleof = args;
+    }
+
+    int32_t type_;
+    uint512 utxo_hash_;
+    alias utxo_hash_ this;
+}
```
Geod24: I think this should be handled one level higher, e.g. the `PublicKey` type has to go.

Geod24 (commented):
Lots of good stuff, but I think we might need to abstract things away on the SCP side a bit more. Stellar used to have all the crypto operations on the SCP layer, they abstracted it away somewhere between v11.2.0 and v16.0.0, but there are some lingering types.

linked0:
Ready for review.
Issue #1309  needs to make the flash nodes use UTXO also instead of PublicKey as the next phase. I will be working on that and submit another issue.

Geod24:
So my recent foray into SCP came from this, as I realized we could greatly simplify our interfacing.
One other thing that came up while reviewing this is that the Flash layer piggyback on the key registration of the registry, which is problematic for us, because it means 1) this change cannot be completed, and 2) the registry is an obvious DoS target.

source/agora/consensus/state/ValidatorSet.d
```diff
@@ -55,7 +55,7 @@ public alias EnrollmentFinder = bool delegate (in Hash enroll_key, out Enrollmen
 
 /// A Height and PublicKey pair to represent expiring Validators
 public alias ExpiringValidator = Tuple!(Height, "enrolled_height",
-    PublicKey, "pubkey");
+    Hash, "utxo", PublicKey, "pubkey");
```
Geod24: Can you extract those changes (add8fd7a6a59f589abb60561dece48298e3e98bd / source/agora/consensus/state/ValidatorSet.d) in their own PR ? We can merge it immediately.

linked0: Sure, I will.

Geod24:
Pretty nice BTW! The first commit does point to a few issues we have, but with a few SCP changes, it should look much nicer.

Geod24:
Can you submit the first commit as its own PR ? Then we can just switch SCP.

linked0:
> Can you submit the first commit as its own PR ? Then we can just switch SCP.

Sure.

Geod24:
After more research and discussion, we decided not to go this way.

PR #2083, opened 2021-05-13
Changed NetworkTopology schemantic

### LINKED ISSUE #2078: makeTestNetwork doesn't properly connect nodes
I might be mistaken, as I didn't test, but it seems 'makeNetworkConfig' makes a network of validators where each validator are connected and also a separate network where all the FullNodes are connected, but these two network doesn't seem to have any common node. Not what I would expect when using NetworkTopology.FullyConnected

ferencdg:
1. FullyConnected: Enrolled validators can now **potentially** connect to all full nodes without discovery
2. MinimallyConnected: Enrolled validators and full nodes are now
properly connected to form a ring

fixes: #2078

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -2131,26 +2131,26 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         .takeExactly(GenesisValidators)
         .map!(en => validatorAddress(en.index, en.value)).array;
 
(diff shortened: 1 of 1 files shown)


AndrejMitrovic (approved):

linked0 (approved):

ferencdg:
fixing some bugs, hold-on pls

ferencdg:
should be ok now, only the MinimallyConnected was affected

Geod24 (approved):

PR #2084, opened 2021-05-13
Fold SlashPolicy into the Ledger

Geod24:
The separation was meant to increase encapsulation but it didn't turn out to do that,
instead we had two tightly coupled classes.


linked0 (approved):
LGTM. There was not a great benefit in separating a class for the slashing policy.

PR #2086, opened 2021-05-13
Ledger.acceptBlock: Properly check for the random seed

Geod24:
The Ledger had two methods with similar name and use: getRandomSeed and getExternalizedRandomSeed.
The former was only used in Ledger.acceptBlock, while the later was used everywhere else the random
seed was needed. Upon closer inspection, it became obvious that 'getRandomSeed' was using
the presently-known Ledger state while getExternalizedRandomSeed had a much more predictable output.
In essence, 'getRandomSeed' could have been 'getCandidateRandomSeed'. This discrepancy could create
issues if a block which slashes a validator was externalized, but we received the pre-image in the
meantime, leading to a different random seed. With the removal of the former 'getRandomSeed' function,
we now have a predictable output (new state won't interfere).
In the process, the verbosely-name 'getExternalizedRandomSeed' was renamed to 'getRandomSeed'.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -534,7 +534,7 @@ extern(D):
                 return;
             }
 
-            Hash random_seed = this.ledger.getExternalizedRandomSeed(
+            Hash random_seed = this.ledger.getRandomSeed(
                 last_block.header.height + 1, con_data.missing_validators);
 
             Transaction[] received_tx_set;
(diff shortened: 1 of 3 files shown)


Geod24:
The two first commits aren't required but show be no brainers.

hewison-chris (approved):
LGTM, I guess I will have some conflicts to resolve in my PR which also overlaps some of the changes here.

linked0 (approved):
LGTM

PR #2087, opened 2021-05-13
Added timing for unit tests

ferencdg:


linked0:
Please add commit message about why timing is needed. I know you intend to check the time a unit test takes but we can know what the commit is for without reading code.

Geod24:
If it's for the CI, you don't really need it:
Screen Shot 2021-05-13 at 18 38 19

ferencdg:
> If it's for the CI, you don't really need it:
> Screen Shot 2021-05-13 at 18 38 19

I actually planned to use this during development, we have test like Simple test that runs 40 seconds, that shouldn't happen for something 'simple'... I hoped that after adding timing, we will be more conscious about how we write the network tests and how long they run.

Geod24 (approved):

PR #2088, opened 2021-05-13
Update Stellar's xdrpp to HEAD

Geod24:
Matches commit 781c0a53a245e3cf1d9f874ff5709fc5ae23d83b which unbreaks the build
with the latest Github image.


Geod24:
Self merging because the CI passed and HEAD is broken.

PR #2089, opened 2021-05-14
Nominator: Externalize block via the node's acceptBlock method

Geod24:
Before this change, the Nominator would directly externalize things through the Ledger.
The major downside of this approach is that any kind of special handling done by the
FullNode.acceptBlock (and Validator.acceptBlock) method is bypassed.
This became obvious then the stats were moved out of the Ledger and were not recorded anymore,
but it also means that the network whitelisting code of the FullNode was bypassed.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -136,6 +136,9 @@ public extern (C++) class Nominator : SCPDriver
     public extern (D) void delegate (in ConsensusData data, in string msg)
         @safe onInvalidNomination;
 
+    /// Delegate called when a block is to be externalized
+    public extern (D) bool delegate (const ref Block) @safe acceptBlock;
+
     /// Nomination start time
     protected TimePoint nomination_start_time;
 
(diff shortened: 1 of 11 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -136,6 +136,9 @@ public extern (C++) class Nominator : SCPDriver
     public extern (D) void delegate (in ConsensusData data, in string msg)
         @safe onInvalidNomination;
 
+    /// Delegate called when a block is to be externalized
+    public extern (D) bool delegate (const ref Block) @safe externalizeBlock;
```
linked0: Isn't the name of `acceptBlock` more proper because the `acceptBlock` of `FullNode` or `Validator` is called in the middle of externalizing the value?

Geod24: Yeah I guess it would be less confusing

Geod24:
Updated:
- `externalizeBlock` => `acceptBlock`
- `assert` for not `null`
- Comment about why it's important

hewison-chris (approved):
LGTM

linked0 (approved):
LGTM

Geod24:
```
Unittesting agora.test.Flash..
#### FATAL ERROR: Connection with peer closed
This node was started at source/agora/test/Flash.d:371
Full error: object.Exception@submodules/localrest/source/geod24/LocalRest.d-mixin-1129(1140): Connection with peer closed
```

🤔

kchulj (approved):

PR #2092, opened 2021-05-16
Nominator: Be more verbose when we are not nominating

Geod24:
This should help better debug the reason(s) why a node is not nominating.
Currently we only see a bunch of 'Nominating X at Y" but nodes can get stuck.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -315,12 +315,15 @@ extern(D):
     {
         this.ledger.prepareNominatingSet(data, MaxTransactionsPerBlock, this.nomination_start_time);
         if (data.tx_set.length < 1)
-            return false;  // not ready to nominate yet
+        {
+            this.log.trace("prepareNominatingSet(): No transaction to nominate yet");
+            return false;
+        }
 
         // check whether the consensus data is valid before nominating it.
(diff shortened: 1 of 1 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -314,12 +314,15 @@ extern(D):
     {
         this.ledger.prepareNominatingSet(data, MaxTransactionsPerBlock, this.nomination_start_time);
         if (data.tx_set.length < 1)
-            return false;  // not ready to nominate yet
+        {
+            this.log.trace("checkNominate(): No transaction to nominate yet");
```
hewison-chris:
```suggestion
            this.log.trace("prepareNominatingSet(): No transaction to nominate yet");
```

hewison-chris (changes requested):
LGTM just a tiny change requested

linked0 (approved):
LGTM

source/agora/consensus/protocol/Nominator.d
Geod24: Good catch!

Geod24:
Fixed method name

hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

PR #2094, opened 2021-05-17
NetworkClient: Use separate logger per validator

Geod24:
This coupled with the ability to dynamically (re)configure logger will
make it easier to track down what is going on with a particular peer.


hewison-chris (approved):
Nice

PR #2097, opened 2021-05-17
Network crawler for publicly available network map

### LINKED ISSUE #251: Publicly available Network map (backend)
As a user, I need to be able to have a view of the network, geographically.

Some of that might come from the monitoring tools ( #255 ) but it could be interesting to integrate that in the advanced administrative interface, or a user-project (see the links in #255 ).
If this is part of the administrative interface, we should explore integrating [geoip](https://www.maxmind.com/en/geoip-demo).

ferencdg:
fixes: #251


source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
...
+    /// a particular node. This maps supplements `discovered_node_list` and
+    /// was added because:
+    ///
+    /// 1. Iterating over `discovered_node_list` is slow.
+    /// 2. Iterating over `discovered_node_list` cannot reliably answer, whether
+    ///    a node has been discovered, as crawling fibers pop elements off of it.
+    private NetworkClientHolder[Address] discovered_node_map;
+
+    /// Contains the crawling results. A particular node is removed from this
+    /// data structure, after it is remporarily banned.
```
hewison-chris:
``` suggestion
    /// data structure, after it is temporarily banned.
```

hewison-chris (approved):
LGTM, just one typo.

source/agora/common/BanManager.d
```diff
@@ -126,12 +126,13 @@ public class BanManager
     ***************************************************************************/
 
     public this (Config config, Clock clock, cstring data_dir,
-                 Logger logger = Logger(ThisModule)) @safe nothrow pure
+        string ban_file_name = "banned.dat",
```
Geod24: Why is the file name passed ? It doesn't make sense to pass both the data dir and the file name, pass either or pass the full path.

source/agora/common/Set.d
```diff
@@ -39,6 +39,13 @@ public struct Set (T)
         this._set[key] = true;
     }
 
+    /// Put all elements in a range into the set
+    public void putRange (R) (R range)
+    {
+        foreach (T item; range)
+            this.put(item);
+    }
+
```
Geod24: You can just do `range.each!(e => set.put(e))`.

source/mmdb/MaxMindDB.d
```diff
@@ -0,0 +1,268 @@
+/*******************************************************************************
+    DLang bindings for MaxMindDB C library generated with jacob-carlborg/dstep.
```
Geod24: Nit: Newline after the `/***...`

source/mmdb/MaxMindDB.d
```diff
@@ -0,0 +1,268 @@
...
+
+    The following simple changes were made after generating the bindings
+
+    1. type `mmdb_uint128_t` is defined as ubyte[16], becuase DLang currently
+       doesn't support 16 byte integers natively yet
```
Geod24: "because" and grammar wise, either currently or yet, not both. (e.g. "because Dlang doesn't support 16 bytes native integers yet").

source/mmdb/MaxMindDB.d
```diff
@@ -0,0 +1,268 @@
...
+
+    2. union member of MMDB_entry_data_s had to be aligned manually to 16
+       bytes boundary
```
Geod24: Why ?

source/agora/common/BanManager.d
```diff
@@ -175,16 +175,18 @@ public class BanManager
 
         Params:
             address = the address to increase the fail count for
+            fail_count_inc = the number with which the failed request count
+                should be increased
 
     ***************************************************************************/
 
-    public void onFailedRequest (Address address) @safe nothrow
+    public void onFailedRequest (Address address, uint fail_count_inc = 1) @safe nothrow
```
Geod24: The commit title is: "Added fail_count_inc to BanManager.onFailedReqest"
Typo asides, this doesn't tell a reader more than the diff would.

Based on the diff, here's an (example from a made up scenario) commit message I would recommend writing:
```
BanManager: Add the ability to weight requests differently

Currently, we weight all requests the same.
However, some requests involve more than just one network requests.
An attacker could slow down a node by only failing the last request in such a scenario,
and delay other requests by as much as it's allowed.
To account for this scenario, we need to be able to pass this information to `onFailedRequest`.
```

This tells you what is being done in the title (on a functional level), why it's being done (commit message), and how it's being done (diff).

source/agora/utils/InetUtils.d
```diff
@@ -234,7 +234,7 @@ version (Windows)
         }
         catch(Exception e)
         {
-            return HostPortTup("",0, HostType.Invalid);
+            return HostPortTup("",0, HostType.Invalid, "");
```
Geod24: You could just return `HostPortTup.init`.

dub.json
```diff
@@ -110,19 +111,6 @@
...
-                "source/agora/consensus/Quorum.d",
-                "source/agora/network/Client.d",
-                "source/agora/network/Manager.d",
-                "source/agora/node/*",
-                "source/agora/flash/*",
-                "source/agora/utils/SCPPrettyPrinter.d",
-                "source/scpp/*",
-                "source/scpd/*",
-                "source/agora/consensus/protocol/Nominator.d",
-                "source/agora/consensus/SCPEnvelopeStore.d",
```
Geod24, line 83: Why is this being deleted ?

dub.json
```diff
@@ -69,6 +69,7 @@
     "dflags": [ "-extern-std=c++14", "-preview=in" ],
     "lflags-posix": [ "-lstdc++", "-lsqlite3" ],
     "lflags-windows": [  "sqlite3.lib", "/nodefaultlib:msvcetd.lib" ],
+    "libs": [ "maxminddb" ],
```
Geod24, line 30: Probably should only be in `server` configuration.

source/agora/test/Crawler.d
```diff
@@ -0,0 +1,72 @@
...
+
+    check_for_success().retryFor(10.seconds);
+}
+
+unittest
+{
+    test_crawling(NetworkTopology.MinimallyConnected);
+
+    // Giving enough time for proper shutdown of nodes
+    Thread.sleep(1.seconds);
```
Geod24: Shutdown ?

source/agora/network/CrawlerTypes.d
```diff
@@ -0,0 +1,98 @@
...
+    public string os;
+
+    /// The time the crawling result
+    public TimePoint crawl_time;
+}
+
+/// Crawling results for all the nodes that are directly or indirectly reachable
+/// from this node. Special serialize/fromBinary methods are provided, because
+/// the current binary serializer cannot deal with associative arrays.
+public static struct CrawlResultHolder
```
Geod24: `static` is redundant outside of another aggregate / function.

source/agora/network/CrawlerTypes.d
```diff
@@ -0,0 +1,98 @@
...
+
+import agora.common.Types : Address, TimePoint;
+import agora.serialization.Serializer;
+
+import std.algorithm : each;
+import std.range : iota;
+import std.traits : KeyType, ValueType;
+
+/// Result obtained by crawling a particular node
+public static struct CrawlResult
```
Geod24: Ditto `static`

source/agora/network/CrawlerTypes.d
```diff
@@ -0,0 +1,98 @@
...
+    /// from this node
+    public CrawlResult[Address] crawl_results;
+
+    ///
+    alias crawl_results this;
+
+    ///
+    public void serialize (scope SerializeDg dg) const @safe
+    {
+        serializePart(cast(ulong) crawl_results.length, dg);
```
Geod24: `serializePart!ulong(crawl_results.length, dg);`
Although I don't see why you'd need that, I don't think that'll be exchanged on the network ?

source/agora/network/NodeLocator.d
```diff
@@ -0,0 +1,285 @@
+/*******************************************************************************
+
+    Contains code to locate the geographilcal location of the nodes in the network.
```
Geod24: "geographi[l]cal"

source/agora/network/NodeLocator.d
```diff
@@ -0,0 +1,285 @@
...
+
+        Params:
+            entry_data = data to convert
+
+        Returns:
+            data converted to string
+
+    ***************************************************************************/
+
+    private string getStringFromEntry (const ref MMDB_entry_data_s entry_data) const
```
Geod24: `in`

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
+/*******************************************************************************
+
+    Contains a network crawler implementation, that tries to determine
+    the geographical location, OS for other reachable network nodes.
```
Geod24: Listing should end with "and", e.g. "a, b, and c", so here "geographical location and OS"

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
...
+
+    The implementation uses a never ending BFS search, where certain nodes
+    can temporarily be banned. Nodes can get temporarily banned, if they become
+    unreachable for a certain period of time, or if they recommend too many
+    unreachable nodes.
```
Geod24: I would add a note that this does not affect the normal operation of the network, e.g. it's a separate `BanManager`.

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
...
+    4. Nodes ignore the port of other nodes during crawling, so an attacker
+       can not start up multiple nodes on the same machine on different ports
+       to distort crawling results.
+    5. Nodes translate domain names to IP addresses during crawling, so an attacker
+       can not start up multiple nodes on multiple subdomains like sub1.attacker.com,
+       sub2.attacker.com with the same IP address to distort crawling results.
+
+    The implementation tries to recover from network failures:
+
+    Whenever a particular node experiences a local network faluire, all other nodes
```
Geod24: failure

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
...
+    The implementaion handles some special cases
+
+    1. Multiple crawling fibers will share the same instance of NodeLocator,
+       so the shutdown of the fibers need to be coordinated to make sure during
+       the shutdown no fibers will try to use an already stopped shared NodeLocator.
+    2. The default behavior of TestNetworkManager.getClient is to asserts with 0.
+       whenever a node would like to create a NetworkClient to an already shut down node.
+       As nodes (and crawlers) stop at different times, the above assert could trigger.
+       In order to avoid this a new config parameter called
+       `TestConf.use_non_assert_get_client` was added.
```
Geod24, line 52: Since we're seeing this failure elsewhere, I think we should fix that.

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
...
+            this.discovered_node_list.removeFront();
+
+            NodeInfo node_info;
+            try
+                // Node might be banned because of too many wrong recommendation,
+                // despite the fact the node is constantly available
+                if (this.banman.isBanned(client_address))
+                    throw new Exception("Client banned");
+                else
+                    node_info = network_client_holder.network_client.getNodeInfo();
```
Geod24: A nested function sounds more reasonable than throwing an `Exception` to follow the same code path.

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
...
+
+    ***************************************************************************/
+
+    private CrawlResult populateCrawlResult (string[] extracted_values) @safe nothrow
+    {
+        CrawlResult crawl_result;
+        foreach (i, ref field; crawl_result.tupleof)
+            static if (is(typeof(field) == string))
+                field = extracted_values[i];
+            else if (is(typeof(field) == TimePoint))
```
Geod24: `else static if`

Otherwise it's a runtime `if`. It's one of those D's gotcha.
Also what happens for other fields ? We just ignore it ? I think we should have a `static assert` for those.

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
...
+    public void stop () @safe @nogc nothrow pure
+    {
+        this.is_shutting_down = true;
+    }
+
+    // Stop the Crawler
+    private void do_stop ()
+    {
+        this.node_locator.stop();
+    }
```
Geod24: Only used once, one liner, and private ? I think this should just be inlined ?

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,526 @@
...
+        foreach (address; addresses)
+            if (auto transformed_address = this.transformAddress(address))
+            {
+                auto network_client = this.getNetworkClient(transformed_address);
+                // Getting the client of an already shut down/never started node
+                // returns null in unittests if `use_non_assert_get_client` is
+                // set to `true`.
+                version (unittest)
+                    if (network_client is null)
+                        break;
```
Geod24, line 436: It's one of those special cases I was talking about. We need to solve this differently.

source/agora/test/Base.d
```diff
@@ -1958,6 +1996,27 @@ public struct TestConf
...
+    public Duration crawling_interval = 2.seconds;
+
+    /// true, if NetworkManager.getClient should return null, as opposed to assert,
+    /// in case the node passed to getClient has already been shut down or
+    /// never created
+    public bool use_non_assert_get_client = false;
+
+    /// true, if NetworkManager should automatically initiate network discovery,
+    /// when the node is started up
+    public bool do_network_discovery = true;
```
Geod24, line 1990: I'm worried that so much of the crawler internal ends up being booleans in the global unittest configuration.

Geod24 (changes requested):
Nice work! I left a few comments.

Other questions:
1) Why do we need to commit the maxmindb file ?
2) Where did it come from and what's the license ?
3) I see that clients can choose to enable crawling, but can they choose to disable being crawlable ?
4) Too much of the crawler ends up being configured in `TestConf`, that's a red flag. `do_network_discovery` and `use_non_assert_get_client` are two obvious things to remove.

ferencdg:
> Nice work! I left a few comments.
> 
> Other questions:
> 
> 1. Why do we need to commit the maxmindb file ?
> 2. Where did it come from and what's the license ?
> 3. I see that clients can choose to enable crawling, but can they choose to disable being crawlable ?
> 4. Too much of the crawler ends up being configured in `TestConf`, that's a red flag. `do_network_discovery` and `use_non_assert_get_client` are two obvious things to remove.

1. we don't need to, in fact it seems we would need to obtain a redistribution license even for the free version of geoIP lite. I will remove the database file from the commit. More about the licensing: https://dev.maxmind.com/geoip/geolite2-free-geolocation-data


2. downloaded from https://www.maxmind.com/en/accounts/current/geoip/downloads after registration, license:
"The GeoLite2 end-user license agreement incorporates components of the Creative Commons Attribution-ShareAlike 4.0 International License. The attribution requirement may be met by including the following in all advertising and documentation mentioning features of or use of GeoLite2 data:"
More information about the license: https://dev.maxmind.com/geoip/geolite2-free-geolocation-data

3. I plan to add that functionality along with others, let me do that.

4. What is your suggestion? How should I fix that?

source/mmdb/MaxMindDB.d
ferencdg: ubyte[16] doesn't seem to force the union to be aligned on 16 byte boundary, but the C object file is probably aligned on 16 bytes. Without the alignment above I received garbage back from the C library

dub.json
ferencdg: Because it caused compile errors and when I tried to selectively remove some files from the list (and not all), then a different compile time error message appeared. After trying for 10 minutes I gave it up... All of this becomes very fiddly and brittle just to try to reduce compile time.

source/agora/test/Crawler.d
ferencdg: I have unit tests where a network is started and shut down multiple times. I wanted to leave some time before the shutdown and the network start.

source/agora/network/CrawlerTypes.d
ferencdg: oh, it was inside before, let me remove...

source/agora/network/CrawlerTypes.d
ferencdg: not sure I understand this one... We need to know the size of the map to know how many elements we can read from it during deserialization.

source/agora/test/Base.d
ferencdg: then what is your suggestion?

dub.json
Geod24: It wasn't to remove compile time, but to remove dependency. But I agree it's brittle. There has been quite a few improvements in both DUB and the code base since this was first written, I'll revisit to see if there's an obvious path forward. If not, I'm not too bothered by the diff, but wanted to understand where it came from.

source/agora/common/Set.d
Geod24: My suggestion here is to delete the method, BTW.

Geod24:
> in fact it seems we would need to obtain a redistribution license even for the free version of geoIP lite

That's what I was worried about. Better to have something that download it if missing than to distribute it ourselves.
For the CI, we can download it, and then use `actions/cache`.

> What is your suggestion? How should I fix that?

Need to wrap my head around that, will come back to it soonish.

source/agora/network/CrawlerTypes.d
Geod24: It's about the `cast`. `serializePart(crawl_results.length, dg);` should work just fine. The only reason I can think of for this `cast` is if we send that data on the network (and hence the other side might have `size_t.sizeof == 4`) or if we hash it.
In case we need it, using `serializePart!ulong` is the way to go as it relies on implicit conversion, and thus is more robust to change.

ferencdg: length is size_t which is platform dependent that is why I used the cast. Even though currently binary serialization only used in network test, there is no need not to handle this case in a general method like serialization.

Geod24: Yes, so:
1) The data is sent on the network and using platform-dependent type is problematic (my original comment asked if that was the case);
2) You can just use explicit template instantiation to avoid the `cast` and make the code more robust.

source/agora/utils/InetUtils.d
ferencdg: I might make sense to have an Invalid enum value in case we want to use it in `final switch`, no?

Geod24: It should probably be the default value of the enum, no ?

dub.json
ferencdg: network/Manager.d pulls in NodelLocator which uses mmdb: currently configurations like 'client' or name-registry would not compile if maxminddb is not linked

ferencdg:
1. fixed according to comments except 
https://github.com/bosagora/agora/pull/2097#issuecomment-855529687
for which I am still waiting for comments

2. rebased

3. added new functionality
3.1 node able to exclude himself from crawling
3.2 nodes provide extra information: os, client name, client version, highest block number, is_ipv4

source/agora/network/Manager.d
```diff
@@ -804,9 +804,11 @@ public class NetworkManager
     ***************************************************************************/
 
     protected BanManager getBanManager (in BanManager.Config banman_conf,
-        Clock clock, cstring data_dir)
+        Clock clock, string data_dir)
```
Geod24: Why the change ?

.gitignore
```diff
@@ -45,4 +45,4 @@ docs/
 source/scpp/extra/DVMChecks.cpp
 
 # Agora log directory when Agora is started up with config.example.yml
-log/ 
+log/
```
Geod24: 👀

source/agora/utils/InetUtils.d
```diff
@@ -234,7 +234,7 @@ version (Windows)
         }
         catch(Exception e)
         {
-            return HostPortTup("",0, HostType.Invalid);
+            return HostPortTup.init;
```
Geod24: Need to re-org the enum though

.gitignore
```diff
@@ -46,3 +46,7 @@ source/scpp/extra/DVMChecks.cpp
 
 # Agora log directory when Agora is started up with config.example.yml
 log/
+
+# Infrastructure project containing GeoIp
+infrastructure/
+
```
Geod24: Why is this in `.gitignore` ?

source/agora/api/FullNode.d
```diff
@@ -41,6 +42,21 @@ public struct NodeInfo
...
+    public bool include_in_network_statistics;
+
+    /// Name of the client
+    public string client_name = "Agora";
+
+    /// Version of the client
+    public string client_ver;
+
+    /// Blockchain height known to the client
+    public Height height;
```
Geod24, line 61: Doesn't that just means we have two endpoints to do the same thing ?

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,570 @@
...
+    // Start the Crawler
+    public void start ()
+    {
+        import std.array : array;
+
+        if (!this.config.node.collect_network_statistics)
+            return;
+
+        // at least 1 seed node needs to be provided
+        this.network_manager.seed_addresses.length || assert(0);
```
Geod24, line 390: Why ? If a node start with only the registry, it won't have any

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,570 @@
...
+        return new NodeLocatorGeoIP(this.config.node.ipdb_path);
+    }
+
+    // Start the Crawler
+    public void start ()
+    {
+        import std.array : array;
+
+        if (!this.config.node.collect_network_statistics)
+            return;
```
Geod24: We shouldn't even get here, the check should be in `FullNode`, not here, so we don't needlessly instantiate that object.

source/agora/network/Crawler.d
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    // Initiate stopping the Crawler
+    public void stop () @safe @nogc nothrow pure
+    {
+        this.is_shutting_down = true;
+    }
+
+    // Stop the Crawler
+    pragma(inline, true)
```
Geod24: What I meant by inline is "remove the method and just put the code in the function", not `pragma(inline)`.

Geod24 (changes requested):
You could submit the MMDB commit on its own if you'd like. Also most commit don't come with a justification, their title is just repeating what the diff shows. Reminder that [this is a must read](https://chris.beams.io/posts/git-commit/).

.gitignore
ferencdg: because during development people will need to check out infrastructure(because it contains the mmdb file), and without checking out, 1 test would fail

source/agora/api/FullNode.d
ferencdg: it would force the crawler to do an extra http call for every single nodes

.gitignore
Geod24: That's why I mentioned that this should soft-fail. Otherwise we're locking out contributors which do not have access to the `infrastructure` repo.

source/agora/api/FullNode.d
Geod24: What I mean is that you could merge them together. For example by providing a `bool` parameter asking for full info, and embedding `Nullable!CrawlerInfo` in `NodeInfo`.

.gitignore
ferencdg: we could have also just created a public repo with that one file, so no-one is 'locked-out' and the test can still run

source/agora/network/Crawler.d
ferencdg: I suppose you mean the name-registry not the test registry. 

The crawler relies on the seed address to initiate the crawling. In theory we could have a 

- Validator
- that has no seeds configured
- we also want to use it as a network crawler

but it is kind of a corner case. I can modify the code though.

source/agora/network/Crawler.d
ferencdg: I think it is cleaner to try to push the details down to lower level classes.

I would rather see:
```D
crawler.start()
```
in FullNode.d
instead of
```D
if (!this.config.node.collect_network_statistics)
{
	crawler = new Crawler(....);
	crawler.start()
}
```
and then checking
```D
if (crawler != null)
```
everywhere in the FullNode if we want to use the Crawler

source/agora/network/Crawler.d
ferencdg: I was thinking a bit more about this: I think we should keep the assert here, so our crawler can always recover from network outages. 

If we allow the following node to act as a crawler:

- Validator
- that has no seeds configured
- we also want to use it as a network crawler 

then what happens if the crawler node has a network outage, and then that outage causes every node in its BFS search tree to be (permanently) removed. Normally that search tree would never be empty, because we have seed nodes(although this node didn't define any). 

We shouldn't treat nodes that were returned from the name registry as seed nodes I think because those nodes can come and go. And if we really want to treat name regitsty nodes as seeds, then it should be done in the Network registry.

source/agora/network/Crawler.d
Geod24: > I think it is cleaner to try to push the details down to lower level classes.

As a general principle, I agree. But in this case, the crawler is something that we "fire and forget", as the node does not request it once it's instantiated. The only moment we need to coordinate with the crawler is in `start` and `stop`, and so two `if`s aren't really a concern.

ferencdg: > > I think it is cleaner to try to push the details down to lower level classes.
> 
> As a general principle, I agree. But in this case, the crawler is something that we "fire and forget", as the node does not request it once it's instantiated. The only moment we need to coordinate with the crawler is in `start` and `stop`, and so two `if`s aren't really a concern.

this is in FullNode:

```D
    /// GET: /network_info
    public override CrawlResultHolder getNetworkInfo () @safe
    {
        this.endpoint_request_stats
            .increaseMetricBy!"agora_endpoint_calls_total"(1, "network_info", "http");
        return this.crawler.getNetworkInfo();
    }
```

now we will need to do a check if there is indeed a crawler instance. This check didn't even needed to be done in in crawler.getNetworkInfo before as we just returned an empty CrawlResultHolder.

anyway, I can make the change

source/agora/common/Config.d
```diff
@@ -538,6 +556,11 @@ private NodeConfig parseNodeConfig (Node* node, in CommandLine cmdln)
     Duration relay_tx_interval = opt!(ulong, "node", "relay_tx_interval_secs")(cmdln, node, 15).seconds;
     const relay_tx_min_fee = Amount(opt!(ulong, "node", "relay_tx_min_fee")(cmdln, node, 0));
     Duration relay_tx_cache_exp = opt!(ulong, "node", "relay_tx_cache_exp_secs")(cmdln, node, 1200).seconds;
+    const collect_network_statistics = opt!(bool, "node", "collect_network_statistics")(cmdln, node, false);
+    const include_in_network_statistics = opt!(bool, "node", "include_in_network_statistics")(cmdln, node, true);
+    const num_of_crawlers = opt!(ubyte, "node", "num_of_crawlers")(cmdln, node, 1);
+    Duration crawling_interval = opt!(ulong, "node", "crawling_interval_secs")(cmdln, node, 3).seconds;
+    const ipdb_path = opt!(string, "node", "ipdb_path")(cmdln, node, "infrastructure/bigfiles/ipdb.mmdb");
```
Geod24, line 563: In general, one should use default values on the struct instead of default values here.

tests/unit/NodeLocator.d
```diff
@@ -0,0 +1,49 @@
...
+    assert(node_locator.extractValues(
+               "211.179.51.66",
+               [
+                   "continent->names->en",
+                   "country->names->en",
+                   "city->names->en",
+                   "location->latitude",
+                   "location->longitude",
+                   ]) ==
+           ["Asia", "South Korea", "Seoul (Namdaemunno 5(o)-ga)", "37.5562", "126.975"]);
```
Geod24, line 46: Isn't that test a little brittle, as MMDB updates are likely to break it ?

tests/unit/NodeLocator.d
```diff
@@ -0,0 +1,49 @@
...
+
+private int main ()
+{
+    auto path = environment.get("AGORA_GEOIP_PATH");
+    if (!path.length)
+    {
+        stderr.writeln("Skipping ", __FILE_FULL_PATH__, " test because it requires GeoIP databases");
+        stderr.writeln("Provide the GeoIP database path via the `AGORA_GEOIP_PATH` environment variable to run it");
+        return 0;
+    }
```
Geod24, line 33: This is what I had in mind for soft-failing test

doc/config.example.yaml
```diff
@@ -58,6 +58,20 @@ node:
...
+  # The number of network crawlers that will be instantiated to collect
+  # statistics about other nodes in the network
+  num_of_crawlers : 3
+
+  # The number of seconds one crawler should wait after successfully contacted a node
+  crawling_interval_secs : 2
+
+  # The path to a file containing IP address -> geographical location mapping
+  ipdb_path : "data/ipdb.mmdb"
+
```
Geod24, line 74: This could be simplified by making it its own section, e.g. like `admin`.

source/agora/network/NodeLocator.d
```diff
@@ -0,0 +1,268 @@
...
+        if (MMDB_SUCCESS != mmdb_error)
+        {
+            log.warn("Got an error from libmaxminddb: {}", MMDB_strerror(mmdb_error));
+            return MissingValue.repeat(paths.length).array();
+        }
+
+        // Extract the data
+        foreach (ind, path; paths)
+        {
+            auto path_splitted = path.split(PathSeparator).map!(path_part => path_part.toStringz()).array();
```
Geod24, line 167: IIUC the MMDB API requires an array of `char*` so why do we have custom path logic using `->` ?

source/agora/network/NodeLocator.d
```diff
@@ -0,0 +1,268 @@
...
+
+        Params:
+            entry_data = data to convert
+
+        Returns:
+            data converted to string
+
+    ***************************************************************************/
+
+    private string getStringFromEntry (in MMDB_entry_data_s entry_data) const
```
Geod24, line 190: This doesn't use anything from the context and could be `static` or a free function right ?

source/agora/network/NodeLocator.d
```diff
@@ -0,0 +1,268 @@
...
+        extractedValues.length = paths.length;
+
+        foreach (ind, path; paths)
+            if (auto it = path in path_map)
+                extractedValues[ind] = *it;
+            else
+                extractedValues[ind] = MissingValue;
+
+        return extractedValues;
+    }
```
Geod24, line 267: This looks like the wrong place to do dependency injection. The injection likely belongs to the place doing the data acquisition, not the place that transforms it.

Geod24 (changes requested):
A few things. I'll have a deeper look in the code soon.

doc/config.example.yaml
```diff
@@ -58,6 +58,20 @@ node:
...
+  # The number of network crawlers that will be instantiated to collect
+  # statistics about other nodes in the network
+  num_of_crawlers : 3
+
+  # The number of seconds one crawler should wait after successfully contacted a node
+  crawling_interval_secs : 2
+
+  # The path to a file containing IP address -> geographical location mapping
+  ipdb_path : "data/ipdb.mmdb"
```
Geod24, line 73: I took a look at it locally, and I think `collect_network_statistics` and `ipdb_path` should be merged together.
As we do not want to give the user the chance to write a configuration that is self-inconsistent, we can base the fact that the user wants to use the frontend off the fact they provided an `ipdb_path` in the config file.
Also, `ipdb_path` => `mmdb_path`, as the later is more commonly used, and specific (`ipdb` would make sense if there was a standard for IP databases and there could be multiple providers).

doc/config.example.yaml
```diff
@@ -58,6 +58,20 @@ node:
   # increased in order to be added to the transaction pool
   double_spent_threshold_pct: 20
 
+  # true, if this node should collect statistics about other
+  # nodes in the network, including their geographical location and OS
+  collect_network_statistics : true
+
+  # The number of network crawlers that will be instantiated to collect
+  # statistics about other nodes in the network
+  num_of_crawlers : 3
```
Geod24, line 67: `crawlers_count`, keeping with the practice of using the unit as a suffix

source/agora/api/FullNode.d
Geod24: This still needs to be addressed

PR #2098, opened 2021-05-17
Restart test: Simplify by storing the stateDB

Geod24:
Instead of individually storing the object, just save the stateDB,
which is intended to be the persistent layer.


hewison-chris (approved):
LGTM

kchulj (approved):

PR #2100, opened 2021-05-20
Fix TransactionReplacement test as CoinBase is not always at index 0

Geod24:
My last minutes changes in PR1881 led to some spurious failures,
as Coinbase seems to not be consistently the first entry in the block.


hewison-chris (approved):
LGTM

kchulj (approved):

linked0 (approved):

PR #2103, opened 2021-05-20
Simplify SCP store by re-using the cache DB

Geod24:
We recently moved dependency injection of the level of the database,
which considerably simplifies the amount of code duplication and
work needed to do testing. The SCP cache was originally forgotten in that process,
this commit rectifies this.


hewison-chris (approved):
LGTM

linked0 (approved):
LGTM

PR #2105, opened 2021-05-21
Revamp Docker image build and push to support releases

Geod24:
With this change, the following will happen:
- The docker 'latest' tag, which is the default, will no longer be pushed to;
- When a PR is merged / something is pushed to a branch, 'devel' will be used instead;
- When a tag is pushed, the matching tag will be pushed to docker hub;
- When pushing a branch or a tag to one's fork, errors will no longer be generated;

The only downside here is to tie up the registry and agora build together,
so that if one fails, both have to be repeated. In practice, this almost never happened,
so it shouldn't be a great concern.

The goal is to latest trigger a push to 'latest' when a release is made.


.github/workflows/devel-docker.yml
```diff
@@ -0,0 +1,54 @@
...
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+
+        # This step will checkout the source with all the submodules
+      - name: '[Agora] Build and push'
+        id: agore_build
+        uses: docker/build-push-action@v2
+        with:
+          push: ${{ steps.login.outcome == 'success' }}
+          tags: bpfk/agora:devel
```
hewison-chris, line 39: Just an idea: Couldn't we add a second tag which includes the github.sha so we can pull for testing.

Geod24: I think that would bloat the repository quite a bit. Plus, it's fairly easy to build locally and push, no ?

hewison-chris: Yes I suppose if you are testing your own build. If you want to share a pushed devel tag then I guess you can still add another tag manually.

hewison-chris (approved):
LGTM

PR #2113, opened 2021-05-25
Fix to use the validator's pre-image when generating the encryption key.

### LINKED ISSUE #2074: When the encrypted key is provided, it should be changed to use the pre-image of the actual validator.
The encrypted key is used to vote in Votera. 
The pre-image of the validator is not included in the process of creating the encrypted key.
We need to modify to generate an encrypted key using the pre-image associated with that node.

TrustHenry:
The encrypted key is used to vote in Votera.
To prevent this pre-image from being pre-released, add the PreimageRevealPeriod + 1 height.

Related to #2074

### DIFF
--- a/source/agora/node/admin/AdminInterface.d
+++ b/source/agora/node/admin/AdminInterface.d
@@ -20,6 +20,7 @@ module agora.node.admin.AdminInterface;
 import agora.api.Admin;
 import agora.common.Types;
 import agora.consensus.data.PreImageInfo;
+import agora.consensus.EnrollmentManager;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.crypto.Schnorr;
@@ -72,6 +73,9 @@ public class AdminInterface : NodeControlAPI
     /// Clock instance
     private Clock clock;
 
+    /// Enrollment manager
+    private EnrollmentManager enroll_man;
+
     /***************************************************************************
 
         Constructor
@@ -79,14 +83,16 @@ public class AdminInterface : NodeControlAPI
         Params:
             key_pair = the keypair of this node
             clock = clock instance
+            enroll_man = the enrollmentManager
 
     ***************************************************************************/
 
-    public this (const KeyPair key_pair, Clock clock)
+    public this (const KeyPair key_pair, Clock clock, EnrollmentManager enroll_man)
     {
         this.log = Logger(__MODULE__);
         this.key_pair = key_pair;
         this.clock = clock;
+        this.enroll_man = enroll_man;
     }
 
     ///
@@ -183,9 +189,8 @@ public class AdminInterface : NodeControlAPI
     public override string encryptionKeyQR (
         string app, ulong height, out string contentType, out string vary)
     {
-        // TODO: PreImage to be released at that height
-        // Or it can be pre-released, so it can be `height + PreimageRevealPeriod`
-        PreImageInfo pre_image;
+        // Gets the pre-image of the requested height.
+        Hash pre_image = this.enroll_man.getOurPreimage(Height(height));
 
         // blake2b hashing
         const Hash key = hashMulti(pre_image, app);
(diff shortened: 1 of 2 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -110,7 +110,7 @@ public class EnrollmentManager
     public EnrollmentPool enroll_pool;
 
     /// Ditto
-    private PreImageCycle cycle;
+    public PreImageCycle cycle;
```
linked0: The data should be encapsulated. I mean it should be `private`.

source/agora/node/admin/AdminInterface.d
```diff
@@ -183,10 +189,9 @@ public class AdminInterface : NodeControlAPI
     public override string encryptionKeyQR (
         string app, ulong height, out string contentType, out string vary)
     {
-        // TODO: PreImage to be released at that height
-        // Or it can be pre-released, so it can be `height + PreimageRevealPeriod`
-        PreImageInfo pre_image;
-
+        // Gets the pre-image of the requested height.
+        // It can be pre-released, so it add to the `Reveal period + 1`
+        Hash pre_image = this.enroll_man.cycle[Height(height + this.enroll_man.PreimageRevealPeriod + 1)];
```
linked0:
```suggestion
        PreImageInfo preimage;
        this.enroll_man.getNextPreimage(preimage, [Height(height - 1)]);
```
and how about using `preimage.hash` afterward.
If it does not suit for your purpose, I would recommend to add a new function in `EM` for that.

linked0 (commented):
Good start.

source/agora/node/admin/AdminInterface.d
TrustHenry: The getNextPreimage function can only import pre-images from an enrolled enrollment cycle range.
So I created a function to get a pre-image of the block height.

TrustHenry: @Geod24 ping

source/agora/node/admin/AdminInterface.d
```diff
@@ -319,13 +326,42 @@ public struct EncryptionKey
...
+    auto man = new EnrollmentManager(stateDB, cacheDB, key_pair, new immutable(ConsensusParams)(20));
+    auto adminInterface = new AdminInterface(key_pair, clock, man);
+    string contentType = "image/svg+xml";
+    string vary = "Accept-Encoding";
+    string enc_qr_svg = adminInterface.encryptionKeyQR("Votera", 100, contentType, vary);
+
+    string sample_enc_qr_svg =
+`<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd">
+<svg xmlns="http://www.w3.org/2000/svg" version="1.1" viewBox="0 0 500 500">
```
Geod24: This is not really testing anything. I'd rather we remove this and have the test on DevOps side.

TrustHenry: Okay, I removed this test code.

source/agora/node/admin/AdminInterface.d
```diff
@@ -183,9 +189,10 @@ public class AdminInterface : NodeControlAPI
     public override string encryptionKeyQR (
         string app, ulong height, out string contentType, out string vary)
     {
-        // TODO: PreImage to be released at that height
-        // Or it can be pre-released, so it can be `height + PreimageRevealPeriod`
-        PreImageInfo pre_image;
+        // Gets the pre-image of the requested height.
+        // It can be pre-released, so it add to the `Reveal period + 1`
+        Hash pre_image = this.enroll_man.getOurPreimage(
+            Height(height + this.enroll_man.PreimageRevealPeriod + 1));
```
Geod24: No it should just be `height`, the app requesting it should compensate for `PreImageRevealPeriod`

TrustHenry: Okay, I updated.

Geod24 (approved):

PR #2119, opened 2021-05-26
Init Tracy by default

omerfirmak:
This is needed to start the automatic stack sampling.

Stack sampling requires running agora as root.

### DIFF
--- a/scripts/build_tracy.d
+++ b/scripts/build_tracy.d
@@ -36,7 +36,7 @@ version (Windows)
     immutable ObjExt = `.obj`;
     immutable CppCmd = [
         `cl`, `/std:c++14`, `/DEBUG`, `/c`, `/D "TRACY_ENABLE"`,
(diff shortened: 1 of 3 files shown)


Geod24:
The first commit shouldn't make a difference. All module ctors are run, if they are compiled it. That's why for example we put `version (unittest):` above everything but the module declaration in `agora/test/`. But having it in `main` makes sense, so just change the commit message.

Geod24:
> Stack sampling requires running agora as root.

Why ?

omerfirmak:
> > Stack sampling requires running agora as root.
> 
> Why ?

Tracy uses linux's tracing facilities, which requires root access.

omerfirmak:
> The first commit shouldn't make a difference. All module ctors are run, if they are compiled it. That's why for example we put `version (unittest):` above everything but the module declaration in `agora/test/`. But having it in `main` makes sense, so just change the commit message.

Done

Geod24 (approved):

Issue #2118, opened 2021-05-26
Get Tracy call stack sampling to work with agora


omerfirmak:

Geod24:
This works now, doesn't it @omerfirmak ?

omerfirmak:
Yeah, this works for the most part. It can be closed.

PR #2120, opened 2021-05-26
Mount debugfs in docker image

omerfirmak:
To get #2118, we need to both run agora as root and mount debugfs inside the container


omerfirmak:
This is not how it is supposed to be done? @Geod24

Geod24:
No, it needs to be done via runtime arg. See the - v (volume) argument in the unit file.

PR #2125, opened 2021-05-31
Fixed config.example.yml

ferencdg:
Previous version was out of date, and no longer worked


doc/config.example.yaml
```diff
@@ -195,31 +198,7 @@ dns:
 ################################################################################
 logging:
   root:
```
ferencdg: even after changing root to 
` - name: root`
agora still returned parsing errors, so removed the whole block

doc/config.example.yaml
```diff
@@ -85,7 +85,7 @@ validator:
   # DO NOT USE THOSE VALUES ANYWHERE
   # Private seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
   # Public address:  GCOQEOHAUFYUAC6G22FJ3GZRNLGVCCLESEJ2AXBIJ5BJNUVTAERPLRIJ
-  seed:    SCT4KKJNYLTQO4TVDPVJQZEONTVVW66YLRWAINWI3FZDY7U4JS4JJEI4
+  seed:    SAUHVPR7O7F2QGLDVXG3DQTVHXESE3ZAWHIIGKT35LCHIPLZBZTAFXJA
```
Geod24: Hum the seed doesn't really matter here, but the `Public address:` comment should be in Bech32.

doc/config.example.yaml
```diff
@@ -135,7 +135,10 @@ flash:
   # Maximum number of seconds to re-try a failed payment / update request
   # before the request is considered failed and is reported to the listener
   # (wallet / etc)
-  max_retry_time: 60
+  max_retry_time: 60000
```
Geod24, line 136: Why ?

doc/config.example.yaml
```diff
@@ -135,7 +135,10 @@ flash:
...
+
+  # Address of the name registry
+  registry_address: http://127.0.0.1:3003
```
Geod24: ~Probably should make this mandatory and remove the network part.~

Or maybe not... But the comment should explain what this is, and its relationship with DNS / network.

doc/config.example.yaml
ferencdg: Agora crashes/asserts without that change
```
core.exception.AssertError@source/agora/crypto/Key.d(69): SecretKey should always be valid Scalar!
----------------
??:? [0x55a794f0dc85]
??:? [0x55a794f35866]
??:? [0x55a794f18fad]
??:? [0x55a794f0bd69]
/home/dan/dev/bosagora/agora/source/agora/crypto/Key.d:69 [0x55a79488adb7]
/home/dan/dev/bosagora/agora/source/agora/common/Config.d:679 [0x55a7948d5e9e]
/home/dan/dev/bosagora/agora/source/agora/common/Config.d:469 [0x55a7948d4a37]
/home/dan/dev/bosagora/agora/source/agora/common/Config.d:353 [0x55a7948360f5]
/home/dan/dev/bosagora/agora/source/agora/node/main.d:134 [0x55a794835a60]
/home/dan/dev/bosagora/agora/source/agora/node/main.d:131 [0x55a794834760]
??:? [0x55a794f18c7b]
??:? [0x55a794f18b77]
??:? [0x55a794f189cd]
/home/dan/dev/thirdparty/ldc2-1.25.1/bin/../import/core/internal/entrypoint.d:42 [0x55a7948376b4]
??:? __libc_start_main [0x7f65de0ff0b2]
??:? [0x55a79479b59d]


```

doc/config.example.yaml
Geod24: > agora still returned parsing errors, so removed the whole block

Well we really can't merge that. The syntax can be seen in the infrastructure repo, or just by reading the code.
The change is simple: Instead of using:
```yaml
- name: agora.foo
  attribute: value
```

We should use:
```yaml
agora.foo:
  attribute: value
```

doc/config.example.yaml
ferencdg: because it is in milliseconds, and agora would crash without that change
```D
enforce(max_retry_time > max_retry_delay, "`max_retry_time` must be greater than `max_retry_delay`");
```

Geod24: It should be in seconds, as the comment says.

doc/config.example.yaml
Geod24: Please update the comment above. And we probably need to check for valid values on parsing, the assert shouldn't be triggered.

doc/config.example.yaml
ferencdg:
```D
    const max_retry_time = get!(Duration, "flash", "max_retry_time", str => str.to!ulong.msecs)
```

ferencdg:
fixed

Geod24 (approved):

kchulj (approved):

PR #2126, opened 2021-05-31
Add UTXO to `Validator.getPublicKey`

Geod24:
In support of #2080

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -273,7 +273,8 @@ public class Validator : FullNode, API
         endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(
             1, "public_key", "http");
 
-        Identity id = Identity(this.config.validator.key_pair.address);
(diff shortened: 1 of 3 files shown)


source/agora/node/Validator.d
```diff
@@ -273,7 +273,8 @@ public class Validator : FullNode, API
         endpoint_request_stats.increaseMetricBy!"agora_endpoint_calls_total"(
             1, "public_key", "http");
 
-        Identity id = Identity(this.config.validator.key_pair.address);
+        Identity id = Identity(this.config.validator.key_pair.address,
+                               this.enroll_man.getEnrollmentKey());
```
linked0, line 277: As discussed on the whiteboard, there should be a workaround for the case where a node does not have its own enrollment key because it's rebuilding the related information from scratch.

linked0 (approved):
Nice, but there should be another PR for the case I commented on. The PR will be my task.

PR #2130, opened 2021-05-31
Talos: Restore ability to navigate between menus

Geod24:
Talos was broken since the dependency update, probably due to an accept-invalid bug being fixed.
The second-to-last commit fixes the bug, the other ones are little fixes found along the way.


kchulj:
There is duplicate content in 1st and 2nd commit.

kchulj (approved):
LGTM

PR #2131, opened 2021-05-31
Flash: Streamline Point/PublicKey usage & co

omerfirmak:
`FlashAPI` already uses `PublicKey`, prefer `PublicKey` over `Point`

More will probably follow.

### DIFF
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -110,7 +110,7 @@ public class Channel
 
     /// @WORKAROUND@: LocalRest issue with timings and sleep not working until
     /// node ctor is done (see Flash restart tests)
-    private FlashAPI delegate (in Point peer_pk, Duration timeout) getFlashClient;
+    private FlashAPI delegate (in PublicKey peer_pk, Duration timeout) getFlashClient;
 
     /// Retry delay algorithm
     private Backoff backoff;
@@ -203,7 +203,7 @@ public class Channel
         OnChannelNotify onChannelNotify,
         OnPaymentComplete onPaymentComplete,
         OnUpdateComplete onUpdateComplete,
-        FlashAPI delegate (in Point peer_pk, Duration timeout) getFlashClient,
+        FlashAPI delegate (in PublicKey peer_pk, Duration timeout) getFlashClient,
         ManagedDatabase db)
     {
         this.db = db;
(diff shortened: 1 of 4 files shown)


source/agora/flash/Channel.d
```diff
@@ -110,7 +110,7 @@ public class Channel
 
     /// @WORKAROUND@: LocalRest issue with timings and sleep not working until
     /// node ctor is done (see Flash restart tests)
```
Geod24: This has been fixed right ?

source/agora/test/Flash.d
```diff
@@ -436,7 +436,7 @@ public class FlashNodeFactory (FlashListenerType = FlashListener)
     public void shutdown ()
     {
         foreach (address; this.addresses)
-            enforce(this.flash_registry.unregister(address.to!string));
+            enforce(this.flash_registry.unregister(PublicKey(address).to!string));
```
Geod24: Isn't it already a `PublicKey` now ?

source/agora/test/Flash.d
```diff
@@ -374,9 +374,9 @@ public class FlashNodeFactory (FlashListenerType = FlashListener)
             &this.flash_registry, &this.listener_registry,
             10.seconds);  // timeout from main thread
 
-        this.addresses ~= pair.V;
+        this.addresses ~= PublicKey(pair.V);
         this.nodes ~= api;
-        this.flash_registry.register(pair.V.to!string, api.listener());
+        this.flash_registry.register(PublicKey(pair.V).to!string, api.listener());
```
Geod24: `this.addresses[$-1].to!string`?

source/agora/flash/Node.d
```diff
@@ -241,10 +241,8 @@ public abstract class FlashNode : FlashControlAPI
         auto secret = SecretKey(scalar);
         auto address = PublicKey(scalar.toPoint()[]);
         this.managed_keys[address] = secret;
-        Channel[Hash] def;
-        this.channels[address] = def;
-        DList!ChannelConfig def2;
-        this.pending_channels[address] = def2;
+        this.channels[address] = typeof(this.channels[address]).init;
+        this.pending_channels[address] = typeof(this.pending_channels[address]).init;
```
Geod24, line 245: Why is it here in the first place ? Does it override something ? Everything is default-initialized in D.

Geod24 (changes requested):

source/agora/test/Flash.d
omerfirmak: Oops, yes

source/agora/flash/Node.d
omerfirmak: Those are complex types, like AA and DList.
These are there so that when you do `this.channels[address].foo()`, you wont throw.

source/agora/flash/Channel.d
omerfirmak: Yep, we fixed that.
But while this works around that, I think it also fixes the race between channel ctor and peer registering itself to the registry. So we can just remove the workaround comments.

Geod24 (approved):
Me likey

PR #2138, opened 2021-06-01
Remove agora.consensus.data.DataPayload

Geod24:
It was essentially a glorified ubyte[] array and was getting in the way.
As shown from the diff, removing it doesn't have any negative impact.
If we choose to add a specialized type for it later, we can always do it,
and it probably should live in the Transaction module like Input and Output.


omerfirmak (approved):

hewison-chris (approved):
LGTM

kchulj (approved):
LGTM.
Faucet needs another update.

Geod24:
Let's give it another try

PR #2142, opened 2021-06-02
Enable TLS if required files are available

omerfirmak:


source/agora/node/Runner.d
```diff
@@ -144,6 +144,15 @@ public Listeners runNode (Config config)
...
         settings.rejectConnectionPredicate = isBannedDg;
+        // Enable TLS
+        if (exists("agora-cert.pem") && exists("agora-key.pem"))
+        {
+            import vibe.stream.tls;
+            settings.tlsContext = createTLSContext(TLSContextKind.server);
+            settings.tlsContext.useCertificateChainFile("agora-cert.pem");
+            settings.tlsContext.usePrivateKeyFile("agora-key.pem");
+            log.info("TLS Enabled..");
+        }
```
Geod24: Should apply to admin interface as well.

Geod24: Also, it should probably just look up a few places for `cert` and `key`

omerfirmak: Updated

source/agora/node/Runner.d
```diff
@@ -201,3 +205,47 @@ private void setVibeLogLevel (ILogger.Level level) @safe
...
+
+        // Enable TLS
+        if (exists(cert_path) && exists(key_path))
+        {
+            ctx = createTLSContext(TLSContextKind.server);
+            ctx.useCertificateChainFile(cert_path);
+            ctx.usePrivateKeyFile(key_path);
+            break;
+        }
+    }
```
Geod24, line 248: This will only work for both certificate and key file at the same level. Is it intended ?

source/agora/node/Runner.d
```diff
@@ -201,3 +205,47 @@ private void setVibeLogLevel (ILogger.Level level) @safe
...
+
+*******************************************************************************/
+
+private TLSContext getTLSContext ()
+{
+    const cert_file = "agora-cert.pem";
+    const cert_search_paths = [
+        "./",
+        "/etc/ssl/certs/",
+        "/etc/pki/tls/certs/"
```
Geod24: Good dlang habit: trailing comas.

source/agora/node/Runner.d
omerfirmak: Yep, I don't think we need to search recursively.

Geod24 (approved):
LGTM
Regarding the `openssl` thing, we don't need to add `openssl` to the container because we won't generate the certificate while *building* Agora, only when testing, in which case we are going to mount a custom config anyway. Does that make sense ?

PR #2148, opened 2021-06-04
Fix #2077: Refactor catch / discovery task to be DRYer

### LINKED ISSUE #2077: FullNode.startPeriodicDiscovery and Validator.startPeriodicDiscovery are almost identical
Validator.startPeriodicDiscovery should be deleted

Geod24:
Those two functions could be refactored to have the timer start in the start method of the node.
~We loose the immediate firing which was explicitly done (see 'avoid delay'), but things will still work.~
EDIT: No, they didn't, and I restored it, and it's much better.
Additionally this fixes a bug where Validators did not respect the node.block_catchup_interval config field and always used 5 seconds.
Since we currently set this field to 2 seconds in the unittest, this should significantly speed up tests that rely on catch-up.

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -375,8 +375,6 @@ public class FullNode : API
 
     public void start ()
     {
-        this.startPeriodicDiscovery();
-        this.startPeriodicCatchup();
         if (config.node.stats_listening_port != 0)
             this.stats_server = this.makeStatsServer();
         this.transaction_relayer.start();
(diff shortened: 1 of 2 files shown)


Geod24:
I think the bug mentioned (catchup config not respected) is why we saw `LocalTransaction` fails so often.
Since `LocalTransaction` relies on catchup, and the timeout it uses is equal to the catchup frequency (5 seconds), any load / scheduling change will affect which part, the check or the catchup, runs first.

hewison-chris (approved):
LGTM, great to get rid of another race condition.

PR #2149, opened 2021-06-04
Block validation: Ensure monotonicity in `missing_validators`, document, refactor

Geod24:
See each commit for the rationale.

### DIFF
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -125,6 +125,9 @@ public string isInvalidReason (in Block block, Engine engine, Height prev_height
         Enrollment.MinValidatorCount)
         return "Block: Insufficient number of active validators";
 
+    if (!block.header.missing_validators.isStrictlyMonotonic())
+        return "Block: Header's missing_validator field is not sorted";
+
     if (!block.txs.isSorted())
         return "Block: Transactions are not sorted";
 
--- a/source/agora/consensus/validation/Block.d
+++ b/source/agora/consensus/validation/Block.d
@@ -85,8 +85,6 @@ version (unittest)
         curr_time_offset = the current time offset
         block_time_tolerance = the proposed block time offset should be less
                 than curr_time_offset + block_time_tolerance
-        getCoinbaseTX = delegate to get the expected Coinbase TX for a given TX
-            set
 
     Returns:
         `null` if the block is valid, a string explaining the reason it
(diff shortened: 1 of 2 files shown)


source/agora/node/Ledger.d
```diff
@@ -765,6 +762,56 @@ public class Ledger
...
+
+        Validate the signature of a block
+
+        This validate that the signature in a block header is consistent with
+        the enrolled validators, and cryptographically correct.
+        Note that since this requires to know which nodes are validators,
+        this method is contextful and can only guarantee the signature
+        of the next block, as the validator set might change after that.
+
+        Implementation_details:
```
linked0, line 783: The underscore is not necessary.

source/agora/node/Ledger.d
```diff
@@ -765,6 +762,56 @@ public class Ledger
...
+
+        Params:
+            block = the block to verify the signature of
+
+        Returns:
+            the error message if block validation failed, otherwise null
+
+    ***************************************************************************/
+
+    protected string validateBlockSignature (in Block block) @safe nothrow
```
linked0: Does it need to be `protected`? It's also good to be `private`, I think.

linked0 (approved):
LGTM with some nits.

source/agora/consensus/validation/Block.d
```diff
@@ -616,11 +610,7 @@ version (unittest)
         return isInvalidReason(block, engine, prev_height, prev_hash, findUTXO,
             checkFee, findEnrollment, enrolled_validators,
             prev_time_offset, (curr_time_offset == ulong.max) ? block.header.time_offset : curr_time_offset,
-            block_time_tolerance,
-            (in Transaction[] tx_set, in uint[] missing_validators)
-            {
-                return (Transaction[]).init;
-            });
```
kchulj, line 623: Why did this check get deleted?

source/agora/node/Ledger.d
```diff
@@ -764,7 +761,62 @@ public class Ledger
...
+
+          The set of signers is defined as all the validators having revealed
+          a pre-image. For this reason, pre-images are allowed and encouraged
+          to be revealed earlier than they are needed (although not too early).
+
+          With the set of signers known, we derive the block-specific `R`
+          by adding the `R` used in the enrollment to `p * B`, where `p` is
+          the pre-image reduced to a scalar and `B` is Curve25519 base point.
+
+          Hence, the signature present in the block is actuall just the
```
kchulj:
```suggestion
          Hence, the signature present in the block is actually just the
```

kchulj (approved):
LGTM + nits

source/agora/node/Ledger.d
Geod24: No it's not, it's how DDOC recognize a section name. When generating the documentation, a second name starts with a single "word". Any underscores are stripped.

source/agora/node/Ledger.d
Geod24: It does not need to, I thought it would be better to let it `protected`. But YAGNI, so I'll change to `private`.

source/agora/consensus/validation/Block.d
Geod24: Because we don't pass the delegate to that function anymore, so we don't need this plumbing.

Geod24:
Fixed

Issue #2134, opened 2021-06-01
Update SCP to the latest release and remove SecretKey


Geod24:
Currently our SCP version matches v16.0.0. We should update it to v17.2.0 (the latest release as of now) and clean up any unneeded / unwanted dependency.
Currently, the only known blocker is `SecretKey` (and its formatting side) which we don't need, but we do need `PublicKey`.
However, Stellar has an abstraction for this (`NodeID`), but it is not used consistently. See for example https://github.com/stellar/stellar-core/pull/3065 for the direction we need to go in. This is especially important in light of #1309.

Geod24:
Done in #2225

PR #2150, opened 2021-06-04
Stellar: Upgrade vendored libs and utils/numeric to HEAD

Geod24:
This is a low effort / high LoC benefit. I just used the update script and verified the output.
I also noted that the numeric update was needed since they removed support for signed types.
I have more incremental updates coming up, including some PRs to stellar-core to reduce our diff.


Geod24:
Part of #2134

PR #2151, opened 2021-06-04
Remove dependency to 'ocean.core.Test'

Geod24:
```
Now that '-checkaction=context' is usable (and the default),
we do not need to use this function anymore and can rely on regular assert.
```

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -63,7 +63,6 @@ import agora.crypto.Key;
 import agora.crypto.Schnorr;
 import agora.utils.Log;
 version (unittest) import agora.utils.Test;
-version (unittest) import ocean.core.Test;
 
 import d2sqlite3.library;
 import d2sqlite3.results;
@@ -1055,7 +1054,7 @@ unittest
     }
 
     Enrollment[] enrolls = man.getEnrollments(EnrollAt1, &utxo_set.peekUTXO);
-    test!"=="(enrolls.length, 3);
+    assert(enrolls.length == 3);
     assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
 
     // get a stored Enrollment object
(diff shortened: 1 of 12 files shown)


Geod24:
Rebased

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1073,26 +1072,27 @@ unittest
...
-    test!"=="(man.validator_set.countActive(EnrollAt9 + 1), 0);
-    test!"=="(man.validator_set.getEnrolledHeight(EnrollAt9, utxo_hash), ulong.max);
+    assert(man.validator_set.countActive(EnrollAt9 + 1) == 0);
+    assert(man.validator_set.getEnrolledHeight(EnrollAt9, utxo_hash) == ulong.max);
     // Add removed enrollment to the pool with the new height
     assert(man.addEnrollment(
         EnrollmentManager.makeEnrollment(utxo_hashes[1], pairs[1], EnrollAt9, params.ValidatorCycle),
             pairs[1].address, EnrollAt9, &utxo_set.peekUTXO));
     // The expired enrollments in the pool from height 1 are cleared on this next call
     // to get enrollments at a highter height. We do have an enrollment at height 9
```
kchulj, line 1082:
```suggestion
    // to get enrollments at a higher height. We do have an enrollment at height 9
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -1073,26 +1072,27 @@ unittest
...
     // One Enrollment was moved to validator set
-    test!"=="(man.validator_set.countActive(EnrollAt9 + 1), 1);
+    assert(man.validator_set.countActive(EnrollAt9 + 1) == 1);
     // Check last block of cycle is still active
-    test!"=="(man.validator_set.countActive(EnrollAt9 + params.ValidatorCycle), 1);
+    assert(man.validator_set.countActive(EnrollAt9 + params.ValidatorCycle) == 1);
     // Check block in next cycle is no longer active
-    test!"=="(man.validator_set.countActive(EnrollAt9 + params.ValidatorCycle + 1), 0);
+    assert(man.validator_set.countActive(EnrollAt9 + params.ValidatorCycle + 1) == 0);
     // Pool shoud now be empty
```
kchulj, line 1094:
```suggestion
    // Pool should now be empty
```

kchulj (approved):
Took a while to review..
LGTM + nits

Geod24:
@kchulj : As our guideline is to have atomic commits, and those lines are unrelated, I'm not going to change them in this commit. Sometimes it's okay to change not-directly-related code in the same commit, for example if you are going to change the line anyway: e.g. if while changing `test!"..."` => `assert` we spotted a typo in a message, it would be okay to have the following diff:
```diff
- test!"=="(a, b, "Expessions are not equal");
+ assert(a == b", Expressions are not equal");
```

Because, you know, we're touching the line anyway.

However, that doesn't mean we shouldn't fix those: https://github.com/bosagora/agora/pull/2159

PR #2157, opened 2021-06-07
Remove two unused `TestConf` fields

Geod24:
While looking at another PR, I remembered that `TestConf` had some fields which are now redundant.
The intent of `TestConf` is to configure the network, that is, parameters which are common to all nodes, and that are relevant to many / most tests. Parameters that would only be used in a single tests do not belong to `TestConf`, unless they are generic enough to be there.


hewison-chris (approved):
LGTM

linked0 (approved):
LGTM

kchulj (approved):

PR #2158, opened 2021-06-07
SCP: Remove ByteSliceHasher, use ShortHash instead

Geod24:
ByteSliceHasher was renamed in upstream commit
9c1c9f01446fdac9f8f17789a51bb925479746f0 (2019-09-22, released as part of v12.1.0),
to ShortHash, as it allowed to hash without serialization,
but wasn't removed as part of the recent SCP upgrade.
This also uncovered that SipHash was not linked in,
as ShortHash is the only file using it, and that the COPYING mention was missing.


Geod24:
This is pretty cut-and-dry so I'm going to go ahead with this (it's also blocking the real enhancement I was working on).

PR #2162, opened 2021-06-07
Default to 20 for ValidatorCycle in unittest

Geod24:
This is the conclusion of a long serie of change by @hewison-chris and myself.


hewison-chris (approved):
LGTM

PR #2165, opened 2021-06-07
Merge EM.getRandomSeed into Ledger.getRandomSeed

Geod24:
This should simplify the random seed code, and make sure we always acquire it in the same way, without silently ignoring some validators.
Marking as a draft for now as I'd like to test it some more and it will trigger other bugs (namely, `receiveEnvelope` might throw, and since it blocks...).

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -551,34 +551,6 @@ public Block makeNewBlock (Transactions)(const ref Block prev_block,
     return block;
 }
 
-/***************************************************************************
-
-    Create the block random seed from the provided pre-images
-
-    It might be that not all validators provide their `pre-image` and in
-    that case the length of the preimages can be less than the number of
-    active validators.
-
-    Params:
-        preimages = preimages sorted by the utxo keys of the validators
-
-    Returns:
-        Hash of all provided pre-images or Hash.init if an exception is thrown
-
-***************************************************************************/
-
-
-public static Hash createRandomSeed (Preimages)(Preimages preimages) @safe
-in
-{
-    static assert (isInputRange!Preimages);
-}
-do
-{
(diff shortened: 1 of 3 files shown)


source/agora/consensus/data/Block.d
```diff
@@ -597,7 +569,7 @@ version (unittest)
         assert(revealed.length == key_pairs.length - missing_validators.length);
         try
         {
-            Hash random_seed = createRandomSeed(pre_images);
+            Hash random_seed = pre_images.reduce!((a, b) => hashMulti(a, b));
```
Geod24, line 572: In the future I would like to remove this duplication (we don't want it to get out of sync), but for the moment this is acceptable. The long term plan would be to make all tests that want to create a block to use `Ledger` so that we never create invalid blocks (because we need so much context to create a block nowadays).

source/agora/node/Ledger.d
```diff
@@ -922,26 +922,46 @@ public class Ledger
...
         }
 
-        return this.enroll_man.getRandomSeed(valid_keys, height);
+        return keys
+            // Filter out entries in `missing_validators`
+            .enumerate.filter!(en => !missing_validators.canFind(en.index))
+            // Lookup
+            .map!(en => getPreimage(en.value))
+            // Generate the final value
+            .reduce!((a , b) => hashMulti(a, b));
```
Geod24, line 964: This is the kind of stuff that make me love this language.

linked0 (approved):
LGTM.
Isn't it better to merge two commits if the merged code is not too complicated because EM.getRandomSeed will be removed in the end?

Geod24:
> Isn't it better to merge two commits if the merged code is not too complicated because EM.getRandomSeed will be removed in the end?

Good point, squashing.

PR #2166, opened 2021-06-07
Use a queue for processing incoming envelopes

### LINKED ISSUE #2021: agora.node.Validator.Validator.receiveEnvelope blocks
It seems that `receiveEnvelope` is blocking, meaning the query will wait until completion (probably of some other network query) before returning to the caller.

```
2021-04-23 15:15:06,141 Info [TimeLog.API] - agora.node.Validator.Validator.receiveEnvelope: Elapsed time: 14 ms, 566 μs, and 8 hnsecs
```

Saw that in #1992 , while all the other calls finished in a sub microsecond time.

Geod24:
```
All incoming SCP envelopes are now added directly to a queue,
which is then consumed by a task. This prevents the clients from
blocking on `receiveEnvelope` calls (which itself might trigger
other `receiveEnvelope` calls), and avoid unnecessary delays.

Fixes #2021
```

This is a rebase of https://github.com/bosagora/agora/pull/2067 but instead of duplicating the store, we simply add a boolean to it.

### DIFF
--- a/source/agora/consensus/SCPEnvelopeStore.d
+++ b/source/agora/consensus/SCPEnvelopeStore.d
@@ -53,7 +53,7 @@ public class SCPEnvelopeStore
         this.db = db;
 
         this.db.execute("CREATE TABLE IF NOT EXISTS scp_envelopes " ~
-            "(seq INTEGER PRIMARY KEY AUTOINCREMENT, envelope BLOB NOT NULL)");
+            "(seq INTEGER PRIMARY KEY AUTOINCREMENT, envelope BLOB NOT NULL, processed INTEGER NOT NULL)");
     }
 
     /***************************************************************************
@@ -63,14 +63,15 @@ public class SCPEnvelopeStore
         First, clean with 'removeAll' before Adding it in new envelopes.
 
         Params:
-            envelope = the envelop to add
+            envelope = the envelope to add
+            processed = If the envelope has been processed
 
         Returns:
             true if the envelope has been added to the database
 
     ***************************************************************************/
 
-    public bool add (const ref SCPEnvelope envelope) @safe nothrow
+    public bool add (const ref SCPEnvelope envelope, bool processed) @safe nothrow
     {
         static ubyte[] envelope_bytes;
 
@@ -88,8 +89,8 @@ public class SCPEnvelopeStore
         try
         {
             () @trusted {
-                db.execute("INSERT INTO scp_envelopes (envelope) VALUES(?)",
-                    envelope_bytes);
+                db.execute("INSERT INTO scp_envelopes (envelope, processed) VALUES(?, ?)",
+                           envelope_bytes, processed);
             }();
         }
         catch (Exception ex)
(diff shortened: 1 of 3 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -186,6 +193,25 @@ extern(D):
         this.restoreSCPState();
         this.nomination_interval = nomination_interval;
         this.acceptBlock = externalize;
+        this.envelope_timer = this.taskman.setTimer(10.msecs,
```
hewison-chris, line 196: Is it worth making this configurable? I think it was the comment I had last time.

source/agora/consensus/SCPEnvelopeStore.d
```diff
@@ -131,18 +132,19 @@ public class SCPEnvelopeStore
...
         () @trusted
         {
             auto results = this.db.execute(
-                "SELECT envelope FROM scp_envelopes");
+                "SELECT envelope, processed FROM scp_envelopes");
 
             foreach (ref row; results)
             {
                 auto env = deserializeFull!(const SCPEnvelope)(row.peek!(ubyte[])(0));
+                auto proc = row.peek!bool(1);
```
hewison-chris: IMO `processed` is better as variable name.

hewison-chris (approved):
LGTM just a couple of suggestions.

linked0 (approved):
LGTM

source/agora/consensus/SCPEnvelopeStore.d
kchulj: Isn't `processed` already used above?

source/agora/consensus/SCPEnvelopeStore.d
```diff
@@ -181,12 +183,12 @@ unittest
...
     foreach (env; envelopes)
     {
-        envelope_store.add(env);
+        envelope_store.add(env, false);
     }
 
     assert(envelope_store.length == 2);
 
-    foreach (const ref SCPEnvelope env; envelope_store)
+    foreach (_, const ref SCPEnvelope env; envelope_store)
```
kchulj, line 191: Is the `_` used for indexing somewhere?

Geod24: Since there is only one `opApply` defined, we need to give 2 arguments to `foreach`. I didn't define a single argument `opApply` because, well, `opApply` is only used in two places, and one of them is this test.

source/agora/consensus/protocol/Nominator.d
Geod24: Actually I gave it some thought and maybe it should be merged with the nomination timer ?

hewison-chris: But the nomination timer is how frequently we check if it is time to nominate where as this is how often to check for envelopes that need processing. The envelope checking probably should be more frequent. Or is the overhead of a second timer the issue?

Geod24: Not worried about the overhead, but worried about race conditions

hewison-chris: We have `isNominating` set which we could check before processing envelopes.

kchulj (approved):
LGTM

source/agora/consensus/protocol/Nominator.d
Geod24: That sounds a bit brittle, but let's leave that discussion for another day.

source/agora/consensus/SCPEnvelopeStore.d
Geod24: Done.

PR #2170, opened 2021-06-08
SCP: Remove many dead types / functions and move towards replacing `Stellar-types.x`

Geod24:
As I was looking at our diff and trying to make some issues easier to debug, it struck me that I've been looking at the wrong level of abstraction: While I originally wanted to remove everything that is not in `Stellar-types.x` / `Stellar-SCP.x` (the later importing the former), as I expected everything else to be built on those foundations, I couldn't find a diff that would be satisfying for upstream (IOW, the changes were very convoluted). However, the fix is actually much simpler than anticipated: All the crypto magic is nicely encapsulated in `Stellar-types.x` (even `NodeID` !) so we can "just" replace this and we'll be good.

Hence, this PR, does (in order):
1) Remove unused types from `Stellar-types.x`;
2) Change `PublicKey` from a semi-complex type to a simple blob;
1.5) As I was doing 2, I realized that the only places that remotely relied on the internals of `PublicKey` were the string formatting, and those were unused, so I nuked them in the second commit to simplify the third.


linked0 (commented):
Great! I will approve this after the tests are green.

Geod24:
Unfortunately the last commit seemed to break some stuff, so I took it out. Now it's only removing a bunch of dead code.

Geod24:
All green!

linked0 (approved):

PR #2173, opened 2021-06-09
KeyPair: Remove 'const' on members

Geod24:
The original intent was to make it read-only,
but the result was that we could not store KeyPair
in an associative array.


kchulj (approved):

linked0 (approved):
Good. It was really needed.

PR #2175, opened 2021-06-09
Logger: Re-use configureLogger in AdminInterface

Geod24:
This just needed a tiny change to configureLogger interface,
but in the rush of implementing the endpoint, that wasn't noticed.


hewison-chris (approved):
LGTM

PR #2179, opened 2021-06-09
Update base32 submodule to reduce v2.097.0 deprecations

Geod24:
The new release deprecates the 'body' keyword, which triggers in a few places,
including in this submodule.


hewison-chris (approved):
LGTM

Issue #2137, opened 2021-06-01
Length upper limit needs to be better handled


Geod24:
This triggers in the live system:
```
Jun 01 06:52:41 eu-002 systemd[1]: Started Agora node 2.
Jun 01 06:52:43 eu-002 docker[229303]: Task terminated with uncaught exception: Value of 'length' exceeds upper bound (35965 > 31184)
Jun 01 06:52:43 eu-002 docker[229303]: Full exception: object.Exception@submodules/serialization/source/agora/serialization/Serializer.d(548): Value of 'length' exceeds upper bound (35965 > 31184)
Jun 01 06:52:43 eu-002 docker[229303]: ----------------
Jun 01 06:52:43 eu-002 docker[229303]: submodules/serialization/source/agora/serialization/Serializer.d:542 @safe ulong agora.serialization.Serializer.deserializeLength(scope const(ubyte)[] delegate(ulong) @safe, ulong) [0x55b81b3fe480]
Jun 01 06:52:43 eu-002 docker[229303]: submodules/serialization/source/agora/serialization/Serializer.d:583 @safe ubyte[] agora.serialization.Serializer.deserializeFull!(ubyte[]).deserializeFull(scope const(ubyte)[] delegate(ulong) @safe, in agora.serialization.Serializer.DeserializerOptions) [0x55b81adfc160]
Jun 01 06:52:43 eu-002 docker[229303]: source/agora/consensus/data/DataPayload.d:90 @safe agora.consensus.data.DataPayload.DataPayload agora.consensus.data.DataPayload.DataPayload.fromBinary!(agora.consensus.data.DataPayload.DataPayload).fromBinary(scope const(ubyte)[] delegate(ulong) @safe, in agora.serialization.Serializer.DeserializerOptions) [0x55b81adfc100]
Jun 01 06:52:43 eu-002 docker[229303]: submodules/serialization/source/agora/serialization/Serializer.d:583 @safe agora.consensus.data.DataPayload.DataPayload agora.serialization.Serializer.deserializeFull!(agora.consensus.data.DataPayload.DataPayload).deserializeFull(scope const(ubyte)[] delegate(ulong) @safe, in agora.serialization.Serializer.DeserializerOptions) [0x55b81ae3c9d0]
[...]

PR #2182, opened 2021-06-10
Minor serialization fixes for transaction

Geod24:
I was looking at #2137 and couldn't make sense of the failure. The only possible things I can think of would be caught by our `testSymmetry` check, so let's try to re-deploy again.


hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

PR #2183, opened 2021-06-10
Make `buildQuorumConfig` to use UTXO instead of PublicKey

linked0:
This change helps use UTXOs in NameRegistry. `Validator` makes a set of UTXO for peers, which will be used to register peers in NetworkManager.

Relates to #1309

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -35,6 +35,7 @@ import agora.common.Task;
 import agora.consensus.data.Block;
 import agora.consensus.data.Params;
 import agora.consensus.data.PreImageInfo;
+import agora.consensus.data.UTXO;
 import agora.consensus.data.ValidatorBlockSig;
 import agora.crypto.Hash;
 import agora.crypto.Key;
@@ -390,7 +391,7 @@ public class NetworkManager
     protected Set!Address connected_peers;
 
     /// Keeps track of Validator keys we're already connected to
-    private Set!PublicKey connected_validator_keys;
+    private Set!Hash connected_validator_keys;
 
     /// All known addresses so far (used for getNodeInfo())
     protected Set!Address known_addresses;
@@ -401,10 +402,10 @@ public class NetworkManager
     /// For a Validator, NodeInfo.state will be Complete only
     /// if it has connected to all of its quorum peers.
     /// See 'minPeersConnected'
-    private Set!PublicKey required_peer_keys;
+    private Set!Hash required_peers;
 
     /// Current quorum set
-    private Set!PublicKey quorum_set_keys;
+    private Set!Hash quorum_set_keys;
 
     /// Address ban manager
     protected BanManager banman;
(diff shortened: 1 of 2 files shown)


source/agora/node/Validator.d
```diff
@@ -200,6 +205,13 @@ public class Validator : FullNode, API
             log.fatal("rebuildQuorumConfig: Exception thrown: {}", e);
             assert(0);
         }
+
+        foreach (hash; keys.filter!(key => key != this_enroll).array)
```
hewison-chris: Is the `.array` required?

hewison-chris (approved):
LGTM

source/agora/node/Validator.d
linked0: I just realized that it isn't. Will remove.

source/agora/node/Validator.d
```diff
@@ -200,6 +205,13 @@ public class Validator : FullNode, API
             log.fatal("rebuildQuorumConfig: Exception thrown: {}", e);
             assert(0);
         }
+
+        foreach (hash; keys.filter!(key => key != this_enroll))
+        {
+            UTXO utxo;
+            assert(this.utxo_set.peekUTXO(hash, utxo));
+            this.required_peer_utxos[hash] = utxo;
+        }
```
Geod24: This should probably be in `buildRequiredKeys` instead (and the removed call be reinstated).

source/agora/node/Validator.d
```diff
@@ -140,8 +140,6 @@ public class Validator : FullNode, API
         static QuorumConfig[] other_qcs;
         this.rebuildQuorumConfig(this.qc, other_qcs, height);
         this.nominator.setQuorumConfig(this.qc, other_qcs);
-        buildRequiredKeys(this.config.validator.key_pair.address, this.qc,
-            this.required_peer_keys);
```
Geod24, line 144: Why the removal ?

source/agora/network/Manager.d
```diff
@@ -642,47 +643,60 @@ public class NetworkManager
...
+                        },
+                        )(taskman, 3, 2.seconds, "Exception happened while trying to get validator addresses");
+                    });
+                }
             }
         }
+        catch (Exception ex)
+        {
+            log.error("Error discovering peers: {}", ex);
+        }
```
Geod24: Looks like the `try` clause is only because of the `foreach` ? If so, `byKeyValue`.

Geod24 (changes requested):

source/agora/node/Validator.d
linked0: Right, I will reinstate it.

source/agora/network/Manager.d
linked0: Will update.

source/agora/node/Validator.d
linked0: Ditto

linked0:
Ready for review.
I reinstated the `buildRequiredKeys` but it needs `utxo_keys` as  `rebuildQuorumConfig` does. So I hand over the variable as a parameter into the two functions in order to remove redundant calls of the functions of the `EM`.

@Geod24 What do you think I should do as the next step for solving this issue?

Geod24 (approved):

Geod24:
> @Geod24 What do you think I should do as the next step for solving this issue?

Unfortunately, I don't think we can switch the registry just yet, so the next step is probably to switch SCP (I've been doing some changes for that purpose), so we can at least refer to validators only by their UTXO hash in Agora (and remove the check that a validator is not already enrolled with a public key).

PR #2185, opened 2021-06-10
Move implementation of `NodeID`-formatting functions from C++ to D

Geod24:
I don't expect this to work right away, as last time I checked our bindings were incomplete, but it did pass on my machine, much to my surprise.


Geod24:
Looks like this could be made to work. We probably need to copy the binding though, as LDC doesn't seem to define `CppRuntime` on Musl (https://github.com/ldc-developers/druntime/blob/645d14c0df55f20e1d4fb72331a88e0d727c40fe/src/core/stdcpp/string.d#L13-L19),

Geod24:
This works \o/

omerfirmak (approved):

linked0 (approved):
LGTM

PR #2188, opened 2021-06-11
Cherry pick some commit from #2097

Geod24:
Those can already be merged, so let's do that and reduce the size of this PR.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -1263,7 +1263,7 @@ public class TestNetworkManager : NetworkManager
     ***************************************************************************/
 
     protected override TestBanManager getBanManager (in BanManager.Config conf,
-        Clock clock, cstring data_dir)
+        Clock clock, string data_dir)
     {
         return new TestBanManager(conf, clock, data_dir);
     }
--- a/tests/unit/BanManager.d
+++ b/tests/unit/BanManager.d
@@ -29,10 +29,12 @@ import core.time;
 ///
 private void main ()
 {
+    import std.path : buildPath;
+
     auto path = makeCleanTempDir();
 
     BanManager.Config conf = { max_failed_requests : 10, ban_duration : 60.seconds };
-    auto banman = new BanManager(conf, new Clock(null, null), path);
+    auto banman = new BanManager(conf, new Clock(null, null), buildPath(path, "banned.dat"));
     banman.load();  // should not throw
 
     const IP = "127.0.0.1";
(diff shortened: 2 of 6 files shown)


source/agora/test/Base.d
```diff
@@ -1263,7 +1263,7 @@ public class TestNetworkManager : NetworkManager
     ***************************************************************************/
 
     protected override TestBanManager getBanManager (in BanManager.Config conf,
-        Clock clock, cstring data_dir)
+        Clock clock, string data_dir)
```
linked0, line 1266: Can you explain what the difference is?

tests/unit/BanManager.d
```diff
@@ -41,7 +43,7 @@ private void main ()
     assert(banman.isBanned(IP));
     banman.dump();
 
-    auto new_banman = new BanManager(conf, new Clock(null, null), path);
+    auto new_banman = new BanManager(conf, new Clock(null, null), buildPath(path, "banned.dat"));
```
linked0, line 46: Isn't it right to use the different file?

linked0 (approved):
LGTM with some questions.
And, a part of the first commit's message is not clear, which is "we might not want a node from receiving blockchain data."

source/agora/test/Base.d
ferencdg: cstring is more permissive than string in an interface, so cstring should be used:
1. you can pass a string to a cstring argument
2. you cannot pass a cstring to a string argument

honestly, I cannot remember why I made this change... I think it is because this.banfile_path is defined as const string and if you have a cstring then you cannot assign it to const string. And changing this.banfile_path to const cstring would cause other methods(which methods take string) to cause compile error.

I also want to be able to call buildPath("a","b") and pass the result to getBanManager() and buildPath returns an immutable array.

ferencdg: I think changing cstring to const cstring in the parameters would have worked too

Geod24: What @ferencdg . I also questioned the change originally, but it is necessary for now. Perhaps later we could just switch the BanManager to use the DB and pass the DB instance, as we already changed for other classes.

Geod24:
Fixed commit message.

tests/unit/BanManager.d
Geod24: No, the test explicitly tests loading.

ferencdg: we use the same file: path was just a directory before and the filename was always banned.dat

linked0: Got it, ty.

PR #2190, opened 2021-06-13
Add the genesis timestamp to the stats

Geod24:
We are moving towards setting the genesis timestamp dynamically
(on TestNet deploy), and hence we should add it to the stats
to make monitoring easier.


hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

PR #2196, opened 2021-06-15
Remove `getActiveValidatorPublicKeys` and use `getValidators`

Geod24:
This should help streamline some code. 
Will submit a follow up PR for `getEnrolledUTXOs` too.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -983,8 +983,9 @@ extern(D):
         // new network connections might need to be established).
         if (!this.acceptBlock(signed_block))
         {
-            log.error("Block was not accepted by node {}", this.kp.address);
-            assert(0, format!"Block was not accepted"());
+            auto msg = this.ledger.validateBlock(signed_block);
+            log.error("Block was not accepted by node {}: {}", this.kp.address, msg);
+            assert(0, "Block was not accepted: " ~ msg);
         }
     }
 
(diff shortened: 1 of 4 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -983,8 +983,9 @@ extern(D):
         // new network connections might need to be established).
         if (!this.acceptBlock(signed_block))
         {
-            log.error("Block was not accepted by node {}", this.kp.address);
-            assert(0, format!"Block was not accepted"());
+            auto msg = this.ledger.validateBlock(signed_block);
```
linked0, line 986: The `acceptBlock` already has the following code. Is calling the `validateBlock` in the above code just for showing the error message explicitly again?
```
if (auto fail_reason = this.validateBlock(block, file, line))
{
    log.trace("Rejected block: {}: {}", fail_reason, block.prettify());
    return false;
}
```

Geod24: Yep, because at the moment, the message might not get printed because the level is `Trace`.

linked0 (approved):
LGTM

source/agora/consensus/protocol/Nominator.d
Geod24: To be precise, the nodes stop responding to any request because they just crash after generating an invalid block, which means the regular log-dumping done by our `assert` handler doesn't happen (it's triggered from the main thread).

hewison-chris (commented):
Nice

PR #2205, opened 2021-06-16
Ledger: Use `getValidators` in `getRandomSeed`

Geod24:
As I am working on removing usages of `getEnrolledUTXOs`, a few things came up:
- We were looking up slashed validators at genesis, getting back `keys.length == 0` and ignoring it;
- We didn't always have our pre-image in the ValidatorSet, which led to self-slashing, and ripple through the code (e.g. #2203 );

Those two bugs are fixed. I also reduced usage of `getEnrolledUTXOs` where it was low risk, and (tests / stats / whitelist) and the last commit is the titular one, which simplifies error handling.

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -496,6 +496,14 @@ public class Validator : FullNode, API
             block.header.time_offset > cur_offset - 3 * this.params.BlockInterval.total!"seconds")
             this.checkAndEnroll(block.header.height);
 
+        // FIXME: Add our pre-image to the validator set so that `getValidators`
+        // works as expected. This will need to be fixed in the Ledger in the
+        // future, as `checkRevealPreimage` has some issues, but doing it here
+        // allows us to unify usage of `getValidators`.
+        PreImageInfo self;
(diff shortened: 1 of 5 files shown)


linked0 (approved):
LGTM

source/agora/node/Validator.d
```diff
@@ -496,6 +496,14 @@ public class Validator : FullNode, API
             block.header.time_offset > cur_offset - 3 * this.params.BlockInterval.total!"seconds")
             this.checkAndEnroll(block.header.height);
 
+        // FIXME: Add our pre-image to the validator set so that `getValidators`
+        // works as expected. This will need to be fixed in the Ledger in the
+        // future, as `checkRevealPreimage` has some issues, but doing it here
```
linked0, line 501: Can you tell me just one of those issues on `checkRevealPreimage`?

Geod24: One example is that it's not correlated with the block time or even the nomination time. Which means we rely on the user's virtue to have it properly configured. For example in network unittests, it's currently 1 second, but we nominate every 100 msecs... Another one is that we *never* ask our peers for preimages, we rely on them sending it (I have a commit to fix this).

PR #2209, opened 2021-06-16
Relay TXs only if node has a Validator peer

### LINKED ISSUE #1726: Node should try to gossip transactions after network state is complete
Consider this scenario:

- User boots up their FullNode / Validator
- They have their wallet send a transaction to only their own node
- The node hasn't completed network discovery (maybe it's slow, network is delayed..). Therefore `putTransaction` accepts the tx in the pool but does not gossip it because it has nobody to gossip to.

End result: the tx is now stuck forever in the pool.

If there are incoming transactions which are accepted in the pool but the network state is incomplete then the node should queue these transactions for gossiping later.

omerfirmak:
Fixes #1726

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -1079,21 +1079,25 @@ public class NetworkManager
     }
 
     ///
-    private bool minPeersConnected ()  pure nothrow @safe
+    private bool minPeersConnected () nothrow @safe
     {
         return this.required_peers.length == 0 &&
-            this.peers[].walkLength >= this.node_config.min_listeners;
+            this.peers[].walkLength >= this.node_config.min_listeners &&
+            this.validators().filter!(node =>
+                !this.banman.isBanned(node.client.address)).count != 0;
     }
 
     private bool peerLimitReached ()  nothrow @safe
     {
         return this.required_peers.length == 0 &&
             this.peers[].filter!(node =>
-                !this.banman.isBanned(node.client.address)).count >= this.node_config.max_listeners;
+                !this.banman.isBanned(node.client.address)).count >= this.node_config.max_listeners &&
+            this.validators().filter!(node =>
+                !this.banman.isBanned(node.client.address)).count != 0;
     }
 
     /// Returns: the list of node IPs this node is connected to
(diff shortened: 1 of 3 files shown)


Geod24:
Doesn't it just push down the issue ? Shouldn't nodes try to periodically sync tx pools instead ?

omerfirmak:
> Doesn't it just push down the issue ? Shouldn't nodes try to periodically sync tx pools instead ?

That may become too costly, no? Especially, when we reach MB levels of tx pool size. Nodes would at least need to share hashes of all the TXs in their pools. What is important is that we share the TXs that are worth externalizing and that is what consensus already does.

My take on this is that tx pools of FullNodes are not important at all. We could probably just make it a Validator thing IMO and FullNodes would only act as relays.

TXs should probably not be valid forever either, they should have an expiration date/block, so that the wallet knows when to retry.

omerfirmak:
I have checked to see if Bitcoin has a mechanism to keep node's mempools in sync and AFAICS there is nothing of the sort.
So we are probably better off with the best effort approach we already have but avoid trying to gossip to an empty set of peers.

ferencdg:
wouldn't it be easier, if the node that is starting up, and haven't finished the network discovery would just not accept the transaction? Wallets should have retry mechanism anyway, no?

omerfirmak:
> 
> 
> wouldn't it be easier, if the node that is starting up, and haven't finished the network discovery would just not accept the transaction? Wallets should have retry mechanism anyway, no?

It may never be able to complete the discovery tho.

omerfirmak:
@Geod24 mentioned the existence of INV messages in BCH. https://www.reddit.com/r/Bitcoincash/comments/jko0ty/bchn_mr_746_merged_faster_transaction_relay/

This makes sense for Bitcoin and its forks as they use PoW but we already do this in the nomination round. So our `Validator` nodes will never suffer from this problem anyways.

Our issue is to somehow get the TX from a `FullNode` to a `Validator` without it ever getting stuck in the `FullNode` forever.

What I think will work is to keep the if check this PR adds and make sure that a `FullNode` always keeps trying to maintain a connection to a `Validator` throughout its lifetime.

Geod24 (approved):

PR #2213, opened 2021-06-18
ValidatorCleanRestart: Ask for UTXOs instead of assuming unspent

Geod24:
```
This test was assuming certain UTXOs were unspent, which could backfire
should modifications lead to them being spent.
In order to make the test more resilient, `TestAPI` now exposes a method
to get enough UTXOs to match a certain amount.
This might need some changes in the future, e.g. w.r.t. fees,
but is a much more robust approach.

Additionally, the test was calling `generateBlock`, which had two undesirable effects:
- It would create another transaction, which *could* compete with the frozen one,
  meaning the frozen tx would not be externalized when we expected it to,
  but the `assertSameBlock` will pass.
- In case the frozen tx was rejected for any reason, it will still pass the
  `assertSameBlock` and hide the real error.
```

This has been sitting on my computer for a bit, figured I should submit it for review. Not 100% sure it fixes the bug though, will need to double check.

### DIFF
--- a/source/agora/test/ValidatorCleanRestart.d
+++ b/source/agora/test/ValidatorCleanRestart.d
@@ -17,6 +17,7 @@ module agora.test.ValidatorCleanRestart;
 version (unittest):
 
 import agora.api.FullNode;
+import agora.common.Amount;
 import agora.consensus.data.Enrollment;
 import agora.consensus.data.Transaction;
 import agora.crypto.Key;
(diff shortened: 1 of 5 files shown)


source/agora/test/ValidatorCleanRestart.d
```diff
@@ -50,18 +51,18 @@ version(none) unittest
...
     // Block 19 we add the freeze utxos for set_b validators
     // prepare frozen outputs for outsider validators to enroll
-    blocks[0].spendable().drop(1)
-        .map!(txb => txb
-            .split(keys).sign(OutputType.Freeze))
-        .each!(tx => set_a[0].putTransaction(tx));
+    TxBuilder txb = TxBuilder(WK.Keys.AAA.address); // Refund
+    utxos.each!(pair => txb.attach(pair.utxo.output, pair.hash));
+    auto toSend = txb.draw(Amount.MinFreezeAmount, keys).sign(OutputType.Freeze);
+    set_a.each!(n => n.putTransaction(toSend));
```
linked0:
```suggestion
    auto to_send = txb.draw(Amount.MinFreezeAmount, keys).sign(OutputType.Freeze);
    set_a.each!(n => n.putTransaction(to_send));
```

linked0 (approved):
LGTM one nit

kchulj (approved):
LGTM

source/agora/test/ValidatorCleanRestart.d
Geod24: Done

PR #2215, opened 2021-06-21
Populate our pre-image to the ValidatorSet after catchup

Geod24:
```
One scenario that was missing in PR2205 was the state right after construction.
Because onAcceptedBlock is not called when replaying blocks (or starting from scratch),
to avoid blocks being re-written to disk and stats being wrong,
the initial state was still faulty. It is now fixed.
```

CC @hewison-chris

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -99,6 +99,13 @@ public class Validator : FullNode, API
             this.clock, this.network, vledger, this.enroll_man, this.taskman);
         this.nominator.onInvalidNomination = &this.invalidNominationHandler;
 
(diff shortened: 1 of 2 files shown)


hewison-chris (approved):
LGTM

hewison-chris:
Looks like some tests fail though.

kchulj (approved):
LGTM
The CI fails, as @hewison-chris mentioned.

Geod24:
Found the "bug": There is a test that test explicitly that pre-images aren't propagated. Removed it, with explanations.

Geod24:
Can I have a re-review ?

linked0 (approved):
LGTM

PR #2216, opened 2021-06-21
Rename pre-image revelation function to improve tests

Geod24:
```
Modifying the behavior of the EnrollmentManager is problematic,
as other parts of the code may depend on it.
For example, the node currently special-case itself in `hasRevealedPreimage`
to work around this. If we want to avoid revealing pre-image,
just override the function that is supposed to reveal them,
not the function that provides them.
```


hewison-chris (approved):
LGTM, makes sense as long as we always fetch `preimages` from `validatorSet` otherwise we would have to override more functions in our test code.

linked0 (approved):
LGTM

kchulj (approved):
LGTM

PR #2218, opened 2021-06-21
SCPDriver: Simplify getHashOf implementation

Geod24:
Was debugging something else when I came across this.
This is almost a copy of `hashMulti`, save for the `foreach`.
I think it was made this way because in the past, `hashPart` didn't support types that implicitly convert to a `ubyte[]` ([this](https://github.com/bosagora/crypto/blob/4899b86ca5b902335c43a91b28107cd2320316af/source/agora/crypto/Hash.d#L159-L163)).


kchulj (approved):

hewison-chris (approved):
LGTM

PR #2220, opened 2021-06-21
SCP: Reduce dependencies, simplify

Geod24:
Some of those fixes have been submitted upstream (https://github.com/stellar/stellar-core/pull/3090). Some of the commits in that PRs have already been integrated here.
The last commit removes a workaround that was introduced when SCP was upgraded.

### DIFF
--- a/source/scpd/scp/LocalNode.d
+++ b/source/scpd/scp/LocalNode.d
@@ -40,11 +40,17 @@ extern(C++, class) public struct LocalNode
     Hash            gSingleQSetHash;  // hash of the singleton qset
     SCPQuorumSetPtr mSingleQSet;      // {{mNodeID}}
 
-    SCP* mSCP;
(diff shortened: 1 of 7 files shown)


Geod24:
Okay this actually uncovered 3 bugs in our code, amazing.

Geod24:
> Okay this actually uncovered 3 bugs in our code, amazing.

Bring that number up a notch. In addition to the logging-related bugs, turns out that:
1) We truncate public key (64 -> 32 bytes) because we follow Stellar's type;
2) Our `toStrKey` output is garbage (which might or might not be due to 1);

Will need to get those fixed before merging this.

linked0:
LGTM, Will approve after the fix is applied.

Geod24:
> Will need to get those fixed before merging this.

Actually, the bug is already present in HEAD, so merging this will not regress anything. At least it will make it consistent, so merging.

PR #2223, opened 2021-06-21
C++ logger: Catch Exceptions thrown while logging

Geod24:
Due to a bug in the logger that has now been fixed,
Exceptions were being thrown, and led to a cryptic abort.
This should make the UX much better while debugging such issues.


hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

PR #2225, opened 2021-06-22
SCP: Remove SecretKey and direct usages of uint256

Geod24:
This removes quite a lot of useless code that has been bloating the SCP layer for a while.
It should make it much clearer what type is used for what, and make changing `NodeID` much easier.
CC @linked0


linked0 (approved):
LGTM. Ty.

kchulj (approved):

PR #2226, opened 2021-06-22
SCP: Remove usage of PublicKey

Geod24:
```
This finally removes the last usages of 'PublicKey',
ensuring that only 'NodeID' is used through the code base.
```

@linked0 : Realized I missed a few occurrences, this should be the last of it.

### DIFF
--- a/source/scpp/src/xdr/Stellar-SCP.h
+++ b/source/scpp/src/xdr/Stellar-SCP.h
@@ -10,8 +10,6 @@
 #include "xdr/Stellar-types.h"
 
 namespace stellar {
-
-using NodeID = PublicKey;
 using Value = xdr::opaque_vec<>;
 
 struct SCPBallot {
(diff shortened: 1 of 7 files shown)


source/scpp/src/xdr/Stellar-SCP.h
```diff
@@ -10,8 +10,6 @@
 #include "xdr/Stellar-types.h"
 
 namespace stellar {
-
-using NodeID = PublicKey;
```
linked0, line 14: Does it need to be `using NodeID = Hash;`?

Geod24: I didn't switch it to `Hash`, it's still a `PublicKey` (`uint256`) under the hood, so your PR is still needed, but will be much simpler after this.

Geod24: This was moved to `Agora-types.h` BTW, `using NodeID = uint256;`

linked0: Got it. Ty.

linked0 (approved):
LGTM

PR #2228, opened 2021-06-28
Move `handshake` (aka `getPublicKey`) endpoint to FullNode

Geod24:
One should never rely on `Exception` for control flow, but we were doing just that when establishing a connection.
As explained in the commit message, the assumptions that led to `getPublicKey` being used the way it is turned out to be wrong:
```
When 'getPublicKey' was first introduced, it was under the assumption that
a node would be either a FullNode or a Validator.
However, this had a major oversight: while a FullNode will never promote
to a Validator internally, from the outside, a node could switch between
a FullNode and a Validator (because an unenrolled Validator is no different
from a FullNode). This led to a few complications in the NetworkManager
and related code, including this hack with HTTPStatusException.
This is now partially fixed by moving the 'handshake' endpoint to FullNode.
```

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -181,11 +181,10 @@ public class NetworkManager
                     import agora.flash.OnionPacket : generateSharedSecret;
                     import libsodium.crypto_auth;
 
-                    // LocalRest will return PublicKey.init,
-                    // vibe.d will throw HTTPStatusException with status == 404
                     const ephemeral_kp = KeyPair.random();
-                    auto id = client.getPublicKey(ephemeral_kp.address);
+                    auto id = client.handshake(ephemeral_kp.address);
 
+                    // No identity, either a full node or not enrolled
                     if (id.key == PublicKey.init)
                         break;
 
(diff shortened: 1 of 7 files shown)


Geod24:
Note: `getPublicKey` was not changed yet as other parts rely on it. I'm still working on simplifying the `NetworkManager`.

hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

linked0 (approved):
LGTM

source/agora/network/Manager.d
```diff
@@ -205,13 +207,6 @@ public class NetworkManager
                 }
                 catch (Exception ex)
                 {
-                    if (auto http = cast(HTTPStatusException)ex)
-                    {
-                        // 404 => API not implemented, it's a FullNode
-                        if (http.status == 404)
-                            break;
-                    }
-
```
Geod24, line 214: Actually since I did that change, I had to change `getPublicKey` to `handshake`, otherwise the full node ended up being banned,

source/agora/network/Manager.d
```diff
@@ -79,6 +76,12 @@ public class NetworkManager
 
         /// Client
         NetworkClient client;
+
+        ///
+        public bool isValidator () const scope @safe pure nothrow @nogc
+        {
+            return this.key != PublicKey.init;
```
omerfirmak, line 83: at some point we should probably make this check if the peer is an active validator too.

Geod24: Yeah I'm thinking of changing it to check for the UTXO hash in the future, but it requires to add a dep to the Ledger.

PR #2231, opened 2021-06-28
Talos: Make it usable by fixing the format and displaying errors

Geod24:
This makes Talos somewhat usable. By default, it will still error out with "Error: Network section must not be empty", however:
- The error is now properly displayed. No more "Error...";
- Iff the user provide a correct entry for the "network" section, a config file is written;
- A couple more issues have been fixed as can be seen from the list of commits;


kchulj (approved):
LGTM.
Couldn't fully review 5th commit due to limited `js` knowledge.

PR #2235, opened 2021-06-29
Fixed nomination bug caused by missing validators

### LINKED ISSUE #2152: Invalid consensus data assertion being triggered
```
Unittesting agora.test.NetworkClient================================ ASSERT HANDLER ===============================
[source/agora/consensus/protocol/Nominator.d:1151] Assertion thrown during test: combineCandidates: Invalid consensus data: The list of missing validators does not match with the local one. [0] != []
/Users/runner/work/agora/agora/source/agora/consensus/protocol/Nominator.d:1151 int agora.consensus.protocol.Nominator.Nominator.combineCandidates(core.stdc.config.__c_ulonglong, ref const(scpd.Cpp.set!(scpd.Cpp.shared_ptr!(scpd.scp.SCPDriver.ValueWrapper).shared_ptr, scpd.scp.SCPDriver.WrappedValuePtrComparator, scpd.Cpp.allocator!(scpd.Cpp.shared_ptr!(scpd.scp.SCPDriver.ValueWrapper).shared_ptr).allocator).set)).__foreachbody4(ref const(scpd.Cpp.shared_ptr!(scpd.scp.SCPDriver.ValueWrapper).shared_ptr)) [0x10df7e104]
/Users/runner/work/agora/agora/source/scpd/Cpp.d:222 _ZNSt3__13setINS_10shared_ptrIN7stellar12ValueWrapperEEENS2_25WrappedValuePtrComparatorENS_9allocatorIS4_EEE7opApply7wrapperEPvRKS4_ [0x10dc41c16]
??:? _Z15cpp_set_foreachINSt3__110shared_ptrIN7stellar12ValueWrapperEEEEiPvS5_S5_ [0x10e6c7352]
/Users/runner/work/agora/agora/source/scpd/Cpp.d:225 const int scpd.Cpp.set!(scpd.Cpp.shared_ptr!(scpd.scp.SCPDriver.ValueWrapper).shared_ptr, scpd.scp.SCPDriver.WrappedValuePtrComparator, scpd.Cpp.allocator!(scpd.Cpp.shared_ptr!(scpd.scp.SCPDriver.ValueWrapper).shared_ptr).allocator).set.opApply(scope int delegate(ref const(scpd.Cpp.shared_ptr!(scpd.scp.SCPDriver.ValueWrapper).shared_ptr))) [0x10dc41bcc]
/Users/runner/work/agora/agora/source/agora/consensus/protocol/Nominator.d:1141 _ZN9Nominator17combineCandidatesEyRKNSt3__13setINS0_10shared_ptrIN7stellar12ValueWrapperEEENS3_25WrappedValuePtrComparatorENS0_9allocatorIS5_EEEE [0x10dcbc0f3]
??:? _ZN7stellar18NominationProtocol15processEnvelopeENSt3__110shared_ptrINS_18SCPEnvelopeWrapperEEE [0x10e7232c8]
??:? _ZN7stellar4Slot15processEnvelopeENSt3__110shared_ptrINS_18SCPEnvelopeWrapperEEEb [0x10e785346]
??:? _ZN7stellar3SCP15receiveEnvelopeENSt3__110shared_ptrINS_18SCPEnvelopeWrapperEEE [0x10e76b62c]
/Users/runner/work/agora/agora/source/agora/consensus/protocol/Nominator.d:575 @trusted void agora.consensus.protocol.Nominator.Nominator.receiveEnvelope(in ref scpd.types.Stellar_SCP.SCPEnvelope) [0x10dcbeb78]
/Users/runner/work/agora/agora/source/agora/node/Validator.d:356 @safe void agora.node.Validator.Validator.receiveEnvelope(scpd.types.Stellar_SCP.SCPEnvelope) [0x10dcb6c2e]
/Users/runner/work/agora/agora/submodules/localrest/source/geod24/LocalRest.d-mixin-667:684 void geod24.LocalRest.RemoteAPI!(agora.test.Base.TestAPI, agora.test.Base.Serializer).RemoteAPI.handleCommand(geod24.LocalRest.Command, agora.test.Base.TestAPI, geod24.LocalRest.FilterAPI, geod24.concurrency.Channel!(geod24.LocalRest.Response).Channel) [0x10dd9d81a]
/Users/runner/work/agora/agora/submodules/localrest/source/geod24/LocalRest.d:801 void geod24.LocalRest.RemoteAPI!(agora.test.Base.TestAPI, agora.test.Base.Serializer).RemoteAPI.spawned!(agora.test.Base.TestValidatorNode).spawned(geod24.concurrency.Channel!(geod24.LocalRest.Connection).Channel, immutable(char)[], int, agora.common.Config.Config, shared(geod24.Registry.Registry!(agora.test.Base.TestAPI).Registry)*, shared(geod24.Registry.Registry!(agora.registry.API.NameRegistryAPI).Registry)*, immutable(agora.consensus.data.Block.Block)[], in ref agora.test.Base.TestConf, shared(ulong)*).runNode().__lambda4().__lambda14() [0x10ddb27e4]
context.d:46 void core.thread.context.Callable.opCall() [0x10eddef9e]
fiber.d:1129 void core.thread.fiber.Fiber.run() [0x10eddf12d]
fiber.d:186 fiber_entryPoint [0x10eddf062]
```

Another ConsensusData resolution issue that needs to be addressed.

ferencdg:
Previously the change in the missing validators caused our SCP votes to
change from 'Yes' to 'No' triggering asserts in the SCP code. This has
been fixed by introducing lower and upper bound of acceptable missing
validators.

fixes: #2152

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -141,6 +141,9 @@ public extern (C++) class Nominator : SCPDriver
     /// Nomination start time
     protected TimePoint nomination_start_time;
 
+    /// The missing validators at the start of the nomination round
+    protected uint[] initial_missing_validators;
+
     /// List of incoming SCPEnvelopes that need to be processed
     private DList!SCPEnvelope queued_envelopes;
 
@@ -336,7 +339,7 @@ extern(D):
         }
 
         // check whether the consensus data is valid before nominating it.
-        if (auto msg = this.ledger.validateConsensusData(data))
+        if (auto msg = this.ledger.validateConsensusData(data, data.missing_validators))
         {
             this.log.error("prepareNominatingSet(): Invalid consensus data: {}. Data: {}",
                     msg, data.prettify);
@@ -409,13 +412,16 @@ extern(D):
             return;
         }
 
-        if (this.nomination_start_time == 0)
+        if (!this.is_nominating)
             this.nomination_start_time = cur_time;
 
         ConsensusData data;
         // `prepareNomintingSet` will log something if it returns `false`
         if (!this.prepareNominatingSet(data))
             return;
+
+        if (!this.is_nominating)
+            this.initial_missing_validators = data.missing_validators;
 
         log.info("Nominating {} at {}", data.prettify, cur_time);
         this.is_nominating = true;
@@ -880,7 +886,7 @@ extern(D):
             return ValidationLevel.kInvalidValue;
         }
 
-        if (auto fail_reason = this.ledger.validateConsensusData(data))
+        if (auto fail_reason = this.ledger.validateConsensusData(data, this.initial_missing_validators))
         {
             log.error("validateValue(): Validation failed: {}. Data: {}",
                 fail_reason, data.prettify);
@@ -969,6 +975,7 @@ extern(D):
         this.gossipBlockSignature(ValidatorBlockSig(height, this.kp.address,
                     this.slot_sigs[height][this.kp.address].s));
         this.nomination_start_time = 0;
+        this.initial_missing_validators = [];
     }
 
     /// function for verifying the block which can be overriden in byzantine unit tests
@@ -1193,10 +1200,7 @@ extern(D):
                 auto data = deserializeFull!ConsensusData(candidate[]);
                 log.trace("Consensus data: {}", data.prettify);
 
-                // Only allowed in unittests, as validating the consensus data
-                // for all the candidates might take long.
-                version (unittest)
-                if (auto msg = this.ledger.validateConsensusData(data))
+                if (auto msg = this.ledger.validateConsensusData(data, this.initial_missing_validators))
                     assert(0, format!"combineCandidates: Invalid consensus data: %s"(
                         msg));
 
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -626,13 +626,16 @@ public class Ledger
 
         Params:
             data = consensus data
+            initial_missing_validators = missing validators at the beginning of
+               the nomination round
 
         Returns:
             the error message if validation failed, otherwise null
 
     ***************************************************************************/
 
-    public string validateConsensusData (in ConsensusData data) @trusted nothrow
+    public string validateConsensusData (in ConsensusData data,
+        in uint[] initial_missing_validators) @trusted nothrow
     {
         const validating = this.getBlockHeight() + 1;
         auto utxo_finder = this.utxo_set.getUTXOFinder();
@@ -685,7 +688,7 @@ public class Ledger
                 return fail_reason;
         }
 
-        if (auto fail_reason = this.validateSlashingData(validating, data))
+        if (auto fail_reason = this.validateSlashingData(validating, data, initial_missing_validators))
             return fail_reason;
 
         return validateBlockTimeOffset(last_block.header.time_offset, data.time_offset,
@@ -698,13 +701,16 @@ public class Ledger
 
         Params:
             data = consensus data
+            initial_missing_validators = missing validators at the beginning of
+               the nomination round
 
         Returns:
             the error message if validation failed, otherwise null
 
     ***************************************************************************/
 
-    public string validateSlashingData (in Height height, in ConsensusData data) @safe nothrow
+    public string validateSlashingData (in Height height, in ConsensusData data,
+        in uint[] initial_missing_validators) @safe nothrow
     {
         // If we are enrolled and not slashing ourselves
         if (this.enroll_man.isEnrolled(height, &this.utxo_set.peekUTXO) && this.isSelfSlashing(height, data))
@@ -713,7 +719,7 @@ public class Ledger
             assert(0);
         }
 
-        return this.isInvalidPreimageRootReason(height, data.missing_validators);
+        return this.isInvalidPreimageRootReason(height, data.missing_validators, initial_missing_validators);
     }
 
     /***************************************************************************
@@ -1228,6 +1234,8 @@ public class Ledger
             height = the height of proposed block
             missing_validators = list of indices to the validator UTXO set
                 which have not revealed the preimage
+            missing_validators_higher_bound = missing validators at the beginning of
+               the nomination round
 
         Returns:
             `null` if the information is valid at the proposed height,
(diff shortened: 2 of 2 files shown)


source/agora/node/Ledger.d
```diff
@@ -2178,7 +2194,45 @@ unittest
...
+    auto no_fee_txs = blocks[$-1].spendable.map!(txb => txb.sign()).array();
+    no_fee_txs.each!(tx => assert(ledger.acceptTransaction(tx)));
+
+    ConsensusData data;
+    ledger.prepareNominatingSet(data, Block.TxsInTestBlock, mock_clock.networkTime());
+
+    assert(ledger.validateConsensusData(data, skip_indexes) is null);
+
+    data.missing_validators = [2,5,7];
+    assert(ledger.validateConsensusData(data, [2,5,7,9]) is null);
```
hewison-chris:
```suggestion
    assert(ledger.validateConsensusData(data, [2,3,5,7,9]) is null);
```

hewison-chris: Not sure this will pass but I think it should.

hewison-chris (commented):
I am not really understanding the lower and upper bounds.

ferencdg:
> I am not really understanding the lower and upper bounds.

The idea here is that, the previous way of implementing what is considered 'acceptable candidate regarding missing validators' was too strict. Instead of equality, NodeA will check the candidate from NodeB in the following way:

Current missing validators in NodeA(=X) ⊆ missing validators in the candidate from NodeB(=Y) ⊆ missing validators in NodeA before the nomination round started(=Z)

If both of those conditions true, then NodeA will accept the candidate.

As the nodes exchanges preimages througout nomination(implemented in a separate ticket), the X and Y will both converge to [] or at least a stable set of missing validators, so:
1. the nomination round will end and not last forever
2. nodes don't change their votes from 'Yes' to 'No' ('No' to 'Yes' still happens, but that is fine in SCP as far as I know and we already rely on that fact with missing transactions)

source/agora/node/Ledger.d
ferencdg: here we are testing: 
[2,5] ⊆ [2,5,7] ⊆ [2,5,7,9] , so of course your suggestion would also pass

[2,5] ⊆ [2,5,7] ⊆ [2,3,5,7,9], but I think no need to change the code

source/agora/consensus/protocol/Nominator.d
```diff
@@ -409,14 +412,17 @@ extern(D):
             return;
         }
 
-        if (this.nomination_start_time == 0)
+        if (!this.is_nominating)
```
linked0, line 415: I think this is an unrelated change.

source/agora/node/Ledger.d
linked0: I think the two calls of `validateConsensusData` including the above one are called on different nodes. So it should be commented, which is kind of confusing.

source/agora/node/Ledger.d
```diff
@@ -1236,31 +1244,39 @@ public class Ledger
...
+        auto sorted_missing_validators = missing_validators.dup().sort();
+        auto sorted_missing_validators_lower_bound = missing_validators_lower_bound.dup().sort();
+        auto sorted_missing_validators_higher_bound = missing_validators_higher_bound.dup().sort();
 
-        if (local_missing_validators != missing_validators)
-            return "The list of missing validators does not match with the local one. " ~
-                assumeWontThrow(to!string(missing_validators)) ~
-                " != " ~ assumeWontThrow(to!string(local_missing_validators));
+        if (!startsWith(sorted_missing_validators, sorted_missing_validators_lower_bound))
+            return "Missing validator mismatch " ~
```
linked0: In the message, you should add additional text like "lower bound" and "current missing validators".

source/agora/node/Ledger.d
```diff
@@ -1236,31 +1244,39 @@ public class Ledger
...
+                assumeWontThrow(to!string(sorted_missing_validators_lower_bound)) ~
+                " is not a subset of " ~ assumeWontThrow(to!string(sorted_missing_validators));
+
+        if (!startsWith(sorted_missing_validators_higher_bound, sorted_missing_validators))
+            return "Missing validator mismatch " ~
```
linked0: In the message, you should add additional text like "higher bound" and "current missing validators".

linked0 (commented):
Good work, some comments.

source/agora/consensus/protocol/Nominator.d
ferencdg: > I think this is an unrelated change.

it is related, I had to start using that boolean to decide if the nomination round has started or not to be consistent with the code a couple of lines below

linked0: > I think this is an unrelated change.

Sorry for the wrong impression. I meant it's not directly related to checking the lists of missing validators. So it should be on another commit. Doesn't it succeed with the original line?

ferencdg: this line was added because of the subset change:
https://github.com/bosagora/agora/pull/2235/files#diff-2c5fcbf7251215d5e7aba3eaa09c23da9d88c3645e5c278b9ef351b9798184f7R423

and the line you commented was changed to be consistent with the line I referenced above

source/agora/node/Ledger.d
hewison-chris: > here we are testing:
> [2,5] ⊆ [2,5,7] ⊆ [2,5,7,9] , so of course your suggestion would also pass
> 
> [2,5] ⊆ [2,5,7] ⊆ [2,3,5,7,9], but I think no need to change the code

Are you sure? The reason why I suggested the change is I suspected it would fail with the current implementation.

hewison-chris: I just checked locally and the test fails with the change. The reason being that the `startsWith` is not same as sub set check.

ferencdg: sorry, I misunderstood, I thought you have a problem with the test. Yes the implementation of the subset had a bug, thanks.

source/agora/node/Ledger.d
```diff
@@ -1236,31 +1244,41 @@ public class Ledger
     ***************************************************************************/
 
     private string isInvalidPreimageRootReason (in Height height,
-        in uint[] missing_validators) @safe nothrow
+        in uint[] missing_validators, in uint[] missing_validators_higher_bound) @safe nothrow
```
hewison-chris, line 1247: IMO it is less confusing if you name this `initial_missing_validators`.

source/agora/node/Ledger.d
```diff
@@ -1236,31 +1244,41 @@ public class Ledger
...
     {
+        import std.algorithm.setops : setDifference;
+
         Hash[] keys;
         if (!this.enroll_man.getEnrolledUTXOs(height, keys) || keys.length == 0)
             assert(0, "Could not retrieve enrollments / no enrollments found");
 
-        uint[] local_missing_validators;
+        uint[] missing_validators_lower_bound;
```
hewison-chris, line 1255: Also here keep the original name.

hewison-chris (changes requested):
LGTM but I think `local_missing` must be subset of `received_missing` must be subset of `initial_missing` should be more clearly documented. The higher and lower bounds just confused me.

source/agora/node/Ledger.d
```diff
@@ -2178,7 +2196,45 @@ unittest
...
+    assert(ledger.validateConsensusData(data, skip_indexes) is null);
+
+    data.missing_validators = [2,5,7];
+    assert(ledger.validateConsensusData(data, [2,3,5,7,9]) is null);
+
+    data.missing_validators = [2,5];
+    assert(ledger.validateConsensusData(data, [2]) == "Higher bound violation - Missing validator mismatch [2, 5] is not a subset of [2]");
+
+    data.missing_validators = [5];
+    assert(ledger.validateConsensusData(data, [2]) == "Lower bound violation - Missing validator mismatch [2, 5] is not a subset of [5]");
```
linked0, line 2229: There are four `validateConsensusData` calls so it would be good to explain the concept of each call.

linked0 (commented):
On the contrary to @hewison-chris'  opinion, the concept of `higher` or `lower` bounds looks good to me because it adds some more meaning to the variables if we know that `initial missing validators` is used for `higher bound` and `current missing validators` for `lower bound`. Let's hear others' opinions.

ferencdg:
ready for merge

added docs as Chris suggested, but kept the variable names - with the docs it is easier to undersand without changing the variable names (at least I think)

source/agora/node/Ledger.d
```diff
@@ -1236,31 +1244,54 @@ public class Ledger
...
         }
 
-        if (local_missing_validators != missing_validators)
-            return "The list of missing validators does not match with the local one. " ~
-                assumeWontThrow(to!string(missing_validators)) ~
-                " != " ~ assumeWontThrow(to!string(local_missing_validators));
+        // The idea here is that, the previous way of implementing what is
+        // considered 'acceptable candidate regarding missing validators'
+        // was too strict. Instead of equality, NodeA will check the candidate
+        // from NodeB in the following way:
```
Geod24: Context fades, and any documentation relying on too much context will eventually become problematic.
Which is why documenting a functionality based on a (assumedly known) previous state is not future-proof.
If you want to mention the previous state, do so explicitly:
> A previous implementation that was simply checking that the local node's missing validators
> was equal to the one received was inadequate because [...]

Or simply document the state of things as it is now:
> NodeA will check the candidate [...]

I do think it's good to mention why a simple equality check is not a good idea, but the rationale should be provided.

Geod24 (approved):
Nice, just one issue w.r.t. doc

source/agora/node/Ledger.d
ferencdg: fixed the doc

PR #2237, opened 2021-07-01
Add proxy configuration

omerfirmak:
This will be used with https://gitlab.com/gitlab-org/security-products/har-recorder for fuzzing in the CI

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -49,6 +49,7 @@ import agora.utils.Utility;
 
 import vibe.http.common;
 import vibe.web.rest;
+import vibe.inet.url;
 
 import std.algorithm;
 import std.array;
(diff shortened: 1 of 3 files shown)


source/agora/network/Manager.d
```diff
@@ -1119,6 +1123,8 @@ public class NetworkManager
         settings.httpClientSettings = new HTTPClientSettings;
         settings.httpClientSettings.connectTimeout = timeout;
         settings.httpClientSettings.readTimeout = timeout;
+        if (this.proxy_url != string.init)
```
```suggestion
        if (this.proxy_url.length)
```

Little piece of D knowledge: `if (myString)` is `true` for `""`, because it's equivalent to `if (myString.ptr !is null)`. Since `string.init` is `null` (and not `""`), if we ended up with an empty string here, the check would pass.

And now that I wrote this, I realize that `URL.parse` should probably be called in the config parsing, so we don't even deal with strings here...

omerfirmak: > And now that I wrote this, I realize that URL.parse should probably be called in the config parsing, so we don't even deal with strings here...

Yeah makes sense, will update

omerfirmak: Updated

Geod24 (approved):

PR #2242, opened 2021-07-05
Change `NodeID` as a `Hash` of `UTXO`

### LINKED ISSUE #1309: Change registries (network, test) to use the UTXO instead of the public key

linked0:
We use the Hash of `UTXO` for `NodeID` instead of `PublicKey` because a user could run multiple `Validator`s so we should provide a way to identify each validator although all the validators come from the same `PublicKey`.

Fixes #1309 

There are two checkpoints.
1. If we should create a new `SCP` object when a new enrollment is used, is it needed to delete the old SCP?
2. Some `QuorumConfig`s is not shuffled in the unit test of `agora.test.QuorumPreimage`, which is needed to check.

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -956,7 +956,7 @@ public SCPQuorumSet toSCPQuorumSet (in QuorumConfig quorum_conf) @safe nothrow
 
     foreach (ref const node; quorum_conf.nodes)
     {
-        auto pub_key = NodeID(node.data[][0 .. NodeID.sizeof]);
+        auto pub_key = NodeID(node[][0 .. NodeID.sizeof]);
         quorum.validators.push_back(pub_key);
     }
 
@@ -985,12 +985,12 @@ public QuorumConfig toQuorumConfig (const ref SCPQuorumSet scp_quorum)
     @safe nothrow
 {
     import std.conv;
-    import scpd.types.Stellar_types : Hash, NodeID;
+    import scpd.types.Stellar_types : NodeID;
 
-    PublicKey[] nodes;
+    Hash[] nodes;
 
     foreach (node; scp_quorum.validators.constIterator)
-        nodes ~= PublicKey(node[]);
+        nodes ~= node;
 
     QuorumConfig[] quorums;
     foreach (ref sub_quorum; scp_quorum.innerSets.constIterator)
@@ -1009,16 +1009,18 @@ public QuorumConfig toQuorumConfig (const ref SCPQuorumSet scp_quorum)
 ///
 unittest
 {
+    import agora.crypto.Hash;
+
     auto quorum = QuorumConfig(2,
-        [PublicKey.fromString("boa1xp9rtxssrry6zqc4yt40h5h3al6enaywnvxfsp0ng8jt0ywyecsszs3fs7r"),
-         PublicKey.fromString("boa1xpc2ugmlve2ttuq6sjl4fznrggf2l5hksk2h2nsakexn690zxg4gss4p0w2")],
+        [Hash("0x11c6b0395c8e1716978c41958eab84e869755c09f7131b3bbdc882a647cb3f2c46c450607c6da71d34d1eab28fbfdf14376b444ef46ed1d0a7d2237ab430ebf5"),
+         Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd")],
         [QuorumConfig(2,
-            [PublicKey.fromString("boa1xp9rtxssrry6zqc4yt40h5h3al6enaywnvxfsp0ng8jt0ywyecsszs3fs7r"),
-             PublicKey.fromString("boa1xpc2ugmlve2ttuq6sjl4fznrggf2l5hksk2h2nsakexn690zxg4gss4p0w2")],
+            [Hash("0x11c6b0395c8e1716978c41958eab84e869755c09f7131b3bbdc882a647cb3f2c46c450607c6da71d34d1eab28fbfdf14376b444ef46ed1d0a7d2237ab430ebf5"),
+             Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd")],
             [QuorumConfig(2,
-                [PublicKey.fromString("boa1xp9rtxssrry6zqc4yt40h5h3al6enaywnvxfsp0ng8jt0ywyecsszs3fs7r"),
-                 PublicKey.fromString("boa1xpc2ugmlve2ttuq6sjl4fznrggf2l5hksk2h2nsakexn690zxg4gss4p0w2"),
-                 PublicKey.fromString("boa1xpc2ugmlve2ttuq6sjl4fznrggf2l5hksk2h2nsakexn690zxg4gss4p0w2")])])]);
+                [Hash("0x11c6b0395c8e1716978c41958eab84e869755c09f7131b3bbdc882a647cb3f2c46c450607c6da71d34d1eab28fbfdf14376b444ef46ed1d0a7d2237ab430ebf5"),
+                 Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd"),
+                 Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd")])])]);
 
     auto scp_quorum = toSCPQuorumSet(quorum);
     assert(scp_quorum.toQuorumConfig() == quorum);
--- a/source/agora/consensus/Quorum.d
+++ b/source/agora/consensus/Quorum.d
@@ -74,9 +74,8 @@ public struct QuorumParams
 
 *******************************************************************************/
 
-public QuorumConfig buildQuorumConfig (in PublicKey key,
-    in Hash[] utxo_keys, scope UTXOFinder finder, in Hash rand_seed,
-    const ref QuorumParams params)
+public QuorumConfig buildQuorumConfig (in Hash key, in Hash[] utxo_keys,
+    scope UTXOFinder finder, in Hash rand_seed, const ref QuorumParams params)
     @safe nothrow
 {
     // special-case: only 1 validator is active
@@ -109,7 +108,7 @@ public QuorumConfig buildQuorumConfig (in PublicKey key,
         if (added[idx])  // skip duplicate
             continue;
 
-        quorum.nodes ~= stakes[idx].key;
+        quorum.nodes ~= stakes[idx].utxo_key;
         added[idx] = true;  // mark used
     }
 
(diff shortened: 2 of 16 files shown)


source/agora/common/Config.d
```diff
@@ -985,12 +985,12 @@ public QuorumConfig toQuorumConfig (const ref SCPQuorumSet scp_quorum)
...
     import std.conv;
-    import scpd.types.Stellar_types : Hash, NodeID;
+    import scpd.types.Stellar_types : NodeID;
 
-    PublicKey[] nodes;
+    Hash[] nodes;
 
     foreach (node; scp_quorum.validators.constIterator)
-        nodes ~= PublicKey(node[]);
+        nodes ~= Hash(node[][0 .. Hash.sizeof]);
```
hewison-chris:
```suggestion
        nodes ~= node;
```

hewison-chris: I think node is already a hash.

source/agora/consensus/Quorum.d
```diff
@@ -274,27 +289,29 @@ unittest
     auto keys = getKeys(64);
     auto amounts = Amount(400000000000uL).repeat(64).array;
     amounts[0] = amounts[$ - 1] = Amount(400000000000uL * 62);
+    PublicKey[Hash] utxo_to_key;
     auto quorums_1 = buildTestQuorums(amounts, keys, hashFull(1),
-        QuorumParams.init, 14);
+        QuorumParams.init, 14, utxo_to_key);
     verifyQuorumsSanity(quorums_1);
     // verified to work but disabled because it runs slow with a
```
hewison-chris, line 294: Did you verify this again?

source/agora/consensus/Quorum.d
```diff
@@ -66,32 +66,33 @@ public struct QuorumParams
...
+        utxo_key = the key of the node for which to generate the quorum
         utxo_keys = the list of UTXO keys of all the active enrollments
         finder = UTXO finder delegate
         rand_seed = the random seed
         params = quorum generator algorithm tweaking parameters
 
 *******************************************************************************/
 
-public QuorumConfig buildQuorumConfig (in PublicKey key,
+public QuorumConfig buildQuorumConfig (in PublicKey key, in Hash utxo_key,
```
hewison-chris: Why do we still need the public key here?

source/scpp/src/xdr/Agora-types.h
```diff
@@ -19,6 +19,6 @@ namespace stellar
 
     // Logical types used by SCP / Agora
     using Hash =      uint512;
-    using NodeID =    uint256; // Currently a `PublicKey`
+    using NodeID =    uint512; // Previously a `PublicKey`
```
hewison-chris:
```suggestion
    using NodeID =    uint512; // Currently hash of `utxo` staked for validator
```

hewison-chris (changes requested):
Nice, I could be wrong of course, but I think we do not need to use public key in quorums at all.

source/agora/common/Config.d
linked0: Updated. Ty.

source/scpp/src/xdr/Agora-types.h
linked0: Updated.

source/agora/consensus/Quorum.d
linked0: I think I should remove the public key. Will update.

source/agora/consensus/Quorum.d
linked0: I tested for `32 nodes` and `64 nodes`. For `64 nodes`, verifying it takes about 1 minute in my local machine.

source/agora/consensus/Quorum.d
linked0: Updated

source/agora/consensus/protocol/Nominator.d
```diff
@@ -306,7 +339,10 @@ extern(D):
     public void stopNominationRound (Height height) @safe nothrow
     {
         this.is_nominating = false;
-        () @trusted { this.scp.stopNomination(height); }();
+        () @trusted {
+            if (this.scp != null)
+                this.scp.stopNomination(height);
```
Geod24, line 343: Can this ever be called if the node doesn't have an SCP object ?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -213,6 +213,39 @@ extern(D):
...
+
+    ***************************************************************************/
+
+    public void recreateSCPObject () nothrow @safe
+    {
+        try
+        {
+            auto self_enroll = this.enroll_man.getEnrollmentKey();
+            if (self_enroll != this.scp_utxo)
+            {
```
```suggestion
            // Either there's already an SCP object, or we're not enrolled
            if (self_enroll == this.scp_utxo)
                return;
```

In general, it's easier to get special cases "out of the way" and use early return if you can.
Reducing the level of indentation tends to reduce cognitive load.
And I think the special case of `Hash.init` should be documented... Which leads to another question: Can this method be called in a situation where `getEnrollmentKey` would return `Hash.init` where it was previously returning a non `init` value ? Because then we would (wrongly) create a new SCP object, correct ?

source/agora/common/Config.d
```diff
@@ -1009,21 +1009,22 @@ public QuorumConfig toQuorumConfig (const ref SCPQuorumSet scp_quorum)
...
-                 PublicKey.fromString("boa1xpc2ugmlve2ttuq6sjl4fznrggf2l5hksk2h2nsakexn690zxg4gss4p0w2"),
-                 PublicKey.fromString("boa1xpc2ugmlve2ttuq6sjl4fznrggf2l5hksk2h2nsakexn690zxg4gss4p0w2")])])]);
+                [Hash("0x11c6b0395c8e1716978c41958eab84e869755c09f7131b3bbdc882a647cb3f2c46c450607c6da71d34d1eab28fbfdf14376b444ef46ed1d0a7d2237ab430ebf5"),
+                 Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd"),
+                 Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd")])])]);
 
     auto scp_quorum = toSCPQuorumSet(quorum);
     assert(scp_quorum.toQuorumConfig() == quorum);
 }
-
```
Geod24: Revert

source/agora/consensus/protocol/Nominator.d
```diff
@@ -577,7 +578,14 @@ extern(D):
             return;  // slot was already externalized or envelope is too new
         }
 
-        const PublicKey public_key = PublicKey(envelope.statement.nodeID[]);
+        const Hash utxo = Hash(envelope.statement.nodeID[][0 .. Hash.sizeof]);
```
Geod24: The (double) slicing is not necessary here.

For `NodeID` we need to do:
`auto node_id = NodeID(self_enroll[][0 .. NodeID.sizeof]);`
Because:
- `self_enroll` is a `Hash` (BitBlob);
- `NodeID` expects a `ubyte[64]`

We can get a `ubyte[]` out of a `BitBlob`, by slicing.
Then, you can get a static array out of a dynamic array if you slice it with compile-time-known values (such as `Hash.sizeof`). So `self_enroll[]` (`ubyte[]`) => `self_enroll[][0 .. 64]`, the type of this expression is *still* `ubyte[]` **but** the compiler knows it can be converted to a static array. It's a form of value-range propagation/analysis. You have the same with literals: `auto x = 42;`, the type of `x` is `int`. But if you do `ubyte x = 42;`, it works, despite `int i; ubyte x = i;` not working.

However, the constructor of `BitBlob` [accepts a slice](https://github.com/Geod24/bitblob/blob/cbf4b05c834fd8b8fb06457d287613dbe7f74b74/source/geod24/bitblob.d#L198), so there is no such requirement as with `NodeID`.

The easiest way to go about it would probably be to allow `BitBlob!X` => `ubyte[X]` conversion, of course, which would remove our need to have this weird slicing statements.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -577,7 +578,14 @@ extern(D):
...
+        UTXO utxo_value;
+        if (!this.ledger.peekUTXO(utxo, utxo_value))
+        {
+            log.trace("Missing UTXO: {} at height {}", utxo, last_block.header.height);
```
Geod24: I think the message should be more particular. It might just be that the validator is not enrolled!
```suggestion
            log.trace("Couldn't find UTXO {} at height {} to validate envelope's signature", utxo, last_block.header.height);
```

source/agora/test/QuorumPreimage.d
```diff
@@ -36,6 +36,15 @@ import std.range;
 import core.thread;
 import core.time;
 
+static immutable A = Hash("0x210f6551d648a4da654da116b100e941e434e4f232b8579439c2ef64b04819bd2782eb3524c7a29c38c347cdf26006bccac54a58a58f103ae7eb5b252eb53b64");
```
Geod24: `static` is redundant at module scope.

source/agora/test/QuorumPreimage.d
```diff
@@ -67,60 +76,43 @@ unittest
...
+        G : "G",
+        H : "H",
+    ];
+
+    string printConvQuorum(in QuorumConfig qc)
+    {
+        string log = "[ ";
+        qc.nodes.each!(node => log ~= node_to_name[node] ~ ", ");
+        log = log[0 .. $ - 2] ~ " ]";
+        return log;
```
Geod24: Something like:
```suggestion
        return format("%-(%s, %)", qc.nodes.map!(n => node_to_name[n])); 
```

Example:
```D
import std.stdio, std.algorithm, std.range;

void main ()
{
 	auto names = [ "A", "B", "C", "D" ];
    writefln("%-(%s, %)", iota(0, 4).map!(n => names[n]));
}
```

source/scpd/scp/SCPDriver.d
```diff
@@ -137,7 +137,7 @@ nothrow:
     {
         import agora.crypto.Key : PublicKey;
```
Geod24: The import should be removed

source/agora/consensus/Quorum.d
```diff
@@ -66,32 +66,31 @@ public struct QuorumParams
...
         finder = UTXO finder delegate
         rand_seed = the random seed
         params = quorum generator algorithm tweaking parameters
 
 *******************************************************************************/
 
-public QuorumConfig buildQuorumConfig (in PublicKey key,
-    in Hash[] utxo_keys, scope UTXOFinder finder, in Hash rand_seed,
-    const ref QuorumParams params)
+public QuorumConfig buildQuorumConfig (in Hash utxo_key, in Hash[] utxo_keys,
```
Geod24: Honestly I don't know if the rename is necessary. We don't consistently use the terms anywhere. `utxo_key` feels weird, as it repeats the type name, essentially (`Hash`).

source/agora/consensus/Quorum.d
```diff
@@ -479,13 +502,14 @@ private NodeStake[] buildStakesDescending (const ref PublicKey filter,
 *******************************************************************************/
 
 version (unittest)
-private QuorumConfig[PublicKey] buildTestQuorums (Range)(Range amounts,
+private QuorumConfig[Hash] buildTestQuorums (Range)(Range amounts,
     const(PublicKey)[] keys, const auto ref Hash rand_seed,
-    const auto ref QuorumParams params, int id)
+    const auto ref QuorumParams params, int id, ref PublicKey[Hash] utxo_to_key)
```
Geod24: Can't you just pass `const(Hash)[] keys` instead of `const(PublicKey)[]` and remove `utxo_to_key` ??

Geod24: Perhaps not, since we generates txs from within this function. But the AA by ref is a red flag.

Geod24 (commented):
Great job! A few comments.

Geod24:
> If we should create a new `SCP` object when a new enrollment is used, is it needed to delete the old SCP?

Ideally yes. It's allocated on the C++ side so it won't be automatically deleted AFAIK. **But** since it's a rare event, and controlled by the node, I think we're fine for the time being with not freeing it.

source/agora/consensus/protocol/Nominator.d
linked0: It's a rare case that new enrollment is used for creating an SCP object. So I will just check it is `Hash.init` or not by removing the new `scp_utxo`. I will add a comment that we should check the situation where new enrollment is used for SCP env.

source/agora/consensus/Quorum.d
linked0: Right. The original meaning for the `key` was `the key of the node for which to generate the quorum` so the name should remain as it was. Will revert.

source/agora/consensus/protocol/Nominator.d
linked0: It happens. I think it's because an `outsider validator` exits without having created its SCP.

linked0: the related error log.
```
Running build/agora-unittests
0   agora-unittests                     0x00000001056f0cbb _D4core7runtime9backtraceFNbPPviZi + 75
1   agora-unittests                     0x00000001056f0c4a _D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPSQCk3sys5posix6signal9siginfo_tPvZv + 58
2   libsystem_platform.dylib            0x00007fff6d40c5fd _sigtramp + 29
3   agora-unittests                     0x000000010505a316 _ZNKSt3__16__treeINS_12__value_typeIyNS_10shared_ptrIN7stellar4SlotEEEEENS_19__map_value_compareIyS6_NS_4lessIyEELb1EEENS_9allocatorIS6_EEE6__rootEv + 22
4   agora-unittests                     0x000000010505d278 _ZNSt3__16__treeINS_12__value_typeIyNS_10shared_ptrIN7stellar4SlotEEEEENS_19__map_value_compareIyS6_NS_4lessIyEELb1EEENS_9allocatorIS6_EEE4findIyEENS_15__tree_iteratorIS6_PNS_11__tree_nodeIS6_PvEElEERKT_ + 40
5   agora-unittests                     0x0000000105055d2d _ZNSt3__13mapIyNS_10shared_ptrIN7stellar4SlotEEENS_4lessIyEENS_9allocatorINS_4pairIKyS4_EEEEE4findERS9_ + 29
6   agora-unittests                     0x000000010505563d _ZN7stellar3SCP7getSlotEyb + 77
7   agora-unittests                     0x0000000105055943 _ZN7stellar3SCP14stopNominationEy + 35
8   agora-unittests                     0x00000001049ead87 _D5agora9consensus8protocol9NominatorQk19stopNominationRoundMFNbNfSQCn6common5Types6HeightZ9__lambda2MFNbNeZv + 23
9   agora-unittests                     0x00000001046f58c8 _D5agora9consensus8protocol9NominatorQk19stopNominationRoundMFNbNfSQCn6common5Types6HeightZv + 56
10  agora-unittests                     0x00000001046eec8b _D5agora4node9ValidatorQk15onAcceptedBlockMFNfIKSQBv9consensus4data5BlockQgbZv + 91
11  agora-unittests                     0x0000000104751347 _D5agora4node6LedgerQh11acceptBlockMFNfIKSQBo9consensus4data5BlockQgAyamZb + 551
12  agora-unittests                     0x00000001048f3b42 _D5agora4node8FullNodeQj9addBlocksMFNfAxSQBn9consensus4data5BlockQgAxSQCqQBdQw12PreImageInfoQoZSQDq6common5Types6Height + 226
13  agora-unittests                     0x00000001047c0a10 _D5agora7network7Manager14NetworkManager13getBlocksFromMFNbSQCg6common5Types6HeightMDFNfAxSQDl9consensus4data5BlockQgAxSQEoQBdQw12PreImageInfoQoZQDiZv + 912
14  agora-unittests                     0x00000001046ed2f3 _D5agora4node8FullNodeQj11catchupTaskMFNbZv + 147
15  agora-unittests                     0x0000000105200397 _D6geod249LocalRest5Timer3runMFZv + 55
16  agora-unittests                     0x00000001056f404d _D4core6thread7context8Callable6opCallMFZv + 61
17  agora-unittests                     0x00000001056f41d9 _D4core6thread5fiber5Fiber3runMFZv + 25
18  agora-unittests                     0x00000001056f4116 fiber_entryPoint + 150
```

source/agora/consensus/Quorum.d
linked0: Will apply `out` instead of `ref`.

source/agora/test/QuorumPreimage.d
linked0: Updated

source/agora/consensus/protocol/Nominator.d
linked0: Ty for the great explanation. :)

linked0: One more question. So is `const Hash utxo = envelope.statement.nodeID;` enough?

linked0:
All green!

source/agora/test/QuorumPreimage.d
```diff
@@ -67,60 +76,32 @@ unittest
         }
     }
 
+    string[Hash] node_to_name = [
+        A : "A", B : "B", C : "C", D : "D", E : "E", F : "F", G : "G", H : "H"];
```
hewison-chris, line 80: Very good.

hewison-chris (approved):
LGTM

Geod24 (approved):

PR #2243, opened 2021-07-05
 Swagger specification for Agora API

omerfirmak:
This can be consumed by various tools

You can render it on https://editor.swagger.io/

### DIFF
--- a/source/agora/api/spec.yaml
+++ b/source/agora/api/spec.yaml
@@ -1,0 +1,653 @@
+openapi: 3.0.0
+info:
+  title: Agora API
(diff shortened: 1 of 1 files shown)


Geod24:
I don't think the fuzzer should run as part of the CI. It should be run at worst manually, or on a cron schedule. Tests should be easily reproducible and the fuzzer is the opposite of that.

Geod24:
Any way we can keep it in sync ? :/

omerfirmak:
> Any way we can keep it in sync ? :/

Through sheer willpower

PR #2244, opened 2021-07-06
Making preimage propagation more reliable

### LINKED ISSUE #2217: making pre-image propagation more reliable
nomination can easily get stuck: there is no guarantee a preimages will reach ALL nodes (the same way there is no guarantee a transaction will reach each nodes). For transactions we have a fallback mechanism: if we see a consensus data with an unknown hash, we try to retrieve the transaction for that unknown hash. Currently there is no similar fallback mechanism for preimages. This means 33% of the nodes might might not have 1 missing preimages, an other 33% might have 2 missing preimages, while 34% have no missing preimages. Now the consnsus is stuck forever

### LINKED ISSUE #2246: CheckSelfSlashing is checked twice when validateValue is called(second check is in validateSlashingData)

ferencdg:
fixes: #2217
fixes: #2246

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -262,6 +262,26 @@ public interface API
 
     /***************************************************************************
 
+        Returns the preimages for the specified enroll keys.
+
+        Params:
+            enroll_keys = Set of enrollment keys. If the set of enroll_keys is
+            null or empty, then all preimages known to the node are returned.
+
+        Returns:
+            The preimages for the specified enroll keys. If the requested node
+            doesn't know the preimage for a specific enroll key, then it will
+            not be included in the result.
+
+        API:
+            GET /preimages_for_enroll_keys
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getPreimagesForEnrollKeys (Set!Hash enroll_keys = Set!Hash.init);
+
+    /***************************************************************************
+
         Get validators' pre-image information
 
         API:
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -243,25 +243,6 @@ public interface API
 
     /***************************************************************************
 
-        Get validator's pre-image information
-
-        API:
-            GET /preimage
-
-        Params:
-            enroll_key = The key for the enrollment in which the pre-image is
-                contained.
-
-        Returns:
-            preimage information of the validator if exists, otherwise
-                PreImageInfo.init
-
-    ***************************************************************************/
-
-    public PreImageInfo getPreimage (Hash enroll_key);
-
-    /***************************************************************************
-
         Returns the preimages for the specified enroll keys.
 
         Params:
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -922,13 +922,6 @@ extern(D):
             return ValidationLevel.kInvalidValue;
         }
 
-        if (this.ledger.isSelfSlashing(Height(slot_idx), data))
-        {
-            log.warn("validateValue(): Marking {} for data slashing us as invalid",
-                     nomination ? "nomination" : "vote");
-            return ValidationLevel.kInvalidValue;
-        }
-
         if (auto fail_reason = this.ledger.validateConsensusData(data, this.initial_missing_validators))
         {
             log.error("validateValue(): Validation failed: {}. Data: {}",
--- a/source/agora/network/Client.d
+++ b/source/agora/network/Client.d
@@ -526,6 +526,29 @@ public class NetworkClient
 
     /***************************************************************************
 
+        Returns the preimages for the specified enroll keys.
+
+        Params:
+            enroll_keys = Set of enrollment keys. If the set of enroll_keys is
+            null or empty, then all preimages known to the node are returned.
+
+        Returns:
+            The preimages for the specified enroll keys. If the requested node
+            doesn't know the preimage for a specific enroll key, then it will
+            not be included in the result.
+
+        API:
+            GET /preimages_for_enroll_keys
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getPreimagesForEnrollKeys (Set!Hash enroll_keys = Set!Hash.init) @trusted nothrow
+    {
+        return this.attemptRequest!(API.getPreimagesForEnrollKeys, Throw.No)(this.api, enroll_keys);
+    }
+
+    /***************************************************************************
+
         Attempt a request up to 'this.max_retries' attempts, and make the task
         wait this.retry_delay between each attempt.
 
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -983,6 +983,33 @@ public class NetworkManager
 
     /***************************************************************************
 
+        Retrieves the missing preimages that other nodes indirectly indicated
+        (through consensus data shared during nomination) they possess.
+
+        Params:
+            ledger = ledger instance
+
+    ***************************************************************************/
+
+    public void retrievePreimages (Ledger ledger) @safe nothrow
+    {
+        auto enroll_keys = ledger.getEnrollKeysForUnknownPreimages();
+
+        foreach (peer; this.peers[])
+        {
+            if (enroll_keys.empty())
+                break;
+
+            foreach (const ref preimage_info; peer.client.getPreimagesForEnrollKeys(enroll_keys))
+                if (ledger.enrollment_manager.addPreimage(preimage_info))
+                    enroll_keys.remove(preimage_info.utxo);
+        }
+
+        () @trusted {enroll_keys.clear();}();
+    }
+
+    /***************************************************************************
+
         Retrieve any missing block signatures from the connected nodes for
         blocks since the last fee payment block.
 
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -96,7 +96,7 @@ public class Ledger
     private EnrollmentManager enroll_man;
 
     /// Property for Enrollment manager
-    @property public EnrollmentManager enrollment_manager () @safe
+    @property public EnrollmentManager enrollment_manager () @safe nothrow
     {
         return this.enroll_man;
     }
@@ -114,6 +114,9 @@ public class Ledger
     /// The new block time_offset has to be greater than the previous block time_offset,
     /// but less than current time + block_time_offset_tolerance
     public Duration block_time_offset_tolerance;
+
+    /// Enrollment keys for unknown preimages for the next height
+    private Set!Hash enrolls_keys_for_unknown_preimages;
 
     /***************************************************************************
 
@@ -700,6 +703,7 @@ public class Ledger
         Check whether the slashing data is valid.
 
         Params:
+            height = height
             data = consensus data
             initial_missing_validators = missing validators at the beginning of
                the nomination round
@@ -712,14 +716,10 @@ public class Ledger
     public string validateSlashingData (in Height height, in ConsensusData data,
         in uint[] initial_missing_validators) @safe nothrow
     {
-        // If we are enrolled and not slashing ourselves
-        if (this.enroll_man.isEnrolled(height, &this.utxo_set.peekUTXO) && this.isSelfSlashing(height, data))
-        {
-            log.fatal("The node is slashing itself.");
-            assert(0);
-        }
+        if (auto res = this.isInvalidPreimageRootReason(height, data.missing_validators, initial_missing_validators))
+            return res;
 
-        return this.isInvalidPreimageRootReason(height, data.missing_validators, initial_missing_validators);
+        return this.isSelfSlashing(height, data);
     }
 
     /***************************************************************************
@@ -728,18 +728,19 @@ public class Ledger
         a node itself
 
         Params:
+            height = the height for which we want to check self slashing
             data = consensus data
 
         Returns:
-            true if the consensus data has the information that is slashing
-            a node itself.
+            error string, if node is trying to slash itself, null otherwise
 
     ***************************************************************************/
 
-    public bool isSelfSlashing (in Height height, in ConsensusData data) @safe nothrow
+    public string isSelfSlashing (in Height height, in ConsensusData data) @safe nothrow
     {
-        const index = this.enroll_man.getIndexOfEnrollment(this.last_block.header.height + 1);
-        return (index != ulong.max && !data.missing_validators.find(index).empty);
+        const index = this.enroll_man.getIndexOfEnrollment(height);
+        return (index != ulong.max && !data.missing_validators.find(index).empty) ?
+            "Node is attempting to slash itself" : null;
     }
 
     /***************************************************************************
@@ -1175,6 +1176,19 @@ public class Ledger
 
     /***************************************************************************
 
+        Get the enrollment keys for unknown preimages
+
+        Returns:
+            the enrollment keys for unknown preimages
+
+    ***************************************************************************/
+
+    public Set!Hash getEnrollKeysForUnknownPreimages () @safe nothrow
+    {
+        return this.enrolls_keys_for_unknown_preimages;
+    }
+    /***************************************************************************
+
         Get the UTXOs of the validators that do not reveal their pre-images
         by indices
 
@@ -1259,6 +1273,11 @@ public class Ledger
             if (!this.hasRevealedPreimage(height, key))
                 missing_validators_lower_bound ~= cast(uint)idx;
         }
+
+        // trying to retrieve preimages that - based on the consensus data -
+        // was shared with other nodes
+        foreach (const validator_ind; setDifference(missing_validators_lower_bound, missing_validators))
+            this.enrolls_keys_for_unknown_preimages.put(keys[validator_ind]);
 
         // NodeA will check the candidate from NodeB in the following way:
         //
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -728,12 +728,14 @@ public class TestAPIManager
         string file = __FILE__, int line = __LINE__)
     {
         static assert (isInputRange!Idxs);
+        import std.algorithm.searching : any;
 
         clients_idxs.each!(idx =>
             enrolls.enumerate.each!((idx_enroll, enroll) {
                 if (clients_idxs.canFind(idx_enroll))
-                    retryFor(this.clients[idx].getPreimage(enroll.utxo_key).height >= height,
-                        timeout, format!"Client #%s has no preimage for client #%s at distance %s"
+                    retryFor(this.clients[idx].getPreimagesForEnrollKeys(Set!Hash.from(enroll.utxo_key.only))
+                        .any!(preimage => preimage.height >= height),
+                            timeout, format!"Client #%s has no preimage for client #%s at distance %s"
                             (idx, idx_enroll, height));
             }));
     }
@@ -2249,16 +2251,6 @@ public class NoPreImageVN : TestValidatorNode
     {
         if (atomicLoad(*this.reveal_preimage))
             super.onPreImageRevealTimer();
-    }
-
-    public override PreImageInfo getPreimage (Hash enroll_key)
-    {
-        if (enroll_key != this.enroll_man.getEnrollmentKey())
-            return super.getPreimage(enroll_key);
-        else if (atomicLoad(*this.reveal_preimage))
-            return super.getPreimage(enroll_key);
-        else
-            throw new Exception("No such PreImage");
     }
 
     /// GET: /preimages
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -1942,6 +1942,10 @@ public struct TestConf
         testing: true,
     };
 
+    /// How often the validator should try to catchup for the preimages for the
+    /// next block
+    public Duration preimage_catchup_interval = 100.seconds;
+
     /// max failed requests before a node is banned
     size_t max_failed_requests = 100;
 
@@ -2064,6 +2068,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
             recurring_enrollment : test_conf.recurring_enrollment,
             preimage_reveal_interval : 1.seconds,  // check revealing frequently
             nomination_interval: 100.msecs,
+            preimage_catchup_interval: test_conf.preimage_catchup_interval,
         };
 
         Config conf =
@@ -2153,6 +2158,24 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
     return net;
 }
 
+/// API manager that creates nodes that are too lazy to reveal preimages
+/// (except when explicitely set to reveal)
+class LazyAPIManager(PreImageVN = NoPreImageVN) : TestAPIManager
+{
+    public shared bool reveal_preimage = false;
+
+    ///
+    mixin ForwardCtor!();
+
+    ///
+    public override void createNewNode (Config conf, string file, int line)
+    {
+        if (conf.validator.enabled == true)
+            this.addNewNode!PreImageVN(conf, &this.reveal_preimage, file, line);
+        else
+            super.createNewNode(conf, file, line);
+    }
+}
 
 /// Returns: the entire ledger from the provided node
 public const(Block)[] getAllBlocks (TestAPI node)
@@ -2247,5 +2270,14 @@ public class NoPreImageVN : TestValidatorNode
         const ek = this.enroll_man.getEnrollmentKey();
         return super.getPreimages(start_height, end_height)
             .filter!(pi => pi.utxo != ek).array();
+    }
+
+    /// GET: /preimages_for_enroll_keys
+    public override PreImageInfo[] getPreimagesForEnrollKeys (Set!Hash enroll_keys = Set!Hash.init) @safe nothrow
+    {
+        if (!atomicLoad(*this.reveal_preimage))
+            enroll_keys.remove(this.enroll_man.getEnrollmentKey());
+
+        return super.getPreimagesForEnrollKeys(enroll_keys);
     }
 }
--- a/source/agora/test/PreimageSharing.d
+++ b/source/agora/test/PreimageSharing.d
@@ -1,0 +1,62 @@
+/*******************************************************************************
+
+    Contains tests for sharing preimages during SCP protocol run.
+
+    Copyright:
+        Copyright (c) 2019-2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.PreimageSharing;
+
+version (unittest):
+
+import agora.common.Config;
+import agora.common.Set;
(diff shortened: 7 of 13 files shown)


source/agora/api/Validator.d
```diff
@@ -120,4 +121,22 @@ public interface API : agora.api.FullNode.API
...
+        Returns:
+            The preimages for the specified enroll keys for the specified height.
+            If the requested node doesn't know the preimage for a specific
+            enroll key, then it will not be included in the result.
+
+        API:
+            GET /preimages_for_enroll_keys
+
+    ***************************************************************************/
+
```
ferencdg: we could be more efficient here by only sharing the indices, but this will unlikely to cause any real problems

source/agora/consensus/protocol/Nominator.d
```diff
@@ -873,13 +873,6 @@ extern(D):
             return ValidationLevel.kInvalidValue;
         }
```
ferencdg, line 924: there were 2 problems with this code:
1. we check for the same thing in the ValidateConsensusData too
2. making the check here would have caused us not to record the missing validators, so I couldn't have implemented my ticket

Because of 2., it is not in a separate commit

source/agora/test/SlashingMisbehavingValidator.d
```diff
@@ -115,23 +115,6 @@ unittest
 ///     some failures.
 unittest
 {
-    static class LazyAPIManager : TestAPIManager
```
ferencdg, line 118: needed to reuse this in my ticket, so that is why there is no separate commit

source/agora/node/Ledger.d
```diff
@@ -722,18 +722,19 @@ public class Ledger
...
 
         Returns:
-            true if the consensus data has the information that is slashing
-            a node itself.
+            error string, if node is trying to slash itself, null otherwise
 
     ***************************************************************************/
 
-    public bool isSelfSlashing (in Height height, in ConsensusData data) @safe nothrow
+    public string isSelfSlashing (in Height height, in ConsensusData data) @safe nothrow
```
ferencdg, line 739: see https://github.com/bosagora/agora/pull/2244/files#r664243457 for why it is not in a separate commit

source/agora/node/Ledger.d
```diff
@@ -1166,6 +1166,10 @@ public class Ledger
         return this.unknown_txs;
     }
 
+    public Set!Hash getEnrollKeysForUnknownPreimages () @safe nothrow
+    {
+        return this.enrolls_keys_for_unknown_preimages;
+    }
```
hewison-chris, line 1189: Please update the commit message to match the actual changes.

source/agora/network/Client.d
```diff
@@ -524,6 +524,27 @@ public class NetworkClient
             start_height, end_height);
     }
 
+    /***************************************************************************
+
+        Returns he preimages for the specified enroll keys for the specified height.
```
hewison-chris:
```suggestion
        Returns the preimages for the specified enroll keys and height.
```

source/agora/network/Manager.d
```diff
@@ -976,6 +976,36 @@ public class NetworkManager
...
     }
 
+    /***************************************************************************
+
+        Retrieves the missing preimages that other nodes indirectly indicated
+        (through consensus data shared during nomination) they possess.
+
+        Params:
+            ledger = ledger instance
+            height = height
```
hewison-chris: `height` is not a param.

source/agora/network/Manager.d
```diff
@@ -976,6 +976,36 @@ public class NetworkManager
...
+
+    ***************************************************************************/
+
+    public void retrievePreimages (Ledger ledger) @safe nothrow
```
hewison-chris, line 994:
```suggestion
    public void retrievePreimages (in Ledger ledger) @safe nothrow
```

source/agora/network/Manager.d
```diff
@@ -976,6 +976,36 @@ public class NetworkManager
...
+    {
+
```
hewison-chris:
```suggestion
```

source/agora/network/Manager.d
```diff
@@ -976,6 +976,36 @@ public class NetworkManager
...
+        auto enroll_keys = ledger.getEnrollKeysForUnknownPreimages();
+
+        foreach (peer; this.peers[])
+        {
+            if (enroll_keys.empty())
```
hewison-chris, line 1000: Why don't we check this before the `foreach`?

hewison-chris (changes requested):
LGTM, just a few small changes requested.

linked0 (commented):
Overall LGTM but first commit has a member variable that is introduced in the next commit, which is `enrolls_keys_for_unknown_preimages`.

source/agora/node/Ledger.d
ferencdg: I staged the wrong chunk there :( fixed by now

source/agora/network/Manager.d
ferencdg: enroll_keys is also changed inside the foor loop, so we need to check it there too

I could check it in front of the loop as well, but entering the foreach have almost no performance impact and for me, the code looks cleaner this way.

source/agora/network/Manager.d
ferencdg: sadly 'in' makes the ledger constant, but I need to call mutable variables methods like: 
```D
ledger.enrollment_manager.addPreimage(preimage_info)
```

ferencdg:
fixed everything else that was suggested

source/agora/node/Ledger.d
hewison-chris: But shouldn't the new function be in the next commit? In fact I don't really see why this adding `nothrow` is not done in the commit that requires it.

ferencdg: right, squashed it

source/agora/network/Manager.d
hewison-chris: This extra line is still there.

hewison-chris (approved):
LGTM

linked0 (approved):

source/agora/node/Ledger.d
```diff
@@ -722,18 +722,19 @@ public class Ledger
         a node itself
 
         Params:
+            height = the heigh for which we want to check self slashing
```
kchulj:
```suggestion
            height = the height for which we want to check self slashing
```

dub.json
```diff
@@ -36,7 +36,6 @@
 	"gdc": "no",
 	"ldc": "~>1.26"
     },
-
```
Geod24: Unrelated

source/agora/api/Validator.d
```diff
@@ -120,4 +121,22 @@ public interface API : agora.api.FullNode.API
...
+    public PreImageInfo[] getPreimagesForEnrollKeys (Set!Hash enroll_keys, ulong height);
```
Geod24: Hum, how does a set get serialized ? I'd expect a `Hash[]` here.

Geod24: Also, `FullNode` already has `getPreimages`. This is equivalent to `getPreimages(height, height)`, isn't it ?

source/agora/test/PreimageSharing.d
```diff
@@ -0,0 +1,63 @@
...
+    Contains tests for sharing preimages during SCP protocol run.
+
+    Copyright:
+        Copyright (c) 2019-2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
```
Geod24:
```suggestion
/*******************************************************************************

    Contains tests for sharing preimages during SCP protocol run.

    Copyright:
        Copyright (c) 2019-2021 BOSAGORA Foundation
        All rights reserved.

    License:
        MIT License. See LICENSE for details.

*******************************************************************************/

module agora.test.PreimageSharing;
```

Geod24 (changes requested):
The API endpoint is duplicated. The rest looks good, will do a second pass once duplication is removed.

source/agora/api/Validator.d
ferencdg: it is not exactly equivalent, because we don't expect many missing preimages, only a few, but probably none at all. getPreimages(height, height) returns every preimage

if and attacker would claim that he knows the missing preimage for NodeA (where in fact NodeA's preimage is missing) getPreimages(height, height) would return all preimages from all nodes

Geod24: But is it worth the extra endpoint ?

ferencdg: > Hum, how does a set get serialized ? I'd expect a `Hash[]` here.

it is a hash array I think
```D
    public string toString () const
    {
        import std.format;

        return format("%(%s, %)",
                      this._set.byKeyValue.filter!(tup => tup.value)
                      .map!(tup => tup.key));
    }
```

ferencdg: > But is it worth the extra endpoint ?

adding an extra endpoint is not ideal (as it is an extra burden on a new client which is not implemented in DLang), but I think it is worth it in this case

if we really want to remove a useless endpoint then I think it should be 

public PreImageInfo getPreimage (Hash enroll_key);

1. this endpoint requires clients to query for each enroll keys
2. height is assumed to be the highest which is not reliable

ferencdg:
rebase #1...

ferencdg:
@Geod24 fixed everything, I think instead of not adding the new getPreimagesForEnrollKeys() method, we should remove the dumber version of it called getPreimage(), which I could do in a separate PR

Geod24:
Well nothing prevents a node from returning more preimages than requested, so the API is only useful for the server, if the client is well behaving.
Having 2 (or 3) endpoints for the same thing is indeed quite ugly. What we currently have:
- `getPreimage` => Returns a single UTXO's preimage, at a given height;
- `getPreimages(from, to)` => Return all preimages for validator between `from` and `to`;
- `getPreimagesForEnrollKeys` => Return all peimages for a certain set of key;

Now there's a few things to take into account:
- A node only needs to return a single `PreImageInfo` for a given UTXO;
- That used not to be the case, which motivated `getPreimages(from, to)`;
- I *think* nodes should return the latest known pre-image regardless of the requested height, and the client should adjust;
- As a result, providing the height to any of those endpoints is *pointless* unless we don't know the validator set;
- If there isn't a case in which we don't know the validator set, we could have an endpoint `getPreimages(Hash[] utxos)` and call it a day;
- But that would require sending *all* validators to a node, which would lead to pretty large (and useless) queries;
- What we could do instead is to have `getPreimages(Hash[] only = null);` where it returns either all preimages (for the node's current validator set) or the `only` subset requested - assumed to be small;

Note that this *could* create issues when catching up, as catching up requires preimages. However, a trivial fix is to request any new validator immediately: Since they tend to be long lived, it will only leads to sporadic, single-argument queries.

Thoughts ?

ferencdg:
getPreimagesForEnrollKeys => Return all peimages for a certain set of key; - this is not true, it only returns for a specified height

ferencdg:
'Well nothing prevents a node from returning more preimages than requested, so the API is only useful for the server, if the client is well behaving.'

but the solution to return ALL preimages is not useful even for well-behaving clients. I think what you said is true for any endpoint, a non well-behaving node can return an endless stream of garbage as long as it can be deserialized.

Geod24:
> this is not true, it only returns for a specified height

Why would we not return the latest preimage ?

ferencdg:
> > this is not true, it only returns for a specified height
> 
> Why would we not return the latest preimage ?

this function is called during the short run of the nomination protocoll, we cannot just ask for the latest one as the latest one is ususally 5 cycles ahead

Geod24:
> this function is called during the short run of the nomination protocoll, we cannot just ask for the latest one as the latest one is ususally 5 cycles ahead

If you know the one 5 cycles ahead, you know the current one and you don't have to request it again.

ferencdg:
> > this is not true, it only returns for a specified height
> 
> Why would we not return the latest preimage ?

I also think that if the requesting node has to depend on the current context of the REQUESTED node (what is the maximum known preimage received by the REQUESTED node so far), then it is much harder to reason about code correctness in the requesting node. For example what if the requested node only knows the preimage for height 5, but the requested node has the preimage for height 7. Now the requested node will reject the returned preimage.

Geod24:
> I also think that if the requesting node has to depend on the current context of the REQUESTED node (what is the maximum known preimage received by the REQUESTED node so far), then it is much harder to reason about code correctness in the requesting node. For example what if the requested node only knows the preimage for height 5, but the requested node has the preimage for height 7. Now the requested node will reject the returned preimage.

Discussed IRL, and cleared up I hope.

ferencdg:
made the requested changes

source/agora/api/FullNode.d
```diff
@@ -260,6 +260,25 @@ public interface API
...
+            enroll key, then it will not be included in the result.
+            If the set of enroll_keys is empty, then all preimages known to the
+            node are returned.
+
+        API:
+            GET /preimages_for_enroll_keys
+
+    ***************************************************************************/
+
+    public PreImageInfo[] getPreimagesForEnrollKeys (Set!Hash enroll_keys);
```
Geod24: Set it to `null` by default, and rename to `getPreimages`

source/agora/api/FullNode.d
```diff
@@ -260,6 +260,25 @@ public interface API
...
+
+        Params:
+            enroll_keys = set of enrollment keys
+
+        Returns:
+            The preimages for the specified enroll keys.
+            If the requested node doesn't know the preimage for a specific
+            enroll key, then it will not be included in the result.
+            If the set of enroll_keys is empty, then all preimages known to the
+            node are returned.
```
Geod24: Should be part of the param documentation

source/agora/api/FullNode.d
```diff
@@ -260,6 +260,25 @@ public interface API
...
 
+    /***************************************************************************
+
+        Params:
+            enroll_keys = set of enrollment keys
+
+        Returns:
+            The preimages for the specified enroll keys.
+            If the requested node doesn't know the preimage for a specific
+            enroll key, then it will not be included in the result.
```
Geod24: This looks like some of this should be the title / description, not in the `Returns` section.

source/agora/api/Validator.d
```diff
@@ -47,6 +47,7 @@ import agora.consensus.data.ValidatorBlockSig;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.crypto.Schnorr;
+import agora.common.Set;
```
Geod24: Remove

source/agora/api/FullNode.d
Geod24: Also I believe we could just use `Hash[]` here instead of a `Set` ?

source/agora/node/Ledger.d
```diff
@@ -1260,6 +1275,11 @@ public class Ledger
                 missing_validators_lower_bound ~= cast(uint)idx;
         }
 
+        // trying to retrieve preimages that - based on the consensus data -
+        // was shared with other nodes
+        foreach (const validator_ind; setDifference(missing_validators_lower_bound, missing_validators))
+            this.enrolls_keys_for_unknown_preimages.put(keys[validator_ind]);
+
```
Geod24, line 1281: This method was not originally modifying any class member, which means it was idempotent.
However, with this addition, it no longer is, and there's an assumption that it will only be called once per (I'm not sure what).
Also, I couldn't find the place where this was reset.

source/agora/node/Ledger.d
```diff
@@ -1173,6 +1174,20 @@ public class Ledger
...
+        Get the enrollment keys for unknown preimages
+
+        Returns:
+            the enrollment keys for unknown preimages
+
+    ***************************************************************************/
+
+
+    public Set!Hash getEnrollKeysForUnknownPreimages () @safe nothrow
```
Geod24, line 1186: Instead of this member function, I think you can just call `getValidators` (since you have access to the Ledger) and check the PI height.

source/agora/test/PreimageSharing.d
```diff
@@ -0,0 +1,63 @@
...
+import agora.test.Base;
+
+import std.exception;
+import std.path : buildPath;
+
+import core.atomic;
+import core.stdc.stdint;
+import core.thread;
+
+import geod24.Registry;
```
Geod24: There's a few imports there that are superfluous

source/agora/test/PreimageSharing.d
```diff
@@ -0,0 +1,63 @@
...
+
+    // create and send txs to nodes
+    auto nodes = network.clients;
+    auto txs = network.blocks[$ - 1].spendable().map!(txb => txb.sign()).array;
+    txs.each!(tx => nodes[0].putTransaction(tx));
+
+    // block should be created even though the validators didn't reveal their
+    // preimages in advance, as preimage sharing is also done during SCP run
+    network.expectHeight(Height(1));
+
```
Geod24: Trailing line

source/agora/test/PreimageSharing.d
```diff
@@ -0,0 +1,63 @@
...
+unittest
+{
+    TestConf conf = {
+        quorum_threshold : 100,
+        preimage_catchup_interval : 1.seconds,
+    };
+
+    // set up nodes
+    auto network = makeTestNetwork!LazyAPIManager(conf);
```
Geod24: Looking at this test, it shouldn't work. `NoPreImageVN` was designed to *not* reveal pre-images, unless told to do so. That's why `getPreimage was overriden.

source/agora/test/Base.d
```diff
@@ -710,7 +710,14 @@ public class TestAPIManager
         clients_idxs.each!(idx =>
             enrolls.enumerate.each!((idx_enroll, enroll) {
                 if (clients_idxs.canFind(idx_enroll))
-                    retryFor(this.clients[idx].getPreimage(enroll.utxo_key).height >= height,
+                    retryFor(
+                        ()
+                        {
+                            Set!Hash enroll_keys;
+                            enroll_keys.put(enroll.utxo_key);
+                            auto preimages = this.clients[idx].getPreimagesForEnrollKeys(enroll_keys);
```
Geod24: You could just use `[enroll.utxo_key]` if it was a `Hash`.

source/agora/test/Base.d
```diff
@@ -1916,6 +1923,10 @@ public struct TestConf
     /// Transaction put into the relay queue will expire, and will be removed
     /// after `relay_tx_cache_exp`.
     Duration relay_tx_cache_exp = 60.minutes;
+
+    /// How often the validator should try to catchup for the preimages for the
+    /// next block
+    public Duration preimage_catchup_interval = 100.seconds;
```
Geod24: You inspired https://github.com/bosagora/agora/pull/2276 BTW :)

source/agora/api/FullNode.d
ferencdg: 1. there are a lot of convenient function on set like setDifference that I use in many places to make the code more clear
2. followed this: getTransactions (Set!Hash tx_hashes); which I think the correct approach

Geod24: `setDifference` works on range, so it will work on arrays. Also, it assumes the input is sorted, which I don't think is the case here ?

Geod24: Regarding point 2, I don't *mind* having `Set` in the API. But:
- [x] It needs to be convenient to use from other languages: Check, because it serializes just like an array;
- [ ] It needs to be convenient to call from our own code: This is the current problem, because calling this with a single value is 3-4 lines of code;
- [x] It needs not to incur more penalty (speed, memory usage, etc...) than an array: I believe this is more-or-less the case right now (not exactly, because one can't manage the underlying memory of an AA, but it's a price we decided to pay), but it's something to keep in mind, especially if you solve the second point;

ferencdg: 'rename to getPreimages' - isn't that a bit ambiguous for a REST interface as we already have 

public override PreImageInfo[] getPreimages (ulong start_height,
        ulong end_height)
        
can Vibe.d even handle this (2 REST interfaces with the same endpoint name)?

source/agora/node/Ledger.d
ferencdg: reset is in NetworkManager.retrievePreimages()
```D
        () @trusted {enroll_keys.clear();}();
```

ferencdg: 'However, with this addition, it no longer is, and there's an assumption that it will only be called once per' 

it is a Set!Hash - there is no assumption (one of the reason why I used Sets)

ferencdg: 'This method was not originally modifying any class member, which means it was idempotent.'

suggestion where it should be?

source/agora/api/FullNode.d
omerfirmak: Vibe.d serializes Sets as dictionaries btw.

source/agora/node/Ledger.d
ferencdg: in enrolls_keys_for_unknown_preimages we store what other validators reported regarding for what validators they have preimage for(and we don't)

I don't think our validator's ledger has that information.

source/agora/test/PreimageSharing.d
ferencdg: You can see in the comment why it works
```D
/// Please note that preimages can still be shared during nomination, if the
/// TestConf.preimage_catchup_interval is set to a low value.
public class NoPreImageVN : TestValidatorNode
```

ferencdg:
@Geod24 pls let me know if you disagree with the comments, so I can make the change

source/agora/test/PreimageSharing.d
Geod24: But then, the other test that uses this class, which relies on preimages *really* not being propagated has its assumptions gutted, doesn't it ?

source/agora/api/FullNode.d
ferencdg: >  Also, it assumes the input is sorted, which I don't think is the case here ?

'it assumes the input is sorted' - it is sorted in our current implementation (checked by a quick look)

`It needs to be convenient to call from our own code: This is the current problem, because calling this with a single value is 3-4 lines of code;` 

extra memory allocation?

source/agora/test/PreimageSharing.d
ferencdg: the default value for TestConf.preimage_catchup_interval is set to the value:

public Duration preimage_catchup_interval = 100.seconds; 

so none of the existing test fails. We could be set it to much higher.

preimage_catchup_interval is more like a switch in NoPreimageVN, if you want no preimage at all, set it to a high value(leave it at default), but you have to option to customize it and share preimages during nomination.

ferencdg: > the default value for TestConf.preimage_catchup_interval is set to the value:
> 
> public Duration preimage_catchup_interval = 100.seconds;
> 
> so none of the existing test fails. We could be set it to much higher.
> 
> preimage_catchup_interval is more like a switch in NoPreimageVN, if you want no preimage at all, set it to a high value(leave it at default), but you have to option to customize it and share ceratain preimages during nomination.

Geod24: Not sure I follow. `NoPreImageVN` should *never* return pre-images if the boolean is set to `false`.
Obviously the rest of the code is unchanged, so it tries to catch-up at a regular interval, but since all validators in the network are `NoPreImageVN`, I would not expect any pre-image to be revealed.

source/agora/test/Base.d
```diff
@@ -2201,16 +2233,6 @@ public class NoPreImageVN : TestValidatorNode
...
-    public override PreImageInfo getPreimage (Hash enroll_key)
-    {
-        if (enroll_key != this.enroll_man.getEnrollmentKey())
-            return super.getPreimage(enroll_key);
-        else if (atomicLoad(*this.reveal_preimage))
-            return super.getPreimage(enroll_key);
-        else
-            throw new Exception("No such PreImage");
-    }
-
```
Geod24, line 2240: Yeah, that should not be removed. It's the whole point of the class.

source/agora/test/Base.d
Geod24: `Set!Hash.from(enroll.utxo_key.only)` should do it ([`only`](https://dlang.org/phobos/std_range.html#only) is in `std.range`).

source/agora/test/PreimageSharing.d
Geod24: > so none of the existing test fails.

They don't fail, but they might start to succeed for the wrong reason.

ferencdg: > Not sure I follow. `NoPreImageVN` should _never_ return pre-images if the boolean is set to `false`.
> Obviously the rest of the code is unchanged, so it tries to catch-up at a regular interval, but since all validators in the network are `NoPreImageVN`, I would not expect any pre-image to be revealed.

NoPreImageVN - there are(or will be) at least 3 ways to reveal preimages
1. sharing preimages in advance in a push way
2. sharing preimages during nomination in a pull way
3. preimage catchup

instead of having NoPreImageVN we should allow all those 3 methods to be enabled/disabled (and not just by a bool value), having 

preimage_catchup_interval = 100.seconds;

should be the first step for that

I don't think we want 2<sup>3</sup> different classes for all the combinations

ferencdg: > > so none of the existing test fails.
> 
> They don't fail, but they might start to succeed for the wrong reason.

:) I don't get this, if we set 
preimage_catchup_interval = 1000000.seconds;

then the only way for a test to succeed for the wrong reason is that the test actually ran 1000000 seconds and succeeded because of nomination preimage sharing

source/agora/test/Base.d
ferencdg: > Yeah, that should not be removed. It's the whole point of the class.
👀 
 
getPreimage (Hash enroll_key) 
was removed from the FullNode's interface...

However it is true that I should have added the new method called
getPreimageWhatever (Set!Hash enroll_keys)

source/agora/test/PreimageSharing.d
ferencdg: if you have any suggestion how you would like to see this, please share in detail and I will implement it that way

ferencdg:
rebase #2 and fixed according to offline discussion

source/agora/test/SlashingMisbehavingValidator.d
Geod24: ~So this change can be reverted nowadays (or put in a separate commit, if you wish).~ Ignore this, I misread the diff.
By the way, it would make sense to have the move as a separate commit, explaining the rationale, which precedes the commit that uses it. It does make reviewing much easier.

source/agora/test/Base.d
```diff
@@ -1941,8 +1941,13 @@ public struct TestConf
 
         // Always set to true, cannot be overriden, but also set here for clarity
         testing: true,
+
```
Geod24: Remove

source/agora/node/Ledger.d
Geod24: Discussed offline, we're going with this approach.

Geod24 (approved):
Just fix the tiny style issue and feel free to merge.

PR #2252, opened 2021-07-09
[WIP] Large enough pre-images not to create new seeds

### LINKED ISSUE #1500: Nodes should be able to recover from exhausting their PreImageCycle

linked0:
Once the preimages in a PreImageCycle are all used, nodes can not enroll again with the same UTXO. So we decided that the number of Pre-Images generated when first enrolling should be large enough that there would unlikely be the need to ever re-enroll with a new pre-image seed. As @hewison-chris suggested, 5 million pre-images are created by default in this PR.

This is the rebased version of PR #1662 which will be closing.

This PR has some problem that makes the total time of unit tests much longer, like increasing 70% of test time because all the preimages of each validator are hashed in the start time. We should prepare well-known preimages for decreasing the test time. So this is why the state is WIP.

Fixes #1500


source/agora/consensus/PreImage.d
```diff
@@ -392,13 +394,30 @@ public struct PreImageCycle
...
+            if (seed == Hash.init)
+            {
+                this.seeds.reset(cycle_seed);
+                this.preimages.reset(this.seeds.byStride[$ - 1 - this.index]);
+            }
+            else
+            {
+                this.preimages.reset(seed);
+            }
+
```
Geod24, line 409: This is why it is quite problematic, in the end, to have WK preimages.
The code here is test-specific but gives no indication of it. Dependency injection should be done via interfaces (in the broad sense, not `interface`), not be adding special cases.

linked0: Would it be better to add `version (unittest)`?

linked0:
Anyway, this PR in WIP does not work as expected. I have to do more analysis about this way and consider another way also.
Feel free to give any opinions.

Geod24:
Superseded by #2291

PR #2263, opened 2021-07-12
CI: Correct and improve MMDB download step

Geod24:
All PRs from fork were failing because the cache action uses the directory as 'path',
not single files like we had configured.
In addition to this fix, the actions/cache repository contains an example which includes
the date, which is exactly what we need to ensure that the MMDB doesn't go stale.


Geod24:
I *think* this will work, will need to rebase one PR *after* one of the CI job completes.

PR #2264, opened 2021-07-12
Talos: Upgrade to node 16

Geod24:
Need to upgrade to node 16 because Homebrew defaults to node 16 by default, so out-of-the-box experience for new users isn't great.


kchulj (approved):
✅

linked0 (approved):

PR #2265, opened 2021-07-12
FlashListener: Add on-chain fees related endpoints

omerfirmak:
```
Once a flash node starts an uncollaborative close attempt,
it will need to add its own fees to the settlement/update TXs. To
do this, it should be able to query the wallet for a reasonable fee
for the TX and the UTXOs that would supply that amount.
```

I am open to any and all suggestions.


Geod24 (approved):

omerfirmak:
Well, that was easy.

kchulj (approved):
LGTM

PR #2272, opened 2021-07-14
Do not store the secret inside `PreImageCycle`

Geod24:
The first commit has been extracted from #2252 
The second one is a result of a discussion with @linked0 yesterday.


hewison-chris (approved):
LGTM, just a typo in the commit description. Also maybe worth noting we do not need to store secret as we no longer support creating a new seed with a new nonce value.

linked0 (approved):
LGTM. Sorry for that I should have done this.

Geod24:
Fixed the typo

kchulj (approved):

PR #2273, opened 2021-07-14
PreImageCycle: Allow creation from a PI + height

Geod24:
This will allow us to extend the default number of pre-images to a very large number,
while still keeping a limited number of pre-images in the test network.

### DIFF
--- a/source/agora/consensus/PreImage.d
+++ b/source/agora/consensus/PreImage.d
@@ -291,17 +291,71 @@ public struct PreImageCycle
 
     public static immutable ulong NumberOfCycles = 100;
 
-    /// Construct an instance with the provided cycle length and number of cycles
+    /***************************************************************************
+
+        Construct an instance based on a secret
+
+        The secret will be hashed in a predictable manner, and this will give
+        the 'seed' hash, which will correspond to the last pre-image the node
+        will be able to reveal (at height `cycles * cycle_length`).
+
+        This overload is provided for convenience. To have more control over
+        the pre-image, see the overload accepting a Hash.
+
+        Params:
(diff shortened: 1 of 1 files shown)


source/agora/consensus/PreImage.d
```diff
@@ -291,17 +291,71 @@ public struct PreImageCycle
...
+
     public this (in Scalar secret, in uint cycle_length, in ulong cycles = PreImageCycle.NumberOfCycles,
-        Height initial_seek = Height(0))
+        in Height initial_seek = Height(0))
+    {
+        // Using this predictable scheme ensures that the pre-image can
+        // be recovered in case of a crash or if another node takes over
+        // (e.g. disk failure from a server leads to a backup server starting
+        // with no state or way to recover it).
+        const cycle_seed = hashMulti(secret, "consensus.preimages", 0);
```
hewison-chris, line 320: Do we need to keep this `nonce` of zero still?

hewison-chris (approved):
LGTM

source/agora/consensus/PreImage.d
Geod24: Not that I can think of. The only upside is to be consistent with how other parameters are computed.

linked0 (approved):
LGTM. I can use the second constructor in my PR. Ty.

source/agora/consensus/PreImage.d
linked0: It's not necessary. I will remove it which should have been removed in the recent commit I have submitted.

kchulj (approved):

PR #2276, opened 2021-07-14
Test: Use NodeConfig directly instead of copying members

Geod24:
```
As we added more elements into NodeConfig, they were also added to TestConf.
The duplication was obvious, and the resolution is to simply embed a NodeConfig
in TestConf as the template used for each nodes.
```

We could probably extend this in the future, but `NodeConfig` was the worst offender.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -795,7 +795,7 @@ public class TestAPIManager
 
     public TimePoint getBlockTimeOffset (Height height)
     {
-        return height * this.test_conf.block_interval_sec;
+        return height * this.test_conf.node.block_interval_sec;
     }
 
     /***************************************************************************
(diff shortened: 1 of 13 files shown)


Geod24:
Can I get a review on this since it simplifies a few other pending PRs ?
The failures are unrelated - looking into them tho.

source/agora/test/Base.d
```diff
@@ -1932,23 +1958,6 @@ public struct TestConf
...
-    Duration relay_tx_interval = 0.seconds;
-
-    /// The minimum amount of fee a transaction has to have to be relayed.
-    /// The fee is adjusted by the transaction size:
-    /// adjusted fee = fee / transaction size in bytes.
-    Amount relay_tx_min_fee = 0;
-
-    /// Transaction put into the relay queue will expire, and will be removed
-    /// after `relay_tx_cache_exp`.
-    Duration relay_tx_cache_exp = 60.minutes;
```
linked0, line 1951: Shouldn't it be contained in the `node` of `TestConf`?

Geod24: You mean, the `60.minutes` value ? Well, I looked at the configuration, and the reason why we have 60 minutes is to ensure that we don't evict anything, essentially. But the default already does that. Ping @ferencdg to check my sanity.

linked0 (approved):
LGTM

hewison-chris (approved):
LGTM

kchulj (approved):

PR #2277, opened 2021-07-14
Allow generation of empty blocks

### LINKED ISSUE #2245: Allow generation of empty blocks
there has been a meeting recently and the conclusion was to allow generation of empty blocks to have a predictable real-life time for

- votera vote deadline
- staked UTXO unfroze deadline
- block reward

all those 3 functionality depends on block height, and we would continue to use the block height

omerfirmak:
Closes #2245

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -304,12 +304,13 @@ public struct Block
 
     public static Hash buildMerkleTree (in Transaction[] txs,
         ref Hash[] merkle_tree) nothrow @safe
-    in
     {
-        assert(txs !is null, "When calling buildMerkleTree, `txs` must not be null.");
-    }
-    do
-    {
+        if (txs.length == 0)
+        {
+            merkle_tree.length = 0;
+            return Hash.init;
+        }
+
         immutable pow2_size = getPow2Aligned(txs.length);
         const MerkleLength = (pow2_size * 2) - 1;
 
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -367,11 +367,6 @@ extern(D):
     protected bool prepareNominatingSet (out ConsensusData data) @safe
     {
         this.ledger.prepareNominatingSet(data, MaxTransactionsPerBlock, this.nomination_start_time);
-        if (data.tx_set.length < 1)
-        {
-            this.log.trace("prepareNominatingSet(): No transaction to nominate yet");
-            return false;
-        }
 
         // check whether the consensus data is valid before nominating it.
         if (auto msg = this.ledger.validateConsensusData(data, data.missing_validators))
@@ -971,10 +966,6 @@ extern(D):
             abort();
         }
 
-        // enrollment data may be empty, but not transaction set
-        if (data.tx_set.length == 0)
-            assert(0, format!"Transaction set empty for slot %s"(height));
-
         log.info("Externalized consensus data set at {}: {}", height, prettify(data));
         try
         {
@@ -1164,6 +1155,11 @@ extern(D):
             if (this.total_adjusted_fee > other.total_adjusted_fee)
                 return -1;
             else if (this.total_adjusted_fee < other.total_adjusted_fee)
+                return 1;
+
+            if (this.consensus_data.tx_set.length > other.consensus_data.tx_set.length)
+                return -1;
+            else if (this.consensus_data.tx_set.length < other.consensus_data.tx_set.length)
                 return 1;
 
             if (this.hash < other.hash)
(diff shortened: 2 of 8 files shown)


source/agora/consensus/data/Block.d
```diff
@@ -304,12 +304,13 @@ public struct Block
...
-        assert(txs !is null, "When calling buildMerkleTree, `txs` must not be null.");
-    }
-    do
     {
+        if (txs.length == 0)
+        {
+            merkle_tree.length = 0;
+            return Hash.init;
+        }
+
```
Geod24, line 313: Won't we have at least ONE transaction in the block though (CB) ?

source/agora/test/Base.d
```diff
@@ -1027,35 +1027,44 @@ public class TestAPIManager
...
-
+        auto last_height = first_client.getBlockHeight();
         // traget height will be one more than previous block
-        Height target_height = last_block.header.height + 1;
+        Height target_height = Height(last_height + 1);
 
-        // Get spendables from last block
-        auto spendables = last_block.spendable().array;
+        while (!no_txs)
+        {
```
Geod24, line 1048: I had a [wat](https://www.destroyallsoftware.com/talks/wat) moment. Nice logic though.

source/agora/test/EmptyBlocks.d
```diff
@@ -0,0 +1,38 @@
...
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto nodes = network.clients;
+    auto node_1 = nodes[0];
+
+    // Create blocks for 2 validator cycles and a couple more
+    auto target_height = Height(2 * GenesisValidatorCycle + 2);
+    network.generateBlocks(target_height, true);
+    network.assertSameBlocks(target_height);
```
Geod24, line 37: This doesn't check that the block only have a CoinBase though ?

source/agora/flash/Channel.d
```diff
@@ -2206,6 +2211,18 @@ LOuter: while (1)
...
+    public Transaction getLastUpdateTx ()
+    {
+        return this.channel_updates[$ - 1].update_tx;
+    }
+
+    version (unittest)
+    public Transaction getTriggerTx ()
+    {
+        return this.channel_updates[0].update_tx;
+    }
```
Geod24: Uuuuuhhhhh we really shouldn't have those

source/agora/test/Fee.d
```diff
@@ -47,6 +48,7 @@ unittest
 
         // send it to one node
         txs.each!(tx => node_1.putTransaction(tx));
+        txs.each!(tx => network.waitForTXPropagation(tx.hashFull()));
```
Geod24: Perhaps add an overload that accepts a `Transaction` then ? ;)

Geod24 (changes requested):
Few things. Also I doubt the commits are atomic (reminder that atomic commits pass the test suite on their own).

source/agora/consensus/data/Block.d
ferencdg: I think no, block reward is only payed in every X block, and fee payment(if there was any) is also done only in certain blocks

source/agora/test/EmptyBlocks.d
omerfirmak: We could, if we really want to be explicit. But since we use `no_txs`, externalization of blocks is enough evidence that it works

I'll add it regardless, for clarity.

source/agora/flash/Channel.d
omerfirmak: Alternative is arbitrary sleeps, pick your poison:)
We already had similar helpers anyways.

omerfirmak: Ok, I might have talked too soon. Good news incoming.

omerfirmak:
Ok, this is much better now. Updated

source/agora/consensus/protocol/Nominator.d
```diff
@@ -1166,6 +1157,11 @@ extern(D):
             else if (this.total_adjusted_fee < other.total_adjusted_fee)
                 return 1;
 
+            if (this.consensus_data.tx_set.length > other.consensus_data.tx_set.length)
+                return -1;
+            else if (this.consensus_data.tx_set.length < other.consensus_data.tx_set.length)
+                return 1;
+
```
Geod24, line 1164: I'm a bit puzzled. If `this` has more transaction than `other`, then `this < other` ? I'd expect the other way around.

Geod24 (changes requested):
Just one thing.

source/agora/consensus/protocol/Nominator.d
omerfirmak: Well, we first `sort()` and get `front()`. So our best candidate should be the "smallest"

Geod24: Right, it's `CandidateHolder`

Geod24 (approved):

PR #2291, opened 2021-07-20
Large enough pre-images not to create new seeds

linked0:
Once the preimages in a `PreImageCycle` are all used, nodes can not enroll again with the same UTXO. So we decided that the number of Pre-Images generated when first enrolling should be large enough that there would unlikely be the need to ever re-enroll with a new pre-image seed. As discussed with the team, 5 million pre-images are created by default in this PR.

This PR has also improved the testing time from [the previous PR ](https://github.com/bosagora/agora/pull/2252), so needs only 20% more time for testing compared to the `v0.x.x` branch. I will improve more about the test time while this PR is being reviewed. The reason why I'm submitting this PR although there's room for improvement is to share the concept the setting seeds for pre-images into `PreImageCycle`s and to get opinions for the concept earlier.

The improved code could be cherry-picked to this PR if there's no confusion.


source/agora/consensus/EnrollmentManager.d
```diff
@@ -929,6 +951,24 @@ public class EnrollmentManager
     }
 }
 
+// A `PreImageCycle` for ramdon keys can be created just with a cycle seed  without generating
```
hewison-chris:
```suggestion
// A `PreImageCycle` for ramdon keys can be created just with a cycle seed without generating
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -929,6 +951,24 @@ public class EnrollmentManager
...
+// 5 million preimages. Thses pre-defined cycle seeds are used for that.
```
hewison-chris:
```suggestion
// 5 million preimages. These pre-defined cycle seeds are used for that.
```

source/agora/consensus/EnrollmentManager.d
```diff
@@ -929,6 +951,24 @@ public class EnrollmentManager
...
+        Hash(`0x3a481cb1576d79002755239b8d4019587d7e5394ddd448f92d5dca74baf742f4899dd53daefda9136e9d3cc6887b3456ffcfc85cfc711e5cf8659998047771cb`),
+        Hash(`0x1a7ebfe71dd438ac96da520917680ff4278c3e4de031890cea2305de009dc750dfd3d9b714ef7b4cf983633a540ca4614ff1eae38d9d121604907f2dc2a7e885`),
+        Hash(`0x64d1b6e98019df73b59c3593e2ca1baea53aeeb3cac18748d3ab50473a64963c76c318913158349caf368730b027eb8c4596b743308a7e53a65e484016c8e2ca`),
+        Hash(`0x0bb70cf7d265661d006bc5bfd5bf70a66c5b32e54d05ee37bf023c4c0bf9b3bc76faf38d656c25e254aae7ea73aabba1902192575da80cfe10a24363d3e41118`),
+        Hash(`0xbb19e8bae8f03ce2aee0c15f83089c87dd208b44d17683ace877696a98680730c4cae9f762366a70dd00885cf0ff2eb228f285885ebda9e6dc11c0288158d8cf`),
+        Hash(`0x90f2a3d469831bad2d27f4698c463dad56c8d942b28859e38c2f4b1bc975605f0641f62aff2edf23326b67ad3b34c47bc43f96167485d4db2529d5e84fc7e994`),
+        Hash(`0xa84c60d88e1c075d8d70c77426c4724ac8bd38f83030fb16a3db71890396548572031907fbf7adba1034a7f9b7b4f2ae11155010cf677ed81cc32728f6ec7a06`),
+        Hash(`0xe39ae981988560d89e85f340c087fea46f1d9eefe73400f2a85edd3d3adee2bc5a3efd6e6ce9de04d73a3fe2554a1b1035450b7be9eaade24bbb789585aa0015`),
+    ];
```
hewison-chris: I think we need some code comments here to explain what cycle these seeds are for.

hewison-chris (commented):
I noticed a couple of small things so far.
I will continue review once I have pulled your branch and take a closer look. It is a tricky one to check just from the code changes as seen in GitHub.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -170,13 +170,23 @@ public class EnrollmentManager
     ***************************************************************************/
 
     public this (ManagedDatabase stateDB, ManagedDatabase cacheDB,
-                 KeyPair key_pair, immutable(ConsensusParams) params)
+        KeyPair key_pair, immutable(ConsensusParams) params,
+        PreImageCycle cycle = PreImageCycle.init)
```
Geod24, line 175: In this case, it would be better to have an overload that just forwards to this.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -193,10 +203,11 @@ public class EnrollmentManager
 
     /// Unittest-only constructor
     version (unittest) private this (
-        KeyPair key_pair, immutable(ConsensusParams) params)
+        KeyPair key_pair, immutable(ConsensusParams) params,
+        PreImageCycle cycle = PreImageCycle.init)
```
Geod24, line 208: Likewise, overload if needed. But in this case, we probably always want to pass a smaller value, so does it need to accept a `cycle` at all ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -958,19 +1002,21 @@ unittest
...
+    auto cycle = PreImageCycle(CommonCycleSeed, Height(params.ValidatorCycle * 2 - 1), params.ValidatorCycle);
+    auto man = new EnrollmentManager(key_pair, params, cycle);
     // Useful constant
     const EnrollAt1 = Height(1);
 
     // check the return value of `getEnrollmentPublicKey`
     assert(key_pair.address == man.getEnrollmentPublicKey());
 
-    // create and add the first Enrollment object
+    // create and add the first Enrollment object]
```
Geod24: ?

Geod24:
~Note that the best way to make sure this is correct would be to not modify any of the test value: That is, use the 2000th PI in the `version (unittest)` constructor and related.~ Nevermind, this won't work.

source/agora/consensus/EnrollmentManager.d
linked0: Right. Looks better

source/agora/consensus/EnrollmentManager.d
```diff
@@ -929,6 +951,28 @@ public class EnrollmentManager
...
+{
+    static immutable CommonCycleSeed = Hash(`0xb7f3802d774665c6ccb4d24f3e6128542185f847e6da4e9c5a0254cfeb885d2536a89be8b882f0917f3ed43dae4760bb3a85b4c5a9e1871872a794fcf5e5236d`);
+}
+
+/// Each `PreImageCycle` for ramdon keys can be created just with a cycle seed without generating
+/// 5 million preimages. These pre-defined cycle seeds are used for that.
+version (unittest)
+{
+    static immutable Hash[] NodeCycleSeeds = [
+        Hash(`0x3a481cb1576d79002755239b8d4019587d7e5394ddd448f92d5dca74baf742f4899dd53daefda9136e9d3cc6887b3456ffcfc85cfc711e5cf8659998047771cb`),
```
Geod24: Logically, I think those would belong to `genesis/Test.d`. But I am still reviewing, any reason why they are in this module ?

source/agora/common/Config.d
```diff
@@ -244,6 +244,12 @@ public struct ValidatorConfig
     /// The seed to use for the keypair of this node
     public immutable KeyPair key_pair;
 
+    /// The seed of PreImageCycle
+    public Hash cycle_seed;
+
+    /// The height of the seed of PreImageCycle
+    public Height cycle_seed_height;
+
```
Geod24: Is this used ? I don't see it being parsed and validated.

source/agora/consensus/EnrollmentManager.d
linked0: The values are kind of random just to be used in the tests in the `EnrollmentManager` module.

source/agora/consensus/PreImage.d
```diff
@@ -487,3 +490,73 @@ unittest
...
+/// which are used for getting 1000 number of pre-images
+version (unittest) public void getCycleSeed (in KeyPair pair, out Hash seed,
+    out Height seed_height)
+{
+    import agora.utils.WellKnownKeys;
+    import std.format;
+
+    seed_height = Height(999); // which means 1000 number of pre-images.
+    if (pair == NODE7)
+        seed = Hash(`0xee7941fc52af778dcafecd2c4695339b25f0be075b8bff010e0aded69a580a4617f80a3c2c068efa4b2fa7161162142732b15b7b192ac35ffe694e74d0946d0d`);
```
Geod24: What does those hardcoded value correspond to ?

linked0: It's the seed for the 1000 pre-images for `NODE7` which is in fact the hash value hashed (`PreImageCycle.PreImageCount` - 1000) times from the code of `hashMulti(secret, "consensus.preimages", 0)`. Anyway I'll documenting it.

source/agora/consensus/EnrollmentManager.d
linked0: Updated.

source/agora/consensus/EnrollmentManager.d
linked0: Updated

source/agora/consensus/EnrollmentManager.d
linked0: Will add more detail

source/agora/consensus/EnrollmentManager.d
```diff
@@ -439,6 +450,17 @@ public class EnrollmentManager
...
+    /// Ditto
+    version (unittest) public static Enrollment makeEnrollment (
+        in Hash utxo, in KeyPair key, in Hash cycle_seed, in Height seed_pos,
+        in Height height, uint cycle_length)
+        @trusted nothrow
+    {
+        // Generate the random seed to use
+        auto cycle = PreImageCycle(cycle_seed, seed_pos, cycle_length);
+        return makeEnrollment(utxo, key, cycle[height], cycle_length);
+    }
```
Geod24: I think you can just replace the previous function, I don't see why we would want to use the production value and generate 5M PIs in tests.
Also I would change the order of parameters to match other overloads:
```D
    version (unittest) public static Enrollment makeEnrollment (
        in Hash utxo, in KeyPair key, in Height height, uint cycle_length,
        in Hash cycle_seed, in Height seed_pos)
        @trusted nothrow
```
It's usually pretty confusing (and dangerous) to have overloads which have swapped parameters.

source/agora/consensus/EnrollmentManager.d
Geod24: I think you could put the values in https://github.com/bosagora/agora/pull/2291#discussion_r673655771 in `genesis/Test.d` and use that everywhere, right ?

linked0: I think so. It may not be needed. I'll look into it.

source/agora/common/Config.d
linked0: It's used here and look into it more about validation.
https://github.com/linked0/agora/blob/06be326146d4da608fc6ea229088e592bfac9be4/source/agora/test/Base.d#L2087-L2101
 If you meant adding the items into the configuration file, It could be in another PR.

source/agora/consensus/EnrollmentManager.d
linked0: Updated.

source/agora/consensus/PreImage.d
```diff
@@ -487,3 +490,73 @@ unittest
...
+version (unittest) public Hash getSeedAt(in Scalar secret, in ulong total_images,
+    in uint cycle_length, in Height at) nothrow
+{
+    auto cycle = PreImageCycle(secret, cycle_length, total_images, Height(0));
+    return cycle[at];
+}
+
+/// To get the cycle seeds for `PreImageCycle`s of some of well-known keys,
+/// which are used for getting 1000 number of pre-images
+version (unittest) public void getCycleSeed (in KeyPair pair, out Hash seed,
```
Geod24: Since this function is only used in `test/Base.d`, move it there and make it private ?

source/agora/consensus/PreImage.d
```diff
@@ -487,3 +490,73 @@ unittest
...
     testPreImageCycle(cycle_length, number_of_cycles);
 }
+
+///
+version (unittest) public Hash getSeedAt(in Scalar secret, in ulong total_images,
+    in uint cycle_length, in Height at) nothrow
+{
+    auto cycle = PreImageCycle(secret, cycle_length, total_images, Height(0));
+    return cycle[at];
+}
```
Geod24: Is this used ?

source/agora/common/Config.d
Geod24: I see, I was expecting that it would be parsed from the configuration file. I guess that makes sense to delay it, just add a comment saying this is currently not parsed ?

source/agora/consensus/PreImage.d
linked0: It's not used now. Will Remove.

source/agora/consensus/PreImage.d
linked0: It could be. I'll try it.

source/agora/common/Config.d
linked0: OK, I will.

source/agora/test/Base.d
```diff
@@ -1589,8 +1590,19 @@ private mixin template TestNodeMixin ()
...
+        {
+            auto cycle = PreImageCycle(this.config.validator.cycle_seed,
+                this.config.validator.cycle_seed_height,
+                this.params.ValidatorCycle);
+            return new EnrollmentManager(this.stateDB, this.cacheDB,
+                this.config.validator.key_pair, this.params, cycle);
+        }
+        else
+            return new EnrollmentManager(this.stateDB, this.cacheDB,
+                this.config.validator.key_pair, this.params);
```
Geod24: This probably belongs to the base node class, doesn't it ?

linked0:
I compared the testing time between v0.x.x and this PR by adding a third commit which prints out the testing time of each unit test and will be deleted later. This is the test result.
[Testing time of unit tests between v0.x.x and PR #2291](https://gist.githubusercontent.com/linked0/792fe0b466ad427a9eefbe7648bce066/raw/a1aab370bcff17be31c7f9e04547d6da44fed0a0/gistfile1.txt).
And I will improve it, which means decrease the testing time, by check the result and using the `perf report` command that @Geod24 suggested and show me how to use.

source/agora/consensus/EnrollmentManager.d
linked0: Updated.

source/agora/common/Config.d
linked0: Updated.

source/agora/consensus/PreImage.d
linked0: Updated

source/agora/consensus/EnrollmentManager.d
Geod24: Ping?

linked0: I intended to explain about this later. I tried to change the code like the following but it doesn't work. An error like `Program exited with code -8` occurred.

```
public this (ManagedDatabase stateDB, ManagedDatabase cacheDB,
       KeyPair key_pair, immutable(ConsensusParams) params,
       PreImageCycle cycle)
 {
     this.log = Logger(__MODULE__);
     assert(params !is null);
     this.params = params;
     this.key_pair = key_pair;

     this.cycle = cycle;

     this.db = stateDB;
     this.validator_set = new ValidatorSet(stateDB, params);
     this.enroll_pool = new EnrollmentPool(cacheDB);


     // create the table for enrollment data for a node itself
         "(key TEXT PRIMARY KEY, val TEXT NOT NULL)");

     // load enrollment key
     this.enroll_key = this.getEnrollmentKey();
 }

 public this (ManagedDatabase stateDB, ManagedDatabase cacheDB,
     KeyPair key_pair, immutable(ConsensusParams) params)
 {
     if (key_pair != KeyPair.init)
         this(stateDB, cacheDB, key_pair, params,
             PreImageCycle(key_pair.secret, params.ValidatorCycle));
     else
         this(stateDB, cacheDB, key_pair, params, PreImageCycle.init);
 }

 /// Unittest-only constructor
 version (unittest) private this (
     KeyPair key_pair, immutable(ConsensusParams) params, PreImageCycle cycle)
 {
     this(new ManagedDatabase(":memory:"), new ManagedDatabase(":memory:"),
         key_pair, params, cycle);
 }

 /// Ditto
 version (unittest) private this (
     KeyPair key_pair, immutable(ConsensusParams) params)
 {
     if (key_pair != KeyPair.init)
         this(key_pair, params, PreImageCycle(key_pair.secret, params.ValidatorCycle));
     else
         this(key_pair, params, PreImageCycle.init);
 }
```

source/agora/consensus/PreImage.d
linked0: Updated.

linked0:
@Geod24 Could you please look at this error as follows. It occurred by adding the third commit which is just for improving the testing time of `agora.node.Ledger` module and will be merged into the first commit.
```
Linking...
To force a rebuild of up-to-date targets, run again with --force.
Running build/agora-unittests 
0   agora-unittests                     0x0000000102f68c2b _D4core7runtime9backtraceFNbPPviZi + 75
1   agora-unittests                     0x0000000102f68bba _D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbiPSQCk3sys5posix6signal9siginfo_tPvZv + 58
2   libsystem_platform.dylib            0x00007fff6f7db5fd _sigtramp + 29
3   agora-unittests                     0x000000010216957d _D5agora4node6Ledger10TestLedger6__ctorMFSQBo6crypto3Key7KeyPairAxSQCn9consensus4data5BlockQgyCQDpQBcQv6Params15ConsensusParamsS4core4time8DurationCQFq7network5ClockQgZCQGlQGiQGgQGc + 1197
4   agora-unittests                     0x0000000102169eec _D5agora4node6Ledger19__unittest_L1720_C1FZv + 348
5   agora-unittests                     0x000000010218542b _D5agora4node6Ledger10__unittestZ + 11
6   agora-unittests                     0x00000001020421c3 _D5agora4test4Base22customModuleUnitTesterFZ7runTestMFSQCbQByQBwQBuFZ7ModTestZb + 371
7   agora-unittests                     0x000000010204368f _D5agora4test4Base22customModuleUnitTesterFZ13runInParallelMFASQCjQCgQCeQCcFZ7ModTestZ10WorkThread3runMFZv + 63
8   agora-unittests                     0x0000000102f6bfbd _D4core6thread7context8Callable6opCallMFZv + 61
9   agora-unittests                     0x0000000102f6fa69 _D4core6thread10threadbase10ThreadBase3runMFZv + 25
10  agora-unittests                     0x0000000102f6cff5 _D4core6thread8osthread6Thread3runMFZv + 21
11  agora-unittests                     0x0000000102f6d5e8 thread_entryPoint + 200
12  libsystem_pthread.dylib             0x00007fff6f7e7109 _pthread_start + 148
Program exited with code -11
```

source/agora/node/Ledger.d
```diff
@@ -1680,13 +1681,21 @@ version (unittest)
                     params.GenesisTimestamp +
                     (blocks.length * params.BlockInterval.total!"seconds"));
 
+            Hash cycle_seed;
+            Height cycle_seed_height;
+            getCycleSeed(key_pair, cycle_seed, cycle_seed_height);
+            assert(cycle_seed != Hash.init);
+            assert(cycle_seed_height != Height(0));
+            auto cycle = PreImageCycle(cycle_seed, cycle_seed_height,
+                this.params.ValidatorCycle);
```
Geod24: `params` not `this.params` because `this.params` is set by the call to `super`, which occurs after this.

source/agora/consensus/EnrollmentManager.d
linked0: We decided to leave the ctor as it is.

source/agora/consensus/EnrollmentManager.d
linked0: I moved it to PreImage.d. I tried to move it to `Base.d` but it gets the cycling dependency.

source/agora/test/Base.d
linked0: Moved it to `FullNode.d`.

source/agora/consensus/EnrollmentManager.d
linked0: In `agora.consensus.data.genesis`, the previous function should be used. So I leave the function as it is. But I changed the order of parameters to match other overloads.

source/agora/consensus/EnrollmentManager.d
linked0: We decided to leave the ctor as it is also.

source/agora/consensus/PreImage.d
linked0: Added more detail.

source/agora/consensus/EnrollmentManager.d
linked0: Added some more detail above the code.

source/agora/test/NetworkManager.d
```diff
@@ -178,6 +179,12 @@ unittest
     node_bad.clearFilter();
     node_validators.each!(node => node.clearFilter());
 
+    // The testing time decreases compared to the last code because setting
+    // seeds to the `PreImageCycle`s make it unnecessary to generate kind of
+    // large number seeds. So we have to add some preparation time to this
+    // network test like this.
+    Thread.sleep(5.seconds);
+
```
hewison-chris: Shouldn't this be removed in the second commit, which I guess will be squashed into first?

linked0: Currently, it should remain because the test gets to need some waiting time with the pre-image generating time decreasing. I think that the `catchupTask` starts kind of late after applying the code of this PR.

linked0:
All green. Please review.

source/agora/consensus/PreImage.d
```diff
@@ -487,3 +490,93 @@ unittest
...
+
+/// To get the cycle seeds for `PreImageCycle`s of some of well-known keys,
+/// which are used for getting 5040 number of pre-images
+public void getCycleSeed (in KeyPair pair, out Hash seed, out Height seed_height)
+{
+    import agora.consensus.PreImage;
+    import agora.utils.WellKnownKeys;
+    import std.format;
+
+    // 10 times of `Validator Cycle` minus 1, which is 5040 - 1
```
hewison-chris: I thought the `ValidatorCycle` was 20 for our  test `GenesisBlock`.

source/agora/consensus/PreImage.d
```diff
@@ -487,3 +490,93 @@ unittest
...
+        seed = Hash(`0xd244e2d05a4520799ba0d927e949087ef125524f1ba1b75d8d3b7e06983496a546b50ef5f2bac1b6db8f81dcadd30b87e67e67e46a42c863a4938c4485fafeeb`);
+    else if (pair == H)
+        seed = Hash(`0x0e2a80082b26abab2b5c1d6e7e6c15789d68c551c318d614ec4bb96e92bcec484ea13e50ff614e33f64c9fa2ff1783ebeb4d480cde7b3285a2475fc19e974532`);
+    else if (pair == J)
+        seed = Hash(`0x2f3273381c16e00509adbd4fb1871a1aa640b6c908ba5837775b02c9cec4e410f8057d0f7f2f742cbae73da9f3086d68cdb4daab54a0665651fcc57f74c6ec29`);
+    else if (pair == K)
+        seed = Hash(`0xc2929018e4cda8ce470d8f6a57052abe05ae4e095cd7f63b2654c7c457d3c085f110a1e8048646bef51cc3ef03442ce3452f267bfd92538b0dbfab0b479560f0`);
+    else
+    {
+        auto cycle = PreImageCycle(pair.secret, 10, PreImageCycle.PreImageCount, Height(0));
```
hewison-chris: So 10 here is the validator cycle? Maybe it is not needed in the `PreImageCycle` anymore as it only depends on the `height`.

source/agora/test/NetworkManager.d
```diff
@@ -178,6 +179,12 @@ unittest
     node_bad.clearFilter();
     node_validators.each!(node => node.clearFilter());
 
+    // The testing time decreases compared to the last code because setting
+    // seeds to the `PreImageCycle`s make it unnecessary to generate kind of
+    // large number seeds. So we have to add some preparation time to this
+    // network test like this.
+    Thread.sleep(8.seconds);
+
```
hewison-chris, line 188: We probably should find out why. If the setup is quicker it should not mean that we need this extra time for this test.

hewison-chris (commented):
Good work. Just a few comments.

source/agora/consensus/PreImage.d
linked0: Right. It could say `252 times of the ValidatorCycle of the test GenesisBlock minus 1, which is 5040 - 1`. In this comment, I intended to say the number of pre-images based on the ValidatorCycle of the Coinnet GenesisBlock. Anyway, the important thing is the number of pre-images like 5040 in this comment.

source/agora/consensus/PreImage.d
linked0: Yes, it's needed because we have seeds for all validator cycles. The validator cycle means the interval between the seeds. If it's not there, we have to have all the pre-images.

source/agora/test/NetworkManager.d
linked0: Right. I will look into it and leave the `TODO` comment about that. Ty.

source/agora/consensus/PreImage.d
hewison-chris: But we can only use `WellKnownKeys` with the test `GenesisBlock`.

source/agora/consensus/PreImage.d
hewison-chris: Perhaps it can be renamed for the purpose it is used. Maybe `seed_interval` as it is no longer coupled with the `enrollment cycle` as far as I can tell.

source/agora/consensus/PreImage.d
linked0: WellKnownKeys are used in some unit tests. But, I will add more detail for clarity.

source/agora/consensus/PreImage.d
linked0: Yes. `seed_interval` could be better because `enrollment cycle` is not related to the interval anymore as you said. But the change of the name is not the main thing of this PR and we can think we use the validator cycle as seed interval. So let's leave it as it is at the moment.

source/agora/consensus/PreImage.d
```diff
@@ -487,3 +490,93 @@ unittest
...
+        Hash(`0xbb19e8bae8f03ce2aee0c15f83089c87dd208b44d17683ace877696a98680730c4cae9f762366a70dd00885cf0ff2eb228f285885ebda9e6dc11c0288158d8cf`),
+        Hash(`0x90f2a3d469831bad2d27f4698c463dad56c8d942b28859e38c2f4b1bc975605f0641f62aff2edf23326b67ad3b34c47bc43f96167485d4db2529d5e84fc7e994`),
+        Hash(`0xa84c60d88e1c075d8d70c77426c4724ac8bd38f83030fb16a3db71890396548572031907fbf7adba1034a7f9b7b4f2ae11155010cf677ed81cc32728f6ec7a06`),
+        Hash(`0xe39ae981988560d89e85f340c087fea46f1d9eefe73400f2a85edd3d3adee2bc5a3efd6e6ce9de04d73a3fe2554a1b1035450b7be9eaade24bbb789585aa0015`),
+    ];
+}
+
+/// To get the cycle seeds for `PreImageCycle`s of some of well-known keys,
+/// which are used for getting 5040 number of pre-images
+public void getCycleSeed (in KeyPair pair, out Hash seed, out Height seed_height)
```
Geod24: Shouldn't this be in the `version (unittest)` ?

linked0: Right, I will add it to the function.

source/agora/test/NetworkManager.d
linked0: Updated

source/agora/consensus/PreImage.d
linked0: Updated

source/agora/consensus/PreImage.d
linked0: Added more details

hewison-chris (approved):
LGTM, I think we should raise an issue to not forget about the sleep in the unit test. Also maybe an issue for the suggested rename of parameter in `PreImageCycle` constructor.

linked0:
> LGTM, I think we should raise an issue to not forget about the sleep in the unit test. Also maybe an issue for the suggested rename of parameter in `PreImageCycle` constructor.

Okay, I will raise issues today and find out there's something to do before that.

source/agora/consensus/PreImage.d
```diff
@@ -341,9 +343,10 @@ public struct PreImageCycle
                  in Height initial_seek = Height(0))
     {
         // Make sure that what we got falls on the boundary of a seed
-        assert((at % cycle_length) == 0,
+        auto requested_length = at + 1;
+        assert((requested_length % cycle_length) == 0,
                "Cannot create a `PreImageCycle` at this height");
-        const cycles = at / cycle_length;
```
Geod24, line 346: That's actually a bug fix, isn't it ? Nice find.

Geod24 (approved):
LGTM

PR #2295, opened 2021-07-21
Stats: Change agora_preimages_gauge to a counter

Geod24:
We were previously using a gauge, as the value could go up and down,
since it was a distance from the enrollment.
Nowadays, the value is absolute (it's height), hence we should use a counter.


kchulj (approved):
LGTM

linked0 (approved):

PR #2303, opened 2021-07-22
Add attributes to TxBuilder.deduct

kchulj:
The attributes are needed in order to call the function from faucet methods.
It is used when trying to deduct fees.

### DIFF
--- a/source/agora/utils/TxBuilder.d
+++ b/source/agora/utils/TxBuilder.d
@@ -459,7 +459,7 @@ public struct TxBuilder
 
     ***************************************************************************/
 
-    public ref typeof(this) deduct (Amount amount)
+    public ref typeof(this) deduct (Amount amount) @safe nothrow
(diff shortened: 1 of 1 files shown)


source/agora/utils/TxBuilder.d
```diff
@@ -459,7 +459,7 @@ public struct TxBuilder
 
     ***************************************************************************/
 
-    public ref typeof(this) deduct (Amount amount)
+    public ref typeof(this) deduct (Amount amount) nothrow @safe
```
Geod24:
```suggestion
    public ref typeof(this) deduct (Amount amount) @safe nothrow
```

kchulj: Fixed~

linked0 (approved):

Geod24 (approved):

PR #2304, opened 2021-07-22
Change high frequency free running timers to on-demand tasks

omerfirmak:
We could maybealso increase their firing delay from 10 msecs to something more reasonable


Geod24:
I think you can reuse the same instance, and call `rearm` on it, instead of calling `setTimer` again.

Geod24:
Also if it's not period, do you need to `stop` it ?

omerfirmak:
> 
> 
> I think you can reuse the same instance, and call `rearm` on it, instead of calling `setTimer` again.

Our common interface only has stop, that might be a vibe.d thing maybe.

https://github.com/bosagora/agora/blob/236b2ee0d98254c498bef5eda9ef5539fd785b10/source/agora/common/Task.d#L96-L105

omerfirmak:
> 
> 
> Also if it's not period, do you need to `stop` it ?

If it is enabled (non-null timer), we dont want it to fire one last time after `shutdown`. This might not work as expected with localrest tho, I'll check.

omerfirmak:
Ok, it works fine in LocalRest too

https://github.com/Geod24/localrest/blob/3b650c53ea8af30061a7feab5af43992ad88655b/source/geod24/LocalRest.d#L537-L538

Geod24:
> Our common interface only has stop

Well, you know what to do :D

Geod24:
> If it is enabled (non-null timer), we dont want it to fire one last time after `shutdown`.

I agree, but that would only happen if we re-enter the event loop.

omerfirmak:
> 
> 
> > If it is enabled (non-null timer), we dont want it to fire one last time after `shutdown`.
> 
> I agree, but that would only happen if we re-enter the event loop.

Then why call it for the periodic one either? I think it's better being safe than sorry :D

source/agora/network/Client.d
```diff
@@ -168,7 +171,9 @@ public class NetworkClient
         this.exception = new Exception(
             format("Request failure to %s after %s attempts", address,
                 max_retries));
-        this.gossip_timer = this.taskman.setTimer(10.msecs, &this.gossipTask, Periodic.Yes);
+        // Create and stop timer immediately
+        this.gossip_timer = this.taskman.setTimer(GossipDelay, &this.gossipTask, Periodic.No);
+        this.gossip_timer.stop();
```
Geod24, line 176: https://github.com/vibe-d/vibe-core/blob/9ca874dc75020b1488a0ac64a1cc5f6471ba15f4/source/vibe/core/core.d#L1243-L1255

But okay, this works :D

source/agora/consensus/protocol/Nominator.d
```diff
@@ -189,12 +192,14 @@ extern(D):
         this.ledger = ledger;
         this.enroll_man = enroll_man;
         this.scp_envelope_store = new SCPEnvelopeStore(cacheDB);
+        // Create and stop timer immediately
+        this.envelope_timer = this.taskman.setTimer(10.msecs,
```
Geod24: `EnvTaskDelay` ?

Geod24 (approved):
LGTM. The `rearm` thing is dangerous, because adding a new method can go wrong very easily, but luckily this doesn't change much.

Geod24:
Did the minor change directly, merging

PR #2311, opened 2021-07-26
Fix 2294: Always provide block height stats

Geod24:
```
Our stats are backed by an AA / map because it is useful to have some stats
not showing by default (e.g. when we log status code).
However, in the case of block stats, we always want to return them,
so we can simplify the code and call 'collect' with the struct directly.
```

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -376,8 +376,6 @@ public class FullNode : API
         this.block_stats.increaseMetricBy!"agora_block_txs_total"(
             block.txs.length);
         this.block_stats.increaseMetricBy!"agora_block_externalized_total"(1);
-        this.block_stats.setMetricTo!"agora_block_height_counter"(
-            block.header.height.value);
     }
 
     /// Convenience function to increase an endpoint stats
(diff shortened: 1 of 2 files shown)


kchulj:
Is using `+=` better than / same as using `increaseMetricBy!`?

Geod24:
The type of the `block_stats` member changed. It used to be `Stats!(BlockStatsValue, NoLabel)` and now it is `BlockStatsValue` (but `BlockStatsValue` was renamed to `BlockStats` which was an alias for `Stats!(BlockStatsValue, NoLabel)`).

`BlockStatsValue` is just a struct with integer fields. So using `+=` is pretty much all you can do, it doesn't have a `increaseMetricBy` member. OTOH, the `Stats` templates provides an `increaseMetricBy` member: https://github.com/bosagora/agora/blob/de0d5a1ce701bb8645701c780f54f2848958ddd6/source/agora/stats/Stats.d#L55-L82

But the point of the `Stats` template is to support labels. Since we don't have any label, we do not need to use it.

kchulj:
Ok makes sense, nice simplification.

kchulj (approved):
✅

linked0 (approved):

PR #2317, opened 2021-07-27
Fix 1753 - Remove validator index maps

### LINKED ISSUE #1753: Excessive memory usage of validator index maps
No elements are ever removed from the following 2 maps and they will grow forever

```D
    /// used for setting the signature bitmask during signature collection
    private ulong[Point][Height] key_to_index;

    /// used for validating the signature
    private Point[ulong][Height] index_to_key;
```
This is not a problem right now, but if we will have 10.000 validators in the future, then 

32*8*2 - size of 2 pubkeys
6*24 - number of blocks per day

32*8*2*6*24*10000/1024/1024 = 703MB/day

Geod24:
Fixes https://github.com/bosagora/agora/issues/1753 and some issues with restart.
Prerequisite to https://github.com/bosagora/agora/pull/2315

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -932,6 +932,7 @@ extern(D):
         log.info("Externalized consensus data set at {}: {}", height, prettify(data));
         try
         {
+            const validators = this.ledger.getValidators(height);
             Hash random_seed = this.ledger.getRandomSeed(
                 height, data.missing_validators);
             Transaction[] externalized_tx_set;
@@ -944,7 +945,7 @@ extern(D):
             }
             const block = makeNewBlock(last_block,
                 externalized_tx_set, data.time_offset, random_seed,
-                this.enroll_man.getCountOfValidators(last_block.header.height + 1),
+                validators.length,
                 data.enrolls, data.missing_validators);
 
             // Now we add our signature and gossip to other nodes
--- a/source/agora/node/Ledger.d
+++ b/source/agora/node/Ledger.d
@@ -473,9 +473,6 @@ public class Ledger
 
         const Height next = block.header.height + 1;
         this.updateSlashedValidatorSet(block);
-        auto keys = this.getValidators(next).map!(vi => vi.address).array();
-        this.log.trace("Update validator lookup maps at height {}: {}", next, keys);
-        this.enroll_man.keymap.update(next, keys);
     }
 
     /***************************************************************************
@@ -1987,7 +1984,8 @@ unittest
         auto params = new immutable(ConsensusParams)(ValidatorCycle);
         const blocks = genBlocksToIndex(ValidatorCycle, params);
         // Enrollment: Insufficient number of active validators
-        assertThrown!Exception(new TestLedger(WK.Keys.A, blocks, params));
+        auto ledger = new TestLedger(WK.Keys.A, blocks, params);
+        assertThrown(ledger.getValidators(Height(ValidatorCycle + 1)));
     }
 }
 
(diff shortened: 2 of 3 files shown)


source/agora/node/Validator.d
```diff
@@ -302,11 +302,11 @@ public class Validator : FullNode, API
         auto signed_validators = BitMask(block.header.validators.count);
         signed_validators.copyFrom(block.header.validators);
 
-        auto node_validator_index = this.nominator.enroll_man
-            .getIndexOfValidator(block.header.height, this.config.validator.key_pair.address);
-
+        const self_utxo = this.enroll_man.getEnrollmentKey();
+        auto validators = this.ledger.getValidators(block.header.height);
+        const ptrdiff_t node_validator_index = validators.countUntil!(v => v.utxo() == self_utxo);
```
hewison-chris: First time I've seen this one. Looks useful.

hewison-chris (approved):
LGTM

Geod24:
Something seems wrong here, I might need to break this down further.

source/agora/node/Validator.d
Geod24: I wish it was named `indexOf` though...

source/agora/consensus/protocol/Nominator.d
```diff
@@ -944,7 +945,7 @@ extern(D):
             }
             const block = makeNewBlock(last_block,
                 externalized_tx_set, data.time_offset, random_seed,
-                this.enroll_man.getCountOfValidators(last_block.header.height + 1),
+                validators.length - data.missing_validators,
```
Geod24: I think this might have been the issue. `makeNewBlock` expects the count of validators to be the number of validators that can sign the block, but `validators.length` is the number of validators *active* (regardless of pre-image revelation) at this height.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -945,7 +945,7 @@ extern(D):
             }
             const block = makeNewBlock(last_block,
                 externalized_tx_set, data.time_offset, random_seed,
-                validators.length,
+                validators.length - data.missing_validators.length,
```
hewison-chris: I think this is wrong as the number of validators set in the BitMask in the block header is not excluding the missing validators.
Sprint #46 - Issues 1669 + 2229 002

hewison-chris (changes requested):
I think the problem is somewhere else.

source/agora/consensus/protocol/Nominator.d
hewison-chris: I think the problem is that a `validator` can be slashed at the `height` of this block when it did not reveal a `preimage`. It then does not get included in the `getValidators` so the count is less than it should be. I think we should slash at next height.

Geod24:
Fixed, should be g2g

source/agora/consensus/protocol/Nominator.d
Geod24: Thanks for the help, indeed that was it.

hewison-chris (approved):
LGTM

source/agora/node/Ledger.d
```diff
@@ -1987,7 +1984,8 @@ unittest
         auto params = new immutable(ConsensusParams)(ValidatorCycle);
         const blocks = genBlocksToIndex(ValidatorCycle, params);
         // Enrollment: Insufficient number of active validators
-        assertThrown!Exception(new TestLedger(WK.Keys.A, blocks, params));
+        auto ledger = new TestLedger(WK.Keys.A, blocks, params);
+        assertThrown(ledger.getValidators(Height(ValidatorCycle)));
```
hewison-chris: Seems this does not throw from ci test results.

linked0:
```suggestion
        assertThrown(ledger.getValidators(Height(ValidatorCycle + 1)));
```

Geod24: Good catch, thanks!

PR #2321, opened 2021-07-27
getValidators: Document behavior and design

Geod24:
```
A common use case for getValidators requires to have the pre-images adjusted.
For example, when deciding which validators haven't revealed their pre-image
for the next block. However, providing a mean to adjust them directly in the
method, while possible, would be counter-productive and dangerous as the index
would then be incorrect.
```


hewison-chris (approved):
LGTM

PR #2325, opened 2021-07-28
Nominator: Print quorum set in log messages

Geod24:
Previously, we were never looking up the quorum set because we always
used the default argument (null) to the struct.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -606,7 +606,7 @@ extern(D):
             return;
         }
 
-        log.trace("Received signed envelope: {}", scpPrettify(&envelope));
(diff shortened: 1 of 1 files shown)


Geod24:
Oops, I forgot that this couldn't work because of differences in hash types, will come back to this later...

hewison-chris (changes requested):
Can we include in the SCPPrettify unit tests?

Geod24:
We do already

hewison-chris:
Then why was there no change for expected output?

Geod24:
Because the tests in SCPPrettify test the correct behavior of `SCPPrettify`, including that the quorum is correctly printed with and without the delegate. So, we know the formatting is correct. The problem here is that we always called `SCPPrettify` without the delegate (all the changes are in the nominator).

PR #2326, opened 2021-07-28
Use `getValidators` when validating & generating slashing data

Geod24:
Instead of getting all the UTXOs, then checking manually if they have revealed a pre-image,
use getValidators and check the height.


Geod24:
Added a commit to remove usage of `hasRevealedPreimage`.

hewison-chris (approved):
Nice

kchulj (approved):

PR #2328, opened 2021-07-28
Introduce a module to throw an exception with a formatted message

Geod24:
It's common to need to format a string while throwing an exception.
But there are two downsides to this:
- The code is more verbose, as an import of `std.format` is needed,
  and the call becomes `throw new Exception(format("...", args))`;
- Creating the `Exception` itself allocates with the GC;
- The formatting allocates too;

For this reason, the `enforce` method provided by `agora.common.Enforce`
is statically allocated when the thread starts, and has a buffer for
allocation. This ensures that we will not allocate (even if the code
is not annotated as `@nogc`) and thus we can avoid triggerring GC
collections on throwing Exceptions.

### DIFF
--- a/source/agora/common/Amount.d
+++ b/source/agora/common/Amount.d
@@ -27,6 +27,8 @@
 
 module agora.common.Amount;
 
+import agora.common.Ensure;
+
 import std.format;
 
 /// Defines a monetary type used in the blockchain
@@ -90,12 +92,11 @@ public struct Amount
     }
 
     /// Support for Vibe.d deserialization
-    public static Amount fromString (scope const(char)[] str) pure @safe
+    public static Amount fromString (scope const(char)[] str) @safe
     {
         import std.conv : to;
         immutable ul = str.to!ulong;
-        if (!Amount.isInRange(ul))
-            throw new Exception("Invalid input value to Amount");
+        ensure(Amount.isInRange(ul), "Invalid input value to Amount: {}", str);
         return Amount(ul, true);
     }
 
@@ -104,8 +105,7 @@ public struct Amount
     public ref Amount opOpAssign (string op : "+") (in Amount other) return
         @safe
     {
-        if (!this.add(other))
-            throw new Exception(format("Amount %d cannot be added to %d", other, this));
+        ensure(this.add(other), "Amount {} cannot be added to {}", other, this);
         return this;
     }
 
@@ -114,8 +114,7 @@ public struct Amount
     public ref Amount opOpAssign (string op : "-") (in Amount other) return
         @safe
     {
-        if (!this.sub(other))
-            throw new Exception(format("Amount %d cannot be subtracted to %d", other, this));
+        ensure(this.sub(other), "Amount {} cannot be subtracted from {}", other, this);
         return this;
     }
 
@@ -124,8 +123,7 @@ public struct Amount
         @safe
     {
         Amount copy = this;
-        if (!copy.add(other))
-            throw new Exception(format("Amount %d cannot be added to %d", other, this));
+        ensure(copy.add(other), "Amount {} cannot be added to {}", other, this);
         return copy;
     }
 
@@ -134,8 +132,7 @@ public struct Amount
         @safe
     {
         Amount copy = this;
-        if (!copy.sub(other))
-            throw new Exception(format("Amount %d cannot be subtracted to %d", other, this));
+        ensure(copy.sub(other), "Amount {} cannot be subtracted from {}", other, this);
         return copy;
     }
 
@@ -466,7 +463,7 @@ nothrow pure @nogc @safe unittest
     assert(val == Amount.UnitPerCoin);
 }
 
-pure @safe unittest
+@safe unittest
 {
     import std.exception;
     assert(Amount.fromString(`5000000000000000`) == Amount.MaxUnitSupply);
--- a/source/agora/common/Ensure.d
+++ b/source/agora/common/Ensure.d
@@ -1,0 +1,60 @@
+/*******************************************************************************
+
+     Throw a statically allocated exception with formatting capacility.
+
+     The function in this module are not `@nogc` / `pure` but they do not
+     allocate if the formatting does not allocate.
+     A static buffer is used by the Exception, so the output length is limited.
+     Do not try to print a full block, for example.
+
+    Copyright:
+        Copyright (c) 2019-2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.Ensure;
+
+import ocean.text.convert.Formatter;
+
+/// Module ctor to initialize the thread-local instance
+static this ()
+{
+    instance = new FormattedException();
+}
+
+/// The underlying instance
+private FormattedException instance;
+
+/// An exception which doesn't allocate but allows formatting a message
+private final class FormattedException : Exception
+{
+    /// Internal buffer, should be enough for most messages
+    private char[2048] buffer;
+    /// Used size
+    private size_t end;
+
+    private this () @safe pure nothrow @nogc
+    {
+        super("An Exception occured");
+    }
+
+    public override const(char)[] message () const return
+        @safe pure nothrow @nogc
+    {
+        return this.buffer[0 .. this.end];
+    }
+}
+
+public void ensure (Args...) (bool exp, string fmt, lazy Args args)
+{
+    if (!exp)
+    {
+        auto res = instance.buffer.snformat(fmt, args);
+        instance.end = res.length;
+        throw instance;
+    }
+}
--- a/source/agora/crypto/Bech32.d
+++ b/source/agora/crypto/Bech32.d
@@ -35,9 +35,9 @@
 
 module agora.crypto.Bech32;
 
+import agora.common.Ensure;
 import agora.crypto.ECC;
 
-import std.format;
 import std.string;
 import std.uni;
 
@@ -134,15 +134,15 @@ public DecodeResult decodeBech32 (in char[] str)
         ubyte c = str[i];
         if (c >= 'a' && c <= 'z') lower = true;
         else if (c >= 'A' && c <= 'Z') upper = true;
-        else if (c < 33 || c > 126)
-            throw new Exception(format("Character '%X' at pos %s outside of valid char range", c, i));
+        else
+            ensure(c >= '!' && c <= '~',
+                    "Character '{X}' at pos {} is outside of valid char range", c, i);
     }
-    if (lower && upper)
-        throw new Exception("Bech32 does not allow mixed lower and upper cases");
+    ensure(lower ^ upper, "Bech32 does not allow mixed lower and upper cases");
 
     auto pos = lastIndexOf(str, '1');
-    if (str.length > 90 || pos == -1 || pos == 0 || pos + 7 > str.length)
-        throw new Exception(format("Invalid HRP for Bech32: %s (pos: %s)", str, pos));
+    ensure(str.length <= 90 && pos != -1 && pos != 0 && pos + 7 <= str.length,
+            "Invalid HRP for Bech32: {} (pos: {})", str, pos);
 
     ubyte[] values;
     values.length = str.length - 1 - pos;
@@ -151,10 +151,8 @@ public DecodeResult decodeBech32 (in char[] str)
         ubyte c = str[i + pos + 1];
         byte rev = CHARSET_REV[c];
 
-        if (rev == -1)
-            throw new Exception(
-                format("Invalid byte (%X) in data at position %s: %s",
-                       c, i + pos + 1, str));
+        ensure(rev != -1,
+                "Invalid byte '{X}' in data at position {}: {}", c, i + pos + 1, str);
 
         values[i] = rev;
     }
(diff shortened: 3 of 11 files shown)


source/agora/common/Amount.d
```diff
@@ -114,8 +114,7 @@ public struct Amount
     public ref Amount opOpAssign (string op : "-") (in Amount other) return
         @safe
     {
-        if (!this.sub(other))
-            throw new Exception(format("Amount %d cannot be subtracted to %d", other, this));
+        enforce(this.sub(other), "Amount {} cannot be subtracted to {}", other, this);
```
Geod24:
```suggestion
        enforce(this.sub(other), "Amount {} cannot be subtracted from {}", other, this);
```

I think (?)

source/agora/common/Amount.d
```diff
@@ -134,8 +132,7 @@ public struct Amount
         @safe
     {
         Amount copy = this;
-        if (!copy.sub(other))
-            throw new Exception(format("Amount %d cannot be subtracted to %d", other, this));
+        enforce(copy.sub(other), "Amount {} cannot be subtracted to {}", other, this);
```
Geod24: `from` ?

source/agora/common/Amount.d
```diff
@@ -90,12 +92,11 @@ public struct Amount
     }
 
     /// Support for Vibe.d deserialization
-    public static Amount fromString (scope const(char)[] str) pure @safe
+    public static Amount fromString (scope const(char)[] str) @safe
```
Geod24, line 95: Since `enforce` access a TLS global, we can't be `pure`.

source/agora/common/Amount.d
```diff
@@ -466,7 +463,7 @@ nothrow pure @nogc @safe unittest
     assert(val == Amount.UnitPerCoin);
 }
 
-pure @safe unittest
+@safe unittest
```
Geod24, line 466: Since `fromString` isn't `pure` anymore

source/agora/common/Enforce.d
```diff
@@ -0,0 +1,60 @@
...
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.common.Enforce;
+
+import ocean.text.convert.Formatter;
+
+/// Module ctor to initialize the thread-local instances
```
Geod24:
```suggestion
/// Module ctor to initialize the thread-local instance
```

source/agora/common/Enforce.d
```diff
@@ -0,0 +1,60 @@
...
+
+public void enforce (Args...) (bool exp, string fmt, lazy Args args)
+{
+    if (!exp)
+    {
+        auto res = instance.buffer.snformat(fmt, args);
+        instance.end = res.length;
+        throw instance;
+    }
+}
```
Geod24: Need to add tests for formatting and lazyness (and lack of allocation).
Could perhaps re-used Ocean ?

source/agora/crypto/Bech32.d
```diff
@@ -134,15 +134,15 @@ public DecodeResult decodeBech32 (in char[] str)
         ubyte c = str[i];
         if (c >= 'a' && c <= 'z') lower = true;
         else if (c >= 'A' && c <= 'Z') upper = true;
-        else if (c < 33 || c > 126)
-            throw new Exception(format("Character '%X' at pos %s outside of valid char range", c, i));
+        else
+            enforce(c >= '!' && c <= '~',
```
Geod24: Slight modification to make it more readable

source/agora/flash/Channel.d
```diff
@@ -326,10 +327,9 @@ public class Channel
 
         scope DeserializeDg dg = (size) @safe
         {
-            if (size > data.length)
-                throw new Exception(
-                    format("Requested %d bytes but only %d bytes available",
-                        size, data.length));
+            enforce(size <= data.length,
+                   "Requested {} bytes but only {} bytes available",
+                   size, data.length);
```
Geod24: This code is repeated in a few place, but I think we have a dedicated overload for this now ?

source/agora/node/BlockStorage.d
```diff
@@ -932,8 +924,9 @@ public class MemBlockStorage : IBlockStorage
 
     public void saveBlock (const ref Block block) @safe
     {
-        if (this.blocks.length != block.header.height)
-            throw new Exception("BlockStorage: Expected blocks in serial order");
+        enforce(this.blocks.length == block.header.height,
+                "BlockStorage: Expected blocks in serial order (known: {}, got: {})",
+                this.blocks.length, block.header.height,);
```
Geod24: I took the liberty to add some information to those, since I went with `git grep -n 'throw new'`

source/agora/node/BlockStorage.d
```diff
@@ -1001,8 +992,7 @@ public class MemBlockStorage : IBlockStorage
         ubyte[Hash.sizeof] hash_bytes = hash[];
 
         auto finds = this.hash_idx[].find!((a, b) => a.hash == b)(hash_bytes);
-        if (finds.empty)
-            throw new Exception(format("Couldn't find hash %s in chain: %s", hash));
```
Geod24, line 1005: This was a bug (orphan format specifier)

source/agora/node/Ledger.d
```diff
@@ -51,7 +52,7 @@ import agora.utils.PrettyPrinter;
 import std.algorithm;
 import std.algorithm.searching : maxElement;
 import std.conv : to;
-import std.exception;
+import std.exception : assumeUnique, assumeWontThrow;
```
Geod24, line 55: Need to do those changes because `std.exception` has `enforce` and so the names conflict.
Perhaps I should just pick another name ? But `enforce` seems like the best fit.

source/agora/crypto/Bech32.d
hewison-chris: Maybe even better to create constants that define these as boundaries (which I am assuming they are).

source/agora/node/Ledger.d
hewison-chris: `ensure` ?

source/agora/crypto/Bech32.d
Geod24: Well it's `man ascii`

source/agora/node/Ledger.d
Geod24: Good idea

Geod24:
Rebased, renamed to `ensure`.

Geod24:
Rebased + fixed a test that was relying on `msg`.
Come to think of it, a few places rely on it, so I'll do a second pass.

Geod24:
Ready to go

source/agora/common/Ensure.d
```diff
@@ -0,0 +1,60 @@
...
+}
+
+/// The underlying instance
+private FormattedException instance;
+
+/// An exception which doesn't allocate but allows formatting a message
+private final class FormattedException : Exception
+{
+    /// Internal buffer, should be enough for most messages
+    private char[2048] buffer;
```
hewison-chris, line 36: What happens when message is too long? Would be good if it just truncates.

source/agora/common/Ensure.d
```diff
@@ -0,0 +1,60 @@
...
+        return this.buffer[0 .. this.end];
+    }
+}
+
+public void ensure (Args...) (bool exp, string fmt, lazy Args args)
+{
+    if (!exp)
+    {
+        auto res = instance.buffer.snformat(fmt, args);
+        instance.end = res.length;
```
hewison-chris, line 57: Maybe here can use min of `res.length` and `2048` ?

hewison-chris (commented):
Will continue to review. Just have a question to start with.

source/agora/common/Ensure.d
Geod24: Yes, that's what `snformat` does:
https://github.com/sociomantic-tsunami/ocean/blob/adb31c84baa2061d07aaa0cb7a7d14c3cc98309b/src/ocean/text/convert/Formatter.d#L182-L184

source/agora/common/Ensure.d
hewison-chris: Not needed as explained above `snformat` will truncate.

Geod24: `snformat` will never return more than the buffer size.
[Here is the source of `snformat`](https://github.com/sociomantic-tsunami/ocean/blob/adb31c84baa2061d07aaa0cb7a7d14c3cc98309b/src/ocean/text/convert/Formatter.d#L200-L215)

Annotated below:
```D
public mstring snformat (Args...) (mstring buffer, cstring fmt, Args args)
{
    size_t start; // Used to keep track of where we are in `buffer`

    // The sink passed to `sformat`, same technique as our serialization library
    scope FormatterSink sink = (cstring s)
    {
        // 0 <= left < buffer.length. If we have filled the buffer, `left` will be 0.
        size_t left = buffer.length - start;
        // Likewise, 0 <= wsize <= s.length
        // If left is 0, `wsize` will be 0 (wsize stands for "write size")
        size_t wsize = left <= s.length ? left : s.length;
        if (wsize > 0)
            buffer[start .. start + wsize] = s[0 .. wsize];
        start += wsize; // If the buffer is full, it means start must be == buffer.length, and we just add 0 to it
    };

    sformat(sink, fmt, args);
     // Here you can see that `min` is not needed since we would get a bounds check error if `start > buffer.length`
    return buffer[0 .. start];
}
```

source/agora/flash/UpdateSigner.d
```diff
@@ -121,10 +122,9 @@ public class UpdateSigner
 
         scope DeserializeDg dg = (size) @safe
         {
-            if (size > data.length)
-                throw new Exception(
-                    format("Requested %d bytes but only %d bytes available",
-                        size, data.length));
+            ensure(size >= data.length,
```
hewison-chris:
```suggestion
            ensure(size <= data.length,
```

hewison-chris (changes requested):

source/agora/flash/UpdateSigner.d
Geod24: Oops

source/agora/node/BlockStorage.d
```diff
@@ -605,8 +600,7 @@ public class BlockStorage : IBlockStorage
             foreach (idx, ref b; data)
             {
                 const size_t pos = (from + idx);
-                if (!this.map(pos / DataSize))
-                    throw new Exception(format("Unabled to map data at position %d", pos));
+                ensure(this.map(pos / DataSize), "Unabled to map data at position %d", pos);
```
hewison-chris:
```suggestion
                ensure(this.map(pos / DataSize), "Unable to map data at position %d", pos);
```

source/agora/node/BlockStorage.d
```diff
@@ -932,8 +924,9 @@ public class MemBlockStorage : IBlockStorage
 
     public void saveBlock (const ref Block block) @safe
     {
-        if (this.blocks.length != block.header.height)
-            throw new Exception("BlockStorage: Expected blocks in serial order");
+        ensure(this.blocks.length == block.header.height,
+                "BlockStorage: Expected blocks in serial order (known: {}, got: {})",
```
hewison-chris:
```suggestion
                "BlockStorage: Expected block at height {} not at height {}",
```

source/agora/flash/UpdateSigner.d
Geod24: Fixed

source/agora/node/BlockStorage.d
```diff
@@ -961,14 +954,16 @@ public class MemBlockStorage : IBlockStorage
     public override void updateBlockSig (in Height height, in Hash hash,
         in Signature sig, in BitMask validators) @safe
     {
-        if (this.blocks.length < height)
-            throw new Exception("No such block");
+        ensure(this.blocks.length >= height,
+                "Signature update: No block at height {} ({})", height, this.blocks.length);
```
hewison-chris:
```suggestion
                "Signature update: Can not update block at height {} as current height is {}.", height, this.blocks.length);
```

hewison-chris (approved):
LGTM, just a few more reword suggestions.

source/agora/node/BlockStorage.d
Geod24: "Can not update block signature at height [...]", dropping the "Signature update" ? Otherwise the "update" is repeated twice.

hewison-chris: Yes even better.

source/agora/node/BlockStorage.d
Geod24: The reason "serial order" is here is to inform the user not only of what is expected, but most importantly, why it is expected. Here it's because we require to `saveBlock` height `N` before we `saveBlock` height `N + 1`.
I agree that the terms are not ideal though, so maybe there's a better way to carry this to the user ?

hewison-chris:
```suggestion
                "BlockStorage: Expected blocks in serial order (next height is {} not {}",
```

source/agora/node/BlockStorage.d
Geod24: Also `{}` instead of `%d`

Geod24:
Fixed messages

hewison-chris (approved):
LGTM

Geod24:
Thanks for the review!

PR #2329, opened 2021-07-28
NoPreImageVN: Don't rely on the caller setting the parameter

Geod24:
In case the caller does not set the enroll_keys parameter,
we will reveal our own pre-image regardless of the boolean.


Geod24:
👋

hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

PR #2332, opened 2021-07-28
Trivial: Remove single-use EM.removeAllValidators

Geod24:
There is no use for this function but to replay the entire blockchain,
so just inline the call inside the Ledger.


kchulj (approved):
LGTM

PR #2335, opened 2021-07-29
Test: Take Logger configuration into account

Geod24:
```
When a test fail, the logs printed to stdout are extremely verbose,
as they include all the trace calls.
This is not the case for the main thread,
which only prints the Error level,
or what we see in production, which uses Info by default.
The reason we were seeing everything is that the configuration
is done in agora.node.Runner, which is not used in unittests.
By slightly duplicating code, we now only print Info level by default,
and an API manager can use a custom log level to get a more verbose output.
We can also, in the future, add a field to configure the default log level
in TestConf, allowing to tailor the logs to a given test.
```


hewison-chris (approved):
LGTM.

kchulj (approved):
LGTM

PR #2336, opened 2021-07-29
Nominator: Return the correct enum value for MaybeValid values

Geod24:
```
This error message floods the log, because it is logged at Error level.
It is currently only returned if there are unknown txs (nominated txs we don't know about).
In this case, the node shouldn't outright reject the nomination,
but return the associated SCP value (`kMaybeValid`).
```

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -902,6 +902,9 @@ extern(D):
 
         if (auto fail_reason = this.ledger.validateConsensusData(data, this.initial_missing_validators))
         {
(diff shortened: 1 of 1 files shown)


omerfirmak:
IIRC, this was intentional.

https://github.com/bosagora/agora/pull/1495#discussion_r558036529

Geod24:
Thanks for the link. Will dive into the code to see how to best handle it. But at the very least we should silence this message.

omerfirmak:
#2748 addresses the original motivation for this PR, which was the missing TX messages flooding the logs.

PR #2337, opened 2021-07-29
Remove some spurious or misleading log messages

Geod24:
Context: Trying to debug any kind of issue in the network tests makes me hate life.
I took a good look at the logs, to try to make sense of it, but there is a lot of noise in it.
The first step was to not print the `Trace` messages unless configured (#2335).
The most common spurious message is addressed in #2336 although that will require more work, so it is in its own PR.
This PR addresses two common messages we're seeing:
- A simple `combineCandidate` with the slot index / height at which it is called;
   Removing it because it always appear after a "Nominating" message, and it should never have been `Info`;
- We get a lot of "Invalid pre-image data" messages, which are just about the height.
   Since the situation can happen in a normal, functioning system, it shouldn't be logged at `Info` level,
   especially since it's harmless. Ideally we should avoid the flood of network communication, but it's another issue.

For example, look at the following screenshot: 2/3rd of the space is taken by useless messages.
Screen Shot 2021-07-29 at 15 55 40

And this is the `combineCandidate` message:
Screen Shot 2021-07-29 at 15 58 09

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -1181,8 +1181,6 @@ extern(D):
     public override ValueWrapperPtr combineCandidates (uint64_t slot_idx,
         ref const(ValueWrapperPtrSet) candidates)
     {
-        log.info("combineCandidates: {}", slot_idx);
-
         try
         {
             CandidateHolder[] candidate_holders;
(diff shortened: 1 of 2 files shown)


hewison-chris (approved):
LGTM

kchulj (approved):
🆗

PR #2339, opened 2021-07-29
Tests: Remove unused TestNominator

Geod24:
This is a relic from the times before stateDB and cacheDB were introduced,
but it is not used anymore and can be removed, along with 'test_start_time'.


hewison-chris (approved):
LGTM

PR #2341, opened 2021-07-30
Trivial: Fix up style issues in TestConf

Geod24:
- Add explicit 'public' visibility;
- Put the standalone field first and document their origin;
- Add a clear separation between the base config and the actual test config;

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -1911,22 +1911,41 @@ public enum NetworkTopology
 public struct TestConf
 {
     /// Network topology to use
-    NetworkTopology topology = NetworkTopology.FullyConnected;
+    public NetworkTopology topology = NetworkTopology.FullyConnected;
 
     /// Extra blocks to generate in addition to the genesis block
-    size_t extra_blocks = 0;
+    public size_t extra_blocks = 0;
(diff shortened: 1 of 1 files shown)


source/agora/test/Base.d
```diff
@@ -1911,22 +1911,41 @@ public enum NetworkTopology
...
-    /// Base values to use for the node configuration
-    NodeConfig node = {
+    /***************************************************************************
+
+        Base configuration for the nodes, matches `agora.common.Config : Config`
+
+    ***************************************************************************/
+
+    /// If the enrollments will be renewed or not at the end of the cycle
+    /// Matches the eponymous field in the `consensus` section.
```
hewison-chris, line 1935: I had to look up `eponymous` but see it means 'named' in this context.

hewison-chris (approved):
LGTM

source/agora/test/Base.d
Geod24: Precisely, "with the same name", e.g. https://en.wikipedia.org/wiki/Metallica#:~:text=Its%20eponymous%20fifth%20album%2C%20Metallica%20(1991)

PR #2342, opened 2021-07-30
Improve logging configuration in network integration tests

Geod24:
This allows to do two things:
- Set the default `LogLevel` for the whole process via environment variable;
- Override, for any logger, the configuration of any logger via `TestConf`;

The usage with `SCP` doesn't actually work now, but that seems to be an issue with the SCP logging.
Tested various combination `dloglevel` and override and it works like a charm.


hewison-chris (approved):
Nice

kchulj (approved):
LGTM.
I like how our logging is constantly improving.

kchulj (approved):
LGTM

PR #2347, opened 2021-08-02
[WIP] Replace public key node_id in SCP messages with an index

### LINKED ISSUE #1191: Replace public key node_id in SCP messages with an index
This is an optimization that is similar to https://github.com/bpfkorea/agora/issues/982.

SCP statements currently contain:
- node_id (64-byte key)
- slot_idx
- signature

However because we are using staking we have the opportunity to optimize this and reduce the network load.

`node_id` can be replaced with a 2-byte `ushort` index - which can encode up to `65535` validators (there is a theoretical limit of 1000 validators, might be relaxed later). The index would be used to look up the node's public key, e.g. `PublicKey[ushort]`. We can assume that for a given slot index there is a corresponding mapping between an index and the public key. We can sort the public keys to generate the index of each key.

The new SCPStatement would then compose of:
- ushort node_idx => maps to the PublicKey
- slot_idx
- signature

In addition to changing SCP code, this will require changes to the Nominator:
- When verifying statement signatures we must have an associated `node idx => public key` map to extract the public key from and thereafter verify the signature
- We might need to keep the last N slot index mappings. In other words, `slot idx => node idx => public key` for the last N slots - where N could be define later. The reason why this is needed is to allow late collection of block signatures. If for example slot 1 was externalized and slot 2 began nomination then we might receive a late schnorr block signature for block 1. In this case we need to be able to verify the SCPStatement's signature.

linked0:
This is for issue #1191 and is in the middle of implementation.

This PR is submitted for a whiteboard before completing.

Fix #1191

### DIFF
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -635,6 +635,56 @@ public class EnrollmentManager
 
     /***************************************************************************
 
+        Get the UTXO corresponding to the UTXO index for the height
+
+        Params:
+            height = the height of proposed block 
+            index = the index to find the UTXO with
+
+        Returns:
+            the `UTXO` corresponding to the index
+
+    ***************************************************************************/
+
+    public Hash getEnrolledUTXOByIndex (in Height height, in uint index)
+        nothrow @safe
+    {
+        Hash[] keys;
+        this.getEnrolledUTXOs(height, keys);
+        if (keys.length > index)
+            return keys[index];
+        else
+            return Hash.init;
+    }
+
+    /***************************************************************************
+
+        Get the index corresponding to the UTXO for the height
+
+        Params:
+            height = the height of proposed block
+            utxo = the UTXO to find the index with
+
+        Returns:
+            the index corresponding to the `UTXO`
+
+    ***************************************************************************/
+
+    public uint getEnrolledUTXOIndex (in Height height, in Hash utxo)
+        nothrow @safe
+    {
+        Hash[] keys;
+        this.getEnrolledUTXOs(height, keys);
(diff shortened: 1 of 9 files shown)


source/agora/consensus/EnrollmentManager.d
```diff
@@ -633,6 +633,56 @@ public class EnrollmentManager
...
+    public Hash getEnrolledUTXOByIndex (in Height height, in uint index)
+        nothrow @safe
+    {
+        Hash[] keys;
+        this.getEnrolledUTXOs(height, keys);
+        if (keys.length > index)
+            return keys[index];
+        else
+            return Hash.init;
+    }
```
Geod24, line 658: Using `Ledger.getValidators(Height)` will return a list of validators with the correct indexes.

source/scpd/types/Stellar_SCP.d
```diff
@@ -384,3 +384,35 @@ static assert(SCPQuorumSet.sizeof == 56);
...
+    static public SCPEnvelope fromSimpleSCPEnvelope (in SimpleSCPEnvelope env, NodeID node_id)
+        nothrow
+    {
+        auto state = SCPStatement(node_id, env.statement.slot_index,
+            cast()env.statement.pledges);
+        return SCPEnvelope(state, env.signature);
+    }
+}
+
+static assert(SimpleSCPEnvelope.sizeof == 232);
```
Geod24, line 418: I would just modify `SCPEnvelope`, because the C++ SCP code uses it an it'll create issues.

linked0:
Closing this PR as we decided to change the type of `NodeID` to `ulong` instead of introducing a new `struct` like `SimpleSCPStatement`.

PR #2352, opened 2021-08-02
Implement SigHash support in the execution engine

omerfirmak:
SigHash was introduced to allow changing
how a signature from an Input is verified.

This is a strict requirement in order to implement:
- SIGHASH_NOINPUT for Eltoo (BIP118)
- SIGHASH_SINGLE for setting the fees for the final
  settlement transaction in the Eltoo protocol.

The basic payment lock types (LockType.Key / LockType.KeyHash)
expect a <signature, sighash> tuple in the
unlock byte array.

For scripts, the Engine now expects a tuple of <signature, sighash>
on the stack when verifying signatures.

Co-authored-by: Andrej Mitrovic <andrej.mitrovich@gmail.com>

### DIFF
--- a/source/agora/script/Signature.d
+++ b/source/agora/script/Signature.d
@@ -1,0 +1,164 @@
+/*******************************************************************************
+
+    Contains the signature definitions and the challenge routine.
+
+    Copyright:
+        Copyright (c) 2020 BOS Platform Foundation Korea
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.script.Signature;
+
+import agora.crypto.Schnorr;
+import agora.crypto.Hash;
+import agora.common.Types;
+import agora.consensus.data.Transaction;
+
+import std.range;
+
+/// Used to select the behavior of the signature creation & validation algorithm
+public enum SigHash : ubyte
+{
+    /// default, signs the entire transaction
+    All = 1 << 0,
+}
+
+/// Contains the Signature and its associated SigHash
+public struct SigPair
+{
+align(1):
+    /// The signature (which also signs the sig_hash below)
+    public Signature signature;
+
+    /// Selects behavior of the signature creation & validation algorithm
+    public SigHash sig_hash;
+
+    /***************************************************************************
+
+        todo: unittest
+
+        Returns:
+            The signature + SigHash as a byte array embeddable in a script
+
+    ***************************************************************************/
+
+    public inout(ubyte)[] opSlice () inout pure nothrow @safe /*@nogc*/
+    {
+        return this.signature.toBlob()[] ~ ubyte(this.sig_hash);
+    }
+}
+
+/*******************************************************************************
+
+    Decodes the raw byte representation of a signature to its
+    `Signature` and `SigHash` parts, and validates the `SigHash`
+    to be one of the known flags or accepted combination of flags.
+
+    Params:
+        bytes = contains the <Signature, SigHash> tuple
+        sig_pair = will contain the signature tuple if the Signature was encoded
+                   correctly and the SigHash is one of the known flags or
+                   accepted combination of flags.
+
+    Returns:
+        null if the signature tuple was decoded correctly,
+        otherwise the string explaining the reason why if decoding failed
+
+*******************************************************************************/
+
+public string decodeSignature (const(ubyte)[] bytes,
+    out SigPair sig_pair) pure nothrow @safe @nogc
+{
+    if (bytes.length != SigPair.sizeof)
+        return "Encoded signature tuple is of the wrong size";
+
+    const sig = toSignature(bytes[0 .. Signature.sizeof]);
+    bytes.popFrontN(Signature.sizeof);
+
+    assert(bytes.length == 1);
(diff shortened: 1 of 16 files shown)


omerfirmak:
This will probably break other projects.

source/agora/script/Signature.d
```diff
@@ -0,0 +1,182 @@
...
+    bytes.popFrontN(Signature.sizeof);
+
+    assert(bytes.length == 1);
+    const SigHash sig_hash = cast(SigHash)bytes[0];
+    if (!isValidSigHash(sig_hash))
+        return "Unknown SigHash";
+
+    sig_pair = SigPair(sig, sig_hash);
+    return null;
+}
```
Geod24, line 92: Looks a lot like `fromBinary`.

source/agora/script/Signature.d
```diff
@@ -0,0 +1,182 @@
...
+            sig_hash = the SigHash
+
+    ***************************************************************************/
+
+    public this (Signature signature, SigHash sig_hash)
+        pure nothrow @safe @nogc
+    {
+        this.signature = signature;
+        this.sig_hash = sig_hash;
+    }
```
Geod24: You don't need to provide a ctor if all it does is to set the members, `struct` have default ctors.

source/agora/script/Signature.d
```diff
@@ -0,0 +1,182 @@
...
+    {
+    case SigHash.All:
+        break;
+
+    default:
+        return false;
+    }
+
+    return true;
+}
```
Geod24: This bothers me, because it means we can have a value that is typed as `SigHash` but is outside the bound of the `enum`. IMO that should not happen to begin with.

source/agora/script/Signature.d
```diff
@@ -0,0 +1,182 @@
...
+        the challenge as a hash
+
+*******************************************************************************/
+
+public Hash getChallenge (in Transaction tx, in SigHash sig_hash = SigHash.All,
+    in ulong input_idx = 0) nothrow @safe
+{
+    assert(input_idx < tx.inputs.length, "Input index is out of range");
+
+    switch (sig_hash)
```
Geod24: You could do a `final switch` here if you ensure that no random byte gets casted to `SigHash` :)

source/agora/script/Signature.d
```diff
@@ -0,0 +1,182 @@
...
+
+    Returns:
+        the challenge as a hash
+
+*******************************************************************************/
+
+public Hash getChallenge (in Transaction tx, in SigHash sig_hash = SigHash.All,
+    in ulong input_idx = 0) nothrow @safe
+{
+    assert(input_idx < tx.inputs.length, "Input index is out of range");
```
Geod24: Dead parameter ?

source/agora/script/Signature.d
```diff
@@ -162,6 +170,14 @@ public Hash getChallenge (in Transaction tx, in SigHash sig_hash = SigHash.All,
 
     switch (sig_hash)
     {
+    case SigHash.NoInput:  // eltoo support
+        Transaction dup;
+        // it's ok, we'll dupe the array before modification
+        () @trusted { dup = *cast(Transaction*)&tx; }();
+        dup.inputs = dup.inputs.dup;
+        dup.inputs[input_idx] = Input.init;  // blank out matching input
+        return hashMulti(dup, sig_hash);
```
Geod24: IMHO the proper way to do this would be via a simple wrapper struct, but we can do this later, since you just moved the code.

Geod24 (commented):
Very nice! I'll need a couple more pass, but this looks pretty good.
One thing I'd like to check, I believe you should be able to stuff a few bytes at the beginning of the signature, because of the range it has.

Geod24:
One point of style: If you require different parameters for different `enum` values, a single function with all parameters is ill suited.
Instead of doing:
```D
public Hash getChallenge (in Transaction tx, in SigHash sig_hash = SigHash.All,
    in ulong input_idx = 0) nothrow @safe;
```

Something like this:
```D
public Hash getChallenge (SigHash sh : SigHash.All) (in Transaction tx) nothrow @safe;
public Hash getChallenge (SigHash sh : SigHash.NoInput) (in Transaction tx, size_t input_index) nothrow @safe;
```

Would do the job. For the default value, `SigHash sh : SigHash.All = SigHash.All` might work, otherwise an empty template param `()` which forwards to `All` does the trick.
The advantage is that the `input_index` param is mandatory when `NoInput` is used.
If you need runtime dispatch, you can use the same trick [as the `Formatter`](https://github.com/sociomantic-tsunami/ocean/blob/adb31c84baa2061d07aaa0cb7a7d14c3cc98309b/src/ocean/text/convert/Formatter.d#L259-L286).

Geod24:
Regarding the template constraint trick, we already use it here: https://github.com/bosagora/agora/blob/f7f995ef4d8de3ac9eda67464e1bb458d9f51cad/source/agora/common/Amount.d#L122-L140

source/agora/script/Signature.d
omerfirmak: But with some validation and is `nothrow`.

Geod24 (approved):

PR #2353, opened 2021-08-03
Change the type of `NodeID` to `ulong`

### LINKED ISSUE #1191: Replace public key node_id in SCP messages with an index

linked0:
This contains the code changing the type of `NodeID` to `ulong` and removing quorum hashes from SCP protocol messages.

Fixes #1191, #982

### DIFF
--- a/source/agora/common/Config.d
+++ b/source/agora/common/Config.d
@@ -976,8 +976,7 @@ public SCPQuorumSet toSCPQuorumSet (in QuorumConfig quorum_conf) @safe nothrow
 
     foreach (ref const node; quorum_conf.nodes)
     {
-        auto pub_key = NodeID(node[][0 .. NodeID.sizeof]);
-        quorum.validators.push_back(pub_key);
+        quorum.validators.push_back(node);
     }
 
     foreach (ref const sub_quorum; quorum_conf.quorums)
@@ -1007,7 +1006,7 @@ public QuorumConfig toQuorumConfig (const ref SCPQuorumSet scp_quorum)
     import std.conv;
     import scpd.types.Stellar_types : NodeID;
 
-    Hash[] nodes;
+    ulong[] nodes;
 
     foreach (node; scp_quorum.validators.constIterator)
         nodes ~= node;
@@ -1031,16 +1030,9 @@ unittest
 {
     import agora.crypto.Hash;
 
-    auto quorum = QuorumConfig(2,
-        [Hash("0x11c6b0395c8e1716978c41958eab84e869755c09f7131b3bbdc882a647cb3f2c46c450607c6da71d34d1eab28fbfdf14376b444ef46ed1d0a7d2237ab430ebf5"),
-         Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd")],
-        [QuorumConfig(2,
-            [Hash("0x11c6b0395c8e1716978c41958eab84e869755c09f7131b3bbdc882a647cb3f2c46c450607c6da71d34d1eab28fbfdf14376b444ef46ed1d0a7d2237ab430ebf5"),
-             Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd")],
-            [QuorumConfig(2,
-                [Hash("0x11c6b0395c8e1716978c41958eab84e869755c09f7131b3bbdc882a647cb3f2c46c450607c6da71d34d1eab28fbfdf14376b444ef46ed1d0a7d2237ab430ebf5"),
-                 Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd"),
-                 Hash("0xdfcada320948a86f6027daf7e5a964a36103ea0e662abaa692212392a280b7c211e56beb2bf83fbc53459603c6750e00cdc194c773f9941dc43b07c6f639e5fd")])])]);
+    auto quorum = QuorumConfig(2, [0, 1, 2],
+        [QuorumConfig(2, [0, 2],
+            [QuorumConfig(2, [0, 1, 1])])]);
 
     auto scp_quorum = toSCPQuorumSet(quorum);
     assert(scp_quorum.toQuorumConfig() == quorum);
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -97,7 +97,7 @@ public extern (C++) class Nominator : SCPDriver
     protected ValidatingLedger ledger;
 
     /// The mapping of all known quorum sets
-    private SCPQuorumSetPtr[Hash] known_quorums;
+    private SCPQuorumSetPtr[NodeID] known_quorums;
 
     private alias TimerType = Slot.timerIDs;
     static assert(TimerType.max == 1);
@@ -272,38 +272,30 @@ extern(D):
 
     /***************************************************************************
 
-        Set or update the quorum configuration
-
-        The node additionally takes the list of all other validator's quorum
-        configurations and saves them for later lookup by getQSet().
+        Update our quorum configuration and store the mapping of all quorums
+        for lookup in getQSet().
 
         Params:
-            quorum = the quorum config for this node
-            other_quorums = the quorum config for all other nodes in the network
+            node_id = the node id of this validator
+            quorums = the mapping of all active validators' quorums
 
     ***************************************************************************/
 
-    public void setQuorumConfig (const ref QuorumConfig quorum,
-        const(QuorumConfig)[] other_quorums) nothrow @safe
+    public void setQuorumConfig (ref const(NodeID) node_id,
+        const(QuorumConfig)[NodeID] quorums) nothrow @safe
     {
         assert(!this.is_nominating);
         () @trusted { this.known_quorums.clear(); }();
 
-        // store the list of other node's quorum hashes
-        foreach (qc; other_quorums)
+        foreach (qpair; quorums.byKeyValue)  // opApply2 is not nothrow..
         {
-            auto quorum_set = buildSCPConfig(qc);
+            auto quorum_set = buildSCPConfig(qpair.value);
             auto shared_set = makeSharedSCPQuorumSet(quorum_set);
-            const hash = this.getHashOfQuorum(quorum_set);
-            this.known_quorums[hash] = shared_set;
-        }
+            this.known_quorums[qpair.key] = shared_set;
 
-        // set up our own quorum
-        auto quorum_set = buildSCPConfig(quorum);
-        () @trusted { this.scp.updateLocalQuorumSet(quorum_set); }();
-        auto shared_set = makeSharedSCPQuorumSet(quorum_set);
-        const hash = this.getHashOfQuorum(quorum_set);
-        this.known_quorums[hash] = shared_set;
+            if (qpair.key == node_id)
+                () @trusted { this.scp.updateLocalQuorumSet(quorum_set); }();
+        }
     }
 
     /***************************************************************************
@@ -1053,18 +1045,19 @@ extern(D):
     /***************************************************************************
 
         Params:
-            qSetHash = the hash of the quorum set
+            node_id = the node's id of the quorum set
 
         Returns:
-            the SCPQuorumSet pointer for the provided quorum set hash
+            the SCPQuorumSetPtr for the provided node ID, or null if not found
 
     ***************************************************************************/
 
-    public override SCPQuorumSetPtr getQSet (ref const(StellarHash) qSetHash)
+    public override SCPQuorumSetPtr getQSet (ref const(NodeID) node_id)
     {
-        if (auto scp_quroum = qSetHash in this.known_quorums)
-            return *scp_quroum;
+        if (auto quorum = node_id in this.known_quorums)
+            return *quorum;
 
+        log.error("getQSet SCPQuorumSetPtr.init, id: {}", node_id);
         return SCPQuorumSetPtr.init;
     }
 
@@ -1374,7 +1367,7 @@ private struct SCPStatementHash
 {
     // sanity check in case a new field gets added.
     // todo: use .tupleof tricks for a more reliable field layout change check
-    static assert(SCPNomination.sizeof == 112);
+    static assert(SCPNomination.sizeof == 48);
 
     /// instance pointer
     private const SCPStatement* st;
@@ -1419,7 +1412,6 @@ private struct SCPStatementHash
         const ref SCPStatement._pledges_t._prepare_t prep,
         scope HashDg dg) @safe pure nothrow @nogc
     {
-        hashPart(prep.quorumSetHash[], dg);
         hashPart(prep.ballot, dg);
 
         /// these two can legitimately be null in the protocol
@@ -1452,7 +1444,6 @@ private struct SCPStatementHash
         hashPart(conf.nPrepared, dg);
         hashPart(conf.nCommit, dg);
         hashPart(conf.nH, dg);
-        hashPart(conf.quorumSetHash[], dg);
     }
 
     /***************************************************************************
@@ -1471,7 +1462,6 @@ private struct SCPStatementHash
     {
         hashPart(ext.commit, dg);
         hashPart(ext.nH, dg);
-        hashPart(ext.commitQuorumSetHash[], dg);
     }
 
     /***************************************************************************
@@ -1487,7 +1477,6 @@ private struct SCPStatementHash
     public static void computeHash (const ref SCPNomination nom, scope HashDg dg)
         @safe pure nothrow @nogc
     {
-        hashPart(nom.quorumSetHash[], dg);
         hashPart(nom.votes[], dg);
         hashPart(nom.accepted[], dg);
     }
@@ -1541,38 +1530,38 @@ private struct SCPEnvelopeHash
     auto getStHash () @trusted { return SCPStatementHash(&st); }
 
     assert(getStHash().hashFull() == Hash.fromString(
-        "0x755a41baaba02e099e86224bec897428a1d985efa44aca8408ab17dd7682e272caf527707dfb02b1f665a3860f8f671e6ce651eaf6d6594126fbccac8dfbdac1"));
+        "0xf3f95c2bcdf182f3c4ab3aa22606983c90676d908bcd053110c560cface4b18ff36d4ab359e3a44332c23ff8280385b8bc22d44ee9ab98878ac947c17e61fe75"));
 
     prep.counter++;
     assert(getStHash().hashFull() == Hash.fromString(
-        "0x7dd0d0832ea5132ca96169bb76f958aba7fc131f2a7113a53fa152db8cb342a4cc853a7b69e2b2bffa0510a1d9665f3771d094315610747056923c455de1f523"));
+        "0xcd9c7719642dadd28f3a3fbbbcaca648f8bf35200041f39ed3c3c3ac7d8cc1c589754cff86063998a25c08bcf93ba6ad752df10b8447cfc5b4d43064debca92e"));
 
     prep_prime.counter++;
     assert(getStHash().hashFull() == Hash.fromString(
-        "0x16d39c2a56a3cf75b85c8f2dc4ad1329713ab21515384f65f4b4e1591a9b20771492ebf26595362807c8fc11ef0028d0fa8bcaeed3f25946d4902acc66bcf55e"));
+        "0xd876318b3559b52fe3dbd7fad5ba2756cfdd1729250c02d0b0b6fe2684eee520c9ce37d787027ef2ff3e79fec1ec73dd500b3ce9abdc724caa7bc45e91b26e9a"));
 
     () @trusted { st.pledges.prepare_.prepared = null; }();
     assert(getStHash().hashFull() == Hash.fromString(
-        "0xc49f187ceda67573afab46466b1e6e7207707174a9a0bf850ea79a5e87d76f56d5f0c13b56a73754d4337416d1322e2780420bb426b604c8db16e427d4fb5416"));
+        "0xca227179325dd0aacec18c20f26ee97d40be1fbe505d685cfb20f297bbf62bdc17fd0da45e6464aa37a24139882ed0cb1db0a3f67b9d63e8b70d726c6dacae21"));
 
     () @trusted { st.pledges.prepare_.preparedPrime = null; }();
     assert(getStHash().hashFull() == Hash.fromString(
-        "0x09e54206c86f7b1a6c796b7a6e56e2b30f465d24e223601e246cc19120d817ac867186f6e7c867f2dbaf7022e9eebc5e560de2fa9f2ab2aab2fdd221272caa5c"));
+        "0x85988fb0978100d2fc5da08c1e7cef35ad3ada741b81411d7b33d0d326454ceda82fd65166b2668416b70eb2fdd89bb1eb059b2831f1e576b5a553be210b95e5"));
 
     () @trusted { st.pledges.nominate_ = SCPNomination.init; }();
     st.pledges.type_ = SCPStatementType.SCP_ST_NOMINATE;
     assert(getStHash().hashFull() == Hash.fromString(
-        "0x104cfee4945c44bd75783bdb2c8ed2de55d6e2346d84430282c1710e6474042079b122e9a24a75d2f2b577cf36b7e1eadefc156c5b2562a06c3c8e2b16f02470"));
+        "0x39f437e1ceb4a6c8f70dae0791c29bd1adbf8f9e053137fc7a3ddd7454e51e47546a438f6da2d563c9bb4c38afa0c2af36340ad2d13efc99edee0d25ab815f7f"));
 
     () @trusted { st.pledges.confirm_ = SCPStatement._pledges_t._confirm_t.init; }();
     st.pledges.type_ = SCPStatementType.SCP_ST_CONFIRM;
     assert(getStHash().hashFull() == Hash.fromString(
-        "0x92a0d0bd78e0b74cfd0664bbb1c4542f492991f260d5916bf290c2929790cfc3875179fb3774de8c664af9b3d135e6edb70d48188874723a3a8e884bad37584c"));
+        "0x4e9785c61b6302695b3b8a8ccabdd3ff55b34c9b235cdb54e64c28a5afe2623791f8f33129e0d791ba0ea04161f402a3e9abf54f0269b2f78a3ad0e899037ac7"));
 
     () @trusted { st.pledges.externalize_ = SCPStatement._pledges_t._externalize_t.init; }();
     st.pledges.type_ = SCPStatementType.SCP_ST_EXTERNALIZE;
     assert(getStHash().hashFull() == Hash.fromString(
-        "0x98fec9d03f95de3fb6e5fefe5afb4aed545532cf1ea0440718a17451480ac4740f9b0389e613eefb1c33247d1ddc111343fd0f19ed26124abb5f8d9a9b15ccb9"));
+        "0x8eb8d71b06c7842a5c257a7475522331f04971805ed9b84c036e1c3e232b0e98ceffa015bf146716f28e1f0c20c5d068fca56c5f00431fd44a260cccc342167a"));
 
     SCPEnvelope env;
     auto getEnvHash () @trusted { return SCPEnvelopeHash(&env); }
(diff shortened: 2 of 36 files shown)


linked0:
This PR has some link errors that I need to deal with and I'd like to have a white-boarding about if necessary.

```
Linking...
  "bool cpp_set_empty<unsigned long>(void const*)", referenced from:
      __D4scpd3Cpp__T3setTmTSQuQr__T4lessTmZQiTSQBnQBl__T9allocatorTmZQnZQCa5emptyMxFNaNbNiZb in agora.o
  "void dtorCPPObject<std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > > >(std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > >*)", referenced from:
      __D4scpd3Cpp__T13unordered_mapTmTSQBf6quorum13QuorumTrackerQp8NodeInfoTSQCrQCp__T10RandHasherTmTSQDqQDo__T4hashTmZQiZQBkTSQEpQEn__T8equal_toTmZQmTSQFoQFm__T9allocatorTSQGjQGh__T4pairTxmTQFxZQnZQBlZQHa8__mixin46__dtorMFNbNiNeZv in agora.o
  "void dtorCPPObject<std::__1::map<unsigned long, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper>, std::__1::less<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper> > > > >(std::__1::map<unsigned long, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper>, std::__1::less<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper> > > >*)", referenced from:
      __D4scpd3Cpp__T3mapTmTSQuQr__T10shared_ptrTSQBp3scp9SCPDriver18SCPEnvelopeWrapperZQCaTSQDgQDe__T4lessTmZQiTSQEbQDz__T9allocatorTSQEwQEu__T4pairTxmTQEvZQnZQBlZQFn8__mixin26__dtorMFNbNiNeZv in agora.o
  "void dtorCPPObject<std::__1::set<unsigned long, std::__1::less<unsigned long>, std::__1::allocator<unsigned long> > >(std::__1::set<unsigned long, std::__1::less<unsigned long>, std::__1::allocator<unsigned long> >*)", referenced from:
      __D4scpd3Cpp__T3setTmTSQuQr__T4lessTmZQiTSQBnQBl__T9allocatorTmZQnZQCa8__mixin46__dtorMFNbNiNeZv in agora.o
  "int cpp_set_foreach<unsigned long>(void*, void*, void*)", referenced from:
      __D4scpd3Cpp__T3setTmTSQuQr__T4lessTmZQiTSQBnQBl__T9allocatorTmZQnZQCa7opApplyMxFMDFKxmZiZi in agora.o
  "void copyCtorCPPObject<std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > > >(std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > >*, std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > >*)", referenced from:
      __D4scpd3Cpp__T13unordered_mapTmTSQBf6quorum13QuorumTrackerQp8NodeInfoTSQCrQCp__T10RandHasherTmTSQDqQDo__T4hashTmZQiZQBkTSQEpQEn__T8equal_toTmZQmTSQFoQFm__T9allocatorTSQGjQGh__T4pairTxmTQFxZQnZQBlZQHa8__mixin46__ctorMFNbNcNiNeKNgSQItQIr__TQIqTmTQIeTQGwTQFcTQEhZQJmZSQKdQKb__TQKaTmTQJoTQIgTQGmTQFrZQKw in agora.o
      __D4scpd3Cpp__T13unordered_mapTmTSQBf6quorum13QuorumTrackerQp8NodeInfoTSQCrQCp__T10RandHasherTmTSQDqQDo__T4hashTmZQiZQBkTSQEpQEn__T8equal_toTmZQmTSQFoQFm__T9allocatorTSQGjQGh__T4pairTxmTQFxZQnZQBlZQHa8__mixin46__ctorMyFNbNcNiNeKNgSQIuQIs__TQIrTmTQIfTQGxTQFdTQEiZQJnZySQKfQKd__TQKcTmTQJqTQIiTQGoTQFtZQKy in agora.o
  "void copyCtorCPPObject<std::__1::map<unsigned long, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper>, std::__1::less<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper> > > > >(std::__1::map<unsigned long, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper>, std::__1::less<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper> > > >*, std::__1::map<unsigned long, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper>, std::__1::less<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper> > > >*)", referenced from:
      __D4scpd3Cpp__T3mapTmTSQuQr__T10shared_ptrTSQBp3scp9SCPDriver18SCPEnvelopeWrapperZQCaTSQDgQDe__T4lessTmZQiTSQEbQDz__T9allocatorTSQEwQEu__T4pairTxmTQEvZQnZQBlZQFn8__mixin26__ctorMFNbNcNiNeKNgSQHgQHe__TQHdTmTQHcTQEuTQEdZQHvZSQImQIk__TQIjTmTQIiTQGaTQFjZQJb in agora.o
      __D4scpd3Cpp__T3mapTmTSQuQr__T10shared_ptrTSQBp3scp9SCPDriver18SCPEnvelopeWrapperZQCaTSQDgQDe__T4lessTmZQiTSQEbQDz__T9allocatorTSQEwQEu__T4pairTxmTQEvZQnZQBlZQFn8__mixin26__ctorMyFNbNcNiNeKNgSQHhQHf__TQHeTmTQHdTQEvTQEeZQHwZySQIoQIm__TQIlTmTQIkTQGcTQFlZQJd in agora.o
  "void copyCtorCPPObject<std::__1::set<unsigned long, std::__1::less<unsigned long>, std::__1::allocator<unsigned long> > >(std::__1::set<unsigned long, std::__1::less<unsigned long>, std::__1::allocator<unsigned long> >*, std::__1::set<unsigned long, std::__1::less<unsigned long>, std::__1::allocator<unsigned long> >*)", referenced from:
      __D4scpd3Cpp__T3setTmTSQuQr__T4lessTmZQiTSQBnQBl__T9allocatorTmZQnZQCa8__mixin46__ctorMFNbNcNiNeKNgSQDtQDr__TQDqTmTQDpTQDaZQEeZSQEvQEt__TQEsTmTQErTQEcZQFg in agora.o
      __D4scpd3Cpp__T3setTmTSQuQr__T4lessTmZQiTSQBnQBl__T9allocatorTmZQnZQCa8__mixin46__ctorMyFNbNcNiNeKNgSQDuQDs__TQDrTmTQDqTQDbZQEfZySQExQEv__TQEuTmTQEtTQEeZQFi in agora.o
  "void defaultCtorCPPObject<std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > > >(std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > >*)", referenced from:
      __D4scpd3Cpp__T13unordered_mapTmTSQBf6quorum13QuorumTrackerQp8NodeInfoTSQCrQCp__T10RandHasherTmTSQDqQDo__T4hashTmZQiZQBkTSQEpQEn__T8equal_toTmZQmTSQFoQFm__T9allocatorTSQGjQGh__T4pairTxmTQFxZQnZQBlZQHa8__mixin46__ctorMFNbNcNiNeEQIqQIo7CppCtorZSQJgQJe__TQJdTmTQIrTQHjTQFpTQEuZQJz in agora.o
  "void defaultCtorCPPObject<std::__1::map<unsigned long, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper>, std::__1::less<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper> > > > >(std::__1::map<unsigned long, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper>, std::__1::less<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, std::__1::shared_ptr<stellar::SCPEnvelopeWrapper> > > >*)", referenced from:
      __D4scpd3Cpp__T3mapTmTSQuQr__T10shared_ptrTSQBp3scp9SCPDriver18SCPEnvelopeWrapperZQCaTSQDgQDe__T4lessTmZQiTSQEbQDz__T9allocatorTSQEwQEu__T4pairTxmTQEvZQnZQBlZQFn8__mixin26__ctorMFNbNcNiNeEQHdQHb7CppCtorZSQHtQHr__TQHqTmTQHpTQFhTQEqZQIi in agora.o
  "void defaultCtorCPPObject<std::__1::set<unsigned long, std::__1::less<unsigned long>, std::__1::allocator<unsigned long> > >(std::__1::set<unsigned long, std::__1::less<unsigned long>, std::__1::allocator<unsigned long> >*)", referenced from:
      __D4scpd3Cpp__T3setTmTSQuQr__T4lessTmZQiTSQBnQBl__T9allocatorTmZQnZQCa8__mixin46__ctorMFNbNcNiNeEQDqQDo7CppCtorZSQEgQEe__TQEdTmTQEcTQDnZQEr in agora.o
  "void cpp_unordered_map_assign<unsigned long, stellar::QuorumTracker::NodeInfo>(void*, unsigned long const&, stellar::QuorumTracker::NodeInfo const&)", referenced from:
      std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > >::opIndexAssign(stellar::QuorumTracker::NodeInfo const&, unsigned long) in agora.o
  "unsigned long cpp_unordered_map_length<unsigned long, stellar::QuorumTracker::NodeInfo>(void const*)", referenced from:
      std::__1::unordered_map<unsigned long, stellar::QuorumTracker::NodeInfo, stellar::RandHasher<unsigned long, std::__1::hash<unsigned long> >, std::__1::equal_to<unsigned long>, std::__1::allocator<std::__1::pair<unsigned long const, stellar::QuorumTracker::NodeInfo> > >::length() in agora.o
  "void push_back<unsigned long const, xdr::xvector<unsigned long, 4294967292u> >(xdr::xvector<unsigned long, 4294967292u>&, unsigned long const&)", referenced from:
      __D4scpd5types5Utils3dupFKxSQzQw11Stellar_SCP12SCPQuorumSetZSQCgQCeQBjQz in agora.o
  "void push_back<unsigned long, std::__1::vector<unsigned long, std::__1::allocator<unsigned long> > >(std::__1::vector<unsigned long, std::__1::allocator<unsigned long> >&, unsigned long&)", referenced from:
      __D4scpd3Cpp__T6vectorTmTSQxQu__T9allocatorTmZQnZQBi9push_backMFNaNbNiNeKmZv in agora.o
      __D4scpd3Cpp__T6vectorTmTSQxQu__T9allocatorTmZQnZQBi9push_backMFNaNbNiNemZv in agora.o
  "stellar::normalizeQSet(stellar::SCPQuorumSet&, unsigned long const*)", referenced from:
      __D5agora9consensus8protocol9NominatorQk14buildSCPConfigFNbNfIKSQCj6common5Types12QuorumConfigZ9__lambda3MFNbNiNeZv in agora.o
  "stellar::SCPDriver::toShortString(xdr::opaque_array<64u> const&) const", referenced from:
      vtable for stellar::SCPDriver in SCPDriver.o
  "stellar::SCPDriver::toStrKey(xdr::opaque_array<64u> const&, bool) const", referenced from:
      vtable for stellar::SCPDriver in SCPDriver.o
```

source/scpp/src/scp/SCPDriver.h
```diff
@@ -100,6 +100,14 @@ class SCPDriver
...
     virtual void emitEnvelope(SCPEnvelope const& envelope) = 0;
 
+    // `getNodeID` gets the UTXO hash for a node with the `node_index`
+    // at the height that is handed in as `slotIndex`.
+    virtual Hash getNodeID(uint64 slotIndex, uint64 nodeIndex) = 0;
+
+    // `getNodeIndex` gets the index of the UTXO hash for a node with
+    // the `node_id` at the height that is handed in as `slotIndex`.
+    virtual uint64 getNodeIndex(uint64 slotIndex, Hash nodeID) = 0;
+
```
Geod24: I don't see this being used ?

source/scpp/src/scp/SCP.h
```diff
@@ -25,7 +25,7 @@ class SCP
     SCPDriver& mDriver;
 
   public:
-    SCP(SCPDriver& driver, NodeID const& nodeID, bool isValidator,
+    SCP(SCPDriver& driver, Hash const& nodeID, bool isValidator,
```
Geod24: That line shouldn't be modified.

source/scpp/src/scp/SCP.cpp
```diff
@@ -18,7 +18,7 @@
 namespace stellar
 {
 
-SCP::SCP(SCPDriver& driver, NodeID const& nodeID, bool isValidator,
+SCP::SCP(SCPDriver& driver, Hash const& nodeID, bool isValidator,
```
Geod24: Ditto, shouldn't be modified

source/scpd/types/Stellar_types.d
```diff
@@ -44,6 +44,6 @@ unittest
     static import agora.crypto.Schnorr;
 
     static assert(agora.crypto.Hash.Hash.sizeof == Hash.sizeof);
-    static assert(agora.crypto.Hash.Hash.sizeof == NodeID.sizeof);
+    static assert(ulong.sizeof == NodeID.sizeof);
```
Geod24: You can remove the line, the other ones are to ensure binary compatibility, but since we use `ulong` directly this is not necessary.

source/scpp/src/scp/SCP.cpp
linked0: Oh, I'll revert it. There might be some mistakes because I'm in the middle of modifying.

source/scpd/scp/SCPDriver.d
```diff
@@ -140,8 +150,9 @@ nothrow:
     {
         try
         {
-            auto slice = Hash(pk).toString();
-            return sliceToStdString(slice.ptr, slice.length);
+            // auto slice = Hash(pk).toString();
+            // return sliceToStdString(slice.ptr, slice.length);
+            return std_string("NODEID");
```
Geod24: For now you can get away with:
```
auto slice = pk.to!string;
return sliceToStdString(slice.ptr, slice.length);
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -232,14 +232,16 @@ extern(D):
...
+        alias scpd.types.Stellar_types.Hash StellarHash;
         try
         {
             // TODO: We should apply the situation where this validator has
             // enrolled with another UTXO after its previous enrollment expired.
             auto self_enroll = this.enroll_man.getEnrollmentKey();
             if (this.scp is null && self_enroll != Hash.init)
             {
-                auto node_id = NodeID(self_enroll[][0 .. NodeID.sizeof]);
+                auto node_id = StellarHash(self_enroll[][0 .. StellarHash.sizeof]);
```
Geod24: That's not correct either

source/scpp/src/scp/SCPDriver.h
linked0: I made it for the case where a `NominationProtocol` could refer to a `NodeID` or an index. But will remove if unnecessary.

source/scpp/src/scp/SCP.h
linked0: Reverted.

source/scpd/types/Stellar_types.d
linked0: Removed

source/scpd/scp/SCPDriver.d
linked0: I'm getting away with it now.

source/agora/consensus/protocol/Nominator.d
linked0: Updated and will modify it more.

linked0:
Only one error remains which I will fix soon.
```
In file included from /Users/hyunjaelee/work/agora/source/scpp/src/quorum/QuorumIntersectionCheckerImpl.cpp:8:
/Users/hyunjaelee/work/agora/source/scpp/src/crypto/KeyUtils.h:25:32: error: reference to type 'const stellar::ByteSlice' could not bind to an lvalue of type 'const unsigned long long'
    return strKey::toStrKey(0, key).value;
                               ^~~
/Users/hyunjaelee/work/agora/source/scpp/src/crypto/KeyUtils.h:31:12: note: in instantiation of function template specialization 'stellar::KeyUtils::toStrKey<unsigned long long>' requested here
    return toStrKey(key).substr(0, 5);
           ^
/Users/hyunjaelee/work/agora/source/scpp/src/quorum/QuorumIntersectionCheckerImpl.cpp:697:31: note: in instantiation of function template specialization 'stellar::KeyUtils::toShortString<unsigned long long>' requested here
    return stellar::KeyUtils::toShortString(mBitNumPubKeys.at(node));
                              ^
/Users/hyunjaelee/work/agora/source/scpp/src/crypto/StrKey.h:27:52: note: passing argument to parameter 'bin' here
SecretValue toStrKey(uint8_t ver, ByteSlice const& bin);
                                                   ^
```

linked0:
I'm sorry but is there any simple way to covert a `unsigned long` to a `ByteSlice`? The previous error occurs in the code of `QuorumIntersectionCheckerImpl.cpp:697` as follows.
```
return stellar::KeyUtils::toShortString(mBitNumPubKeys.at(node));
```

linked0:
With the support of @Geod24 about the mismatch `NodeID` and `ByteSlice`(e6bb6ae7a719da9fdb43c91902b6aa794e34e54f), I can so much progress in this PR. But there's one more link error about the bridge functions defined in the `DUtils.cpp` file.
I tried to modify the file for the link error and fail to solve the error. Can't adding `CPPSIZEOFINST(uint512);` solve the error?
```
  "cppSizeOf(xdr::opaque_array<64u>&)", referenced from:
      __D4scpd5tests8SizeTest17__unittest_L26_C1FZv in agora-unittests.o
```

omerfirmak:
> Can't adding CPPSIZEOFINST(uint512); solve the error?

AFAICS, it should

source/scpp/src/scp/LocalNode.cpp
```diff
@@ -99,7 +99,7 @@ LocalNode::computeWeight(uint64 m, uint64 total, uint64 threshold)
 // if a validator is repeated multiple times its weight is only the
 // weight of the first occurrence
 uint64
-LocalNode::getNodeWeight(NodeID const& nodeID, SCPQuorumSet const& qset)
+LocalNode::getNodeWeight(NodeID const& nodeID, Hash const& nodeHash, SCPQuorumSet const& qset)
```
Geod24: Why ?

source/scpp/src/scp/QuorumSetUtils.cpp
```diff
@@ -134,14 +134,14 @@ namespace
 //      { t:1, { innerSet } } into innerSet
 
 void
-normalizeQSetSimplify(SCPQuorumSet& qSet, NodeID const* idToRemove)
+normalizeQSetSimplify(SCPQuorumSet& qSet, bool remove, NodeID const idToRemove)
```
Geod24: I don't see why those changes are necessary either ?

source/scpp/src/scp/LocalNode.cpp
linked0: I wanted the algorithm calculating the weight to be as it was, which uses the hash of UTXO.

Geod24: No it does not

linked0: I'll check it again.

Geod24: The parameter you added, `nodeHash`, is not used anywhere, is it ?

source/scpp/src/scp/QuorumSetUtils.cpp
linked0: Previously, the default value of `idToRemove` was `nullptr`, so it can be used to determine whether to remove the node or not. But we can not call this function in the situation where there's no node to be removed because the type `NodeID` is changed into `ulong`.

Geod24: Why not ? It passes a point to a `NodeID`, I don't see what's different here ?

linked0: Right, we don't need the new parameter, I'll check it out again.

source/scpp/src/scp/LocalNode.cpp
linked0: Removed the unnecessary parameter.

linked0:
The modification of `source/scpd/tests/GlueTypes.d ` should be restored. But without the modification, this PR has a link error. This could be solved in the middle of reviews, I think.
```
  "cppSizeOf(xdr::opaque_array<64u>&)", referenced from:
      __D4scpd5tests8SizeTest17__unittest_L26_C1FZv in agora-unittests.o
```

linked0:
I have completed applying the code that removes quorum hashes from SCP protocol messages from [the PR #1192](https://github.com/bosagora/agora/pull/1192), which solves issue #982. I'm working on refining all the codes in commits in this PR.

linked0:
Ready for review! :)

source/agora/common/Config.d
```diff
@@ -1007,7 +1006,7 @@ public QuorumConfig toQuorumConfig (const ref SCPQuorumSet scp_quorum)
     import std.conv;
     import scpd.types.Stellar_types : NodeID;
 
-    Hash[] nodes;
+    ulong[] nodes;
```
Geod24, line 1009: NodeID

source/agora/consensus/protocol/Nominator.d
```diff
@@ -228,24 +228,33 @@ extern(D):
         checks that the UTXO key for enrollment exists and new SCP should be
         created because of new enrollment of this node.
 
+        Params:
+            height = block height for creating SCP
+
     ***************************************************************************/
 
-    public void createSCPObject () nothrow @safe
+    public void createSCPObject (in Height height) nothrow @safe
```
Geod24: Probably needs a different name, like `updateSCPObject` ?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -1081,6 +1096,123 @@ extern(D):
...
+        Block block;
+        try
+        {
+            block = this.ledger.getBlocksFrom(Height(0))[0];
+        }
+        catch (Exception exc)
+        {
+            log.fatal("Exception happened with calling `getBlocksFrom` on `getNodeIndex: {}", exc);
+            assert(0);
+        }
```
Geod24: That will only work for initially enrolled validator, will it not ?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -1081,6 +1096,123 @@ extern(D):
...
+        if (slot_index == 0)
+        {
+            foreach (idx, enroll; block.header.enrollments)
+            {
+                if (enroll.utxo_key == node_hash)
+                    return idx;
+            }
+            log.fatal("Could not find a UTXO for this node at Genesis");
+            assert(0);
+        }
```
Geod24: The only call I can see is at L243 in validator, which means we will never call this with `slot_index == 0`.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -228,24 +228,33 @@ extern(D):
...
+    public void createSCPObject (in Height height) nothrow @safe
     {
+        import scpd.types.Stellar_types;
         try
         {
             // TODO: We should apply the situation where this validator has
             // enrolled with another UTXO after its previous enrollment expired.
-            auto self_enroll = this.enroll_man.getEnrollmentKey();
-            if (this.scp is null && self_enroll != Hash.init)
+            const node_id = this.getSelfUTXOIndex(height + 1);
```
Geod24: Looking at the call chain, I think the caller of `createSCPObject` already has all the infos needed.

Geod24 (changes requested):
Need to review the last commit, but first 3 LGTM, comments mostly on the 4th.
Last commit also includes an openssl downgrade which was committed accidentally I suppose.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -98,6 +98,7 @@ public extern (C++) class Nominator : SCPDriver
 
     /// The mapping of all known quorum sets
     private SCPQuorumSetPtr[Hash] known_quorums;
+    private SCPQuorumSetPtr[NodeID] all_quorums;
```
Geod24: Why the new member ? Isn't it replacing `known_quorums` ?

linked0: Reverted to `known_quorums`.

source/agora/consensus/protocol/Nominator.d
linked0: Updated.

source/agora/consensus/protocol/Nominator.d
linked0: This is also called at L160 in `Validator.d`.

source/agora/consensus/protocol/Nominator.d
linked0: I don't think we need `getSelfUTXOIndex` with your comment. I removed it.

source/agora/consensus/protocol/Nominator.d
linked0: Removed `getSelfUTXOIndex`.

source/agora/consensus/protocol/Nominator.d
linked0: Ditto

source/agora/common/Config.d
linked0: I'd like to remain as it is because there's some problem when sorting the node in Quorum.d.
```
quorum.nodes.sort;
```

linked0: I tried it several ways. But there are a couple of things to handle after changing that. So I think it would be OK to remain as it is.
The problems:
1. Serializer/Deserializer should be changed.
2. New code is needed to sort the array of `NodeID`.

source/scpp/extra/DSCPUtils.cpp
```diff
@@ -9,7 +9,7 @@
 using namespace xdr;
 using namespace stellar;
 
-SCP* createSCP(SCPDriver* driver, NodeID const& nodeID, bool isValidator, SCPQuorumSet const& qSetLocal)
+SCP* createSCP(SCPDriver* driver, NodeID nodeID, bool isValidator, SCPQuorumSet const& qSetLocal)
```
Geod24, line 12: Why the change ?

source/scpd/quorum/QuorumIntersectionChecker.d
```diff
@@ -19,6 +19,17 @@ import scpd.Cpp;
 import scpd.quorum.QuorumTracker;
 import scpd.types.Stellar_types;
 
+extern (C++, "_GLOBAL__N_1") {
```
Geod24, line 25: DDOC

source/agora/consensus/protocol/Nominator.d
```diff
@@ -1081,6 +1103,61 @@ extern(D):
...
+                b0 = this.ledger.getBlocksFrom(Height(0))[0];
+            }
+            catch (Exception exc)
+            {
+                log.error("Exception happened with getting block at requested height: {}", exc);
+            }
+            assert(b0.header.enrollments.length > index);
+            utxo = b0.header.enrollments[index].utxo_key;
+        }
+        else
```
Geod24: Why is `slot_index == 1` treated specially ?

Geod24 (commented):
Nice. I'll give it a shot soon, but besides the one comment about the branch, LGTM.

source/scpp/extra/DSCPUtils.cpp
linked0: It generated an error like this.
```
Linking...
  "createSCP(stellar::SCPDriver*, unsigned long long, bool, stellar::SCPQuorumSet const&)", referenced from:
      __D5agora9consensus8protocol9NominatorQk15updateSCPObjectMFNbNfISQCk6common5Types6HeightIE4core4stdc6config13__c_ulonglongZ9__lambda5MFNeZv in agora-unittests.o
```

source/agora/consensus/protocol/Nominator.d
linked0: Removed it.

source/scpd/quorum/QuorumIntersectionChecker.d
linked0: I'm sorry but I have no idea how I DDOC this. Could you let me know the right comment for that?

source/scpp/src/scp/BallotProtocol.cpp
```diff
@@ -1564,27 +1561,6 @@ BallotProtocol::hasPreparedBallot(SCPBallot const& ballot,
...
-    {
-    case SCP_ST_PREPARE:
-        h = st.pledges.prepare().quorumSetHash;
-        break;
-    case SCP_ST_CONFIRM:
-        h = st.pledges.confirm().quorumSetHash;
-        break;
-    case SCP_ST_EXTERNALIZE:
-        h = st.pledges.externalize().commitQuorumSetHash;
-        break;
```
Geod24, line 1581: Do you know why Stellar had this difference ?

Geod24 (approved):
LGTM, just one question.

source/scpd/quorum/QuorumIntersectionChecker.d
Geod24:
```
// Quorum intersection checker implementation
// This method was originally implemented in C++ but since the changes to `NodeID` had to be moved to D.
```

Something like this.

source/scpp/src/scp/BallotProtocol.cpp
linked0: For the line, I don't know the reason which I should analyze more about.

linked0: The `commitQuorumSetHash` is the same value as `quorumSetHash` of other types of SCPStatement. It's never changed as long as the quorum set remains the same. So it should be removed.

https://github.com/bosagora/agora/blob/bfe95e853eef28f8487d9a2fc8897b86d2852554/source/scpp/src/scp/BallotProtocol.cpp#L560-L576

PR #2357, opened 2021-08-03
Update Vibed

omerfirmak:
This allows `in` parameter in our REST interfaces


Geod24:
Wanna try adding a few `in` here and there too ?

source/agora/flash/Channel.d
```diff
@@ -148,7 +148,7 @@ public class Channel
         ManagedDatabase db)
     {
         this.flash_conf = flash_conf;
-        this.conf = conf;
+        this.conf = cast()conf;
```
Geod24: Hum, `in` means `scope`, so not a good idea. Also `cast`s are evil, why do we need this ?

omerfirmak: without the cast;

```
source/agora/flash/Channel.d(151,21): Error: cannot implicitly convert expression `conf` of type `const(ChannelConfig)` to `ChannelConfig
```

It can't be copied for some reason I guess.

Geod24: Yeah, let's leave it out of the `in` changes then

omerfirmak: Done

Geod24 (approved):

kchulj (approved):
LGTM

PR #2358, opened 2021-08-04
FullNode: Accept some parameters as 'in'

Geod24:


kchulj (approved):
✅

source/agora/cli/client/SendTxProcess.d
```diff
@@ -225,13 +226,13 @@ unittest
         /// Contains the transaction cache
         private Transaction[Hash] tx_cache;
 
-        public override void putTransaction (Transaction tx) @safe
+        public override void putTransaction (in Transaction tx) @safe
         {
-            this.tx_cache[hashFull(tx)] = tx;
+            this.tx_cache[hashFull(tx)] = tx.serializeFull().deserializeFull!Transaction;
```
linked0, line 231: Why do we serialize and deserialize the transaction?

kchulj: Good question..

Geod24: To deep copy it. That's the only way to get a mutable instance from a const one. And that's the only place where we need one.

linked0 (approved):

Issue #267, opened 2019-08-23
Implement auto-update software for Agora


bpalaggi:
As a node operator, I need a software that keeps an automatically updated agora node on my server in order to have seamless integration with the chain.

**Definition of Done:**
- ~[ ] The auto-updater is separate from Agora~
- [ ] The auto-updater is the default package pointed to on the website
- ~[ ] The auto-updater can never be updated~
- [ ] The auto-updater never fails
- [ ] The auto-updater can rollback a failing upgrade
- [ ] The auto-updater uses multiple mirrors
- [ ] The auto-updater supports server-side downtime

Geod24:
We might be able to leverage https://github.com/containrrr/watchtower with a special boot sequence (migration scripts).

PR #2364, opened 2021-08-09
Part of #267: Implement in Agora version upgrader

Geod24:
```
In order to fix the issue, we need two components:
- The ability to update the node's persistent storage to accomodate for format / schema changes;
- The ability to automatically download and install a new update whenever available;

This implements the former functionality. The later will be handled outside of Agora.
```

This still need:
- Proper testing
- A fix to `semver`: https://github.com/dcarp/semver/pull/5

Submitting in case anyone want to give feedback on the design.

### DIFF
--- a/source/agora/node/Versioning.d
+++ b/source/agora/node/Versioning.d
@@ -1,0 +1,160 @@
+/*******************************************************************************
+
+    Apply version upgrades to a node
+
+    To allow for a smooth upgrade path for Agora node, there is a built-in
+    upgrade system: When a node starts, it checks a specific table in stateDB,
+    and gradually run upgrade scripts as needed.
+
+    Copyright:
+        Copyright (c) 2019-2021 BOSAGORA Foundation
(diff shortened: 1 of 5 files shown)


source/agora/node/Versioning.d
```diff
@@ -0,0 +1,160 @@
...
+        const size_t downgrades = 1;
+        log.info("Need to apply {} downgrades from {} to {}", downgrades, oldVers, vers);
+    }
+    stateDB.setMetadataVersion(current);
+}
+
+/// Set the current version in the metadata
+private void setMetadataVersion (ManagedDatabase stateDB, string version_)
+{
+    stateDB.execute("INSERT OR REPLACE INTO metadata (key, value) VALUES ('version', ?)", version_);
```
mkykadir, line 159: Sqlite have [this](https://www.sqlite.org/pragma.html#pragma_user_version)

PR #2373, opened 2021-08-10
Fix `ensure` to actually write the error message + trivial fixes

Geod24:
Better reviewed commit-by-commit.
The first 6 are pretty trivial and not worth the CI time. The last two are fixing `ensure` to work as one would expect, that is, report the correct file/line and actually write the message.
Note that the stack trace will still point to `ensure` itself but it will be transparent otherwise.

### DIFF
--- a/source/agora/common/Ensure.d
+++ b/source/agora/common/Ensure.d
@@ -49,12 +49,15 @@ private final class FormattedException : Exception
     }
 }
 
-public void ensure (Args...) (bool exp, string fmt, lazy Args args)
+public void ensure (Args...) (bool exp, string fmt, lazy Args args,
+    string file = __FILE__, typeof(__LINE__) line = __LINE__)
 {
     if (!exp)
     {
         auto res = instance.buffer.snformat(fmt, args);
         instance.end = res.length;
+        instance.file = file;
+        instance.line = line;
         throw instance;
     }
 }
--- a/source/agora/common/Ensure.d
+++ b/source/agora/common/Ensure.d
@@ -42,10 +42,60 @@ private final class FormattedException : Exception
         super("An Exception occured");
     }
 
+    /// Overrides `Throwable.message`
     public override const(char)[] message () const return
         @safe pure nothrow @nogc
     {
         return this.buffer[0 .. this.end];
+    }
+
+    /***************************************************************************
+
+        Overrides `Throwable.toString` sink overload
+
+        Unfortunately, `Throwable.toString` does not call `message`,
+        instead calling `msg` directly, which means our overload of `message`
+        does not help. We need to overload this method and replicate the logic
+        in the original code (see `object.d` in druntime for the original).
+
+        Params:
+          sink = The sink to send the piece-meal string to
+
+    ***************************************************************************/
+
+    public override void toString (scope void delegate(in char[]) sink) const scope
+    {
+        import core.internal.string : unsignedToTempString;
+
+        char[20] buffer = void;
+
+        sink(typeid(this).name);
+        sink("@"); sink(file);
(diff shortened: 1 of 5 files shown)


source/agora/common/Ensure.d
```diff
@@ -47,14 +47,49 @@ private final class FormattedException : Exception
     {
         return this.buffer[0 .. this.end];
     }
+
+    /// Overrides `Throwable.toString` because it uses `msg`...
+    public override void toString (scope void delegate(in char[]) sink) const scope
+    {
+        import core.internal.string : unsignedToTempString;
+
+        char[20] tmpBuff = void;
```
hewison-chris:
```suggestion
        char[20] line_number = void;
```

source/agora/common/Ensure.d
```diff
@@ -47,14 +47,49 @@ private final class FormattedException : Exception
...
+
+        sink(typeid(this).name);
+        sink("@"); sink(file);
+        sink("("); sink(unsignedToTempString(line, tmpBuff)); sink(")");
+
+        if (this.end > 0)
+        {
+            sink(": "); sink(this.message());
+        }
+        if (this.info)
```
hewison-chris, line 84: Where is info from?

hewison-chris (approved):
LGTM just a small suggestion and question.

source/agora/common/Ensure.d
kchulj: I'm assuming `class Exception`, which this one inherited from?

Geod24: https://github.com/dlang/druntime/blob/934b4d24bc9a24d93b1d7c380d6155cc0252f9c3/src/object.d#L2283-L2310

source/agora/common/Ensure.d
Geod24: It's almost a pure copy of https://github.com/dlang/druntime/blob/934b4d24bc9a24d93b1d7c380d6155cc0252f9c3/src/object.d#L2439-L2474

source/agora/common/Ensure.d
```diff
@@ -47,14 +47,49 @@ private final class FormattedException : Exception
...
+
+        char[20] tmpBuff = void;
+
+        sink(typeid(this).name);
+        sink("@"); sink(file);
+        sink("("); sink(unsignedToTempString(line, tmpBuff)); sink(")");
+
+        if (this.end > 0)
+        {
+            sink(": "); sink(this.message());
```
kchulj, line 80: is using a custom `message()` more efficient than using the `msg` in `Throwable`?

kchulj (approved):
LGTM w/ a question

source/agora/common/Ensure.d
Geod24: No, because `message` is a (virtual) function call while `msg` is a field.
But `message` was introduced for users that needed something like this (buffer + format).
Unfortunately, `toString` wasn't changed to use it.

Geod24: FYI: https://github.com/dlang/druntime/commit/e7e49cd206dec5324f9f4baf56099e211d8f9c9e

source/agora/common/Ensure.d
hewison-chris: I think that is why it lacks documentation. Most of Agora code has more descriptive names and comments to make things easier to quickly understand.

source/agora/common/Ensure.d
hewison-chris: I was hinting for a comment above this line.

Geod24:
Fixed the style issues

source/agora/common/Ensure.d
hewison-chris: Thanks.

PR #2374, opened 2021-08-10
Ledger: Merge slashing of UTXOSet / ValidatorSet together

Geod24:
```
This refactors the slashing to be performed in one single function,
instead of two functions called from the respective update methods.

The rationale for doing so is to prevent any slashed validator
to have an impact on the block, which means slashing needs to be done first.

Once that change was applied, it also was obvious that most of the functions
were redundant with one another, hence why they were merged.
This also allows to remove `Ledger.getMissingValidatorsUTXOs` which was here
only to deal with the duplication. Additionally, using `getValidators` and
a `filter` call makes this trivial. Finally, `getValidators used to be called
needlessly (when there was no slashing informations), and this was fixed as well.
```

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -374,6 +374,7 @@ public class Ledger
         {
             // rollback on failure within the scope of the db transactions
             scope (failure) ManagedDatabase.rollback();
+            this.applySlashing(block.header);
             this.updateUTXOSet(block);
             this.updateValidatorSet(block);
             ManagedDatabase.commitBatch();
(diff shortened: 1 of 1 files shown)


source/agora/consensus/Ledger.d
```diff
@@ -406,63 +407,79 @@ public class Ledger
...
+        - Second, its stake is consumed: One refund is created to the key
+          controlling the stake, and a penalty is sent to the commons budget.
 
-        Update the UTXOs of validators that are to be slashed
+        This is the first action that happens during block externalization,
+        so that slashed UTXOs are not spent by transactions.
 
         Params:
-            block = the block to update the UTXO set with
+            header = The `BlockHeader` containing the slashing informations
```
```suggestion
            header = The `BlockHeader` containing the slashing information
```

hewison-chris (approved):
LGTM just tiny doc change

source/agora/consensus/Ledger.d
Geod24: Thanks, fixed

PR #2375, opened 2021-08-10
Set: Returns whether an element was removed

Geod24:
This is the standard behavior of AA, and our Set implementation should follow it.

### DIFF
--- a/source/agora/common/Set.d
+++ b/source/agora/common/Set.d
@@ -39,10 +39,11 @@ public struct Set (T)
         this._set[key] = true;
     }
 
-    /// Remove an element from the set
-    public void remove (T key)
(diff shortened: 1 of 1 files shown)


kchulj:
Was the function removing, but not returning anything without the `return`?

kchulj (approved):
LGTM w/ a question

hewison-chris (approved):
LGTM

Geod24:
> Was the function removing, but not returning anything without the `return`?

It had the same behavior (removing if possible, no-op otherwise), but just not making the result available to the caller.
There's no point in not making the result available to the caller, as it doesn't improve encapsulation, it's just a pointless limitation.

Issue #2200, opened 2021-06-15
Flash: off-chain transactions should include fees


omerfirmak:
Flash nodes pre-sign the off-chain transactions, later to be used in a non-collaborative close attempt. Since this involves timelocks, timely externalization of those transactions are key to the security of Flash Layer. In a congested network, including fees to the TX is the only way to somewhat ensure that.

Trigger/update/settlement transactions should include some (probably dynamically determined) fees to be able to function properly.

omerfirmak:
We could interface with Stoa as it already [provides](https://github.com/bosagora/stoa/blob/7107c0a63ff66a0f6a3d2e188b598d41a18f7ad9/src/Stoa.ts#L139) a fee calculation interface. Assuming it will be dynamic, we can probably use that and add a margin on top of that just to be sure.

omerfirmak:
From https://github.com/bosagora/agora/pull/2227;

> Any fees associated with a non-collaborative closing attempt should be handled differently and paid by the peer that is attempting to close the channel.

With our current sequence signatures using the whole TX hash, it is not possible to add new inputs/outputs to a settlement/update TX after it is signed. We should come up with a scheme that allows adding new inputs/outputs but ensures that we cannot modify the existing ones. Since existing inputs/outputs are of equal value, there will be no risk of channel funds being stolen/rerouted.

omerfirmak:
I have decided to only hash the channel funds related inputs/outputs and use TX `payload` field to determine which ones are to be hashed or not. Will probably have a PR soon.

omerfirmak:
> 
> 
> I have decided to only hash the channel funds related inputs/outputs and use TX `payload` field to determine which ones are to be hashed or not. Will probably have a PR soon.

See https://github.com/bosagora/agora/pull/2227/commits/d698dd14352adab0441ff740386780a4178cb16a

omerfirmak:
#2227 implements this for funding and closing TXs but it is better to implement fees for update/settlement TXs after #2281 is resolved. So this issue is half in-review half blocked.

Geod24:
Fixed in #2383

PR #2383, opened 2021-08-12
Flash: Add fees to update TXs before publishing them 

omerfirmak:
Part of #2200

### DIFF
--- a/source/agora/flash/Node.d
+++ b/source/agora/flash/Node.d
@@ -180,7 +180,8 @@ public abstract class FlashNode : FlashControlAPI
             &this.getFlashClient, engine, taskman,
             &this.putTransaction, &this.paymentRouter,
             &this.onChannelNotify,
-            &this.onPaymentComplete, &this.onUpdateComplete);
+            &this.onPaymentComplete, &this.onUpdateComplete,
+            &this.getFeeUTXOs);
 
         this.network = new Network((Hash chan_id, Point from) {
             if (auto updates = chan_id in this.channel_updates)
@@ -591,7 +592,8 @@ public abstract class FlashNode : FlashControlAPI
         auto channel = new Channel(this.conf, chan_conf, key_pair,
             priv_nonce, peer_nonce, peer, this.engine, this.taskman,
             &this.putTransaction, &this.paymentRouter, &this.onChannelNotify,
-            &this.onPaymentComplete, &this.onUpdateComplete, this.db);
+            &this.onPaymentComplete, &this.onUpdateComplete,  &this.getFeeUTXOs,
+            this.db);
 
         this.channels[recv_pk][chan_conf.chan_id] = channel;
         this.network.addChannel(chan_conf);
@@ -1194,7 +1196,7 @@ public abstract class FlashNode : FlashControlAPI
         auto channel = new Channel(this.conf, chan_conf, key_pair,
             priv_nonce, result.value, peer, this.engine, this.taskman,
             &this.putTransaction, &this.paymentRouter, &this.onChannelNotify,
-            &this.onPaymentComplete, &this.onUpdateComplete, this.db);
+            &this.onPaymentComplete, &this.onUpdateComplete, &this.getFeeUTXOs, this.db);
         this.channels[reg_pk][chan_conf.chan_id] = channel;
     }
 
@@ -1292,6 +1294,19 @@ public abstract class FlashNode : FlashControlAPI
             return false;
 
         return true;
+    }
+
+    ///
+    protected FeeUTXOs getFeeUTXOs (PublicKey pk, ulong tx_size)
+    {
+        auto per_byte = this.listener.getEstimatedTxFee();
+        if(!per_byte.mul(tx_size))
+            return FeeUTXOs.init;
+        auto utxos = this.listener.getFeeUTXOs(pk, per_byte);
+        // reuse total_value as refund amount
+        if (!utxos.total_value.sub(per_byte))
+            utxos.total_value = Amount(0);
+        return utxos;
     }
 }
 
--- a/source/agora/flash/Node.d
+++ b/source/agora/flash/Node.d
@@ -760,12 +760,12 @@ public abstract class FlashNode : FlashControlAPI
     }
 
     /// See `FlashAPI.requestUpdateSig`
-    public override Result!Signature requestUpdateSig (PublicKey sender_pk,
+    public override Result!SigPair requestUpdateSig (PublicKey sender_pk,
         PublicKey recv_pk, /* in */ Hash chan_id, /* in */ uint seq_id) @trusted
     {
         auto secret_key = recv_pk in this.managed_keys;
         if (secret_key is null)
-            return Result!Signature(ErrorCode.KeyNotRecognized,
+            return Result!SigPair(ErrorCode.KeyNotRecognized,
                 format("The provided key %s is not managed by this "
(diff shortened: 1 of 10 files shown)


source/agora/script/Signature.d
```diff
@@ -142,39 +213,147 @@ unittest
...
+    import std.exception : assumeWontThrow;
+
+    if (sig_hash != SigHash.All)
+    {
+        assert(input_idx < tx.inputs.length, "Input index is out of range");
+        if (sig_hash & SigHash.Single)
+            assert(output_idx < tx.outputs.length, "Output index is out of range");
+    }
+
+    Transaction dup = () @trusted { return *cast(Transaction*)&tx; }();
```
Geod24, line 239: We really need to get rid of this...

source/agora/flash/UpdateSigner.d
```diff
@@ -480,17 +483,19 @@ public class UpdateSigner
         // Note that an update tx with seq 0 do not exist.
         if (this.seq_id == 0)
         {
-            return sign(this.kp.secret, this.conf.pair_pk, nonce_pair_pk,
-                priv_nonce.update.v, update_tx.getChallenge());
+            const sig_hash = cast(SigHash)(SigHash.Single | SigHash.AnyoneCanPay);
```
Geod24: Nope, a flag is not an enum. I wish D had something like C#'s flag, but it currently doesn't.
Casting an out-of-range value to an `enum` type is like taking a chainsaw to the type system, and I'd rather we do not do that.

source/agora/flash/UpdateSigner.d
```diff
@@ -687,15 +698,15 @@ public class UpdateSigner
...
     {
         // if the current sequence is 0 then the update tx is a trigger tx that
         // only needs a multi-sig and does not require a sequence.
         // an update tx with seq 0 do not exist.
-        if (this.seq_id == 0)
+        if (seq_id == 0)
             return genKeyUnlock(update_multi_sig);
         else
-            return createUnlockUpdate(update_multi_sig, this.seq_id);
+            return createUnlockUpdate(update_multi_sig, seq_id);
```
Geod24: Minor, but that `else` is superfluous (code would work the same without it).

source/agora/script/Signature.d
```diff
@@ -74,23 +98,64 @@ align(1):
 *******************************************************************************/
 
 public string decodeSignature (const(ubyte)[] bytes,
-    out SigPair sig_pair) pure nothrow @safe @nogc
+    out SigPair sig_pair, out ulong pop_count) pure nothrow @safe @nogc
 {
-    if (bytes.length != SigPair.sizeof)
+    pop_count = 0;
```
Geod24: `out` does that for you

source/agora/script/Signature.d
```diff
@@ -74,23 +98,64 @@ align(1):
...
 
-    sig_pair = SigPair(sig, sig_hash);
+    ulong output_idx;
+    if (sig_hash & SigHash.Single && bytes.length >= ulong.sizeof)
+    {
+        output_idx = littleEndianToNative!ulong(bytes[0 .. ulong.sizeof]);
+        pop_count += ulong.sizeof;
+    }
+    else if (sig_hash & SigHash.Single)
+        return "Encoded signature does not have output idx";
```
Geod24:
```suggestion
    if (sig_hash & SigHash.Single)
    {
        if (bytes.length < ulong.sizeof)
            return "Encoded signature does not have output idx";
        output_idx = littleEndianToNative!ulong(bytes[0 .. ulong.sizeof]);
        pop_count += ulong.sizeof;
    }
```

Would be equivalent, right ?

source/agora/script/Signature.d
```diff
@@ -74,23 +98,64 @@ align(1):
...
+    auto serialized = pair[].dup;
+
+    SigPair decoded_pair;
+    ulong consumed;
+
+    assert(decodeSignature(serialized, decoded_pair, consumed) is null);
+    assert(pair == decoded_pair);
+    assert(consumed == serialized.length);
+
+    serialized ~= ubyte(44);
```
Geod24: Don't think you need the `ubyte()` here, the compiler knows the type of the LHS.

source/agora/script/Signature.d
```diff
@@ -142,39 +213,147 @@ unittest
...
-        dup.inputs = dup.inputs.dup;
-        dup.inputs[input_idx] = Input.init;  // blank out matching input
+        dup.inputs = dup.inputs[0..input_idx] ~ dup.inputs[input_idx + 1..$]; // blank out matching input
+        return hashMulti(dup, sig_hash);
+    // sign all inputs and a single output
+    case SigHash.Single:
+        dup.outputs = dup.outputs[output_idx .. output_idx + 1];
+        return hashMulti(dup, sig_hash);
+    // sign a single input and a single output
+    case SigHash.Single | SigHash.AnyoneCanPay:
```
Geod24: One reason why using OOB values for `enum` is bad: The compiler could decide that this is definitely unreachable. Not to mention, `final switch` is useless here.

source/agora/test/Flash.d
```diff
@@ -418,8 +418,9 @@ public class FlashNodeFactory (FlashListenerType = FlashListener)
     public RemoteAPI!TestFlashListenerAPI createFlashListener (
         Listener : TestFlashListenerAPI)(string address)
     {
+        const string agora_address = format("Validator #%s (%s)", 0, genesis_validator_keys[0].address);
         RemoteAPI!TestFlashListenerAPI api
-            = RemoteAPI!TestFlashListenerAPI.spawn!Listener(5.seconds);
+            = RemoteAPI!TestFlashListenerAPI.spawn!Listener(this.agora_registry, agora_address, 5.seconds);
```
Geod24, line 423: Any hope we can merge this custom impl into `Base.d` ? :D

source/agora/flash/Node.d
```diff
@@ -1292,6 +1295,19 @@ public abstract class FlashNode : FlashControlAPI
...
+    {
+        auto per_byte = this.listener.getEstimatedTxFee();
+        if(!per_byte.mul(tx_size))
+            return FeeUTXOs.init;
+        auto utxos = this.listener.getFeeUTXOs(pk, per_byte);
+        // reuse total_value as refund amount
+        if (!utxos.total_value.sub(per_byte))
+            utxos.total_value = Amount(0);
+        return utxos;
+    }
```
Geod24, line 1310: Why do we have two implementations of this ?

source/agora/flash/UpdateSigner.d
omerfirmak: I explicitly defined them.

source/agora/flash/Node.d
omerfirmak: this one takes `tx_size`
`FlashListener.getFeeUTXOs` takes flat `Amount`

source/agora/script/Signature.d
omerfirmak: Deep-dup thru ser/deser is not possible atm.
`TxBuilder` is signing invalid TXs while trying to calculate minimum fees and that triggers assertions in ctors.

https://github.com/bosagora/agora/blob/be48645a0c1e921bb562c5ed10873bf7ceccf3de/source/agora/utils/TxBuilder.d#L598

source/agora/test/Flash.d
omerfirmak: `FlashNodeFactory` as a whole?

Issue #2380, opened 2021-08-12
Set the keep-alive of the stats endpoint independently


Geod24:
Seen on the name-registry, probably affects Agora as well:
```
Aug 12 04:12:22 eu-001.bosagora.io docker[578903]: [main(MCNT) dbv] route match: /metrics -> GET /metrics []
Aug 12 04:12:32 eu-001.bosagora.io docker[578903]: [main(MCNT) dbg] Keep-alive connection timed out!
Aug 12 04:12:37 eu-001.bosagora.io docker[578903]: [main(M15D) dbv] route match: /metrics -> GET /metrics []
Aug 12 04:12:47 eu-001.bosagora.io docker[578903]: [main(M15D) dbg] Keep-alive connection timed out!
```

The keep-alive seems to be 10 seconds, however stats are queried every 15 seconds.

Geod24:
https://github.com/vibe-d/vibe.d/issues/2593

Geod24:
We worked around it in #2384 , so closing as there's unlikely to be any issue caused by the new default.

PR #2384, opened 2021-08-13
Stats server: Set keep alive to 70 seconds

Geod24:
Workaround issue #2380.


kchulj:
Need an import for this?
`source/agora/stats/Server.d(48,39): Error: no property `seconds` for type `int``

Geod24:
Missed the `core.time` import indeed.

kchulj (approved):
LG

PR #2387, opened 2021-08-13
openssl: Update Deimos bindings to v2.0.1+1.1.0h

Geod24:
We were using a very outdated bindings of OpenSSL.
Sadly the latest ones don't compile with Vibe.d,
but this is already a step forward.


Geod24:
@omerfirmak : That might be the cause of all our pain and suffering.

PR #2391, opened 2021-08-18
dub.json: Reduce exclusion and minor clean up

Geod24:
The exclusion isn't necessary now that we fixed DUB, and the version script should be moved to `scripts`.


kchulj:
`source/agora/node/FullNode.d`
`source/agora/node/main.d`
might need additional changes..

PR #2398, opened 2021-08-23
Minor improvements to config handling

Geod24:
Pretty simple improvements before the big chunk.

PR #2402, opened 2021-08-23
Implement an automatic config filler instead of doing it manually

Geod24:
```
Our configuration started pretty simple, and organically grew over time.
Originally, we wrote code to read the YAML `Node` and fill the `struct` / arrays / value,
as it was the most straightforward and efficient approach.
However, our config file has now reached a size where there is a lot of boilerplate
and repetition needed to achieve this.

Thus, this changes the implementation of the configuration to generate code from the type definition.
`parseConfig` will now walk the provided `struct` recursively, attempt to read each field,
use the default value if a sensible one is provided, or throw otherwise.

As a result, a few attributes needed to be added to the `Config` et al. type definitions
to inform the config filler of which fields were optional despite not having a different default,
as well as how to convert some fields (most notably `Duration`).

To support this change, `agora.common.Config` now only contains the config filling code,
with the type definitions having been moved to `agora.node.Config`.
To reduce dependencies, a new module, `agora.common.ConfigAttributes` has also been added.
```

The first 4 commits are part of #2398 . This was made so that implementing the Registry's config file will be ~simpler~ trivial.


Geod24:
Leaving it as a draft as there's more testing to be done.

Geod24:
I have added a pretty-printer so that issues will be easy to debug, and correctness easier to prove.
Example output:
Screen Shot 2021-08-24 at 15 49 45

source/agora/common/BanManager.d
```diff
@@ -38,7 +39,7 @@ public class BanManager
         public size_t max_failed_requests = 100;
 
         /// How long does a ban lasts, in seconds (default: 1 day)
-        public Duration ban_duration = 1.days;
+        public @fromSeconds Duration ban_duration = 1.days;
```
hewison-chris, line 42: Where is `@fromSeconds` defined and what does the @ mean?

kchulj: It is defined in `ConfigAttributes.d`.

Geod24: For ref: https://dlang.org/spec/attribute.html#uda

source/agora/common/ConfigAttributes.d
```diff
@@ -0,0 +1,204 @@
...
+
+public struct Name
+{
+    ///
+    public string name;
+}
+
+/*******************************************************************************
+
+    Provides a mean to convert a field from a `string` to a complex type
```
hewison-chris:
```suggestion
    Provides a means to convert a field from a `string` to a complex type
```

source/agora/flash/Config.d
```diff
@@ -39,7 +40,7 @@ public struct FlashConfig
     public immutable string[] addresses_to_register;
 
     /// Timeout for requests
-    public Duration timeout = 10.seconds;
+    public @fromSeconds Duration timeout = 10.seconds;
```
hewison-chris: Maybe we should use `@fromMsecs` for these values just in case we want to test with less than a second.

source/agora/node/Config.d
```diff
@@ -0,0 +1,662 @@
...
+    /// Maximum number of listeners to connect to
+    public size_t max_listeners = 10;
+
+    /// The local address where the stats server (currently Prometheus)
+    /// is going to connect to, for example: http://0.0.0.0:8008
+    /// It can also be set to 0 do disable listening
+    public @Optional ushort stats_listening_port;
+
+    /// Time to wait between request retries
+    public @fromSeconds Duration retry_delay = 3.seconds;
```
hewison-chris: Also maybe `@fromMsecs`

source/agora/node/Config.d
```diff
@@ -0,0 +1,662 @@
...
+
+    /// Path to the data directory to store metadata and blockchain data
+    public string data_dir = ".cache";
+
+    /// The duration between requests for doing periodic network discovery
+    public @fromSeconds Duration network_discovery_interval = 5.seconds;
+
+    /// The duration between requests for retrieving the latest blocks
+    /// from all other nodes
+    public @fromSeconds Duration block_catchup_interval = 20.seconds;
```
hewison-chris, line 200: Also maybe `@fromMsecs`

source/agora/node/Config.d
```diff
@@ -0,0 +1,662 @@
...
+    // increased in order to be added to the transaction pool
+    public ubyte double_spent_threshold_pct = 20;
+
+    /// The maximum number of transactions relayed in every batch.
+    /// Value 0 means no limit.
+    public @Optional uint relay_tx_max_num;
+
+    /// Transaction relay batch is triggered in every `relay_tx_interval`.
+    /// Value 0 means, the transaction will be relayed immediately.
+    public @fromSeconds @Optional Duration relay_tx_interval;
```
hewison-chris, line 218: Also maybe `@fromMsecs`

source/agora/node/Config.d
```diff
@@ -0,0 +1,662 @@
...
+
+    /// The minimum amount of fee a transaction has to have to be relayed.
+    /// The fee is adjusted by the transaction size:
+    /// adjusted fee = fee / transaction size in bytes.
+    public @Optional Amount relay_tx_min_fee;
+
+    /// Transaction put into the relay queue will expire, and will be removed
+    /// after `relay_tx_cache_exp`.
+    public @fromSeconds @Optional Duration relay_tx_cache_exp;
```
hewison-chris, line 228: Also maybe `@fromMsecs`

source/agora/node/Config.d
```diff
@@ -0,0 +1,662 @@
...
+    public @Optional immutable string[] addresses_to_register;
+
+    // Registry address
+    public string registry_address;
+
+    // If the enrollments will be renewed or not at the end of the cycle
+    public bool recurring_enrollment = true;
+
+    /// How often should the periodic preimage reveal timer trigger (in seconds)
+    public @fromSeconds Duration preimage_reveal_interval = 10.seconds;
```
hewison-chris, line 257: Also maybe `@fromMsecs`

source/agora/node/Config.d
```diff
@@ -0,0 +1,662 @@
...
+
+    /// How often the nomination timer should trigger, in seconds
+    public @fromSeconds Duration nomination_interval = 5.seconds;
```
hewison-chris, line 260: Also maybe `@fromMsecs`.

source/agora/node/Config.d
```diff
@@ -0,0 +1,662 @@
...
+
+    /// How often the validator should try to catchup for the preimages for the
+    /// next block
+    public @fromSeconds Duration preimage_catchup_interval = 2.seconds;
```
hewison-chris, line 264: Also maybe `@fromMsecs`

hewison-chris (changes requested):
Looks very good. I would just prefer some `Durations` to be `msecs` just in case. It will be harder to change once we start creating configs with seconds and then later want to change to msecs.

source/agora/node/Config.d
Geod24: Actually this was linked to field "block_catchup_interval_secs".
I need to enable strict mode to catch those...

Geod24:
```
	FlashConfig: `parseMapping` called for 'flash' (node entries: 11)
agora.common.Ensure.FormattedException@source/agora/common/Config.d(267): Unexpected key 'seed' in section 'flash'. Valid keys are: ["enabled", "registry_address", "addresses_to_register", "timeout", "control_address", "control_port", "listener_address", "min_funding", "max_funding", "min_settle_time", "max_settle_time", "max_retry_time", "max_retry_delay", "retry_multiplier"]
----------------
```

I originally made "strict" mode disabled by default because there was much to implement. But now that those things have been implemented, best to turn it back on before the merge. It caught a few misnamed fields in the documented config file, and one unused field, shown above.

source/agora/node/Config.d
Geod24: Ditto this was a `_secs` field

source/agora/node/Config.d
Geod24: Ditto `_secs`

source/agora/node/Config.d
Geod24: The example config file uses seconds, so let's leave it as seconds and re-evaluate in a different PR, so this one can focus on the framework and not how it's used.

source/agora/node/Config.d
Geod24: This was parsed as `seconds` before.

source/agora/node/Config.d
Geod24: This was parsed as `seconds` before

Geod24:
Updated and addressed all comments. Enabled `strict` mode by default caught a few error. With the config dumper, I can also see that some fields are undocumented. Some are expected, such as the whole `consensus` section, or `testing`, but not those:
```
FlashConfig:
		control_address: Field is missing from node
		Using default value '127.0.0.1' for optional field 'control_address'
		control_port: Field is missing from node
		Using default value '2828' for optional field 'control_port'
```

```
ValidatorConfig:
		nomination_interval: Field is missing from node
		Using default value '5 secs' for optional field 'nomination_interval'
```

```
	NodeConfig: `parseMapping` called for 'node' (node entries: 15)
		commons_budget_address: Field is missing from node
		Using default value 'boa1xqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqa9jq8m' for optional field 'commons_budget_address'
```

Geod24:
A future enhancement would be to generate the example config file from the struct, but that isn't simple to do, and we don't really need it right now.

source/agora/common/Config.d
```diff
@@ -81,1059 +136,693 @@ public struct CommandLine
...
+    !hasConverter!(FR.Ref) && !hasFromString!(FR.Type) && !hasStringCtor!(FR.Type);
 
-    {
-        immutable conf_example = `
-node:
-  interfaces:
-    - type: http
-      address: 0.0.0.0
-      port: 2926
-  data_dir: .cache
```
Geod24, line 604: Strict mode blew a fuse on this, so I have removed the nested `interfaces` as it is incorrect.

Geod24:
```
Unittesting agora.test.QuorumPreimage================================ ASSERT HANDLER ===============================
[source/agora/test/Base.d:1286] Assertion thrown during test: Trying to access node at address 'Validator #5 (boa1xzval4nvru2ej9m0rptq7hatukkavemryvct4f8smyy3ky9ct5u0s8w6gfy)' from 'Validator #5 (boa1xzval4nvru2ej9m0rptq7hatukkavemryvct4f8smyy3ky9ct5u0s8w6gfy)' without first creating it
source/agora/test/Base.d:1286 agora.test.Base.TestAPI agora.test.Base.TestNetworkManager.getClient(immutable(char)[], core.time.Duration) [0x10093485f]
source/agora/network/Manager.d:167 void agora.network.Manager.NetworkManager.ConnectionTask.connect_canthrow() [0x100a3216e]
source/agora/network/Manager.d:158 nothrow void agora.network.Manager.NetworkManager.ConnectionTask.connect() [0x100a3208b]
core/thread/context.d:46 void core.thread.context.Callable.opCall() [0x1016372cc]
core/thread/fiber.d:1141 void core.thread.fiber.Fiber.run() [0x101637458]
core/thread/fiber.d:198 fiber_entryPoint [0x101637395]
#### FATAL ERROR: Trying to access node at address 'Validator #5 (boa1xzval4nvru2ej9m0rptq7hatukkavemryvct4f8smyy3ky9ct5u0s8w6gfy)' from 'Validator #5 (boa1xzval4nvru2ej9m0rptq7hatukkavemryvct4f8smyy3ky9ct5u0s8w6gfy)' without first creating it
./ci/run.sh: line 21: 12703 Abort trap: 6           ./build/agora-unittests
+ ./build/agora-unittests
This node was started at source/agora/test/QuorumShuffle.d:39
Full error: core.exception.AssertError@source/agora/test/Base.d(1286): Trying to access node at address 'Validator #5 (boa1xzval4nvru2ej9m0rptq7hatukkavemryvct4f8smyy3ky9ct5u0s8w6gfy)' from 'Validator #5 (boa1xzval4nvru2ej9m0rptq7hatukkavemryvct4f8smyy3ky9ct5u0s8w6gfy)' without first creating it
----------------
source/agora/test/Base.d:171 nothrow void agora.test.Base.testAssertHandler(immutable(char)[], ulong, immutable(char)[]) [0x10092ff7e]
core/exception.d:448 onAssertErrorMsg [0x10160de8b]
core/exception.d:590 _d_assert_msg [0x10160e215]
source/agora/test/Base.d:1286 agora.test.Base.TestAPI agora.test.Base.TestNetworkManager.getClient(immutable(char)[], core.time.Duration) [0x10093485f]
source/agora/network/Manager.d:167 void agora.network.Manager.NetworkManager.ConnectionTask.connect_canthrow() [0x100a3216e]
source/agora/network/Manager.d:158 nothrow void agora.network.Manager.NetworkManager.ConnectionTask.connect() [0x100a3208b]
core/thread/context.d:46 void core.thread.context.Callable.opCall() [0x1016372cc]
core/thread/fiber.d:1141 void core.thread.fiber.Fiber.run() [0x101637458]
core/thread/fiber.d:198 fiber_entryPoint [0x101637395]
Unittesting scpd.types.Stellar_types (took 1 hnsec)
```

Blast from the past.

Geod24:
Okay, this is ready!

doc/config.example.yaml
```diff
@@ -236,14 +238,18 @@ logging:
 ################################################################################
 event_handlers:
   # URLs to push a data when a block is externalized. (path is "/block_externalized")
-  block_externalized:
-    - http://127.0.0.1:3836/block_externalized
+  BlockExternalized:
+    addresses:
+      - http://127.0.0.1:3836/block_externalized
```
hewison-chris, line 243: ~~Why do we need multiple addresses?~~ Ignore this.

tests/system/node/4/config.yaml
```diff
@@ -11,7 +11,6 @@ node:
   timeout: 5000
   # Path to the data directory (if the path doesn't exist it will be created)
   data_dir: .cache
-  block_interval_sec: 5
```
hewison-chris, line 14: Strictly speaking shouldn't this be a separate commit?

hewison-chris (approved):
LGTM

source/agora/flash/Config.d
```diff
@@ -39,7 +40,7 @@ public struct FlashConfig
     public immutable string[] addresses_to_register;
 
     /// Timeout for requests
-    public Duration timeout = 10.seconds;
+    public @fromMsecs Duration timeout = 10.seconds;
```
kchulj, line 43: Does the `@fromMsecs` convert the `seconds` to `msecs`?

hewison-chris: No it converts `msecs` as string from a `config.yaml` to a `Duration`. The default here is specified in `seconds` but could be any valid `Duration`.

tests/system/node/4/config.yaml
Geod24: I deserved it.

source/agora/flash/Config.d
kchulj: Understood. Is the `Duration` type specified by calling it (`.seconds`)?

Geod24:
Commit split done @hewison-chris :)

hewison-chris (approved):
LGTM

source/agora/flash/Config.d
Geod24: @kchulj : A `Duration` is just a `struct` that stores `hnsecs`:
https://github.com/dlang/druntime/blob/32299128cfebc9886703256ccda3654cfa05c76e/src/core/time.d#L1701-L1714

That's all there is. hnsecs stands for hecto-nanoseconds (100 ns). So one millisecond is 10_000 hnsecs, and one second is 10_000_000 hnsecs. So all `seconds` do is return a `Duration` which has the argument times 10M.

kchulj (approved):
LGTM

PR #2403, opened 2021-08-23
Fee: Make validator shares worth a fixed amount

### LINKED ISSUE #2240: TX fee calculation vulnerability
One validator can cause all the transaction fees to go to the Foundation and prevent other validators including itself to receive any transaction fee.

The validator can chose a prime number greater than 40000 Boa(40000.000_000_0) like the following
40000.000_001_9

If the total tx fee for the block was less then 40000 BOA and we have at least 1 validator whose stake is a relative prime to 40000.000_001_9 (reasonable assumption), then the tx fee will be 0

```D
        Amount share_stake = Amount.gcd(stake_amounts);
        // Total "share" count staked
        const total_shares = sum_stake.count(share_stake);

        // tx_fees now equals "share value"
        tx_fees.div(total_shares);
```

omerfirmak:
Now Validators will have to stake a minimum additional
amount to get a bigger share from the fees of a block.

Fixes #2240

### DIFF
--- a/source/agora/consensus/Fee.d
+++ b/source/agora/consensus/Fee.d
@@ -287,6 +287,7 @@ public class FeeManager
     private Amount[] getValidatorFees (Amount tot_fee, Amount tot_data_fee,
         UTXO[] stakes) nothrow @safe
     {
+        import std.numeric : gcd;
         // no stakes, no fees
         if (stakes.length == 0)
             return [];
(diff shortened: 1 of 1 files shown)


kchulj:
The logic LGTM. Are there any tests for this?

hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

omerfirmak:
> 
> 
> The logic LGTM. Are there any tests for this?

I made small changes to existing tests and that should be enough.

source/agora/consensus/Fee.d
```diff
@@ -297,22 +298,20 @@ public class FeeManager
...
         tx_fees.percentage(this.params.ValidatorTXFeeCut);
 
-        Amount sum_stake;
-        Amount[] stake_amounts = stakes.map!(utxo => utxo.output.value).array;
-        stake_amounts.each!(stake => sum_stake += stake);
+        // A Validator with MinFreezeAmount of coins staked will get 100 "shares"
+        // this fixes the stake amount that is needed to get a share
+        Amount share_stake = Amount.MinFreezeAmount;
+        share_stake.div(100);
+        // this will ignore any remaning part of the stake that is not worth a share
```
Geod24, line 305: 👍 for the comment

source/agora/consensus/Fee.d
```diff
@@ -525,10 +524,11 @@ public class FeeManager
         auto utxoset = new TestUTXOSet();
 
         Amount double_stake = Amount.MinFreezeAmount; assert(double_stake.mul(2));
+        Amount stake_with_excess = Amount.MinFreezeAmount; assert(double_stake.add(Amount(19))); // this should not change distribution
```
Geod24: Is the `assert` be on `double_stake` or `stake_with_excess` ?

Geod24: In any case, you can use the operators overloads here if you want, since it's not testing the addition / multiplication themselves.

source/agora/consensus/Fee.d
```diff
@@ -287,6 +287,7 @@ public class FeeManager
     private Amount[] getValidatorFees (Amount tot_fee, Amount tot_data_fee,
         UTXO[] stakes) nothrow @safe
     {
+        import std.numeric : gcd;
```
Geod24, line 290: Why the direct `gcd` usage instead of `Amount.gdc` ? Is it because `Amount.gcd` doesn't accept a range ?

omerfirmak: Because they are not `Amount`s anymore. They are plain `ulong`s, shares. `Amount.gcd()` was only used here, we might as well remove it.

Geod24 (approved):
Sounds good

PR #2404, opened 2021-08-23
TransactionPool: Try to remove() non-existing TXs

omerfirmak:
Code path will correctly handle the case where the requested TX is
not in the pool anyway. So no need to query the database only to
check for existence of the TX.

Additionally, in the case where we don't have an externalized TX
in our pool but have TXs that spend the same inputs we should still
continue to remove those double-spent TXs as if we had the original TX
in the pool.


linked0 (approved):
LGTM

kchulj (approved):
LGTM

Geod24 (approved):

PR #2407, opened 2021-08-24
Test: Flush stdout after destroying the lock

Geod24:
```
While this works locally, we are not seeing the flush in Github CI,
where it is most important. Currently, the line is only printed
once the timing's newline is printed.
```

Not sure if this will work, or if I need to go with https://github.com/Geod24/localrest/commit/3cceb64faad36084427ffd10db6703ad0b10e0eb or if it's just not possible... Let's see.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -262,18 +262,18 @@ private UnitTestResult customModuleUnitTester ()
 
             if (chatty)
             {
+                scope(exit) stdout.flush();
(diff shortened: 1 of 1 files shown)


Geod24:
Okay, this doesn't work, going to try `fflush` a bit later, cancelled the workflow for now to lighten the load on the CI.

Geod24:
Not really worth spending more time on this.

PR #2409, opened 2021-08-25
Only instantiate the Ledger once in Validator

Geod24:
We were instantiating it twice, which could be problematic,
and is definitely wasteful. We know we have a ValidatingLedger
so a simple dynamic cast will do.


linked0 (approved):
LGTM

hewison-chris (approved):
LGTM

kchulj (approved):
🟢

PR #2414, opened 2021-08-26
Nominator: Ignore early nominations

omerfirmak:
Although our nodes wait for block creation time before
nominating new values, they might be sent nomination envelopes
before the expected block time. They can simply ignore those.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -653,6 +653,19 @@ extern(D):
                 return; // We dont have all the TXs for this block. Try to catchup
             }
         }
+        else if (envelope.statement.pledges.type_ == SCPStatementType.SCP_ST_NOMINATE)
+        {
+            // show some tolerance to early nominations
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM

Geod24:
What if the nomination arrives mere seconds before ?
I feel like we should allow some jitter, especially considering nomination would only start at the time the block should be externalized according to the code.

omerfirmak:
> 
> 
> What if the nomination arrives mere seconds before ?
> I feel like we should allow some jitter, especially considering nomination would only start at the time the block should be externalized according to the code.

Added  %5 of the block interval as tolerance.

Geod24:
Right, but still not sure what's the goal ? Nodes can only nominate for the next block, right ? Is it a DoS vector ?

omerfirmak:
> 
> 
> Right, but still not sure what's the goal ? Nodes can only nominate for the next block, right ? Is it a DoS vector ?

If we don't reject these early nominations a single node would be able to send a nomination and trigger block creation anytime, no?

kchulj (approved):

PR #2416, opened 2021-08-26
Use average fees of TXs in pool as acceptance criteria

### LINKED ISSUE #1959: Agora mempool is unlimited in size
this could easily lead to DoS attacks, we should 
. limit the size of the pool
. evict old entries periodically
. replace transactions in the mempool based on (tx fee/size) when a new transaction is goshiped to us and the pool is at it's maximum size
. transactions received through getTransaction during consensus should probably be included in the pool even, if the pool would go above the limit

omerfirmak:
This will allow to protect against TX flood attacks.
    
With double_spent_threshold_pct and max_fee_undercut_pct, for an
attacker to flood more and more transactions to the network they would
have to pay increasingly more fees.

With double_spent_threshold_pct of %20, an attacker would have to pay almost 80M times
more than the average TX in the pool to be able to create 100 double-spent TXs.

With this, we can keep our pool size unlimited, but make it unpractical to use it as a
DoS vector.

Fixes #1959

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -339,25 +339,19 @@ public class Ledger
 
     ***************************************************************************/
 
-    public bool acceptTransaction (in Transaction tx,
+    public string acceptTransaction (in Transaction tx,
         in ubyte double_spent_threshold_pct = 0,
         in ushort min_fee_pct = 0) @safe
     {
         const Height expected_height = this.getBlockHeight() + 1;
-        string reason;
         auto tx_hash = hashFull(tx);
 
         // If we were looking for this TX, stop
         this.unknown_txs.remove(tx_hash);
 
-        scope (exit)
-            if (reason)
-                log.info("Rejected tx. Reason: {}. Tx: {}, txHash: {}",
-                    reason, tx, tx_hash);
-
         if (tx.isCoinbase ||
             this.pool.hasTransactionHash(tx_hash))
-            return false;
+            return "Coinbase or existing TX";
 
         Amount fee_rate;
         auto checkAndSave = (in Transaction tx, Amount tx_fee)
@@ -368,10 +362,10 @@ public class Ledger
             return this.fee_man.check(tx, tx_fee);
         };
 
-        if ((reason = tx.isInvalidReason(this.engine,
+        if (auto reason = tx.isInvalidReason(this.engine,
                 this.utxo_set.getUTXOFinder(),
-                expected_height, checkAndSave)) !is null)
-            return false;
+                expected_height, checkAndSave))
+            return reason;
 
         auto min_fee = this.pool.getAverageFeeRate();
         if (!min_fee.percentage(min_fee_pct))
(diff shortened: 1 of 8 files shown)


omerfirmak:
This assumes we have more than a single block worth of TXs in the pool. So, once we put a limit to block size, it should be taken in to account as well.

omerfirmak:
Rebased

source/agora/consensus/Ledger.d
```diff
@@ -354,21 +368,24 @@ public class Ledger
...
-            (reason = tx.isInvalidReason(this.engine,
+        if ((reason = tx.isInvalidReason(this.engine,
                 this.utxo_set.getUTXOFinder(),
-                expected_height, checkAndSave)) !is null ||
-            !this.pool.add(tx, fee_rate))
+                expected_height, checkAndSave)) !is null)
+            return false;
+
+        auto min_fee = this.pool.getAverageFeeRate();
+        assert(max_fee_undercut_pct <= 100);
```
Geod24: We're not validating this in the config file parsing

source/agora/consensus/Ledger.d
```diff
@@ -340,10 +340,24 @@ public class Ledger
     ***************************************************************************/
 
     public bool acceptTransaction (in Transaction tx,
-        in ubyte double_spent_threshold_pct = 0) @safe
+        in ubyte double_spent_threshold_pct = 0,
+        in ubyte max_fee_undercut_pct = 100) @safe
     {
         const Height expected_height = this.getBlockHeight() + 1;
         string reason;
```
Geod24: Perhaps it's time to just change the return type ?

Geod24:
The aim LGTM, just a few notes about the form. In general we should try to make things fool-proof, which means that:
- Parameter default value should be justifiable from the interface PoV (e.g. it's  not reasonable to add a default value that does something incorrect);
- Things should be hard to misuse;

I see a lot of default values and I'm not sure they all pass the first point. In particular, when doing serialization, having the fee rate set to 0 seems off, especially since it can be computer directly inside the function (avoiding the need for the parameter in the first place). Regarding the second part, the undercut percentage makes sense from the PoV of usage, but it bounds the code to a very narrow set of values. It would be harder to use if, instead of having an undercut percentage, we'd have a "minimum percentage of the fees", so that "20% lower" would be "80% of the fee". This would also allow to express the desire of the node operator to have tx that are slightly higher than those in the pools (e.g. 101 becomes "1% higher than average).

However I'm not sure there's a need for such a use case - but with this definition, it makes all possible input to the `Ledger` function valid input, instead of limiting it to [0 - 100].

omerfirmak:
> 
> 
> Couldn't this function compute it ? This way you don't have to add the extra parameter and there's no risk of forgetting to pass the value ?

Ledger already calculates that while trying to validate the incoming TX. We would be doing it needlessly and, to me, it doesn't feel like it's pool's job, since it is just dumb storage.

Default value was there just to avoid changing bunch of tests, but I will just bite the bullet and make it required I guess

source/agora/test/LocalTransactions.d
```diff
@@ -88,11 +88,12 @@ unittest
...
         mixin ForwardCtor!();
 
-        public override bool acceptTransaction (in Transaction tx) @safe
+        public override string acceptTransaction (in Transaction tx,
+            in ubyte double_spent_threshold_pct = 0, in ushort min_fee_pct = 0) @safe
         {
             // Dont accept any incoming TX
             log.info("Picky node ignoring TX {}", tx);
-            return false;
+            return "Nah";
```
Geod24, line 96: :D

Geod24 (approved):

PR #2419, opened 2021-08-27
Catch up with missing preimages regardless of nomination status

Geod24:
Previously, we were only catching up with pre-images we've seen as missing
through side effects of the `isInvalidPreimageRootReason`.
This `private` member of the `Ledger` was then cleared in the `NetworkManager`.

Instead of relying on this tight coupling and hard to follow logic,
and as we want to try to catch-up with all missing pre-images anyways,
we can just add a task in `Validator` which behaves similarly to the block catchup.

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -15,6 +15,7 @@ module agora.node.Validator;
 
 import agora.api.Validator;
 import agora.common.Amount;
+import agora.common.Set;
 import agora.common.Task;
 import agora.common.Types;
 import agora.consensus.data.Block;
@@ -45,6 +46,7 @@ import scpd.types.Stellar_SCP;
 import scpd.types.Stellar_types : NodeID;
 
 import std.algorithm;
+import std.range : array, enumerate;
 
 import core.stdc.stdlib : abort;
 import core.time;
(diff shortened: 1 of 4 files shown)


source/agora/node/Validator.d
```diff
@@ -248,6 +251,76 @@ public class Validator : FullNode, API
...
+
+        auto missing = validators.filter!(en => en.preimage.height < next_height).array();
+        if (!missing.length)
+            return;
+
+        log.warn("Currently missing pre-images for next height ({}): {}",
+                 next_height, missing);
+
+        auto query = this.network.peers[]
+            .map!(peer => peer.client.getPreimages(next_height, next_height + 100));
```
hewison-chris: Why limit to pre-images 100 higher ? Any future pre-image will do.

source/agora/test/PreimageSharing.d
```diff
@@ -20,18 +20,18 @@ import agora.consensus.data.PreImageInfo;
...
-    /// GET: /preimages_for_enroll_keys
-    public override PreImageInfo[] getPreimagesForEnrollKeys (Set!Hash enroll_keys = Set!Hash.init) @safe nothrow
-    {
-        return TestValidatorNode.getPreimagesForEnrollKeys(enroll_keys);
-    }
+    ///
+    public override void onPreImageRevealTimer () @safe {}
+
+    /// To be extra sure, we also disable receiving a pre-image
+    /// so that node node may gossip them
```
hewison-chris:
```suggestion
    /// so that the node may gossip them
```

hewison-chris (approved):
LGTM

source/agora/node/Validator.d
Geod24: There's no way to express this in the API. Perhaps I should just call it with `Height.max`

Geod24: Thanks, new version looks much better.

linked0 (approved):
Nice!

Geod24:
> Thanks, new version looks much better.

This should have been preceded by a `git push`. Done  now.

hewison-chris (approved):
LGTM

kchulj (approved):

PR #2420, opened 2021-08-27
[Testing #2400] Fixed minimum fee check for empty payload

kchulj:
This PR is to test #2400 with only one commit.


kchulj:
🟢 so closing.

PR #2423, opened 2021-08-27
Config: Better error message, builtin Duration parsing, SetInfo fields

Geod24:
Also... COLORS!

This is what the new error message looks like with `agora-config-dumper`:
Screen Shot 2021-08-27 at 20 45 40

And with  plain Agora:
Screen Shot 2021-08-27 at 20 53 04

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -37,25 +37,30 @@ node:
   stats_listening_port: 9110
   # The new block time offset has to be greater than the previous block time offset
   # but less than current time + block_time_offset_tolerance
-  block_time_offset_tolerance_seconds: 60
+  block_time_offset_tolerance:
+    seconds: 60
   # The duration between requests for doing periodic network discovery
-  network_discovery_interval_seconds: 5
+  network_discovery_interval:
+    seconds: 5
   # The duration between requests for retrieving the latest blocks
   # from all other nodes
-  block_catchup_interval_seconds: 20
+  block_catchup_interval:
+    seconds: 20
   # The maximum number of transactions relayed in every batch.
   # Value 0 means no limit.
   relay_tx_max_num : 100
   # Transaction relay batch is triggered in every `relay_tx_interval`.
   # Value 0 means, the transaction will be relayed immediately.
-  relay_tx_interval_seconds : 30
+  relay_tx_interval:
+    seconds : 30
   # The minimum amount of fee a transaction has to have to be relayed.
   # The fee is adjusted by the transaction size:
   # adjusted fee = fee / transaction size in bytes.
   relay_tx_min_fee : 0
   # Transaction put into the relay queue will expire, and will be removed
   # after `relay_tx_cache_exp`.
-  relay_tx_cache_exp_seconds : 1200
+  relay_tx_cache_exp:
+    seconds : 1200
   # The percentage by which the double spend transaction's fee should be
   # increased in order to be added to the transaction pool
   double_spent_threshold_pct: 20
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -48,19 +48,19 @@ node:
     seconds: 20
   # The maximum number of transactions relayed in every batch.
   # Value 0 means no limit.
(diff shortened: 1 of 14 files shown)


source/agora/common/Config.d
```diff
@@ -381,7 +381,14 @@ private FR.Type parseField (alias FR)
     if (node.nodeID == NodeID.invalid)
         throw new Exception(format("Node type is invalid: %s", node));
 
-    static if (hasConverter!(FR.Ref))
+    // If we reached this, it  means the field is set, set just recurse
```
hewison-chris:
```suggestion
    // If we reached this, it means the field is set so just recurse
```

source/agora/common/ConfigAttributes.d
```diff
@@ -112,6 +112,66 @@ public struct Name
...
+        public @Optional long hours;
+        public @Optional long minutes;
+        public           long seconds = 42;
+        public @Optional long msecs;
+        public @Optional long usecs;
+        public @Optional long hnsecs;
+        public @Optional long nsecs;
+    }
+    ```
+    In  this case, it would be impossible to know if any field was explicitly
```
hewison-chris:
```suggestion
    In this case, it would be impossible to know if any field was explicitly
```

doc/config.example.yaml
```diff
@@ -35,25 +37,25 @@ node:
   stats_listening_port: 9110
   # The new block time offset has to be greater than the previous block time offset
   # but less than current time + block_time_offset_tolerance_secs
-  block_time_offset_tolerance_secs: 60
+  block_time_offset_tolerance_seconds: 60
```
hewison-chris: In my opinion using the new approach for all `Durations` is better as it not only makes it consistent, the names are shorter, and it removes the need to debate if a new config field should be limited to a certain unit.

source/agora/common/Config.d
```diff
@@ -47,6 +47,51 @@
...
+          minutes: 510
+      ```
+      Provided that the definition of the field is:
+      ```
+      public Duration sleepFor;
+      ```
+
+      In the second form, the field should have a suffix composed of an
+      underscore ('_'), followed by a unit name as defined in `core.time`.
+      This can be either the field name directly,  or a name override.
```
hewison-chris:
```suggestion
      This can be either the field name directly, or a name override.
```

hewison-chris (approved):
Very good, I just think we should consider if the second form of `Durations` is really needed.

doc/config.example.yaml
Geod24: I'll add a commit to switch them all. But I think in some situations, it might make sense to restrict things to a certain unit. Also, it wouldn't surprise me if some people had the exact opposite PoV. Since I want to isolate the config framework in its own repository, and the work is already there, perhaps let's keep it but not use it ?

hewison-chris: Yes I was going to suggest that perhaps by convention we do not use in our `Agora` to keep it consistent, but I agree it might be useful for other projects to have this option.

Geod24:
Updated with requested fixes (+ extra commit)

doc/config.example.yaml
```diff
@@ -37,25 +37,30 @@ node:
...
+  block_catchup_interval:
+    seconds: 20
   # The maximum number of transactions relayed in every batch.
   # Value 0 means no limit.
   relay_tx_max_num : 100
   # Transaction relay batch is triggered in every `relay_tx_interval`.
   # Value 0 means, the transaction will be relayed immediately.
-  relay_tx_interval_seconds : 30
+  relay_tx_interval:
+    seconds : 30
```
kchulj: I think we should unify the syntax of the `Duration`.
`seconds: 30` or  `seconds : 30` ?

linked0 (approved):
LGTM

doc/config.example.yaml
Geod24: Added a commit since the problem was pre-existing and affected other entries.

kchulj (approved):

PR #2427, opened 2021-08-27
Flash test: Simplify usage of identities

Geod24:
```
When the Flash layer started, we had a mix of types for keys.
This was unified under a single KeyPair type, which contains
the PublicKey (address) and SecretKey (secret) under the hood.
Those types themselves are wrappers around Point and Scalar,
respectively. The Flash layer used to deal with the lower layer
types because it needed to perform certain cryptographic operations
which were not available to the old SecretKey / PublicKey.
We can now greatly simplify this code by removing any mention
of this past difference. However, since some vanity keys do not
exists (such as a key starting with 'B'), there is no Bob.
As a result, Bob was renamed to Charlie, and Charlie was renamed to Diego.
Since Alice, Bob, Charlie, Diego and co are just placeholders for A, B, C, D,
we use WK.Keys.{A,B,C,D} whenever possible, and the longer names for variables.
```

### DIFF
--- a/source/agora/test/Flash.d
+++ b/source/agora/test/Flash.d
@@ -377,7 +377,7 @@ public class FlashNodeFactory (FlashListenerType = FlashListener)
 
     /// Create a new flash node user
     public RemoteAPI!TestFlashAPI create (FlashNodeImpl = TestFlashNode)
-        (const Pair pair, string agora_address,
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

Issue #2363, opened 2021-08-08
Unify / clear up the REST API


Geod24:
Currently, we have multiple conventions used in the REST API, and most of them have problem, e.g.:
- `putTransaction`: Usage of `PUT` is wrong here, as a `PUT` should send a specific resource. But we just "submit" a transaction, so we should use `POST`. Something like `PUT /transaction/${TXID}` would be correct though. [More info](https://stackoverflow.com/questions/630453/what-is-the-difference-between-post-and-put-in-http#:~:text=POST%20means%20%22create%20new%22%20as,the%20server%20to%20create%20it.).
- `getEnrollment`: This gets the pending enrollments, not the externalized ones. I don't know if we want to provide an API to provide a specific enrollment that has already be externalized, but I think it might be confusing that this only returns enrollments in the pool;
- There is no way to get all pending enrollments (similar to https://github.com/bosagora/agora/issues/1481);
- `receivePreimage` is `POST reveice_preimage` => Should be `POST /preimage` or even `PUT /preimage/${UTXO_ID}`;
- `getTransactions` => Like `getEnrollment`, should indicate that they are pending txs. Also, #1481

There are probably other issues there, so the cleanup should be done by someone familiar with REST.
For anyone not familiar, [this is pretty great](https://www.ics.uci.edu/~fielding/pubs/dissertation/rest_arch_style.htm).

Geod24:
Done in #2428 (mostly), #2511 and #2519 .

PR #2428, opened 2021-08-28
Part of #2363: Rename and organize FullNode & Validator APIs

Geod24:
This is the first half of #2363 which also mentions that the names used for getting txs and enrollments aren't great, since a user might expect to get an externalized tx rather than one in the pool.
The main part is a major rename for consistency. See the commit message for a detailed explanation about the whys and whats.

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -119,24 +119,19 @@ public interface API
     /***************************************************************************
 
         Returns:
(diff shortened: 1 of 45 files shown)


Geod24:
Reinstated `putTransaction` endpoint so that Faucet builds.

kchulj (approved):
🟢 and LGTM.
Very nice explanation.

PR #2429, opened 2021-08-28
Reduce ManagedDatabase usage, improve shutdown, move SCPEnvelopeStore

Geod24:
This PR does 3 different things in 3 commits, but since they followed from one another, and they would conflict otherwise, I left them in a single PR.
I was looking at the SEGV which triggers while shutting down in the unittests, and came across `ManagedDatabase`.
`ManagedDatabase` is a global object which holds a list of databases.
Except that now we only have two databases objects, unlike before when we opened one DB per table, so we could get rid of it. There are only two users anyways: Ledger and `SCPEnvelopeStore`. The later was trivial to change, so I did that. In the process I fixed an old pet peeve, which is that `SCPEnveloperStore` was living in `consensus` while everything SCP related is in `protocol`.


source/agora/consensus/protocol/EnvelopeStore.d
```diff
@@ -56,6 +56,46 @@ public class SCPEnvelopeStore
             "(seq INTEGER PRIMARY KEY AUTOINCREMENT, envelope BLOB NOT NULL, processed INTEGER NOT NULL)");
     }
 
+    /***************************************************************************
+
+        Lock / Unlock the underlying database for updates
+
+        This stores uses the cacheDB for storing the envelopes accross restart.
```
hewison-chris:
```suggestion
        This store uses the cacheDB for storing the envelopes across restarts.
```

source/agora/consensus/protocol/EnvelopeStore.d
```diff
@@ -56,6 +56,46 @@ public class SCPEnvelopeStore
...
+        When performing an update, one needs to first lock the DB first
```
hewison-chris:
```suggestion
        When performing an update, one needs to first lock the DB
```

hewison-chris (approved):
LGTM just a couple of typos in comments.

Geod24:
Fixed

PR #2430, opened 2021-08-30
Fee: Minor cleanup + renamed 'adjusted fee' to 'rate'

Geod24:
Was going through the code for several reason and noticed the use of 'adjusted fee', which I found confusing.
Decided to renamed it to what it actually is, a fee rate.


source/agora/consensus/Fee.d
```diff
@@ -435,29 +435,31 @@ public class FeeManager
...
+        If the fee is not an exact multiplicator of the tx size,
+        the remainder will be ignored.
 
         Params:
-            tx = transaction for which we want to calculate the adjusted fee
-            peekUTXO = UTXO finder
-            tot_fee = adjusted fee
+            tx = transaction for which we want to calculate the fee rate
+            peekUTXO = UTXO finder (with or without replay protection)
+            rate = This transaction minimum fee rate
```
hewison-chris:
```suggestion
            rate = This transaction's fee rate
```

hewison-chris: I know we are ignoring the remainder but it is misleading when we have `minimum fee rate` as `Consensus Param`

source/agora/consensus/Fee.d
```diff
@@ -435,29 +435,31 @@ public class FeeManager
...
     ***************************************************************************/
 
-    public string getAdjustedTXFee (in Transaction tx, scope UTXOFinder peekUTXO,
-        out Amount tot_fee) nothrow @safe
+    public string getTxFeeRate (in Transaction tx, scope UTXOFinder peekUTXO,
+        out Amount rate) nothrow @safe
     {
         try
-            tot_fee = tx.getFee(peekUTXO);
+            rate = tx.getFee(peekUTXO);
```
hewison-chris, line 459: Just a comment.
At first I thought there was an error in this statement. But it is correct as we need to create the `Amount` before we do the division. 
I wish we returned the result in `Amount` calculations so we could avoid confusing code statements like this one. Not part of this PR but I think we could look at making `Amount` easier to work with.

hewison-chris (approved):
LGTM

source/agora/consensus/Fee.d
Geod24: Right, should be better worded.

source/agora/consensus/Fee.d
Geod24: Yep, had the same thought. Will add a comment.

Geod24:
Updated

kchulj (approved):
LGTM
CI needs a re-run.

PR #2433, opened 2021-08-31
Remove unnecessary assert in getNodeUTXO

linked0:
There is no reason for the assertion to occur just because there is no matching UTXO for the nodeID at some height.
So removed the assertion and added more detail to logging.

It's related to [this PR](https://github.com/bosagora/agora/pull/2417#issuecomment-908188362).


Geod24 (approved):

kchulj (approved):
LGTM

kchulj:
`ubuntu-20.04, ldc-master` needs a re-run 😞

PR #2435, opened 2021-08-31
Registry: Add a configuration file

Geod24:
And voilà what it was all for: We can trivially add a configuration file to the registry.
The next step will improve the DNS server itself to dynamically define when it is authoritative (and which zones it should follow).

### DIFF
--- a/source/agora/registry/main.d
+++ b/source/agora/registry/main.d
@@ -13,12 +13,14 @@
 
 module agora.registry.main;
 
+import agora.common.Config;
 import agora.common.DNS;
 import agora.registry.API;
 import agora.registry.Config;
 import agora.registry.Server;
 import agora.serialization.Serializer;
 import agora.stats.Server;
+import agora.utils.Log;
 
 import vibe.core.core;
 import vibe.core.net;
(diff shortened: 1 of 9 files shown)


kchulj:
Is there a plan to add a `yaml` file?

Geod24:
If the config file grows too big, yes. So far it's rather small.

Geod24:
I've run the config-dumper on https://github.com/bosagora/infrastructure/pull/171 though and it went as expected.
I also added a few more sections just to make sure, seems to work just fine.

source/agora/registry/main.d
```diff
@@ -84,16 +105,17 @@ private int main (string[] strargs)
...
 *******************************************************************************/
 
-private void runDNSServer_canThrow (NameRegistry registry)
+private void runDNSServer_canThrow (in Config config, NameRegistry registry)
 {
     // The `listenUDP` needs to be in the `runTask` otherwise we get
     // a fatal error due to a bug in vibe-core (see comment #2):
     /// https://github.com/vibe-d/vibe-core/issues/289
-    auto udp = listenUDP(53);
+    auto udp = listenUDP(config.dns.port, config.dns.address);
```
kchulj, line 118: why did this become two parameters from one?

Geod24: Because the default value is `"0.0.0.0"`, and now I've made it configurable.
https://vibed.org/api/vibe.core.net/listenUDP

kchulj (approved):
LGTM

linked0 (approved):
LGTM

PR #2437, opened 2021-09-02
Config fixes: Special case 'root' name & pass a YAML `Node` to `Converter`

Geod24:
See the rationale in each commit.


Geod24:
Note that those two commits were created while working on two unrelated piece of work, but since they are both config-related, and trivial, I put them in the same PR to save CI time. LMK if anyone prefer to have them split.

kchulj:
I think same PR is ok since they are both related to `Config`.

kchulj:
Is the CI failure legit..?

Geod24:
Given that both failures are in Flash, I doubt so. The third one is a timeout, and one of the 4 passed. So I re-ran it.

linked0 (approved):

kchulj (approved):

PR #2441, opened 2021-09-02
Restore enrollments from other nodes

### LINKED ISSUE #755: Recover enrollments in abnormal failure scenarios
We have a recovery code for validators that restores the information from the chain. A node can restore the validators' information from the ledger by reading the `enrollments` section of a header in a block. But we don't have a recovery code for enrollments, or candidates waiting to be a validator. The information can be restored from requesting other nodes for that, which is the best solution for that. But, there might be a better solution.

Definition of Done:

- [x] Get the information of enrollments from other nodes
- [x] Save the information into the `EnrollmentPool`

How to implement:

- Add an API endpoint getting the information from other nodes.
- Determine the time when the API must be used.
- Save the information into the 'EnrollmentPool' class.
- Consider an interface or an abstract class for processing abnormal situations.

linked0:
Added the API endpoint called `getEnrollments` to get missing enrollments from other nodes and make the network manager periodically call the function on other nodes to add received enrollments into the enrollment pool.

Fixes #755

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -364,4 +364,21 @@ public interface API
 
     @method(HTTPMethod.GET)
     public Enrollment getEnrollment (in Hash enroll_hash);
+
+    /***************************************************************************
+
+        API:
+            GET /enrollments
+
+        Params:
+            height = the height of proposed block
+            exclude_enrolls = the enrollments that should be excluded
+
+        Returns:
+            All the enrollments for the height in the enrollment pool
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.GET)
+    public Enrollment[] getEnrollments (ulong height, Set!uint exclude_enrolls);
 }
--- a/source/agora/consensus/EnrollmentManager.d
+++ b/source/agora/consensus/EnrollmentManager.d
@@ -46,6 +46,7 @@
 module agora.consensus.EnrollmentManager;
 
 import agora.common.ManagedDatabase;
+import agora.common.Set;
 import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.Enrollment;
@@ -114,6 +115,15 @@ public class EnrollmentManager
 
     /// Parameters for consensus-critical constants
     private immutable(ConsensusParams) params;
+
+    /// Current height
+    private Height current_height;
+
+    /// Indices of Frozen UTXOs which are sorted by hashes of UTXOs
+    private uint[Hash] utxo_key_to_idx;
+
+    /// Enrollments of Frozen UTXOs by index
+    private Enrollment[uint] utxo_idx_to_enroll;
 
     /***************************************************************************
 
@@ -222,8 +232,15 @@ public class EnrollmentManager
             return false;
         }
 
-        return this.enroll_pool.add(enroll, Height(available), finder,
+        auto result = this.enroll_pool.add(enroll, Height(available), finder,
                                     &this.validator_set.findRecentEnrollment);
+
+        // Update the map of index to UTXO if the UTXO is already frozen
+        if (result && this.current_height + 1 == Height(available) &&
+            enroll.utxo_key in this.utxo_key_to_idx)
+            this.utxo_idx_to_enroll[this.utxo_key_to_idx[enroll.utxo_key]] = enroll;
+
+        return result;
     }
 
     /***************************************************************************
@@ -732,6 +749,85 @@ public class EnrollmentManager
     public EnrollmentFinder getEnrollmentFinder () @trusted nothrow
     {
         return &this.validator_set.findRecentEnrollment;
+    }
+
+    /***************************************************************************
+
+        Undate the map of frozen UTXOs
+
+        Params:
+            height = the current block height
+            utxos =  the frozen UTXOs from the utxo set
+
+    ***************************************************************************/
+
+    public void updateFrozenUTXO (in Height height, in UTXO[Hash] utxos)
+        @trusted nothrow
+    {
+        uint idx = 0;
+        this.current_height = height;
+        this.utxo_key_to_idx.clear();
+        this.utxo_idx_to_enroll.clear();
+
+        utxos.keys.sort.each!((key) {
+            this.utxo_key_to_idx[key] = idx;
+            this.utxo_idx_to_enroll[idx] = this.enroll_pool.getEnrollment(key, height);
+            idx++;
+        });
+    }
+
+    /***************************************************************************
+
+        Get the indices of enrollments of the node and future ones
+
+        Params:
+            height = the height intended for the enrollments
+            enroll_set = the indices of enrollments of the node
+
+    ***************************************************************************/
+
+    public void getEnrollmentIndices (Height height, ref Set!uint enroll_set)
+        @safe nothrow
+    {
+        try
+        {
+            foreach (idx, value; this.utxo_idx_to_enroll)
+            {
+                if (value != Enrollment.init)
+                    enroll_set.put(idx);
+            }
+        }
+        catch (Exception ex)
+        {
+            log.warn("Caught Exception while calling getEnrolmentIndices {}", ex);
+        }
+    }
+
+    /***************************************************************************
+
+        Get all the enrollments excluding the unwanted enrollments
+
+        Params:
+            height = the height intended for the enrollments
+            exclude_enrolls = the indices of unwanted enrollments
+
+        Returns:
+            The enrollments excluding the unwanted and `future` ones
+
+    ***************************************************************************/
+
+    public Enrollment[] getExclusiveEnrollments (in Height height,
+        Set!uint exclude_enrolls) @safe nothrow
+    {
+        Enrollment[] enrolls;
+        Set!uint enroll_set;
+        getEnrollmentIndices(height, enroll_set);
+
+        exclude_enrolls.each!(idx => enroll_set.remove(idx));
+        foreach (idx; enroll_set)
+            enrolls ~= utxo_idx_to_enroll[idx];
+
+        return enrolls;
     }
 }
 
--- a/source/agora/consensus/pool/Enrollment.d
+++ b/source/agora/consensus/pool/Enrollment.d
@@ -283,6 +283,43 @@ public class EnrollmentPool
 
     /***************************************************************************
 
+        Get the enrollment with the key and the height
+
+        Params:
+            enroll_hash = key for the enrollment which has the frozen UTXO
+            height = the height of proposed block
+
+        Returns:
+            Return an `Enrollment` if the enrollment is found, otherwise
+                `Enrollment.init`
+
+    ***************************************************************************/
+
+    public Enrollment getEnrollment (in Hash enroll_hash, in Height height)
+        @trusted nothrow
+    {
+        try
+        {
+            auto results = this.db.execute("SELECT key, val FROM enrollment_pool " ~
+                "WHERE key = ? AND avail_height = ?", enroll_hash, height.value);
+
+            foreach (row; results)
+            {
+                return deserializeFull!Enrollment(row.peek!(ubyte[])(1));
+            }
+        }
+        catch (Exception ex)
+        {
+            log.error("Exception occured on getEnrollment: {}, " ~
+                "Key for enrollment: {}", ex.msg, enroll_hash);
+            return Enrollment.init;
+        }
+
+        return Enrollment.init;
+    }
+
+    /***************************************************************************
+
         Get the height when the enrollment will be available
 
         Params:
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -31,10 +31,13 @@ import agora.common.ManagedDatabase;
 import agora.common.Set;
 import agora.common.Task;
 import agora.consensus.data.Block;
+import agora.consensus.data.Enrollment;
 import agora.consensus.data.Params;
 import agora.consensus.data.PreImageInfo;
 import agora.consensus.data.UTXO;
 import agora.consensus.data.ValidatorBlockSig;
+import agora.consensus.EnrollmentManager;
+import agora.consensus.state.UTXOCache;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.network.Clock;
@@ -1046,6 +1049,29 @@ public class NetworkManager
             log.error("getMissingBlockSigs: Exception thrown : {}", e.msg);
         }
         return headers_updated;
+    }
+
+    /// Retrieve any missing enrollments
+    public void getEnrollments (Height height, EnrollmentManager enroll_man,
+        scope UTXOFinder finder) @safe nothrow
+    {
+        auto this_enrolls = enroll_man.enroll_pool.getEnrollments(height);
+        Set!uint exclude_indices;
+        enroll_man.getEnrollmentIndices(height, exclude_indices);
+
+        foreach (peer; this.peers)
+        {
+            auto peer_enrolls = peer.client.getEnrollments(height, exclude_indices);
+            foreach (enroll ; peer_enrolls)
+            {
+                if (this_enrolls.canFind(enroll))
+                    continue;
+
+                if (enroll_man.enroll_pool.add(enroll, height, finder,
+                        enroll_man.getEnrollmentFinder()))
+                    this_enrolls ~= enroll;
+            }
+        }
     }
 
     /// Shut down timers & dump the metadata
--- a/source/agora/node/Config.d
+++ b/source/agora/node/Config.d
@@ -231,6 +231,9 @@ public struct NodeConfig
     /// from all other nodes
     public Duration block_catchup_interval = 20.seconds;
 
+    /// The duration between requests for retrieving the missing enrollments
+    public Duration enrollment_catchup_interval = 40.seconds;
+
     /// The new block time offset has to be greater than the previous block time offset,
     /// but less than current time + block_time_offset_tolerance
     public Duration block_time_offset_tolerance = 60.seconds;
@@ -641,12 +644,15 @@ unittest
 node:
   data_dir: .cache
   commons_budget_address: boa1xrzwvvw6l6d9k84ansqgs9yrtsetpv44wfn8zm9a7lehuej3ssskxth867s
+  enrollment_catchup_interval:
+    seconds: 60
 network:
  - "something"
 `;
     auto config = parseConfigString!Config(conf_example, "/dev/null");
     assert(config.node.min_listeners == 2);
     assert(config.node.max_listeners == 10);
+    assert(config.node.enrollment_catchup_interval == 60.seconds);
     assert(config.node.data_dir == ".cache");
     assert(config.node.commons_budget_address.toString() == "boa1xrzwvvw6l6d9k84ansqgs9yrtsetpv44wfn8zm9a7lehuej3ssskxth867s");
 }
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -421,6 +421,8 @@ public class FullNode : API
             this.config.node.network_discovery_interval, &this.discoveryTask, Periodic.Yes);
         this.timers ~= this.taskman.setTimer(
             this.config.node.block_catchup_interval, &this.catchupTask, Periodic.Yes);
+        this.timers ~= this.taskman.setTimer(
+            this.config.node.enrollment_catchup_interval, &this.catchupEnrollmentTask, Periodic.Yes);
 
         // Immediately run discovery to avoid delays at startup
         this.taskman.runTask(&this.discoveryTask);
@@ -507,6 +509,22 @@ public class FullNode : API
         {
             log.error("Error sending updated block headers:{}", e);
         }
+    }
+
+    /***************************************************************************
+
+        Periodically retrieve the missing enrollments and apply them to the
+        enrollment manager.
+
+    ***************************************************************************/
+
+    protected void catchupEnrollmentTask () nothrow
+    {
+        if (this.network.peers.empty())  // no clients yet (discovery)
+            return;
+
+        this.network.getEnrollments(this.ledger.getBlockHeight() + 1, 
+            this.enroll_man, &this.utxo_set.peekUTXO);
     }
 
     /***************************************************************************
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -1059,6 +1059,16 @@ public class FullNode : API
         return this.enroll_man.getEnrollment(enroll_hash);
     }
 
+    /// GET: /enrollments
+    public override Enrollment[] getEnrollments (ulong height, Set!uint exclude_enrolls)
+        @safe
+    {
+        if (height != this.getBlockHeight() + 1)
+            return Enrollment[].init;
+
+        return this.enroll_man.getExclusiveEnrollments(Height(height), exclude_enrolls);
+    }
+
     /// POST /preimage
     public override void postPreimage (in PreImageInfo preimage) @safe
     {
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -2024,6 +2024,9 @@ public struct TestConf
         // Catchup needs to happens much more frequently than in production
         block_catchup_interval: 1.seconds,
 
+        // Catchup period for missing enrollments
+        enrollment_catchup_interval: 1.seconds,
+
         // The default is much longer, but in unittests latency is negligible
         retry_delay: 300.msecs,
         timeout: 300.msecs,
--- a/source/agora/test/GetEnrollments.d
+++ b/source/agora/test/GetEnrollments.d
@@ -1,0 +1,129 @@
+/*******************************************************************************
+
+    Contains tests for getting missing enrollments from the network
+
+    Copyright:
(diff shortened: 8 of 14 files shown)


source/agora/network/Manager.d
```diff
@@ -1063,6 +1066,39 @@ public class NetworkManager
         return headers_updated;
     }
 
+    /// Retrieve any missing enrollments from randomly selected nodes
+    public void getEnrollments (EnrollmentManager enroll_man,
+        scope UTXOFinder finder) @safe nothrow
+    {
+        // TODO: This should be part of the configuration.
+        static uint MinNodes = 3;
```
linked0: I thought that getting enrollments from all other nodes could impose a burden on the network. So I'm sampling the nodes which will get the request, or `getEnrollments`. @Geod24  How do you think of this approach?

source/agora/consensus/data/Enrollment.d
```diff
@@ -70,6 +70,21 @@ public struct Enrollment
...
+*******************************************************************************/
+
+public struct PoolEnrollment
+{
+    /// An enrollment data in the pool
+    public Enrollment enrollment;
+
+    /// The available height set on being stored in the pool
+    public Height avail_height;
+}
```
Geod24: I don't follow why we need to get the `avail_height` from other nodes ?

source/agora/api/FullNode.d
```diff
@@ -359,4 +359,17 @@ public interface API
...
+
+        API:
+            GET /enrollments
+
+        Returns:
+            All the enrollments in the enrollment pool
+
+    ***************************************************************************/
+
+    @method(HTTPMethod.GET)
```
Geod24, line 382: This is the default

source/agora/test/GetEnrollments.d
```diff
@@ -0,0 +1,80 @@
+module agora.test.GetEnrollments;
```
Geod24, line 14: DDOC

source/agora/node/FullNode.d
```diff
@@ -500,6 +500,7 @@ public class FullNode : API
         {
             log.error("Error sending updated block headers:{}", e);
         }
+        this.network.getEnrollments(this.enroll_man, this.utxo_set.getUTXOFinder());
```
Geod24: I think it should be its own task, so that it doesn't starve because of block catchup

source/agora/network/Manager.d
Geod24: Getting all enrollments also means that we most likely always have needless data transfer even when we're up to date, which isn't optimal.

linked0: OK, I see what you mean. I will find a way to store missing enrollments that will be requested later from other nodes.

source/agora/consensus/data/Enrollment.d
linked0: The `avail_height` is needed to store enrollments into the pool because a requesting node could have a different height and can not set the exact value for the field.

source/agora/node/FullNode.d
linked0: Modified.

source/agora/consensus/data/Enrollment.d
linked0: Anyway, I'll look into more whether the field could be removed.

source/agora/api/FullNode.d
linked0: Removed

source/agora/test/GetEnrollments.d
linked0: Added DDOC

source/agora/network/Manager.d
linked0: Had a different approach and solved it.

source/agora/consensus/data/Enrollment.d
linked0: I think it's needed. So I remain the field as it is.

linked0:
Ready for review

source/agora/node/Config.d
```diff
@@ -224,6 +224,10 @@ public struct NodeConfig
     /// from all other nodes
     public Duration block_catchup_interval = 20.seconds;
 
+    /// The number of the other nodes from which a node can retrive missing
```
hewison-chris: I think you need to update this comment.

source/agora/node/FullNode.d
```diff
@@ -1043,6 +1043,19 @@ public class FullNode : API
         return this.enroll_man.getEnrollment(enroll_hash);
     }
 
+    /// GET: /enrollments
+    public override PoolEnrollment[] getEnrollments (Set!Hash enroll_keys) @safe
```
hewison-chris: Please add to source/api/spec.yaml.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -111,6 +112,9 @@ public class EnrollmentManager
     /// Ditto
     private PreImageCycle cycle;
 
+    /// Missing enrollments appearing from a failure in adding an enrollment
```
hewison-chris:
```suggestion
    /// Missing enrollments discovered from other nodes
```

source/agora/test/GetEnrollments.d
```diff
@@ -0,0 +1,138 @@
...
+    assert(gotten_enroll == Enrollment.init);
+
+    bad_node.clearFilter();
+
+    // enable `catchupTask` for the bad validator and
+    // make sure the bad validator is up to date
+    CustomAPIManager.enable_catchup = true;
+    network.expectHeight([GenesisValidators - 1], Height(GenesisValidatorCycle - 1));
+
+    Thread.sleep(conf.node.enrollment_catchup_interval);
```
hewison-chris: How long will this sleep? I think the default is too long for the test and should be overridden to maybe 500.msecs?

hewison-chris: Also I think it is better to use retry with timeout rather than sleep.

hewison-chris (changes requested):
Looks good just a few requests.

source/agora/node/Config.d
linked0: Oh, I missed it. Will update.

source/agora/node/FullNode.d
linked0: Will update.

source/agora/consensus/EnrollmentManager.d
linked0: Will update.

source/agora/test/GetEnrollments.d
linked0: I will try to use retry.

source/agora/consensus/EnrollmentManager.d
hewison-chris:
```suggestion
    /// Set to store `utxo hashes` of enrollments that fail to be added when initially received
```

source/agora/test/Base.d
```diff
@@ -2006,6 +2006,9 @@ public struct TestConf
         // Catchup needs to happens much more frequently than in production
         block_catchup_interval: 1.seconds,
 
+        // Catchup period for missing enrollments
+        enrollment_catchup_interval: 1.seconds,
```
hewison-chris, line 2028: 👍

source/agora/test/GetEnrollments.d
```diff
@@ -0,0 +1,137 @@
...
+
+    auto gotten_enroll = bad_node.getEnrollment(enroll.utxo_key);
+    assert(gotten_enroll == Enrollment.init);
+
+    // make sure the bad validator is up to date    
+    network.expectHeight([GenesisValidators - 1], Height(GenesisValidatorCycle - 1));
+
+    // check the missing enrollment to be restored
+    retryFor(bad_node.getEnrollment(enroll.utxo_key) == enroll,
+        5.seconds, format!"The enrollment (%s) not in pool of the bad validator"
```
hewison-chris:
```suggestion
        2 * conf.node.enrollment_catchup_interval, format!"The enrollment (%s) not in pool of the bad validator"
```

hewison-chris: It should happen within 2 intervals.

hewison-chris (approved):

source/agora/node/Config.d
linked0: Updated.

source/agora/node/FullNode.d
linked0: Added.

source/agora/test/GetEnrollments.d
linked0: Used retry. It's better.

source/agora/consensus/EnrollmentManager.d
linked0: Modified with a shorter message.

source/agora/test/GetEnrollments.d
linked0: Updated. Ty.

source/agora/consensus/data/Enrollment.d
Geod24: The reason I'm asking is because we cannot trust the remote node to provide the right information.
What if the node sends us a right enrollment but pretends it will be available in 3 million blocks ?

source/agora/consensus/pool/Enrollment.d
```diff
@@ -17,6 +17,7 @@
 module agora.consensus.pool.Enrollment;
 
 import agora.common.ManagedDatabase;
+import agora.common.Set;
```
Geod24: Obviously this is not used in this commit.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -738,6 +771,42 @@ public class EnrollmentManager
...
+        Set!Hash local_missing_enrolls;
+        foreach (key; this.missing_enrolls)
+        {
+            if (this.getEnrollment(key) == Enrollment.init)
+                local_missing_enrolls.put(key);
+        }
+
+        this.missing_enrolls = local_missing_enrolls;
+        return missing_enrolls;
+    }
```
Geod24: Why the `Set` duplication ?

source/agora/node/FullNode.d
```diff
@@ -1026,7 +1043,12 @@ public class FullNode : API
         this.recordReq("postEnrollment");
 
         UTXO utxo;
-        this.utxo_set.peekUTXO(enroll.utxo_key, utxo);
+        if (!this.utxo_set.peekUTXO(enroll.utxo_key, utxo))
+        {
+            this.enroll_man.addMissingEnrollment(enroll.utxo_key);
```
Geod24: Wait that is not a "missing enrollment", it's a missing UTXO referenced in an enrollment

source/agora/node/FullNode.d
```diff
@@ -502,6 +504,21 @@ public class FullNode : API
...
+        enrollment manager.
+
+    ***************************************************************************/
+
+    protected void catchupEnrollmentTask () nothrow
+    {
+        if (this.network.peers.empty())  // no clients yet (discovery)
+            return;
+
+        this.network.getEnrollments(this.enroll_man, this.utxo_set.getUTXOFinder());
```
Geod24: This looks very dangerous. We can't really use the UTXO finder in tasks as it is not re-entrant.
Meaning that the following may occur:
```
Task A is called, starts to use the UTXOFinder, and yield mid-task;
Task B is called, starts to use the UTXOFinder, which resets the list of double spend, and yield mid-task;
Task A resumes, continues using the UTXOFinder, unaware that it lost replay protection.
```

source/agora/network/Manager.d
```diff
@@ -1063,6 +1066,44 @@ public class NetworkManager
         return headers_updated;
     }
 
+    /// Retrieve all the missing enrollments from the connected nodes
+    public void getEnrollments (EnrollmentManager enroll_man,
+        scope UTXOFinder finder) @safe nothrow
+    {
+        NodeConnInfo[] conns = this.peers.array;
+        ulong[] conn_indices = conns.length.iota.array;
```
Geod24: So this essentially creates an array being `[0, 1, 2, 3, ...]` ?
And it *only* uses it for a `foreach`. We should try to avoid pointless memory allocations like this.
A simple way to do this is:
```D
foreach (idx; 0 .. conns.length)
```

But then, the only use of `idx` in that loop is for `conns[idx]`, so why not simply:
```D
foreach (conn; this.peers)
```

And avoid two superfluous memory allocations ?

source/agora/network/Manager.d
```diff
@@ -1063,6 +1066,44 @@ public class NetworkManager
...
+        // Retrieve missing enrollments from other nodes
+        foreach (idx; conn_indices)
+        {
+            if (missing_enrolls.length == 0)
+                break;
+
+            auto pool_enrolls = conns[idx].client.getEnrollments(missing_enrolls);
+            foreach (enroll; pool_enrolls)
+            {
+                import std.stdio;
```
Geod24: Leftover

source/agora/test/GetEnrollments.d
```diff
@@ -0,0 +1,137 @@
...
+            this.enable_catchup = enable_catchup;
+            super(args);
+        }
+
+        ///
+        protected override void catchupTask () nothrow
+        {
+            if (atomicLoad(*this.enable_catchup))
+                super.catchupTask();
+        }
```
Geod24, line 49: Why do we override `catchupTask` and not `catchupEnrollment` ?

Geod24 (changes requested):
I don't understand the logic of the last commit. This is not catching up `Enrollment`, since the only place where `addMissingEnrollment` is called is when we receive an `Enrollment` that we don't have an UTXO for. So when we do so, we'll reject it, and ask other nodes to send it to us again... What am I missing ?

source/agora/consensus/EnrollmentManager.d
linked0: The `missing_enrollments` will be used later so it must have the latest missing enrollments. If you meant another way, please let me know.

source/agora/node/FullNode.d
linked0: I see what you mean. I will try to save missing `Enrollment`s instead of UTXOs.

source/agora/node/FullNode.d
linked0: Right, I missed it, I will use peekUTXO instead.

source/agora/test/GetEnrollments.d
linked0: It's because I wanna make the bad node one block behind the latest height.

source/agora/consensus/data/Enrollment.d
linked0: After considering your opinion. I could use a new table for missing enrollments that have `Enrollment` and `avail_height`. How do you think of creating new table for that? I mean nodes can just communicate with `Enrollment`, not `PoolEnrollment`.

linked0: Anyway, I will find a way to use `Enrollment` in the API endpoint, `getEnrollments`, and remove this structure.

source/agora/consensus/EnrollmentManager.d
Geod24: I am not sure why we need a local `Set` for this. This seems to replicate what was done with "missing transactions", but the problem was different: Missing transactions uses a `Set` because we see the hashes in the SCP statement but we can't match a data structure. On the other hand, we haven't seen the hashes of `Enrollment` that are missing, we don't even know if they exist or not.

linked0: I see. I think we need to store `Enrollment`s for the missing enrollments instead of hashes as commented on other threads. For `Set` duplication, I didn't know that `remove` function can be used in the iteration. I understand the local variable is not needed now. Thx for pointing out the problem.

source/agora/node/FullNode.d
linked0: Updated. It uses `peekUTXO` now.

source/agora/network/Manager.d
linked0: Removed.

source/agora/network/Manager.d
linked0: Updated.

source/agora/consensus/pool/Enrollment.d
linked0: Removed.

linked0:
I applied a table for orphan enrollments and pushed the code in this PR, but @Geod24 gave an opinion that it's not in fact needed IRL.
@Geod24 Could you explain it in more detail?

linked0:
@Geod24 Do you totally disagree with this approach? Could you review this once?

source/agora/consensus/pool/Enrollment.d
```diff
@@ -191,6 +245,28 @@ public class EnrollmentPool
...
+    /***************************************************************************
+
+        Remove the orphan enrollment with the given key from the orphan pool
+
+        Params:
+            enroll_hash = key for an enrollment data to remove
+
+    ***************************************************************************/
+
+    public void removeOrphan (in Hash enroll_hash) @trusted nothrow
```
linked0: Additionally, we need the function named `removeOrphanExpired` or to add the logic removing orphan enrollments to the `removeExpired` function.

Geod24:
From what I can see from the diff:
- Whenever we receive an `Enrollment` which does not pass validation, we store it in a table, regardless of **why** is wasn't valid (https://github.com/bosagora/agora/pull/2441/files#diff-71a378ff4dee5d52c4fd906503085d5635555f0ed418e138439cb7f78dc575dbR96);
- We only remove entries from this table if we add a validated enrollment (https://github.com/bosagora/agora/pull/2441/files#diff-71a378ff4dee5d52c4fd906503085d5635555f0ed418e138439cb7f78dc575dbR153);
- Although, we also override data if the same UTXO is used (https://github.com/bosagora/agora/pull/2441/files#diff-71a378ff4dee5d52c4fd906503085d5635555f0ed418e138439cb7f78dc575dbR188-R192);
- We have a periodic table which allocates a whole AA with the content of this database (https://github.com/bosagora/agora/pull/2441/files#diff-7b4de96b302d2ab907d0084dd3697c8dbd7ea2359812b46ec950c0e3c94a17c7R1077);
- We request all our peers for the enrollments we already know about (???): https://github.com/bosagora/agora/pull/2441/files#diff-7b4de96b302d2ab907d0084dd3697c8dbd7ea2359812b46ec950c0e3c94a17c7R1091

Provided the above is correct, there are many problems that would ensue from this:
1) First and foremost, we **know** the content of the `Enrollment` structure. Why are we asking our peers to send us something we know about ?
2) Even if there was a good reason for **1** (I can't see any), we don't take into account **why** validation fails. It could be a wrong signature, in which case, there's no point in storing it;
3) Since we don't take into account the failure reason, this is a glaring DoS vector: One can just send a random piece of data with `utxo_key == 0x00...001`, then `0x.00...002`, etc... and we will store **everything** in the database. We will then quickly run of memory because of the AA allocations, or of disk space because we essentially would store `(Enrollment.sizeof +sqlite3 overhead) * 2^(64-1)`;

linked0:
> From what I can see from the diff:
> ...

I agree with your opinion. Sorry for not thinking to that extent about that in advance. Will think about it more.

linked0:
Ready for review.
I applied another approach which is not perfect but better than the current status. Getting enrollments could make the network congested so I considered making the burden of the network as small as possible by getting enrollments from other nodes only when other nodes have more enrollments.

linked0:
Ready for review.
This is a more improved version that is more compact and removes the unnecessary `PoolEnrollment` structure.

source/agora/node/FullNode.d
```diff
@@ -1062,6 +1062,17 @@ public class FullNode : API
...
+    public override Enrollment[] getEnrollments (ulong height, ulong enroll_count)
+        @safe
+    {
+        auto enrolls = this.enroll_man.enroll_pool.getEnrollments(Height(height));
+        if (enrolls.length > enroll_count)
+            return enrolls;
+        
+        return Enrollment[].init;
+    }
+
```
Geod24: I don't understand this API. I don't see why it would make sense from the point of view of either the client, or the server.

In the client:
- You probably want to set this to a low value (I would just use 0?) because otherwise you might not get any result;

In the server:
- You always allocate, then check the length, then discard the allocated array if the user-provided value is too high;
- If an attacker repeatedly sends you `ulong.max`, you'll just end up allocating like crazy;

So the endpoint is an easy DoS vector, and doesn't make sense to the client either. What am I missing ?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -226,6 +226,27 @@ public class EnrollmentManager
...
+
+    ***************************************************************************/
+
+    public bool addEnrollment (in Enrollment enroll, in Height avail_height,
+        scope UTXOFinder peekUTXO) @safe nothrow
+    {
+        return this.enroll_pool.add(enroll, avail_height, peekUTXO,
+                                    &this.validator_set.findRecentEnrollment);
+    }
+
```
Geod24: I've been trying to remove those trampoline functions from the EM for the past year.
The EM role is to manage a validator's enrollment, not to be an interface for the ValidatorSet.
Well, not anymore, although that was the original design, it proved inadequate.

source/agora/network/Manager.d
```diff
@@ -1044,6 +1056,26 @@ public class NetworkManager
...
+        foreach (peer; this.peers)
+        {
+            auto peer_enrolls = peer.client.getEnrollments(height, enroll_keys.length);
+            foreach (enroll ; peer_enrolls)
+            {
+                if (!enroll_keys.canFind(enroll.utxo_key) && 
+                    enroll_man.addEnrollment(enroll, height, finder))
+                    enroll_keys ~= enroll.utxo_key;
+            }
+        }
```
Geod24: So this code:
1) Get a (newly allocated) array of all the enrollments;
2) Allocate a new array of the same size with just the `utxo_key`;
3) Queries *all peers* (no matter what), and for each peer:
4) Get all their enrollments (but only if they have more enrollments... If they have less, but a different set, no luck);
5) Add each enrollment, after doing a linear search in `enroll_keys`;

Not only is this approach extremely wasteful, as the pool checks if there's already an enrollment for this UTXO, hence the `getEnrollments` => `each` / `canFind` are superfluous, but it also won't work if, for example, we have 5 enrollments, and the other node has 4 **totally different** Enrollments.

source/agora/node/FullNode.d
```diff
@@ -506,6 +506,26 @@ public class FullNode : API
         {
             log.error("Error sending updated block headers:{}", e);
         }
+
+        // try to catch up enrollments only when there some blocks received
```
Geod24: Why ?

Geod24 (changes requested):
As mentioned in the commit, this allocates way too much, but it also won't work.

linked0:
Thanks for the review. Will request a whiteboard after reviewing the comments.

source/agora/consensus/EnrollmentManager.d
linked0: Got it. Removed it. Thx.

source/agora/node/FullNode.d
linked0: Modified the code vulnerable to DoS attack.

source/agora/node/FullNode.d
linked0: It was for efficient networking but I took another approach.

source/agora/network/Manager.d
linked0: Updated.
But there remains the issue that needless data transfer occurs even when we're up to date, which isn't optimal. It might need a whiteboard.

omerfirmak:
Is this really necessary?
Nodes already broadcast everything in their enrollment pool during the nomination phase anyway.

linked0:
> Is this really necessary? Nodes already broadcast everything in their enrollment pool during the nomination phase anyway.

Yes, it is because we need to synchronize the enrollment pool as far as possible before the nomination phase.

linked0:
Ready for review.

omerfirmak:
I still don't get it.

Enrollments are nominated as they are, so no hashing like Transactions. Even if a node has no enrollment in it's pool, it can validate and vote on other nominations.

Also, Enrollments are valid only on a single Block. So after any new Block, pools are cleared, effectively syncing the whole network.

linked0:
You're right, but if there are any missing enrollments in the pool on a validator that already exist on other nodes, the validator can not make the proper candidates be validators in the next block by nominating consensus data. Am I missing something?

omerfirmak:
That particular node will fail to nominate those Validators, yes. But some nodes will do. Our `combineCandidates` should prioritze the nominations with enrollments over the ones with no enrollments.

I still feel like our `combineCandidates` is lacking a lot of flexibility. It should be aggregating TXs and Enrollments from all the nominations and creating an optimal one from all of those, not just picking one of the nominations like we do right now, according to a suboptimal criteria.

With all these "syncing", we are going against the nature of a distributed system. It is completely fine (and probably better, since it acts as a load balance) to have different states in different nodes, as long as we update the chain with a somewhat globally optimal set of TXs and enrollments.

linked0:
> Our combineCandidates should prioritze the nominations with enrollments over the ones with no enrollments.

I think it's a good point for the situation. I will think of it and try to apply it after discussing more with @Geod24. Thanks for the good idea and please let me know if you think I'm still missing something.

PR #2442, opened 2021-09-02
Flash tests: Also set the nodes' min_fee_pct to 0

Geod24:
Also comes with two improvements to the documentation / messages we get.
@omerfirmak Please review.

### DIFF
--- a/source/agora/node/Config.d
+++ b/source/agora/node/Config.d
@@ -228,8 +228,32 @@ public struct NodeConfig
     // increased in order to be added to the transaction pool
     public ubyte double_spent_threshold_pct = 20;
 
-    // The minimum percentage of average fee in the pool which the incoming
-    // transaction should include
+    /***************************************************************************
+
+        Minimum fee percentage for transactions to be added to the pool
+
+        When accepting incoming transactions, only accept transactions where the
+        fee is at least `min_fee_pct` percent of the current transaction pool.
+        This setting can be tuned to prevent DoS attack.
+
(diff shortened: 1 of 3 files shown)


source/agora/node/Config.d
```diff
@@ -228,8 +228,31 @@ public struct NodeConfig
...
+        fee is at least `min_fee_pct` percent of the current transaction pool.
+        This setting can be tuned to prevent DoS attack.
+
+        Setting this to 0 sets the mempool to be unlimited in size.
+
+        Example:
+        The transaction pool has two transactions, `TxA` which is 1000 bytes and
+        has a fee of 15_000_000 (fee rate: 15_000) and another one which is 200
+        bytes and comes with a fee of 1_000_000 (fee rate: 5_000).
+        Hence, the average fee rate of the transaction pool is 13_333 / byte.
```
omerfirmak:
```suggestion
        Hence, the average fee rate of the transaction pool is 10_000 / byte.
```

Our average is not weighted. I wonder if it should be.

Geod24: That's what I suspected. Couldn't someone then rather easily lower one's requirement by sending a stream of txs which are just at the threshold, provided the pool's threshold is < 100 ?

omerfirmak: Yeah, they could. But, once the network is mature enough, they would need a lot UTXOs to be able to make any significant impact.

Geod24:
Fixed

Geod24:
Self merging because it gets the CI back on tracks.

Issue #2289, opened 2021-07-20
The name registry should have protection against DoS attacks


Geod24:
Currently the name registry will accept any payload that contains a valid signature as valid.
This means that anyone can generate millions of keys and overload the node registry with registrations.
This fact is used / exploited by the flash layer to register payment channels.

Additionally, even for valid registration, there should be a mechanism to clean up expired validators.

ferencdg:
and what is your suggested solution here?

1. unless the registry have access to the up to date blockchain, how would the registy know what public keys are allowed? Also if it is (still) used by the flash layer how do you plan to support that functionality?

2. if we want to protect against DoS we have to do more things than described in the description for example check for the size of the payload registered. Unlike Agora nodes, Name Registry is not a distributed component, so much easier to execute DoS attack.

Fortunately the Name Registry is not a critical component(at least for non-flash layer purposes, and flash layer was not known at the point of design) as we have distributed DoS resistant fallbacks like network discovery.

ferencdg:
even things like trying to check whether the IP of the sender is matching the address he tries to register wouldn't work, as the original sender might be behind a firewall

ferencdg:
I think I am a bit opposed to the idea of trying to put too much faith in a central, single-point of failure system like the Registry while building a distributed system. Not to mention all the effort that would make into trying to make this central entity reliably fend off DoS attacks.

If anything, then I think we should just throw away this entire idea of Name Registry(we already have hardcoded/configurable seed servers). Along with that we should just plugin our DoS resistant, distributed network crawler(already written) that can keep a more or less up-to-date picture of our network all the time. We can even persist that list in case ALL of our seed servers are down.

omerfirmak:
I think we shouldn't try to be smart and just make NameRegistry a flash enabled FullNode to cover all bases.

ferencdg:
that would solve the access to the blockchain problem, but how would an agora node know which node/nameserver to connect? As long as only one or a couple of nameservers are configured in the node's config it is still regarded as a central component

Geod24:
> and what is your suggested solution here?

That's what I wanted to explore with you guys :)

> unless the registry have access to the up to date blockchain, how would the registy know what public keys are allowed? Also if it is (still) used by the flash layer how do you plan to support that functionality?

Unfortunately, it seems that the only way would be to give the registry access to the blockchain data. However, block headers should be sufficient.
[...]

PR #2443, opened 2021-09-02
Registry: Reject any non-enrolled public key from registering 

omerfirmak:
Part of #2289 

Will follow up with a Flash related PR.


doc/config.example.yaml
```diff
@@ -269,3 +269,21 @@ event_handlers:
...
+  enabled: true
+
+  # DNS server configuration
+  dns:
+    enabled: true
+    address: 0.0.0.0
+    port: 53
+    authoritative:
+      - Zone1
+      - Zone2
```
```suggestion
      - "validators.bosagora.io"
      - "flash.bosagora.io"
```

Geod24 (commented):
Neat. Looks broadly LGTM, a few items stood out but I need to look deeper into it.
So I assume you don't have a use-case for a standalone registry ?
Also we probably need to change the UX a bit, and fold `dns` into `RegistryConfig`, but that can be done later. Since ports < 1024 are privileged under Linux, the nasty effect of having the registry on by default is that it needs to run as root (by default).

omerfirmak:
>  So I assume you don't have a use-case for a standalone registry ?

Yeah, I see no reason why we should have that. But, we either need to fix Flash to not rely on registry or have a seperate registry  hat validates open channels for Flash.

IIUC, it works fine now, because Flash tests only use `LocalRest.Registry`, not `NameRegistry`

Geod24:
Rebased and squashed related commit (otherwise the CI wouldn't pass on the first and second commit).

Geod24 (approved):

Geod24:
@omerfirmak : Would it be complex to relax the restriction to allow not-yet-enrolled validators to register ?
This would allow a validator to ensure it's reachable from the second it becomes a validator.

omerfirmak:
> 
> 
> @omerfirmak : Would it be complex to relax the restriction to allow not-yet-enrolled validators to register ?
> This would allow a validator to ensure it's reachable from the second it becomes a validator.

We could also check the enrollment pool. I'll take a look.

source/agora/node/Runner.d
```diff
@@ -109,6 +116,30 @@ public Listeners runNode (Config config)
...
+    TCPListener dnstcplistener;
+    scope (exit)
+        if (config.registry.enabled && config.registry.dns.enabled)
+        {
+            // Here, we might need to interrupt the task to correctly shut down,
+            // however this throws an exception in the task that needs to be explicitly
+            // handled. And it doesn't help with the error messages printed by Vibe.d.
+            //dnstask.interrupt();
+            dnstcplistener.stopListening();
+        }
```
Geod24, line 128: How did I miss this ? 🤦 
This closes the TCP listener as soon as this function exits. Instead, the listener should be stored in the `result`.

omerfirmak: Shoot

PR #2444, opened 2021-09-03
Trivial: Print the correct line / column for unexpected key

Geod24:
Before it was  using the line / column infos of the containing node,
e.g. `0:0` at the document root.


linked0 (approved):
LGTM

PR #2446, opened 2021-09-03
Fix deploy action to also tag the registry

Geod24:
Otherwise one might be left wondering why things aren't working properly.


kchulj (approved):
Looks green

PR #2453, opened 2021-09-06
Fix #2447: Implement CNAME / A record distinction in registry + others

### LINKED ISSUE #2447: Registry: Implement CNAME / A RR distinction
Currently we register a string, which can be either an IP or a domain name.
The former should get us an A record, the later should be a CNAME.
We should also check the specs to see if we could / should look up the A record matching the CNAME.

Geod24:
The aim was to implement a fix for 2447 and 2451. 2451 proved harder than expected, as we need to start the transfer by sending SOA records. So this implement most of the SOA machinery.

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -208,13 +208,15 @@ public class NameRegistry: NameRegistryAPI
 
         Params:
           query = The query received by the server
-
-        Returns:
-          An answer that matches the query
+          sender = A delegate that allows to send a `Message` to the client.
+                   A server may send multiple `Message`s as a response to a
+                   single query, e.g. when doing zone transfer.
 
     ***************************************************************************/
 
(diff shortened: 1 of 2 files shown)


Geod24:
There was a bug due to a botched rebase, fixed now. Time to take this out for another spin.

source/agora/node/Registry.d
```diff
@@ -273,21 +282,48 @@ public class NameRegistry: NameRegistryAPI
 
         auto ptr = public_key in registry_map;
         // We are authoritative, so we can set `NameError`
-        if (!ptr)
+        if (!ptr || !(*ptr).payload.data.addresses.length)
```
omerfirmak, line 294: I don't think that is possible, payloads with no addresses are rejected anyways.

omerfirmak (approved):

Issue #2256, opened 2021-07-12
Merge 'dns' and registries


Geod24:
Currently our configuration allow to have a DNS seed, as well as registries.
This is a clear duplication of functionality, and they should most likely be merged.
Additionally, the registry currently only accept one address, and we might want to extend that as well as move it to another section so a full node can use it.

Geod24:
In order for this issue to be fixed, we need:
- To be able to lookup a node from its key using our own DNS client (https://github.com/bosagora/agora/issues/2558)
- Our client needs to be recursive and caching, which probably can reuse work done on https://github.com/bosagora/agora/issues/2449;
- The registry needs to support `agora://`, `https://` and `http://` scheme, hence we need https://github.com/bosagora/agora/issues/2566

More infos in https://github.com/bosagora/agora/pull/2516

Geod24:
Note that https://github.com/bosagora/agora/issues/2320 is also closely related.

hewison-chris:
@mkykadir : Can we add here why this is blocked?

mkykadir:
@hewison-chris 

Currently `dns` field in config is used as a seed value; after #2874 is resolved we can remove it and resolve this issue. I don't think this issue should present in the Sprint before resolving the other issue.

PR #2454, opened 2021-09-06
Flash: Drop ThinFlashNode & AgoraFlashNode

omerfirmak:
```
We now only have a single Flash node class called FlashNode.
Anything related to a agora node is handled by delagates, so we still
can have a ThinFlashNode like operation.

This also makes use of the getFlashNode & putFlashNode endpoints of the
NameRegistry. Nodes now need a open channel to be able to register themselves.
In case a node has no open channels, they pass their address to their peer as
they are openning their first channel to bootstrap the communication.
```

I initially wanted to only use the new registry APIs but, having multiple Flash nodes made that harder and didn't make sense. We can simply have a single class and abstract away the network related stuff using delegates.

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -261,33 +261,6 @@ public class FullNode : API
         this.ledger = this.makeLedger();
         // Note: Needs to be instantiated after `ledger` as it depends on it
         this.transaction_relayer = this.makeTransactionRelayer();
-        // If we have handlers defined in the config
-        if (config.event_handlers.length > 0)
-        {
-            // Make `BlockExternalizedHandler`s from config
-            config.event_handlers.filter!(h => h.type == HandlerType.BlockExternalized)
-                .each!(handler => handler.addresses
-                    .each!((string address) =>
-                        this.block_handlers[address] = this.network.getBlockExternalizedHandler(address)));
-
-            // Make `BlockHeaderUpdatedHandler`s from config
-            config.event_handlers.filter!(h => h.type == HandlerType.BlockHeaderUpdated)
-                .each!(handler => handler.addresses
-                    .each!((string address) =>
-                        this.block_header_handlers[address] = this.network.getBlockHeaderUpdatedHandler(address)));
-
-            // Make `PreImageReceivedHandler`s from config
-            config.event_handlers.filter!(h => h.type == HandlerType.PreimageReceived)
-                .each!(handler => handler.addresses
-                    .each!((string address) =>
-                        this.preimage_handlers[address] = this.network.getPreImageReceivedHandler(address)));
-
-            // Make `TransactionReceivedHandler`s from config
-            config.event_handlers.filter!(h => h.type == HandlerType.TransactionReceived)
-                .each!(handler => handler.addresses
-                    .each!((string address) =>
-                        this.transaction_handlers[address] = this.network.getTransactionReceivedHandler(address)));
-        }
 
         Utils.getCollectorRegistry().addCollector(&this.collectAppStats);
         Utils.getCollectorRegistry().addCollector(&this.collectStats);
@@ -401,6 +374,34 @@ public class FullNode : API
 
     public void start ()
     {
+        // If we have handlers defined in the config
+        if (config.event_handlers.length > 0)
+        {
+            // Make `BlockExternalizedHandler`s from config
+            config.event_handlers.filter!(h => h.type == HandlerType.BlockExternalized)
+                .each!(handler => handler.addresses
+                    .each!((string address) =>
+                        this.block_handlers[address] = this.network.getBlockExternalizedHandler(address)));
+
+            // Make `BlockHeaderUpdatedHandler`s from config
+            config.event_handlers.filter!(h => h.type == HandlerType.BlockHeaderUpdated)
+                .each!(handler => handler.addresses
+                    .each!((string address) =>
+                        this.block_header_handlers[address] = this.network.getBlockHeaderUpdatedHandler(address)));
+
+            // Make `PreImageReceivedHandler`s from config
+            config.event_handlers.filter!(h => h.type == HandlerType.PreimageReceived)
+                .each!(handler => handler.addresses
+                    .each!((string address) =>
+                        this.preimage_handlers[address] = this.network.getPreImageReceivedHandler(address)));
+
+            // Make `TransactionReceivedHandler`s from config
+            config.event_handlers.filter!(h => h.type == HandlerType.TransactionReceived)
+                .each!(handler => handler.addresses
+                    .each!((string address) =>
+                        this.transaction_handlers[address] = this.network.getTransactionReceivedHandler(address)));
+        }
+
         if (config.node.stats_listening_port != 0)
             this.stats_server = this.makeStatsServer();
         this.transaction_relayer.start();
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -63,7 +63,7 @@ mixin AddLogger!();
 public class Channel
 {
     /// Used to publish funding / trigger / update / settlement txs to blockchain
-    public const void delegate (Transaction) txPublisher;
+    public const void delegate (in Transaction) txPublisher;
 
     /// Database
     private ManagedDatabase db;
@@ -114,7 +114,8 @@ public class Channel
     private OnUpdateComplete onUpdateComplete;
 
     /// Fetch the Flash Client for the peer
-    private FlashAPI delegate (in PublicKey peer_pk, Duration timeout) getFlashClient;
+    private FlashAPI delegate (in PublicKey peer_pk, Duration timeout,
+        string address = null) getFlashClient;
 
     /// Retry delay algorithm
     private Backoff backoff;
@@ -151,7 +152,7 @@ public class Channel
     public this (FlashConfig flash_conf, ChannelConfig conf, in KeyPair kp,
         PrivateNonce priv_nonce, PublicNonce peer_nonce, FlashAPI peer,
         Engine engine, ITaskManager taskman,
-        void delegate (Transaction) txPublisher,
+        void delegate (in Transaction) txPublisher,
         PaymentRouter paymentRouter,
         OnChannelNotify onChannelNotify,
         OnPaymentComplete onPaymentComplete,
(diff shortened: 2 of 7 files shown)


Geod24:
Regarding the first commit (and 3rd one):

I think it would actually be better to move to a "realm"-based approach. I've been thinking about how to address #2256 on the config level, and currently we have `registry_address` and `addresses_to_register` in both `ValidatorConfig` and `FlashConfig`. With the switch to a pure DNS registry, one can find the name registry through regular `gethostname` call, all they need to provide is (essentially) what domain is the root domain for their chain.

Extending this, we could have `realm` (or a better name) replace `registry_address`. The value would be e.g. `testnet.bosagora.io`, `coinnet.bosagora.io`, `flash.coinnet.bosagora.io`, etc... The registry could then be the same, or separate. What do you think ?

omerfirmak:
> The registry could then be the same, or separate. What do you think ?

Could it to be the same? We need additional data to validate Flash payloads and the validation logic is different. Even if we query the same DNS server to resolve the hostnames, wouldn't we still need some sort of a way to distiinguish Flash and Agora nodes while regiistering them thru our REST endpoints ?

Geod24:
> We need additional data to validate Flash [...]

Hum right, we can't fully unify it.

dub.json
```diff
@@ -64,6 +64,7 @@
                 "source/agora/consensus/Quorum.d",
                 "source/agora/network/Client.d",
                 "source/agora/network/Manager.d",
+                "source/agora/api/Registry.d",
```
Geod24: Alphabetical order.

Geod24: Also why is this needed ?

omerfirmak:
Rebased, added an attempt to make `ValidatorRecurringEnrollment` less of an annoyance.
Hopefully will be green.

omerfirmak:
```
./build/agora-unittests(_D4core7runtime18runModuleUnitTestsUZ19unittestSegvHandlerUNbNiiPSQCm3sys5posix6signal9siginfo_tPvZv+0x3a)[0x564d381e0d8a]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x153c0)[0x7f38d36e33c0]
./build/agora-unittests(_D5agora4node9ValidatorQk19preImageCatchupTaskMFNbZ10__lambda12MFNbNfSQCq9consensus4data12PreImageInfoQoZv+0x29)[0x564d3766f5e9]
./build/agora-unittests(_D3std9algorithm9iteration__T4eachS_D5agora4node9ValidatorQk19preImageCatchupTaskMFNbZ10__lambda12MFNbNfSQCq9consensus4data12PreImageInfoQoZvZ__TQEmTAQBuZQEvMFNbNfKQpZEQGk8typecons__T4FlagVAyaa4_65616368ZQv+0x6e)[0x564d3766d7fe]
./build/agora-unittests(_D5agora4node9ValidatorQk19preImageCatchupTaskMFNbZv+0x31e)[0x564d3753288e]
./build/agora-unittests(_D6geod249LocalRest5Timer3runMFZv+0x60)[0x564d37c934b0]
./build/agora-unittests(_D4core6thread7context8Callable6opCallMFZv+0x3d)[0x564d3824226d]
./build/agora-unittests(_D4core6thread5fiber5Fiber3runMFZv+0x19)[0x564d381e3939]
./build/agora-unittests(fiber_entryPoint+0x96)[0x564d381e3876]
./ci/run.sh: line 21:  5300 Segmentation fault      (core dumped) ./build/agora-unittests
Unittesting agora.test.QuorumPreimage
```

```
Module tests failed: agora.test.NetworkDiscovery
core.exception.AssertError@source/agora/test/NetworkDiscovery.d(146): Node Validator #5 (boa1xzval4nvru2ej9m0rptq7hatukkavemryvct4f8smyy3ky9ct5u0s8w6gfy) did not ban Validator #0 (boa1xrval7gwhjz4k9raqukcnv2n4rl4fxt74m2y9eay6l5mqdf4gntnzhhscrh)
----------------
/home/runner/work/agora/agora/source/agora/test/Base.d:171 nothrow void agora.test.Base.testAssertHandler(immutable(char)[], ulong, immutable(char)[]) [0x55d6c451e6a5]
exception.d:448 onAssertErrorMsg [0x55d6c5133c47]
exception.d:590 _d_assert_msg [0x55d6c5134075]
/home/runner/work/agora/agora/source/agora/test/NetworkDiscovery.d:146 void agora.test.NetworkDiscovery.__unittest_L89_C1() [0x55d6c43b1551]
/home/runner/work/agora/agora/source/agora/test/NetworkDiscovery.d [0x55d6c44e34a4]
/home/runner/work/agora/agora/source/agora/test/Base.d:275 bool agora.test.Base.customModuleUnitTester().runTest(agora.test.Base.customModuleUnitTester().ModTest) [0x55d6c451ef12]
/home/runner/work/agora/agora/source/agora/test/Base.d:302 core.runtime.UnitTestResult agora.test.Base.customModuleUnitTester() [0x55d6c4305586]
runtime.d:607 runModuleUnitTests [0x55d6c5137800]
dmain2.d:551 void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).runAll() [0x55d6c51517c2]
dmain2.d:525 void rt.dmain2._d_run_main2(char[][], ulong, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x55d6c5151742]
dmain2.d:584 _d_run_main2 [0x55d6c51515f9]
dmain2.d:384 _d_run_main [0x55d6c51512b3]
/opt/hostedtoolcache/dc/ldc2-1.26.0/x64/ldc2-1.26.0-linux-x86_64/bin/../import/core/internal/entrypoint.d:42 main [0x55d6c4529654]
??:? __libc_start_main [0x7efc8598a0b2]
??:? _start [0x55d6c3def5dd]
```

source/agora/node/Registry.d
```diff
@@ -194,7 +197,84 @@ public class NameRegistry: NameRegistryAPI
...
+        Register network addresses corresponding to a public key
+
+        Params:
+            registry_payload =
+                the data we want to register with the name registry server
+            channel =
+                a known channel of the registering public key
+
+        API:
+            PUT /flash_node
```
Geod24: The API should be `POST`.
https://restfulapi.net/rest-put-vs-post/

source/agora/test/Base.d
```diff
@@ -1231,6 +1231,13 @@ public class TestAPIManager
     {
         this.ensureTxInPool(iota(this.clients.length), hash);
     }
+
+    /// Ditto
+    public void ensureTxInPool (in Transaction tx)
+    {
+        iota(this.clients.length).each!(idx => this.clients[idx].postTransaction(tx));
+        this.ensureTxInPool(tx.hashFull());
+    }
```
Geod24: This was recently changed, so I guess bad rebase ?

source/agora/test/Base.d
```diff
@@ -1051,14 +1051,14 @@ public class TestAPIManager
 
     ***************************************************************************/
 
-    void addBlock (bool no_txs = false,
+    public void addBlock (bool no_txs = false,
```
Geod24: Those changes, while welcome, would be better in an isolated "Style fixes" commit.

source/agora/test/Base.d
```diff
@@ -618,6 +618,18 @@ public class TestAPIManager
...
+
+        Returns:
+            A pointer to the network registry for name registries
+
+    ***************************************************************************/
+
+    public Registry!NameRegistryAPI* getNRegRegistry ()
+    {
+        return &this.nreg;
+    }
```
Geod24: Why is it a pointer ? Wouldn't `ref` be better ?

source/agora/node/FullNode.d
```diff
@@ -401,6 +374,34 @@ public class FullNode : API
 
     public void start ()
     {
+        // If we have handlers defined in the config
```
Geod24, line 377: Why the move ?

Geod24 (commented):
First round of review. Overall looks very nice!
Would probably be better to split this PR up a bit so we can merge it gradually, as it's quite the behemoth.

dub.json
Geod24: Okay, I looked more into this. It fails because the moduleinfo of `agora.flash.Node` is missing.
It's missing because it is referenced from the `api` module, as it imports it on Registry.d:L20.
Will comment there.

source/agora/api/Registry.d
```diff
@@ -17,6 +17,7 @@ import agora.common.Types;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.crypto.Schnorr: Signature;
+import agora.flash.Node : KnownChannel;
```
Geod24: This means that the interface imports the implementation, which is a red flag.
If `KnownChannel` is part of the API, then it should be isolated (in another module) from the implementation.
The problem is that `KnownChannel` also contains `ChannelConfig`. Looking at the type, the `Config` module stands out a bit. It mixes the actual configuration file definitions (mirroring `agora.node.Config`) and Channel configuration, which, although it might be present in the configuration as well, is more importantly part of the interface.

source/agora/test/Base.d
omerfirmak: It's a leftover anyways, will delete.

source/agora/node/FullNode.d
omerfirmak: We have to create Flash nodes after Agora nodes are created. So at the time FullNode ctor runs, there are no Flash nodes registered, even if they are put in the config as block handlers. So defering the lookup to `start()` makes sure that all nodes (Agora & Flash) are instantiated and registered.

source/agora/node/Registry.d
omerfirmak: I'll do that in a separate commit since `/validator` is `PUT` as well.

source/agora/test/Base.d
omerfirmak: should've used `postAndEnsureTxInPool`

source/agora/api/Registry.d
omerfirmak: I moved all that to Flash api module, in a separate commit.

omerfirmak:
Rebased and dropped some commits.

Geod24 (approved):
Neat!

PR #2456, opened 2021-09-06
ValidatorSet: Remove getPreimageAt

Geod24:
We do not need getPreimageAt, as most places use getValidators nowadays.
As can be seen from the diff, it was only used in one place outside
unittest, and was actually redundant with a previous getValidators call.


hewison-chris (approved):
LGTM

kchulj (commented):
LGTM

PR #2460, opened 2021-09-07
Simple changes extracted from #2453

Geod24:
Those are fairly straightforward and do not lead to a change in behavior, but should make reviewing #2453 easier.


omerfirmak (approved):

kchulj (approved):
LGTM

PR #2467, opened 2021-09-07
DNS: Allow users to configure a zone's SOA record

Geod24:
```
Since we're already parsing zone informations,
allow the users to specify each zone's SOA record.
```

This was also extracted from #2453 to make reviewing easier, and because we can just do all the v0.21.0 changes at once in production.

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -285,5 +285,15 @@ registry:
     address: 0.0.0.0
     port: 53
     authoritative:
-    - "validators.bosagora.io"
-    - "flash.bosagora.io"
+      validators.bosagora.io:
+        email: dns@validators.bosaagora.io
+      flash.bosagora.io:
+        email: dns.flash.bosagora.io
+        refresh:
+          minutes: 1
+        retry:
+          seconds: 30
+        expire:
+          minutes: 10
+        minimum:
+          minutes: 1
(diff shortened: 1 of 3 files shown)


doc/config.example.yaml
```diff
@@ -285,5 +285,15 @@ registry:
...
+      flash.bosagora.io:
+        email: dns.flash.bosagora.io
+        refresh:
+          minutes: 1
+        retry:
+          seconds: 30
+        expire:
+          minutes: 10
+        minimum:
+          minutes: 1
```
Geod24, line 299: The whole registry config needs to be better documented, but the condition is pre-existing, so keeping it for another PR.

doc/config.example.yaml
```diff
@@ -285,5 +285,15 @@ registry:
     address: 0.0.0.0
     port: 53
     authoritative:
-    - "validators.bosagora.io"
-    - "flash.bosagora.io"
+      validators.bosagora.io:
+        email: dns@validators.bosaagora.io
+      flash.bosagora.io:
+        email: dns.flash.bosagora.io
```
hewison-chris, line 291: This isn't an email.

linked0 (approved):
I reviewed it a while ago and LGTM.

hewison-chris (approved):
LGTM

doc/config.example.yaml
Geod24: The DNS `RNAME` record isn't, either. Email addresses are converted to domains, replacing `@` with a `.`.
So this example is to show that both the human-friendly `dns@...` and the `RNAME` approach are supported.

kchulj (approved):
LGTM

PR #2469, opened 2021-09-07
Fix 2248 - Ensure we cannot be DoS through far pre-image reveal

Geod24:
Bundled a couple of minor improvements along with it.

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -64,6 +64,9 @@ node:
   # The percentage by which the double spend transaction's fee should be
   # increased in order to be added to the transaction pool
   double_spent_threshold_pct: 20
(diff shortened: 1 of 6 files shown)


omerfirmak:
Wouldnt this be a problem if we fall back a lot and try to catch up?

Geod24:
> Wouldnt this be a problem if we fall back a lot and try to catch up?

Thought about this during development, and I didn't think so, because we fetch pre-image by height.
However, if the validators don't change, the nodes return the latest pre-image (it's a useful optimization), which means that actually, yes, it would be an issue.

Geod24:
Superseded by #2669 , the small fixes are now in https://github.com/bosagora/agora/pull/2693 .

PR #2474, opened 2021-09-09
Test: Relax assertions in ValidatorRecurringEnrollment

omerfirmak:
to hopefully make it less flakey.

### DIFF
--- a/source/agora/test/ValidatorRecurringEnrollment.d
+++ b/source/agora/test/ValidatorRecurringEnrollment.d
@@ -162,6 +162,7 @@ unittest
 // create another enrollment request for the next block
 unittest
 {
+    import std.algorithm.mutation : reverse;
     static class SocialDistancingNominator : Nominator
     {
         mixin ForwardCtor!();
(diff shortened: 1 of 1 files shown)


Geod24:
So, some enrollments are not making it to the blocks and the validator is not enrolled for one or more block ?

omerfirmak:
> 
> 
> So, some enrollments are not making it to the blocks and the validator is not enrolled for one or more block ?

Yep, it tests gaps in enrollment essentially.

Geod24:
Any way we can wait for enrollment propagation instead of working around it ? Similar to how we do it with TXs.

omerfirmak:
> 
> 
> Any way we can wait for enrollment propagation instead of working around it ? Similar to how we do it with TXs.

I wouldnt say this is a workaround. This can happen in a real life scenario and what those assertions test still fully verifies our intented behaviour. Waiting for propagation sounds more like a workaround IMHO.

hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

Geod24 (approved):
Fair enough

PR #2479, opened 2021-09-10
Accept `future` enrollment

### LINKED ISSUE #1975: TransactionPool and EnrollmentPool should accept "future" transactions
One issue that is showing in our tests is the following scenario:
- Nodes produce blocks, arriving towards the end of their cycle (let's assume cycle = 20);
- One node becomes non-responsive after externalizing height 18;
- A frozen tx is sent at height 19 and externalized;
- Node comes back online;
- An enrollment is sent using said frozen tx;
- Node will reject enrollment;

This is for enrollment, but transactions suffer from the same issue. The tx pool should be greatly improved to deal with those scenarios.

linked0:
There could be a scenario where a node does not have an
frozen UTXO externalized for an enrollment and receives
the enrollment. We improve to deal with the scenario by
making a node accept the enrollment with a transaction
that has the UTXO in the pool.

Fixes #1975

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -986,11 +986,50 @@ public class FullNode : API
     {
         this.recordReq("postEnrollment");
 
+        scope UTXOFinder utxo_finder;
+        scope TestUTXOSet extra_set;
+
         UTXO utxo;
-        this.utxo_set.peekUTXO(enroll.utxo_key, utxo);
+        if (this.utxo_set.peekUTXO(enroll.utxo_key, utxo))
+        {
+            utxo_finder = this.utxo_set.getUTXOFinder();
+        }
+        else 
+        {
+            /// FIXME: Use a proper type and sensible memory allocation pattern
+            // We create a extra UTXO set using the transaction pool if there
+            // is no UTXO in the UTXO set. We only use transactions that have
+            // a frozen UTXO with the right amount.
+            version (all)
+            {
+                extra_set = new TestUTXOSet();
+                foreach (ref Hash hash, ref Transaction tx; this.pool)
+                {
+                    if (!tx.isFreeze())
+                        continue;
+
+                    foreach (output; tx.outputs)
+                        if (output.type == OutputType.Freeze && 
+                            output.value >= Amount.MinFreezeAmount)
+                        {
+                            extra_set.put(tx);
+                            break;
+                        }
+                }
+                extra_set.peekUTXO(enroll.utxo_key, utxo);
+                utxo_finder = extra_set.getUTXOFinder();
+            }
+        }
+
+        if (utxo == UTXO.init)
+        {
+            log.info("Found no UTXO for the enrollment: {}", prettify(enroll));
+            return;
+        }
+
         const utxo_address = utxo.output.address;
         if (this.enroll_man.addEnrollment(enroll, utxo_address,
-            this.ledger.getBlockHeight() + 1, this.utxo_set.getUTXOFinder()))
+            this.ledger.getBlockHeight() + 1, utxo_finder))
         {
             log.info("Accepted enrollment: {}", prettify(enroll));
             this.network.peers.each!(p => p.client.sendEnrollment(enroll));
--- a/source/agora/test/FutureEnrollment.d
+++ b/source/agora/test/FutureEnrollment.d
@@ -1,0 +1,137 @@
+/*******************************************************************************
+
+    Contains tests for accepting "future" enrollments
+
+    Copyright:
+        Copyright (c) 2019-2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.FutureEnrollment;
+
+version (unittest):
+
+import agora.api.FullNode;
+import agora.common.Amount;
+import agora.consensus.data.Block;
+import agora.consensus.data.Enrollment;
+import agora.consensus.data.Transaction;
+import agora.crypto.Key;
+import agora.node.Config;
+import agora.test.Base;
+
+import core.atomic : atomicLoad;
+
+/// Situation: A delayed validator is a block behind the latest height(19)
+///     where other nodes have a block that contains a frozen UTXO for
+///     the enrollment for the outsider. The validator only has the `tx`
+///     in the pool for the frozen UTXO. At that time, the outsider tries
+///     to enroll and the validator receives the enrollment.
+/// Expectation: The validator accepts the enrollment even though there is
+///     not the frozen UTXO in the UTXO set because the transaction that
+///     contains the UTXO exists in the transaction pool. And the block
+///     having the enrollment is externalized for all the nodes.
+unittest
+{
+    static class CustomValidator : TestValidatorNode
+    {
+        mixin ForwardCtor!();
+
+        private shared bool* enable_catchup;
+
+        ///
+        public this (Parameters!(TestValidatorNode.__ctor) args,
+            shared(bool)* enable_catchup)
+        {
+            this.enable_catchup = enable_catchup;
+            super(args);
+        }
+
+        ///
+        protected override void catchupTask () nothrow
+        {
+            if (atomicLoad(*this.enable_catchup))
+                super.catchupTask();
+        }
+    }
+
+    static class CustomAPIManager : TestAPIManager
+    {
+        mixin ForwardCtor!();
+
+        public static shared bool enable_catchup = false;
+
+        /// set base class
+        public override void createNewNode (Config conf, string file, int line)
+        {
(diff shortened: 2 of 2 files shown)


source/agora/test/FutureEnrollment.d
```diff
@@ -0,0 +1,114 @@
...
+import agora.api.FullNode;
+import agora.common.Amount;
+import agora.common.Config;
+import agora.consensus.data.Block;
+import agora.consensus.data.Enrollment;
+import agora.consensus.data.Transaction;
+import agora.crypto.Key;
+import agora.test.Base;
+
+/// Situation: A bad validator are a block behind the latest height(19)
```
hewison-chris:
```suggestion
/// Situation: A slow validator is a block behind the latest height(19)
```

source/agora/test/FutureEnrollment.d
```diff
@@ -0,0 +1,114 @@
...
+///     where other nodes has a block that contain a frozen UTXO for
```
hewison-chris:
```suggestion
///     where other nodes have a block that contains a frozen UTXO for
```

source/agora/test/FutureEnrollment.d
```diff
@@ -0,0 +1,114 @@
...
+///     the enrollment for the outsider. The bad validator only has
```
hewison-chris:
```suggestion
///     the enrollment for the outsider. The slow validator only has
```

source/agora/test/FutureEnrollment.d
```diff
@@ -0,0 +1,114 @@
...
+///     the `tx` in the pool for the frozen UTXO. At that time, the
+///     outsider tries to enroll and the bad validator receives the
+///     enrollment.
+/// Expectation: The bad validator accepts the enrollment even though
```
hewison-chris:
```suggestion
/// Expectation: The slow validator accepts the enrollment even though
```

source/agora/test/FutureEnrollment.d
```diff
@@ -0,0 +1,114 @@
...
+        recurring_enrollment : false
+    };
+    auto all_validators = GenesisValidators + conf.outsider_validators;
+    auto network = makeTestNetwork!CustomAPIManager(conf);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto bad_node = GenesisValidators - 1;
```
hewison-chris:
```suggestion
    auto slow_node = GenesisValidators - 1;
```

hewison-chris (changes requested):
LGTM but I suggested a few typo corrections and a rename.

linked0:
@hewison-chris thanks for the correction.

hewison-chris (approved):
LGTM

source/agora/node/FullNode.d
```diff
@@ -1026,11 +1026,36 @@ public class FullNode : API
     {
         this.recordReq("postEnrollment");
 
+        /// FIXME: Use a proper type and sensible memory allocation pattern
+        version (all)
+        {
+            scope extraSet = new TestUTXOSet();
+            foreach (ref Hash hash, ref Transaction tx; this.pool)
+                extraSet.put(tx);
+            scope extraFinder = extraSet.getUTXOFinder();
```
Geod24: There's two main problems with this approach:
- We populate the `extraSet` with the content of the transaction pool even before we need it, so every `postEnrollment` will pay the price;
- We essentially duplicate the whole `pool` based on user input, which can create high delays and memory allocation;

source/agora/node/FullNode.d
```diff
@@ -1026,11 +1026,36 @@ public class FullNode : API
...
+            scope UTXOFinder enrollmentsUTXOFinder =
+                (in Hash utxo, out UTXO val)
+                {
+                    if (this.utxo_set.peekUTXO(utxo, val))
```
Geod24: Why is this using `peekUTXO` ? It just bypasses any replay protection.

Geod24 (changes requested):

source/agora/node/FullNode.d
linked0: Right. We should use `UTXOFinder`.

source/agora/node/FullNode.d
linked0: Ah, It's better to use `extraSet` only when we fail to find a UTXO with the finder of the original utxo set. Ty.

linked0: > We essentially duplicate the whole pool based on user input, which can create high delays and memory allocation;

Do you mean that we could maintain a future utxo set in the transaction pool?

linked0: > > We essentially duplicate the whole pool based on user input, which can create high delays and memory allocation;
> 
> Do you mean that we could maintain a future utxo set in the transaction pool?

I don't think you meant that. Never mind.

Geod24: No I did not, but it's a potential solution, **iff** we only store the hashes of UTXO that are valid enrollment candidate (freeze txs with the right amount).

linked0: Oh, I didn't think about it to that extent. Ty for the suggestion!

linked0: Updated.

linked0:
Ready for review. 
Please tell me if there's something more to do.

source/agora/test/FutureEnrollment.d
```diff
@@ -0,0 +1,114 @@
...
+import agora.consensus.data.Transaction;
+import agora.crypto.Key;
+import agora.test.Base;
+
+/// Situation: A slow validator is a block behind the latest height(19)
+///     where other nodes have a block that contains a frozen UTXO for
+///     the enrollment for the outsider. The slow validator only has
+///     the `tx` in the pool for the frozen UTXO. At that time, the
+///     outsider tries to enroll and the bad validator receives the
+///     enrollment.
```
Geod24: Why is the validator referred to as "bad" or "slow" ? It isn't either.
Also, the fix checks the TxPool, which is not really tested here IIUC.
It'd be better to let everyone re-enroll, but have an outsider send a freezing transaction, then an enrollment, and check that it correctly gets externalized.

source/agora/test/FutureEnrollment.d
```diff
@@ -0,0 +1,114 @@
...
+    assert(enroll != Enrollment.init);
+
+    // re-enroll all the validators
+    iota(GenesisValidators).each!(i => network.enroll(iota(GenesisValidators), i));
+
+    // check that the bad validator has the enrollment of the outsider
+    auto gotten_enroll = network.clients[slow_node].getEnrollment(enroll.utxo_key);
+    assert(gotten_enroll == enroll);
+
+    network.clients[slow_node].clearFilter();
```
Geod24: There should be a check that a block is externalized with an enrollment matching the frozen UTXO.

Geod24 (changes requested):
Much better! The test is also quite clear. There are few changes needed on the test, but overall looks good.

source/agora/test/FutureEnrollment.d
linked0: Right. Generating a block having the enrollment makes the test complete.

Geod24 (approved):

source/agora/test/FutureEnrollment.d
```diff
@@ -0,0 +1,138 @@
...
+
+*******************************************************************************/
+
+module agora.test.FutureEnrollment;
+
+version (unittest):
+
+import agora.api.FullNode;
+import agora.common.Amount;
+import agora.common.Config;
```
Geod24: `node.Config`

source/agora/node/FullNode.d
```diff
@@ -986,11 +986,50 @@ public class FullNode : API
...
 
+        scope UTXOFinder utxo_finder;
+        scope TestUTXOSet extra_set;
+
         UTXO utxo;
-        this.utxo_set.peekUTXO(enroll.utxo_key, utxo);
+        if (this.utxo_set.peekUTXO(enroll.utxo_key, utxo))
+        {
+            utxo_finder = this.utxo_set.getUTXOFinder();
+        }
```
linked0, line 996:
```suggestion
        if (this.utxo_set.peekUTXO(enroll.utxo_key, utxo))
        {
            utxo_finder = (in Hash hash, out Output op)
            {
                 if (hash == enroll.utxo_key)
                 {
                     op = enroll;
                     return true;
                 }
                 return false;
        }
```

linked0:
Fixed the test failure by modifying the test code.

source/agora/node/FullNode.d
linked0: Will apply this suggestion in other PR.

Geod24 (approved):

PR #2392, opened 2021-08-19
Make the name registry a proper DNS server

Geod24:
```
The function of the name registry was very similar to that of a DNS server.
However, it was working at a higher level, which means the clients depended
on both the DNS server and the registry working.
Combining both reduces the chances of error.

Additionally, this will allow a few key improvements:
- Nodes will be able to get an automatic domain name, e.g. `IDENTIFIER.net.bosagora.io`,
  where `IDENTIFIER` is an encoded version of their UTXO hash;
- This will in turn allow to automatically enable HTTPS for the admin interface,
  allowing a simple way for node operators to connect to their nodes from anywhere;
- It will integrate with the existing DNS seed, for which we have a client but no server;
- We get replication for free, as we can setup a slave DNS server which has no knowledge of Agora;
```

In other words, this is a step towards fixing:
- https://github.com/bosagora/agora/issues/2292
- https://github.com/bosagora/agora/issues/2290
- https://github.com/bosagora/agora/issues/2289
- https://github.com/bosagora/agora/issues/2256

PR #2483, opened 2021-09-13
Not empty array for addresses of Agora servers

### LINKED ISSUE #2395: NetUtils' `getAllIPs` returns an empty array on the servers
Tested on the server as part of #2392 , I got an empty array. This might be due to Agora running in Docker though.
One solution would be to integrate https://github.com/Kripth/my-ip .
This is a blocker for the deployment of TestNet.

linked0:
Tested on the server as part of #2392 , we got an empty array. This might be due to Agora running in Docker though.
This was a blocker for the deployment of TestNet. So we solve this problem by applying [the `ipify` API](https://www.ipify.org/).

Fixes #2395

### DIFF
--- a/source/agora/utils/InetUtils.d
+++ b/source/agora/utils/InetUtils.d
@@ -17,20 +17,30 @@ import agora.utils.Log;
 
 import vibe.inet.url : URL;
 
+import core.stdc.string;
+
 import std.algorithm;
 import std.algorithm.searching;
 import std.array;
 import std.ascii : isAlpha, isDigit;
 import std.conv;
+import std.datetime : Clock, dur;
+import std.file : tempDir, exists, isFile, read, write;
+import std.path : buildPath;
 import std.range : zip, iota;
 import std.regex : matchFirst, regex;
 import std.string : indexOf, strip;
 import std.socket;
 import std.typecons : tuple, Tuple;
 
-import core.stdc.string;
 
 mixin AddLogger!();
+
+private immutable string cache;
+
+shared static this() {
+	cache = buildPath(tempDir(), ".dub.my-ip");
+}
 
 ///
 public enum HostType : ubyte
@@ -39,6 +49,17 @@ public enum HostType : ubyte
     IPv4,
     IPv6,
     Domain,
+}
+
+struct Service {
+	enum ipify = Service("api.ipify.org", "/");							/// https://www.ipify.org/
+	enum plain_text_ip = Service("plain-text-ip.com", "/");				/// http://about.plain-text-ip.com/
+	enum icanhazip = Service("icanhazip.com", "/");						/// http://icanhazip.com/
+	enum whatismyipaddress = Service("bot.whatismyipaddress.com", "/");	/// https://whatismyipaddress.com/api
+	enum amazonws = Service("checkip.amazonaws.com", "/");				/// http://checkip.amazonaws.com/
+
+	string host;
+	string path;
 }
 
 struct InetUtils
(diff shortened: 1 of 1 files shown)


Geod24:
For caching metadata, we should not use an ad-hoc file, but rather `cacheDB`.

linked0:
> For caching metadata, we should not use an ad-hoc file, but rather `cacheDB`.

Ah, right. I'll apply it.

source/agora/utils/InetUtils.d
```diff
@@ -139,10 +160,97 @@ struct InetUtils
...
+            != Socket.ERROR)
+        {
+            char[] buffer = new char[512];
+            ptrdiff_t recv, body_;
+
+            if ((recv = socket.receive(buffer)) != Socket.ERROR && (body_ = buffer[0..recv].indexOf("\r\n\r\n")) != -1)
+                return buffer[body_+4..recv].idup.strip;
+        }
+
+        throw new SocketException("Could not send or receive data");
```
Geod24, line 199: Why are we sending raw HTTP requests instead of using Vibe.d ? Especially considering those are blocking.

source/agora/utils/InetUtils.d
```diff
@@ -86,7 +107,7 @@ struct InetUtils
                 ips ~= ip;
             }
         }
-
+        ips ~= publicAddresses();
```
Geod24, line 110: This should be a fallback, as there's no need to do this kind of query unless we didn't get something usable from getifaddr.

Geod24: Also we probably want to report that situation to the user.

source/agora/utils/InetUtils.d
linked0: OK, I agree that Non-blocking is right for us and we should use Vibe.d.

source/agora/utils/InetUtils.d
linked0: Will think of it.

source/agora/utils/InetUtils.d
```diff
@@ -41,6 +51,17 @@ public enum HostType : ubyte
     Domain,
 }
 
+struct Service {
+	enum ipify = Service("api.ipify.org", "/");							/// https://www.ipify.org/
+	enum plain_text_ip = Service("plain-text-ip.com", "/");				/// http://about.plain-text-ip.com/
+	enum icanhazip = Service("icanhazip.com", "/");						/// http://icanhazip.com/
+	enum whatismyipaddress = Service("bot.whatismyipaddress.com", "/");	/// https://whatismyipaddress.com/api
+	enum amazonws = Service("checkip.amazonaws.com", "/");				/// http://checkip.amazonaws.com/
```
hewison-chris, line 59: Is the plan to try each of these services until a valid `ip` is returned ?

linked0: Actually not but we can. I will discuss it with @Geod24.

source/agora/utils/InetUtils.d
linked0: Modified.

source/agora/utils/InetUtils.d
```diff
@@ -41,6 +51,17 @@ public enum HostType : ubyte
     Domain,
 }
 
+struct Service {
+	enum ipify = Service("api.ipify.org", "/");							/// https://www.ipify.org/
```
mkykadir, line 55: This cannot return IPv6 adresses, `api64.ipify.org` should be used

mkykadir (changes requested):

mkykadir:
Issue is resolved with #3239

PR #2485, opened 2021-09-13
 Flash: Implement private channels

### LINKED ISSUE #2421: Flash: Implement private channels
End users should be able to configure their channels to be private.

- [ ] Private channels should not be gossiped
- [ ] Flash nodes that know of a private channel (i.e peers) should not use it as an intermediate hop while routing
- [ ] If a node receives a routing request for a private channel, it should reject the request

omerfirmak:
Fixes #2421


omerfirmak:
Rebased

source/agora/flash/Node.d
```diff
@@ -469,14 +469,21 @@ public class FlashNode : FlashControlAPI
     /// Handle an outgoing gossip event
     private void handleGossip (GossipEvent event)
     {
-        foreach (peer; this.known_peers.byKeyValue)
+        auto should_gossip = (in ChannelConfig conf, in PublicKey key)
```
Geod24: Use `scope` to avoid memory allocation

Geod24 (approved):
Just the one thing, feel free to self merge afterwards.

source/agora/flash/Node.d
```diff
@@ -469,14 +469,21 @@ public class FlashNode : FlashControlAPI
     /// Handle an outgoing gossip event
     private void handleGossip (GossipEvent event)
     {
-        foreach (peer; this.known_peers.byKeyValue)
+        scope should_gossip = (in ChannelConfig conf, in PublicKey key)
```
omerfirmak, line 472: I feel like I haven't thought this through. We need to gossip the channel so that a private channel can receive multi hop payments. 🤦 Will revert and add a test

PR #2487, opened 2021-09-14
Registry: Make it use its dedicated `FullNode` and use the `Ledger` directly

Geod24:
The first commit is unrelated, but trivial enough that I lumped it here.

The last commit was my goal, however in the process of doing so I realized that the test suite was relying on calling a different node directly for validation! Besides being quite slow, it's also brittle (as `nodes[0]` is likely to be tested).
So the third commit fixes this, and the second commit was just an easy step in that direction.

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -13,13 +13,13 @@
 
 module agora.node.Registry;
 
-import agora.api.FullNode : FullNodeAPI = API;
 import agora.api.Registry;
 import agora.common.DNS;
 import agora.common.Ensure;
 import agora.common.ManagedDatabase;
 import agora.common.Types;
 import agora.consensus.data.ValidatorInfo;
+import agora.consensus.Ledger;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.crypto.Schnorr: Signature;
@@ -62,7 +62,7 @@ public class NameRegistry: NameRegistryAPI
     private ZoneData flash;
 
     ///
-    private FullNodeAPI agora_node;
+    private Ledger ledger;
 
     ///
     private Height validator_info_height;
@@ -71,14 +71,15 @@ public class NameRegistry: NameRegistryAPI
     private ValidatorInfo[] validator_info;
 
     ///
-    public this (string realm, RegistryConfig config, FullNodeAPI agora_node, 
+    public this (string realm, RegistryConfig config, Ledger ledger,
         ManagedDatabase cache_db)
     {
         assert(config.enabled, "Registry instantiated but not enabled");
 
         this.config = config;
         this.log = Logger(__MODULE__);
-        this.agora_node = agora_node;
+
+        this.ledger = ledger;
         this.validators = ZoneData("validator", cache_db);
         this.flash = ZoneData("flash", cache_db);
         Utils.getCollectorRegistry().addCollector(&this.collectStats);
@@ -203,10 +204,10 @@ public class NameRegistry: NameRegistryAPI
             this.validators.get(registry_payload.data.public_key));
 
         // Last step is to check the state of the chain
-        auto last_height = this.agora_node.getBlockHeight() + 1;
+        auto last_height = this.ledger.getBlockHeight() + 1;
         if (last_height > this.validator_info_height || this.validator_info.length == 0)
         {
-            this.validator_info = this.agora_node.getValidators(last_height);
+            this.validator_info = this.ledger.getValidators(last_height);
             this.validator_info_height = last_height;
         }
         ensure(this.validator_info.map!(info => info.address)
@@ -264,11 +265,12 @@ public class NameRegistry: NameRegistryAPI
 
     public override void postFlashNode (RegistryPayload registry_payload, KnownChannel channel)
     {
-        TYPE payload_type = this.ensureValidPayload(registry_payload, 
+        TYPE payload_type = this.ensureValidPayload(registry_payload,
             this.flash.get(registry_payload.data.public_key));
 
-        ensure(isValidChannelOpen(channel.conf, this.agora_node.getBlock(channel.height)),
-            "Not a valid channel");
+        auto range = this.ledger.getBlocksFrom(channel.height);
+        ensure(!range.empty && isValidChannelOpen(channel.conf, range.front),
+               "Not a valid channel");
 
         // register data
         log.info("Registering network addresses: {} for Flash public key: {}", registry_payload.data.addresses,
@@ -380,10 +382,10 @@ public class NameRegistry: NameRegistryAPI
 
         foreach (payload; zone)
         {
-            reply.answers ~= this.convertPayload(payload, format("%s.%s", 
+            reply.answers ~= this.convertPayload(payload, format("%s.%s",
                 payload.payload.data.public_key, zone.root.value));
         }
-            
+
         reply.answers ~= soa;
         reply.header.RCODE = Header.RCode.NoError;
     }
@@ -724,13 +726,13 @@ private struct ZoneData
     {
         this.db = cache_db;
 
-        this.query_count = format("SELECT COUNT(*) FROM registry_%s_signature", 
+        this.query_count = format("SELECT COUNT(*) FROM registry_%s_signature",
             type_table_name);
 
         this.query_registry_get = format("SELECT pubkey " ~
             "FROM registry_%s_signature", type_table_name);
 
-        this.query_payload = format("SELECT signature, sequence, address, type " ~ 
+        this.query_payload = format("SELECT signature, sequence, address, type " ~
             "FROM registry_%s_addresses l " ~
             "INNER JOIN registry_%s_signature r ON l.pubkey = r.pubkey " ~
             "WHERE l.pubkey = ?", type_table_name, type_table_name);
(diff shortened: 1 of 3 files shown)


source/agora/node/Runner.d
```diff
@@ -128,12 +126,13 @@ public Listeners runNode (Config config)
...
-        router.registerRestInterface!(NameRegistryAPI)(result.registry);
+        auto reg = result.node.getRegistry();
+        assert(reg !is null);
+        router.registerRestInterface(reg);
         if (config.registry.dns.enabled)
         {
-            /* auto dnstask = */ runTask(() => runDNSServer(config.registry, result.registry));
-            dnstcplistener = listenTCP(config.registry.dns.port, (conn) => conn.runTCPDNSServer(result.registry),
+            /* auto dnstask = */ runTask(() => runDNSServer(config.registry, reg));
+            dnstcplistener = listenTCP(config.registry.dns.port, (conn) => conn.runTCPDNSServer(reg),
```
hewison-chris: Not part of your changes but `dns_tcp_listener` is easier to read.

source/agora/test/Base.d
```diff
@@ -1542,8 +1552,8 @@ public struct UTXOPair
 /// (multiple-inheritance is not supported in D)
 private mixin template TestNodeMixin ()
 {
-    ///
-    protected AnyRegistry* registry;
+    /// The network registry (note that the parent class has a `registry` member)
+    protected AnyRegistry* nregistry;
```
hewison-chris, line 1532: Maybe `name_registry` ?

hewison-chris (approved):
LGTM

source/agora/node/Runner.d
Geod24: Indeed. I'm going to shuffle this a bit more soon, so I'll rename the variables in the process.

source/agora/test/Base.d
Geod24: It's actually network_registry :)

Geod24:
Looks like there's some extra fix needed for Flash.

Geod24:
So this fails on [restart](https://github.com/bosagora/agora/blob/582e795aa0eebb28ec64235aaa8a796a8e98bbca/source/agora/test/Flash.d#L1464), any idea why @omerfirmak ?

omerfirmak:
``` diff
(ldc-1.27.0)omer@DESKTOP-H1S5IJD:~/bpf/agora$ git diff source/agora/test/Base.d
diff --git a/source/agora/test/Base.d b/source/agora/test/Base.d
index 77cf39238..7abc59e28 100644
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -864,6 +864,7 @@ public class TestAPIManager
         assert(conf.interfaces.length == 1);
         assert(this.registry.register(conf.interfaces[0].address, api.listener()));
         this.dns = api;
+        this.nodes ~= NodePair(conf.interfaces[0].address, new RemoteAPI!TestAPI(api.listener()), time, api);
     }

@@ -2235,23 +2236,23 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)

     auto net = new APIManager(blocks, test_conf, validator_configs[0].consensus.genesis_timestamp, eArgs);

+    // Then all the nodes
+    foreach (ref conf; all_configs)
+        net.createNewNode(conf, file, line);
+
     // First, create the name registry
     Config registry_config = {
       banman : ban_conf,
       node : makeNodeConfig(),
-      network : null,
+      network : net.nodes.map!(pair => pair.address).array.dup,
       logging: test_conf.logging,
       event_handlers : test_conf.event_handlers,
       registry: { enabled: true, },
     };
     net.createNameRegistry(registry_config, file, line);

-    // Then all the nodes
-    foreach (ref conf; all_configs)
-        net.createNewNode(conf, file, line);
-
     return net;
 }
```

This works with https://github.com/Geod24/localrest/pull/129

Geod24:
Hum, the registry should be created first, shouldn't it ? Currently it requires a `network` config I guess, but this is a restriction that should be lifted.

Geod24:
Also I wanted to avoid having the DNS in the `nodes` array as we often iterate on it.

omerfirmak:
> 
> 
> Also I wanted to avoid having the DNS in the `nodes` array as we often iterate on it.

Why? It is just a regular `FullNode` at this point. Validating that it follows the network as expected is important IMO.

Geod24:
> Why?

Mostly because `nodes[0]` is special in that it's the most requested node.
Also, it's not just a regular full node - It knows all the address and is a PoF, which means we shouldn't `restart` or make it `sleep`.

Geod24:
But we're adding it as the last node, so it's fine 🤦

Geod24:
It seems to consistently fail `ValidatorRecurrringEnrollment` now, oh joy.

omerfirmak:
> 
> 
> It seems to consistently fail `ValidatorRecurrringEnrollment` now, oh joy.

Probably because registry node never receives any `Enrollment`s

omerfirmak:
Maybe nodes should always use registry address in discovery to discover them as fUllnodes

Geod24:
> Probably because registry node never receives any Enrollments

To be precise, we were iterating on all the nodes (including registry), and getting their enrollment. Then submitting it to the network. The one returned by registry should have been `Enrollment.init`, which was rightfully rejected, leading to a timeout on the externalization.

Now, do `FullNode` have a requirement to have txs and enrollments propagated to them ? I don't believe so.
A validator MUST to ensure that it is well-connected (it has all the required validators for SCP to work).
Besides that, a Validator MAY choose to connect to more nodes, validators or full node, for gossipping. But there is no guarantee it will do so.

Geod24:
I believe this needs https://github.com/bosagora/agora/issues/2524 to move forward. Or longer timeouts.

source/agora/test/Base.d
```diff
@@ -2152,7 +2152,22 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
     auto net = new APIManager(blocks, test_conf, validator_configs[0].consensus.genesis_timestamp, eArgs);
     foreach (ref conf; all_configs)
         net.createNewNode(conf, file, line);
+<<<<<<< HEAD
```
hewison-chris: Oops.

hewison-chris (changes requested):
Conflict to resolve

omerfirmak:
> 
> 
> I believe this needs #2524 to move forward. Or longer timeouts.

Maybe give this a try?

``` diff
diff --git a/source/agora/consensus/Ledger.d b/source/agora/consensus/Ledger.d
index 671c25ab5..1287168bb 100644
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -85,7 +85,7 @@ public class Ledger
     private Block last_block;
 
     /// UTXO set
-    private UTXOCache utxo_set;
+    public UTXOCache utxo_set;
 
     // Clock instance
     private Clock clock;
diff --git a/source/agora/node/Registry.d b/source/agora/node/Registry.d
index 6cd541beb..438cfd747 100644
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -187,7 +187,7 @@ public class NameRegistry: NameRegistryAPI
     {
         import std.algorithm;
-
+        import agora.consensus.data.UTXO;
         TYPE payload_type = this.ensureValidPayload(registry_payload, this.validators.map);
 
         // Last step is to check the state of the chain
@@ -197,8 +197,14 @@ public class NameRegistry: NameRegistryAPI
             this.validator_info = this.ledger.getValidators(last_height);
             this.validator_info_height = last_height;
         }
+        UTXO utxo;
         ensure(this.validator_info.map!(info => info.address)
-            .canFind(registry_payload.data.public_key), "Not an enrolled validator");
+            .canFind(registry_payload.data.public_key) ||
+            this.ledger.enrollment_manager.enroll_pool
+                .getEnrollments(last_height).canFind!(
+                    enroll => this.ledger.utxo_set.peekUTXO(enroll.utxo_key, utxo)
+                        && (utxo.output.address == registry_payload.data.public_key)
+                ), "Not an enrolled validator");
 
         // register data
         log.info("Registering addresses {}: {} for public key: {}", payload_type,

```

omerfirmak:
and this

``` diff
diff --git a/source/agora/test/Base.d b/source/agora/test/Base.d
index 90bd55bbe..e5e76250f 100644
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -2122,7 +2122,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         .map!(en => validatorAddress(en.index, en.value)).array;

     auto full_node_addresses = test_conf.full_nodes.iota.map!(
-        idx => fullNodeAddress(idx)).array;
+        idx => fullNodeAddress(idx)).array ~ ["name.registry"];

     // full nodes and enrolled validators will connect to other enrolled validators
     // and other full nodes (but not to outsider nodes)
```

source/agora/node/Registry.d
```diff
@@ -253,8 +254,9 @@ public class NameRegistry: NameRegistryAPI
     {
         TYPE payload_type = this.ensureValidPayload(registry_payload, this.flash.map);
 
-        ensure(isValidChannelOpen(channel.conf, this.agora_node.getBlock(channel.height)),
-            "Not a valid channel");
+        auto range = this.ledger.getBlocksFrom(channel.height);
```
mkykadir, line 271: Previously it was returning `HTTP:400` when `channel.height` is wrong, it will result `HTTP:500` now.

Geod24:
Found the bug, this should work now

omerfirmak:
Can you rebase on https://github.com/omerfirmak/agora/tree/registry-fullnode-instantiation?

Geod24:
Superseded by #2578

PR #2489, opened 2021-09-14
Remove `cycle_length` field from `Enrollment`

### LINKED ISSUE #1548: Agora should ensure that validator cycle and "preimage cycle" are consistent
The preimage cycle (a.k.a. number of blocks that will belong to the same pre-image "chain") must be evenly divisible by the validator cycle otherwise we'll have a nasty edge case.

kchulj:
Agora should ensure that `validator_cycle` and `preimage_cycle` are consistent.
The `preimage_cycle` must be evenly divisible by the `validator_cycle`.
The `cycle_length` field can be removed because there is the same field in `ConsensusConfig`.

Fixes #1548


kchulj:
Still in testing stage, will mark as ready when ready.

kchulj:
Ready for review (it was 🟢 before).
Note that:
- `cycle_length` parameter of the `PreImageCycle` constructor is now defaulted to 20, because some instances were getting the value from `Enrollment`'s `cycle_length`.
- `cycle_length`, `cycles` of the `PreImageCycle` still exist, because `unittest`s still use `cycle_length * cycles`.
- Ping for DDOC update review to check if the concepts are correct.

Geod24 (approved):
LGTM

Issue #2422, opened 2021-08-27
Flash: Delegate all signing operations to the wallet


omerfirmak:
Currently, wallet shares its seed with the Flash node for it to create all necessary signatures during a channel's lifetime.

This assumes that each user runs their own Flash node, or fully trusts the node operator. Ideally we want an end user to able to use a third-party service for its Flash needs, a node.

The biggest issue here is the on-chain fees that Flash nodes currently add to the transactions before they are published to the network, for unilateral close operations. For that, node needs access to the private key in real-time. Since user's wallet can be offline for most of the time and cannot be relied on, third-party node operator would have to cover those on-chain fees and charge its customer accordingly.

- [ ] Flash node should delegate all signature creation to the wallet.
- [ ] On-chain fees should be handled by the node operator if necessary

omerfirmak:
This also should include decryption and encrpytion of Onion packets.

Geod24:
@omerfirmak : Wasn't this fully covered by #2493 ?

omerfirmak:
> 
> 
> @omerfirmak : Wasn't this fully covered by #2493 ?

No this still has ways to go.

PR #2493, opened 2021-09-15
Flash: ChannelUpdates are now signed externally

omerfirmak:
Part of #2422


Geod24 (changes requested):
LGTM bar the side effect in `assert`.

omerfirmak:
> 
> 
> LGTM bar the side effect in `assert`.

Updated

Geod24 (approved):

PR #2495, opened 2021-09-15
Registry: Document and flatten configuration

Geod24:
```
The registry configuration was a bit lacking, and this has been fixed.
Additionally, since the whole point of the 'registry' is now to have an 'enabled' boolean,
and then contain the DNS configuration, the two configurations were flattened.
One reason to do so is that we want to move towards the registry being mostly DNS-based
(as opposed to mostly HTTP-based, as currently), if not only DNS-based.
Hence, the registry and the DNS server are now tied together.
The original reason for allowing disabling the DNS server was to not impair
production if a bug was found in the DNS server, but this concern is no more.
```

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -272,28 +272,55 @@ event_handlers:
 
 ################################################################################
 ##                             Registry configuration                         ##
-## Configuration for the name registry.                                       ##
 ##                                                                            ##
+## The registry is a DNS server used to resolve the address of validators or  ##
+## Flash nodes. It is also a Full Node (or validator if enabled), as it needs ##
+## to validate registration requests.                                         ##
+##                                                                            ##
+## The registry is optional, and disabled by default. Enabling it exposes an  ##
+## extra API for registration, and a DNS server (on port 53 by default).      ##
 ################################################################################
 registry:
   # If this node should also act as a registry
   enabled: true
 
-  # DNS server configuration
-  dns:
-    enabled: true
-    address: 0.0.0.0
-    port: 53
-    authoritative:
-      validators.bosagora.io:
-        email: dns@validators.bosaagora.io
-      flash.bosagora.io:
-        email: dns.flash.bosagora.io
-        refresh:
-          minutes: 1
-        retry:
-          seconds: 30
-        expire:
-          minutes: 10
-        minimum:
-          minutes: 1
+  # The address to bind the DNS server to (default: 0.0.0.0)
+  # On Linux system with systemd, one might want to specify the local IP,
+  # as systemd-resolvd binds to 127.0.0.53:53, which conflicts with this.
+  address: 0.0.0.0
+  # Port to bind to (default: 53, standard DNS port)
+  # Note that ports < 1024 are privileged ports and might require root powers
+  port: 53
+  # Zones for which this server is authoritative
+  authoritative:
+    # The zone responsible for validators registration
+    validators.bosagora.io:
+      ## The fields below define the SOA record for the DNS entry.
+      ##
+      ## If you are not familiar with DNS, simply enter a valid email address,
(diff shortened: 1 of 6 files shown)


source/agora/node/Config.d
```diff
@@ -482,6 +412,63 @@ public struct DNSConfig
...
+                   zone.name, zone.refresh, intMaxSecs);
+            ensure(zone.retry <= intMaxSecs,
+                   "registry.authoritative: Zone '%s' field 'retry' ({}) should be at most {}",
+                   zone.name, zone.retry, intMaxSecs);
+            ensure(zone.expire <= intMaxSecs,
+                   "registry.authoritative: Zone '%s' field 'expire' ({}) should be at most {}",
+                   zone.name, zone.expire, intMaxSecs);
+            ensure(zone.minimum <= intMaxSecs,
+                   "registry.authoritative: Zone '%s' field 'minimum' ({}) should be at most {}",
+                   zone.name, zone.minimum, uintMaxSecs);
```
hewison-chris, line 469:
```suggested
                   zone.name, zone.minimum, intMaxSecs);
```

hewison-chris (changes requested):
LGTM just a typo

source/agora/node/Runner.d
```diff
@@ -130,12 +130,9 @@ public Listeners runNode (Config config)
     {
         result.registry = new NameRegistry(config.registry, result.node);
         router.registerRestInterface!(NameRegistryAPI)(result.registry);
-        if (config.registry.dns.enabled)
-        {
-            /* auto dnstask = */ runTask(() => runDNSServer(config.registry, result.registry));
-            result.tcp ~= listenTCP(config.registry.dns.port, (conn) => conn.runTCPDNSServer(result.registry),
-                config.registry.dns.address);
-        }
+        /* auto dnstask = */ runTask(() => runDNSServer(config.registry, result.registry));
```
kchulj, line 133: Is the `auto dnstask` an implementation for the future ?

linked0 (approved):
LGTM

doc/config.example.yaml
```diff
@@ -272,28 +272,55 @@ event_handlers:
...
+    # The zone responsible for validators registration
+    validators.bosagora.io:
+      ## The fields below define the SOA record for the DNS entry.
+      ##
+      ## If you are not familiar with DNS, simply enter a valid email address,
+      ## and use the default values.
+      ##
+      ## Those fields are ignored if the server is not authoritative for the zone.
+
+      # Email of the DNS server admnistrator
```
kchulj:
```suggestion
      # Email of the DNS server administrator
```

source/agora/node/Runner.d
Geod24: It's a bit of a tricky issue with the UDP code.
The TCP code spawns fibers for us and our client code run in it. But the UDP code is a bit more ancient and doesn't do that. Instead, we need to manually run the server and listen in it. Which means we need to run a task.
When the node shuts down, we should interrupt that task. However, interrupt throws an Exception, so we would need to handle that. So I've let it for a future enhancement.

kchulj (approved):
LGTM

source/agora/node/Config.d
Geod24: Good eye! Although the error is actually on the comparison, as `minimum` is unsigned:

> MINIMUM         The unsigned 32 bit minimum TTL field that should be
>                           exported with any RR from this zone.

[Source](https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.13).
Added a commit to fix it.

doc/config.example.yaml
Geod24: Fixed

hewison-chris (approved):
LGTM

PR #2497, opened 2021-09-16
NetMan: Remove redundant connected_peers variable

Geod24:
```
We already have 'peers' which contains 'address', so this is redundant.
Additionally, we do not actually know if the peer is connected:
it might have timed out / been disconnected, and since we never clean
this Set, we might end up with stale information.
The only benefit we're loosing is the amortized O(1) lookup.
```


linked0 (approved):
LGTM

kchulj (approved):
🆗

PR #2505, opened 2021-09-17
Flash: Revert selective gossipping for private channels

omerfirmak:
Since our path finding is source based, when a private channel
funder wants to receive a payment through their private channel
source would have to know about existence of the channel.

### DIFF
--- a/source/agora/test/Flash.d
+++ b/source/agora/test/Flash.d
@@ -594,7 +594,8 @@ private TestConf flashTestConf ()
     conf.event_handlers = [
         EventHandlerConfig(HandlerType.BlockExternalized, [WK.Keys.A.address.to!string()]),
         EventHandlerConfig(HandlerType.BlockExternalized, [WK.Keys.C.address.to!string()]),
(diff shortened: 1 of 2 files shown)


source/agora/test/Flash.d
```diff
@@ -1917,3 +1918,133 @@ unittest
...
+    eomer.waitForChannelDiscovery(alice_charlie_chan_id);
+
+    // begin off-chain transactions
+    auto inv_1 = diego.createNewInvoice(WK.Keys.D.address, Amount(2_000),
+        time_t.max, "payment 1");
+
+    // This would have to use Charlie => Diego channel as a hop, but channel is private
+    alice.payInvoice(WK.Keys.A.address, inv_1.value);
+    auto inv_res = network.listener.waitUntilNotified(inv_1.value);
+    assert(inv_res == ErrorCode.PathNotFound, format("Payment %s didn't fail", inv_res));
```
Geod24, line 2035: Note that nowadays, with `-checkaction=context`, you get a nice output even without the message as a parameter.

Geod24 (approved):

kchulj (approved):

PR #2506, opened 2021-09-17
Flash: Introduce node key pair

omerfirmak:
We are gradually removing any dependency on user KeyPair in
Channel and the Flash node. But adding fees to the update and
settlement TXs require access to KeyPair in realtime. So Flash node
will be configured to use a single KeyPair to be used for all fee
bumping operations.

### DIFF
--- a/source/agora/flash/Config.d
+++ b/source/agora/flash/Config.d
@@ -73,6 +73,11 @@ public struct FlashConfig
 
     /// Multiplier for the truncating exponential backoff retrying algorithm
     public uint retry_multiplier = 10;
+
+    /// The seed to use for the keypair of this node
+    /// This will only be used for paying on-chain fees in uncollaborative close attempts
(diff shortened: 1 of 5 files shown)


Geod24:
I'm not fully sure how this would / should work. Let's touch base later today.

source/agora/flash/Config.d
```diff
@@ -73,6 +73,10 @@ public struct FlashConfig
 
     /// Multiplier for the truncating exponential backoff retrying algorithm
     public uint retry_multiplier = 10;
+
+    /// The seed to use for the keypair of this node
+    @Converter!KeyPair((value) => KeyPair.fromSeed(SecretKey.fromString(value.as!string)))
+    public @Name("seed") immutable KeyPair key_pair;
```
Geod24: Could you add the field to the example config, especially considering that it's a required parameter.
Also document what this key is for (only paying fees).

Geod24 (approved):

PR #2507, opened 2021-09-23
Nominator: Print quorum set in log messages

Geod24:
```
Previously, we were never looking up the quorum set because we always
used the default argument (null) to the struct.
```

Finally got around to fixing this.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -617,7 +617,7 @@ extern(D):
             return;
         }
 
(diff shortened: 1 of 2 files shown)


linked0:
LGTM.
The signature of `scpPrettify` of `SCPPrettyPrinter` should be changed also.

linked0 (approved):

Geod24:
To not have a default value ?

linked0:
> To not have a default value ?

I meant that. But never mind. There's something I missed.

Issue #2455, opened 2021-09-06
Long distance block catchup failed


Geod24:
```
2021-09-07 00:54:54,665 Error [agora.consensus.Ledger] - Validator 0x210f6551d648a4da654da116b100e941e434e4f232b8579439c2ef64b04819bd2782eb3524c7a29c38c347cdf26006bccac54a58a58f103ae7eb5b252eb53b64 (index: 0, address: boa1xrval7gwhjz4k9raqukcnv2n4rl4fxt74m2y9eay6l5mqdf4gntnzhhscrh) missing pre-image for height 1 (known: 0)
2021-09-07 00:54:54,665 Error [agora.consensus.Ledger] - Validator 0x3b44d65edb3361dd91441ab4f449eeda55644026624c4b8ae12ecf0264fa8a228dbf672ef97e2c4f87fb98ad7099e17b7f9ba7dbe8479672066912b1ea24ba77 (index: 1, address: boa1xzval2a3cdxv28n6slr62wlczslk3juvk7cu05qt3z55ty2rlfqfc6egsh2) missing pre-image for height 1 (known: 0)
2021-09-07 00:54:54,665 Error [agora.consensus.Ledger] - Validator 0x7bacc99e9bf827f0fa6dc6a77303d2e6ba6f1591277b896a9305a9e200853986fe9527fd551077a4ac2b511633ada4190a7b82cddaf606171336e1efba87ea8f (index: 2, address: boa1xzval3ah8z7ewhuzx6mywveyr79f24w49rdypwgurhjkr8z2ke2mycftv9n) missing pre-image for height 1 (known: 0)
2021-09-07 00:54:54,666 Error [agora.consensus.Ledger] - Validator 0x9b2726e79f05abc107b6531486a46c977414e13ed9f3ee994ec14504964f86fcf9464055b891b9c34020feb72535c300ff19e8b5167eb9d202db1a053d746b2c (index: 3, address: boa1xrval6hd8szdektyz69fnqjwqfejhu4rvrpwlahh9rhaazzpvs5g6lh34l5) missing pre-image for height 1 (known: 0)
[...]

PR #2511, opened 2021-09-23
Remove register listener

omerfirmak:


source/agora/node/Validator.d
```diff
@@ -555,7 +555,12 @@ public class Validator : FullNode, API
 
         // regenerate the quorums
         if (validators_changed || need_shuffle)
+        {
+            // Try to register immediately if we are a validator for the next cycle
+            if (this.enroll_man.isEnrolled(block.header.height + 1, &this.utxo_set.peekUTXO))
+                () @trusted { this.network.onRegisterName(); } ();
```
Geod24: A `@system` function is a function for which we can't prove that there won't be any memory corruption for any combination of parameter. A `@safe` function is one where we can prove it. A `@trusted` lambda is usually used to signify that the call is `@safe` with this set of parameters.

This is easier to demonstrate with C function, let's say `strdup`.
If you do:
```D
void main () @safe
{
    auto ptr = () @trusted { return strdup("Hello World"); }();
}
```

A human can easily look at this, and knowing that string literals are always `\0`-terminated, the call will not result in any memory corruption.

Now when I look at this block of code, I am confused, because `onRegisterName` does not take any parameter.
So the only way that it could be `@system` and proven `@safe` here would be if calling `onRegisterName` was not `@safe` depending on the state of `this`. That would obviously be a very big breach of encapsulation.

So it seems to me that in this case, `onRegisterName` should have been made `@safe`, instead of being wrapped here.

source/agora/test/ManyValidators.d
```diff
@@ -75,6 +80,10 @@ void manyValidators (size_t validators)
             format("Node %s has validator count %s. Expected: %s",
                 idx, node.countActive(Height(GenesisValidatorCycle + 1)), validators)));
 
+    // Wait for nodes to run a discovery task and update their required peers
+    Thread.sleep(3.seconds);
+    network.waitForDiscovery();
```
Geod24, line 85: Shouldn't `waitForDiscovery` handle the sleeping period ?

omerfirmak: Even after block externalization, required_peers in NetworkManager is not updated until the next call to `discover()` which happens periodically in discovery timer. So between the block externalization and the next timer tick, NetworkManager still thinks the network state is complete, even though quorums were shuffled in the last block and makes `waitForDiscovery()` return earlier than it should.

Geod24:
First commit fixes #2455 so thanks for that.
One thing though is that if a node falls too much behind *and* there has been a lot of changes in the validator set, what do we send ? Everything from the looks of it.

Also, nitpick in the commit message:
> if it falls too behind

Much ?

source/agora/test/ManyValidators.d
Geod24: So.... Shouldn't we fix this ?

Geod24:
Very nice! Not much to say. The Integration test failed though, I've re-triggered it, let's see if it was spurious, but I doubt so.

Geod24:
Looks like the integration failure is legit @omerfirmak

omerfirmak:
> 
> 
> Looks like the integration failure is legit @omerfirmak

There are no errors in the logs, it seems nodes are not shutting down properly.

Geod24 (approved):

PR #2513, opened 2021-09-24
Fix #2363: Unify / clear up the REST API

### LINKED ISSUE #2363: Unify / clear up the REST API

Geod24:
The bulk of it has been done in #2428 , and the only point not addressed from the issue is its own issue.

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -1032,10 +1032,10 @@ public class FullNode : API
         }
     }
 
-    /// GET: /preimages_for_enroll_keys
-    public override PreImageInfo[] getPreimagesForEnrollKeys (Set!Hash enroll_keys = Set!Hash.init) @safe nothrow
+    /// GET: /preimages
+    public override PreImageInfo[] getPreimages (Set!Hash enroll_keys = Set!Hash.init) @safe nothrow
     {
-        this.recordReq("preimages_for_enroll_keys");
+        this.recordReq("preimages");
 
         // if enroll_keys is empty, then all preimages should be returned
         if (enroll_keys.empty())
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -1057,13 +1057,27 @@ public class FullNode : API
         return preimage_infos;
     }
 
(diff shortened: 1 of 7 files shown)


linked0 (approved):
Neat!

source/agora/node/FullNode.d
```diff
@@ -1057,13 +1057,27 @@ public class FullNode : API
...
-        return this.enroll_man.getValidatorPreimages(Height(start_height),
-            Height(end_height)).array();
+        this.recordReq("preimages_range");
+
+        const known = this.ledger.getBlockHeight();
+        // We have no data that could match this query
+        if (known < start_height)
+            return null;
+
+        // Bounds check end_height and make the API prractical
```
kchulj, line 1072:
```suggestion
        // Bounds check end_height and make the API practical
```

kchulj (approved):
LGTM one typo

PR #2514, opened 2021-09-24
Registry: Move instantiation to FullNode

Geod24:
```
By keeping a reference to the Registry from the FullNode, we can ensure that
the FullNode is able to use it (e.g. in its network manager) and ideally
control it (e.g. forward shutdown() instructions), instead of having them
separated as currently is.
```

Extracted from #2487 to make it move forward.

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -1081,7 +1081,7 @@ public class FullNode : API
         if (known < start_height)
             return null;
 
-        // Bounds check end_height and make the API prractical
+        // Bounds check end_height and make the API practical
         if (known < end_height)
             end_height = known;
         else if (end_height < start_height)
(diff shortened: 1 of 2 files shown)


source/agora/node/FullNode.d
```diff
@@ -274,6 +283,9 @@ public class FullNode : API
             // Use second precision to simplify aggregation
             Clock.currTime!(ClockType.second)(UTC()).toISOString(),
         );
+
+        if (config.registry.enabled)
+            this.registry = new NameRegistry(config.registry, this);
```
linked0:
```suggestion
            this.registry = new NameRegistry(config.node.realm, config.registry, this);
```
There is a compile error about this.

kchulj: LGTM

Geod24: Thanks, fixed

Geod24:
Also added a commit to fix the typo @kchulj spotted

kchulj (approved):
✅

linked0 (approved):
LGTM.

PR #2515, opened 2021-09-24
Remove EM tests that is no longer relevant

Geod24:
This test is described as testing 'getRandomSeed' which is no longer in the EM.
Additionally, most of its code seems to be redundant with PreImageCycle test code.


linked0 (approved):
LGTM. I've also noticed that it's redundant.

PR #2516, opened 2021-09-24
Document DNS scheme used

Geod24:
I was looking into how we could be remove any HTTP lookup, and it looks like either `SRV` or `URI` would do the trick.
However `SRV` assumes the scheme is know, and that's not currently the case for us.
So I started adding the needed definitions, and wrote documentation to increase the bus factor.

TODO:
- Specify A / AAAA record role (they need to be returned as extra if available);
- Specify "User B" accessing admin interface from his browser;

### DIFF
--- a/doc/DNS.md
+++ b/doc/DNS.md
@@ -1,0 +1,117 @@
+## DNS localization scheme
+
+Agora is a distributed, decentralized network, where nodes enter by "enrolling".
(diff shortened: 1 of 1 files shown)


Geod24:
This is just documentation, but it's what we've been working on for months, so maybe time to bring it in ? :)

mkykadir (approved):

PR #2526, opened 2021-09-28
RPC: Reuse established TCP connections

omerfirmak:


source/agora/network/RPC.d
```diff
@@ -385,7 +388,14 @@ private void handleThrow (API) (scope API api, ref TCPConnection stream)
...
 
-    log.trace("[{}] Handling a new request: {}", stream.peerAddress, stream.leastSize());
+    // use the existing readTimeout in leastSize()
+    // if this is a new connection, handle() will have set it to the configuration value
+    // if it is a reused connection it will be set to the keep alive period.
+    log.info("[{}] Handling a new request: {}", stream.peerAddress, stream.leastSize());
+    // after the initial data arrives, reduce the timeout to the configured amount
+    stream.readTimeout = timeout;
+    // We will reuse this connection, keep the connection alive for 10 minutes after handling a request
+    scope (exit) stream.readTimeout = 10.minutes;
```
Geod24, line 398: Not sure I follow. The connection is with a specific client, so we are bounding ourselves to the client's keep-alive, right ?

omerfirmak: Yeah, we are. We could enable keep alive on server as well tho.

Geod24: Would that make sense to make this a parameter of the `interface` (with a sensible default value) ?

omerfirmak: As an agora config? If we are gonna parameterize this, max number of connections would be a better one and we would drop older ones as we reach that limit

Geod24 (approved):

source/agora/network/RPC.d
Geod24: Well, here's a few questions:
1) Does the number of connection take into account validators connection or any connections ?
2) How can we prevent a single client (with multiple IPs) from hogging one specific nodes ?

In general, whatever limits we set should be for unauthenticated clients (full nodes). See for example https://github.com/bosagora/agora/issues/2523 .
The rationale being, if we have let's say 5 other nodes in our quorums, we need to allow at least those 5 connections. That's the bare minimum to be functional. But there are two other kinds of connections that may exist:
1) Connections to other validators, e.g. before quorum re-shuffle;
2) Connections to full nodes;

The first one is something we can control quite well, because those nodes are authenticated and so if we allow 10 "extra" connections we know it can't be abused.
On the other hand, full nodes currently work on a first-come-first-serve basis, so if a new validators come online, and a client establish 10 connections (or whatever the limit is) to us, they could simply prevent us from having other peers connected by using those connections for trivial things (sending invalid transactions for example).

omerfirmak: What I was referring to are TCP connections, which only limit the number of concurrent requests a node can handle. We can serve unlimited number of clients even if we maintain only a single TCP connection at a time.

Dropping idle connections will help preventing a DoS attack. We can track how long a request takes and if it reaches a certain threshold we can allow it to be dropped if a new connection comes and we are at our maximum connection limit.

PR #2527, opened 2021-09-28
Test: Add Registry to network as a FullNode

omerfirmak:
This will make Validators gossip TXs/enrollments/preimages
to the name registry node.


omerfirmak:
wait this should be a part of #2487, not a standalone PR. Closing.

PR #2528, opened 2021-09-28
Tests: Reduce amount of imports boilerplate

Geod24:
Writing a tests historically required a lot of boilerplate,
but this was gradually trimmed down as our test framework evolved.
This is another attempt at trimming it down by removing a lot of
no-longer-needed imports, as well as adding common ones to Base
as public imports (e.g. almost every tests use 'Transaction').

### DIFF
--- a/source/agora/test/BanManager.d
+++ b/source/agora/test/BanManager.d
@@ -16,9 +16,6 @@ module agora.test.BanManager;
 
 version (unittest):
 
-import agora.common.Types;
-import agora.consensus.data.Transaction;
-import agora.crypto.Hash;
 import agora.test.Base;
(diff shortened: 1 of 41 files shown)


kchulj:
`Base.d`: What do you think of grouping `public import agora.utils.Utility : retryFor;`
under
`/* The following imports are frequently needed in tests */` ?

linked0 (approved):
LGTM

Geod24:
@kchulj : I'll probably do another round of cleanup in the future. For the moment, that `utils` import comes with a rationale (the comment), hence why it's slightly separate.

kchulj (approved):
LGTM

PR #2529, opened 2021-09-28
Tests: Improve discovery failure error message

Geod24:
This not only tells us what node has not completed discovery, but also their currrent NodeInfo.
We can also avoid the try / catch / rethrow by just passing the file and line parameters to retryFor.


kchulj (approved):
LGTM
Looks like CI needs a re-run

PR #2536, opened 2021-09-29
Registry: Always use qname with CNAME record

Geod24:
```
The original code came from a misinterpretation of the spec:
it says 'copy [...], change QNAME [...], and go back [...]'.

But those things should not happen simultaneously, so the copy should happen
with the proper QNAME, then the algorithm should be restarted with a new QNAME.
However, this is done to support A records in the same domain, e.g. when
'a.example.com' is a CNAME for 'b.example.com' and the latter has an 'A' record.

Currently, 'PUBKEY.validators.REALM' cannot point to 'XXXX.validators.REALM',
as it could only point to another pubkey. It could in the future, relying on
domain matching to select the right key, but this first require us to fix
the validation of record added to the registry.
```

This is a bug spotted when I tested AXFR. Currently deploying a test registry to confirm.

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -34,6 +34,7 @@ import std.algorithm.iteration : map, splitter;
 import std.algorithm.searching : endsWith;
 import std.array : replace;
 import std.datetime;
+import std.format;
 import std.range : zip;
 import std.socket;
 static import std.uni;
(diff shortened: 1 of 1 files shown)


Geod24:
This is wrong, AXFR should provide `format("%s.%s.%s.", key, zoneName, realm)`, not just `key.toString()` as `qname`.

kchulj (approved):
LGTM

Geod24:
Since the qname needs to be fully qualified when doing AXFR, I added a little workaround for #2537 in the process.

kchulj (approved):

PR #2543, opened 2021-09-30
Minor tests improvements

Geod24:
One of the test that tends to fail not too often, but often enough to be noticeable, is `EnrollDifferentUTXO`.
Turns out it was another case of `postTransaction`... Or so I think.


Geod24:
Two of them failed on Flash on the first try but the automatic retry (in the script) means they went green "first try"

hewison-chris (approved):
LGTM

Geod24:
The second commit fixed https://github.com/bosagora/agora/issues/2426

PR #2545, opened 2021-10-01
Move name registry to persistent storage

### LINKED ISSUE #2290: Name registry should have persistence
Currently, restarting the name registry means loosing all data.

mkykadir:
Fixes #2290 

`Validator` and `Flash` nodes publish their registry information. Data is separately stored for different types even they publish same payload. Name registry information was stored on memory (Hash map) previously, storage is moved to persistent storage (Cache DB).

1. `ZoneData` is extended to provide DB access interface
2. Each node type has two tables, table schemes are same for different types, total of 4 tables are added to DB

### Database updates

**Validator and Flash Signature Table**
| Field Name | SQL Type | Attributes |
|------------|----------|------------|
| pubkey     | TEXT     |  PRIMARY KEY |
| signature  | TEXT     |  NOT NULL   |
| sequence   | INTEGER  | NOT NULL   |

**Validator and Flash Addresses Table**
| Field Name | SQL Type | Attributes               |
|------------|----------|--------------------------|
| pubkey     | TEXT     | PRIMARY KEY, FOREIGN KEY |
| address    | TEXT     | PRIMARY KEY, NOT NULL    |
| type       | INTEGER  | NOT NULL                 |

Addresses table is related with Signature table through `pubkey` foreign key. `pubkey` and `address` attributes are composite primary keys. `DELETE` operation on Signature table is cascaded to Addresses table to prevent stale address checks.

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -36,7 +36,7 @@ import std.array : replace;
 import std.datetime;
 import std.format;
 import std.range : zip;
-import std.socket;
+import std.socket : InternetAddress;
 static import std.uni;
 
 /// Implementation of `NameRegistryAPI` using associative arrays
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -378,16 +378,11 @@ public class NameRegistry: NameRegistryAPI
         auto soa = zone.toRR();
         reply.answers ~= soa;
 
-        () @trusted {
-            // Row and ResultRange has `@system` methods
-            ResultRange registries = zone.getRegistries();
-            foreach (registry; registries)
-            {
-                PublicKey registry_pub_key = PublicKey.fromString(registry["pubkey"].as!(string));
-                TypedPayload payload = zone.get(registry_pub_key);
-                reply.answers ~= this.convertPayload(payload, format("%s.%s", registry_pub_key, zone.root.value));
-            }
-        } ();
+        foreach (payload; zone)
+        {
+            reply.answers ~= this.convertPayload(payload, format("%s.%s", 
+                payload.payload.data.public_key, zone.root.value));
+        }
             
         reply.answers ~= soa;
         reply.header.RCODE = Header.RCode.NoError;
@@ -772,16 +767,23 @@ private struct ZoneData
         return results.empty ? 0 : results.oneValue!(ulong);
     }
 
-    /***************************************************************************
+    public int opApply(scope int delegate(TypedPayload) dg) @trusted
+    {
+        int result = 0;
 
-         Returns:
-           All registries in persistent storage
+        auto query_results = this.db.execute(this.query_registry_get);
 
-    ***************************************************************************/
+        foreach (row; query_results)
+        {
+            auto registry_pub_key = PublicKey.fromString(row["pubkey"].as!string);
+            auto payload = this.get(registry_pub_key);
+            result = dg(payload);
 
-    public ResultRange getRegistries () @trusted
-    {            
-        return this.db.execute(this.query_registry_get);
+            if (result)
+                break;
+        }
+
+        return result;
     }
 
     /***************************************************************************
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -17,10 +17,12 @@ import agora.api.FullNode : FullNodeAPI = API;
 import agora.api.Registry;
 import agora.common.DNS;
 import agora.common.Ensure;
+import agora.common.ManagedDatabase;
 import agora.common.Types;
 import agora.consensus.data.ValidatorInfo;
 import agora.crypto.Hash;
 import agora.crypto.Key;
+import agora.crypto.Schnorr: Signature;
 import agora.node.Config;
 import agora.flash.api.FlashAPI;
 import agora.flash.Node;
@@ -32,12 +34,17 @@ import agora.utils.Log;
 import std.algorithm.comparison : among, equal;
 import std.algorithm.iteration : map, splitter;
 import std.algorithm.searching : endsWith;
-import std.array : replace;
+import std.array : array, replace;
+import std.conv;
 import std.datetime;
 import std.format;
 import std.range : zip;
 import std.socket : InternetAddress;
+import std.string;
+
 static import std.uni;
+
+import d2sqlite3 : ResultRange;
 
 /// Implementation of `NameRegistryAPI` using associative arrays
 public class NameRegistry: NameRegistryAPI
@@ -679,6 +686,170 @@ private struct ZoneData
 
     /// The content of the zone
     public TypedPayload[PublicKey] map;
+
+    /// Query for registry count interface
+    private string query_count;
+
+    /// Query for getting all registries
+    private string query_registry_get;
+
+    /// Query for getting payload
+    private string query_payload;
+
+    /// Query for adding registry signature table
+    private string query_signature_add;
+
+    /// Query for adding registry to addresses table
+    private string query_addresses_add;
+
+    /// Database to store data
+    private ManagedDatabase db;
+
+    /***************************************************************************
+
+         Params:
+           type_table_name = Registry table for table name
+           cache_db = Database instance
+
+    ***************************************************************************/
+
+    public this (string type_table_name, ManagedDatabase cache_db)
+    {
+        this.db = cache_db;
+
+        this.query_count = format("SELECT COUNT(*) FROM registry_%s_signature", 
+            type_table_name);
+
+        this.query_registry_get = format("SELECT pubkey " ~
+            "FROM registry_%s_signature", type_table_name);
+
+        this.query_payload = format("SELECT signature, sequence, address, type " ~ 
+            "FROM registry_%s_addresses l " ~
+            "INNER JOIN registry_%s_signature r ON l.pubkey = r.pubkey " ~
+            "WHERE l.pubkey = ?", type_table_name, type_table_name);
+
+        this.query_signature_add = format("REPLACE INTO registry_%s_signature " ~
+            "(pubkey, signature, sequence) VALUES (?, ?, ?)", type_table_name);
+
+        this.query_addresses_add = format("REPLACE INTO registry_%s_addresses " ~
+                    "(pubkey, address, type) VALUES (?, ?, ?)", type_table_name);
+
+        string query_sig_create = format("CREATE TABLE IF NOT EXISTS registry_%s_signature " ~
+            "(pubkey TEXT, signature TEXT NOT NULL, sequence INTEGER NOT NULL, " ~
+            "PRIMARY KEY(pubkey))", type_table_name);
+        
+        string query_addr_create = format("CREATE TABLE IF NOT EXISTS registry_%s_addresses " ~
+            "(pubkey TEXT, address TEXT NOT NULL, type INTEGER NOT NULL, " ~
+            "FOREIGN KEY(pubkey) REFERENCES registry_%s_signature(pubkey) ON DELETE CASCADE, " ~
+            "PRIMARY KEY(pubkey, address))", type_table_name, type_table_name);
+
+        this.db.execute(query_sig_create);
+        this.db.execute(query_addr_create);
+    }
+
+    /***************************************************************************
+
+         Returns:
+           Total number of name registries
+
+    ***************************************************************************/
+
+    public ulong count () @trusted
+    {
+        auto results = this.db.execute(this.query_count);
+        return results.empty ? 0 : results.oneValue!(ulong);
+    }
+
+    /***************************************************************************
+
+         Returns:
+           All registries in persistent storage
+
+    ***************************************************************************/
+
+    public ResultRange getRegistries () @trusted
+    {            
+        return this.db.execute(this.query_registry_get);
+    }
+
+    /***************************************************************************
+
+         Get payload data from persistent storage
+
+         Params:
+           public_key = the public key that was used to register
+                         the network addresses
+
+        Returns:
+            TypedPayload of name registry associated with `public_key`
+
+    ***************************************************************************/
+
+    public TypedPayload get (PublicKey public_key) @trusted
+    {
+        auto results = this.db.execute(this.query_payload, public_key);
+
+        if (results.empty)
+            return TypedPayload.init;
+
+        // Address loop consumes data, gather following first
+        const TYPE node_type = to!TYPE(results.front["type"].as!ushort);
+        const ulong sequence = results.front["sequence"].as!ulong;
+        const Signature signature = Signature.fromString(results.front["signature"].as!string);
+
+        const auto addresses = results.map!(r => r["address"].as!Address).array;
+
+        const RegistryPayload payload =
+        {
+            data:
+            {
+                public_key : public_key,
+                addresses : addresses,
+                seq : sequence,
+            },
+            signature: signature,
+        };
+
+        // CNAME and A cannot exist at the same time for a node
+        TypedPayload typed_payload =
+        {
+            type: node_type,
+            payload: payload,
+        };
+
+        return typed_payload;
+    }
+
+    /***************************************************************************
+
+         Updates matching payload in the persistent storage. Payload is added
+         to the persistent storage when no matching payload is found.
+
+         Params:
+           payload = Payload to update
+
+    ***************************************************************************/
+
+    public void update (TypedPayload payload) @trusted
+    {
+        if (payload.payload.data.addresses.length == 0)
+            return;
+
+        db.execute(this.query_signature_add, 
+            payload.payload.data.public_key,
+            payload.payload.signature,
+            payload.payload.data.seq);
+
+        // There is no need to check for stale addresses 
+        // since `REPLACE INTO` is used and `DELETE` is cascaded.
+        foreach (address; payload.payload.data.addresses)
+        {
+            db.execute(this.query_addresses_add,
+                payload.payload.data.public_key,
+                address,
+                payload.type.to!ushort);
+        }
+    }
 
     /// Fill this from the configuration
     public void fill (in string name, in ZoneConfig config, uint serial)
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -71,13 +71,16 @@ public class NameRegistry: NameRegistryAPI
     private ValidatorInfo[] validator_info;
 
     ///
-    public this (string realm, RegistryConfig config, FullNodeAPI agora_node)
+    public this (string realm, RegistryConfig config, FullNodeAPI agora_node, 
+        ManagedDatabase cache_db)
     {
         assert(config.enabled, "Registry instantiated but not enabled");
 
         this.config = config;
         this.log = Logger(__MODULE__);
         this.agora_node = agora_node;
+        this.validators = ZoneData("validator", cache_db);
+        this.flash = ZoneData("flash", cache_db);
         Utils.getCollectorRegistry().addCollector(&this.collectStats);
 
         const vname = "validators." ~ realm;
@@ -112,21 +115,21 @@ public class NameRegistry: NameRegistryAPI
     private void collectStats (Collector collector)
     {
         RegistryStats stats;
-        stats.registry_validator_record_count = this.validators.map.length;
-        stats.registry_flash_record_count = this.flash.map.length;
+        stats.registry_validator_record_count = this.validators.count();
+        stats.registry_flash_record_count = this.flash.count();
         collector.collect(stats);
     }
 
     /// Returns: throws if payload is not valid
     protected TYPE ensureValidPayload (in RegistryPayload payload,
-        TypedPayload[PublicKey] map) @safe
+        TypedPayload previous) @safe
     {
         // verify signature
         ensure(payload.verifySignature(payload.data.public_key),
                 "Incorrect signature for payload");
 
         // check if we received stale data
-        if (auto previous = payload.data.public_key in map)
+        if (previous != TypedPayload.init)
             ensure(previous.payload.data.seq <= payload.data.seq,
                 "registry already has a more up-to-date version of the data");
 
@@ -165,10 +168,11 @@ public class NameRegistry: NameRegistryAPI
 
     public override const(RegistryPayload) getValidator (PublicKey public_key)
     {
-        if (auto ptr = public_key in this.validators.map)
+        TypedPayload payload = this.validators.get(public_key);
+        if (payload != TypedPayload.init)
         {
-            log.trace("Successfull GET /validator: {} => {}", public_key, *ptr);
-            return (*ptr).payload;
+            log.trace("Successfull GET /validator: {} => {}", public_key, payload);
+            return payload.payload;
         }
         log.trace("Unsuccessfull GET /validators: {}", public_key);
         return RegistryPayload.init;
@@ -195,7 +199,8 @@ public class NameRegistry: NameRegistryAPI
     {
         import std.algorithm;
 
-        TYPE payload_type = this.ensureValidPayload(registry_payload, this.validators.map);
+        TYPE payload_type = this.ensureValidPayload(registry_payload,
+            this.validators.get(registry_payload.data.public_key));
 
         // Last step is to check the state of the chain
         auto last_height = this.agora_node.getBlockHeight() + 1;
@@ -210,8 +215,7 @@ public class NameRegistry: NameRegistryAPI
         // register data
         log.info("Registering addresses {}: {} for public key: {}", payload_type,
                  registry_payload.data.addresses, registry_payload.data.public_key);
-        this.validators.map[registry_payload.data.public_key] =
-            TypedPayload(payload_type, registry_payload);
+        this.validators.update(TypedPayload(payload_type, registry_payload));
     }
 
     /***************************************************************************
@@ -233,10 +237,11 @@ public class NameRegistry: NameRegistryAPI
 
     public override const(RegistryPayload) getFlashNode (PublicKey public_key)
     {
-        if (auto ptr = public_key in this.flash.map)
+        TypedPayload payload = this.flash.get(public_key);
+        if (payload != TypedPayload.init)
         {
-            log.trace("Successfull GET /flash_node: {} => {}", public_key, *ptr);
-            return (*ptr).payload;
+            log.trace("Successfull GET /flash_node: {} => {}", public_key, payload);
+            return payload.payload;
         }
         log.trace("Unsuccessfull GET /flash_node: {}", public_key);
         return RegistryPayload.init;
@@ -259,7 +264,8 @@ public class NameRegistry: NameRegistryAPI
 
     public override void postFlashNode (RegistryPayload registry_payload, KnownChannel channel)
     {
-        TYPE payload_type = this.ensureValidPayload(registry_payload, this.flash.map);
+        TYPE payload_type = this.ensureValidPayload(registry_payload, 
+            this.flash.get(registry_payload.data.public_key));
 
         ensure(isValidChannelOpen(channel.conf, this.agora_node.getBlock(channel.height)),
             "Not a valid channel");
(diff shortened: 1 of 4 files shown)


source/agora/node/Registry.d
```diff
@@ -802,3 +819,338 @@ private struct ZoneData
...
+           payload = Payload to save
+
+    ***************************************************************************/
+    public void addRegistry (TypedPayload payload) @safe;
+    
+    public ZoneData getZoneData () @safe;
+}
+
+/// Manage persistent info of `validators`
+private class ValidatorZone : ZoneInterface{
```
kchulj:
```suggestion
private class ValidatorZone : ZoneInterface
{
```

source/agora/node/Registry.d
```diff
@@ -28,6 +28,8 @@ import agora.serialization.Serializer;
 import agora.stats.Registry;
 import agora.stats.Utils;
 import agora.utils.Log;
+import agora.common.ManagedDatabase;
+import agora.crypto.Schnorr: Signature;
```
Geod24: Minor style issues: We usually try to keep imports alphabetically ordered. They are also grouped by package, and usually ordered by reverse order of dependency (e.g. `agora` > `vibe` > `std` > `core`). Although the order between grouping can vary, please keep the alphabetical order in the import list.

source/agora/node/Registry.d
```diff
@@ -802,3 +808,340 @@ private struct ZoneData
...
+
+private interface ZoneInterface
+{
+    /***************************************************************************
+
+         Returns:
+           Total number of name registries
+
+    ***************************************************************************/
+    public ulong getRegistryCount () @safe;
```
Geod24: Leave a newline between DDOC and the prototype

Geod24:
My main question here is, why the interface and two classes ?

Before, data were stored in `ZoneData`. `ZoneData` handles the common parts of `validators` and `flash` zones handling, and the only thing that actually differs is how we validate inclusion in the map (frozen UTXO for validators, ownership of a channel-looking UTXO for flash).

I could see why a `class` could be desirable if we wanted to somehow abstract the validation, but that's not the case here. AFAICS, the two classes share 99% of their code, and the only difference is the name of the table(s) used. With this in mind, wouldn't it make much more sense to keep the previous struct, but pass the table name as a parameter ?

mkykadir:
I also didn't like code duplication but these are reasons;

- `d2sqlite3` doesn't support table name parameterization, I could construct a query string using table name but it seemed bad practice (SQL injection) to me, even parameter will only passed by us (no user input)
- `ZoneData` was the data storage previously and it could provide the data directly without any interface. Current storage is DB now and its data can be accessed through an interface (queries). Thus, two classes can easily be switched to ORM classes in future.
- I implemented an interface due to code duplication actually. 

If query construction is OK for us, I have an alternative implementation.

Geod24:
> If query construction is OK for us, I have an alternative implementation.

Yep it is, since as you mentioned, there's not risk of user input. From a quick look this new implementation looks much nicer, will review soon.

source/agora/node/Registry.d
```diff
@@ -36,7 +36,7 @@ import std.array : replace;
 import std.datetime;
 import std.format;
 import std.range : zip;
-import std.socket;
+import std.socket : InternetAddress;
```
Geod24, line 42: I don't mind reducing the scope of  the import, but in general, commit messages *should* come with an explanation. In this case, I am wondering, did you do it because it was more hygienic (which is valid) or because you had an issue (e.g. conflicting symbols).

BTW I usually review commit-by-commit, so this applies to the first commit.

source/agora/node/Registry.d
```diff
@@ -37,8 +39,13 @@ import std.datetime;
 import std.format;
 import std.range : zip;
 import std.socket : InternetAddress;
+import std.conv;
+import std.string;
+
 static import std.uni;
 
+import d2sqlite3;
```
Geod24: Let's try not to import the whole package. What part of it is needed ?

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
     /// The content of the zone
     public TypedPayload[PublicKey] map;
 
+    private string type_table_name;
+    private ManagedDatabase db;
+    private Logger log;
```
Geod24: All members should have DDOC. When the use can be derived from the name of the field / common usage, like here, you can simply use `///`
E.g.:
```D
///
private string type_table_name;

///
private ManagedDatabase db;

// [...]
```

The reason we do this is so that fields show up in the generated documentation (fields without docs do not show up).

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+
+    public this(string type_table_name, ManagedDatabase db, Logger logger)
```
Geod24: All declarations should have a DDOC, even if trivial. Additionally, always include a space between the function name and the opening parenthesis of the parameters when defining or declaring it.
This allow to very easily find a method's declaration:
```shell
$ grep -nR 'makeNewBlock (' source/
```

source/agora/node/Registry.d
```diff
@@ -801,4 +996,4 @@ private struct ZoneData
             return null;
         return this.matches(parentDomain) ? child : null;
     }
-}
+}
```
Geod24: Revert this

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
 
+    private string type_table_name;
+    private ManagedDatabase db;
+    private Logger log;
+
+    public this(string type_table_name, ManagedDatabase db, Logger logger)
+    {
+        this.type_table_name = type_table_name;
+        this.db = db;
+        this.log = logger;
```
Geod24: This is good, but since it is your first PR I am also adding comments for general knowledge. So you don't have to change this code.

It is quite easy to mistakenly self-assign in constructors. For example, consider the following:
```D
class Something
{
    ///
    private Height enrolled_height;
    
    ///
    public this (Height enrolled_heigth)
    {
        this.enrolled_height = enrolled_height;
    }
```

Do you see the bug ? The parameter has a typo, meaning the assignment will always self-assign (hence the value will be `0`). For some reason, *some* places in our code use a different convenient, for example `enrolled_height_` for the parameter, so that it's harder to get it wrong. But most places use the approach of naming the parameters after the field.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+
+        try
+        {
+            this.db.execute(query_signature);
+            this.db.execute(query_addresses);
+        }
+        catch (Exception ex)
+        {
+            this.log.warn("ManagedDatabase operation error: {}", ex.msg);
+        }
```
Geod24: I think, in this case, the appropriate course of action might just be to terminate the application.
Otherwise, the registry will be completely non-function, will it not ?

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+    }
+
+    /***************************************************************************
+
+         Returns:
+           Total number of name registries
+
+    ***************************************************************************/
+    public ulong getRegistryCount () @trusted
```
Geod24: Style: One newline in between please

Geod24: Additionally, the usual name used is `length` (as in, the array property).
Using `count` might also make sense to show that it is computationally expensive.
In that case, I'd say we should go with either `length ()` or `count ()`, whichever you prefer.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+            {
+                return results.oneValue!(ulong);
+            }
+        }
+        catch (Exception ex)
+        {
+            this.log.warn("ManagedDatabase operation error: {}", ex.msg);
+        }
+
+        return 0;
```
Geod24: It seems a bit off to me to swallow the Exception and return an in-band value. I know a lot of place swallow exceptions, but it's quite an anti pattern, I'd say let it propagate.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+    }
+
+    /***************************************************************************
+
+         Returns:
+           All registries in persistent storage
+
+    ***************************************************************************/
+    public ResultRange getRegistries () @trusted
```
Geod24: Style: Missing newline.

Regarding the name, it seems a bit ambiguous. Did you mean "registered" ? The registry is the container.
Also I do not see this function used, what is it for ?

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+
+         Params:
+           public_key = the public key that was used to register
+                         the network addresses
+
+        Returns:
+            TypedPayload of name registry associated with `public_key`
+
+    ***************************************************************************/
+    public TypedPayload getRegistryTypedPayload (PublicKey public_key) @trusted
```
Geod24: Style: Newline in between

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+    {
+        string query_signature = format("SELECT signature, sequence FROM registry_%s_signature " ~
+            "WHERE pubkey = ?", this.type_table_name);
+        
+        string query_addresses = format("SELECT address, type FROM registry_%s_addresses " ~
+            "WHERE pubkey = ?", this.type_table_name);
```
Geod24: I would recommend to just pre-allocate all those queries in the ctor

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+        {
+            auto signature_results = this.db.execute(query_signature, public_key.toString());
+
+            if (signature_results.empty)
+                return TypedPayload.init;
+
+            auto addresses_results = this.db.execute(query_addresses, public_key.toString());
+
+            if (addresses_results.empty)
+                return TypedPayload.init;
```
Geod24: Looks like we could just do a `JOIN` here ?

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+
+    /***************************************************************************
+
+         Save name registry payload to the persistent storage
+
+         Params:
+           payload = Payload to save
+
+    ***************************************************************************/
+    public void addRegistry (TypedPayload payload) @trusted
```
Geod24: Ditto style

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+    {
+        string query_signature = format("REPLACE INTO registry_%s_signature " ~
+                "(pubkey, signature, sequence) VALUES (?, ?, ?)", this.type_table_name);
+
+        string query_addresses = format("REPLACE INTO registry_%s_addresses " ~
+                    "(pubkey, address, type) VALUES (?, ?, ?)", this.type_table_name);
```
Geod24: Ditto prealloc

source/agora/node/Registry.d
```diff
@@ -680,6 +687,194 @@ private struct ZoneData
...
+
+        if (payload.payload.data.addresses.length > 0)
```
Geod24: In general prefer early returns, e.g.
```
if (payload.data.addresses.length == 0)
    return;
```

It reduces indentation and gets the special cases "out of the way".

Geod24 (changes requested):
Globally looks good, just a few things.
Also are the commit atomic ? As in, do they each compile and pass tests on their own ?
The documentation commit, I feel, should be folded into the commit that introduces the code.

source/agora/node/Registry.d
mkykadir: Due to conflicting symbols between `agora.common.Types.Address` and `std.socket.Address`

source/agora/node/Registry.d
mkykadir: `doAXFR` uses it to gather Payload for each registry

mkykadir:
Commits are atomic,
- First commit just reduces the import scope, nothing is effected
- Second commit just implements the interfaces but keeps already proven `map` so no code that uses `map` is updated
- Third commit converts `map` usages to newly implemented interfaces and doesn't change anything in the implementation from second commit
- Last commit implements `opApply` for `ZoneData`

source/agora/node/Registry.d
mkykadir: Naming seems consistent with `addRegistry` naming, open for suggestions.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,171 @@ private struct ZoneData
...
+           type_table_name = Registry table for table name
+           cache_db = Database instance
+
+    ***************************************************************************/
+
+    public this (string type_table_name, ManagedDatabase cache_db)
+    {
+        this.db = cache_db;
+
+        this.query_count = format("SELECT COUNT(*) FROM registry_%s_signature", type_table_name);
```
omerfirmak: This works, but templated `ZoneData` could be a better idea. That would let you inline these.

`"SELECT COUNT(*) FROM registry_" ~ TemplateTableName ~ "_signature"` would work perfectly fine with no runtime allocations IIRC.

mkykadir: It's good idea, but `ZoneData` will be initiated two times at most in a Registry and allocations will only made during initialization. Last commit implements the suggestion and I think it results in cleaner `ZoneData` attributes.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,171 @@ private struct ZoneData
...
+        // Address loop consumes data, gather following first
+        const TYPE node_type = to!TYPE(results.front["type"].as!(ushort));
+        const ulong sequence = results.front["sequence"].as!(ulong);
+        const Signature signature = Signature.fromString(results.front["signature"].as!(string));
+
+        Address[] addresses;
+        foreach (Row row; results)
+        {
+            addresses ~= row["address"].as!(Address);
+        }
```
Geod24: Two notes:
1) This seems to be the only usage of `Row`. You could just use `foreach (row; results)` and the type should be inferred (usually... There are some exceptions, and I guess you might have hit that limitation?).
2) A more idiomatic way to write this would be `auto addresses = results.map!(r => r["address"].as!Address).array;`.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,171 @@ private struct ZoneData
...
+
+    public ulong count () @trusted
+    {
+        auto results = this.db.execute(this.query_count);
+        if (!results.empty)
+        {
+            return results.oneValue!(ulong);
+        }
+
+        return 0;
```
Geod24: Prefer early return:
```D
if (results.empty)
    return 0;
```

In this case, a ternary would also be okay.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,171 @@ private struct ZoneData
...
+    public TypedPayload getRegistryTypedPayload (PublicKey public_key) @trusted
+    {
+        auto results = this.db.execute(this.query_payload, public_key.toString());
+
+        if (results.empty)
+            return TypedPayload.init;
+
+        // Address loop consumes data, gather following first
+        const TYPE node_type = to!TYPE(results.front["type"].as!(ushort));
+        const ulong sequence = results.front["sequence"].as!(ulong);
```
Geod24: Style nit: Parenthesis are optional for single template arguments, so people tend to write `as!ulong` much more.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,171 @@ private struct ZoneData
...
+            addresses ~= row["address"].as!(Address);
+        }
+
+        const RegistryPayload payload =
+        {
+            data:
+            {
+                public_key : public_key,
+                addresses : addresses,
+                seq : sequence
```
Geod24: Style nitpick: Use trailing commas, they make later addition more git-friendly:
`seq : sequence,`

We mostly use them in struct initializers and enum declarations.

source/agora/node/Registry.d
```diff
@@ -680,6 +687,171 @@ private struct ZoneData
...
+        if (payload.payload.data.addresses.length <= 0)
+            return;
+
+        db.execute(this.query_signature_add, 
+            payload.payload.data.public_key.toString(),
+            payload.payload.signature.toString(),
+            payload.payload.data.seq);
+
+        // There is no need to check for stale addresses 
+        // since `REPLACE INTO` is used and `DELETE` is cascaded.
```
Geod24, line 858: 👍 for the comment

source/agora/node/Registry.d
```diff
@@ -680,6 +687,171 @@ private struct ZoneData
...
+        foreach (Address address; payload.payload.data.addresses)
+        {
+            db.execute(this.query_addresses_add,
+                payload.payload.data.public_key.toString(),
```
Geod24: You don't have to use `toString` here. `d2sqlite` uses the `toString(sink)` approach.

## How does it work ?

The common signature for `toString()` is `string toString() const;`, however this has a major downside: It forces immutability of the return, and hence usually means allocation. That's pretty bad, especially if the only intent is to print it to stdout after.

So we often use composition via delegate. We define `toString` as:
```D
void toString (scope void delegate (scope const char[]) @safe sink) const
```

That's a bit more complicated, but it allows us to avoid allocation. The `toString` method can pass "pieces" of the string to `sink` to build the string without needing to allocate. For example:
```D
public struct Person
{
    ///
    public string name;
    
    ///
    public string city;
    
    ///
    void toString (scope void delegate (scope const char[]) @safe sink) const
    {
        sink("Name: ");
        sink(this.name.length ? this.name : "(null)");
        sink("City: ");
        sink(this.city.length ? this.city : "(null)");
    }
}
````

Now, if you call this like this:
```D
import core.stdc.stdio;

void main ()
{
    scope print = (scope const char[] data) {
        printf("%.*s", cast(int) data.length, data.ptr);
    }
    Person inst;
    inst.toString(print);
    printf("\n");
}
```

This will not allocate. Even better, you can easily compose with this approach:
```D
public struct Group
{
    ///
    public Person[] members;
    
    ///
    void toString (scope void delegate (scope const char[]) @safe sink) const
    {
        foreach (const ref p; this.members)
        {
            sink("New person: ");
            p.toString(s); // No need to get a `string` allocated in `Person.toString` just to append it somewhere else!
            sink("\n");
        }
    }
}
```

Now that's a lot of manual work and doesn't work with `int` and co... But luckily, some library functions accept a `sink`, such as `formattedWrite`:
```D

public struct Group
{
    ///
    public Person[] members;
    
    ///
    void toString (scope void delegate (scope const char[]) @safe sink) const
    {
        // formattedWrite will just pass the data piecemeal to the `sink`, avoiding intermediate allocations
        foreach (const ref p; this.members)
            formattedWrite(sink, "New person: %s\n", p);
    }
}
```

The great thing about this technique is that it works for pretty much everything where composition is needed. We use it for `toString` a lot, but also for hashing (search for `computeHash`, called from [here](https://github.com/bosagora/crypto/blob/b05147f8844ee8a4e4fd6751df69f66ba30001bd/source/agora/crypto/Hash.d#L139-L140) or binary serialization (look up `serialize`, called from [here](https://github.com/bosagora/serialization/blob/1cb6c19d4b51b82e40438c08d27cdcfc9e487ca7/source/agora/serialization/Serializer.d#L408-L410)).

The only downside with this approach is that you have to stick to the delegate's attributes, unless you template the delegate (but this has other issues - story for another time). So in our case, the delegates have a fixed set of attributes which we never diverge from (and from `toString`, that means only `@safe`).

source/agora/node/Registry.d
```diff
@@ -680,6 +687,171 @@ private struct ZoneData
...
+            type: node_type,
+            payload: payload
+        };
+
+        return typed_payload;
+    }
+
+    /***************************************************************************
+
+         Save name registry payload to the persistent storage
```
Geod24: "Save payload [...]"

source/agora/node/Registry.d
```diff
@@ -677,8 +698,145 @@ private struct ZoneData
...
+
+         Params:
+           payload = Payload to save
+
+    ***************************************************************************/
+
+    public void addRegistry (TypedPayload payload) @trusted
```
Geod24: Regarding naming, I'd say either `update`, `register`, `addAddress`, etc...

source/agora/node/Registry.d
```diff
@@ -666,7 +687,7 @@ private struct TypedPayload
 }
 
 /// Contains infos related to either `validators` or `flash`
-private struct ZoneData
+private struct ZoneData (string TemplateTableName = "")
```
Geod24: In this scenario, I don't think a default value is mandated. You usually *want* the user to provide a default value, because not doing so is most likely to be a bug than not. But then again, see the other notes about templating based on a value.

source/agora/node/Registry.d
```diff
@@ -475,8 +496,8 @@ public class NameRegistry: NameRegistryAPI
 
     ***************************************************************************/
 
-private PublicKey parsePublicKeyFromDomain (in char[] domain,
-    ZoneData zone) @safe
+private PublicKey parsePublicKeyFromDomain (string TemplateRegistry = "") (in char[] domain,
+    ZoneData!(TemplateRegistry) zone) @safe
```
Geod24: Personally I am not too keen on templating a `struct` based on some internal variable. It always comes back to bite you, exactly like it does here. Since D has a nominal type system, you loose substitution for what amounts to the value of a variable.

But if you really want to go with this approach, a better way would be to template the whole `zone` type, e.g.
```D
private PublicKey parsePublicKeyFromDomain (ZoneDataT) (in char[] domain,
    ZoneDataT zone) @safe
```

And let the compiler infer it. You *can* get fancy and put template constraints, either by doing deconstruction:
```D
private PublicKey parsePublicKeyFromDomain (string TableName, ZoneDataT : ZoneData!TableName ) (in char[] domain,
    ZoneDataT zone) @safe
```

(Something like this... Not tested, but it definitely can be expressed) or with constraints:
```D
private PublicKey parsePublicKeyFromDomain (ZoneDataT) (in char[] domain,
    ZoneDataT zone) @safe
    if (is(ZoneDataT : ZoneData!TableName, TableName)
```

But if you're going that far, might as well pass `TableName` as a runtime value to the constructor, OR hide the problem by making the method a member of `ZoneData`.

source/agora/node/Registry.d
```diff
@@ -677,8 +698,145 @@ private struct ZoneData
...
+    }
+
+    /***************************************************************************
+
+         Returns:
+           All registries in persistent storage
+
+    ***************************************************************************/
+
+    public ResultRange getRegistries () @trusted
```
Geod24: I see now why this is needed. It would probably be better to implement an `opApply` then:
https://github.com/bosagora/agora/blob/23f48f8dd0b4f6d1999e2b735a0bfb94cbe4ce47/source/agora/consensus/pool/Transaction.d#L279-L301

source/agora/node/Registry.d
```diff
@@ -677,8 +698,145 @@ private struct ZoneData
...
+         Params:
+           public_key = the public key that was used to register
+                         the network addresses
+
+        Returns:
+            TypedPayload of name registry associated with `public_key`
+
+    ***************************************************************************/
+
+    public TypedPayload getRegistryTypedPayload (PublicKey public_key) @trusted
```
Geod24: Regarding naming:

I would recommend to try avoiding repeating the return type name in the method name, and focus on the semantic meaning of the function (This is a soft rule, to consider with good judgement, because for some simple functions it *does* make sense to repeat the type name).

Here you have a function that takes a `PublicKey` and returns a `TypedPayload`. Obviously there's a relationship between the parameter and the return, so you could go with a generic name (`lookup` ? `get` like AAs ? After all it is the main function for this map-like object), or something more domain-specific (e.g. `findRegistered`).

The last option would be to mimic built-in property. For example, for an associative array, one can do `value in myAA` and get a pointer. If you wanted to do this, you can implement the operator overload. However this would require to handle the lifetime of the return value, so it's probably more trouble than it's worth.

source/agora/node/Registry.d
```diff
@@ -359,14 +370,24 @@ public class NameRegistry: NameRegistryAPI
 
     ***************************************************************************/
 
-    private void doAXFR (in ZoneData zone, ref Message reply) @safe
+    private void doAXFR (string TemplateRegistry = "") (ZoneData!(TemplateRegistry) zone, ref Message reply) @safe
```
Geod24: I see you removed `in`. I suppose it's because the methods of `ZoneData` are not `const` ?

source/agora/node/Registry.d
```diff
@@ -359,14 +370,24 @@ public class NameRegistry: NameRegistryAPI
...
     {
         log.info("Performing AXFR for {} ({} entries)",
-                 zone.root.value, zone.map.length);
+                 zone.root.value, zone.count());
```
Geod24, line 377: If you go with `opApply`, you might want to move this message in two:
`Performing AXFR for {}`
-> Add `+1` to a count at each iteration
`AXFR for {} done: {} entries`

Or just remove the count completely 🤷

Geod24 (changes requested):
Nice improvement! A few more comments.

source/agora/node/Registry.d
mkykadir: Yes, I couldn't make `ZoneData` methods `const` because of `d2sqlite` s methods were not `const`

source/agora/node/Registry.d
mkykadir: `count` is staying for `collectStats` eitherway.

source/agora/node/Registry.d
```diff
@@ -64,13 +71,15 @@ public class NameRegistry: NameRegistryAPI
     private ValidatorInfo[] validator_info;
 
     ///
-    public this (string realm, RegistryConfig config, FullNodeAPI agora_node)
+    public this (string realm, RegistryConfig config, FullNodeAPI agora_node, ManagedDatabase cache_db)
```
linked0:
```suggestion
    public this (string realm, RegistryConfig config, FullNodeAPI agora_node,
        ManagedDatabase cache_db)
```
It exceeds the limit of 80 characters in a line. We do not strictly force 80 characters per line but we recommend it. So if adding a newline doesn't make the code ugly, it would be good to add a newline.

source/agora/node/Registry.d
```diff
@@ -359,14 +370,19 @@ public class NameRegistry: NameRegistryAPI
...
-                 zone.root.value, zone.map.length);
+                 zone.root.value, zone.count());
         auto soa = zone.toRR();
         reply.answers ~= soa;
-        foreach (const ref key, const ref value; zone.map)
-            reply.answers ~= this.convertPayload(value, format("%s.%s", key, zone.root.value));
+
+        foreach (payload; zone)
+        {
+            reply.answers ~= this.convertPayload(payload, format("%s.%s", payload.payload.data.public_key,
```
linked0: Ditto

source/agora/node/Registry.d
```diff
@@ -677,8 +693,169 @@ private struct ZoneData
...
+                    "(pubkey, address, type) VALUES (?, ?, ?)", type_table_name);
+
+        string query_signature_create = format("CREATE TABLE IF NOT EXISTS registry_%s_signature " ~
+            "(pubkey TEXT, signature TEXT NOT NULL, sequence INTEGER NOT NULL, " ~
+            "PRIMARY KEY(pubkey))", type_table_name);
+        
+        string query_addresses_create = format("CREATE TABLE IF NOT EXISTS registry_%s_addresses " ~
+            "(pubkey TEXT, address TEXT NOT NULL, type INTEGER NOT NULL, " ~
+            "FOREIGN KEY(pubkey) REFERENCES registry_%s_signature(pubkey) ON DELETE CASCADE, " ~
+            "PRIMARY KEY(pubkey, address))", type_table_name, type_table_name);
```
linked0: Ditto.
Please refer to SQL statements in other files like, `ValidatorSet.d`

source/agora/test/Base.d
```diff
@@ -845,7 +845,7 @@ public class TestAPIManager
         {
             auto listener = reg.locate!TestAPI(agora_addr);
             assert(listener != typeof(listener).init);
-            super("localrest", config, new RemoteAPI!TestAPI(listener, timeout));
+            super("localrest", config, new RemoteAPI!TestAPI(listener, timeout), new ManagedDatabase(":memory:"));
```
linked0: Ditto

linked0 (approved):
Good work!
Some comments about the `80 character in line`

Geod24 (approved):

Issue #2498, opened 2021-09-16
Duplicated entries in `node_infos` because of trailing slash


Geod24:
A full node doing `GET /node_info`:

```
{"state":1,"addresses":{"_set":{"https://v7.bosagora.io/":true,"https://v3.bosagora.io/":true,"https://v5.bosagora.io/":true,"https://v4.bosagora.io":true,"https://v2.bosagora.io":true,"https://v2.bosagora.io/":true,"https://v6.bosagora.io":true,"https://v3.bosagora.io":true,"https://v5.bosagora.io":true,"https://v7.bosagora.io":true,"https://v4.bosagora.io/":true,"https://v6.bosagora.io/":true}}}
```

We should normalize host names so that there is always a trailing slash.
This is probably because we register slashless addresses: https://github.com/bosagora/infrastructure/blob/7da14eb4d73819c66db935d6a483ab289ebd714e/ansible/roles/agora/templates/config.yaml#L58-L61

But the registry has a static configuration with slashes:
https://github.com/bosagora/infrastructure/blob/7da14eb4d73819c66db935d6a483ab289ebd714e/ansible/roles/agora/templates/config.yaml#L81-L89

PR #2546, opened 2021-10-01
Set: Implement `toString` and `fromString` methods

### LINKED ISSUE #2499: Format `Set` as a array
As seen in https://github.com/bosagora/agora/issues/2498 , `Set!T` is formatted as an AA.
This is both hard to read and extremely wasteful, especially when calling this method:
https://github.com/bosagora/agora/blob/001db65be907b8c4d65e7ab76d9672fd0f6c6d41/source/agora/api/FullNode.d#L222-L240

Which puts the `Set` in the query string. Instead, we should format a `Set` as an array.

kchulj:
As seen in #2498, `Set!T` is formatted as an `AA`. This is both hard to read and extremely wasteful.
Now, `Set` is formatted as an array, using the updated `toString` and newly implemented `fromString` methods.

e.g. a `Set!string` such as `"_set":{"Hello":true,"world":true}`
will now be formatted as `["Hello", "world"]`.

Fixes #2499

### DIFF
--- a/source/agora/common/Set.d
+++ b/source/agora/common/Set.d
@@ -25,10 +25,15 @@ import std.typecons;
 
 import core.stdc.string;
 
+import vibe.data.json;
+
 /// simplified Set() code with some convenience methods,
 /// could use a drop-in implementation later.
 public struct Set (T)
 {
+    /// Helper type for `toString`
+    private alias SinkT = void delegate(scope const(char)[] v) @safe;
+
     ///
     bool[T] _set;
     alias _set this;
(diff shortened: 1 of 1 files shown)


source/agora/common/Set.d
```diff
@@ -236,6 +263,28 @@ unittest
     old_set.each!(value => assert(value in new_set));
 }
 
+/// toString and fromString test for Set!int
+unittest
+{    
+    auto old_set = Set!uint.from([1, 3, 5, 7, 9]);
+    auto str = old_set.toString();
```
Geod24, line 270: The actual value should be tested, because those will pass as long as they are symmetric.

source/agora/common/Set.d
```diff
@@ -51,11 +56,33 @@ public struct Set (T)
...
+    /// Ditto
+    public void toString (scope void delegate(const(char)[]) @safe sink) const
+    {
+        import std.format : formattedWrite;
 
-        return format("%(%s, %)",
-                      this._set.byKeyValue.filter!(tup => tup.value)
-                      .map!(tup => tup.key));
+        formattedWrite(sink, "[%(%s, %)]", this._set.byKeyValue().filter!(
+            kv => kv.value).map!(kv => kv.key));
```
Geod24, line 71: Why is one side using a policy and the other side is not ?

kchulj: The `deserializeWithPolicy` method requires a policy as a parameter, but the contrary `serializeWithPolicy`  method was found to be not necessary for the `toString`.

Should it be used here?

source/agora/common/Set.d
kchulj: Added a check to test the actual value.

kchulj:
Rebased to the latest commit.

Geod24 (approved):

hewison-chris (commented):
LGTM

PR #2547, opened 2021-10-03
Simplify block externalization code and improve some test code along the way

Geod24:
It started with `buildBlock`, but as I was doing some changes, I was hit by other issues:
- The usual spurious failure due to `postTransaction`;
- The unittest `abort`ing with no message whatsoever when a consensus rule is broken;
- Lack of information when the `Ledger` test fails;
- That ugly useless `nothrow` which didn't make sense;
- Redundancy with `-checkaction=context`;

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -1012,6 +1012,36 @@ public class Ledger
 
     /***************************************************************************
 
+        Create a new block based on the current previous block.
+
+        This function only builds a block and will not externalize it.
+        See `acceptBlock` for this.
+
+        Params:
+          txs = An `InputRange` of `Transaction`s
+          time_offset = The `time_offset` of the block compared to the old one
+          enrollments = New enrollments for this block (can be `null`)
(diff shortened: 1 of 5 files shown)


linked0 (approved):
LGTM.
I'd like to know the meaning of `the free function` in the following sentence.
> The Ledger has all the context required to create a Block.
> Instead of calling the free function, we can offer the use the ability
> to create a block via this function, which simplifies the Nominator
> and will simplify tests.

Geod24:
`the free function` => `makeNewBlock`

hewison-chris (approved):
LGTM

kchulj (approved):
LGTM

hewison-chris:
> LGTM. I'd like to know the meaning of `the free function` in the following sentence.
> 
> > The Ledger has all the context required to create a Block.
> > Instead of calling the free function, we can offer the use the ability
> > to create a block via this function, which simplifies the Nominator
> > and will simplify tests.

`Free function` = `non member function`

PR #2552, opened 2021-10-05
BlockHeader: Replace 'missing_validators' and 'random_seed' with 'preimages'

Geod24:
```
The original idea was that we wouldn't need to store the individual pre-images in the block,
and missing_validators would seldom be populated.

However, we ended up needing the individual pre-images, and we currently have multiple
issues related to catchup with unknown pre-images.

This definitely fixes the problem by storing the pre-images in the block.
The downside of this approach is that the BlockHeader's size is now linearly
proportional to the number of validators.

However, as validators are expected to be long-lived, a future optimization would
be to remove earlier pre-images from storages when a node publishes new pre-images,
essentially reducing the storage to one `Hash` per unique enrollment.
```


source/agora/consensus/Ledger.d
```diff
@@ -1058,7 +1098,7 @@ public class Ledger
 
     ***************************************************************************/
 
-    public Hash getRandomSeed (in Height height, in uint[] missing_validators)
+    private Hash[] getRandomSeed (in Height height, in uint[] missing_validators)
```
```suggestion
    private Hash[] getPreImages (in Height height, in uint[] missing_validators)
```

hewison-chris (approved):
LGTM, just a rename suggestion.

source/agora/consensus/Ledger.d
```diff
@@ -827,15 +828,46 @@ public class Ledger
...
-                return "Block: Header's random seed does not match that of known pre-images";
+            const validators = this.getValidators(block.header.height);
+            if (validators.length != block.header.preimages.length)
+                return "Block: Number of preimages does not match active validators";
+            foreach (idx, const ref hash; block.header.preimages)
+            {
+                if (hash is Hash.init)
+                {
+                    // TODO: Check that the block contains a slashing transaction
+                }
```
Geod24, line 842: Should open an issue for this.

source/agora/consensus/Ledger.d
```diff
@@ -827,15 +828,46 @@ public class Ledger
...
+                    pi.height = block.header.height;
+                    pi.hash = hash;
+                    if (!this.addPreimage(pi))
+                        return "Block: Preimages include an invalid non-revealed pre-image";
+                }
+                else
+                {
+                    // TODO: By caching the 'current' hash, we can prevent a semi
+                    // DoS if a node reveal a pre-image far in the future and then
+                    // keep on submitting wrong blocks.
```
Geod24, line 856: And for this

source/agora/consensus/Ledger.d
```diff
@@ -827,15 +828,46 @@ public class Ledger
...
+                    if (hash.hashFull() !is validators[idx].preimage[Height(block.header.height - 1)])
```
Geod24: Why is it hashing here ? :|

hewison-chris: Verifying the pre-image is valid.

source/agora/consensus/Ledger.d
Geod24: https://github.com/bosagora/agora/issues/2554

source/agora/consensus/Ledger.d
Geod24: Actually this whole method is unused now, will remove.

source/agora/consensus/Ledger.d
Geod24: We could just compare them.

linked0:
Great! But the compiler error needs to be fixed.

linked0 (approved):

kchulj (approved):

source/agora/consensus/Ledger.d
hewison-chris: Yes, as the check that it `hashed` is the previous is performed when receiving the `pre-image`.

Geod24:
Compilation issue should be fixed 🤞

source/agora/consensus/data/Block.d
```diff
@@ -92,10 +96,7 @@ public struct BlockHeader
         dg(this.merkle_root[]);
         foreach (enrollment; this.enrollments)
             hashPart(enrollment, dg);
-        dg(this.random_seed[]);
-        hashPart(this.validators.count, dg); // Include just the count of possible signers
-        foreach (validator; this.missing_validators)
-            hashPart(validator, dg);
+        hashPart(this.preimages, dg);
```
TrustHenry, line 99: Is this hashing order right?
According to the order defined above, the order must be after the block height so that the order is not confused.

Geod24: Damn, good catch! Will fix.

PR #2553, opened 2021-10-05
Fix #2284: Schnorr signatures misusage leaks private keys

### LINKED ISSUE #2284: Misusing Schnorr signatures causes leakage of the Validator's private key
image

An attacker can easily calculate any validator's private key by 

1. observing signatures on any 2 blocks from the same validator during the 1008 block enrollment cycle AND
2. observing preimages on the same 2 blocks AND
3. observing the block hashes on the same 2 blocks

All those information above are publicly available to anyone.
After observing those information, the following 2 modular equations can be created:

s<sub>10</sub> = rc + p<sub>10</sub> + (c<sub>10</sub>*v)
s<sub>11</sub> = rc + p<sub>11</sub> + (c<sub>11</sub>*v)

s<sub>10</sub> = signature on block at height 10
p<sub>10</sub> = preimage for block at height 10
c<sub>10</sub> = block hash for block at height 10
v = private key
rc = unknown constant for the 1008 block enrollment cycle

those 2 equations can be solved for rc and v

here is a D problem to demostrate

```D
unittest
{
    import agora.consensus.data.genesis.Test: GenesisBlock;
    import agora.crypto.ECC: Scalar, Point;
    import agora.crypto.Schnorr;
    import agora.utils.Test;
    import std.format;
    Hash random_seed1 = "seed10".hashFull();
    Hash random_seed2 = "seed11".hashFull();
    Hash preimage10 = "preimage10".hashFull();
    Hash preimage11 = "preimage11".hashFull();
    const TimeOffset = 1;
    const Validators = genesis_validator_keys.length;
    // Generate two blocks
    auto block10 = GenesisBlock.makeNewBlock(
        genesisSpendable().take(1).map!(txb => txb.sign()), TimeOffset, random_seed1, Validators);
    auto block11 = block10.makeNewBlock(
        genesisSpendable().take(1).map!(txb => txb.sign()), TimeOffset, random_seed2, Validators);
    Scalar v = genesis_validator_keys[0].secret;
    assert(v.isValid(), "v is not a valid Scalar!");
    Point V = v.toPoint();
    const Scalar rc = Scalar(hashMulti(v, "consensus.signature.noise", 0));
    const Scalar r10 = rc + Scalar(preimage10);
    const Point R10 = r10.toPoint();
    const Scalar r11 = rc + Scalar(preimage11);
    const Point R11 = r11.toPoint();
    // Two messages
    Scalar c10 = block10.hashFull();
    Scalar c11 = block11.hashFull();
    assert(c10 != c11);
    Signature sig10 = block10.header.createBlockSignature(v, preimage10);
    Signature sig11 = block11.header.createBlockSignature(v, preimage11);
    // Verify signatures
    assert(verify(sig10, c10, V));
    assert(verify(sig11, c11, V));
    // Calculate the private key
    // v = (s10 - s11 + p11 - p10) / (c10 - c11)
    Scalar inverseBlockDiff = c10 - c11;
    inverseBlockDiff = inverseBlockDiff.invert();
    Scalar private_key_calculated = (sig10.s - sig11.s + Scalar(preimage11) - Scalar(preimage10)) * inverseBlockDiff;
    assert(private_key_calculated == v,
        format!"Key %s is not matching key %s"
        (private_key_calculated.toString(PrintMode.Clear), v.toString(PrintMode.Clear)));
}
```

Geod24:
Based on #2552

Only the last commit belongs to this PR.
```
This effectively changes the formula used for signing blocks, turning it upside-down:
Instead of offsetting the 'R' by the pre-image as a point,
we commit to the pre-image being the 's' of the signature,
and provide 'R' such as 'S == c * R + X'.
In order to find the adequate 'R' value, one has to know both 'x' and 'r',
as one may only derive CR from the known values, and CR cannot be divided by 'c'.
In the process of this change, the no-longer-used 'getCommitmentNonce' was removed.
```

### DIFF
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -457,7 +457,6 @@ public struct Block
 ///
 unittest
 {
-    import agora.crypto.Schnorr;
     immutable Hash merkle =
         Hash(`0xdb6e67f59fe0b30676037e4970705df8287f0de38298dcc09e50a8e85413` ~
         `959ca4c52a9fa1edbe6a47cbb6b5e9b2a19b4d0877cc1f5955a7166fe6884eecd2c3`);
@@ -825,7 +824,6 @@ unittest
 {
     import agora.consensus.data.genesis.Test: GenesisBlock;
     import agora.crypto.ECC: Scalar, Point;
-    import agora.crypto.Schnorr;
     import agora.utils.Test;
     import std.format;
 
--- a/source/agora/consensus/data/Block.d
+++ b/source/agora/consensus/data/Block.d
@@ -111,26 +111,21 @@ public struct BlockHeader
         Params:
             secret_key = node's secret
             preimage = preimage at the block height for the signing validator
-            offset = enrollment cycle offset
 
         Returns:
             A new signature derived from the provided parameters
 
     ***************************************************************************/
 
-    public Signature sign (in Scalar secret_key, in Hash preimage,
-        ulong offset = 0) const @safe nothrow
+    public Signature sign (in Scalar secret_key, in Hash preimage)
+        const @safe nothrow
     {
-        static import agora.crypto.Schnorr;
-
         // challenge = Hash(block) to Scalar
         const Scalar challenge = this.hashFull();
-        // rc = r used in signing the commitment
-        const Scalar rc = Scalar(hashMulti(secret_key, "consensus.signature.noise", offset));
-        const Scalar reduced_preimage = Scalar(preimage);
-        const Scalar r = rc + reduced_preimage; // make it unique for block height
-        const Point R = r.toPoint();
-        return agora.crypto.Schnorr.sign(secret_key, R, r, challenge);
+        const Scalar s = Scalar(preimage);
+        const Scalar rc = s - secret_key;
+        const Scalar r = rc * challenge.invert();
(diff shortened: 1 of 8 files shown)


hewison-chris (approved):
Excellent :-)

hewison-chris (changes requested):
Block IO unit tests also need to be updated.

source/agora/consensus/data/Block.d
```diff
@@ -892,8 +903,8 @@ unittest
     Signature sig2 = block2.header.sign(v, preimages[0]);
 
     // Verify signatures
-    assert(block1.header.verify(genesis_validator_keys[0].address, sig1));
-    assert(block2.header.verify(genesis_validator_keys[0].address, sig2));
+    assert(block1.header.verify(genesis_validator_keys[0].address, block1.header.preimages[0], sig1.R));
+    assert(block2.header.verify(genesis_validator_keys[0].address, block1.header.preimages[0], sig2.R));
 
     // Calculate the private key by subtraction
     // `s = r + (v * c)`
```
hewison-chris: This will need to be updated to show it still works.

hewison-chris (approved):
LGTM just a small update to show revealing private key when signing two blocks at same height.

hewison-chris:
Screen Shot 2021-10-07 at 10 13 09

PR #2556, opened 2021-10-05
Tests: Move well known pre-images to the WK namespace

Geod24:
```
This makes it consistent with well-known keypair and a bit easier to use,
as one doesn't have to import agora.consensus.validation.Block.
```

Requirement to fix https://github.com/bosagora/agora/pull/2552 properly.


hewison-chris (approved):
LGTM

linked0 (approved):

kchulj (approved):
LGTM

PR #2559, opened 2021-10-05
RPC: RPCConnections can now handle requests in both directions

### LINKED ISSUE #2534: Transactions do not propagate to an outsider node
After the commit [9dd1e4ffc](https://github.com/bosagora/agora/commit/9dd1e4ffcdc90844e5d554e970d509f0448876fb), the following code fails with the log like,
"Client #6 did not receive tx in expected time for height 19".

```
    TestConf conf = {
        outsider_validators : 1,
    };

    auto network = makeTestNetwork!TestAPIManager(conf);
    network.start();
    network.waitForDiscovery();

    network.generateBlocks(Height(GenesisValidatorCycle - 2));
    network.generateBlocks(iota(GenesisValidators + 1), Height(GenesisValidatorCycle - 1));
```

The error log
```
core.exception.AssertError@source/agora/test/Base.d(1070): Check condition failed after timeout of 4 secs and 40 attempts: [source/agora/test/ClearFilter.d:42] Client #6 did not receive tx in expected time for height 19
----------------
??:? void agora.utils.Utility.retryFor!(core.exception.AssertError).retryFor(lazy bool, core.time.Duration, lazy immutable(char)[], immutable(char)[], ulong) [0x10b0d22ee]
??:? void agora.test.Base.TestAPIManager.addBlock!(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result).addBlock(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result, bool, immutable(char)[], int).__lambda13!(ulong).__lambda13(ulong) [0x10b6e91b3]
??:? std.typecons.Flag!("each").Flag std.algorithm.iteration.each!(agora.test.Base.TestAPIManager.addBlock!(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result).addBlock(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result, bool, immutable(char)[], int).__lambda13).each!(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result).each(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result) [0x10b6e8e07]
??:? void agora.test.Base.TestAPIManager.addBlock!(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result).addBlock(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result, bool, immutable(char)[], int) [0x10b6e8afd]
??:? void agora.test.Base.TestAPIManager.generateBlocks!(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result).generateBlocks(std.range.iota!(ulong, ulong).iota(ulong, ulong).Result, agora.common.Types.Height, bool, immutable(char)[], int).__lambda9!(ulong).__lambda9(ulong) [0x10b6e88d1]
```

omerfirmak:
No concurrent request in the same direction.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -25,6 +25,7 @@ module agora.network.Manager;
 import agora.api.Handlers;
 import agora.api.Registry;
 import agora.api.Validator;
+import agora.api.FullNode;
 import agora.common.BanManager;
 import agora.common.Types;
 import agora.common.ManagedDatabase;
@@ -418,8 +419,10 @@ public class NetworkManager
     /// Proxy to be used for outgoing Agora connections
     protected URL proxy_url;
 
+    protected agora.api.FullNode.API owner_node;
+
     /// Ctor
-    public this (in Config config, ManagedDatabase cache, ITaskManager taskman, Clock clock)
+    public this (in Config config, ManagedDatabase cache, ITaskManager taskman, Clock clock, agora.api.FullNode.API owner_node)
     {
         this.log = Logger(__MODULE__);
         this.taskman = taskman;
@@ -432,6 +435,7 @@ public class NetworkManager
             node_config.data_dir);
         this.discovery_task = new AddressDiscoveryTask(&this.addAddresses);
         this.clock = clock;
+        this.owner_node = owner_node;
         this.banman.load();
 
         this.cacheDB.execute(
@@ -1106,19 +1110,27 @@ public class NetworkManager
 
     ***************************************************************************/
 
-    protected API getClient (Address address, Duration timeout)
+    protected agora.api.Validator.API getClient (Address address, Duration timeout)
     {
         import std.algorithm.searching;
-        import vibe.inet.url;
         const url = URL.parse(address);
 
         if (url.schema == "tcp")
         {
-            import agora.network.RPC;
-            return new RPCClient!API(
+            auto owner_validator = cast (agora.api.Validator.API) this.owner_node;
+
+            return owner_validator ?
+                new RPCClient!(agora.api.Validator.API)(
                 url.host, url.port,
                 /* Disabled, we have our own method: */ 0.seconds, 1,
-                timeout, timeout, timeout, 3 /* Hard coded max tcp connections*/);
+                timeout, timeout, timeout, 3 /* Hard coded max tcp connections*/,
+                owner_validator)
+                :
+                new RPCClient!(agora.api.Validator.API)(
+                url.host, url.port,
+                /* Disabled, we have our own method: */ 0.seconds, 1,
+                timeout, timeout, timeout, 3 /* Hard coded max tcp connections*/,
+                this.owner_node);
         }
 
         if (url.schema.startsWith("http"))
@@ -1130,7 +1142,7 @@ public class NetworkManager
             settings.httpClientSettings.connectTimeout = timeout;
             settings.httpClientSettings.readTimeout = timeout;
             settings.httpClientSettings.proxyURL = this.proxy_url;
-            return new RestInterfaceClient!API(settings);
+            return new RestInterfaceClient!(agora.api.Validator.API)(settings);
         }
         assert(0, "Unknown agora schema");
     }
@@ -1180,8 +1192,8 @@ public class NetworkManager
     ***************************************************************************/
 
     public NetworkClient getNetworkClient (ITaskManager taskman,
-        BanManager banman, Address address, API api, Duration retry,
-        size_t max_retries)
+        BanManager banman, Address address, agora.api.Validator.API api,
+        Duration retry, size_t max_retries)
     {
         return new NetworkClient(taskman, banman, address, api, retry,
             max_retries);
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -40,6 +40,7 @@ import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.network.Clock;
 import agora.network.Client;
+import agora.network.RPC;
 import agora.node.Config;
 import agora.consensus.Ledger;
 import agora.utils.InetUtils;
@@ -57,7 +58,7 @@ import std.datetime.stopwatch;
 import std.exception;
 import std.format;
 import std.random;
-import std.range : walkLength;
+import std.range : walkLength, zip;
 
 import core.stdc.time;
 import core.time;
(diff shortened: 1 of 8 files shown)


omerfirmak:
@Geod24 , I can't push to our `vibe-core` fork.

Geod24:
@omerfirmak : Fixed

source/agora/network/RPC.d
```diff
@@ -298,6 +289,12 @@ private class RPCConnection
 
     ///
     alias conn this;
+
+    ///
+    this (TCPConnection conn) @safe
```
Geod24: Attributes. Especially `nothrow` because the compiler generates much more efficient code for `nothrow` ctors.

source/agora/network/Manager.d
```diff
@@ -1101,7 +1105,7 @@ public class NetworkManager
 
     ***************************************************************************/
 
-    protected API getClient (Address address, Duration timeout)
+    protected agora.api.Validator.API getClient (Address address, Duration timeout)
```
Geod24, line 1153: Why does this return a `Validator` API ? What happens for full nodes ?

Geod24 (commented):
I see the main problem is that it creates a dependency between the `NetworkManager` and the calling node. Will take a deeper look tomorrow after the release.

source/agora/network/Manager.d
omerfirmak: `API` was always `Validator.API` , we just have to specify it explicitly now that I have also imported `FullNode.API`.

> What happens for full nodes ?

Nothing, we just don't call `Validator` only endpoints on them.

omerfirmak:
> I see the main problem is that it creates a dependency between the NetworkManager and the calling node. Will take a deeper look tomorrow after the release.

Yeah, I don't like that either. I'll update it if I find a cleaner solution.

omerfirmak:
Updated with discovery thru incoming RPC conns. Integration is happy.

Geod24:
> RPC: Remove RPC client lookup table

Why ?

Geod24:
Please upstream the vibe-core fixes.

omerfirmak:
> 
> 
> > RPC: Remove RPC client lookup table
> 
> Why ?

Dropped it.

> Please upstream the vibe-core fixes.

Will do

source/agora/network/Manager.d
```diff
@@ -129,6 +133,19 @@ public class NetworkManager
             @safe pure nothrow @nogc
         {
             this.address = address;
+            this.api = null;
+            this.onHandshakeComplete = onHandshakeComplete;
+            this.onFailedRequest = onFailedRequest;
```
```suggestion
            this(null, address, onHandshakeComplete, onFailedRequest);
```

source/agora/network/Manager.d
```diff
@@ -129,6 +133,19 @@ public class NetworkManager
...
+        }
+
+        public this (agora.api.Validator.API api,
+            void delegate (scope ref NodeConnInfo node) onHandshakeComplete,
+            bool delegate (in Address address) onFailedRequest,
+            Address address = Address.init)
```
Geod24: Do we need a default value ?
If not, keep the parameters in the same order (e.g. `address` after `api`).

Geod24 (approved):
One last thing

Geod24:
Rebased it and addressed my comments.

omerfirmak:
So this is back

```
[source/agora/test/Base.d:1339] Assertion thrown during test: Trying to access node at address 'Validator-1.boa1xzval2a3cdxv28n6slr62wlczslk3juvk7cu05qt3z55ty2rlfqfc6egsh2.localrest' from 'Validator-1.boa1xzval2a3cdxv28n6slr62wlczslk3juvk7cu05qt3z55ty2rlfqfc6egsh2.localrest' without first creating it
source/agora/test/Base.d:1339 agora.test.Base.TestAPI agora.test.Base.TestNetworkManager.getClient(immutable(char)[], core.time.Duration) [0x103c58d9f]
source/agora/network/Manager.d:186 void agora.network.Manager.NetworkManager.ConnectionTask.connect_canthrow() [0x103d97471]
source/agora/network/Manager.d:177 nothrow void agora.network.Manager.NetworkManager.ConnectionTask.connect() [0x103d9735b]
core/thread/context.d:46 void core.thread.context.Callable.opCall() [0x104a2bb9c]
core/thread/fiber.d:1141 void core.thread.fiber.Fiber.run() [0x104a2bd28]
core/thread/fiber.d:198 fiber_entryPoint [0x104a2bc65]
#### FATAL ERROR: Trying to access node at address 'Validator-1.boa1xzval2a3cdxv28n6slr62wlczslk3juvk7cu05qt3z55ty2rlfqfc6egsh2.localrest' from 'Validator-1.boa1xzval2a3cdxv28n6slr62wlczslk3juvk7cu05qt3z55ty2rlfqfc6egsh2.localrest' without first creating it
This node was started at source/agora/test/BlockRewards.d:45
Full error: core.exception.AssertError@source/agora/test/Base.d(1339): Trying to access node at address 'Validator-1.boa1xzval2a3cdxv28n6slr62wlczslk3juvk7cu05qt3z55ty2rlfqfc6egsh2.localrest' from 'Validator-1.boa1xzval2a3cdxv28n6slr62wlczslk3juvk7cu05qt3z55ty2rlfqfc6egsh2.localrest' without first creating it
```

Geod24:
Yeah, but the worrying part is integration: Maybe my rebase was incorrect, as it seems to always re-establish connections.

Geod24:
Also, it's a node trying to connect to itself. They should ban themselves. I'll take a look.

omerfirmak:
> Yeah, but the worrying part is integration: Maybe my rebase was incorrect, as it seems to always re-establish connections.

That wasnt the case, I will take a look.

omerfirmak:
Ok, I see what is happening. Incoming connections from FullNodes cannot be merged as they can not be identified by a key. So every time a FullNode initiates a connections, that is registered as a different peer. Which easily maxes the listener limit of a node.

omerfirmak:
Also, now that we have both TCP and HTTP endpoints, they both count as a separate peer and take up a peer slot.

Geod24:
Should / can we prioritize the tcp connection in that situation ?

Geod24:
> Ok, I see what is happening. Incoming connections from FullNodes cannot be merged as they can not be identified by a key. So every time a FullNode initiates a connections, that is registered as a different peer. Which easily maxes the listener limit of a node.

Why does it keeps on initiating connections tho ?

Geod24 (approved):

PR #2562, opened 2021-10-06
Workaround SEGV in unittests by not nullifying the fields on shutdown

Geod24:
We're seeing frequent SEGV in unittests because of leftover tasks
trying to access null fields. Until we can properly clean them up,
disable this nullification.


Geod24:
This is green 🥳

linked0 (approved):
Congratulations on all green! LGTM.

kchulj (approved):

PR #2564, opened 2021-10-07
Trivial: Slightly simplify NodePair

Geod24:
```
There's no way to have two fields containing the same object.
If the intent is to restrict access to control methods,
exposing a function is better suited, as was done here.
```


hewison-chris (approved):
LGTM

linked0 (approved):
LGTM

PR #2575, opened 2021-10-12
Extend Vibe's URL implementation for normalization and agora schema

### LINKED ISSUE #2566: Normalize URLs used through Agora
Currently, we pass URL around as string, which can lead to issues such as #2498 .
We also have multiple type of interfaces, which means we need to instantiate a different client depending on the schema.

What we should do:
- Always use a unified (canonical) representation when we print URLs as string;
- Use a specific schema for the TCP connection, e.g. `agora://<domain>[:port]/`, where the default port is `2826`;
- Support `http` and `https` schema, and instantiate the client appropriately;
- Always emit the trailing slash;
- Only emit the port if it differs from the default port for the protocol (e.g. if we see `http://foo.bar:80` we normalize it to `http://foo.bar/`;

Any other things you can think of @mkykadir ?

mkykadir:
Fixes #2566 
Vibe.d's URL implementation is extended. Some of the Vibe.d's URL limitations are fixed;

- Users should input plain URLs, percent-encoding will be handled (`http://example.com/hello~🌍` is valid)
- Puny-codes are not supported (`http://i👍c.io/` is invalid)
- Plain URL will be untouched, normalized representation will be calculated when requested as follows
- URL schema [must be](https://github.com/vibe-d/vibe.d/issues/2619) lowercase and must exist
- `agora` schema support is added with default port as `2826`

**URL Normalization**

1. Lowercase schema and host name (`http://eXAMplE.com` -> `http://example.com/`)
2. Strip default port for schema (`agora://myserver.com:2826` -> `agora://myserver.com/`)
3. Normalize URL path (`/path/./try//` -> `/path/try/`)
4. Add trailing slash (`http://example.com/mypath` -> `http://example.com/mypath/`)
5. Strip files (`http://example.com/path/inner.html` -> `http://example.com/path/`) 
6. Strip anchor (`http://example.com/path/inner.html#anchor` -> `http://example.com/path/`)

**What can be done?**
Following items can also be implemented but didn't implement due to low probability of use cases,

1. Reverse lookup for IP addresses (`http://142.251.36.238/` -> `http://google.com/`)
2. Punycode support
3. Query ordering

### DIFF
--- a/source/agora/common/Types.d
+++ b/source/agora/common/Types.d
@@ -20,6 +20,11 @@ public import agora.crypto.Types;
 
 import geod24.bitblob;
 
+import std.array : appender;
+import std.conv : to;
+import std.string: lastIndexOf, toLower;
+import std.uri: encode;
+
 /// Represents a specific point in time, it should be changed to time_t
 /// after time_t became platform independent
 public alias TimePoint = ulong;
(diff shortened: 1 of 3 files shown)


source/agora/common/Types.d
```diff
@@ -30,6 +35,180 @@ public alias cstring = const(char)[];
...
+    /// The port part of the URL (optional, supports `agora` schema)
+    @property ushort port() const nothrow 
+    {
+        return inner.port ? inner.port : defaultPort(inner.schema); 
+    }
+
+    /// ditto
+    @property port(ushort v) nothrow { inner.port = v; }
+
+    /// Get the default port for the given schema or 0 (supports `agora` schema)
```
omerfirmak, line 75: IMO, anything related to the new agora schema might be better suited to live in our vibe.d fork, if not upstream. That way we would not need the nested structs and `normalize()` could be just a standalone function.

Geod24:
> Plain URL will be untouched, normalized representation will be calculated when requested as follows

What I had in mind was to just normalize it straight after reading the string from the config file.

> Strip files (http://example.com/path/inner.html -> http://example.com/path/)

Should we ?
How do you differentiate between `http://example.com/path/inner.ext` and `http://example.com/path/inner.ext/`) ?
I don't think we should concern ourselves with what I assume is a domain-specific addition. Just use the normalized URL as-is.

> Strip anchor (http://example.com/path/inner.html#anchor -> http://example.com/path/)

Same reasoning as above. If someone added an anchor, well let's use this anchor, it won't work anyways.

Note that URLs for the `agora://` scheme shouldn't include path either, because the path is an HTTP thing, but we want, in the near future, to have:
- `agora://` => Port 2826, TCP binary protocol;
- `https://` => Port 443, REST API
- `http://` => Port 80, REST API

Hence I see normalization done in two steps:
1) Regular URL normalization, which you describe;
2) Scheme-specific validation, which would be point 5 and 6 (and path for `agora://`).

Does that make sense ?

source/agora/common/Types.d
```diff
@@ -30,6 +35,180 @@ public alias cstring = const(char)[];
 /// A network address
 public alias Address = string;
 
+/// Extension of Vibe.d's URL with normalization and `agora` schema capabilities
+public struct URL
+{
```
Geod24, line 40: You probably want to put a `@safe:` at the top of this struct ?

source/agora/common/Types.d
Geod24: We could just just this function: https://github.com/vibe-d/vibe.d/blob/c899425f3af3eafe7194b6b4c6663f746f6775e1/inet/vibe/inet/url.d#L510-L514

We could (should?) also extend the function to include the default port, that sounds fairly natural to me.

Geod24 (commented):
Yeah, I think it's better to just normalize ahead of time and then use an URL.
One way to do it would be to just call `normalize` in the constructor and use that in `Config` and any network communication. Then `alias inner this;`

source/agora/common/Types.d
mkykadir: `registerCommonInternetSchema` is required for parsing URLs starting with `agora` schema properly on Vibe's side (unit tests have an example for that). But registering a schema is not enough for default ports.

Geod24: Which is why I mention extending it by adding a parameter:
```D
void registerCommonInternetSchema(string schema, ushort default_port)
```

mkykadir: Extending this way will require constructing mapping between schema and default port at module level. This function is not part of the `URL` structure but part of the URL module and default port is part of the `URL` structure. Thus, relationship between default port and schema is constructed within the structure. There are no mappings between schema and default ports in Vibe's URL module currently.

mkykadir:
#2590 overwrites this

Issue #2576, opened 2021-10-12
Clear up registration for validator once their frozen UTXO is spent


Geod24:
This depends on #2524 .

Once a frozen UTXO has been spent, the associated registration should be purged.
This could be done after a "grace" period of a few blocks if there is any concern of race condition.

PR #2578, opened 2021-10-13
Registry: Relax rules to allow valid stakes to register

### LINKED ISSUE #2524: Registry: Relax rules to allow valid stakes to register
Currently, only enrolled validators may register. However, this means that nodes are not immediately reachable when they enroll, and there is a delay in between the moment a block is externalized and nodes can get an address to contact those new peers.
If the block interval is too low, or one's retry delay too high, it means a node might miss one or more consensus rounds.

omerfirmak:
Builds on top of #2487

Fixes #2524

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -1677,7 +1677,8 @@ version (unittest)
             const(Block)[] blocks = null,
             immutable(ConsensusParams) params_ = null,
             Duration block_time_offset_tolerance_dur = 600.seconds,
-            Clock mock_clock = null)
+            Clock mock_clock = null,
+            void delegate (in Block, bool) @safe onAcceptedBlock = null)
         {
             const params = (params_ !is null)
                 ? params_
@@ -1712,7 +1713,7 @@ version (unittest)
                 new TransactionPool(cacheDB),
                 stateDB,
                 mock_clock,
-                block_time_offset_tolerance_dur, null);
+                block_time_offset_tolerance_dur, onAcceptedBlock);
         }
 
         ///
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -1309,6 +1309,24 @@ public class Ledger
         return lastPaidHeight(this.getBlockHeight, this.params.PayoutPeriod);
     }
 
+    /***************************************************************************
+
+        Get an UTXO, no double-spend protection.
+
+        Params:
+            hash = the hash of the UTXO (`hashMulti(tx_hash, index)`)
+            value = the UTXO
+
+        Returns:
+            true if the UTXO was found
+
+    ***************************************************************************/
+
+    public bool peekUTXO (in Hash utxo, out UTXO value) nothrow @safe
+    {
+        return this.utxo_set.peekUTXO(utxo, value);
+    }
+
     version (unittest):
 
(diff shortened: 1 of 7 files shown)


omerfirmak:
Alternative to commit#4 is to reject any validators that can not prove a stake, I am not completely sure which way is the right way to go. Feel free to comment.

omerfirmak:
Added the fix for #2576

source/agora/test/Base.d
```diff
@@ -2234,7 +2234,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         .map!(en => validatorAddress(en.index, en.value)).array;
 
     auto full_node_addresses = test_conf.full_nodes.iota.map!(
-        idx => fullNodeAddress(idx)).array;
+        idx => fullNodeAddress(idx)).array ~ ["name.registry"];
```
Geod24, line 2237: Doesn't seem to belong to this commit ?

source/agora/node/Registry.d
```diff
@@ -210,8 +210,14 @@ public class NameRegistry: NameRegistryAPI
             this.validator_info = this.ledger.getValidators(last_height);
             this.validator_info_height = last_height;
         }
+        UTXO utxo;
         ensure(this.validator_info.map!(info => info.address)
-            .canFind(registry_payload.data.public_key), "Not an enrolled validator");
+            .canFind(registry_payload.data.public_key) ||
+            this.ledger.enrollment_manager.enroll_pool
+                .getEnrollments(last_height).canFind!(
```
Geod24: Why do we check the enrollment pool ?

source/agora/consensus/Ledger.d
```diff
@@ -85,7 +85,7 @@ public class Ledger
     private Block last_block;
 
     /// UTXO set
-    private UTXOCache utxo_set;
+    public UTXOCache utxo_set;
```
Geod24: Just move `ValidatingLedger.peekUTXO` in `Ledger` and you can avoid exposing internals

source/agora/node/Registry.d
```diff
@@ -198,8 +197,7 @@ public class NameRegistry: NameRegistryAPI
 
     public override void postValidator (RegistryPayload registry_payload)
     {
-        import std.algorithm;
-        import agora.consensus.data.UTXO;
+        import std.range;
```
Geod24: Move to top level

Geod24 (changes requested):

source/agora/node/Registry.d
omerfirmak: IIRC, that was how we discussed to go about this initially.

This way we don't need to do any checks on the frozen UTXO. If the enrollment is in the pool already, we know for sure that it satisfies all the requirements.

Geod24: But we still risk a race condition if the node's `Enrollment` doesn't make it to the registry.
Of course this can be trivially worked around by the client by sending the `Enrollment` first, but it looks like an un-needed edge case.

Geod24: I guess as a first step it would be fine, though, if the test suite goes green with this approach.

omerfirmak: All green apart from the weird hang in between tests.

Geod24 (approved):

PR #2580, opened 2021-10-14
Ledger: Minor improvements to reduce `makeNewTestBlock` dependencies

Geod24:
It became evident a while ago that we need to better expose the Ledger's ability to generate blocks.
The original approach taken by `makeNewBlock` doesn't quite scale, and requires passing a lot of context nowadays.
In tests, this becomes even more problematic, as we for example need valid block (signed by a majority of validators) for some tests.

For a while, `makeNewTestBlock` was used to work around that problem, but it just shift things. Moving those functionalities to the `Ledger` seems to be the best way forward. Hence the few commits here, and the many that will follow.

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -1798,11 +1798,6 @@ unittest
 
     Block invalid_block;  // default-initialized should be invalid
     assert(ledger.acceptBlock(invalid_block));
-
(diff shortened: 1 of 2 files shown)


linked0 (approved):
LGTM

kchulj (approved):
LGTM

kchulj (approved):
LGTM

PR #2582, opened 2021-10-15
Trivial: Validate parameters to NameRegistry constructor

Geod24:
Previously we were passing an empty realm in some tests.
This was noticed while investigating something unrelated,
and validation for all constructor parameters was added in the process.


linked0 (approved):
LGTM

kchulj (approved):

PR #2585, opened 2021-10-15
Ledger: Cache the result of getValidators

Geod24:
```
Most calls to `getValidators` are done for the current (or next) height,
and historical data are rarely looked up.
We can take advantage of this by caching based on the height,
to reduce the number of memory allocations or database lookups.
```

I'm a bit on the fence about this, since it can defeat some expectations (as seen in `addPreimages` invalidation).
But we call this method a lot and either we need to start passing the array around, or caching like here.

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -116,6 +116,9 @@ public class Ledger
     /// but less than current time + block_time_offset_tolerance
     public Duration block_time_offset_tolerance;
 
+    /// Cache the result of `getValidators` to allow frequent calls
+    protected HeightCache!(ValidatorInfo[]) validatorCache;
+
     /***************************************************************************
 
         Constructor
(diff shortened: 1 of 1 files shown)


linked0 (approved):
Nice!

source/agora/consensus/Ledger.d
```diff
@@ -116,6 +116,9 @@ public class Ledger
     /// but less than current time + block_time_offset_tolerance
     public Duration block_time_offset_tolerance;
 
+    /// Cache the result of `getValidators` to allow frequent calls
+    protected HeightCache!(ValidatorInfo[]) validatorCache;
```
omerfirmak, line 120: Wouldnt it be a less intrusive and more intuitive to have ValidatorSet to do the caching, stuff like addPreimage would be easier to deal with too.

Geod24: Yeah that makes much more sense 🤔

Geod24:
Superseded by #3025

PR #2590, opened 2021-10-18
Change Address alias to Vibe URL

mkykadir:
Part of #2566 

- Remove `string` alias of `Address`
- Extend Vibe's `URL` with serialization under `Address` structure

It might be better to change URL expecting inputs of `Config` from `string` to `Address`. It is not a problem for mapping types (as in `Proxy` config) but sequence types cannot be parsed with string inputs into `URL` struct yet. This way we would be able to throw error at configuration parsing when user is provided invalid URL from config file.

### DIFF
--- a/source/agora/common/BanManager.d
+++ b/source/agora/common/BanManager.d
@@ -61,7 +61,7 @@ public class BanManager
     private static struct BannedList
     {
         /// Internal implementation relies on AA
-        private Status[Address] data;
+        private Status[string] data;
 
         /// Serialization hook
         public void serialize (scope SerializeDg dg) const @safe
@@ -86,7 +86,7 @@ public class BanManager
             size_t length = deserializeLength(dg, opts.maxLength);
             foreach (idx; 0 .. length)
             {
-                Address key = deserializeFull!Address(dg);
+                string key = deserializeFull!string(dg);
                 Status value = deserializeFull!Status(dg);
                 ret.data[key] = value;
             }
@@ -196,9 +196,9 @@ public class BanManager
                     {
                        auto resolved_address = addr_info.address.toAddrString();
                        if (addr_info.address.addressFamily == AddressFamily.INET6)
-                           resolved_address = InetUtils.expandIPv6(resolved_address);
+                           resolved_address = "[" ~ InetUtils.expandIPv6(resolved_address) ~ "]";
 
-                       this.ban(resolved_address);
+                       this.ban(Address(address.schema ~ resolved_address));
                     }
             }
             catch (Exception e)
@@ -303,9 +303,9 @@ public class BanManager
 
     ***************************************************************************/
 
-    public bool isBanned (Address address) @safe nothrow @nogc
+    public bool isBanned (Address address) @safe nothrow
     {
-        if (auto stats = address in this.ips.data)
+        if (auto stats = address.host in this.ips.data)
             return stats.banned_until > this.getCurTime();
 
         return false;
@@ -324,9 +324,9 @@ public class BanManager
 
     ***************************************************************************/
 
-    public TimePoint getUnbanTime (Address address) @safe nothrow pure @nogc
+    public TimePoint getUnbanTime (Address address) @safe nothrow pure
     {
-        if (auto stats = address in this.ips.data)
+        if (auto stats = address.host in this.ips.data)
             return stats.banned_until;
 
         return 0;
@@ -365,7 +365,7 @@ public class BanManager
     private Status* get (Address address) @trusted nothrow pure
     {
         scope(failure) assert(0);  // it will never throw
-        return &this.ips.data.require(address, Status.init);
+        return &this.ips.data.require(address.host, Status.init);
     }
 }
 
@@ -381,9 +381,9 @@ unittest
         public override void load () { }
     }
 
-    const node1 = "node-1";
-    const node2 = "node-2";
-    const whitelistNode = "whitelist-node";
+    const node1 = Address("agora://node-1");
+    const node2 = Address("agora://node-2");
+    const whitelistNode = Address("agora://whitelist-node");
 
     auto banman = new UnitBanMan();
     banman.whitelist(whitelistNode);
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -237,7 +237,8 @@ public class NetworkManager
             const is_validator = key != PublicKey.init;
             if (is_validator)
             {
-                if (address.length && key == this.outer.validator_config.key_pair.address)
+                if (address !is Address.init 
+                    && key == this.outer.validator_config.key_pair.address)
                 {
                     // either we connected to ourself, or someone else is pretending
                     // to be us
@@ -460,7 +461,7 @@ public class NetworkManager
         foreach (ref row; results)
         {
             const address = row.peek!(string)(0);
-            this.addAddress(address);
+            this.addAddress(Address(address));
         }
 
         // add the IP seeds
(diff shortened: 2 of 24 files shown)


source/agora/common/BanManager.d
```diff
@@ -381,50 +381,54 @@ unittest
         public override void load () { }
     }
 
+    myAddress node1 = myAddress("agora://node-1");
+    myAddress node2 = myAddress("agora://node-2");
+    myAddress whitelistNode = myAddress("agora://whitelist-node");
```
Geod24:
```suggestion
    const node1 = myAddress("agora://node-1");
    const node2 = myAddress("agora://node-2");
    const whitelistNode = myAddress("agora://whitelist-node");
```

Can you create a commit with those changes (using `Address` instead of `myAddress` of course) and submit it in its own PR ? That should be a no-brainer to merge.

source/agora/flash/Config.d
```diff
@@ -50,7 +52,7 @@ public struct FlashConfig
     public ushort control_port = 0xB0C;
 
     /// Address to the listener which will receive payment / update notifications
-    public string listener_address;
+    public @Optional string listener_address;
```
Geod24: ?

mkykadir: Stale change, will be reverted

Geod24:
```
node-7_1  | 2021-10-25T09:34:05.844974693Z [main(m8Tl) INF] Listening for requests on http://0.0.0.0:7826/
node-7_1  | 2021-10-25T09:34:05.851552114Z Timer callback failed: Empty URL.
node-7_1  | 2021-10-25T09:34:06.085775498Z Full exception: object.Exception@submodules/vibe.d/inet/vibe/inet/url.d(145): Empty URL.
node-7_1  | 2021-10-25T09:34:06.085810700Z ----------------
node-7_1  | 2021-10-25T09:34:06.085815001Z /usr/include/d/std/exception.d:513 pure @safe void std.exception.bailOut!(Exception).bailOut(immutable(char)[], ulong, scope const(char)[]) [0x5622c5a56010]
node-7_1  | 2021-10-25T09:34:06.085818501Z /usr/include/d/std/exception.d:433 pure @safe bool std.exception.enforce!().enforce!(bool).enforce(bool, lazy const(char)[], immutable(char)[], ulong) [0x5622c5a55f90]
node-7_1  | 2021-10-25T09:34:06.085821901Z submodules/vibe.d/inet/vibe/inet/url.d:142 ref @safe vibe.inet.url.URL vibe.inet.url.URL.__ctor(immutable(char)[]) [0x5622c64d4c80]
node-7_1  | 2021-10-25T09:34:06.085825201Z source/agora/common/Types.d:51 ref @safe agora.common.Types.myAddress agora.common.Types.myAddress.__ctor(immutable(char)[]) [0x5622c5d1ecb0]
node-7_1  | 2021-10-25T09:34:06.085828202Z source/agora/network/Manager.d:667 agora.common.Task.ITimer agora.network.Manager.NetworkManager.startPeriodicNameRegistration() [0x5622c5d138e0]
```

source/agora/network/Manager.d
```diff
@@ -667,7 +667,7 @@ public class NetworkManager
     public ITimer startPeriodicNameRegistration ()
     {
         this.registry_client = this.getNameRegistryClient(
-            this.validator_config.registry_address, 2.seconds);
+            myAddress(this.validator_config.registry_address), 2.seconds);
```
Geod24: So this throws because `registry_address` is `null`.

mkykadir: Yes, I've updated check wrong. It was checking in `getNameRegistryClient`, will update

source/agora/common/Set.d
```diff
@@ -66,21 +66,23 @@ public struct Set (T)
     public void toString (scope void delegate(const(char)[]) @safe sink) const
     {
         import std.format : formattedWrite;
+        import std.conv : to;
 
         formattedWrite(sink, "[%(%s, %)]", this._set.byKeyValue().filter!(
-            kv => kv.value).map!(kv => kv.key));
+            kv => kv.value).map!(kv => kv.key.to!string));
```
Geod24: Wouldn't:
```suggestion
        formattedWrite(sink, `[%("%s", %)]`, this._set.byKeyValue().filter!(
            kv => kv.value).map!(kv => kv.key));
```

Be better as it would not allocate an intermediate string ?

mkykadir: This would result in double quotation marks in `Set!string`

Geod24: Is it the only case ?

source/agora/node/Config.d
```diff
@@ -630,7 +629,7 @@ network:
   - http://192.168.0.42:2826
 `;
         auto conf = parseConfigString!Config(conf_str, "/dev/null");
-        assert(conf.network == [ `http://192.168.0.42:2826` ]);
+        assert(conf.network == [ "http://192.168.0.42:2826" ]);
```
Geod24: Why the change ?

source/agora/network/Manager.d
```diff
@@ -237,7 +237,8 @@ public class NetworkManager
             const is_validator = key != PublicKey.init;
             if (is_validator)
             {
-                if (address.length && key == this.outer.validator_config.key_pair.address)
+                if (address != Address.init
```
Geod24: Prefer `!is` for comparison default-initialized `struct`.

source/agora/node/Runner.d
```diff
@@ -139,7 +140,10 @@ public Listeners runNode (Config config)
     }
 
     bool delegate (in NetworkAddress address) @safe nothrow isBannedDg = (in address) @safe nothrow {
-        return result.node.getBanManager().isBanned(address.toAddressString());
+        try
+            return result.node.getBanManager().isBanned(Address("agora://" ~ address.toAddressString()));
```
Geod24, line 144: This should just use the `host`.

source/agora/test/Base.d
```diff
@@ -2271,7 +2273,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
       node : makeNodeConfig(),
       interfaces: [ makeInterfaceConfig("name.registry") ],
       consensus: makeConsensusConfig(),
-      network : net.nodes.map!(pair => pair.address).array.dup,
+      network : cast(immutable) net.nodes.map!(pair => "http://"~pair.address).array.dup,
```
```suggestion
      network : net.nodes.map!(pair => "http://"~pair.address).array.idup,
```

Geod24: Or:
```suggestion
      network : cast(immutable) net.nodes.map!(pair => "http://"~pair.address).array,
```

source/agora/test/Base.d
```diff
@@ -532,7 +532,7 @@ public class TestBanManager : BanManager
 public struct NodePair
 {
     ///
-    public Address address;
+    public string address;
```
Geod24, line 535: Why this change ?

source/agora/test/Base.d
```diff
@@ -2199,24 +2201,24 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
...
         };
 
         return conf;
     }
 
-    string validatorAddress (size_t idx, KeyPair key)
+    Address validatorAddress (size_t idx, KeyPair key)
     {
-        return format("Validator-%s.%s.localrest", idx, key.address);
+        return Address(format("http://Validator-%s.%s.localrest", idx, key.address));
```
Geod24, line 2216: Since Localrest uses the binary protocol, I think it would make more sense to use `agora://` here. Or its own `localrest://` scheme. But the current approach works, so let's keep that for another day.

Geod24 (approved):
Broadly LGTM. Just a few minor points.

source/agora/test/Base.d
Geod24: Note that this applies to other instances in this file as well.

source/agora/test/Base.d
mkykadir: `NodePair`s are [generated from](https://github.com/bosagora/agora/blob/ef4662921f3aaaae2d733ab3055b899d2f690829/source/agora/test/Base.d#L875) interfaces configuration only with addresses (without schema or port information)

Geod24 (approved):

PR #2599, opened 2021-10-25
Ledger: Respect missing_validators in buildBlock

Geod24:
```
Prior to this change, we implicitly relied on the synchronization of the ValidatorSet
and the missing_validators being passed to the function.
```

Let's see how this fares.

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -1070,12 +1070,22 @@ public class Ledger
         const height = this.getBlockHeight() + 1;
         const validators = this.getValidators(height);
 
-        Hash[] preimages = validators.map!(
-            (in ValidatorInfo validator)
+        Hash[] preimages = validators.enumerate.map!(
+            (in entry)
             {
-                if (validator.preimage.height < height)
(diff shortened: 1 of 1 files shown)


source/agora/consensus/Ledger.d
```diff
@@ -1070,11 +1070,21 @@ public class Ledger
...
+                if (entry.value.preimage.height < height)
+                {
+                    ensure(false,
+                           "buildBlock: Missing pre-image ({} < {}) for index {} ('{}') " ~
+                           "but index is not in missing_validators ({})",
+                           entry.value.preimage.height, height,
+                           entry.index, entry.value.utxo, missing_validators);
+                }
+
                 return validator.preimage[height];
```
linked0:
```suggestion
                return entry.value.preimage[height];
```

linked0 (approved):
LGTM, but one suggestion.

source/agora/consensus/Ledger.d
Geod24: Ah, thanks! Fixed.

PR #2604, opened 2021-10-26
Fix #2572: Use only IPv6 to avoid dual stack issue

### LINKED ISSUE #2572: Fix "Could not listen to address" spurious message on startup
photo_2021-10-11_14-51-01

This triggers because Vibe.d automatically listens to "all interface" and so tries to listen to both `[::]:$PORT` and `0.0.0.0:$PORT`, and only one succeeds. The fix is most likely to provide the bind address explicitly.

Geod24:
As mentioned in the issue, Vibe.d's default is problematic, so override it.
This still accepts IPv4 connections on Linux and Windows.

### DIFF
--- a/source/agora/stats/Server.d
+++ b/source/agora/stats/Server.d
@@ -40,6 +40,9 @@ public class StatsServer
         router.get("/metrics", &handle_metrics);
 
         auto settings = new HTTPServerSettings;
+        // Vibe.d uses both IPv6 and IPv4 all-interfaces at once, which causes
+        // problems on Linux because of dual stack (the second listen fails)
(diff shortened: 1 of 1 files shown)


mkykadir (approved):

Geod24:
@mkykadir mentioned it would unfortunately not work on Windows. For the time being, we let Windows aside, due to a host of issues, and our recommendation for Windows users will be to use Docker for the time being. It is also configurable in Linux, so it *could* lead to problems. But since there is no good solution for us, the only way forward would be to give the user's control over this. Until we do, this is better than the status quo.

PR #2605, opened 2021-10-26
Represent `Set` values as string in JSON

mkykadir:
Formatter only puts quotation marks when it gets `string` typed keys.

### DIFF
--- a/source/agora/common/Set.d
+++ b/source/agora/common/Set.d
@@ -67,20 +67,21 @@ public struct Set (T)
     {
         import std.format : formattedWrite;
 
-        formattedWrite(sink, "[%(%s, %)]", this._set.byKeyValue().filter!(
+        formattedWrite(sink, `[%-("%s", %)"]`, this._set.byKeyValue().filter!(
             kv => kv.value).map!(kv => kv.key));
     }
 
     /// Support for Vibe.d deserialization
     public static SetT fromString (SetT = Set) (string str) @safe
     {
+        import std.conv : to;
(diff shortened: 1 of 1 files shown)


source/agora/common/Set.d
```diff
@@ -268,7 +270,7 @@ unittest
 {    
     auto old_set = Set!uint.from([1, 3, 5, 7, 9]);
     auto str = old_set.toString();
-    assert(str == "[7, 5, 3, 1, 9]");
+    assert(str == "[\"7\", \"5\", \"3\", \"1\", \"9\"]");
```
Geod24:
```suggestion
    assert(str == `["7", "5", "3", "1", "9"]`);
```

source/agora/common/Set.d
```diff
@@ -66,21 +66,23 @@ public struct Set (T)
     public void toString (scope void delegate(const(char)[]) @safe sink) const
     {
         import std.format : formattedWrite;
+        import std.conv : to;
 
         formattedWrite(sink, "[%(%s, %)]", this._set.byKeyValue().filter!(
-            kv => kv.value).map!(kv => kv.key));
+            kv => kv.value).map!(kv => kv.key.to!string));
```
Geod24: See my question on the other PR about intermediate allocation

mkykadir: Double quotes in `Set!string` can be resolved by `%-(%)`, quotation mark before `]` is required for last item in the list

Geod24 (approved):

PR #2607, opened 2021-10-27
Add a configuration for vanity miner and fix a deprecation message

Geod24:
Not much to it: I was debugging something else and the `vanity` main stood in the way, realized it was because there was no configuration for it. Additionally, the deprecation message was triggered in the CI. We're not doing it wrong, but we're not doing it clearly either.


hewison-chris (approved):
LGTM

PR #2618, opened 2021-11-01
DNS: Add more record definition

Geod24:
Extracted from https://github.com/bosagora/agora/pull/2516 and fairly straightforward.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -28,7 +28,19 @@ import std.format;
 import std.string;
 static import std.utf;
 
-/// https://datatracker.ietf.org/doc/html/rfc1035#section-3.2.2
+/*******************************************************************************
+
+    List of record types.
+
+    Originally defined in RFC1035, then extended through many RFCs.
+    A good overview can be obtained from the Wikipedia article linked below.
+
+    See_Also:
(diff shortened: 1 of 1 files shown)


source/agora/common/DNS.d
```diff
@@ -136,6 +177,45 @@ public struct RDATA
...
+{
+    /// The priority of this target host (lowest is better)
+    public ushort priority;
+
+    /// A server selection mechanism for entries with the same priority
+    public ushort weight;
+
+    /// This field holds the URI of the target (RFC 3986)
+    /// Resolution of the URI is according to the definitions for the Scheme of the URI.
+    public string target;
```
mkykadir, line 216: `Address` can be used here

mkykadir (approved):

source/agora/common/DNS.d
Geod24: Yes it could, but I think it's better not to do it here. This module is currently almost standalone: The only two dependencies, outside of stdlib, are `Ensure` (itself a leaf module) and `Serialization` (a library).
There is a possibility that we will later move this to its own library / somewhere else, so let's keep it as a string and have Agora validate it independently for the time being.

PR #2619, opened 2021-11-01
Allow registry node to connect peers from registry

### LINKED ISSUE #2482: Registry should be able to start without a network peer
When we start a `FullNode`, it expects to either know about a registry or have statically defined network peers.
However, if the `FullNode` *is* the registry, we have a chicken-and-egg problem. Hence, a FullNode should be able to start without any known peer provided it's a registry.

Part of #2256

mkykadir:
Fixes #2482

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -760,13 +760,12 @@ public class NetworkManager
 
         while (this.todo_addresses.length)
         {
-            scope (success)
-                this.taskman.wait(this.node_config.retry_delay);
-
             if (this.connection_tasks.length >= MaxConnectionTasks)
             {
                 log.info("Connection task limit reached. Trying again in {}..",
                     this.node_config.retry_delay);
+
+                this.taskman.wait(this.node_config.retry_delay);
                 continue;
             }
 
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -43,6 +43,7 @@ import agora.network.Clock;
 import agora.network.Client;
 import agora.network.RPC;
 import agora.node.Config;
+import agora.node.Registry : NameRegistry;
 import agora.consensus.Ledger;
 import agora.utils.InetUtils;
 import agora.utils.Log;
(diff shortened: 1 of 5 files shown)


source/agora/node/FullNode.d
```diff
@@ -396,7 +396,7 @@ public class FullNode : API
                 .each!(handler => handler.addresses
                     .each!((string address) {
                         auto url = Address(address); // TODO normalize
-                        this.block_header_handlers[url] = this.network.getBlockHeaderUpdatedHandler(url);
```
Geod24: Whitespace fixes in their own commit.

source/agora/network/Manager.d
```diff
@@ -760,6 +761,17 @@ public class NetworkManager
             return true;
         }
 
+        try
+        {
+            auto registry = (cast(FullNode) this.owner_node).getRegistry();
+            if (registry !is null)
+                foreach(addr; registry.getValidatorsAddresses())
+                    this.addAddress(addr);
```
Geod24: It's a bit bothering that we would:
- `cast` an interface to a `class` type (that's a failure of encapsulation);
- Expose an internal member of a class via `public` (I already wasn't happy with `package`);

source/agora/node/FullNode.d
mkykadir: Editor did it automatically on save, I will revert

source/agora/network/Manager.d
omerfirmak: You could easily pass these to `discover()`, as parameters (which is already called by the `owner_node`). No need to cast or expose internals.

source/agora/node/Registry.d
```diff
@@ -756,6 +776,16 @@ private struct ZoneData
         return typed_payload;
     }
 
+    public Address[] getAddresses ()
+    {
+        auto results = this.db.execute(this.query_addresses_get);
+
+        if (results.empty)
+            return new Address[0];
```
Geod24: `return null` ?

Geod24 (approved):
LGTM, just the return, and then feel free to self merge

PR #2621, opened 2021-11-01
DNS message header, bitwise operation fixes

mkykadir:
Fixes bitwise operations in DNS message `Header`. After this fixes, my test fails with 

```
BadResponse: A DNS query response does not respond to the question asked.
```

for following query and answer

``` 
Fullfilled DNS query: 
{ header: ID: a760 QUERY rd NoError, Query: 1, Answer: 0, Authority: 0, Additional: 0, questions: [{ qname: { value: "**masked**.validators.localhost" }, qtype: const(QTYPE).A, qclass: const(QCLASS).IN }], answers: [], authorities: [], additionals: [] } 
=> 
{ header: ID: a760 QUERY qr aa rd NoError, Query: 0, Answer: 1, Authority: 0, Additional: 0, questions: [], answers: [{ name: { value: "**masked**.validators.localhost" }, type: TYPE.A, class_: CLASS.IN, ttl: 0, rdata: [172, 72, 12, 123] }], authorities: [], additionals: [] }
```

Further investigation might be required. Since it returned `FormatError` previously, it couldn't get to the `BadResponse` exception.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -252,7 +252,7 @@ public struct Header
     /// query (0), or a response (1).
     public bool QR () const scope { return !!(this.field2 & 0b1_0000_0_0_0_0_000_0000); }
     /// Ditto
-    public void QR (bool val) scope { this.field2 |= 1 << (ushort.sizeof * 8 - 1); }
+    public void QR (bool val) scope { this.field2 |= val << (ushort.sizeof * 8 - 1); }
 
     /// A four bit field that specifies kind of query in this message.
     /// This value is set by the originator of a query and copied into the response.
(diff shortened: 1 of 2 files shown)


source/agora/common/DNS.d
```diff
@@ -359,7 +359,10 @@ public struct Header
         return cast(RCode) val;
     }
     /// Ditto
-    public void RCODE (RCode val) scope { this.field2 |= val; }
+    public void RCODE (RCode val) scope
+    {
+        this.field2 = (0xfff0 & this.field2) | val;
+    }
```
Geod24, line 365: How is that different from the previous code though ? `val.max == 5`, which fits in 4 bits.

Geod24:
The second commit LGTM. The first commit, I'm not sure. The previous behavior was to have it default to `FormatError` and only change the response code if we found a match. So IIUC, the error changed from the server returning an error code to the server returning invalid data and the client reporting it.

Geod24:
Also, as a general note:
> Use input value instead of always true

This is a descriptive message of what the change is, but it fails a bit short in practice, because it doesn't indicate where the change happens, which make it hard to filter in/out while looking for regressions or categorize using `git log`.

source/agora/common/DNS.d
mkykadir: It was not possible to return back to `NoError` (0) once a return code is set. Or setting `FormatError` first and `ServerFailure` later would result in `NameError`

Geod24: Missed that. Makes perfect sense.

Geod24 (approved):

mkykadir:
> The second commit LGTM. The first commit, I'm not sure. The previous behavior was to have it default to `FormatError` and only change the response code if we found a match. So IIUC, the error changed from the server returning an error code to the server returning invalid data and the client reporting it.

It couldn't return `NoError` due to bitwise operation bug. I believe client was checking for rcode first, now it returns `NoError` and client is now aware of answer is not for the question asked; I will look into that.

PR #2623, opened 2021-11-02
DNS: Fix bad response with missing questions

mkykadir:
Fixes Bad response issue from [here](https://github.com/bosagora/agora/pull/2621#issuecomment-957112958)

Reply [shall include](https://datatracker.ietf.org/doc/html/rfc1034#section-6.2.1) questions from query

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -355,6 +355,8 @@ public class NameRegistry: NameRegistryAPI
                 log.warn("DNS: Ignoring query for unknown QTYPE: {}", q);
                 reply.header.RCODE = Header.RCode.NotImplemented;
             }
+
+            reply.questions ~= q;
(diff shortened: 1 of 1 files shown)


source/agora/node/Registry.d
```diff
@@ -355,6 +355,8 @@ public class NameRegistry: NameRegistryAPI
                 log.warn("DNS: Ignoring query for unknown QTYPE: {}", q);
                 reply.header.RCODE = Header.RCode.NotImplemented;
             }
+
+            reply.questions ~= q;
```
Geod24, line 359: What about AXFR ?

Geod24: Just checked, AXFR include the query too.

Geod24 (approved):

PR #2628, opened 2021-11-03
Logger: Add convenience functions to enable console/file output

Geod24:
This can be useful when someone is developping on Agora and want to do
standalone test, without having to either call `configureLogger` or replace
the log calls with writeln calls.


linked0 (approved):
LGTM

Issue #2625, opened 2021-11-02
DNS: Add support for EDNS(0)


Geod24:
https://datatracker.ietf.org/doc/html/rfc6891

Require for https://github.com/bosagora/agora/issues/2558 as many server will just ignore us otherwise.

PR #2630, opened 2021-11-03
DNS: Add type definitions for EDNS(0)

Geod24:
The RFC itself is pretty simple. The following is enough for usage in a client, integration on the server side might need a bit more. Part of #2625 .

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -584,6 +584,75 @@ public struct ResourceRecord
     }
 }
 
+/// The OPT opcode is a special ResourceRecord with its own semantic
+/// See https://datatracker.ietf.org/doc/html/rfc6891
+public struct OPTRR
+{
+    @safe pure nothrow @nogc:
+
+    /***************************************************************************
+
+        The underlying ResourceRecord
+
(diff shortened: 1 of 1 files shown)


source/agora/common/DNS.d
```diff
@@ -585,6 +584,75 @@ public struct ResourceRecord
...
+    /// Currently only BADVERS (0x01) is supported
+    public ubyte extendedRCODE () scope const
+    {
+        return this.record.ttl >> 24;
+    }
+
+    /// Ditto
+    public void extendedRCODE (ubyte upperByteValue) scope
+    {
+        this.record.ttl = (this.record.ttl & 0x00FF_FFFF) | upperByteValue;
```
mkykadir: Shifting required
`this.record.ttl = (this.record.ttl & 0x00FF_FFFF) | (upperByteValue << 24);`

mkykadir (changes requested):

source/agora/common/DNS.d
Geod24: Good catch

Geod24: Rebased, fixed.

mkykadir (approved):

PR #2637, opened 2021-11-03
DNS: Add support for AD and CD fields in the header

Geod24:
https://www.freesoft.org/CIE/RFC/2065/40.htm

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -397,7 +397,17 @@ public struct Header
     public void RA (bool val) scope { this.field2 |= val << (ushort.sizeof * 8 - 9); }
 
     /// Reserved for future use. Must be zero in all queries and responses.
-    public ubyte Z () const scope { return (this.field2 & 0b0_0000_0_0_0_0_111_0000) >> 4; }
+    public ubyte Z () const scope { return (this.field2 & 0b0_0000_0_0_0_0_100_0000) >> 4; }
+
+    /// Authenticated Data, initially defined in RFC2065
+    public bool AD () const scope { return !!(this.field2 & (1 << 5)); }
+    /// Ditto
+    public void AD (bool val) scope { this.field2 = (this.field2 & ~(1 << 5)) | (val << 5); }
+
+    /// Checking Disabled, initially defined in RFC2065
+    public bool CD () const scope { return !!(this.field2 & (1 << 4)); }
+    /// Ditto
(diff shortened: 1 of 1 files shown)


source/agora/common/DNS.d
```diff
@@ -399,6 +399,16 @@ public struct Header
     /// Reserved for future use. Must be zero in all queries and responses.
     public ubyte Z () const scope { return (this.field2 & 0b0_0000_0_0_0_0_111_0000) >> 4; }
 
+    /// Authenticated Data, initially defined in RFC2065
+    public bool AD () const scope { return !!(this.field2 & (1 << 5)); }
+    /// Ditto
+    public void AD (bool val) scope { this.field2 = (this.field2 & ~(1 << 5)) | (1 << 5); }
```
mkykadir: `val` should be used instead of directly `1`
`this.field2 = (this.field2 & ~(1 << 5)) | (val << 5);`

source/agora/common/DNS.d
```diff
@@ -399,6 +399,16 @@ public struct Header
...
+
+    /// Checking Disabled, initially defined in RFC2065
+    public bool CD () const scope { return !!(this.field2 & (1 << 4)); }
+    /// Ditto
+    public void CD (bool val) scope { this.field2 = (this.field2 & ~(1 << 4)) | (1 << 4); }
```
mkykadir: So both getter and setter knows `CD` is on 4th bit, should we store this info on a mapping (like static enum)?

Geod24: I think it would make sense to have a `mixin` for this.

source/agora/common/DNS.d
Geod24: Riiiiiight... Fixed!

mkykadir (approved):

source/agora/common/DNS.d
```diff
@@ -399,6 +399,16 @@ public struct Header
     /// Reserved for future use. Must be zero in all queries and responses.
     public ubyte Z () const scope { return (this.field2 & 0b0_0000_0_0_0_0_111_0000) >> 4; }
```
mkykadir: This needs an update I think,
https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.1

Geod24: Fixed

mkykadir (approved):

PR #2644, opened 2021-11-05
Fix 2624:  Implement support for DNS name compression

Geod24:
```
The RFC specifies that conformant implementations may not generate
compressed names, but must be able to understand them.
As we are moving towards having the DNS code be a consumer of messages,
and not only a producers, support is becoming a blocker.

As a result, we loose the ability to deserialize individual structs,
and deserialization must only happen at the `Message` level.
```

The second commit is a giant test, taken from real world data.
Compression is actually used inside of it, as can be seen from the screenshot below:

Screen Shot 2021-11-05 at 20 16 42

The `0xc00c` value translates to `0b1100_0000_0000_1100`. The top two `11` are to indicate compression is in use.
The bottom `1100` means it is pointing at offset 12 (in decimal). The header is exactly 12 bytes long, and is followed by a question RR, which first member is the domain.


Geod24:
Tested extensively, so moving ahead with this one.

PR #2645, opened 2021-11-05
Fix #2320: Improve address registration UX

### LINKED ISSUE #2320: Improve `addresses_to_register` configuration UX
Currently, we have `addresses_to_register` under `validator` config.
This has a few issues:
- It requires to specify the protocol being used, e.g. `http://` or `https://`;
- It essentially duplicates configuration (because we have `interfaces` entries);
- Misconfiguration can easily result in hard-to-debug issues;

One suggestion would be to move the address to register to the interface entry, so that we can remove the duplication we have with the protocol. Also bear in mind that our `interface` type might be `http` but if a certificate is present, it's actually going to bind to `https`.

Another is to look up methods to find the IP address of the node, or even its domain name.
Such functions include:
- [gethostname](https://man7.org/linux/man-pages/man2/gethostname.2.html);
- [getaddrinfo](https://man7.org/linux/man-pages/man3/getaddrinfo.3.html);
- [rDNS lookup](https://en.wikipedia.org/wiki/Reverse_DNS_lookup);
- Another external service (à la http://www.mon-ip.com/en/my-ip/ );

Example configuration:
```YAML
interfaces:
  - type:    http
    address: 0.0.0.0
    port:    2826
    register: true
```
The above would attempt to first determine the domain name, then the IP address of the listening interface;

```YAML
interfaces:
  - type:    http
    address: 0.0.0.0
    port:    2826
    register: "node1.bosagora.io"
```
The above would assume that the listening interface is bound to the specified domain name.

Geod24:
```
This removes 'addresses_to_register' from validator, and instead add it to
the related entry in 'interfaces'.
It simplifies things for the user as the port does not need to be repeated,
nor the scheme, and when adding / removing interfaces, only one place
needs to be edited. The port was also made optional now that we have default ports.
```

### DIFF
--- a/source/agora/node/Config.d
+++ b/source/agora/node/Config.d
@@ -32,6 +32,7 @@ import std.algorithm.iteration : splitter;
 import std.algorithm.searching : all;
 import std.exception;
 import std.getopt;
+import std.string : indexOf;
 import std.traits : hasUnsharedAliasing;
 import std.uni : isAlphaNum;
 
@@ -123,11 +124,15 @@ public struct Config
     public void validate () @safe const scope
     {
         if (this.validator.enabled)
+        {
+            enforce(this.interfaces.length > 0,
+                    "Validators must have at least one entry in the 'interfaces' section");
             enforce(this.network.length ||
                     this.validator.registry_address != string.init ||
                     // Allow single-network validator (assume this is NODE6)
                     this.node.limit_test_validators == 1,
                     "Either the network section must not be empty, or 'validator.registry_address' must be set");
+        }
         else
             enforce(this.network.length, "Network section must not be empty");
 
@@ -170,8 +175,27 @@ public struct InterfaceConfig
      /// Bind address
     public string address;
 
-    /// Bind port
-    public ushort port;
+    /***************************************************************************
+
+        Bind port, default to the scheme's default
+
+        Default values are 80 for `http`, 443 for `https`, and 2826 for `tcp`.
+
+    ***************************************************************************/
+
+    public @Optional ushort port;
+
+    /***************************************************************************
+
+        The public address of this interface
+
+        If this binds to an interface that is not publicly available,
+        this value needs to be set for the node to correctly register itself
+        on the registry.
+
+    ***************************************************************************/
+
+    public string hostname;
 
     /// Default values when none is given in the config file
     private static immutable InterfaceConfig[Type.max] Default = [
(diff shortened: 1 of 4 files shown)


source/agora/node/Config.d
```diff
@@ -170,15 +175,74 @@ public struct InterfaceConfig
...
+
+        The public address of this interface
+
+        If this binds to an interface that is not publicly available,
+        this value needs to be set for the node to correctly register itself
+        on the registry.
+
+    ***************************************************************************/
+
+    public string hostname;
```
mkykadir, line 198: A little bit confusing name since it can also include path for `http` schema

source/agora/node/Config.d
```diff
@@ -170,15 +175,74 @@ public struct InterfaceConfig
...
+            break;
+        case Type.https:
+            schema = "https";
+            break;
+        case Type.tcp:
+            schema = "agora";
+            break;
+        }
+
+        return Address(URL(schema, host, this.port, InetPath(path)));
```
mkykadir, line 244: This limits user to use of same port on its machine and public address

mkykadir:
Actually I've encountered with this while normalizing URLs and [test suite provides](https://github.com/mkykadir/agora/blob/d2991d99646dfea0b148bd17142709717236003b/source/agora/test/Base.d#L2193) hostname-like address to `address` of interface

source/agora/node/Config.d
Geod24: We should probably fuzzy parse `hostname` (or whatever we rename it to) and patch it as needed.
So if one wants to have a different port for their public address, they can just give it in `hostname`.

source/agora/node/Config.d
Geod24: Suggestions ? `public_address` ? Verbose but correct.

source/agora/node/Config.d
Geod24: Issue also mentions `register`.

source/agora/node/Config.d
mkykadir: I'm thinking why we need this. I thought a node should be accessible (through its interface) from all addresses that it registers. Why binding with an address at interface config?

Geod24: `hostname` should really just be an override. Ideally, what would happen is:
- User doesn't specify anything and get the default interface(s);
- If user specify something, it's what interface to bind to / what port;
- The node detects this and uses the IP address and configured port to register;

However, there are some cases where it doesn't work:
- Node is behind some sort of proxy: Nginx, a docker container, etc... And it cannot know its final IP / port (our case);
- The user wants to register a domain, not just the IP address;

We could potentially detect the domain via the reverse lookup, but there could be multiple domains pointing to the same host (like us) and virtual host matching.

That's why the override is in place, so the user can have the final say if their config doesn't match the default.

Geod24:
Superseded by #3215

PR #2648, opened 2021-11-08
Trivial: Move some DNS-related function in the right aggregate

Geod24:
As I was moving forward with the DNS implementation, those felt like they weren't where they belong.


mkykadir (approved):
Lgtm

PR #2650, opened 2021-11-08
Registry: Segment operations by ZoneData

Geod24:
Operations were previously done at the top level,
however it means zones could not implement their own logic
(and a lot of duplication was necessary).
It also meant that flash nodes couldn't have their key looked up
in practice if they were not validators as well.

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -83,8 +83,8 @@ public class NameRegistry: NameRegistryAPI
         this.log = Logger(__MODULE__);
 
         this.ledger = ledger;
-        this.validators = ZoneData("validator", cache_db);
-        this.flash = ZoneData("flash", cache_db);
+        this.validators = ZoneData("validator", cache_db, log);
+        this.flash = ZoneData("flash", cache_db, log);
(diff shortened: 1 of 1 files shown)


source/agora/node/Registry.d
```diff
@@ -331,32 +333,25 @@ public class NameRegistry: NameRegistryAPI
...
-                reply.header.RCODE = this.getValidatorDNSRecord(q, reply);
-                if (reply.header.RCODE != Header.RCode.NoError)
-                    break;
+                log.warn("DNS: Ignoring query for unknown QTYPE: {}", q);
+                reply.header.RCODE = Header.RCode.NotImplemented;
+                break;
             }
+
+            if (this.validators.owns(q.qname))
+                this.validators.answer(q, reply);
```
mkykadir, line 344: Setting `RCODE` seems a little bit scattered, `getKeyDNSRecord` returns the value even it also gets `reply` as `ref` parameter as in `answer` and `doAXFR`

mkykadir (approved):

source/agora/node/Registry.d
Geod24: Yeah, it bothers me too, let's try to clean that up.

PR #2652, opened 2021-11-08
Normalize URLs in Agora

### LINKED ISSUE #2566: Normalize URLs used through Agora

mkykadir:
Fixes #2566 

- Upgrades vibe.d to URL normalization capable version (now we can configure `https://hello-🌍.com/뉴스` URLs)
- Normalizes URLs used through Agora

By default, we will expect folder endpoints from configuration file and always add a trailing slash to URLs

### DIFF
--- a/source/agora/common/Types.d
+++ b/source/agora/common/Types.d
@@ -35,20 +35,21 @@ public alias TimePoint = ulong;
 /// An array of const characters
 public alias cstring = const(char)[];
 
-/// A network address, extended version of Vibe.d's URL with serialization
+/// A normalized network address, extended version of Vibe.d's URL with serialization
 public struct Address
 {
 @safe:
     URL inner;
 
-    public this (URL url) immutable
+    public this (URL url)
     {
         this.inner = url;
+        this.inner.normalize(true);
     }
 
     public this (string url)
     {
-        this.inner = URL(url);
+        this(URL(url));
     }
 
     /// Serialization hook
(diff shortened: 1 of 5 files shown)


Geod24:
Maybe we should just normalize in the constructor ?

Geod24:
Needs a Faucet upgrade of Vibe.d

Geod24:
Screen Shot 2021-11-09 at 15 44 01

That looks a bit off.

source/agora/common/Types.d
```diff
@@ -35,22 +35,31 @@ public alias TimePoint = ulong;
...
     }
 
+    /// Returns the normalized form of the Address, see `vibe.d/URL`
+    // public Address normalized () const
+    // {
+    //     Address ret = this;
+    //     ret.normalize(true);
+    //     return ret;
+    // }
+
```
Geod24: Remove ?

Geod24 (approved):

PR #2660, opened 2021-11-10
AXFR should respond when `qname` matches zone

mkykadir:
See commit

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -354,18 +354,15 @@ public class NameRegistry: NameRegistryAPI
                 break;
             }
 
-            if (this.validators.owns(q.qname))
-                this.validators.answer(q, reply);
-            else if (this.flash.owns(q.qname))
-                this.flash.answer(q, reply);
-            else
+            reply.header.RCODE = this.validators.answer(q, reply);
+            if (reply.header.RCODE == Header.RCode.Refused)
+                reply.header.RCODE = this.flash.answer(q, reply);
+
+            if (reply.header.RCODE == Header.RCode.Refused)
             {
                 log.warn("Refusing {} for unknown zone: {}", q.qtype, q.qname);
-                reply.header.RCODE = Header.RCode.Refused;
-            }
-
-            if (reply.header.RCODE != Header.RCode.NoError)
                 break;
+            }
         }
         reply.fill(query.header);
         log.trace("{} DNS query: {} => {}",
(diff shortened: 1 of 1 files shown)


source/agora/node/Registry.d
```diff
@@ -354,18 +354,15 @@ public class NameRegistry: NameRegistryAPI
...
-            }
+            reply.header.RCODE = this.validators.answer(q, reply);
+            if (reply.header.RCODE != Header.RCode.NoError)
+                reply.header.RCODE = this.flash.answer(q, reply);
 
             if (reply.header.RCODE != Header.RCode.NoError)
+            {
+                log.warn("Refusing {} for unknown zone: {}", q.qtype, q.qname);
                 break;
+            }
```
Geod24: I think it's not sound to depend so much on the internals here. The top level should match the `ZoneData` instead of cascading through them. It would avoid falling back to `flash` when a request can be fulfilled (e.g. a query for an inexistent validator will still go through the `flash` `ZoneData`.

mkykadir: What do you think about checking for `Refused` instead of `NoError`?
Like;
`if (reply.header.RCODE == Header.RCode.Refused)`

Geod24: It has the same problem. For example, we could implement whitelisting for AXFR later, and we would then return `Refused` for queries for our zone that do not match the whitelist.

source/agora/node/Registry.d
```diff
@@ -646,14 +643,25 @@ private struct ZoneData
...
+    public Header.RCode answer (in Question q, ref Message reply) @safe
     {
         if (q.qtype == QTYPE.AXFR)
-            this.doAXFR(reply);
+            return this.matches(q.qname) ? this.doAXFR(reply) : Header.RCode.Refused;
+        else if (this.owns(q.qname))
+            return this.getKeyDNSRecord(q, reply);
         else
-            reply.header.RCODE = this.getKeyDNSRecord(q, reply);
+            return Header.RCode.Refused;
```
Geod24, line 664: Bit of style guideline:
Less indentation is better (the code flows more easily). Hence, when using early returns like this, the last branch does not need to be in a `else`.

Geod24 (approved):
Merging because it works, and it's not wrong in itself. While I believe it can be improved, this is still under active development, and will likely change again soon, so making it perfect wouldn't be the best use of our time.

PR #2664, opened 2021-11-11
NetworkManager: Address should never be .init

Geod24:
```
The ConnectionTask should never have an unknown Address,
hence some of this code is dead. Add an assert to validate
this assumption.
```

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -142,6 +142,7 @@ public class NetworkManager
             bool delegate (in Address address) onFailedRequest)
             @safe pure nothrow @nogc
         {
+            assert(address !is Address.init);
             this.address = address;
             this.api = api;
             this.onHandshakeComplete = onHandshakeComplete;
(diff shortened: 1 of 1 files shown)


Geod24:
Note that from what I've seen in the log, it is sometimes `init`, which is wrong. I want to see the node crash in this case.

hewison-chris (approved):
LGTM

source/agora/network/Manager.d
```diff
@@ -142,6 +142,7 @@ public class NetworkManager
             bool delegate (in Address address) onFailedRequest)
             @safe pure nothrow @nogc
         {
+            assert(address !is Address.init);
```
mkykadir, line 145: [RPC uses](https://github.com/bosagora/agora/blob/c795379eff36548751843c196a303f2d04f0457e/source/agora/network/Manager.d#L1390) `Address.init`, discussed this a little with @omerfirmak yesterday.

omerfirmak: Those are used for the incoming connections where we don't know what address server is reachable on.

hewison-chris: @omerfirmak  So we should not ban or whitelist these?

omerfirmak: We should not. Instead of banning them, we should probably remove that peer from the peer list,

Geod24:
Superseded by https://github.com/bosagora/agora/pull/2845

PR #2667, opened 2021-11-13
Refactor Registry

mkykadir:
Refactor Registry code to

- Separate registry and zone clearly
- Move all zone initialization into its ctor (e.g. `fill`)
- Fix [the issue](https://github.com/bosagora/agora/pull/2660#discussion_r746573025) mentioned in #2660

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -70,6 +70,11 @@ public class NameRegistry: NameRegistryAPI
     ///
     private ValidatorInfo[] validator_info;
 
+    /// Supported DNS query types
+    private immutable QTYPE[] supported_query_types = [
+        QTYPE.A, QTYPE.CNAME, QTYPE.AXFR, QTYPE.ALL,
+    ];
+
     ///
     public this (string realm, RegistryConfig config, Ledger ledger,
         ManagedDatabase cache_db)
@@ -383,7 +388,7 @@ public class NameRegistry: NameRegistryAPI
                 break;
             }
 
-            if (!q.qtype.among(QTYPE.A, QTYPE.CNAME, QTYPE.AXFR, QTYPE.ALL))
+            if (!supported_query_types.canFind(q.qtype))
             {
                 log.warn("DNS: Ignoring query for unknown QTYPE: {}", q);
                 reply.header.RCODE = Header.RCode.NotImplemented;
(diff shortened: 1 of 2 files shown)


source/agora/node/Registry.d
```diff
@@ -407,6 +407,33 @@ public class NameRegistry: NameRegistryAPI
...
+            return &this.validators.answer_owns;
+        else if (this.validators.matches(name))
+            return &this.validators.answer_matches;
+        else if (this.flash.owns(name))
+            return &this.flash.answer_owns;
+        else if (this.flash.matches(name))
+            return &this.flash.answer_matches;
+        else
+            return null;
+    }
```
Geod24: This will do 4 comparisons for every non-matching zones. It feels a bit odd to compare twice.
I really think the better fix would be to have a way to bridge `owns` and `matches` since they are obvious duplication.

source/agora/node/Registry.d
```diff
@@ -611,45 +596,63 @@ private struct ZoneData
...
             "(pubkey TEXT, address TEXT NOT NULL, type INTEGER NOT NULL, " ~
             "FOREIGN KEY(pubkey) REFERENCES registry_%s_signature(pubkey) ON DELETE CASCADE, " ~
-            "PRIMARY KEY(pubkey, address))", type_table_name, type_table_name);
+            "PRIMARY KEY(pubkey, address))", zone_name, zone_name);
 
         this.db.execute(query_sig_create);
         this.db.execute(query_addr_create);
+
+        // Serial's value wraps around so the cast is safe
+        const serial = cast(uint) Clock.currTime(UTC()).toUnixTime();
```
Geod24, line 688: Note that the reason why `serial` was passed down was so that they would both have the same serial. However, considering it again, I'm not sure it really is a problem.

source/agora/node/Registry.d
mkykadir: New method of the zone finding will resolve this problem now and also opens a way to implement even more sub-zones in future; e.g. `verytrusted.validators.localhost` next to `validators.localhost`

source/agora/node/Registry.d
mkykadir: `SERIAL` is a way to detect data changes, since these are separate zones I don't think this would be a problem

hewison-chris (approved):
LGTM, just a typo in second commit message (separately)

Issue #559, opened 2020-02-19
[C++] Some types definitions are missing copy constructor


Geod24:
The most important being, obviously, `vector`. `shared_ptr` is affected too, and anything that use RC.
I'm going to go over them soon, but there are a few upstream bugs I need to sort out first.

omerfirmak:
I think all of our STL bindings, apart from `std::vector`, now have proper copy ctors.

omerfirmak:
this is blocking us from going forward with `std::vector` bindings https://issues.dlang.org/show_bug.cgi?id=22505

omerfirmak:
https://github.com/dlang/druntime/blob/1a2f2158345ece9fc3eec098a55519644bf6dfd0/src/rt/lifetime.d#L2066

this only does postblitting (if possible) after extending the array, no check for elaborate copy ctors.

omerfirmak:
This is now blocked, waiting for a new compiler release to include both https://github.com/dlang/phobos/pull/8316 and https://github.com/dlang/phobos/pull/8315

omerfirmak:
done in #2668

PR #2668, opened 2021-11-14
CPP: Add copy ctors for std::vector and SCPStatement

omerfirmak:
New `dlist.d` is only there until https://github.com/dlang/phobos/pull/8315 is merged and included in a release.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -568,7 +568,7 @@ extern(D):
         {
             if (!proc)
             {
-                this.queued_envelopes.insertBack(cast()envelope);
+                this.queued_envelopes.insertBack(envelope.serializeFull.deserializeFull!SCPEnvelope());
                 this.envelope_timer.rearm(EnvTaskDelay, Periodic.No);
                 continue;
             }
@@ -1106,22 +1106,14 @@ extern(D):
 
     public override void emitEnvelope (ref const(SCPEnvelope) envelope) nothrow
     {
-        SCPEnvelope env = cast()envelope;
         log.trace("Emitting envelope: {}", scpPrettify(&envelope, &this.getQSet));
-
+        SCPEnvelope copy;
         try
-        {
-            // deep-dup as SCP stores pointers to memory on the stack
-            env.statement.pledges = deserializeFull!(SCPStatement._pledges_t)(
-                serializeFull(env.statement.pledges));
-        }
-        catch (Exception ex)
-        {
-            assert(0, ex.to!string);
-        }
-
+            copy = envelope.serializeFull.deserializeFull!SCPEnvelope();
+        catch (Exception e)
+            assert(0);
         log.dbg("{}: peers are {}", __PRETTY_FUNCTION__, this.network.peers[]);
-        this.network.validators().each!(c => c.sendEnvelope(env));
+        this.network.validators().each!(v => v.sendEnvelope(copy));
     }
 
     /***************************************************************************
--- a/source/scpd/Cpp.d
+++ b/source/scpd/Cpp.d
@@ -41,12 +41,11 @@ public enum CppCtor { Use = 0 }
 public alias std_string = basic_string!char;
 
 extern(C++) {
-nothrow @nogc:
-    void defaultCtorCPPObject(T) (T* ptr);
-    void dtorCPPObject(T) (T* ptr);
-    void opAssignCPPObject(T) (T* lhs, T* rhs);
-    void copyCtorCPPObject(T) (T* ptr, inout(T)* rhs);
-    void copyCtorCPPObject(T) (immutable(T)* ptr, inout(T)* rhs);
+nothrow @nogc pure:
+    void defaultCtorCPPObject(T) (inout(T)* ptr);
+    void dtorCPPObject(T) (inout(T)* ptr);
+    void opAssignCPPObject(T) (inout(T)* lhs, inout(T)* rhs);
+    void copyCtorCPPObject(T) (inout(T)* ptr, inout(T)* rhs);
     int getCPPSizeof(T) () @safe;
     std_string sliceToStdString (const(char)* ptr, size_t length);
 }
@@ -55,31 +54,31 @@ private mixin template CPPBindingMixin (T, bool Copyable = true, bool DefaultCon
 {
     extern(D)
     {
-        static if (!DefaultConstructable) @disable this();
-        this (CppCtor use) @trusted nothrow @nogc
+        import std.traits : Fields, FieldNameTuple;
+
+        static if (!DefaultConstructable)
+            @disable this ();
+
+        this (CppCtor use) inout @trusted nothrow @nogc pure
         {
             defaultCtorCPPObject!T(&this);
         }
 
-        ~this () @trusted nothrow @nogc
+        ~this () inout @trusted nothrow @nogc pure
         {
             dtorCPPObject!T(&this);
         }
 
+        @disable this (this);
+
         static if (Copyable)
-        this (ref return scope inout T rhs) @trusted nothrow @nogc
+        this (ref return scope inout T rhs) inout @trusted nothrow @nogc pure
         {
             copyCtorCPPObject!T(&this, &rhs);
         }
 
         static if (Copyable)
-        this (ref return scope inout T rhs) immutable @trusted nothrow @nogc
-        {
-            copyCtorCPPObject!T(&this, &rhs);
-        }
-
-        static if (Copyable)
-        void opAssign()(auto ref T rhs) @trusted nothrow @nogc
+        void opAssign()(auto ref inout(T) rhs) inout @trusted nothrow @nogc pure
         {
             opAssignCPPObject!T(&this, &rhs);
         }
@@ -325,6 +324,8 @@ extern(C++, (StdNamespace)) extern(C++, class) struct vector (T, Alloc = allocat
 
     alias ElementType = T;
 
+    mixin CPPBindingMixin!(vector, true, true);
+
     extern(D)
     {
         /// TODO: Separate from `vector` definition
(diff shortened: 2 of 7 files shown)


omerfirmak:
This compiles fine on ldc 1.27.0 :/

omerfirmak:
This commit breaks it https://github.com/ldc-developers/phobos/commit/73e51d395ebee36029f337b3b97bd5efbcb694a4

omerfirmak:
Now also depends on https://github.com/dlang/phobos/pull/8316

omerfirmak:
They (https://github.com/dlang/phobos/pull/8315 and https://github.com/dlang/phobos/pull/8316) are both merged, I will rerun this when they land in `ldc-master`.

omerfirmak:
dmd-2.098.1 is released, giving this a spin again.

omerfirmak:
Works fine with RPC & LocalRest, but with HTTP endpoints we are not externalizing any blocks.

omerfirmak:
Fixed the bug that caused envelopes to get partially serialized in vibe.d
But there is now a runtime crash when using HTTP interfaces between validators, RPC works fine with no crashes.

omerfirmak:
> But there is now a runtime crash when using HTTP 

Have been working on this for the past two days, using dmd-2.098.1

[This ](https://issues.dlang.org/show_bug.cgi?id=22505)came back to bite us. Tried to fix the issue itself but turns out, it is a really involved one. Added a workaround for it in the meantime.

Marking ready for reviews before the new ldc release.

source/agora/consensus/protocol/Nominator.d
```diff
@@ -565,7 +565,7 @@ extern(D):
         {
             if (!proc)
             {
-                this.queued_envelopes.insertBack(cast()envelope);
+                this.queued_envelopes.insertBack(envelope.serializeFull.deserializeFull!SCPEnvelope());
```
hewison-chris, line 571:
```suggestion
                this.queued_envelopes.insertBack(envelope.clone());
```

source/agora/consensus/protocol/Nominator.d
```diff
@@ -1084,21 +1084,13 @@ extern(D):
...
-            env.statement.pledges = deserializeFull!(SCPStatement._pledges_t)(
-                serializeFull(env.statement.pledges));
-        }
-        catch (Exception ex)
-        {
-            assert(0, ex.to!string);
-        }
-
-        this.network.validators().each!(v => v.client.sendEnvelope(env));
+            copy = envelope.serializeFull.deserializeFull!SCPEnvelope();
```
hewison-chris, line 1112:
```suggestion
            copy = envelope.clone();
```

hewison-chris (approved):
LGTM, maybe make  use of clone() though :-)

omerfirmak:
> LGTM, maybe make use of clone() though :-)

Actually we dont need serializer for deep copies anymore. We have copy ctors for all CPP types now :)

omerfirmak:
Updated ldc to the newest release, once the alpine ldc issue is fixed we can probably finally get this merged.

source/scpd/Cpp.d
```diff
@@ -497,8 +498,10 @@ unittest
     iota(5).each!((num) {ubyte b = cast(ubyte)num; vec_ubyte.push_back(b);});
     auto serialized = vec_ubyte.toString();
     assert(serialized == `"AAECAwQ="`, "actual serialized: " ~ serialized);
-    auto deserialized = vec_ubyte.fromString(serialized)[];
-    assert(deserialized == [0, 1, 2, 3, 4], "actual deserialized: " ~ to!string(deserialized));
+    vector!ubyte deserialized = vec_ubyte.fromString(serialized);
+    assert(deserialized[] == [0, 1, 2, 3, 4], "actual deserialized: " ~ to!string(deserialized));
```
Geod24: You can remove the message now

source/scpd/types/Stellar_SCP.d
```diff
@@ -102,6 +102,71 @@ struct SCPStatement {
...
+                case SCPStatementType.SCP_ST_NOMINATE:
+                    destroy(this.nominate_);
+                    break;
+            }
+        }
+
+        this (ref return scope inout(_pledges_t) rhs) inout @trusted nothrow @nogc pure
+        {
+            this.type_ = rhs.type_;
+            if (this.type_== SCPStatementType.SCP_ST_PREPARE)
```
Geod24, line 127: `final switch` ?

source/scpd/types/Stellar_SCP.d
```diff
@@ -102,6 +102,71 @@ struct SCPStatement {
...
+            else if (this.type_== SCPStatementType.SCP_ST_EXTERNALIZE)
+                this.externalize_ = rhs.externalize_;
+            else if (this.type_== SCPStatementType.SCP_ST_NOMINATE)
+                this.nominate_ = rhs.nominate_;
+        }
+
+        void opAssign()(auto ref _pledges_t rhs) @trusted nothrow @nogc pure
+        {
+            this.type_ = rhs.type_;
+            if (this.type_== SCPStatementType.SCP_ST_PREPARE)
```
Geod24, line 140: Ditto `final switch`

source/scpd/types/Stellar_SCP.d
```diff
@@ -226,76 +291,7 @@ struct SCPStatement {
...
-
-            We need to initialize using a literal to account for type
-            constructors, but we can't initialize the `union` in a generic way
-            (because we need to use a different name based on the `type`).
-            The normal solution is to put it in a `switch`, but since we
-            declare multiple variable (one per `switch` branch),
-            NRVO is disabled.
-            The solution is to use `static if` to ensure the compiler only sees
-            one temporary and does NRVO on this function, which in turn enables
-            NRVO on the caller.
```
Geod24, line 255: I had forgotten about this black magic. I remember going through a lot of suffering to make sure it works, so I'll give it a spin tomorrow to make sure we do not regress.

source/scpd/Cpp.d
```diff
@@ -69,7 +69,13 @@ private mixin template CPPBindingMixin (T, bool Copyable = true, bool DefaultCon
...
         }
 
-        @disable this (this);
+        // Workaround for https://issues.dlang.org/show_bug.cgi?id=22505
+        static if (Copyable)
+        this (this) @trusted nothrow @nogc pure
+        {
+            T temp = T(this);
+            copyCtorCPPObject!T(&this, &temp);
+        }
```
Geod24: Waait what ? We have a copy ctor, a postblit should never be called.

Geod24 (commented):
Need to give this some testing

source/scpd/Cpp.d
omerfirmak: `druntime` works in mysterious ways

Geod24:
Can you make a PR with the first commit so I can do the CI juggling already ?

omerfirmak:
Fixes #559

source/scpd/types/Stellar_SCP.d
omerfirmak: Fails to compile, a false positive tho

https://github.com/bosagora/agora/runs/4849358883?check_suite_focus=true#step:11:99

omerfirmak:
Rebased and finally ready to be merged I think.

omerfirmak:
@Geod24 , we need this merged to get rid of the serializer deep dups, they are costing us a lot of cpu time.

Geod24:
Fails to build though

source/scpd/Cpp.d
Geod24: Do you know what calls it ?

omerfirmak: Dynamic array runtime, while trying to relocate the array.

omerfirmak:
> Fails to build though

Can't figure out why tho, we have `inout` copy ctors for `SCPStatement.pledges`, so SCPEnvelope should be copyable right?

omerfirmak:
Dropped the last commit which was failing to compile

source/scpd/types/Utils.d
```diff
@@ -31,14 +31,9 @@ public inout(opaque_vec!()) toVec (scope ref inout(Hash) data) nothrow @nogc
...
-    {
-        base: {
-            _start: data.ptr,
-            _end: data.ptr + data.length,
-            _end_of_storage: data.ptr + data.length,
-        },
-    };
+    inout opaque_vec!() ret;
+    foreach (elem; data)
+        ret.base.push_back(cast(ubyte) elem);
```
Geod24, line 36: This is a rather major change, as the previous version did not allocate

source/scpd/types/Stellar_SCP.d
Geod24: For ref:
```
Type 'FlashControlAPI' implements more than one interface: make sure the one describing the REST server is the first one
source/scpd/types/Stellar_SCP.d(130,17): Error: inout field `prepare_` initialization is not allowed in loops or after labels
source/scpd/types/Stellar_SCP.d(133,17): Error: inout field `confirm_` initialized multiple times
       Previous initialization is here.
source/scpd/types/Stellar_SCP.d(136,17): Error: inout field `externalize_` initialized multiple times
       Previous initialization is here.
source/scpd/types/Stellar_SCP.d(139,17): Error: inout field `nominate_` initialized multiple times
       Previous initialization is here.
ldc2 failed with exit code 1.
```

Geod24: Yeah I think it's another instance of https://issues.dlang.org/show_bug.cgi?id=21080

source/scpd/types/Utils.d
omerfirmak: Previous version was wrong tho, giving `std::vector` control of a GC allocated buffer seems dangerous.

Geod24: True, I thought there was a clever trick that was used that make this legit, but I can't find it (and everything I find confirms it was wrong). Can we at least call `reserve` before pushing bytes by bytes though ? :D

PR #2669, opened 2021-11-15
ValidatorSet: Reject preimages that are too distant to the prev one

### LINKED ISSUE #2248: getPreimage() method call can force a node to hash excessively
NodeA can call NodeB(attacker) who can return a preimage for Height 'ulong - 1', and causing NodeA to hash 'ulong - 1 - height known by NodeA' times

omerfirmak:
Fixes #2248

This will make sure that an attacker can't fool us into hashing
excessively. The genuine case where a node may want to reveal a
preimage too early is now discouraged since that preimage will not
be accepted and may cause the validator to be slashed.

### DIFF
--- a/source/agora/consensus/state/ValidatorSet.d
+++ b/source/agora/consensus/state/ValidatorSet.d
@@ -529,6 +529,10 @@ public class ValidatorSet
         if (prev_preimage.height >= preimage.height)
             return false;
 
+        // Ignore preimages beyond the current cycle
+        if (preimage.height - prev_preimage.height > this.params.ValidatorCycle)
+            return false;
+
         if (auto reason = isInvalidReason(preimage, prev_preimage))
         {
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM

Geod24:
Wasn't https://github.com/bosagora/agora/pull/2469 a bit more complete ? Not only did it make the value configurable, but it also let the door open for the node to add far pre-images by itself ?

Geod24:
Also the problem mentioned here: https://github.com/bosagora/agora/pull/2469#issuecomment-914311529 still applies.

omerfirmak:
> Also the problem mentioned here: [#2469 (comment)](https://github.com/bosagora/agora/pull/2469#issuecomment-914311529) still applies.

But we now include preimages in the header, so catchup should work just fine. Am I right?

Geod24:
That's right, indeed.

PR #2674, opened 2021-11-16
TransactionPool: implement getFrom()

### LINKED ISSUE #1481: We need endpoints that provide pending transactions
We were able to provide pending transactions in Stoa using `pushTransaction`, which was addressed in Issue https://github.com/bpfkorea/agora/issues/1403. 
It was also used to filter the UTXO used in pending transactions when UTXO was provided.

However, transactions received when Stoa is stopped do not exist on Stoa even if Stoa is restarted.
Therefore, Stoa requires the ability to automatically synchronize to solve this problem.

To do this, we need to add API endpoints that provide pending transactions to Agora.

And Agora does not have the time to receive transactions. 
It also provides the time to receive the transactions needed in Stoa to help achieve full synchronization.

omerfirmak:
Fixes #1481

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -347,6 +347,18 @@ public interface API
 
     public Transaction[] getTransactions (Set!Hash tx_hashes);
 
+     /***************************************************************************
+
+         Params:
+             from = starting hash
+
+         Returns:
+             Transactions in the pool that have a larger hash value
+
+     ***************************************************************************/
+
+    public Transaction[] getTransactions (Hash from);
+
     /***************************************************************************
 
         Get an enrollment data if the data exists in the enrollment pool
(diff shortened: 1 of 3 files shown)


mkykadir (approved):

mkykadir (approved):
LGTM but there is a comment in the issue,

> This also will need pagination.

omerfirmak:
> 
> 
> LGTM but there is a comment in the issue,
> 
> > This also will need pagination.

https://github.com/bosagora/agora/pull/2674/commits/faff9c82a60ec3e4d616c58cf73080c49e4f89fd#diff-495f1a28109d946e08a66735141da184d0f65916b585f729363b723b94fdc8cbR487

`count` parameter handles that. a client can fetch the whole pool with multiple calls similar to;

https://github.com/bosagora/agora/pull/2674/commits/faff9c82a60ec3e4d616c58cf73080c49e4f89fd#diff-495f1a28109d946e08a66735141da184d0f65916b585f729363b723b94fdc8cbR579-R589

omerfirmak:
CC @MichaelKim20

PR #2675, opened 2021-11-16
DNS: Fix Header flag bitwise operations

mkykadir:
56204502f1be26adea056f2ca8ec526dcc793368 was first try to fix header flags but it couldn't clear a flag after it was set. fe30ed6d46065296e0586869fd9c7309db2cdcde implements a better way of handling bitwise operations.

This commit generalizes those changes for other bits in the field.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -370,12 +370,30 @@ public struct Header
     /// to match up replies to outstanding queries.
     public ushort ID;
 
+    /***************************************************************************
+        
+        Function prototype for setting or getting specified bit(s) of field
+
+        Params:
+            position = LSB position for reqested bit(s)
+            T = Type to cast when getting bit(s), default is `bool`
+            mask = Bitmask to select bit(s) to be effected starting from `position`, default is 1-bit
+
+    ***************************************************************************/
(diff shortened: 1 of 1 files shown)


source/agora/common/DNS.d
```diff
@@ -370,12 +371,30 @@ public struct Header
...
+        
+        Function prototype for setting or getting specified bit(s) of field
+
+        Params:
+            T = Type to cast when getting bit(s)
+            position = LSB position for reqested bit(s)
+            mask = Bitmask to select bit(s) to be effected starting from `position`
+
+    ***************************************************************************/
+    private mixin template BitGetterSetter (T, ubyte position, ushort mask = 0x1)
```
omerfirmak: if `mask` has default value of 1, you could also make `T` a `bool` by default

omerfirmak (approved):
Merge at will

source/agora/common/DNS.d
```diff
@@ -347,6 +347,7 @@ public struct Message
 
 /// https://datatracker.ietf.org/doc/html/rfc1035#section-4.1.1
 public struct Header
+
```
omerfirmak: Oops

omerfirmak (changes requested):

omerfirmak (approved):

PR #2678, opened 2021-11-17
Fix errors of fee calculation on flash transactions

### LINKED ISSUE #2494: Remove 0 fee special casing in Flash tests
https://github.com/bosagora/agora/blob/b72bd7c2718e33bf962cf90b2453c086dd6fc90c/source/agora/test/Flash.d#L591-L593

This has to go to mimic real-world scenarios.

linked0:
This is a PR in the middle of solving the #2494 which needs some information for the problem that occurs in collecting close transaction signatures. I solved the errors in calculating fees on `funding tx`, `update tx`, and `settle tx`. But in the middle of solving the problem on the `close tx`, I might need to get some clues or prerequisite knowledge from @omerfirmak.

I will open another PR and make the PR fail on the second unit test in source/agora/test/Flash.d intentionally for some discussion. Before that, can you review the code for solving the problem of fee calculation on `funding tx`, `update tx`, and `settle tx`?
 
Fixes #2494

### DIFF
--- a/source/agora/flash/Channel.d
+++ b/source/agora/flash/Channel.d
@@ -2233,12 +2233,19 @@ LOuter: while (1)
         auto update_input = update_tx.inputs[0];
         auto update_ouput = update_tx.outputs[0];
 
-        auto utxos = this.getFeeUTXOs(update_tx.sizeInBytes());
+        // calculate fee
+        auto max_size = update_tx.sizeInBytes();
+        auto mock_refund_output = Output(Amount.init, this.own_pk);
+        max_size += mock_refund_output.sizeInBytes();
+        auto mock_update_unlock = this.update_signer.makeUpdateUnlock(SigPair.init, update.seq_id);
+        max_size += mock_update_unlock.sizeInBytes();
+        auto utxos = this.getFeeUTXOs(max_size);
+
         update_tx.inputs ~= utxos.utxos.map!(hash => Input(hash)).array;
         update_tx.inputs.sort();
 
         // check for refund amount
-        if (utxos.total_value > 0.coins)
+        if (utxos.total_value.sub(utxos.total_fee) && utxos.total_value > 0.coins)
             update_tx.outputs ~= [Output(utxos.total_value, this.own_pk)];
         update_tx.outputs.sort();
 
--- a/source/agora/flash/Node.d
+++ b/source/agora/flash/Node.d
@@ -1418,9 +1418,7 @@ public class FlashNode : FlashControlAPI
             return FeeUTXOs.init;
         // Always pay with the node key
         auto utxos = this.listener.getFeeUTXOs(this.conf.key_pair.address, per_byte);
-        // reuse total_value as refund amount
-        if (!utxos.total_value.sub(per_byte))
-            utxos.total_value = Amount(0);
+
         return utxos;
     }
 
--- a/source/agora/flash/Scripts.d
+++ b/source/agora/flash/Scripts.d
@@ -177,7 +177,8 @@ public Unlock createUnlockSettle (Signature sig, in ulong seq_id,
 public Transaction createFundingTx (in UTXO utxo, in Hash utxo_hash,
     in Amount capacity, in Point pair_pk, Amount fee) @safe nothrow
 {
-    auto inputs = [Input(utxo_hash)];
+    auto fee_rate = fee;
+    auto inputs = [Input(utxo_hash, Unlock(SigPair.init[]))];
     auto outputs = [Output(capacity,
         Lock(LockType.Key, pair_pk[].dup))];
 
@@ -193,7 +194,16 @@ public Transaction createFundingTx (in UTXO utxo, in Hash utxo_hash,
     if (rem_funds > 0.coins)
         outputs ~= Output(rem_funds, utxo.output.lock);
 
-    return Transaction(inputs, outputs);
+    auto funding_tx = Transaction(inputs, outputs);
+
+    version (unittest)
+    {
+        // check the calculated fee for the tx size
+        if (!fee_rate.mul(funding_tx.sizeInBytes()) || fee_rate < fee)
+            assert(0);
+    }
+
+    return funding_tx;
 }
 
 /*******************************************************************************
--- a/source/agora/test/Flash.d
+++ b/source/agora/test/Flash.d
@@ -40,6 +40,7 @@ import agora.flash.Types;
 import agora.script.Engine;
 import agora.script.Lock;
 import agora.script.Script;
+import agora.script.Signature;
 import agora.serialization.Serializer;
 import agora.test.Base;
 import agora.utils.Log;
@@ -560,6 +561,7 @@ private class FlashListener : TestFlashListenerAPI
     public FeeUTXOs getFeeUTXOs (PublicKey pk, Amount amount)
     {
         auto last_height = agora_node.getBlockHeight();
+        auto per_byte = this.getEstimatedTxFee();
 
         FeeUTXOs utxos;
         do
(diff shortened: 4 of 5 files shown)


source/agora/flash/Scripts.d
```diff
@@ -177,7 +177,7 @@ public Unlock createUnlockSettle (Signature sig, in ulong seq_id,
 public Transaction createFundingTx (in UTXO utxo, in Hash utxo_hash,
     in Amount capacity, in Point pair_pk, Amount fee) @safe nothrow
 {
-    auto inputs = [Input(utxo_hash)];
+    auto inputs = [Input(utxo_hash, Unlock(SigPair.init[]))];
```
omerfirmak, line 181: Can you add a small test to verify that:

`createFundingTx(..., fee). getFeeRate()  == fee`

So that we don't break this in the future

source/agora/flash/Channel.d
```diff
@@ -2233,7 +2233,14 @@ LOuter: while (1)
...
         auto update_ouput = update_tx.outputs[0];
 
-        auto utxos = this.getFeeUTXOs(update_tx.sizeInBytes());
+        // calculate fee
+        auto max_size = update_tx.sizeInBytes();
+        auto mock_refund_output = Output(Amount.init, this.own_pk);
+        max_size += mock_refund_output.sizeInBytes();
+        auto mock_update_unlock = this.update_signer.makeUpdateUnlock(SigPair.init, update.seq_id);
+        max_size += mock_update_unlock.sizeInBytes();
+        auto utxos = this.getFeeUTXOs(max_size);
```
omerfirmak, line 2242: This is a chicken&egg problem. We can't know how many UTXOs `getFeeUTXOs` will return. So any calculations we do here is an approximation of the TX size. We have to check if our TX size is the same as the expected size after adding fees, or delegate that work to the getFeeUTXOs itself and assume that returned UTXOs are enough to cover the additional inputs and outputs.

source/agora/flash/Scripts.d
linked0: Will add a test for that.

source/agora/flash/Channel.d
linked0: Right, I missed a calculation of the fee for outputs. In fact, the code in `getFeeUTXOs` calculates the fee for inputs. I will add it and check that the returned UTXOs are enough. Thanks.

linked0: After looking into the code again, the output for the refund of the fee was also calculated as follows.
`
auto mock_refund_output = Output(Amount.init, this.own_pk);
`
So only thing to be needed in this part is to add checking that the returned UTXOs are enough.

source/agora/flash/Scripts.d
linked0: Updated.

source/agora/flash/Channel.d
linked0: Updated.

source/agora/flash/Scripts.d
```diff
@@ -193,7 +194,12 @@ public Transaction createFundingTx (in UTXO utxo, in Hash utxo_hash,
     if (rem_funds > 0.coins)
         outputs ~= Output(rem_funds, utxo.output.lock);
 
-    return Transaction(inputs, outputs);
+    // check the calculated fee for the tx size
+    auto funding_tx = Transaction(inputs, outputs);
+    if (!fee_rate.mul(funding_tx.sizeInBytes()) || fee_rate != fee)
```
omerfirmak: Oops, I overlooked something when I left the previous comment. As `max_size` name suggests, we actually might overpay here, if there is no refund output. So the check should be `fee_rate >= fee`.

and let's put this in a `version (unittest)` so that we don't crash for this when running in production.

source/agora/flash/Node.d
```diff
@@ -1418,8 +1419,15 @@ public class FlashNode : FlashControlAPI
             return FeeUTXOs.init;
         // Always pay with the node key
         auto utxos = this.listener.getFeeUTXOs(this.conf.key_pair.address, per_byte);
+
+        // calculate a expected fee for checking the fee is correct
+        per_byte = this.listener.getEstimatedTxFee();
```
omerfirmak: Calling `getEstimatedTxFee` multiple times might result in different values to be returned, so it should only be called once throughout the `getFeeUTXOs`

source/agora/flash/Node.d
```diff
@@ -1418,8 +1419,15 @@ public class FlashNode : FlashControlAPI
...
+        utxos.utxos.each!(hash =>
+            tx_size += Input(hash, genKeyUnlock(SigPair.init)).sizeInBytes());
+        per_byte.mul(tx_size);
+
         // reuse total_value as refund amount
-        if (!utxos.total_value.sub(per_byte))
+        if (!utxos.total_value.sub(utxos.total_fee) || per_byte != utxos.total_fee)
```
omerfirmak: If we are gonna add a new `total_fee` field, don't subtract that from the `total_value`. So that the caller knows how much the UTXO set it received is worth and how much fees it should pay with them. Caller should do the `total_value - total_fee` to properly create the refund output.

source/agora/test/Flash.d
```diff
@@ -569,12 +570,17 @@ private class FlashListener : TestFlashListenerAPI
             foreach (idx, output; tx.outputs)
                 if (output.address() == pk)
                 {
-                    utxos.utxos ~= UTXO.getHash(tx.hashFull(), idx);
+                    auto utxo = UTXO.getHash(tx.hashFull(), idx);
+                    utxos.utxos ~= utxo;
                     utxos.total_value += output.value;
+                    auto per_byte = this.getEstimatedTxFee();
```
omerfirmak: Same comment about the `getEstimatedTxFee`

source/agora/flash/Channel.d
```diff
@@ -2233,7 +2233,14 @@ LOuter: while (1)
         auto update_input = update_tx.inputs[0];
```
omerfirmak, line 2233: Typo in commit message "undate"

source/agora/flash/Channel.d
```diff
@@ -777,6 +777,38 @@ LOuter: while (1)
             auto settle_tx = this.channel_updates[$ - 1].settle_tx;
             // point the input to the last update utxo
             settle_tx.inputs[0].utxo = this.last_externalized_update_utxo;
+            auto settle_input = settle_tx.inputs[0];
```
omerfirmak: Settle txs can't have fees added to them this way. They are not the same with the update TXs. Let's drop this commit from this PR and discuss it on another one after merging the others.

source/agora/test/Flash.d
```diff
@@ -585,21 +590,28 @@ private class FlashListener : TestFlashListenerAPI
...
+private TestConf flashTestConf (bool noFee = true)
 {
     import agora.node.Config;
 
     TestConf conf;
     conf.consensus.quorum_threshold = 100;
-    // TODO: remove this line when fees are handled
-    conf.consensus.min_fee = Amount(0);
-    conf.node.min_fee_pct = 0;
+    if (noFee)
```
omerfirmak: There is no reason to have seperate tests for TXs with/without fees. It also leads to duplicated tests. I would say drop this commit from this PR too.

omerfirmak (changes requested):

source/agora/test/Flash.d
linked0: Right, this `if` part is temporary. The goal for issue #2494  is to remove all this `if` part as the comment, `// TODO: remove this line when fees are handled` suggested. My current blocking problem about the issue is to fix the error [the test PR](https://github.com/bosagora/agora/pull/2679) makes. And, we need to fix the problem about `settle tx` also in other way [as you commented](https://github.com/bosagora/agora/pull/2678#discussion_r752095583).
Thank you for the review.

linked0:
I'll refine this PR to drop some commits at this time and will request review again. Let talk about the remaining problems in the issue again. @omerfirmak Thanks :)

source/agora/flash/Scripts.d
linked0: Updated.

source/agora/flash/Channel.d
linked0: Fixed

source/agora/flash/Channel.d
linked0: Will discuss it next time.

source/agora/flash/Node.d
linked0: I agree. So I moved the code to the caller and removed the check code for the expected fee because it's not so useful.

linked0:
Ready for review.
After this is merged, we will discuss how to deal with the fee calculation of `settle tx` and `close tx`.

omerfirmak (approved):

omerfirmak:
a great read on the topic @linked0 
https://yakshaver.org/2021/07/26/first.html

PR #2687, opened 2021-11-19
DNS: Refactor zone gathering from AA for safe access

mkykadir:


source/agora/node/Registry.d
```diff
@@ -115,8 +115,8 @@ public class NameRegistry: NameRegistryAPI
     private void collectStats (Collector collector)
     {
         RegistryStats stats;
-        stats.registry_validator_record_count = this.zones[this.validators].count();
-        stats.registry_flash_record_count = this.zones[this.flash].count();
+        stats.registry_validator_record_count = (this.validators in this.zones).count();
```
omerfirmak: I think changes like these are not needed. There won't be any copying unless you do;

`ValueType value = AA[Key];`

Feel free to revert such changes or merge.

omerfirmak (approved):

source/agora/node/Registry.d
```diff
@@ -459,7 +460,7 @@ public class NameRegistry: NameRegistryAPI
     public void onAcceptedBlock (in Block, bool)
         @safe
     {
-        auto validators_zone = this.zones[this.validators];
+        auto validators_zone = this.validators in this.zones;
```
Geod24, line 463: This is not safer BTW, because D dereferences the pointer automatically. The safe way would include a null check.

source/agora/node/Registry.d
```diff
@@ -430,8 +430,9 @@ public class NameRegistry: NameRegistryAPI
     
     auto findZone (Domain name, bool matches = true) @safe
     {
-        if (name in this.zones)
-            return matches ? &this.zones[name].answer_matches : &this.zones[name].answer_owns;
+        auto matching_zone = name in this.zones;
+        if (matching_zone)
+            return matches ? &matching_zone.answer_matches : &matching_zone.answer_owns;
```
Geod24, line 435: When using `in`, a very nice idiom is:
```D
if (auto matching_zone = name in this.zones)
    // Use the variable, which is only accessible within this scope.
```

source/agora/node/Registry.d
omerfirmak: My reasoning was that they are initialized in ctor which would make the null check superfluous. It might have been better to assert(0), if they turn out to be null tho.

Geod24: I'm not sure I follow ? `validators_zone` is now a `ZoneData*`. If we know the zone is always present, why the switch from `id[key]` to `key in id` ?

omerfirmak: To avoid copying it, So that we don't accidentally modify the copy and leave the ZoneData in `this.zones` untouched

PR #2688, opened 2021-11-19
Validator: Reveal preimages on new block events

omerfirmak:
On enrollment blocks, network is missing preimage of the newly
enrolled validator for the next block. In our production environment,
our block time is short enough that sometimes reveal timer is not triggered
before creation of the next block, resulting in the newly enrolled validator
to be slashed immediately.

onPreImageRevealTimer seems like re-entrant so this should not be a problem
that we call it from outside of the timer.


hewison-chris (approved):
LGTM
Trying a re-run of ci unit tests as single failure is likely unrelated.

linked0 (commented):
LGTM

PR #2689, opened 2021-11-22
RPC: Proper timeout handling while waiting for responses

### LINKED ISSUE #2653: PROD env often gets stuck on a block height with no errors logged
Since we have fixed some of the issues that we detected were blocking the progress of the blockchain we no longer see in the logs why the next block is sometimes stuck in nomination. We should add some other logging info or debug the situation more to find out the cause.

omerfirmak:


source/agora/network/Client.d
```diff
@@ -589,34 +588,40 @@ public class NetworkClient
...
     }
+
+    ///
+    public bool connected () pure nothrow @safe @nogc
+    {
+        return this.connections.length > 0;
+    }
+
+    ///
+    @property public auto address () const pure nothrow @safe @nogc
```
hewison-chris: Maybe `addresses` is better name?

hewison-chris (approved):
LGTM

source/agora/network/Client.d
omerfirmak: Done in a separate commit.

omerfirmak:
This might have a positive affect on our production stability.

source/agora/network/Client.d
```diff
@@ -531,20 +523,20 @@ public class NetworkClient
...
-        LogLevel log_level = LogLevel.Trace, API, Args...)
-        (API[] apis, auto ref Args args, string file = __FILE__, uint line = __LINE__)
+        LogLevel log_level = LogLevel.Trace, Args...)
+        (auto ref Args args, string file = __FILE__, uint line = __LINE__)
     {
         import std.traits;
         enum name = __traits(identifier, endpoint);
         alias T = ReturnType!(__traits(getMember, API, name));
 
+        foreach (conn; this.connections.dup)
```
Geod24: Why the `dup` each time we do a request ?

source/agora/network/Client.d
```diff
@@ -559,22 +551,29 @@ public class NetworkClient
...
                 {
                     // nothing we can do
                 }
 
+                if (conn.address == Address.init)
+                {
+                    try this.connections = this.connections.remove!(c => c.api == conn.api);
+                    catch (Exception e) log.trace("Failed to remove dead connection");
+                    break;
+                }
```
Geod24: I guess that's why the `dup`. But still, we can't `dup` every time just because there is a chance a connection gets removed. You can just use a `foreach` / `while` loop instead, or swap the inner loops (iterating on connections inside of retry instead of the other way around).

source/agora/network/Client.d
```diff
@@ -589,34 +588,40 @@ public class NetworkClient
...
         existing_rpc.merge(incoming_peer);
-        if (this.address[rpc_idx] == Address.init)
-            this.address[rpc_idx] = incoming.address[0];
+        if (this.connections[rpc_idx].address == Address.init)
+            this.connections[rpc_idx].address = incoming.connections[0].address;
         return true;
     }
+
+    ///
+    public bool connected () pure nothrow @safe @nogc
```
Geod24: `const` ?

source/agora/network/RPC.d
```diff
@@ -131,7 +131,7 @@ public class RPCClient (API) : API
         public Duration retry_delay  = 1.seconds;
 
         ///
-        public uint concurrency = 3;
+        public uint concurrency = 1;
```
Geod24: Why this change ?

Geod24 (changes requested):
Two main things

source/agora/network/Client.d
omerfirmak: It is not only this loop that may modify the connection list tho, another fiber could do that too.

source/agora/network/RPC.d
omerfirmak: We use `Config.init` for incoming connections, for which we can't enable concurrency since we don't know the server address. If we do, client tries to initiate a new connection to an unindentified server and fails as expected.

https://github.com/bosagora/agora/commit/c5e36c49dea4a569f398974e9974d015e7116b31#diff-f5162195974e34bb5a3926eac58610027532467ff12d53569827d5867072b62cR199

omerfirmak: We already set it to 3 for outgoing RPC connections.

https://github.com/bosagora/agora/commit/aa4ceb271bb0214b30e026975fcf70c51075d464#diff-7b4de96b302d2ab907d0084dd3697c8dbd7ea2359812b46ec950c0e3c94a17c7R1126

source/agora/network/Client.d
omerfirmak: This might be solvable with DList and slicing.

omerfirmak: @Geod24 , forgot to ping you on this. How do you like the last commit?

omerfirmak: Ok, I'll take a completely different approach on this. Incoming connections will only be used to discover new server address, and not persist in the system.

omerfirmak: Dropped related commit, will have it in its own PR

Geod24 (approved):
Neat

PR #2690, opened 2021-11-22
DNS: Refactor registry to manage TCP/UDP bindings itself

mkykadir:
Previously TCP/UDP port bindings were managed through `Runner`, Registry
will be updated to support secondary servers and resolvers which require
a client implementation. Client can send UDP packets through same port.

- UDP and TCP handlers moved from `Runner` to `Registry`
- Registry shutdown sequence is implemented to properly drop ports
- Registry will be started when node is started

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -33,6 +33,9 @@ import agora.stats.Registry;
 import agora.stats.Utils;
 import agora.utils.Log;
 
+import vibe.core.core;
+import vibe.core.net;
+
 import std.algorithm;
 import std.array : array, replace;
 import std.conv;
@@ -40,6 +43,7 @@ import std.datetime;
 import std.format;
 import std.range;
 import std.socket : InternetAddress;
+import std.stdio;
 import std.string;
 
 static import std.uni;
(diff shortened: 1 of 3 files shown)


source/agora/node/Registry.d
```diff
@@ -33,13 +33,17 @@ import agora.stats.Registry;
 import agora.stats.Utils;
 import agora.utils.Log;
 
+import vibe.core.core;
+import vibe.core.net;
```
Geod24, line 37: That's the big issue. The reason those were separated was to allow for dependency injection.
Having a straight `{TCP,UDP}Connection` in the registry means that we loose the nice separation we had between the business code and the IO code.

mkykadir: I get it. We will need DNS client to ask queries to an another server within a DNS server. For example, AXFR is one of them, which also requires TCP. I thought that whole name registry in a single packet would be easier to manage.
Apart from TCP/UDP, we have REST (HTTP) interface implemented within registry itself.

Geod24: Yes, we need the client to be abstracted as well, but I don't understand why the refactoring here, as it's server side ?

mkykadir: I was going to implement client into Registry also 😃 Closing PR now

PR #2691, opened 2021-11-22
Registry: Implement SOA query support

### LINKED ISSUE #2448: Registry: Return SOA records for authoritative zones
Otherwise DNS recursion doesn't work.

mkykadir:
Fixes #2448 

Adds support for SOA queries. Zone serial is updated when data is **really** updated or a record is removed from database.

Note:
Currently SOA query is not checking for Zone configuration if its authoritative, because current configuration parameters are not enough (secondaries are also authoritative, only resolvers are not authoritative)

### DIFF
--- a/source/agora/api/Registry.d
+++ b/source/agora/api/Registry.d
@@ -19,6 +19,9 @@ import agora.crypto.Key;
 import agora.crypto.Schnorr: Signature;
 import agora.flash.api.FlashAPI;
 
+import std.algorithm.comparison : isPermutation, cmp;
+import std.range : zip;
+
 ///
 public struct RegistryPayloadData
 {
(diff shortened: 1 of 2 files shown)


source/agora/api/Registry.d
```diff
@@ -30,6 +32,14 @@ public struct RegistryPayloadData
...
     /// monotonically increasing sequence number
     public ulong seq;
+
+    /// Compares payload data, ignores `seq` and permutations of `addresses` 
+    ///   are consiederd equal
+    bool opEquals (in RegistryPayloadData other) const @safe
+    {
+        return (this.public_key == other.public_key)
+            && (this.addresses.isPermutation(other.addresses));
+    }
```
Geod24: If we implement `opEquals`, we generally implement `opCmp` as well.

source/agora/api/Registry.d
```diff
@@ -30,6 +32,14 @@ public struct RegistryPayloadData
 
     /// monotonically increasing sequence number
     public ulong seq;
+
+    /// Compares payload data, ignores `seq` and permutations of `addresses` 
+    ///   are consiederd equal
```
Geod24: Why the extra indent ? Also typo.

source/agora/node/Registry.d
```diff
@@ -928,6 +939,16 @@ private struct ZoneData
         if (payload.payload.data.addresses.length == 0)
             return;
 
+        if (this.get(payload.payload.data.public_key) == payload)
+        { // Payload is equal to Zone's data, no need to update
```
Geod24: Newline

Geod24 (approved):
This does three things that could be broken down in their own commits:
- Add an `opEqual` to a `struct`;
- Reply to `SOA` request;
- Avoid overriding payload data when not needed;

Concept LGTM.

source/agora/api/Registry.d
```diff
@@ -30,6 +33,48 @@ public struct RegistryPayloadData
...
+
+            if (this.seq != other.seq)
+                return this.seq < other.seq ? -1 : 1;
+
+            if (this.addresses.length != other.addresses.length)
+                return this.addresses.length < other.addresses.length ? -1 : 1;
+            
+            foreach (thisAddr, otherAddr; this.addresses.zip(other.addresses))
+            {
+                auto comparison = cmp(thisAddr.toString, otherAddr.toString);
```
Geod24: Let's avoid `toString` as it allocates

source/agora/api/Registry.d
```diff
@@ -30,6 +33,48 @@ public struct RegistryPayloadData
...
+                // will eventually find non-equal addresses since opEquals failed
+                if (comparison == 0) continue;
+                return comparison;
```
Geod24:
```D
if (auto c = cmp(...))
    return c;
```

source/agora/node/Registry.d
```diff
@@ -833,6 +842,7 @@ private struct ZoneData
             return Header.RCode.NameError;
 
         reply.answers ~= payload.toRR(question.qname);
+        reply.authorities ~= this.toRR(); // optional
```
Geod24, line 845: I think most servers don't include this, but reading the RFC, this looks correct.

Geod24 (approved):
Just the allocation.

Geod24 (approved):
Merge at will

PR #2693, opened 2021-11-23
Minor fixes extracted from #2469

Geod24:
https://github.com/bosagora/agora/pull/2469 has been superseded, but included two small commits which are now submitted here.


linked0 (approved):
LGTM

PR #2695, opened 2021-11-23
DNS: Add a factory function for `ResourceRecord`

Geod24:
```
`ResourceRecord`s are the core of the DNS system, and are essentially
an `enum` type when framed from a D point of view.
However, they are currently defined following the wire format,
which is less than an ideal API for the user.
As a step towards having a better API for them, a factory function needs
to be introduced, hiding the underlying details / boilerplate.
```

Draft as I'm testing.


Geod24:
Works nicely

Geod24:
Rebased on HEAD.

linked0 (approved):
LGTM

PR #2696, opened 2021-11-23
Override `opAssign` of `Domain` to always add root domain

mkykadir:
Commit 1cc5276d4d006555615f7bf8c33ea470ce2375ef implements always adding
root domain (trailing dot) but missed overriding `opAssign`. Thus
assignments missed trailing dot, which causes removing last character from
domain during serialization.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -371,7 +371,7 @@ public struct Header
     public ushort ID;
 
     /***************************************************************************
-        
+
         Function prototype for setting or getting specified bit(s) of field
 
         Params:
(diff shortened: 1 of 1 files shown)


source/agora/common/DNS.d
```diff
@@ -371,7 +371,7 @@ public struct Header
     public ushort ID;
 
     /***************************************************************************
-
```
Geod24, line 374: Separate commit

Geod24 (commented):
Actually I think the `alias this` is the issue. I'll raise a PR to remove it.

linked0 (approved):
LGTM

mkykadir:
This is obsolete due to #2698

PR #2697, opened 2021-11-24
Always instantiate a name registry

Geod24:
As we want to cache the addresses of validators locally,
and we have to implement the logic in the registry for secondary servers,
it makes sense to re-use this logic for clients as well.

With this change, the `enabled` boolean now means "publicly available",
rather than "internally instantiate a NameRegistry object".


mkykadir (commented):
This might be a bit early, since there is no transferring capability yet.

mkykadir (approved):

PR #2698, opened 2021-11-24
Remove `alias this` from DNS' `Domain` and other fixes

Geod24:
The `alias this` proved more problematic than anything else, and led to issues with assignment.
Removing it makes the code easier to follow and remove the need for overriding `opAssign`.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -753,12 +753,12 @@ public struct Domain
     public const(char)[] value = ".";
 
     /// Creates domain name with always trailing root
-    public this (inout const(char)[] v) @safe inout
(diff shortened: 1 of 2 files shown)


mkykadir (approved):
LGTM, direct usage of `Domain` with the help of `alias` looks better than always asking for `value`, I think. But in the end, its a `struct` :smile: 

If this gets merged #2696 becomes obsolete.

Geod24:
@mkykadir : I plan to add `opApply` support so we can iterate on the domain by label, which would also simplify some other code. In the end, if you are holding a `Domain`, it's because you want to use it within the DNS code, and we should make that easy. If the path of least resistance is dropping back to use `string`, we failed.

PR #2699, opened 2021-11-24
Extend registry config to allow specifying `mname` and the realm zone

### LINKED ISSUE #2537: Registry: SOA's `mname` is incorrect
The `mname` should be one of the authoritative NS for the zone, not the zone itself.
E.g. we currently use this:
https://github.com/bosagora/agora/blob/ebae6647674271f435d7c18668b5092b8707262c/source/agora/node/Registry.d#L636-L640

But we should instead lookup the `NS` record for the `realm`.

Geod24:
See each commit.

The first one is required for recursion to work properly, and the second one fixes #2537

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -303,17 +303,37 @@ registry:
   # Port to bind to (default: 53, standard DNS port)
   # Note that ports < 1024 are privileged ports and might require root powers
   port: 53
+
+  ## The fields below define 3 SOA records, one for each zone required by the network.
+  ##
+  ## Those can typically be left empty, except for the primary server of the zone.
+  ## If non-empty, both `primary` and `email` must be provided.
+  ##
+  ## Others fields are ignored if the server is not authoritative for the zone.
+
   # The zone responsible for validators registration
   validators:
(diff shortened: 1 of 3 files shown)


doc/config.example.yaml
```diff
@@ -303,17 +303,37 @@ registry:
...
-    ## Those fields are ignored if the server is not `authoritative` for the zone.
+    # The *hostname* (not IP address) of the primary name server for this zone
+    #
+    # If this is set, `email` must be set too. If this isn't set, `email` must be unset.
+    #
+    # The value here is typically the hostname of the server itself.
+    # For example, if the node that read this configuration will be reachable via
+    # `agora.example.com`, then this string can be used.
+    # However, the typical naming convention would be to use `ns1.example.com`.
+    primary: ns1.bosagora.io
```
mkykadir, line 336: I think we can have enumerated registry types `primary`, `secondary` and `caching` since

- a `primary` can have `trusted` servers for zone transfer
- a `secondary` server can be `primary` for another `secondary`
- we might have multiple `primary`ies configured to do transfers from as a backup
- only `primary` can have a SOA configured

mkykadir (approved):

PR #2636, opened 2021-11-03
Add a client DNS resolver

Geod24:
Still a long way to go before this is ready, but a simple proof of concept which works with some example.
Mainly for the benefit of @mkykadir as we're both working on the name registry.

Doing:
`dub -c dns-resolver -- google.com.`

Will return significant results from `8.8.8.8` and `8.8.4.4`, the others will not (for various reasons).
Going to submit the name compression fixes in a separate PR soon.

PR #2703, opened 2021-11-26
Implement secondary and caching registry

### LINKED ISSUE #2449: Registry: Implement secondary zones
A registry should be able to be non-authoritative, however we need to see how this is best implemented, especially taking https://github.com/bosagora/agora/pull/2443 into account.

### LINKED ISSUE #2647: DNS: Implement whitelisting for AXFR
We should have a way to restrict AXFR to whitelisted clients, so that it doesn't get abused.
A config file entry, probably at the registry level, as it doesn't make much sense to have it at the zone level, would be needed.

```yaml
registry:
  enabled: true
  whitelist:
    - 1.2.3.4
    - foo.bar
```

mkykadir:
This PR implements secondary and caching registry. Fixes of #2449 and Fixes #2647
With secondary registries we can create backup registry servers to improve availability, also this implementation enables us to create a hidden registry.

- It uses `DNSResolver`, depends on #2636 
- Requests AXFR over UDP
- SOA RR is checked periodically, depends on #2730 
- POST endpoints are redirected to actual primary
- Removes two `boolean` flags from configuration and creates three type registries
  - primary registry server, only this will have database modifying capabilities, authoritative
  - secondary registry server, this will do AXFR query to its primaries; secondary servers can be primary for other secondaries
  - caching server, not implemented yet but will be non-authoritative
- Previous Registry cache is removed when Zone type is updated
- Caching zone starts with clean storage and adds to cache according to query results up to TTL seconds of the record.
- Makes Registry enabled by default with caching zones

**DNSResolver**

- [ ] RFC recommends TCP for AXFR, currently using UDP

### DIFF
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -392,3 +392,8 @@ registry:
   # Zone responsible for realm
   realm:
     authoritative: false
+
+    # List of servers to update records cache from
+    # Default DNS resolvers are (i.e. 8.8.8.8 / 1.1.1.1) used when not configured
+    query_servers:
+      - ns1.bosagora.io
--- a/doc/config.example.yaml
+++ b/doc/config.example.yaml
@@ -347,6 +347,7 @@ registry:
     allow_transfer:
       - 127.0.0.1
 
+    # SOA record for the zone, only applicable for primary servers
     soa:
       # Email of the DNS server administrator
       # This field is required for authoritative servers.
@@ -368,6 +369,26 @@ registry:
 
   # Zone responsible for flash nodes registration
   flash:
-    # Having separate authoritative server increases redundancy and reduces load,
+    # Having one or more secondary servers increase redundancy and reduces load,
     # hence this configuration, while unlikely in a small to medium network, is supported.
+    authoritative: true
+
+    # Redirect server for REST endpoints of the registry
+    redirect_register: http://nsHidden.bosagora.io
+
+    # List of servers to AXFR zone transfer from
+    #
+    # Servers are tried in the configuration order until zone transfer can be
+    # completed from a server.
+    query_servers:
+      - ns1.bosagora.io
+
+    # A secondary server can be an upstream of an another server, thus secondary
+    # servers can also limit AXFR zone transfers from them to a configured
+    # whitelist
+    allow_transfer:
+      - 127.0.0.1
+
+  # Zone responsible for realm
+  realm:
     authoritative: false
--- a/source/agora/node/Config.d
+++ b/source/agora/node/Config.d
@@ -462,6 +462,10 @@ public struct RegistryConfig
                    "registry.{}: Authoritative zones require both 'primary' and " ~
                    "'email' fields to be set", name);
 
+            // Caching registry, nothing to validate
+            if (!zone.authoritative.set || zone.authoritative.value == false)
+                return;
+
             // Secondary registry
             if (!zone.primary.set && zone.authoritative.set)
             {
@@ -548,8 +552,8 @@ public struct ZoneConfig
     /// Servers that are allowed to do AXFR queries, applicable for authoritatives
     public @Optional immutable IPAddress[] allow_transfer;
 
-    /// Servers to zone transfer (AXFR) from, only applicaple and
-    /// required for secondary servers
+    /// Servers to zone transfer (AXFR) from, required for secondary servers
+    /// or servers to update record cache from for caching servers
     public @Optional immutable string[] query_servers;
 
     /// Rest interface of an Agora node with name registering capability,
--- a/source/agora/node/Config.d
+++ b/source/agora/node/Config.d
@@ -462,9 +462,20 @@ public struct RegistryConfig
                    "registry.{}: Authoritative zones require both 'primary' and " ~
                    "'email' fields to be set", name);
 
-            // If we're not primary, there's no configuration to validate
-            if (!zone.primary.set)
+            // Secondary registry
+            if (!zone.primary.set && zone.authoritative.set)
+            {
+                ensure(zone.query_servers.length > 0,
+                    "registry.{} is secondary and at least one primary is required",
+                    name);
+
+                ensure(zone.redirect_register.length > 0,
+                    "registry.{} is secondary (authoritative), thus requires an " ~
+                    "address registering endpoint to an Agora node",
+                    name);
+
                 return;
+            }
 
             ensure(!zone.authoritative.set || zone.authoritative.value == true,
                    "registry.{}.authoritative: Cannot be false for primary server", name);
@@ -508,7 +519,7 @@ public struct RegistryConfig
 public struct ZoneConfig
 {
     /// IP address configuration type for `allow_transfer`
-    private struct IPAddress
+    public struct IPAddress
     {
         /// Address
         public string ip;
@@ -536,6 +547,14 @@ public struct ZoneConfig
 
     /// Servers that are allowed to do AXFR queries, applicable for authoritatives
     public @Optional immutable IPAddress[] allow_transfer;
+
+    /// Servers to zone transfer (AXFR) from, only applicaple and
+    /// required for secondary servers
+    public @Optional immutable string[] query_servers;
+
+    /// Rest interface of an Agora node with name registering capability,
+    /// probably one of the `query_servers`; required for secondary servers
+    public @Optional immutable string redirect_register;
 
     /// SOA Configuration of the zone, required for master server
     /// All `Duration` values are precise to the second
@@ -577,7 +596,7 @@ public struct ZoneConfig
     }
 
     /// Ditto
-    public SOAConfig soa;
+    public @Optional SOAConfig soa;
 }
 
 //
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -18,6 +18,7 @@ import agora.common.Amount;
 import agora.common.DNS;
 import agora.common.Ensure;
 import agora.common.ManagedDatabase;
+import agora.common.Task;
 import agora.common.Types;
 import agora.consensus.data.Block;
 import agora.consensus.data.UTXO;
@@ -26,6 +27,8 @@ import agora.consensus.Ledger;
 import agora.crypto.Hash;
 import agora.crypto.Key;
 import agora.crypto.Schnorr: Signature;
+import agora.network.DNSResolver;
+import agora.network.Manager;
 import agora.node.Config;
 import agora.flash.api.FlashAPI;
 import agora.serialization.Serializer;
@@ -42,9 +45,13 @@ import std.range;
 import std.socket : InternetAddress;
 import std.string;
 
+import core.time;
+
 static import std.uni;
 
 import d2sqlite3 : ResultRange;
+import vibe.http.client;
+import vibe.web.rest;
 
 /// Implementation of `NameRegistryAPI` using associative arrays
 public class NameRegistry: NameRegistryAPI
@@ -91,7 +98,7 @@ public class NameRegistry: NameRegistryAPI
 
     ///
     public this (Domain realm, RegistryConfig config, NodeLedger ledger,
-        ManagedDatabase cache_db)
+        ManagedDatabase cache_db, ITaskManager taskman, NetworkManager network)
     {
         assert(ledger !is null);
         assert(cache_db !is null);
@@ -107,11 +114,11 @@ public class NameRegistry: NameRegistryAPI
 
         this.zones = [
             ZoneData("realm", this.realm,
-                this.config.realm, cache_db, log),
+                this.config.realm, cache_db, log, taskman, network),
             ZoneData("validator", this.validators,
-                this.config.validators, cache_db, log),
+                this.config.validators, cache_db, log, taskman, network),
             ZoneData("flash",  this.flash,
-                this.config.flash, cache_db, log)
+                this.config.flash, cache_db, log, taskman, network)
         ];
 
         Utils.getCollectorRegistry().addCollector(&this.collectStats);
@@ -225,11 +232,14 @@ public class NameRegistry: NameRegistryAPI
 
     public override void postValidator (RegistryPayload registry_payload)
     {
-        ensure(this.zones[ZoneIndex.Validator].type == ZoneType.primary,
-            "Couldn't register, server is not primary for the zone");
-
         TYPE payload_type = this.ensureValidPayload(registry_payload,
             this.zones[ZoneIndex.Validator].get(registry_payload.data.public_key));
+
+        if (this.zones[ZoneIndex.Validator].type == ZoneType.secondary)
+        {
+            this.zones[ZoneIndex.Validator].redirect_register.postValidator(registry_payload);
+            return;
+        }
 
         // Last step is to check the state of the chain
         auto last_height = this.ledger.height() + 1;
@@ -306,11 +316,14 @@ public class NameRegistry: NameRegistryAPI
 
     public override void postFlashNode (RegistryPayload registry_payload, KnownChannel channel)
     {
-        ensure(this.zones[ZoneIndex.Flash].type == ZoneType.primary,
-           "Couldn't register, server is not primary for the zone");
-
         TYPE payload_type = this.ensureValidPayload(registry_payload,
             this.zones[ZoneIndex.Flash].get(registry_payload.data.public_key));
+
+        if (this.zones[ZoneIndex.Flash].type == ZoneType.secondary)
+        {
+            this.zones[ZoneIndex.Flash].redirect_register.postFlashNode(registry_payload, channel);
+            return;
+        }
 
         auto range = this.ledger.getBlocksFrom(channel.height);
         ensure(!range.empty, "Channel is at height {} but local Ledger is at height {}",
@@ -488,15 +501,30 @@ public class NameRegistry: NameRegistryAPI
 
     ***************************************************************************/
 
-    public void onAcceptedBlock (in Block, bool)
+    public void onAcceptedBlock (in Block, bool validators_changed)
         @safe
     {
+        ZoneType validator_type = this.zones[ZoneIndex.Validator].type;
+
         if (this.zones[ZoneIndex.Validator].type == ZoneType.primary)
         {
             this.zones[ZoneIndex.Validator].each!((TypedPayload tpayload) {
                 if (this.ledger.getPenaltyDeposit(tpayload.utxo) == 0.coins)
                     this.zones[ZoneIndex.Validator].remove(tpayload.payload.data.public_key);
             });
+        }
+        else if (validator_type == ZoneType.secondary
+                    && validators_changed)
+        {
+            // Even this manipulates the SOA timings, we can think it as a
+            // NOTIFY request of the DNS, new node is found and zone needs update
+            () @trusted {
+                if (this.zones[ZoneIndex.Validator].soa_update_timer.pending)
+                {
+                    this.zones[ZoneIndex.Validator].soa_update_timer.stop;
+                    this.zones[ZoneIndex.Validator].updateSOA();
+                }
+            } ();
         }
     }
 }
@@ -712,7 +740,7 @@ private enum ZoneType
     /// When both `authoritative` and `SOA` configurations are set
     primary = 1,
 
-    ///
+    /// `authoritative` is set but `SOA` configuration is not
     secondary = 2,
 }
 
@@ -758,8 +786,26 @@ private struct ZoneData
     /// Query for getting all registered network addresses
     private string query_addresses_get;
 
+    /// Query for clean up zone before AXFR, only for secondary zone
+    private string query_axfr_cleanup;
+
     /// Database to store data
     private ManagedDatabase db;
+
+    /// DNS resolver to send request, only for secondary
+    private DNSResolver resolver;
+
+    /// Timer for requesting SOA from a primary to check serial, only for secondary
+    private ITimer soa_update_timer;
+
+    /// Timer for disabling zone when SOA check cannot be completed, only for secondary
+    private ITimer expire_timer;
+
+    /// Task manager to manage timers
+    private ITaskManager taskman;
+
+    /// REST API of a primary registry to redirect API calls, only for secondary
+    private NameRegistryAPI redirect_register;
 
     /***************************************************************************
 
@@ -770,17 +816,22 @@ private struct ZoneData
     ***************************************************************************/
 
     public this (string zone_name, Domain root, ZoneConfig config,
-        ManagedDatabase cache_db, Logger logger)
+        ManagedDatabase cache_db, Logger logger, ITaskManager taskman,
+        NetworkManager network)
     {
         this.db = cache_db;
         this.log = logger;
         this.config = config;
         this.root = root;
+        this.taskman = taskman;
+        this.soa = SOA.init;
 
         if (this.config.authoritative)
         {
             if (this.config.soa.email.set)
                 this.type = ZoneType.primary;
+            else
+                this.type = ZoneType.secondary;
         }
 
         static string serverType (ZoneType zone_type)
@@ -788,7 +839,8 @@ private struct ZoneData
             switch (zone_type)
             {
                 case ZoneType.primary: return "primary (authoritative)";
-                default: return "secondary";
+                case ZoneType.secondary: return "secondary (authoritative)";
+                default: return "caching";
             }
         }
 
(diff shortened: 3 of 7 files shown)


mkykadir:
@Geod24 this is ready for review (even it depends on WIP Resolver :) )

source/agora/common/Types.d
```diff
@@ -34,6 +34,7 @@ public T clone (T)(in T input)
 shared static this ()
 {
     registerCommonInternetSchema("agora", 2826);
+    registerCommonInternetSchema("dns", 53);
```
Geod24: I've included this commit in my rebase

source/agora/node/Registry.d
```diff
@@ -751,14 +753,28 @@ private struct ZoneData
             "FOREIGN KEY(pubkey) REFERENCES registry_%s_utxo(pubkey) ON DELETE CASCADE, " ~
             "PRIMARY KEY(pubkey, address))", zone_name, zone_name);
 
+        // Initialize common fields
+        this.soa = this.config.fromConfig(this.root);
+
         this.db.execute(query_sig_create);
         this.db.execute(query_addr_create);
+        this.updateSOA();
```
Geod24: I think this call should be right after `this.soa = ...` (L757).

source/agora/node/Registry.d
```diff
@@ -525,12 +527,12 @@ unittest
...
 
 /// Converts a `ZoneConfig` to an `SOA` record
-private SOA fromConfig (in ZoneConfig zone, Domain name, uint serial) @safe pure
+private SOA fromConfig (in ZoneConfig zone, Domain name) @safe pure
 {
     SOA soa;
     soa.mname = Domain(format("ns1.%s", name.value));
     soa.rname = Domain(zone.soa.email.value.replace('@', '.'));
-    soa.serial = serial;
+    soa.serial = 0;
```
Geod24: Just rm the line

devel/config-fullnode.yaml
```diff
@@ -33,9 +33,13 @@ validator:
...
   flash:
-    email: github@bosagora.io
+    type: primary
+    soa:
+      email: github@bosagora.io
   validators:
-    email: github@bosagora.io
+    type: primary
+    soa:
+      email: github@bosagora.io
```
Geod24: The commit message should explain why this is done.
I'm reviewing commit-per-commit and currently I can't tell why we should do this. If an `SOA` is specified, surely it's a `primary` zone. If it's `authoritative` but no `SOA` is specified, surely it's a secondary zone ?

mkykadir:
Rebased

source/agora/node/Registry.d
```diff
@@ -737,14 +741,21 @@ private struct ZoneData
         this.config = config;
         this.root = root;
 
-        static string serverType (bool auth, bool primary)
+        static string serverType (ZoneConfig.Type zone_type)
         {
-            return primary ? "primary" : (auth ? "secondary" : "caching");
+            switch (zone_type)
+            {
+                case ZoneConfig.Type.disabled: return "disabled";
```
Geod24: The idea is that we never disable it.

source/agora/node/Registry.d
```diff
@@ -129,8 +136,8 @@ public class NameRegistry: NameRegistryAPI
     private void collectStats (Collector collector)
     {
         RegistryStats stats;
-        stats.registry_validator_record_count = this.zones[this.validators].count();
-        stats.registry_flash_record_count = this.zones[this.flash].count();
+        stats.registry_validator_record_count = this.zones[1].count();
+        stats.registry_flash_record_count = this.zones[2].count();
```
Geod24: I wouldn't mind this refactoring, but in its own commit. Also we should use an `enum` here.

source/agora/node/Runner.d
```diff
@@ -320,7 +320,7 @@ private void runDNSServer_canThrow (in RegistryConfig config, NameRegistry regis
             auto pack = udp.recv(buffer, ppeer);
             auto query = deserializeFull!Message(pack);
             registry.answerQuestions(
-                query,
+                query, peer.toAddressString,
```
Geod24: Does it do a reverse DNS lookup or just spits out an IP ?

source/agora/node/Registry.d
```diff
@@ -351,8 +351,8 @@ public class NameRegistry: NameRegistryAPI
     {
         Message reply;
         reply.header.RCODE = Header.RCode.FormatError;
-        reply.header.RA = false; // TODO: Implement
-        reply.header.AA = true;  // TODO: Make configurable
+        reply.header.RA = false;
+        reply.header.AA = true;
```
Geod24: Caching shouldn't be AA should it ?

Geod24 (commented):
Did a first pass.

source/agora/node/Registry.d
mkykadir: [This](https://github.com/bosagora/agora/pull/2703/files#diff-9a7d511b8407367630fb104c60937f8c304fb9d6eed6ff2281d46541c93513f9R1174) handles that. I will move setting those fields into `answer` method within the zone.

devel/config-fullnode.yaml
mkykadir: It doesn't differ, I thought that this way would be more clearer to differentiate between different zone types (while configuring and coding) commit message states that but if it seems unnecessary to you, I can revert.

source/agora/node/Registry.d
mkykadir: This was required due to new additions to `ZoneConfig`; will be separated.

source/agora/node/Registry.d
mkykadir: That's outdated; #2788 changed this in a way that parameters are passed directly to Ctor

source/agora/node/Runner.d
mkykadir: Just string representation of the IP

mkykadir:
Rebased, I needed to remove `unittest` of the Registry since it now requires a TaskManager to set-up timers. #2790 will open a door for adding proper Registry tests under `tests`

source/agora/node/Registry.d
mkykadir: Separated and merged in new PR #2825

devel/config-fullnode.yaml
mkykadir: Reverted as requested

mkykadir:
Rebased

tests/system/node/0/config.yaml
```diff
@@ -48,6 +48,26 @@ validator:
...
+registry:
+  enabled: true
+  port: 3053
+
+  validators:
+    authoritative: true
+    primary: integration.testnet
+
+    allow_transfer:
+      - node-2
```
mkykadir: This field requires IP address to be provided.

source/agora/test/Base.d
```diff
@@ -2312,7 +2313,18 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
...
-      registry: { enabled: true, },
+      registry: {
+          validators: {
+              authoritative: true,
+              primary: "test.name.registry",
+              soa: { email: SetInfo!string("test@testnet"), },
+          },
+          flash: {
+              authoritative: true,
+              primary: "test.name.registry",
```
Geod24: Is that correct ? Shouldn't that be `name.registry`, as this one is the authoritative registry ?

source/agora/node/Registry.d
```diff
@@ -576,13 +578,13 @@ unittest
...
 /// Converts a `ZoneConfig` to an `SOA` record
-private SOA fromConfig (in ZoneConfig zone, Domain name, uint serial) @safe
+private SOA fromConfig (in ZoneConfig zone, Domain name) @safe
 {
     return SOA(
         // mname, rname
         Domain.fromString(zone.primary),
         Domain.fromString(zone.soa.email.value.replace('@', '.')),
-        serial,
+        0,
```
Geod24: Add a comment saying it's set from `updateSOA` ? Or better, move `updateSOA` at the end of `fromConfig` ?

source/agora/node/Registry.d
```diff
@@ -823,14 +825,28 @@ private struct ZoneData
...
+
+    public void updateSOA () @trusted
+    {
         if (this.type == ZoneType.primary)
         {
-            // Serial's value wraps around so the cast is safe
-            const serial = cast(uint) Clock.currTime(UTC()).toUnixTime();
-            this.soa = this.config.fromConfig(this.root, serial);
+            this.soa.serial = cast(uint) Clock.currTime(UTC()).toUnixTime();
+            return;
```
Geod24:
```D
if (this.type != ZoneType.primary)
    return;
this.soa.serial = ...;
```

The `else` is not required anymore since we're no longer in the constructor.

doc/config.example.yaml
```diff
@@ -363,6 +364,20 @@ registry:
 
   # Zone responsible for flash nodes registration
   flash:
-    # Having separate authoritative server increases redundancy and reduces load,
+    # Having a secondary server increases redundancy and reduces load,
```
Geod24: "one or more secondary servers" ?

source/agora/node/Config.d
```diff
@@ -463,9 +463,20 @@ public struct RegistryConfig
...
+            if (!zone.primary.set && zone.authoritative.set)
+            {
+                ensure(zone.primary_servers.length > 0,
+                    "registry.{} is secondary and at least one primary is required",
+                    name);
+
+                ensure(zone.redirect_primary.length > 0,
+                    "registry.{} is secondary (authoritative), thus requires an " ~
+                    "address registering endpoint to an Agora node",
+                    name);
```
Geod24: Perhaps we should make it default to `primary` ? It seems that in most configuration, that'd be the same.

source/agora/node/Config.d
```diff
@@ -463,9 +463,20 @@ public struct RegistryConfig
...
                    "'email' fields to be set", name);
 
-            // If we're not primary, there's no configuration to validate
-            if (!zone.primary.set)
+            // Secondary registry
+            if (!zone.primary.set && zone.authoritative.set)
+            {
+                ensure(zone.primary_servers.length > 0,
+                    "registry.{} is secondary and at least one primary is required",
+                    name);
```
Geod24: Hold on... We have `primary`, `primary_servers`, and `redirect_primary` ? That's a bit much, isn't it ?

source/agora/node/Registry.d
```diff
@@ -226,6 +232,12 @@ public class NameRegistry: NameRegistryAPI
 
     public override void postValidator (RegistryPayload registry_payload)
     {
+        if (this.zones[ZoneIndex.Validator].type == ZoneType.secondary)
+        {
+            this.zones[ZoneIndex.Validator].redirect_primary.postValidator(registry_payload);
+            return;
+        }
```
Geod24: Any redirect should happen after we've checked the payload is valid (at the very least the signature), otherwise it's a very obvious DDoS vector.

source/agora/node/Registry.d
```diff
@@ -314,6 +332,7 @@ public class NameRegistry: NameRegistryAPI
         // register data
         log.info("Registering network addresses: {} for Flash public key: {}", registry_payload.data.addresses,
             registry_payload.data.public_key.toString());
+
```
Geod24: ?

source/agora/node/Registry.d
```diff
@@ -583,7 +617,9 @@ private SOA fromConfig (in ZoneConfig zone, Domain name) @safe
     return SOA(
         // mname, rname
         Domain.fromString(zone.primary),
-        Domain.fromString(zone.soa.email.value.replace('@', '.')),
+        (zone.soa.email.length > 0) ?
+            Domain.fromString(zone.soa.email.value.replace('@', '.'))
+            : Domain.init, // Secondary will get this from primary
```
Geod24: It will get the rest from primary too, so maybe we should just not be calling `fromConfig` for non-primary ?

source/agora/node/Runner.d
```diff
@@ -305,7 +305,8 @@ private void runDNSServer_canThrow (in RegistryConfig config, NameRegistry regis
     // The `listenUDP` needs to be in the `runTask` otherwise we get
     // a fatal error due to a bug in vibe-core (see comment #2):
     /// https://github.com/vibe-d/vibe-core/issues/289
-    auto udp = listenUDP(config.port, config.address);
+    auto udp = listenUDP(config.port, config.address,
+        UDPListenOptions.reuseAddress | UDPListenOptions.reusePort);
```
Geod24, line 309: Own commit please

doc/config.example.yaml
```diff
@@ -342,6 +342,14 @@ registry:
     allow_transfer:
       - 127.0.0.1
 
+    # Whitelist for AXFR zone transfers
+    #
+    # Only servers listed in this field will be allowed to initiate an AXFR zone
+    # transfer from this server. Server will not allow zone transfer to any
+    # requester if not configured. Applicable for primary and secondary servers.
+    allow_transfer:
+      - ns2.bosagora.io
```
Geod24: Can't be a domain, can it ?

Geod24: Also should be merged in the commit that implements it

Geod24 (changes requested):
First two commits are G2G, need to do a bit more review, but left a few comments already.

source/agora/node/Config.d
mkykadir: Yes, when I was started `primary` wasn't there, thus I need to think over that again. `primary_servers` can have multiple servers configured as a backup.

source/agora/node/Registry.d
mkykadir: Non-primaries call this to initialize timing fields of their SOA as a backup value (in-case of they couldn't reach any primary to get actual SOA)

doc/config.example.yaml
mkykadir: Implementing commit merged with #2825, ~can separate it to another PR~ actually that PR had documentation already

source/agora/node/Registry.d
```diff
@@ -577,7 +580,8 @@ private SOA fromConfig (in ZoneConfig zone, Domain name, uint serial) @safe
 {
     return SOA(
         // mname, rname
-        Domain.fromString(zone.primary), Domain.fromString(zone.soa.email.value.replace('@', '.')),
+        Domain.fromString(zone.primary),
+        Domain.fromString(zone.soa.email.value.replace('@', '.')),
```
Geod24: This is a style change. Either make it its own commit or revert it.

source/agora/node/Registry.d
```diff
@@ -752,14 +766,23 @@ private struct ZoneData
...
+            if (this.config.soa.email.set)
+                this.type = ZoneType.primary;
+        }
+
+        static string serverType (ZoneType zone_type)
+        {
+            switch (zone_type)
+            {
+                case ZoneType.primary: return "primary (authoritative)";
+                default: return "Unknown";
```
Geod24: I know this is fixed in a later commit, but we should not return "Unknown" here.

source/agora/node/Registry.d
mkykadir: `fromConfig` is module scope and I don't think it would be good to move it into `ZoneData`. Since `Serial` is not a configuration value, I will add a comment.

mkykadir:
Rebased now depends on #2907

source/agora/node/Registry.d
mkykadir: I can either set default values for time values of `SOA` struct in `DNS` or keep this way of initializing it. `SOAConfig` already have default values which I wanted to utilize.
[This](https://github.com/mkykadir/agora/blob/a767a72dae19e96e6c423c70fe3223f99401de12/source/agora/node/Registry.d#L1026) is the reason why we need some default values.

source/agora/node/Config.d
mkykadir: Currently, `primary` is for NS record which cannot cover secondary servers right now (thinking a way)
`primary_servers` are list of DNS servers for a secondary to perform AXFR from and recursion endpoint for cachings; if one is not reachable second one will be tried
`redirect_primary` must be an TCP/HTTP endpoint that has `postValidator/postFlashNode` available for a secondary server. Thus this cannot replace `primary` directly; but I am thinking for `primary` and `primary_servers` right now.

mkykadir:
Rebased and resolved conversations except `primary` and `redirect_primary`

doc/config.example.yaml
```diff
@@ -368,6 +369,26 @@ registry:
 
   # Zone responsible for flash nodes registration
   flash:
-    # Having separate authoritative server increases redundancy and reduces load,
+    # Having one or more secondary servers increase redundancy and reduces load,
     # hence this configuration, while unlikely in a small to medium network, is supported.
+    authoritative: true
+
+    # Redirect server for REST endpoints of the registry
+    redirect_primary: http://nsHidden.bosagora.io
```
Geod24: Technically we don't have to redirect to a primary, so the name is a bit of a misnomer. But let's leave it as is for now.

source/agora/node/Registry.d
```diff
@@ -834,31 +887,159 @@ private struct ZoneData
...
+                    format("DROP TABLE IF EXISTS registry_%s_utxo", zone_name));
+                this.db.execute(
+                    format("DROP TABLE IF EXISTS registry_%s_addresses", zone_name));
+            }
+
+            // DNS resolver is used to get SOA RR and performing AXFR
+            auto peer_addrs = this.config.primary_servers.map!(
+                peer => Address("dns://" ~ peer)
+            ).array();
+            this.resolver = node.makeDNSResolver(peer_addrs);
```
Geod24: Why doesn't the `Node` do that itself ?

source/agora/node/Registry.d
```diff
@@ -834,31 +887,159 @@ private struct ZoneData
...
+            auto settings = new RestInterfaceSettings;
+            settings.baseURL = Address(this.config.redirect_primary);
+            settings.httpClientSettings = new HTTPClientSettings;
+            settings.httpClientSettings.connectTimeout = 2.seconds;
+            settings.httpClientSettings.readTimeout = 2.seconds;
+            this.redirect_primary = new RestInterfaceClient!NameRegistryAPI(settings);
+
+            this.expire_timer = this.taskman.createTimer(&this.disable);
+
+            this.soa_update_timer = this.taskman.setTimer(0.seconds, &this.updateSOA);
```
Geod24: Not going to work in unittests

Geod24 (commented):
Second commit: Nope, we should not add Agora-specific definitions / value to `common.DNS`, especially with the provided reasoning. If we can't reach a primary, relying on semi arbitrary defaults seems prone to many issues.

source/agora/node/Registry.d
mkykadir: It's per-Zone. Each zone can have different list of primaries.

doc/config.example.yaml
```diff
@@ -368,6 +369,26 @@ registry:
...
+
+    # List of servers to AXFR zone transfer from
+    #
+    # Servers are tried in the configuration order until zone transfer can be
+    # completed from a server.
+    primary_servers:
+      - ns1.bosagora.io
+
+    # A secondary server can be primary of an another server, thus secondary
```
Geod24: No it can't. It can be an "upstream" server, but primary == ultimate authority on the zone.

doc/config.example.yaml
mkykadir: Renamed this and `primary_servers` and better documented

Geod24 (approved):
LGTM

Geod24:
```
2022-01-27 09:28:19,226 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'unittest.bosagora.io.'
2022-01-27 09:28:19,226 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_realm_utxo in "SELECT * FROM registry_realm_utxo WHERE utxo = ?"
2022-01-27 09:28:19,226 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'validators.unittest.bosagora.io.'
2022-01-27 09:28:19,226 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_validator_utxo in "SELECT * FROM registry_validator_utxo WHERE utxo = ?"
2022-01-27 09:28:19,227 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'flash.unittest.bosagora.io.'
2022-01-27 09:28:19,227 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_flash_utxo in "SELECT * FROM registry_flash_utxo WHERE utxo = ?"
```

Geod24:
The previous message logs were for one of the validator, but those ones are for the registry:
```
Registry logs:
Log for node: 10.8.8.8
======================================================================
2022-01-27 09:37:06,644 Info [agora.consensus.state.Ledger] - Last known block: #0 (0x8365f069fe37ee02f2c4dc6ad816702088fab5fc875c3c67b01f82c285aa2d90b605f57e068139eba1f20ce20578d712f75be4d8568c8f3a7a34604e72aa3175)
2022-01-27 09:37:06,647 Info [agora.consensus.state.Ledger] - Beginning externalization of block #0
2022-01-27 09:37:06,647 Info [agora.consensus.state.Ledger] - Transactions: 3 - Enrollments: 6
2022-01-27 09:37:06,647 Info [agora.consensus.state.Ledger] - Validators: Active: 0 - Signing: 000000 - Slashed: 0
2022-01-27 09:37:06,653 Info [agora.consensus.state.Ledger] - Completed externalization of block #0
2022-01-27 09:37:06,654 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'unittest.bosagora.io.'
2022-01-27 09:37:06,654 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_realm_utxo in "SELECT * FROM registry_realm_utxo WHERE utxo = ?"
2022-01-27 09:37:06,654 Info [agora.node.Registry] - Registry is primary (authoritative) DNS server for zone 'validators.unittest.bosagora.io.'
2022-01-27 09:37:06,654 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_validator_utxo in "SELECT * FROM registry_validator_utxo WHERE utxo = ?"
2022-01-27 09:37:06,655 Info [agora.node.Registry] - Registry is primary (authoritative) DNS server for zone 'flash.unittest.bosagora.io.'
2022-01-27 09:37:06,655 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_flash_utxo in "SELECT * FROM registry_flash_utxo WHERE utxo = ?"
```

Besides the same errors, the following sounds wrong:
> 2022-01-27 09:37:06,654 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'unittest.bosagora.io.'

mkykadir:
> ```
> 2022-01-27 09:28:19,226 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'unittest.bosagora.io.'
> 2022-01-27 09:28:19,226 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_realm_utxo in "SELECT * FROM registry_realm_utxo WHERE utxo = ?"
> 2022-01-27 09:28:19,226 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'validators.unittest.bosagora.io.'
> 2022-01-27 09:28:19,226 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_validator_utxo in "SELECT * FROM registry_validator_utxo WHERE utxo = ?"
> 2022-01-27 09:28:19,227 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'flash.unittest.bosagora.io.'
> 2022-01-27 09:28:19,227 Error [agora.consensus.pool.Transaction] - SQLite error: (1) no such table: registry_flash_utxo in "SELECT * FROM registry_flash_utxo WHERE utxo = ?"
> ```

These logs are from #2822 ; I am checking for a table to understand if our configuration is changed from a primary to another

mkykadir:
> > 2022-01-27 09:37:06,654 Info [agora.node.Registry] - Registry is caching (non-authoritative) DNS server for zone 'unittest.bosagora.io.'

Unittest config realm configured in [that way](https://github.com/bosagora/agora/blob/51a7ba60966c85f6bc315cf039dd5c380e1d64ce/source/agora/test/Base.d#L2073)

Geod24:
Alright, we can fix this later. I've retriggered the CI to try to get a green run.

Geod24:
MacOS passed.

PR #2705, opened 2021-11-26
DScanner: Clear up remaining warnings

Geod24:
See each commit for what steps are taken. In practice, we end up with:
```
% dscanner --skipTests --config devel/dscanner.ini --styleCheck source
source/agora/consensus/data/Block.d(161:35)[warn]: Argument 2 is named 'sig', but this is the name of parameter 3
source/agora/consensus/data/Block.d(394:48)[warn]: Avoid subtracting from '.length' as it may be unsigned.
source/agora/crypto/Bech32.d(149:25)[warn]: Avoid subtracting from '.length' as it may be unsigned.
source/agora/test/Base.d(308:16)[warn]: Catching Error or Throwable is almost always a bad idea.
source/agora/network/RPC.d(455:43)[warn]: Parameter api is never used.
source/agora/cli/vanity/main.d(261:57)[warn]: Avoid subtracting from '.length' as it may be unsigned.
source/agora/flash/Channel.d(829:57)[warn]: Parameter tx is never used.
source/agora/flash/OnionPacket.d(162:47)[warn]: Parameter payment_hash is never used.
source/agora/flash/OnionPacket.d(184:29)[warn]: Avoid subtracting from '.length' as it may be unsigned.
source/agora/flash/Node.d(1022:53)[warn]: Avoid subtracting from '.length' as it may be unsigned.
source/agora/flash/Node.d(1321:44)[warn]: Parameter reg_pk is never used.
source/agora/common/DNS.d(852:58)[warn]: Avoid subtracting from '.length' as it may be unsigned.
source/agora/common/Set.d(309:28)[warn]: Avoid subtracting from '.length' as it may be unsigned.
source/agora/common/Ensure.d(94:20)[warn]: Catching Error or Throwable is almost always a bad idea.
```

I have checked every instance, except for the "unused parameters" on flash, which I've forwarded to @omerfirmak .
Either they are correct, or a bug in D-Scanner.

### DIFF
--- a/source/agora/flash/api/FlashAPI.d
+++ b/source/agora/flash/api/FlashAPI.d
@@ -175,6 +175,13 @@ public struct ChannelUpdate
             && this.fixed_fee == other.fixed_fee
             && this.proportional_fee == other.proportional_fee;
     }
+
+    /// Required when `opEquals` is implemented
+    public size_t toHash () const scope @trusted pure nothrow @nogc
+    {
+        static assert(this.chan_id.sizeof >= size_t.sizeof);
+        return *(cast(size_t*) this.chan_id[].ptr);
+    }
 }
 
 unittest
--- a/devel/dscanner.ini
+++ b/devel/dscanner.ini
@@ -51,7 +51,7 @@ comma_expression_check="enabled"
 ;  used with D versions older than 2.071.0
 local_import_check="disabled"
 ; Checks for variables that could be declared immutable
-could_be_immutable_check="enabled"
+could_be_immutable_check="disabled"
 ; Checks for redundant expressions in if statements
 redundant_if_check="enabled"
 ; Checks for redundant parenthesis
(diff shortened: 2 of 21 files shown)


source/agora/flash/api/FlashAPI.d
```diff
@@ -175,6 +175,12 @@ public struct ChannelUpdate
             && this.fixed_fee == other.fixed_fee
             && this.proportional_fee == other.proportional_fee;
     }
+
+    /// Required when `opEquals` is implemented
+    public size_t toHash () const scope @trusted pure nothrow @nogc
+    {
+        return *(cast(size_t*) this.chan_id[].ptr);
```
omerfirmak, line 183: Could use a `static assert(sizeof(this.chan_id) >= sizeof(size_t))`

omerfirmak (approved):

source/agora/flash/api/FlashAPI.d
Geod24: Done

PR #2707, opened 2021-11-27
DScanner: Clear up remaining Flash related warnings

omerfirmak:


source/agora/flash/Channel.d
```diff
@@ -826,9 +826,8 @@ LOuter: while (1)
 
     ***************************************************************************/
 
-    private void onSettleTxExternalized (in Transaction tx)
+    private void onSettleTxExternalized (in Transaction)
```
Geod24, line 829: So why is there a parameter in the first place ? Is it used as a delegate ?

omerfirmak: We could remove it, but I left it there just in case we want to do something with it in the future.

Geod24 (approved):

PR #2708, opened 2021-11-28
DNS: Actually use the configuration value 'primary'

Geod24:
We started reading that value in a previous PR, but didn't make use of it.


Geod24:
Giving this a speedy merge as it's blocking v0.28.0 and I just tested it successfully live.

PR #2709, opened 2021-11-29
Correct documentation for return value of `acceptBlock`

mkykadir:
Small mistake that caught my eye

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -275,7 +275,7 @@ public class Ledger
             block = the block to add
 
         Returns:
-            true if the block was accepted
+            an error message if the block is not accepted, otherwise null
 
     ***************************************************************************/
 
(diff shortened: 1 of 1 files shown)


source/agora/consensus/Ledger.d
```diff
@@ -275,7 +275,7 @@ public class Ledger
             block = the block to add
 
         Returns:
-            true if the block was accepted
+            the error message if block is not accepted, otherwise null
```
linked0:
```suggestion
            an error message if the block is not accepted, otherwise null
```
It would be better, IMO.

linked0 (approved):
LGTM but one suggestion.

source/agora/consensus/Ledger.d
mkykadir: Then, also updated for `validateBlock`

omerfirmak (approved):

PR #2711, opened 2021-11-30
Deserialize RDATA for supported types

mkykadir:
For A, SOA and CNAME records RDATA will provide deserialized versions. This will help us to deserialize Resource Records fetched through AXFR query (`Domain` requires special deserializer context).

This PR also fixes an error introduced with 1bccc2c2c0a7c1642ef7cdee718728a3471b191e which caused A records to appear as CNAME records in the answer and IP octets sent in reverse order for addresses.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -167,6 +167,19 @@ public struct SOA
     /// The unsigned 32 bit minimum TTL field that should be
     /// exported with any RR from this zone.
     public uint minimum;
+
+    /// Support for network serialization
+    public static T fromBinary (T) (scope ref DNSDeserializerContext ctx) @safe
+    {
+        return T(Domain.fromBinary!(typeof(T.mname))(ctx),
+            Domain.fromBinary!(typeof(T.rname))(ctx),
+            deserializeFull!(uint)(&ctx.read, ctx.options),
+            deserializeFull!(int)(&ctx.read, ctx.options),
+            deserializeFull!(int)(&ctx.read, ctx.options),
+            deserializeFull!(int)(&ctx.read, ctx.options),
+            deserializeFull!(uint)(&ctx.read, ctx.options),
+        );
+    }
 }
 
 /// https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.14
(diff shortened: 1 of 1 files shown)


source/agora/common/DNS.d
```diff
@@ -611,8 +611,9 @@ public struct ResourceRecord
     public static ResourceRecord make (TYPE type : TYPE.A) (
         Domain name, uint ttl, uint[] ipv4...) @trusted
     {
+
```
Geod24: ?

source/agora/common/DNS.d
```diff
@@ -167,6 +167,18 @@ public struct SOA
...
+    {
+        return T(Domain.fromBinary!(typeof(T.mname))(ctx),
+            Domain.fromBinary!(typeof(T.rname))(ctx),
+            deserializeFull!(uint)(&ctx.read, ctx.options),
+            deserializeFull!(int)(&ctx.read, ctx.options),
+            deserializeFull!(int)(&ctx.read, ctx.options),
+            deserializeFull!(int)(&ctx.read, ctx.options),
+            deserializeFull!(uint)(&ctx.read, ctx.options),
+        );
+    }
```
Geod24: Good catch. Just add `///` so it shows up in documentation.

source/agora/common/DNS.d
```diff
@@ -656,28 +698,95 @@ public struct ResourceRecord
...
         return T(
-            tmp.name, tmp.type, tmp.class_, tmp.ttl,
-            () @trusted { return cast(typeof(T.rdata)) ctx.read(rdlength); }(),
+            tmp.name, tmp.type, tmp.class_, tmp.ttl, tmp_data,
         );
     }
 
+    /// Retruns serialized Rdata, original value is returned when Rdata already
+    /// stores serialized data
+    public const(ubyte[]) serializedRdata () const @safe
```
Geod24: Why don't you implement `serialize` instead ?

Geod24 (approved):
Generally LVGTM. Biggest point is to use `serialize` as it's understood by the serializer, after that feel free to self merge.

Issue #2714, opened 2021-11-30
Registry should only answer for asked query type


mkykadir:
Currently, since `ZoneData.get` does not have query type provided, Registry answers with the type in the storage even it is not what queried. Following query should result in empty answer section

```
;; QUESTION SECTION:
;**masked**.validators.localhost. IN CNAME

;; ANSWER SECTION:
**masked**.validators.localhost. 0	IN A 127.0.0.1
```

PR #2715, opened 2021-11-30
Registry: Only answer with records matching query type

mkykadir:
This PR also refactors `QTYPE` in DNS to use base values from `TYPE`


source/agora/node/Registry.d
```diff
@@ -872,6 +875,11 @@ private struct ZoneData
 
         // Address loop consumes data, gather following first
         const TYPE node_type = to!TYPE(results.front["type"].as!ushort);
+
+        // Check query type; QTYPE is superset of TYPE, casting is OK
+        if (type != QTYPE.ALL && type != cast(QTYPE) node_type)
+            return TypedPayload.init;
+
```
Geod24: This is not quite correct though:

> CNAME RRs cause special action in DNS software.  When a name server
fails to find a desired RR in the resource set associated with the
domain name, it checks to see if the resource set consists of a CNAME
record with a matching class.  If so, the name server includes the CNAME
record in the response and restarts the query at the domain name
specified in the data field of the CNAME record.  The one exception to
this rule is that queries which match the CNAME type are not restarted.

[RFC1034, 3.6.2. Aliases and canonical names](https://datatracker.ietf.org/doc/html/rfc1034#section-3.6.2)

mkykadir: We're looking per zone basis, thus if we're not planning to allow a node to have CNAME record pointing to another public key within the same zone, section 3.6.2 is not applying for us. Section 3.6.2 can only be applied for following condition for us and I don't know if it is achievable;

```
boa1pubkey.validators.localhost. IN CNAME boa1pubkey2.validators.localhost.
boa2pubkey.validators.localhost. IN A 127.0.01
```

Geod24 (approved):

linked0:
It fixed #2714

PR #2716, opened 2021-11-30
Signature: Introduce OmitSingle

omerfirmak:
This will let us sign multiple outputs while letting one of them
to be open to modifications.

Need for this comes from the settlement TXs in Flash. Settlement
may include multiple outputs for the channel funds which are signed
at the time of the micro-transaction and we may need to add a refund
output for the inputs we add to fund the fees for the externalization
of the tx.

### DIFF
--- a/source/agora/script/Signature.d
+++ b/source/agora/script/Signature.d
@@ -39,9 +39,13 @@ public enum SigHash : ubyte
     /// Modifier that can only be used with other SigHash types
     AnyoneCanPay = 1 << 3,
 
+    /// Omits signing an output
+    OmitSingle = 1 << 4,
+
     /// Combined types
     Single_AnyoneCanPay = Single | AnyoneCanPay,
     Single_NoInput_AnyoneCanPay = Single | NoInput | AnyoneCanPay,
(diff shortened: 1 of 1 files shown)


omerfirmak:
Similar to `SIGHASH_GROUP` discussed for Eltoo implementations in bitcoin mailing lists, but we don't need to define a range of outputs for our use case, and it brings unnecessary complexity.

linked0 (approved):
LGTM

source/agora/script/Signature.d
```diff
@@ -78,7 +82,13 @@ public struct SigPair
     public inout(ubyte)[] opSlice () inout pure nothrow @safe /*@nogc*/
     {
         return this.signature.toBlob()[] ~ ubyte(this.sig_hash)
-            ~ ((sig_hash & SigHash.Single) ? nativeToLittleEndian(this.output_idx).dup : []);
+            ~ (SigPair.usesOutputIdx(this.sig_hash) ? nativeToLittleEndian(this.output_idx).dup : []);
```
Geod24: We probably shouldn't be allocating in an `opSlice`, but that's an improvement for another day

PR #2719, opened 2021-12-01
DNS: Merge two definitions of RDATA

Geod24:
```
The new union approach is superior, but some lingering declarations
were left behind.
```

I also did a few minor changes: using visibility qualifiers, adding / moving documentation, and renaming `cname` to just `name` as it's used by other records (e.g. `NS`).

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -182,21 +182,6 @@ public struct SOA
     }
 }
 
-/// https://datatracker.ietf.org/doc/html/rfc1035#section-3.3.14
-public struct RDATA
-{
-    /// A 32 bit Internet address.
-    public uint address;
-}
-
-/// https://datatracker.ietf.org/doc/html/rfc3596#section-2.2
-public struct AAAARDATA
-{
-    /// A 128 bit IPv6 address is encoded in the data portion of an AAAA
-    /// resource record in network byte order (high-order byte first).
-    public ulong[2] address;
-}
-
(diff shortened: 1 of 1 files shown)


mkykadir (approved):

omerfirmak:
This apparently broke docker build

```
source/agora/common/DNS.d(670,13): Error: field `RDATA.a` cannot access pointers in `@safe` code that overlap other fields
source/agora/common/DNS.d(675,13): Error: field `RDATA.name` cannot access pointers in `@safe` code that overlap other fields
source/agora/common/DNS.d(680,13): Error: field `RDATA.soa` cannot access pointers in `@safe` code that overlap other fields
Type 'FlashControlAPI' implements more than one interface: make sure the one describing the REST server is the first one
```

linked0:
It needs to upgrade the `ldc2` to the latest version.

omerfirmak:
> It needs to upgrade the `ldc2` to the latest version.

Weirdly, it is already on v.1.28.0

linked0:
I had the same problem on the local machine yesterday. I solved the problem with `brew update & brew upgrade`, so I thought the version was old. But it could be aroused from another cause, I guess.

mkykadir:
Actual PR that added mentioned code is #2711 BTW, it didn't throw any errors on my machine

Geod24:
As discussed on Slack, it was an outdated `agora-builder` locally.

PR #2722, opened 2021-12-02
Manager: Try to connect to last known Validators

omerfirmak:
If we are a not-enrolled Validator with no network configuration
we should atleast try to connect to last known validators to catch up.

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -1337,6 +1337,27 @@ public class Ledger
         return this.utxo_set.peekUTXO(utxo, value);
     }
 
+    /// Returns: UTXOs for validator active at the given height
+    public UTXO[Hash] getEnrolledUTXOs (in Height height) @safe nothrow
+    {
+        UTXO[Hash] utxos;
+        Hash[] keys;
+        if (this.enroll_man.validator_set.getEnrolledUTXOs(height, keys))
+            foreach (key; keys)
+            {
+                UTXO val;
+                assert(this.peekUTXO(key, val));
+                utxos[key] = val;
+            }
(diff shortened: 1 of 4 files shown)


source/agora/network/Manager.d
```diff
@@ -759,11 +759,8 @@ public class NetworkManager
         {
             if (this.connection_tasks.length >= MaxConnectionTasks)
             {
-                log.info("Connection task limit reached. Trying again in {}..",
-                    this.node_config.retry_delay);
-
-                this.taskman.wait(this.node_config.retry_delay);
-                continue;
+                log.info("Connection task limit reached. will trying again..");
```
Geod24: "Will try again". So when will it try again ?

linked0 (approved):
Looks good!

mkykadir:
A different approach, since #2697 always enables registry and we soon will have a caching mechanism in registry this might not be needed in future. Until then, it LGTM.

source/agora/consensus/Ledger.d
```diff
@@ -1337,6 +1337,27 @@ public class Ledger
...
+            }
+        return utxos;
+    }
+
+    /// Ditto
+    public UTXO[Hash] getEnrolledUTXOs () @safe nothrow
+    {
+        return this.getEnrolledUTXOs(this.getBlockHeight() + 1);
+    }
+
```
Geod24, line 1360: We really ought to stop allocating AAs all over the place though.

Geod24 (approved):
Just one typo

source/agora/network/Manager.d
omerfirmak: Updated

PR #2723, opened 2021-12-02
Add `TransactionResult` struct for result of posting transaction

### LINKED ISSUE #1082: Transaction pool & gossiping might need to be slightly improved
Thanks to https://github.com/Geod24/localrest/pull/90 I've found an edge-case in the Validator:

Consider a case of 3 validator nodes:
- Nodes reach consensus on block height 1
- Externalize messages are propagated throughout the network
- One node has networking issues towards the other nodes and has not received the externalize message yet
- Node 1 is at block height 0, nodes 2 and 3 are at block height 1.

In this situations nodes [2, 3] will accept transactions which reference UTXOs from block height 1, but node 0 will not. 

Visual example:

```
N2 --- N3
 \    /
   N1
   |
   User
```

In this case the network between N1 and N2&N3 is broken, but the User can still reach N1. If this user sends a payment to N1 then this TX is rejected and **not** gossiped to the rest of the network (regardless of networking issues) and it's also **not** added to N1's pool. Ideally we want N1 to eventually store this TX in its pool.

I can think of several solutions:

A) Periodically sync the TX pools between nodes (might be costly, but could be minimized via techniques such as merkle root)
B) If the transactions are otherwise well-structured keep them in a pending state for some time before purging them (in case the Validator node is behind the rest of the network).

Actually for the B solution what makes things complicated is that we can't verify the TX signature because it signs the UTXO but we don't know which Public key it is until we find that UTXO. And so it can become a DDoS vector.

Open to ideas & discussion.

linked0:
We need to know the details about why the posting transaction has failed.
So we introduce a new structure for identifying whether the operation
succeeds and the reason when the operation fails.

Fixes #1082

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -289,9 +289,12 @@ public interface API
         Params:
             tx = a Transaction being posted
 
+        Returns:
+            a TransactionResult object, the result of posting the transaction
+
     ***************************************************************************/
 
-    public void postTransaction (in Transaction tx);
+    public TransactionResult postTransaction (in Transaction tx);
 
     /***************************************************************************
 
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -286,6 +286,9 @@ public interface API
         API:
             POST /transaction
 
+        Params:
+            tx = a Transaction being posted
+
     ***************************************************************************/
 
     public void postTransaction (in Transaction tx);
--- a/source/agora/consensus/data/Transaction.d
+++ b/source/agora/consensus/data/Transaction.d
@@ -377,6 +377,20 @@ unittest
     assert(expected2 == tx_freeze_hash, expected2.toString());
 }
 
+/// The information of the result of posting a transaction
+public struct TransactionResult
+{
+    public enum Status : uint
+    {
+        Accepted,
+        Duplicated,
+        Rejected,
+    }
+    
+    Status status;
+    string reason;
+}
+
 /*******************************************************************************
 
     Get sum of `Output`
--- a/source/agora/test/Flash.d
+++ b/source/agora/test/Flash.d
@@ -312,12 +312,17 @@ public class TestFlashNode : FlashNode, TestFlashAPI
     }
 
     ///
-    private void postTransaction (in Transaction tx)
+    private TransactionResult postTransaction (in Transaction tx)
     {
         if (this.allow_publish)
-            this.agora_node.postTransaction(tx);
+        {
+            return this.agora_node.postTransaction(tx);
+        }
         else
+        {
             log.info("Skipping publishing {}", tx);
+            return TransactionResult(TransactionResult.Status.Accepted);
+        }
     }
 }
 
--- a/source/agora/test/PostTransactionResult.d
+++ b/source/agora/test/PostTransactionResult.d
@@ -1,0 +1,63 @@
+/*******************************************************************************
+
+    Contains tests for the result of posting transactions which can be one
+    of these status: `Accepted``, `Duplicated`, and `Rejected`.
+
+    Copyright:
+        Copyright (c) 2019-2021 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.PostTransactionResult;
+
+version (unittest):
+
+import agora.api.Validator;
+import agora.consensus.data.genesis.Test;
+import agora.crypto.Hash;
+import agora.crypto.Key;
+import agora.test.Base;
+import agora.utils.Test;
+
+import core.thread.osthread : Thread;
+import std.algorithm.searching;
+
+/// test for posting transactions, waiting for the various results
+unittest
+{
+    TestConf conf = TestConf.init;
+    auto network = makeTestNetwork!TestAPIManager(conf);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto nodes = network.clients;
+    auto input_txs = GenesisBlock.txs.filter!(tx => tx.isPayment).array;
+    auto output_addr = WK.Keys.AA.address;
+
+    // create a transaction with the fee rate of 10000
+    auto tx = TxBuilder(input_txs[0], 0).feeRate(Amount(10_000))
+        .draw(Amount(100_000), [output_addr])
+        .sign();
+
+    // post the transaction, expect `Accepted`
+    auto result = nodes[0].postTransaction(tx);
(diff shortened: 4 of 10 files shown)


source/agora/consensus/data/Transaction.d
```diff
@@ -377,6 +377,21 @@ unittest
     assert(expected2 == tx_freeze_hash, expected2.toString());
 }
 
+// The status of the result of posting a transaction
```
mkykadir: One `/` is missing
`///`

source/agora/consensus/data/Transaction.d
```diff
@@ -377,6 +377,21 @@ unittest
...
+public enum TransactionStatus : uint
```
mkykadir: I couldn't directly relate this naming and enum with `postTransaction` and `TransactionResult`, how about renaming it as `Status` inside `TransactionResult`?

```
public struct TransactionResult
{
   public enum Status
   {...
```

source/agora/test/Flash.d
```diff
@@ -312,12 +312,16 @@ public class TestFlashNode : FlashNode, TestFlashAPI
     }
 
     ///
-    private void postTransaction (in Transaction tx)
+    private TransactionResult postTransaction (in Transaction tx)
     {
+        auto result = TransactionResult(TransactionStatus.Accepted);
```
mkykadir: Local variable can be removed

mkykadir (commented):
Return value is never used, one case is [here](https://github.com/bosagora/agora/blob/982431b12cc9d7303b1ff733b1f5fd042b71bab0/source/agora/cli/client/SendTxProcess.d#L204)

source/agora/test/PostTransactionResult.d
```diff
@@ -0,0 +1,61 @@
...
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    auto nodes = network.clients;
+    auto input_txs = GenesisBlock.txs.filter!(tx => tx.isPayment).array;
+    auto output_addr = WK.Keys.AA.address;
+
+    // create a transaction with the fee of 10000
```
hewison-chris:
```suggestion
    // create a transaction with the fee rate of 10000
```

source/agora/test/PostTransactionResult.d
```diff
@@ -0,0 +1,61 @@
...
+    auto tx = TxBuilder(input_txs[0], 0).feeRate(Amount(10_000))
+        .draw(Amount(100_000), [output_addr])
+        .sign();
+
+    // post the transaction, waiting for `Accepted`
+    auto result = nodes[0].postTransaction(tx);
+    assert(result == TransactionResult(TransactionStatus.Accepted));
+
+    // post a double spend transaction, waiting for `Duplicated`
```
hewison-chris:
```suggestion
    // post the same transaction again, expect `Duplicated`
```

source/agora/test/PostTransactionResult.d
```diff
@@ -0,0 +1,61 @@
...
+    auto nodes = network.clients;
+    auto input_txs = GenesisBlock.txs.filter!(tx => tx.isPayment).array;
+    auto output_addr = WK.Keys.AA.address;
+
+    // create a transaction with the fee of 10000
+    auto tx = TxBuilder(input_txs[0], 0).feeRate(Amount(10_000))
+        .draw(Amount(100_000), [output_addr])
+        .sign();
+
+    // post the transaction, waiting for `Accepted`
```
hewison-chris:
```suggestion
    // post the transaction, expect `Accepted`
```

source/agora/test/PostTransactionResult.d
```diff
@@ -0,0 +1,61 @@
...
+    // post a double spend transaction, waiting for `Duplicated`
+    result = nodes[0].postTransaction(tx);
+    assert(result == TransactionResult(TransactionStatus.Duplicated));
+
+    // post a transaction with no fee, waiting for `Rejected`
+    tx = TxBuilder(input_txs[0], 1).feeRate(Amount(0))
+        .draw(Amount(100_000), [output_addr])
+        .sign();
+    result = nodes[0].postTransaction(tx);
+    assert(result.status == TransactionResult.status.Rejected);
```
hewison-chris, line 61: We should also check the reason.

hewison-chris (changes requested):
Nice, just a few small requests.

source/agora/consensus/data/Transaction.d
linked0: Good suggestion! Will apply it.

source/agora/test/PostTransactionResult.d
linked0: I was thinking about checking the reason also but that could break the test if the content of the reason changes even slightly. Let me think of a better way.

hewison-chris: It does not need to check the exact string but that it contains something relevant. If someone makes a breaking change to the message then the test will fail and it can be updated. It is better to have a test that can fail with a code change than no test at all.

source/agora/api/FullNode.d
```diff
@@ -289,9 +289,12 @@ public interface API
         Params:
             tx = a Transaction being posted
 
+        Returns:
+            an TransactionResult object, the result on posting the transaction
```
hewison-chris:
```suggestion
            a TransactionResult object, the result of posting the transaction
```

source/agora/test/PostTransactionResult.d
linked0: It's a simple way to check the reason. I applied it. Thanks.

source/agora/test/PostTransactionResult.d
```diff
@@ -0,0 +1,63 @@
...
+    result = nodes[0].postTransaction(tx);
+    assert(result == TransactionResult(TransactionResult.Status.Duplicated));
+
+    // post a transaction with no fee, expect `Rejected`
+    tx = TxBuilder(input_txs[0], 1).feeRate(Amount(0))
+        .draw(Amount(100_000), [output_addr])
+        .sign();
+    result = nodes[0].postTransaction(tx);
+    assert(result.status == TransactionResult.status.Rejected);
+    assert(!result.reason.find("Fee rate is less than minimum").empty);
```
hewison-chris, line 62: 👍

hewison-chris (approved):
LGTM

linked0:
> Return value is never used, one case is [here](https://github.com/bosagora/agora/blob/982431b12cc9d7303b1ff733b1f5fd042b71bab0/source/agora/cli/client/SendTxProcess.d#L204)

@mkykadir Modified based on your comment. If there's something more needed about that, please let me know. Thanks.

mkykadir (approved):

PR #2725, opened 2021-12-06
Validator: Push BlockHeader on new signatures received

omerfirmak:
This probably should not be a validator only thing, but since it uses Nominator, it will have to do for now.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -690,27 +690,31 @@ extern(D):
         Params:
             block_sig = the structure with details of the block signature
 
+        Returns:
+            Updated block header
+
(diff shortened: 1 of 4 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -690,27 +690,31 @@ extern(D):
         Params:
             block_sig = the structure with details of the block signature
 
+        Returns:
+            updated header
```
TrustHenry:
```suggestion
            Updated block header
```

TrustHenry (approved):
LGTM, Thanks

linked0 (approved):
LGTM

hewison-chris (commented):
LGTM

PR #2730, opened 2021-12-07
Extend timer interface

mkykadir:
Include more (useful) interfaces from Vibe.d's timer

- `createTimer`: Creating timer without arming
- `pending`: State of the timer after arming (if waiting to be fired)

With this PR, also timer handlers are converted to `@safe nothrow` delegations. Since we didn't have those attributes previously, most of the handlers are unsafe and can throw exceptions in Agora. I've forced them to be `@safe nothrow` but it became a huge change set (300+ line changes and increasing) which led to still allowing current handler types in a `trusted` attribute and logging exceptions thrown in the handler.

### DIFF
--- a/source/agora/common/Task.d
+++ b/source/agora/common/Task.d
@@ -140,4 +140,15 @@ public interface ITimer
     ***************************************************************************/
 
     void rearm (Duration timeout, bool periodic) nothrow;
+
+    /***************************************************************************
+
+        See_also: https://vibed.org/api/vibe.core.core/Timer.pending
+
+        Returns:
+            True if timer is armed and pending to fire
+
+    ***************************************************************************/
+
+    public bool pending () @safe nothrow;
 }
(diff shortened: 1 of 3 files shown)


source/agora/common/Task.d
```diff
@@ -124,4 +140,15 @@ public interface ITimer
     ***************************************************************************/
 
     void rearm (Duration timeout, bool periodic) nothrow;
```
mkykadir: All interfaces should be `@safe` and `nothrow`

omerfirmak (approved):
I believe there should be a corresponding PR to LocalRest.

mkykadir:
> I believe there should be a corresponding PR to LocalRest.

Yes, here [it is](https://github.com/Geod24/localrest/pull/130)

omerfirmak (approved):

PR #2734, opened 2021-12-09
Change toString sinks to use 'in' + support extra loglevel

Geod24:
```
Now that the bug we worked around (prevent in versions up to v2.096.0) has been fixed,
we can switch back to using 'in' for sink delegates.
As we are updating Ocean, two more log levels have been added,
and need to be handled in the final switch for the code to compile.
Hence, support for it was added at the same time.
```


mkykadir (approved):

Geod24:
There's no good way to fix `Integration`, so I'll just force merge and will update Faucet right after when this is approved.

PR #2736, opened 2021-12-14
Implement Chain ID

### LINKED ISSUE #2728: Prevent alien attacks
According to the audit report:

> There is no distinction between the protocols of different networks(eg. Testnet/Mainnet), which may lead to the pollution of the peer address pool.

Followed by:

> Regarding 4.2.3 "Alien Attack" audit,
> This vulnerability may seriously reduce the performance of the p2p network, if the network is large, a single node may receive requests from a large number of other network nodes, and never stop, we think it is necessary to add chainid/magic value so that the protocol can distinguish different networks.

omerfirmak:
Chain ID is hashed together with all user data which makes
signatures for different networks to be unique.

Initial idea was to include genesis hash instead of a magic number
but that creates a chicken&egg problem since genesis hash of different
network with exact same genesis block should be different as well.

Fixes #2728


source/agora/consensus/PreImage.d
```diff
@@ -538,41 +523,43 @@ public void getCycleSeed (in KeyPair pair, in uint validator_cycle, out Hash see
...
+        seed = Hash(`0x3256f762cee8e5345361d3db3d500ce7fb1c9591a0f051a264e383f764dc8c41033ce5f094917f58311c8597a450784c92bdfae947ca9c39a4ce62373a8631b4`);
     else if (pair == K)
-        seed = Hash(`0x72d84ce0dfc74f856909a3f49dffdb8b68c39212f1229fddec3fb071c3bff84983645f8caab335189a710502c1020ec4b855f70f8b1493718848aa9dcc761fa7`);
+        seed = Hash(`0xf32b9715a5ba6c8c26ac9dd92179c9e549cdbbe18066c02db7dc6fff62e83b8c3ac4052265b1320af39b93f3a70e2f9b5f860c39d1f2b27a7e66d3554628bcec`);
     else
     {
         auto cycle = PreImageCycle(pair.secret, 20, PreImageCycle.PreImageCount, Height(0));
         seed = cycle[seed_height];
-        throw new Exception(format!"There's no seed for %s. The Hash of (Hash(`%s`) should be added."
+        writeln(format!"There's no seed for %s. The Hash of (Hash(`%s`) should be added."
```
linked0, line 562: I guess the original code is more suitable for our tests because we should add the missing seed for the new key after the exception.

linked0 (commented):
Broadly LGTM and thanks for refactoring the `makeEnrollment` and removing some weird cycle seeds. I think the first commit could be split into two or three commits.

source/agora/consensus/PreImage.d
omerfirmak: I think there is no reason to interrupt the test just for missing seeds, logging the seed so that they can be updated when convenient is a better approach imo.

omerfirmak:
> Broadly LGTM and thanks for refactoring the `makeEnrollment` and removing some weird cycle seeds. I think the first commit could be split into two or three commits.

Splitted into two

hewison-chris (approved):
LGTM

linked0 (approved):
LGTM

PR #2737, opened 2021-12-14
Actions: checkout faucet and agora recursively

omerfirmak:
Required since https://github.com/bosagora/faucet/pull/155

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -15,7 +15,7 @@ jobs:
         path: agora
         # For Codecov
         fetch-depth: 2
-        submodules: true
+        submodules: recursive
 
     - name: Prepare compiler
       uses: dlang-community/setup-dlang@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/integration.yml
```diff
@@ -32,7 +32,7 @@ jobs:
       with:
         path: faucet
         repository: bosagora/faucet
-        submodules: true
+        submodules: recursive
 
     - name: Update Faucet submodule to point to our clone
```
Geod24, line 37: I think you need to run `git submodule update --init` before the "For debug output", otherwise the problem will still show.

omerfirmak: https://github.com/bosagora/agora/runs/4520765196?check_suite_focus=true it builds fine

Geod24: Now try updating a submodules with a breaking change

omerfirmak: `git submodule update --init`

Added that in as well

omerfirmak: > Now try updating a submodules with a breaking change

🤣 https://github.com/bosagora/agora/runs/4520736505?check_suite_focus=true

Geod24 (approved):

PR #2738, opened 2021-12-14
Actions: Properly update agora submodules in faucet

omerfirmak:
https://github.com/bosagora/agora/pull/2737 was merged before I could update, this is proven to work in https://github.com/bosagora/agora/runs/4521427363?check_suite_focus=true


hewison-chris (approved):
LGTM

PR #2741, opened 2021-12-15
Trivial: Fix quotes around date in CI config

Geod24:
Editors don't like the look of this and colorize the whole file.


mkykadir (approved):

hewison-chris (approved):
LGTM

PR #2756, opened 2021-12-20
Fix LocalRest commit hash

omerfirmak:
We messed it up somehow.


mkykadir:
The reason was I missed to force push changes to the PR on LocalRest side, which caused unknown commit hash on agora side.

mkykadir (approved):

PR #2759, opened 2021-12-22
Trivial updates

mkykadir:
See each commit for details

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -1401,27 +1401,6 @@ public class Ledger
             &this.getPenaltyDeposit, utxo_finder);
     }
 
-    /***************************************************************************
-
-        Add an enrollment data to the enrollment pool
-
-        Params:
-            enroll = the enrollment data to add
-            pubkey = the public key of the enrollment
-            height = block height for enrollment
-
-        Returns:
-            true if the enrollment data has been added to the enrollment pool
-
-    ***************************************************************************/
-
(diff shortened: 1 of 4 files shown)


Geod24:
Can you add the motivation for the second commit ? I'm wondering whether `addEnrollment` belongs to the basic `Ledger`, as a mirror to `addTransaction`. Another approach would be to have a basic `Ledger` without any pool (if that makes sense), and a derived class with pools (given that for a client tracking the chain, a pool might not make sense).

mkykadir:
`addEnrollment` of Ledger is only used through `TestLedger` in tests. Source code uses enrollment manager only.

linked0 (approved):
LGTM

omerfirmak:
> `addEnrollment` of Ledger is only used through `TestLedger` in tests. Source code uses enrollment manager only.

`TestLedger` already exposes enrollment manager thru the `enrollment_manager` getter. We might as well remove addEnrollment entirely.

source/agora/consensus/Ledger.d
```diff
@@ -1401,27 +1401,6 @@ public class Ledger
...
-            enroll = the enrollment data to add
-            pubkey = the public key of the enrollment
-            height = block height for enrollment
-
-        Returns:
-            true if the enrollment data has been added to the enrollment pool
-
-    ***************************************************************************/
-
-    public bool addEnrollment (in Enrollment enroll, in PublicKey pubkey,
```
omerfirmak, line 1418: Commit message should be updated as well, feel free to self merge after that.

omerfirmak (approved):

PR #2762, opened 2021-12-24
Flash: Remove 0 fee special casing in Flash tests 

omerfirmak:


source/agora/test/Flash.d
```diff
@@ -593,19 +593,17 @@ private class FlashListener : TestFlashListenerAPI
...
+private TestConf flashTestConf () @safe
 {
     import agora.node.Config;
 
     TestConf conf;
     conf.consensus.quorum_threshold = 100;
-    // TODO: remove this line when fees are handled
-    conf.consensus.min_fee = Amount(0);
-    conf.node.min_fee_pct = 0;
+    conf.consensus.min_fee = Amount(1);
```
Geod24: Much nicer, but it still looks like a hack though. Lowering the minimum fee is not something we should be doing if we can avoid it. Can we avoid it ?

source/agora/test/Flash.d
```diff
@@ -421,6 +421,7 @@ public class FlashNodeFactory : TestAPIManager
 
     public override void printLogs (string file = __FILE__, int line = __LINE__)
     {
+        super.printLogs();
```
Geod24, line 427: We're going to have 2 "START OF LOGS" which will be confusing.

linked0:
Thanks for addressing this issue. I guess the last thing we should handle is to add fees in closing `tx.

omerfirmak:
Updated per comments

omerfirmak:
> Thanks for addressing this issue. I guess the last thing we should handle is to add fees in closing `tx.

Closing tx already has fees tho?

linked0:
> > Thanks for addressing this issue. I guess the last thing we should handle is to add fees in closing `tx.
> 
> Closing tx already has fees tho?

Right, it has. But if we set a kind of big fee rate, we get an error on waiting for the `closed` state on unit tests. You can see what I mean on [my commit for testing](https://github.com/bosagora/agora/pull/2679/commits/0c53323d4fbcca963a3b55e4b95ae7e7ec1645c7).

omerfirmak:
> > > Thanks for addressing this issue. I guess the last thing we should handle is to add fees in closing `tx.
> > 
> > 
> > Closing tx already has fees tho?
> 
> Right, it has. But if we set a kind of big fee rate, we get an error on waiting for the `closed` state on unit tests. You can see what I mean on [my commit for testing](https://github.com/bosagora/agora/pull/2679/commits/4f2ea316e52d075fdf025e5edd9dff9056f9eb0d).

I see, I will address that in this PR as well.

omerfirmak:
@linked0 , see the last commit

linked0:
Looks good. I will approve it with the tests green. Thanks.

omerfirmak:
> Looks good. I will approve it with the tests green. Thanks.

Green

linked0 (approved):

PR #2763, opened 2021-12-27
Implement caching registry

### LINKED ISSUE #2449: Registry: Implement secondary zones

mkykadir:
Depends on #2703 
Fixes #2449 

- [ ] Code cleanup


mkykadir:
#2703 also implements caching

PR #2764, opened 2021-12-27
Enable UnlockNotify for d2sqlite3

omerfirmak:
If our DB is locked by another process, we should gracefully wait
for it to be unlocked instead of erroring right away.

This will help NODE0 in production, DB of which is also accessed by
Grafana.

Homebrew sqlite has the native support disabled so we use the fake
UnlockNotify provided by the D bindings.


omerfirmak:
I will deploy this to Node0 to see if it helps

mkykadir:
[Unlock-notify](https://www.sqlite.org/unlock_notify.html) is only applicable for [shared-cache mode](https://www.sqlite.org/sharedcache.html) applications which is only available in single process (multi-threaded) environments. Since Grafana is another process than Agora, this might not solve the problem. Another alternative is [busy handler](https://www.sqlite.org/c3ref/busy_handler.html).

omerfirmak:
Any solution sqlite will provide will block our event loop, so I am working on a custom solution.

PR #2767, opened 2021-12-28
StateDB: Write operations should always fail hard

omerfirmak:
If any of them fails while adding a new block to the Ledger,
we should be aware and rollback.


omerfirmak:
This may not be an exhaustive list, I only followed the code inside the begin/rollback scope in `Ledger.addValidatedBlock`

mkykadir (approved):

PR #2768, opened 2021-12-29
Tidy up SOA configuration fields into its own structure

mkykadir:
Secondary and caching registries will not have a SOA configuration, tidying up the fields will help us to easily configure and validate for those types.

Part of #2449

### DIFF
--- a/devel/config-single.yaml
+++ b/devel/config-single.yaml
@@ -43,11 +43,13 @@ registry:
   validators:
     authoritative: true
     primary: localhost
-    email: no@no.no
+    soa:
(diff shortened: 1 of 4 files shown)


Geod24 (changes requested):
Since the second commit depends on the first commit, I think they can be merged.
Splitting the commits here doesn't make much sense, as they are not atomic: After the first commit is merged, the `devel` config file is not usable.
Additionally, the example config (in `doc`) needs to be updated, in the same commit.

Geod24 (approved):

Issue #2686, opened 2021-11-19
Flash nodes just rely on registry for peer authentication


mkykadir:
Validator nodes authenticate themselves also during `handshake`. Flash nodes should not just rely on registry since there are planned changes on registry architecture (disabling `GET` service from REST and use of only DNS). Payload authentication through DNS requires DNSSEC implementation that requires chain of trust which may result in more expensive operations for peer authentication.

PR #2769, opened 2021-12-29
Remove peer validation through Registry

mkykadir:
Current way of peer validation uses both registry and handshake for Validator nodes and just rely on registry (#2686) for Flash nodes. Relying on registry will not be achievable when we move to DNS from current GET interface and DNS has its own DNSSEC method for that. Registry itself will still validate registering nodes by signature.

Part of #2449

### DIFF
--- a/source/agora/flash/Node.d
+++ b/source/agora/flash/Node.d
@@ -602,12 +602,6 @@ public class FlashNode : FlashControlAPI
                 return null;
             }
 
-            if (!payload.verifySignature(peer_pk))
-            {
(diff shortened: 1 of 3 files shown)


Geod24 (approved):
LGTM. However, do we discard the client if the handshake fails ?

mkykadir:
[Yes](https://github.com/bosagora/agora/blob/6011caaee41a93ea5c5e361f080894b4faaac2aa/source/agora/network/Manager.d#L200), address gets banned and client is discarded when handshake fails

PR #2770, opened 2021-12-29
Increase DNS compatibility of Registry storage

mkykadir:
Secondary and caching zones will use DNS interface for registry communication. `TypedPayload` is Registry's data storage type and compatible with REST payload, this change set adds DNS payload compatibility and TTL (Time-to-Live) support for the data storage.

Part of #2449

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -528,7 +528,7 @@ private SOA fromConfig (in ZoneConfig zone, Domain name, uint serial) @safe pure
     return soa;
 }
 
-/// Associate a `RegistryPayload` with internal data
+/// Internal registry data
 private struct TypedPayload
 {
     /// The DNS RR TYPE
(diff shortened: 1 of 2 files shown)


source/agora/node/Registry.d
```diff
@@ -540,6 +540,70 @@ private struct TypedPayload
...
+        Construct an instance of an `TypedPayload` from REST payload
+
+        Params:
+            type = Resource record type
+            payload = Payload data of API
+            utxo = UTXO of the validator (optional)
+
+    ***************************************************************************/
+
+    public this (TYPE type, RegistryPayload payload, Hash utxo = Hash.init) @safe
```
omerfirmak: This is not needed as compiler already generates this by default.

Geod24: Missing other attributes.
Another approach would be to rename the other ctor into `make` (a `static` function) so that you keep the default ctor. I wish we had something like:
```D
public this () = default;
```

as C++ does :(

source/agora/node/Registry.d
```diff
@@ -540,6 +540,70 @@ private struct TypedPayload
...
+        (Resource record)
+
+        Params:
+            rr = DNS Resource record
+            pubKeyParser = Delegate for parsing public key from RR domain
+
+    ***************************************************************************/
+
+    public this (ResourceRecord rr,
+        PublicKey delegate(in char[]) const scope @safe pubKeyParser) @trusted
```
Geod24: What in that function needs to be trusted ?

source/agora/node/Registry.d
mkykadir: Since I've added a ctor (from RR) compiler stops generating default

omerfirmak: Ah yes, my bad

source/agora/node/Registry.d
```diff
@@ -540,6 +540,53 @@ private struct TypedPayload
...
+            reg_payload.data.addresses ~= Address("http://" ~
+                                            cast(string) address.value);
+        }
+        else if (rr.type == TYPE.A)
+        {
+            import std.socket : InternetAddress;
+            auto addresses = rr.rdata.a;
+            reg_payload.data.addresses = addresses.map!(
+                (addr) {
+                    auto in_addr = new InternetAddress(addr,
```
omerfirmak: `scope in_addr` to avoid allocation?

omerfirmak (approved):

PR #2771, opened 2021-12-29
Tracy: Demangle D symbols

Geod24:
Still needs to be tested.

https://github.com/wolfpld/tracy/issues/279

### DIFF
--- a/source/agora/utils/TracyAPI.d
+++ b/source/agora/utils/TracyAPI.d
@@ -19,6 +19,49 @@ module agora.utils.TracyAPI;
 extern(C):
 @system:
 nothrow:
+
+version (Tracy_NoDemangle) {}
+else
+{
+    /***************************************************************************
+
+        Demangles a symbol
+
+        Support demangling either a D or a C++ symbol.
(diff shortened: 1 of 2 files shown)


source/agora/utils/TracyAPI.d
```diff
@@ -19,6 +19,49 @@ module agora.utils.TracyAPI;
...
+            return result.length > 0;
+        }
+        int status;
+        __cxa_demangle(mangled, out_, &len, &status);
+        return status == 0;
+    }
+}
+
+/// https://itanium-cxx-abi.github.io/cxx-abi/abi.html#demangler
+private char* __cxa_demangle (const(char)* mangled_name, char* buf,
```
omerfirmak, line 62: This requires the `buf` to be mallocated. If I am not mistaken, Tracy allocates it on stack.

mkykadir (approved):
D's demangling is another type of mangling, not so clear :woozy_face: 
Also we've something like this, that I also couldn't demangle [from here](https://dlang.org/spec/abi.html#name_mangling)
image

Geod24:
Oh that's https://github.com/ldc-developers/ldc/issues/3669

Issue #2772, opened 2021-12-29
Ledger should not accept any block if previous height was not signed by majority of validators


omerfirmak:
We do check if signatures of previous height have reached majority while validating nominated values, but same should be applied to the blocks we catch up on.

Additionally if the last block we have in storage was not signed by majority (which we allow thru externalization), we should still try to catch up on that height and if we get a block that was signed by majority, we should be able to replace the block in our storage.

A malicious node could sign multiple blocks at the same height, we should come up with a scheme that discourages or handles that scenario as well.

Geod24:
> we should be able to replace the block in our storage.

I assume that part is not done yet ? Reverting changes made to the validator set and UTXO set is going to be very hard, isn't it ?

omerfirmak:
Yeah that part is missing, I am hoping sqlite to have a snapshot-like feature so that we can easily revert all the changes. If not we will probably need to keep the an in-memory diff that last block would cause on the state and commit that to DB only when next block is externalized, which means the previous block reached majority sigs.

hewison-chris:
After some discussion it has been decided to look into how to only externalize when a majority of block signatures have been collected. Maybe between the `ballot` phases of `prepare` and `commit` the signatures can be gossiped and the `commit` only `emitted` once the majority is reached. Then when the `externalize` happens the nodes can externalize with a block which has a majority.

hewison-chris:
The solution implemented was to sign the block when the SCP `Externalize` single event occurs. The next nomination round then waits until the `pending block` has a majority of signatures and is written to the ledger (externalized).

PR #2597, opened 2021-10-22
Introduce Freezing fee

omerfirmak:
Freezing TXs now have to include SlashAmountPenatly amount of fee
for each Freeze output they have. This amount will be refunded to them
if they are not slashed for the duration of their validation cycle.

Part of #2554

PR #2774, opened 2021-12-30
Ledger: hasMajoritySignatures

omerfirmak:
There are multiple points that we check if signatures have reach
majority, it is better to have a single implementation.

Towards #2772

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -841,12 +841,10 @@ public class Ledger
                 return "Block: Number of preimages does not match active validators";
             foreach (idx, const ref hash; block.header.preimages)
             {
-                if (hash is Hash.init)
-                {
-                    // TODO: Check that the block contains a slashing transaction
-                }
+                if (hash is Hash.init) // Slashed
+                    continue;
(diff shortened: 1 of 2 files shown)


source/agora/consensus/Ledger.d
```diff
@@ -1019,6 +1017,49 @@ public class Ledger
...
+            return false;
+        else if (height == this.last_block.header.height) // most common case
+            return this.hasMajoritySignature(this.last_block.header);
+
+        Block block;
+        try
+            block = this.storage.readBlock(height);
+        catch (Exception exc)
+            return false;
+        return this.hasMajoritySignature(block.header);
```
mkykadir: I've searched where else `block` is set? What about
```
try
    return this.hasMajoritySignature(this.storage.readBlock(height));
```

mkykadir (approved):
LGTM except commit messages;

1. I don't know if its adding or removing :smile: 
2. Instead of `old`, `invalid` with reference to PR --> #2597 
3. :heavy_check_mark: 
4. Don't have any idea about what to change but, needs a new title

PR #2781, opened 2022-01-03
DNS: Add URI RR support

mkykadir:
URI RR support is added according to [RFC7553](https://datatracker.ietf.org/doc/html/rfc7553)

This change will add `_` character support to `Domain` labels, even though it is not supported by Domain RFC; DNS uses that to identify non-domain, i.e. service, labels.

Towards #2320

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -1035,12 +1035,13 @@ public struct Domain
             }();
             buffer[count + label.length] = '.';
 
-            // Only [a-z], [A-Z], [0-9], or "-" (dash)
+            // Only [a-z], [A-Z], [0-9], or "-" (dash), or "_" (underscore)
+            // "_" is to support service labels
             // https://datatracker.ietf.org/doc/html/rfc1034#section-3.5
             foreach (idx, char c; buffer[count .. count + label.length])
             {
                 ensure((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') ||
-                    (c >= '0' && c <= '9') || c == '-',
+                    (c >= '0' && c <= '9') || c == '-' || (c == '_' && idx == 0),
                     "Invalid char '{}' in domain name at index {}", c, count + idx);
             }
             count += label.length + 1;
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -20,6 +20,7 @@
 module agora.common.DNS;
 
 import agora.common.Ensure;
+import agora.common.Types : Address;
 import agora.serialization.Serializer;
 
 import std.algorithm.iteration;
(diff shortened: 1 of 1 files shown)


source/agora/common/DNS.d
```diff
@@ -703,6 +716,12 @@ public struct ResourceRecord
             this.soa = val;
         }
 
+        /// Ditto
+        public this (URIRDATA val) @safe
```
Geod24: `inout` / attributes, see `SOA`

source/agora/common/DNS.d
```diff
@@ -616,6 +617,15 @@ public struct ResourceRecord
         return ResourceRecord(name, TYPE.A, CLASS.IN, ttl, RDATA(ipv4));
     }
 
+    /// Make a record of URI type
+    public static ResourceRecord make (TYPE type : TYPE.URI) (
+        Domain name, uint ttl, Address uri) @safe
+    {
+        return ResourceRecord(name, TYPE.URI, CLASS.IN, ttl, RDATA(URIRDATA(
+            0, 0, uri
```
Geod24: Those two parameters should be parameters to `make` (with sensible default values).

source/agora/common/DNS.d
```diff
@@ -823,17 +848,28 @@ unittest
     ips[0] = InternetAddress.parse(test_ip);
 
     ResourceRecord a_rr = ResourceRecord.make!(TYPE.A)(Domain("localhost"), 0, ips);
+
+    auto uri_target = Address("agora://test.local:1234");
+    ResourceRecord uri_rr = ResourceRecord.make!(TYPE.URI)(
+        Domain("_agora._tcp"), 0, uri_target
```
Geod24, line 854: Sorry for not catching that earlier, but why are we using `Domain` here actually ?

mkykadir: We don't have to but, RFC states that domains are **extended** in a way that it can accept service labels. We can really extend `Domain` with a, maybe, `Service` type but only difference would be allowing `_` as first character.

source/agora/common/DNS.d
```diff
@@ -616,6 +617,16 @@ public struct ResourceRecord
         return ResourceRecord(name, TYPE.A, CLASS.IN, ttl, RDATA(ipv4));
     }
 
+    /// Make a record of URI type
+    public static ResourceRecord make (TYPE type : TYPE.URI) (
+        Domain name, uint ttl, Address uri, ushort prio = 0,
+        ushort weight = ushort.max) @safe
```
Geod24: Why `max` ? `0` sounds like a better default, as not caring about the weight probably means that the entry is more important than one that sets it.

Geod24 (approved):
LGTM. Just change the weight and merge at will.

source/agora/common/DNS.d
mkykadir: Value towards `0` for `priority` means higher priority but value towards `max` higher priority for `weight`

Geod24: So default-created records will be at highest priority ?

PR #2782, opened 2022-01-03
Report address of the object to be freed to Tracy

### LINKED ISSUE #2778: Tracy improper instrumentation while memory tracing
Tracy throws `improper instrumentation` error after short time. Reason is reporting of previously allocated but not freed address.

mkykadir:
Garbage sweeping works per-page basis but allocation is object based. Thus, object addresses is reported to Tracy and freeing also needs to report object address. Previously it was reporting page addresses to Tracy during sweeping. This is the reason why it was passing with `COLLECT_PRINTF` debug flag.

Fixes #2778

### DIFF
--- a/source/agora/utils/gc/GC.d
+++ b/source/agora/utils/gc/GC.d
@@ -4206,6 +4206,12 @@ debug (LOGGING)
         private void log_malloc(void *p, size_t size) nothrow
         {
             //debug(PRINTF) printf("+log_malloc(p = %p, size = %zd)\n", p, size);
+            if (OPFAIL != current.find(p))
+            {
+                debug(PRINTF) printf("Memory allocation without free'ing at %p\n", p);
+                return;
+            }
+
             Log log;
 
             log.p = p;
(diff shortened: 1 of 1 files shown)


source/agora/utils/gc/GC.d
```diff
@@ -2483,12 +2484,13 @@ struct Gcx
                                     assert(core.bitop.bt(toFree.ptr, i));
 
                                     debug(COLLECT_PRINTF) printf("\tcollecting %p\n", p);
-                                    leakDetector.log_free(sentinel_add(p));
```
Geod24: That means that `leakDetector` doesn't quite work right ?

mkykadir: Leak detector is not checking for double allocations, I’ve found the problem by updating `log_malloc`. I ~can~ also send that with the PR.

Geod24:
Wouldn't it be better to just set `hasFinalizer` to `true`, always ? The idea is to minimize the diff with upstream, which is [here](https://github.com/dlang/druntime/tree/master/src/core/internal/gc).

Geod24 (approved):

Geod24:
Interestingly I just found this: https://github.com/dlang/druntime/commit/6558d800f8b4db57bb8954511b7c5987590a0480

PR #2787, opened 2022-01-04
Registry: Compare record domains in case-insensitive manner

### LINKED ISSUE #2786: DNS Domains should be case insensitive
But records should be stored as it is;

`findZone`, `ZoneData.get` with its SQL query should be updated and `PublicKey.fromString` should be checked.

mkykadir:
https://datatracker.ietf.org/doc/html/rfc1035#section-2.3.3

Fixes #2786 

Bech32 changeset is required to prevent disrupting secondary registry implementation (i.e. zone transferring)

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -920,6 +920,23 @@ public struct Domain
             this.value = cast(string) (v ~ '.');
     }
 
+    size_t toHash () const @safe pure nothrow
+    {
+        import std.ascii : toLower;
+        char[255] buffer;
+
+        foreach (ix, char c; this.value)
+            buffer[ix] = toLower(c);
+
+        return buffer.hashOf;
+    }
+
+    bool opEquals (in Domain other) const nothrow @safe
+    {
+        import std.uni : sicmp;
+        return (sicmp(other.value, value) == 0);
+    }
+
     /// Returns: A forward range allowing to iterate by label
     /// Iteration is done in reverse order from the string ordering,
     /// meaning for `bosagora.io.`, the initialized state is root (empty string)
--- a/source/agora/crypto/Bech32.d
+++ b/source/agora/crypto/Bech32.d
@@ -129,17 +129,12 @@ public DecodeResult decodeBech32 (in char[] str)
 {
     import std.conv;
 
-    bool lower = false, upper = false;
     for (size_t i = 0; i < str.length; ++i)
     {
         ubyte c = str[i];
-        if (c >= 'a' && c <= 'z') lower = true;
-        else if (c >= 'A' && c <= 'Z') upper = true;
-        else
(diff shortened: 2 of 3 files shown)


source/agora/node/Registry.d
```diff
@@ -328,6 +328,9 @@ public class NameRegistry: NameRegistryAPI
         reply.header.RA = false; // TODO: Implement
         reply.header.AA = true;  // TODO: Make configurable
 
+        // Domain names should be case insensitive
+        // See
```
omerfirmak: see what?

mkykadir: you saw

source/agora/node/Registry.d
```diff
@@ -392,7 +395,8 @@ public class NameRegistry: NameRegistryAPI
                 break;
             }
 
-            auto answer = this.findZone(q.qname);
+            auto lowercase = Domain(q.qname.value.toLower);
+            auto answer = this.findZone(lowercase);
```
Geod24: Any chance we can use `sicmp` to avoid the allocation ?

source/agora/node/Registry.d
```diff
@@ -434,9 +434,12 @@ public class NameRegistry: NameRegistryAPI
 
     auto findZone (Domain name, bool matches = true) @safe
     {
-        auto matching_zone = name in this.zones;
-        if (matching_zone)
-            return matches ? &matching_zone.answer_matches : &matching_zone.answer_owns;
+        foreach (zone; this.zones.byKeyValue())
+        {
+            if (0 == std.uni.sicmp(zone.value.root.value, name.value))
```
Geod24: If you implement `opEqual` you don't need to change the lookup.

source/agora/common/DNS.d
```diff
@@ -920,6 +920,23 @@ public struct Domain
...
 
+    size_t toHash () const @safe pure nothrow
+    {
+        import std.ascii : toLower;
+        char[255] buffer;
+
+        foreach (ix, char c; this.value)
+            buffer[ix] = toLower(c);
+
+        return buffer.hashOf;
```
Geod24, line 931:
```D
buffer[0 .. this.value.length] = this.value[];
return buffer[0 .. this.value.length].hashOf();
```

source/agora/common/DNS.d
```diff
@@ -920,6 +920,23 @@ public struct Domain
...
+    }
+
+    bool opEquals (in Domain other) const nothrow @safe
+    {
+        import std.uni : sicmp;
+        return (0 == sicmp(other.value, value));
```
Geod24: [The force in you, I see](https://en.wikipedia.org/wiki/Yoda_conditions).

Geod24 (approved):

PR #2788, opened 2022-01-04
DNS: More validation in Domain

Geod24:
```
Properly validate the domain name when creating an instance.
There are now two ways to create a Domain:
- `fromSafeString`: Will validate its argument and `Domain` will be
  a wrapper around it.
- `fromString`: Does what the constructor was previously doing,
  plus some extra validation. Will always duplicate the input.

The constructor has been disabled has it was too error prone.
```

@mkykadir : I am not too happy about how this looks, so suggestions are welcome.
The main idea is that we should have a way to instantiate a Domain from a pre-validated string without allocating.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -740,11 +740,12 @@ public struct ResourceRecord
             deserializeFull!(TYPE)(&ctx.read, ctx.options),
             deserializeFull!(CLASS)(&ctx.read, ctx.options),
             deserializeFull!(uint)(&ctx.read, ctx.options),
+            T.RDATA(null),
         );
 
         auto rdlength = deserializeFull!(ushort)(&ctx.read, ctx.options);
 
-        ResourceRecord.RDATA tmp_data;
+        ResourceRecord.RDATA tmp_data = null;
         () @trusted
         {
             switch (tmp.type)
@@ -894,7 +895,8 @@ public struct OPTRR
 
     ***************************************************************************/
 
-    public ResourceRecord record = ResourceRecord(Domain.init, TYPE.OPT, cast(CLASS) 4096);
+    public ResourceRecord record = ResourceRecord(
+        Domain.init, TYPE.OPT, cast(CLASS) 4096, 0, ResourceRecord.RDATA(null));
 
     /// Requestor's UDP payload size
     public ushort payloadSize () scope const
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -813,13 +813,13 @@ unittest
 {
     checkFromBinary!Message();
 
-    auto root = Domain(".");
+    auto root = Domain.fromSafeString(".");
     assert(root.serializeFull() == [0]);
 
-    auto dlang = Domain("dlang.org.");
+    auto dlang = Domain.fromSafeString("dlang.org.");
     assert(dlang.serializeFull() == [ubyte(5), 'd', 'l', 'a', 'n', 'g', 3, 'o', 'r', 'g', 0 ]);
 
-    auto dlang2 = Domain("dlang.org"); // No trailing dot
+    auto dlang2 = Domain.fromString("dlang.org"); // No trailing dot
     assert(dlang.serializeFull() == dlang2.serializeFull());
 
     auto lroot = root.byLabel();
@@ -848,11 +848,12 @@ unittest
     auto ips = new uint[](1);
     ips[0] = InternetAddress.parse(test_ip);
 
-    ResourceRecord a_rr = ResourceRecord.make!(TYPE.A)(Domain("localhost"), 0, ips);
+    ResourceRecord a_rr = ResourceRecord.make!(TYPE.A)(
+        Domain.fromSafeString("localhost."), 0, ips);
 
     auto uri_target = Address("agora://test.local:1234");
     ResourceRecord uri_rr = ResourceRecord.make!(TYPE.URI)(
-        Domain("_agora._tcp"), 0, uri_target
+        Domain.fromSafeString("_agora._tcp."), 0, uri_target
     );
     Message msg;
     msg.answers ~= a_rr;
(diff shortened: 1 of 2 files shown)


mkykadir (approved):

source/agora/common/DNS.d
```diff
@@ -720,11 +720,12 @@ public struct ResourceRecord
             deserializeFull!(TYPE)(&ctx.read, ctx.options),
             deserializeFull!(CLASS)(&ctx.read, ctx.options),
             deserializeFull!(uint)(&ctx.read, ctx.options),
+            T.RDATA(null),
```
mkykadir, line 743: Separated commit, maybe

source/agora/common/DNS.d
```diff
@@ -911,13 +913,90 @@ public struct Domain
     /// The domain name
     public const(char)[] value = ".";
 
-    /// Creates domain name with always trailing root
-    public this (scope inout const(char)[] v) @trusted inout pure nothrow
+    @disable this () @safe pure nothrow @nogc;
+    @disable this (inout(char)[]) inout @safe pure nothrow @nogc;
```
mkykadir, line 954: Default `ctor` would be enough

source/agora/common/DNS.d
```diff
@@ -911,13 +913,90 @@ public struct Domain
...
+        {
+            // We can't use `splitter` because it's not `nothrow @nogc`
+            if (c != '.') continue;
+            // This assert is also present in `checkLabel` but the error message
+            // is less precise.
+            const chunk = v[lastLabelIdx .. idx];
+            assert(chunk.length < 64,
+                   "Domain instantiated with a label longer than 64 characters");
+            auto errIdx = Domain.checkLabel(chunk);
+            if (errIdx < chunk.length)
```
mkykadir: The way it is checked can look similar to version in `fromString`

source/agora/common/DNS.d
```diff
@@ -911,13 +913,90 @@ public struct Domain
...
+        // This is a copy for `fromSafeString`, but uses exceptions
+        size_t lastLabelIdx = 0;
+        foreach (idx, char c; str)
+        {
+            if (c != '.') continue;
+            const chunk = str[lastLabelIdx .. idx];
+            ensure(chunk.length < 64,
+                   "Domain instantiated with a label longer than 64 characters");
+            auto errIdx = Domain.checkLabel(chunk);
+            ensure(errIdx >= chunk.length,
```
mkykadir: Why `>=`? L#1149 checks with `==`

source/agora/node/Registry.d
```diff
@@ -851,7 +854,7 @@ private struct ZoneData
 
         foreach (const ref payload; this)
         {
-            reply.answers ~= payload.toRR(Domain(format("%s.%s",
+            reply.answers ~= payload.toRR(Domain.fromString(format("%s.%s",
```
mkykadir, line 899: This can also be `fromSafeString` for consistency with L#619.
PublicKey validates itself and `this.root` is already a `Domain`

mkykadir (approved):

source/agora/common/DNS.d
Geod24: `@disable`ing the default ctor of `Domain` means that every struct that contains a `Domain` is not default-constructible anymore. But sure, will split it out in a prior commit.

source/agora/common/DNS.d
Geod24: I will double check, but I think it wasn't, or that it created nonsensical errors.

source/agora/common/DNS.d
Geod24: It shouldn't make a diff, but will change.

source/agora/node/Registry.d
```diff
@@ -479,8 +479,8 @@ unittest
 
     const AStr = WK.Keys.A.address.toString();
     // We only need a valid mname
-    ZoneData zone;
-    zone.root = Domain("net.bosagora.io");
+    ZoneData zone = void; // FIXME
```
Geod24: Oops

source/agora/common/DNS.d
Geod24: Now I remember. It will give you an error like this:
```
onlineapp.d(7): Error: constructor `onlineapp.Container.this()` is not callable using argument types `(typeof(null))`
onlineapp.d(7):        expected 0 argument(s), not 1
```

Instead of:
```
onlineapp.d(8): Error: constructor `onlineapp.Container.this` cannot be used because it is annotated with `@disable`
```

source/agora/node/Registry.d
Geod24: Added a commit to fix this.

source/agora/node/Registry.d
```diff
@@ -472,45 +472,89 @@ public class NameRegistry: NameRegistryAPI
...
+    enum NoHRPPublicKeyStringLength = 63 - "boa1".length;
+
+    static PublicKey tryParse (in char[] data)
+    {
+        try return PublicKey.fromString(data);
+        catch (Exception exc) return PublicKey.init;
+    }
+
+    // `findZone` ensures, `domain` is owned by us;
+    // thus this method should NEVER be `public`
```
mkykadir, line 503: This can be removed

source/agora/node/Registry.d
```diff
@@ -472,60 +472,104 @@ public class NameRegistry: NameRegistryAPI
...
```
mkykadir: These should be removed

Geod24: Good eye, fixed!

Geod24:
Rebased

source/agora/node/Registry.d
```diff
@@ -472,60 +472,102 @@ public class NameRegistry: NameRegistryAPI
...
+
+    Params:
+      domain = The full domain name to parse
+
+    Returns:
+      A valid `PublicKey`, or `PublicKey.init`
+
+*******************************************************************************/
+
+private PublicKey extractPublicKey (in char[] domain) @safe
```
mkykadir, line 487: Regarding this change, since #2770 was using this as a function delegate in `TypedPayload.make`, that function can be changed to remove delegate parameter and directly calling this.

Geod24: Do we even call this function ? It would have stopped compiling if we did.

Geod24:
Merging as removing the unused function can be done in another commit.

source/agora/node/Registry.d
mkykadir: Not right now but it will be called, I can do this in separate PR

PR #2790, opened 2022-01-05
Test: DNS Server in test networks

### LINKED ISSUE #2789: DNS server and DNSResolver should be LocalRest-ready
So that we dont duplicate the logic in NetworkClient/NetworkManager

omerfirmak:
Fixes #2789

### DIFF
--- a/source/agora/network/DNSResolver.d
+++ b/source/agora/network/DNSResolver.d
@@ -184,7 +184,7 @@ public abstract class DNSResolver
         catch (std.socket.SocketException) {}
 
         // We need to execute a query, and parse the resulting DNS record
-        auto results = this.query(address.host);
+        auto results = this.query(address.host, QTYPE.A);
         ensure(results.length > 0, "Could not resolve host name '{}' (address: '{}')",
                address.host, address);
 
--- a/source/agora/network/DNSResolver.d
+++ b/source/agora/network/DNSResolver.d
@@ -199,8 +199,8 @@ public abstract class DNSResolver
             {
                 ensure(res.name.value.length > 0,
                        "Found a CNAME to the root domain (empty) while resolving '{}'", address);
-                log.trace("Address '{}': Found CNAME '{}'", address, res.name);
-                address.host = cast(string) res.name.value;
+                log.trace("Address '{}': Found CNAME '{}'", address, res.rdata.name);
+                address.host = cast(string) res.rdata.name.toString();
                 return this.resolve(address);
             }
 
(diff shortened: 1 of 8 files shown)


omerfirmak:
Added a very simple testbed as well, DNS can now be unittested.

Next step is to integrate DNSResolver into Agora.

omerfirmak:
> Next step is to integrate DNSResolver into Agora.

Which will have it's own PR at some point, hopefully soon.

source/agora/network/Manager.d
```diff
@@ -463,6 +464,7 @@ public class NetworkManager
         this.discovery_task = new AddressDiscoveryTask(&this.addAddresses);
         this.clock = clock;
         this.owner_node = owner_node;
+        this.resolver = resolver;
```
mkykadir, line 468: Ok, this is a decision point and may be conflicting with #2703 's caching registry.
I think (we might also rename it but) we should decide tasks of the `DNSResolver`. Currently it asks queries to configured peers and needs an implementation to cache and manage TTLs (which the mentioned PR implements it on `Registry` since we already have a storage there). After the mentioned PR we can have all nodes to have registry with type `caching` enabled by default, thus requesting a `getValidator` will eventually pass through its own registry and manager will not need this.

omerfirmak: Well, we could just add the local caching registry to `resolv.conf` and utilize it that way.
NetworkClient will have to handle expiration of TTL. But I beleive those are not within the scope of this PR

mkykadir: Yeah, #2703 already implements a caching registry with TTL support.
^ With that, then this PR also shouldn't include a resolver in the Manager, might be too early for mentioned reasons.

omerfirmak:
Updated, a FullNode running on my local machine can connect to test network with no problems.

source/agora/network/DNSResolver.d
```diff
@@ -0,0 +1,283 @@
...
+    {
+        this.log = Log.lookup(__MODULE__);
+        this.log.enableConsole();
+    }
+
+    ///
+    public abstract ResourceRecord[] query (const(char)[] name, QTYPE type = QTYPE.ALL) @trusted;
+
+    ///
+    public Message buildQuery (const(char)[] name, QTYPE type = QTYPE.ALL) @safe
```
mkykadir, line 155: #2872 marks it as a `protected`, which should be `public` I believe

omerfirmak:
Commits of this PR was mostly  merged in #2872 and the remaining are most likely will never make in. So closing this one

PR #2792, opened 2022-01-05
Handlers: Use exponential backoff to account for misconfiguration

Geod24:
It's possible that handlers are down for some (perhaps extended)
period of time, which leads to Agora absolutely flooding the log
file with error message, drowning any useful information there might be.
Adding exponential backoff ensures we handle this much more gracefully.


Geod24:
Tested live.

linked0 (approved):
LGTM

PR #2793, opened 2022-01-06
Nominator: Use hasMajoritySignature instead of inlining it

Geod24:
I saw an assert getting triggered in the tests:
Screen Shot 2022-01-06 at 9 33 57

Since this should never happen, the first step was to look at the code and notice the duplication. I'm not sure that it will solve the issue, but it's a step forward.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -477,17 +477,12 @@ extern(D):
             return;
         }
 
-        if (slot_idx > 1) // Genesis block is not signed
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM

linked0 (approved):
LGTM

omerfirmak:
This probably wont fix it, I think this is actually a scenario we overlooked.
Network externalizes a new value (which means other nodes have enough signatures) before one of the nodes received enough signatures.

PR #2794, opened 2022-01-06
FutureEnrollment: Use retryFor to avoid race condition

Geod24:
Screen Shot 2022-01-06 at 11 25 49


hewison-chris (approved):
LGTM

linked0 (approved):
LGTM. It needed the code that accounts for the race condition. Thanks.

Issue #2306, opened 2021-07-23
Unittests sometimes deadlock


Geod24:
I'm seeing unittest getting stuck and just timing out.

Example from recently:
```
Unittesting agora.test.InvalidBlockSigByzantine (took 4 secs, 398 ms, 642 μs, and 2 hnsecs)
Unittesting agora.test.GossipProtocol (took 5 secs, 565 ms, 918 μs, and 1 hnsec)
Unittesting agora.test.GenesisBlock (took 2 secs, 487 ms, 500 μs, and 7 hnsecs)
Error: The operation was canceled.
```

```
Unittesting agora.test.SlashingMisbehavingValidator (took 7 secs, 615 ms, 32 μs, and 4 hnsecs)
Unittesting agora.test.Simple (took 28 secs, 815 ms, 837 μs, and 5 hnsecs)
Unittesting agora.test.Script (took 22 secs, 548 ms, 304 μs, and 8 hnsecs)
Unittesting agora.test.RestoreSlashingInfo (took 22 secs, 275 ms, 644 μs, and 6 hnsecs)
Error: The operation was canceled.
```

I've also seen it happen in `RestoreSCPState` so I don't think it's module-dependent.

Geod24:
This hasn't happened in a while so we probably solved it in one of the CI stabilization PR.

omerfirmak:
https://github.com/bosagora/agora/pull/2416/checks?check_run_id=3440852508

Geod24:
WHY

omerfirmak:
I have seen this happen right after `VariableBlockSize` multiple times

```
Unittesting agora.common.Amount (took 67 μs)
Unittesting agora.client.SendTxProcess (took 673 μs and 6 hnsecs)
Unittesting agora.test.VariableBlockSize (took 3 secs, 712 ms, 376 μs, and 3 hnsecs)
```

Geod24:
> I have seen this happen right after `VariableBlockSize` multiple times

I don't think it's related to `VariableBlockSize` itself, but more to what `VariableBlockSize` is: It's the very first network integration test to be run. And from the log above, it completes, so it must get stuck on the second test. Hence my hypothesis is that either the shutdown sequence is the cause, or we somehow don't properly isolate tests / nodes from one another.

PR #2795, opened 2022-01-06
ValidatorCleanRestart: Raise the timeout from 10 to 30 seconds

Geod24:
This test often errors out, possibly because it runs too slowly.
This is an attempt at finding out if that is the case or if there
is another cause for the frequent failures.

### DIFF
--- a/source/agora/test/ValidatorCleanRestart.d
+++ b/source/agora/test/ValidatorCleanRestart.d
@@ -164,5 +164,5 @@ unittest
 
     // The new block has been inserted to the ledger with the approval
     // of the node_1, although node_2 was shutdown.
-    network.expectHeightAndPreImg(Height(4), network.blocks[0].header);
+    network.expectHeightAndPreImg(Height(4), network.blocks[0].header, 30.seconds);
 }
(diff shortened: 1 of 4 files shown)


source/agora/test/ValidatorCleanRestart.d
```diff
@@ -164,5 +164,5 @@ unittest
 
     // The new block has been inserted to the ledger with the approval
     // of the node_1, although node_2 was shutdown.
-    network.expectHeightAndPreImg(Height(4), network.blocks[0].header);
+    network.expectHeightAndPreImg(Height(4), network.blocks[0].header, 30.seconds);
```
linked0, line 167: How about adding a comment like "TODO: could be restored to 10 seconds".

linked0 (approved):
LGTM with a suggestion.

Geod24:
```
Module tests failed: agora.test.ValidatorCleanRestart
core.exception.AssertError@source/agora/test/ValidatorCleanRestart.d(167): Check condition failed after timeout of 30 secs and 300 attempts: Node 0 has block height 3. Expected: 4
```

Well that's not going to solve anything it seems

linked0:
The failure of the unit test `ValidatorCleanRestart` is also contained to issue #2306, I guess. Right?

Geod24:
This will not work.

PR #2796, opened 2022-01-06
Reduce the pre-image flood

### LINKED ISSUE #2356: Reduce pre-image flood
We currently take a rather aggressive approach to pre-image propagation (read: we just send it, all the time), which leads to a lot of unnecessary traffic. We should remember which nodes received the pre-images and prevent flooding them.

linked0:
We previously took a rather aggressive approach to pre-image
propagation, which leads to a lot of unnecessary traffic. So
We check that the pre-images to be sent are already posted in
order to prevent the too heavy traffic.

Fixes #2356

### DIFF
--- a/source/agora/network/Client.d
+++ b/source/agora/network/Client.d
@@ -156,6 +156,9 @@ public class NetworkClient
     /// Gossip delay
     private enum GossipDelay = 10.msecs;
 
+    /// Heights of last gossiped pre-images for staked UTXOs
+    private Height[Hash] last_preimages;
+
     /***************************************************************************
 
         Constructor.
@@ -227,7 +230,28 @@ public class NetworkClient
             break;
 
         case Preimage:
-            this.attemptRequest!(API.postPreimage, Throw.No)(event.preimage);
+            // Clean up the `last_preimages` periodically
+            // Note: It clears `last_preimages` everyday based on the CoinNet.
+            if (event.preimage.height % Height(144) == 0)
+                this.last_preimages.clear();
+
+            // We do not post the preimage if it is already posted in order to
+            // prevent calls of `postPreimage` from flooding
+            Height last_height;
+            try
+            {
+                last_height = this.last_preimages.get(event.preimage.utxo, Height(0));
+            }
+            catch (Exception e)
+            {
+                log.trace("Exception caught while calling on last_preimages: ", e.msg);
+            }
+            if (event.preimage.height > last_height)
+            {
+                auto height = this.attemptRequest!(API.postPreimage, Throw.No)(event.preimage);
+                if (height >= event.preimage.height)
+                    this.last_preimages[event.preimage.utxo] = event.preimage.height;
+            }
             break;
 
         default:
--- a/source/agora/test/PreImageNoFlood.d
+++ b/source/agora/test/PreImageNoFlood.d
@@ -1,0 +1,124 @@
+/*******************************************************************************
+
+    Test for the flood of the posting pre-images not happening
+
+    Copyright:
+        Copyright (c) 2019-2022 BOSAGORA Foundation
+        All rights reserved.
+
+    License:
+        MIT License. See LICENSE for details.
+
+*******************************************************************************/
+
+module agora.test.PreImageNoFlood;
+
+version (unittest):
+
+import agora.consensus.data.PreImageInfo;
+import agora.test.Base;
+
+import core.atomic;
+import core.thread;
+
+/// A validator should not receive the `postPreimage` calls after it has
+/// already received the same pre-images so that the flood of the posting
+/// pre-images does not happen.
+unittest
+{
+    static class CustomValidator : TestValidatorNode
+    {
+        mixin ForwardCtor!();
+
+        // The pointer to the number of the times the `postPreimage` is called
+        private shared int* post_preimage_count;
+
+        ///
+        public this (Parameters!(TestValidatorNode.__ctor) args,
+            shared(int)* count)
+        {
+            this.post_preimage_count = count;
+            super(args);
+        }
+
+        /// This overridden function counts the number to be called
+        public override Height postPreimage (in PreImageInfo preimage) @safe
+        {
+            atomicOp!("+=")(*this.post_preimage_count, 1);
+            return super.postPreimage(preimage);
+        }
+    }
+
+    static class CustomAPIManager : TestAPIManager
+    {
+        mixin ForwardCtor!();
+
+        // The number of times the `postPreimage` is called
+        public static shared int post_preimage_count;
+
+        /// set base class
+        public override void createNewNode (Config conf, string file, int line)
+        {
+            if (this.nodes.length == 0)
+                this.addNewNode!CustomValidator(conf, &post_preimage_count,
+                    file, line);
+            else
+                super.createNewNode(conf, file, line);
+        }
+    }
+
+    TestConf config;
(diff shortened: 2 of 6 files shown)


hewison-chris (approved):
LGTM

linked0:
Ready for review with all green.

source/agora/network/Client.d
```diff
@@ -223,7 +226,19 @@ public class NetworkClient
             break;
 
         case Preimage:
-            this.attemptRequest!(API.postPreimage, Throw.No)(event.preimage);
+            // Clean up the `last_preimages` priodically
+            // Note: It clears `last_preimages` everyday based on the CoinNet.
+            if (event.preimage.height % Height(144) == 0)
+                this.last_preimages.clear();
```
Geod24: `periodically`

source/agora/network/Client.d
```diff
@@ -223,7 +226,19 @@ public class NetworkClient
...
+
+            // We do not post the preimage if it is already posted in order to
+            // prevent calls of `postPreimage` from flooding
+            if (event.preimage.utxo !in this.last_preimages ||  
+                event.preimage.height > this.last_preimages[event.preimage.utxo].height)
```
Geod24: We only need to store the `Height` actually, not the full `PreImageInfo`, right ?

source/agora/test/PreImageCatchup.d
```diff
@@ -0,0 +1,64 @@
...
+*******************************************************************************/
+
+module agora.test.PreImageCatchup;
+
+version (unittest):
+
+import agora.consensus.data.PreImageInfo;
+import agora.test.Base;
+
+/// A validator should retrive the pre-images even when the `postPreimage`
```
Geod24: `retrieve`

source/agora/test/PreImageCatchup.d
```diff
@@ -0,0 +1,64 @@
+/*******************************************************************************
+
+    Check that a validator can retrieve missing pre-images when it could not
+    receive pre-images through `postPreimage`. We altered the algorithm of
+    distributing pre-images so that a validator does not continuously send
+    pre-images any more without checking that the previous pre-image is
+    posted normally. So the network should survive the situation where
+    some pre-images are missing.
```
Geod24: Small DDOC note: The title/summary should be short, and the longer explanation should follow.
E.g.
```D

    Some high level overview about this, in a single sentence, if possible.

    A longer explanation that can span as many sentences as one want.
    And obviously multiple line. Give as much details as you wish, really.

    Specific_notes:
      For specific topic, it's possible to have sub-heading using the same
      approach as `Params` and `Returns`. In fact, `Params`, `Returns`,
      and the likes are simply recognized sub heading and don't need to come
      in a specific order.

```

source/agora/test/PreImageCatchup.d
```diff
@@ -0,0 +1,64 @@
...
+
+        /// It makes the situation where the node doesn't receive pre-images
+        /// and miss some preimages from some point.
+        public override bool postPreimage (in PreImageInfo preimage) @safe
+        {
+            if (this.ledger.getBlockHeight() <= 1)
+                return super.postPreimage(preimage);
+            return true;
+        }
+    }
```
Geod24: What is this testing ? The golden rule of testing is that your test should fail if the fix is not present.
From the look of it, this test would pass even without the fix.

Geod24 (changes requested):
Test is lacking

linked0:
> Test is lacking

Thanks for the review. I will make a unit test more proper. Anyway, do you think this approach for keeping pre-images from flooding is right?

Geod24:
Yes

source/agora/test/PreImageCatchup.d
linked0: Modified the test based on the comment.

source/agora/test/PreImageCatchup.d
linked0: Simplified the summary of the unit test because the goal of the test has been made clearer.

source/agora/test/PreImageNoFlood.d
```diff
@@ -0,0 +1,84 @@
...
+    auto network = makeTestNetwork!CustomAPIManager(config);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+    network.generateBlocks(Height(1));
+
+    // The count of calling `postPreimage` does not change after a while
+    Thread.sleep(config.preimage_reveal_interval);
+    auto prev_post_count = atomicLoad(CustomAPIManager.post_preimage_count);
```
hewison-chris, line 80: Can we not check the expected count is the number of other validators here?

hewison-chris (changes requested):
I think it would be better to check the exact number of posts.
Also, the test you previously had for pre-image catchup is maybe worth adding back as a commit if it is not already covered by another test.

linked0:
> I think it would be better to check the exact number of posts.

I don't think it's meaningful because I found that the number of posts could be changed with just a little change in the environment, like some unintended delay based on machines on which it is tested.

> Also, the test you previously had for pre-image catchup is maybe worth adding back as a commit if it is not already covered by another test.

I agree with you. Will add it again.

source/agora/test/PreImageNoFlood.d
linked0: I understand what you mean but I found that the expected count changes every time in the middle of running the unit test with checking the count.

linked0:
Applied suggestions to the code.

hewison-chris (approved):
LGTM

source/agora/network/Client.d
```diff
@@ -156,6 +156,9 @@ public class NetworkClient
     /// Gossip delay
     private enum GossipDelay = 10.msecs;
 
+    /// Heights of last gossiped pre-images
```
Geod24: Mention that the key is the stake

source/agora/network/Client.d
```diff
@@ -227,7 +230,20 @@ public class NetworkClient
...
+
+            // We do not post the preimage if it is already posted in order to
+            // prevent calls of `postPreimage` from flooding
+            if (event.preimage.utxo !in this.last_preimages ||
+                event.preimage.height > this.last_preimages[event.preimage.utxo])
+            {
+                auto height = this.attemptRequest!(API.postPreimage, Throw.No)(event.preimage);
+                if (height >= event.preimage.height)
+                    this.last_preimages[event.preimage.utxo] = event.preimage.height;
+            }
```
Geod24:
```suggestion
            if (event.preimage.height > this.last_preimages.get(event.preimage.utxo, Height(0)))
            {
                auto height = this.attemptRequest!(API.postPreimage, Throw.No)(event.preimage);
                if (height >= event.preimage.height)
                    this.last_preimages[event.preimage.utxo] = event.preimage.height;
            }
```

Geod24 (approved):
Just a couple style changes, then feel free to self-merge

source/agora/network/Client.d
linked0: Will change to `/// Heights of last gossiped pre-images for staked UTXOs`. If it's not proper, please let me know.

source/agora/network/Client.d
linked0: The `get` function is not `nothrow`. Isn't it better to leave as it is?

Geod24 (approved):

PR #2798, opened 2022-01-06
Trivial: Rename Ledger.getLastBlock to lastBlock

Geod24:
Having a shorter name will be welcome, even thought the 'last block'
should be seldom used. Additionally, since 'lastBlock' returns by 'ref',
it is much more akin to a property function.


linked0 (approved):
LGTM

PR #2803, opened 2022-01-06
Nominator: Skip externalization if missing sigs for previous height

omerfirmak:
```
Let the catchup process handle it
```

Root cause of the issue PR #2793 was trying to address, as explained [here](https://github.com/bosagora/agora/pull/2793#issuecomment-1006342118)

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -932,6 +932,13 @@ extern(D):
                 height, last_height);
             return;  // slot was already externalized or envelope is too new
         }
+        else if (!this.ledger.hasMajoritySignature(last_height))
+        {
+            log.trace("valueExternalized: Will not externalize envelope with slot id {} as we are " ~
+                "missing signagures for height {}", height, last_height);
+            return;
+        }
+
         ConsensusData data = void;
         try
             data = deserializeFull!ConsensusData(value[]);


source/agora/consensus/protocol/Nominator.d
```diff
@@ -932,6 +932,13 @@ extern(D):
                 height, last_height);
             return;  // slot was already externalized or envelope is too new
         }
+        else if (!this.ledger.hasMajoritySignature(last_height))
+        {
+            log.trace("valueExternalized: Will not externalize envelope with slot id {} as we are " ~
+                "missing signagures for height {}", height, last_height);
+            return;
+        }
```
Geod24, line 940: I'm confused, should we even reach `valueExternalized` if we can't externalize ?

hewison-chris: Here is where we `validateValue` ~but do not block `ValueExternalized`~ in fact it should right?: 
https://github.com/bosagora/agora/blob/4269bf72d6b146a24d70a518bd8fa7006dcec9fa/source/agora/consensus/protocol/Nominator.d#L900-L903

Geod24:
Let's move ahead as it fixes the CI.

source/agora/consensus/protocol/Nominator.d
omerfirmak: `validateValue` only represents a single node's view on a value, for now we have a homogenous network but a value might be "invalid" for a single node yet be externalized by the rest of the network.

PR #2804, opened 2022-01-07
Remove the base `Ledger` dependency on `EnrollmentManager`

Geod24:
Obviously, better reviewed commit-by-commit. There's a large diff, but code is mostly moved. Checking it out locally with a good `diff` tool makes the PR **almost** trivial.

There are a few issues with our current `Ledger` class:
- It depends on `ConsensusData`, which is an SCP-only struct;
- It depends on `EnrollmentManager`, which is only required for validating;
- It's massive and hard to make sense of;
- It has bit-rotten unittests;

On a functional level, it is hardly usable as a library because of the dependencies it bring. This can be seen by the difficulty of writing unittests, but also, by the fact we're not using it in Faucet. Another aspect of the unneeded dependencies is that we have a `EnrollmentManager` in the `FullNode` (wat), and we pass `this.config.validator` even though we know it's disabled ([wat](https://www.destroyallsoftware.com/talks/wat)).

This PR fixes some of those issues, by moving the most basic functionalities, those that depend on `ValidatorSet` and `UTXOSet`, to the base `Ledger` class, and leaving the rest of the monster behind in `NodeLedger`. This allowed me to progressively weed out dependencies, and introduce methods were needed.

The next step, after this PR, will be to remove `EnrollmentManager` from `NodeLedger`. This means that a `FullNode` will then be able to instantiate a `NodeLedger` without creating extra dependencies. In a future PR, we could even make `ValidatingLedger` have *multiple* EM, allowing it to manage multiple stakes at once... But let's not get ahead of ourselves.

Currently, what's blocking `NodeLedger` to get rid of `EnrollmentManager` is the fact that the registry uses a function that should be reserved to `ValidatingLedger`, `getCandidateEnrollments`:
https://github.com/bosagora/agora/blob/51774f8762ab34ab1a311eacd77e6ea4263843e9/source/agora/node/Registry.d#L235-L237

I have a few ideas on how to tackle this, but it requires more functional changes, and I think this PR is already quite large, despite having been split into some smaller PRs already (most recently #2799, #2800, #2801, but there were a million others before).


Geod24:
Looks like it's still too big and I missed a few things. Will split it into smaller PRs.

source/agora/consensus/Ledger.d
```diff
@@ -243,6 +243,28 @@ public class Ledger
...
+        returned by `getValidators` as it avoids any intermediate allocation.
+        An 'active' validator is one that hasn't been slashed,
+        and may sign a block. A validator that enrolls, reveal all its
+        pre-images but never sign a block will still be counted as 'active'.
+
+        Params:
+          height = Height at which to query the validator set.
+                   Accurate results are only guaranteed for
+                   `height <= this.getBlockHeight() + 1`.
+
```
linked0: Should have a doc for the return value

Geod24: Fixed in https://github.com/bosagora/agora/pull/2805

Geod24:
All green 🎉

hewison-chris (approved):
LGTM

PR #2805, opened 2022-01-07
Ledger: add a function to count active validators

Geod24:
```
This is a common enough use case that such a function is warranted.
More importantly, it removes a dependency from FullNode to the EnrollmentManager.
```

Extracted from https://github.com/bosagora/agora/pull/2804


linked0 (approved):
Looks good.

hewison-chris (approved):
LGTM

PR #2806, opened 2022-01-07
Introduce NodeLedger as an intermediate Ledger type

Geod24:
```
A NodeLedger is a more involved Ledger than the basic Ledger,
as the basic Ledger should be usable by a client without some of
the machinery present in the NodeLedger, most notably the two pools.

As a first step, and instant win to decouple things, we can move
`onValidatedBlock` to the `NodeLedger`, as a client would most likely
perform operations sequentially (call `acceptBlock` then check state).
```

Extracted from https://github.com/bosagora/agora/pull/2804

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -85,10 +85,6 @@ public class Ledger
     /// Enrollment manager
     protected EnrollmentManager enroll_man;
 
-    /// If not null call this delegate
-    /// A block was externalized
-    protected void delegate (in Block, bool) @safe onAcceptedBlock;
-
     /// Parameters for consensus-critical constants
     protected immutable(ConsensusParams) params;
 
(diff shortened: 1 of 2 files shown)


source/agora/consensus/Ledger.d
```diff
@@ -1485,6 +1468,61 @@ public class Ledger
     }
 }
 
+/// The Ledger class held by a node
```
hewison-chris, line 1471: Due to the large size of  this file is it maybe better to create `NodeLedger.d` file?

source/agora/consensus/Ledger.d
```diff
@@ -1516,7 +1554,7 @@ unittest
 
 *******************************************************************************/
 
-public class ValidatingLedger : Ledger
+public class ValidatingLedger : NodeLedger
```
hewison-chris, line 1557: Also maybe a new file. But maybe not part of this commit.

hewison-chris (approved):
LGTM

source/agora/consensus/Ledger.d
Geod24: Check the linked PR, it does that in one of the last commits, but first I need to break down the dependencies.

hewison-chris: Great minds ...

PR #2807, opened 2022-01-07
Ledger: Gradually move pool / validation related utilities from `Ledger` to `NodeLedger`

Geod24:
This is the main refactoring that comes from https://github.com/bosagora/agora/pull/2804
Each commit moves a specific set of functionalities to `NodeLedger`. First the transaction pool, then the enrollment pool (which the `EnrollmentManager` acts as a proxy for), and finally some misc functions / types.
The result is a "clean" `Ledger`, which we will be able to isolate in its own module, and will not require an EM / key to instantiate.

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -76,8 +76,8 @@ public class Ledger
     /// UTXO set
     protected UTXOCache utxo_set;
 
-    /// Enrollment manager
-    protected EnrollmentManager enroll_man;
+    /// The object controlling the validator set
+    protected ValidatorSet validator_set;
 
     /// Parameters for consensus-critical constants
     protected immutable(ConsensusParams) params;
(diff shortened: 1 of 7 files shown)


source/agora/consensus/Ledger.d
```diff
@@ -1005,6 +994,16 @@ public class Ledger
 /// The Ledger class held by a node
 public class NodeLedger : Ledger
```
linked0, line 995: It would be better to document the roles of the `NodeLedger` in more detail and so is for `Ledger` because we need to know the difference of their roles explicitly, which might be applied in later PRs, I guess.

linked0 (approved):
LGTM

source/agora/consensus/Ledger.d
Geod24: Good point, the documentation is severely lacking. I will try to improve it after the split.

hewison-chris (approved):
LGTM

PR #2811, opened 2022-01-07
DNS: Fix domain name decompression and better RDATA toString

omerfirmak:


Geod24:
Do we need the second commit ? It looks like a lot of changes for a marginal gain.

Geod24:
And if you want to fix `toString`, do it at the RR level, where you know the type, instead of duplicating it.

omerfirmak:
> Do we need the second commit ? It looks like a lot of changes for a marginal gain.

The output was incorrect, it would print "1 bytes" for an IPv4 addr

> And if you want to fix toString, do it at the RR level, where you know the type, instead of duplicating it.

But then you would have to manually print other fields as well, no?

Geod24:
> But then you would have to manually print other fields as well, no?

Yeah but that's not really a big deal is it ?

omerfirmak:
> Do we need the second commit

Dropped

Geod24 (approved):
Can you add a test ?

PR #2813, opened 2022-01-07
Change `node.realm` to be of type `Domain`

Geod24:
And improve the validation of `Domain` in the process.


Geod24:
https://github.com/bosagora/config/pull/15 will make the config experience much nicer.

PR #2817, opened 2022-01-10
FullNode: Remove makeFeeManager

Geod24:
This was originally needed because we didn't have proper dependency injection:
we were passing a path around and each class was instantiating their own DB.
We now have a unified DB (stateDB) which we pass to classes,
so the need for makeFeeManager has disappeared.

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -126,9 +126,6 @@ public class FullNode : API
     /// Enrollment manager
     protected EnrollmentManager enroll_man;
 
-    /// Fee manager
-    protected FeeManager fee_man;
-
     /// Set of unspent transaction outputs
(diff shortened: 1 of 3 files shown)


source/agora/node/FullNode.d
```diff
@@ -948,7 +930,8 @@ public class FullNode : API
     protected NodeLedger makeLedger ()
     {
         return new NodeLedger(params, this.engine, this.utxo_set, this.storage,
-            this.enroll_man, this.pool, this.fee_man, &this.onAcceptedBlock);
+            this.enroll_man, this.pool, new FeeManager(this.stateDB, this.params),
```
linked0, line 933: The `makeLedger` in `Validator.d` should be also modified like this.

hewison-chris (approved):
LGTM

PR #2818, opened 2022-01-10
DNS: Send NS record pointing to primary

Geod24:
```
Those are currently missing and prevent recursion from working correctly.
There is currently no protection against misconfiguration.
```

This is not ideal, because we should allow multiple `NS` records for example, but it will do for now.


mkykadir (approved):
LGTM but, need to think about how to inform about secondaries. Generally NS queries also return secondary servers.

PR #2820, opened 2022-01-10
Usability: Allow to start Agora with `-c /dev/null`

Geod24:
This is to allow a zeroconf Agora to "just work".
Obviously this usage is not the recommended one, as every restart
means re-syncing the whole chain, but for the foreseable future
it will be a welcome tool to have users test our chain.


linked0 (approved):
LGTM

PR #2823, opened 2022-01-10
Backoff: Change the algorithm from a class to a function

Geod24:
Using a class for this is quite overkill, and seems to only have been
done to make the utility testable, which can be achieved just as well
by using a template parameter.

### DIFF
--- a/source/agora/utils/Backoff.d
+++ b/source/agora/utils/Backoff.d
@@ -19,68 +19,56 @@ import libsodium.randombytes;
 
 import std.algorithm : min;
 
-/// Ditto
-public class Backoff
+/*******************************************************************************
+
+    Get the delay to apply before retrying
+
+    The unit is purposely left unspecified: Calling code can use `.seconds`,
+    `.msecs`, or whichever unit is deemed appropriate.
+
+    The algorithm is simply:
+    ```
+    sleep_time = random(0, min(max_delay, base * (2 ^^ attempt));
+    ```
+
+    Params:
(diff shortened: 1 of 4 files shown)


source/agora/utils/Backoff.d
```diff
@@ -19,68 +19,55 @@ import libsodium.randombytes;
...
+    ```
+    sleep_time = random(0, min(max_delay, base * (2 ^^ attempt));
+    ```
 
-    ***************************************************************************/
+    Params:
+        RandomFunc = A function that takes a `uint` as an input and returns a
+                     value between `0` and the provided value, uniformly
+        attempt = the attempt number (can be 0, which will result in 0)
+        base    =
```
hewison-chris: ?

source/agora/utils/Backoff.d
```diff
@@ -19,68 +19,55 @@ import libsodium.randombytes;
...
-    ***************************************************************************/
+*******************************************************************************/
 
-    protected uint getRandom (uint value)
-    {
-        return () @trusted { return randombytes_uniform(value); }();
-    }
+public uint getRandom (uint value) @safe nothrow @nogc
+{
+    return () @trusted { return randombytes_uniform(value); }();
```
omerfirmak, line 71: I would not call this "jitter", maybe we should take a look at this later.

Geod24:
Fixed documentation

linked0 (approved):
LGTM

PR #2825, opened 2022-01-10
Limit AXFR to whitelisted nodes

### LINKED ISSUE #2647: DNS: Implement whitelisting for AXFR

mkykadir:
Fixes #2647

### DIFF
--- a/source/agora/node/Config.d
+++ b/source/agora/node/Config.d
@@ -458,6 +458,8 @@ public struct RegistryConfig
 
         static void validateZone (string name, in ZoneConfig zone)
         {
+            import std.socket : parseAddress;
+
             // Those two fields must be both set or both unset
             ensure(zone.primary.set == zone.soa.email.set,
                    "registry.{}: Authoritative zones require both 'primary' and " ~
@@ -508,11 +510,35 @@ public struct RegistryConfig
 /// All `Duration` values are precise to the second
 public struct ZoneConfig
 {
+    /// IP address configuration type for `allow_transfer`
+    private struct IPAddress
+    {
+        /// Address
+        public string ip;
+
+        /// Validate `str` as IPv4 or IPv6 address
+        public static IPAddress fromString (string str) @safe
+        {
+            import agora.common.DNS : guessAddressType;
+            auto type = guessAddressType(str);
+            if (type != TYPE.A && type != TYPE.AAAA)
+                throw new Exception("Only IP address is expected");
+
+            return IPAddress(str);
+        }
+
+        ///
+        public alias ip this;
+    }
+
     /// Whether this server is authoritative
     public SetInfo!bool authoritative;
 
     /// CNAME of the primary server for this zone
     public SetInfo!string primary;
+
+    /// Servers that are allowed to do AXFR queries, applicable for authoritatives
+    public @Optional immutable IPAddress[] allow_transfer;
 
     /// SOA Configuration of the zone, required for master server
     /// All `Duration` values are precise to the second
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -56,8 +56,16 @@ public class NameRegistry: NameRegistryAPI
     ///
     protected RegistryConfig config;
 
+    /// Indexes for `zones`
+    private enum ZoneIndex
+    {
+        Realm = 0,
+        Validator = 1,
+        Flash = 2,
+    }
+
     /// Zones of the registry
-    private ZoneData[Domain] zones;
+    private ZoneData[ZoneIndex.max + 1] zones;
 
     /// The domain for `realm`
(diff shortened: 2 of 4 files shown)


source/agora/node/Runner.d
```diff
@@ -319,7 +319,7 @@ private void runDNSServer_canThrow (in RegistryConfig config, NameRegistry regis
             auto pack = udp.recv(buffer, ppeer);
             auto query = deserializeFull!Message(pack);
             registry.answerQuestions(
-                query,
+                query, peer.toAddressString,
```
mkykadir, line 322: Need to consider [this](https://github.com/bosagora/agora/pull/2703#discussion_r780897531)

mkykadir: I will also consider reverse proxy situation

source/agora/node/Runner.d
```diff
@@ -420,7 +420,7 @@ private void runTCPDNSServer_canThrow (TCPConnection conn, NameRegistry registry
 
         auto query = deserializeFull!Message(buffer[0 .. size]);
         registry.answerQuestions(
-            query,
+            query, conn.peerAddress(),
```
mkykadir: this returns with port number also

source/agora/node/Runner.d
mkykadir: I've checked Nginx about reverse proxy on stream sockets (i.e. TCP and UDP) and stream configs doesn't have a `forwarded for` feature like in the HTTP configurations. Either client IP is directed or proxy IP is directed, thus we cannot do anything about it through Agora.

mkykadir:
Rebased

source/agora/node/Registry.d
```diff
@@ -56,8 +56,16 @@ public class NameRegistry: NameRegistryAPI
...
+    private enum ZoneIndex
+    {
+        REALM = 0,
+        VALIDATOR = 1,
+        FLASH = 2,
+    }
+
     /// Zones of the registry
-    private ZoneData[Domain] zones;
+    private ZoneData[3] zones;
```
omerfirmak: Would `ZoneIndex.max + 1` work?

linked0 (approved):
It's green. LGTM.

source/agora/node/Registry.d
```diff
@@ -56,8 +56,16 @@ public class NameRegistry: NameRegistryAPI
     ///
     protected RegistryConfig config;
 
+    /// Indexes for `zones`
+    private enum ZoneIndex
+    {
+        REALM = 0,
```
Geod24: Why the uppercase ?

source/agora/node/Registry.d
```diff
@@ -438,9 +445,10 @@ public class NameRegistry: NameRegistryAPI
 
     auto findZone (Domain name, bool matches = true) @safe
     {
-        auto matching_zone = name in this.zones;
-        if (matching_zone)
-            return matches ? &matching_zone.answer_matches : &matching_zone.answer_owns;
+        foreach (i, zone; this.zones)
```
Geod24: `const ref zone` to avoid copying ?

source/agora/node/Config.d
```diff
@@ -514,6 +514,9 @@ public struct ZoneConfig
     /// CNAME of the primary server for this zone
     public SetInfo!string primary;
 
+    /// Servers that are allowed to do AXFR queries, applicable for authoritatives
+    public @Optional immutable string[] allow_transfer;
```
Geod24: Need to validate that it's using IP addresses tho

Geod24 (approved):
Broadly LGTM.
Please add validation on the config level so it's hard / impossible to misconfigure.

source/agora/node/Config.d
```diff
@@ -467,6 +469,16 @@ public struct RegistryConfig
...
 
+            foreach (addr; zone.allow_transfer)
+            {
+                try
+                    parseAddress(addr);
+                catch (Exception)
+                    ensure(0,
+                        "allow_transfer: Field must consist from IP addresses, {} is not an IP address",
+                        addr);
+            }
```
Geod24: I think you could get away with `guessAddressType`, which doesn't `throw`.
Also if you create a simple type for it with a `fromString` / `this (string)` and throw from there, you will get the file/line and field name printed for you by the config parser.

mkykadir: I picked second approach, which looks better during user configuration I think.

Geod24 (approved):

PR #2827, opened 2022-01-10
DNS: Add AAAA record to the list of accepted types

Geod24:
While we don't store them at the moment, this allow to remove
the warning that shows up quite a bit in the logs due to recursion.


linked0 (approved):
LGTM

PR #2828, opened 2022-01-10
DNS: Return a range of Address from getAddresses

Geod24:
Avoid needless alllocations as the range is going to be iterated over.

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -193,7 +193,7 @@ public class NameRegistry: NameRegistryAPI
 
     ***************************************************************************/
 
-    public Address[] getValidatorsAddresses ()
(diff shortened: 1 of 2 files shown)


source/agora/node/Registry.d
```diff
@@ -193,7 +193,7 @@ public class NameRegistry: NameRegistryAPI
 
     ***************************************************************************/
 
-    public Address[] getValidatorsAddresses ()
+    public auto validatorsAddresses ()
```
omerfirmak, line 196: Is this really necessary? They can all be reached at `pubkey.validators.realm` anyways

Geod24: The change or the method itself ?

mkykadir (approved):

PR #2829, opened 2022-01-10
DNS: Correctly deserialize NS records

Geod24:


omerfirmak (approved):
`ebadrebase` should be a default branch git creates after every rebase

PR #2830, opened 2022-01-10
Ledger: Limit TX set size in nominations

### LINKED ISSUE #2712: Block size should be limited
It will probably take some time reach to a level of network activity for this to matter

omerfirmak:
Fixes #2712

Point made in https://github.com/bosagora/agora/pull/2416#issuecomment-906494622 still stands, will need a separate PR

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -1601,52 +1601,52 @@ unittest
     import agora.consensus.PreImage;
 
     {
-    auto params = new immutable(ConsensusParams)(20, 7, 80, Amount(10));
-    const(Block)[] blocks = [ GenesisBlock ];
-    scope ledger = new TestLedger(genesis_validator_keys[0], blocks, params);
+        auto params = new immutable(ConsensusParams)(20, 7, 80, Amount(10));
+        const(Block)[] blocks = [ GenesisBlock ];
+        scope ledger = new TestLedger(genesis_validator_keys[0], blocks, params);
 
-    ushort min_fee_pct = 80;
+        ushort min_fee_pct = 80;
 
-    auto no_fee_tx = genesisSpendable().front().refund(WK.Keys[0].address).feeRate(0.coins).sign();
-    assert(ledger.acceptTransaction(no_fee_tx, 0, min_fee_pct) !is null);
+        auto no_fee_tx = genesisSpendable().front().refund(WK.Keys[0].address).feeRate(0.coins).sign();
+        assert(ledger.acceptTransaction(no_fee_tx, 0, min_fee_pct) !is null);
 
-    auto average_tx = genesisSpendable().front().refund(WK.Keys[0].address).deduct(10.coins).sign();
-    assert(ledger.acceptTransaction(average_tx, 0, min_fee_pct) is null);
+        auto average_tx = genesisSpendable().front().refund(WK.Keys[0].address).deduct(10.coins).sign();
+        assert(ledger.acceptTransaction(average_tx, 0, min_fee_pct) is null);
 
-    // switch to a different input, even with low fees this should be accepted since pool is almost empty
-    auto different_tx = genesisSpendable().dropOne().front().refund(WK.Keys[0].address).deduct(1.coins).sign();
-    assert(ledger.acceptTransaction(different_tx, 0, min_fee_pct) is null);
+        // switch to a different input, even with low fees this should be accepted since pool is almost empty
+        auto different_tx = genesisSpendable().dropOne().front().refund(WK.Keys[0].address).deduct(1.coins).sign();
+        assert(ledger.acceptTransaction(different_tx, 0, min_fee_pct) is null);
 
-    // lower than average, but enough
-    auto enough_fee_tx = genesisSpendable().dropOne().front().refund(WK.Keys[0].address).deduct(9.coins).sign();
-    assert(ledger.acceptTransaction(enough_fee_tx, 0, min_fee_pct) is null);
+        // lower than average, but enough
+        auto enough_fee_tx = genesisSpendable().dropOne().front().refund(WK.Keys[0].address).deduct(9.coins).sign();
+        assert(ledger.acceptTransaction(enough_fee_tx, 0, min_fee_pct) is null);
 
-    // overwrite the old TX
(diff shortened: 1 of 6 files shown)


omerfirmak:
Added a second commit to address https://github.com/bosagora/agora/pull/2416#issuecomment-906494622

source/agora/consensus/Ledger.d
```diff
@@ -228,7 +228,8 @@ public class NodeLedger : Ledger
         if (auto err = this.fee_man.getTxFeeRate(tx, this.utxo_set.getUTXOFinder(),
             &this.getPenaltyDeposit, fee_rate))
             return err;
-        if (fee_rate < min_fee)
+        const size_budget = this.params.MaxTxSetSize * 1024;
```
```suggestion
        const size_limit = this.params.MaxTxSetSize * 1024;
```

source/agora/consensus/Ledger.d
```diff
@@ -228,7 +228,8 @@ public class NodeLedger : Ledger
...
+        if (fee_rate < min_fee && this.pool.getPoolSize() >= size_budget) // If pool is sparsely populated, accept any fee.
```
hewison-chris: Isn't there a risk that no fees will be paid with this change?

source/agora/consensus/Ledger.d
```diff
@@ -365,6 +366,11 @@ public class NodeLedger : Ledger
         if (auto fail_reason = this.getValidTXSet(data, tx_set, utxo_finder))
             return fail_reason;
 
+        const tx_set_size = tx_set.map!(tx => tx.sizeInBytes()).sum();
+        const size_budget = this.params.MaxTxSetSize * 1024;
```
```suggestion
        const size_limit = this.params.MaxTxSetSize * 1024;
```

source/agora/consensus/Ledger.d
```diff
@@ -757,11 +763,11 @@ public class ValidatingLedger : NodeLedger
 
     ***************************************************************************/
 
-    public Hash[] getCandidateTransactions (in Height height, ulong max_txs,
-        scope UTXOFinder utxo_finder) @safe
+    public Hash[] getCandidateTransactions (in Height height, scope UTXOFinder utxo_finder) @safe
     {
         Hash[] result;
         Amount tot_fee, tot_data_fee;
+        size_t size_budget = this.params.MaxTxSetSize * 1024; // cant overflow
```
hewison-chris, line 784:
```suggestion
        size_t size_limit = this.params.MaxTxSetSize * 1024; // cant overflow
```

source/agora/consensus/Ledger.d
```diff
@@ -861,33 +867,28 @@ public class ValidatingLedger : NodeLedger
     }
 
     /// Generate a new block by creating transactions, then calling `forceCreateBlock`
-    private Transaction[] makeTestBlock (
-        Transaction[] last_txs, ulong txs = Block.TxsInTestBlock)
+    private Transaction[] makeTestBlock (Transaction[] last_txs)
```
hewison-chris, line 884: Nice, finally removed. :-)

hewison-chris (approved):
LGTM, just not sure we should allow lower then min fee rate when tx set size is below the limit.

source/agora/consensus/Ledger.d
```diff
@@ -757,11 +762,11 @@ public class ValidatingLedger : NodeLedger
 
     ***************************************************************************/
 
-    public Hash[] getCandidateTransactions (in Height height, ulong max_txs,
-        scope UTXOFinder utxo_finder) @safe
+    public Hash[] getCandidateTransactions (in Height height, scope UTXOFinder utxo_finder) @safe
```
linked0: Two-line is better, I guess.

source/agora/consensus/Ledger.d
```diff
@@ -1594,6 +1596,29 @@ unittest
     auto average_tx = genesisSpendable().front().refund(WK.Keys[0].address).deduct(10.coins).sign();
     assert(ledger.acceptTransaction(average_tx, 0, min_fee_pct) is null);
 
+    // switch to a different input, even with low fees this should be accepted since pool is almost empty
```
linked0: All the added lines in the bracket should be indented.

source/agora/consensus/pool/Transaction.d
```diff
@@ -475,6 +475,25 @@ public class TransactionPool
         return 0.coins;
     }
 
+    /***************************************************************************
+
+        Returns:
+            Total size of the transactions in the pool
```
linked0: Would be better to add the unit of size, like "in bytes"

source/agora/consensus/Ledger.d
```diff
@@ -1727,6 +1727,51 @@ unittest
     }
 
     // should be able to create a new block after previous height receives majority signatures
-    ledger.forceCreateBlock(0);
+    ledger.forceCreateBlock();
     assert(ledger.getBlockHeight() == 2);
 }
+
+unittest
```
linked0: Isn't it better to explain what this unit test is for?

linked0 (commented):
Broadly good. Left some suggestions.
And please add the commit message on the first commit.

source/agora/consensus/Ledger.d
omerfirmak: I will add a test for it.

omerfirmak: added a simple check for it, FeeManager handles that and does not allow any TXs with fees below `ConsensusConfig.min_fee`

omerfirmak:
> not sure we should allow lower then min fee rate when tx set size is below the limit

We need that to be able to talk about an average of fees in the pool. If we don't do that, a single high paying TX can block any other TX to be accepted in a block.

source/agora/consensus/Ledger.d
omerfirmak: I applied previous suggestions but since we keep subtracting from this, it actually represents the budget we have left. So it seem like a better name.

source/agora/consensus/Ledger.d
omerfirmak: They result in a hard to follow diff tho

linked0: Indentation is the basic requirement, I guess. But it's not the critical one.

linked0 (approved):
LGTM

omerfirmak:
Rebased

PR #2839, opened 2022-01-11
Explicitly transfer target height of an Enrollment

### LINKED ISSUE #2838: First enrollment of a node can associate a PreImage with wrong height
`checkAndEnroll` creates a new enrollment and sends it to the network, since there is no previous enrollment nodes accept it regardless of it's commitment.

If externalization of the enrollment gets delayed for even a single block, the commitment in the enrollment is associated with the wrong height, causing any PreImage revealed after that to be deemed invalid.

omerfirmak:
Previously, target height was assumed to be the next height. After
next block is externalized, enrollment was removed from the pool, even
if not externalized.

If an Enrollment gets gossiped after externalization of it's target
height that the originating node intended, it could be externalized
at a later height. Which caused the commitment in the Enrollment to be
associated with a wrong height and made impossible to reveal any new
PreImages.

By sending target height explicitly along the enrollment, we guarantee
that a delay in externalization of Enrollments don't doom the Validator.

Fixes #2838


hewison-chris (approved):
This works but would it not be better to just add this height to the `Enrollment` struct?

Geod24 (changes requested):
This doesn't really fix the issue: If a byzantine validator changes the value and gossip it faster than us, we could end up in the same situation (and get slashed).
This kind of issues is why we originally had enrollment cycles. Adding the height to the Enrollment was something we tried very hard to avoid in the past.

omerfirmak:
> : If a byzantine validator changes the value and gossip it faster than us, we could end up in the same situation (and get slashed)

Yep, thanks for pointing that out.

> Adding the height to the Enrollment was something we tried very hard to avoid in the past.

Let's just sign the intended height as well but not add it to the struct, nodes will try to validate Enrollment with the assumption that it is intended for the next height.

So `enroll_sig = hashMulti(height, enrollment).sign();`

Geod24:
That's slick, I like it a lot.

source/agora/api/FullNode.d
```diff
@@ -322,7 +322,7 @@ public interface API
 
     ***************************************************************************/
 
-    public void postEnrollment (in Enrollment enroll);
+    public void postEnrollment (in Enrollment enroll, in Height avail_height);
```
linked0, line 332: Needs documentation for the `avail_height`.

source/agora/node/FullNode.d
```diff
@@ -1048,10 +1048,14 @@ public class FullNode : API
...
 
     /// POST /enrollment
-    public override void postEnrollment (in Enrollment enroll) @safe
+    public override void postEnrollment (in Enrollment enroll, in Height avail_height) @safe
     {
         this.recordReq("postEnrollment");
 
+        // Ignore enrollments targeting a previous Height
+        if (avail_height <= this.ledger.getBlockHeight())
+            return;
```
linked0, line 1067: It would be better to add a log for the ignoring.

omerfirmak:
> So enroll_sig = hashMulti(height, enrollment).sign();

Added a second commit implementing that, we may squash it before merging

linked0 (approved):
Great but the second commit could be split into two commits.

omerfirmak:
> Great but the second commit could be split into two commits.

I thought of it as pretty atomic, how should I do that?

source/agora/consensus/EnrollmentManager.d
```diff
@@ -819,48 +818,42 @@ unittest
...
-    // Adding enrollments in reverse order of utxo order to show it still works
-    const EnrollAt10 = Height(10);
-    ordered_enrollments.sort!("a.utxo_key > b.utxo_key");
-    foreach (idx, ordered_enroll; ordered_enrollments)
-        assert(man.addEnrollment(ordered_enroll, pairs[idx].address, EnrollAt10,
-            &utxo_set.peekUTXO, getPenaltyDeposit));
-    enrolls = man.getEnrollments(EnrollAt10, &utxo_set.peekUTXO, getPenaltyDeposit);
-    assert(enrolls.length == 3);
-    assert(enrolls.isStrictlyMonotonic!("a.utxo_key < b.utxo_key"));
-
```
linked0, line 832: I thought that this is unrelated to this commit.

source/agora/consensus/EnrollmentManager.d
```diff
@@ -898,41 +891,37 @@ unittest
 
     Height height = Height(2);
 
-    Enrollment[] enrollments;
-    foreach (idx, kp; pairs[0 .. 3])
-    {
-        enrollments ~= EnrollmentManager.makeEnrollment(
-            utxo_hashes[idx], kp, height, params.ValidatorCycle);
-    }
-
```
linked0, line 907: Ditto.

linked0 (commented):
I spotted the unrelated modification to the second commit. But it could be thought of as atomic based on the point of view. Good job, go ahead as it is.

source/agora/consensus/EnrollmentManager.d
omerfirmak: Those are both required to make the tests pass with this change.

omerfirmak:
Added a simple test as well

Geod24 (approved):

Geod24:
Can you rebase ?

omerfirmak:
Rebased

PR #2842, opened 2022-01-12
NetMan: Remove getNetworkClient

Geod24:
Originally, getNetworkClient was intended as the dependency injection point.
However, over time, it has evolved from the 'network layer' into a higher level,
as it now manages multiple connection, the gossip queue, and retries.
The true dependency injection point is 'NetworkManager.getClient'.


omerfirmak (approved):

linked0 (approved):
LGTM

PR #2843, opened 2022-01-12
Reduce testing for `ValidatorRecurringEnrollment`

### LINKED ISSUE #1977: Test `agora.test.ValidatorRecurringEnrollment` is sleep-bound (4m15s)
```
2021-04-18T11:56:18.2318560Z Unittesting agora.test.ValidatorRecurringEnrollment..
2021-04-18T12:00:03.9434600Z Unittesting agora.test.ValidatorCount..
```

This is 4m15s of run time on MacOSX, which is 33% of the CI run time on a good day (less on a bad day).
It takes almost the same time on Ubuntu, which leads me to believe that it's not due to the inherent complexity.

linked0:
The `ValidatorRecurringEnrollment` test has an intrinsic characteristic that causes so much testing time like 1 minute and 24 secs on my local machine, which is the most time-consuming test. The test has 125 times of `expectHeightAndPreImg` calls that have `expectHeight` that takes 400~500 msecs. 

By the way, we try to reduce the testing time by removing the unnecessary communication time for testing recurring enrollments in this PR, which reduces 30% of the testing time from 1 minute and 24 secs to 58 secs. We could reduce the testing time of other unit tests based on this PR.

EDIT: We removed the unnecessary test in the first commit because the removed test is also checked almost same way in [this unit test](https://github.com/bosagora/agora/blob/d293d3f94fed6e7d9e7c4ba15a9441c930a955cb/source/agora/test/ValidatorRecurringEnrollment.d#L149-L152).

Fixes #1977

### DIFF
--- a/source/agora/test/ValidatorRecurringEnrollment.d
+++ b/source/agora/test/ValidatorRecurringEnrollment.d
@@ -72,32 +72,11 @@ unittest
     auto blocks = node_0.getBlocksFrom(0, 2);
     assert(blocks.length == 1);
 
-    Transaction[] txs;
-
-    void createAndExpectNewBlock (Height new_height)
-    {
-        // create enough tx's for a single block
-        txs = blocks[new_height - 1].spendable().map!(txb => txb.sign())
-            .array();
-
-        // send it to one node
-        txs.each!(tx => node_0.postTransaction(tx));
-
-        network.expectHeightAndPreImg(new_height, blocks[0].header);
-
-        // add next block
-        blocks ~= node_0.getBlocksFrom(new_height, 1);
-    }
-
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM, but can you put link to test that covers the one you are removing please. Also how much time is now saved?

linked0:
> LGTM, but can you put link to test that covers the one you are removing please. Also how much time is now saved?

I will put the link to test and this PR reduces 30% of the testing time from 1 minute and 24 secs to 58 secs. The main reason that time is saved is to remove the network time by `postTransaction` that is not called by setting `no_txs` to `true`.

linked0:
I put the link to the reference test with the `EDIT` tag in the summary.

PR #2845, opened 2022-01-13
NetworkManager: Remove `merge`, delay `NetworkClient` instantiation

Geod24:
The goal here was to remove the notion of "merge", and instead rely on `NetworkClient` being added to.
In the process, `NetworkClient` gained a (sticky) identity and `NodeConnInfo` was removed.
I hope this makes the code simpler. The test currently fails on my machine (spuriously), so I'm going to break this down to make sure no commit introduce a regression.

Next steps are to:
- Allow for `peers` to contain `NetworkClient` with empty connection array;
- Eagerly instantiate `NetworkClient` on `required_peers`;
- Move what `ConnectionTask` is currently doing inside `NetworkClient` (somehow);
- Validate the incoming UTXO, creating four tiers: quorums, validators, nodes with a stake, full nodes;
- Rework the limits for the tiers above;
- Change the nominator to gossip only to our quorum;

### DIFF
--- a/source/agora/api/FullNode.d
+++ b/source/agora/api/FullNode.d
@@ -49,13 +49,19 @@ public struct NodeInfo
 public struct Identity
 {
     /// Public Key of the node
-    PublicKey key;
+    public PublicKey key;
 
     /// UTXO that is used as collateral
-    Hash utxo;
+    public Hash utxo;
 
     /// MAC
-    ubyte[] mac;
+    public ubyte[] mac;
+
+    /// Whether or not this is in the `init` state (meaning a full node)
+    public bool opCast (T : bool) () const scope @safe pure nothrow @nogc
+    {
+        return this !is Identity.init;
+    }
 }
 
 /*******************************************************************************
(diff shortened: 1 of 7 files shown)


mkykadir (commented):
1st commit -> Typo in method name; should be `onHandshakeComplete`
2nd commit -> makes `NetworkClient.handshake` useless; why not remove it?
4th commit -> not atomic, `Manager` still uses `NetworkClient.connections`
5th commit -> not atomic, `Manager` still uses `NetworkClient.connections`

Geod24:
> 2nd commit -> makes NetworkClient.handshake useless; why not remove it?

Good point.

> 4th commit -> not atomic, Manager still uses NetworkClient.connections
> 5th commit -> not atomic, Manager still uses NetworkClient.connections

They are atomic, because what the constructor used to do is now done by `merge`, which is called right after instantiation of `NetworkClient`.

Geod24:
Ended up with #2985

PR #2848, opened 2022-01-13
Ledger: Move FeeManager / UTXOCache instantiation in the Ledger

Geod24:
```
The instantiation happening in the node is a leftover of the time
where dependency injection was done at the class level.
Now that we just pass a ManagedDatabase, there is no need for it.
If tests want to implement a specific behavior for one of those
classes, they can re-introduce the dependency injection in the Ledger.
```

### DIFF
--- a/source/agora/consensus/Ledger.d
+++ b/source/agora/consensus/Ledger.d
@@ -91,21 +91,19 @@ public class NodeLedger : Ledger
 
         Params:
             params = the consensus-critical constants
-            engine = script execution engine
-            utxo_set = the set of unspent outputs
+            database = State database
             storage = the block storage
+            engine = script execution engine
             enroll_man = the enrollmentManager
             pool = the transaction pool
-            fee_man = the FeeManager
             onAcceptedBlock = optional delegate to call
                               when a block was added to the ledger
 
     ***************************************************************************/
 
     public this (immutable(ConsensusParams) params,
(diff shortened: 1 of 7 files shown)


Geod24:
Rebased

source/agora/consensus/Ledger.d
```diff
@@ -91,21 +91,19 @@ public class NodeLedger : Ledger
...
                               when a block was added to the ledger
 
     ***************************************************************************/
 
     public this (immutable(ConsensusParams) params,
-        Engine engine, UTXOSet utxo_set, IBlockStorage storage,
-        EnrollmentManager enroll_man, TransactionPool pool,
-        FeeManager fee_man,
+        ManagedDatabase database, IBlockStorage storage,
+        Engine engine, EnrollmentManager enroll_man, TransactionPool pool,
```
omerfirmak, line 106: We could do the same for `engine` as well, it is not stateful and pretty lightweight too. There is no reason to share it.

omerfirmak (approved):

source/agora/consensus/Ledger.d
omerfirmak: Engine could even be disposible, created only in the scope that it is needed.

Geod24: Yeah I wanted to do it for Engine too, but it is used in the `FlashNode`... I wanted to have a look, but didn't consider duplicating it.

PR #2849, opened 2022-01-14
Dockerfile: Temporarily use alpine 3.15

Geod24:
The reason we use edge is to match 'docker-agora-builder'.
However, upstream has broken the LDC package and we are currently
waiting on them to merge the fix.
Luckily, the last Alpine release was 2021-11-24,
and no LDC update has happened since.


Geod24:
> Error relocating /usr/local/bin/agora: _ZSt28__throw_bad_array_new_lengthv: symbol not found

FFS

PR #2850, opened 2022-01-14
CI: Extend config file check to devel files

Geod24:
```
Those files are easy to break when a change happens in Agora.
Although they are in devel (and hence, subject to laxer rules),
we can easily avoid any breakage by adding this check to the CI,
as the check is extremely cheap.
```

As discussed with @hewison-chris yesterday.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -140,7 +140,10 @@ jobs:
 
     - name: Test example configuration file
       if: matrix.run_extra_checks && runner.os != 'Windows'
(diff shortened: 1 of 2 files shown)


hewison-chris (approved):
LGTM

hewison-chris:
It is already helping:-
```
devel/config-fullnode.yaml(37:4): registry.validators.email: Key is not a valid member of this section. There are 3 valid keys: authoritative, primary, soa
```

linked0 (approved):
LGTM

PR #2851, opened 2022-01-14
NetClient: Remove address/api from ctor, use merge

Geod24:
```
We want to move to supporting connection-less NetworkClient,
e.g. so that we can have objects to represent missing nodes,
hence we need to first remove the dependency on an existing
connection from the constructor.
```

Part of https://github.com/bosagora/agora/pull/2845


hewison-chris (approved):
LGTM

linked0 (approved):

PR #2854, opened 2022-01-14
NetworkManager: Store nodes identity directly in `NetworkClient`

Geod24:
One step towards eliminating `NodeConnInfo`.
Part of https://github.com/bosagora/agora/pull/2845

### DIFF
--- a/source/agora/network/Client.d
+++ b/source/agora/network/Client.d
@@ -153,6 +153,9 @@ public class NetworkClient
     /// Logger uniquely identifying this client
     private Logger log;
 
+    /// The identity associated with this node - may be empty
+    private Identity identity_;
+
     /// Gossip delay
     private enum GossipDelay = 10.msecs;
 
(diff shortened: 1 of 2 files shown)


omerfirmak (approved):

source/agora/network/Client.d
```diff
@@ -146,6 +146,9 @@ public class NetworkClient
     /// Logger uniquely identifying this client
     private Logger log;
 
+    /// The identity associated with this node - may be empty
+    private Identity identity_;
```
linked0, line 157: Can you explain the meaning of adding `_` in the name other than other member variables?

linked0 (approved):
LGTM

source/agora/network/Client.d
Geod24: Because otherwise it would conflict with the method that is named `identity`.

Geod24:
Looks to me like the CI failures are not spurious.

Geod24:
So, after some work on this, it's clear this approach here is not going to work because we currently accept any node that present us with a public key as a validator, which requires deeper/more functional changes. Closing and will open other PRs to address the issue.

PR #2855, opened 2022-01-14
RPC: Check if sending peer is banned

mkykadir:
RPC node ban was only checked during sending phase, now also receivers
will be checking for the banned senders.

### DIFF
--- a/source/agora/network/RPC.d
+++ b/source/agora/network/RPC.d
@@ -31,6 +31,7 @@ import ocean.text.convert.Formatter;
 import vibe.core.net;
 import vibe.core.connectionpool;
 import vibe.core.sync;
+import vibe.http.server : RejectConnectionPredicate;
 
 import std.traits;
 import core.stdc.stdio;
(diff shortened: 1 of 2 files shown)


source/agora/network/RPC.d
```diff
@@ -332,6 +333,9 @@ private class RPCConnection
     ///
     private TCPConnection conn;
 
+    /// Check if peer is banned on receive
+    private RejectConnectionPredicate isBannedDg;
```
omerfirmak: I don't think we need this. Since this is to reject incoming connections, only point we need this is `listenRPC`

source/agora/network/RPC.d
```diff
@@ -384,17 +389,19 @@ private class RPCConnection
...
 
 *******************************************************************************/
 
 public TCPListener listenRPC (API) (API impl, string address, ushort port,
-    Duration timeout, void delegate (agora.api.Validator.API api) @safe nothrow discoverFromClient)
+    Duration timeout, void delegate (agora.api.Validator.API api) @safe nothrow discoverFromClient,
+    RejectConnectionPredicate isBannedDg)
 {
     auto callback = (TCPConnection stream) @safe nothrow {
         try stream.readTimeout = timeout;
```
omerfirmak, line 405: Isnt this the right place to reject them?

source/agora/network/RPC.d
```diff
@@ -384,13 +385,18 @@ private class RPCConnection
...
 
 *******************************************************************************/
 
 public TCPListener listenRPC (API) (API impl, string address, ushort port,
-    Duration timeout, void delegate (agora.api.Validator.API api) @safe nothrow discoverFromClient)
+    Duration timeout, void delegate (agora.api.Validator.API api) @safe nothrow discoverFromClient,
+    RejectConnectionPredicate isBannedDg)
 {
     auto callback = (TCPConnection stream) @safe nothrow {
+        if (isBannedDg(stream.remoteAddress))
```
omerfirmak, line 397: Can we log this? So that we dont pull our hair out when we cant figure out why we are dropping connections.

mkykadir: Trace or Debug?

omerfirmak (approved):

PR #2856, opened 2022-01-14
Test: Add missing cycle seed for NODE10

omerfirmak:
Used in https://github.com/bosagora/agora/pull/2839/files#diff-9907a82d8ed41d8ddaa0f6c18006c1b3151985618f52a3e172440dd2c5562bc5R1757


source/agora/consensus/PreImage.d
```diff
@@ -537,6 +537,8 @@ public void getCycleSeed (in KeyPair pair, in uint validator_cycle, out Hash see
         seed = Hash(`0x6917a265e852182cf7a309378905ccc23a9f05967c1acec3854a9a54ebc59e75a3beea61443d04de90866595b8c710e89734f2d5bfa1a215461380eeb5db6c4a`);
     else if (pair == NODE7)
         seed = Hash(`0xa656589184be1055d12e60821ccf933023665db1903892b5814c6bbcc0e3bcd47067b6c4a6a2241c1f8e174bf417dc11edf03715ab9c50c7923500cf987fc717`);
+    else if (pair == NODE10)
+        seed = Hash(`0xfe425761e0aee608ca3fbad5a358bb19d0e29758257849ec002f38390cf37e118c2868410f29ff71b3e5e377b8f7b5b42b10411e171526d368b3c4257d46bdd3`);
```
Geod24, line 541: Shouldn't the other WK nodes also be added ?

omerfirmak: If they were needed, it would have let us know. We can add them lazily as they are needed.

PR #2858, opened 2022-01-17
Trivial: Rename TestUTXOSet to MemoryUTXOSet

Geod24:
As it is already used for non-testing purpose.


linked0 (approved):
LGTM

omerfirmak:
Faucet needs an update as well, https://github.com/bosagora/agora/runs/4836349900?check_suite_focus=true#step:7:38

Geod24:
Rebased, should go green now that Faucet doesn't use it anymore.

PR #2859, opened 2022-01-17
Registry: Accept staking nodes even if they are not enrolled

Geod24:
One of the issue we are seeing live is that non-enrolled validators cannot yet register, leading to jumpy DNS.
In order to avoid this issue, we should rely on the stake being present, as it's much more long-lived.

### DIFF
--- a/source/agora/consensus/state/Ledger.d
+++ b/source/agora/consensus/state/Ledger.d
@@ -587,6 +587,19 @@ public class Ledger
         }
     }
 
+    /// Returns: Whether the `utxo` can be used as a stake for a validator
+    public bool isStake (in Hash hash, in Output utxo) scope @safe
+    {
+        if (utxo.type != OutputType.Freeze)
+            return false;
+        if (utxo.value < Amount.MinFreezeAmount)
+            return false;
+        EnrollmentState last_enrollment;
+        if (this.validator_set.findRecentEnrollment(hash, last_enrollment))
+            return last_enrollment.slashed_height == 0;
+        return true;
+    }
+
     ///
     public Amount getPenaltyDeposit (Hash utxo) @safe nothrow
     {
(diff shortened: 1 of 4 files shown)


mkykadir (approved):

source/agora/consensus/state/Ledger.d
```diff
@@ -587,6 +587,19 @@ public class Ledger
         }
     }
 
+    /// Returns: Whether the `utxo` can be used as a stake for a validator
+    public bool isStake (in Hash hash, in Output utxo) scope @safe
```
omerfirmak, line 591: `getPenaltyDeposit() > 0` would be equivalent

omerfirmak (approved):

source/agora/consensus/state/Ledger.d
Geod24: I have questions about that name... It's what it seems to be doing but it is not at all clear what other role(s) it takes ?
Also this doesn't do the UTXO lookup.

source/agora/consensus/state/Ledger.d
```diff
@@ -1027,3 +1040,80 @@ unittest
...
+{
+    import agora.common.Set;
+    import std.typecons;
+
+    /// Type of data we keep track of
+    public struct Tracked
+    {
+        /// Hash of the UTXO
+        public Hash hash;
+        /// Output itself
```
linked0, line 1064: Need a line break.

source/agora/consensus/state/Ledger.d
```diff
@@ -1027,3 +1040,80 @@ unittest
...
+        tx.outputs.enumerate
+            .map!((in tup) => tuple(UTXO.getHash(txHash, tup.index), tup.value))
+            .each!((in hash, in utxo) => this.externalize(hash, utxo));
+    }
+
+    /// Process a single `Output`
+    public void externalize (in Hash hash, in Output output) @safe
+    {
+        if (this.include(hash, output))
+            this.data_.put(Tracked(hash, output));
```
linked0, line 1118: Isn't there any possibility that the `data_` would be ever-growing to some critical point?

source/agora/consensus/state/Ledger.d
omerfirmak: It has a history:) At first it was called `hasPenaltyDeposit` and returned a `bool`. But after I decided to let the melting TX spend the deposit, I needed a way to get how much the deposit was as well. Deposit is fixed and configured in `ConsensusParams` but our TX validation routine have no access to it.

source/agora/consensus/state/Ledger.d
Geod24: Fixed

source/agora/consensus/state/Ledger.d
Geod24: It's more likely that `UTXOSet` reach that threshold way before this struct does, as this is only a (likely small) subset of `UTXOSet`.

linked0: Right. But `updateUTXOCache` has the `remove` operation but it doesn't. So I was just a little worried about that.

linked0 (approved):
LGTM

source/agora/consensus/state/Ledger.d
Geod24: See line 1107

PR #2860, opened 2022-01-17
Instantiate the Engine directly in the Ledger

Geod24:
Since @omerfirmak suggested it...


omerfirmak (approved):

Geod24:
This should be the last touch in making the Faucet transition from `UTXOSet` to `Ledger` nice and easy.

Issue #1675, opened 2021-02-19
BanManager is incompatible with reverse-proxying and load balancing


Geod24:
If someone wished to setup a reverse proxy in front of Agora, it would probably come crashing as soon as the `BanManager` would be involved, as it would end up banning the proxy.
I suggest that we had a (configured) whitelist (we need a whitelisting mechanism for #1674 anyways) and handle https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For and https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded correctly.

mkykadir:
Suggested header fields are only for HTTP, since we're moving towards to a RPC (and DNS on registry) based communication those fields still would not be enough. There is a newer protocol called [Proxy Protocol](https://www.haproxy.com/blog/haproxy/proxy-protocol/) that we can start to support. This protocol will still work with HTTP communications since it covers TCP communication basically. 

In addition to that, we should provide a documentation about putting Agora behind a proxy to aid users about some configuration fields.

mkykadir:
PR #2863 implements a solution for RPC connections inside Agora. Since we have a HTTP interface available also [this](https://github.com/vibe-d/vibe.d/issues/2639) needs to be addressed first, because we're using HTTP interface through generated REST API. If [this](https://github.com/vibe-d/vibe-core/issues/307) also gets addressed, solution within RPC is not needed.

PR #2863, opened 2022-01-17
Proxy protocol for RPC

mkykadir:
Towards #1675 

[Proxy Protocol](http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt) is a protocol that provides a way to access actual client's addresses over TCP streams when Agora is behind reverse proxy/load balancer. This allows us to properly ban manage peers.

The protocol should actually be part of Vibe.d's underlying stream implementation. Until then, this implementation can help us. The protocol actually has more recent version 2 but Nginx and most of the applications (like curl) doesn't support it yet. 

Needs testing, will try soon; let's have discussion on it.

### DIFF
--- a/source/agora/network/RPC.d
+++ b/source/agora/network/RPC.d
@@ -33,11 +33,90 @@ import vibe.core.connectionpool;
 import vibe.core.sync;
 import vibe.http.server : RejectConnectionPredicate;
 
+import std.conv : to;
+import std.socket : Address, AddressFamily, parseAddress;
 import std.traits;
 import core.stdc.stdio;
 import core.time;
 
 mixin AddLogger!();
+
+struct ProxyProtocol
+{
+    /// Protocol v2 signature
+    private immutable ubyte[12] v2_signature = [
+        13, 10, 13, 10, 0, 13, 10, 81, 85, 73, 84, 10
+    ];
+
+    /// Source address
+    Address src;
+
+    /// Destination address
+    Address dst;
+
+    ///
+    public this (TCPConnection stream) @trusted
+    {
+        ubyte[108] buffer;
+
+        stream.read(buffer[0 .. 12]); // Read signature
+
+        ensure(buffer[0 .. 12] != v2_signature,
+            "Proxy Protocol v2 is not yet supported");
+
+        // Read into buffer starting from `last_ix` until `ch`
+        uint readUntil (uint last_ix, char ch)
+        {
+            uint inner_ix = last_ix;
+            ubyte[1] mb = ['\0'];
+
+            while (cast(char) mb[0] != ch)
+            {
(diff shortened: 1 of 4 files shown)


source/agora/network/RPC.d
```diff
@@ -33,12 +33,91 @@ import vibe.core.connectionpool;
...
+
+    /// Destination address
+    Address dst;
+
+    ///
+    public this (TCPConnection stream) @trusted
+    {
+        ubyte[108] buffer;
+
+        stream.read(buffer[0..12]); // Read signature
```
mkykadir: Actually, I don't like reading stream data during ctor'ing a struct. But this way we can drop connections early, see `ensure`s below.

source/agora/network/RPC.d
```diff
@@ -33,12 +33,91 @@ import vibe.core.connectionpool;
...
+
+        ensure(buffer[0..12] != v2_signature,
+            "Proxy Protocol v2 is not yet supported");
+
+        // Read into buffer starting from `last_ix` until `ch`
+        uint readUntil (uint last_ix, char ch)
```
mkykadir, line 68: Vibe.d really needs [this stuff](https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readline), noted

source/agora/network/RPC.d
linked0:
```suggestion
        stream.read(buffer[0 .. 12]); // Read signature
```

linked0 (approved):
LGTM

mkykadir:
Tested and it works!
```RPC connection through ProxyProtocol: 192.XX.X.X1:5XX69```

Ready to review, fixed a little bug.

File: source/agora/node/Config.d [hunk 3]
```diff
@@ -174,6 +174,9 @@ public struct InterfaceConfig
     /// Bind port
     public ushort port;
 
+    /// Proxy Protocol V1
+    public bool proxy_proto;
```
omerfirmak, line 178: We probably could get away without configuring this explicitly. `RPCConnection` could easily eavesdrop on the reads and extract any ProxyProtocol data that may be in the buffers. Overriding `read` & `remoteAddress` could be enough IMO.

but like you said, this should be part of vibe.d, not in agora. Do we need this immediately? maybe we can hold on until we can implement this in vibed.

mkykadir: Protocol states
```
The receiver MUST be configured to only receive the protocol described in this
specification and MUST not try to guess whether the protocol header is present
or not.
```

omerfirmak: Rebase, then we can merge

mkykadir: Rebased

PR #2872, opened 2022-01-18
Test: Implement LocalRest DNSResolver

omerfirmak:


source/agora/network/DNSResolver.d
```diff
@@ -68,8 +65,7 @@ public final class DNSResolver
 
     public this (const Address[] peers = DefaultDNS)
     {
-        this.log = Log.lookup(__MODULE__);
-        this.log.enableConsole();
+        super();
```
Geod24: This is implicit, no need to call it explicitly.

source/agora/network/DNSResolver.d
```diff
@@ -144,6 +128,48 @@ public final class DNSResolver
...
+    ***************************************************************************/
+
+    public this ()
+    {
+        this.log = Log.lookup(__MODULE__);
+        this.log.enableConsole();
+    }
+
+    ///
+    public abstract ResourceRecord[] query (const(char)[] name, QTYPE type = QTYPE.ALL) @trusted;
```
Geod24: `@safe`, it doesn't make sense to mark an `abstract` method as `@trusted`

source/agora/network/DNSResolver.d
```diff
@@ -144,6 +128,48 @@ public final class DNSResolver
...
+
+    ///
+    public Message buildQuery (const(char)[] name, QTYPE type = QTYPE.ALL) @safe
```
Geod24: Should this be `protected` ?

source/agora/test/Base.d
```diff
@@ -2413,6 +2417,13 @@ public class NoGossipTransactionRelayer : TransactionRelayer
 /// Interface combining both the `TestAPI` and the `NameRegistryAPI` to expose both at once
 package interface FullRegistryAPI : TestAPI, NameRegistryAPI {}
 
+/// Mimics a DNS UDP Socket
+struct DNSQuery
```
Geod24: `private`

source/agora/test/Base.d
```diff
@@ -2413,6 +2417,13 @@ public class NoGossipTransactionRelayer : TransactionRelayer
...
+{
+    Message msg;
+    Channel!Message response_chan;
```
Geod24: Style :)

source/agora/test/Base.d
```diff
@@ -2447,3 +2458,55 @@ public class RegistryNode : TestFullNode, FullRegistryAPI
         return this.registry.postFlashNode(registry_payload, channel);
     }
 }
+
+///
+public final class LocalRestDNSResolver : DNSResolver
+{
+    ///
+    Channel!DNSQuery dns_chan;
```
Geod24: `private`

source/agora/test/Base.d
```diff
@@ -2447,3 +2458,55 @@ public class RegistryNode : TestFullNode, FullRegistryAPI
...
+        Instantiate a new object of this type
+
+        Params:
+          dns_chan = Channel that DNS server listens to
+
+    ***************************************************************************/
+
+    public this (Channel!DNSQuery dns_chan)
+    {
+        super();
```
Geod24: Implicit

source/agora/test/Base.d
```diff
@@ -2447,3 +2458,55 @@ public class RegistryNode : TestFullNode, FullRegistryAPI
...
+
+        DNSQuery q = DNSQuery(msg, new Channel!Message());
+        ensure(this.dns_chan.write(q), "DNS write channel closed");
+        Message answer;
+        ensure(q.response_chan.read(answer, 5.seconds), "Failed to get DNS response");
+        log.trace("Got response from for '{}'({}): {}", name, type, answer);
+        if (answer.header.RCODE == Header.RCode.NoError)
+            return answer.answers;
+        return null;
+    }
```
Geod24: I think you need to setup the listener too, but I don't see it done in that commit ?
Also, considering the code duplication, perhaps an approach where `query` is not the method overriden, but a new `sendMessage` (which just does network communication) is would make more sense ?

Geod24 (changes requested):

source/agora/test/Base.d
omerfirmak: Pushed new commits for the listener, but `query` already only does network communication as it is, there is no DNS logic there.

Geod24: Because most got moved to `buildQuery`. What's left is logging and the `RCODE` check.

Geod24: But it's not a blocker to move forward with this PR, just something that came to mind while reading the code.

omerfirmak: I will add a small commit to eliminate the duplicate buildQuery calls as well

source/agora/test/Base.d
```diff
@@ -1656,6 +1665,12 @@ private mixin template TestNodeMixin ()
         return super.makeEnrollmentManager();
     }
 
+    /// Returns an instance of a DNSResolver
+    protected override DNSResolver makeDNSResolver ()
+    {
+        return new LocalRestDNSResolver(this.dns_chan);
```
Geod24, line 1677: Why are we passing the `Channel` around instead of doing a registry lookup (Localrest registry, obv) ?

source/agora/test/Base.d
omerfirmak: Done

source/agora/node/FullNode.d
```diff
@@ -877,6 +878,20 @@ public class FullNode : API
...
+    /***************************************************************************
+
+        Returns an instance of a DNSResolver
+
+        Returns:
+            the dns resolver
+
+    ***************************************************************************/
+
+    protected DNSResolver makeDNSResolver ()
```
Geod24, line 875: Well technically it's not instantiated (looking at the commit subject) because this is never called.

source/agora/test/Base.d
```diff
@@ -2484,6 +2484,40 @@ public class RegistryNode : TestFullNode, FullRegistryAPI
...
+                this.registry.answerQuestions(
+                    query.msg, "peer.localrest",
+                    (in Message msg) @trusted { query.response_chan.write(msg.clone()); });
+            }
+            catch (Exception exc)
+            {
+                scope (failure) assert(0);
+                stderr.writeln("Exception thrown while handling query: ", exc);
+            }
+        }
```
Geod24, line 2527: Will this play nicely with `shutdown` ?

source/agora/test/DNS.d
```diff
@@ -0,0 +1,38 @@
...
+version (unittest):
+
+import agora.test.Base;
+import geod24.LocalRest;
+import geod24.concurrency;
+
+/// DNS test
+unittest
+{
+    import std.stdio;
```
Geod24: Leftover ?

source/agora/test/DNS.d
```diff
@@ -0,0 +1,38 @@
...
+    auto network = makeTestNetwork!TestAPIManager(conf);
+    network.start();
+    scope(exit) network.shutdown();
+    scope(failure) network.printLogs();
+    network.waitForDiscovery();
+
+    FiberScheduler scheduler = new FiberScheduler;
+    scheduler.start(
+        () {
+            auto addr = network.resolve(Address("agora://boa1xzval2a3cdxv28n6slr62wlczslk3juvk7cu05qt3z55ty2rlfqfc6egsh2.validators.unittest.bosagora.io"));
```
Geod24, line 33: Check the value that is returned ? Also we probably need a retry as the nodes might not have registered themselves even at the end of discovery (although that is unlikely).

Geod24 (changes requested):
Looks mostly

source/agora/test/Base.d
omerfirmak: Because `registry` only works for `Listener`s and that requires a `BindChn` which are only available for `spawn`d LocalRest threads (nodes). And LocalRest requires an interface

source/agora/test/Base.d
omerfirmak: Yep, it should. shutdown()  terminates the LocalRest thread, even if there are Fibers in queue

source/agora/test/Base.d
omerfirmak: Also, `FullNode`s dont have access to the registry client

Geod24 (approved):

Geod24:
I just need this to pass the test once before merging.

Geod24:
Something it seriously doesn't seem interested in.

omerfirmak:
> Something it seriously doesn't seem interested in.

I'll see if it is a regression

omerfirmak:
> > Something it seriously doesn't seem interested in.
> 
> I'll see if it is a regression

Should be good now. This is what was missing 

https://github.com/bosagora/agora/pull/2872/commits/383f77813aaac40df90f598b18eabf08196b9f05#diff-2cb6a9d42ca703fefdc67b1c1062e63c52c36cb15156b0e48d4cf36b9ace51e8R651

Issue #2868, opened 2022-01-18
Implement peer re-negotiation


Geod24:
It is possible for one of our peers to change stake:
https://github.com/bosagora/agora/blob/14459dd88a04797457a51f6e793d8b827b8d2d91/source/agora/node/Validator.d#L337-L339

For example, in the following set of events:
- Node starts with a pubkey, registers as validator because of #2867 ;
- A stake is created, the node picks it up and try to enroll (if `recurring_enrollment` is enabled);
- The stake might get destroyed, and another one created, and the node will pick it up, too;

At any point, the node might establish connections, meaning it can be seen from different clients with different, potentially outdated, stakes.
Currently there is no way to re-negotiate a connection (re-do the 'handshake'), nor is there a clean way to tear-down all connections and restart them.

Geod24:
To do this, we might have to dig into both the RPC client and Vibe.d.
It's acceptable to only implement it for the RPC client for now if necessary, but it will need to *also* work with Localrest.

The first step I was looking into was subclassing `RestInterfaceClient`, but it's only useful if we can have a persistent connection, which means storing the `HTTPClient`, which is not currently the case:
https://github.com/vibe-d/vibe.d/blob/b3818391b397616b20dd0c9b842e369b1e74a6d6/web/vibe/web/rest.d#L2051-L2053

Called from:
https://github.com/vibe-d/vibe.d/blob/b3818391b397616b20dd0c9b842e369b1e74a6d6/web/vibe/web/rest.d#L822-L824

Which is a method we could `override`, but currently we'd need to copy-paste a lot of code to make it work.

omerfirmak:
> nor is there a clean way to tear-down all connections and restart them.

I am wondering if LocalRest channels could be used to make this easier to implement, a dynamic set of peers.

PR #2877, opened 2022-01-18
Validator: Store and record change to stake

Geod24:
```
Currently, changes to stake are indirectly recorded and handshake may start
to return different result any time. With this change, we ensure that we
have a single point from which changes are performed, and tracked.
```

@omerfirmak : This was my foray into https://github.com/bosagora/agora/issues/2868 as well as https://github.com/vibe-d/vibe.d/pull/2638 and https://github.com/vibe-d/vibe.d/pull/2637 (with the underlying rationale that I wanted to break down `vibe.web.rest : request` so we could re-use it in an override of `RestInterfaceClient.request`).

If this doesn't make sense, feel free to close.

### DIFF
--- a/source/agora/node/Validator.d
+++ b/source/agora/node/Validator.d
@@ -82,6 +82,9 @@ public class Validator : FullNode, API
     /// but LocalScheduler is not instantiated yet.
     private bool started;
 
+    /// The identity we present on `handshake`
+    private Identity identity;
+
     /// Ctor
     public this (Config config)
     {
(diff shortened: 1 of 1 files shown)


Geod24:
Rebased

mkykadir (approved):

omerfirmak:
Well, the existing implementation was consistent IMO. A validator would always enroll with the stake that it presents in a `handshake` call. I didnt see the point but I held back reviewing this until I started working on #2868

omerfirmak:
I believe this introduced a bug

Some nodes are reporting missing peers although they are connected to all 32 nodes, verified thru `node_info`.

```
Jan 30 10:42:21 eu-005 agora[2436465]: 2022-01-30 10:42:19,603 Info [agora.network.Manager] - Doing periodic network discovery: 31 required peers requested, 6 missing, known 32
```

PR #2881, opened 2022-01-19
Ledger: Do not error/throw when re-adding the HEAD block (esp. Genesis)

Geod24:
```
When Faucet was switched over to use a Ledger,
it was found that trying to add genesis to the Ledger
results in an Exception being thrown.
While Faucet could be changed not to do that,
it seems like a common (yet harmless) error would be
to attempt 'acceptBlock' with the Ledger HEAD.
This is now handled more gracefully.
```

There's still an issue with the Ledger (it fails at the end of the cycle), but at least we get 20 blocks in, which is better than 0 :)

### DIFF
--- a/source/agora/consensus/state/Ledger.d
+++ b/source/agora/consensus/state/Ledger.d
@@ -162,6 +162,16 @@ public class Ledger
         }
     }
 
+    /// Create an instance of the Ledger suitable for unittests
+    version (unittest) public this ()
+    {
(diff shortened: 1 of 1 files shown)


linked0 (approved):
LGTM. The first sentence(`When Faucet was switched over the use a Ledger,`) of the commit message should be corrected if I'm not mistaken.

Geod24:
Fixed commit message.

PR #2883, opened 2022-01-19
Limit `Address` type to only accept schemes supported by Agora

### LINKED ISSUE #2882: Expected Address schema should be limited
Currently we can even have addresses starting with `ftp`, this will cause problems with the addition of the `URI` records #2869  for the registry. We should only expect addresses with schemes among `http`, `https` and `agora` within Agora.

mkykadir:
Fixes #2882 

Let's see if we have invalid scheme usages in tests first

### DIFF
--- a/source/agora/common/Types.d
+++ b/source/agora/common/Types.d
@@ -14,6 +14,7 @@
 module agora.common.Types;
 
 import agora.common.Amount;
+import agora.common.Ensure : ensure;
 import agora.crypto.ECC;
 import agora.crypto.Key;
 import agora.crypto.Schnorr;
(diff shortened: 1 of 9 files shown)


Geod24 (changes requested):
Need to swap both commits otherwise test won't pass on the first commit. Then feel free to self-merge.

Geod24:
Also this means we still accept `dns://` addresses for non-DNS situation (e.g. `network:`).

mkykadir:
> Also this means we still accept `dns://` addresses for non-DNS situation (e.g. `network:`).

Those can be validated through `Config`

mkykadir:
> Need to swap both commits otherwise test won't pass on the first commit. Then feel free to self-merge.

Swapped

omerfirmak:
We can lose the `tcp` from here as well then

https://github.com/omerfirmak/agora/commit/d9c2b82cb018d20e0f080f39d21627d908649883

omerfirmak (approved):
Still needs @Geod24 to approve

linked0 (approved):
LGTM

Geod24 (approved):

PR #2884, opened 2022-01-19
Ledger: Rename 'getBlockHeight' to just 'height' 

Geod24:
Got tired of the overly verbose name, and better to do it before `Faucet` depends on `Ledger.


omerfirmak (approved):

linked0 (approved):
Looks good.

PR #2885, opened 2022-01-19
Vibed URL: Account for port in opEquals

omerfirmak:
Otherwise Address("agora://host:port1") ==
	Address("agora://host:port2") would be `true`


mkykadir (approved):

linked0 (approved):
LGTM

PR #2887, opened 2022-01-19
Stats: Make address configurable

omerfirmak:
Before, it was listening on all interfaces. With native binaries
in production we should have a way to select the interface, docker
used to do that.


Geod24 (approved):

Geod24:
> source/faucet/main.d(549,25): Error: constructor `agora.stats.Server.StatsServer.this(string address, ushort port)` is not callable using argument types `(ushort)`
> source/faucet/main.d(549,25):        cannot pass argument `config.tx_generator.stats_port` of type `ushort` to parameter `string address`

Ah!

PR #2890, opened 2022-01-20
DNS: Fix `URRIRDATA` address serialization

mkykadir:
Default serialization of the `Address` adds length of the address
string as first byte which is not expected in URI record of the DNS.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -810,7 +810,11 @@ public struct ResourceRecord
                     tmp_data.soa = SOA.fromBinary!(SOA)(ctx);
                     break;
                 case TYPE.URI:
-                    tmp_data.uri = deserializeFull!(URIRDATA)(&ctx.read, ctx.options);
+                    tmp_data.uri = URIRDATA(
+                        deserializeFull!(typeof(URIRDATA.priority))(&ctx.read, ctx.options),
+                        deserializeFull!(typeof(URIRDATA.weight))(&ctx.read, ctx.options),
+                        Address(cast(string) ctx.read(rdlength - (URIRDATA.priority.sizeof + URIRDATA.weight.sizeof)))
+                    );
                     break;
                 default:
                     tmp_data.binary = cast(ubyte[]) ctx.read(rdlength);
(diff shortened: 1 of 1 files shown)


Geod24 (approved):

Geod24:
Wait, you need to override `fromBinary` too :D

mkykadir:
> Wait, you need to override `fromBinary` too :D

Since a context (`rdlength`) is needed deserialization is within `ResourceRecord` but serialization remained in `URIRDATA` itself which doesn't look right. Should I move the serialization also within `ResourceRecord`?

Geod24:
Let's do that

mkykadir:
Updated

source/agora/common/DNS.d
```diff
@@ -845,7 +849,11 @@ public struct ResourceRecord
                 case TYPE.SOA:
                     return this.rdata.soa.serializeFull(CompactMode.No);
                 case TYPE.URI:
-                    return this.rdata.uri.serializeFull(CompactMode.No);
+                    ubyte[] buffer;
+                    buffer ~= this.rdata.uri.priority.serializeFull(CompactMode.No);
+                    buffer ~= this.rdata.uri.weight.serializeFull(CompactMode.No);
+                    buffer ~= this.rdata.uri.target.toString().representation;
```
Geod24, line 855: So, this allocates, which we'd like to avoid. Can we avoid it ?
Taking a look at `URL` shows there's an overload that accepts an `OutputRange`:
https://github.com/vibe-d/vibe.d/blob/83753ac1f58e2b8abd26cca1f0730a67d09ace64/inet/vibe/inet/url.d#L381-L382

(Well to be fair [we put it here](https://github.com/vibe-d/vibe.d/commit/0bd793c7ee15483b5e0de8f8ec5a382f9e916119)).

So if we can wrap our `dg` in an `OutputRange`, we're good. Note that the `OutputRange` is a structural concept (like any range), hence there's many ways one can approach this. E.g.
```D
static struct ORWrapper
{
    private SerializeDg dg;
    
    public void put (char c) @trusted { dg((cast(ubyte*) &c)[0 .. 1]); }
    public void put (in char[] s) @safe { dg(s.representation); }
}
```

Geod24: Although... Now this comment doesn't make sense anymore, since the other places allocate too.

PR #2891, opened 2022-01-20
Move ChainID to ConsensusConfig

Geod24:
Fairly trivial.
BTW @omerfirmak I think the whole `setHashMagic` should have been better in Agora, instead of being in the library.


source/agora/consensus/data/Params.d
```diff
@@ -113,11 +113,16 @@ public immutable class ConsensusParams
...
 {
+    /// Time at which the Genesis block was created
     public ulong genesis_timestamp = 1609_459_200; // 2021-01-01:00:00:00 GMT
 
     /// The cycle length for a validator
     public uint validator_cycle = 1008;
 
+    /// Chain ID to be used while calculating hashes
+    /// Used to protect against replay attacks between networks
+    public ulong chain_id = 0; // 1: CoinNet, 1: TestNet, etc...
```
linked0: The values should be different for CoinNet and TestNet in the comment, I guess. And we should set the value for TestNet by default in unit tests.

Geod24: Fixed the comment. As for setting the value, I think we could do that in another PR, as this just moves stuff.
Looking at it again, wouldn't using `genesis_timestamp` work @omerfirmak ?

omerfirmak: Practically it should, but would be a PITA for us. That would invalidate genesis signatures every time we change it in production. And we change it pretty frequently.

Geod24:
> node-7_1  | 2022-01-21T05:51:42.994312143Z config.yaml(33:2): consensus.chain_id: Required key was not found in configuration of command line arguments

~Thank you Faucet for catching the mistake~ :D
Actually every node did.

omerfirmak (approved):

Issue #2869, opened 2022-01-18
Generate `SRV`/`URI` record on node registration


Geod24:
The registry needs to automatically generate `SRV` records when `postValidator` / `postFlashNode` are called, as defined in https://github.com/bosagora/agora/blob/14459dd88a04797457a51f6e793d8b827b8d2d91/doc/DNS.md#locating-another-node

mkykadir:
Will CNAME and A records still meaningful for us after this with the current Config fields (which expects `Address`)? Or, we might ease configuration field from `Address` to (i.e.) `string` and then work on that.

Geod24:
Depends on the way we want to go. I think we can eliminate `SRV` records because they don't support the path component of the HTTP interface. So provided we generate `URI` record, do we want to have it under `boa1....coinnet.validators.bosagora.io.` or `XXX.boa1...coinnet.validators.bosagora.io.`.
`XXX` could be anything, so we could use `_agora` to somewhat mimic the `SRV` approach. If we do this, then having `boa1...` be either a `CNAME` (if we're provided with a domain name) or an `A` record (if we're provided with an IP) can make sense.

Geod24:
Under this scheme, you would have the following:

1) `agora://192.168.0.1/`
    - `$KEY.$REALM` : A record (`192.168.0.1`);
    - `_agora.$KEY.$REALM` : URI record (`agora://$KEY.$REALM`);
2) `agora://best.validator.example.com:1234/api/`
    - `$KEY.$REALM` : CNAME record (`best.validator.example.com`);
    - `_agora.$KEY.$REALM` : URI record (`agora://$KEY.$REALM:1234/api/`);
    
The alternative is much simpler: Just store the `URI` that is provided. But it will not play nicely with other services (e.g. admin interface / votera).

mkykadir:
It seems like this will require some architectural changes over Registry. I think it is better to have #2933 beforehand.

PR #2894, opened 2022-01-20
DNS: Add getter for service and protocol labels

mkykadir:
`Domain` can include service and protocol labels; `LabelRange` is
a backward range; service and protocol labels are guaranteed to be in front.

Towards #2869

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -28,6 +28,7 @@ import std.bitmanip;
 import std.format;
 import std.range;
 import std.string;
+import std.typecons : Tuple;
 
 /*******************************************************************************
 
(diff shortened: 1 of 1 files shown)


omerfirmak (approved):

source/agora/common/DNS.d
```diff
@@ -1409,6 +1448,9 @@ unittest
                           "0123456789abcdef0123456789abcde." ~
                           "0123456789abcdef0123456789abcde." ~
                           "0123456789abcdef0123456789abcde")); // Final '.'
+
+    assert(Domain.fromSafeString("_agora._tcp.bosagora.io.").serviceLabel() == "agora");
+    assert(Domain.fromSafeString("_agora._tcp.bosagora.io.").protocolLabel() == "tcp");
```
omerfirmak: Could use some tests with no service/proto labels then feel free to merge

omerfirmak (approved):

source/agora/common/DNS.d
```diff
@@ -1107,6 +1108,35 @@ public struct Domain
...
+
+        Get service and protocol labels
+
+        Returns:
+          A tuple containing both service and protocol labels, empty `ServiceProtoTuple`
+          is returned when one of them is missing.
+
+    ***************************************************************************/
+
+    public ServiceProtoTuple serviceAndProtocolLabels () const @safe
```
omerfirmak: could this be `nothrow`?

source/agora/common/DNS.d
```diff
@@ -1107,6 +1108,40 @@ public struct Domain
...
+            splitted.popFront();
+            const proto = splitted.front();
+
+            if (proto[0] != '_')
+                return ServiceProtoTuple.init;
+
+            return ServiceProtoTuple(service[1 .. $], proto[1 .. $]);
+        }
+        catch (Exception)
+            return ServiceProtoTuple.init;
```
Geod24: The only thing that could throw is if UTF-8 decoding fails, so I would make it an `assert(0)`, unless I'm missing something ?

source/agora/common/DNS.d
```diff
@@ -1107,6 +1108,40 @@ public struct Domain
...
+    {
+        if (this.value[0] != '_')
+            return ServiceProtoTuple.init;
+
+        try
+        {
+            auto splitted = this.value.splitter('.');
+            const service = splitted.front();
+            splitted.popFront();
+            const proto = splitted.front();
```
Geod24, line 1134: That's not correct, you need to check if the range is `empty` at some point.

mkykadir: Since we always have a trailing `.` we always have an empty label; but this still misses check for empty label, added.

Geod24: Try "." and it still `assert`s.

Geod24 (approved):

PR #2895, opened 2022-01-20
Refactor: UTXO probing in TransactionPool

omerfirmak:
Existing implementation was buggy, had paths that could SEGV us.

### DIFF
--- a/source/agora/consensus/pool/Transaction.d
+++ b/source/agora/consensus/pool/Transaction.d
@@ -18,6 +18,7 @@ import agora.common.Amount;
 import agora.common.ManagedDatabase;
 import agora.common.Types;
 import agora.common.Set;
+import agora.consensus.data.UTXO;
 import agora.consensus.data.Transaction;
 import agora.consensus.data.genesis.Test;
 import agora.crypto.Hash;
@@ -94,6 +95,9 @@ public class TransactionPool
     /// A delegate to select one of the double spent TXs
     public DoubleSpentSelector selector;
 
+    /// UTXO set
+    protected Output[Hash] utxo_set;
+
     /***************************************************************************
 
         Params:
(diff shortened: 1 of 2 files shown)


source/agora/node/FullNode.d
```diff
@@ -1005,66 +1005,24 @@ public class FullNode : API
...
-            {
-                foreach (ref Hash hash, ref Transaction tx; this.pool)
-                {
-                    if (!tx.isFreeze())
-                        continue;
-
-                    foreach (size_t idx, output_; tx.outputs)
-                    {
-                        if (output_.type == OutputType.Freeze &&
-                            output_.value >= Amount.MinFreezeAmount)
```
mkykadir, line 1056: What about these checks?

omerfirmak: enrollment validation routine checks that anyways, so they were redundant

mkykadir (approved):

linked0 (approved):
LGTM. Thanks.

source/agora/consensus/pool/Transaction.d
```diff
@@ -570,6 +580,29 @@ public class TransactionPool
...
+
+    public bool peekUTXO (in Hash utxo, out UTXO value) nothrow @safe
+    {
+        if (auto output = utxo in this.utxo_set)
+        {
+            value.output = *output;
+            return true;
+        }
+        return false;
+    }
```
Geod24, line 605: Looks to me like we should just have another `UTXOCache` here ?

omerfirmak: That was my initial thought too, but to remove a UTXO we would either need to construct a dummy spending TX or expose the private `remove` UTXOCache has. Either way didnt feel right, so this.

I can change it to use UTXOCache if you are comfortable with one of those.

Geod24: That answers my question

PR #2897, opened 2022-01-21
Move simulatePreimages to utils.test

Geod24:


Geod24:
Now with added süß.

source/agora/utils/Test.d
```diff
@@ -442,3 +444,34 @@ unittest
...
+                     This can be used to test slashing.
+
+*******************************************************************************/
+
+public void simulatePreimages (Ledger ledger, in Height height,
+                               uint[] skip_indexes = null) @safe
+{
+    ledger.getValidators(height).enumerate.each!((idx, val)
+    {
+        if (skip_indexes.length == 0 || !skip_indexes.canFind(idx))
```
linked0:
```suggestion
        if (!skip_indexes.canFind(idx))
```
I think there is some reason to add the condition more though.

linked0 (approved):
LGTM

source/agora/utils/Test.d
Geod24: Added as a separate commit to avoid moving and modifying.

PR #2903, opened 2022-01-24
Update config dependency to configy

Geod24:


source/agora/node/main.d
```diff
@@ -133,27 +133,13 @@ private int main (string[] args)
         return runEventLoop();
     }
 
-    Nullable!Config config = ()
+    Nullable!Config configN = ()
```
hewison-chris, line 136: What does the `N` represent here?

Geod24: `Nullable`

hewison-chris (approved):
LGTM

Issue #2425, opened 2021-08-27
Flash unittests sometimes crashes


Geod24:
```
Unittesting agora.test.Flash#### FATAL ERROR: Connection with peer closed
This node was started at source/agora/test/Flash.d:399
This most likely means that the node crashed due to an uncaught exception
If not, please file a bug at https://github.com/Geod24/localrest/
Full error: object.Exception@submodules/localrest/source/geod24/LocalRest.d-mixin-1179(1190): Connection with peer closed
----------------
submodules/localrest/source/geod24/LocalRest.d-mixin-1179:1190 @trusted geod24.LocalRest.Response geod24.LocalRest.RemoteAPI!(agora.test.Flash.TestFlashAPI, agora.test.Base.Serializer).RemoteAPI.confirmChannelUpdate(agora.crypto.Key.PublicKey, agora.crypto.Key.PublicKey, geod24.bitblob.BitBlob!(64uL).BitBlob, uint).__lambda6() [0x10f28d091]
submodules/localrest/source/geod24/LocalRest.d-mixin-1179:1185 @safe agora.flash.Types.Result!(bool).Result geod24.LocalRest.RemoteAPI!(agora.test.Flash.TestFlashAPI, agora.test.Base.Serializer).RemoteAPI.confirmChannelUpdate(agora.crypto.Key.PublicKey, agora.crypto.Key.PublicKey, geod24.bitblob.BitBlob!(64uL).BitBlob, uint) [0x10f28cb33]
submodules/localrest/source/geod24/LocalRest.d-mixin-1179:1180 _DThn16_6geod249LocalRest__T9RemoteAPITC5agora4test5Flash12TestFlashAPITSQBhQBe4Base10SerializerZQCr20confirmChannelUpdateMFNfSQDj6crypto3Key9PublicKeyQzSQFq7bitblob__T7BitBlobVmi64ZQokZSQFr5flash5Types__T6ResultTbZQk [0x10f28e505]
[...]

PR #2906, opened 2022-01-24
Add log for the result of Flash transactions

linked0:
This is needed to analyze the failure of the operations on Flash channels.

Part of #2425


hewison-chris (approved):
LGTM

PR #2912, opened 2022-01-25
Trivial: QoL refactors

omerfirmak:


mkykadir (approved):

linked0 (approved):
LGTM

doc/config.example.yaml
```diff
@@ -211,7 +211,7 @@ network:
   - http://192.168.0.44:2828
 dns:
   # Supported value: FQDN seed
-  - seed.bosagora.io
+  - dns://seed.bosagora.io
```
Geod24: I'm not sure that's a QoL improvement. For `network`, it makes sense, since there can be multiple IPs, but for this field, it can only ever be `<domain-or-ip>[:<port>]?`

source/agora/node/Config.d
```diff
@@ -100,11 +100,11 @@ public struct Config
     public RegistryConfig registry;
 
     /// The list of IPs for use with network discovery
-    public @Optional immutable string[] network;
+    public @Optional immutable Address[] network;
 
     /// The list of DNS FQDN seeds for use with network discovery
     @Name("dns")
-    public @Optional immutable string[] dns_seeds;
+    public @Optional immutable Address[] dns_seeds;
```
Geod24: We should nuke this field, eventually, so I'm less fussed about it being changed to `Address` if it simplifies your life.

doc/config.example.yaml
omerfirmak: I only care about the `network` config actually, will revert

PR #2913, opened 2022-01-25
Manager: Move setIdentity call to onHandshakeComplete

omerfirmak:
The newly created client could be merged to an existing one in
onHandshakeComplete, by moving the setIdentity call (which logs the event)
we are making sure that temporary clients dont spam logs


hewison-chris (approved):
LGTM

PR #2915, opened 2022-01-26
ManagedDatabase: Do not begin/commit/rollback all DBs at once 

Geod24:
Was looking more into `ManagedDatabase` and the pain it brings and realized that it should/could go.
The reason we needed it was due to the multiple `Database` instances we used to hold, which is not the case anymore.
So first step is to get rid of those wrappers, and the next thing to take care of will be finalization. That's why I lumped in the second commit, as we will need to properly shut-down before we can remove our reliance on the module dtor.


linked0 (approved):
LGTM

omerfirmak (approved):

Geod24:
```
----------------
```

PR #2916, opened 2022-01-26
Test: Fix registry network config

omerfirmak:
```
It depended on manager's node list, which is now populated
after the registry is created.
```

See https://github.com/bosagora/agora/commit/209ad7a225d7e57afaf280cbe2c0fa44c8ba03bc

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -2276,7 +2276,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
         .map!(en => validatorAddress(en.index, en.value)).array;
 
     auto full_node_addresses = test_conf.full_nodes.iota.map!(
-        idx => fullNodeAddress(idx)).array ~ [Address("http://name.registry")];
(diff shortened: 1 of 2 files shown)


source/agora/test/Base.d
```diff
@@ -2320,7 +2320,7 @@ public APIManager makeTestNetwork (APIManager : TestAPIManager = TestAPIManager)
       node : makeNodeConfig(),
       interfaces: [ makeInterfaceConfig("name.registry") ],
       consensus: makeConsensusConfig(),
-      network : net.nodes.map!(pair => Address("http://"~pair.address)).array.idup,
+      network : connect_addresses.filter!(addr => addr != Address("http://name.registry")).array.idup,
```
Geod24: Or we could just remove the registry from `full_node_addresses` (L2278)

omerfirmak: Let me see if that would break anything

omerfirmak: Done

Geod24:
Passes on MacOS.

Geod24 (approved):

PR #2917, opened 2022-01-26
Provide more information when a Flash node registration is rejected

Geod24:
Started to fail locally, I assume @omerfirmak 's fix #2916 will fix it, but at least we can see it working :)

### DIFF
--- a/source/agora/flash/Node.d
+++ b/source/agora/flash/Node.d
@@ -1248,7 +1248,6 @@ public class FlashNode : FlashControlAPI
             gen_hash        : this.genesis_hash,
             funder_pk       : funding_utxo.output.address,
             peer_pk         : recv_pk,
-            pair_pk         : funding_utxo.output.address + recv_pk,
             num_peers       : num_peers,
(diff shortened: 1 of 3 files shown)


source/agora/flash/Node.d
```diff
@@ -775,10 +775,19 @@ public class FlashNode : FlashControlAPI
                         open.conf.chan_id.flashPrettify);
 
                 try
-                    if (!isValidChannelOpen(open.conf, this.getBlock(open.height)))
+                    if (auto err = open.conf.isNotValidOpenReason(this.getBlock(open.height).txs))
+                    {
+                        log.warn("Skipping gossipping of {} because is is not valid/open: {}",
```
linked0:
```suggestion
                        log.warn("Skipping gossipping of {} because is not valid/open: {}",
```

linked0 (approved):
LGTM, one type.

source/agora/flash/Node.d
Geod24: Fixed

PR #2918, opened 2022-01-26
Tests: Compare block by hashes, not identity

Geod24:
```
It's possible that some validators are behind / a bit slow,
and miss a signature or more, which is completely fine as
long as they have a majority of them.
```

Seen this one trigger in one of my PR.


linked0 (approved):
LGTM

Geod24:
Both Ubuntu and MacOS went green first try. Either the stars are aligned or this was one of the main issue we had in the CI.

PR #2919, opened 2022-01-26
Manager: Allow ConnectionTask to be spawned with just a PublicKey

omerfirmak:
```
Which itself will do all the Registry/DNS queries
```

This will prevent overwhelming the registry


Geod24:
Shouldn't the connection tasks eventually be merged into `NetworkClient` ?

omerfirmak:
> Shouldn't the connection tasks eventually be merged into `NetworkClient` ?

It could be yes, ideally it should.

PR #2920, opened 2022-01-26
Simplify / Improve some Flash test, give an IP to the registry in unittests

Geod24:
The end goal was to move all nodes to have an IP, the registry seemed like the best place to start.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -892,7 +892,7 @@ public class TestAPIManager
 
         The name registry is a full node that exposes a more advanced interface.
         It is not part of the `nodes` array, but is reachable as a designed
-        address (currently "name.registry").
+        address (currently "10.8.8.8").
 
         Params:
           conf = The `FullNode` configuration to use.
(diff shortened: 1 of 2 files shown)


source/agora/test/Base.d
```diff
@@ -892,7 +892,7 @@ public class TestAPIManager
 
         The name registry is a full node that exposes a more advanced interface.
         It is not part of the `nodes` array, but is reachable as a designed
-        address (currently "name.registry").
+        address (currently "10.8.8.8").
```
omerfirmak, line 895: https://github.com/bosagora/agora/pull/2900/commits/f4436f3863929908a9ee83d39fa2c8f0ef1ca836 was setting it to `8.8.8.8`

omerfirmak (approved):

source/agora/test/Base.d
Geod24: Yep, but I figured using the private IP range would be safer, *just in case*

PR #2922, opened 2022-01-26
RPC: ensure stream is connected before calling handle()

omerfirmak:
Fixes a segfault where connection is closed and we call
readTimeout on it.


omerfirmak:
Deployed, seem to be working fine.

Geod24 (approved):

PR #2924, opened 2022-01-26
Doc: Add documentation about reverse proxy configuration

### LINKED ISSUE #2837: Provide documentation about Agora behind Proxy
We should provide a documentation about putting Agora behind a proxy to aid users about some configuration fields.

_Originally posted by @mkykadir in https://github.com/bosagora/agora/issues/1675#issuecomment-1009744633_

Main idea here is users can have several different reverse proxy (or load balancer) configurations in front of the Agora and this might prevent Agora to detect actual peer addresses. There are several applications (Forwarded for HTTP headers and [Proxy Protocol](https://github.com/bosagora/agora/blob/7fdff995445b6f1ba695fe9e46500c75f00f1eca/source/agora/common/Ensure.d#L111) for TCP/UDP streams) to inform applications behind the proxy about the actual addresses. After Agora starts to support those features, we will need to provide a documentation about supported features and configuration of them. 

Which Proxy Protocol document also states;

> Implementers should be very careful about not trying to automatically detect
whether they have to decode the header or not, but rather they must only rely
on a configuration parameter.

mkykadir:
Fixes #2837

### DIFF
--- a/doc/Proxy.md
+++ b/doc/Proxy.md
@@ -1,0 +1,42 @@
+## Agora behind a reverse proxy (Incoming connections)
+
+Agora communicates with other nodes through its configured `interfaces` which can be
+either HTTP or RPC (TCP) based. Some configurations might require putting an Agora
+node behind a reverse proxy. However, an Agora node depends on a ban manager to
+temporarily disable communications with bad-behaving nodes. Those nodes are
+banned through their public addresses. An Agora node cannot know public address
+of a peer when it is behind a reverse proxy with a default configuration.
+
+Agora supports [Proxy Protocol V1](http://www.haproxy.org/download/1.8/doc/proxy-protocol.txt)
+for RPC interfaces. Proxy Protocol allows an Agora node to know public address
+of the peer even it is behind a reverse proxy.
+
+Only RPC interfaces and Proxy Protocol (V1) is supported. Preferred reverse proxy
+implementation must support proxy for TCP streams and Proxy Protocol (V1).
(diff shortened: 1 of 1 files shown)


doc/Proxy.md
```diff
@@ -0,0 +1,42 @@
+## Agora behind a reverse proxy (Incoming connections)
+
+Agora communicates other nodes through its configured `interfaces` which can be
```
Geod24: with other nodes*

doc/Proxy.md
```diff
@@ -0,0 +1,42 @@
...
+either HTTP or RPC (TCP) based. Some configurations might require putting an Agora
+node behind a reverse proxy. An Agora node utilizes ban manager to temporarily
```
Geod24: "However, and Agora node depends on a ban manager to [...]"

Since we're presenting the way two features conflict, the "however" serves to underline the contradiction.

doc/Proxy.md
```diff
@@ -0,0 +1,42 @@
...
+stream {
+    server {
+        listen <ip>:<port>;
+        proxy_pass 127.0.0.1:<agora_tcp_port>;
+        proxy_protocol on;
+    }
+}
+```
+
+For more information, check [Nginx documentation.](https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/)
```
Geod24: ~check~ "see the" or "refer to".

Geod24 (approved):
LGTM just a couple nits

PR #2925, opened 2022-01-27
Fix flash unittest crashes on CI

linked0:
There was some situation where a necessary `update tx` does not reach 
validators because some block for the UTXO that an `update tx` refers to
has not yet been externalized before the tx is received. It is because the
unittests in the Flash make blocks without transactions so a block can be
externalized with the necessary txs. So we remove the cause of missing txs.

Relates to #2425 
We can close the issue if clashes on the unittests do not appear for a while.

### DIFF
--- a/source/agora/test/Flash.d
+++ b/source/agora/test/Flash.d
@@ -712,7 +712,7 @@ unittest
 
     // at this point charlie will automatically publish the latest update tx
     // and then a settlement will be published (but only after time lock expires)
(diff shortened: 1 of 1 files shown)


linked0:
This PR is very simple and has no side effects on other codes. So we can merge this if the CI tests pass and see if the crashes on Flash disappear on many other PRs. I hope the tests pass at least in this PR.

hewison-chris (approved):
LGTM

PR #2927, opened 2022-01-27
Flash test: Limit number of handlers to number of nodes

Geod24:
This would not lead to failures, but probably delays the nodes as their
requests will fail.


linked0 (approved):
LGTM

PR #2928, opened 2022-01-27
NetworkClient: Upgrade final message of client failure to 'warn'

Geod24:
Because when we look at the logs, it is drowned in a sea of other 'info' messages,
but can be quite useful for us to know why a test failed.


linked0 (approved):
LGTM, some fix is needed in the commit message, I guess.

Geod24:
Fixed message.

hewison-chris (approved):
LGTM

PR #2934, opened 2022-01-27
DNS: Remove CNAME limitation from resolver

mkykadir:
`resolveHost` is capable of resolving address of a name and this limitation
prevents users from providing name server names (i.e. `ns1.bosagora.io`)
in `query_servers` field.

### DIFF
--- a/source/agora/network/DNSResolver.d
+++ b/source/agora/network/DNSResolver.d
@@ -71,10 +71,6 @@ public final class VibeDNSResolver : DNSResolver
         // underlying address.
         foreach (idx, ref res; this.resolvers)
         {
-            assert(peers[idx].host.guessAddressType != TYPE.CNAME,
-                   "Address '" ~ peers[idx].host ~
-                   "' is not an IP address and can't be used for DNS resolution");
-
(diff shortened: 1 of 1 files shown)


Geod24:
`resolveName` is a bit of an issue, because it will block if it needs to resolve a name.
The parameters provided to the constructor should be seen like the `nameserver` in `/etc/resolv.conf`.
One way to improve this would be to actually read and parse `resolv.conf`, but I thought it was too involved.
Another would be to have an `addResolver` that doesn't have this limitation (because it already has existing resolvers).

Geod24:
@mkykadir : What should we do with this PR ?

mkykadir:
> @mkykadir : What should we do with this PR ?

I agree with your previous comment but name resolution only happens during construction. DNSResolver is used during starting up Registry, thus I don't know if blocking will be a big problem for us. But I can add a comment to ctor if we go this way. It is not a mandatory change but makes configuration easier.

Geod24 (approved):
Something I haven't considered: The host name could be in `/etc/hosts` or on a network resolver. We have the case with the integration tests. So this actually should be removed.

PR #2937, opened 2022-01-28
Always use IP addresses in tests (+ some minor improvements)

Geod24:
Last row of "random name" => IP address conversion.


linked0 (approved):
LGTM

linked0:
If the failure is repeating on `ValidatorCleanRestart`, we could merge this.

PR #2939, opened 2022-01-28
Registry: Add a function to register a validator without checking signature

Geod24:
```
Since we want to add the result of our lookups, and our lookups won't get us a signature,
we need to be able to bypass it from within Agora.
```

First part of https://github.com/bosagora/agora/pull/2900

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -145,10 +145,6 @@ public class NameRegistry: NameRegistryAPI
     protected TYPE ensureValidPayload (in RegistryPayload payload,
         TypedPayload previous) @safe
     {
-        // verify signature
-        ensure(payload.verifySignature(payload.data.public_key),
-                "Incorrect signature for payload");
-
         // check if we received stale data
         if (previous != TypedPayload.init)
             ensure(previous.payload.data.seq <= payload.data.seq,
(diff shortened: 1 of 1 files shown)


mkykadir (approved):

source/agora/node/Registry.d
```diff
@@ -235,6 +231,14 @@ public class NameRegistry: NameRegistryAPI
         ensure(this.zones[ZoneIndex.Validator].type != ZoneType.caching,
             "Couldn't register, server is not authoritative for the zone");
 
+        ensure(registry_payload.verifySignature(registry_payload.data.public_key),
+               "Incorrect signature for payload");
+        this.registerValidator(registry_payload);
+    }
+
+    /// Similar to `postValidator`, but does not perform signature verification
+    public void registerValidator (RegistryPayload registry_payload) @safe
```
omerfirmak, line 240: We should be careful not to return the records that we obtained this way, as a response to an external request. I am not sure if that is possible, but just came to my mind.

omerfirmak (approved):

source/agora/node/Registry.d
Geod24: Hum, why ? If we're caching, anything goes. If we're authoritative, either this shouldn't get called (there's no upstream) or our upstream should be another authoritative server.

omerfirmak: I wasnt sure if that was possible in practice anyways.

> our upstream should be another authoritative server.

if they are to be trusted, I dont see a problem either

PR #2944, opened 2022-01-31
Nominator: Nominate until a deadline

omerfirmak:
Nominator will nominate new values only for the first half of the
block interval, to allocate rest of the interval for Ballot protocol
to terminate.

We also ensure that each node nominates atleast a single value for
each slot, regardless of the time.


mkykadir (approved):

omerfirmak:
This didnt work quite as expected. Convering to draft

hewison-chris:
I think this was a good idea. However in the `network unit tests` we should probably advance the `mocked Clock` to a time after the halfway point during the test to benefit from this.

PR #2946, opened 2022-01-31
DNS: Remove buffer limit of answer through TCP

### LINKED ISSUE #2935: DNS TCP buffer is not enough for AXFR zone transfer
Currently we're [using](https://github.com/bosagora/agora/blob/f55909e30fc22df2be88acb1ce777690b34b5bfc/source/agora/node/Runner.d#L395) 4096 bytes buffer from stack, which might not be enough for zone transfer when we have increasing number of records (current test condition is just 32 records).

Buffer might needed to be increased or zone transfers happen through IXFR.

mkykadir:
Fixes #2935 

This solution **allocates**, previous implementation was better for segmented TCP stream (DG is invoked per ResourceRecord) but segmented messaging still need to put whole message length to first message.

### DIFF
--- a/source/agora/node/Runner.d
+++ b/source/agora/node/Runner.d
@@ -393,14 +393,6 @@ private void runTCPDNSServer (TCPConnection conn, NameRegistry registry) @truste
 private void runTCPDNSServer_canThrow (TCPConnection conn, NameRegistry registry) @trusted
 {
     ubyte[4096] buffer;
-    ushort length = 2;
-    scope writer = (in ubyte[] data) @safe {
-        ensure(data.length <= (buffer.length - length),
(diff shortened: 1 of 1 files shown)


source/agora/node/Runner.d
```diff
@@ -422,15 +414,15 @@ private void runTCPDNSServer_canThrow (TCPConnection conn, NameRegistry registry
         auto query = deserializeFull!Message(buffer[0 .. size]);
         registry.answerQuestions(
             query, conn.remoteAddress.toAddressString(),
-            (in Message msg) @safe => msg.serializePart(writer, CompactMode.No),
+            (in Message msg) @trusted
+            {
+                import std.stdio : writeln;
```
Geod24: Leftover?

Geod24 (approved):

PR #2951, opened 2022-02-01
GeoIP: Update actions/cache to work on all OS

mkykadir:


Geod24:
You also need to define a matrix

omerfirmak (approved):

PR #2954, opened 2022-02-02
Registry: Increment SOA serial per register

### LINKED ISSUE #2953: Primary registry SOA serial update is not high resolution enough
[Current implementation](https://github.com/bosagora/agora/blob/ba7fb8fff4f6ece23bcbcca23dc5680452479728/source/agora/node/Registry.d#L1034) uses UNIX Epoch time as SOA serial number; which might not be high resolution enough. During DNS test implementations, since `LocalRest` is fairly fast, I've found out that SOA serial is not always updated when multiple different node registrations occur in same second; which prevents a secondary registry to detect zone changes and transfer new records. 

See following logs for an example;
```
2022-02-02 13:45:26,431 Info [agora.node.Registry] - Registering addresses TYPE.A: [agora://10.0.0.0/] for public key: boa1x**
2022-02-02 13:45:26,431 Info [agora.node.Registry] - Primary SOA updated 1643798726
2022-02-02 13:45:26,434 Info [agora.node.Registry] - Registering addresses TYPE.A: [agora://10.0.0.1/] for public key: boa1x**
2022-02-02 13:45:26,434 Info [agora.node.Registry] - Primary SOA updated 1643798726
```
here both registrations occur in same second and SOA serial is still same.

[RFC#1912](https://datatracker.ietf.org/doc/html/rfc1912#section-2.2) suggests a serial update method; which might be preferred.

mkykadir:
when registries happen in same second.

Just UNIX Epoch time is not high resolution enough for testing conditions
and some network conditions (i.e. both nodes register at same second).
RFC 1912 recommendation for serial format also is not high resolution enough.

Fixes #2953

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -635,7 +635,7 @@ private SOA fromConfig (in ZoneConfig zone, Domain name) @safe
     return SOA(
         // mname, rname
         Domain.fromString(zone.primary), Domain.fromString(zone.soa.email.value.replace('@', '.')),
-        0, // Serial is not a config value, Will be set through `updateSOA`
+        cast(uint) Clock.currTime(UTC()).toUnixTime(),
         // Casts are safe as the values are validated during config parsing
         cast(int) zone.soa.refresh.total!"seconds",
         cast(int) zone.soa.retry.total!"seconds",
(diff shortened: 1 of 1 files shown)


source/agora/node/Registry.d
```diff
@@ -1031,7 +1028,7 @@ private struct ZoneData
     {
         if (this.type == ZoneType.primary)
         {
-            this.soa.serial = cast(uint) Clock.currTime(UTC()).toUnixTime();
+            this.soa.serial += 1;
```
Geod24: Using the UNIX time is quite handy though, to know about the last update.
Perhaps `max(time, serial + 1)` would be the best of both worlds ?

Geod24 (approved):

PR #2959, opened 2022-02-03
Remove some ulong usage, use `Duration` and `TimePoint` instead

Geod24:
Not much to it, but I saw that we had similar patterns in a few places, so I did a small cleanup.


hewison-chris (approved):
LGTM

PR #2961, opened 2022-02-03
Tests: Use LocalRest for DNS queries, not bare Channels

Geod24:
```
LocalRest was intended as a way to simplify RPC calls, no mattter the type
of communication going on under the hood, as it simply replaces it.
So instead of using bare Channel for communication, just use the same RPC
approach, by exposing one method for each communication method.
```

Haven't tested it yet, but here's the idea.

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -636,12 +636,6 @@ public class TestAPIManager
     /// The name registry used in this network
     public NodePair dns;
 
-    /// The channel that the dns server listens to
-    public Channel!DNSQuery dns_chan;
-
-    /// A DNS resolver to query registry
-    public LocalRestDNSResolver dns_resolver;
-
     /// Used by the unittests in order to directly interact with the nodes,
     /// without trying to handshake or do any automatic network discovery.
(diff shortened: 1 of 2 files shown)


source/agora/test/Base.d
```diff
@@ -931,11 +928,9 @@ public class TestAPIManager
         assert(conf.interfaces.length == 1);
 
         auto time = new shared(TimePoint)(this.initial_time);
-        this.dns_chan = new Channel!DNSQuery();
-        this.dns_resolver = new LocalRestDNSResolver(this.dns_chan);
```
mkykadir, line 935: `test/DNS.d` uses this during tests

Geod24: Fixed

mkykadir (approved):
This is cool :sunglasses:

omerfirmak:
We now either need NS records in unittests or move `registry_address` out of `ValidatorConfig`

Geod24:
> We now either need NS records in unittests or move registry_address out of ValidatorConfig

I started doing this, then hit a configy bug, which turned into a DMD bug. So WIP.

PR #2963, opened 2022-02-04
Trivial: Remove unused test type DNSQuery

Geod24:
Its usage was removed in a previous commit, but the type was left behind.


Geod24:
Self merging as this should be fairly straightforward.

PR #2968, opened 2022-02-04
Move config for registry_address from validator to node section

Geod24:
In 'validator', it won't even be parsed if the section is not enabled.
Since all nodes rely on a validator, having it in node makes much more sense.


hewison-chris (approved):
LGTM

Geod24:
Neat, we should probably make a new release soon

PR #2969, opened 2022-02-04
RPC: Move RPCConnection delegate out of the constructor

Geod24:
This should make the code much easier to follow.
Best viewed with whitespace diff off.


linked0 (approved):
LGTM

PR #2976, opened 2022-02-04
Registry EDNS(0) fixes

mkykadir:
See each commit

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -955,7 +955,7 @@ public struct OPTRR
     ***************************************************************************/
 
     public ResourceRecord record = ResourceRecord(
-        Domain.init, TYPE.OPT, cast(CLASS) 4096, 0, ResourceRecord.RDATA(null));
+        Domain.init, TYPE.OPT, cast(CLASS) 2048, 0, ResourceRecord.RDATA(null));
 
     /// Requestor's UDP payload size
     public ushort payloadSize () scope const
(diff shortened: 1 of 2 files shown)


linked0 (approved):
LGTM

source/agora/common/DNS.d
```diff
@@ -955,7 +955,7 @@ public struct OPTRR
     ***************************************************************************/
 
     public ResourceRecord record = ResourceRecord(
-        Domain.init, TYPE.OPT, cast(CLASS) 4096, 0, ResourceRecord.RDATA(null));
+        Domain.init, TYPE.OPT, cast(CLASS) 2048, 0, ResourceRecord.RDATA(null));
```
Geod24, line 958: Would be nice to also define the buffer in terms of `OPTRR.init.payloadSize()` so that we don't have to keep those values in sync. And I think 4096 is more common.

source/agora/node/Registry.d
```diff
@@ -413,7 +413,7 @@ public class NameRegistry: NameRegistryAPI
                         goto BAILOUT;
                     }
                     // Ignore the DO bit for now
-                    payloadSize = min(opt.payloadSize(), ushort(512));
+                    payloadSize = max(opt.payloadSize(), ushort(512));
```
Geod24, line 416: I remember looking at this `min` long and hard and having a very good reason for it. Should have added a comment.

Looking at the code, it was to avoid DOS: If we take the client's value at face value, we essentially allow any size of response to be written. Using `min` ensures that we never send messages higher than 512 bytes.

So instead, this should be: `min(max(opt.payloadSize(), ushort(512)), OPTRR.init.payloadSize())`.

PR #2977, opened 2022-02-04
EDNS(0): Trivial fixes

mkykadir:
I've missed some reviews in #2976 due to early merge :disappointed: 

See each commit.

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -413,7 +413,9 @@ public class NameRegistry: NameRegistryAPI
                         goto BAILOUT;
                     }
                     // Ignore the DO bit for now
-                    payloadSize = max(opt.payloadSize(), ushort(512));
+                    // `min` is to prevent DOS attack (request with huge payload size)
+                    payloadSize = min(max(opt.payloadSize(), ushort(512)),
+                        responseOPT.payloadSize());
                     reply.additionals ~= responseOPT.record;
                 }
             }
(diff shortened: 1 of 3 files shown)


source/agora/node/Registry.d
```diff
@@ -413,7 +413,9 @@ public class NameRegistry: NameRegistryAPI
                         goto BAILOUT;
                     }
                     // Ignore the DO bit for now
-                    payloadSize = max(opt.payloadSize(), ushort(512));
+                    // `min` is to prevent DOS attack (request with huge payload size)
+                    payloadSize = min(max(opt.payloadSize(), ushort(512)),
+                        OPTRR.init.payloadSize());
```
Geod24: Actually, looking at the code again, `responseOPT.payloadSize()` is more correct (same value in the end).
`OPT` is a way to advertise the max payload size for both clients and server. A server that would send a record with a payload higher than its own OPT `payloadSize()` would be buggy.

Geod24 (approved):
Just one nit.

source/agora/node/Registry.d
mkykadir: `payloadSize` is not responded back, always `responseOPT` is used and its `OPTRR.init`

Geod24: Yes but if we ever change the OPT we respond, using `responseOPT.payloadSize()` will still be valid, while `OPTRR.init` will not.

mkykadir: Updated

Geod24 (approved):
Merge at will

PR #2978, opened 2022-02-04
Nominator: Nomination timer changes

omerfirmak:
see commit-by-commit

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -465,16 +465,6 @@ extern(D):
         }
 
         const cur_time = this.clock.networkTime();
-        const genesis_timestamp = this.params.GenesisTimestamp;
-
-        if (cur_time < genesis_timestamp)
-        {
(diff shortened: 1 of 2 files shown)


omerfirmak:
Didnt fix our particular problem, but could be a good addition nonetheless.

Geod24 (commented):
Can you explain commit 2 (the revert) ?
Also tiny type in the field name in commit 3 msg.

omerfirmak:
> Can you explain commit 2 (the revert) ?
> Also tiny type in the field name in commit 3 msg.

We should keep nominating until atleast a single value makes it to balloting phase. If we stop nominating halfway and no Value has made it to the balloting phase, we are doomed. We should find another way to limit the number of nominations.

hewison-chris (approved):
LGTM

linked0 (approved):
LGTM

Geod24:
Trying to get a green run, rebased on v0.x.x

Geod24 (approved):

PR #2979, opened 2022-02-04
DNS test updates

mkykadir:
See each commit

### DIFF
--- a/source/agora/test/DNS.d
+++ b/source/agora/test/DNS.d
@@ -39,41 +39,31 @@ unittest
         Address("agora://" ~ node_addr ~ ".validators.unittest.bosagora.io"));
     assert(addr.host.startsWith("10.0.0."));
 
-    Message query;
-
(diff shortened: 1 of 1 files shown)


source/agora/test/DNS.d
```diff
@@ -157,4 +157,7 @@ unittest
     // CNAME record
     result = resolver.query(node_addr ~ ".validators.unittest.bosagora.io.", QTYPE.CNAME);
     assert(result.length == 0);
+
+    result = resolver.query("validators.unittest.bosagora.io.", QTYPE.AXFR);
+    assert(result is null);
```
Geod24, line 165: How come there's no result?

mkykadir: Resolver sets return value to null when RCode is different than NoError. Caching registry doesn’t have AXFR capabilities.

Geod24 (approved):

PR #2980, opened 2022-02-04
Add  TCP support to `DNSResolver` and improve UDP

### LINKED ISSUE #2945: Zone transferring should use TCP
Even RFC#5936 is not prohibiting UDP usage for zone transferring, it is suggested and generally accepted that zone transfers should use TCP. This requires `DNSResolver` to have support for TCP usage and #2935 to be fixed.

mkykadir:
See each commit for details

- Adds TCP support to DNS resolver
- AXFR zone transfer over TCP
- Retries through TCP when UDP returns with truncation

Depends on #2979 

Fixes #2945

### DIFF
--- a/source/agora/network/DNSResolver.d
+++ b/source/agora/network/DNSResolver.d
@@ -78,14 +78,14 @@ public final class VibeDNSResolver : DNSResolver
 
     /***************************************************************************
 
-        Query the server with given `msg` and return the response
+        Query the server over UDP with given `msg` and return the response
 
         Params:
             msg = DNS message
 
     ***************************************************************************/
 
-    public override ResourceRecord[] query (Message msg) @trusted
+    public override ResourceRecord[] queryUDP (Message msg) @trusted
     {
         ubyte[16384] buffer;
         foreach (ref peer; this.resolvers)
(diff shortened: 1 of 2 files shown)


source/agora/network/DNSResolver.d
```diff
@@ -117,6 +123,62 @@ public final class VibeDNSResolver : DNSResolver
         }
         return null;
     }
+
+    public ResourceRecord[] queryTCP (Message msg) @trusted
```
Geod24: DDOC

source/agora/network/DNSResolver.d
```diff
@@ -86,10 +86,16 @@ public final class VibeDNSResolver : DNSResolver
...
 
     ***************************************************************************/
 
-    public override ResourceRecord[] query (Message msg) @trusted
+    public override ResourceRecord[] query (Message msg, bool tcp = false) @trusted
+    {
+        return !tcp ? this.queryUDP(msg) : this.queryTCP(msg);
+    }
+
+    public ResourceRecord[] queryUDP (Message msg) @trusted
```
Geod24: DDOC

source/agora/network/DNSResolver.d
```diff
@@ -86,10 +86,16 @@ public final class VibeDNSResolver : DNSResolver
 
         Params:
             msg = DNS message
+            tcp = Query the server with TCP, default is UDP
 
     ***************************************************************************/
 
-    public override ResourceRecord[] query (Message msg) @trusted
+    public override ResourceRecord[] query (Message msg, bool tcp = false) @trusted
```
Geod24: If we add one more level of indirection, I think we should make `query` `final` and have `queryUDP` / `queryTCP` be the points of override. Otherwise `query` becomes a [leaky abstraction](http://www.joelonsoftware.com/articles/LeakyAbstractions.html).

source/agora/test/Base.d
```diff
@@ -2631,12 +2631,12 @@ public final class LocalRestDNSResolver : DNSResolver
 
         Params:
             msg = DNS message
+            tcp = Query the server with TCP, default is UDP
 
     ***************************************************************************/
 
-    public override ResourceRecord[] query (Message msg) @trusted
+    public override ResourceRecord[] query (Message msg, bool tcp = false) @trusted
```
Geod24: Looks like this belongs to the previous commit, because otherwise it's not an override.

Geod24 (changes requested):
A few fixes needed

mkykadir:
Rebased and fixed issues

source/agora/network/DNSResolver.d
```diff
@@ -101,6 +101,11 @@ public final class VibeDNSResolver : DNSResolver
                 auto answer = response.deserializeFull!Message();
                 log.trace("Got response from '{}' for '{}' : {}",
                           peer.address, msg, answer);
+
+                // Message is truncated, retry over TCP
+                if (answer.header.TC == true)
+                    return this.queryTCP(msg);
+
```
Geod24: I think that should be handled at the higher level. Having something called `queryUDP` perform a TCP query is surprising. And since `answer.header.RCODE` will be `NoError` (correct me if I'm mistaken), the higher level `query` can handle the retry.

source/agora/test/Base.d
```diff
@@ -2643,6 +2643,9 @@ public final class LocalRestDNSResolver : DNSResolver
         {
             auto answer = !tcp ? p.queryUDP(msg) : p.queryTCP(msg);
             log.trace("Got response from for '{}' : {}", msg, answer);
+            if (answer.header.TC == true)
+                answer = p.queryTCP(msg);
```
Geod24: What if `tcp is true` ?

Geod24 (approved):
Just two small things to fix, otherwise LGTM.

source/agora/network/DNSResolver.d
mkykadir: I need to update return type for `queryUDP` and `queryTCP` then; it would be good to have another look after that.

mkykadir: I've updated, last 2 commits can be reviewed

Geod24 (approved):

PR #2982, opened 2022-02-07
Move nodes identity in `NetworkClient` and remove `NodeConnInfo`

Geod24:
The idea is that, eventually, we will have the `NetworkClient` responsible for anything related to a peer.
Currently it handles the gossip queue, and store all the connections, but there are some relics which make us instantiate `NetworkClient` too early, and we're almost to the point where we can remove them.
Another thing I'd like to do is more some of `ConnectionTask` to the `NetworkClient`, as often we want to connect to a certain validator, and we know beforehand what key we expect.


hewison-chris (approved):
LGTM

PR #2985, opened 2022-02-07
NetMan: Delay NetworkClient instantiation to remove tryMerge

Geod24:
By delaying the NetworkClient instantiation, we can actually simplify the code,
removing the need for the tryMerge function and making the code easier to follow.
This also paves the way to remove the duplication between clients and required_peers,
as in the future we would like to have connection-less NetworkClient that handles
their own 'ConnectionTask'.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -300,20 +300,50 @@ public class NetworkManager
         in Address address, agora.api.Validator.API api,
         in Hash utxo, in PublicKey key)
     {
+        log.dbg("onHandshakeComplete: {} - (k: {}, utxo: {})", address, key, utxo);
+        this.connection_tasks.remove(address);
+
+        // We have an authenticated client, maybe we already have a client for it
+        if (key !is PublicKey.init)
+        {
+            auto existing_peers = this.peers[].find!(p => p.identity.key == key);
+            if (!existing_peers.empty())
+            {
+                auto prev = existing_peers.front();
+                // There's three possibilities:
+                // a) We don't have an UTXO for this peer and we just got one,
+                //    so record it.
+                // b) We have an UTXO for this peer and got one, don't do anything
+                // c) We have an UTXO for this peer and didn't get one (or got a different one),
+                //    drop the new connection, as it could be another node
+                //    (e.g. backup) not wanting to be known as validator
+                //
+                // Case (c) shows that using keys for identification is tricky,
+                // but we had to do so due to DNS limitations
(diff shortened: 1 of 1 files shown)


source/agora/network/Manager.d
```diff
@@ -300,21 +300,51 @@ public class NetworkManager
...
+            auto existing_peers = this.peers[].find!(p => p.identity.key == key);
+            if (!existing_peers.empty())
+            {
+                auto prev = existing_peers.front();
+                // There's three possibilities:
+                // a) We don't have an UTXO for this peer and we just got one,
+                //    so record it.
+                // b) We have an UTXO for this peer and got one, don't do anything
+                // c) We have an UTXO for this peer and didn't get one (or got a different one),
+                //    drop the new connection, as it could be another node
```
omerfirmak, line 318: Why can't we assume the public keys to be unique? We dont let same public key to be enrolled with multiple UTXOs, so why would we try to handle that here.

source/agora/network/Manager.d
```diff
@@ -300,21 +300,51 @@ public class NetworkManager
...
+                // but we had to do so due to DNS limitations
+                if (prev.identity.utxo !is Hash.init)
+                {
+                    if (utxo != prev.identity.utxo)
+                        return; // Drop it
+                }
+                else if (utxo !is Hash.init)
+                {
+                    prev.setIdentity(utxo, key);
+                    this.required_peers.remove(utxo);
```
omerfirmak, line 331: We can do this at the very beginning (like `conn_tasks.remove()`)and not duplicate in on L339.

omerfirmak (approved):

source/agora/network/Manager.d
Geod24: Because we don't allow multiple enrollments with the same public key solely as an artificial limitation because so many other places assume uniqueness. But there's no real reason to forbid it other than a design mistake early on.

source/agora/network/Manager.d
Geod24: Actually, L339 means that this will always be the second removal, so I just deleted that line.

source/agora/network/Manager.d
Geod24: And so, why does the registry use public keys ? Because DNS is limited to 63 chars per label, and the string representation of our `Hash` is 128 chars, while a key is <= 62. There are other ways to deal with it (e.g. use another, shorter hash function just for the DNS name) but at the time it was the path of least resistance.

PR #2987, opened 2022-02-07
Trivial: Version out an unittest if Vibe is not available

Geod24:
Makes usage of agora.consensus as a library easier,
as otherwise the user will need a dependency on Vibe.d
for `dub test` to work.

### DIFF
--- a/source/agora/consensus/data/Transaction.d
+++ b/source/agora/consensus/data/Transaction.d
@@ -378,7 +378,7 @@ public struct TransactionResult
         Duplicated,
         Rejected,
     }
-    
+
     Status status;
     string reason;
 }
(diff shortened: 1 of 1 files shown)


linked0 (commented):
Where is the dependency to be removed, which was needed for `dub test` to work?

Geod24:
> Where is the dependency to be removed

`dub test` works fine in Agora. But Agora has a gazillion of dependency. The core of Agora, `agora.consensus`, is supposed to be usable as a library. That's why it should not depend on, for example, `agora.node`. There's a few failures to do so here and there, but overall, it is fairly contained. Likewise, if you look for imports to `vibe` in `agora.consensus`, you will see exactly one: this one. Which means that, by simply versioning it out when `Vibe` is not available (and the user will not care about Vibe's JSON serialization of `Transaction` if they don't use `Vibe`), we avoid a compiler error for library users, who do not depend on Vibe.d, when they run `dub test`.

linked0:
I see what you mean. Thanks for the explanation.

linked0 (approved):
LGTM

PR #2988, opened 2022-02-07
CPP: bind std::vector::reserve()

omerfirmak:
So that we can have a single allocation when building vectors
on D side.

### DIFF
--- a/source/scpp/extra/DUtils.cpp
+++ b/source/scpp/extra/DUtils.cpp
@@ -147,14 +147,18 @@ CPPOBJECTINST(std::set<NodeID>);
 CPPOBJECTINST(std::set<SCPBallot>);
 CPPOBJECTINST(std::set<unsigned int>);
 
-CPPOBJECTINST(std::vector<unsigned char>);
-CPPOBJECTINST(std::vector<Value>);
-CPPOBJECTINST(std::vector<unsigned long>);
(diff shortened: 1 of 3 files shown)


Geod24 (approved):

omerfirmak:
and of course mac fails linking.

source/scpp/extra/DUtils.cpp
```diff
@@ -147,15 +147,19 @@ CPPOBJECTINST(std::set<NodeID>);
...
-CPPOBJECTINST(std::vector<Value>);
-CPPOBJECTINST(std::vector<unsigned long>);
-CPPOBJECTINST(std::vector<unsigned long long>);
-CPPOBJECTINST(std::vector<SCPQuorumSet>);
-CPPOBJECTINST(std::vector<SCPEnvelope>);
-CPPOBJECTINST(std::vector<Slot::HistoricalStatement>);
 CPPOBJECTINST(stellar::SCPStatement::_pledges_t);
 
+#define CPPVECINST(T) template class T; \
+                      CPPOBJECTINST(T)
```
Geod24, line 153: Why this BTW ?

omerfirmak: to force generation of all `std::vector` symbols, otherwise only our wrappers (`defaultCtorCPPObject` et al) get generated.

PR #3001, opened 2022-02-08
Registry: Fix putting multiple A records into single RR

### LINKED ISSUE #2993: A records are combined within RDATA of same RR
RFC 1035 [states](https://datatracker.ietf.org/doc/html/rfc1035#section-3.4.1) that each should be in its own ResourceRecord (RR).

mkykadir:
Each A record should be returned in its own RR

Fixes #2993

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -692,17 +692,8 @@ private struct TypedPayload
         }
         else if (rr.type == TYPE.A)
         {
-            import std.socket : InternetAddress;
-            auto addresses = rr.rdata.a;
-            reg_payload.addresses = addresses.map!(
-                (addr) {
-                    scope in_addr = new InternetAddress(addr,
-                                        InternetAddress.PORT_ANY);
-
-                    // TODO SRV is needed to keep intact
-                    return Address("http://" ~ in_addr.toAddrString());
-                }
-            ).array;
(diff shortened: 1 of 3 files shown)


source/agora/node/Registry.d
```diff
@@ -693,37 +693,36 @@ private struct TypedPayload
...
-                (addr) {
-                    scope in_addr = new InternetAddress(addr,
-                                        InternetAddress.PORT_ANY);
-
-                    // TODO SRV is needed to keep intact
-                    return Address("http://" ~ in_addr.toAddrString());
-                }
-            ).array;
+            scope in_addr = new InternetAddress(rr.rdata.a,
+                                InternetAddress.PORT_ANY);
```
Geod24: `struct IPv4` can also be used here 👀

source/agora/node/Registry.d
```diff
@@ -693,37 +693,36 @@ private struct TypedPayload
...
+
+            // TODO SRV is needed to keep intact
+            reg_payload.addresses ~= Address("http://" ~ in_addr.toAddrString());
```
Geod24: In particular `Address(format("http://%s", IPv4(rr.rdata.a)))`.

Geod24 (approved):
나이스

PR #3003, opened 2022-02-08
Test: Filter UTXOs by lock type in getSpendables

omerfirmak:
```
Should fix crashes in Flash where TxBuilder tries to sign a
non-LockType.Key output
```

cc @linked0

### DIFF
--- a/source/agora/test/Base.d
+++ b/source/agora/test/Base.d
@@ -58,6 +58,7 @@ import agora.node.FullNode;
 import agora.node.Registry;
 import agora.node.TransactionRelayer;
 import agora.node.Validator;
+public static import agora.script.Lock;
 import agora.utils.Log;
 import agora.utils.PrettyPrinter;
 public import agora.utils.Utility : retryFor;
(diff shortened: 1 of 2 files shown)


source/agora/test/Base.d
```diff
@@ -1138,7 +1139,8 @@ public class TestAPIManager
 
         if (!no_txs)
         {
-            auto utxo_pairs = first_client.getSpendables(1.coins, OutputType.Payment);
+            auto utxo_pairs = first_client.getSpendables(1.coins, OutputType.Payment,
+                agora.script.Lock.LockType.Key);
```
linked0, line 1143: `agora.script.Lock.LockType.Key` is default to the `lock_type` parameter on the `getSpendables`, so can't we remove the last argument?

linked0 (commented):
Thanks, this is a time saver. There's one request for the code.

source/agora/test/Base.d
omerfirmak: Sadly no, `RemoteAPI` (LocalRest) generator does not work with default values.

linked0: I see.

linked0 (approved):

PR #3004, opened 2022-02-08
Registry API: Return the payload data directly

Geod24:
Currently, we return a RegistryPayload where the signature is always init.
This was a necessary change from the original design as zone transfer over
DNS does not allow to send those custom records.
We can just return a RegistryPayloadData directly and save ourselve some trouble.

### DIFF
--- a/source/agora/api/Registry.d
+++ b/source/agora/api/Registry.d
@@ -149,7 +149,7 @@ public interface NameRegistryAPI
 
     ***************************************************************************/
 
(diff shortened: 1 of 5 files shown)


mkykadir (approved):
I also plan to move towards to returning `ResourceRecord` from DB accessing methods like `get` and `update`. A cleaner Registry is near future! 😄

Geod24:
That'd be even better ! I was just writing some tests for a bug fix and got annoyed, figured the fix was easy and non-controversial.

PR #3005, opened 2022-02-08
Move 'struct IP' to Types and re-use it

Geod24:
```
It's a handy tool to format a string, and might be useful in other places,
so just place it in Types as its dependency are minimal.
```

@mkykadir : I just spotted it while reviewing your PR, so if you want to wait until yours is merged LMK, since it's obviously going to conflict.

### DIFF
--- a/source/agora/common/Types.d
+++ b/source/agora/common/Types.d
@@ -92,6 +92,58 @@ public struct Address
     public alias inner this;
 }
 
+/// A simple wrapper struct to string-represent IPv4 without allocating
+public struct IPv4
+{
+    import std.format;
+
+    public uint value;
+
+    /// Returns: An IPv4 in network byte order
+    public static IPv4 fromHost (uint v) @safe pure nothrow @nogc
+    {
+        import core.sys.posix.arpa.inet : htonl;
+        return IPv4(htonl(v));
+    }
+
+    /// Provides a string representation of this IP
+    public void toString (scope void delegate(in char[]) @safe sink)
+        const scope @safe
+    {
+        formattedWrite!"%s.%s.%s.%s"(sink,
+            this.value >> 24 & 0xFF, this.value >> 16 & 0xFF,
+            this.value >>  8 & 0xFF, this.value >>  0 & 0xFF);
+    }
+
+    public string toString () const scope @safe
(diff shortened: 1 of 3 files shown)


mkykadir:
Won't be a problem, we can first merge this

source/agora/common/Types.d
```diff
@@ -92,6 +92,58 @@ public struct Address
...
+        };
+        this.toString(sink);
+        return buffer[0 .. offset].idup;
+    }
+}
+
+///
+unittest
+{
+    IPv4 all = IPv4(0);
```
mkykadir, line 135: would it be better to force use of `fromHost` or have a ctor?

source/agora/network/DNSResolver.d
```diff
@@ -161,9 +160,7 @@ public abstract class DNSResolver
                 // Just pick the first record.
                 // TODO: What is expected when there's more than one record ?
                 // Should we use it as a fall back, round robin, something else?
-                const straddr = format("%d.%d.%d.%d",
-                    res.rdata.a[0] >> 24 & 0xFF, res.rdata.a[0] >> 16 & 0xFF,
-                    res.rdata.a[0] >>  8 & 0xFF, res.rdata.a[0] >>  0 & 0xFF);
+                const straddr = IPv4(res.rdata.a[0]).toString();
```
mkykadir, line 163: `fromHost`?

source/agora/common/Types.d
Geod24: I was wondering this. Having a disabled ctor causes all sort of problem so I went with the simplest approach for the time being.

source/agora/network/DNSResolver.d
Geod24: Oops

Geod24: I guess this shows the need for a disabled ctor :)

Geod24: I agreed too quickly, this is actually correct, or at the very least doing the same thing as before.
I suggest we merge this and take a look at whether it was doing the right thing in the first place ?

mkykadir (approved):

PR #3006, opened 2022-02-08
Small refactoring and fixes w.r.t dependency injection

Geod24:
Pretty simple, see each commit.


Geod24:
One of them passed 👀

linked0 (approved):
LGTM, there is a spurious failure on CI tho that does not occur in my local machine.

PR #3009, opened 2022-02-08
Add URI support to Registry

### LINKED ISSUE #2869: Generate `SRV`/`URI` record on node registration

mkykadir:
Adds URI support to registry, second commit ended up with a big changeset but commit message tries to explain why, I can try to split it when requested

```
URI record creates a chicken-egg problem, thus several other changes
are happened within this commit. URI records store full address, thus
A and CNAME records are not needed to store full addresses. Some
architectural changes are needed to achieve URI support and continue
supporting REST interface.

- SQL Foreign Key relation between utxo and addresses tables are broken
- Only primary zones will have utxo table
- Persistent storage methods are using ResourceRecord instead of TypedPayload
- TypedPayload can only be constructed with URI records from Registry side
- REST/GET interface will only return URI record
```

Fixes #2869

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -53,6 +53,9 @@ import d2sqlite3 : ResultRange;
 import vibe.http.client;
 import vibe.web.rest;
 
+/// Service and protocol label format used for URI resource record name
+private immutable string uri_service_label = "_agora._tcp.%s";
+
 /// Implementation of `NameRegistryAPI` using associative arrays
 public class NameRegistry: NameRegistryAPI
 {
@@ -97,6 +100,7 @@ public class NameRegistry: NameRegistryAPI
     /// Supported DNS query types
     private immutable QTYPE[] supported_query_types = [
         QTYPE.A, QTYPE.AAAA, QTYPE.CNAME, QTYPE.AXFR, QTYPE.ALL, QTYPE.SOA, QTYPE.NS,
+        QTYPE.URI,
     ];
 
     ///
@@ -180,7 +184,7 @@ public class NameRegistry: NameRegistryAPI
     /// Implementation of `NameRegistryAPI.getValidator`
     public override const(RegistryPayloadData) getValidator (PublicKey public_key)
     {
-        TypedPayload payload = this.zones[ZoneIndex.Validator].get(public_key);
+        TypedPayload payload = this.zones[ZoneIndex.Validator].getPayload(public_key);
         if (payload != TypedPayload.init)
         {
             log.trace("Successfull GET /validator: {} => {}", public_key, payload);
@@ -193,7 +197,7 @@ public class NameRegistry: NameRegistryAPI
     /// Internal endpoint, mimics `getValidator` but forward the query if needed
     public final const(RegistryPayloadData) getValidatorInternal (in PublicKey key)
     {
-        TypedPayload payload = this.zones[ZoneIndex.Validator].get(key);
+        TypedPayload payload = this.zones[ZoneIndex.Validator].getPayload(key);
         if (payload != TypedPayload.init)
             return payload.payload;
 
@@ -244,7 +248,7 @@ public class NameRegistry: NameRegistryAPI
     public void registerValidator (RegistryPayloadData data, Signature sig) @safe
     {
         TYPE payload_type = this.ensureValidPayload(data,
-            this.zones[ZoneIndex.Validator].get(data.public_key));
+            this.zones[ZoneIndex.Validator].getPayload(data.public_key));
 
         if (this.zones[ZoneIndex.Validator].type == ZoneType.secondary)
         {
@@ -263,7 +267,7 @@ public class NameRegistry: NameRegistryAPI
         const stake = this.getStake(data.public_key);
         ensure(stake !is Hash.init, "Couldn't find an existing stake to match this key");
 
-        this.zones[ZoneIndex.Validator].update(TypedPayload(payload_type, data, stake));
+        this.zones[ZoneIndex.Validator].updatePayload(TypedPayload(payload_type, data, stake));
         this.zones[ZoneIndex.Validator].updateSOA();
     }
 
@@ -285,7 +289,7 @@ public class NameRegistry: NameRegistryAPI
     /// Implementation of `NameRegistryAPI.getFlashNode`
     public override const(RegistryPayloadData) getFlashNode (PublicKey public_key)
     {
-        TypedPayload payload = this.zones[ZoneIndex.Flash].get(public_key);
+        TypedPayload payload = this.zones[ZoneIndex.Flash].getPayload(public_key);
         if (payload != TypedPayload.init)
         {
             log.trace("Successfull GET /flash_node: {} => {}", public_key, payload);
@@ -304,7 +308,7 @@ public class NameRegistry: NameRegistryAPI
         ensure(data.verify(sig), "Incorrect signature for payload");
 
         TYPE payload_type = this.ensureValidPayload(data,
-            this.zones[ZoneIndex.Flash].get(data.public_key));
+            this.zones[ZoneIndex.Flash].getPayload(data.public_key));
 
         if (this.zones[ZoneIndex.Flash].type == ZoneType.secondary)
         {
@@ -321,7 +325,7 @@ public class NameRegistry: NameRegistryAPI
         // register data
         log.info("Registering network addresses: {} for Flash public key: {}", data.addresses,
             data.public_key);
-        this.zones[ZoneIndex.Flash].update(TypedPayload(payload_type, data));
+        this.zones[ZoneIndex.Flash].updatePayload(TypedPayload(payload_type, data));
         this.zones[ZoneIndex.Flash].updateSOA();
     }
 
@@ -495,7 +499,13 @@ public class NameRegistry: NameRegistryAPI
 
         if (this.zones[ZoneIndex.Validator].type == ZoneType.primary)
         {
-            this.zones[ZoneIndex.Validator].each!((TypedPayload tpayload) {
+            this.zones[ZoneIndex.Validator].each!((ResourceRecord[] rrs) {
+                // That is not the performant way, until another approach this can
+                // help us to go through URI records
+                auto tpayload = this.zones[ZoneIndex.Validator].getPayload(
+                    // Query in `opApply` guarantees that `rrs` cannot be empty
+                    rrs[0].name.extractPublicKey()
+                );
                 if (this.ledger.getPenaltyDeposit(tpayload.utxo) == 0.coins)
                     this.zones[ZoneIndex.Validator].remove(tpayload.payload.public_key);
             });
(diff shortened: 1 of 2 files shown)


source/agora/node/Registry.d
```diff
@@ -512,14 +512,26 @@ private PublicKey extractPublicKey (in char[] domain) @safe
...
-            return PublicKey.init;
+        // Includes service and protocol labels
+        if (domain.serviceAndProtocolLabels() != Domain.ServiceProtoTuple.init)
+        {
+            splitted.popFront(); // Pop service
+            splitted.popFront(); // Pop proto
+            str = splitted.front;
+        }
+
+        if (str)
```
omerfirmak: I dont think this is needed, if `str == null`, `str.length` is 0

Geod24: (not so) fun fact:
```D
void main ()
{
 	string str = "";
    if (str)
        assert(0); // Triggers
}
```

That's why we should **always** use `.length` when checking for empty string, not `if (str)`, `if (str is null)`, or `if (str == string.init)`.

mkykadir: Below lines are checking for `.length` already; thus removing

Geod24:
Reviewing. First commit LGTM though, so feel free to submit it separately and self-merge if you want to work iteratively.

source/agora/node/Registry.d
```diff
@@ -94,6 +94,7 @@ public class NameRegistry: NameRegistryAPI
     /// Supported DNS query types
     private immutable QTYPE[] supported_query_types = [
         QTYPE.A, QTYPE.AAAA, QTYPE.CNAME, QTYPE.AXFR, QTYPE.ALL, QTYPE.SOA, QTYPE.NS,
+        QTYPE.URI
```
Geod24: Trailing commas, always.

source/agora/node/Registry.d
```diff
@@ -289,8 +290,8 @@ public class NameRegistry: NameRegistryAPI
 
         // register data
         log.info("Registering network addresses: {} for Flash public key: {}", data.addresses,
-            data.public_key);
-        this.zones[ZoneIndex.Flash].update(TypedPayload(payload_type, data));
+            data.public_key.toString());
```
Geod24: Why?

source/agora/node/Registry.d
```diff
@@ -515,7 +517,7 @@ private PublicKey extractPublicKey (in Domain domain) @safe
     auto splitted = domain.value.splitter('.');
     if (auto str = splitted.front)
     {
-        // Includes service and protocol labels
+        // Check if `domain` includes service and protocol labels
```
Geod24: Belongs to the first commit, not the second

source/agora/node/Registry.d
```diff
@@ -1335,51 +1337,100 @@ private struct ZoneData
...
 
-        // Check query type; QTYPE is superset of TYPE, casting is OK
-        // RFC#1034 : Section 3.6.2
-        if (node_type != TYPE.CNAME
-            && type != QTYPE.ALL && type != cast(QTYPE) node_type)
+    public TypedPayload getPayload (PublicKey public_key) @trusted
+    {
+        auto records = this.get(public_key, QTYPE.URI);
+
+        if (records is null)
```
Geod24: Check `length` here

source/agora/node/Registry.d
```diff
@@ -604,47 +606,37 @@ private struct TypedPayload
...
             pubKeyParser = Delegate for parsing public key from RR domain
 
     ***************************************************************************/
 
-    public static TypedPayload make (ResourceRecord rr)
+    public static TypedPayload make (ResourceRecord[] rrs)
     {
-        auto public_key = rr.name.extractPublicKey();
+        ensure(rrs !is null || rrs.length > 0,
+            "No records are provided to make TypedPayload");
```
Geod24: Redundant condition (null / length).
Also it looks like something the caller should check, not the callee. In the presence of array / nullable data, this is the approach I usually take:

Functions like this can have one of two behavior: They should either be no-op in the face of `null` data, or not no-op.

If they are not no-op, it means `null` is considered an in-band value. An example would be a function that either can do something with the `null` value (rare) or has some sort of fallback (like the ill-fated https://github.com/bosagora/agora/pull/2981).

Some other functions want to be no-ops in the face of `null`. An example of this is `startPeriodicNameRegistration`, where we just `return` and don't do anything if `config.node.registry_address` is not configured.

Now what is this function doing: It is neither no-op (since it throws) nor is the `null` value in-band.
My instinct here is to ask the question: "Should we have reached **this point** or should we have stopped before?". By checking the caller, I see that the check is also performed in the caller, so we should never have reached this in the first place. And IMO, considering the tradeoffs, that's a sound thing to do. Hence, this shouldn't be an `ensure`, it should be an `assert`, because passing out-of-band data should have been caught earlier.

Now this is not an absolute rule: Some functions will accept out-of-band data and then check if they are in-band. But that is usually their sole focus. One good example is Vibe.d, which needs to accept arbitrary string and turn them into a structured view of the HTTP request. If you were to get an `HTTPClientRequest` from Vibe.d with data that violate the HTTP specs, you would be right to call it a Vibe.d bug.

Geod24 (commented):
Ouch. That duplication is going to bite us hard down the road, but I see what problem you had. I have a couple of ideas on how we can address it, let me sketch it. Also, I think the commit that introduce the tests should come with the feature it introduces.

source/agora/node/Registry.d
```diff
@@ -464,7 +465,8 @@ public class NameRegistry: NameRegistryAPI
 
         if (this.zones[ZoneIndex.Validator].type == ZoneType.primary)
         {
-            this.zones[ZoneIndex.Validator].each!((TypedPayload tpayload) {
+            this.zones[ZoneIndex.Validator].each!((ResourceRecord[] rrs) {
+                auto tpayload = TypedPayload.make(rrs);
```
mkykadir: This introduces a bug, should use `getPayload` here

source/agora/node/Registry.d
```diff
@@ -711,6 +705,13 @@ private struct TypedPayload
             assert(this.payload.addresses.length == 1);
             dg(ResourceRecord.make!(TYPE.CNAME)(name, this.payload.ttl,
                 Domain.fromString(this.payload.addresses[0].host)));
+
+            // Add URI record along with
+            dg(ResourceRecord.make!(TYPE.URI)(
+                Domain.fromString(format("_%s._tcp.%s", this.payload.addresses[0].schema, name.toString)),
```
Geod24: The first element is the name of the service, in our case, that should be Agora, shouldn't it ?

mkykadir: I thought that HTTP would be another service and only RPC would be Agora; but its wrong and switched to `_agora._tcp`.

source/agora/node/Registry.d
```diff
@@ -711,6 +708,13 @@ private struct TypedPayload
             assert(this.payload.addresses.length == 1);
             dg(ResourceRecord.make!(TYPE.CNAME)(name, this.payload.ttl,
                 Domain.fromString(this.payload.addresses[0].host)));
+
+            // Add URI record along with
+            dg(ResourceRecord.make!(TYPE.URI)(
+                Domain.fromString(format(uri_service_label, name.toString)),
```
Geod24: Note: You do not need to call `toString` here, `format` will do that for you.
In fact, it should call the `toString` overload with a `sink`, reducing the amount of allocations.

Geod24: Another easy improvement: We know that an address can only be 256 chars long at most. So we could format this into a static buffer because it can never exceed 256 chars (something we do not check BTW, and could cause problems).

source/agora/node/Registry.d
```diff
@@ -903,11 +930,14 @@ private struct ZoneData
         string query_addr_create = format("CREATE TABLE IF NOT EXISTS registry_%s_addresses " ~
             "(pubkey TEXT, address TEXT NOT NULL, type INTEGER NOT NULL, " ~
             "ttl INTEGER NOT NULL, expires INTEGER, " ~
-            "FOREIGN KEY(pubkey) REFERENCES registry_%s_utxo(pubkey) ON DELETE CASCADE, " ~
-            "PRIMARY KEY(pubkey, address))", zone_name, zone_name);
+            "PRIMARY KEY(pubkey, address))", zone_name);
 
-        // Initialize common fields
+        // Creation of signature table can be limited to only primary zones
```
Geod24: "signature table" ?

source/agora/node/Registry.d
```diff
@@ -1357,7 +1366,7 @@ private struct ZoneData
 
     /***************************************************************************
 
-         Get payload data from persistent storage
+         Get ResourceRecord's from persistent storage
```
Geod24: Why the possessive instead of plural ?

source/agora/node/Registry.d
```diff
@@ -1366,51 +1375,100 @@ private struct ZoneData
            type = the type of the record that was requested, default is `ALL`
 
         Returns:
-            TypedPayload with type `type` of name registry associated with
-            `public_key`. All records are returned when `type` is `ALL`
+            ResourceRecord array with type `type` of name registry associated with
```
Geod24, line 1378: Technically, it will match `type`, but it *is* not type, because `type` can be a wildcard.

source/agora/node/Registry.d
```diff
@@ -1442,46 +1501,104 @@ private struct ZoneData
...
+                    assert(0, "RR type is not supported for permenant storage");
+            }
+
             db.execute(this.query_addresses_add,
-                payload.payload.public_key,
-                address,
-                payload.type.to!ushort,
-                payload.payload.ttl,
-                payload.expires);
+                rr.name.extractPublicKey().toString,
```
Geod24: Likewise, `toString` should not be called here.

Geod24 (changes requested):
Few nits, feel free to self merge

source/agora/node/Registry.d
mkykadir: > (something we do not check BTW, and could cause problems).

`Domain.fromString` has `ensure` for that

mkykadir: > So we could format this into a static buffer because it can never exceed 256 chars

`Domain.fromString` checks `length` of the array not `strlen`, thus I think static array would cause problems there.

source/agora/node/Registry.d
mkykadir: I only know `ALL` condition and its commented

source/agora/node/Registry.d
Geod24:
```D
            ensure(uri_service_label.length + name.toString <= Domain.MaxLength, "Useful msg");
            char[Domain.MaxLength] buffer;
            dg(ResourceRecord.make!(TYPE.URI)(
                Domain.fromString(snformat(buffer, uri_service_label, name)),
```

This would not allocate, and work as expected. It does not allocate because Ocean's `snformat` will write to a static array up to its length, and it returns a slice to the subset of the array that has been written to.
I am not saying we should use it here - There are plenty of memory optimizations needed and I don't think we are there yet. But I wanted to share the pattern.

Geod24 (approved):

PR #3010, opened 2022-02-09
Registry: Remove table drop in case of type change

Geod24:
This piece of code was dropping the registry tables iff the type of the registry
was changed between two runs. There's apparently no technical reason to do so,
as all types use the same table format.
This code created problems because it consistently prints 3 error messages when a node
starts, which can be startling to users. This could be avoided by checking for the
table existence before actually doing the SELECT, however, when considering the logic,
it stood out as no other node has that behavior (e.g. switching between full node and validator).
In this case, I think it is better to simply not do this and let the operator
wipe the table clean (by removing cacheDB) if that behavior is desired.

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -929,32 +929,6 @@ private struct ZoneData
             "FOREIGN KEY(pubkey) REFERENCES registry_%s_utxo(pubkey) ON DELETE CASCADE, " ~
             "PRIMARY KEY(pubkey, address))", zone_name, zone_name);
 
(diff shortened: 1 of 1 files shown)


linked0 (approved):
LGTM

mkykadir (approved):
Even I don’t expect type to change much between runs; this might cause stale data problem. But it looks like this problem will be gone with URI support.

Geod24:
I think we might end up with some "metadata" table on the node to account for those situations: For example, to allow nodes to auto-upgrade their database when restarted with a new release.

PR #3011, opened 2022-02-09
NetCli: Lazily remove banned connections

Geod24:
```
If the connection is banned, we should just drop it and restart the loop.
```

Best viewed with whitespace diff off.

### DIFF
--- a/source/agora/network/Client.d
+++ b/source/agora/network/Client.d
@@ -604,35 +604,51 @@ public class NetworkClient
         enum name = __traits(identifier, endpoint);
         alias T = ReturnType!(__traits(getMember, API, name));
 
-        foreach (idx; 1 .. this.max_retries + 1)
+        T onError ()
         {
-            foreach (conn; this.connections)
+            static if (DT == Throw.Yes)
+            {
+                this.exception.file = file;
+                this.exception.line = line;
+                throw this.exception;
+            }
+            else static if (!is(T == void))
+                return T.init;
+        }
+
+    RETRY: foreach (idx; 1 .. this.max_retries + 1)
+        {
+            // Clients without connections should not wait and instead error out
+            if (!this.connections.length)
+                return onError();
+
+            foreach (index, conn; this.connections)
             {
                 if (this.banman.isBanned(conn.address))
                 {
-                    log.dbg("Client.attemptRequest '{}' skipped as {} is banned: {}/{}", name, conn.address, idx, this.max_retries);
+                    this.connections.remove(index);
+                    log.warn("Removing banned address {} while performing {}, addresses left: {}",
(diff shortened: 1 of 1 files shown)


source/agora/network/Client.d
```diff
@@ -604,35 +604,51 @@ public class NetworkClient
...
+            // Clients without connections should not wait and instead error out
+            if (!this.connections.length)
+                return onError();
+
+            foreach (index, conn; this.connections)
             {
                 if (this.banman.isBanned(conn.address))
                 {
-                    log.dbg("Client.attemptRequest '{}' skipped as {} is banned: {}/{}", name, conn.address, idx, this.max_retries);
+                    this.connections.remove(index);
```
linked0: We can use this connection again after the banning duration ends. So why should it be removed?

source/agora/network/Client.d
```diff
@@ -604,35 +604,51 @@ public class NetworkClient
...
+                    log.warn("Removing banned address {} while performing {}, addresses left: {}",
+                             conn.address, name, this.addresses);
+                    continue RETRY;
```
omerfirmak: This fixes the problem only for this instance of the loop, but we are removing from a list that could potentially be traversed by multiple fibers at any given time.

omerfirmak: @Geod24 , added a new commit to fix this and rebased.

omerfirmak: we can squash if you like it.

source/agora/network/Client.d
```diff
@@ -603,36 +606,62 @@ public class NetworkClient
...
+                    {
+                        curr_conn_version = this.connections_version;
+                        continue RETRY;
+                    }
+                }
+
                 if (this.banman.isBanned(conn.address))
                 {
-                    log.dbg("Client.attemptRequest '{}' skipped as {} is banned: {}/{}", name, conn.address, idx, this.max_retries);
+                    this.connections.remove(index);
```
omerfirmak: We shouldnt delete by index tho, it should be by value.

source/agora/network/Client.d
```diff
@@ -624,12 +628,21 @@ public class NetworkClient
 
             foreach (index, conn; this.connections)
             {
+                scope (exit)
+                {
+                    if (this.connections_version > curr_conn_version)
+                    {
+                        curr_conn_version = this.connections_version;
+                        continue RETRY;
```
Geod24: We should not jump from a `scope (exit)`

source/agora/network/Client.d
omerfirmak: done

source/agora/network/Client.d
omerfirmak: updated

Issue #2992, opened 2022-02-07
Integration test throws errors on SOA record query


Geod24:
```
node-7_1  | 2022-02-07 11:32:59,741 Info [boa1xrval7gwhjz4k9raqukcnv2n4rl4fxt74m2y9eay6l5mqdf4gntnzhhscrh] - Accepted tx: Inputs (1): 0x8107...ccd9:0xd361...6d14
node-7_1  | Outputs (1): boa1xrj0050f...er6u(23,086,422.0,175,273)<Payment>, txHash: 0xebabfddb660aacc723be6744bce927e9a94102c89b227f9bc2d54be43db6a39ef5fde555939cfafb1b3ab11fbfcc634971856023aae2d1d6a31d1b5c5dfbbd4b
node-7_1  | 2022-02-07 11:33:00,173 Warn [agora.network.DNSResolver] - Network error while resolving '{ header: ID: 966d QUERY rd NoError, Query: 1, Answer: 0, Authority: 0, Additional: 1, questions: [{ qname: integration.bosagora.io., qtype: QTYPE.SOA, qclass: QCLASS.IN }], answers: [], authorities: [], additionals: [OPT (payload: 4096)] }' using '172.18.0.5:1826': object.Exception@submodules/vibe-core/source/vibe/core/net.d(1096): Failed to receive packet.
node-7_1  | ----------------
node-7_1  | /usr/include/d/std/exception.d:513 pure @safe noreturn std.exception.bailOut!(Exception).bailOut(immutable(char)[], ulong, scope const(char)[]) [0x55f552b74560]
node-7_1  | /usr/include/d/std/exception.d:433 pure @safe bool std.exception.enforce!().enforce!(bool).enforce(bool, lazy const(char)[], immutable(char)[], ulong) [0x55f5528e4150]
node-7_1  | submodules/vibe-core/source/vibe/core/net.d:1071 @safe ubyte[] vibe.core.net.UDPConnection.recv(core.time.Duration, ubyte[], vibe.core.net.NetworkAddress*) [0x55f55350adb0]
node-7_1  | source/agora/network/DNSResolver.d:88 @trusted agora.common.DNS.Message agora.network.DNSResolver.VibeDNSResolver.queryUDP(agora.common.DNS.Message) [0x55f552cc2630]
[...]

PR #2981, opened 2022-02-07
Fix #2962: Use registry address as default DNS resolver

Geod24:
The code previously was falling back to outside DNS servers,
which means the registry wasn't used when it should have been.

PR #3015, opened 2022-02-09
Fix #2989 - Revert PR #2981

### LINKED ISSUE #2989: Every node is required to provide a `registry_address`
I don't think this was intended, `DNSResolver` has default DNS resolvers configured thus it shouldn't be required to have `registry_address` field for every node.

https://github.com/bosagora/agora/blob/133499d7c8083398e147be90bcb2874d86487cbc/source/agora/network/Manager.d#L695-L696

Geod24:
PR #2981 assumed that `registry_address` contained the address of a DNS server,
irrespective of the actual schema. This broke the live system as DNS request
are now sent over port 80. Revert to the default DNS resolvers,
as it will work properly, although it will issue wrong requests
in integration tests.

### DIFF
--- a/source/agora/network/VibeManager.d
+++ b/source/agora/network/VibeManager.d
@@ -67,9 +67,7 @@ public final class VibeNetworkManager : NetworkManager
     /// See `NetworkManager.makeDNSResolver`
     public override DNSResolver makeDNSResolver (Address[] peers = null)
     {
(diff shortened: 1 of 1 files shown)


mkykadir (approved):

mkykadir:
This fixes #2992

PR #3016, opened 2022-02-09
Update LocalRest to support default arg and use them

Geod24:
Per popular demand.


omerfirmak (approved):

Geod24:
Passed locally:
> dsinglethreaded=true dchatty=true dub test   711.83s user 63.77s system 103% cpu 12:31.90 total

PR #2984, opened 2022-02-07
Optimize `NetworkManager.discover` memory usage

Geod24:
It doesn't really fix the design issues, but at least it should reduce memory usage quite a bit.

PR #3022, opened 2022-02-10
NetMan: Only update known validators on block externalized

Geod24:
```
Currently we allocate an associative array every time we call discover,
even if nothing has changed, and we call discover quite often.
```

This is part of #2984 but the other commit fails, however this one should have the most impact on memory usage.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -247,6 +247,9 @@ public class NetworkManager
     /// Current quorum set
     private Set!Hash quorum_set_keys;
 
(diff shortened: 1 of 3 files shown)


Geod24:
We should probably just store a Ledger in the `NetworkManager` in the end...

hewison-chris (approved):
LGTM

Geod24:
Previous commit was making `TimeDrift` fail, this one makes `NetworkDiscovery` fail :)

Geod24:
Done in https://github.com/bosagora/agora/pull/3080

PR #3025, opened 2022-02-10
ValidatorSet: Cache result of getValidators()

omerfirmak:
This is called pretty frequently, we will try to respond from
cache as an optimization.


Geod24:
Rebased

Geod24 (approved):

PR #3038, opened 2022-02-11
Centralize all TestNet specialization behind a switch

Geod24:
```
This should have been done this way in the first place.
The realm in Config was left to TestNet but could be changed back anytime.
```

CC @mkykadir

### DIFF
--- a/devel/testnet/config.yaml
+++ b/devel/testnet/config.yaml
@@ -13,7 +13,7 @@
 node:
   realm: "testnet.bosagora.io"
   testing: true
-  data_dir: .fullnode/data/
+  data_dir: data/
   registry_address: http://ns1.bosagora.io/
 
 consensus:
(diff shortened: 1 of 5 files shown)


devel/testnet/config.yaml
```diff
@@ -13,7 +13,7 @@
 node:
   realm: "testnet.bosagora.io"
   testing: true
-  data_dir: .fullnode/data/
+  data_dir: data/
```
mkykadir, line 16: 👍

mkykadir (approved):

devel/testnet/config.yaml
Geod24: Will apply to `--testnet` as well

Geod24:
Fun fact: If we go with modifying `dub.json` to have extra string import path, we could do:
```
parseString(mixin("devel/testnet/config.yaml"));
```

And the configuration file would be embedded into Agora :D

Geod24:
Updated

PR #3040, opened 2022-02-11
Correct doc and image on how to join TestNet

linked0:
Modified the text and the image for binaries as the version of the binaries have changed.


Geod24:
It's actually v0.34.1, but I don't think it makes sense to update that image for every release. I'd rather we reword it and keep the old image.

Geod24:
Also https://bosplatformfoundation.slack.com/archives/GHBDRKY1F/p1644568819715289

linked0:
> It's actually v0.34.1, but I don't think it makes sense to update that image for every release. I'd rather we reword it and keep the old image.

Right, I updated only the version number and @TrustHenry's request has been already applied in the previous PR. This is just fixing the text in the README. So I'm merging it.

PR #3041, opened 2022-02-11
Registry: Add AAAA record support

### LINKED ISSUE #2990: Registry cannot represent `A` and `AAAA` records within same payload
Due to single type field, both types cannot be represented within same payload

https://github.com/bosagora/agora/blob/133499d7c8083398e147be90bcb2874d86487cbc/source/agora/node/Registry.d#L1429-L1430

mkykadir:
Add AAAA record support to registry. This fixes #2990 and it is the reason that it depends on #3009 which changes underlying structure to support records with different type for same public key.

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -618,6 +618,13 @@ public struct ResourceRecord
         return ResourceRecord(name, TYPE.A, CLASS.IN, ttl, RDATA(ipv4));
     }
 
+    /// Make a record of AAAA type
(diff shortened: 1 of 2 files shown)


Geod24:
I don't follow, #2781 was merged over a month ago ?

Geod24:
Ah you meant #3009

Geod24 (approved):
Merge at will once rebased on `v0.x.x`

mkykadir:
Rebased, lets have som checks pass

PR #3043, opened 2022-02-14
Manager.validators(): filter out non-enrolled validators 

omerfirmak:


omerfirmak:
We now remove them from our peer list anyways

PR #3045, opened 2022-02-15
Update genesis_timestamp to match TestNet

Geod24:


linked0 (approved):

Geod24:
Actually that was the wrong place to do it, fixed it now

PR #3047, opened 2022-02-15
Use URI records in Manager

mkykadir:
Utilize Registry DNS interface and get URI records for connecting peers

This changeset will reduce load on a primary Registry node, each node will cache the query results in their own Registry.

Depends on #3009

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -27,6 +27,7 @@ import agora.api.Handlers;
 import agora.api.Registry;
 import agora.api.Validator;
 import agora.common.BanManager;
+import agora.common.DNS;
 import agora.common.ManagedDatabase;
 import agora.common.Set;
 import agora.common.Task;
@@ -599,22 +600,27 @@ public class NetworkManager
                     const ckey = key;
                     try
                     {
-                        auto payload = this.registry.getValidatorInternal(ckey);
-                        if (payload == RegistryPayloadData.init)
-                        {
-                            log.warn("Could not find mapping in registry for key {}", ckey);
-                            return;
-                        }
+                        this.registry.getValidatorInternal(ckey,
+                            (in Message msg) @trusted {
+                                if (msg.header.RCODE != Header.RCode.NoError)
+                                {
+                                    log.warn("Could not find mapping in registry for key {}", ckey);
+                                    return;
+                                }
 
-                        if (payload.public_key != ckey)
-                        {
-                            log.error("Registry answered with the wrong key: {} => {}",
-                                      ckey, payload);
-                            return;
-                        }
+                                foreach (rr; msg.answers)
+                                {
+                                    if (rr.type != TYPE.URI)
+                                    {
+                                        log.warn("Registry returned non-URI record ({}), ignoring",
+                                            rr.type);
+                                        continue;
+                                    }
 
-                        foreach (addr; payload.addresses)
-                            this.addAddress(addr);
+                                    this.addAddress(rr.rdata.uri.target);
+                                }
+                            }
+                        );
                     }
                     catch (Exception exc)
                         log.error("Exception happened while looking up address for {}: {}",
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -195,29 +195,17 @@ public class NameRegistry: NameRegistryAPI
     }
 
     /// Internal endpoint, mimics `getValidator` but forward the query if needed
-    public final const(RegistryPayloadData) getValidatorInternal (in PublicKey key)
+    public final void getValidatorInternal (in PublicKey key,
+        scope void delegate (in Message) @safe dg)
     {
-        TypedPayload payload = this.zones[ZoneIndex.Validator].getPayload(key);
-        if (payload != TypedPayload.init)
-            return payload.payload;
+        Message msg;
+        msg.questions ~= Question(
+            Domain.fromString(format("%s.%s", key, this.validators)),
+            QTYPE.URI, QCLASS.IN
+        );
+        msg.fill(msg.header);
 
-        // If we're not caching, we're either primary or secondary.
-        // If we're primary, there's obviously no upstream to ask.
-        // If we're secondary, we ask the primary, we do not want other zone data
-        // to get out of sync (e.g. the serial), so we only query,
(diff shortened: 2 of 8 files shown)


source/agora/node/Registry.d
```diff
@@ -195,29 +195,17 @@ public class NameRegistry: NameRegistryAPI
     }
 
     /// Internal endpoint, mimics `getValidator` but forward the query if needed
-    public final const(RegistryPayloadData) getValidatorInternal (in PublicKey key)
+    public final void getValidatorInternal (in PublicKey key,
```
omerfirmak, line 198: Let's return a `bool` here,, so that we dont need the `retVal` trick at the caller site?

mkykadir: Not needed since #3053 removes `retry`

omerfirmak:
This still returns a CNAME and results in a vibed DNS lookup, could registry resolve the CNAME as well using the DNSResolver and cache that too?

Geod24:
IMO this should be handled client side, e.g. I sketched it as https://github.com/Geod24/agora/commit/e54fcb7a342a7c26065dd58ff8e7177c39cfffba#diff-7bac49d84db926243af8e35c6d23c21b8becdff1fb00e08fd123313678ab2970R742

Regarding the CNAME, the registry needs to return a `CNAME` if that's what was in `addresses_to_register`, but the `DNSResolver` needs to be made recursive.

mkykadir:
> IMO this should be handled client side, e.g. I sketched it as [Geod24@e54fcb7#diff-7bac49d84db926243af8e35c6d23c21b8becdff1fb00e08fd123313678ab2970R742](https://github.com/Geod24/agora/commit/e54fcb7a342a7c26065dd58ff8e7177c39cfffba#diff-7bac49d84db926243af8e35c6d23c21b8becdff1fb00e08fd123313678ab2970R742)

That feels like duplication, Registry (with caching zones) is also a client.

mkykadir:
Rebase on top #3071 and it fixed a problem in integration test which was resulting in (host name - port number) mismatch in peers list. I will investigate how it resolved that, probably a concurrency issue was happening.

mkykadir:
Ready to review, thanks to #3095; fixes #2992

source/agora/node/Registry.d
```diff
@@ -195,29 +195,17 @@ public class NameRegistry: NameRegistryAPI
...
-            return RegistryPayloadData.init;
-
-        auto upstream = this.client.getValidator(key);
-        if (upstream !is RegistryPayloadData.init &&
-            this.zones[ZoneIndex.Validator].type == ZoneType.caching)
-            this.registerValidator(upstream, Signature.init);
-        return upstream;
+        Message msg;
+        msg.questions ~= Question(
+            Domain.fromString(format("%s.%s", key, this.validators)),
```
mkykadir, line 203: Actually, this should have service and proto labels; but currently we allow this query for URI records to gather both `http` and `agora`

source/agora/network/Manager.d
```diff
@@ -570,22 +571,27 @@ public class NetworkManager
...
-                        }
-
-                        if (payload.public_key != ckey)
-                        {
-                            log.error("Registry answered with the wrong key: {} => {}",
-                                      ckey, payload);
-                            return;
-                        }
+                        this.registry.getValidatorInternal(ckey,
+                            (in Message msg) @trusted {
```
omerfirmak, line 604: I feel like the `NetworkManager` should not be dealing with DNS `Message`s directly, I still think using `DNSResolver` here (or in `NetworkClient`) is the sanest way forward.

mkykadir: I agree that manager shouldn't be dealing with any DNS stuff

Using `DNSResolver` within Manager would end up in following chain (for caching node)
```
Manager -> DNSResolver -> Registry (checks cache) -> DNSResolver -> out network reach other node
```
and it would just be, not so necessary, additional step. We can still resolve dealing with DNS stuff in manager directly using the provided registry interface but it would end up with array allocations. Instead of requiring a delegate it can return array of addresses directly.

Geod24:
There's a few things I don't follow with this PR:
1) It completely removes caching AFAICS (??)
2) It makes the `NetworkManager` deal with DNS details, instead of using the `RegistryAPI` interface;

> That feels like duplication, Registry (with caching zones) is also a client.

Yep, but maybe the `Registry` shouldn't use a `RegistryAPI` to contact an upstream registry.
I was experimenting with multiple approach, another possibility is to change `getValidator` to return `const(Address)[]`.

mkykadir:
@Geod24 

1. No, `getValidatorInternal` calls `answerQuestion` internally which will check for the registry cache
2. Yep, I don't like it either; `getValidatorInternal` can return `Address[]` but I thought this would be another unnecessary allocation thus I used a delegate

> but maybe the Registry shouldn't use a RegistryAPI to contact an upstream registry

Hm, caching is happening within Registry itself, not `DNSResolver`.

Geod24:
> No, `getValidatorInternal` calls `answerQuestion` internally which will check for the registry cache

Where ? [I don't see it in this function](https://github.com/bosagora/agora/blob/a154651edaacc086395e5820f1430751c0e9d2b5/source/agora/node/Registry.d#L338)

> `getValidatorInternal` can return `Address[]` but I thought this would be another unnecessary allocation thus I used a delegate

But the delegate unconditionally allocates (`format` and `~=`).

mkykadir:
> Where ? [I don't see it in this function](https://github.com/bosagora/agora/blob/a154651edaacc086395e5820f1430751c0e9d2b5/source/agora/node/Registry.d#L338)

[`answer`](https://github.com/bosagora/agora/blob/a154651edaacc086395e5820f1430751c0e9d2b5/source/agora/node/Registry.d#L423) > [here](https://github.com/bosagora/agora/blob/a154651edaacc086395e5820f1430751c0e9d2b5/source/agora/node/Registry.d#L1273-L1276) checks for the cache, if no record can be found it asks and caches from other registry node

> But the delegate unconditionally allocates (format and ~=).

I can refactor it to return `Address[]` but wouldn't we end up with duplicating `Message.answers` to keep `Manager` away from DNS stuff?

Geod24 (approved):

PR #3050, opened 2022-02-15
Improve TestNet configuration and allow dynamic ConsensusParams

Geod24:
This fixes two major issues with the current TestNet setting:
- A network reset requires a new release and users to manually upgrade;
- Using '--testnet' does not work when a one wants to run a validator;

We fix both issues by first adding a new endpoint to the API, which returns the ConsensusParams.
The current approach is not ideal as ConsensusParams and ConsensusConfig are slightly different,
but it will cover our immediate need. The second issue is fixed by only overriding the consensus
section whenever we detect there is a configuration file we can read.
The realm, the registry address, and node.testing still need to be set correctly,
but the sample config file does this and people are likely to start off of it.

### DIFF
--- a/source/agora/node/main.d
+++ b/source/agora/node/main.d
@@ -137,26 +137,7 @@ private int main (string[] args)
     Nullable!Config configN = ()
     {
         if (cmdln.testnet)
-        {
-            import agora.common.DNS;
-            import core.time;
-
-            Config defaultConfig = {
-                node: {
-                    testing: true,
-                    realm: Domain.fromSafeString("testnet.bosagora.io."),
-                    registry_address: Address("http://ns1.bosagora.io"),
-                },
-                network: TestNetNodes,
-                consensus: {
-                    validator_cycle: 20,
-                    block_interval: 1.minutes,
-                    // Mon Feb 14 2022 04:30:00 GMT.
-                    genesis_timestamp: 1644_813_000,
-                },
-            };
-            return Nullable!Config(defaultConfig);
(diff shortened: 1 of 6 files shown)


Geod24:
This works and can be tested (deployed on ns1).

source/agora/node/main.d
```diff
@@ -322,3 +303,61 @@ shared static this ()
...
+{
+    import agora.api.FullNode;
+    import agora.common.DNS;
+    import vibe.web.rest;
+    static import std.file;
+    import core.time;
+
+    // If the user provided a configuration, they likely want to run
+    // a validator, or have some special config (e.g. interfaces, logging...).
+    // In this case, we just override the `consensus` part, and set `network`
```
omerfirmak, line 318: should we `ensure(config.consensus == ConsensusConfig.init)` before overriding or log the new consensus config?

Geod24: I decided against the former because this option is for non-technical people so it should "just work".
As for the later, we've been logging the config since forever in Faucet and it hasn't been useful a single time. I added a logging statement when I tested it to make sure things worked as expected, but I don't think a permanent one is going to be more useful than in Faucet, unless you feel strongly about it.

omerfirmak: Faucet's config is static tho, this one is dynamically fetched and if a user wants to look up what config they are running they have no means to do that, right?

omerfirmak (approved):

PR #3052, opened 2022-02-15
Trivial: Improve 'Connection task limit reached' message

Geod24:
It is not directly obvious what is wrong when the message is triggered,
and it is currently being triggered in production.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -609,8 +609,12 @@ public class NetworkManager
         {
             if (this.connection_tasks.length >= MaxConnectionTasks)
             {
-                log.info("Connection task limit reached. Will trying again in {}..",
-                    this.config.node.network_discovery_interval);
+                log.info("Connection task limit reached ({}/{}). Will try again in {}. {} addresses in queue.",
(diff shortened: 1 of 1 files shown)


source/agora/network/Manager.d
```diff
@@ -609,8 +609,10 @@ public class NetworkManager
         {
             if (this.connection_tasks.length >= MaxConnectionTasks)
             {
-                log.info("Connection task limit reached. Will trying again in {}..",
-                    this.config.node.network_discovery_interval);
+                log.info("Connection task limit reached ({}/{}). Will try again in {}. {} addresses in queue.",
+                         this.connection_tasks.length, MaxConnectionTasks,
+                         this.config.node.network_discovery_interval,
+                         this.todo_addresses.length);
```
omerfirmak, line 615: It would be nice if we could log `this.connection_tasks.keys` and `this.todo_addresses`

omerfirmak (approved):

source/agora/network/Manager.d
Geod24: Updated

Issue #3030, opened 2022-02-10
FullNode cannot find other peers when just `registry` is enabled


mkykadir:
NetworkManager is responsible for discovering other peers. When a FullNode is started from scratch with just `registry` configuration enabled and no `network` peer is configured; it loops over

```Doing periodic network discovery: 0 required peers requested, 0 missing, known 6``` 

message and cannot find any peer. Main [cause](https://github.com/bosagora/agora/blob/6b59d77a4bb8231b3b86304ad4aed4c182a36c08/source/agora/network/Manager.d#L537) of this is `registry_client` is not initialized within `NetworkManager`, it is [only](https://github.com/bosagora/agora/blob/6b59d77a4bb8231b3b86304ad4aed4c182a36c08/source/agora/network/Manager.d#L505) initialized when node is `Validator`.

Geod24:
Fixed in: https://github.com/bosagora/agora/pull/3053

PR #3053, opened 2022-02-15
Fix #3048, #3030: Set 'registry_address' as explicitly set in TestNet

### LINKED ISSUE #3048: Warnings about lookup failure spamming the log
Since https://github.com/bosagora/agora/pull/3028 this is constantly triggering: https://github.com/bosagora/agora/blob/f061636d4643f8a7054479792ce35d2c9247dedc/source/agora/network/Manager.d#L566

Geod24:
I think it's time to fix configy. In the meantime, this fixes Agora. Packed a few improvements along the way.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -554,31 +554,32 @@ public class NetworkManager
                 if (key == this.config.validator.key_pair.address)
                     continue;
 
-                taskman.runTask
-                ({
+                taskman.runTask(
+                () nothrow {
                     // https://github.com/bosagora/agora/issues/2197
                     const ckey = key;
-                    retry!
-                    ({
+                    try
+                    {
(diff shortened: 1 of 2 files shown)


source/agora/network/Manager.d
```diff
@@ -554,31 +554,32 @@ public class NetworkManager
                 if (key == this.config.validator.key_pair.address)
                     continue;
 
-                taskman.runTask
-                ({
+                taskman.runTask(
```
omerfirmak, line 557: We probably shouldnt spawn a task either

omerfirmak (approved):

source/agora/network/Manager.d
Geod24: The reason for spawning a task is to avoid doing all clients serially. Not great but I see the reasoning. Will look into removing it soon.

PR #3058, opened 2022-02-16
Add instructions for TestNet in Korean

Geod24:
@linked0 : Would appreciate a review. I'll leave a couple note, too.

PR #3060, opened 2022-02-17
Fix the release version on TestNet README

linked0:
Fixed the version to the latest `v0.35.0`.


Geod24:
I originally went for "at the time of writing" so we would not have to continuously update it.
But a much better option is just to link to "latest", as done [here](https://github.com/bosagora/agora/pull/3058/commits/4635d0aa632847cddceab40d6a39f65702e9b12f).

Geod24:
Superseded by #3058

PR #3061, opened 2022-02-17
Refactor Nominator timer start/stop/arm cycles

Geod24:
The previous code would have triggered a 'log' message on every call to 'checkNomination',
which would be a lot of noise for the user.
The names were also not matching the functions anymore, and some documentation was missing.


hewison-chris (approved):
LGTM

PR #3067, opened 2022-02-17
Clock: Use UTC time everywhere

Geod24:
I was actually working on using the `Ledger.height` vs expected height (last commit) when I noticed our call to `time`, which doesn't handle timezone, meaning people joining the network would see massive offsets.

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -1131,7 +1131,7 @@ public class FullNode : API
     public override TimePoint getLocalTime () @safe nothrow
     {
         this.recordReq("local_time");
-        return this.clock.localTime();
+        return this.clock.utcTime();
     }
 
     /***************************************************************************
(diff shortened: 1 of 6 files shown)


source/agora/node/FullNode.d
```diff
@@ -1131,7 +1131,7 @@ public class FullNode : API
     public override TimePoint getLocalTime () @safe nothrow
     {
         this.recordReq("local_time");
```
mkykadir, line 1133: Trivial

```suggestion
        this.recordReq("utc_time");
```

mkykadir (approved):

source/agora/node/FullNode.d
Geod24: No, the endpoint is still `GET /local_time`. We can change it later but I tend to avoid those and just pack them in one commit / release.

hewison-chris (approved):
LGTM

PR #3068, opened 2022-02-17
Nominator: Wait increasingly more between nomination timer ticks

omerfirmak:
Hoping that it will allow network to reach consensus before a new
nomination extends the round.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -1469,7 +1469,7 @@ extern(D):
 
     ***************************************************************************/
 
-    protected override milliseconds computeTimeout(uint32_t roundNumber)
+    protected override milliseconds computeTimeout (uint32_t roundNumber)
     {
         auto base = 1.seconds;
         // double the timeout every 16 validators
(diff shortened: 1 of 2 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -499,7 +504,8 @@ extern(D):
     {
         scope (exit)
         {
-            this.armTaskTimer(TimersIdx.Nomination, this.nomination_interval);
+            this.armTaskTimer(TimersIdx.Nomination,
+                this.nomination_interval * (1 + this.nom_fire_count));
```
Geod24: Exponential backoff maybe ?

omerfirmak: Our "jitter" implementation is plain wrong, it will need to be fixed first. We can switch to using exponential backoff after that. I will have a seperate PR for it.

omerfirmak:
Superseded by https://github.com/bosagora/agora/commit/1aeb251517c06db272cdd00d40c7a716f5be2bdf

PR #3071, opened 2022-02-17
Manager: Add a range to iterate over mutable peer list

omerfirmak:
We used to be not able to remove from peer list since multiple
fibers could be iterating over it at any moment without any
synchronization between them. With PeerRange, NetworkManager
can remove/add to the peer_list and invalidate all ranges over the
list by incrementing the peer_list_version.

Ranges will start back from the beginning, so caller can iterate
over the same peers multiple times but that probably we a rare
occurrence and even if it did happen, not much would change.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -234,7 +234,11 @@ public class NetworkManager
     protected ConnectionTask[Address] connection_tasks;
 
     /// All connected nodes (Validators & FullNodes)
-    public DList!NetworkClient peers;
+    private DList!NetworkClient peer_list;
+
+    /// Version of peer_list
+    /// incremented to signal PeerRanges that they are invalidated
+    private uint peer_list_version;
 
     /// The list of addresses we have not yet tried to connect to
     protected Set!Address todo_addresses;
@@ -282,6 +286,12 @@ public class NetworkManager
             this.addAddresses(resolveDNSSeeds(config.dns_seeds, this.log));
     }
 
+    ///
+    public auto peers () @safe nothrow pure
+    {
+        return PeerRange(&this.peer_list_version, &this.peer_list);
+    }
+
     /// Returns an already instantiated version of the BanManager
     /// (please also see `NetworkManager.getBanMananger()`)
     public BanManager getBanManager () @safe @nogc nothrow pure
@@ -300,7 +310,7 @@ public class NetworkManager
         // We have an authenticated client, maybe we already have a client for it
         if (key !is PublicKey.init)
         {
-            auto existing_peers = this.peers[].find!(p => p.identity.key == key);
+            auto existing_peers = this.peers.find!(p => p.identity.key == key);
             if (!existing_peers.empty())
             {
                 auto prev = existing_peers.front();
@@ -341,7 +351,7 @@ public class NetworkManager
 
         if (!client.connections.any!(conn => address == Address.init))
         {
-            this.peers.insertBack(client);
+            this.peer_list.insertBack(client);
 
             if (key !is PublicKey.init)
             {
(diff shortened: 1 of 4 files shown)


source/agora/network/Manager.d
```diff
@@ -1213,3 +1223,52 @@ private Set!Address resolveDNSSeeds (in string[] dns_seeds, ref Logger log)
 
     return resolved_ips;
 }
+
+/// A range over the list of peers that can be invalidated
+struct PeerRange
```
Geod24: `public`

source/agora/network/Manager.d
```diff
@@ -1213,3 +1223,52 @@ private Set!Address resolveDNSSeeds (in string[] dns_seeds, ref Logger log)
...
+{
+    private
+    {
+        uint* manager_version;
```
Geod24: Visibility per field / function please

source/agora/network/Manager.d
```diff
@@ -234,7 +234,11 @@ public class NetworkManager
     protected ConnectionTask[Address] connection_tasks;
 
     /// All connected nodes (Validators & FullNodes)
-    public DList!NetworkClient peers;
+    private DList!NetworkClient peer_list;
+
+    /// Version of peer_list
+    /// incremented to signal PeerRanges that they are invalidated
+    private uint peer_list_version;
```
Geod24, line 241: I think you could store a `PeerRange` here directly, couldn't you ? `manager_version` would then just be an internal pointer.

source/agora/network/Manager.d
```diff
@@ -1213,3 +1223,52 @@ private Set!Address resolveDNSSeeds (in string[] dns_seeds, ref Logger log)
...
+
+    this (uint* manager_version, DList!NetworkClient* peers) nothrow @safe pure
+    {
+        this.manager_version = manager_version;
+        this.peers = peers;
+        this.range_version = *this.manager_version;
+        this.range = (*this.peers)[];
+    }
+
+    void popFront() nothrow @safe
```
Geod24: Style (space before parenthesis)

Geod24:
> Ranges will start back from the beginning, so caller can iterate
> over the same peers multiple times but that probably we a rare
> occurrence and even if it did happen, not much would change.

Couldn't we end up with an infinite loop if someone was to frequently connect / disconnect ?

Geod24 (commented):
Few comments, the idea is interesting, did you test it live ?

omerfirmak:
> > Ranges will start back from the beginning, so caller can iterate
> > over the same peers multiple times but that probably we a rare
> > occurrence and even if it did happen, not much would change.
> 
> Couldn't we end up with an infinite loop if someone was to frequently connect / disconnect ?

Technically we could, but we are the one who connects and disconnects so it cant be triggered by an attacker. But we could put an upper limit on how many times `PeerRange` could start over.

source/agora/network/Manager.d
omerfirmak: > manager_version would then just be an internal pointer.
 
 what do you mean exactly? I feel like `PeerRange` definitely requires pointers to `manager_version` and the `manager_peer_list`.

Geod24: Something along this line: https://github.com/Geod24/agora/commit/cb49d6ae5b0b216234a2870052ce6a9e07f93a08
But I'm not sure it's worth it :D

Geod24:
I don't see either `manager_version` or `range_version` being changed tho

omerfirmak:
> I don't see either `manager_version` or `range_version` being changed tho

because we dont remove any peers, we just ban them. Adding to the end of the list is safe to do even with active ranges.
right now I am working on removing peers that we've banned.

Geod24:
Fair enough. Please fix the style issues and let's merge. As for the diff I showed, take it or leave it, up to you.

source/agora/network/Manager.d
omerfirmak: That will probably work but made my head hurt :D

omerfirmak:
> Please fix the style issues

done

Geod24 (approved):

PR #3072, opened 2022-02-18
Disabled Tracy by default

omerfirmak:


dub.json
```diff
@@ -43,6 +43,16 @@
             "targetName": "agora",
             "mainSourceFile": "source/agora/node/main.d",
             "debugVersions": [ "AgoraDebugAPIs" ],
```
Geod24, line 45: Maybe remove the debug API too ?

omerfirmak: I would rather keep it during the testnet, in case we run in to a consensus problem.

Geod24 (approved):

PR #3080, opened 2022-02-21
Manager: Remove non-enrolled validators from peer list

### LINKED ISSUE #2868: Implement peer re-negotiation

omerfirmak:
Will work both when a validator is no longer enrolled or changes
its stake.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -268,6 +268,9 @@ public class NetworkManager
 
     protected agora.api.FullNode.API owner_node;
 
+    /// Most recent validator set
+    protected UTXO[Hash] last_known_validator_utxos;
+
     /// Ctor
     public this (in Config config, ManagedDatabase cache, ITaskManager taskman, Clock clock, agora.api.FullNode.API owner_node)
     {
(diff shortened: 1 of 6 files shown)


omerfirmak:
Fixes #2868

source/agora/network/Manager.d
```diff
@@ -535,11 +538,21 @@ public class NetworkManager
...
 
+    /// Called when validator set is changed
+    public void onValidatorSetChanged (ref UTXO[Hash] validator_utxos) @trusted
+    {
+        this.last_known_validator_utxos = validator_utxos;
+
+        foreach (val; this.validators())
+            if (val.identity.utxo !in this.last_known_validator_utxos)
+                if (this.peer_list.linearRemoveElement(val))
+                    this.peer_list_version++;
```
Geod24: Doesn't look like this is closing the connection though, just removing it from the list ?

source/agora/network/Manager.d
```diff
@@ -535,11 +538,21 @@ public class NetworkManager
             &this.onRegisterName, Periodic.Yes);
     }
 
+    /// Called when validator set is changed
+    public void onValidatorSetChanged (ref UTXO[Hash] validator_utxos) @trusted
```
Geod24: Why `ref` ?

omerfirmak: Oops, AA's are already reference types right?

Geod24: Genau

source/agora/network/Manager.d
omerfirmak: updated to call `shutdown()` on the client

omerfirmak:
green

hewison-chris (approved):
LGTM

Geod24 (approved):

PR #3082, opened 2022-02-21
Manager: Simplify removal of connection tasks

omerfirmak:
I have seen lingering connection tasks in production. We remove them
in various points and they are mostly guarded by a condition. This way
we ensure they get removed from the AA when they exit.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -154,6 +154,7 @@ public class NetworkManager
             catch (Exception exc)
                 log.error("Unexpected exception while contacting {}: {}",
                           this.address, exc);
+            this.outer.connection_tasks.remove(this.address);
         }
 
         /// Ditto, just behind a trampoline to avoid function-wide try/catch
@@ -211,7 +212,6 @@ public class NetworkManager
                 {
                     // either we connected to ourself, or someone else is pretending
                     // to be us
-                    this.outer.connection_tasks.remove(address);
                     this.outer.banman.ban(address);
                     return;
                 }
@@ -305,7 +305,6 @@ public class NetworkManager
         in Hash utxo, in PublicKey key)
     {
         log.dbg("onHandshakeComplete: {} - (k: {}, utxo: {})", address, key, utxo);
-        this.connection_tasks.remove(address);
 
         // We have an authenticated client, maybe we already have a client for it
         if (key !is PublicKey.init)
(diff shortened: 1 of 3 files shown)


source/agora/network/Manager.d
```diff
@@ -162,9 +163,7 @@ public class NetworkManager
             if (this.api is null)
                 this.api = this.outer.makeClient(this.address);
 
-            PublicKey key;
-            Hash utxo;
-            while (1)
+            foreach (_; 0..this.outer.config.node.max_retries)
```
mkykadir:
```suggestion
            foreach (_; 0 .. this.outer.config.node.max_retries)
```

source/agora/network/Manager.d
```diff
@@ -599,11 +595,7 @@ public class NetworkManager
         bool onFailedRequest (in Address address) nothrow
         {
             if (this.banman.isBanned(address))
```
mkykadir:
```suggestion
            return !(this.banman.isBanned(address))
```

mkykadir (approved):
Trivial changes

source/agora/network/Manager.d
```diff
@@ -190,9 +189,20 @@ public class NetworkManager
...
-                    key = id.key;
-                    break;
+                    if (id.key != PublicKey.init)
+                    {
+                        if (this.address !is Address.init
+                            && id.key == this.outer.config.validator.key_pair.address)
+                        {
+                            // either we connected to ourself, or someone else is pretending
+                            // to be us
+                            this.outer.banman.ban(address);
```
Geod24: We'd still have a problem with whitelisting here, right ?

Geod24 (approved):
LGTM, just @mkykadir 's review.

Geod24:
Also `connection`

source/agora/network/Manager.d
omerfirmak: The way we whitelist validators is not the best. Unless we were already connected to it, whitelisting has no affect,

https://github.com/bosagora/agora/blob/88b23cd06fd9fe47025dbe88b5865c829e2f5b3b/source/agora/network/Manager.d#L1148-L1152

omerfirmak:
Updated

Geod24:
Timeout failed for both run, re-triggered, but I don't think it's spurious.

omerfirmak:
green

PR #3084, opened 2022-02-21
Validator: Reveal preimages according to expected height

### LINKED ISSUE #3077: Integration test should use a recent (dynamic) genesis timestamp
Otherwise the Ledger will go crazy thinking it's many weeks late.

omerfirmak:
When network falls behind on blocks, it creates a lot of blocks
in a short span and this results in some nodes to be slashed in the
process. If we reveal the preimage of the expected height instead
of the ledger height, we should decrease the chance of us getting
slashed in this scenario.

### DIFF
--- a/ci/system_integration_test.d
+++ b/ci/system_integration_test.d
@@ -4,8 +4,11 @@ module system_integration_test;
 private:
 
 import std.algorithm;
+import std.array;
+import std.conv;
 import std.file;
 import std.format;
+import std.datetime;
 import std.path;
 import std.process;
 import std.stdio;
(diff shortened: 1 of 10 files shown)


Geod24:
LGTM, but integration failure is probably because https://github.com/bosagora/agora/issues/3077 + the limitation on long-range pre-image to avoid DoS.

Geod24 (approved):

omerfirmak:
@Geod24 , added a second commit to fix #3077 
will merge when CI is green

ci/system_integration_test.d
```diff
@@ -49,6 +52,17 @@ immutable Cleanup = [ "rm", "-rf", IntegrationPath.buildPath("node/0/.cache/"),
                       IntegrationPath.buildPath("node/6/.cache/"),
                       IntegrationPath.buildPath("node/7/.cache/"),
 ];
+auto SetGenesisTimestamp = ["sed", "-i",
+    "s/genesis_timestamp: [0-9]\\+/genesis_timestamp: curr_time/g",
```
Geod24, line 56: Wild, didn't think of this, neat.

Geod24: Well, with the exception that running this script dirties the repository, but I guess that's okay because it's in the CI script.

source/agora/node/Validator.d
```diff
@@ -685,7 +685,7 @@ public class Validator : FullNode, API
     {
         PreImageInfo preimage;
         if (this.enroll_man.getNextPreimage(preimage,
-            this.ledger.height()))
+            max(this.ledger.height(), this.ledger.expectedHeight(this.clock.utcTime()))))
```
hewison-chris, line 688: I think this is what is causing the recent `ci` failures. If the expected height is beyond the end of the enrollment cycle then `getNextPreimage` returns `ulong.max` as there is no enrollment at this height yet. Rather than revert this commit I will create a PR to fix it.

PR #3087, opened 2022-02-22
Backoff: Fix jitter

omerfirmak:
Apply 5% jitter to the delay we calculated.

### DIFF
--- a/source/agora/utils/Backoff.d
+++ b/source/agora/utils/Backoff.d
@@ -28,7 +28,8 @@ import std.algorithm : min;
 
     The algorithm is simply:
     ```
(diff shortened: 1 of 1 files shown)


source/agora/utils/Backoff.d
```diff
@@ -62,13 +63,13 @@ public uint getDelay (alias RandomFunc = getRandom)
       value = the input value
 
     Returns:
-      A random value between [0, value], inclusive
+      A random value between [0, value / 20], inclusive
 
 *******************************************************************************/
 
 public uint getRandom (uint value) @safe nothrow @nogc
```
hewison-chris: How about renaming to jitter?

hewison-chris (approved):
LGTM

PR #3088, opened 2022-02-22
Nominator: Move D symbols to extern (D) scope

omerfirmak:
So we dont need to `extern (D)` them explicitly


hewison-chris (approved):
Was thinking of doing the same recently. Tricky to review but I trust it is just moved code.

Geod24 (approved):

PR #3089, opened 2022-02-22
Nominator: override SCP computeTimeout

omerfirmak:
similar to the default implementation but also scales with the
number of validators in the network.


hewison-chris (approved):
LGTM

PR #3090, opened 2022-02-22
Update and improve instructions for testnet

Geod24:
Use https as http gives a 301, mention that -p should be used for Docker,
and provide a template to make it even easier to become a validator.


mkykadir (approved):

linked0:
Will update the Korean version of README.

PR #3096, opened 2022-02-23
Integration tests: Use node-0 as DNS server

Geod24:
Let's see what the CI thinks of this.


Geod24:
Well we probably need to cross the Rubicon and parse `/etc/resolv.conf` for this to work.

PR #3107, opened 2022-02-24
Doc: Update TestNet documentation about consensus parameters

mkykadir:
Consensus parameters are needed to sync with network and `--testnet` flag
does this by default. Documentation was misleading when providing own
configuration.

### DIFF
--- a/devel/testnet/README.md
+++ b/devel/testnet/README.md
@@ -71,6 +71,7 @@ Agora can be extensively configured via the use of a configuration file.
 By default, Agora will look for `config.yaml` in the directory it is started in, but using the `-c`
 command line option, you can give Agora any configuration file.
 Using `--testnet` is a special way to make Agora start without a configuration file.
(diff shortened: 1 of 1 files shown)


devel/testnet/README.md
```diff
@@ -71,6 +71,8 @@ Agora can be extensively configured via the use of a configuration file.
 By default, Agora will look for `config.yaml` in the directory it is started in, but using the `-c`
 command line option, you can give Agora any configuration file.
 Using `--testnet` is a special way to make Agora start without a configuration file.
+`--testnet` flag is still required to fetch `consensus` parameters from test network,
+even a configuration file is provided.
```
Geod24: That should be "even **if**". But I'd rephrase it as just saying something along the line of"Always use `--testnet`"

mkykadir: Updated

Geod24 (approved):

Issue #2942, opened 2022-01-28
Race condition on shutdown can cause SEGV


Geod24:
The small commit introduced a regression:
```
Unittesting agora.test.Flash#### FATAL ERROR: index [0] exceeds array of length 0
This node was started at source/agora/test/Flash.d:914
This most likely means that the node crashed due to an uncaught exception
If not, please file a bug at https://github.com/Geod24/localrest/
Full error: core.exception.ArrayIndexError@source/agora/node/FullNode.d(1159): index [0] exceeds array of length 0
----------------
core/exception.d:875 onArrayIndexError [0x11132fc6c]
core/exception.d:848 _d_arraybounds_index [0x1113301b5]
source/agora/node/FullNode.d:1159 nothrow @safe void agora.node.FullNode.FullNode.pushBlock(const(agora.consensus.data.Block.Block)).__lambda7() [0x110665b09]
core/thread/context.d:46 void core.thread.context.Callable.opCall() [0x11135c34c]
core/thread/fiber.d:1151 void core.thread.fiber.Fiber.run() [0x11135c4d8]
core/thread/fiber.d:198 fiber_entryPoint [0x11135c415]
./ci/run.sh: line 21:  3795 Abort trap: 6           ./build/agora-unittests
```

_Originally posted by @Geod24 in https://github.com/bosagora/agora/issues/2915#issuecomment-1024529269_

This is because the task is still running (but shouldn't).

Geod24:
We probably need a way to shut down the task manager. We might even want to, IDK, name tasks ?

hewison-chris:
I think this was fixed by 
https://github.com/bosagora/agora/commit/e86c0d06d1761bd4470628cf85d4d2d7052b3e63
and
https://github.com/bosagora/agora/commit/093103ba12d25ef0cc85056a07bfb3d1951bcdb1

hewison-chris:
However this might still be related https://github.com/bosagora/agora/issues/3079

linked0:
This could be solved by PR #3109.

linked0:
This is probably solved by [PR #3109](https://github.com/bosagora/agora/pull/3109). So closing this.

linked0:
I see the issue happens again. So reopening this.

PR #3109, opened 2022-02-25
Prevent possible crashes about handlers on shutting down

### LINKED ISSUE #3079: Check if we need to set the array of timers to null
This cannot prevent accessing the array after it being `null` because the flow could already be in the `foreach`, for example. So it would be better to not set `null` to the arrays in the `shutdown` as not setting `null` to the arrays makes no harm, I think.

_Originally posted by @linked0 in https://github.com/bosagora/agora/pull/3078#discussion_r810819032_

linked0:
This prevents a node from crashing when it accesses the event handlers while it's shutting down.

Fixes #3079, #2942

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -1212,8 +1212,6 @@ public class FullNode : API
 
     private void pushBlock (const Block block) @trusted
     {
-        if (this.is_shutting_down)
-            return;
         const now = StdClock.currTime();
         foreach (index, ref handler; this.block_handlers)
         {
@@ -1223,6 +1221,9 @@ public class FullNode : API
             this.taskman.runTask({
                 // Work around potential DMD bug
                 const idx = index;
+                if (this.block_handlers.length <= idx)
+                    return;
+
                 try
                 {
                     this.block_handlers[idx].client.pushBlock(block);
@@ -1255,8 +1256,6 @@ public class FullNode : API
 
     protected void pushBlockHeader (in BlockHeader header) @trusted
     {
-        if (this.is_shutting_down)
-            return;
         const now = StdClock.currTime();
         foreach (index, ref handler; this.block_header_handlers)
         {
@@ -1266,6 +1265,9 @@ public class FullNode : API
             this.taskman.runTask({
                 // Work around potential DMD bug
                 const idx = index;
+                if (this.block_header_handlers.length <= idx)
+                    return;
+
                 try
                 {
                     this.block_header_handlers[idx].client.pushBlockHeader(header);
@@ -1296,8 +1298,6 @@ public class FullNode : API
 
     protected void pushPreImage (const PreImageInfo pre_image) @trusted
     {
-        if (this.is_shutting_down)
-            return;
         const now = StdClock.currTime();
         foreach (index, ref handler; this.preimage_handlers)
         {
@@ -1307,6 +1307,9 @@ public class FullNode : API
             this.taskman.runTask({
                 // Work around potential DMD bug
                 const idx = index;
+                if (this.preimage_handlers.length <= idx)
+                    return;
+
                 try
                 {
                     this.preimage_handlers[idx].client.pushPreImage(pre_image);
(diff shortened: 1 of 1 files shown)


source/agora/node/FullNode.d
```diff
@@ -1205,8 +1205,6 @@ public class FullNode : API
 
     private void pushBlock (const Block block) @trusted
     {
-        if (this.is_shutting_down)
-            return;
```
hewison-chris, line 1216: Why not keep this anyway? We are shutting down so no need to continue.

linked0: I thought as you do but I think it's not so big impact because it could do only one iteration even if it enters the `foreach` scope. But it's better to keep it.

source/agora/node/FullNode.d
```diff
@@ -1205,27 +1205,32 @@ public class FullNode : API
...
                 continue;
 
             this.taskman.runTask({
                 // Work around potential DMD bug
                 const idx = index;
+                if (this.block_handlers.length <= idx)
+                    return;
+
+                auto _handler = this.block_handlers[idx];
+
```
hewison-chris: You can remove these other changes now as if not shutting down the array will not be null.

hewison-chris (changes requested):
I think moving the check for shutting down is sufficient to prevent the chance of crashing.

source/agora/node/FullNode.d
linked0: It's for accessing the `this.block_handlers` after the variable has been set to null which means the handler is an empty array. This is kind of related to #2942.

hewison-chris (approved):
LGTM

linked0:
This PR is continuously failing on CI since last Friday, especially on `ValidatorRecurringEnrollment`. I think the failure is totally unrelated to this PR. So this can be merged. @hewison-chris Can you merge this if you agree with me?

The failure logs are as follows. This failure also happens frequently on other PRs.
```
Module tests failed: agora.test.ValidatorRecurringEnrollment
core.exception.AssertError@source/agora/test/Base.d(797): Check condition failed after timeout of 10 secs and 100 attempts: Client #3 (agora://10.0.0.3/) has no preimage for client #2 (agora://10.0.0.2/) at height 21
```
and
```
Module tests failed: agora.test.ValidatorRecurringEnrollment
core.exception.AssertError@source/agora/test/ValidatorRecurringEnrollment.d(134): Check condition failed after timeout of 10 secs and 100 attempts: Node 0 has block height 21. Expected: 22
----------------
source/agora/utils/Utility.d:146 void agora.utils.Utility.retryFor(lazy bool, core.time.Duration, lazy immutable(char)[], immutable(char)[], ulong) [0x5632b5e7afbf]
<OMITTED>
source/agora/test/ValidatorRecurringEnrollment.d:134 void agora.test.ValidatorRecurringEnrollment.__unittest_L85_C1() [0x5632b63d08ee]
```

hewison-chris:
> This PR is continuously failing on CI since last Friday, especially on `ValidatorRecurringEnrollment`. I think the failure is totally unrelated to this PR. So this can be merged. @hewison-chris Can you merge this if you agree with me?
> 
> The failure logs are as follows. This failure also happens frequently on other PRs.
> 
> ```
> Module tests failed: agora.test.ValidatorRecurringEnrollment
> core.exception.AssertError@source/agora/test/Base.d(797): Check condition failed after timeout of 10 secs and 100 attempts: Client #3 (agora://10.0.0.3/) has no preimage for client #2 (agora://10.0.0.2/) at height 21
> ```
> 
> and
> 
> ```
> Module tests failed: agora.test.ValidatorRecurringEnrollment
> core.exception.AssertError@source/agora/test/ValidatorRecurringEnrollment.d(134): Check condition failed after timeout of 10 secs and 100 attempts: Node 0 has block height 21. Expected: 22
> ```

I have started to look into why these `ValidatorRecurringEnrollment` tests are failing and hope to solve it today. If you really think we need the changes in this PR to be merged then can you add a commit to just skip the two failing tests?  Then we can make sure all the other tests that run after them are still passing.

linked0:
This is an important and simple bug fix but this fails constantly on `ValidatorRecurringEnrollment` which is not related to this PR. Fixing the failure for the test might need more time so I'm merging it with one approval.

PR #3112, opened 2022-02-25
Revert 3110 signature logging

hewison-chris:


linked0 (approved):

hewison-chris:
Instead of revert will make new PR.

PR #3115, opened 2022-02-26
DNS: Compare ResourceRecords with their data

mkykadir:
ResourceRecord equality check was always failing since it was impossible to compare `rdata` without checking for `type` first.


mkykadir:
OK, this is **not required** for resolving any issues but simple equality check of `ResourceRecord` is not working without this. This is a trivial enhancement; we can close without merging.

omerfirmak (approved):

omerfirmak:
Merge when we get a green CI

PR #3118, opened 2022-02-28
Registry: Allow configuration of multiple NS records

### LINKED ISSUE #3102: All primary name servers should be returned for `NS` query
When a zone is queried for NS records, it should return all configured authoritative name servers. Currently secondary registry nodes are not publishing their name to a primary registry zone; thus when NS record is queried only primary registry zone is returned.

For example;
```
$ dig validators.testnet.bosagora.io -t NS
;; ANSWER SECTION:
validators.testnet.bosagora.io. 0 IN    NS      ns1.bosagora.io.
ns1.bosagora.io.        0       IN      A       188.165.202.109
```
above query should also include NS record for `ns2.bosagora.io`

### LINKED ISSUE #3103: AXFR zone transfer should include `NS` records
NS records are not stored inside DB, thus AXFR zone transfer misses that.

mkykadir:
NS record indicates authoritative DNS servers for a zone; since Agora
allows multiple distributed authoritative servers for zones it should
also allow multiple NS records to be configured.

Fixes #3102 and Fixes #3103

### DIFF
--- a/devel/config-single.yaml
+++ b/devel/config-single.yaml
@@ -43,12 +43,14 @@ registry:
   port: 5335
   validators:
     authoritative: true
-    primary: localhost
+    nameservers:
+      - localhost
     soa:
       email: no@no.no
   flash:
     authoritative: true
-    primary: localhost
+    nameservers:
+      - localhost
     soa:
       email: no@no.no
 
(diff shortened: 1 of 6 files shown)


Geod24 (approved):
I have a few reservations here. Namely, I think we should just ahead and store only RR in the registry table. Having the NS being config only is inconsistent and will bite us back. But the rest looks sane and working, so let's move ahead with this. Thanks!

Geod24:
Rebased, trying to get a green run.

PR #3120, opened 2022-02-28
Skip validation for enrollments in the pool

omerfirmak:
```
and add new enrollments we encounter in nominations to the
pool
```

This should save us from a signature and commitment validation.


omerfirmak:
Added a test as well

mkykadir (approved):

PR #3148, opened 2022-03-01
Ledger: Dont validate UTXO still exists for nodes to be slashed

omerfirmak:
Stakes are locked for the duration of their enrollment, there is no
way they are spent.


Geod24 (approved):

linked0 (approved):
LGTM

PR #3156, opened 2022-03-02
PeerRange: Fix calling popFront on empty range

omerfirmak:
checkIfInvalidated() could cause range to be empty.


omerfirmak:
Actually same could happen in `front()`, but I dont have a fix for it yet.

mkykadir (approved):

PR #3165, opened 2022-03-04
Upgrade checking handlers for Stoa

### LINKED ISSUE #2942: Race condition on shutdown can cause SEGV

linked0:
After shutting down nodes, there are cases where nodes crash
because the handlers for Stoa are nullified. So this is a way
to keep from crashing by using the handler structure itself, not
the index.

Fixes #2942

### DIFF
--- a/source/agora/node/FullNode.d
+++ b/source/agora/node/FullNode.d
@@ -1222,19 +1222,22 @@ public class FullNode : API
             this.taskman.runTask({
                 // Work around potential DMD bug
                 const idx = index;
-                if (this.block_handlers.length <= idx)
+                HandlerInfo!(BlockExternalizedHandler) handler;
+                if (this.block_handlers.length > idx)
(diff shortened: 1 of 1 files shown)


linked0:
Ready for review. The spurious error on CI happens tho.

Geod24:
This is not fixing the issue IMO, for the same reason that previous changes didn't fix the issue.
After Ctrl+C is called and the shutdown is initiated, nothing should be called, because we cannot sanely shut down in the signal handler if things are going to fire randomly afterwards. However, we're seeing via this segv that things DO get called. Additionally, https://github.com/bosagora/agora/issues/3158 seems to suggest that this trigger in prod, too.

PR #3166, opened 2022-03-04
FullNode: Update ledger after accepting header

### LINKED ISSUE #3164: FullNode: Block signature update is repeated even block is accepted at first
FullNode's `acceptHeader` doesn't update `BlockHeader` within Ledger; thus even Signature is fetched successfully update request is sent repeatedly for the same block.

https://github.com/bosagora/agora/blob/4bd472597ec6b63097f7b9fc850c471c60cf140f/source/agora/node/FullNode.d#L620-L623

mkykadir:
Fixes #3164 and reduces GC operation in `doCatchUp`.

Following memory trace is from `v0.x.x` where `doCatchUp` always checks again for missing signatures between peers
v0.x.x

Following memory trace is longer duration than the above one but GC impact is improved, accepted block headers are removed from AA between peers
new

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -903,7 +903,7 @@ public class NetworkManager
 
     public void getMissingBlockSigs (Ledger ledger,
         scope ulong delegate(BlockHeader) @safe extra_sigs,
-        scope void delegate(BlockHeader) @safe acceptHeader) @safe nothrow
+        scope string delegate(BlockHeader) @safe acceptHeader) @safe nothrow
     {
         import std.algorithm;
         import std.conv;
@@ -949,8 +949,14 @@ public class NetworkManager
                             {
                                 try
                                 {
-                                    log.trace("getMissingBlockSigs: updating header signature: {} validators: {}", header.signature, header.validators);
-                                    acceptHeader(header);
+                                    if (auto res = acceptHeader(header))
+                                        log.dbg("getMissingBlockSigs: couldn't update header ({})", header.height);
+                                    else
+                                    {
+                                        log.trace("getMissingBlockSigs: updated header ({}) signature: {} validators: {}",
+                                            header.height, header.signature, header.validators);
+                                        missing_heights.remove(header.height);
+                                    }
                                 }
                                 catch (Exception e)
                                 {
@@ -958,7 +964,6 @@ public class NetworkManager
                                 }
                             }
                         }
-                        missing_heights = heightsMissingSigs();
                         if (missing_heights.empty)
                             break;
                     }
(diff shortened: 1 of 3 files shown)


source/agora/network/Manager.d
```diff
@@ -949,16 +949,19 @@ public class NetworkManager
                             {
                                 try
                                 {
-                                    log.trace("getMissingBlockSigs: updating header signature: {} validators: {}", header.signature, header.validators);
-                                    acceptHeader(header);
+                                    auto res = acceptHeader(header);
+                                    if (res is null)
```
Geod24: Do we want to log the error, if any ? Possibly at a low level (trace) since it's data supplied by an untrusted party ?

source/agora/network/Manager.d
```diff
@@ -949,16 +949,19 @@ public class NetworkManager
...
+                                    {
+                                        log.trace("getMissingBlockSigs: updating header signature: {} validators: {}", header.signature, header.validators);
```
Geod24: Technically it's updated, since the operation has already been done

source/agora/network/Manager.d
Geod24: Oh it's done in a later commit

source/agora/network/Manager.d
```diff
@@ -952,9 +952,11 @@ public class NetworkManager
                                     auto res = acceptHeader(header);
                                     if (res is null)
                                     {
-                                        log.trace("getMissingBlockSigs: updating header signature: {} validators: {}", header.signature, header.validators);
+                                        log.trace("getMissingBlockSigs: updating header ({}) signature: {} validators: {}", header.height, header.signature, header.validators);
                                         missing_heights.remove(header.height);
                                     }
+                                    else
+                                        log.dbg("getMissingBlockSigs: couldn't update header ({})", header.height);
```
Geod24: Actually I'd suggest:
```D
if (auto res = acceptHeader(header))
    log.dbg("...");
else
{
    // ...
}
```

Geod24 (approved):
Looks good. Two things:
- Wanna fix the typo in `Recieved` ?
- The GC pattern is worrying: Fixing the bug makes it go smooth, but it's just because it doesn't fail to fetch signatures anymore. When we do fail in the future, the issue will manifest itself;

mkykadir:
> * The GC pattern is worrying: Fixing the bug makes it go smooth, but it's just because it doesn't fail to fetch signatures anymore. When we do fail in the future, the issue will manifest itself;

That spikes still happen through `getMissingBlockSigs`, this fix is one step improvement; still looking for another improvements.

hewison-chris (approved):
LGTM

PR #3184, opened 2022-03-14
Make TimePoint a struct instead of an alias

Geod24:
Had this diff locally since forever. Ideally we'll make it contain a `DateTime` as well but there are a few obstacles to that.

### DIFF
--- a/source/agora/network/Manager.d
+++ b/source/agora/network/Manager.d
@@ -482,9 +482,9 @@ public class NetworkManager
             if (node_time == 0)
                 continue;  // request failed
 
-            const req_delay = (this.clock.utcTime() - req_start).seconds;
+            const req_delay = (this.clock.utcTime() - req_start);
             const dist_delay = req_delay / 2;  // divide evently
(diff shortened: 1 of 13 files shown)


linked0 (approved):
LGTM

source/agora/network/Manager.d
```diff
@@ -482,10 +482,10 @@ public class NetworkManager
             if (node_time == 0)
                 continue;  // request failed
 
-            const req_delay = this.clock.utcTime() - req_start;
+            const req_delay = (this.clock.utcTime() - req_start).seconds;
             const dist_delay = req_delay / 2;  // divide evently
-            const offset = (node_time - dist_delay) - req_start;
-            offsets ~= TimeInfo(node.identity.key, node_time, req_delay, offset.seconds);
+            const offset = (node_time - req_start).seconds - dist_delay;
```
omerfirmak: We should probably rename `getLocalTime`, as it has been returning UTC time for some time.

omerfirmak (approved):

PR #3186, opened 2022-03-14
Trivial: Remove dependency to ocean.transition

Geod24:
This module was only imported once in the whole program,
and only for trivial aliases.


hewison-chris (approved):
LGTM

PR #3191, opened 2022-03-14
Increase SCP slot timeout

omerfirmak:
Our timeouts were a bit agressive for a globally distributed arbitrarily large network. Let's see if this will help decrease our externalization time for the current state of the production.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -1543,7 +1543,7 @@ extern(D):
 
     ***************************************************************************/
 
-    protected override milliseconds computeTimeout(uint32_t roundNumber)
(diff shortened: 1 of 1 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -1546,8 +1546,8 @@ extern(D):
     protected override milliseconds computeTimeout(uint32_t roundNumber)
```
mkykadir:
```suggestion
    protected override milliseconds computeTimeout (uint32_t roundNumber)
```
just minor, not necessary

mkykadir (approved):

hewison-chris:
Integration test restarted as it failed.

PR #3194, opened 2022-03-14
Manager: Remove AA usage while signature updating

mkykadir:
AA causes heavy GC operations and signature update is frequently checked.
Thus, AA is replaced with `find`.


hewison-chris (approved):
LGTM

Geod24 (approved):

PR #3201, opened 2022-03-15
ValidatorSet: Bypass far preimage check on adding new enrollment

omerfirmak:
If a Validator stays unenrolled more than `ValidatorCycle` amount of
blocks, it would never be able to reveal new preimages after that point.

### DIFF
--- a/source/agora/consensus/state/ValidatorSet.d
+++ b/source/agora/consensus/state/ValidatorSet.d
@@ -141,12 +141,10 @@ public class ValidatorSet
             return "A validator with the same public key is already enrolled";
 
         () @trusted {
-            if (this.getPreimage(enroll.utxo_key) == PreImageInfo.init)
-                this.db.execute("INSERT INTO preimages " ~
+            if (this.getPreimage(enroll.utxo_key).height <= height)
+                this.db.execute("INSERT OR REPLACE INTO preimages " ~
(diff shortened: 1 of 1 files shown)


omerfirmak:
This is the root cause of our full nodes in production falling behind.

Geod24 (changes requested):
This reverts https://github.com/bosagora/agora/commit/5784d240096667ad0195c7316fa3f5696e64c335

omerfirmak:
> This reverts [5784d24](https://github.com/bosagora/agora/commit/5784d240096667ad0195c7316fa3f5696e64c335)

Updated to also conform to the behaviour https://github.com/bosagora/agora/commit/5784d240096667ad0195c7316fa3f5696e64c335 expects

hewison-chris (commented):
Is the title still valid? Also at least one test needs an update as the failure looks related.

omerfirmak:
> Is the title still valid?

Yeah, I think it sill is. We are already checking if commitment is valid while validating the enrollment. There is no reason to use `addPreImage()` and force another validation with far preimage guard.

> Also at least one test needs an update as the failure looks related.

updated, should be good now.

omerfirmak:
Green

Geod24 (approved):

hewison-chris (commented):
LGTM

PR #3207, opened 2022-03-17
Internalize Prometheus dependencies

Geod24:
As mentioned recently, I am attempting to drop our Ocean dependency, and besides the formatter / logger, Prometheus is the only thing we use from it.


hewison-chris (approved):
LGTM

PR #3208, opened 2022-03-17
Manager: Query peer heights lazily while catching up

omerfirmak:
```
If we have unresponsive peers, the initial act of querying
each peer for their ledger heights could take a lot of time.
```

could help https://github.com/bosagora/agora/issues/3206


omerfirmak:
me ❤️ macOS

mkykadir (approved):
We lost sorting but I agree that its not big of a deal.

PR #3210, opened 2022-03-18
NetMan: Optimize signature catchup

Geod24:
Found a bug in signature catchup: Namely, we do not verify the header we receive is valid.
This can easily crash the node, which is how I noticed it. While looking deeper, I also noticed that this function is quite expensive, so here's an attempt at optimizing it. See each commit for a description. Still need some testing.


source/agora/consensus/Ledger.d
```diff
@@ -664,6 +696,76 @@ public class NodeLedger : Ledger
...
+
+            private this (const BitMask bm)
+            {
+                this.data = bm;
+                this.size = this.data.setCount();
+                this.popFront(); // Initialize `index`
+            }
+
+            ///
+            public bool   empty () const scope { return this.size == 0; }
```
hewison-chris, line 728: Extra whitespace

source/agora/consensus/Ledger.d
```diff
@@ -664,6 +696,76 @@ public class NodeLedger : Ledger
...
+
+            ///
+            public Height front () { return Height(0); }
+
+            ///
+            public void   popFront () scope @safe pure nothrow @nogc
```
hewison-chris, line 734: Whitespace

hewison-chris (commented):
First pass on this draft looks ok.

hewison-chris:
Closing as new PR coming soon to address this.

PR #3215, opened 2022-03-21
Improve address registry UX

### LINKED ISSUE #2320: Improve `addresses_to_register` configuration UX

mkykadir:
Fixes #2320 
Depends on #3239 

After this PR, #3240 can be considered. 

This PR misses Flash configuration intentionally due to current Flash architecture. It ends up with code duplication and unnecessarily doubles memory usage for the same data (i.e. `NetworkManager.addresses_to_register` will be same and required for `Flash`).

There are three options for configuring how a node registers itself to public registry; by default all interfaces are considered as public:

1. Default (no specific configuration from user)
    `node.register` or `node.public_addresses` are not configured by user thus all interfaces are registered. If an interface has
    `0.0.0.0` configured as its `address` than public IP address of the host will be used. In registry, this will result in A record
    for each public interface and associated URI record (including schema and port)
2. `node.register` is configured by user
    Each public interface will be registered with value in the `node.register` as `address` of the interface. This will result in single
    CNAME record and associated URI records for each public interface (including schema and port)
3. `node.public_addresses` is configured by user
    Interfaces are ignored and addresses in this field is registered as it was before in `addresses_to_register` method.

### DIFF
--- a/devel/config-single.yaml
+++ b/devel/config-single.yaml
@@ -34,8 +34,6 @@ validator:
   enabled: true
   # We use first of Genesis Block enrollments: val5: boa1xrval5rzmma29zh4aqgv3mvcarhwa0w8rgthy3l9vaj3fywf9894ycmjkm8
   seed: SAZO4WA5SXUN3J6XBXZCFPXJJZ62IQAJYAG2KBVWF2QZKTU2WK3QTNMV
-  addresses_to_register:
-    - http://127.0.0.1:2826/
 
 registry:
   public: true
--- a/devel/testnet/README.KR.md
+++ b/devel/testnet/README.KR.md
@@ -1,4 +1,4 @@
-# 노드 운영 방법 및 조건 
+# 노드 운영 방법 및 조건
 
 보스아고라 노드인 아고라(Agora)를 테스트 하는데 필요한 단계와 테스트 네트워크(‘TestNet’)에서 Foundation 노드와 상호 작용하는 방법을 안내드립니다.
 
@@ -38,7 +38,7 @@ Ubuntu Linux 20.04 및 MacOS-11(Big Sur)용 네이티브 바이너리(Native Bin
 
 터미널을 열고 `docker run bosagora/agora --testnet` 명령을 사용합니다.
 
-노드가 시작되고 여러 메시지를 화면에 나타나야 합니다. 약 몇초 동안 네트워크와 동기화를 시작하여 최신 상태의 체인을 다운로드해야 합니다. 
+노드가 시작되고 여러 메시지를 화면에 나타나야 합니다. 약 몇초 동안 네트워크와 동기화를 시작하여 최신 상태의 체인을 다운로드해야 합니다.
 
 ![초기 블록 다운로드시 로그 메시지](./IBD.png)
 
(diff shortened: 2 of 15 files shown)


Geod24:
I was thinking about it the other day, and there's a few options:
1) Having a `register: {true, false}` in the interface itself reduces duplication, but we could run into problems since the IP provided is likely to be a netmask (e.g. 0.0.0.0) than an actual IP address;
2) Even if we get an actual IP address OR we can detect it (we'd have to handle native as well as node in Docker), the user might want to use a hostname;
3) In addition to the above IP/hostname discrepancy, a user could also have a different port, thanks to reverse-proxying;
4) Finally, users might have addresses to register which are completely different from the `interfaces` setting: For example if they live behind a load balancer, or, if we go wild, if they have a message broker sitting between their public endpoint and their node;

For this reason, I was inclined to go with the following solution:
- By default, assume an interface is to be registered. Interfaces not to be registered should have `public: false`;
- If an interface is to be registered, check if `register` has an address in it, and use that;
- If `register` is not set, detect the IP, and attempt to self-connect: If it fails, hard error;
- Add (or keep) a way to register arbitrary addresses: This can be done by keeping `addresses_to_register` (although I'd suggest another name, such as `public_addresses`, or adding another `type` in `interfaces`, e.g. `public`, to represent addresses to register but not to listen to;

This is just a brain dump, I'll look in the PR in more details now :)

mkykadir:
Here is the summary of what I've done in this PR:

1. Added `register` boolean flag; interface is only registered when this is set and node is a Validator (will work on Flash nodes)
2. Added `hostname` field and `addresses_to_register` is moved to interface. One can only be set, hostname acts as a CNAME record and `addresses_to_register` is for extreme scenarios
3. If none is provided from 2, IP address is used in interface config; or public IP address is fetched when config is `0.0.0.0` or `::`

It looks mostly similar to @Geod24 thoughts with different namings 😄

mkykadir:
Per-interface configuration allows us to configure a CNAME for one interface and use IPs for another interface thus Payload includes CNAME and A records together; which Registry [doesn't like](https://github.com/bosagora/agora/blob/ea0f589ff3aaa1331aa15cb325978b3fbbff09fe/source/agora/node/Registry.d#L176-L178) and integration test catches that.

hewison-chris:
I think you need to rebase and then update the new `ngrok` instructions recently merged as it mentions `addresses_to_register:`

mkykadir:
> I think you need to rebase and then update the new `ngrok` instructions recently merged as it mentions `addresses_to_register:`

Updated

hewison-chris (approved):
LGTM

PR #3220, opened 2022-03-23
Remove Ocean dependency, use dtext instead

Geod24:
Should reduce our footprint a bit.


Geod24:
Before:
> -rwxr-xr-x  2 geod24  staff   79769112 Mar 23 15:30 agora

After:
> -rwxr-xr-x  2 geod24  staff   79285712 Mar 23 15:28 agora

Not as much as I'd hoped but still 500 kb of binary size gone, and hopefully a few module ctors.

mkykadir (approved):

Geod24:
Integration will have to be forced (and Faucet fixed)

Geod24:
All CIs (but integration) are green so going ahead with this.

PR #3222, opened 2022-03-23
Ledger: eliminate DB ops in getTxFeeRate()

omerfirmak:
Which used tp slow down combineCandidates()

### DIFF
--- a/source/agora/consensus/pool/Transaction.d
+++ b/source/agora/consensus/pool/Transaction.d
@@ -93,8 +93,18 @@ public class TransactionPool
     /// Keeps track of which TXs spend which inputs
     private Set!Hash[Hash] spenders;
 
+    /// Data associated with TXs in the pool for fast access
+    public struct KnownTx
+    {
+        /// Inputs
+        Set!Hash inputs;
+
+        /// Fee rate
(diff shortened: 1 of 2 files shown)


source/agora/consensus/pool/Transaction.d
```diff
@@ -125,7 +135,8 @@ public class TransactionPool
             this.updateSpenderList(tx);
             auto tx_hash = tx.hashFull();
             iota(tx.outputs.length).each!(idx => utxo_set[UTXO.getHash(tx_hash, idx)] = tx.outputs[idx]);
-            this.known_txs[tx_hash] = Set!Hash.from(tx.inputs.map!(input => input.utxo));
+            Amount fee_rate = Amount(this.db.execute("SELECT fee FROM tx_pool WHERE key = ?", tx_hash).oneValue!ulong);
```
mkykadir, line 138: Noting this down for future startup/shutdown improvements; these kind of IO bound loops in ctors seems problematic for node startup procedure. But no effects for the current condition

mkykadir (approved):

Geod24 (approved):

omerfirmak:
oom?

PR #3231, opened 2022-03-24
Nominator: Fix combineCandidates introducing new Values

omerfirmak:
We were serializing the holder instead of ConsensusData

Nominator now also scores the validated values to be used
as a nomination criteria for new values to be nominated.

Towards https://github.com/bosagora/agora/issues/3230


hewison-chris (approved):
LGTM

hewison-chris:
I rebased on upstream and pushed as actions had got stuck.

PR #3232, opened 2022-03-24
Stats: Merge `stats` interface with `http[s]`

### LINKED ISSUE #2888: Allow merging `stats` interface with an `http[s]` one
Currently, we can have both `stats` and `http` interface:
https://github.com/bosagora/agora/blob/1d67da7e16f0233d20bbb9f0175c08bdecd81dc6/doc/config.example.yaml#L67-L78

However, we might want to have them on the same port (e.g. if the interface is meant to be public).
In which case, the following should use the same `URLRouter`:
```YAML
interfaces:
  - type: http
    address: "0.0.0.0"
    port: 8080
  - type: stats
    address: "0.0.0.0"
    port: 8080
```

Questions:
- What happens when `https` is used instead of `http` ? I'm tempted to say same behavior.
- What happens when the `address` differs but the `port` is the same ? Having both `127.0.0.1` and `0.0.0.0` will conflict (because `0.0.0.0` includes `127.0.0.1`). Perhaps we can just error out.

Side note: Having a way to bind to an interface instead of an address would be *really* nicer.

mkykadir:
Stats interface is also a http[s] interface, before it was required to
configure as a separate interface. This change ease configuration of
interfaces and still allows configuring stats interface on
separate address/port.

Fixes #2888
This changeset also allows us to configure multiple stats interface

If one wants to access stats server from same HTTP interface, following configuration can be used;
```yaml
interfaces:
    - type: http
       address: 0.0.0.0
       port: 8080
       stats: true
```
stats server can also be accessed through separate interface with the following configuration;
```yaml
interfaces:
    - type: http
       address: 0.0.0.0
       port: 8080
    - type: http
       address: 0.0.0.0
       port: 3000
       stats: true
```


mkykadir:
Faucet will need an update, reason of Integration test failing

hewison-chris (approved):
LGTM

Geod24:
My problem with this change is that it forces an `http[s]` interface. We know that our http[s] interface are more fragile / risky because they go through Vibe.d. That's why it was a separate interface to begin with.

Geod24:
Reading again, `stats` is still here, so ignore the above comment.

PR #3233, opened 2022-03-25
EnrollmentManager: Remove some trivial wrapper

Geod24:
The diff should say it all.

### DIFF
--- a/source/agora/consensus/protocol/Nominator.d
+++ b/source/agora/consensus/protocol/Nominator.d
@@ -238,10 +238,9 @@ extern(D):
         Utils.getCollectorRegistry().addCollector(&this.collectStats);
 
         // Find the node id of this validator and create an SCPObject
-        Hash[] utxo_keys;
-        this.enroll_man.getEnrolledUTXOs(Height(1), utxo_keys);
         const this_utxo = this.enroll_man.getEnrollmentKey();
-        NodeID node_id = utxo_keys.countUntil(this_utxo);
+        NodeID node_id = ledger.getValidators(Height(1))
+            .map!(vi => vi.utxo).countUntil(this_utxo);
         this.updateSCPObject(node_id);
         this.restoreSCPState();
         this.slot_stat.clear();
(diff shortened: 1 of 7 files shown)


source/agora/consensus/protocol/Nominator.d
```diff
@@ -952,7 +952,7 @@ extern(D):
     protected Signature signBlock (in BlockHeader header) @safe nothrow
     {
         return header.sign(this.kp.secret,
-            this.enroll_man.getOurPreimage(header.height));
+            this.enroll_man.getPreimage(header.height).hash);
```
omerfirmak, line 954: Is this really equivalent? We are hashing a `PreImageInfo`, right?

source/agora/consensus/protocol/Nominator.d
```diff
@@ -238,10 +238,9 @@ extern(D):
         Utils.getCollectorRegistry().addCollector(&this.collectStats);
 
         // Find the node id of this validator and create an SCPObject
-        Hash[] utxo_keys;
-        this.enroll_man.getEnrolledUTXOs(Height(1), utxo_keys);
         const this_utxo = this.enroll_man.getEnrollmentKey();
-        NodeID node_id = utxo_keys.countUntil(this_utxo);
+        NodeID node_id = ledger.getValidators(Height(1))
```
hewison-chris, line 242: Not sure why this is not ledger height rather than Height(1).

source/agora/consensus/protocol/Nominator.d
Geod24: `getPreimage`: `return PreImageInfo(this.enroll_key, this.cycle[height], height);`
`getOurPreimage`: `return this.cycle[height];`

Since the second argument is `hash`, I believe it is, yes.

omerfirmak: wait, I confused it with `hashFull` 🤦

hewison-chris (commented):
If `Nominator` is constructed with `ledger` at height greater than 0 then it looks like the `node_id` will be wrong. Although this is the same behaviour as before I think it is worth checking.

source/agora/consensus/protocol/Nominator.d
Geod24: @hewison-chris : Yes, that bothered me as well, but I think it would be better to handle that in a separate PR.

hewison-chris (approved):
LGTM, I raised an issue for the Nominator ctor code.

PR #3238, opened 2022-03-28
TxBuilder: Remove hard dependency to WellKnown keys

Geod24:
The last step missing to allow the TxBuilder to be truly independent
from the Well-Known keys, yet offer a similar API.


mkykadir:
Moved to #3256

PR #3242, opened 2022-03-30
Nominator: Purge externalized SCP slots

omerfirmak:
SCP holds state for each height and needs them to be
explicitly cleared. This was probably the reason of our
ever growing memory consumption.


linked0 (approved):
LGTM

hewison-chris (commented):
Good find.

omerfirmak:
Memory usage is still increasing tho, it will need more investigation.

PR #3248, opened 2022-03-31
Stats: Append scrape duration to the end of `/metrics` response

omerfirmak:


Geod24 (approved):

Geod24:
Although https://dlang.org/phobos/std_datetime_stopwatch.html#StopWatch might be more suited

omerfirmak:
> Although https://dlang.org/phobos/std_datetime_stopwatch.html#StopWatch might be more suited

this was what docs suggested

```
// If stopping or resetting the StopWatch is not required, then
// MonoTime can easily be used by itself without StopWatch.
auto before = MonoTime.currTime;
// do stuff...
auto timeElapsed = MonoTime.currTime - before;
```

Geod24:
Looking at this again, I have to question the motive. Do you expect scrape to take long ?
I think that the delays you might be seeing is in Vibe.d handling the request, which means by the time we get to `MonoTime.start`, most of the delay has already passed.

omerfirmak:
@Geod24, yeap I suspected the DB operations in `/metrics` to block the process for a long time occasionally. But havent seen anything proving that.

PR #3250, opened 2022-03-31
Move Tracy into its own submodule

### LINKED ISSUE #1756: Move Tracy and the tracking GC to its own submodule
It could be useful for other purposes / people, and can be cleanly separated.
This will also allow us to expose different kind of configuration, most importantly `TRACY_ON_DEMAND`.

mkykadir:
Fixes #1756 

Tracy is moved into its own [submodule](https://github.com/bosagora/tracyd) with GC changes

- New repo is created `tracyd`
- Tracy is only dependent when required (compiled with `traced-server` config)
- Previous GC config is still valid


source/agora/node/main.d
```diff
@@ -30,7 +30,11 @@ import agora.node.FullNode;
 import agora.node.Validator;
 import agora.node.Runner;
 import agora.utils.Workarounds;
-import agora.utils.TracyAPI;
+
+version (AgoraTracy)
+{
+  import tracyd;
+}
```
Geod24: The import is not used it seems ?

dub.json
```diff
@@ -44,25 +44,19 @@
...
-                "$DUB scripts/build_tracy.d",
-            ],
-            "sourceFiles-posix": [
-                "build/TracyClient.o"
-            ],
-            "sourceFiles-windows": [
-                "build/TracyClient.obj"
-            ],
-            "excludedSourceFiles": [ "source/agora/cli/*" ]
+            "versions": [ "AgoraTracy" ],
```
Geod24: Dub gives you a `Have_tracyd` version automatically.

mkykadir: Cool, is this documented BTW?

Geod24:
Few things w.r.t. the submodule:
- Why does it include a copy of (C++) Tracy ? I'd say a submodule would be better. Best would be if there was a way to install `tracy` as a lib with a `pkg-config` file.
- Needs LICENSE (MIT);
- I would not expose `package.d`. One of the reason is that it's just an alias to API, the second is that top level modules tend to get in the way. E.g.:
```D
module writeln;

import std.stdio;

void main ()
{
 	writeln("Does not compile");   
}
```

That's why I recommend [vanity packages](https://forum.dlang.org/thread/jdkmtgftmwtwaxxqhlju@forum.dlang.org).

mkykadir:
> * Why does it include a copy of (C++) Tracy ? I'd say a submodule would be better. Best would be if there was a way to install `tracy` as a lib with a `pkg-config` file.

I pass over that intentionally for not creating a `git` dependency to build library but currently it depends on a lot already.

> * Needs LICENSE (MIT);

Added

hewison-chris (approved):
LGTM

PR #3252, opened 2022-04-01
Trivial: Be more verbose when hitting an 'assert' in VibeManager

Geod24:


mkykadir (approved):

source/agora/network/VibeManager.d
```diff
@@ -98,7 +98,7 @@ public final class VibeNetworkManager : NetworkManager
             return new RestInterfaceClient!(agora.api.Validator.API)(
                 this.makeRestInterfaceSettings(url));
 
-        assert(0, "Unknown agora schema");
+        assert(0, "Unknown agora schema: " ~ url.toString());
```
omerfirmak, line 101: Asserting on this is not really a good idea, address is supplied externally and we do no checks on the schema before this point.

PR #3262, opened 2022-04-06
Fix typo on TestNet README

linked0:
I didn't apply the spell-checked version of README in the last commit. So commit again.

### DIFF
--- a/devel/testnet/README.md
+++ b/devel/testnet/README.md
@@ -143,7 +143,7 @@ points to your own IP address.
 As you have read in the previous section, you can run a validator, but if the validator runs on
 a local computer without a publicly exposed ip, it cannot be reached from other validators.
 One option is to use a service like [ngrok](https://ngrok.com) to expose a local port to the public. If you only run
-a fullnode, not a validator, please skip this section.
+a full node, not a validator, please skip this section.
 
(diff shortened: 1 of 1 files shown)


devel/testnet/README.md
```diff
@@ -143,7 +143,7 @@ points to your own IP address.
 As you have read in the previous section, you can run a validator, but if the validator runs on
 a local computer without a publicly exposed ip, it cannot be reached from other validators.
 One option is to use a service like [ngrok](https://ngrok.com) to expose a local port to the public. If you only run
-a fullnode, not a validator, please skip this section.
+a full node, not a validator, please skip this section.
```
hewison-chris, line 146: I think `fullnode` was correct here.

hewison-chris (changes requested):

devel/testnet/README.md
linked0: Out of curiosity, can you explain the difference between them?

hewison-chris: Looking at the rest of the `Readme` it does use `full node` and not `fullnode` (the name we use for the class in the code) so let's keep it consistent.

hewison-chris (approved):
LGTM

PR #3264, opened 2022-04-06
Registry: Prevent registering loopback addresses

### LINKED ISSUE #3261: Prevent nodes to register loopback addresses
Currently one can configure its node to register `127.0.0.1`  address to public registry. This is not an addressed issue in DNS standards but this creates a problem that this configuration will cause a some another node to end-up connecting itself or any other application on its local machine. 

But we also need to keep this ability to continue supporting local testing environments. `testing` flag in configuration can be used as a toggle for that.

mkykadir:
to reduce the attack surface on nodes' host machines.
Loopback address registry is still allowed when `node.testing` is set
(i.e. Registry and network is in testing mode) to still support local
testing networks.

Fixes #3261

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -103,15 +103,20 @@ public class NameRegistry: NameRegistryAPI
         QTYPE.URI,
     ];
 
+    /// Registry is in testing mode and allows loopback addresses
+    private bool testing;
+
     ///
     public this (Domain realm, RegistryConfig config, NodeLedger ledger,
-        ManagedDatabase cache_db, ITaskManager taskman, NetworkManager network)
+        ManagedDatabase cache_db, ITaskManager taskman, NetworkManager network,
+        bool testing = false)
     {
         assert(ledger !is null);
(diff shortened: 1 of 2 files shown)


source/agora/node/Registry.d
```diff
@@ -173,6 +178,18 @@ public class NameRegistry: NameRegistryAPI
...
+
+            // Prevent loopback addresses to be registered
+            if (!this.testing)
+            {
+                // It is safe to compare case-insensitive since `Address` is normalized
+                ensure(
+                    (this_type == TYPE.CNAME && addr.host != "localhost")
+                    || (this_type == TYPE.A && !addr.host.startsWith("127."))
+                    || (this_type == TYPE.AAAA && (addr.host != "::1" && addr.host != "0:0:0:0:0:0:0:1"))
+                , "Cannot register a loopback addresses");
```
linked0:
```suggestion
                , "Cannot register a loopback address");
```

linked0 (approved):
LGTM

PR #3266, opened 2022-04-06
Fix #3265 : Add ctor for AAAA type to RDATA

### LINKED ISSUE #3265: Registry: Abbreviated IPv6 addresses are registered incorrectly
`::` is abbreviated IPv6 address which is equivalent `0.0.0.0` in IPv4. Abbreviated addresses are stored incorrectly;

`::` > `1000::788f:2207:367f:0` with wrong RR type

mkykadir:
AAAA field is separated from plain binary

### DIFF
--- a/source/agora/common/DNS.d
+++ b/source/agora/common/DNS.d
@@ -721,6 +721,12 @@ public struct ResourceRecord
         }
 
         /// Ditto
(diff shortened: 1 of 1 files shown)


hewison-chris (approved):
LGTM

linked0 (approved):

PR #3267, opened 2022-04-06
Add `additive` field to the TestNet configuration

linked0:
There was no log message while running a validator on TestNet. It was because the `additive` field was missing in the `logging` config of a configuration file.


hewison-chris (approved):
LGTM

PR #3271, opened 2022-04-07
Add missing `public_addresses` field to README

linked0:
This adds the missing `public_addresses` field to README which is needed to run a validator.

### DIFF
--- a/devel/testnet/README.KR.md
+++ b/devel/testnet/README.KR.md
@@ -83,6 +83,10 @@ docker run -v $(pwd):/agora/ bosagora/agora -c my_config.yaml
 
 위의 작업이 완료되면, 구성 파일을 다음과 같이 변경하면 됩니다.
 ```yaml
+node:
+  public_addresses:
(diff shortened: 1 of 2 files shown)


devel/testnet/README.KR.md
```diff
@@ -83,6 +83,10 @@ docker run -v $(pwd):/agora/ bosagora/agora -c my_config.yaml
 
 위의 작업이 완료되면, 구성 파일을 다음과 같이 변경하면 됩니다.
 ```yaml
+node:
+  public_addresses:
+    - "agora://my-address.example.com:2826"
```
mkykadir, line 88: Since `agora://` already have default port 2826 we don't need to specify it but it wouldn't change anything.

mkykadir (approved):

devel/testnet/README.KR.md
linked0: Right but it's just an example, so I could use other ports like 3826, or 4826.

PR #3276, opened 2022-04-10
RPC: Rework to implement a packet based protocol

### LINKED ISSUE #3253: RPC: Revamp to implement a LocalRest-like request handling
Currently, we support only a single request in-flight per connection. LocalRest does this differently and can handle multiple requests thru a single connection. LocalRest's approach is more robust in certain edge cases where a request to node A from node B triggers a new request to node B.

omerfirmak:
This will allow us to have multiple requests in-flight on
a single connection and allow infinite nested RPC requests

### DIFF
--- a/source/agora/network/Client.d
+++ b/source/agora/network/Client.d
@@ -712,36 +712,11 @@ public class NetworkClient
     {
         import std.range;
 
-        if (this.tryMergeRPC(address, api))
-            return true;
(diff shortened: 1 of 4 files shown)


omerfirmak:
Fixes #3253

hewison-chris (approved):
LGTM

Issue #3280, opened 2022-04-12
Implement the endpoints for Web Admin Interface


hewison-chris:
We provides the `Admin` interface now in [Admin.d](https://github.com/bosagora/agora/blob/v0.x.x/source/agora/api/Admin.d). But the requirement for the endpoints has changed so we support the change in Agora.

linked0:
[This PR](https://github.com/bosagora/agora/pull/2478) is for the Web Admin interface that uses the Agora Admin endpoints.

linked0:
Completed with PR [#3294](https://github.com/bosagora/agora/pull/3294)

PR #3294, opened 2022-04-14
Apply the changed requirement for the Admin endpoints

linked0:
Added the new `login` endpoint and changed the name and return value for the existing `validator` and `encriptionkey` endpoints.

Relates #3280

### DIFF
--- a/source/agora/api/Admin.d
+++ b/source/agora/api/Admin.d
@@ -83,8 +83,8 @@ public interface NodeControlAPI
     @method(HTTPMethod.GET)
     @resultSerializer!(RawStringSerializer.serialize,
                        RawStringSerializer.deserialize,
-                       "image/svg+xml")()
-    public string loginQR (@viaHeader("Content-Type") out string contentType,
+                       "application/json")()
+    public string validator (@viaHeader("Content-Type") out string contentType,
                            @viaHeader("Vary") out string vary);
 
     /***************************************************************************
@@ -99,8 +99,8 @@ public interface NodeControlAPI
     @method(HTTPMethod.GET)
     @resultSerializer!(RawStringSerializer.serialize,
                        RawStringSerializer.deserialize,
-                       "image/svg+xml")()
-    public string encryptionKeyQR (string app, ulong height,
+                       "application/json")()
+    public string encryptionkey (string app, ulong height,
                                    @viaHeader("Content-Type") out string contentType,
                                    @viaHeader("Vary") out string vary);
 }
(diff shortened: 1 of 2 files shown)


source/agora/api/Admin.d
```diff
@@ -44,6 +44,24 @@ public interface NodeControlAPI
         @viaQuery("console") Nullable!bool console = Nullable!(bool).init,
         @viaQuery("file") Nullable!string file = Nullable!(string).init);
 
+    /***************************************************************************
+
+        Log in to the admin server
+
+        This is dummy endpoint for checking the username and password in
```
hewison-chris:
```suggestion
        This is endpoint used only for checking the username and password in
```

source/agora/node/admin/AdminInterface.d
```diff
@@ -119,6 +120,11 @@ public class AdminInterface : NodeControlAPI
         configureLogger(config, !console.isNull() || !file.isNull());
     }
 
+    // GET: /login
+    public override void login ()
+    {
+    }
```
hewison-chris, line 126: Here we should use the AUTH header to find out if user id and password are correct.

hewison-chris (changes requested):
If I understand correctly, the login page is only used to give the user feedback if the credentials are incorrect, otherwise it will store those credentials in the session to use in the AUTH header to pass for other endpoints.

source/agora/node/admin/AdminInterface.d
linked0: We use AUTH header for all the requests. So I introduced the dummy endpoint. It will return an HTTP Status Code(200 or 401). But, I will find a more reasonable way to handle as you mean and the requirement is described.
```
bool checkPassword (string user, string password) @safe
{
    return user == config.admin.username && password == config.admin.pwd;
}
auto adminrouter = new URLRouter();
adminrouter.any("*", performBasicAuth("Agora Admin", &checkPassword));
```

source/agora/api/Admin.d
linked0: Updated

source/agora/node/admin/AdminInterface.d
hewison-chris: After thinking more about this we should probably return a JWT which can be used on the subsequent requests to the other endpoints. But for now I think what you have is ok to proceed.

hewison-chris (approved):
LGTM

source/agora/api/Admin.d
mkykadir: Updated in next commit btw

linked0: Ah, I will update, Thanks.

PR #3302, opened 2022-04-15
Config: Make LoggerConfig `additive` flag default true

### LINKED ISSUE #3295: Root log is not propagated to console
Even following config, logs are not displayed on console; since #3243 

````
logging:
  root:
    level: Info
    console: true
    file: log/root.log
````

mkykadir:
This is more intuitive and easies the configuration process.

Fixes #3295 

> Perhaps the default should be `additive` so that it behaves as expected.

_Originally posted by @hewison-chris in https://github.com/bosagora/agora/issues/3295#issuecomment-1099822597_

### DIFF
--- a/source/agora/utils/Log.d
+++ b/source/agora/utils/Log.d
@@ -224,7 +224,7 @@ public struct LoggerConfig
     public @Optional string file;
 
     /// Whether this logger should be additive or not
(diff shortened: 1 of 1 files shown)


linked0 (approved):
LGTM

PR #3303, opened 2022-04-18
Mutex for the handlers for Stoa

### LINKED ISSUE #2942: Race condition on shutdown can cause SEGV

linked0:
After shutting down nodes, there are cases where nodes crash
because the handlers for Stoa are nullified. So this is a way
to keep from crashing by using a `synchronized` mutex when
accessing the handlers.

Fixes #2942


hewison-chris (approved):
LGTM

PR #3304, opened 2022-04-18
Manager: Check registry cache immediately during discovery

mkykadir:
to reduce Registry load. Even this might directly block the discovery
task (Registry might not have records cached yet), this condition is
expected to mostly happen during node initialization.


hewison-chris (approved):
LGTM

linked0 (approved):

PR #3305, opened 2022-04-18
Trivial: Remove `TransactionRelayerFeeImpl` unnecessary ctor field

mkykadir:
`start_timers` field was only used for the module unittest;
this test doesn't invoke the `start` anyway.


hewison-chris (approved):
LGTM

PR #3306, opened 2022-04-18
Trivial: Use `createTimer` for Client gossip timer

mkykadir:
It is immediately stopped after setting, `createTimer` is more suitable
for those conditions.


omerfirmak (approved):

hewison-chris (approved):
LGTM

linked0 (approved):

Issue #3158, opened 2022-03-03
Control + C does not lead to immediate shutdown


Geod24:
```
2022-03-03 07:22:10,568 Info [agora.consensus.state.Ledger] - Completed externalization of block #561
Received signal 2, shutting down listeners...
[main(/uLN) INF] Stopped to listen for HTTP requests on 0.0.0.0:8080
[main(/uLN) INF] Stopped to listen for HTTP requests on :::44235
[main(/uLN) INF] Stopped to listen for HTTP requests on 0.0.0.0:45213
Calling node shutdown procedure...
2022-03-03 07:22:11,241 Info [agora.node.FullNode] - Shutting down..
2022-03-03 07:22:11,241 Info [agora.common.Task] - Tasks started: 400
Terminating event loop...
Received signal 2, shutting down listeners...
Calling node shutdown procedure...
Terminating event loop...
2022-03-03 07:22:11,717 Info [agora.consensus.state.Ledger] - Beginning externalization of block #562 (previous block signatures: 11111111111111111111111111111111)
2022-03-03 07:22:11,717 Info [agora.consensus.state.Ledger] - Transactions: 1 - Enrollments: 5
2022-03-03 07:22:11,719 Info [agora.consensus.state.Ledger] - Validators: Active: 27 - Signing: 11111111111111111111111111111111 - Slashed: 0
2022-03-03 07:22:11,814 Info [agora.consensus.state.Ledger] - Completed externalization of block #562
2022-03-03 07:22:12,940 Info [agora.consensus.state.Ledger] - Beginning externalization of block #563 (previous block signatures: 11111111111111111111111111111111)
2022-03-03 07:22:12,940 Info [agora.consensus.state.Ledger] - Transactions: 1 - Enrollments: 0
2022-03-03 07:22:12,942 Info [agora.consensus.state.Ledger] - Validators: Active: 32 - Signing: 11111111111111111111111111111111 - Slashed: 0
2022-03-03 07:22:12,978 Info [agora.consensus.state.Ledger] - Completed externalization of block #563
2022-03-03 07:22:14,282 Info [agora.consensus.state.Ledger] - Beginning externalization of block #564 (previous block signatures: 11111111111111111111111111111111)
2022-03-03 07:22:14,282 Info [agora.consensus.state.Ledger] - Transactions: 1 - Enrollments: 21
[...]

PR #3307, opened 2022-04-18
Stop TCP interfaces and RPC connections on shutdown

mkykadir:
Towards a better shutdown procedure

This will reduce the number of eventcore leaks during shutdown, only DNS UDP connection will be remaining as leaking.

Part of #3158 

depends on #3311 and https://github.com/bosagora/localrest/pull/6

### DIFF
--- a/source/agora/network/VibeManager.d
+++ b/source/agora/network/VibeManager.d
@@ -59,6 +59,21 @@ shared static this ()
 /// And implementation of `agora.network.Manager : NetworkManager` using Vibe.d
 public final class VibeNetworkManager : NetworkManager
 {
+    /// A stub class for overriding `shutdown` for the REST client
+    private class ValidatorRestClient : RestInterfaceClient!(agora.api.Validator.API)
+    {
+        this (RestInterfaceSettings settings)
+        {
+            super(settings);
+        }
+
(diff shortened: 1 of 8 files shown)


source/agora/network/VibeManager.d
```diff
@@ -59,13 +59,23 @@ shared static this ()
 /// And implementation of `agora.network.Manager : NetworkManager` using Vibe.d
 public final class VibeNetworkManager : NetworkManager
 {
+    private RPCClient!(agora.api.Validator.API)[] rpc_clients;
```
omerfirmak: Cant we just `close()` the socket in `RPCClient` dtor?

mkykadir: I think all interface and connections should be closed before invoking node's shutdown procedure. But this might cause holding an unnecessary reference for cases where a connection [is banned ](https://github.com/bosagora/agora/blob/03f36b7f6103396707cde00906226a68ab43d9c5/source/agora/network/Client.d#L657). I will think a better approach.

mkykadir: I've updated it with new approach, ready to review

hewison-chris (approved):
LGTM

PR #3311, opened 2022-04-19
Add RPC support to Agora client

### LINKED ISSUE #3310: Agora client doesn't support RPC interface
Only REST interface is available

mkykadir:
also includes some fixes

see each commit

Fixes #3310


hewison-chris (approved):
LGTM

PR #3320, opened 2022-05-11
Implement DNS seed

### LINKED ISSUE #2874: Registry: Implement server-side seeding
When a node first start, from a clean configuration, it needs to connect to a set of peer and download the chain.
We implemented this client side a while ago (https://github.com/bosagora/agora/pull/19) but it was never implemented server side.

We should be able to do an `A` / `AAAA` query on a zone (probably the `realm` itself, but open to discussion, as the same idea might apply to `flash`) and get a "well distributed" set of nodes. Well distributed meaning that at least X validators are included (X to be decided based on the number of active validators), and the rest being staking nodes (ideally, but this requires #2867). Obviously there must be a maximum numbers of nodes that will be included in the response, and the nodes should be round-robin-ed on a timely basis (not per request, I think, to avoid people using it as a way to discover the network).

Note that the main part of this issue can be done without #2867 : it's fine if the initial implementation just return validators.

mkykadir:
Fixes #2874 

Adds DNS seed support like Bitcoin seeding. Parameters can be tweaked.

- Adds new config field to Registry
DNS seed update interval
- A or AAAA query will return randomly selected validator nodes
- Maximum of 10 nodes will be returned to seed queries

### DIFF
--- a/source/agora/node/Registry.d
+++ b/source/agora/node/Registry.d
@@ -813,6 +813,9 @@ private struct ZoneData
     /// Query for getting records with matching QTYPE
     private string query_records_get_typed;
 
+    /// Query for getting all records matching QTYPE
+    private string query_records_get_all_typed;
+
     /// Query for getting registry utxo
     private string query_utxo_get;
 
@@ -913,6 +916,9 @@ private struct ZoneData
 
         this.query_records_get_typed = format("SELECT address, type, ttl " ~
             "FROM registry_%s_addresses WHERE pubkey = ? AND type = ?", zone_name);
+
+        this.query_records_get_all_typed = format("SELECT pubkey, address, type, ttl " ~
+            "FROM registry_%s_addresses WHERE type = ?", zone_name);
 
         this.query_utxo_get = format("SELECT sequence, utxo " ~
             "FROM registry_%s_utxo WHERE pubkey = ?", zone_name);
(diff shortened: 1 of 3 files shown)


source/agora/node/Registry.d
```diff
@@ -914,6 +917,9 @@ private struct ZoneData
         this.query_records_get_typed = format("SELECT address, type, ttl " ~
             "FROM registry_%s_addresses WHERE pubkey = ? AND type = ?", zone_name);
```
hewison-chris, line 972: Although not changed in this PR. It looks like zone_name will be used as pubkey.

source/agora/node/Registry.d
```diff
@@ -914,6 +917,9 @@ private struct ZoneData
...
 
+        this.query_records_get_all_typed = format("SELECT pubkey, address, type, ttl " ~
+            "FROM registry_%s_addresses WHERE type = ?", zone_name);
+
         this.query_utxo_get = format("SELECT sequence, utxo " ~
             "FROM registry_%s_utxo WHERE pubkey = ?", zone_name);
```
hewison-chris, line 978: Also here the number of params is correct but wrong variable I think.

hewison-chris (changes requested):
Can you check the questions I have first. Then I'll review again.

source/agora/node/Registry.d
mkykadir: No, it is for `registry_%s_addresses`

source/agora/node/Registry.d
mkykadir: As first comment

source/agora/node/Registry.d
hewison-chris: 🤦

source/agora/node/Registry.d
hewison-chris: 🤦‍♂️ 🤦‍♂️

hewison-chris (approved):
LGTM

PR #3322, opened 2022-05-16
RPC: Write lock Shared TCPConnection

mkykadir:
Write is invoked before previous write operation without a concurrency
lock; which caused nodes to core-dump with
`Overwriting notification callback.`

### DIFF
--- a/source/agora/network/RPC.d
+++ b/source/agora/network/RPC.d
@@ -483,13 +483,14 @@ private class SharedTCPConnection
 {
     private TCPConnection stream;
     alias stream this;
-    private TaskMutex wmutex;
+    public TaskMutex wmutex;
 
     this (TCPConnection stream) @safe nothrow
     {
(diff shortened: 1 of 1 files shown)


source/agora/network/RPC.d
```diff
@@ -648,7 +649,9 @@ private void handleThrow (API) (scope API api, SharedTCPConnection stream, Durat
                     {
                         mixin("auto foo = ", CallMixin);
                         packet.is_response = true;
+                        stream.wmutex.lock();
                         stream.write(serializeFull(packet) ~ serializeFull(foo));
+                        scope (exit) stream.wmutex.unlock();
```
omerfirmak, line 653: That should be before the `write`

omerfirmak (approved):

hewison-chris (commented):
LGTM

PR #3323, opened 2022-06-13
DNS: Remove seed configuration

### LINKED ISSUE #2256: Merge 'dns' and registries

mkykadir:
Fixes #2256 

Registry answers for seed queries since #3320 and `NetworkManager` adds addresses from node registry for initial peer finding already. This configuration field is not needed anymore.


hewison-chris (approved):
LGTM

